PR #4080, opened 2025-06-24
feat: upgrade arrow object_store to 0.12.2

jackye1995:
0.12.2 adds additional logging that would be helpful for debugging actual error thrown from object storage. It also brings in a few breaking changes such as changing `Range<usize>` to `Range<u64>`.


Cargo.toml
```diff
@@ -132,8 +132,7 @@ mockall = { version = "0.13.1" }
 mock_instant = { version = "0.3.1", features = ["sync"] }
 moka = { version = "0.12", features = ["future", "sync"] }
 num-traits = "0.2"
-# Set min to prevent use of versions with CVE-2024-41178
```
jackye1995, line 135: this CVE is fixed in 0.11.0

rust/lance-io/src/object_reader.rs
```diff
@@ -136,7 +136,7 @@ impl Reader for CloudObjectReader {
                 let meta = self
                     .do_with_retry(|| self.object_store.head(&self.path))
                     .await?;
-                Ok(meta.size)
+                Ok(meta.size as usize)
```
jackye1995, line 139: The Lance Reader still requires usize. 

```
    /// Object/File Size.
    async fn size(&self) -> object_store::Result<usize>;

    /// Read a range of bytes from the object.
    ///
    /// TODO: change to read_at()?
    async fn get_range(&self, range: Range<usize>) -> object_store::Result<Bytes>;
```

It might be worth discussing if we should also change the trait to use u64. Let me know if we want to do that I can change in this PR as well.

wjones127: Yeah we need to do a refactor at some point to fix all the integers. There's a lot of place where we use `usize` where we shouldn't.

wjones127 (approved):

rust/lance-io/src/object_reader.rs
jackye1995: Ended up touching too many files, created an issue #4082 to do it separately, so we can merge this asap.

Issue #4081, opened 2025-06-24
Make `max_fragment_id` optional


wjones127:
Earlier we found there is a bug with https://github.com/lancedb/lance/issues/2718

We need to make sure that `max_fragment_id` is preserved when explicitly set. Right now, it's ambiguous whether it's set to zero or unset.

In particular, we should test:

1. Fragment id zero isn't re-used:
    1. Create a dataset with 1 fragment
    2. Delete all rows
    3. Append another fragment
    4. Assert new fragment has id 1 not 0
2. Fragment id is never reset, even if all rows are deleted
    1. Create dataset with N fragments
    2. Delete all rows
    3. Append more fragments
    4. Assert new fragments have ids >= N

PR #4083, opened 2025-06-24
fix: make max_fragment_id optional to prevent fragment ID reuse

wjones127:
## Summary

Fixes issue #4081 by making `max_fragment_id` optional to distinguish between "unset" and "set to 0".

Previously, when all rows were deleted from a dataset, the `max_fragment_id` could be incorrectly reset, causing fragment IDs to be reused. This led to index corruption where indices pointed to the wrong rows after data was re-inserted.

## Changes Made

- **Protobuf Schema**: Changed `uint32 max_fragment_id = 11;` to `optional uint32 max_fragment_id = 11;`
- **Rust Implementation**: Updated to use `Option<u32>` with proper high water mark semantics
- **High Water Mark Logic**: Ensures `max_fragment_id` only increases, never decreases, even when all fragments are deleted
- **Comprehensive Tests**: Added tests for both scenarios mentioned in the issue

## Test Plan

Added two comprehensive tests that verify the exact scenarios from issue #4081:

1. **`test_fragment_id_zero_not_reused`**: Verifies fragment ID 0 isn't reused after deletion
   - Create dataset with 1 fragment (ID 0)
   - Delete all rows  
   - Append new fragment → gets ID 1 (not 0)

2. **`test_fragment_id_never_reset`**: Verifies fragment IDs never reset after deletion
   - Create dataset with 3 fragments (IDs 0, 1, 2)
   - Delete all rows
   - Append new fragments → get IDs 3, 4 (continuing sequence)

All existing tests pass, ensuring no regressions.

### Forward Compatibility ✅
- **Old versions reading new data**: When `max_fragment_id` is unset, old versions see default value (0) - safe
- **No crashes**: Unlike sentinel values, 0 won't cause memory exhaustion in range operations

### Backward Compatibility ✅  
- **New versions reading old data**: All existing manifests have the field set, continue to work
- **Gradual migration**: Old and new versions can interoperate safely

The proto3 `optional` approach provides clean semantics without breaking changes.

### DIFF
--- a/rust/lance-table/src/format/manifest.rs
+++ b/rust/lance-table/src/format/manifest.rs
@@ -70,8 +70,9 @@ pub struct Manifest {
     /// The writer flags
     pub writer_feature_flags: u64,
 
-    /// The max fragment id used so far
-    pub max_fragment_id: u32,
+    /// The max fragment id used so far  
+    /// None means never set, Some(0) means max ID used so far is 0
+    pub max_fragment_id: Option<u32>,
 
     /// The path to the transaction file, relative to the root of the dataset
     pub transaction_file: Option<String>,
@@ -135,7 +136,7 @@ impl Manifest {
             tag: None,
             reader_feature_flags: 0,
             writer_feature_flags: 0,
-            max_fragment_id: 0,
+            max_fragment_id: None,
             transaction_file: None,
             fragment_offsets,
             next_row_id: 0,
@@ -230,8 +231,18 @@ impl Manifest {
             .try_into()
             .unwrap();
 
-        if max_fragment_id > self.max_fragment_id {
-            self.max_fragment_id = max_fragment_id;
+        match self.max_fragment_id {
+            None => {
+                // First time being set
+                self.max_fragment_id = Some(max_fragment_id);
+            }
+            Some(current_max) => {
+                // Only update if the computed max is greater than current
+                // This preserves the high water mark even when fragments are deleted
+                if max_fragment_id > current_max {
+                    self.max_fragment_id = Some(max_fragment_id);
+                }
+            }
         }
     }
 
@@ -240,12 +251,11 @@ impl Manifest {
     ///
     /// This will return None if there are no fragments.
     pub fn max_fragment_id(&self) -> Option<u64> {
-        if self.max_fragment_id == 0 {
-            // It might not have been updated, so the best we can do is recompute
-            // it from the fragment list.
-            self.fragments.iter().map(|f| f.id).max()
+        if let Some(max_id) = self.max_fragment_id {
+            Some(max_id.into())
         } else {
-            Some(self.max_fragment_id.into())
+            // Not yet set, compute from fragment list
+            self.fragments.iter().map(|f| f.id).max()
         }
     }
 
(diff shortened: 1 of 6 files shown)


rust/lance/src/dataset.rs
```diff
@@ -6553,4 +6553,155 @@ mod tests {
...
+        .unwrap();
+        let data3 = RecordBatch::try_new(
+            schema.clone(),
+            vec![Arc::new(UInt32Array::from_iter_values(20..30))],
+        )
+        .unwrap();
+        let batches = RecordBatchIterator::new(
+            vec![data1, data2, data3].into_iter().map(Ok),
+            schema.clone(),
+        );
```
wjones127: You don't need to create 3 separate batches. The `max_rows_per_file` will handle splitting into fragments.
```suggestion
        // Create dataset with 3 fragments (N=3)
        let data = RecordBatch::try_new(
            schema.clone(),
            vec![Arc::new(UInt32Array::from_iter_values(0..30))],
        )
        .unwrap();
        .unwrap();
        let batches = RecordBatchIterator::new(
            vec![Ok(data)],
            schema.clone(),
        );
```

rust/lance/src/dataset.rs
```diff
@@ -6553,4 +6553,155 @@ mod tests {
...
+        let data4 = RecordBatch::try_new(
+            schema.clone(),
+            vec![Arc::new(UInt32Array::from_iter_values(100..110))],
+        )
+        .unwrap();
+        let data5 = RecordBatch::try_new(
+            schema.clone(),
+            vec![Arc::new(UInt32Array::from_iter_values(110..120))],
+        )
+        .unwrap();
```
wjones127: Similar here, doesn't need to be separate batches to get multiple fragments.

rust/lance-table/src/format/manifest.rs
```diff
@@ -221,31 +222,46 @@ impl Manifest {
...
-        if self.max_fragment_id == 0 {
-            // It might not have been updated, so the best we can do is recompute
-            // it from the fragment list.
-            self.fragments.iter().map(|f| f.id).max()
+        if let Some(max_id) = self.max_fragment_id {
+            // Return the stored high water mark
+            Some(max_id.into())
         } else {
-            Some(self.max_fragment_id.into())
+            // Not yet set, compute from fragment list
```
westonpace, line 263: In what scenario do we end up with fragments but no max_fragment_id?

rust/lance/src/index/prefilter.rs
```diff
@@ -62,7 +62,7 @@ impl DatasetPreFilter {
     ) -> Self {
         let mut fragments = RoaringBitmap::new();
         if indices.iter().any(|idx| idx.fragment_bitmap.is_none()) {
-            fragments.insert_range(0..dataset.manifest.max_fragment_id);
+            fragments.insert_range(0..dataset.manifest.max_fragment_id.unwrap_or(0));
```
westonpace, line 65: It looks like we unwrap to 0 in all cases where it has never been set (e.g. returns None).  Do we want to move this logic into `max_fragment_id`?

rust/lance-table/src/format/manifest.rs
wjones127: This should only return `None` if an empty dataset was written with Lance v0.4.9 or earlier. Added a test for that. If there is a non-empty fragment list, then it should return `Some()`.

rust/lance/src/index/prefilter.rs
wjones127: In all of these cases it makes sense. It might not universally make sense. It's meant to differentiate between:

```
fragments: []
```

and 
```
fragments: [Fragment(id=0)]
```

So want to keep that distinction

PR #4086, opened 2025-06-25
feat: support sql api for dataset

### LINKED ISSUE #3992: Support SQL api for dataset
Currently, lance does not support a complete SQL. If users want to use SQL query data. They need to fetch data from Lance into Arrow, then use duckdb to query Arrow. It's not user-friendly.

It would be better to have it. And I am going to implement it.

yanghua:
Closes #3992

### DIFF
--- a/rust/lance-core/src/error.rs
+++ b/rust/lance-core/src/error.rs
@@ -111,11 +111,6 @@ pub enum Error {
         minor_version: u16,
         location: Location,
     },
-    #[snafu(display("DataFusion inner error: {source}, {location}"))]
-    DataFusionInnerError {
-        source: BoxedError,
-        location: Location,
-    },
 }
 
 impl Error {
--- a/rust/lance-core/src/error.rs
+++ b/rust/lance-core/src/error.rs
@@ -111,6 +111,11 @@ pub enum Error {
         minor_version: u16,
         location: Location,
     },
+    #[snafu(display("DataFusion inner error: {source}, {location}"))]
+    DataFusionInnerError {
+        source: BoxedError,
+        location: Location,
+    },
 }
 
 impl Error {
--- a/rust/lance/src/dataset.rs
+++ b/rust/lance/src/dataset.rs
@@ -75,7 +75,7 @@ use self::refs::Tags;
 use self::scanner::{DatasetRecordBatchStream, Scanner};
 use self::transaction::{Operation, Transaction};
 use self::write::write_fragments_internal;
-use crate::dataset::sql::SqlOptions;
+use crate::dataset::sql::SqlBuilder;
 use crate::datatypes::Schema;
 use crate::error::box_error;
 use crate::io::commit::{
@@ -1455,8 +1455,8 @@ impl Dataset {
     /// Run a SQL query against the dataset.
     /// The underlying SQL engine is DataFusion.
     /// Please refer to the DataFusion documentation for supported SQL syntax.
-    pub fn sql(&mut self, sql: &str) -> SqlOptions {
-        SqlOptions {
+    pub fn sql(&mut self, sql: &str) -> SqlBuilder {
+        SqlBuilder {
             dataset: Some(self.clone()),
             sql: sql.to_string(),
             ..Default::default()
@@ -1906,9 +1906,7 @@ mod tests {
     use crate::dataset::transaction::DataReplacementGroup;
     use crate::dataset::WriteMode::Overwrite;
     use crate::index::vector::VectorIndexParams;
-    use crate::utils::test::{
-        copy_test_data_to_tmp, DatagenExt, FragmentCount, FragmentRowCount, TestDatasetGenerator,
-    };
+    use crate::utils::test::copy_test_data_to_tmp;
 
     use arrow::array::{as_struct_array, AsArray, GenericListBuilder, GenericStringBuilder};
     use arrow::compute::concat_batches;
@@ -1945,7 +1943,6 @@ mod tests {
     use lance_table::format::{DataFile, WriterVersion};
 
     use all_asserts::assert_true;
-    use arrow_array::types::Int64Type;
     use lance_testing::datagen::generate_random_array;
     use pretty_assertions::assert_eq;
     use rand::seq::SliceRandom;
@@ -6541,54 +6538,5 @@ mod tests {
         assert_eq!(dataset.get_fragments()[0].id(), 3);
         assert_eq!(dataset.get_fragments()[1].id(), 4);
         assert_eq!(dataset.manifest.max_fragment_id(), Some(4));
-    }
-
-    #[tokio::test]
-    async fn test_sql() {
-        let test_dir = tempdir().unwrap();
-        let test_uri = test_dir.path().to_str().unwrap();
-        let mut ds = gen()
-            .col("x", array::step::<Int32Type>())
-            .col("y", array::step_custom::<Int32Type>(0, 2))
-            .into_dataset(
-                test_uri,
-                FragmentCount::from(10),
-                FragmentRowCount::from(10),
-            )
-            .await
-            .unwrap();
-
-        let results = ds
-            .sql("SELECT SUM(x) FROM foo WHERE y > 100")
-            .table_name("foo")
-            .execute()
-            .await
-            .unwrap()
-            .collect()
-            .await
-            .unwrap();
-        assert_eq!(results.len(), 1);
-        let results = results.into_iter().next().unwrap();
-        assert_eq!(results.num_columns(), 1);
-        assert_eq!(results.num_rows(), 1);
-        // SUM(0..100) - SUM(0..50) = 3675
-        assert_eq!(results.column(0).as_primitive::<Int64Type>().value(0), 3675);
-
-        let results = ds
-            .sql("SELECT x, y, _rowid, _rowaddr FROM foo where y > 100")
-            .table_name("foo")
-            .with_row_id(true)
-            .with_row_addr(true)
-            .execute()
-            .await
-            .unwrap()
-            .collect()
-            .await
-            .unwrap();
-        assert_eq!(results.len(), 1);
-        let results = results.into_iter().next().unwrap();
-        assert_eq!(results.num_columns(), 4);
-        assert_true!(results.column(2).as_primitive::<UInt64Type>().value(0) > 100);
-        assert_true!(results.column(3).as_primitive::<UInt64Type>().value(0) > 100);
     }
 }
--- a/rust/lance/src/dataset.rs
+++ b/rust/lance/src/dataset.rs
@@ -7,6 +7,7 @@
 use arrow_array::{RecordBatch, RecordBatchReader};
 use byteorder::{ByteOrder, LittleEndian};
 use chrono::{prelude::*, Duration};
+use datafusion::prelude::SessionContext;
 use deepsize::DeepSizeOf;
 use futures::future::BoxFuture;
 use futures::stream::{self, BoxStream, StreamExt, TryStreamExt};
@@ -74,6 +75,7 @@ use self::refs::Tags;
 use self::scanner::{DatasetRecordBatchStream, Scanner};
 use self::transaction::{Operation, Transaction};
 use self::write::write_fragments_internal;
+use crate::datafusion::LanceTableProvider;
 use crate::datatypes::Schema;
 use crate::error::box_error;
 use crate::io::commit::{
@@ -1449,6 +1451,28 @@ impl Dataset {
         *self = self.checkout_version(latest_version).await?;
         Ok(())
     }
+
+    /// Run a SQL query against the dataset.
+    pub async fn sql(
+        &self,
+        sql: &str,
+        table: &str,
+        with_row_id: bool,
+        with_row_addr: bool,
+    ) -> Result<Vec<RecordBatch>> {
+        let ctx = SessionContext::new();
+        ctx.register_table(
+            table,
+            Arc::new(LanceTableProvider::new(
+                Arc::new(self.clone()),
+                with_row_id,
+                with_row_addr,
+            )),
+        )?;
+        let df = ctx.sql(sql).await?;
+        let result = df.collect().await?;
+        Ok(result)
+    }
 }
 
 pub(crate) struct NewTransactionResult<'a> {
@@ -1893,7 +1917,9 @@ mod tests {
     use crate::dataset::transaction::DataReplacementGroup;
     use crate::dataset::WriteMode::Overwrite;
     use crate::index::vector::VectorIndexParams;
-    use crate::utils::test::copy_test_data_to_tmp;
+    use crate::utils::test::{
+        copy_test_data_to_tmp, DatagenExt, FragmentCount, FragmentRowCount, TestDatasetGenerator,
+    };
 
     use arrow::array::{as_struct_array, AsArray, GenericListBuilder, GenericStringBuilder};
     use arrow::compute::concat_batches;
@@ -1930,6 +1956,7 @@ mod tests {
     use lance_table::format::{DataFile, WriterVersion};
 
     use all_asserts::assert_true;
+    use arrow_array::types::Int64Type;
     use lance_testing::datagen::generate_random_array;
     use pretty_assertions::assert_eq;
     use rand::seq::SliceRandom;
@@ -6395,6 +6422,7 @@ mod tests {
         );
     }
 
+<<<<<<< HEAD
     #[rstest]
     #[tokio::test]
     async fn test_fragment_id_zero_not_reused() {
@@ -6479,7 +6507,7 @@ mod tests {
             schema.clone(),
             vec![Arc::new(UInt32Array::from_iter_values(0..30))],
         )
-        .unwrap();
+            .unwrap();
         let batches = RecordBatchIterator::new(vec![Ok(data)], schema.clone());
         let write_params = WriteParams {
             max_rows_per_file: 10, // Force multiple fragments
@@ -6509,7 +6537,7 @@ mod tests {
             schema.clone(),
             vec![Arc::new(UInt32Array::from_iter_values(100..120))],
         )
-        .unwrap();
+            .unwrap();
         let batches = RecordBatchIterator::new(vec![Ok(data)], schema.clone());
         let write_params = WriteParams {
             mode: WriteMode::Append,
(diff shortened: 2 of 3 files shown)


rust/lance/src/dataset.rs
```diff
@@ -1485,6 +1487,28 @@ impl Dataset {
...
     }
+
+    /// Run a SQL query against the dataset.
+    pub async fn sql(
+        &self,
+        sql: &str,
+        table: &str,
+        with_row_id: bool,
+        with_row_addr: bool,
+    ) -> Result<Vec<RecordBatch>> {
```
wjones127: I think I'd prefer a builder API. Partly because it's easier to see the parameters. But also because adding and removing optional parameters isn't a breaking change. It's also easier to deprecate individual parameters when using builders.

```rust
dataset.sql("SELECT * FROM dataset").execute()
dataset.sql("SELECT * FROM table").table_name("table").execute()
dataset.sql("SELECT _rowaddr FROM dataset").with_row_addr().execute()
```

rust/lance/src/dataset.rs
```diff
@@ -1485,6 +1487,28 @@ impl Dataset {
         *self = self.checkout_version(latest_version).await?;
         Ok(())
     }
+
+    /// Run a SQL query against the dataset.
```
jackye1995, line 1455: nit: should specify this is DataFusion SQL syntax

rust/lance/src/dataset.rs
```diff
@@ -1485,6 +1487,28 @@ impl Dataset {
...
+    pub async fn sql(
+        &self,
+        sql: &str,
+        table: &str,
+        with_row_id: bool,
+        with_row_addr: bool,
+    ) -> Result<Vec<RecordBatch>> {
+        let ctx = SessionContext::new();
+        ctx.register_table(
```
jackye1995: Regardless of this API, we should probably be more explicit about how to run DataFusion SQL in documentation. This API makes people easy to run SQL for a single table, but when people want to do joins they still need to craft the session and register the table provider by themselves.

I guess it has not been clear to me that do we treat DataFusion as an internal implementation detail, or do we want to make it more explicit and public for people to use it directly? @wjones127

wjones127: There's two uses of DataFusion:

1. **As a component**: We the low-level components to build our own database operations
2. **As an integration**: We integrate into it as an engine, like we might do for DuckDB or Spark. For example: https://github.com/renato2099/lancedb/blob/21a723a89fcf35ce0ddcabf204a3445ecc577b7c/docs/src/guides/sql_querying.md

We do both of these. I think the one thing that is confusing is that when we do (1), we use our own SQL parser. So, for example, the `filter` on our `Scanner` uses a custom SQL parser, and that might have different syntax than the built-in one for DataFusion.

But otherwise, I think we should feel free to say we are using DataFusion SQL. Essentially, this method is a convenience over what is described in these docs: https://github.com/renato2099/lancedb/blob/21a723a89fcf35ce0ddcabf204a3445ecc577b7c/docs/src/guides/sql_querying.md

rust/lance/src/dataset.rs
westonpace: Do we want to return a stream here instead of forcing a collect?  This limits the API to results that fit in memory.

We could also return some kind of "query results" object which has `into_stream` and `collect` methods.  This way users still have the convenience of `collect` but aren't forced to use it.

rust/lance/src/dataset.rs
```diff
@@ -1485,6 +1487,28 @@ impl Dataset {
         *self = self.checkout_version(latest_version).await?;
         Ok(())
     }
+
+    /// Run a SQL query against the dataset.
+    pub async fn sql(
+        &self,
+        sql: &str,
+        table: &str,
```
westonpace: We will need to document the purpose of the `table` parameter.  This is not an obvious parameter to someone running an SQL query.

Another approach could be to auto-detect the table name and reject queries that reference more than one table.

rust/lance/src/dataset.rs
westonpace: Will's answer is good.  Another way to think about it is that we could, in theory, switch to a different query engine in the future (or even abandon query engines in general) and still be a datafusion table provider.

rust/lance/src/dataset.rs
```diff
@@ -1485,6 +1487,28 @@ impl Dataset {
         *self = self.checkout_version(latest_version).await?;
         Ok(())
     }
+
+    /// Run a SQL query against the dataset.
+    pub async fn sql(
```
westonpace: There are various config options when running an SQL query (e.g. how much RAM to use before spilling?  Is spilling even enabled?  How many threads should be used for compute operations?

Do you think this method will grow to address these options?  If so we might want to add a basic `SqlOptions` struct which today can just be `with_row_id` and `with_row_addr`

westonpace (commented):
I agree with @jackye1995 that SQL tends to be more of a database/catalog-level operation and not a table operation.

However, as a convenience (to enable things like sorting and aggregation) it's probably ok as a sort of utility API.

rust/lance/src/dataset.rs
yanghua: sounds reasonable

rust/lance/src/dataset.rs
yanghua: Yes, I will use `builder pattern` to convergence the potential config options.

yanghua:
The background of this requirement is that we need to let users run SQLs on the lightweight sample lance datasets.

The current way lacks flexibility. We use lance scanner to filter an arrow dataset, then forward it to DuckDB to run a SQL on it.

Jay-ju:
@yanghua I have a question. Can the SQL mode here support fts and fast_search?

Jay-ju:
@yanghua @jackye1995 Can this be updated? I want to do something based on this PR.

yanghua:
Will refactor the PR, a little busy, sorry.

jackye1995:
@Jay-ju can't you just use the DataFusion integration directly? https://lancedb.github.io/lance/integrations/datafusion/

rust/lance-core/src/error.rs
```diff
@@ -111,6 +111,11 @@ pub enum Error {
         minor_version: u16,
         location: Location,
     },
+    #[snafu(display("DataFusion inner error: {source}, {location}"))]
+    DataFusionInnerError {
```
jackye1995: currently I think all datafusion errors are converted to execution error already?

```rust
#[cfg(feature = "datafusion")]
impl From<Error> for datafusion_common::DataFusionError {
    #[track_caller]
    fn from(e: Error) -> Self {
        Self::Execution(e.to_string())
    }
}
```

rust/lance/src/dataset.rs
```diff
@@ -6526,4 +6542,53 @@ mod tests {
         assert_eq!(dataset.get_fragments()[1].id(), 4);
         assert_eq!(dataset.manifest.max_fragment_id(), Some(4));
     }
+
+    #[tokio::test]
+    async fn test_sql() {
+        let test_dir = tempdir().unwrap();
```
jackye1995: we can just write to a dataset at in-memory location to avoid having a tempdir

rust/lance/src/dataset/sql.rs
```diff
@@ -0,0 +1,97 @@
...
+use std::sync::Arc;
+
+/// Customize the params of dataset's sql API.
+#[derive(Clone, Debug)]
+pub struct SqlOptions {
+    /// the dataset to run the SQL query
+    pub(crate) dataset: Option<Dataset>,
+
+    /// the SQL query to run
+    pub(crate) sql: String,
```
jackye1995, line 19: why is the SQL string a part of the SQL option? Maybe it makes more sense to call the struct `SqlBuilder` based on how it is used?

rust/lance/src/dataset/sql.rs
```diff
@@ -0,0 +1,97 @@
...
+
+    /// the name of the table to register in the datafusion context
+    pub(crate) table_name: String,
```
jackye1995, line 22: I think this should be optional, given we have a default that is `dataset`

rust/lance/src/dataset/sql.rs
```diff
@@ -0,0 +1,97 @@
...
+        let ctx = SessionContext::new();
+        ctx.register_table(
+            self.table_name,
+            Arc::new(LanceTableProvider::new(
+                Arc::new(self.dataset.unwrap()),
+                self.row_id,
+                self.row_addr,
+            )),
+        )?;
+        let df = ctx.sql(&self.sql).await?;
```
jackye1995, line 77: why not just return df? So user can also do things like `explain` if they want.

rust/lance/src/dataset/sql.rs
yanghua: Yes, I thought `SqlOptionBuilder` or `SqlBuilder` may be a better name. Actually, this PR is not prepared. I just did some refactor before going bed, lol.

rust/lance/src/dataset/sql.rs
yanghua: This is based on Weston's comments:

> We could also return some kind of "query results" object which has into_stream and collect methods. This way users still have the convenience of collect but aren't forced to use it.

IMO, this may be a trade-off. Returning the `df` object would be very free and more flexible. But it requires the users to be very familiar with `datafusion`'s interfaces. Returning a simple object of the query result, users may not lose some flexibility, but they are not confused by many APIs coming from `df`.

jackye1995: what if we make query result hold the DataFrame, and offer the related methods to wrap important DataFusion APIs? Because I think at least `explain` is pretty important, this is basically a dry run feature so people can know what will be executed if the execution could be expensive.

yanghua: Sounds reasonable, and that would make the query result to be extendability.

yanghua:
@jackye1995 please take a look when you have time.

rust/lance/src/dataset/sql.rs
```diff
@@ -0,0 +1,171 @@
...
+            .try_collect::<Vec<_>>()
+            .await?)
+    }
+
+    pub fn dataframe(&self) -> &DataFrame {
+        &self.dataframe
+    }
+
+    pub async fn explain(&self, verbose: bool, analyze: bool) -> DataFrame {
+        self.dataframe.clone().explain(verbose, analyze).unwrap()
```
jackye1995: I think based on what we discussed before, we don't want to expose the dataframe itself. For explain, the returned DataFrame has one string column named "plan". So I think we can do something like

```rust
    let explained_df = df.explain(false, false)?;
    let batches = explained_df.collect().await?;
    let mut lines = Vec::new();
    for batch in &batches {
        let column = batch.column(0);
        let array = column
            .as_any()
            .downcast_ref::<StringArray>()
            .expect("Expected StringArray in 'plan' column for DataFrame.explain");
        for i in 0..array.len() {
            lines.push(array.value(i).to_string());
        }
    }

    Ok(lines.join("\n"))
```

rust/lance/src/dataset/sql.rs
```diff
@@ -0,0 +1,171 @@
...
+            .unwrap()
+            .try_collect::<Vec<_>>()
+            .await?)
+    }
+
+    pub fn dataframe(&self) -> &DataFrame {
+        &self.dataframe
+    }
+
+    pub async fn explain(&self, verbose: bool, analyze: bool) -> DataFrame {
```
jackye1995: I think it's probably better to also consume self, make this `into_explain_plan(self, ...)`, because for example if it is analyze, it actually runs the plan.

rust/lance/src/dataset/sql.rs
```diff
@@ -0,0 +1,171 @@
...
+        Ok(self
+            .dataframe
+            .execute_stream()
+            .await
+            .unwrap()
+            .try_collect::<Vec<_>>()
+            .await?)
+    }
+
+    pub fn dataframe(&self) -> &DataFrame {
```
jackye1995: based on what we discussed before, we don't want to expose the dataframe itself. should we just remove this?

rust/lance/src/dataset/sql.rs
```diff
@@ -0,0 +1,171 @@
...
+impl QueryResult {
+    pub fn new(dataframe: DataFrame) -> Self {
+        Self { dataframe }
+    }
+
+    pub async fn into_stream(self) -> SendableRecordBatchStream {
+        self.dataframe.execute_stream().await.unwrap()
+    }
+
+    pub async fn collect(self) -> lance_core::Result<Vec<RecordBatch>> {
```
jackye1995: nit: what about `into_batch_rows`

rust/lance/src/dataset/sql.rs
```diff
@@ -0,0 +1,171 @@
...
+            dataset: None,
+            sql: "".to_string(),
+            table_name: Some("dataset".to_string()),
+            row_id: false,
+            row_addr: false,
+        }
+    }
+}
+
+pub struct QueryResult {
```
jackye1995: when this is returned, it is not the result yet since it has not been executed. What about calling this struct `SqlQuery`?

rust/lance/src/dataset/sql.rs
```diff
@@ -0,0 +1,171 @@
...
+                self.row_id,
+                self.row_addr,
+            )),
+        )?;
+        let df = ctx.sql(&self.sql).await?;
+        Ok(QueryResult::new(df))
+    }
+}
+
+impl Default for SqlBuilder {
```
jackye1995: I think instead of setting default, we should just use 

```rust
SqlBuilder::new(dataset, sql)
   .table_name(....)
   .with_row_id(....)
   .execute()
```

This makes sure dataset and sql are required inputs, and the other ones are optional, and you can set those default values in the `new` method. What do you think?

rust/lance/src/dataset/sql.rs
yanghua: I remember that discussion. After thinking of your comments, I changed my mind to expose `DataFrame` as an optional one to make it more flexible for users. The previous discussion is about just **only** returning a `DataFrame` to users. WDYT?

jackye1995: Then what about make this method into_dataframe()? So it's not exactly tied to the internal.

yanghua: sounds good!

rust/lance/src/dataset/sql.rs
```diff
@@ -0,0 +1,171 @@
...
+    }
+
+    /// Specify if the query result should include the internal row address.
+    /// If true, the query result will include an additional column named "_rowaddr".
+    pub fn with_row_addr(mut self, row_addr: bool) -> Self {
+        self.row_addr = row_addr;
+        self
+    }
+
+    pub async fn execute(self) -> lance_core::Result<QueryResult> {
```
jackye1995: just one other thing, I think this should not be called `execute` but just `build`, because the SQL query has not really been fully executed at this point.

rust/lance/src/dataset/sql.rs
```diff
@@ -0,0 +1,171 @@
...
+use crate::Dataset;
+use arrow_array::RecordBatch;
+use datafusion::dataframe::DataFrame;
+use datafusion::execution::SendableRecordBatchStream;
+use datafusion::prelude::SessionContext;
+use std::sync::Arc;
+
+/// A SQL builder to prepare options for running SQL queries against a Lance dataset.
+#[derive(Clone, Debug)]
+pub struct SqlBuilder {
```
jackye1995: nit: to match the result `SqlQuery`, maybe call this a `SqlQueryBuilder`?

rust/lance/src/dataset/sql.rs
yanghua: What about `into_batch_records`?

yanghua:
@jackye1995 Addressed your comments, please take a look.

rust/lance/src/dataset/sql.rs
```diff
@@ -0,0 +1,214 @@
...
+                row_id,
+                row_addr,
+            )),
+        )?;
+        let df = ctx.sql(&self.sql).await?;
+        Ok(SqlQueryBuilder::new(df))
+    }
+}
+
+pub struct SqlQueryBuilder {
```
jackye1995: sorry maybe I was not clear, I was saying the `SqlBuilder` could be renamed as `SqlQueryBuilder` to match the result of `build` that is a `SqlQuery`, so this struct should still be named `SqlQuery`.

rust/lance/src/dataset/sql.rs
```diff
@@ -0,0 +1,214 @@
...
+    /// If true, the query result will include the internal row address
+    pub(crate) row_addr: Option<bool>,
+}
+
+impl SqlBuilder {
+    pub fn new(dataset: Dataset, sql: &str) -> Self {
+        Self {
+            dataset,
+            sql: sql.to_string(),
+            table_name: Some("dataset".to_string()),
```
jackye1995: in this setup, `table_name`, `row_id`, `row_addr` don't need to be an option.

rust/lance/src/dataset/sql.rs
```diff
@@ -0,0 +1,214 @@
...
+    pub(crate) dataset: Dataset,
+
+    /// The SQL query to run
+    pub(crate) sql: String,
+
+    /// the name of the table to register in the datafusion context
+    pub(crate) table_name: Option<String>,
+
+    /// If true, the query result will include the internal row id
+    pub(crate) row_id: Option<bool>,
```
jackye1995: should be named `with_row_id` and `with_row_addr` to match the inputs for `LanceTableProvider`

rust/lance/src/dataset/sql.rs
yanghua: Yes, it seems we can use the default value, and make these fields non-option.

rust/lance/src/dataset.rs
```diff
@@ -60,6 +60,7 @@ pub mod refs;
 pub(crate) mod rowids;
 pub mod scanner;
 mod schema_evolution;
+mod sql;
```
jackye1995: I think this mod needs to be public to access SqlQuery and SqlQueryBuilder?

jackye1995 (approved):
looks good to me! just 1 minor comment for the mod

Issue #3905, opened 2025-05-29
Implement copy that crosses filesystem boundaries


wjones127:
Right now, we use ObjectStore's `copy` implementation to move index files:

https://github.com/lancedb/lance/blob/bf2f72ec2645701ce810bb81b6b651c4fa06977a/rust/lance-index/src/scalar/lance_format.rs#L269

However, this can fail on Windows if temporary directory are on one drive but data is on another:

https://github.com/lancedb/lancedb/actions/runs/15332697378/job/43143056146?pr=2404

> [LocalFileSystem::copy](https://docs.rs/object_store/0.11.2/object_store/local/struct.LocalFileSystem.html#method.copy) is implemented using [std::fs::hard_link](https://doc.rust-lang.org/nightly/std/fs/fn.hard_link.html), and therefore does not support copying across filesystem boundaries. [^1]

[^1]: https://docs.rs/object_store/0.11.2/object_store/local/struct.LocalFileSystem.html#cross-filesystem-copy

We already have places where we carve out local filesystems:

https://github.com/lancedb/lance/blob/bf2f72ec2645701ce810bb81b6b651c4fa06977a/rust/lance-io/src/object_store.rs#L523-L526

We can do something similar with `copy`.

PR #4088, opened 2025-06-25
fix: use std::fs::copy for local filesystem cross-filesystem copies

wjones127:
## Summary
Fixes issue #3905 where ObjectStore::copy would fail when copying files across filesystem boundaries (e.g., Windows different drives) because it relied on hard links internally.

## Changes
- Add `copy_file` function in `lance-io/src/local.rs` using `std::fs::copy` which supports cross-filesystem copies
- Modify `ObjectStore::copy` to use local copy function for local filesystems  
- Add comprehensive unit tests for cross-filesystem copy functionality
- Add regression test for FTS index creation that exercises the problematic code path

## Test plan
- [x] Unit tests pass for cross-filesystem copy scenarios
- [x] Regression test passes for FTS index creation
- [x] All existing tests continue to pass
- [x] Code passes clippy and formatting checks

The fix ensures that FTS index creation will work reliably even when temporary directories are on different filesystems, resolving the Windows cross-drive issue described in the original bug report.

### DIFF
--- a/rust/lance-io/src/local.rs
+++ b/rust/lance-io/src/local.rs
@@ -49,6 +49,28 @@ pub fn remove_dir_all(path: &Path) -> Result<()> {
     Ok(())
 }
 
+/// Copy a file from one location to another, supporting cross-filesystem copies.
+///
+/// Unlike hard links, this function works across filesystem boundaries.
+pub fn copy_file(from: &Path, to: &Path) -> Result<()> {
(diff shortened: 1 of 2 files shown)


vishnu2000vcube:
Hi @wjones127 @jackye1995 Is there anyway I can add this change to my pylance, since this issue is stopping me to use lanced.

Thanks 
Vishnu

Issue #3350, opened 2025-01-07
Fuzz testing and finish implementation for large_string / large_binary in 2.1


westonpace:
There is at least one `todo!()` (and I think several) related to 64-bit offsets in 2.1 encodings.  We should be able to easily extend the 2.0 fuzz tests to include 2.1 and fill out the 64-bit implementations.

westonpace:
Make sure to also test dictionary encoding of 64-bit binary types

HaochengLIU:
This problem has two parts. The above CR will support the generic string compression and the next one will support FSST.

HaochengLIU:
I think this issue could be closed? @westonpace

westonpace:
Agreed, thanks for your help!

PR #4092, opened 2025-06-26
feat: add 64-bit offset support to FSST compression for large binary/string data in V2.1

HaochengLIU:
This PR addresses https://github.com/lancedb/lance/issues/3350. 
This enhancement enables Lance to handle datasets with string/binary fields exceeding 4GB in total size while maintaining optimal compression ratios through FSST.
It breaks the binary data layout for binary/string data in lance V2.1(unstable) format.

It also absorbs PR #4095 to not error on highly compressible FSST by Weston.

# Summary
* Extends FSST compression support to handle large binary and string data by implementing 64-bit offset support alongside the existing 32-bit offsets for miniblock, fullzip and dictionary encoding. This enables efficient compression of variable-width data that exceeds the 4GB limit imposed by 32-bit offsets. 

# Changes Made
* Made FSST compression generic - Updated FSST implementation to work with both 32-bit and 64-bit offset types using generic programming
* Updated compression strategy - Modified DefaultCompressionStrategy to support both 32-bit and 64-bit offsets in variable-width data blocks
* Enhanced decompressors - Added auto-detection of offset bit width in FsstMiniBlockDecompressor with layered decompression architecture
* Extended block compressor - Updated BinaryBlockDecompressor and related components to handle 64-bit offsets
* Updated primitive encoder - Enhanced PrimitiveStructuralEncoder to support 64-bit offset encoding
Added dependencies - Included arrow-array dependency in FSST Cargo.toml

# Breaking Change Context
It introduces a new u8 byte - used for `bits_per_offset` - to the current datablock compression memory layout. Now it supports both 32 and 64 bits. Since Lance 2.1 is on an unstable version, it should not affect any existing prod users.
```
/// ----------------------------------------------------------------------------------------
/// | bits_per_offset(NEW!) | number of values  | bytes_start_offset | offsets data | bytes data
/// ----------------------------------------------------------------------------------------
/// |       1 byte    | bits_per_offset/8 | bits_per_offset/8  |  offsets_len | dat_len
/// ----------------------------------------------------------------------------------------
```

# Test Results
✅ All existing ones are passing
✅ A new test:`test_fsst_64_bit_offsets` in fsst.rs is passing
✅ A new test: `test_large_binary_fsst` in binary.rs is passing

# API Impact
None

### DIFF
--- a/rust/lance-encoding/src/compression_algo/fsst/src/fsst.rs
+++ b/rust/lance-encoding/src/compression_algo/fsst/src/fsst.rs
@@ -18,9 +18,8 @@ const FSST_CODE_MASK: u16 = FSST_CODE_MAX - 1;
 const FSST_SAMPLETARGET: usize = 1 << 14;
 const FSST_SAMPLEMAXSZ: usize = 2 * FSST_SAMPLETARGET;
 
-// if the input size is less than 4MB, we mark the file header and copy the input to the output as is
-//  TODO: FIX ME.
-pub const FSST_LEAST_INPUT_SIZE: usize = 32 * 1024; // 4MB
+// if the input size is less than 32 KB, we mark the file header and copy the input to the output as is
+pub const FSST_LEAST_INPUT_SIZE: usize = 32 * 1024;
 
 // if the max length of the input strings are less than `FSST_LEAST_INPUT_MAX_LENGTH`, we shouldn't use FSST.
 pub const FSST_LEAST_INPUT_MAX_LENGTH: u64 = 5;
@@ -559,9 +558,8 @@ fn build_symbol_table<T: OffsetSizeTrait>(
 ) -> io::Result<Box<SymbolTable>> {
     let mut st = SymbolTable::new();
     let mut best_table = SymbolTable::new();
-    // TODO FIX ME!!!!
-    //let mut best_gain = -1 * T::from_usize(FSST_SAMPLEMAXSZ).unwrap(); // worst case (everything exception)
-    let mut best_gain = T::from_usize(0).unwrap(); // Start with 0, will be updated during iteration
+    // worst case (everything exception), will be updated later
+    let mut best_gain = T::zero() - T::from_usize(FSST_SAMPLEMAXSZ).unwrap();
 
     let mut byte_histo = [0; 256];
     for c in &sample_buf {
@@ -592,7 +590,7 @@ fn build_symbol_table<T: OffsetSizeTrait>(
             let mut prev_code = st.find_longest_symbol_from_char_slice(&word[curr..]);
             curr += st.symbols[prev_code as usize].symbol_len() as usize;
 
-            // Fix the gain calculation - avoid arithmetic on Option<T>
+            // Avoid arithmetic on Option<T>
             let symbol_len = st.symbols[prev_code as usize].symbol_len() as usize;
             let escape_cost = if is_escape_code(prev_code) { 1 } else { 0 };
             let gain_contribution = symbol_len.saturating_sub(1 + escape_cost);
@@ -631,7 +629,7 @@ fn build_symbol_table<T: OffsetSizeTrait>(
                     symbol_len = st.symbols[curr_code as usize].symbol_len();
                 }
 
-                // Fix the gain calculation
+                // Avoid arithmetic on Option<T>
                 let symbol_len_usize = symbol_len as usize;
                 let escape_cost = if is_escape_code(curr_code) { 1 } else { 0 };
                 let gain_contribution = symbol_len_usize.saturating_sub(1 + escape_cost);
@@ -828,12 +826,6 @@ fn decompress_bulk<T: OffsetSizeTrait>(
             let escape_mask = (next_block & 0x80808080u32)
                 & ((((!next_block) & 0x7F7F7F7Fu32) + 0x7F7F7F7Fu32) ^ 0x80808080u32);
             if escape_mask == 0 {
-                // Ensure we have enough space for all 4 bytes (worst case: 4 * MAX_SYMBOL_LENGTH)
-                let required_space = *out_curr + 4 * MAX_SYMBOL_LENGTH;
-                if required_space > out.len() {
-                    out.resize(required_space.max(out.len() * 2), 0);
-                }
-
                 // 0th byte
                 code = compressed_strs[in_curr] as usize;
                 len = lens[code] as usize;
@@ -874,12 +866,6 @@ fn decompress_bulk<T: OffsetSizeTrait>(
                 in_curr += 1;
                 *out_curr += len;
             } else {
-                // Ensure space needed for escape cases (worst case: 3 * MAX_SYMBOL_LENGTH + 1 escape byte)
-                let required_space = *out_curr + 3 * MAX_SYMBOL_LENGTH + 1;
-                if required_space > out.len() {
-                    out.resize(required_space.max(out.len() * 2), 0);
-                }
-
                 let first_escape_pos = escape_mask.trailing_zeros() >> 3;
                 if first_escape_pos == 3 {
                     // 0th byte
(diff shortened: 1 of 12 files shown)


HaochengLIU:
TODO: check big strings with function test_small_strings(...) after https://github.com/lancedb/lance/pull/4095/files is merged.

rust/lance-encoding/src/encodings/physical/binary.rs
```diff
@@ -439,15 +530,32 @@ pub mod tests {
...
         field_metadata.insert(COMPRESSION_META_KEY.to_string(), "fsst".into());
+        let field = Field::new("", data_type, true).with_metadata(field_metadata);
+        check_round_trip_encoding_random(field, LanceFileVersion::V2_1).await;
+    }
 
-        let field = Field::new("", DataType::Utf8, true).with_metadata(field_metadata);
+    #[rstest]
+    #[test_log::test(tokio::test)]
+    async fn test_large_binary_fsst(
+        #[values(STRUCTURAL_ENCODING_MINIBLOCK, STRUCTURAL_ENCODING_FULLZIP)]
```
HaochengLIU, line 540: How could i ensure that it's testing both dictionary and miniblock?
Right now only by manually running could i l know that it's doing such work

rust/lance-encoding/src/compression_algo/fsst/src/fsst.rs
```diff
@@ -968,6 +984,10 @@ fn decompress_bulk(
         if in_curr < in_end {
             // last code cannot be an escape code
             let code = compressed_strs[in_curr] as usize;
+            let required_space = *out_curr + lens[code] as usize;
+            if required_space > out.len() {
+                out.resize(required_space.max(out.len() * 2), 0);
+            }
```
westonpace: Do we still need this?  When would we not have enough size?

rust/lance-encoding/src/encodings/logical/primitive.rs
```diff
@@ -549,14 +549,35 @@ impl DecodePageTask for DecodeMiniBlockTask {
...
+                        });
+                    }
+                    64 => {
+                        let indices = fixed_width_data_block.data.borrow_to_typed_slice::<i64>();
+                        let indices = indices.as_ref();
+
+                        indices.iter().for_each(|&idx| {
+                            data_builder.append(dictionary, idx as u64..idx as u64 + 1);
+                        });
+                    }
```
westonpace, line 570: No harm in this, but I'm not sure we should ever be creating 64-bit dictionary indices.

rust/lance-encoding/src/encodings/logical/primitive.rs
```diff
@@ -3824,7 +3847,58 @@ impl PrimitiveStructuralEncoder {
...
+                        let mut indices_buffer =
+                            Vec::with_capacity(variable_width_data_block.num_values as usize);
+
+                        offsets
+                            .iter()
+                            .zip(offsets.iter().skip(1))
+                            .for_each(|(&start, &end)| {
+                                let key =
+                                    &variable_width_data_block.data[start as usize..end as usize];
+                                let idx: i64 = *map.entry(U8SliceKey(key)).or_insert_with(|| {
```
westonpace, line 4028: This should be `i32`.

rust/lance-encoding/src/encodings/logical/primitive.rs
```diff
@@ -3824,7 +3847,58 @@ impl PrimitiveStructuralEncoder {
...
+                            data: LanceBuffer::reinterpret_vec(dictionary_buffer),
+                            offsets: LanceBuffer::reinterpret_vec(dictionary_offsets_buffer),
+                            bits_per_offset: 64,
+                            num_values: curr_idx as u64,
+                            block_info: BlockInfo::default(),
+                        });
+
+                        let mut indices_data_block = DataBlock::FixedWidth(FixedWidthDataBlock {
+                            data: LanceBuffer::reinterpret_vec(indices_buffer),
+                            bits_per_value: 64,
```
westonpace, line 4047: This can be 32 bits per value

rust/lance-encoding/src/encodings/physical/binary.rs
```diff
@@ -289,44 +302,89 @@ impl MiniBlockDecompressor for BinaryMiniBlockDecompressor {
 }
 
 /// Most basic encoding for variable-width data which does no compression at all
+/// The DataBlock memory layout looks like below:
+/// ----------------------------------------------------------------------------------------
+/// | bits_per_offset | number of values  | bytes_start_offset | offsets data | bytes data
+/// ----------------------------------------------------------------------------------------
+/// |       1 byte    | bits_per_offset/8 | bits_per_offset/8  |  offsets_len | dat_len
```
westonpace: Does the `number of values` need to change with the data type?  Can it always be 32 bits?  A single page will never have more than 2^32 values in it.

rust/lance-encoding/src/encodings/physical/binary.rs
```diff
@@ -289,44 +302,89 @@ impl MiniBlockDecompressor for BinaryMiniBlockDecompressor {
...
+                        Ok(LanceBuffer::Owned(output))
+                    }
+                    64 => {
+                        let offsets = variable_width_data.offsets.borrow_to_typed_slice::<u64>();
+                        let offsets = offsets.as_ref();
+                        // the first bit stores the bits_per_offset, the next 8 bytes store the number of values,
+                        // then 8 bytes for bytes_start_offset,
+                        // then offsets data, then bytes data.
+
+                        let bytes_start_offset = 1 + 8 + 8 + std::mem::size_of_val(offsets) as u64;
```
westonpace, line 350: Seems unlikely this will need to be a u64.  Can we make this a u32?

rust/lance-encoding/src/encodings/physical/binary.rs
westonpace: https://github.com/lancedb/lance/issues/4249

rust/lance-encoding/src/encodings/physical/fsst.rs
```diff
@@ -43,37 +43,78 @@ impl FsstCompressed {
...
+                            num_values: variable_width.num_values,
+                            block_info: BlockInfo::new(),
+                        };
+
+                        Ok(Self {
+                            data: compressed,
+                            symbol_table,
+                        })
+                    }
+                    64 => {
```
westonpace, line 80: This is more for a follow-up but I question the need to compress 64 bit offsets into 64 bit offsets.  Generally we have one of two cases:

* The data is using 64 bit offsets but the actual values all fit within 32 bit offsets (user is being paranoid and over-allocating their offsets)
* The data truly doesn't fit into 64 bit offsets (there are very large values in this array we are given)

In the first case we could compress into 32 bit offsets and be fine (as long as we decompress back into the right thing when reading)

In the second case we probably shouldn't be using FSST in this way anyways.  A single symbol table for a string that large is likely to be ineffective.

This second problem is a more general problem with FSST that we can tackle in a follow-up and let's address this concern then.

westonpace: This is more about simplification than performance.  We should be bitpacking offsets after FSST in either case.

rust/lance-encoding/src/compression_algo/fsst/src/fsst.rs
HaochengLIU: removed

rust/lance-encoding/src/encodings/logical/primitive.rs
HaochengLIU: Discussed offline, we will keep 64 here

rust/lance-encoding/src/encodings/logical/primitive.rs
HaochengLIU: ditto as above

rust/lance-encoding/src/encodings/physical/binary.rs
HaochengLIU: I use `std::env::set_var("LANCE_ENCODING_DICT_DIVISOR", "1");` as a workaround right now to ensure dict  encoding is tested

westonpace (approved):
I've been thinking over the idea that we might have more than 2^32 values in a page and I can't come up with any good reason we would ever want to allow it.  Also, if we decide that is possible, then it is possible even in other places (such as fixed sized values and strings with 4-byte offsets) that we aren't accounting for today.

I'd like to consolidate this logic on the assertion that a single page will never have more than 2^32 values.

However, I don't want to bike shed.  I'll do it in a follow-up.

westonpace:
Hmm...then again, we do have...

https://github.com/lancedb/lance/blob/7069045672da668a39efb98aec6fd6687f324a44/protos/file2.proto#L18

So maybe we need to consolidate on 64 bits

PR #4095, opened 2025-06-26
fix: don't error on highly compressible FSST

### LINKED ISSUE #4079: Random take with v2.1 encounters 'output offsets buffer too small for FSST decoder' error.
I write a unit test to show the case. test_take_rows_2_0 works fine while test_take_rows_2_1 encounters error 'output offsets buffer too small for FSST decoder'.

Is my test case wrong, or is there a bug in v2.1? Please help.

https://github.com/wojiaodoubao/lance/blob/take-fsst-error/rust/lance/src/take_fsst_error.rs

```
running 1 test
test take_fsst_error::tests::test_take_rows_2_1 ... FAILED

failures:

---- take_fsst_error::tests::test_take_rows_2_1 stdout ----

thread 'take_fsst_error::tests::test_take_rows_2_1' panicked at rust/lance/src/take_fsst_error.rs:27:88:
called `Result::unwrap()` on an `Err` value: Internal { message: "Error decoding batch: LanceError(IO): output offsets buffer too small for FSST decoder, rust/lance-encoding/src/encodings/physical/fsst.rs:235:9", location: Location { file: "rust/lance-encoding/src/decoder.rs", line: 2372, column: 31 } }


failures:
    take_fsst_error::tests::test_take_rows_2_1

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 557 filtered out; finished in 1.61s
```

westonpace:
If we got a lot of highly compressible strings (user example was small prefix + counter) then its possible we have 1024 values in a single mini-block.  The old logic assumed we would have at most 1024 output offsets and, in this case, we have 1025 and this would fail.

However, it's actually very easy to know exactly how many output offsets we will have (it's the same as the input) and so this sets the expected value exactly.  We also add a test case to trigger this situation.

Closes #4079

### DIFF
--- a/rust/lance-encoding/src/compression_algo/fsst/src/fsst.rs
+++ b/rust/lance-encoding/src/compression_algo/fsst/src/fsst.rs
@@ -19,7 +19,7 @@ const FSST_SAMPLETARGET: usize = 1 << 14;
 const FSST_SAMPLEMAXSZ: usize = 2 * FSST_SAMPLETARGET;
 
 // if the input size is less than 4MB, we mark the file header and copy the input to the output as is
-pub const FSST_LEAST_INPUT_SIZE: usize = 4 * 1024 * 1024; // 4MB
+pub const FSST_LEAST_INPUT_SIZE: usize = 32 * 1024; // 4MB
 
 // if the max length of the input strings are less than `FSST_LEAST_INPUT_MAX_LENGTH`, we shouldn't use FSST.
 pub const FSST_LEAST_INPUT_MAX_LENGTH: u64 = 5;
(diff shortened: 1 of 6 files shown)


rust/lance-encoding/src/compression_algo/fsst/src/fsst.rs
```diff
@@ -19,7 +19,7 @@ const FSST_SAMPLETARGET: usize = 1 << 14;
 const FSST_SAMPLEMAXSZ: usize = 2 * FSST_SAMPLETARGET;
 
 // if the input size is less than 4MB, we mark the file header and copy the input to the output as is
-pub const FSST_LEAST_INPUT_SIZE: usize = 4 * 1024 * 1024; // 4MB
+pub const FSST_LEAST_INPUT_SIZE: usize = 32 * 1024; // 4MB
```
HaochengLIU, line 22: r/4MB/32KB/g

rust/lance-encoding/src/encodings/logical/primitive.rs
```diff
@@ -79,6 +79,8 @@ pub mod fullzip;
 pub mod miniblock;
 
 const FILL_BYTE: u8 = 0xFE;
+pub const MINIBLOCK_MAX_VALUES: u64 = 4 * 1024;
```
HaochengLIU, line 82: where are these two consts used?

HaochengLIU:
I will absorb this CR into mine #4092

PR #4096, opened 2025-06-27
feat: support IVF_SQ index

BubbleCal:


rust/lance/src/index/vector/ivf.rs
```diff
@@ -516,15 +536,6 @@ pub(crate) async fn optimize_vector_indices_v2(
             .build()
             .await?;
         }
-        (sub_index_type, quantizer_type) => {
```
BubbleCal, line 519: we now cover all types

PR #4098, opened 2025-06-29
chore: update tantivy version to 0.24.1

Dig-Doug:


Cargo.lock
```diff
@@ -4823,8 +4854,8 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "2ad1f34fb1d46c07e05ea1abb848078adfa3c19530d1cf7a689f97541f6e1a3e"
 dependencies = [
  "lindera",
- "tantivy",
- "tantivy-tokenizer-api",
+ "tantivy 0.22.0",
```
Xuanwo: Can we bump this together to avoid have two versions of `tantivy`?

Cargo.toml
```diff
@@ -150,7 +150,7 @@ serde = { version = "^1" }
 serde_json = { version = "1" }
 shellexpand = "3.0"
 snafu = "0.8"
-tantivy = { version = "0.22.0", features = ["stopwords"] }
+tantivy = { version = "^0.24.1", features = ["stopwords"] }
```
Xuanwo: Using `0.24.1` is enough, we don't need to have a `^` here.

Xuanwo (commented):
Thank you @Dig-Doug for working on this. This change looks mostly good to me.

Dig-Doug:
Updated.

Cargo.lock
```diff
@@ -188,9 +188,9 @@ checksum = "7c02d123df017efcdfbd739ef81735b36c5ba83ec3c59c80a9d7ecc718f92e50"
 
 [[package]]
 name = "arrow"
-version = "55.1.0"
+version = "55.2.0"
```
Xuanwo: Hi, I'm assuming this change isn't related? `arrow` is a critical dependency for us, and I hope we can update it in a separate PR.

We can handle this PR as follows:

```shell
# Checkout the unchanged Cargo.lock
git checkout Cargo.lock
# Run any cargo command to trigger an update of 'Cargo.lock'
cargo check
```

Please avoid running `cargo update` to prevent any unexpected upgrades.

Dig-Doug: Done.

Xuanwo (approved):
Thank you @Dig-Doug! This PR LGTM now.

Xuanwo:
The Python regression is not included in our CI. I will investigate.

Issue #3579, opened 2025-03-20
Allow users to request repetition index be cached in 2.1


westonpace:
Currently we cache the repetition index for small data types (mini block layout) and pay the 2-IOP cost for large data types (full zip).  If the data is in remote storage and users have enough RAM then it may make sense to cache the repetition index in all scenarios.

Xuanwo:
https://github.com/lancedb/lance/pull/4104 is implemented so that all repetition indexes are cached in memory when available. Do you think we should add a configuration option for this?

westonpace:
Yes, I think we can expose it in https://github.com/lancedb/lance/blob/baa89013fd5b580ac9ffeaff339fb1554e6957c4/rust/lance-file/src/v2/reader.rs#L315

westonpace:
Now that https://github.com/lancedb/lance/pull/4287 has merged is this ready to be closed?

Xuanwo:
> Now that [#4287](https://github.com/lancedb/lance/pull/4287) has merged is this ready to be closed?

Yeah, I'll expose this setting to python and see how it works.

PR #4104, opened 2025-06-30
feat(encoding): cache repetition index for FullZip encoding

Xuanwo:
## Summary

This PR introduces caching for repetition index in FullZip encoding to reduce I/O operations when reading variable-length data from remote storage.

## Why this change?

When reading variable-length data (like strings) with FullZip encoding, the system needs to load a repetition
index to determine byte offsets. Currently, this index is loaded from storage for every read operation, causing
2 additional I/O requests per query. For remote storage scenarios with sufficient RAM, caching this index can
significantly improve performance.

## What's changed?

- Added `FullZipCacheableState` struct to store decoded repetition index in memory
- Implemented cache initialization in `FullZipScheduler::initialize()` that loads and decodes the repetition
index once
- Added `schedule_ranges_with_cached_rep()` method that uses cached data directly without I/O
- Modified `schedule_ranges_rep()` to check for cached data first before falling back to disk reads
- Integrated with existing `CachedPageData` trait system for cache management

This is the foundation work for issue #3579. The caching is currently automatic when a repetition index exists, with configuration options coming in follow-up PRs.

---

I changed this PR into disable repetition index cache by default so users won't be affected. I will re-add this part after configration from `field metadata` and `ReadParams` has been added.

### DIFF
--- a/rust/lance-encoding/src/encodings/logical/primitive.rs
+++ b/rust/lance-encoding/src/encodings/logical/primitive.rs
@@ -1932,15 +1932,12 @@ impl FullZipScheduler {
     ) -> Result<BoxFuture<'static, Result<Box<dyn StructuralPageDecoder>>>> {
         // Check if we have cached repetition index data
         if let Some(cached_state) = &self.cached_state {
-            if let Some(rep_index_buffer) = &cached_state.rep_index_buffer {
-                // Use cached data directly
-                return self.schedule_ranges_with_cached_rep(
-                    ranges,
-                    io,
-                    rep_index_buffer,
-                    rep_index.bytes_per_value,
-                );
-            }
+            return self.schedule_ranges_with_cached_rep(
+                ranges,
+                io,
+                &cached_state.rep_index_buffer,
+                rep_index.bytes_per_value,
+            );
         }
 
         // Fall back to loading from disk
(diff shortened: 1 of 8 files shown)


rust/lance-encoding/src/encodings/logical/primitive.rs
```diff
@@ -1822,13 +1833,117 @@ impl FullZipScheduler {
...
+                        details,
+                        data,
+                        num_rows,
+                        bits_per_offset,
+                        bits_per_offset,
+                    )) as Box<dyn StructuralPageDecoder>)
+                }
+            }
+        }
+        .boxed())
```
westonpace, line 1923: Is there anyway we can consolidate this logic across the two implementations?  That way we don't have duplication?

rust/lance-encoding/src/encodings/logical/primitive.rs
```diff
@@ -1908,16 +2023,58 @@ impl FullZipScheduler {
     }
 }
 
+/// Cacheable state for FullZip encoding, storing the decoded repetition index
+#[derive(Debug)]
+struct FullZipCacheableState {
+    /// The decoded repetition index, if present
+    rep_index: Option<RepetitionIndex>,
```
westonpace: This struct (`RepetitionIndex`) is actually specific to mini-block (we can maybe rename it `MiniBlockRepetitionIndex`) and shouldn't be needed here.

rust/lance-encoding/src/encodings/logical/primitive.rs
```diff
@@ -1908,16 +2023,58 @@ impl FullZipScheduler {
...
+    /// The raw repetition index buffer for future decoding
+    rep_index_buffer: Option<LanceBuffer>,
```
westonpace: I don't think this needs to be optional.  If there is no repetition index then we should skip the cached state entirely.  This way we only create a cached state if a repetition index is present (and don't need an option)

Xuanwo:
Let me polish a bit before merging.

rust/lance-encoding/src/encodings/logical/primitive.rs
Xuanwo: I will try in the following PRs.

Xuanwo:
Test `index::append::tests::test_query_delta_indices::index_params_2_VectorIndexParams__with_ivf_hnsw_sq_params_MetricType__L2_IvfBui` seems flaky. Can't reproduce locally.

PR #4109, opened 2025-07-01
chore: replace lazy_static with std::sync::LazyLock

Xuanwo:
This PR will replace lazy_static with std::sync::LazyLock.

lazy_static is known to compile more slowly and relies on macros, which makes maintenance more difficult. Now that there is a new standard API that can fully replace it, this change is appropriate.

### DIFF
--- a/rust/lance/src/index/vector/ivf/io.rs
+++ b/rust/lance/src/index/vector/ivf/io.rs
@@ -50,7 +50,7 @@ use crate::Result;
 
 // TODO: make it configurable, limit by the number of CPU cores & memory
 static HNSW_PARTITIONS_BUILD_PARALLEL: LazyLock<usize> =
-    LazyLock::new(|| get_num_compute_intensive_cpus());
+    LazyLock::new(get_num_compute_intensive_cpus);
 
 /// Merge streams with the same partition id and collect PQ codes and row IDs.
 async fn merge_streams(


python/Cargo.toml
```diff
@@ -3,7 +3,7 @@ name = "pylance"
 version = "0.31.0"
 edition = "2021"
 authors = ["Lance Devs <dev@lancedb.com>"]
-rust-version = "1.65"
```
Xuanwo, line 6: Is there a reason we keep the `rust-version` of python to 1.65?

Xuanwo:
CI failed for:

> (raylet) [2025-07-01 07:37:28,581 E 11809 11827] (raylet) file_system_monitor.cc:111: /tmp/ray/session_2025-07-01_07-36-00_817069_11731 is over 95% full, available space: 531230720; capacity: 76887154688. Object creation will fail if spilling is required.

BubbleCal (commented):
can we remove lazy_static from `Cargo.toml`?
or are we still using it somewhere after this?

Xuanwo:
> can we remove lazy_static from `Cargo.toml`? or are we still using it somewhere after this?

Yes, we can. Let me perform the double check.

python/Cargo.toml
westonpace: Not that I'm aware of.  Our MSRV policy is to match Datafusion.  Looks like we just haven't had the need yet and so never bumped python.

westonpace:
Approval assuming CI passes.  Can't tell if there are CI failures or if Github is just acting up.

python/Cargo.toml
Xuanwo: Cool, let's just bump it. Do you think I need to split in two PRs?

westonpace: No need.

PR #4113, opened 2025-07-01
perf: add dataset random take benchmark

yanghua:


python/python/benchmarks/test_take.py
```diff
@@ -0,0 +1,82 @@
...
+        [
+            pa.field("i", pa.int32(), nullable=False),
+            pa.field("f", pa.float32(), nullable=False),
+            pa.field("s", pa.binary(), nullable=False),
+            pa.field("fsl", pa.list_(pa.float32(), 2), nullable=False),
+            pa.field("blob", pa.binary(), nullable=False),
+        ]
+    )
+
+    batch_size = BATCH_SIZE
```
jackye1995: seems like we should also make batch size an input?

python/python/benchmarks/test_take.py
```diff
@@ -0,0 +1,82 @@
...
+    )
+
+
+def gen_ranges(total_rows, num_rows):
+    return np.random.choice(total_rows, num_rows, replace=False)
+
+
+@pytest.mark.benchmark()
+@pytest.mark.parametrize("file_size", (1024 * 1024, 1024))
+@pytest.mark.parametrize("lance_format_version", [("stable", "V2_0"), ("2.1", "V2_1")])
```
jackye1995: nit: should use 2.0 instead of "stable" to be explicit

python/python/benchmarks/test_take.py
```diff
@@ -0,0 +1,82 @@
...
+@pytest.mark.parametrize("num_rows", [1, 10, 100, 1000])
+def test_dataset_take(benchmark, file_size, lance_format_version, num_rows):
+    data_storage_version, version_name = lance_format_version
+    path = "memory://test.lance"
```
jackye1995: why not just make scheme another parameter so we don't need another PR for enabling local disk benchmark?

python/python/benchmarks/test_take.py
```diff
@@ -0,0 +1,95 @@
...
+def gen_ranges(total_rows, num_rows):
+    return np.random.choice(total_rows, num_rows, replace=False)
+
+
+@pytest.mark.benchmark()
+@pytest.mark.parametrize("file_size", (1024 * 1024, 1024))
+@pytest.mark.parametrize("lance_format_version", [("2.0", "V2_0"), ("2.1", "V2_1")])
+@pytest.mark.parametrize("num_rows", [1, 10, 100, 1000])
+@pytest.mark.parametrize("batch_size", [512, 1024, 2048])
+@pytest.mark.parametrize("scheme", ["memory", "file"])
```
jackye1995: I think you need a prefix, rather than a scheme here, because if we want to test object store it will need to be a storage prefix that includes the bucket name.

yanghua: I tend to place the tests that integrate with the object store into a single test file. Some considered reason:

* Access object store needs to specify credentials.
* memory and file schemes are out of the box to run the bench, no more information or third-party services. I tend to think the object store belongs to the integration test.

WDYT?

jackye1995: that sounds reasonable. I also like the approach of using environment variable so user can set it to run against cloud if necessary and don't need to re-write the function again. I remember we proposed that in the rust benchmark, but probably did not fully merge or finish it. What do you think about doing that approach?

yanghua: > I remember we proposed that in the rust benchmark, but probably did not fully merge or finish it.

I guess you mentioned #4143 

> What do you think about doing that approach?

Agree to reduce duplicated code. Let me have a try at supporting object store.

python/python/benchmarks/test_take.py
```diff
@@ -0,0 +1,128 @@
...
+            [arr_i, arr_f, arr_s, arr_fsl, arr_blob], schema=schema
+        )
+        batches.append(batch)
+
+    table = pa.Table.from_batches(batches, schema=schema)
+
+    storage_options = None
+    if not path.startswith(("memory://", "file://")):
+        storage_options = {
+            "access_key_id": os.getenv("ACCESS_KEY_ID", ""),
```
yanghua, line 72: IMO, We can let Pylance check these config options. If missed, it would throw an exception.

yanghua:
Have tested the object store bench in Volcengine's object store service(TOS).

jackye1995 (approved):
looks good to me

PR #4116, opened 2025-07-01
fix: allow rowid in columns for scanner

Jay-ju:
Currently, the scanner does not support passing in hidden columns, such as rowid and rowaddr, which will result in an error.
```
self = <lance.dataset.LanceScanner object at 0x7f330d7cbf60>

    def to_reader(self) -> pa.RecordBatchReader:
>       return self._scanner.to_pyarrow()
E       ValueError: LanceError(Schema): Schema error: No field named _rowid. Valid fields are data_item_id., /data00/code/emr/lance/rust/lance-datafusion/src/projection.rs:171:25

```

This PR mainly allows both with_row_id in the scanner and rowid in the columns. The reason for this is that when performing push - down on the computing side, the schema in the dataframe is generally pushed directly to the columns, rather than setting properties specifically for a certain with_row_id or with_row_addr. 

Therefore, there is a hope to make a compatibility on the lance side, similar to allowing such behavior. 
```
lance_ds = lance.dataset(path)
table = lance_ds.scanner(
        with_row_id=True, columns=["data_item_id", "_rowid"]
    ).to_table()
```

### DIFF
--- a/python/python/tests/test_lance.py
+++ b/python/python/tests/test_lance.py
@@ -248,3 +248,34 @@ def test_io_counters(tmp_path):
     dataset.to_table()
     assert lance.iops_counter() > starting_iops
     assert lance.bytes_read_counter() > starting_bytes
+
+
+@pytest.mark.parametrize(
+    "row_param, column_name",
+    [("with_row_id", "_rowid"), ("with_row_address", "_rowaddr")],
+)
+def test_with_row_id(tmp_path, row_param, column_name):
+    path = tmp_path / "test_with_inner_column.lance"
+    data = pa.table(
+        {
+            "data_item_id": pa.array([1001, 1002, 1003]),
+            "a": pa.array([1, 2, 3]),
+        }
+    )
+
+    lance.write_dataset(data, path)
+    lance_ds = lance.dataset(path)
+
+    scanner_args = {row_param: True, "columns": ["data_item_id", column_name]}
+    table = lance_ds.scanner(**scanner_args).to_table()
+    assert table.column(column_name).type == pa.uint64()
+
+    scanner_args = {row_param: True, "columns": ["data_item_id"]}
+    table = lance_ds.scanner(**scanner_args).to_table()
+    assert column_name in table.column_names
+    assert table.column(column_name).type == pa.uint64()
+
+    with pytest.raises((ValueError, OSError), match="without enabling"):
+        lance_ds.scanner(
+            **{row_param: False, "columns": ["data_item_id", column_name]}
+        ).to_table()
--- a/python/src/dataset.rs
+++ b/python/src/dataset.rs
@@ -559,6 +559,15 @@ impl Dataset {
         strict_batch_size: Option<bool>,
     ) -> PyResult<Scanner> {
         let mut scanner: LanceScanner = self_.ds.scan();
+
+        if with_row_id.unwrap_or(false) {
+            scanner.with_row_id();
+        }
+
+        if with_row_address.unwrap_or(false) {
+            scanner.with_row_address();
+        }
+
         match (columns, columns_with_transform) {
             (Some(_), Some(_)) => {
                 return Err(PyValueError::new_err(
@@ -675,14 +684,6 @@ impl Dataset {
         }
 
         scanner.scan_in_order(scan_in_order.unwrap_or(true));
-
-        if with_row_id.unwrap_or(false) {
-            scanner.with_row_id();
-        }
-
-        if with_row_address.unwrap_or(false) {
-            scanner.with_row_address();
-        }
 
         if let Some(use_stats) = use_stats {
             scanner.use_stats(use_stats);
--- a/rust/lance-core/src/datatypes/schema.rs
+++ b/rust/lance-core/src/datatypes/schema.rs
@@ -192,7 +192,7 @@ impl Schema {
                 } else {
                     candidates.push(projected_field)
                 }
-            } else if err_on_missing {
+            } else if err_on_missing && first != ROW_ID && first != ROW_ADDR {
                 return Err(Error::Schema {
                     message: format!("Column {} does not exist", col.as_ref()),
                     location: location!(),
--- a/rust/lance/src/dataset/scanner.rs
+++ b/rust/lance/src/dataset/scanner.rs
@@ -459,12 +459,37 @@ impl Scanner {
     ///
     /// Only select the specified columns. If not specified, all columns will be scanned.
     pub fn project<T: AsRef<str>>(&mut self, columns: &[T]) -> Result<&mut Self> {
-        self.project_with_transform(
-            &columns
-                .iter()
-                .map(|c| (c.as_ref(), escape_column_name(c.as_ref())))
-                .collect::<Vec<_>>(),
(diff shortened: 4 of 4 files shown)


LuQQiu:
@wjones127 are we planning to expose row_id and row_address to end users? will they subject to change?

Jay-ju:
> @wjones127 are we planning to expose row_id and row_address to end users? will they subject to change?

@LuQQiu At the current stage, as I understand it, it can only be disclosed. It may not be necessary in the future (if there is a business key), because the screening of a lot of pre-training data is based on the set of row_id to identify which data needs to be selected for different datasets.

westonpace:
> @wjones127 are we planning to expose row_id and row_address to end users? will they subject to change?

We haven't exposed them in `lancedb` but in `lance` we expose both and assume the user is advanced enough to understand the risks.

rust/lance-core/src/datatypes/schema.rs
```diff
@@ -1081,8 +1085,9 @@ impl Projection {
 
     /// Convert the projection into an Arrow schema
     pub fn to_arrow_schema(&self) -> Result<arrow_schema::Schema> {
-        let mut arrow_schema: arrow_schema::Schema = (&self.to_schema()).into();
+        let arrow_schema: arrow_schema::Schema = (&self.to_schema()).into();
         // Should we be adding row_id / row_addr on to_schema?
+        let mut arrow_schema = arrow_schema.clone();
```
westonpace: What is this change for?

Jay-ju:
> I think this only works if `with_row_id` is also passed. For example, this would fail...
> 
> ```
> lance_ds.scanner(columns=["data_item_id", "_rowid"]).to_table()
> ```
> 
> When we see the column named `_rowid` do we want to automatically set `with_row_id` to true? (same with `_rowaddr`)


@westonpace Well, there should indeed be such a restriction.

rust/lance-core/src/datatypes/schema.rs
Jay-ju: It has been deleted here

Jay-ju:
@westonpace Could you help review it when you have time?

Jay-ju:
@westonpace @jackye1995 Do you have any questions regarding this PR? Could you please take a look at the changes here? Thank you first

python/python/tests/test_lance.py
```diff
@@ -248,3 +248,22 @@ def test_io_counters(tmp_path):
...
+            "data_item_id": pa.array([1001, 1002, 1003]),
+            "a": pa.array([1, 2, 3]),
+        }
+    )
+
+    lance.write_dataset(data, path)
+
+    lance_ds = lance.dataset(path)
+    table = lance_ds.scanner(
+        with_row_id=True, columns=["data_item_id", "_rowid"]
```
jackye1995: I think we should also test a negative case of not setting with_row_id and expect an error

python/python/tests/test_lance.py
```diff
@@ -248,3 +248,22 @@ def test_io_counters(tmp_path):
     dataset.to_table()
     assert lance.iops_counter() > starting_iops
     assert lance.bytes_read_counter() > starting_bytes
+
+
+def test_with_row_id(tmp_path):
```
jackye1995: we should also have a test of `test_with_row_addr`

python/src/dataset.rs
```diff
@@ -665,14 +674,6 @@ impl Dataset {
 
         scanner.scan_in_order(scan_in_order.unwrap_or(true));
 
-        if with_row_id.unwrap_or(false) {
```
jackye1995, line 679: this change is just moving the code up? Does it serve any functional purpose?

rust/lance/src/dataset/scanner.rs
```diff
@@ -474,9 +474,19 @@ impl Scanner {
         &mut self,
         columns: &[(impl AsRef<str>, impl AsRef<str>)],
     ) -> Result<&mut Self> {
+        let filtered_columns: Vec<_> = columns
```
jackye1995, line 502: we should have a test in `scanner.rs` to verify this behavior

Jay-ju: @jackye1995  Some test cases in rs have been added here

python/python/tests/test_lance.py
Jay-ju: add it

python/python/tests/test_lance.py
Jay-ju: add it

python/src/dataset.rs
Jay-ju: Yes, it does work. Originally, the retrieval of with_row_id was after the project. Now it has been moved up here.

```python
(Some(c), None) => {
                scanner
                    .project(&c)
                    .map_err(|err| PyValueError::new_err(err.to_string()))?;
            }
            (None, Some(ct)) => {
                scanner
                    .project_with_transform(&ct)
                    .map_err(|err| PyValueError::new_err(err.to_string()))?;
            }
``` 

In front of this code, if `with_row_id=True`, some `_rowid` will be automatically added to the project.


@jackye1995 Please help review it when you have time. It has been modified according to the comment.

rust/lance/src/dataset/scanner.rs
```diff
@@ -6393,4 +6431,78 @@ mod test {
...
+        let mut scanner = dataset.scan();
+        scanner.with_row_id = false;
+        scanner.with_row_address = false;
+        let result = scanner.project(&[if with_row_id { ROW_ID } else { ROW_ADDR }]);
+        assert!(result.is_err());
+    }
+
+    #[rstest]
+    #[tokio::test]
+    async fn test_with_row_id() {
```
jackye1995: nit: it feels unnecessary to test this independently with a helper function. Can we just have one test `test_with_row_id_and_address` that test the 3 cases in 1 pass?

python/python/tests/test_lance.py
```diff
@@ -248,3 +248,34 @@ def test_io_counters(tmp_path):
     dataset.to_table()
     assert lance.iops_counter() > starting_iops
     assert lance.bytes_read_counter() > starting_bytes
+
+
+@pytest.mark.parametrize(
+    "row_param, column_name",
+    [("with_row_id", "_rowid"), ("with_row_address", "_rowaddr")],
+)
+def test_with_row_id(tmp_path, row_param, column_name):
```
jackye1995, line 257: nit: `test_with_row_id_and_address`, given we are now testing both

rust/lance/src/dataset/scanner.rs
Jay-ju: fix it

python/python/tests/test_lance.py
Jay-ju: add it

Jay-ju: @jackye1995 Please help take a look when you have time. I've been a bit busy this week, so I'm a bit late. Sorry about that.

jackye1995 (approved):
looks good to me

jackye1995:
CI failure is unrelated, merging.

Issue #3950, opened 2025-06-06
Epic: Improving integration with Java ecosystem


majin1102:
As metioned in #3742 This is the epic issue for collecting features/enhancements to improve Java ecosystem integration.

I  will help to collect related issues/PR after 2025.06.06 and update the issue list periodically. Everyone is welcome to add relevant issues here. I'll also periodically update this issue list.

Updating this issue is for convenience and good, not necessary and should not block any contribution work

## Meaningful issues
* #4201
* #4202
* #3944
* #3957
* #4050

## Java api
* #3907
* #3945
* #3958 
* #3969 
* #4039 
* #4207 
* #4119
* #4328

## Transaction
* #4219
* #4327

## Merge insert(upsert)
* #4172

majin1102:
For now the Lance java binding is quite catched up and I think this epic is good to close

PR #4119, opened 2025-07-02
feat(java): support replace schema and field metadata

majin1102:
epic: #3950

### DIFF
--- a/java/core/src/main/java/com/lancedb/lance/Dataset.java
+++ b/java/core/src/main/java/com/lancedb/lance/Dataset.java
@@ -772,6 +772,34 @@ public class Dataset implements Closeable {
     return new Tags();
   }
 
+  /**
+   * Replace the schema metadata of the dataset.
+   *
+   * @param metadata the new table metadata
+   */
+  public void replaceSchemaMetadata(Map<String, String> metadata) {
+    try (LockManager.WriteLock writeLock = lockManager.acquireWriteLock()) {
+      Preconditions.checkArgument(nativeDatasetHandle != 0, "Dataset is closed");
+      nativeReplaceSchemaMetadata(metadata);
+    }
+  }
+
+  private native void nativeReplaceSchemaMetadata(Map<String, String> metadata);
+
+  /**
+   * Replace target field metadata of the dataset. This method won't affect fields not in the map
+   *
+   * @param fieldMetadataMap field name to metadata map
+   */
+  public void replaceFieldMetadata(Map<String, Map<String, String>> fieldMetadataMap) {
+    try (LockManager.WriteLock writeLock = lockManager.acquireWriteLock()) {
+      Preconditions.checkArgument(nativeDatasetHandle != 0, "Dataset is closed");
+      nativeReplaceFieldMetadata(fieldMetadataMap);
+    }
+  }
+
+  private native void nativeReplaceFieldMetadata(Map<String, Map<String, String>> fieldMetadataMap);
+
   /** Tag operations of the dataset. */
   public class Tags {
 
--- a/java/core/src/test/java/com/lancedb/lance/DatasetTest.java
+++ b/java/core/src/test/java/com/lancedb/lance/DatasetTest.java
@@ -836,4 +836,65 @@ public class DatasetTest {
       assertFalse(currentConfig.containsKey("key1"));
     }
   }
+
+  @Test
+  void testReplaceSchemaMetadata() {
+    String testMethodName = new Object() {}.getClass().getEnclosingMethod().getName();
+    String datasetPath = tempDir.resolve(testMethodName).toString();
+    try (RootAllocator allocator = new RootAllocator(Long.MAX_VALUE)) {
+      TestUtils.SimpleTestDataset testDataset =
+          new TestUtils.SimpleTestDataset(allocator, datasetPath);
+      dataset = testDataset.createEmptyDataset();
+      assertEquals(1, dataset.version());
+      Map<String, String> replaceMetadata = new HashMap<>();
+      replaceMetadata.put("key1", "value1");
+      replaceMetadata.put("key2", "value2");
+      dataset.replaceSchemaMetadata(replaceMetadata);
+      assertEquals(2, dataset.version());
+      Map<String, String> currentMetadata = dataset.getSchema().getCustomMetadata();
+      for (String configKey : currentMetadata.keySet()) {
+        assertEquals(currentMetadata.get(configKey), replaceMetadata.get(configKey));
+      }
+      assertEquals(replaceMetadata.size(), currentMetadata.size());
+
+      Map<String, String> replaceConfig2 = new HashMap<>();
+      replaceConfig2.put("key1", "value3");
+      dataset.replaceSchemaMetadata(replaceConfig2);
+      currentMetadata = dataset.getSchema().getCustomMetadata();
+      assertEquals(3, dataset.version());
+      assertEquals(1, currentMetadata.size());
+      assertEquals("value3", currentMetadata.get("key1"));
+    }
+  }
+
+  @Test
+  void testReplaceFieldConfig() {
+    String testMethodName = new Object() {}.getClass().getEnclosingMethod().getName();
+    String datasetPath = tempDir.resolve(testMethodName).toString();
+    try (RootAllocator allocator = new RootAllocator(Long.MAX_VALUE)) {
+      TestUtils.SimpleTestDataset testDataset =
+          new TestUtils.SimpleTestDataset(allocator, datasetPath);
+      dataset = testDataset.createEmptyDataset();
+      assertEquals(1, dataset.version());
+      Field field = dataset.getSchema().getFields().get(0);
+      Map<String, String> replaceMetadata = new HashMap<>();
+      replaceMetadata.put("key1", "value1");
+      replaceMetadata.put("key2", "value2");
+      dataset.replaceFieldMetadata(Collections.singletonMap(field.getName(), replaceMetadata));
+      assertEquals(2, dataset.version());
+      Map<String, String> currentMetadata = dataset.getSchema().getFields().get(0).getMetadata();
+      for (String configKey : currentMetadata.keySet()) {
+        assertEquals(currentMetadata.get(configKey), replaceMetadata.get(configKey));
+      }
+      assertEquals(replaceMetadata.size(), currentMetadata.size());
+
+      Map<String, String> replaceConfig2 = new HashMap<>();
+      replaceConfig2.put("key1", "value3");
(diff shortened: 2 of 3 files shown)


java/core/src/main/java/com/lancedb/lance/Dataset.java
```diff
@@ -772,6 +772,34 @@ public Tags tags() {
...
+  }
+
+  private native void nativeReplaceSchemaMetadata(Map<String, String> metadata);
+
+  /**
+   * Replace target field metadata of the dataset. This method won't affect fields not in the map
+   *
+   * @param fieldMetadataMap field name to metadata map
+   */
+  public void replaceFieldMetadata(Map<String, Map<String, String>> fieldMetadataMap) {
```
eddyxu: Does this work with nested fields?

majin1102: Ahhh..
Now I understand why use field id as the key. Thanks for pointing out this miss

How about use a.b.c...to express nested fields?  (in Spark secenarios  this approach would be more friendly to config compression).
Or after introducing lance schema to Java we still use field ids as the key, Which do you prefer?
@eddyxu @yanghua @jackye1995

java/core/src/main/java/com/lancedb/lance/Dataset.java
```diff
@@ -772,6 +772,34 @@ public Tags tags() {
     return new Tags();
   }
 
+  /**
+   * Replace the schema metadata of the dataset.
+   *
+   * @param metadata the new table metadata
+   */
+  public void replaceSchemaMetadata(Map<String, String> metadata) {
```
yanghua, line 795: Does `updateSchemaMetadata` look better than `replaceXXX`?

majin1102: I think replace and update are different meanings. Let's say the original map is
A-> V1;
B-> V2;

Replace it with:
C-> V3

The result would only contain C. While update means A, B, upsert means A, B, C

majin1102: Additionally
1. rust api uses replace sematics
2. if use update, delete interface would be necessary like deleleConfigKeys

yanghua: After checking Rust's implementation, I think your thought is reasonable. The arg of `replace_schema_metadata` in Rust is `upsert_values`, which may be misunderstood by users.

yanghua:
```
Operation::UpdateConfig {
            upsert_values: None,
            delete_keys: None,
            schema_metadata: Some(HashMap::from_iter(upsert_values)),
            field_metadata: None,
        }
```

Upsert and replace are distinguished here.

So the `replace` is reasonable.

java/core/src/main/java/com/lancedb/lance/Dataset.java
yanghua: > How about use a.b.c...to express nested fields? (in Spark secenarios this approach would be more friendly to config compression).

Not sure if there is another `delimiter` besides `.`. If it exists, we may need to specify it.

Or in the Java module, we follow native rules. Spark module is one of the users of the Java module.

java/core/src/main/java/com/lancedb/lance/Dataset.java
majin1102: > After checking Rust's implementation, I think your thought is reasonable. The arg of `replace_schema_metadata` in Rust is `upsert_values`, which may be misunderstood by users.

I‘ve also seen that.
I will submit a PR to improve that

java/core/src/main/java/com/lancedb/lance/Dataset.java
majin1102: Use the field id after #4202

majin1102:
@jackye1995 PTAL when you have time

java/core/lance-jni/Cargo.lock
jackye1995, line 1: this change seems unintended?

java/core/src/test/java/com/lancedb/lance/DatasetTest.java
```diff
@@ -848,4 +849,65 @@ void testGetLanceSchema() {
       assertEquals(testDataset.getSchema(), dataset.getLanceSchema().asArrowSchema());
     }
   }
+
+  @Test
+  void testReplaceSchemaMetadata() {
+    String testMethodName = new Object() {}.getClass().getEnclosingMethod().getName();
```
jackye1995, line 854: can just use `void testXXX(@TempDir Path tempDir) { ...} `

java/core/src/test/java/com/lancedb/lance/DatasetTest.java
```diff
@@ -848,4 +849,65 @@ void testGetLanceSchema() {
...
+      replaceMetadata.put("key2", "value2");
+      dataset.replaceFieldMetadata(Collections.singletonMap(field.getId(), replaceMetadata));
+      assertEquals(2, dataset.version());
+      Map<String, String> currentMetadata = dataset.getSchema().getFields().get(0).getMetadata();
+      for (String configKey : currentMetadata.keySet()) {
+        assertEquals(currentMetadata.get(configKey), replaceMetadata.get(configKey));
+      }
+      assertEquals(replaceMetadata.size(), currentMetadata.size());
+
+      Map<String, String> replaceConfig2 = new HashMap<>();
```
jackye1995, line 903: should also test behavior when trying to update a field ID that does not exist

java/core/lance-jni/Cargo.lock
majin1102: Yes, that's a mistake

java/core/src/test/java/com/lancedb/lance/DatasetTest.java
majin1102: Yes.

I think that would be better.

Do we need to use this pattern just in this PR or raise another PR to change the whole DatasetTest?  Or just modify the whole DatasetTest in this PR. I used to keep PRs small, but not preferable.

java/core/src/test/java/com/lancedb/lance/DatasetTest.java
majin1102: Copy

majin1102:
@jackye1995 This is ready for review again. 2 TODOs left:

1. Use the (@TempDir Path tempDir) through DatasetTest
2. Replace field id which not exists doesn't throw error(I think better change rust dataset behaviour)

jackye1995:
Agree we should change the behavior. Do you want to do it as a part of this PR?

majin1102:
> Agree we should change the behavior. Do you want to do it as a part of this PR?

Sure

majin1102:
> Agree we should change the behavior. Do you want to do it as a part of this PR?

https://github.com/lancedb/lance/blob/main/rust/lance-table/src/format/manifest.rs

I followed the code path, there's a comment on line 216:

image

I believe there's context @westonpace knows. It seems designed as this

westonpace:
> I believe there's context @westonpace knows. It seems designed as this

I don't recall why I wrote it that way.  I can't think of any reason not to error.  In fact, it looks like we error on missing field higher up in the python layer:

```
    fn replace_field_metadata(
        &mut self,
        field_name: &str,
        metadata: HashMap<String, String>,
    ) -> PyResult<()> {
        let mut new_self = self.ds.as_ref().clone();
        let field = new_self
            .schema()
            .field(field_name)
            .ok_or_else(|| PyKeyError::new_err(format!("Field \"{}\" not found", field_name)))?;
```

I'm +1 on changing rust behavior if we want to error

rust/lance-table/src/format/manifest.rs
```diff
@@ -214,9 +214,19 @@ impl Manifest {
...
+        &mut self,
+        field_id: i32,
+        new_metadata: HashMap<String, String>,
+    ) -> Result<()> {
         if let Some(field) = self.schema.field_by_id_mut(field_id) {
             field.metadata = new_metadata;
+            Ok(())
+        } else {
+            Err(Error::invalid_input(
+                format!("field with id {} for replaceFieldMetadata", field_id),
```
jackye1995: the error message seems not clear, can we use "Field with id {} does not exist"?

java/core/src/test/java/com/lancedb/lance/DatasetTest.java
jackye1995: Added an issue tracking this, we can do it separately

jackye1995:
Mostly looks good to me, just a nit

rust/lance-table/src/format/manifest.rs
majin1102: The actual message is like :

java.lang.IllegalArgumentException: Invalid user input: field with id 2147483647 for replaceFieldMetadata, /Users/Nathan/workspace/lance/rust/lance-table/src/format/manifest.rs:228:17

Because we used an invaild_input function, the message has been formatted. I don't find a better error type for this case. I can construct a formatted error type for this case if you don't feel heavy?

jackye1995: The main reason I was asking that is because it does not say why the field is is invalid, which is because it does not exist in the schema

jackye1995: Also I don't quite understand why we want to add "for replaceFieldMetadata" in the message, that is the java method name you are putting in the rust.

majin1102: Yeah....I guessed this was a bad inference by code assistance and I didn't notice

majin1102:
> Mostly looks good to me, just a nit

please take a look when you have time.

btw, I usually see the linux-build(nightly) ci failed like this, is there any context here?
image

rust/lance-table/src/format/manifest.rs
```diff
@@ -214,9 +214,19 @@ impl Manifest {
...
+    pub fn replace_field_metadata(
+        &mut self,
+        field_id: i32,
+        new_metadata: HashMap<String, String>,
+    ) -> Result<()> {
         if let Some(field) = self.schema.field_by_id_mut(field_id) {
             field.metadata = new_metadata;
+            Ok(())
+        } else {
+            Err(Error::FieldNotExists {
```
jackye1995: sorry for the back and forth, I did not describe clearly. I think it is not worth creating a dedicated error for this, this is an invalid input of the replace_field_metadata function, so the original choice was correct. But we just need to have a more clear message like the new one you have - "Field does not exists: {field_id}"

jackye1995:
> btw, I usually see the linux-build(nightly) ci failed like this, is there any context here?

taking a look

rust/lance-table/src/format/manifest.rs
majin1102: my bad

Issue #2067, opened 2024-03-13
Trying to open non-existent dataset should return Error::NotFound instead of Error::IO


chebbyChefNEQ:
we have a blanket conversion of `object_store::Error` -> `Error::IO`.

PR #4126, opened 2025-07-02
added match pattern to convert to DatasetNotFound when appropriate rather blanket converting to IO Errors

saresend:
Aims to address #2067

### DIFF
--- a/rust/lance-core/src/error.rs
+++ b/rust/lance-core/src/error.rs
@@ -225,9 +225,19 @@ impl From<std::io::Error> for Error {
 impl From<object_store::Error> for Error {
     #[track_caller]
     fn from(e: object_store::Error) -> Self {
-        Self::IO {
-            source: box_error(e),
-            location: std::panic::Location::caller().to_snafu_location(),
+        match e {
+            object_store::Error::NotFound {
+                ref path,
+                source: _,
+            } => Self::DatasetNotFound {
+                path: path.clone(),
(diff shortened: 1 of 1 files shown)


rust/lance-core/src/error.rs
```diff
@@ -225,9 +225,19 @@ impl From<std::io::Error> for Error {
 impl From<object_store::Error> for Error {
     #[track_caller]
     fn from(e: object_store::Error) -> Self {
-        Self::IO {
-            source: box_error(e),
-            location: std::panic::Location::caller().to_snafu_location(),
+        match e {
+            object_store::Error::NotFound {
```
Xuanwo, line 229: Hi, I'm afraid this isn't a good fix because the `NotFound` error is so general that there could be many different reasons for encountering it. Converting it to `DatasetNotFound` may cause confusion.

saresend: Sure - do you have an understanding of what categories the `object_store::Error::NotFound` covers? My impression is that this covers essentially cases where the path passed doesn't point to any actual file, which is more or less the semantics of `DatasetNotFound`. What are the other cases?

Xuanwo: Hi, it's quite a common error to encounter a `not found` error. For example, users might remove a file from S3, resulting in a `not found` error. In any case, I don't think we should handle this error in this manner.

PR #4127, opened 2025-07-02
feat: use readthedocs to host pylance generated docs

### LINKED ISSUE #4055: Move pylance API documentation to readthedocs
Today any documentation change requires rebuilding pylance (which essentially builds all rust modules), because today the pylance documentation is generated together in the lance documentation website. 

I think we could publish pylance documentation independently to readthedocs which hosts most of the pypi project documentation, just like the rust documentation is published to docs.rs. (We should probably also do that for the Java to javadoc.io)

By making this change, Lance documentation is just pure text and can be built much faster. This would also simplify the migration of the doc site to mkdocs (#3871).

What do we think?

jackye1995:
Closes #4055

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -1,0 +1,28 @@
+# Read the Docs configuration file
+# See https://docs.readthedocs.io/en/stable/config-file/v2.html for details
+
+# Required
+version: 2
+
+# Set the OS, Python version, and other tools you might need
+build:
+  os: ubuntu-24.04
+  tools:
+    python: "3.13"
+    rust: "1.86"
+  jobs:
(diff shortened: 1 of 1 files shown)


Xuanwo (approved):

python/docs/dataset.rst
```diff
@@ -0,0 +1,8 @@
+Lance Dataset
```
Xuanwo, line 1: So we will migrate to rtd first and then migrating to mardown, right?

jackye1995: this will just be in rtd, it's just easier navigation for the generated doc so I kept it there. In general in the main Lance doc my expectation is that we put multi-language blocks for all our code examples

PR #4128, opened 2025-07-03
fix: panic message when LanceFileWriter is not closed and then got garbage collected

wojiaodoubao:
When I use LanceFileWriter and failed to close it, then the writer will panic when it got garbage collected. The program won't crash, but it leaves a panic error message.

I'll use a sample to illustrate it.
```
import lance.file as lf
import pyarrow as pa
import gc

OUT_PATH = '/tmp/output.lance'


def file_writer_gc():
    writer = lf.LanceFileWriter(OUT_PATH)
    records = []
    for i in range(1_000_000):
        records.append({"name": f"name-{i}", "age": i})
    table = pa.Table.from_pylist(records)
    writer.write_batch(table)

    print('Some error occurs and fail to close writer, then writer got garbage collected.')
    del writer
    gc.collect()
    print('Garbage collect successfully.')


if __name__ == '__main__':
    file_writer_gc()
    print('Sample done')

```

We'll get an error message in output.
```
Some error occurs, failed to close writer, then writer got garbage collected.

thread '<unnamed>' panicked at /data00/home/lijinglun/Project/lance-community/rust/lance-io/src/object_writer.rs:330:17:
there is no reactor running, must be called from the context of a Tokio 1.x runtime
  File "/home/lijinglun/tmp/filewriter_panic.py", line 17, in file_writer_gc
    del writer
        ^^^^^^
pyo3_runtime.PanicException: there is no reactor running, must be called from the context of a Tokio 1.x runtime
Garbage collect successfully.
Sample done
```


rust/lance-io/src/object_writer.rs
```diff
@@ -327,9 +332,14 @@ impl Drop for ObjectWriter {
...
             if let UploadState::InProgress { mut upload, .. } = state {
-                tokio::task::spawn(async move {
+                let abort = async move {
                     let _ = upload.abort().await;
-                });
+                };
+                if let Ok(handle) = Handle::try_current() {
+                    handle.spawn(abort);
+                } else {
+                    futures::executor::block_on(abort);
```
Xuanwo: Calling `futures::executor::block_on` can result in tokio hang while many drop on `ObjectWriter` are running.

I suggest we handling this part this way:

- Add new API `async fn abort()` for ObjectWriter.
- Only call `abort` while runtime is alive (Handle::try_current())
- At the python side, call `abort()` during drop in `RT.runtime.block_on()`

wojiaodoubao: Thanks a lot for your professional guidance. Let me try the abort way.

wojiaodoubao:
Hi @Xuanwo , could you help take a look, thanks very much.

rust/lance-io/src/object_writer.rs
```diff
@@ -152,6 +154,10 @@ impl UploadState {
             _ => unreachable!(),
         };
     }
+
+    fn any_to_abort(&mut self) -> Self {
```
Xuanwo: Hi, I have a feeling that we don't need to have this?

rust/lance-io/src/object_writer.rs
```diff
@@ -317,19 +323,36 @@ impl ObjectWriter {
...
         }
     }
+
+    pub async fn abort(&mut self) {
+        if let Some(abort_task) = self.inner_abort() {
+            abort_task.await;
+        }
+    }
+
+    fn inner_abort(&mut self) -> Option<BoxFuture<'static, ()>> {
```
Xuanwo: Maybe we can just implment in `abort`?

rust/lance-io/src/object_writer.rs
```diff
@@ -317,19 +323,36 @@ impl ObjectWriter {
...
-        if matches!(self.state, UploadState::InProgress { .. }) {
-            // Take ownership of the state.
-            let state =
-                std::mem::replace(&mut self.state, UploadState::Done(WriteResult::default()));
-            if let UploadState::InProgress { mut upload, .. } = state {
-                tokio::task::spawn(async move {
-                    let _ = upload.abort().await;
+        if let Some(abort_task) = self.inner_abort() {
+            if let Ok(handle) = Handle::try_current() {
+                handle.spawn(async move {
```
Xuanwo: How about calling `self.abort()` here directly?

rust/lance-io/src/object_writer.rs
```diff
@@ -317,19 +323,36 @@ impl ObjectWriter {
...
+                    abort_task.await;
                 });
+            } else {
+                warn!("Failed to abort object writer because runtime not found.");
```
Xuanwo: I don't think we need to warn here. The cleanup logic is best effort and we can just ignore it if failed.

python/src/file.rs
```diff
@@ -320,6 +320,11 @@ impl LanceFileWriter {
         self.inner_lock()?.add_schema_metadata(key, value);
         Ok(())
     }
+
+    pub fn abort(&self) -> PyResult<()> {
```
Xuanwo: We can call `abort` in `Drop` at the `RT.runtime`.

Xuanwo (commented):
Thank you, we are almost here.

python/src/file.rs
wojiaodoubao: Got it. Do you mean we should use Drop instead of an explicit abort method? It makes sense to me, let me fix it.

rust/lance-io/src/object_writer.rs
wojiaodoubao: I found this ```impl UploadState``` block is used for '/// Methods for state transitions'. So I think we can put all state transitions together, including 'from any state to abort', for better readability.

Maybe the method name 'any_to_abort' is not clear? Would 'force_to_abort' be better?

 What do you think?

rust/lance-io/src/object_writer.rs
wojiaodoubao: Makes sense to me, let me fix it.

rust/lance-io/src/object_writer.rs
wojiaodoubao: Call self.abort() directly will got a lifetime error because self escapes the method body. So I make ```inner_abort``` to generate an owned object as a bridge.
```
error[E0521]: borrowed data escapes outside of method
   --> /data00/home/lijinglun/Project/lance-community/rust/lance-io/src/object_writer.rs:351:17
    |
348 |       fn drop(&mut self) {
    |               ---------
    |               |
    |               `self` is a reference that is only valid in the method body
    |               let's call the lifetime of this reference `'1`
...
351 | /                 handle.spawn(async move {
352 | |                     self.abort().await;
353 | |                     // abort_task.await;
354 | |                 });
    | |                  ^
    | |                  |
    | |__________________`self` escapes the method body here
    |                    argument requires that `'1` must outlive `'static`
```

wojiaodoubao:
Hi @Xuanwo , sorry to bother. Could you help taking a look when you have time, thanks very much.

PR #4131, opened 2025-07-03
feat: expose max_page_bytes arg for python LanceFileWriter

wojiaodoubao:
We can expose max_page_bytes for python LanceFileWriter.


python/python/lance/file.py
```diff
@@ -201,6 +201,7 @@ def __init__(
         data_cache_bytes: Optional[int] = None,
         version: Optional[str] = None,
         storage_options: Optional[Dict[str, str]] = None,
+        max_page_bytes: Optional[int] = None,
```
yanghua, line 204: We may miss a doc for this arg?

python/python/lance/file.py
```diff
@@ -225,6 +226,9 @@ def __init__(
         storage_options : optional, dict
             Extra options to be used for a particular storage connection. This is
             used to store connection parameters like credentials, endpoint, etc.
+        max_page_bytes : optional, int
```
yanghua, line 229: It would be better to point out what the default value is.

yanghua (approved):
LGTM

Issue #4134, opened 2025-07-03
Python CI out of disk space


jackye1995:
https://github.com/lancedb/lance/actions/runs/16062288527/job/45330498233?pr=4132

codephage2020:
It is possible that we opt not to cache the 'target' directory of the workspace. 
Add the following attributes: `cache-targets: "false"`; 
refer [here](https://github.com/Swatinem/rust-cache?tab=readme-ov-file#:~:text=%23%20Determines%20whether%20workspace%20%60target%60%20directories%20are%20cached.%0A%20%20%20%20%23%20If%20%60false%60%2C%20only%20the%20cargo%20registry%20will%20be%20cached.%0A%20%20%20%20%23%20default%3A%20%22true%22%0A%20%20%20%20cache%2Dtargets%3A%20%22%22).

Xuanwo:
```
 # Determines whether workspace `target` directories are cached.
    # If `false`, only the cargo registry will be cached.
    # default: "true"
    cache-targets: ""
```

Since the sparse index is used by default, there's no need to cache the index. Disabling cache targets is essentially the same as disabling the Rust cache action altogether. I think we can simply enable `cache-workspace-crates` and avoid caching the built artifacts.

Xuanwo:
The other idea I have is to split lance-jni out of the workspace like we do for pylance.

jackye1995:
Good point about lance-jni. It seems fine to remove? I can give a try

PR #4139, opened 2025-07-04
ci: polish workflow by reduce cache size and remove not needed actions

Xuanwo:
close https://github.com/lancedb/lance/issues/4134


.github/workflows/java.yml
```diff
@@ -39,10 +39,6 @@ jobs:
         run: |
           sudo apt update
           sudo apt install -y protobuf-compiler libssl-dev
-      # pin the toolchain version to avoid surprises
```
jackye1995, line 42: Would there be any side effect of this removal?

jackye1995 (approved):
apart from the question, looks good to me

.github/workflows/java.yml
westonpace: If we don't run `setup-rust-toolchain` then it will use whatever happens to be on the runner.  The main annoyance is when changes are made to rustfmt / clippy.  If our dev machines aren't in sync with our CI machines, or our CI machines are not in sync with each other, it can be difficult to get things to pass.

.github/workflows/rust.yml
```diff
@@ -85,18 +84,19 @@ jobs:
...
       # pin the toolchain version to avoid surprises
-      - uses: actions-rust-lang/setup-rust-toolchain@v1
-        with:
-          toolchain: ${{ matrix.toolchain }}
-          # When using the nightly toolchain, do not surface warnings as check annotations
-          matcher: ${{ matrix.toolchain != 'nightly'}}
+      - name: Setup rust toolchain
+        run: |
+          rustup toolchain install ${{ matrix.toolchain }}
+          rustup default ${{ matrix.toolchain }}
```
westonpace, line 90: Why is this better than `setup-rust-toolchain`?

westonpace (commented):
I think `cache-targets: false` and `cache-workspace-crates: false` is good.  But why remove `setup-rust-toolchain`?

.github/workflows/java.yml
Xuanwo: > If we don't run `setup-rust-toolchain` then it will use whatever happens to be on the runner.

Since we already have a `rust-toolchain.toml` file and I have added `clippy` and `fmt` to it:

https://github.com/lancedb/lance/pull/4139/files#diff-2b1bde2cf3a858b7bf7424cb8bcbf01f35b94dc80b925d9432cbab3319ca9b4eL2-L4

`rustup` will ensure that we use version 1.86's clippy and rustfmt, matching exactly what we use in development.

I believe this action is no longer necessary now that `rust-toolchain.toml` has been stabilized.

.github/workflows/rust.yml
Xuanwo: `setup-rust-toolchain` didn't do much beyond what `rustup` plus `rust-cache` already provide. I find it's clearer to see what's happening by calling `rustup` and using `rust-cache` directly.

PR #4140, opened 2025-07-04
chore: remove dead cargo.lock files

Xuanwo:
I personally feel those files are not used anywhere anymore.


jackye1995 (approved):
makes sense

westonpace:
The `Cargo.lock` files are used to hold dependent versions constant so that the GHA cache works more consistently.  Without them, the concern is that the CI jobs will take longer as the cache is more likely to be out of date.  I don't have any hard data though.

jackye1995:
With #4146 at least I think we can remove the jni one since the Cargo.lock will be collocated with the Cargo.toml.

westonpace:
Oh!  I see.  These are in child directories.

westonpace (approved):

PR #4142, opened 2025-07-04
refactor: refactor full zip schedule_ranges_rep

Xuanwo:
Follow-up of https://github.com/lancedb/lance/pull/4104, make the code more clear.

This PR is based on https://github.com/lancedb/lance/pull/4141, should be merged after that.


rust/lance-encoding/src/encodings/logical/primitive.rs
```diff
@@ -1628,7 +1628,7 @@ impl StructuralPageScheduler for MiniBlockScheduler {
     }
 }
 
-#[derive(Debug)]
+#[derive(Debug, Clone, Copy)]
```
BubbleCal, line 1631: `Copy` implies `Clone`

Xuanwo: Yep, IDE will always suggest `Clone` and `Copy` at the same time 🤣

PR #4143, opened 2025-07-04
feat(rust): support object store local mode for take bench

yanghua:


rust/lance/benches/take.rs
```diff
@@ -27,6 +27,8 @@ use std::sync::Arc;
 #[cfg(target_os = "linux")]
 use std::time::Duration;
 
+const ENV_OBJECT_STORE_SCHEME: &str = "OBJECT_STORE_SCHEME";
```
yanghua, line 30: There are two options to support object store local mode:

* Option 1: choose one type of object store(`memory` or `local`, default is memory) in one bench run, the code is simple;
* Option 2: support two object store types in one bench run, it needs one more loop(it seems `criterion` is not friendly for parameterize) before creating the  dataset;

Currently, chose option 1.

jackye1995: I am wondering if there is a good way to combine the 2 options. Not sure if it would be useful to you, but for us it's definitely useful to compare for example S3 vs Azure vs GCS performance using the same code.

I think the core issue is that we are doing inner loops after inner loops, instead of having a flat function that produces one execution, and then allocate different variables at the top.

For example, the ideal `bench_random_single_take_with_file_reader` should be

```rust
fn bench_random_single_take_with_file_reader(
  c: &mut Criterion, 
  version: LanceFileVersion, 
  num_batches: u64,
  file_size: u64,
  file_scheme: &str,
  num_rows_per_take: u64) {
...
}
```

and then we allocate all the variables that need to be run all together through some macro so that we can avoid the nested loops:

```
macro_rules! define_bench {
    ($name:ident, $version:expr, $nb:expr, $fs:expr, $scheme:expr, $take:expr) => {
        fn $name(c: &mut Criterion) {
            bench_random_single_take_with_file_reader(c, $version, $nb, $fs, $scheme, $take)
        }
    };
}

define_bench!(bench_case_1, LanceFileVersion::V1, 10, 100_000, "file", 1000);
define_bench!(bench_case_2, LanceFileVersion::V2, 1, 1_000_000, "memory", 1);
```

what do you think?

yanghua: Let me have a try.

PR #4145, opened 2025-07-04
feat(rust): support trace dataset events

yanghua:


yanghua:
Still WIP, not ready for review.

jackye1995:
let me know when this is ready 😄

rust/lance-core/src/utils/tracing.rs
```diff
@@ -75,4 +75,11 @@ pub const IO_TYPE_OPEN_MEM_WAL: &str = "open_mem_wal_index";
 pub const IO_TYPE_LOAD_VECTOR_PART: &str = "load_vector_part";
 pub const IO_TYPE_LOAD_SCALAR_PART: &str = "load_scalar_part";
 pub const TRACE_EXECUTION: &str = "lance::execution";
+pub const TRACE_DATASET_EVENTS: &str = "lance::dataset_events";
+pub const DATASET_EVENT_CREATE_OR_WRITE: &str = "dataset_create_or_write";
```
yanghua: We may track `fragment` and `data file` related events in the future. So I added `dataset_` prefix.

yanghua:
@jackye1995 Please take a look.

rust/lance-core/src/utils/tracing.rs
jackye1995: but you can always know the event is for a dataset from the `target = lance::dataset_events`?

rust/lance-core/src/utils/tracing.rs
```diff
@@ -75,4 +75,11 @@ pub const IO_TYPE_OPEN_MEM_WAL: &str = "open_mem_wal_index";
 pub const IO_TYPE_LOAD_VECTOR_PART: &str = "load_vector_part";
 pub const IO_TYPE_LOAD_SCALAR_PART: &str = "load_scalar_part";
 pub const TRACE_EXECUTION: &str = "lance::execution";
+pub const TRACE_DATASET_EVENTS: &str = "lance::dataset_events";
```
jackye1995: I think all these new contents need to be after `EXECUTION_PLAN_RUN`, so that it is after everything about `TRACE_EXECUTION`

rust/lance/src/dataset.rs
```diff
@@ -528,9 +533,16 @@ impl Dataset {
     ///
     pub async fn write(
         batches: impl RecordBatchReader + Send + 'static,
-        dest: impl Into<WriteDestination<'_>>,
+        dest: impl Into<WriteDestination<'_>> + Clone,
         params: Option<WriteParams>,
     ) -> Result<Self> {
+        if let Some(dataset) = dest.clone().into().dataset() {
```
jackye1995: it it possible for us to separate the create and write events? by checking if the stream is empty.

rust/lance/src/dataset.rs
```diff
@@ -528,9 +533,16 @@ impl Dataset {
...
+            info!(target: TRACE_DATASET_EVENTS, event=DATASET_EVENT_CREATE_OR_WRITE, uri=dataset.uri());
```
jackye1995: I think it's also important to trace the write mode, what do you think?

rust/lance-core/src/utils/tracing.rs
```diff
@@ -75,4 +75,11 @@ pub const IO_TYPE_OPEN_MEM_WAL: &str = "open_mem_wal_index";
...
+pub const DATASET_EVENT_CREATE_OR_WRITE: &str = "dataset_create_or_write";
+pub const DATASET_EVENT_OPEN: &str = "dataset_open";
+pub const DATASET_EVENT_DROP_COLUMN: &str = "dataset_drop_column";
+pub const DATASET_EVENT_DELETE_ROW: &str = "dataset_delete_row";
```
jackye1995: nit: no need for `_row`, delete already means delete rows

rust/lance/src/dataset/write/commit.rs
```diff
@@ -23,6 +14,14 @@ use crate::{
     session::Session,
     Dataset, Error, Result,
 };
+use lance_core::utils::mask::RowIdTreeMap;
```
jackye1995: looks like there are quite a few changes in import order without any actual changes across files, are these unintentional?

rust/lance/src/dataset.rs
yanghua: If the stream is empty, it represents a creation. However, if the stream is not empty, the semantics may not be clear. It could be a combination of creation, writing, or pure writing.

If we want to distinguish them precisely. We may need to push down this judgment into `CommitBuilder#execute`. There are two questions if we trace in `CommitBuilder#execute`:

* Not sure if the trace would contain events that come from the fragments.
* semantic: we want to catch a motivation(triggered but may not be successful, that means a `before` event) or a result(that means a `after` event); I prefer to catch a `before` event, especially for **open event**;

rust/lance/src/dataset.rs
yanghua: agree, if we can.

rust/lance/src/dataset/write/commit.rs
yanghua: yes, I found it, that's because I tried tracing in commit, but I changed the decision. It's strange, the format and code style check not recognize it.

rust/lance/src/dataset.rs
jackye1995: I see, thanks for the explanation. In that case I think we should just say this is just a `dataset_write` event. And in `CommitBuilder.execute`, we can track a `new_version` event that tracks the new version created and the associated operation type. If the new version is 0, it means it is a create event. We will also be able to track a detached version, or know if the dataset version creation actually succeeded or not. Seems in general beneficial.

rust/lance-core/src/utils/tracing.rs
```diff
@@ -76,3 +76,12 @@ pub const IO_TYPE_LOAD_VECTOR_PART: &str = "load_vector_part";
 pub const IO_TYPE_LOAD_SCALAR_PART: &str = "load_scalar_part";
 pub const TRACE_EXECUTION: &str = "lance::execution";
 pub const EXECUTION_PLAN_RUN: &str = "plan_run";
+pub const TRACE_DATASET_EVENTS: &str = "lance::dataset_events";
+pub const CREATE_EVENT: &str = "create";
+pub const WRITE_EVENT: &str = "write";
+pub const NEW_VERSION_EVENT: &str = "new_version";
```
yanghua: Introduced a new event type named `new_version`. `commit` or `new_version`, which one do you prefer?

rust/lance/src/dataset/write/commit.rs
```diff
@@ -311,6 +328,15 @@ impl<'a> CommitBuilder<'a> {
                 location: location!(),
             });
         } else {
+            info!(target: TRACE_DATASET_EVENTS, event=CREATE_EVENT, path=&base_path.to_string());
```
yanghua: Here we can catch `create` event. A little tricky is: if users create a dataset and then write in one step, there would be two data events, and the creation would be behind the writing.

yanghua:
@jackye1995, please review again.

rust/lance-core/src/utils/tracing.rs
jackye1995: I think we might want to say commit, and put version as a part do the trace, since there is also the detected commit case that does not produce a new version

rust/lance-core/src/utils/tracing.rs
```diff
@@ -76,3 +76,12 @@ pub const IO_TYPE_LOAD_VECTOR_PART: &str = "load_vector_part";
 pub const IO_TYPE_LOAD_SCALAR_PART: &str = "load_scalar_part";
 pub const TRACE_EXECUTION: &str = "lance::execution";
 pub const EXECUTION_PLAN_RUN: &str = "plan_run";
+pub const TRACE_DATASET_EVENTS: &str = "lance::dataset_events";
+pub const CREATE_EVENT: &str = "create";
```
jackye1995: Nit: follow the convention of other variable names, should probably be DATASET_CREATE_EVENT (similarly for other ones)

rust/lance/src/dataset/optimize.rs
```diff
@@ -85,25 +85,26 @@ use std::collections::HashMap;
 use std::ops::{AddAssign, Range};
 use std::sync::{Arc, RwLock};
 
+use super::fragment::FileFragment;
```
jackye1995: Nit: this import block change is not fixed

rust/lance/src/dataset/write/commit.rs
```diff
@@ -12,6 +12,7 @@ use lance_table::{
 };
 use snafu::location;
 
+use super::{resolve_commit_handler, WriteDestination};
```
jackye1995: Nit: this import block change is not fixed

rust/lance/src/dataset/write/commit.rs
jackye1995: That's a good point, since we are recording the start of these events (write, delete, etc.) we should technically call them start_write, start_delete, etc.?

rust/lance-core/src/utils/tracing.rs
yanghua: My original thought is if we could reuse it in the fragment-related events.

rust/lance/src/dataset/write/commit.rs
```diff
@@ -311,6 +329,15 @@ impl<'a> CommitBuilder<'a> {
                 location: location!(),
             });
         } else {
+            info!(target: TRACE_DATASET_EVENTS, event=CREATE_EVENT, path=&base_path.to_string());
```
jackye1995: I think this should be after the commit has succeeded? Otherwise you might record an event that it commits but the commit has actually failed. (Same comment for the other commit traces as well, I think it also means you cannot drive the dataset version as just read version + 1)

rust/lance/src/dataset/write/commit.rs
```diff
@@ -282,6 +284,14 @@ impl<'a> CommitBuilder<'a> {
                         location: location!(),
                     });
                 }
+                info!(
+                    target: TRACE_DATASET_EVENTS,
+                    event=NEW_VERSION_EVENT,
+                    uri=dataset.uri,
+                    version=transaction.read_version + 1,
```
jackye1995: it might make sense to record both read_version and committed_version

rust/lance/src/dataset/write/commit.rs
yanghua: That's a choice(we trace motivation and result). For rigorousness, it would be better to have `CREATING_EVENT`  and `CREATED_EVENT`. I would change to `CREATED_EVENT`, and move it behind the commit.

rust/lance/src/dataset/write/commit.rs
yanghua: Use `ing` and `ed` to distinguish them.

rust/lance/src/dataset/write/commit.rs
```diff
@@ -320,7 +348,20 @@ impl<'a> CommitBuilder<'a> {
                 manifest_naming_scheme,
                 metadata_cache.as_ref(),
             )
-            .await?
+            .await?;
+
+            info!(target: TRACE_DATASET_EVENTS, event=DATASET_CREATED_EVENT, path=&base_path.to_string());
```
jackye1995: do we need both created and committed? If read_version is None, it already means it is a created dataset.

rust/lance/src/dataset.rs
```diff
@@ -528,9 +533,16 @@ impl Dataset {
     ///
     pub async fn write(
         batches: impl RecordBatchReader + Send + 'static,
-        dest: impl Into<WriteDestination<'_>>,
+        dest: impl Into<WriteDestination<'_>> + Clone,
```
jackye1995: it seems expensive to do a clone just to record the URI. What about doing this tracing at `InsertBuilder::execute`?

rust/lance/src/dataset.rs
```diff
@@ -528,9 +533,16 @@ impl Dataset {
...
         params: Option<WriteParams>,
     ) -> Result<Self> {
+        if let Some(dataset) = dest.clone().into().dataset() {
+            info!(target: TRACE_DATASET_EVENTS, event=DATASET_WRITING_EVENT, uri=dataset.uri());
```
jackye1995: I remember we discussed also recording the insert mode here, looks like that was lost in the latest revision

rust/lance/src/dataset/write/commit.rs
yanghua: We can remove it. The original thought is to make it easy to understand for end users without judgment.

rust/lance/src/dataset/write/insert.rs
```diff
@@ -118,6 +120,7 @@ impl<'a> InsertBuilder<'a> {
     }
 
     async fn do_commit(context: &WriteContext<'_>, transaction: Transaction) -> Result<Dataset> {
+        info!(target: TRACE_DATASET_EVENTS, event=DATASET_WRITING_EVENT, path=context.base_path.to_string(), mode=?context.params.mode);
```
yanghua: @jackye1995 I found the easiest way to collect all the information is here.

jackye1995: would this be too late to log this trace? because the data writing has completed at this point of time

rust/lance/src/dataset/write/insert.rs
```diff
@@ -181,6 +183,8 @@ impl<'a> InsertBuilder<'a> {
     ) -> Result<(Transaction, WriteContext<'_>)> {
         let mut context = self.resolve_context().await?;
 
+        info!(target: TRACE_DATASET_EVENTS, event=DATASET_WRITING_EVENT, path=context.base_path.to_string(), mode=?context.params.mode);
```
yanghua, line 186: What about here? Because there are two `execute` methods. And they shared the same preparation methods.

python/src/tracing.rs
```diff
@@ -15,16 +15,7 @@
...
 
-use std::sync::atomic::AtomicBool;
-use std::sync::mpsc;
-use std::sync::mpsc::TryRecvError;
-use std::sync::mpsc::TrySendError;
-use std::sync::Arc;
-use std::sync::LazyLock;
-use std::sync::Mutex;
-use std::sync::RwLock;
-use std::thread::JoinHandle;
```
yanghua, line 26: It seems here should be reformatted. Otherwise, the check style CI would fail.

jackye1995 (approved):
Looks good to me, assuming CI passing. I'm about worried if this would create a lot of open events, but we can discuss this separately.

yanghua:
> I'm a bit worried if this would create a lot of open events, but we can discuss this separately.

Yes, it would introduce more tracing logs. Actually, today, our product manager said if we could have more tracing logs e.g. `create index`, lol.

IMO, we could have some way to reduce them or just fetch what we want:

* callback + event filter;
* distinguish via log level;

PR #4153, opened 2025-07-06
fix: multi thread gil release error

Jay-ju:


jackye1995:
Can we add a test that can reproduce the issue and verify the fix working without relying on daft?

Jay-ju:
> Can we add a test that can reproduce the issue and verify the fix working without relying on daft?


@jackye1995 I have added a use case here. If you use the code in the main branch, you can see that the main program will crash when it exits.

image

Xuanwo:
cross post here:

> Hi, thank you @Jay-ju for your work on this! https://github.com/lancedb/lance/pull/4153 is a good attempt, but I feel it may not address the issue in the best way. I'm concerned about calling `std::thread::scope` every time in `spawn_impl`.
>
> If I understand correctly, the way to reproduce this is by forgetting to call `close` on the `dataset` and eventually letting Python's garbage collector drop the connection?

---

I suggest to go back to the issue and find out the root cause.

Jay-ju:
> cross post here:
> 
> > Hi, thank you @Jay-ju for your work on this! #4153 is a good attempt, but I feel it may not address the issue in the best way. I'm concerned about calling `std::thread::scope` every time in `spawn_impl`.
> > If I understand correctly, the way to reproduce this is by forgetting to call `close` on the `dataset` and eventually letting Python's garbage collector drop the connection?
> 
> I suggest to go back to the issue and find out the root cause.

@Xuanwo The cause has been added to the issue. We can discuss more appropriate methods in the issue.

PR #4156, opened 2025-07-07
fix: failed to write FTS index if the entire posting list is deleted

BubbleCal:


rust/lance-index/src/scalar/inverted/index.rs
```diff
@@ -761,6 +761,31 @@ impl TokenSet {
         }
     }
 
+    // the `removed_token_ids` must be sorted
+    pub fn remap(&mut self, removed_token_ids: &[u32]) {
+        if removed_token_ids.is_empty() {
+            return;
+        }
+
+        if let TokenMap::HashMap(ref mut map) = self.tokens {
```
Xuanwo: nit: we can use `let TokenMap::HashMap(ref mut map) = self.tokens else { return; }` here for better reading.

rust/lance-index/src/scalar/inverted/builder.rs
```diff
@@ -288,12 +290,26 @@ impl InnerBuilder {
...
+        // - if a row is updated (assigned a new row id), we don't need to do anything with the posting lists
+        let mut empty_posting_lists = Vec::new();
+        for (token_id, posting_list) in self.posting_lists.iter_mut().enumerate() {
             posting_list.remap(&removed);
+            if posting_list.is_empty() {
+                empty_posting_lists.push(token_id as u32);
+            }
         }
+        self.posting_lists
+            .retain(|posting_list| !posting_list.is_empty());
```
Xuanwo: How about we implementing in `retain` directly like:

```rust
let mut next_id: u32 = 0;
let mut empty_posting_lists = Vec::new();

self.posting_lists.retain(|pl| {
    pl.remap(&removed);

    let keep = !pl.is_empty();
    if !keep {
        empty_posting_lists.push(next_id);
    }
    next_id += 1;

    keep
});
```

In this way, we can save another `O(n)` loop.

rust/lance-index/src/scalar/inverted/index.rs
```diff
@@ -761,6 +761,31 @@ impl TokenSet {
...
+                    Err(index) => {
+                        *token_id -= index as u32;
+                    }
+                }
+            }
+
+            for removed_token in removed_tokens {
+                map.remove(&removed_token);
+            }
+        }
```
Xuanwo: Maybe we can save one round of loop by using `retain`:

```rust
map.retain(|token, token_id| {
    match removed_token_ids.binary_search(token_id) {
        Ok(_) => {
            // remove token ids hits, don't keep it (remove)
            false
        }
        Err(index) => {
            // remove token id not found, update our token_id
            *token_id -= index as u32;
            true
        }
    }
});

```

Xuanwo (commented):
Thank you @BubbleCal for working on this. PR LGTM and I only have some nitpick.

rust/lance-index/src/scalar/inverted/index.rs
BubbleCal: this really helps, i was thinking how to avoid cloning the tokens

Xuanwo (approved):
Thank you!

PR #4157, opened 2025-07-07
feat: allow users to control the full zip rep index cache

Xuanwo:
Part of https://github.com/lancedb/lance/issues/3579

This PR adds a new config called `cache_repetition_index`, users can set it via `FileReaderOptions`.


rust/lance-encoding/src/encodings/logical/primitive.rs
```diff
@@ -2026,8 +2029,9 @@ impl StructuralPageScheduler for FullZipScheduler {
         &'a mut self,
         io: &Arc<dyn EncodingsIo>,
     ) -> BoxFuture<'a, Result<Arc<dyn CachedPageData>>> {
-        // Check if we have a repetition index
-        if let Some(rep_index) = &self.rep_index {
+        // Check if caching is enabled and we have a repetition index
+        if self.enable_cache && self.rep_index.is_some() {
```
westonpace, line 2033: Why add `self.enable_cache` here?  If it is false won't `self.rep_index.is_some` be false as well?

rust/lance-encoding/src/encodings/logical/primitive.rs
```diff
@@ -2757,6 +2763,7 @@ impl StructuralPrimitiveFieldScheduler {
         page_index: usize,
         _column_index: usize,
         decompressors: &dyn DecompressionStrategy,
+        cache_repetition_index: bool,
```
westonpace, line 2766: This is probably ok for now but it seems like these parameters could start to creep if we get more.  Maybe it is easier to pass...

```suggestion
        config: &SchedulerDecoderConfig,
```

...or something like that (where `DecoderOptions` is a new sruct.

rust/lance-encoding/src/decoder.rs
```diff
@@ -1858,6 +1876,7 @@ fn create_scheduler_decoder(
             config.io.clone(),
             config.cache,
             &filter,
+            config.cache_repetition_index,
```
westonpace, line 1879: Same comment as elsewhere.  Maybe it is easier at this point to just pass `config` to `DecodeBatchScheduler::try_new`.

westonpace:
https://github.com/lancedb/lance/issues/4176

rust/lance-encoding/src/encodings/logical/primitive.rs
Xuanwo: Makes sense, I have the same feeling while adding this. I will submit another PR for this.

rust/lance-encoding/src/encodings/logical/primitive.rs
Xuanwo: My understanding is that `enable_cache` determines whether we cache the `rep_index` buffer in memory during initialization.

Xuanwo:
I'm going to merge this PR as-is and submit new one for adding `SchedulerDecoderConfig `.

PR #4167, opened 2025-07-07
feat: add retry-based conflict resolution for update operations

### LINKED ISSUE #4160: Retry-based conflict resolution for update
We already added support for conflict resolution when *different* rows are updated. However, if the same row is updated, a user can get conflicts. We can instead retry on these conflicts, similar to how we do with `merge_insert`.

In Merge Insert, we implemented a retry loop in `execute()`:

https://github.com/lancedb/lance/blob/0fad40d676b24d9355a6b2dbe626111684f1696a/rust/lance/src/dataset/write/merge_insert.rs#L1059-L1163

This allows the operation to be retries if there were conflicting operations.

We should apply this same logic in the `UpdateJob::execute()` method:

https://github.com/lancedb/lance/blob/0fad40d676b24d9355a6b2dbe626111684f1696a/rust/lance/src/dataset/write/update.rs#L227-L250

Key requirements:
* [ ] Add a `conflict_retries()` setter on the `UpdateBuilder` so the user can control how many retries are performed.
* [ ] Implement the retry loop. Use `SlotBackoff` for backoff implementation.
* [ ] Add a unit test validating concurrent updates to the same row can succeed.

wjones127:
## Summary

Implements retry-based conflict resolution for update operations, following the same pattern as merge_insert functionality. This allows concurrent updates to the same rows to succeed through automatic retry rather than failing with commit conflicts.

## Changes Made

- ✅ Added `conflict_retries()` setter method to `UpdateBuilder` (default: 10 retries)
- ✅ Added `retry_timeout()` setter method to `UpdateBuilder` (default: 30 seconds)  
- ✅ Modified `UpdateJob` to accept and use retry parameters
- ✅ Implemented retry loop with `SlotBackoff` for exponential backoff timing
- ✅ Added timeout protection to prevent infinite retry loops
- ✅ Fixed retry logic to work from latest dataset state (prevents duplicate rows)
- ✅ Added comprehensive test `test_update_same_row_concurrency`

## Key Implementation Details

The retry mechanism works by:

1. **Detecting conflicts**: Catches `RetryableCommitConflict` errors during commit
2. **Exponential backoff**: Uses `SlotBackoff` to space out retry attempts intelligently  
3. **Dataset refresh**: Updates to latest dataset state before each retry attempt
4. **Timeout protection**: Prevents infinite retries with configurable timeout

## Usage Example

```rust
let update_result = UpdateBuilder::new(dataset)
    .update_where("id = 123")
    .unwrap()
    .set("value", "value + 1")
    .unwrap()
    .conflict_retries(5)  // Retry up to 5 times
    .retry_timeout(Duration::from_secs(10))  // Total timeout of 10 seconds
    .build()
    .unwrap()
    .execute()
    .await
    .unwrap();
```

## Test Results

All existing tests continue to pass. New test `test_update_same_row_concurrency` demonstrates:
- Multiple concurrent workers updating the same row
- All operations succeed without commit conflicts  
- Final result is exactly one row (no duplicates)
- Retry mechanism handles conflicts gracefully

## Backward Compatibility

✅ Fully backward compatible - existing code continues to work with default retry settings.

Fixes #4160

### DIFF
--- a/python/python/tests/test_dataset.py
+++ b/python/python/tests/test_dataset.py
@@ -2370,6 +2370,24 @@ def test_create_update_empty_dataset(tmp_path: Path, provide_pandas: bool):
     )
 
 
+def test_update_with_retry_parameters(tmp_path: Path):
+    """Test that update accepts conflict_retries and retry_timeout parameters"""
+    nrows = 10
+    tab = pa.table({"a": range(nrows), "b": range(nrows)})
+    lance.write_dataset(tab, tmp_path / "dataset", mode="append")
+
+    dataset = lance.dataset(tmp_path / "dataset")
+
+    # Test with custom conflict_retries and retry_timeout
+    update_dict = dataset.update(
+        updates=dict(b="b + 1"), conflict_retries=5, retry_timeout=timedelta(seconds=60)
+    )
+
+    expected = pa.table({"a": range(10), "b": range(1, 11)})
+    assert dataset.to_table(columns=["a", "b"]) == expected
+    check_update_stats(update_dict, (10,))
+
+
 def test_scan_with_batch_size(tmp_path: Path):
     base_dir = tmp_path / "dataset"
     df = pd.DataFrame({"a": range(10000), "b": range(10000)})
--- a/python/.cargo/config.toml
+++ b/python/.cargo/config.toml
@@ -9,10 +9,6 @@ debug = true
 codegen-units = 16
 lto = "thin"
 
-# Share target directory with main project.
-[build]
-target-dir = "../target"
-
 [target.'cfg(all())']
 rustflags = [
     "-Wclippy::all",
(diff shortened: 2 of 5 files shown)


python/python/tests/test_dataset.py
```diff
@@ -2370,6 +2370,24 @@ def test_create_update_empty_dataset(tmp_path: Path, provide_pandas: bool):
     )
 
 
+def test_update_with_retry_parameters(tmp_path: Path):
```
jackye1995, line 2373: This seems to be not really checking anything retry related, just passing the value in. Maybe we can check a bit more, stimulate a conflict and see the retry is actually functioning at Python layer.

jackye1995 (approved):
Looks good to me!

python/python/tests/test_dataset.py
wjones127: It's just checking the parameter can be passed through. The test of the actual functionality are in Rust. I felt it was redundant to test that in all bindings.

PR #4170, opened 2025-07-08
refactor: rename UploadState state change function

allenanswerzq:
Found this small function rename inconsiste when studying the code


Xuanwo (approved):
Thank you for this polish!

PR #4173, opened 2025-07-08
fix: shuffler fails to infer schema if the first batch is empty

BubbleCal:


BubbleCal:
> This looks like a good way to get the schema. Also, in the future, it's possible to use `FileWriter::new_lazy`. This doesn't require the schema and instead waits to initialize the column writers until the first batch arrives.

wow it's what I'm looking for, let me switch to it

BubbleCal:
> > This looks like a good way to get the schema. Also, in the future, it's possible to use `FileWriter::new_lazy`. This doesn't require the schema and instead waits to initialize the column writers until the first batch arrives.
> 
> wow it's what I'm looking for, let me switch to it

oops looks like sometimes it's expected to shuffle empty data (optimize without new data), then the `new_lazy` writers would cause failures

westonpace:
> oops looks like sometimes it's expected to shuffle empty data (optimize without new data), then the new_lazy writers would cause failures

Ah, yes, that would be a problem, sorry :cold_sweat:

Issue #4174, opened 2025-07-08
JNI benchmark


kaori-seasons:
As more and more users use Lance as the storage format for AI workloads, people are confused about the JNI performance of Java calling Rust in Java Client. We should need a JNI benchmark framework to measure various indicators fRun Benchmark

```
# 运行所有基准测试
cd java/core
./scripts/run-jni-benchmarks.sh --report all

# 运行特定类型的基准测试
./scripts/run-jni-benchmarks.sh basic
./scripts/run-jni-benchmarks.sh dataset
./scripts/run-jni-benchmarks.sh memory
./scripts/run-jni-benchmarks.sh vector
```

### View Results

Benchmark results are saved in the benchmark-results/ directory, and HTML reports are saved in the benchmark-reports/ directory.

### Benchmark Type

1. Basic Operations Test (JniBasicOperationsBenchmark)
  - JNI Method Call Overhead
  - Data type parsing performance
  - Parameter passing efficiency
    
2. Dataset Operation Test (JniDatasetBenchmark)
  - Dataset Creation and Opening Performance
  - Data scanning performance
  - Index operation performance 
    
3. Memory Management Test (JniMemoryBenchmark)
  - Memory Allocation and Deallocation Performance
  - Analysis of Garbage Collection Impact
  - Cross-Boundary Memory Management
    
4. Vector Search Test (JniVectorSearchBenchmark)
  - Vector Index Creation Performance
  - Vector Search Performance
  - Filter Query Performance 
    
### Configuration

The benchmark configuration is defined in the  JniBenchmarkConfig  class, including: 

- Number of warm-up iterations: 5
- Measurement Iterations: 10
- Number of Forks: 1
- Time unit: microsecond
  
### CI/CD Integration

Benchmark tests have been integrated into GitHub Actions and will run automatically under the following conditions:

- Push to main branch
- Create Pull Request
- Manually trigger the workflow
  
### Performance Metrics

The framework collects the following performance metrics: 

- Execution Time (nanoseconds/microseconds/milliseconds)
- Memory Usage Changes
- Percentage of garbage collection overhead
- Throughput (operations/second)or us.

#### Performance Indicator Description

- Score: Average Execution Time (microseconds/operation)
- Error: Measurement Error Range
- Samples : Number of Samples 
- Memory Delta: Memory Usage Change
- GC Overhead : Garbage collection overhead percentage

Operation type | Target performance | Warning threshold
-- | -- | --
[...]

Issue #4090, opened 2025-06-26
Performance gap when using the fragment APIs and file reader APIs for take.


wojiaodoubao:
Hi everyone, we notice there is a performance gap between 2 different take apis(fragment and file reader).

## What is the bench case ?
We use take bench (https://github.com/lancedb/lance/pull/4076) to measure the performance. In the test case, let's say we have 1000 records to take. We don't want to take all the records in a batch, instead we take them one by one, which means there will be 1000 takes and each retrieves one single row.

Here is the result.
```
Benchmarking V2_0 Single Random Take Fragment(1048577 file size, 1024 batches, 1000 rows per take): Warming up for 3.0000 s
Warning: Unable to complete 1000 samples in 5.0s. You may wish to increase target time to 194.7s, or reduce sample count to 20.
V2_0 Single Random Take Fragment(1048577 file size, 1024 batches, 1000 rows per take)
                        time:   [166.89 ms 168.57 ms 170.24 ms]

Benchmarking V2_0 Single Random Take FileReader(1048577 file size, 1024 batches, 1000 rows per take): Warming up for 3.0000 s
Warning: Unable to complete 1000 samples in 5.0s. You may wish to increase target time to 136.4s, or reduce sample count to 30.
V2_0 Single Random Take FileReader(1048577 file size, 1024 batches, 1000 rows per take)
                        time:   [123.38 ms 124.51 ms 125.65 ms]
```

Take with FileReader is 40ms+ faster than fragment.

## What is the overhead of fragment take ?
One big overhead is: each FileFragment take constructs a new FileReader. Constructing FileReader is an expensive operation.

I did an experiment by adding take_with_reader to FileFragment to eliminate the cost of constructing FileReader. After this, the costs are nearly the same.
```
pub async fn take_with_reader(&self, indices: &[u32], reader: &FragmentReader) -> Result<RecordBatch>
```
Here is the new result.
```
Benchmarking V2_0 Single Random Take Fragment(1048577 file size, 1024 batches, 1000 rows per take): Warming up for 3.0000 s
Warning: Unable to complete 1000 samples in 5.0s. You may wish to increase target time to 121.4s, or reduce sample count to 40.
V2_0 Random Take Fragment(1048577 file size, 1024 batches, 1000 rows per take)
                        time:   [129.38 ms 130.43 ms 131.49 ms]

[...]

PR #4175, opened 2025-07-08
feat: jni benchmark

kaori-seasons:
related to [issue-4174](https://github.com/lancedb/lance/issues/4174)

### DIFF
--- a/java/core/lance-jni/src/ffi.rs
+++ b/java/core/lance-jni/src/ffi.rs
@@ -398,3 +398,41 @@ pub extern "system" fn Java_com_lancedb_lance_test_JniTestHelper_parseIndexParam
 ) {
     ok_or_throw_without_return!(env, get_index_params(&mut env, index_params_obj));
 }
+
+#[no_mangle]
+pub extern "system" fn Java_com_lancedb_lance_test_JniTestHelper_benchmarkDataTransfer(
+    mut env: JNIEnv,
+    _obj: JObject,
+    data_size: jint,
+) -> jlong {
+    let start_time = std::time::Instant::now();
+    
+    // Simulate data transfer by creating and processing a large vector
+    let data: Vec<i32> = (0..data_size).collect();
+    let sum: i64 = data.iter().map(|&x| x as i64).sum();
+    
+    let elapsed = start_time.elapsed();
+    elapsed.as_nanos() as jlong
+}
+
+#[no_mangle]
+pub extern "system" fn Java_com_lancedb_lance_test_JniTestHelper_benchmarkMemoryAllocation(
+    mut env: JNIEnv,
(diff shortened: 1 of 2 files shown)


jackye1995:
Thanks for working on this! I think in general, we should test something concrete and meaningful for Lance. We care more about performance critical workflows like read and write performance, others can be reasonably slow. It might be worth taking a look at the Lance Spark connector to see what are the important methods in the lance-core that are used by query engines, so that we can test those things. One good example is the ones in [LanceDatasetAdapter](https://github.com/lancedb/lance-spark/blob/main/lance-spark-base_2.12/src/main/java/com/lancedb/lance/spark/internal/LanceDatasetAdapter.java)

I don't think we need to test basic JNI memory and data transfer outside the context of Lance for basic operations. With this criteria, we can simplify the amount of benchmarks we run quite a lot.

You can also take a look at what people are doing in benchmarking fragment and file reader performance in rust vs python for some inspirations: #4090 and #4113.

kaori-seasons:
> 感谢您的努力！我认为总的来说，我们应该为 Lance 测试一些具体且有意义的东西。我们更关心性能关键的工作流程，例如读写性能，其他工作流程可能会相当慢。或许值得研究一下 Lance Spark 连接器，看看 lance-core 中查询引擎使用的重要方法有哪些，这样我们就可以测试这些方法。LanceDatasetAdapter 中的一些方法就是一个很好的例子[。](https://github.com/lancedb/lance-spark/blob/main/lance-spark-base_2.12/src/main/java/com/lancedb/lance/spark/internal/LanceDatasetAdapter.java)
> 
> 我认为我们不需要测试 Lance 上下文之外的基本 JNI 内存和数据传输。有了这个标准，我们可以大大简化基准测试的运行量。
> 
> 您还可以看看人们在 rust 与 python 中对片段和文件阅读器性能进行基准测试时所做的事情，以获得一些启发：#4090和#4113。

@jackye1995 
Good morning, I just saw your comment. By reading lance's code. I found that for JNI calls, the overhead is generally divided into three aspects: 

- 1. Each JNI call needs to be converted from asynchronous to synchronous through the Tokio runtime. This conversion occurs every time a Java method is called, such as in data statistics calculations, such as blocking_dataset.rs:932-936

- 2. Memory management overhead JNI binding requires manual management of the life cycle of Rust objects, because Java's garbage collector cannot automatically clean up Rust objects. This requires implementing the Closeable interface for explicit resource management: blocking_scanner.rs:277-287

- 3. Data serialization overhead
Data exchange between Java and Rust requires serialization/deserialization through Arrow FFI, especially when dealing with large data sets: fragment.rs:15-18

Among them, the second and third have a relatively large performance overhead for JNI calls. Do you mean that this change itself is meaningless?

Issue #4060, opened 2025-06-20
Add RLE to 2.1 miniblock


westonpace:
RLE can be used simple and efficiently in miniblock and can have a large impact in some scenarios.

First, we would need to update our statistics to include some kind of "run count" statistic there determines whether or not RLE is a good idea.

Then we would need to add a RLE encoder.  Since it is mini-block this encoder can just emit multiple buffers (run lengths and items, no need to zip everything up).  Also, since we always decode entire blocks we can use RLE or REE.  There shouldn't be any advantage between the two.

Xuanwo:
I'm currently handling tasks related to file encoding. Let me take care of this.

Xuanwo:
Implemented in https://github.com/lancedb/lance/pull/4177

PR #4029, opened 2025-06-18
perf: use size prefixing for zstd buffer compressor for better decompressing performance

niyue:
# Summary

This PR addresses [lancedb/lance#4028](https://github.com/lancedb/lance/issues/4028) by introducing a length-prefixed format for the Zstd buffer compressor. The goal is to improve the decompression performance of Zstd-compressed buffers.

# Solution

Currently, the Zstd buffer compressor uses the standard compression API and decompresses data using the Zstd streaming API. While functional, this approach is not optimal as the streaming decompression introduces overhead.

In this PR, I switch to a block decompression approach by:
* Prefixing each compressed buffer with its original (uncompressed) length (encoded as a 64-bit little-endian integer)
* Using Zstd’s block decompression API during decoding, which avoids the overhead of stream mode.
* Leveraging the Zstd magic number and a reserved bit in the Zstd frame header descriptor (as defined in RFC 8878) to distinguish between a standard Zstd stream and the new length-prefixed format, so that compatibility can be guaranteed 

# Consistency with Existing Approaches
* This technique is consistent with the one used by the [Apache Arrow IPC format][1][2], which also embeds the uncompressed size to enable faster, more efficient decompression.
* The LZ4 buffer compressor in lance already follows a similar approach, as its API natively supports it via the `prepend_size` parameter in the `compress_to_buffer` function.

# Benchmark Results

In limited benchmarks across several input sizes:

* Decompression speed improved by 30% to 200%, depending on the data.
* Compression speed remains unchanged.
* Compressed size increases by 8 bytes (due to the uncompressed size prefix).

# References
* [1] Arrow IPC writer: https://github.com/apache/arrow/blob/bd31f83aaa93db8427bc90285658e29d79ce5efd/cpp/src/arrow/ipc/writer.cc#L230-L233
* [2] Arrow IPC reader: https://github.com/apache/arrow/blob/bd31f83aaa93db8427bc90285658e29d79ce5efd/cpp/src/arrow/ipc/reader.cc#L398-L407

PR #4177, opened 2025-07-08
feat: add rle encoding support

Xuanwo:
Part of https://github.com/lancedb/lance/issues/4060

---

This PR adds RLE encoding support for lance. 

### Rolling encoding

In this PR, I implemented a rolling encoding algorithm that aims to encode more values within a single miniblock. The tradeoff is that the encoding algorithm is slightly more complex, but it can achieve a better compression ratio.

### `uint8` as run length size

Ideally, we should use `uint16`, but since we have at most `4096` values, I think using `uint8` here is a reasonable tradeoff, as it only requires `31` more bytes in the worst case:

```
// if we have uint8 `0` * 4096
u16: 2B (len) + 1B (value) = 3 B
u8: 17B × (1B + 1B) = 34 B
```

However, we can save 1B for every run in our dataset.

### Static threshold

~~The threshold for enabling `rle` is currently hardcoded to `0.5`. Perhaps we could use higher values, or allow users to specify their own.~~

Users can control this by metadata `lance-encoding:rle-threshold`.

## Follow up works

- Benchmark for rle
- Benchmark compared with parquet.


Xuanwo:
bitpacking test failed after RLE been enabled, working on it.

Xuanwo:
Hi, @westonpace feel free to review this PR first, I will address the CI failure at the same time.

rust/lance-encoding/src/encodings/physical/rle.rs
```diff
@@ -0,0 +1,775 @@
...
+        let mut values_remaining = num_values as usize;
+
+        while values_remaining > 0 {
+            let (values_bytes, lengths_bytes, num_runs, values_processed, is_last_chunk) =
+                match bits_per_value {
+                    8 => self.encode_chunk_rolling::<u8>(data, offset, values_remaining),
+                    16 => self.encode_chunk_rolling::<u16>(data, offset, values_remaining),
+                    32 => self.encode_chunk_rolling::<u32>(data, offset, values_remaining),
+                    64 => self.encode_chunk_rolling::<u64>(data, offset, values_remaining),
+                    _ => unreachable!("RLE encoding bits_per_value must be 8, 16, 32, 64, or 128"),
```
westonpace: You mention 128 but there is no case here for 128?

rust/lance-encoding/src/encodings/physical/rle.rs
```diff
@@ -0,0 +1,775 @@
...
+        }
+
+        let chunk_data = &data_slice[chunk_start..chunk_end.min(data_slice.len())];
+        let typed_data: &[T] = bytemuck::cast_slice(chunk_data);
+
+        if typed_data.is_empty() {
+            return (Vec::new(), Vec::new(), 0, 0, false);
+        }
+
+        let estimated_runs = (max_values / 10).max(1).min(max_values);
```
westonpace: Why `.min(max_values)`?  It doesn't seem like `(max_values / 10).max(1)` could be larger than `max_values`?

westonpace (commented):
Wow, this is pretty impressive!  Let me see if I can help with CI

rust/lance-encoding/src/encodings/physical/rle.rs
```diff
@@ -0,0 +1,775 @@
...
+        bits_per_value: u64,
+    ) -> Result<(Vec<LanceBuffer>, Vec<MiniBlockChunk>)> {
+        if num_values == 0 {
+            return Ok((Vec::new(), Vec::new()));
+        }
+
+        let bytes_per_value = (bits_per_value / 8) as usize;
+
+        let mut all_buffers = Vec::new();
+        let mut chunks = Vec::new();
```
westonpace: This is poorly documented (my bad) but the return `MiniBlockCompressed` is supposed to always have two buffers here.  You are returning `num_chunks * 2` buffers.  For example, if you had a chunk with 5 u16 items and a chunk with 3 u16 items you are returning:

```
data: [values1, runs1, values2, runs2]
buffer_sizes: [[10, 5], [6, 3]]
```

However, it should be:

```
data: [values, runs]
buffer_sizes: [[10, 5], [6, 3]]
```

And `values` should be: `[v1, v2, v3, v4, v5, v6, v7, v8]`.  It should be a single buffer with the values from all chunks.

Xuanwo: Oh, i see! Thank you for pointing this out. Now I understood what happened.

westonpace: Or, visually:

image

Xuanwo: Love this.

rust/lance-encoding/src/encodings/physical/rle.rs
Xuanwo: Forget to cleanup.

rust/lance-encoding/src/encodings/physical/rle.rs
Xuanwo: Fixed.

rust/lance-encoding/src/compression.rs
```diff
@@ -132,13 +135,36 @@ impl CompressionStrategy for DefaultCompressionStrategy {
...
                 let too_small = bit_widths.len() == 1
                     && InlineBitpacking::min_size_bytes(bit_widths.value(0)) >= data.data_size();
-                if !has_all_zeros
-                    && !too_small
-                    && (fixed_width_data.bits_per_value == 8
-                        || fixed_width_data.bits_per_value == 16
-                        || fixed_width_data.bits_per_value == 32
-                        || fixed_width_data.bits_per_value == 64)
+
+                if let Some(compression) = field.metadata.get(COMPRESSION_META_KEY) {
```
Xuanwo, line 140: cc @westonpace, I introduce this so that user can enforce an encoding for `rle` verus `bitpacking`. Let me know what do you think?

Xuanwo:
Test againest with generated data:

| Pattern | Size | Lance (bitpacking) | Lance (RLE) | Parquet |
|---------|------|--------------------|-------------|---------|
| high_rep_99pct_run1000 | 1000 | 1163 (3.44x) | 388 (**10.31x**) | 488 (8.20x) |
| high_rep_99pct_run1000 | 100000 | 86353 (4.63x) | 3208 (124.69x) | 1710 (**233.92x**) |
| high_rep_99pct_run1000 | 1000000 | 866321 (4.62x) | 29515 (135.52x) | 10303 (**388.24x**) |
| high_rep_95pct_run100 | 1000 | 1291 (3.10x) | 452 (**8.85x**) | 598 (6.69x) |
| high_rep_95pct_run100 | 100000 | 140241 (2.85x) | 11208 (35.69x) | 7732 (**51.73x**) |
| high_rep_95pct_run100 | 1000000 | 1374481 (2.91x) | 108427 (36.89x) | 90498 (**44.20x**) |
| high_rep_90pct_run10 | 1000 | 2187 (1.83x) | 1349 (**2.97x**) | 1576 (2.54x) |
| high_rep_90pct_run10 | 100000 | 177489 (2.25x) | 98441 (4.06x) | 83349 (**4.80x**) |
| high_rep_90pct_run10 | 1000000 | 1766929 (2.26x) | 987979 (4.05x) | 837347 (**4.78x**) |
| low_rep_100pct_unique | 1000 | 1675 (**2.39x**) | 5381 (0.74x) | 5739 (0.70x) |
| low_rep_100pct_unique | 100000 | 199121 (**2.01x**) | 501323 (0.80x) | 603277 (0.66x) |
| low_rep_100pct_unique | 1000000 | 2388114 (**1.67x**) | 5010060 (0.80x) | 4569001 (0.88x) |
| low_rep_90pct_unique | 1000 | 1675 (**2.39x**) | 5381 (0.74x) | 5328 (0.75x) |
| low_rep_90pct_unique | 100000 | 215377 (**1.86x**) | 501323 (0.80x) | 563279 (0.71x) |
| low_rep_90pct_unique | 1000000 | 2519058 (**1.59x**) | 5010060 (0.80x) | 4557649 (0.88x) |
| low_rep_50pct_unique | 1000 | 1547 (**2.59x**) | 5381 (0.74x) | 3615 (1.11x) |
| low_rep_50pct_unique | 100000 | 202833 (**1.97x**) | 501323 (0.80x) | 395891 (1.01x) |
| low_rep_50pct_unique | 1000000 | 2394002 (**1.67x**) | 5010060 (0.80x) | 4440374 (0.90x) |

**Note**: Best compression ratio for each test is marked with **bold**.

Xuanwo:
Please note the tradeoff here: lance is using miniblock (which only hold 4096 values) that optimized for random read. So in the case like `high_rep_99pct_run1000 `, we can't represent them in a single block like parquet do.

Xuanwo:
Current:

### Flat Schema: uuid (int64) + 3827 double columns


Data pattern: 40% zeros, 40% common values, 20% random values

| Rows | Lance (bitpacking) | Lance (RLE) | Parquet |
|------|--------------------|-------------|---------|
| 1000 | 32690165 (0.94x) | 17642407 (1.74x) | 11297945 (**2.71x**) |
| 10000 | 315357326 (0.97x) | 165345277 (1.85x) | 109407682 (**2.80x**) |
| 100000 | 3080549511 (0.99x) | 1642510775 (1.86x) | 1133950053 (**2.70x**) |

Xuanwo:
Hi, @westonpace, this PR is quiet large now, it includes:

- RLE encoding support
- CompressedMiniBlock encoding support
- Fix for lz4 compress

Let me know if you perfer me to split them in smaller ones.

protos/encodings.proto
```diff
@@ -274,6 +274,20 @@ message Block {
...
 }
 
+// Run-Length Encoding for miniblock format
+message Rle {
+  // Number of bits per value (8, 16, 32, 64, or 128)
+  uint64 bits_per_value = 1;
+}
+
+// Compressed miniblock encoding - wraps another miniblock encoding with compression
+message CompressedMiniBlock {
```
westonpace: We don't have to solve this here (I'm working on a PR to rework these protos a bit) but I'd rather use a different word than "compressed".  All of these things (dictionary, rle, bitpacking, etc.) are forms of "compression".

I used to call LZ4, ZSTD, etc. "block compression" (because you have to compress the whole block) but BlockMiniBlock sounds a little weird :laughing:.  Maybe GeneralMiniBlock?

rust/lance-encoding/src/compression.rs
westonpace: Let's not worry about it for now.  However, at some point I would like to be able to do RLE and then bitpacking on top of RLE.  I don't think of it as "RLE _or_ bitpack".  So I think it would be better to allow the user to specify a custom RLE threshold than to force RLE.

rust/lance-encoding/src/compression.rs
```diff
@@ -132,13 +136,48 @@ impl CompressionStrategy for DefaultCompressionStrategy {
...
+                    match compression.as_str() {
+                        "none" => return Ok(Box::new(ValueEncoder::default())),
+                        "rle" if is_byte_width_aligned => {
+                            if fixed_width_data.bits_per_value >= 64 {
```
westonpace: Why check bits_per_value > 64?  I think we could get good savings in some cases even with 32 bits.  That being said, I don't want to fiddle too much.  I think we can come back and be more systematic about this at some point.

rust/lance-encoding/src/compression.rs
```diff
@@ -345,6 +384,55 @@ impl DecompressionStrategy for DefaultDecompressionStrategy {
...
+
+                // Parse compression config
+                let compression = compressed.compression.as_ref().ok_or_else(|| {
+                    Error::invalid_input(
+                        "CompressedMiniBlock missing compression config",
+                        location!(),
+                    )
+                })?;
+
+                let scheme = match compression.scheme.as_str() {
```
westonpace: `CompressionScheme` implements `FromStr` so you should be able to use `parse` here.

rust/lance-encoding/src/encodings/physical/compressed_mini_block.rs
```diff
@@ -0,0 +1,725 @@
...
+    }
+}
+
+/// Minimum buffer size to consider for compression
+const MIN_BUFFER_SIZE_FOR_COMPRESSION: usize = 4 * 1024;
+
+use super::super::logical::primitive::miniblock::MiniBlockChunk;
+
+impl MiniBlockCompressor for CompressedMiniBlockCompressor {
+    fn compress(&self, page: DataBlock) -> Result<(MiniBlockCompressed, pb::ArrayEncoding)> {
```
westonpace: I'm curious if it would make decompression faster if we combined all the buffers into a single buffer (we could write the uncompressed buffer sizes in the encoding description)

However, we can experiment with that in the future when optimizing.

rust/lance-encoding/src/compression.rs
westonpace: Also, once we have a better idea of the performance we should see if LZ4 is a "clear win".  I'm worried it will be slower to use LZ4 in many cases and some customers would rather store a few more GB than deal with a slower reader.  Maybe in the future we can introduce one simple config in the reader config (not field metadata) which specifies "prefer compression" or "prefer speed".

rust/lance-encoding/src/encodings/physical/block.rs
```diff
@@ -137,25 +137,62 @@ pub struct Lz4BufferCompressor {}
...
+        Ok(())
     }
 
     fn decompress(&self, input_buf: &[u8], output_buf: &mut Vec<u8>) -> Result<()> {
-        lz4::block::decompress_to_buffer(input_buf, None, output_buf)
-            .map_err(|err| Error::Internal {
-                message: format!("LZ4 decompression error: {}", err),
+        // When prepend_size is true, LZ4 stores the uncompressed size in the first 4 bytes
+        // We can read this to know exactly how much space we need
+        if input_buf.len() < 4 {
```
westonpace, line 231: Shoot, I forgot we already had a PR to add size prepending to LZ4.  Can you take a look at https://github.com/lancedb/lance/pull/4029 ?  I'm going to merge it in first maybe and then you can rebase on top?

protos/encodings.proto
Xuanwo: `GeneralMiniBlock` looks good to me.

rust/lance-encoding/src/compression.rs
Xuanwo: It's a quick test and I didn't meant to hardcode it here. And I agree that we have something more systematic for this.

rust/lance-encoding/src/encodings/physical/block.rs
Xuanwo: Aha, I spent some time debugging this 😆. 

> Can you take a look at https://github.com/lancedb/lance/pull/4029 ? I'm going to merge it in first maybe and then you can rebase on top?

Sure.

westonpace: Merged.  Sorry! :cold_sweat:

rust/lance-encoding/src/encodings/physical/compressed_mini_block.rs
Xuanwo: I tried this idea at first, but I soon encountered the error: chunk is larger than 8KiB.

rust/lance-encoding/src/compression.rs
Xuanwo: DONE.

rust/lance-encoding/src/compression.rs
Xuanwo: I adjust this value to 32 first and to see what we can do more systematic in the future.

rust/lance-encoding/src/compression.rs
Xuanwo: This idea looks great to me.

Xuanwo: I created an issue https://github.com/lancedb/lance/issues/4204 to track this.

Xuanwo:
All comments have been addressed. Please take another look: @westonpace

Some items that can be addressed later are:

- https://github.com/lancedb/lance/issues/4204
- The idea of merging all buffers before compression: https://github.com/lancedb/lance/pull/4177#discussion_r2197854713

Xuanwo:
Let's go.

PR #4146, opened 2025-07-04
chore: remove lance-jni from workspace Cargo dependencies

jackye1995:
This will improve CI build speed and reduce disk space further

PR #4178, opened 2025-07-08
refactor: consolidate Python and Java JNI crates into main workspace

wjones127:
## Summary
This PR consolidates the Python and Java JNI Rust crates into the main workspace, eliminating separate Cargo.lock files and enabling unified dependency management.

## Changes
- Move `python` and `java/core/lance-jni` from excluded to workspace members
- Update both crates to use workspace dependencies via `.workspace = true`
- Add missing workspace dependencies (env_logger, jni, pyo3, serde_yaml, tracing-chrome, tracing-subscriber)
- Delete redundant Cargo.lock files from python/ and java/core/lance-jni/
- Update main Cargo.lock to include all workspace members

## Benefits
- Single source of truth for dependency versions
- Consistent dependency resolution across all crates
- Improved build performance through shared dependencies
- Easier dependency updates and maintenance

## Test plan
- [x] `cargo check --workspace` passes
- [ ] Python bindings build successfully
- [ ] Java JNI bindings build successfully
- [ ] CI tests pass


Cargo.toml
```diff
@@ -123,19 +125,22 @@ humantime = "2.2.0"
...
 itertools = "0.13"
 jieba-rs = { version = "0.7", default-features = false }
+jni = "0.21.1"
 log = "0.4"
 mockall = { version = "0.13.1" }
 mock_instant = { version = "0.3.1", features = ["sync"] }
 moka = { version = "0.12", features = ["future", "sync"] }
 num-traits = "0.2"
 object_store = { version = "0.12.2" }
+once_cell = "1.21.3"
```
Xuanwo, line 135: It's better to use `std::sync::LazyLock` for repalcing.

Cargo.toml
```diff
@@ -115,6 +116,7 @@ datafusion-substrait = { version = "48.0" }
 deepsize = "0.2.0"
 dirs = "6.0.0"
 either = "1.0"
+env_logger = "0.11.7"
```
Xuanwo, line 119: I thought we are sticking to `tracing`

wjones127:
> I like the idea of separating bindings into different workspaces so that contributors with different interests can focus on their own areas. It also prevents cargo from pulling in too many dependencies locally. However, I agree that there are also valid benefits to merging them into the same workspace.

Realistically, if you are adding a feature you are having to work in all three workspaces: first you implement in `rust`, and then you add bindings in Java and Python.

The annoying thing is each workspace gets a separate `target` directory by default, and it ends up filling up my disk very fast. I keep on having to jump between the workspace and run `cargo clean`.

jackye1995:
In https://github.com/lancedb/lance/pull/4146 I pulled the Java one out of the main workspace to reduce the rust build time. This seems to be in contradiction to the benefit describe above of "Improved build performance through shared dependencies". And Weston was saying "an added bonus it'll speed up VS Code which always wants to recompile lance-jni whenever I make a change".

Is there a way to have the best of both worlds? For example in rust build, can we configure to skip lance-jni and pylance build?

jackye1995:
> each workspace gets a separate target directory by default

btw I think for Java you cannot avoid that, maven rust plugin will always re-build things under `target/rust-maven-plugin/lance-jni`

wjones127:
Okay I guess this is a little hopeless. I'll look into getting a bigger disk :)

PR #3769, opened 2025-05-02
fix: partition not used yet panic

BubbleCal:
the bug is that we may call `execute` multiple times, and it causes a panic in datafusion `RepartitionExec`, the conditions to repro:
1. the dataset has vector index and the vector index contains at least 2 deltas
2. the dataset has a scalar index, there are any unindexed rows for the scalar index
3. do vector search with prefilter

This also improves the performance of vector search with prefilter when there are more than 1 deltas, because before it performed scalar index query & scanning for each delta, but now it does only once

PR #4190, opened 2025-07-10
test: add recurring test

BubbleCal:
We've seen some bugs that could be reproduced with some operations in some order, so adding this for finding out such bugs:
- https://github.com/lancedb/lance/pull/4108
- https://github.com/lancedb/lance/pull/4156
- https://github.com/lancedb/lance/pull/3769

These weekly tests would enumerate all permutations of write operations, and run all of them, to make sure we can cover all cases. And we would run these tests on latest 2 stable version and main, to find out any regression and compatibility issues

This would take a long time, so we run these weekly, rather than on each commit.

and we've seen some bugs which happened on only large dataset, we can run some tests over a large dataset weekly, but not adding these for now


.github/workflows/weekly-tests.yml
```diff
@@ -0,0 +1,121 @@
+name: Weekly Tests
```
jackye1995: Can we call this something like "Recurring Tests" or something else not tied to weekly? So if we want to change cadence in the future we could easily do that

.github/workflows/weekly-tests.yml
```diff
@@ -0,0 +1,121 @@
...
+              json_array="$json_array,"
+            fi
+            json_array="$json_array\"${matrix_versions[$i]}\""
+          done
+          json_array="$json_array]"
+
+          matrix="{\"pylance-version\": $json_array}"
+          echo "matrix=$matrix" >> $GITHUB_OUTPUT
+
+  # This job is used to test the weekly tests on the latest 2 stable versions of LanceDB,
```
jackye1995: nit: "latest 2 stable versions of Lance"

.github/workflows/weekly-tests.yml
```diff
@@ -0,0 +1,121 @@
...
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.12"
+      - uses: Swatinem/rust-cache@v2
+        with:
+          workspaces: python
+      - name: Install Lance
+        run: pip install -e ".[tests]"
+      - name: Run weekly tests
```
jackye1995: is there any good way we can report the result of the test to an email or slack/discord notification? Otherwise my worry is that we will forget it and just not care about the errors.

BubbleCal: we can send the results to a slack channel, will work on this in the next PR, I will keep eyes on it for now

python/python/tests/weekly/test_recurring.py
wjones127, line 1: What are some examples of issues this would catch?

BubbleCal: > What are some examples of issues this would catch?

this is inspired by the bug https://github.com/lancedb/lance/pull/3769, and https://github.com/lancedb/lance/pull/4108 and https://github.com/lancedb/lance/pull/4156 could be detected (not consistently).

I may add more write operations if this test wound take too long time so that it can catch more cases

PR #4200, opened 2025-07-11
docs: separate Quickstart into three tutorials

davidmyriel:
Only Python so far - will add ts and rs in a few days.

I merged Installation with Quickstart and made it a simple data management tutorial.

Versioning was taken out and turned into a separate tutorial with additional explanations.

Vector search and indexing is a separate topic now.


docs/src/start/quickstart.md
```diff
@@ -1,8 +1,31 @@
...
+The easiest way to get started with Lance is via our Python SDK `pylance`:
+
+```bash
+pip install pylance
+```
+
+For the latest features and bug fixes, you can install the preview version:
+
+```bash
+pip install --pre --extra-index-url https://pypi.fury.io/lancedb/ pylance
```
jackye1995: `lancedb/ pylance` extra space

docs/src/start/versioning.md
```diff
@@ -0,0 +1,90 @@
...
+
+# Versioning Your Datasets with Lance
+
+Lance supports versioning natively, allowing you to track changes over time. 
+
+In this tutorial, you'll learn how to append new data to existing datasets while preserving historical versions and access specific versions using version numbers or meaningful tags. You'll also understand how to implement proper data governance practices with Lance's native versioning capabilities.
+
+## Set Up Your Environment
+
+First, you should import the necessary libraries:
```
jackye1995: maybe include `pip install` instructions?

jackye1995: Actually maybe this should be a part of the quickstart setup

docs/src/start/search.md
jackye1995, line 1: we should name this file as `vector-search.md` then, since I think it's also important to talk about full text search, there is literally zero documentation about that right now.

docs/src/start/search.md
```diff
@@ -0,0 +1,258 @@
...
+1  575538  [2.0, 0.0, 1.0, 42.0, 3.0, 38.0, 152.0, 27.0, ...  76908.0
+2  241428  [11.0, 0.0, 2.0, 118.0, 11.0, 108.0, 116.0, 21...  92877.0
+...
+```
+
+Without the index, the search will scan throughout the whole dataset to compute the distance between each data point. For practical real-time performance with, you will get much better performance with an ANN index.
+
+## Build the Search Index
+
+You should create an ANN index to dramatically speed up vector search operations while maintaining high accuracy. Lance supports `IVF_PQ`, `IVF_HNSW_PQ` and `IVF_HNSW_SQ` indexes.
```
jackye1995: @BubbleCal is this list complete? I think we also have other combos like IVF_FLAT but not sure if that is still recommended for any case.

docs/src/start/search.md
```diff
@@ -0,0 +1,258 @@
...
+
+!!! note "Performance Note"
+    Your actual numbers will vary by your storage. These numbers are from local disk on an M2 MacBook Air. If you're querying S3 directly, HDD, or network drives, performance will be slower.
+
+## Tune the Search Parameters
+
+You need to adjust search parameters to balance between speed and accuracy, finding the optimal settings for your use case.
+
+The latency vs recall is tunable via:
+- **nprobes**: how many IVF partitions to search
```
jackye1995: this is out of date, now there is min and max nprobes

jackye1995: This raises a question. Is there a way we can execute the doc as a part of the CI to make sure it is up to date?

jackye1995: https://github.com/danielfrg/mkdocs-jupyter

docs/src/start/search.md
davidmyriel: We have flat, but im not sure if we want to cover every single detail in the Quickstart.

The Quickstart is supposed to be a minimal document to achieve victory and move on to detailed docs.

docs/src/start/search.md
jackye1995: (just to be clear this will be a lot of effort, we don't need to do it in this PR, but want to align the direction)

docs/src/start/search.md
BubbleCal: I'd like to post only IVF_PQ here to make it simple.
the full list is:
- IVF_FLAT
- IVF_SQ
- IVF_PQ
- IVF_HNSW_SQ
- IVF_HNSW_PQ

we will deprecate IVF_HSNW_PQ soon

jackye1995: yeah in that case let's just say something like "we will use IVF_PQ as an example". Once we have more comprehensive documentation of all the supported indexes, we can add another sentence of "see the full list of supported at XYZ"

docs/src/start/search.md
davidmyriel: understood

jackye1995 (approved):

PR #4207, opened 2025-07-11
feat(java): support get lance schema with field ids

### LINKED ISSUE #4202: Brings java lance schema into java module
For now, lance java(including spark) uses arrow schema to build dataset and retrieval from dataset.

However, when we need to config compression or other field level configurations, the field ids need to be known, which is stored in lance dominated schema. That's why we need to bring lance schema into java module, which has been done in python and rust modules.

The type transformation between java and rust could be dirty. Try best to make it light

majin1102:
close #4202

### DIFF
--- a/java/core/src/main/java/com/lancedb/lance/schema/LanceField.java
+++ b/java/core/src/main/java/com/lancedb/lance/schema/LanceField.java
@@ -1,0 +1,81 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.lancedb.lance.schema;
+
+import org.apache.arrow.vector.types.pojo.ArrowType;
+import org.apache.arrow.vector.types.pojo.DictionaryEncoding;
+import org.apache.arrow.vector.types.pojo.Field;
+import org.apache.arrow.vector.types.pojo.FieldType;
+
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+public class LanceField {
+
+  private final int id;
+  private final String name;
+  private final boolean nullable;
+  private final ArrowType type;
+  private final DictionaryEncoding dictionary;
+  private final Map<String, String> metadata;
+  private final List<LanceField> children;
+
+  LanceField(
+      int id,
+      String name,
+      boolean nullable,
+      ArrowType type,
+      DictionaryEncoding dictionary,
+      Map<String, String> metadata,
+      List<LanceField> children) {
+    this.id = id;
+    this.name = name;
+    this.nullable = nullable;
+    this.type = type;
+    this.dictionary = dictionary;
+    this.metadata = metadata;
+    this.children = children;
+  }
+
+  public int getId() {
+    return id;
+  }
+
+  public boolean isNullable() {
+    return nullable;
+  }
+
+  public ArrowType getType() {
+    return type;
+  }
+
+  public DictionaryEncoding getDictionary() {
+    return dictionary;
+  }
+
+  public Map<String, String> getMetadata() {
+    return metadata;
+  }
+
+  public Field asArrowField() {
+    List<Field> arrowChildren =
+        children.stream().map(LanceField::asArrowField).collect(Collectors.toList());
+    return new Field(name, new FieldType(nullable, type, dictionary, metadata), arrowChildren);
+  }
+
+  public List<LanceField> getChildren() {
+    return children;
+  }
+}
--- a/java/core/lance-jni/src/blocking_dataset.rs
+++ b/java/core/lance-jni/src/blocking_dataset.rs
@@ -14,8 +14,11 @@
 
 use crate::error::{Error, Result};
 use crate::ffi::JNIEnvExt;
+use crate::schema::convert_to_java_field;
 use crate::traits::{export_vec, import_vec, FromJObjectWithEnv, FromJString};
-use crate::utils::{extract_storage_options, extract_write_params, get_index_params, to_rust_map};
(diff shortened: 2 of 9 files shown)


majin1102:
Please take a look when you have time @jackye1995 @yanghua

java/core/src/main/java/com/lancedb/lance/schema/LanceField.java
```diff
@@ -0,0 +1,81 @@
...
+import org.apache.arrow.vector.types.pojo.ArrowType;
+import org.apache.arrow.vector.types.pojo.DictionaryEncoding;
+import org.apache.arrow.vector.types.pojo.Field;
+import org.apache.arrow.vector.types.pojo.FieldType;
+
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+public class LanceField {
```
jackye1995, line 26: I think we want to be exactly the same as the rust model:

```
pub struct Field {
    pub name: String,
    pub id: i32,
    // TODO: Find way to move these next three fields to private
    pub parent_id: i32,
    pub logical_type: LogicalType,
    pub metadata: HashMap<String, String>,
    pub encoding: Option<Encoding>,
    pub nullable: bool,

    pub children: Vec<Field>,

    /// Dictionary value array if this field is dictionary.
    pub dictionary: Option<Dictionary>,
    pub storage_class: StorageClass,
    pub unenforced_primary_key: bool,
}
```

this is missing some fields, and fields like `dictionary` is not matching exactly.

java/core/src/main/java/com/lancedb/lance/schema/LanceField.java
```diff
@@ -0,0 +1,81 @@
...
+
+  private final int id;
+  private final String name;
+  private final boolean nullable;
+  private final ArrowType type;
```
jackye1995, line 32: similar comment, I think we need a `logicalType` and a class `LogicalType`, and logical type should have a field `name` (corresponding to the unnamed string in `lance_core::datatypes::LogicalType`) to record the string type name.

westonpace: :cold_sweat: I'd really like to get rid of `lance_core::datatypes::LogicalType` from Rust.  It's a relic I don't think we need anymore.  Originally, we had a slightly different type system (because we couldn't support all Arrow data types), but at this point we implement most of the Arrow type system.  I think it's much simpler to just use Arrow data types and error when we hit a type we don't expect.

java/core/src/main/java/com/lancedb/lance/schema/LanceField.java
majin1102: Sure, I originally thought keeping the same with arrow fields with the additional info like field id could be right but conservative.  Let me look deep into the model details. 

BTW, how do we know the exactly primary key sequence. like:
primary key(A, B), we need to know A before B right? by field id or what others? @jackye1995

java/core/src/main/java/com/lancedb/lance/schema/LanceField.java
majin1102: > 😰 I'd really like to get rid of `lance_core::datatypes::LogicalType` from Rust. It's a relic I don't think we need anymore. Originally, we had a slightly different type system (because we couldn't support all Arrow data types), but at this point we implement most of the Arrow type system. I think it's much simpler to just use Arrow data types and error when we hit a type we don't expect.

Yeah, I found map and union types are not supported yet. Are they in the plan? @westonpace 
I also believe keeping ArrowType could be more friendly to engines and users so that they don't need to deal with two almost same type systems. What do you think? @jackye1995

java/core/src/main/java/com/lancedb/lance/schema/LanceField.java
majin1102: Hi @westonpace.

I didn't see encoding, dictonary,  unenforced_primary_key putting in the metadata when converting to an arrow field. 

I also noticed you said only field id is needed for lance schema.

What do you think of the upper three aligning with JAVA lance shema, are they unnecessary for schema readers? (especially dictionary, there may be an Array in it. I wish you could elaberase more about this)

westonpace: Dictionary and encoding are not important (only used for legacy 0.1 format).  `unenforced_primary_key` could be important.  We should probably include this.

> (especially dictionary, there may be an Array in it. I wish you could elaberase more about this)

In the 0.1 format we implemented dictionary at the table level.  We would store indices in the file and use the dictionary from the manifest to decode.

This field is unused in 2.0 (the current default) and 2.1 (the next default) and will be phased out at some point if we drop support for 0.1.

java/core/src/main/java/com/lancedb/lance/schema/LanceField.java
westonpace: > Yeah, I found map and union types are not supported yet. Are they in the plan?

I think we will add them when someone needs them.  Map will be easy.  Union is a little more tricky but we can support.

jackye1995: > I think it's much simpler to just use Arrow data types and error when we hit a type we don't expect.

I see, that makes sense. I thought we want to reserve the right to extend the Arrow types with that setup, but if we want to be fully Arrow then let's just use Arrow type.

java/core/src/main/java/com/lancedb/lance/schema/LanceField.java
jackye1995: > primary key(A, B), we need to know A before B right? by field id or what others? @jackye1995

I think primary key is expressed as a set, no ordering is needed. Because either `A, B` or `B, A` uniquely identifies a row.

java/core/src/main/java/com/lancedb/lance/schema/LanceField.java
westonpace: If we want to extend Arrow we can use Arrow's builtin extension type mechanism.

jackye1995 (approved):
looks good to me

java/core/src/main/java/com/lancedb/lance/schema/LanceField.java
majin1102: > > primary key(A, B), we need to know A before B right? by field id or what others? @jackye1995
> 
> I think primary key is expressed as a set, no ordering is needed. Because either `A, B` or `B, A` uniquely identifies a row.

Yes, makes sense.....this is datalake, not database...

PR #4212, opened 2025-07-11
fix: encode_batch should add external buffer after task run

allenanswerzq:
external_buffers passed to async tasks, without asyn task even starts run, take buffer from it is not correct


Xuanwo:
Hi, can you elaborate more details about this change? Or is there a reproduction for this bug?

PR #4213, opened 2025-07-12
fix: expand pyright coverage and fix type annotations

TaoKevinKK:
fix: expand pyright coverage and fix type annotations

- Add 8 files to pyright include list in pyproject.toml
- Fix type issues in log.py, types.py, util.py, and torch/__init__.py
- Resolve circular import between dependencies.py and torch module
- Improve type safety across multiple Python modules

Resolves TODO comment about expanding pyright file coverage.

### DIFF
--- a/python/python/lance/types.py
+++ b/python/python/lance/types.py
@@ -3,16 +3,14 @@
 
 from __future__ import annotations
 
-from typing import TYPE_CHECKING, Iterable, Optional, Union
-
-import pyarrow as pa
-from pyarrow import RecordBatch
-
-from . import dataset
-from .dependencies import _check_for_pandas
-from .dependencies import pandas as pd
+from typing import TYPE_CHECKING, Any, Iterable, Optional, Union
 
 if TYPE_CHECKING:
+    import pyarrow as pa  # type: ignore[import-untyped]
+    from pyarrow import RecordBatch
+    from . import dataset
+    from .dependencies import pandas as pd
+    
     ReaderLike = Union[
         pd.Timestamp,
         pa.Table,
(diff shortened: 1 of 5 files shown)


python/python/lance/types.py
```diff
@@ -53,15 +58,19 @@ def _casting_recordbatch_iter(
 
 
 def _coerce_reader(
-    data_obj: ReaderLike, schema: Optional[pa.Schema] = None
-) -> pa.RecordBatchReader:
+    data_obj: Any, schema: Optional["pa.Schema"] = None
```
BubbleCal, line 61: why do we need to change `ReaderLike` to `Any`?

python/python/lance/types.py
```diff
@@ -53,15 +58,19 @@ def _casting_recordbatch_iter(
...
     elif isinstance(data_obj, pa.Table):
         return data_obj.to_reader()
     elif isinstance(data_obj, pa.RecordBatch):
         return pa.Table.from_batches([data_obj]).to_reader()
-    elif isinstance(data_obj, dataset.LanceDataset):
+    elif (
+        hasattr(dataset, 'LanceDataset') 
+        and type(data_obj).__name__ == 'LanceDataset'
+        and hasattr(data_obj, 'scanner')
+    ):
```
BubbleCal, line 73: are these conditions equal to `isinstance(data_obj, dataset.LanceDataset)`?

python/python/lance/types.py
```diff
@@ -72,6 +81,7 @@ def _coerce_reader(
     elif (
         type(data_obj).__module__.startswith("polars")
         and data_obj.__class__.__name__ == "DataFrame"
+        and hasattr(data_obj, "to_arrow")
```
BubbleCal, line 84: is it possible an object satisfies the 2 conditions above but doesn't has `to_arrow` method?

PR #4214, opened 2025-07-12
docs: add complete blob writing example and fix take_blobs usage

TaoKevinKK:
This PR improves the blob documentation by adding a complete example for writing blob data and fixing the `take_blobs` API usage.

- **Added complete blob writing example**: Previously, the documentation only showed how to create a blob schema but didn't demonstrate how to actually write blob data to a dataset
- **Fixed `take_blobs` method usage**: Corrected the method call from `ds.take_blobs([1], "video")` to `ds.take_blobs("video", ids=[1])` to match the correct API signature
- **Improved user guidance**: Added clear comments instructing users to prepare their own video file instead of assuming a specific file exists

## Before
The blob documentation was incomplete and contained incorrect API usage, making it difficult for users to understand how to write blob data.

## After
Users now have a complete, working example that shows:
1. How to create a blob schema
2. How to read a video file and write it as blob data
3. How to correctly retrieve blob data using the proper API syntax

### DIFF
--- a/docs/src/guide/blob.md
+++ b/docs/src/guide/blob.md
@@ -25,6 +25,30 @@ schema = pa.schema(
 )
 ```
 
+To write blob data to a Lance dataset, create a PyArrow table with the blob schema and use `lance.write_dataset`:
+
+```python
+import lance
+
+# First, prepare a video file, e.g., "your_video.mp4"
+# Then read the video file content
+with open("your_video.mp4", 'rb') as f:
+    video_data = f.read()
+
+# Create table with blob data
+table = pa.table({
+    "id": [1],
+    "video": [video_data],
+}, schema=schema)
+
+# Write to Lance dataset
+ds = lance.write_dataset(
+    table,
+    "./youtube.lance",
+    schema=schema
+)
+```
+
 To fetch blobs from a Lance dataset, you can use `lance.dataset.LanceDataset.take_blobs`.
 
 For example, it's easy to use `BlobFile` to extract frames from a video file without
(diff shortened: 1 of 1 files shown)


docs/src/guide/blob.md
```diff
@@ -25,6 +25,30 @@ schema = pa.schema(
...
 ```
 
+To write blob data to a Lance dataset, create a PyArrow table with the blob schema and use `lance.write_dataset`:
+
+```python
+import lance
+
+# First, prepare a video file, e.g., "your_video.mp4"
+# Then read the video file content
+with open("your_video.mp4", 'rb') as f:
```
BubbleCal: it would be great to download a small video file in this example, so that people can run it directly.

TaoKevinKK: Thanks for the suggestion! I've updated the example to include automatic downloading of a sample video file.

docs/src/guide/blob.md
```diff
@@ -25,6 +25,37 @@ schema = pa.schema(
...
+To write blob data to a Lance dataset, create a PyArrow table with the blob schema and use `lance.write_dataset`:
+
+```python
+import lance
+
+# First, download a sample video file for testing
+# wget https://file-examples.com/storage/fe272c835c68711ef9aa603/2017/04/file_example_MP4_480_1_5MG.mp4
+import urllib.request
+urllib.request.urlretrieve(
+    "https://file-examples.com/storage/fe272c835c68711ef9aa603/2017/04/file_example_MP4_480_1_5MG.mp4",
```
BubbleCal: it seems this file is not public, ran into `urllib.error.HTTPError: HTTP Error 403: Forbidden` when running this example

TaoKevinKK: Thanks for the feedback! I've updated the example to use a more reliable approach

BubbleCal (approved):

PR #4218, opened 2025-07-13
chore: add some compile commands for convenient development

Jay-ju:
image
The compiled content is still relatively large, so some compilation cleanup commands have been added here.

### DIFF
--- a/python/Makefile
+++ b/python/Makefile
@@ -20,6 +20,14 @@ format: format-python
 	cargo fmt
 .PHONY: format
 
+build:
+	maturin develop
+.PHONY: build
+
+clean:
+	rm -rf ./target
+.PHONY: clean
+
 format-python:
 	ruff format python
 	ruff check --fix python


python/Makefile
```diff
@@ -20,6 +20,14 @@ format: format-python
 	cargo fmt
 .PHONY: format
 
+build:
+	maturin develop
```
jackye1995, line 24: @Xuanwo is adding `uv` support in #4221 , might be worth directly adding it there or waiting for that merged first?

Xuanwo: I think it's better to add it here directly, so we won't need to keep any PRs pending.

python/Makefile
```diff
@@ -20,6 +20,14 @@ format: format-python
...
+.PHONY: build
+
+clean:
```
Xuanwo, line 27: I wish we can remove this after `cargo gc` been stablized.

PR #4219, opened 2025-07-14
feat(java): brings transaction api to Java module and support project

### LINKED ISSUE #4201: Brings transaction api for java module
This is necessary in many distributed work like compaction and projection for big data engines

majin1102:
Close #4201


java/core/lance-jni/src/transaction.rs
```diff
@@ -0,0 +1,115 @@
...
+        let jstring = JString::from(message);
+        let rust_str = env.get_string(&jstring)?.to_string_lossy().into_owned();
+        Some(rust_str)
+    };
+    Ok(Transaction {
+        read_version: read_ver as u64,
+        uuid,
+        operation: op,
+        blobs_op,
+        tag: message,
```
jackye1995: tag should be described as `tag` also in Java, not a new term `message`

java/core/src/main/java/com/lancedb/lance/Dataset.java
```diff
@@ -265,6 +265,27 @@ public static native Dataset commitOverwrite(
...
+  }
+
+  /**
+   * Create a new operation at current version for the dataset. The dataset itself will not refresh
+   * after the operation committed.
+   *
+   * @param message The description of the operation.
+   * @return A new instance of {@link Transaction} linked to the opened dataset.
+   */
+  public Transaction newTransaction(String message) {
```
jackye1995: why expose a specific one with message? We can just expose a 

```
public Transaction.Builder newTransactionBuilder() {
    return new Transaction.Builder(this);
  }
```

java/core/lance-jni/src/transaction.rs
majin1102: my conern is that when I set a tag and a dataset into one transaction. how to make people know the tag IS NOT version related and it actually acts like a transaction extra label？
if I understand wrong，please correct me.

java/core/src/main/java/com/lancedb/lance/Dataset.java
majin1102: I think what I aim to do is directly get a transaction obj by newTransaction. Then the usage could be simple.

 Your suggestion seems more flexible. I will follow it

java/core/lance-jni/src/transaction.rs
jackye1995: Yeah this is a confusing part indeed. I think we can start with what Python does, just omit the `tag` field for now so it is not exposed to end users. Even if we want to commit with a message, as we discussed previously, we want to do that as a part of the Transaction properties workstream instead, so we should not hijack the tag field for that purpose. What do you think?

majin1102: > Yeah this is a confusing part indeed. I think we can start with what Python does, just omit the `tag` field for now so it is not exposed to end users. Even if we want to commit with a message, as we discussed previously, we want to do that as a part of the Transaction properties workstream instead, so we should not hijack the tag field for that purpose. What do you think?

Sure, I'm OK with this approach

java/core/src/main/java/com/lancedb/lance/Dataset.java
majin1102: > why expose a specific one with message? We can just expose a
> 
> ```
> public Transaction.Builder newTransactionBuilder() {
>     return new Transaction.Builder(this);
>   }
> ```

As discussed above,  there's no need to expose the api with message.
I think a newTransaction is enough for now. For further properties usage, I think we could just:

```
 txn = committedDataset.newTransaction()
        .newProject(new Schema(fieldList))
        .setProperties(.....)
        .commit()
```
This could be concise
WDYT @jackye1995

jackye1995: > .newProject(new Schema(fieldList))

what does it mean to have a project for a transaction?

jackye1995: oh nvm it's the Project operation, I misunderstood

jackye1995:
```
.newProject(new Schema(fieldList))
        .setProperties(.....)
```

so this is mutating the transaction after it is built? It feels a bit weird to me, because with a builder usually the assumption is the object is immutable after built. 

It feels like this is trying to be similar to the Iceberg transaction java API? (sorry if it's not, but the terminologies used are really similar).

java/core/src/main/java/com/lancedb/lance/operation/Projection.java
```diff
@@ -0,0 +1,58 @@
...
+import org.apache.arrow.c.ArrowSchema;
+import org.apache.arrow.c.Data;
+import org.apache.arrow.memory.BufferAllocator;
+import org.apache.arrow.vector.types.pojo.Schema;
+
+/**
+ * Project to a new schema. This only changes the schema, not the data. This operation could be used
+ * to: 1. add/remove columns. The data will be removed after compaction. 2. modify column positions
+ */
+public class Projection implements Operation {
```
jackye1995: let's align the class name with the actual enum name, like `Project` in this case

java/core/src/main/java/com/lancedb/lance/Dataset.java
majin1102: > It feels like this is trying to be similar to the Iceberg transaction java API? (sorry if it's not, but the terminologies used are really similar).

I must say I am quite influenced by Iceberg, but not this one. I'm trying to avoid 

xxxxx
     xxxx
    .build()
    .newXXXX
     xxxx
     .commt()

Cause in my mind build() or commit() should be the last calling in one statement. If directly get a transaction from newTransaction(), that makes one concise statement to me. While this may be limited when we want to extend build properties and setProperties() is not clear for operation or transaction. 

Let me just return the builder

majin1102: The newXXX usage is mainly for the multi-statement extension. #3734 

Maybe it is a little earlier. I did some job for multi-statement prototype because we hava product scenario(update a whole column)

If you prefer the rust one-statement style. I will be OK with that

jackye1995: If we want to fully align with the way we run transaction in rust, I think we should probably introduce the commit builder concept in Java, because that is what is used to do multi-statement transaction with `execute_batch`. 

With that model, each transaction in Java should be kept simple, just builder style to accumulate all the information needed, and then fed into the commit builder to execute one transaction, or multiple transactions.

java/core/src/main/java/com/lancedb/lance/operation/Project.java
```diff
@@ -0,0 +1,58 @@
...
+/**
+ * Project to a new schema. This only changes the schema, not the data. This operation could be used
+ * to: 1. add/remove columns. The data will be removed after compaction. 2. modify column positions
+ */
+public class Project implements Operation {
+
+  private final Schema schema;
+  private final ArrowSchema cSchema;
+
+  public Project(Schema schema, BufferAllocator allocator) {
```
jackye1995: should we also use builder style here? Because for more complicated objects there could be many inputs, and we should keep all operations consistent.

java/core/src/main/java/com/lancedb/lance/operation/Project.java
```diff
@@ -0,0 +1,58 @@
...
+    this.schema = schema;
+    this.cSchema = ArrowSchema.allocateNew(allocator);
```
jackye1995: I did not fully understand, why have a `cSchema`?

java/core/src/main/java/com/lancedb/lance/Dataset.java
jackye1995: I would suggest, let's just do the multi-statement case implementation here to see how things work end to end, have 2 different operations and commit them. It would be hard to imagine how things extend to multi-statement with just this PR.

majin1102: > If we want to fully align with the way we run transaction in rust

Actually I did consider alignment with the concepts in rust as a goal, not including style and calling ways. Be honest, I really don't think the rust transaction api is very friendly. Let's imagine when we want to operate one update-column transaction:

```
dataset.newTransaction()
     .newProject(xxx)
     .mergeColumn(xxx)
     .commit()

```
Or

```
List<Operation> opList = new ArrayList<>();
opList.add(ProjectBuilder.newBuilder().schema(xxxx).build());
opList.add(mergeColumn.newBuilder().column(xxxx).build());
Transaction tx = TransactionBuilder.builder(dataset).operations(opList).build();
CommitBuilder commitBuilder = new CommitBuilder(xxxx);
commit.executeBatch(tx);
```

Honestly I'm not very comfortable to deal with the code below for unnecesary concepts and POJOS to construct and understand. I think operation and commit related POJO could be hidden behind(one-statment or multi, execute or execute_batch I don't think that should be an API thing) transaction api so that caller could be simplier and elegent.

I understand rust has its native style and I respect that. I think JAVA could provide a simpler style.

> I think we should probably introduce the commit builder concept in Java

I'm not sure we should do that according to the JNI pattern for now. I think the current pattern is we build the JObject and use it in rust jni code.

majin1102: And I don't think newXXXX() should be one-one with operation.

For example, right now update_meta is one opereation including updating table config, replacing schema config and replacing field configs. It could be used as

```
dataset.newTransaction()
     .newUpdateConfig(xxx)
     .newReplaceSchemaConfig(xxx)
     .newReplaceFieldsConfig(xxxx)
     .commit()
```
When calling like this, the underlying is one operation.

People don't need to know how to construct operations. They could be hidden and Transaction api could provide the better calling way to construct them.

majin1102: > I would suggest, let's just do the multi-statement case implementation here to see how things work end to end, have 2 different operations and commit them. It would be hard to imagine how things extend to multi-statement with just this PR.

You mean implement more opreations and see the usage in uni-test maybe?

jackye1995: Let's see if we can simply what you wrote. Firstly there is only one operation per transaction you don't need a list. Secondly, secondly, we can add syntax sugar to say dataset.newCommit().addTxn(...).addTxn(...).execute. Would that be better?

java/core/src/main/java/com/lancedb/lance/operation/Project.java
majin1102: This cSchema is to pass the schema to rust jni as zero-copy.

Do we have better ways or cSchema is not a good name?

java/core/src/main/java/com/lancedb/lance/Dataset.java
majin1102: > Let's see if we can simply what you wrote. Firstly there is only one operation per transaction you don't need a list. Secondly, secondly, we can add syntax sugar to say dataset.newCommit().addTxn(...).addTxn(...).execute. Would that be better?

I think better than the build() storm code. 

1. dataset.newCommit().execute() is one transaction or multi? I mean the ACID granularity. I think what we want is one execute() one ACID which means one transaction? but what you are calling is addTxn and andTx. That could make things confusing.
2. The key differenece from mine(maybe we could call Iceberg like?)  I think is use Commit to replace Transaction, use transaction to replace operation(which I think cause semantics confusing). and let callers build operations themself. I assume the key point is to make operations more exposed to upper modules for flexibilities concern? If so I think we can discuss the risks that transaction wraps operations. Or we just want to strictly align with rust style as one goal? Or be different from Iceberg(seriously what I am afraid is to be not as good as Iceberg in this case not be like it :()

java/core/src/main/java/com/lancedb/lance/operation/Project.java
majin1102: Yes, that makes sense to me

java/core/src/main/java/com/lancedb/lance/operation/Projection.java
majin1102: Copy

java/core/src/main/java/com/lancedb/lance/Dataset.java
majin1102: > Let's see if we can simply what you wrote. Firstly there is only one operation per transaction you don't need a list. Secondly, secondly, we can add syntax sugar to say dataset.newCommit().addTxn(...).addTxn(...).execute. Would that be better?

One more issue is the read version, addTxn() more than once could introduce more than one read version, which should not be involved in multi-statement trasaction. We could check and throw exceptions. But I think the whole thing is just redundant and unnecessary.

majin1102:
> Let's see if we can simply what you wrote. Firstly there is only one operation per transaction you don't need a list. Secondly, secondly, we can add syntax sugar to say dataset.newCommit().addTxn(...).addTxn(...).execute. Would that be better?

What's our next step?  I'm ready for this newCommit and addTransaction interface. I think eventually they do the same things except the priority is concept consistency or simplicity. I agree that the former takes priority

java/core/src/main/java/com/lancedb/lance/Dataset.java
```diff
@@ -266,6 +266,16 @@ public static native Dataset commitOverwrite(
...
       Map<String, String> storageOptions);
 
+  /**
+   * Create a new transaction builder at current version for the dataset. The dataset itself will
+   * not refresh after the transaction committed.
+   *
+   * @return A new instance of {@link Transaction.Builder} linked to the opened dataset.
+   */
+  public Transaction.Builder newTransactionBuilder() {
+    return new Transaction.Builder(this).readVersion(version());
```
jackye1995, line 276: I think you should be able to also set the UUID?

jackye1995: oh nvm it is set in the builder

java/core/src/main/java/com/lancedb/lance/Transaction.java
```diff
@@ -0,0 +1,123 @@
...
+
+  public Operation blobsOperation() {
+    return blobOp;
+  }
+
+  public Map<String, String> writeParams() {
+    return writeParams;
+  }
+
+  public Transaction newProject(Schema schema) {
```
jackye1995: Given that we have not fully decided how we willl chain multiple transaction, I feel it is better that we don't do a syntax like this to give an illusion that we can "chain transactions" (We can always add later). 

Also typically with a builder people expect the object to be immutable after build, and a `newXXX` to create a new object that inherits some stats of the source object. But here ` Transaction txn = dataset.newTransactionBuilder().build().newProject(new Schema(fieldList));` we are actually mutating the state of the already built transaction, that is also a bit confusing.

As a simple first step integration, can we just in the builder do:

```java
Transaction txn = dataset.newTransactionBuilder()
.operation(Project.builder().schema(schema).allocator(dataset.allocator).build())
.build()
```

what do you think?

java/core/src/main/java/com/lancedb/lance/Dataset.java
majin1102: yes. I think the uuid generating should be align with rust logics. Openness to set may be a risk

java/core/src/main/java/com/lancedb/lance/Transaction.java
majin1102: I think this is resonable. I think there's a note for this case:

```
Transaction txn = dataset.newTransactionBuilder()
.operation(Project.builder(dataset).schema(schema).build())
.build()
```

1. Avoid users or engines to set their own allocator or create a new one(avoid danging allocators)
2. Make the dataset a neligable paramter to build the project operation
3. The dataset now is necessary for JNI usage

I think this situation may suite some other operations cases. WDYT

majin1102: oh，never mind
 
The dataset is necessary for transaction, not operation. Project did need an allocator. but it didn't release.

I think We'd better give it a managed allocator(do not expose,just pass transaction dataset allocator). Or create a new one and release it after commit(creating allocator is heavy as said). 

What do you prefer?

jackye1995: > I think We'd better give it a managed allocator(do not expose,just pass transaction dataset allocator).

+1

jackye1995 (approved):
looks good to me!

jackye1995:
looks like some tests are faling, let me know when that is fixed I can go merge it!

PR #4220, opened 2025-07-14
Fix: Handle case-insensitive 'is not null' filter for column[0]

marygmalana:
When I was creating index with accelerator, I was getting raise NotImplementedError(f"Can't yet run filter <{filter}> in-memory") error. But apparently it's because my column name has capital characters.

### DIFF
--- a/python/python/lance/sampler.py
+++ b/python/python/lance/sampler.py
@@ -143,7 +143,7 @@ def _filtered_efficient_sample(
             limit=(end - start),
             batch_size=shard_size,
         )
-        if len(columns) == 1 and filter.lower() == f"{columns[0]} is not null":
+        if len(columns) == 1 and filter.lower() == f"{columns[0]}.lower() is not null":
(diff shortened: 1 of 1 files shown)


python/python/lance/sampler.py
```diff
@@ -143,7 +143,7 @@ def _filtered_efficient_sample(
             limit=(end - start),
             batch_size=shard_size,
         )
-        if len(columns) == 1 and filter.lower() == f"{columns[0]} is not null":
+        if len(columns) == 1 and filter.lower() == f"{columns[0]}.lower() is not null":
```
Xuanwo, line 146: Hi, it looks like we are comparing with `NAME.lower() is not null`, should we inlcude `.lower()` in the `{columns[0].lower()}`?

wjones127:
Thanks for making a PR. Looks like we got this issue fixed in https://github.com/lancedb/lance/pull/4278/files

Issue #4138, opened 2025-07-04
Migrate our python workflow to uv


Xuanwo:
We use `pip install -r requirements.txt` in many places. Is there a specific reason for this? If not, would it make sense to switch to `uv` instead? `uv` is fast and stable, and it can help us achieve reproducible Python dependencies.

westonpace:
I think switching to uv would be fine.

jackye1995:
I always hear uv is great, and I also use it myself (because of the hype), but never got the time to know all the details of why. I think currently we just use `requirements.txt` in docs and benchmarks, because they might have conflicting dependencies, and people might not be interested in installing all dependencies if they just need to work on one thing. Is that something uv can solve?

Xuanwo:
>  I think currently we just use requirements.txt in docs and benchmarks, because they might have conflicting dependencies

I will take a look on this.

But in the worst cases, we can maintain two separate `pyproject.toml` files, which is no different from our current approaches.

> people might not be interested in installing all dependencies if they just need to work on one thing. Is that something uv can solve?

I think we can organize our dependencies into groups: https://docs.astral.sh/uv/concepts/projects/dependencies/#dependency-groups

```toml
[dependency-groups]
dev = [
  "pytest"
]
lint = [
  "ruff"
]
```

PR #4221, opened 2025-07-14
chore: use uv for dep management

Xuanwo:
Close https://github.com/lancedb/lance/issues/4138

This PR only introduces `uv`, and in subsequent ones, we will gather any missing dependencies in `pyconfig.toml` like `uv pip install --system -r docs/requirements.txt`.

### DIFF
--- a/.github/workflows/build_linux_wheel/action.yml
+++ b/.github/workflows/build_linux_wheel/action.yml
@@ -34,6 +34,7 @@ runs:
         target: x86_64-unknown-linux-gnu
         manylinux: ${{ inputs.manylinux }}
         args: ${{ inputs.args }}
+        container: off
         before-script-linux: |
           set -e
           yum install -y openssl-devel \
@@ -52,6 +53,7 @@ runs:
           -e CC=clang
           -e CXX=clang++
         args: ${{ inputs.args }}
+        container: off
         before-script-linux: |
           set -e
           yum install -y openssl-devel clang \
@@ -67,6 +69,7 @@ runs:
         target: aarch64-unknown-linux-gnu
         manylinux: ${{ inputs.manylinux }}
         args: ${{ inputs.args }}
+        container: off
         before-script-linux: |
           set -e
           yum install -y openssl-devel clang \
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -259,6 +259,6 @@ jobs:
       - uses: ./.github/workflows/build_linux_wheel
       - name: Install dependencies
         run: |
-          uv --no-project sync --group ray
+          uv sync --no-project --group ray
           uv pip install torch --index-url https://pypi.org/simple --extra-index-url https://download.pytorch.org/whl/cpu
       - uses: ./.github/workflows/run_integtests


python/DEVELOPMENT.md
```diff
@@ -7,7 +7,7 @@ This project is built with [maturin](https://github.com/PyO3/maturin).
 It can be built in development mode with:
 
 ```shell
-maturin develop
+uv run maturin develop
```
jackye1995, line 10: we also need to update the website contributing pgae related to python development (maybe we should just combine the 2 pages?)

python/Makefile
```diff
@@ -5,33 +5,33 @@ else
 endif
```
jackye1995, line 5: we should add a `make lock` in this case

.gitignore
```diff
@@ -92,7 +92,9 @@ target
 .ccls-cache/
 
 python/venv
+python/.venv
 test_data/venv
+test_data/.venv
```
jackye1995, line 97: should we also just add `docs/.venv` instead of doing `uv pip install --system -r docs/requirements.txt`?

PsiACE (commented):
Using `[dependency-groups]` in pyproject.toml for uv to manage development, documentation, and test dependencies is a great idea. 

for big project, use uv workspace

Xuanwo:
Thank you, will take a look over `uv workspace`

python/Makefile
Xuanwo: uv will sync lock by default so we don't need to worry about it.

jackye1995:
@Xuanwo I just moved to use uv workspace for lance namespace: https://github.com/lancedb/lance-namespace

It works pretty well with a shared `uv.lock`, everything is installed with a single `uv sync --all-packages`, seems pretty nice so far.

Xuanwo:
cc @jackye1995 to take another look

python/pyproject.toml
```diff
@@ -59,11 +59,53 @@ tests = [
...
+# Development and linting tools
+dev = [
+    "maturin>=1.4",
+    "ruff==0.11.2",
+    "pyright",
+    "ipython",
+]
+
+# Optional heavy dependencies
+ray = ["ray[data]<2.38; python_version<'3.12'"]
```
wjones127, line 98: I hope eventually we can put `tensorflow`, `duckdb`, `polars` and others over here. I think our "core test dependencies" really should just by `pytest` and `pytest-benchmark`.

jackye1995: what would be the difference? When we run a test that involves one of these dependencies, we still would need to install it. Do you expect we also separate out those tests agaisnt tensorflow, duckdb, etc. so normal pytest only tests core functionalities?

wjones127: Yeah I think it would be nice to have `test` just cover all core tests.

Then an `integration-tests` dependency list to install heavy depencies like `tensorflow` and `torch`.

It's pretty easy in `pytest` to skip a test if a dependency is not found using `pytest.importorskip()`: https://docs.pytest.org/en/stable/how-to/skipping.html#skipping-on-a-missing-import-dependency

wjones127 (approved):
This looks great!

Xuanwo:
Will work on a new one.

PR #4223, opened 2025-07-14
fix: phrase query may miss some results

BubbleCal:
this fixes 2 bugs:
- AND query and phrase query may miss results because of float precision (summing up floats in diff order may get diff results). With this, WAND would just check whether the query hits all posting lists, doesn't rely on the score threshold
- there's a bug may cause we get less positions after decompressing

fix https://github.com/lancedb/lancedb/issues/2483#issuecomment-3057589374

### DIFF
--- a/rust/lance-index/src/scalar/inverted/wand.rs
+++ b/rust/lance-index/src/scalar/inverted/wand.rs
@@ -337,8 +337,8 @@ impl<'a, S: Scorer> Wand<'a, S> {
                 continue;
             }
 
-            if let Some(slop) = params.phrase_slop
-                && !self.check_positions(slop as i32)
+            if params.phrase_slop.is_some()
+                && !self.check_positions(params.phrase_slop.unwrap() as i32)
             {
                 self.move_preceding(pivot, doc.doc_id() + 1);
                 continue;


rust/lance-index/src/scalar/inverted/index.rs
```diff
@@ -1489,11 +1476,22 @@ impl PostingListBuilder {
...
 
-        if let Some(mut position_builder) = position_builder {
+        if let Some(positions) = self.positions.as_ref() {
+            let mut position_builder = ListBuilder::new(ListBuilder::with_capacity(
+                LargeBinaryBuilder::new(),
+                length,
+            ));
+            for index in 0..length {
+                let positions = positions.get(index);
+                let compressed = compress_positions(positions)?;
```
Xuanwo: I spent sometime to understand the code here. How about using different names for `positions` and the value returned by `positions.get(index)`?

rust/lance-index/src/scalar/inverted/index.rs
```diff
@@ -1548,7 +1546,7 @@ impl PositionBuilder {
 
     pub fn size(&self) -> usize {
         std::mem::size_of::<u32>() * self.positions.len()
-            + std::mem::size_of::<i32>() * (self.offsets.len() - 1)
+            + std::mem::size_of::<i32>() * self.offsets.len()
```
Xuanwo, line 1549: Should `u32` have the same size with `i32`? Can we just `self.positions.len()+self.offsets.len()`?

rust/lance-index/src/scalar/inverted/wand.rs
```diff
@@ -345,6 +336,14 @@ impl<'a, S: Scorer> Wand<'a, S> {
                 self.move_preceding(pivot, doc.doc_id() + 1);
                 continue;
             }
+
+            if params.phrase_slop.is_some()
+                && !self.check_positions(params.phrase_slop.unwrap() as i32)
```
Xuanwo, line 341: Let's try new syntex: `if let Some(phrase_slop) = params.phrase_slop && !self.check_positions(phrase_slop as i32)`

rust/lance-index/src/scalar/inverted/wand.rs
```diff
@@ -453,6 +452,14 @@ impl<'a, S: Scorer> Wand<'a, S> {
     // are greater than or equal to the threshold
     #[instrument(level = "debug", skip_all)]
     fn find_pivot_term(&self) -> Option<usize> {
+        if self.operator == Operator::And {
+            // for AND query, we always require all terms to be present in the document,
+            // so the pivot is always the last term as long as no posting list is exhausted
+            if self.postings.len() == self.num_terms {
+                return Some(self.num_terms - 1);
```
westonpace, line 459: Is there any concern that `self.num_terms` could be zero?

rust/lance-index/src/scalar/inverted/wand.rs
```diff
@@ -546,7 +551,16 @@ impl<'a, S: Scorer> Wand<'a, S> {
                 let (Some(last), Some(next)) = (last, next) else {
                     return false;
                 };
-                max_relative_pos = Some(std::cmp::max(last, next) - slop);
+
+                let move_to = if last > next {
+                    last
+                } else {
+                    std::cmp::max(last + 1, next - slop)
```
Xuanwo, line 558: nit: how about using `(last+1).max(next - slop)` for better reading?

westonpace (approved):
Not sure I fully understand wand but looks good from a high level overview

rust/lance-index/src/scalar/inverted/wand.rs
```diff
@@ -546,7 +551,16 @@ impl<'a, S: Scorer> Wand<'a, S> {
...
+                };
+                max_relative_pos = match max_relative_pos {
```
Xuanwo: nit: I'm thinking of something like `max_relative_pos.unwrap_or(move_to).max(move_to)`

rust/lance-index/src/scalar/inverted/wand.rs
BubbleCal: oh good point, let me try it out

Xuanwo (commented):
Mostly LGTM, only some nits.

rust/lance-index/src/scalar/inverted/wand.rs
BubbleCal: added a test case for this, it passed, FTS search would check whether the posting lists are empty before performing WAND

rust/lance-index/src/scalar/inverted/index.rs
BubbleCal: just to keep it for not confusing people

rust/lance-index/src/scalar/inverted/wand.rs
BubbleCal: it's not equal
it should be `last` if `last > next`

Xuanwo: Oh, sorry for the wrong code range. I'm just referring to `std::cmp::max(last + 1, next - slop)`

rust/lance-index/src/scalar/inverted/wand.rs
BubbleCal: I found `max_relative_pos.max(Some(move_to))` works!

rust/lance-index/src/scalar/inverted/wand.rs
BubbleCal: oh I like this, fixed!

rust/lance-index/src/scalar/inverted/index.rs
BubbleCal: yeah it confuses, modified

rust/lance-index/src/scalar/inverted/wand.rs
BubbleCal: it seems we haven't upgrade toolchain in CI, changed it back for now

rust/lance-index/src/scalar/inverted/wand.rs
Xuanwo: Cool!

Xuanwo (approved):

PR #4229, opened 2025-07-14
fix: handle index statistics for system indexes

jackye1995:
Currently calling `index_statistics` on system indices cause failure because index out of bound for getting column.

### DIFF
--- a/rust/lance/src/index.rs
+++ b/rust/lance/src/index.rs
@@ -711,7 +711,7 @@ impl DatasetIndexExt for Dataset {
             let index = self
                 .open_frag_reuse_index(&NoOpMetricsCollector)
                 .await?
-                .unwrap();
+                .expect("FragmentReuse index does not exist");
             return serde_json::to_string(&index.statistics()?).map_err(|e| Error::Index {
                 message: format!("Failed to serialize index statistics: {}", e),
                 location: location!(),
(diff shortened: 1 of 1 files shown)


rust/lance/src/index.rs
```diff
@@ -707,6 +707,28 @@ impl DatasetIndexExt for Dataset {
             });
         }
 
+        if index_name == FRAG_REUSE_INDEX_NAME {
+            let index = self
+                .open_frag_reuse_index(&NoOpMetricsCollector)
+                .await?
+                .unwrap();
```
westonpace: Should we handle the "no frag reuse index" case better?  Maybe return a specific error message like "there is no fragment reuse index"?

rust/lance/src/index.rs
```diff
@@ -707,6 +707,28 @@ impl DatasetIndexExt for Dataset {
...
+                message: format!("Failed to serialize index statistics: {}", e),
+                location: location!(),
+            });
+        }
+
+        if index_name == MEM_WAL_INDEX_NAME {
+            let index = self
+                .open_mem_wal_index(&NoOpMetricsCollector)
+                .await?
+                .unwrap();
```
westonpace: Same question, should we avoid unwrap here?

westonpace (approved):
Might be nice to avoid the unwraps but maybe I'm misunderstanding something.  If a user asks for an index and it doesn't exist we should give some meaningful error and not crash.

rust/lance/src/index.rs
jackye1995: Sure. I was assuming when this is invoked there is an assumption that the index exists, but probably better to make it explicit

PR #4230, opened 2025-07-15
feat: Add utoipa for structs

LuQQiu:
Utoipa are used for auto generating open api.
All the structs in request and repsonses are needed to add "ToSchema" annotation.
However, openapi cannot deal with recursive struct, use "#[schema(value_type = Object)]" to work around the recursive dependencies, e.g. the FTS query


rust/lance-index/src/scalar/inverted/query.rs
```diff
@@ -119,6 +120,8 @@ pub enum FtsQuery {
     // compound queries
     Boost(BoostQuery),
     MultiMatch(MultiMatchQuery),
+    // Openapi generator does not support recursive schema
+    #[schema(value_type = Object)]
```
LuQQiu: @BubbleCal could you help provide some guidance on:
if BoostQuery and BooleanQuery cannot be used for FTS, how much of the FTS query functionality would be affected?

rust/lance/src/arrow/json.rs
```diff
@@ -13,13 +13,15 @@ use serde::{Deserialize, Serialize};
...
+use utoipa::ToSchema;
 
 /// JSON representation of an Apache Arrow [DataType].
-#[derive(Serialize, Deserialize, Debug)]
+#[derive(Serialize, Deserialize, Debug, ToSchema)]
 pub struct JsonDataType {
     #[serde(rename = "type")]
     type_: String,
     #[serde(skip_serializing_if = "Option::is_none")]
+    #[schema(value_type = Object)]
```
LuQQiu: Schema also has recursive dependencies

jackye1995:
Sorry I misclicked the approve button.... Still reviewing...

jackye1995:
> openapi cannot deal with recursive struct, use "#[schema(value_type = Object)]" to work around the recursive dependencies, e.g. the FTS query

Is this a limitation of OpenAPI, or a limitation of utoipa? Because when I was doing Iceberg OpenAPI it does have recursive structure and it seems fine. For example for schema: https://github.com/apache/iceberg/blob/main/open-api/rest-catalog-open-api.yaml#L2086-L2160

If this is a limitation of utoipa, I think we should just hand-craft the OpenAPI definition so that the generated client can have proper models generated. Otherwise it will be complicated to extend upon the generated client.

rust/lance-index/src/scalar/inverted/query.rs
BubbleCal: Boost may not affect much
But Boolean really matters, we can't do query like A and/or B without Boolean query

LuQQiu:
@jackye1995 Confirmed that it's a limitation of utoipa, will hand craft those recursive structs

jackye1995:
Does that mean we will close this PR and directly add the definition in lance namespace?

LuQQiu:
@jackye1995 Please take a look the updated hard-crafted recursive structs, validated that it can correctly generate the openapi.yml script

wkalt:
I'm not sure it makes sense to add a utoipa dependency to lancedb. There are alernatives to utoipa right? Shouldn't consumers of lancedb should be able to choose among them? Or maybe I am misunderstanding

jackye1995:
wow I did not expect you to implement here, I thought we will just do the hard-coded OpenAPI spec in https://github.com/lancedb/lance-namespace/pull/134. I am a bit confused now, was that generated by this code here?

LuQQiu:
Close this PR, this is server side changes

Issue #4224, opened 2025-07-14
Implement byte stream split encoding for double.


Xuanwo:
[`byte stream split`](https://parquet.apache.org/docs/file-format/data-pages/encodings/#byte-stream-split-byte_stream_split--9) is an encoding method supported by Parquet. It works well for float and double types because the bits of doubles tend to have similar patterns in the same positions. This is beneficial for subsequent compression algorithms like LZ4 and Snappy.

PR #4236, opened 2025-07-15
feat: implement byte stream split for doubles

Xuanwo:
close https://github.com/lancedb/lance/issues/4224

This PR only adds this feature and not enable by default.

## Benchmark

test cases:

```shell
Flat Schema: uuid (int64) + 3827 double columns
Data pattern: 40% zeros, 40% common values, 20% random values
```

bss + lz4

| Rows | Lance | Parquet |
|------|-------------|---------|
| 1000 | 21807702 (1.40x) | 11297945 (**2.71x**) |
| 10000 | 201906557 (1.52x) | 109407682 (**2.80x**) |
| 100000 | 2003744202 (1.53x) | 1133950053 (**2.70x**) |

bss + zstd

| Rows | Lance | Parquet |
|------|-------------|---------|
| 1000 | 14432739 (2.12x) | 11297945 (**2.71x**) |
| 10000 | 123655678 (2.48x) | 109407682 (**2.80x**) |
| 100000 | 1217373281 (2.52x) | 1133950053 (**2.70x**) |


rust/lance-encoding/src/encodings/physical/byte_stream_split.rs
```diff
@@ -0,0 +1,385 @@
...
+//! - Byte stream 3: `[3F, 40, 40, 40]` (all fourth bytes)
+//!
+//! Output: `[00, 00, 00, 00, 00, 00, 00, 00, 80, 00, 40, 80, 3F, 40, 40, 40]`
+//!
+//! ## Compression Benefits
+//!
+//! BSS itself doesn't compress data - it reorders it. The compression benefit
+//! comes when BSS is combined with general-purpose compression (e.g., LZ4):
+//!
+//! 1. **Timestamps**: Sequential timestamps have similar high-order bytes
```
alamb, line 40: this is a great shout out, btw

jackye1995: @Xuanwo does that mean we should not limit this to floating-point values? Currently the above says "optimized for floating-point data compression"

Xuanwo: Maybe, but I suspect other encodings might perform better with u64 or u32. As far as I know, we rarely see BSS used outside of floating point. In fact, it's never actually been proven true 🤣

jackye1995: yeah technically there is delta encoding for ints (actually did we implement that? I think no). Would be curious to see how these actually compare, it feels like BSS has the advantage that it can better leverage SIMD.

Xuanwo: > delta encoding for ints (actually did we implement that? I think no).

Not implemented yet.

> Would be curious to see how these actually compare, it feels like BSS has the advantage that it can better leverage SIMD.

Makes sense. Will give it a try.

rust/lance-encoding/src/encodings/physical/value.rs
```diff
@@ -819,6 +819,12 @@ pub(crate) mod tests {
         #[values(LanceFileVersion::V2_0, LanceFileVersion::V2_1)] version: LanceFileVersion,
     ) {
         for data_type in PRIMITIVE_TYPES {
+            // skip tests for types using byte stream split.
+            //
+            // FIXME: we should have better config to control this.
+            if version == LanceFileVersion::V2_1 && data_type.is_floating() {
+                continue;
+            }
```
westonpace: Why do we need to skip these tests?  I didn't think you hooked byte stream split into the encoding strategy yet?

rust/lance-encoding/src/encodings/physical/byte_stream_split.rs
```diff
@@ -0,0 +1,380 @@
...
+        // For f64 (8 bytes): 8186 / 8 = 1023, so max chunk = 512 (power of 2)
+        match self.bits_per_value {
+            32 => 1024,
+            64 => 512,
+            _ => unreachable!("ByteStreamSplit only supports 32 or 64 bit values"),
+        }
+    }
+}
+
+impl MiniBlockCompressor for ByteStreamSplitEncoder {
```
westonpace, line 106: So this works, but there might be a simpler way to implement this.  I often try to think of these encodings running layers (or maybe the right term is cascading).  For example, here is the diagram from the BtrBlocks paper:

image

What you've done here is you've fused the byte stream split encoder and the `ValueEncoder` (which has the chunking logic) into a single encoding (reimplementing the chunking logic).  You could simplify it by allowing the downstream encoder to be provided by the caller.  For example:

```
#[derive(Debug, Clone)]
pub struct ByteStreamSplitEncoderMiniBlock {
  bits_per_value: usize,
  inner: Box<dyn MiniBlockCompressor>,
}

...

impl MiniBlockCompressor for ByteStreamSplitEncoderMiniBlock {
  fn compress(&self, page: DataBlock) -> Result<(MiniBlockCompressed, pb::ArrayEncoding)> {
    if DataBlock::FixedWidth(data) = page {
      // Do the actual byte stream split on the data buffer and make a new data buffer (no chunking)
      let split_buffer: Buffer = split_values(data.data);
      let split_data = DataBlock::FixedWidth(FixedWidthDataBlock {
        data: split_buffer,
        bits_per_value: data.bits_per_value,
        num_values: data.num_values,
        block_info: BlockInfo::new()
      };
      let (compressed, descr) = self.inner.compress(split_data);
      let wrapped_descr = ProtobufUtils::byte_stream_split(descr);
      Ok((compressed, wrapped_descr))
    } else {
      // error
    }
  }
}
```

This also makes it possible to do things like BSS -> Bitpacking + delta or some other layers of encoding (I don't know if that ever makes sense or we would need to do that).

westonpace (approved):
I have a potential suggestion but we could also proceed as is and simplify later.

rust/lance-encoding/src/encodings/physical/value.rs
Xuanwo: Oh, right. I used to enable them by default so users could test them, but later I decided not to enable them by default. I'll revert those changes.

rust/lance-encoding/src/encodings/physical/byte_stream_split.rs
Xuanwo: I love this idea. Please let me create a new issue to track it. I want to run a benchmark on this, but this PR is currently blocking me. I plan to merge this PR first so that my benchmark and this optimization can be implemented in parallel.

westonpace: Go for it

PR #4238, opened 2025-07-15
docs: add fts tutorial

davidmyriel:
Basic three document full text search with inverted index and an explanation on various booleans, metadata etc.
Optimization tips as well.

### DIFF
--- a/docs/src/quickstart/full-text-search.md
+++ b/docs/src/quickstart/full-text-search.md
@@ -72,7 +72,7 @@ ds.create_scalar_index(
     - **INVERTED**: Traditional inverted index that maps terms to document IDs
     - **FTS**: Full Text Search index with additional linguistic processing capabilities
 
-The index creation process analyzes your text data and builds an efficient lookup structure that maps words to the documents containing them. This enables sub-second search performance even on large datasets.
+The index creation process analyzes your text data and builds an efficient lookup structure that maps words to the documents containing them. This enables high-performance full text search, even on large datasets.
 
 !!! warning "Index Creation Time"
     Index creation time depends on the size of your text data. For large datasets, this process may take several minutes, but the performance benefits are substantial.
@@ -101,6 +101,13 @@ ds.create_scalar_index(
 - **simple**: Splits tokens on whitespace and punctuation
 - **whitespace**: Splits tokens only on whitespace
 - **raw**: No tokenization (useful for exact matching)
+
+LanceDB also supports multilingual tokenization:
+
+- **jieba/default**: Chinese text tokenization using Jieba
+- **lindera/ipadic**: Japanese text tokenization using Lindera with IPAdic dictionary
+- **lindera/ko-dic**: Korean text tokenization using Lindera with Ko-dic dictionary
+- **lindera/unidic**: Japanese text tokenization using Lindera with UniDic dictionary
 
 ### Language Processing Features
 


docs/src/quickstart/full-text-search.md
```diff
@@ -0,0 +1,242 @@
...
+    column="text",
+    index_type="INVERTED",  # or "FTS"
+)
+```
+
+!!! note "Index Types"
+    - **INVERTED**: Traditional inverted index that maps terms to document IDs
+    - **FTS**: Full Text Search index with additional linguistic processing capabilities
+
+The index creation process analyzes your text data and builds an efficient lookup structure that maps words to the documents containing them. This enables sub-second search performance even on large datasets.
```
jackye1995: "sub-second" feels either slow or fast depending on which storage level people are comparing against. Can we just say "enables high-performance full text search"?

docs/src/quickstart/full-text-search.md
```diff
@@ -0,0 +1,242 @@
...
+```python
+ds = lance.dataset("/tmp/docs.lance")
+print(ds.schema)
+```
+
+This will show you the schema of your dataset, including the data types and column information. Understanding your schema is crucial for building effective indexes.
+
+## Build the Full-Text Search Index
+
+To enable fast full-text search operations, you need to create a scalar index on your text column. Lance supports two types of full-text search indexes:
```
jackye1995, line 62: @BubbleCal could you take a look if this is accurate?

docs/src/quickstart/full-text-search.md
```diff
@@ -0,0 +1,242 @@
...
+    max_token_length=40,         # Maximum token length
+    lower_case=True,             # Convert to lowercase
+    stem=False,                  # Enable stemming
+    remove_stop_words=False,     # Remove stop words
+    ascii_folding=False          # Convert non-ASCII to ASCII
+)
+```
+
+### Tokenizer Options
+
```
jackye1995, line 100: this is missing the lindera and jieba tokenizer options.

docs/src/quickstart/full-text-search.md
davidmyriel: Also another question - is it absolutely necessary to create an index? Or can I do brute force?

I got some type of an error when I didnt create one.

davidmyriel:
@BubbleCal @jackye1995 

Addressed comments, feel free to merge.

docs/src/quickstart/full-text-search.md
BubbleCal: It's required to create the index to do full-text search

The INVERTED index and FTS index are the same, it's just aliases. We can't use INVERTED here because FTS alias was added after it

docs/src/quickstart/full-text-search.md
```diff
@@ -0,0 +1,249 @@
...
+    language="English",           # For stemming and stop words
+    max_token_length=40,         # Maximum token length
+    lower_case=True,             # Convert to lowercase
+    stem=False,                  # Enable stemming
+    remove_stop_words=False,     # Remove stop words
+    ascii_folding=False          # Convert non-ASCII to ASCII
+)
+```
+
+### Tokenizer Options
```
BubbleCal, line 99: we just added ngram tokenizer for FTS, do we want to mention them here?
https://github.com/lancedb/lance/commit/d517f9b483c8f479390d91ca5ecfe36950a4ecbf#diff-7a5382f9cf3432567ed0df86a54435c0752992420e3397a99ae23f799563a256R175-R196

docs/src/quickstart/full-text-search.md
jackye1995: what about ngram? do we want to still mention that?

docs/src/quickstart/full-text-search.md
```diff
@@ -0,0 +1,249 @@
...
+
+## Advanced Search Features
+
+### Boolean Search Operators
+
+You can use boolean operators in your search queries:
+
+```python
+# Search for documents containing both "fox" AND "dog"
+table = ds.to_table(full_text_query="fox AND dog")
```
BubbleCal, line 144: boolean query is supported by a structured FTS query, we don't parse the string. ref [here](https://lancedb.com/documentation/guides/search/full-text-search.html#boolean-queries)

docs/src/quickstart/full-text-search.md
```diff
@@ -0,0 +1,249 @@
...
+table = ds.to_table(full_text_query='"quick brown fox"')
+```
+
+### Wildcard Search
+
+Use wildcards for pattern matching:
+
+```python
+# Search for words starting with "fox"
+table = ds.to_table(full_text_query="fox*")
```
BubbleCal, line 168: we support fuzzy search by https://lancedb.com/documentation/guides/search/full-text-search.html#4-basic-and-fuzzy-search

and prefix matching is supported, but not sufix matching

docs/src/quickstart/full-text-search.md
```diff
@@ -0,0 +1,249 @@
...
+# Search for documents containing "fox" OR "cat"
+table = ds.to_table(full_text_query="fox OR cat")
+
+# Search for documents containing "fox" but NOT "lazy"
+table = ds.to_table(full_text_query="fox NOT lazy")
+```
+
+### Phrase Search
+
+For exact phrase matching (requires `with_position=True` in index creation):
```
BubbleCal, line 155: we may want to mention that `remove_stop_words=True` may make phrase query miss some results that containing stop words

jackye1995:
@davidmyriel this doc feels more like a part of the user guide. The vector search page goes through a concrete example, but here this is going through all the features. Is there a way we can make this more similar to the vector search quickstart, and move some of the contents here to user guide?

docs/src/quickstart/full-text-search.md
BubbleCal: there 2 things about ngram:
- ngram index: it's another index type, which is used for finding rows containing a substring
- ngram tokenizer for FTS index: we can mention we support it

PR #4240, opened 2025-07-16
feat: add snappy general compressor support

Xuanwo:
This PR will add snappy compressor support for lance 2.1


Xuanwo:
Not needed.

PR #4242, opened 2025-07-16
feat: add file reader cache

wojiaodoubao:
Add FileReader cache to optimize Single Take (one row by each take).

### DIFF
--- a/rust/lance/src/dataset/fragment.rs
+++ b/rust/lance/src/dataset/fragment.rs
@@ -6,9 +6,10 @@
 pub mod write;
 
 use std::borrow::Cow;
+use std::cmp::min;
 use std::collections::{BTreeMap, HashMap, HashSet};
 use std::ops::Range;
-use std::sync::Arc;
+use std::sync::{Arc, LazyLock};
 
 use arrow::compute::concat_batches;
 use arrow_array::cast::as_primitive_array;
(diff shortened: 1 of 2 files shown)


wojiaodoubao:
In the Single Take Bench, the time costs could be optimized from 314ms to 258ms. See https://github.com/lancedb/lance/issues/4090 for more details.

Xuanwo:
Hi, as I understand it, we're trying to save an `open` syscall for the local file system. I'm a bit concerned about this because the `open` call for S3 is almost cost-free. If we're aiming to optimize this, should we consider placing the cache at the storage level instead? That way, we wouldn't have to modify our upper reader logic.

wojiaodoubao:
Sorry I think I made a mistake, so I need to modify this comment. I did the take bench using 'memory', so there is no syscall cost, thus 'ObjectStore Cache' may not help.

The idea of optimizing 'open syscall' is great. So I implemented an 'ObjectStore Cache'(https://github.com/wojiaodoubao/lance/blob/issue-4090-experiment/rust/lance-io/src/object_store.rs#L420) and ran the bench on ssd. It did optimizes some cost, but I think 'open syscall' only taking up a small fraction of the expense, so there is a gap between 'ObjectStore Cache' and 'FileReader Cache'.

The new experiment result can be found at https://github.com/lancedb/lance/issues/4090#issuecomment-3092415146.

wojiaodoubao:
Hi @Xuanwo , sorry to bother you, I have updated the test result(https://github.com/lancedb/lance/pull/4242#issuecomment-3083330319), could you help taking a look when you have time?  Thanks a lot.

Xuanwo:
> Hi @Xuanwo , sorry to bother you, I have updated the test result([#4242 (comment)](https://github.com/lancedb/lance/pull/4242#issuecomment-3083330319)), could you help taking a look when you have time? Thanks a lot.

Thank you, I'll take a look.

Issue #4162, opened 2025-07-07
Add zone map secondary index


westonpace:
Zone maps record the minimum and maximum for a "zone".  It's also sometimes just called statistics based pushdown.  It is an inexact index (similar to ngram).

Traditionally this is handled in file metadata co-resident with pages.  However, in lance, it would be much simpler for these to be treated the same as any other index.

The index could be a single file with a simple schema (min_value, max_value) and a rows_per_zone property that is stored in the metadata (this could be configurable, start with some default like 1024 or 4096 or 10000 or whatever).

Searches then identify all zones which are possible candidates given the min/max for that zone.

Note: if the data is not sorted by the index column then this index should still function but will be very ineffective (it will not skip very many zones).

Compared to btree / bitmap this index is only effective if the data is clustered (sorted) by the index column but requires much less space and does not require sorted data for training.

jackye1995:
What about other stats like null count, NaN count? I think they should also be a part of this

westonpace:
> What about other stats like null count, NaN count? I think they should also be a part of this

Yes, both null count and NaN count are useful for some filters (e.g. is_null) and we should include those.

However, I do also think "general column statistics" is another feature entirely and don't want to imply or suggest that zone map indexes are a replacement for column statistics.

jackye1995:
> I do also think "general column statistics" is another feature entirely and don't want to imply or suggest that zone map indexes are a replacement for column statistics.

Oh interesting, I thought this is basically serving the same purpose as column stats. Could you elaborate a bit more about the differences?

westonpace:
> Oh interesting, I thought this is basically serving the same purpose as column stats. Could you elaborate a bit more about the differences?

Let me make a new issue for that.

PR #4244, opened 2025-07-16
feat: add zonemap filter support to Lance at table level

HaochengLIU:
# Summary
This PR introduces an inexact index called `ZoneMap` which breaks the column into fixed-size chunks called zones and stores summary statistics for each zone (min, max, null_count, nan_count, fragment_id). It's very small but only effective if the column is at least approximately in sorted order. The zone size is configurable via environment variable `LANCE_ZONEMAP_DEFAULT_ROWS_PER_ZONE`.

Address #4162 and #4360

# Changes Made
* Add the zonemap filter in Rust
* Exposed the zonemap filter to Python
* Make re_map(...) function configurable for all scalar filters
* Make the recheck logic configurable on SargableQuery by moving the logic from filter level to SargableQueryParser

# Breaking Changes
* None

# Test Results
* Add thorough tests in ZoneMap.rs
* Add new Python tests in test_scalar_index.py

# API Impact
* A new filter called `ZONEMAP` is introduced to the Python create_scalar_index(...) function

### DIFF
--- a/python/python/tests/test_scalar_index.py
+++ b/python/python/tests/test_scalar_index.py
@@ -1327,9 +1327,6 @@ def test_zonemap_index(tmp_path: Path):
 
     # Verify zonemap statistics
     zonemap_stats = index_stats["indices"][0]
-    assert "max_zonemap_size" in zonemap_stats
-    assert "num_zones" in zonemap_stats
-    assert "type" in zonemap_stats
     assert zonemap_stats["max_zonemap_size"] == 8192
     assert zonemap_stats["num_zones"] == 2  # Should have 2 zones (8192 rows + 1 row)
     assert zonemap_stats["type"] == "ZoneMap"
--- a/rust/lance-datagen/src/generator.rs
+++ b/rust/lance-datagen/src/generator.rs
@@ -389,6 +389,8 @@ pub trait ArrayGeneratorExt {
     /// Replaces the validity bitmap of generated arrays with the inverse of `nulls`, cycling if needed
     fn with_nulls(self, nulls: &[bool]) -> Box<dyn ArrayGenerator>;
     /// Replaces the values of generated arrays with NaN values, cycling if needed
+    ///
+    /// Will have no effect if the data type is not a floating point data type
     fn with_nans(self, nans: &[bool]) -> Box<dyn ArrayGenerator>;
     /// Replaces the validity bitmap of generated arrays with `validity`, cycling if needed
     fn with_validity(self, nulls: &[bool]) -> Box<dyn ArrayGenerator>;
--- a/rust/lance-index/benches/zonemap.rs
+++ b/rust/lance-index/benches/zonemap.rs
@@ -36,7 +36,7 @@ fn bench_zonemap(c: &mut Criterion) {
         ))
     });
 
-    // Generate random integers for the zonemap index
+    // Generate sequential integers for the zonemap index
     let data_col = arrow_array::Int32Array::from_iter_values(0..TOTAL as i32);
 
     let row_addr_col = Arc::new(UInt64Array::from(
--- a/rust/lance-index/src/scalar/zonemap.rs
+++ b/rust/lance-index/src/scalar/zonemap.rs
@@ -483,7 +483,6 @@ impl ScalarIndex for ZoneMapIndex {
         query: &dyn AnyQuery,
         metrics: &dyn MetricsCollector,
     ) -> Result<SearchResult> {
-        println!("ZoneMapIndex::search!!!!!");
         metrics.record_comparisons(self.zones.len());
         let query = query.as_any().downcast_ref::<SargableQuery>().unwrap();
 
@@ -749,11 +748,10 @@ impl ZoneMapIndexBuilder {
 
                 // Check if there is enough data from the current fragment to fill the current zone
                 let desired = if let Some(idx) = next_fragment_index {
-                    // We found a fragment boundary, update cur_fragment_id and reset zone start
-                    let next_fragment_id = row_addrs_array.value(idx) >> 32;
-                    self.cur_fragment_id = next_fragment_id;
-                    // Don't create an empty zone, just process the data up to the boundary
-                    idx - array_offset
+                    self.cur_fragment_id = row_addrs_array.value(idx) >> 32;
+                    // Take the minimum between distance to boundary and space left in zone
+                    // to ensure we don't exceed the zone size limit
+                    std::cmp::min(idx - array_offset, empty_rows_left_in_cur_zone)
                 } else {
                     empty_rows_left_in_cur_zone
                 };
@@ -2061,5 +2059,4 @@ mod tests {
             &[4294967296, 4294967297, 4294967298, 4294967299, 4294967300]
         );
     }
-    // Create another test following test_page_cache in  btree.rs
 }
--- a/rust/lance/src/dataset/index.rs
+++ b/rust/lance/src/dataset/index.rs
@@ -43,10 +43,7 @@ impl DatasetIndexRemapper {
         mapping: &HashMap<u64, Option<u64>>,
     ) -> Result<Option<RemappedIndex>> {
         let new_uuid = remap_index(&self.dataset, &index.uuid, mapping).await?;
-        match new_uuid {
-            Some(uuid) => Ok(Some(RemappedIndex::new(index.uuid, uuid))),
-            None => Ok(None), // No remapping occurred
-        }
+        Ok(new_uuid.map(|uuid| RemappedIndex::new(index.uuid, uuid)))
     }
 }
 


rust/lance/src/index/scalar.rs
```diff
@@ -102,16 +113,25 @@ impl TrainingRequest {
...
         );
 
-        let ordering = match sort {
-            true => Some(vec![ColumnOrdering::asc_nulls_first(self.column.clone())]),
-            false => None,
+        let scan = if aligned {
+            // When aligned is true, add row_address column
+            scan.with_row_id()
+                .with_row_address()
+                .order_by(Some(vec![ColumnOrdering::asc_nulls_first(
```
HaochengLIU, line 168: remove it?

protos/table.proto
```diff
@@ -362,24 +379,26 @@ message ExternalFile {
...
+// hard-coded (e.g. filenames) or stored in the index itself.  However, we may
+// want to add more details in the future, in particular we can add details that
+// may be useful for planning queries (e.g. don't force us to load the index
+// until we know we need it)
 message BTreeIndexDetails {}
 message BitmapIndexDetails {}
 message LabelListIndexDetails {}
 message InvertedIndexDetails {}
 message NGramIndexDetails {}
+message ZoneMapIndexDetails {}
```
HaochengLIU, line 377: the only new change in this file

rust/lance-core/src/utils/testing.rs
```diff
@@ -12,7 +12,7 @@ use futures::{StreamExt, TryStreamExt};
 use object_store::path::Path;
 use object_store::{
     Error as OSError, GetOptions, GetResult, ListResult, MultipartUpload, ObjectMeta, ObjectStore,
-    PutMultipartOpts, PutOptions, PutPayload, PutResult, Result as OSResult,
+    PutOptions, PutPayload, PutResult, Result as OSResult,
```
HaochengLIU: According to clippy, `PutMultipartOpts` is deprecated

rust/lance-index/src/scalar.rs
```diff
@@ -214,10 +219,6 @@ pub trait AnyQuery: std::fmt::Debug + Any + Send + Sync {
     fn to_expr(&self, col: String) -> Expr;
     /// Compare this query to another query
     fn dyn_eq(&self, other: &dyn AnyQuery) -> bool;
-    /// If true, the query results are inexact and will need rechecked
-    fn needs_recheck(&self) -> bool {
```
HaochengLIU, line 218: dead code

rust/lance-index/src/scalar/lance_format.rs
```diff
@@ -331,20 +331,10 @@ pub mod tests {
     use datafusion_common::ScalarValue;
     use futures::FutureExt;
     use lance_core::utils::mask::RowIdTreeMap;
+    use lance_core::ROW_ADDR;
     use lance_datagen::{array, gen_batch, ArrayGeneratorExt, BatchCount, ByteCount, RowCount};
     use tempfile::{tempdir, TempDir};
 
-    fn test_store(tempdir: &TempDir) -> Arc<dyn IndexStore> {
```
HaochengLIU, line 337: moved into test section

rust/lance/src/index.rs
```diff
@@ -1442,30 +1449,34 @@ impl DatasetIndexInternalExt for Dataset {
...
+                        as Box<dyn ScalarQueryParser>,
+                    _ => continue,
+                },
+                _ => {
+                    // inexact index filter
+                    let needs_recheck = matches!(index_type, ScalarIndexType::ZoneMap);
+                    Box::new(SargableQueryParser::new(index.name.clone(), needs_recheck))
+                        as Box<dyn ScalarQueryParser>
                 }
-                _ => Box::new(SargableQueryParser::new(index.name.clone()))
```
HaochengLIU, line 1378: all cases are covered. This code will irritate the linter.

rust/lance-index/src/scalar/zonemap.rs
```diff
@@ -0,0 +1,1628 @@
...
+struct ZoneMapStatistics {
+    min: ScalarValue,
+    max: ScalarValue,
+    null_count: u32,
+    fragment_id: u64,
+    zone_size: usize,
+}
+
+impl DeepSizeOf for ZoneMapStatistics {
+    fn deep_size_of_children(&self, _context: &mut deepsize::Context) -> usize {
```
HaochengLIU, line 57: I'm not sure about the fn implementation here

rust/lance-index/benches/zonemap.rs
```diff
@@ -0,0 +1,122 @@
+// SPDX-License-Identifier: Apache-2.0
```
HaochengLIU, line 1: sample output
```
Gnuplot not found, using plotters backend
train/zonemap_train(1000000)
                        time:   [913.61 µs 929.59 µs 947.05 µs]
                        change: [-21.243% -19.322% -17.382%] (p = 0.00 < 0.05)
                        Performance has improved.

search/zonemap_search(1000000)
                        time:   [1.0365 µs 1.0459 µs 1.0630 µs]
                        change: [-87.130% -86.475% -85.741%] (p = 0.00 < 0.05)
                        Performance has improved.
  1 (10.00%) high mild

```

HaochengLIU:
Hey folks @westonpace @jackye1995 @Xuanwo , could you review the PR when you have time? I will try to populate the proper commit messages either tonight or tomorrow, ty

The key file is lance-index/src/scalar/zonemap.rs

protos/table.proto
jackye1995, line 1: looks like there are quite a few line changes for the comments in the file. Is there some sort of new linter used here?

rust/lance-index/src/scalar/zonemap.rs
```diff
@@ -0,0 +1,1629 @@
...
+
+const ZONEMAP_FILENAME: &str = "zonemap.lance";
+const ZONEMAP_SIZE_META_KEY: &str = "zonemap_size";
+
+/// Basic stats about zonemap index
+#[derive(Debug, PartialEq, Clone)]
+struct ZoneMapStatistics {
+    min: ScalarValue,
+    max: ScalarValue,
+    null_count: u32,
```
jackye1995, line 51: I think we should also have a nan_count for floats

rust/lance-index/src/scalar/zonemap.rs
```diff
@@ -0,0 +1,1629 @@
...
+        metrics: &dyn MetricsCollector,
+    ) -> Result<SearchResult> {
+        metrics.record_comparisons(self.zones.len());
+        let query = query.as_any().downcast_ref::<SargableQuery>().unwrap();
+
+        let mut row_id_tree_map = RowIdTreeMap::new();
+        let mut current_fragment_id = None;
+        let mut cumulative_offset = 0;
+
+        // Loop through zones sequentially to calculate offsets efficiently
```
jackye1995: I was expecting the zone_start to be stored alongside the zone size, so that each zone map stats is self-describing. But looks like in the implementation here we do not store the zone_start, and instead we assume that the zones of a fragment are listed sequentially in order and calculate offset based on the previous size. Why is this more efficient? I would imagine if I use the zone map index to do pruning if I don't need to calculate zone start sequentially it could be faster since I can prune all the zones in parallel and directly send to downstream for execution.

rust/lance-index/src/scalar/zonemap.rs
```diff
@@ -0,0 +1,1629 @@
...
+        )?))
+    }
+
+    /// Remap the row ids, creating a new remapped version of this index in `dest_store`
+    async fn remap(
+        &self,
+        _mapping: &HashMap<u64, Option<u64>>,
+        _dest_store: &dyn IndexStore,
+    ) -> Result<()> {
+        Err(Error::InvalidInput {
```
jackye1995, line 547: Do we plan to make this a "primary index" (if I remember Weston's term correctly) and always update this index synchronously at write time? If so I think this is not really "not supported", but because we always update synchronously, remap is not necessary and should be a no-op.

protos/table.proto
HaochengLIU: could you guide which linter should i use to reformat it? Currently i use whatever cursor uses

rust/lance-index/src/scalar/zonemap.rs
HaochengLIU: The initial impl assumes every zone has the same size, then Weston and I decided to relax it to be flexible. I'm down to store zone_start to further improve it, will add it to the next round of review

rust/lance-index/src/scalar/zonemap.rs
HaochengLIU: I'm still trying to grip the gist of this `remap` Incantation, I will defer it to Weston

protos/table.proto
jackye1995: currently we don't lint the proto files. I think we should just revert those changes and have only the change related to zone map

rust/lance-index/src/scalar/zonemap.rs
jackye1995: @westonpace thoughts?

python/python/lance/dataset.py
```diff
@@ -1896,6 +1897,10 @@ def create_scalar_index(
           creates a bitmap for each ngram in the string.  By default we use trigrams.
           This index can currently speed up queries using the ``contains`` function
           in filters.
+        * ``ZONEMAP``. This index breaks the column into fixed-size chunks called zones
+          and store summary statistics for each zone(min, max, null_count, fragment_id).
+          It's inexact but enables efficient filtering by eliminating zones that do not
+          contain matching values.
```
westonpace: We should mention something about sorting here.  Perhaps...

```suggestion
        * ``ZONEMAP``. This index breaks the column into fixed-size chunks called zones
          and stores summary statistics for each zone(min, max, null_count, fragment_id).
          It's very small but only effective if the column is at least approximately in sorted
          order.
```

python/python/tests/test_scalar_index.py
```diff
@@ -1310,6 +1310,42 @@ def test_with(tbl: pa.Table):
     test_with(tbl)
 
 
+def test_zonemap_index(tmp_path: Path):
+    """Test create zonemap index"""
+    os.environ["LANCE_ZONEMAP_DEFAULT_ROWS_PER_ZONE"] = "100"
```
westonpace: I'm generally nervous about setting environment variables in unit tests.  Unit tests may run in parallel and environment variables are process-wide.  Although I don't actually recall if our python tests run in parallel.

python/python/tests/test_scalar_index.py
```diff
@@ -1310,6 +1310,42 @@ def test_with(tbl: pa.Table):
...
+    assert "num_zones" in zonemap_stats
+    assert "type" in zonemap_stats
+    assert zonemap_stats["max_zonemap_size"] == 100
+    assert zonemap_stats["num_zones"] == 2  # Should have 2 zones (100 rows + 1 row)
+    assert zonemap_stats["type"] == "ZoneMap"
+
+    # Test that the zonemap index is being used in the query plan
+    scanner = dataset.scanner(filter="values > 50", prefilter=True)
+    plan = scanner.explain_plan()
+    assert "ZoneMapIndex" in plan or "zonemap" in plan.lower()
```
westonpace: Why is there an `or` here?  This should be deterministic and only have one possibility right?

rust/lance-index/src/lib.rs
```diff
@@ -96,6 +96,8 @@ pub enum IndexType {
 
     MemWal = 7,
 
+    ZoneMap = 8, // ZoneMap, Sargable filter
```
westonpace:
```suggestion
    ZoneMap = 8, // ZoneMap
```

for consistency (many of the above indexes are also sargable)?

rust/lance/src/index.rs
```diff
@@ -265,6 +265,10 @@ pub(crate) async fn remap_index(
             let scalar_index = dataset
                 .open_scalar_index(&field.name, &index_id.to_string(), &NoOpMetricsCollector)
                 .await?;
+            if !scalar_index.can_remap() {
+                // If the index cannot be remapped, we just return the same index ID.
+                return Ok(*index_id);
+            }
```
westonpace: I don't remember if this is correct.  Let's create a python unit test to cover remapping.  It should look something like...

* Create a dataset with 5 fragments
* Train a zone map index
* Confirm the zone map index is used if you search the dataset
* Run compaction
* Confirm the zone map index is **not** used if you search the dataset
* Run optimize indices
* Confirm the zone map index is used if you search the dataset

rust/lance/src/index/scalar.rs
```diff
@@ -79,6 +87,8 @@ impl TrainingRequest {
         self: Box<Self>,
         chunk_size: u32,
         sort: bool,
+        // based on aligned, we can add a row_address column to the scan and it's order by row_address
+        aligned: bool,
```
westonpace: Can both `sort` and `aligned` be true?  If not, maybe a small enum would be more clear?

```
enum OrderMode {
  Unordered,
  Ordered,
  Aligned,
}
```

rust/lance-index/src/scalar/zonemap.rs
```diff
@@ -0,0 +1,1629 @@
...
+    max: ScalarValue,
+    null_count: u32,
+    fragment_id: u64,
+    zone_size: usize,
+}
+
+impl DeepSizeOf for ZoneMapStatistics {
+    fn deep_size_of_children(&self, _context: &mut deepsize::Context) -> usize {
+        // Estimate sizes for ScalarValue
+        let min_size = match &self.min {
```
westonpace: Oops, I deleted your comment here saying you were unsure about the impl (I was trying to delete my reply :laughing:)

Anyways, we have an impl for `DeepSizeOf` for `ScalarValue` in `OrderableScalarValue` so you can just reuse that logic.  It's something like `self.min.size() - std::mem::size_of::<ScalarValue>()`

rust/lance-index/src/scalar/zonemap.rs
```diff
@@ -0,0 +1,1629 @@
...
+            ScalarValue::Int64(_) => 8,
+            ScalarValue::Float32(_) => 4,
+            ScalarValue::Float64(_) => 8,
+            ScalarValue::Utf8(Some(s)) => s.len(),
+            ScalarValue::Utf8(None) => 0,
+            ScalarValue::LargeUtf8(None) => 0,
+            _ => 16, // Default estimate
+        };
+
+        min_size + max_size + 4 + 8 + 8
```
westonpace:
```suggestion
        min_size + max_size
```

The `deep_size_of_children` method only needs to keep track of things that wouldn't be included in `std::mem::size_of`.  So you don't need the `4 + 8 + 8`.

E.g. in C++ speak we are looking for anything that is heap-allocated.

rust/lance-index/src/scalar/zonemap.rs
```diff
@@ -0,0 +1,1629 @@
...
+    }
+}
+
+/// ZoneMap index
+/// At high level it's a columnar database technique for predicate push down and scan pruning.
```
westonpace, line 72: This is quite a high level description :laughing:

rust/lance-index/src/scalar/zonemap.rs
westonpace: Yep, I hadn't thought this through either.  I agree storing the start will be useful.

rust/lance-index/src/scalar/zonemap.rs
```diff
@@ -0,0 +1,1629 @@
...
+    ) -> Result<Arc<Self>>
+    where
+        Self: Sized,
+    {
+        let index_file = store.open_index_file(ZONEMAP_FILENAME).await?;
+        let zone_maps = index_file
+            .read_range(0..index_file.num_rows(), None)
+            .await?;
+        let file_schema = index_file.schema();
+        // zonemap_size is expected to be a usize (or possibly u64), parsed from a string in the metadata
```
westonpace: This is up to us right?  Let's be concrete and assert `u64`.

rust/lance-index/src/scalar/zonemap.rs
westonpace: I don't know if I had a term for it.  For now, at least, we are not adding any kind of index-on-write.  A zone map cannot do remap whether we make it a primary index or not so I think "not supported" is fine.  We've added a `can_remap` so this method should never be called.  Instead, if a fragment covered by a zone map is compacted, it is a no-op.  The old fragment remains in the index's bitmap and the new fragment is not covered by the index.  It can be included by running an update.

rust/lance-index/src/scalar/zonemap.rs
```diff
@@ -0,0 +1,1629 @@
...
+                    // We found a fragment boundary, update cur_fragment_id
+                    let next_fragment_id = row_addrs_array.value(idx) >> 32;
+                    self.cur_fragment_id = next_fragment_id;
+                    // Don't create an empty zone, just process the data up to the boundary
+                    idx - array_offset
+                } else {
+                    empty_rows_left_in_cur_zone
+                };
+                // desired = {}, next_fragment_index = {:?}, remaining = {}",
+                // desired, next_fragment_index, remaining
```
westonpace: Is this comment leftover?

rust/lance-index/src/scalar/zonemap.rs
```diff
@@ -0,0 +1,1629 @@
...
+                    if self.cur_zone_offset > 0 {
+                        self.new_map(self.cur_fragment_id - 1)?;
+                    }
+                    // Let the loop run again
+                    // to find the next fragment boundary
+                    continue;
+                }
+                array_offset += desired;
+                remaining = remaining.saturating_sub(desired);
+                // array_offset = {}, remaining = {}", array_offset, remaining
```
westonpace: ?

rust/lance-index/src/scalar/zonemap.rs
```diff
@@ -0,0 +1,1629 @@
...
+            }
+        }
+        // Create the final map
+        if self.cur_zone_offset > 0 {
+            self.new_map(self.cur_fragment_id)?;
+        }
+
+        // TODO return anything here???
```
westonpace: Nope :smile:

protos/table.proto
HaochengLIU: k, reverted

rust/lance-index/src/scalar/zonemap.rs
HaochengLIU: good catch
lol dead debug code, let me remove it

rust/lance-index/src/scalar/zonemap.rs
HaochengLIU: added

rust/lance-index/src/scalar/zonemap.rs
HaochengLIU: removed

rust/lance-index/src/scalar/zonemap.rs
HaochengLIU: updated

python/python/lance/dataset.py
HaochengLIU: reworded

python/python/tests/test_scalar_index.py
HaochengLIU: humm, removed, it's indeed a safety hazard

python/python/tests/test_scalar_index.py
HaochengLIU: "ZoneMapIndex" is removed since it does not exist

rust/lance-index/src/lib.rs
HaochengLIU: fixed

rust/lance/src/index/scalar.rs
HaochengLIU: I love the idea. Compacted

rust/lance-index/src/scalar/zonemap.rs
HaochengLIU: ah, now it's clear

rust/lance-index/src/scalar/zonemap.rs
HaochengLIU: fixed

rust/lance-index/src/scalar/zonemap.rs
HaochengLIU: removed

rust/lance-index/src/scalar/zonemap.rs
HaochengLIU: fixed

python/python/tests/test_scalar_index.py
```diff
@@ -1310,6 +1310,124 @@ def test_with(tbl: pa.Table):
...
+        start_val = i * 1000
+        end_val = (i + 1) * 1000
+        tbl = pa.Table.from_arrays(
+            [pa.array([j for j in range(start_val, end_val)])], 
+            names=["values"]
+        )
+        if i == 0:
+            dataset = lance.write_dataset(tbl, tmp_path / "dataset", max_rows_per_file=1000)
+        else:
+            dataset = lance.write_dataset(tbl, tmp_path / "dataset", mode="append", max_rows_per_file=1000)
```
westonpace: You can just do...

```
tbl = pa.Table.from_arrays(
  [pa.array(range(0, 5000))],
  names=["values"]
)
lance.write_dataset(tbl, tmp_path / "dataset", max_rows_per_file=1000)
```

It will split into 5 fragments already due to `max_rows_per_file`.

python/python/tests/test_scalar_index.py
```diff
@@ -1310,6 +1310,124 @@ def test_with(tbl: pa.Table):
...
+    
+    # After compaction, check the fragment structure
+    fragments_after = dataset.get_fragments()
+    print(f"Fragments after compaction: {len(fragments_after)}")
+    
+    # Check if the zone map index is still being used
+    scanner = dataset.scanner(filter="values > 2500", prefilter=True)
+    plan = scanner.explain_plan()
+    print(f"Query plan after compaction: {plan}")
+    assert "zonemap" in plan.lower()
```
westonpace: It should no longer be used at this point.  I would expect this to be `assert "zonemap" not in plan.lower()`

python/python/tests/test_scalar_index.py
```diff
@@ -1310,6 +1310,124 @@ def test_with(tbl: pa.Table):
...
+    
+    # The index might not work correctly after compaction, so let's test without prefilter
+    scanner_no_prefilter = dataset.scanner(filter="values > 2500", prefilter=False)
+    print(f"Query plan without prefilter: {scanner_no_prefilter.explain_plan()}")
+    result_no_prefilter = scanner_no_prefilter.to_table()
+    print(f"Results without prefilter: {result_no_prefilter.num_rows} rows")
```
westonpace: The query should work, it just won't use the index.

HaochengLIU: reworked the test

rust/lance/src/index.rs
HaochengLIU: A new test `test_zonemap_index_remapping` is introduced

rust/lance-datagen/src/generator.rs
```diff
@@ -287,6 +388,8 @@ pub trait ArrayGeneratorExt {
     fn with_random_nulls(self, null_probability: f64) -> Box<dyn ArrayGenerator>;
     /// Replaces the validity bitmap of generated arrays with the inverse of `nulls`, cycling if needed
     fn with_nulls(self, nulls: &[bool]) -> Box<dyn ArrayGenerator>;
+    /// Replaces the values of generated arrays with NaN values, cycling if needed
+    fn with_nans(self, nans: &[bool]) -> Box<dyn ArrayGenerator>;
```
westonpace:
```suggestion
    /// Replaces the values of generated arrays with NaN values, cycling if needed
    ///
    /// Will have no effect if the data type is not a floating point data type
    fn with_nans(self, nans: &[bool]) -> Box<dyn ArrayGenerator>;
```

rust/lance/src/dataset/index.rs
```diff
@@ -41,9 +41,12 @@ impl DatasetIndexRemapper {
...
         index: &Index,
         mapping: &HashMap<u64, Option<u64>>,
-    ) -> Result<RemappedIndex> {
+    ) -> Result<Option<RemappedIndex>> {
         let new_uuid = remap_index(&self.dataset, &index.uuid, mapping).await?;
-        Ok(RemappedIndex::new(index.uuid, new_uuid))
+        match new_uuid {
+            Some(uuid) => Ok(Some(RemappedIndex::new(index.uuid, uuid))),
+            None => Ok(None), // No remapping occurred
+        }
```
westonpace: Minor nit: this could be a map call...

```
Ok(new_uuid.map(|uuid| RemappedIndex::new(index.uuid, uuid)))
```

westonpace (approved):

HaochengLIU:
✋ Finally in!

jackye1995:
🎉 🎉 🎉 🎉 🎉 🎉

PR #4245, opened 2025-07-17
feat: add claude md for future agentic work

Xuanwo:
This PR aims to add `CLAUDE.md` to help the code agent collaborate more effectively with humans.

Most of the content was generated by Claude itself; my own contributions are in the `Project Vision` and `Project Requirements` sections.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -5,11 +5,11 @@ This file provides guidance to Claude Code (claude.ai/code) when working with co
 ## Project Overview
 
 Lance is a modern columnar data format optimized for ML workflows and datasets. It provides:
-- 100x faster random access than Parquet
-- Native vector search capabilities (IVF, HNSW, PQ)
-- Zero-copy automatic versioning
-- Deep integration with Apache Arrow, Pandas, Polars, DuckDB, Ray
-- Cloud-native with S3, Azure, GCS support
+
+- High-performance random access
+- Vector search
+- Zero-copy, automatic versioning
+- Ecosystem integrations
 
 ## Project Vision
 


Xuanwo:
This should also work for other code agents. I assume they're intelligent enough to read CLAUDE.md as well.

jackye1995:
I remember @wjones127 also has one, might be worth merging the 2?

Xuanwo:
> I remember @wjones127 also has one, might be worth merging the 2?

Yeah, I have requested review from @wjones127.

CLAUDE.md
```diff
@@ -0,0 +1,134 @@
...
+cargo test -- --nocapture          # Show println! output
+
+# Formatting & Linting
+cargo fmt                          # Format code
+cargo fmt -- --check              # Check formatting
+cargo clippy -- -D warnings       # Run linter
+
+# Dependency Checking
+cargo deny check                   # Audit dependencies
+```
```
wjones127: To save on context window, I would condense this:

* Check for build errors: `cargo check --all --tests --benches`
* Run tests: `cargo test`
* Run specific test: `cargo test -p <package> <test_name>`
* Lint: `cargo clippy --all --tests --benches -- -D warnings`
* Format: `cargo fmt --all`

CLAUDE.md
```diff
@@ -0,0 +1,134 @@
...
+ruff format python               # Format Python code
+ruff check python                # Run linter
+ruff check --fix python          # Auto-fix issues
+pyright                          # Type checking
+
+# Benchmarking
+pip install -e python[benchmarks]
+pytest python/benchmarks         # Run quick benchmarks
+pytest python/benchmarks -m "not slow"  # Skip slow benchmarks
+```
```
wjones127: I would recomend moving this into the `python/CLAUDE.md` folder, so it only comes into context when working in that directory. Also, could we point it to use the `Makefile`?

Here's what I have in mine:

```markdown
Use the makefile for most actions:

* Build: `maturin develop`
* Test: `make test`
* Run single test: `pytest python/tests/<test_file>.py::<test_name>`
* Doctest: `make doctest`
* Lint: `make lint`
* Format: `make format`

```

CLAUDE.md
```diff
@@ -0,0 +1,134 @@
...
+
+### Building Wheels
+
+```bash
+# Linux x86_64 (requires Docker)
+docker run -v $(pwd):/io ghcr.io/pyo3/maturin:latest build --release -m python/Cargo.toml --features <features>
+
+# macOS
+maturin build --release -m python/Cargo.toml
+```
```
wjones127: I think we can omit this. It generally shouldn't need to build the wheels.

CLAUDE.md
```diff
@@ -0,0 +1,134 @@
...
+
+## Key Technical Details
+
+1. **Async-first Architecture**: Heavy use of tokio and async/await throughout Rust codebase
+2. **Arrow-native**: All data operations work directly with Apache Arrow arrays
+3. **Version Control**: Every write creates a new version with manifest tracking
+4. **Indexing**: Supports both vector indices (for similarity search) and scalar indices (BTree, inverted)
+5. **Encoding**: Custom encodings optimized for ML data patterns
+6. **Object Store**: Unified interface for local, S3, Azure, GCS storage
+
```
wjones127, line 85: I'd also add these tips:

```markdown

## Development tips

Code standards:
* Be mindful of memory use:
  * When dealing with streams of `RecordBatch`, avoid collecting all data into
    memory whenever possible.
  * Use `RoaringBitmap` instead `HashSet<u32>`.

Tests:
* When writing unit tests, prefer using the `memory://` URI instead of creating
  a temporary directory.
* Use rstest to generate parameterized tests to cover more cases with fewer lines
  of code.
    * Use syntax `#[case::{name}(...)]` to provide human-readable names for each case.
* For backwards compatibility, use the `test_data` directory to check in datasets
  written with older library version.
    * Check in a `datagen.py` that creates the test data. It should assert the
      version of Lance used as part of the script.
    * Use `pip install pylance=={version}` and then run `python datagen.py` to
      create the dataset. The data files should be checked into git.
    * Use `copy_test_data_to_tmp` to read this data in Lance
```

CLAUDE.md
Xuanwo: Applied.

CLAUDE.md
Xuanwo: Makes sense, removed.

CLAUDE.md
Xuanwo: Nice!

CLAUDE.md
```diff
@@ -0,0 +1,114 @@
+# CLAUDE.md
+
+This file provides guidance to Claude Code (claude.ai/code) when working with code in this repository.
+
+## Project Overview
+
+Lance is a modern columnar data format optimized for ML workflows and datasets. It provides:
+- 100x faster random access than Parquet
+- Native vector search capabilities (IVF, HNSW, PQ)
```
jackye1995: this is not up to date with the latest doc https://github.com/lancedb/lance/blob/main/docs/src/index.md?plain=1#L23

We should probably add a line in contributing guide that if we edit anything we should keep this file in sync.

Xuanwo: I'm now thinking this doesn't need to be here, so let's just remove this section.

jackye1995 (approved):

PR #4248, opened 2025-07-17
perf: improve btree performance for range query

BubbleCal:


rust/lance-index/src/scalar/expression.rs
```diff
@@ -1007,61 +1007,62 @@ fn visit_comparison(
...
-                Some(Between {
-                    expr: left_comparison.right.clone(),
-                    low: left_comparison.left.clone(),
-                    high: right_comparison.left.clone(),
-                    negated: false,
-                })
-            } else {
-                None
-            }
+            (Bound::Included(right_value), Bound::Included(left_value))
```
Xuanwo, line 1050: ~~Order is wrong?~~

Is it better to swap the left and right values before comparing, or are there other constraints we should consider?

rust/lance-index/src/scalar/expression.rs
```diff
@@ -1007,61 +1007,62 @@ fn visit_comparison(
...
-    match (left_comparison.op, right_comparison.op) {
+    let (left_col, dt, parser) = maybe_indexed_column(&left_expr.left, index_info)?;
+    let right_col = maybe_column(&right_expr.left)?;
+
+    if left_col != right_col {
+        return None;
+    }
+
+    let left_value = maybe_scalar(&left_expr.right, dt)?;
+    let right_value = maybe_scalar(&right_expr.right, dt)?;
```
westonpace, line 1031: This assumes we always have `column op literal` and never `literal op column` but I think that's the way the DF normalizer works so we are ok.

westonpace (approved):
Thanks for digging into this.  This has tripped me up a few times in the past.

PR #4253, opened 2025-07-17
feat: extend merge_insert fast path for UpdateIf and update-only operations

### LINKED ISSUE #4196: Optimize merge_insert with only updates and with update condition
## Background

We are migrating from an implementation of merge insert that manually manipulates streams to one that constructs a DataFusion plan and runs optimizers on that plan for optimize it.

The old code path is in `execute_uncommitted_impl`:

https://github.com/lancedb/lance/blob/94dcaf5e735480f9f61216b4e4f90358dbeac2ba/rust/lance/src/dataset/write/merge_insert.rs#L1337-L1340

While the new codepath is under `execute_uncommitted_v2`:

https://github.com/lancedb/lance/blob/94dcaf5e735480f9f61216b4e4f90358dbeac2ba/rust/lance/src/dataset/write/merge_insert.rs#L1352-L1371

## Task

**Goal**: get the fast path working for both:

1. `self.params.insert_not_matched = false` (Just update, without insert)
2. `matches!(self.params.when_matched, WhenMatched::UpdateIf(_))`

For (1), we should make sure we switch the `JoinType` to `JoinType::Inner`:

https://github.com/lancedb/lance/blob/94dcaf5e735480f9f61216b4e4f90358dbeac2ba/rust/lance/src/dataset/write/merge_insert.rs#L1243-L1245

For (2), we can add the condition in `merge_insert_action`:

https://github.com/lancedb/lance/blob/94dcaf5e735480f9f61216b4e4f90358dbeac2ba/rust/lance/src/dataset/write/merge_insert/assign_action.rs#L51-L53

wjones127:
## Summary

Extends the DataFusion plan-based optimization (`execute_uncommitted_v2`) to support two important merge_insert scenarios:

- **Update-only operations** (`insert_not_matched = false`) with Inner join optimization
- **Conditional updates** (`WhenMatched::UpdateIf`) with expression transformation support

## Key Improvements

- **Schema resolution fix**: Added `transform_struct_to_qualified_expr` function to convert struct field access expressions to qualified column references, solving the critical issue where DataFusion interpreted `source.value` as struct field access instead of qualified column references
- **Fast path extension**: Extended `can_use_fast_path` conditions to include `WhenMatched::UpdateIf(_)` 
- **Join optimization**: Implemented conditional join type logic - uses Inner join when `insert_not_matched = false` instead of Right join for better performance
- **Comprehensive testing**: Added thorough test coverage for both optimization scenarios

## Performance Impact

This enables significant performance improvements for common merge_insert patterns by migrating from manual stream manipulation to DataFusion plan-based execution for:
- Update-only operations (no inserts)
- Conditional updates based on source/target column values

## Test Plan

- ✅ All existing 20 merge_insert tests pass
- ✅ Added `test_fast_path_update_only` - validates Inner join optimization
- ✅ Added `test_fast_path_conditional_update` - validates UpdateIf expression handling
- ✅ Verified no performance regressions in existing functionality
- ✅ Clippy and formatting checks pass

## Related Issues

Fixes #4196


rust/lance/src/dataset/write/merge_insert.rs
```diff
@@ -192,27 +192,13 @@ pub enum WhenMatched {
     DoNothing,
     /// The row is updated (similar to UpdateAll) only for rows where the expression evaluates to
     /// true
-    UpdateIf(Expr),
+    UpdateIf(String),
```
wjones127, line 195: Because each code path needs to parse the expression a little differently, I am deferring the expression parsing till later for now. When we delete the old code path later, we can remove this.

rust/lance-datafusion/src/planner.rs
```diff
@@ -238,28 +238,50 @@ impl ContextProvider for LanceContextProvider {
 pub struct Planner {
     schema: SchemaRef,
     context_provider: LanceContextProvider,
+    enable_relations: bool,
```
LuQQiu, line 241: Minor: could we have a description here for true/false behavior

rust/lance/src/dataset/write/merge_insert.rs
```diff
@@ -1240,9 +1226,18 @@ impl MergeInsertJob {
         let source_df = session_ctx.read_one_shot(source)?;
         let source_df_aliased = source_df.alias("source")?;
         let scan_aliased = scan.alias("target")?;
+        let join_type = if self.params.insert_not_matched {
+            JoinType::Right
+        } else {
+            JoinType::Inner
```
LuQQiu, line 1232: Good call

rust/lance-datafusion/src/planner.rs
```diff
@@ -238,28 +238,50 @@ impl ContextProvider for LanceContextProvider {
...
-        let mut column = col(&idents[0].value);
-        for ident in &idents[1..] {
-            column = Expr::ScalarFunction(ScalarFunction {
-                args: vec![
-                    column,
-                    Expr::Literal(ScalarValue::Utf8(Some(ident.value.clone())), None),
-                ],
-                func: Arc::new(ScalarUDF::new_from_impl(GetFieldFunc::default())),
-            });
+    pub fn with_enable_relations(mut self, enable_relations: bool) -> Self {
```
wjones127, line 258:
```suggestion
    /// If passed with `true`, then the first identifier in column reference
    /// is parsed as the relation. For example, `table.field.inner` will be
    /// read as the nested field `field.inner` (`inner` on struct field `field`)
    /// on the `table` relation. If `false` (the default), then no relations
    /// are used and all identifiers are assumed to be a nested column path.
    pub fn with_enable_relations(mut self, enable_relations: bool) -> Self {
```

PR #4257, opened 2025-07-18
feat!: support shallow_clone in dataset

majin1102:
This is the prototype of shallow clone.

Design document: #4256

I did tests on some real datasets. It's OK to do some basic read and write. We may need some deeper testcases.

If there are confusing concepts or details. Please let me know or refer to the document.

@jackye1995  cc @wojiaodoubao


jackye1995:
This is actually different from what I was thinking about doing shallow clone. Looks like what we are doing here is that we add a reference object so that the data and delete file paths are resolved using the reference, instead of using the table's root location.

I was thinking about instead of introducing a Clone operation and a Reference object in manifest, when we do shallow clone, we convert all the relative paths to absolute paths. And we fix the reader to allow reading absolute path instead of always resolving to a relative path. I think that is also how Delta does shallow clone: https://github.com/delta-io/delta/blob/master/spark/src/main/scala/org/apache/spark/sql/delta/commands/CloneTableBase.scala#L147-L151

Overall, my feeling is that using absolute paths is better because with reference, it is unclear how we will resolve file paths after we write new fragments, some needs to be resolved using the reference dataset, some needs to use the current new dataset, and things get complicated quickly. Also absolute paths will be useful also for other cases like import files, so it's in general a good feature to have.

What do you think?

jackye1995:
cc @wjones127 if you have any thoughts on this

majin1102:
> I was thinking about instead of introducing a Clone operation and a Reference object in manifest, when we do shallow clone, we convert all the relative paths to absolute paths. And we fix the reader to allow reading absolute path instead of always resolving to a relative path. I think that is also how Delta does shallow clone: https://github.com/delta-io/delta/blob/master/spark/src/main/scala/org/apache/spark/sql/delta/commands/CloneTableBase.scala#L147-L151

Thanks for the delta input. I will look into it.

Let's take it into two parts:
### Clone operation. 
I think for now every manifest/version has a related transaction. What do we put in the cloned dataset as the initialized transaction like Overwrite if we don't have a clone operation? It will help if you can elaborate more about your thoughts.(I think this may be unrelated to the reference thing?)
### Absolute path. 
I did some work to use absolute path in this case. And some mind came up to me:
1. when we talk about absolute path, we actually use a dataset path to get a data directory and the filename. What if we just make the source dataset as an absolute path. This seems lighter and friendly to goverment tools. We get absolute file paths from reference eventually in this case.
2. I was thinking: what if the cloned dataset overwritten or totally updated. Then it comes to a normal dataset which has no actual relations with the source dataset. Somehow I thought we must use absolute path here is some sort of limitation. Or let's say absolute path could be an independent feature or property. I thought it may be not that necessary to bound to this case.
3. For deletion file, there is not even a relative path, the file name is assembled. I need to add a path to the spec. I think it will work, just a little heavy for this case.(Acutally this is the first problem I met)
4. Later I tried to figure out if there are blocking points to make this reference happen. I found the logics is quite clear:
    - If fragment_id > reference.max_fragment_id, use the cloned dataset relative path. Otherwise, use the source dataset path.
    - if deletion_file.read_version >= reference.version, use the cloned dataset relative path. Otherwise, use the source dataset path.

This is my mindmap.  And I thought this reference was quite lance unique thing, because of the fragment id and the deletion file construction

> it is unclear how we will resolve file paths after we write new fragments, some needs to be resolved using the reference dataset

For now I think the 'unclear' thing is a simple condition for both data file and deletion file. But I maybe wrong or misunderstand something. I'm not insistent on this reference methenism. But I think it will help if we can talk deeper into it and do some comparision.

> Also absolute paths will be useful also for other cases like import files, so it's in general a good feature to have.

I think I need to learn more from scenarios like import files you metioned. Apparently I don't get the whole picture. Overall I‘m willing to embrace using absolute paths in each datafiles, and I'm open to improve this prototype by anyways.

jackye1995:
> I think for now every manifest/version has a related transaction. What do we put in the cloned dataset as the initialized transaction like Overwrite if we don't have a clone operation? It will help if you can elaborate more about your thoughts.(I think this may be unrelated to the reference thing?)

Yeah My original thought is that it will just reflect the latest operation done to the original table. But yeah I think you are right it is better to just describe it is a clone from the source table.

> what if the cloned dataset overwritten or totally updated. Then it comes to a normal dataset which has no actual relations with the source dataset.

yeah I am only thinking using absolute path for the initial manifest content. if there are new fragments after new operations to the cloned table, then those will take relative path, and that is not ambiguous because we always use the source location to derive the full path if it is relative.

> For deletion file, there is not even a relative path, the file name is assembled. I need to add a path to the spec. I think it will work, just a little heavy for this case.

ah good point. I did not know that the delete file path is constructed instead of a relative path. 

Seems like if there is some importance in using the  constructed path for delete file, then I agree using reference is a good idea, the design looks like a good start. I think it needs to be a list of reference instead of just one, because there can be situations like I clone a table and then after a while I want to clone again, and then we will need to resolve file paths from multiple tables. And you might need to create some more actions like `DropReference` to fully remove a reference.

But it feels like hard-coding the delete file path could be limiting, if we want to do features like multi-tier storage and move files around tiers then it will become a blocker again. @wjones127 what were the considerations around this if you could share the historical context?

> I think I need to learn more from scenarios like import files you metioned. Apparently I don't get the whole picture. 

I am talking about features like https://iceberg.apache.org/docs/latest/spark-procedures/#add_files.

majin1102:
>  I think it needs to be a list of reference instead of just one, because there can be situations like I clone a table and then after a while I want to clone again, and then we will need to resolve file paths from multiple tables. 

Yes I think this is a good point. In my mind this was a nested reference like:

```
Reference {
  source_path: xxx
  version: 10
  max_fragment_id: 100
  ......
  reference: {
    source_path: xxxxxx
    version: 5
    max_fragment:70
    ......
   }
}
```

I think a reference list would be better

> And you might need to create some more actions like DropReference to fully remove a reference

I thougt a reference could be deleted automatically when the refered versions are cleaned up in the cloned dataset. Can you elaborate a little about the 'DropReference' scenario?

majin1102:
I think a field id list should be included in Reference struct for the case of merging column:

branch drawio

We will use the source root if the data file fragment id < reference.max_fragment_id and datafile field ids are all included in the reference

jackye1995:
I chatted with Will and Weston lately, here is a summary:

## Hard coded path

In general for the long term, we should have a reference for all file paths instead of having file paths. Originally paths like delete file are hard-coded to save manifest size.

For now, regarding the shallow clone feature, we should aim for just copying the files with hard coded paths to the cloned dataset. Over time, as we move more files as reference, shallow clone will copy less files.

## Relative paths for different bases

The easiest way to handle this is to just store absolute path, but that could inflate the overall manifest size. The reference approach you introduced is one way to handle that, Weston suggested another potentially simpler way:

For each path, we store the tuple of (base, path), where base is not the base path itself, but just a symbol. At manifest level, we store a base to URI mapping to resolve it. For example, consider mapping `{"other", "/tmp/bar"}`, then a path `(other, data/fragment1.lance)` will be resolved to `/tmp/bar/data/fragment1.lance`.

At protobuf deifnition level, this probably means an optional base in places like

```proto
message DataFile {
  // Relative path to the base
  string path = 1;
  ...
  // optional base, if not the root table location
  optional string base = 7;
}
```

For cloning/branching, we can rewrite the manifest to add a base for the paths in the manifest.

## Other optimizations

We should also add compression to manifest to keep the manifest size in control. Currently protobuf does very little with string compression.

@majin1102 please let us know what you think, if you agree with these ideas or have any other suggestions!

majin1102:
This approach is highly impressive to me—it seems to cover everything
- manifest size issue
- reference complexity
- extension for importing files.
- potentially the base list could be modified easily if we had the migration scenarios

I think some details could be discussed:
1. The initialized version number of cloned dataset. For example we clone the version of 3, will the cloned dataset initialized at 3 or 1. I think both ways are reasonable. For me I picked 3 for a clear relationship with the source dataset in the prototype.
2. I think this approach will leave reference only a string for base path. The new definition as I understand is like:
    `optional repeated string base_paths`
     Do we agree? @jackye1995 @westonpace 
3. Do we need a deep clone as metioned in #4256?(just copy the target snapshot using object_store api).  I think this approach is quite useful for those rich guys. WDYT @jackye1995

majin1102:
Hi, @jackye1995 

This PR is ready for review and the design document has been updated as well #4256 . PTAL when you have time.
cc @Xuanwo  I'm grateful if you can help reviewing

protos/table.proto
```diff
@@ -153,6 +153,7 @@ message Manifest {
   // If this value is 0 then there are no blob fields.
   uint64 blob_dataset_version = 17;
 
+  repeated string reference_paths = 18;
```
jackye1995: I was originally thinking this is a map (technically something like repeated tuple of string, string in protobuf). The reason is that if we use an index to reference the base path, we will run into the issue that we can never remove a reference path? Since that will change the position of the path_base_index. What do you think?

protos/table.proto
```diff
@@ -320,6 +321,8 @@ message DataFile {
   //
   // When this is zero, it should be interpreted as "unknown".
   uint64 file_size_bytes = 6;
+  // The base path of the data file. Used when the file is imported or referred from another dataset
+  optional uint32 path_base_index = 7;
```
jackye1995: nit: I think we can use a simpler name like `base` as long as the documentation is clear about what it means. 
nit2: When we say "Used when the file is imported or referred from another dataset", we should probably also mention that it should look up the actual base in `reference_paths`

protos/transaction.proto
```diff
@@ -154,6 +154,15 @@ message Transaction {
     uint32 num_fragments = 1;
   }
 
+  // An operation that clones a dataset.
+  message Clone {
```
jackye1995, line 158: nit: should add documentation for the meaning of these fields. I did not fully understand what these mean until I read the actual impl.

protos/table.proto
jackye1995: Also I think this field should be called `base_paths`. "reference" is specific to cloning, but this "base" concept can be used for other cases like multi-tier storage.

rust/lance/src/dataset.rs
```diff
@@ -1200,6 +1201,54 @@ impl Dataset {
         self.base.child(DATA_DIR)
     }
 
+    pub(crate) fn data_file_dir(&self, data_file: &DataFile) -> Result<Path> {
```
jackye1995, line 1203: Have you tested with indexes? We also have hard-coded paths and relative path references in indexes, I think we will need similar handling there.

jackye1995:
Really nice, thank you for continuing to push on this important feature! I looked through the rust part. Would it make more sense if we only have the rust changes here without the python and java changes, so we can iterate faster on the review?

protos/table.proto
majin1102: > I was originally thinking this is a map (technically something like repeated tuple of string, string in protobuf). The reason is that if we use an index to reference the base path, we will run into the issue that we can never remove a reference path? Since that will change the position of the path_base_index. What do you think?

I was using a map originally. But I encountered the issue that who could give the path key and be responsable that it won't conflict. Let users do that seems not reasonable. If let system decides I think an integer would be enough.

You are right about the remove case. I think we could use a `HashMap<u32, String>` as the base_paths type. What do you think.

majin1102: > Also I think this field should be called `base_paths`. "reference" is specific to cloning, but this "base" concept can be used for other cases like multi-tier storage.

Agree.
Actually I was on the fence about choosing base_paths or reference_paths

majin1102:
> Really nice, thank you for continuing to push on this important feature! I looked through the rust part. Would it make more sense if we only have the rust changes here without the python and java changes, so we can iterate faster on the review?

Yeah! 
I'll try this approach. Actually, I remember hitting some CI issues last time I worked on Rust—let me tackle that

protos/table.proto
majin1102: > nit: I think we can use a simpler name like base as long as the documentation is clear about what it means.

Since we're aligning rust and protobuf field names now, I suggest naming this integer field base_index. How do you feel?

protos/table.proto
jackye1995: > I think we could use a HashMap<u32, String> as the base_paths type. What do you think.

sounds good! Can we use a list of entries instead of map here? I think that is in general more backwards compatible than a map, like

```
repeated BasePath base_paths = 18;

message BasePath {
  uint32 id,
  string name,
  string path,
}
```

In this way, we can also assign a meaningful `name` to the base path. For the clone us case, this can just be for example the `ref_name`

protos/table.proto
jackye1995: > I suggest naming this integer field base_index. How do you feel?

+1

jackye1995:
> Actually, I remember hitting some CI issues last time I worked on Rust—let me tackle that

I see, if that's the case then let's just review everything all together

protos/table.proto
jackye1995: oh actually, if you agree with using entries instead of a map in base paths, then we should call this `base_id` instead of `base_idx`

majin1102: > oh actually, if you agree with using entries instead of a map in base paths, then we should call this `base_id` instead of `base_idx`

+1

protos/table.proto
majin1102: Totally agree with the BasePath

I think a map in manifest would be better for generating paths, or we will scan the paths to get the right one. The impact might be minor for now, but it feels unnecessary to introduce this overhead. What do you think?

majin1102:
> > Actually, I remember hitting some CI issues last time I worked on Rust—let me tackle that
> 
> I see, if that's the case then let's just review everything all together

The java and python modules must adapt to the new DataFile and DeletionFile struct(unlike Java, they don't use a constructor for compatibility). I'm afraid we can not spilt the PR now.

protos/table.proto
jackye1995: oh when loaded into Manifest struct, we can use a map for sure. I am taking about the protobuf definition only

majin1102:
> Have you tested with indexes? We also have hard-coded paths and relative path references in indexes, I think we will need similar handling there.

I'm currently working on index relocation (I initially planned to address indexes in a separate step). After analysis, I found that modifying index-related files would significantly increase the PR size and complicate the review process. Would it be acceptable to defer index handling to the next phase?  

In my opinion, shallow cloning without index support could already serve as a complete feature for experimental scenarios (apologies for mentioning this in the initial discussion draft and later removing it).  

For now, the cloned manifest will have `index_section` set to `None`, allowing the new dataset to build its own indexes independently of the source.  

I've adopted all the comments except the index. Please let me know if you have any concerns or alternative suggestions

protos/transaction.proto
```diff
@@ -154,6 +154,21 @@ message Transaction {
...
+    // if true, only clone the target manifest and don't copy the data. use base_paths field to reference the data.
+    // if false, copy the data and manifest at the target version. This should be a totally oss copy operation.
```
jackye1995: nit: what do you mean by "totally oss copy operation"?

rust/lance-table/src/format/fragment.rs
```diff
@@ -186,6 +206,21 @@ pub struct DeletionFile {
     pub file_type: DeletionFileType,
     /// Number of deleted rows in this file. If None, this is unknown.
     pub num_deleted_rows: Option<usize>,
+    pub base_id: Option<u32>,
+}
+
+impl Display for DeletionFile {
```
jackye1995: nit: why do we need this? this is basically exactly the same as the the Debug format

rust/lance-table/src/format/manifest.rs
```diff
@@ -176,6 +181,71 @@ impl Manifest {
...
+            .collect::<Vec<_>>();
+
+        Self {
+            schema: self.schema.clone(),
+            local_schema: self.local_schema.clone(),
+            version: self.version,
+            writer_version: self.writer_version.clone(),
+            fragments: Arc::new(cloned_fragments),
+            version_aux_data: self.version_aux_data,
+            index_section: None,
```
jackye1995, line 227: nit: please add a TODO since we plan to do it in a separated PR

rust/lance/src/dataset.rs
```diff
@@ -2670,6 +2740,99 @@ mod tests {
         )
     }
 
+    #[rstest]
+    #[tokio::test]
+    async fn clone_dataset(
+        #[values(LanceFileVersion::Legacy, LanceFileVersion::Stable)]
```
jackye1995, line 2757: I don't think the file format version makes a difference here?

rust/lance/src/dataset.rs
```diff
@@ -2670,6 +2740,99 @@ mod tests {
...
+        let test_dir = tempdir().unwrap();
+        let clone_path = test_dir.path().join("clone");
+        let cloned_dir = clone_path.to_str().unwrap();
+
+        let schema = Arc::new(ArrowSchema::new(vec![ArrowField::new(
+            "i",
+            DataType::Int32,
+            false,
+        )]));
+        let batches = vec![RecordBatch::try_new(
```
jackye1995: nit: please start to use `lance_datagen::gen_batch()` for all the tests going forward

rust/lance/src/dataset.rs
```diff
@@ -2670,6 +2740,99 @@ mod tests {
         )
     }
 
+    #[rstest]
+    #[tokio::test]
+    async fn clone_dataset(
```
jackye1995: nit: shallow_clone_dataset?

protos/table.proto
```diff
@@ -153,8 +153,27 @@ message Manifest {
...
+  // base_paths[id = 0] + /data/ + file.path
+  // the key(a.k.a index) starts from 0, increased by 1 for each new base path.
+  repeated BasePath base_paths = 18;
 } // Manifest
 
+// external dataset base path
+message BasePath {
+  uint32 id = 1;
+  string name = 2;
+  // in case that cloning and importing paths are used in different ways
```
jackye1995: nit: should explain this a bit more, I didn't understand what this means at all until reading the code. But I agree this is a nice feature for imported files!

rust/lance/src/io/deletion.rs
```diff
@@ -13,6 +13,7 @@ pub async fn read_dataset_deletion_file(
     fragment_id: u64,
     deletion_file: &DeletionFile,
 ) -> lance_core::Result<Arc<DeletionVector>> {
+    let deletion_dir = dataset.deletion_file_base_dir(deletion_file)?;
```
jackye1995: this is misleading, `read_deletion_file` takes in the dataset directory, not the _deletion directory:

```
/// Get the file path for a deletion file. This is relative to the dataset root.
pub fn deletion_file_path(base: &Path, fragment_id: u64, deletion_file: &DeletionFile) -> Path {
    let DeletionFile {
        read_version,
        id,
        file_type,
        ..
    } = deletion_file;
    let suffix = file_type.suffix();
    base.child(DELETION_DIRS)
        .child(format!("{fragment_id}-{read_version}-{id}.{suffix}"))
}
```

I think we should call the `deletion_file_base_dir` function `dataset_dir_for_deletion`, and then

```rust
let dataset_dir = dataset.dataset_dir_for_deletion(deletion_file)?;

read_deletion_file(
                fragment_id,
                deletion_file,
                &dataset_dir,
                dataset.object_store.as_ref(),
            )
```

For clarity.

rust/lance/src/dataset.rs
```diff
@@ -2670,6 +2740,99 @@ mod tests {
...
+            .await
+            .unwrap();
+
+        let expected_batch = RecordBatch::try_new(
+            schema.clone(),
+            vec![Arc::new(Int32Array::from_iter_values(0..40))],
+        )
+        .unwrap();
+
+        let actual_ds = Dataset::open(test_uri).await.unwrap();
```
jackye1995: the contents tested below do not make much sense to me... I think we should test (1) cloned dataset can be read after cloning, (2) cloned dataset can be written to again after cloning and then read back results correctly in a hybrid base paths case, (3) original dataset can be written again after cloning without impact, (4) we can clone multiple times, which will create multiple base paths, and then read and write still works.

rust/lance/src/dataset/transaction.rs
```diff
@@ -1239,6 +1349,9 @@ impl Transaction {
                 });
 
         match &self.operation {
+            Operation::Clone { .. } => {
+                unreachable!()
```
jackye1995: nit: should provide a more meaningful error like `Error::Internal`

rust/lance/src/io/commit/conflict_resolver.rs
```diff
@@ -435,9 +438,13 @@ impl<'a> TransactionRebase<'a> {
                 Operation::Merge { .. } => {
                     Err(self.retryable_conflict_err(other_transaction, other_version, location!()))
                 }
-                Operation::Overwrite { .. } | Operation::Restore { .. } => Err(
-                    self.incompatible_conflict_err(other_transaction, other_version, location!())
-                ),
+                Operation::Overwrite { .. }
+                | Operation::Restore { .. }
+                | Operation::Clone { .. } => Err(self.incompatible_conflict_err(
```
jackye1995: I think all operations should be compatible with Clone? Because Cloning does not have real effect against the source dataset, any concurrent operation should be able to just rebase and succeed.

jackye1995: Oh actually I think you are right. This is against the new table, so it should only happen once as the first transaction and it should conflict against all other transactions. We should make sure that is the case.

rust/lance/src/dataset.rs
majin1102: The transaction CommitBuilder need to use the property of lance_file_version:

```
        let builder = CommitBuilder::new(WriteDestination::Uri(target_path))
            .with_store_params(store_params)
            .with_object_store(Arc::new(self.object_store().clone()))
            .with_commit_handler(self.commit_handler.clone())
            .with_storage_format(self.manifest.data_storage_format.lance_file_version()?);
```

1. The lance file version must be inherented from the source dataset. So I think different format version is necessary.
2. The CommitBuilder uses this property and the .execute() uses it. I think applying appling the two versions is more safe

java/core/src/main/java/com/lancedb/lance/fragment/DeletionFile.java
```diff
@@ -25,12 +26,19 @@ public class DeletionFile implements Serializable {
...
   private final DeletionFileType fileType;
+  private final Integer pathBaseIndex;
 
-  public DeletionFile(long id, long readVersion, Long numDeletedRows, DeletionFileType fileType) {
+  public DeletionFile(
+      long id,
+      long readVersion,
+      Long numDeletedRows,
+      DeletionFileType fileType,
+      Integer pathBaseIndex) {
```
jackye1995: nit: variable name not upated, should be `baseId`

majin1102:
This is ready for another review. @jackye1995

protos/transaction.proto
```diff
@@ -154,6 +154,28 @@ message Transaction {
     uint32 num_fragments = 1;
   }
 
+  // An operation that clones a dataset.
+  message Clone {
+    // - true:  Performs a metadata-only clone (copies manifest without data files).
+    //          The cloned dataset references original data through `base_paths`,
+    //          suitable for experimental scenarios or rapid metadata migration.
+    // - false: Performs a full deep clone using the underlying object storage's native
```
jackye1995, line 162: if we actually do deep clone like what is described here, can we still commit the cloned dataset with this operation? It seems like we cannot, because we are just bulk-copying the directory to somewhere else?

rust/lance-table/src/format/fragment.rs
```diff
@@ -41,9 +41,24 @@ pub struct DataFile {
 
     /// The size of the file in bytes, if known.
     pub file_size_bytes: CachedFileSize,
+
+    /// The base path of the datafile, when the datafile is outside the dataset.
+    pub base_id: Option<u32>,
 }
 
 impl DataFile {
+    pub fn refer(datafile: &Self, base_id: u32) -> Self {
```
jackye1995: nit: it is not clear to me that `refer` actually means "refer to a base ID", can we use a more explicit function name like `with_base_id`?

protos/table.proto
```diff
@@ -152,8 +152,30 @@ message Manifest {
...
 } // Manifest
 
+// external dataset base path
+message BasePath {
+  uint32 id = 1;
+  string name = 2;
+  // Flag indicating whether this path is a dataset root path or file directory:
+  // - true:  Path is a dataset root (actual files under subdirectories like `data`, '_deletions')
+  // - false: Path is a direct file directory (scenario like importing files)
+  bool dataset_base = 3;
```
jackye1995: this name feels a bit confusing, because there is a "base path", and then this "dataset base" is a boolean, and also we don't really have a concept of "dataset base" so far, we have been calling it a "database root". Can we name it something like "is_dataset_root"?

rust/lance/src/dataset.rs
```diff
@@ -1199,6 +1200,51 @@ impl Dataset {
...
+                    Error::invalid_input(
+                        format!(
+                            "base_path id {} not found for deletion_file {:?}",
+                            base_id, deletion_file
+                        ),
+                        location!(),
+                    )
+                })?;
+
+                Ok(Path::parse(base_path.path.as_str())?)
```
jackye1995, line 1251: we should add a check that the BasePath must be a dataset root.

rust/lance/src/dataset.rs
```diff
@@ -1605,6 +1651,30 @@ impl Dataset {
         Ok(())
     }
 
+    pub async fn shallow_clone(
+        &mut self,
+        target_path: &str,
+        ref_name: &str,
+        store_params: ObjectStoreParams,
+    ) -> Result<Self> {
+        // self.tags.create(ref_name, self.version().version).await?;
```
jackye1995: remove this line? Or I think we can have an optional flag of `create_tag` which can run this if equal to true.

rust/lance/src/dataset.rs
```diff
@@ -1605,6 +1651,30 @@ impl Dataset {
         Ok(())
     }
 
+    pub async fn shallow_clone(
```
jackye1995, line 1667: we should add documentation given this is a public API

jackye1995 (approved):
I left some minor nit comments, but apart from that this looks good to me!

jackye1995:
Oh one more thing I forgot to mention, I think for this feature we need to update the `reader_feature_flags` and `writer_feature_flags` accordingly. We should not let an old reader read a cloned dataset since that will cause correctness issue (most likely it will fail). We should also not let an old writer write to a cloned dataset, since that will revert the base paths set in the manifest. I am okay we do that in this PR or separately, up to you.

protos/transaction.proto
majin1102: In my mental mind, we should copy each datafiles at the target version instead of bulk-copying the directory which could benifit:

1. Compared to copying the whole directory, we only focus the files at the target version which makes the cloning operation lightweight somehow.
2. Compared to read-and-write, we don't need to load arrow data and this should be faster somehow.
3. Moreover, I think we may provide some properties to config if we need multiple versions or include index, that will make this function more flexable.

What do you think?

protos/table.proto
majin1102: I must say the "is_dataset_root" flashed in my mind. I did't recall why did't choose that

rust/lance-table/src/format/fragment.rs
majin1102: Actually this is not used anymore. Now we clone the datafiles and directly set the value:

```
    let cloned_fragments = self
            .fragments
            .as_ref()
            .iter()
            .map(|fragment| {
                let mut cloned_fragment = fragment.clone();
                cloned_fragment.files = cloned_fragment
                    .files
                    .into_iter()
                    .map(|mut file| {
                        file.base_id = Some(new_base_id);
                        file
                    })
                    .collect();

                if let Some(mut deletion) = cloned_fragment.deletion_file.take() {
                    deletion.base_id = Some(new_base_id);
                    cloned_fragment.deletion_file = Some(deletion);
                }

                cloned_fragment
            })
```

I will delete it

rust/lance/src/dataset.rs
majin1102: Yeah, I think this is a to-be-discussed left behind.

Do you think we should bind shallow_clone to tags? or we could directly shallow_clone from a u32 version. (make the parameter as an `impl Into<refs::Ref>` like checkout_version)

Of course the lifetime of files are not garenteed. I think this is an authurity issue:  normally the shallow_cloning user should be a reader on the source dataset and shallow_clone itself like a reading operation on the source, binding to tags means this operation must depend on a tag write/manage operation, not that reasonable. Consider the case users only want to do a very short experiment on the latest verion, I think it could be independent to tags.

Also IMO auto create tag in shallow_clone is not very reasonable according to the least privilege princeple.

What do you think about this? This would make the ref_name as an `Option<String>` which quite makes sense to me @jackye1995

majin1102:
Newest update: address comments with two discussions left behind:

1. deep_clone description
2. tag decoupling

> We should also not let an old writer write to a cloned dataset, since that will revert the base paths set in the manifest. I am okay we do that in this PR or separately, up to you.

Then let me raise another PR to do this（immediately after this PR). Thank you

rust/lance/src/dataset.rs
jackye1995: Sounds good to me, agree separating it makes more sense

protos/transaction.proto
jackye1995: Today many users actually do directly copy a Lance dataset to cloud. This is especially popular for ML scientists that do the initial dataset locally then copy it and hand over to a team to bulk load the rest. I think what you say makes sense. Maybe if the storage supports it we can also copy the entire table directory except the version directory and then commit the deep cloned manifest.

Regarding configs to include index, technically that could be an option but I don't know why we would desire no index if we support it. 

For multiple version, that I am not so sure what would be the use case. Especially we consider the commit history to be linear. It's hard to comprehend what it means to have multiple Clone transactions on top of each other. 

But overall I think we don't need to hang on too long here, we can discuss these more in the Discussion thread.

majin1102:
> Sounds good to me, agree separating it makes more sense

Please take another look when you have time

jackye1995:
Thanks for all the work, looking forward to the next steps!!

Issue #4136, opened 2025-07-04
Expand and refresh Lance format spec


jackye1995:
With the documentation refactoring, I have moved the Lance format document to its separated section, to prepare for expanding it to multiple documents, describe things in more details and make sure it is up to date. 

My current thinking is to first separate it into table and file format. For table format, we talk about the basic manifest structure, indexes, transactions. For file format, we talk about the basic file layout, encodings, etc.

Xuanwo:
I somehow think myself is the best candidate for this issue 🤩, assigned.

jackye1995:
I am planning to do the table format part. If you can handle the file format part that would be perfect.

Xuanwo:
> I am planning to do the table format part. If you can handle the file format part that would be perfect.

Great!

jackye1995:
Actually I have another PR I am working on right now to fully close this. Reopen this until that PR is published and merged.

PR #4264, opened 2025-07-19
docs: separate table and file format spec

jackye1995:
The first step of #4136


docs/mkdocs.yml
```diff
@@ -76,20 +76,20 @@ nav:
       - Getting Started with Lance: quickstart/index.md
       - Versioning: quickstart/versioning.md
       - Vector Search: quickstart/vector-search.md
-    - Format Specification: format/index.md
```
jackye1995, line 79: the leveling was a bit odd right now, seems like user guide, format spec and community guide should not be nested in getting started. fixed along the way.

docs/src/format/table.md
```diff
@@ -0,0 +1,272 @@
+# Table Format
+
+## Dataset Directory
+
+A `Lance Dataset` is organized in a directory.
```
Xuanwo, line 5: Do we need to clarify that Lance Table and Lance Dataset are the same thing? (Are they?)

jackye1995:
Yeah that's my plan, this just separates things to modules first

docs/src/format/table.md
jackye1995: Yeah they are the same thing

PR #4267, opened 2025-07-20
feat: expose sql api for python dataset

Jay-ju:


Jay-ju:
@jackye1995 This PR mainly opens the SQL interface to the Python module. Personally, I think SQL is a quite useful function, so I made it exposed. Please take a look when you have time.

python/python/lance/dataset.py
```diff
@@ -2762,6 +2788,118 @@ def drop(
...
+    """
+    An executable SQL query.
+
+    This is created by calling :meth:`SqlQueryBuilder.build`.
+    """
+
+    def __init__(self, query):
+        self._query = query
+
+    def to_list(self) -> list[pa.RecordBatch]:
```
jackye1995: nit: `to_list` returning record batches and `to_batches` returning a reader seems a bit confusing, why not just use the same name as the rust ones?

src/dataset.rs
jackye1995, line 1: looks like unintentional change?

Jay-ju: yes, delete it

python/python/lance/dataset.py
Jay-ju: Actually, I don't think the Rust - side method can well represent the Python expression here. However, it still seems reasonable. Here, the Rust method name has been changed from `into_stream` to `to_stream_reader`.  Please take a look again.

jackye1995 (approved):
looks good to me!

PR #4272, opened 2025-07-21
chore: add format detail

Jay-ju:
There was no error in my local format, but there has been an error on the remote end. So, the details page has been printed out here.

This PR Before 
image

This PR After
image


Xuanwo (approved):
Thank you!

Xuanwo:
I believe the issue here is that we might be using different versions of `ruff`, which can be resolved by locking the versions of all the tools we're using with `uv`.

Issue #4176, opened 2025-07-08
Add `FileReaderOptions` to `ReadParams`


westonpace:
There are a few options that can be used to fine-tune the file reader's performance (e.g. caching full zip repetition index or skipping validation).  These are not configurable in the dataset API.   By adding a `Option<FileReaderOptions>` field to `ReadParams` we could allow users to specify these options if they desire.

Xuanwo:
I will work on this to get our rep index cache working

PR #4287, opened 2025-07-22
feat: add file reader options in ReadParams

Xuanwo:
Close https://github.com/lancedb/lance/issues/4176

This PR:

- Add `FileReaderOptions` in `ReadParams` so users can control the read behavior.
- Add `DecoderConfig` to refactor a bit of internal code.


rust/lance-file/src/v2/reader.rs
```diff
@@ -313,10 +313,7 @@ impl ReaderProjection {
 
 #[derive(Clone, Debug, Default)]
 pub struct FileReaderOptions {
```
westonpace, line 316: Not related to your PR but let's put a basic docstring on this type since is is part of public API?

westonpace (approved):
There's no test (to make sure the read params are passed down correctly) but I can't think of any easy way to verify that the read params are actually set.  Maybe when we add python bindings we can make a test that turns on `cache_repetition_index` and measures the number of bytes read from disk (this can be returned by scanner).

rust/lance-file/src/v2/reader.rs
Xuanwo: Added.

PR #4289, opened 2025-07-22
fix: fix can not scan and take correct rowids after compaction when enable move stable rowid feature

yanghua:


yanghua:
Open for watching CI, still WIP, not ready for review

yanghua:
@wjones127 Would you please take a look?

rust/lance-table/src/rowids.rs
```diff
@@ -439,6 +439,17 @@ impl RowIdSequence {
         }
         ranges
     }
+
+    /// Convert a range with bitmap segment to sorted array
+    /// so that we can exclude invalid row ids.
+    pub fn convert_rwb_to_sorted_array(&mut self) -> Result<()> {
```
jackye1995: sorry do not have time to look into details at the moment, could you explain a bit about why this way to fix? It seems like when enable stable row ID, the column produces more rows than expected when scanning. In that case my intuition is that some logic about RangeWithBitmap is wrong and should be fixed, instead of converting to the SortedArray?

yanghua: > the column produces more rows than expected when scanning

Yes, the phenomenon is what you said. 

Here, https://github.com/lancedb/lance/blob/f86f196b8a41a06b36af66262b283e2289fb3026/rust/lance/src/dataset/optimize.rs#L803

In this method, it would remove deleted rows. Some deleted rows would be marked as `RangeWithBitmap`.
While, in this method:

https://github.com/lancedb/lance/blob/f86f196b8a41a06b36af66262b283e2289fb3026/rust/lance/src/dataset/optimize.rs#L846

It would slice the `RangeWithBitmap`, the `RangeWithBitmap` actually contains the `deleted rows`(bit is '0', and length is total rows). So when slice, it seems it's hard to exclude them.

Currently, I tried to convert to `SortedArray` to solve it. Also open for discussion.

jackye1995: sorry I wanted to check some ideas but somehow pushed to your branch... feel free to revert...

but overall it seems like the slice logic is not correct, the start and end should not be calculated like that for RangeWithHoles and RangeWithBitmap because of the holes. I vibe coded the fix and it seems like it is working. But I agree I don't know if this would be more efficient or not compared to directly convert it to a sorted array. I guess there has to be some threshold to decided what type of segment to convert to after slicing to make the storage efficient and perf also reasonably good.

yanghua: > sorry I wanted to check some ideas but somehow pushed to your branch

Please feel free to do this.

> I guess there has to be some threshold to decided what type of segment to convert to after slicing to make the storage efficient.

Sounds reasonable. Do we need any input from others?

jackye1995: Looks like Will already defined that:

```
/// Different ways to represent a sequence of distinct u64s.
///
/// This is designed to be especially efficient for sequences that are sorted,
/// but not meaningfully larger than a Vec<u64> in the worst case.
///
/// The representation is chosen based on the properties of the sequence:
///                                                           
///  Sorted?───►Yes ───►Contiguous?─► Yes─► Range            
///    │                ▼                                 
///    │                No                                
///    │                ▼                                 
///    │              Dense?─────► Yes─► RangeWithBitmap/RangeWithHoles
///    │                ▼                                 
///    │                No─────────────► SortedArray      
///    ▼                                                    
///    No──────────────────────────────► Array            
///
/// "Dense" is decided based on the estimated byte size of the representation.
///
/// Size of RangeWithBitMap for N values:
///     8 bytes + 8 bytes + ceil((max - min) / 8) bytes
/// Size of SortedArray for N values (assuming u16 packed):
///     8 bytes + 8 bytes + 8 bytes + 2 bytes * N
///
```

so during slicing we could try to find a way to estimate if the end result should be a RangeWithBitMap or SortedArray.

Looks like it is not clearly defined about when to use RangeWithBitmap vs RangeWithHoles, but we should be able to derive the math and update the definition.

yanghua: > Looks like it is not clearly defined about when to use RangeWithBitmap vs RangeWithHoles, but we should be able to derive the math and update the definition.

It seems the logic is here: https://github.com/lancedb/lance/blob/bdfb8fc58f0a20830dea9d1164bac3edc778498e/rust/lance-table/src/rowids/segment.rs#L164

jackye1995: Oh nice!

yanghua: I refactored the `slice` method with the existing `from_slice`. Seems it works. If you agree, I would refactor this PR based on this!

jackye1995: Sounds good! I was thinking if `U64Segment::SortedArray` and `U64Segment::Array` can be optimized through their `EncodedU64Array::slice`, but it seems like it will recreate a `Vec<64>` anyway, so I think your implementation should be equivalent and it's much cleaner.

rust/lance-table/src/rowids/segment.rs
```diff
@@ -273,49 +273,119 @@ impl U64Segment {
...
+        if len == 0 {
+            return Self::Range(0..0);
+        }
 
-                let start = holes.binary_search(start).unwrap_or_else(|x| x) as u64;
-                let end = holes.binary_search(end).unwrap_or_else(|x| x) as u64;
-                let holes_len = end - start;
+        let slice_iter = self.iter().skip(offset).take(len);
+        
+        let values: Vec<u64> = slice_iter.collect();
```
jackye1995: nit: why not just `let values = self.iter().skip(offset).take(len).collect<Vec<_>>()`

yanghua:
Hi @jackye1995, please take a look, thanks!

Issue #3727, opened 2025-04-23
Rework data loaders


westonpace:
We have a basic data loader for pytorch.  However, this data loader has several limitations and is not very intuitive.  There are also other training libraries (e.g. Ray Train, TensorFlow, etc.) and it would be good to ensure a consistent experience across all of them.  This issue captures a lot of feedback (and we welcome more!) that we have gathered over the years in hopes of motivating a new design.

## Clearly document fork-safety

Spawn is considerably slower than fork.  In addition, users have often missed our guidelines on fork safety in the past and ended up frustrated with a hung process.  I think https://github.com/lancedb/lance/pull/3584 will help here but there is more we could do.

* Improve our documentation on fork safety (we should improve our documentation on data loading in general, see below)
* Make lance fork-safe.  I _think_ this would mostly involve resetting static state (e.g. RT, LOOP) after a fork.  However, this is very difficult to guarantee in general and would likely be an on-going challenge.
* Document how to use forkserver.  It is very possible to safely use forkserver in lance.  One just needs to take care to ensure that lance is loaded _after_ the forkserver is started.  Improved documentation on data loading should demonstrate this.

## Provide both map-based and iterator-based datasets

We currently only have an implementation of torch's iterator-based datasets.  Several users have created their own map-based data loaders and these are useful in a variety of situations.  However, properly implementing a map-based data loader can be tricky (e.g. need to understand row addresses and pin to version, should use slightly hidden `__getitems__` method, not clear how to do shuffling with restarts, etc.)

## Simplify filtering & shuffling

The filtering and shuffling in the current data loader is confusing.  Pytorch's sampler is confusing.  I think a much simpler model could be devised in Lance.  From Lance perspective, these are the shuffling scenarios I am aware of and the approaches that should be taken.

| Scenario | Approach |
| --- | --- |
| No shuffling, No filter | If there is no shuffling and there is no filter then an iterable dataset should be used for best performance |
| No shuffling, With filter | If there is no shuffling but there is a filter then we should decide between scan (iterable dataset) and take (map dataset) based on the filter selectivity (assuming there is an index on the filter we can do this cheaply) |
| Shard shuffling, No filter | If there is no filter and the user wants shard shuffling then an iterable dataset should be used.  The shard size should be completely independent of the fragment size.  There is no benefit to joining these two things and it muddies the waters.  Offsets & limits should be used for sharding. |
[...]

PR #4291, opened 2025-07-22
docs: torch dist training guide

AyushExel:


docs/src/dist-training-torch/ddp_intro.md
```diff
@@ -0,0 +1,63 @@
...
+
+*   **Lance Integration:** `lance.torch.data.SafeLanceDataset`
+*   **Best For:** Finite, indexable datasets. This covers the vast majority of standard use cases, such as image classification (ImageNet, FOOD101), text classification, or any scenario where the dataset is a discrete collection of samples.
+
+### Iterable-Style: `torch.utils.data.IterableDataset`
+An iterable-style dataset works like a Python generator. It implements the `__iter__(self)` method, which yields data items sequentially as a stream. It often has no known length and does not support random access to a specific index. This makes it ideal for handling data that doesn't fit the indexed collection model.
+
+*   **Lance Integration:** `lance.torch.data.LanceDataset`
+*   **Best For:** True streaming data sources (e.g., from a database cursor or real-time log), extremely large datasets where building a full index is impractical, or when implementing complex, custom sampling logic that goes beyond what standard PyTorch samplers offer.
```
westonpace: The `Best For` section here is a little confusing.  What is an infinite Lance dataset?  All Lance datasets can be either map style or iterable style.  You explain how to choose them in the next section, do you also need a "best for" here?

docs/src/dist-training-torch/ddp_intro.md
```diff
@@ -0,0 +1,63 @@
...
+| **Performance** | **Highest**. The recommended default for throughput. | **Lower**. I/O is serialized in a single process. |
+| **Parallelism** | **High**. Data decoding is parallelized across multiple workers. | **Low**. Data is fetched and decoded sequentially in a single stream. |
+| **Shuffling** | Natively supported via PyTorch samplers (e.g., `DistributedSampler`).  | Requires custom implementation (e.g., buffering data). |
+| **Epoch Length** | Deterministic. The dataset has a known length (`__len__`).  | Not deterministic by default. Requires manual loop control. |
+| **PyTorch `num_workers`** | `> 0` (Recommended for performance). | `0` (mutli-worker requires custom implementation). |
+| **Sampler** | PyTorch `DistributedSampler`. | Custom (E.g. Lance `ShardedBatchSampler`. ) |
+
+It is recommended to use map‑style datasets by default. Map‑style give you their size ahead of time, easier to shuffle, and allow for easy parallel loading. But if you're reading from a massive data source remotely and/or you need to use custom sampling logic, you should use iterable style dataloader. Common techniques and pitfalls of each approach is discussed in their respective sections.
```
westonpace: Reading this table I don't see any real reason to use an iterable style dataset for Lance?  Doesn't pytorch support a custom sampler for map datasets?

docs/src/dist-training-torch/iterable_style.md
```diff
@@ -0,0 +1,90 @@
+# Training on Stream data with `LanceDataset` (Iterable-Style)
+
+The basic iterable-style pattern is for use cases where you need to stream data sequentially or implement custom sampling logic. This can be due to various reasons like - your data is tool large to fit in a node or in a massively parallel multi-node multi-GPU setting, you need to read data from a central source, etc. It uses the `LanceDataset` and a Lance-native sampler.
```
westonpace: > your data is tool large to fit in a node or in a massively parallel multi-node multi-GPU setting, you need to read data from a central source

I don't understand these reasons.  A map dataset does not materialize the entire dataset in memory does it?

docs/src/dist-training-torch/iterable_style.md
```diff
@@ -0,0 +1,90 @@
...
+
+**Important:** This simple pattern requires setting `num_workers=0` in the `DataLoader`, which means data will be loaded serially. This is significantly slower than the map-style approach and should only be used when the flexibility of iterable datasets is a requirement.
```
westonpace, line 5: I thought `LanceDataset` supported multiple workers?

westonpace (commented):
I'm not sure about the differences described between our map and iterable datasets.  That being said, I'm not sure why the user wouldn't always use the map dataset.

I think the only advantage to the iterable dataset is its ability to do shards which can increase performance (fewer IOPS).  However, this guide describes `LanceDataset` as lower performance.

Note: I'm also working on https://github.com/lancedb/lance/issues/3727 at the moment although its probably going to lance in `lancedb` and not `lance` (so we can still have this document).

docs/src/dist-training-torch/iterable_style.md
AyushExel: This is not a lance thing, its more of a Pytorch iterable dataset thing in general.  in DDP setting if you set num_workers>1, it'll double count the batches. So there multiple workers need to be synced manually. See this discussion i've added at the bottom - https://discuss.pytorch.org/t/iterable-pytorch-dataset-with-multiple-workers/135475/3

docs/src/dist-training-torch/ddp_intro.md
AyushExel: Hmm yeah let me rephrase that a little.. The main intention of the intro part is to not be lance specific. We get questions like - whats the difference bw the different approaches and why should I use one over other? 
The general recommendation is that - Always use map-style unless you can't. Its probably better to simply say that lance does not materialize the entire dataset in memory, so you should always stick to Map-style

AyushExel:
>  I think the only advantage to the iterable dataset is its ability to do shards which can increase performance (fewer IOPS). However, this guide describes LanceDataset as lower performance.

I'm not sure if this is true in a DDP setting. To be able to truly use multiple workers per GPU, you'll need to open multiple streams and sync them, something like [this](https://docs.pytorch.org/docs/stable/data.html#torch.utils.data.IterableDataset) in pytorch i thnk..something like this? Or is there another way recommended for lance datasets


```
class LanceIterableDDP(IterableDataset):
    def __init__(self, path, batch_size, ddp_sampler, rank, world_size, num_workers):
        super().__init__()
        self.path = path
        self.batch_size = batch_size
        self.ddp_sampler = ddp_sampler
        self.rank = rank
        self.world_size = world_size
        self.num_workers = num_workers

    def __iter__(self):
        w = get_worker_info()
        worker_id = w.id if w else 0
        num_workers = w.num_workers if w else 1

        shard_index = self.rank * num_workers + worker_id
        total_shards = self.world_size * num_workers

        print(f"[Rank {self.rank}] Worker {worker_id} using shard {shard_index}/{total_shards}")

        ds = LanceDataset(
            self.path,
            batch_size=self.batch_size,
            to_tensor_fn=decode_tensor_image,
            sampler=self.ddp_sampler,
            shard_granularity="fragment",
            rank=shard_index,
            world_size=total_shards
        )

        return iter(ds)
```
So it'll only be faster if they implement something like this ?

docs/src/dist-training-torch/iterable_style.md
AyushExel: Yeah let me rephrase this part to be lance specific

docs/src/dist-training-torch/ddp_intro.md
AyushExel: yeah makes sense.. removed it from here

AyushExel:
> I think the only advantage to the iterable dataset is its ability to do shards which can increase performance (fewer IOPS). However, this guide describes LanceDataset as lower performance.

@westonpace  Okay just went through the `LanceDataset` implementation in lance.. I see what you mean now. It adds a sharded sampler automatically if not provided in DDP mode unlike vanilla torch. I'll redo this PR.

btw the work you're doing on LanceDB dataloaders, will that deprecate Lance<>torch integration? If so, I'll just directly try that and not bother with this.

Issue #4159, opened 2025-07-07
Retry-based conflict resolution for Delete


wjones127:
We can add basic support for concurrent deletes by retrying them when there are conflicts. Resolving non-overlapping conflicts optimally is left to a later issue https://github.com/lancedb/lance/issues/3796

In Merge Insert, we implemented a retry loop in `execute()`:

https://github.com/lancedb/lance/blob/0fad40d676b24d9355a6b2dbe626111684f1696a/rust/lance/src/dataset/write/merge_insert.rs#L1059-L1163

This allows the operation to be retries if there were conflicting operations.

We should apply this same logic in the `delete` dataset method:

https://github.com/lancedb/lance/blob/0fad40d676b24d9355a6b2dbe626111684f1696a/rust/lance/src/dataset/write/delete.rs#L13

Key requirements:
* [ ] Create a `DeleteBuilder` similar to `UpdateBuilder` or `MergeInsertBuilder`
* [ ] Add a `conflict_retries()` setter on the builder so the user can control how many retries are performed.
* [ ] Implement the retry loop. Use `SlotBackoff` for backoff implementation.
* [ ] Add a unit test validating concurrent deletes can succeed.

wjones127:
Closed #4292

PR #4292, opened 2025-07-22
feat: add retry-based conflict resolution for delete operations (#4159)

wjones127:
## Summary

This PR adds retry-based conflict resolution for delete operations, addressing issue #4159. Delete operations now support automatic retries on commit conflicts, similar to existing update and merge_insert operations.

## Changes

- **New retry infrastructure**: Added `retry.rs` module with common retry logic
- **DeleteBuilder**: New fluent API for configuring delete operations with retry support  
- **Comprehensive tests**: Added tests for concurrent deletes and timeout behavior
- **Backwards compatibility**: Legacy `delete()` method still works unchanged

## Usage

```rust
// New builder API with retry support
let new_dataset = DeleteBuilder::new(dataset)
    .predicate("age > 50")
    .conflict_retries(5)
    .retry_timeout(Duration::from_secs(60))
    .execute()
    .await?;

// Legacy API still works (no retries)
dataset.delete("age > 50").await?;
```

## Test Plan

- [x] All existing tests pass
- [x] New concurrent delete test with retries  
- [x] New retry timeout test
- [x] Code passes clippy and formatting checks

The retry mechanism follows the same patterns as existing update/merge_insert operations for consistency.

### DIFF
--- a/rust/lance/src/dataset/write/delete.rs
+++ b/rust/lance/src/dataset/write/delete.rs
@@ -9,11 +9,15 @@ use crate::{
 use datafusion::logical_expr::Expr;
 use datafusion::scalar::ScalarValue;
 use futures::{StreamExt, TryStreamExt};
-use lance_core::Result;
+use lance_core::{Error, Result};
 use lance_table::format::Fragment;
 use roaring::RoaringTreemap;
 use std::collections::BTreeMap;
 use std::sync::{Arc, RwLock};
+use std::time::Duration;
+
+use super::retry::{execute_with_retry, RetryConfig, RetryExecutor};
+use super::CommitBuilder;
 
 /// Apply deletions to fragments based on a RoaringTreemap of row IDs.
 ///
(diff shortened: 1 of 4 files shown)


python/python/tests/test_dataset.py
```diff
@@ -323,9 +323,6 @@ def test_checkout(tmp_path: Path):
     assert ds1.version == 2
     assert ds1.to_table() == pa.table({"a": [0, 2]})
 
-    with pytest.raises(IOError):
-        ds2.delete("a = 2")
-
```
wjones127, line 328: This was saying that if we checked out an old version, we couldn't do a write. This was failing, but not quite for correct reasons. Currently, we don't have a way to differentiate checking out an old version on purpose and being not on the latest version because a concurrent writer has performed a write. We should fix this in a later PR.

python/python/lance/dataset.py
```diff
@@ -1501,7 +1501,12 @@ def drop_columns(self, columns: List[str]):
         # Indices might have changed
         self._list_indices_res = None
 
-    def delete(self, predicate: Union[str, pa.compute.Expression]):
+    def delete(
+        self,
+        predicate: Union[str, pa.compute.Expression],
+        conflict_retries: int = 10,
```
wjones127, line 1508:
```suggestion
        *,
        conflict_retries: int = 10,
```

python/python/tests/test_dataset.py
jackye1995: > Currently, we don't have a way to differentiate checking out an old version on purpose and being not on the latest version because a concurrent writer has performed a write.

I have actually been relying on this to do testing of conflicts. What would be the difference between these 2 cases?

> ds2.delete("a = 2")

With the row level conflict resolution, shouldn't this succeed because they don't touch the same row?

rust/lance/src/dataset/write/delete.rs
```diff
@@ -64,82 +68,212 @@ async fn apply_deletions(
...
+
+    /// Set the timeout for retry operations
+    pub fn retry_timeout(mut self, timeout: Duration) -> Self {
+        self.retry_timeout = timeout;
+        self
+    }
+
+    /// Execute the delete operation
+    pub async fn execute(self) -> Result<Arc<Dataset>> {
+        let predicate = self.predicate.ok_or_else(|| Error::InvalidInput {
```
jackye1995: predicate is required, then why not directly do

```rust
pub fn new(dataset: Arc<Dataset>, predicate: impl Into<String>) -> Self {
        Self {
            dataset,
            predicate,
            conflict_retries: 10,
            retry_timeout: Duration::from_secs(30),
        }
    }
```

python/python/tests/test_dataset.py
wjones127: > What would be the difference between these 2 cases?

If a user intentionally checked out an old version, we should not allow writes. That's actually how we handle things in LanceDB: There's a `Latest` and a `TimeTravel` mode: https://github.com/lancedb/lancedb/blob/296205ef96377b7681c6f4d77bfe97a7aeea16f8/rust/lancedb/src/table/dataset.rs#L26-L35

And if you try to get a mutable reference to the dataset while in `TimeTravel` mode, we raise an error:

https://github.com/lancedb/lancedb/blob/296205ef96377b7681c6f4d77bfe97a7aeea16f8/rust/lancedb/src/table/dataset.rs#L225-L234

rust/lance/src/dataset/write/retry.rs
```diff
@@ -0,0 +1,134 @@
...
+    fn default() -> Self {
+        Self {
+            max_retries: 10,
+            retry_timeout: Duration::from_secs(30),
+        }
+    }
+}
+
+/// Trait for operations that can be retried on commit conflicts
+pub trait RetryExecutor: Clone {
```
jackye1995, line 34: nice refactoring!

python/python/tests/test_dataset.py
wjones127: > With the row level conflict resolution, shouldn't this succeed because they don't touch the same row?

I should succeed, yeah. I just deleted the condition because what that was testing is if you checked out an old version and then tried to do a write it would fail.

rust/lance/src/dataset/write/retry.rs
```diff
@@ -0,0 +1,134 @@
...
+#[derive(Debug, Clone)]
+pub struct RetryConfig {
+    pub max_retries: u32,
+    pub retry_timeout: Duration,
+}
+
+impl Default for RetryConfig {
+    fn default() -> Self {
+        Self {
+            max_retries: 10,
```
jackye1995, line 27: we also set the same defaults separately at delete, merge-insert, update params level, and then there is another higher level default in python. Since we have this shared retry config now, can we just make the upper levels configs optional so it all default to this automatically?

wjones127: We could. Honestly, I like having the defaults visible in the Python signature. That way users can see immediately what the default is, without having to look it up in the docs.

python/python/tests/test_dataset.py
jackye1995: > And if you try to get a mutable reference to the dataset while in TimeTravel mode, we raise an error

I see. Are you suggesting you plan to also add such guard in Lance? (or is it already there and I just did not notice?)

It feels like there can be advanced use cases like streaming, where you will have checkpoints, and when resuming a streaming job, you want to intentionally resume at the last checkpointed version and then write data so you can go through conflict resolution against commits happened after the checkpointed version.

jackye1995 (approved):
mostly looks good to me, pending rebase

rust/lance/src/dataset/write/delete.rs
```diff
@@ -66,88 +70,195 @@ async fn apply_deletions(
...
+}
+
+impl RetryExecutor for DeleteJob {
+    type Data = DeleteData;
+    type Result = Arc<Dataset>;
+
+    async fn execute_impl(&self) -> Result<Self::Data> {
+        // Create a single scanner for the entire dataset
+        let mut scanner = self.dataset.scan();
+        scanner.with_row_id().filter(&self.predicate)?;
```
wjones127, line 161: This should be made an empty projection. Otherwise, we are going to read all columns, which would be wasteful. We just need `.with_row_id()`.

wjones127: We should exclusively project the row id, since we don't need to scan other columns.
```suggestion
        scanner.project(&[ROW_ID]).filter(&self.predicate)?;
```

rust/lance/src/dataset/write/delete.rs
```diff
@@ -162,244 +273,324 @@ mod tests {
         data_storage_version: LanceFileVersion,
         #[values(false, true)] with_scalar_index: bool,
     ) {
-        fn sequence_data(range: Range<u32>) -> RecordBatch {
-            let schema = Arc::new(ArrowSchema::new(vec![
-                ArrowField::new("i", DataType::UInt32, false),
```
wjones127: The changes to this test and others seem unnecessary. Please leave all existing tests unchanged.

python/python/tests/test_dataset.py
wjones127: > Are you suggesting you plan to also add such guard in Lance?

Yeah, was contemplating that. We don't have anything like that.

> It feels like there can be advanced use cases like streaming, where you will have checkpoints, and when resuming a streaming job, you want to intentionally resume at the last checkpointed version and then write data so you can go through conflict resolution against commits happened after the checkpointed version.

Oh that's an interesting idea. I have no immediate plans to change things, so no rush to flesh this out.

PR #4295, opened 2025-07-22
feat: implement `explain_plan` and `analyze_plan` for `merge_insert` operations

### LINKED ISSUE #4195: Provide `explain_plan` and `analyze_plan` for merge_insert
Now that at least some queries generate a full DataFusion plan for execution, we can add `explain_plan` and `analyze_plan` to the list.

`explain_plan()` should be straightforward: We should just generate the plan using `create_plan` and then run explain on it.

https://github.com/lancedb/lance/blob/94dcaf5e735480f9f61216b4e4f90358dbeac2ba/rust/lance/src/dataset/write/merge_insert.rs#L1222-L1225

Note that we can only call `create_plan` in some circumstances, guarded currently here:

https://github.com/lancedb/lance/blob/94dcaf5e735480f9f61216b4e4f90358dbeac2ba/rust/lance/src/dataset/write/merge_insert.rs#L1352-L1365

`analyze_plan()` requires a bit more care. What we should do is run analyze but make sure we don't pass the transaction to `Commit`.

You can see how we do analyze plan with scans here:

https://github.com/lancedb/lance/blob/94dcaf5e735480f9f61216b4e4f90358dbeac2ba/rust/lance-datafusion/src/exec.rs#L392-L415

wjones127:
## Summary

Implements `explain_plan` and `analyze_plan` methods for merge_insert operations to support debugging and performance analysis.

## New APIs

- `explain_plan(schema: Optional[Schema] = None, verbose: bool = False)` - generates execution plans without executing. Schema defaults to dataset schema when not provided.
- `analyze_plan(source: ReaderLike, *, schema: Optional[Schema] = None)` - executes plan with real data and collects performance metrics including `bytes_written` and `num_files_written`

Both methods only work when fast path conditions are met (exact schema match, no scalar index, etc.) and return `Error::NotSupported` otherwise.

### explain_plan() Usage

```python
import lance
import pyarrow as pa

# Create dataset
data = pa.table([
    pa.array([1, 2, 3]),
    pa.array(['alice', 'bob', 'charlie']),
    pa.array([100.0, 200.0, 300.0])
], names=['id', 'name', 'score'])

dataset = lance.write_dataset(data, '/tmp/example')
builder = dataset.merge_insert('id').when_matched_update_all().when_not_matched_insert_all()

# Generate execution plan (uses dataset schema by default)
plan = builder.explain_plan()
print(plan)
```

**Output:**
```
MergeInsert: on=[id], when_matched=UpdateAll, when_not_matched=InsertAll, when_not_matched_by_source=Keep
  CoalescePartitionsExec
    ProjectionExec: expr=[_rowaddr@1 as _rowaddr, id@2 as id, name@3 as name, score@4 as score, CASE WHEN __common_expr_1@0 AND _rowaddr@1 IS NULL THEN 2 WHEN __common_expr_1@0 AND _rowaddr@1 IS NOT NULL THEN 1 ELSE 0 END as action]
      ProjectionExec: expr=[id@1 IS NOT NULL as __common_expr_1, _rowaddr@0 as _rowaddr, id@1 as id, name@2 as name, score@3 as score]
        CoalesceBatchesExec: target_batch_size=8192
          HashJoinExec: mode=CollectLeft, join_type=Right, on=[(id@0, id@0)], projection=[_rowaddr@1, id@2, name@3, score@4]
            LanceRead: uri=example/data, projection=[id], num_fragments=1, range_before=None, range_after=None, row_id=false, row_addr=true, full_filter=--, refine_filter=--
            RepartitionExec: partitioning=RoundRobinBatch(32), input_partitions=1
              StreamingTableExec: partition_sizes=1, projection=[id, name, score]
```

### analyze_plan() Usage

```python
# Create source data for analysis  
source_data = pa.table([
    pa.array([1, 4]),  # 1 matches existing, 4 is new
    pa.array(['alice_updated', 'david']),
    pa.array([150.0, 400.0])
], names=['id', 'name', 'score'])

# Analyze performance with real data
analysis = builder.analyze_plan(source_data)
print(analysis)
```

**Output:**
```
MergeInsert: on=[id], when_matched=UpdateAll, when_not_matched=InsertAll, when_not_matched_by_source=Keep, metrics=[output_rows=0, elapsed_compute=1ns, bytes_written=821, num_deleted_rows=0, num_files_written=1, num_inserted_rows=1, num_updated_rows=1]
  CoalescePartitionsExec, metrics=[output_rows=2, elapsed_compute=48.842µs]
    ProjectionExec: expr=[_rowaddr@1 as _rowaddr, id@2 as id, name@3 as name, score@4 as score, CASE WHEN __common_expr_1@0 AND _rowaddr@1 IS NULL THEN 2 WHEN __common_expr_1@0 AND _rowaddr@1 IS NOT NULL THEN 1 ELSE 0 END as action], metrics=[output_rows=2, elapsed_compute=226.974µs]
      ProjectionExec: expr=[id@1 IS NOT NULL as __common_expr_1, _rowaddr@0 as _rowaddr, id@1 as id, name@2 as name, score@3 as score], metrics=[output_rows=2, elapsed_compute=12.477µs]
        CoalesceBatchesExec: target_batch_size=8192, metrics=[output_rows=2, elapsed_compute=22.547µs]
          HashJoinExec: mode=CollectLeft, join_type=Right, on=[(id@0, id@0)], projection=[_rowaddr@1, id@2, name@3, score@4], metrics=[output_rows=2, build_input_batches=1, build_input_rows=3, input_batches=1, input_rows=2, output_batches=1, build_mem_used=172, build_time=930.562µs, join_time=203.691µs]
            LanceRead: uri=example/data, projection=[id], num_fragments=1, range_before=None, range_after=None, row_id=false, row_addr=true, full_filter=--, refine_filter=--, metrics=[output_rows=3, elapsed_compute=964.093µs, bytes_read=845, fragments_scanned=1, iops=2, ranges_scanned=1, requests=2, rows_scanned=3, task_wait_time=963.714µs]
            RepartitionExec: partitioning=RoundRobinBatch(32), input_partitions=1, metrics=[fetch_time=120.756µs, repartition_time=1ns, send_time=8.809µs]
              StreamingTableExec: partition_sizes=1, projection=[id, name, score], metrics=[]
```

## Key Features

- **Optional Schema**: `explain_plan()` defaults to dataset schema for convenience
- **Performance Metrics**: `analyze_plan()` includes `bytes_written` and `num_files_written` for storage analysis  
- **Clean Output**: AnalyzeExec/TracedExec wrapper lines are filtered out for better readability
- **Fast Path Only**: Methods work with full schema match, no scalar index, and keep unmatched rows

## Implementation Details

- Extracted `can_use_create_plan()` method to centralize fast path validation logic
- Added comprehensive Rust unit tests including plan structure validation
- Python bindings provide the same API surface with optional parameters
- Uses `lance_datagen` for test data generation

Fixes #4195


docs/merge_insert_analyze_plan_examples.md
wjones127, line 1: We don't need to check this in.

docs/merge_insert_analyze_plan_examples.md
```diff
@@ -0,0 +1,83 @@
...
+Merge insert configuration:
+- `when_matched: UpdateAll`
+- `when_not_matched: InsertAll` 
+- `when_not_matched_by_source: Keep` (default)
+
+## analyze_plan(data_stream, verbose=false)
+
+```
+AnalyzeExec verbose=true, metrics=[]
+  TracedExec, metrics=[]
```
wjones127: Can we find a way to remove these lines? They are just confusing.

rust/lance/src/dataset/write/merge_insert.rs
```diff
@@ -1501,6 +1522,82 @@ impl MergeInsertJob {
...
+    /// * `verbose` - If true, provides more detailed information in the analysis output
+    ///
+    /// # Errors
+    ///
+    /// Returns Error::NotSupported if the merge insert configuration doesn't support
+    /// the fast path required for plan generation.
+    pub async fn analyze_plan(
+        &self,
+        source: SendableRecordBatchStream,
+        _verbose: bool,
```
wjones127: If we don't use it, let's remove the `verbose` parameter from `analyze_plan`.

rust/lance/src/dataset/write/merge_insert.rs
```diff
@@ -3138,4 +3241,405 @@ mod tests {
             StreamingTableExec: partition_sizes=1, projection=[value, key]"
         ).await.unwrap();
     }
+
+    #[tokio::test]
+    async fn test_explain_plan() {
```
wjones127: It seems like we should be able to consolidate these test cases. For example, the  `test_explain_plan_string_output` could be part of the basic `test_explain_plan`.

rust/lance/src/dataset/write/merge_insert.rs
```diff
@@ -3138,4 +3241,405 @@ mod tests {
...
+        )
+        .unwrap();
+
+        let dataset = Dataset::write(
+            RecordBatchIterator::new(vec![Ok(batch.clone())], schema.clone()),
+            "memory://test_explain_plan",
+            None,
+        )
+        .await
+        .unwrap();
```
wjones127: For brevity, use `lance_datagen` to create the test dataset. Here's an example from another test:

https://github.com/lancedb/lance/blob/5d72cbf6b6f147ff09169fb86c0cb8b24933546f/rust/lance/src/dataset/optimize.rs#L2809-L2817

wkalt:
@wjones127 what's the reason for not supporting it if the index exists? I would think an important usecase would be understanding if the index is getting used.

wjones127:
> @wjones127 what's the reason for not supporting it if the index exists? I would think an important usecase would be understanding if the index is getting used.

We are migrating from the old code path, which doesn't use query plans, to a new code path, which does. We can only explain/analyze plan if there is a plan.

It's too much work to do the migration in one shot, so I am doing it case by case. You can track the progress of this work in this milestone:

https://github.com/lancedb/lance/milestone/7

So it will eventually be supported when indices exist.

wkalt (approved):

python/python/lance/dataset.py
```diff
@@ -205,6 +205,153 @@ def retry_timeout(self, timeout: timedelta) -> "MergeInsertBuilder":
...
+            ProjectionExec: expr=[_rowaddr@1 as _rowaddr, id@2 as id, ...]
+              ProjectionExec: expr=[id@1 IS NOT NULL as __common_expr_1, ...]
+                CoalesceBatchesExec: target_batch_size=...
+                  HashJoinExec: mode=CollectLeft, join_type=Right, ...
+                    ...
+        """
+        if schema is None:
+            return super(MergeInsertBuilder, self).explain_plan(verbose=verbose)
+        else:
+            return super(MergeInsertBuilder, self).explain_plan(schema, verbose=verbose)
```
westonpace:
```suggestion
        return super(MergeInsertBuilder, self).explain_plan(schema, verbose=verbose)
```
Do you need the if check here?  Since the default value of `schema` is `None` it seems redundant

python/python/lance/dataset.py
```diff
@@ -205,6 +205,153 @@ def retry_timeout(self, timeout: timedelta) -> "MergeInsertBuilder":
...
+        schema: Optional[pa.Schema] = None,
+    ) -> str:
+        """
+        Generate the execution plan and analyze its performance with real data.
+
+        This method creates the execution plan using the provided source data,
+        executes it to collect performance metrics, and returns the analysis
+        as a formatted string. This is useful for understanding the actual
+        performance characteristics of the merge insert operation with your data.
+
```
westonpace, line 291: Should we add a warning that this will execute the operation and insert new data, modifying the dataset?  I feel like it's slightly different from query analyze_plan in that regard.

python/python/lance/dataset.py
```diff
@@ -205,6 +205,153 @@ def retry_timeout(self, timeout: timedelta) -> "MergeInsertBuilder":
...
+        Parameters
+        ----------
+        data_obj : ReaderLike
+            The source data that would be merged into the dataset. This parameter
```
westonpace, line 302:
```suggestion
            The source data that will be merged into the dataset. This parameter
```

python/python/lance/dataset.py
```diff
@@ -205,6 +205,153 @@ def retry_timeout(self, timeout: timedelta) -> "MergeInsertBuilder":
...
+        - num_inserted_rows: rows added that didn't match existing data
+        - num_updated_rows: existing rows that were updated
+        - num_deleted_rows: rows removed (when using delete conditions)
+
+        Key metrics in LanceRead operations:
+        - bytes_read: bytes read from storage
+        - fragments_scanned: number of data fragments accessed
+        - rows_scanned: total rows examined during the scan
+        - iops: number of I/O operations performed
+        - requests: number of storage requests made
```
westonpace, line 359: This seems useful but maybe a bit out of context.  Perhaps a sentence above explaining what these key metrics are?

westonpace (approved):
Nicely done.  Just one note on wording to make it clear that `analyze_plan` can modify the dataset.

python/python/lance/dataset.py
wjones127: This wasn't very clear, but I actually designed this *not* to modify the dataset. It runs the operation and writes data files. But it doesn't attempt to commit the transaction. (The execution plan itself doesn't do the commit.)

westonpace: Ah, I missed that, and agree it is a good idea.

PR #4297, opened 2025-07-23
ci: enable crates.io trust publishing

Xuanwo:
This PR will enable trust publishing for all our rust crates. After this change, we don't need to have static key from crates.io anymore.


BubbleCal (approved):
let's make a preview release after merging this

PR #4298, opened 2025-07-23
feat: add loading trace event

yanghua:


rust/lance/src/dataset/builder.rs
```diff
@@ -271,6 +271,7 @@ impl DatasetBuilder {
 
     #[instrument(skip_all)]
     pub async fn load(mut self) -> Result<Dataset> {
```
yanghua, line 281: The codes like this `ds = lance.dataset("/tmp/test.lance")` will not trigger `opening` event. It would trigger this load method(both `write` and `read` paths would trigger this method). So it would be better to add a `loading` event.

freesinger (approved):

rust/lance-core/src/utils/tracing.rs
```diff
@@ -84,3 +84,4 @@ pub const DATASET_DROPPING_COLUMN_EVENT: &str = "dropping_column";
 pub const DATASET_DELETING_EVENT: &str = "deleting";
 pub const DATASET_COMPACTING_EVENT: &str = "compacting";
 pub const DATASET_CLEANING_EVENT: &str = "cleaning";
+pub const DATASET_LOADING_EVENT: &str = "loading";
```
wjones127, line 86: Could you describe the different between opening and loading? I think that would be confusing to most people.

rust/lance/src/dataset/builder.rs
wjones127: The code path with the existing `opening` event also calls `load()`:

https://github.com/lancedb/lance/blob/184c16a6891349f9a22bc891bae2e2514275d58e/rust/lance/src/dataset.rs#L352-L355

So why not just move the `opening` event to the `load()` method?

rust/lance-core/src/utils/tracing.rs
wjones127: Actually, is there a good reason to have these two as separate events?

jackye1995: yeah I think we should only have one. If we want to differentiate between cached open vs non-cached open, Weston has a good point at https://github.com/lancedb/lance/discussions/4234#discussioncomment-13760219 that we should trace that through file opening trace events.

rust/lance/src/dataset/builder.rs
yanghua: The original intention is to track only the `read` operation. I misunderstood that `open` can do this. The `load` method is called in both `read` and `write` operations. 

Finally, I found that the `loading` event traces in the `load` method can also be accepted.

Now my thought is to introduce `loading` in the `load` method, and remove `opening`(if a dataset is first written, `opening` may be confused). WDYT?

rust/lance-core/src/utils/tracing.rs
yanghua: > Could you describe the different between opening and loading?

It's misunderstood. I thought we had a clear path to distinguish `write` and `read`.

* `writing` for only writing data;
* `opening` for only opening the existing dataset;

The reality is:

* currently, read and write are both called `load` method;
* The current `opening` could not track the main scenario(`lance.dataset('xxxx')`)

I did not remove the `opening` event. Because I'm not sure if the `opening` is still useful for someone who called the `open` method in Rust.

rust/lance/src/dataset/builder.rs
jackye1995: > The original intention is to track only the read operation

if that is the case, I think we should add tracing event in the scanner

> The load method is called in both read and write operations.

`Dataset::open` also eventually calls `DatasetBuilder::load`, so I think we can remove the one in `Dataset::open`. At that point, it seems like it's just a matter of calling it `open` or `load`, I am fine with both.

yanghua: > so I think we can remove the one in Dataset::open

Yes.

> At that point, it seems like it's just a matter of calling it open or load, I am fine with both.

I prefer `load`.

yanghua:
@jackye1995 please take a look, thanks~

wjones127 (approved):

PR #4301, opened 2025-07-23
feat(java): support scalar index for java sdk

NIDIEJIE:


NIDIEJIE:
[issue #4243 ](https://github.com/lancedb/lance/issues/4243)

java/core/src/main/java/com/lancedb/lance/index/IndexParams.java
```diff
@@ -29,6 +29,16 @@ private IndexParams(Builder builder) {
     this.vectorIndexParams = builder.vectorIndexParams;
   }
 
+  /**
+   * Creates an empty IndexParams instance with default values. This is useful for non-vector index
+   * types where IndexParams can be empty.
+   *
+   * @return an empty IndexParams instance
+   */
+  public static IndexParams empty() {
```
jackye1995: This feels like a hack. Currently the `IndexParams` is basically `VectorIndexParams` because it supports vector index only. I think to support scalar, we should properly make `IndexParams` an interface, and have different implementations for different index types. What do you think?

NIDIEJIE: Yes, but currently I've found that scalar indexes basically don't require parameters except for inverted indexes. So I wrote it this way. This is a simple implementation method. The inverted index supported at the back can be directly loaded into it. Do we need to make the granularity very fine.

jackye1995: Thought about this a bit more, probably need some historical context from @LuQQiu, why do we have distance type both in `IndexParams` and `VectorIndexParams`? Is the intent to have `IndexParams` be a parent holding different index params for different types?

Also there was `distanceType` in both `IndexParams` and `VectorIndexParams`, is that a mistake or have some reasons for that?

jackye1995: Sorry forgot to circle back on this @NIDIEJIE . I think you have a good point. Looks like the intention is for `IndexParams` to hold different specific index parameters, and the `distanceType` should be removed and use the one in `VectorIndexParams`. With that, I think the current code makes sense to have an empty `IndexParams`.

java/core/src/main/java/com/lancedb/lance/Dataset.java
```diff
@@ -572,7 +572,23 @@ public void createIndex(
...
-      nativeCreateIndex(columns, indexType.getValue(), name, params, replace);
+      boolean isVectorIndex =
+          indexType == IndexType.VECTOR
+              || indexType == IndexType.IVF_FLAT
+              || indexType == IndexType.IVF_SQ
+              || indexType == IndexType.IVF_PQ
+              || indexType == IndexType.IVF_HNSW_SQ
+              || indexType == IndexType.IVF_HNSW_PQ;
+
+      if (isVectorIndex) {
```
jackye1995: the params should always be non-null, even for scalar index we will initialize an empty one

NIDIEJIE: Got it.

jackye1995:
Hi @NIDIEJIE, just want to check do you still plan to work on this?

NIDIEJIE:
> Hi @NIDIEJIE, just want to check do you still plan to work on this?

I'll keep doing it. I've been a bit busy these days. I'll finish it this weekend

jackye1995:
Sounds good, looking forward to it!

java/core/src/main/java/com/lancedb/lance/index/IndexParams.java
```diff
@@ -42,7 +49,6 @@ public Builder() {}
      * @return this builder
      */
     public Builder setDistanceType(DistanceType distanceType) {
```
jackye1995: we should just remove the whole method

java/core/src/main/java/com/lancedb/lance/index/IndexParams.java
jackye1995: Feels like `empty()` is not the right name for what it means, this is representing the default parameter not empty. I think we should just add a normal `public static Builder builder()` method so that users can do `IndexParams.builder().build()` to represent the default params. (since default is a keyword in java)

jackye1995: Actually I remember now the convention is to have a `public static IndexParams create()` to represent default values. But the builder one is still good to have I think.

java/core/src/main/java/com/lancedb/lance/Dataset.java
```diff
@@ -556,12 +556,12 @@ public void restore() {
...
 
   /**
-   * Creates a new index on the dataset. Only vector indexes are supported.
+   * Creates a new index on the dataset.
    *
    * @param columns the columns to index from
    * @param indexType the index type
    * @param name the name of the created index
-   * @param params index params
+   * @param params index params, can be null for non-vector index types
```
jackye1995: I think it will not be null? since user will use `IndexParams.create()` it will always be a non-null value. It does not handle null in the JNI code but I think it's probably better to enforce it, because we explicitly use `Optional` for nullable objects in methods.

java/core/src/main/java/com/lancedb/lance/index/IndexParams.java
NIDIEJIE: Got it.

java/core/src/main/java/com/lancedb/lance/Dataset.java
NIDIEJIE: Forgot to update the comments. I will recover it .

NIDIEJIE:
> Mostly looks good to me! Can you add a few tests for creating scalar indexes?

Ok.

java/core/src/main/java/com/lancedb/lance/index/IndexParams.java
NIDIEJIE: Got it.

NIDIEJIE:
@jackye1995 PTAL when you have a chance. Thanks!

java/src/main/java/com/lancedb/lance/index/IndexParams.java
```diff
@@ -21,31 +21,36 @@
...
     this.vectorIndexParams = builder.vectorIndexParams;
   }
 
+  /**
+   * Creates IndexParams instance with default values. This is useful for non-vector index types
+   * where vectorIndexParams can be empty.
+   *
+   * @return an empty IndexParams instance without vectorIndexParams
+   */
+  public static IndexParams create() {
```
jackye1995, line 36: sorry things has changed a lot since the last review, as you can see we have made scalar index pluggable. So now we can actually do something like this:

```
IndexParams.builder()
  .setScalarIndexParams(ScalarIndexParams.builder()
        .indexType(...)
        .paramsJson(...)
        .build();
```

We can still keep this create method to mean nothing, but the description should be changed to something like this to be more generic in description:

```
  /**
   * Creates IndexParams instance with default values.
   *
   * @return an empty IndexParams instance
   */
```

java/lance-jni/src/blocking_dataset.rs
```diff
@@ -631,7 +632,18 @@ fn inner_create_index(
...
+    let params: Box<dyn IndexParams> = match index_type {
+        IndexType::BTree => Box::new(ScalarIndexParams::new(ScalarIndexType::BTree)),
+        IndexType::Bitmap => Box::new(ScalarIndexParams::new(ScalarIndexType::Bitmap)),
+        IndexType::LabelList => Box::new(ScalarIndexParams::new(ScalarIndexType::LabelList)),
+        // TODO: using default params for inverted index temporarily, need to extract params from params_jobj
+        IndexType::Inverted => Box::new(ScalarIndexParams::new(ScalarIndexType::Inverted)),
+        IndexType::NGram => Box::new(ScalarIndexParams::new(ScalarIndexType::NGram)),
+        // scalar index alias for btree index
+        IndexType::Scalar => Box::new(ScalarIndexParams::new(ScalarIndexType::BTree)),
+        _ => get_index_params(env, params_jobj)?,
```
jackye1995, line 645: we should rename this to `get_vector_index_params`, and we need another one `get_scalar_index_params` which will convert a Java `ScalarIndexParams` to Rust `ScalarIndexParams`. 

Inside `get_scalar_index_params`, we can have this logic here if ScalarIndexParams is empty, but if user has provided the param, it should be honored.

jackye1995 (commented):
Thank you very much for continuing this effort! Sorry we recently did a refactoring, so I added a few comments to make this PR more aligned with the latest structure. Sorry for keep making you update this, really appreciate your help!

jackye1995:
Merged https://github.com/lancedb/lance/pull/4683 as coauthor, thanks for the work!

PR #4307, opened 2025-07-24
[WIP] chore(rust): distinguish row id and row addr

yanghua:


yanghua:
Still WIP...not ready for review.

rust/lance/src/dataset/optimize.rs
```diff
@@ -755,11 +755,11 @@ async fn rewrite_files(
...
 
         if options.defer_index_remap {
-            let mut changed_row_addrs = Vec::with_capacity(row_ids.serialized_size());
-            row_ids.serialize_into(&mut changed_row_addrs)?;
+            let mut changed_row_addrs = Vec::with_capacity(row_addrs.serialized_size());
+            row_addrs.serialize_into(&mut changed_row_addrs)?;
             (None, Some(changed_row_addrs))
         } else {
-            let row_id_map = remapping::transpose_row_ids(row_ids, &fragments, &new_fragments);
+            let row_id_map = remapping::transpose_row_ids(row_addrs, &fragments, &new_fragments);
```
yanghua: @jackye1995 Shall we rename this to `transpose_row_addrs`?

jackye1995: yeah I think that makes sense, the stable row ID path does not use this anyway it's purposely used for row address

yanghua:
Maybe we should do a big refactor for `mask.rs`:

* RowIdTreeMap -> RowAddrTreeMap;
* RowIdSelection -> RowAddrSelection;
* RowIdMask -> RowAddrMask;

cc @jackye1995

yanghua:
The scope of refactoring seems larger and larger(I found it also touches `index` sub-module). What about splitting it into multiple sub-tasks? @jackye1995

jackye1995:
> The scope of refactoring seems larger and larger(I found it also touches index sub-module). What about splitting it into multiple sub-tasks?

For sure!

PR #4309, opened 2025-07-24
fix: support count(*)

wojiaodoubao:
Currently 'SELECT count(*) from table' will throw error below.
```
Err(IO { source: Internal("Physical input schema should be the same as the one converted from logical input schema. Differences: \n\t- Different number of fields: (physical) 3 vs (logical) 0"), location: Location { file: "rust/lance/src/bin/sql_test.rs", line: 23, column: 19 } })
```

That's because 'count(*)' uses an empty projection, and currently we don't support it.


rust/lance/src/dataset/scanner.rs
```diff
@@ -472,6 +472,18 @@ impl Scanner {
         self.fragments.is_some()
     }
 
+    /// Empty Projection
+    ///
+    /// The first column will be scanned.
```
jackye1995: feels like we should request the row address column since we don't really need a scan on this. Ideally this should be a query optimization to just use Lance metadata to return result, but that can be a separated feature.

wojiaodoubao: > feels like we should request the row address column since we don't really need a scan on this

Thanks, I'll try the row address.

> Ideally this should be a query optimization to just use Lance metadata to return result

Yes, If there is no filter, then the metadata should be enough. Otherwise, we still need scanner to filter the results.

jackye1995: Oh yeah good point, if there is a filter we should use the filter column since that needs to be scanned anyway

wojiaodoubao: > if there is a filter we should use the filter column since that needs to be scanned anyway

If I understand correctly, building '_rowaddr' in a scan is very light weight. So the cost of scanning with 'filter' and 'filter column' + '_rowaddr' should be the same. 

So shall we always use row address column like you suggested before? @jackye1995

rust/lance/src/datafusion/dataframe.rs
```diff
@@ -96,7 +96,9 @@ impl TableProvider for LanceTableProvider {
                     columns.push(self.full_schema.field(*field_idx).name());
                 }
             }
-            if !columns.is_empty() {
+            if projection.is_empty() {
```
jackye1995: this condition can happen before 

```
let mut columns = Vec::with_capacity(projection.len());
            for field_idx in projection {
                if Some(*field_idx) == self.row_id_idx {
                    scan.with_row_id();
                } else if Some(*field_idx) == self.row_addr_idx {
                    scan.with_row_address();
                } else {
                    columns.push(self.full_schema.field(*field_idx).name());
                }
            }
```

to avoid initializing the columns and computing all the logic above.

rust/lance/src/dataset/scanner.rs
```diff
@@ -477,6 +477,15 @@ impl Scanner {
         self.fragments.is_some()
     }
 
+    /// Empty Projection
+    ///
+    /// The row_address will be scanned.
+    pub fn empty_project(&mut self) -> Result<&mut Self> {
+        self.with_row_address();
```
jackye1995, line 484: you also need to `self.project(&vec![ROW_ADDR])` in order to not select the other columns

wojiaodoubao: Thanks! Let me fix it.

rust/lance/src/dataset/scanner.rs
```diff
@@ -477,6 +477,16 @@ impl Scanner {
...
+        let _ = self.project(&[ROW_ADDR]);
```
jackye1995: I think you need to do `self.project(&[ROW_ADDR])?` in order to make sure the operation has actually succeeded.

Or one other easy way is just return it as the last line like

```rust
pub fn empty_project(&mut self) -> Result<&mut Self> {
        self.with_row_address();
        self.projection_plan.final_projection_is_empty = true;
        self.project(&[ROW_ADDR])
}

File: rust/lance-datafusion/src/projection.rs [hunk 5]
```diff
@@ -29,6 +29,11 @@ pub struct ProjectionPlan {
     base: Arc<dyn Projectable>,
     /// The physical schema that must be loaded from the dataset
     pub physical_projection: Projection,
+    /// True if the final schema of datafusion execution plan is empty.
+    /// e.g.
+    ///   select count(*) from dataset.
+    ///   select from dataset.
```
jackye1995: nit: is this a valid SQL query?

wojiaodoubao: It is weird, but Datafusion can accept this sql, result is like this.
```
Ok([RecordBatch { schema: Schema { fields: [], metadata: {} }, columns: [], row_count: 6 }])
```

wojiaodoubao: Let me remove it, it is confusing.

Issue #4302, opened 2025-07-23
Lance Encoding Configuration System


Xuanwo:
Lance now supports multiple encoding methods such as RLE, bit-packing, and BSS. Each has its own advantages and trade-offs, making it difficult to choose a default that works well in every scenario. 

My current idea is:

- First build an encoding configuration system that allows users to statically select encodings for their data, based on column names or data types. 
- Next, we can benchmark different datasets to determine suitable default settings. 
- This way, we can offer guidance to users on making the right choices, such as using byte stream splits together with zstd for time series data. 
- Finally, we can evaluate table statistics automatically and perform a second round of encoding during compaction.

This issue covers the first part: building an encoding configuration system. 

## Guide-level explanation

Users can control data encoding based on the name and type of each column as follows:

```python
compression_overrides = {
    "columns": {
        "*_id": "fsst",         
        "embedding*": "none",    
        "logs": "zstd:6",        
    },
    "types": {
        "string": "lz4",  
        "double": "rle+lz4",    
        "float": "bit_packing+byte_stream_split+zstd",     
    }
}
dataset.write(data, compression_overrides=compression_overrides)
```

## Reference-level explanation

We will add two hashmaps (one for `columns` and the other for `types`) in `DefaultCompressionStrategy` and select the compressor based on the settings. 

The hashmaps will be passed down through: `FileWriterOptions` -> `FieldEncodingStrategy` -> `StructuralEncodingStrategy` -> `DefaultCompressionStrategy`.

### Configure via column metadata

We can support `lance-compression:rle+lz4` through column metadata. However, I feel this approach makes it difficult to discover possible values and increases the risk of incorrect usage. It also becomes tedious for users when they have 3,000 columns to configure.

westonpace:
My hope is that default settings will work well enough for most scenarios.  If users have to specify encodings then they are probably doing something advanced.

> We can support lance-compression:rle+lz4 through column metadata. However, I feel this approach makes it difficult to discover possible values and increases the risk of incorrect usage. It also becomes tedious for users when they have 3,000 columns to configure.

Can we have both?  I think the advantage of column metadata is that it doesn't have to be specified each time a write occurs (since the config is set on the dataset schema).  If a user wants to write a column in a particular way it seems likely they will continue to want to write a column in the same way.

Also, for the type overrides, perhaps they can be set on the dataset schema metadata as well?

**Re: type overrides**

[...]

PR #4310, opened 2025-07-24
feat: add compression override support for encoding

Xuanwo:
Part of https://github.com/lancedb/lance/issues/4302

This PR adds the compression override supprot for encoding at the rust side.

In general, I did the following changes:

- Add `CompressionOverrides` along with `CompressionSpec` which can specify the compression methods user want to use based on column name and types.
- Add `ConfiguredCompressionStrategy` that can apply compression specs in order, like `rle -> lz4`
- End to end tests from the FileWriter side.

Things I think can follow-up:

- Expose settings to python users
- Allow configure from field metadata

### DIFF
--- a/rust/lance-encoding/src/compression.rs
+++ b/rust/lance-encoding/src/compression.rs
@@ -116,21 +116,19 @@ pub trait CompressionStrategy: Send + Sync + std::fmt::Debug {
 
 #[derive(Debug, Default)]
 pub struct DefaultCompressionStrategy {
-    /// Optional user-configured compression parameters
-    params: Option<CompressionParams>,
+    /// User-configured compression parameters
+    params: CompressionParams,
 }
 
 impl DefaultCompressionStrategy {
     /// Create a new compression strategy with default behavior
     pub fn new() -> Self {
-        Self { params: None }
+        Self::default()
     }
 
     /// Create a new compression strategy with user-configured parameters
     pub fn with_params(params: CompressionParams) -> Self {
-        Self {
-            params: Some(params),
-        }
+        Self { params }
     }
 
     /// Build compressor based on parameters for fixed-width data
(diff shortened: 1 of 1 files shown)


rust/lance-encoding/src/compression_config.rs
```diff
@@ -0,0 +1,246 @@
...
+    /// RLE threshold (0.0-1.0)
+    /// When run_count < num_values * threshold, RLE will be used
+    pub rle_threshold: Option<f64>,
+
+    /// General compression scheme: "lz4", "zstd", "none"
+    pub compression: Option<String>,
+
+    /// Compression level (only for schemes that support it, e.g., zstd)
+    pub compression_level: Option<i32>,
+}
```
westonpace, line 35: In the old approach we used field metadata and so configuration parameters were in a string-string hashmap.  One thing we had considered at one point that seemed interesting to people was eventually getting to the point where new encodings could be added as plugins.

If we did want to do plugins then we could easily support new configuration parameters because it was just a string-string hashmap.

Can this approach also work with plugins?

rust/lance-encoding/src/configured_compression_strategy.rs
```diff
@@ -0,0 +1,470 @@
...
+
+    fn create_block_compressor(
+        &self,
+        field: &Field,
+        data: &DataBlock,
+    ) -> Result<(Box<dyn BlockCompressor>, pb::ArrayEncoding)> {
+        // Block compression doesn't use parameters currently
+        self.default_strategy.create_block_compressor(field, data)
+    }
+}
```
westonpace: This code seems pretty similar to what's already in `DefaultCompressionStrategy`.  Instead of making a new `ConfiguredCompressionStrategy` can we just modify the existing default strategy to use compression params?

westonpace (commented):
A few high level questions

rust/lance-encoding/src/compression_config.rs
Xuanwo: Nice question. I think it depends on how we register those plugins; maybe we can register possible options at the same time?

rust/lance-encoding/src/configured_compression_strategy.rs
Xuanwo: I added `ConfiguredCompressionStrategy` to avoid the existing logic, but I'm perfectly fine with just adding it to `DefaultCompressionStrategy` instead. I'll go ahead and make this change.

rust/lance-encoding/src/compression.rs
```diff
@@ -114,14 +115,181 @@ pub trait CompressionStrategy: Send + Sync + std::fmt::Debug {
...
+                    // FSL doesn't support compression currently
+                    return Ok(Box::new(ValueEncoder::default()));
+                }
+                _ => {
+                    // Fall through to default behavior for unsupported types
+                }
+            }
+        }
+
+        // Default behavior (no user parameters)
```
westonpace: There is still duplication here.  Everything below this point is duplicated with everything above this point.  I think the decision tree for choosing which compression to use could get rather complicated and I want to avoid having it in two different places.

What if we do this (change `params` from `Option<T>` to `T` with default)...

```
pub struct DefaultCompressionStrategy {
    /// Optional user-configured compression parameters
    params: CompressionParams,
}

impl DefaultCompressionStrategy {
    /// Create a new compression strategy with default behavior
    pub fn new() -> Self {
        Self { params: CompressionParams::default() }
    }

    /// Create a new compression strategy with user-configured parameters
    pub fn with_params(params: CompressionParams) -> Self {
        Self {
            params,
        }
    }
```

Then we can add a default impl for `CompressionParams`.

Xuanwo: Makes sense to me, I'll implement this and merge.

PR #4311, opened 2025-07-24
fix: add hnsw range search support

LeoReeYang:
- fix https://github.com/lancedb/lancedb/issues/2473

### DIFF
--- a/rust/examples/src/hnsw.rs
+++ b/rust/examples/src/hnsw.rs
@@ -131,7 +131,7 @@ async fn main() {
             upper_bound: None,
         };
         let results: HashSet<u32> = hnsw
-            .search_basic(q.clone(), k, params, None, vector_store.as_ref())
+            .search_basic(q.clone(), k, &params, None, vector_store.as_ref())
             .unwrap()
             .iter()
             .map(|node| node.id)
--- a/rust/examples/src/hnsw.rs
+++ b/rust/examples/src/hnsw.rs
@@ -20,7 +20,10 @@ use lance::Dataset;
 use lance_index::vector::v3::subindex::IvfSubIndex;
 use lance_index::vector::{
     flat::storage::FlatFloatStorage,
-    hnsw::{builder::HnswBuildParams, HNSW},
+    hnsw::{
+        builder::{HnswBuildParams, HnswQueryParams},
+        HNSW,
+    },
 };
 use lance_linalg::distance::DistanceType;
 
@@ -122,8 +125,13 @@ async fn main() {
         let construct_time = now.elapsed().as_secs_f32();
         let now = std::time::Instant::now();
         // 3. Perform vector search with different parameters and compute the ground truth using L2 distance search
+        let params = HnswQueryParams {
+            ef,
+            lower_bound: None,
+            upper_bound: None,
+        };
         let results: HashSet<u32> = hnsw
-            .search_basic(q.clone(), k, ef, None, vector_store.as_ref())
+            .search_basic(q.clone(), k, params, None, vector_store.as_ref())
             .unwrap()
             .iter()
             .map(|node| node.id)
--- a/rust/lance-index/src/vector/graph.rs
+++ b/rust/lance-index/src/vector/graph.rs
@@ -288,7 +288,11 @@ pub fn beam_search(
     let upper_bound: OrderedFloat = params.upper_bound.unwrap_or(f32::MAX).into();
 
     let mut results = BinaryHeap::with_capacity(k);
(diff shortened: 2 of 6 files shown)


rust/lance-index/src/vector/hnsw/builder.rs
```diff
@@ -165,7 +165,7 @@ impl HNSW {
         &self,
         query: ArrayRef,
         k: usize,
-        ef: usize,
+        params: HnswQueryParams,
```
BubbleCal: it's worth making this a reference

rust/examples/src/hnsw.rs
```diff
@@ -123,7 +126,17 @@ async fn main() {
...
+                q.clone(),
+                k,
+                HnswQueryParams {
+                    ef,
+                    lower_bound: None,
+                    upper_bound: None,
+                },
+                None,
+                vector_store.as_ref(),
+            )
```
BubbleCal:
```suggestion
let params = HnswQueryParams{ ef, lower_bound: None, upper_bound: None};
let results: HashSet<u32> = hnsw
            .search_basic(
                q.clone(),
                k,
                &params,
                None,
                vector_store.as_ref(),
            )
```

rust/lance-index/src/vector/hnsw/builder.rs
```diff
@@ -703,7 +721,7 @@ impl IvfSubIndex for HNSW {
                 prefilter_bitset.expect("the prefilter bitset must be set for flat search");
             self.flat_search(storage, query, k, prefilter_bitset)
```
BubbleCal: we need to handle distance range in `flat_search` as well, or it will be still ignored if the query is with a filter that matching less than 10% data

BubbleCal (changes requested):

rust/lance-index/src/vector/graph.rs
```diff
@@ -312,7 +319,10 @@ pub fn beam_search(
 
         let process_neighbor = |neighbor: u32| {
             visited.insert(neighbor);
-            let dist = dist_calc.distance(neighbor).into();
+            let dist: OrderedFloat = dist_calc.distance(neighbor).into();
+            if is_range_search && (dist < lower_bound || dist > upper_bound) {
```
BubbleCal:
```suggestion
            if dist < lower_bound || dist >= upper_bound {
```

BubbleCal: this causes your test failed

rust/lance-index/src/vector/hnsw/builder.rs
```diff
@@ -543,13 +547,27 @@ impl Graph for HnswBottomView<'_> {
 #[derive(Debug)]
```
BubbleCal:
```suggestion
#[derive(Debug, Clone, Copy)]
```
then we don't need to implement `Clone` manually

rust/lance-index/src/vector/graph.rs
```diff
@@ -312,7 +318,10 @@ pub fn beam_search(
 
         let process_neighbor = |neighbor: u32| {
             visited.insert(neighbor);
-            let dist = dist_calc.distance(neighbor).into();
+            let dist: OrderedFloat = dist_calc.distance(neighbor).into();
+            if dist < lower_bound || dist >= upper_bound {
+                return;
+            }
             if dist <= furthest || results.len() < k {
                 if bitset
```
BubbleCal, line 327:
```suggestion
                if dist >= lower_bound && dist < upper_bound && bitset
```

BubbleCal: move the check of distance range here,
the current impl would break the graph because it returns early, we don't add the unmatched node into `results`, but still need to add it into `candidates`

BubbleCal (changes requested):

rust/lance-index/src/vector/hnsw/builder.rs
```diff
@@ -205,7 +205,7 @@ impl HNSW {
         &self,
         query: ArrayRef,
         k: usize,
-        ef: usize,
+        params: HnswQueryParams,
```
BubbleCal: let's make this reference as well

rust/lance-index/src/vector/hnsw/builder.rs
```diff
@@ -241,6 +241,7 @@ impl HNSW {
         query: ArrayRef,
         k: usize,
         prefilter_bitset: Visited,
+        params: HnswQueryParams,
```
BubbleCal: ref as well

BubbleCal (approved):

BubbleCal:
Thanks for the nice work! @LeoReeYang

PR #4312, opened 2025-07-24
chore: add CLAUDE.md files to subdirectories

wjones127:


java/CLAUDE.md
```diff
@@ -0,0 +1,9 @@
+## Structure
+Java code: `core/src`
+Rust JNI bindings: `core/lance-jni`
+
+## Commands
+format: `mvn spotless:apply && cargo fmt --all`
```
jackye1995: this should use `./mvnw` instead of `mvn` so it can run without the need to install maven separately

PR #4318, opened 2025-07-25
fix: implement limit and offset support for FTS and vector search

### LINKED ISSUE #4317: incorrect offset and limit behavior when using lancedb
There seems to be a bug in lance `offset` and `limit` behavior that is causing bugs in lancedb.  Here is a test case repro:

https://github.com/lancedb/lancedb/pull/2545/files#diff-255df6db5ecd2b3d7438614f8467e6ffc8ed3e9a25144fd9697e359db091a985R723

There is also an original bug report here:

https://github.com/lancedb/lancedb/issues/2459

Upstream PR for tests and tantivy fix, but lacks lance fix: https://github.com/lancedb/lancedb/pull/2545

wjones127:
Fixes limit and offset functionality for Full-Text Search (FTS) and vector search operations in Lance scanner.

Previously, `limit` and `offset` parameters were not working correctly because:
- FTS search applied its own limit early, truncating results before offset could be applied
- Vector search returned inconsistent order if any vector had equal distance to the query vector.

Users who paginate vector search still need to set `k > offset + limit`.

I've also added a nice unit test to validate offset and limit functionality for basic queries.

Fixes #4317

### DIFF
--- a/rust/lance/src/dataset/scanner.rs
+++ b/rust/lance/src/dataset/scanner.rs
@@ -870,6 +870,11 @@ impl Scanner {
         Ok(self)
     }
 
+    #[cfg(test)]
+    fn nearest_mut(&mut self) -> Option<&mut Query> {
+        self.nearest.as_mut()
+    }
+
     /// Set the distance thresholds for the nearest neighbor search.
     pub fn distance_range(
         &mut self,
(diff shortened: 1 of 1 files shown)


wjones127:
So, this is current failing for an interesting issue: `SortExec` doesn't do a stable sort. If any vectors happen to have identical distance to the query vector in a KNN query, then we get inconsistent results between queries.

wkalt (approved):

PR #4321, opened 2025-07-28
perf: improve L2 distance perf

BubbleCal:
this improves the L2 performance by 10%+ but slows the L2 performance with small dimensions 5% (8, 16).
8, 16 dimensions are so rare in vector search so it's worth.

this also avoids dynamic dispatching and any heap allocations when calculating batch L2 distances

this also fix some incorrect benchmark:
- `iter::count()` may be optimized if the length is already known, so replace it with `iter::fold`
- `collect()` may impact the benchmark result significantly, so also replace it with `iter::fold`


Xuanwo (approved):
Nice change

Issue #4319, opened 2025-07-25
feat: introduce RabitQ and set it as default BQ implementation


BubbleCal:
- [ ] implement [RabitQ](https://arxiv.org/abs/2405.12497)
- [ ] integrated it with lance vector index
- [ ] support RabitQ with multiple bits by padding
- [ ] benchmark RabitQ vs PQ
- [ ] improve multiple bits impl by [Extended-RabitQ](https://arxiv.org/abs/2409.09913)

PR #4322, opened 2025-07-28
feat: add matrix type

BubbleCal:
RabitQ needs matrix but only transpose and multiply
related to #4319

### DIFF
--- a/rust/lance-linalg/src/matrix.rs
+++ b/rust/lance-linalg/src/matrix.rs
@@ -124,7 +124,7 @@ impl<'a, T: ArrowPrimitiveType> Iterator for MatrixVectorIter<'a, T> {
     }
 
     fn size_hint(&self) -> (usize, Option<usize>) {
-        let remaining = (self.end - self.current) / self.step;
+        let remaining = (self.end - self.current).div_ceil(self.step);
         (remaining, Some(remaining))
     }
 }
--- a/rust/lance-linalg/src/matrix.rs
+++ b/rust/lance-linalg/src/matrix.rs
@@ -6,8 +6,12 @@ use std::ops::Mul;
 
 use arrow_array::{ArrowNativeTypeOp, ArrowPrimitiveType, PrimitiveArray};
 
+/// A matrix is a 2D array of elements.
+/// It's a thin wrapper around a vector of elements.
 pub struct Matrix<'a, T: ArrowPrimitiveType> {
     data: Cow<'a, [T::Native]>,
+    // (row, col) without considering transposed
+    // transposing won't change the internal shape, but method `shape()` will return the transposed shape
     shape: (usize, usize),
     transposed: bool,
 }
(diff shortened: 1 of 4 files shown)


rust/lance-linalg/src/matrix.rs
```diff
@@ -0,0 +1,174 @@
...
+// SPDX-FileCopyrightText: Copyright The Lance Authors
+
+use std::borrow::Cow;
+use std::ops::Mul;
+
+use arrow_array::{ArrowNativeTypeOp, ArrowPrimitiveType, PrimitiveArray};
+
+pub struct Matrix<'a, T: ArrowPrimitiveType> {
+    data: Cow<'a, [T::Native]>,
+    shape: (usize, usize),
```
Xuanwo, line 15: Can you add some comments about the shape? I'm a bit confused about the row and column.

rust/lance-linalg/src/matrix.rs
```diff
@@ -0,0 +1,174 @@
...
+        self.shape.0 * self.shape.1
+    }
+
+    pub fn is_empty(&self) -> bool {
+        self.len() == 0
+    }
+
+    pub fn row(&self, i: usize) -> impl Iterator<Item = &T::Native> {
+        if self.transposed {
+            self.data[i..self.len()].iter().step_by(self.shape.1)
```
Xuanwo: Do we need to assert `i < self.shape().0` here?

rust/lance-linalg/src/matrix.rs
```diff
@@ -0,0 +1,174 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright The Lance Authors
+
+use std::borrow::Cow;
+use std::ops::Mul;
+
+use arrow_array::{ArrowNativeTypeOp, ArrowPrimitiveType, PrimitiveArray};
+
+pub struct Matrix<'a, T: ArrowPrimitiveType> {
```
Xuanwo, line 11: This seems to be a public API, how about adding some docs for it?

rust/lance-linalg/src/matrix.rs
```diff
@@ -0,0 +1,174 @@
...
+        }
+    }
+
+    pub fn col(&self, j: usize) -> impl Iterator<Item = &T::Native> {
+        if self.transposed {
+            self.data[j * self.shape.1..(j + 1) * self.shape.1]
+                .iter()
+                .step_by(1)
+        } else {
+            self.data[j..self.len()].iter().step_by(self.shape.1)
```
Xuanwo: I suggest we implement our own `MatrixIter` instead of using `step_by`:

- Make code easier to read and maintain.
- `MatrixIter` can remove the use of `step_by(1)`
- `MatrixIter` can implement `ExactSizeIterator` directly for correct size.

I vibed one here for a quick start:

```rust
pub struct ColumnIter<'a, T> {
    data: &'a [T],
    current: usize,
    end: usize,
    step: usize,
}

impl<'a, T> Iterator for ColumnIter<'a, T> {
    type Item = &'a T;
    
    fn next(&mut self) -> Option<Self::Item> {
        if self.current < self.end {
            let item = &self.data[self.current];
            self.current += self.step;
            Some(item)
        } else {
            None
        }
    }
    
    fn size_hint(&self) -> (usize, Option<usize>) {
        let remaining = (self.end - self.current + self.step - 1) / self.step;
        (remaining, Some(remaining))
    }
}
```

rust/lance-linalg/src/matrix.rs
BubbleCal: I think we don't need that with the new iterator, it would return None if `i` is out of the bounds

rust/lance-linalg/src/matrix.rs
BubbleCal: it really does, added

PR #4325, opened 2025-07-28
chore: box pin large future block in test_index_impl

### LINKED ISSUE #4324: chore: test_index() easily cause async large future blocks
background: I have a PR failed in CI recently due to clippy:large_features, that fails in v2.rs tests , initially I think this was caused by my additional properties added. 
With the instruction by @Xuanwo 

I Installed a nightly rust toolchain and use 

` RUSTFLAGS="-Zprint-type-sizes"  cargo test --color=always --message-format=json-diagnostic-rendered-ansi --no-run --lib index::vector::ivf::v2::tests::test_create_ivf_hnsw_flat --profile test --manifest-path rust/lance/Cargo.toml`

to investigate the future block size

```
print-type-size type: `{async fn body of index::vector::ivf::v2::tests::test_index()}`: 11744 bytes, alignment: 16 bytes

print-type-size         local `.__awaitee`: 11120 bytes, alignment: 16 bytes, type: {async fn body of test_index_impl<UInt8Type>()}

print-type-size     local `.__awaitee`: 10224 bytes, alignment: 16 bytes, type: {async fn body of index::vector::ivf::v2::tests::generate_test_dataset<UInt8Type>()}

```

Looks like due to test_index_impl -> generate_test_dataset() was a large future block, and the upper tests may easily reach the threshold. This blocked my PR CI merge. So raise a quick box pin to bypass this.

chenghao-guo:
closes #4324


BubbleCal:
I think the main issue is that, the write path now is deep so that the future would be too large, unfortunately `test_index` is the last straw, but any other calls like `test_index` would also run into the same issue.

So I'd like to make `Dataset::write` to wrap its call into a Box, so that we can fix this issue for all possible use cases, not only `test_index`:
```
    pub async fn write(
        batches: impl RecordBatchReader + Send + 'static,
        dest: impl Into<WriteDestination<'_>>,
        params: Option<WriteParams>,
    ) -> Result<Self> {
        let mut builder = InsertBuilder::new(dest);
        if let Some(params) = &params {
            builder = builder.with_params(params);
        }
        Box::pin(builder
            .execute_stream(Box::new(batches) as Box<dyn RecordBatchReader + Send>))
            .await
    }
```

chenghao-guo:
> I think the main issue is that, the write path now is deep so that the future would be too large, unfortunately `test_index` is the last straw, but any other calls like `test_index` would also run into the same issue.
> 
> So I'd like to make `Dataset::write` to wrap its call into a Box, so that we can fix this issue for all possible use cases, not only `test_index`:
> 
> ```
>     pub async fn write(
>         batches: impl RecordBatchReader + Send + 'static,
>         dest: impl Into<WriteDestination<'_>>,
>         params: Option<WriteParams>,
>     ) -> Result<Self> {
>         let mut builder = InsertBuilder::new(dest);
>         if let Some(params) = &params {
>             builder = builder.with_params(params);
>         }
>         Box::pin(builder
>             .execute_stream(Box::new(batches) as Box<dyn RecordBatchReader + Send>))
>             .await
>     }
> ```

Thanks @BubbleCal ! I fully agree, that's better solution

chenghao-guo:
I've relocated the box-pin location as proposed by @BubbleCal is the more common practice and it also handles more test cases.

after changes, Verify the initial future size as below


```
type: `{async fn body of index::vector::ivf::v2::tests::test_index()}`: 4048 bytes, alignment: 16 bytes
-->
local `.__awaitee`: 3440 bytes, alignment: 16 bytes, type: {async fn body of test_index_impl<UInt8Type>()}
-->
local `.__awaitee`: 832 bytes, alignment: 8 bytes, type: {async fn body of index::vector::ivf::v2::tests::generate_test_dataset<UInt8Type>()}

```

wait for CI running and I'll check if any other issues.

BubbleCal (approved):

PR #4328, opened 2025-07-29
feat(java): expose sql api for java api

fangbo:
Add sql query for java api.

epic: https://github.com/lancedb/lance/issues/3950

See:

https://github.com/lancedb/lance/pull/4267
https://github.com/lancedb/lance/pull/4086

### DIFF
--- a/java/core/src/main/java/com/lancedb/lance/Dataset.java
+++ b/java/core/src/main/java/com/lancedb/lance/Dataset.java
@@ -21,7 +21,6 @@ import com.lancedb.lance.ipc.ScanOptions;
 import com.lancedb.lance.schema.ColumnAlteration;
 import com.lancedb.lance.schema.LanceSchema;
 import com.lancedb.lance.schema.SqlExpressions;
-import com.lancedb.lance.sql.SqlQuery;
 
 import org.apache.arrow.c.ArrowArrayStream;
 import org.apache.arrow.c.ArrowSchema;
@@ -898,7 +897,7 @@ public class Dataset implements Closeable {
   }
 
   public SqlQuery sql(String sql) {
-    return new SqlQuery(allocator, this, sql);
+    return new SqlQuery(this, sql);
   }
 
   private native void nativeCreateTag(String tag, long version);
--- a/java/core/src/main/java/com/lancedb/lance/Dataset.java
+++ b/java/core/src/main/java/com/lancedb/lance/Dataset.java
@@ -897,8 +897,8 @@ public class Dataset implements Closeable {
     }
   }
 
-  public SqlQuery sql(String sql, String tableName) {
-    return new SqlQuery(allocator, this, sql, tableName);
+  public SqlQuery sql(String sql) {
+    return new SqlQuery(allocator, this, sql);
   }
 
   private native void nativeCreateTag(String tag, long version);
--- a/java/core/src/main/java/com/lancedb/lance/Dataset.java
+++ b/java/core/src/main/java/com/lancedb/lance/Dataset.java
@@ -896,6 +896,15 @@ public class Dataset implements Closeable {
     }
   }
 
+  /**
+   * Execute SQL query on the dataset.
+   * The underlying SQL engine is DataFusion.
+   * Please refer to the DataFusion documentation for supported SQL syntax.
+   *
+   * @param sql SQL SELECT statement to execute. The query must reference the dataset
+   *            as the FROM clause or omit the FROM clause to query the current dataset.
+   * @return a SqlQuery with default table name is "dataset".
+   */
   public SqlQuery sql(String sql) {
     return new SqlQuery(this, sql);
   }
--- a/java/core/src/main/java/com/lancedb/lance/Dataset.java
+++ b/java/core/src/main/java/com/lancedb/lance/Dataset.java
@@ -900,8 +900,8 @@ public class Dataset implements Closeable {
    * Execute SQL query on the dataset. The underlying SQL engine is DataFusion. Please refer to the
    * DataFusion documentation for supported SQL syntax.
    *
-   * @param sql SELECT statement to execute. The default FROM table name is `dataset`, for
-   *     example: SELECT * FROM `dataset` LIMIT 10. If FROM table name is a custom value, the {@link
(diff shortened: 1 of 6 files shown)


fangbo:
@yanghua Please review this pr

java/core/src/main/java/com/lancedb/lance/Dataset.java
```diff
@@ -896,6 +897,10 @@ public long getVersion(String tag) {
     }
   }
 
+  public SqlQuery sql(String sql, String tableName) {
```
yanghua: In rust, we make the `tableName` an optional arg. If users do not specify, the default name is `dataset`.

java/core/src/test/java/com/lancedb/lance/SqlQueryTest.java
```diff
@@ -0,0 +1,97 @@
...
+        reader.getVectorSchemaRoot().getSchema().toString());
+    int rowCount = 0;
+    while (reader.loadNextBatch()) {
+      rowCount += reader.getVectorSchemaRoot().getRowCount();
+    }
+    Assertions.assertEquals(rowCount, 40);
+    reader.close();
+
+    // Test agg query
+    reader = dataset.sql("select sum(id) from " + NAME, NAME).intoBatchRecords();
```
yanghua: Can we also verify the result of the sum operation?

java/core/src/test/java/com/lancedb/lance/SqlQueryTest.java
```diff
@@ -0,0 +1,97 @@
...
+    if (dataset != null) {
+      dataset.close();
+    }
+    if (allocator != null) {
+      allocator.close();
+    }
+  }
+
+  @Test
+  public void testToRecordBatches() throws IOException {
```
yanghua, line 56: We can also verify `with_row_id` and `with_row_addr`.

yanghua:
@fangbo, thanks for your contribution! Left some comments.

java/core/src/main/java/com/lancedb/lance/Dataset.java
fangbo: > In rust, we make the `tableName` an optional arg. If users do not specify, the default name is `dataset`.

If the name of the Dataset is custom and the tableName is not specified, the query will report an error. Therefore, I think it is necessary to enforce the specification of the correct tableName in terms of the interface's semantics to avoid misunderstandings by users.

java/core/src/main/java/com/lancedb/lance/sql/SqlQuery.java
```diff
@@ -0,0 +1,79 @@
...
+
+public class SqlQuery {
+  private BufferAllocator allocator;
+  private Dataset dataset;
+  private String sql;
+  private String table;
+  private boolean withRowId = false;
+  private boolean withRowAddr = false;
+
+  public SqlQuery(BufferAllocator allocator, Dataset dataset, String sql, String table) {
```
majin1102: I think we may reuse allocator in the dataset? The issue here is who could be responsable for releasing this allocator

java/core/src/main/java/com/lancedb/lance/Dataset.java
majin1102: I think this would be challenged if rust or python doesn't have a table name parameter.

java/core/src/main/java/com/lancedb/lance/sql/SqlQuery.java
fangbo: The allocator is reused from Dataset's allocator.

majin1102: I mean this parameter of allocator may be unnecessary and expose some risk. We could directly get it from dataset.

fangbo: Does it mean that Dataset provides a new method like getAllocator() which returns the Allocator, and SqlQuery uses getAllocator() to get the shared Allocator ?

java/core/src/main/java/com/lancedb/lance/Dataset.java
fangbo: tableName has been changed to Optional in order to keep consistency with python and rust.

java/core/src/main/java/com/lancedb/lance/sql/SqlQuery.java
majin1102: Ye, that should be an option.

I think right now you could just use dataset.allocator if you put that code in the same package. There has been use cases like this.

fangbo: Good suggestion. The SqlQuery has been move to package com.lancedb.lance and it uses dataset.allocator.

java/core/src/main/java/com/lancedb/lance/SqlQuery.java
```diff
@@ -0,0 +1,81 @@
...
+import org.apache.arrow.c.Data;
+import org.apache.arrow.vector.ipc.ArrowReader;
+
+import java.io.IOException;
+import java.util.Optional;
+
+public class SqlQuery {
+  private Dataset dataset;
+  private String sql;
+  private String table;
```
jackye1995: default table name is `dataset`

java/core/src/main/java/com/lancedb/lance/Dataset.java
```diff
@@ -896,6 +896,10 @@ public long getVersion(String tag) {
     }
   }
 
+  public SqlQuery sql(String sql) {
```
yanghua, line 908: We could add some comments about this api is uses datafusion, and the default table name is `dataset`.

yanghua (approved):
LGTM, if we can add a comment for the public api. It would be better.

java/core/src/main/java/com/lancedb/lance/Dataset.java
yanghua: IMO, we can tell users the default name in the `@param sql`'s comment. And tell users they can configure some options to build a final sql query for `@return`'s comment.

PR #4335, opened 2025-07-29
feat: expose decoder config at python side

Xuanwo:
This PR is part of https://github.com/lancedb/lance/issues/3579

We will expose decoder config at python side so our python users can control them.


westonpace (commented):
Hmm, looking at this now, in python, I am wondering if these should be scanner options or not.

What do you think about making these part of the `ReadParams` in the dataset?

Partly to reduce complexity (scanner already has many options)

Partly because it makes it easier to just say "I always want to cache repetition index" and not have to re-specify it every scan.

It is slightly less flexible but if a user really wants to change params between reads they can use multiple dataset instances.

Xuanwo:
> What do you think about making these part of the `ReadParams` in the dataset?

I think that's a good idea, let's see how it look like then.

Xuanwo:
Please let me know if this what you want

python/python/lance/dataset.py
```diff
@@ -287,30 +293,38 @@ def __getstate__(self):
...
+        self._default_scan_options = default_scan_options
+        self._read_params = read_params
 
     def __copy__(self):
         ds = LanceDataset.__new__(LanceDataset)
         ds._uri = self._uri
         ds._storage_options = self._storage_options
         ds._ds = copy.copy(self._ds)
         ds._default_scan_options = self._default_scan_options
+        ds._read_params = getattr(self, "_read_params", None)
```
westonpace: Why `getattr` here instead of `ds._read_params = self._read_params.copy()`?

python/python/tests/test_dataset.py
```diff
@@ -3729,6 +3729,40 @@ def test_dataset_sql(tmp_path: Path):
...
+    assert result.num_rows == 100
+
+    # Test that scanner inherits options from dataset
+    dataset = lance.dataset(
+        tmp_path / "test",
+        read_params={"cache_repetition_index": True, "validate_on_decode": True},
+    )
+    scanner = dataset.scanner()
+    result = scanner.to_table()
+    assert result.num_rows == 100
```
westonpace: Is there any good way to validate the options are getting passed down?  For `cache_repetition_index` we could look at the iops generated by a scan (`lance.iops_counter` could be useful here)?

Xuanwo: Makes sense, added.

python/python/lance/dataset.py
Xuanwo: Fixed, not good at python 🤣

PR #4336, opened 2025-07-29
refactor: use previous instead v2 for our old encoding algo

Xuanwo:
We are currently working on the `v2.1` format, but many of our code paths are still using `v2`, which is actually the previous implementation. This PR renames the old legacy v2 code to `previous` to make it clearer.

I find this useful because it can prevent LLMs from referring to or searching within incorrect code paths.


HaochengLIU:
+1 on I ran into similar issues as well with LLM.
One thought: How do you feel calling it `v2_legacy`? `previous` is a very general word and i can foresee problems when we have 2.x, 3.x... down the road

westonpace:
>  One thought: How do you feel calling it v2_legacy? previous is a very general word and i can foresee problems when we have 2.x, 3.x... down the road

I actually am probably to blame here as I think Xuanwo was going to use legacy originally :smile: I was a little bit worried because, so far, legacy has specifically referred to the 0.1 version, which is very different, and has a lot more limitations.  We have various "legacy paths" in the scanner to handle it.

However, if it's two against one now then I'm happy going with legacy.  My feelings are not that strong :smile:

HaochengLIU:
Need an index mapping from version -> nickname soon! 😸

Xuanwo:
> i can foresee problems when we have 2.x, 3.x... down the road

`previous` looks good for now. And I will leave this task for people who will working on v3 to think 😆

PR #4339, opened 2025-07-29
chore: remove file level statistics documentation

HaochengLIU:
Remove dead file level statistics documentation.

### DIFF
--- a/docs/src/format/file.md
+++ b/docs/src/format/file.md
@@ -86,7 +86,7 @@ benchmarking upcoming features.
 The following values are supported:
 
 | Version        | Minimal Lance Version | Maximum Lance Version | Description                                                                                                                                  |
-|----------------|-----------------------|-----------------------|----------------------------------------------------------------------------------------------------------------------------------------------|
+| -------------- | --------------------- | --------------------- | -------------------------------------------------------------------------------------------------------------------------------------------- |
 | 0.1            | Any                   | Any                   | This is the initial Lance format.                                                                                                            |
 | 2.0            | 0.16.0                | Any                   | Rework of the Lance file format that removed row groups and introduced null support for lists, fixed size lists, and primitives              |
 | 2.1 (unstable) | None                  | Any                   | Enhances integer and string compression, adds support for nulls in struct fields, and improves random access performance with nested fields. |
@@ -113,7 +113,7 @@ are not yet supported (since there is no dictionary field encoding)
 ### Encodings Available
 
 | Encoding Name   | Encoding Type  | What it does                                                                                                                                | Supported Versions | When it is applied                                                                      |
-|-----------------|----------------|---------------------------------------------------------------------------------------------------------------------------------------------|--------------------|-----------------------------------------------------------------------------------------|
+| --------------- | -------------- | ------------------------------------------------------------------------------------------------------------------------------------------- | ------------------ | --------------------------------------------------------------------------------------- |
 | Basic struct    | Field encoding | Encodes non-nullable struct data                                                                                                            | >= 2.0             | Default encoding for structs                                                            |
 | List            | Field encoding | Encodes lists (nullable or non-nullable)                                                                                                    | >= 2.0             | Default encoding for lists                                                              |
 | Basic Primitive | Field encoding | Encodes primitive data types using separate validity array                                                                                  | >= 2.0             | Default encoding for primitive data types                                               |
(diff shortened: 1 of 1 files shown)


docs/src/format/file.md
```diff
@@ -155,8 +155,8 @@ There are special behaviors for different data types to account for nulls:
    an array containing just the value `"abcd"` could have a truncated min of `"abc"` and max of `"abd"`.
    If there is no truncated value greater than the maximum value, then instead use null for the maximum.
 
-!!! warning
-The `min` and `max` values are not guaranteed to be within the array; they are simply upper and lower bounds. Two common cases where they are not contained in the array is if the min or max original value was deleted and when binary data is truncated. Therefore, statistic should not be used to compute queries such as `SELECT max(col) FROM table`.
+!!! warning "Statistics min/max are not always present in the data"
```
HaochengLIU: [It](https://lancedb.github.io/lance/format/file/#statistics) was not rendered properly

new block
image

docs/src/format/file.md
```diff
@@ -121,7 +121,7 @@ are not yet supported (since there is no dictionary field encoding)
...
 | Dictionary      | Array encoding | Encodes data using a dictionary array and an indices array which is useful for large data types with few unique values                      | >= 2.0             | Used on string pages with fewer than 100 unique elements                                |
 | Packed struct   | Array encoding | Encodes a struct with fixed-width fields in a row-major format making random access more efficient                                          | >= 2.0             | Only used on struct types if the field metadata attribute `"packed"` is set to `"true"` |
-| Fsst            | Array encoding | Compresses binary data by identifying common substrings (of 8 bytes or less) and encoding them as symbols                                   | >= 2.1             | Used on string pages that are not dictionary encoded                                    |
+| Fsst            | Array encoding | Compresses binary data by identifying common substrings (of 8 bytes or less) and encoding them as symbols                                   | >= 2.1             | Used on string pages                                                                    |
```
HaochengLIU: I'm 99.99% sure that FSST could be applied to dictionary encoding
https://github.com/lancedb/lance/blob/main/rust/lance-encoding/src/encodings/physical/binary.rs#L556

The pipeline is dict -> FSST -> miniblock -> flush to disk

docs/src/format/file.md
HaochengLIU: Out of curiosity, do you know where is the file level statistics built in the Rust code?

HaochengLIU:
@jackye1995 Could you pls take a look? Taking a break after staring at zonemap for too long :(

docs/src/format/file.md
jackye1995: @Xuanwo I am not super familiar with this, can you confirm?

docs/src/format/file.md
jackye1995: This is at https://github.com/lancedb/lance/blob/v0.27.0/rust/lance-encoding-datafusion/src/zone.rs, but is now removed. 

Also from Weston, there's some old 0.1 statistics code at https://github.com/lancedb/lance/blob/main/rust%2Flance-file%2Fsrc%2Fwriter%2Fstatistics.rs

jackye1995 (approved):
thanks for the fix, pending confirmation of Fsst

docs/src/format/file.md
HaochengLIU: if the whole block is obsolete, should we remove it now?

docs/src/format/file.md
Xuanwo: If dictionary encoding is enabled, we will use fixed-size integers along with a dictionary, which means `fsst` will not be applied to them. I tend to not change here.

HaochengLIU: @Xuanwo Good catch  👍 

I rechecked the test I added and realize that it has some issues that never trigger the FSST codepath. The wording is reverted and I will remove the broken test in another PR sending to you

HaochengLIU: @Xuanwo On top of your mind, do you know where are these [`dictionary_data_block`](https://github.com/lancedb/lance/blob/main/rust/lance-encoding/src/encodings/logical/primitive.rs#L4192-L4200) flushed to disk?, would like to update my notes

docs/src/format/file.md
jackye1995: Oh good point, we should be good to remove it for now

HaochengLIU: 👍  Removed @jackye1995

PR #4340, opened 2025-07-29
feat: allow empty projection

westonpace:
Recent changes made it so we didn't report an error in `project` if a column did not exist.  This made it possible to get an error "no columns selected" when all columns asked for did not exist.  This was quite confusing.

This PR cleans up that behavior by reworking the projection slightly.

Previously, if a user requested zero columns, then they would receive an error "no columns selected and with_row_id is False, there is nothing to scan".  With this PR we now return empty record batches (with the correct number of rows).

### DIFF
--- a/python/python/tests/test_dataset.py
+++ b/python/python/tests/test_dataset.py
@@ -2737,9 +2737,11 @@ def test_scan_no_columns(tmp_path: Path):
     for batch in batches:
         assert batch.schema == expected_schema
 
-    # if with_row_id is not True then columns=[] is an error
-    with pytest.raises(ValueError, match="no columns were selected"):
-        dataset.scanner(columns=[]).to_table()
+    # Can specify nothing at all to get empty batches (with correct counts)
+    batches = list(dataset.scanner(columns=[], batch_size=10).to_batches())
+    assert len(batches) == 10
+    for batch in batches:
+        assert batch.num_rows == 10
 
     # also test with deleted data to make sure deleted ids not included
     dataset.delete("a = 5")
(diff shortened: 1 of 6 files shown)


python/python/tests/test_dataset.py
```diff
@@ -2737,9 +2737,11 @@ def test_scan_no_columns(tmp_path: Path):
     for batch in batches:
         assert batch.schema == expected_schema
 
-    # if with_row_id is not True then columns=[] is an error
-    with pytest.raises(ValueError, match="no columns were selected"):
-        dataset.scanner(columns=[]).to_table()
+    # Can specify nothing at all to get empty batches (with correct counts)
+    batches = list(dataset.scanner(columns=[], batch_size=10).to_batches())
```
wjones127, line 2741: Should we assert anything about the schema?

wjones127:
> BREAKING CHANGE: I could be convinced this isn't a breaking change but I think it technically is. Previously, if a user requested zero columns, then they would receive an error "no columns selected and with_row_id is False, there is nothing to scan". With this PR we now return empty record batches (with the correct number of rows).

Hmmm I feel like going from "Not supported error" to something that works generally isn't a breaking change. I'm fine if you want to label is as, though not sure what workflow it breaks.

westonpace:
> Hmmm I feel like going from "Not supported error" to something that works generally isn't a breaking change. I'm fine if you want to label is as, though not sure what workflow it breaks.

In that case I think I'll make it non-breaking.  I want to integrate this into lancedb and don't want to bother forcing a release.

python/python/tests/test_dataset.py
westonpace: Added a check to ensure it had no columns

Xuanwo:
Please allow me to remove the `breaking-change` tag from this PR.

westonpace:
> Please allow me to remove the breaking-change tag from this PR.

Ah, thank you!

PR #4343, opened 2025-07-30
feat: add FragmentSession

wojiaodoubao:
See https://github.com/lancedb/lance/issues/4090#issuecomment-3135227336 for more details

### DIFF
--- a/rust/lance/benches/take.rs
+++ b/rust/lance/benches/take.rs
@@ -26,14 +26,7 @@ use rand::Rng;
 use std::sync::Arc;
 #[cfg(target_os = "linux")]
 use std::time::Duration;
-use tokio::runtime::Runtime;
 
-type BenchEnvResult = (
-    Runtime,
-    u64,
-    u64,
-    Arc<dyn Fn(u64, u64, usize) -> Vec<Vec<u32>>>,
-);
 const BATCH_SIZE: u64 = 1024;
 
 fn gen_ranges(num_rows: u64, file_size: u64, n: usize) -> Vec<u64> {
@@ -48,7 +41,7 @@ fn gen_ranges(num_rows: u64, file_size: u64, n: usize) -> Vec<u64> {
 }
 
 fn bench_random_take_with_dataset(c: &mut Criterion) {
-    let rt = Runtime::new().unwrap();
+    let rt = tokio::runtime::Runtime::new().unwrap();
     let num_batches = 1024;
 
     for file_size in [1024 * 1024, 1024] {
@@ -73,7 +66,7 @@ fn bench_random_take_with_dataset(c: &mut Criterion) {
 
 fn dataset_take(
     c: &mut Criterion,
-    rt: &Runtime,
+    rt: &tokio::runtime::Runtime,
     file_size: usize,
     num_batches: usize,
     version: LanceFileVersion,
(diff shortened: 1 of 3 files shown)


rust/lance/src/dataset/fragment/session.rs
```diff
@@ -0,0 +1,204 @@
...
+};
+use arrow_array::RecordBatch;
+use lance_core::datatypes::Schema;
+use lance_core::Result;
+use std::borrow::Cow;
+use std::sync::Arc;
+
+/// A session of [`FileFragment`]
+///
+/// A [`FragmentSession`] maintains states for better performance in continuous take requests.
```
Xuanwo: This is the result, but not the purpose of `FragmentSession`. 

I suggest we update the comments to clarify that the following tone:

FragmentSession manages a short-lived session, allowing us to maintain internal states instead of creating a new one each time. This API works well for users making repeated requests over the same projected schema.

(Just for an example)

rust/lance/src/dataset/fragment.rs
```diff
@@ -1257,6 +1259,17 @@ impl FileFragment {
         Ok(())
     }
 
+    /// Open a session for take.
+    ///
+    /// A [`FragmentSession`] maintains states for better performance in continuous take requests.
```
Xuanwo: The same

rust/lance/src/dataset/fragment.rs
```diff
@@ -2347,7 +2365,7 @@ impl FragmentReader {
 }
 
 #[cfg(test)]
-mod tests {
+pub mod tests {
```
Xuanwo: Why was this change made? It's quite strange to make tests mod `pub`.

rust/lance/src/dataset/fragment/session.rs
```diff
@@ -0,0 +1,204 @@
...
+        with_row_address: bool,
+    ) -> Result<Self> {
+        let reader = fragment
+            .open(
+                projection,
+                FragReadConfig::default().with_row_address(with_row_address),
+            )
+            .await?;
+
+        let deletion_vector = fragment.get_deletion_vector().await?;
```
Xuanwo: Let's try

```rust
let sorted_deleted_ids = fragment
    .get_deletion_vector()
    .await?
    .map(|dv| {
        let mut ids: Vec<_> = dv.as_ref().iter().cloned().collect();
        ids.sort();
        ids
    });
```

rust/lance/benches/take.rs
```diff
@@ -26,7 +26,14 @@ use rand::Rng;
 use std::sync::Arc;
 #[cfg(target_os = "linux")]
 use std::time::Duration;
+use tokio::runtime::Runtime;
 
+type BenchEnvResult = (
```
Xuanwo: I suggest we don't update bench code in this PR.

Xuanwo (commented):
Thank you for working on this

rust/lance/src/dataset/fragment.rs
wojiaodoubao: Thanks your reminding. I used to use ```crate::dataset::fragment::tests::create_dataset``` in unit tests of fragment/session.rs. 

After a search, I find in lance we don't import tests from other mod. Let me fix this.

Xuanwo (approved):
Thank you @wojiaodoubao for working on this!

PR #4344, opened 2025-07-30
feat: support RabitQ quantization

BubbleCal:
related to https://github.com/lancedb/lance/issues/4319

this also:
- introduces `ndarray` for matrix
- upgrade `rand` for generating random orthogonal matrix

### DIFF
--- a/rust/lance-index/src/vector.rs
+++ b/rust/lance-index/src/vector.rs
@@ -175,7 +175,8 @@ pub trait VectorIndex: Send + Sync + std::fmt::Debug + Index {
     /// that are most likely to contain the nearest neighbors (e.g. the closest
     /// partitions to the query vector).
     ///
-    /// The results should be in sorted order from closest to farthest.
+    /// Return the partition ids and the distances between the query and the centroids,
+    /// the results should be in sorted order from closest to farthest.
     fn find_partitions(&self, query: &Query) -> Result<(UInt32Array, Float32Array)>;
 
     /// Get the total number of partitions in the index.


rust/lance-index/src/vector/bq/builder.rs
```diff
@@ -0,0 +1,370 @@
...
+    }
+}
+
+fn random_normal_matrix(n: usize) -> ndarray::Array2<f32> {
+    let mut rng = rand::rng();
+    let normal = rand_distr::Normal::new(0.0, 1.0).unwrap();
+    ndarray::Array2::from_shape_simple_fn((n, n), || normal.sample(&mut rng))
+}
+
+fn householder_qr(a: ndarray::Array2<f32>) -> (ndarray::Array2<f32>, ndarray::Array2<f32>) {
```
wjones127: I'm a little surprised there's no direct tests for some of this linear algebra functions. Maybe we should add a few?

rust/lance-index/src/vector/bq/storage.rs
```diff
@@ -0,0 +1,686 @@
...
+    T::Native: AsPrimitive<f32>,
+{
+    let mut dist_table = vec![0.0; qc.len() * 4];
+    qc.chunks_exact(SEGMENT_LENGTH)
+        .zip(dist_table.chunks_exact_mut(SEGMENT_NUM_CODES))
+        .for_each(|(sub_vec, dist_table)| {
+            (1..16).for_each(|j| {
+                dist_table[j] = dist_table[j - lowbit(j)] + sub_vec[LOWBIT_IDX[j]].as_();
+            })
+        });
```
wjones127: This code could use a little more comments:

* Should `1..16` be `1..SEGMENT_NUM_CODES`?
* Why do we skip 0 in the `for_each` loop?
* What does `j - lowbit(j)` do?

Maybe give a concrete example in the loop so we understand bit better.

rust/lance-index/src/vector/bq/storage.rs
```diff
@@ -0,0 +1,686 @@
...
+fn lowbit(x: usize) -> usize {
+    1 << x.trailing_zeros()
+}
+
+#[inline]
+pub fn build_dist_table_direct<T: ArrowFloatType>(qc: &[T::Native]) -> Vec<f32>
+where
+    T::Native: AsPrimitive<f32>,
+{
+    let mut dist_table = vec![0.0; qc.len() * 4];
```
wjones127, line 217: Is the 4 here supposed to be `SEGMENT_LENGTH` or something else?

rust/lance-index/src/vector/bq/storage.rs
```diff
@@ -0,0 +1,686 @@
...
+    0, /*0111*/
+    3, /*1000*/
+    0, /*1001*/
+    1, /*1010*/
+    0, /*1011*/
+    2, /*1100*/
+    0, /*1101*/
+    1, /*1110*/
+    0, /*1111*/
+]; // all possible combination for a 4 bit string
```
wjones127: Why not just use [trailing_zeros](https://doc.rust-lang.org/std/primitive.u32.html#method.trailing_zeros)? Don't some platforms have an instruction for this?

Or, at least use it to pre-compute this array?

```suggestion
const LOWBIT_IDX: [usize; 16] = {
    let mut array = [0; 16];
    let mut i = 0;
    while i < 16 {
        array[i] = i.trailing_zeros();
        i += 1;
    }
    array
};
```

rust/lance-index/benches/rq.rs
```diff
@@ -0,0 +1,120 @@
...
+        //     )
+        //     .as_str(),
+        //     |b| {
+        //         b.iter(|| {
+        //             for i in 0..TOTAL {
+        //                 black_box(dist_calc.distance(i as u32));
+        //             }
+        //         })
+        //     },
+        // );
```
westonpace: ?

rust/lance-index/src/vector/bq/builder.rs
```diff
@@ -0,0 +1,370 @@
...
+fn random_orthogonal<T: ArrowFloatType>(n: usize) -> ndarray::Array2<T::Native>
+where
+    T::Native: FromPrimitive,
+{
+    let a = random_normal_matrix(n);
+    let (q, _) = householder_qr(a);
+
+    // cast f32 matrix to T::Native matrix
+    q.mapv(|v| T::Native::from_f32(v).unwrap())
+}
```
westonpace, line 413: Should we have unit tests here of the quantizer?

rust/lance-index/src/vector/bq/storage.rs
```diff
@@ -0,0 +1,686 @@
...
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct RabitQuantizationMetadata {
+    #[serde(skip)]
+    pub rotate_mat: Option<FixedSizeListArray>,
+    pub rotate_mat_position: u32,
+    pub num_bits: u8,
+    // number of vectors in the vector store,
+    // we need to store this because we need to add padding to the vector codes
+    // pub n: usize,
```
westonpace: ?

rust/lance-index/src/vector/bq/storage.rs
```diff
@@ -0,0 +1,686 @@
...
+    // we split the query codes into d/4 chunks, each chunk is with 4 elements,
+    // then dist_table[i][j] is the distance between the i-th query code and the code j
+    dist_table: Vec<f32>,
+    add_factors: &'a [f32],
+    scale_factors: &'a [f32],
+    query_factor: f32,
+
+    // sq_min: f32,
+    // sq_scale: f32,
+    // sq_sum: f32,
```
westonpace: ?

rust/lance-index/src/vector/bq/storage.rs
```diff
@@ -0,0 +1,686 @@
...
+                dist_table[current_code] + next_dist_table[next_code]
+            })
+            .sum::<f32>();
+
+        // distance between quantized vector and query vector
+        let dist_vq_qr = (2.0 * dist - self.sum_q) / self.sqrt_d;
+        dist_vq_qr * self.scale_factors[id] + self.add_factors[id] + self.query_factor
+    }
+
+    #[inline(never)]
```
westonpace: Were these hints tuned with benchmarks?  Or are they just guesses?

rust/lance-index/src/vector/ivf.rs
```diff
@@ -242,6 +251,60 @@ impl IvfTransformer {
...
+        transforms.push(Arc::new(ResidualTransform::new(
+            centroids.clone(),
+            PART_ID_COLUMN,
+            vector_column,
+        )));
+
+        // transforms.push(Arc::new(ExtendDimensionTransformer::new(
+        //     vector_column,
+        //     rq.code_dim(),
+        // )));
```
westonpace: ?

rust/lance-linalg/src/simd/dist_table.rs
```diff
@@ -0,0 +1,156 @@
...
+
+extern "C" {
+    #[cfg(all(kernel_support = "avx512", target_arch = "x86_64"))]
+    pub fn sum_4bit_dist_table_32bytes_batch_avx512(
+        codes: *const u8,
+        code_length: usize,
+        dist_table: *const u8,
+        dists: *mut u16,
+    );
+}
```
westonpace, line 172: Can we add unit tests here to at least illuminate what these functions are supposed to do?

rust/lance/src/dataset/scanner.rs
```diff
@@ -3429,16 +3431,21 @@ impl Scanner {
 
     #[instrument(level = "info", skip(self))]
     pub async fn analyze_plan(&self) -> Result<String> {
+        let start = Instant::now();
         let plan = self.create_plan().await?;
+        log::info!("create plan took {:?}", start.elapsed());
```
westonpace: Should these log statements be at debug level?

rust/lance/src/io/exec/knn.rs
```diff
@@ -841,6 +862,7 @@ impl ANNIvfSubIndexExec {
                             DataFusionError::Execution(format!("Failed to calculate KNN: {}", e))
                         })
                         .await?;
+                    log::info!("partition {} took {:?}", part_id, start.elapsed());
```
westonpace: Should this be debug?

rust/lance-index/src/vector/bq/storage.rs
BubbleCal: added more comments to explain these

rust/lance-index/src/vector/bq/storage.rs
BubbleCal: didn't use it because don't want to do type cast every time, changed to use it for pre-computing

rust/lance-index/src/vector/bq/builder.rs
BubbleCal: added

rust/lance-index/src/vector/bq/storage.rs
BubbleCal: yes, added more comments to explain this

rust/lance-index/benches/rq.rs
BubbleCal: fixed

rust/lance-index/src/vector/bq/storage.rs
BubbleCal: removed

rust/lance-index/src/vector/bq/storage.rs
BubbleCal: removed

rust/lance-index/src/vector/bq/storage.rs
BubbleCal: yes it's for benchmarking, changed to `always`

rust/lance-index/src/vector/ivf.rs
BubbleCal: removed

rust/lance/src/dataset/scanner.rs
BubbleCal: removed it, just for debugging

BubbleCal:
@westonpace @wjones127 thanks for the feedbacks, added more tests and they did capture some bugs!

rust/lance-linalg/src/simd/dist_table.rs
```diff
@@ -0,0 +1,212 @@
...
+
+#[cfg(target_arch = "x86_64")]
+use std::arch::x86_64::*;
+
+#[allow(unused_imports)]
+use lance_core::utils::cpu::{SimdSupport, SIMD_SUPPORT};
+
+pub const PERM0: [usize; 16] = [0, 8, 1, 9, 2, 10, 3, 11, 4, 12, 5, 13, 6, 14, 7, 15];
+pub const PERM0_INVERSE: [usize; 16] = [0, 2, 4, 6, 1, 3, 5, 7, 8, 10, 12, 14, 9, 11, 13, 15];
+pub const BATCH_SIZE: usize = 32;
```
jbapple, line 12: Does this need to be `pub`?

rust/lance-linalg/src/simd/dist_table.rs
```diff
@@ -0,0 +1,212 @@
...
+
+// This function is used to sum the distance table for 4-bit codes.
```
jbapple, line 14: It would help to say where the reader can find the definition of "the distance table".

rust/lance-linalg/src/simd/dist_table.rs
```diff
@@ -0,0 +1,212 @@
...
+    accu2 = _mm256_sub_epi16(accu2, _mm256_slli_epi16(accu3, 8));
+    let dis1 = _mm256_add_epi16(
+        _mm256_permute2f128_si256(accu2, accu3, 0x21),
+        _mm256_blend_epi32(accu2, accu3, 0xF0),
+    );
+
+    _mm256_storeu_si256(dists.as_mut_ptr().add(16) as *mut __m256i, dis1);
+}
+
+extern "C" {
```
jbapple, line 164: Can you add a comment explaining why this is in C but the AVX2 version is in Rust?

rust/lance-index/src/vector/bq/builder.rs
```diff
@@ -0,0 +1,416 @@
...
+    }
+}
+
+fn random_normal_matrix(n: usize) -> ndarray::Array2<f64> {
+    let mut rng = rand::rng();
+    let normal = rand_distr::Normal::new(0.0, 1.0).unwrap();
+    ndarray::Array2::from_shape_simple_fn((n, n), || normal.sample(&mut rng))
+}
+
+fn householder_qr(a: ndarray::Array2<f64>) -> (ndarray::Array2<f64>, ndarray::Array2<f64>) {
```
jbapple, line 316: Can you say more about this in relation to the standard notation for Householder transformations, which take more than one parameter?

rust/lance-index/src/vector/bq/builder.rs
```diff
@@ -0,0 +1,416 @@
...
+    pub fn new<T: ArrowFloatType>(num_bits: u8, dim: i32) -> Self {
+        // we don't need to calculate the inverse of P,
+        // just take the generated matrix as P^{-1}
+        let code_dim = dim * num_bits as i32;
+        let rotate_mat = random_orthogonal::<T>(code_dim as usize);
+        let (rotate_mat, _) = rotate_mat.into_raw_vec_and_offset();
+
+        let rotate_mat = match T::FLOAT_TYPE {
+            FloatType::Float16 => {
+                let rotate_mat = T::ArrayType::from(rotate_mat);
```
jbapple: Nit: if these cases are identical, can they be combined?

rust/lance-index/src/vector/bq/builder.rs
```diff
@@ -0,0 +1,416 @@
...
+        let rotated_vectors = rotate_mat.dot(&vectors);
+
+        let quantized_vectors = rotated_vectors.t().mapv(|v| v.as_().is_sign_positive());
+        let bv: BitVec<u8, Lsb0> = BitVec::from_iter(quantized_vectors);
+
+        let codes = UInt8Array::from(bv.into_vec());
+        debug_assert_eq!(codes.len(), n * self.code_dim() / 8);
+        Ok(Arc::new(FixedSizeListArray::try_new_from_values(
+            codes,
+            self.code_dim() as i32 / 8, // num_bits -> num_bytes
```
jbapple: Nit: prefer `u8::BITS`?

jbapple: (Here and elsewhere)

rust/lance-index/src/vector/bq/builder.rs
```diff
@@ -0,0 +1,416 @@
...
+        let h_r = h.dot(&r_block);
+        r.slice_mut(s![k.., k..]).assign(&h_r);
+
+        // Apply transformation to Q
+        let q_block = q.slice(s![.., k..]).to_owned();
+        let q_h = q_block.dot(&h);
+        q.slice_mut(s![.., k..]).assign(&q_h);
+    }
+
+    (q, r)
```
jbapple, line 355: It looks like `r` isn't used in the one current caller

rust/lance-index/src/vector/bq/storage.rs
```diff
@@ -0,0 +1,691 @@
...
+use crate::vector::storage::{DistCalculator, VectorStore};
+
+pub const RABIT_METADATA_KEY: &str = "lance:rabit";
+pub const RABIT_CODE_COLUMN: &str = "_rabit_codes";
+pub const SEGMENT_LENGTH: usize = 4;
+pub const SEGMENT_NUM_CODES: usize = 1 << SEGMENT_LENGTH;
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct RabitQuantizationMetadata {
+    #[serde(skip)]
```
jbapple, line 46: Can you leave a comment on why the skip?

rust/lance-index/src/vector/bq/storage.rs
```diff
@@ -0,0 +1,691 @@
...
+        .iter()
+        .cloned()
+        .minmax_by(|a, b| a.total_cmp(b))
+        .into_option()
+        .unwrap();
+    // this happens if the query is all zeros
+    if qmin == qmax {
+        return (qmin, qmax, vec![0; dist_table.len()]);
+    }
+    let factor = 255.0 / (qmax - qmin);
```
jbapple, line 260: Can you add a comment about `factor` - where does this come from? Is this just normalizing to 0.0-255.0?

jbapple (changes requested):

rust/lance-linalg/src/simd/dist_table.rs
BubbleCal: Yes, need to use these from the other crate (lance-index)

rust/lance-linalg/src/simd/dist_table.rs
BubbleCal: added description here

rust/lance-index/src/vector/bq/builder.rs
BubbleCal: did you mean that `alpha` param? that param is calculated at runtime (line 326 `sign * x_norm`), just didn't explicitly show it in the code.

rust/lance-index/src/vector/bq/builder.rs
BubbleCal: yes it's not used anywhere for now, just reserve it for keeping the function's semantic

rust/lance-index/src/vector/bq/storage.rs
BubbleCal: Yes, right, added more comments for this method

rust/lance-linalg/src/simd/dist_table.rs
BubbleCal: added more comments here,
it's because we are on rust 1.83, but AVX512 is supported since 1.89

rust/lance-linalg/src/simd/dist_table.rs
BubbleCal: added

rust/lance-index/src/vector/bq/builder.rs
jbapple: I mean it would be helpful to relate this code to https://en.wikipedia.org/wiki/Householder_transformation#Definition, which takes two vectors and returns a vector, especially because "Householder" is not referenced in the RabitQ paper.

rust/lance-index/src/vector/bq/builder.rs
```diff
@@ -0,0 +1,412 @@
...
+            metadata: metadata.clone(),
+        }))
+    }
+
+    fn field(&self) -> Field {
+        Field::new(
+            RABIT_CODE_COLUMN,
+            DataType::FixedSizeList(
+                Arc::new(Field::new("item", DataType::UInt8, true)),
+                self.code_dim() as i32 / 8, // num_bits -> num_bytes
```
jbapple: Nit: u8::BITS

rust/lance-index/benches/pq_dist_table.rs
```diff
@@ -18,8 +18,8 @@ use rand::{prelude::StdRng, Rng, SeedableRng};
 #[cfg(target_os = "linux")]
 use pprof::criterion::{Output, PProfProfiler};
 
-const PQ: usize = 96;
-const DIM: usize = 1536;
+const DIM: usize = 128;
+const PQ: usize = DIM / 8;
```
jbapple, line 22: nit: u8::BITS

rust/lance-index/benches/rq.rs
```diff
@@ -0,0 +1,120 @@
...
+
+fn mock_rq_storage(num_bits: u8) -> RabitQuantizationStorage {
+    // generate random rq codes
+    let rq = RabitQuantizer::new::<Float32Type>(num_bits, DIM as i32);
+    let builder = BatchGeneratorBuilder::new()
+        .col(ROW_ID, lance_datagen::array::step::<UInt64Type>())
+        .col(
+            RABIT_CODE_COLUMN,
+            rand_type(&fixed_size_list_type(
+                (DIM * num_bits as usize / 8) as i32,
```
jbapple: nit: u8::BITS

rust/lance-index/src/vector/bq/storage.rs
jbapple: Is this right? My intuition is that this should be `(256.0/(qmax-qmin)).floor()`. That's a small difference, obviously. For example, for `d, qmax-qmin = 3,4`, we get 191 vs 192.

rust/lance-core/src/utils/cpu.rs
```diff
@@ -4,18 +4,20 @@
...
     Neon,
     Sse,
     Avx2,
     Avx512,
+    Avx512FP16,
     Lsx,
     Lasx,
 }
 
 /// Support for FP16 SIMD operations
```
jbapple: Is this comment out of date now?

rust/lance-index/Cargo.toml
```diff
@@ -95,6 +98,12 @@ pprof.workspace = true
 # docs.rs uses an older version of Ubuntu that does not have the necessary protoc version
 features = ["protoc"]
 
+[profile.release]
```
jbapple: Elsewhere, `debug = true` is only on `release-with-debug` and `bench`, not `release`.

rust/lance-index/src/vector/bq/builder.rs
```diff
@@ -0,0 +1,412 @@
...
+
+    fn transform<T: ArrowFloatType>(
+        &self,
+        residual_vectors: &FixedSizeListArray,
+    ) -> Result<ArrayRef>
+    where
+        T::Native: AsPrimitive<f32>,
+    {
+        // we don't need to normalize the residual vectors,
+        // because the signal of P^{-1} x v_r is the same as P^{-1} x v_r / ||v_r||
```
jbapple: How do we know the signal is the same?

jbapple (changes requested):

rust/lance-index/src/vector/bq/storage.rs
jbapple: (or maybe `255.99/(qmax-qmin)` so we never get `256`)

rust/lance-index/src/vector/bq/builder.rs
BubbleCal: I was referring this: https://en.wikipedia.org/wiki/Householder_transformation#QR_decomposition

rust/lance-index/src/vector/bq/builder.rs
BubbleCal: say the former is $x$, then the latter is $\frac{x}{||v_r||}$, so the signals are all the same

rust/lance-index/benches/pq_dist_table.rs
BubbleCal: here it's `DIM/8`, which means each subvector is with 8 dims

rust/lance-index/src/vector/bq/builder.rs
jbapple: Could you elaborate?

rust/lance-index/src/vector/bq/storage.rs
jbapple: Was this marked as resolved (without a response or a change) by mistake?

jbapple (changes requested):

rust/lance-index/src/vector/bq/storage.rs
BubbleCal: Oh... it's a mistake

`256` won't work because that would give us 256 which can't be represent by `u8`.
`255.99` is probably just the same as `255.0`.

jbapple:
```
>>> sum([abs(round((255.0/m) * i) - math.floor((255.99/m) * i)) for m in range(1,256) for i in range(0,m+1)])
7803
```

rust/lance-index/src/vector/bq/builder.rs
BubbleCal: no matter what $x, \lVert v \rVert$ is, e.g. $x = (-1, 2)$ and $\lVert v \rVert = 2$, then $x/\lVert v \rVert = (-0.5, 1)$, the signal of each dim just keeps the same because $\lVert v \rVert > 0$ (I'm assuming no zero vector)

jbapple: Ah, I see! Nit: can you change the comment to say "sign" instead of "signal"? I think that is the more common term, and "signal" doesn't appear in the paper.

rust/lance-index/src/vector/bq/storage.rs
BubbleCal: selected `255.0 / (qmax - qmin)` because the min will be normalized to 0, and the max to 255.
`round` would perform better because it reduces the loss (say 1.8 will be normalized to 2 (loss=0.2) rather than 1 (loss=0.8))

rust/lance-index/src/vector/bq/builder.rs
BubbleCal: kk done!

rust/lance-index/src/vector/bq/storage.rs
```diff
@@ -0,0 +1,691 @@
...
+    // every 4 bits (SEGMENT_LENGTH) is a segment, and we need to compute the distance between the segment and all the codes
+    // so there are dim/4 segments, and the number of codes is 16 (2^{SEGMENT_LENGTH}),
+    // so we have dim/4 * 16 = dim * 4 elements in the dist_table
+    let mut dist_table = vec![0.0; qc.len() * 4];
+    qc.chunks_exact(SEGMENT_LENGTH)
+        .zip(dist_table.chunks_exact_mut(SEGMENT_NUM_CODES))
+        .for_each(|(sub_vec, dist_table)| {
+            // skip 0 because it's always 0
+            (1..SEGMENT_NUM_CODES).for_each(|j| {
+                // this is a little bit tricky,
```
jbapple: Probably anything tricky should have a test.

rust/lance-index/src/vector/bq/storage.rs
BubbleCal: just the same as the implementation of rabitq repo: 
https://github.com/VectorDB-NTU/RaBitQ-Library/blob/de7846836136e214e88a6898d7215927de4b47f0/rabitqlib/index/lut.hpp#L38

https://github.com/VectorDB-NTU/RaBitQ-Library/blob/de7846836136e214e88a6898d7215927de4b47f0/rabitqlib/utils/space.hpp#L19

jbapple: Here's another way to think about it: we want the distribution of values in the domain to be the same as the range, right? So a perfectly evenly distributed domain should be perfectly evenly distributed in the range. But round with 255.0 doesn't accomplish that, while floor with 255.99 gets as close as possible.

```
>>> ys = [len([i for i in range(0,10000) if round((255.0/9999) * i) == j]) for j in range(0,256)]
>>> max(ys)
40
>>> min(ys)
20
```

and

```
>>> zs = [len([i for i in range(0,10000) if math.floor((255.99/9999) * i) == j]) for j in range(0,256)]
>>> min(zs)
39
>>> max(zs)
40
```

rust/lance-linalg/build.rs
```diff
@@ -61,6 +55,14 @@ fn main() -> Result<(), String> {
             // generated the AVX512 version of the f16 kernels.
             println!("cargo:rustc-cfg=kernel_support=\"avx512\"");
         };
+        if let Err(err) = build_dist_table_with_flags("avx512", &["-march=native"]) {
+            println!(
+                "cargo:warning=Skipping build of AVX-512 dist_table.  Clang/GCC too old or compiler does not support sapphirerapids architecture.  Error: {}",
```
jbapple: AVX512 is not sapphirerapids and up - AMD chips support it, and so do some older Intel chips like Skylake server. But, more importantly, compilers could fail for other reasons. Maybe just say that the compile failed without guessing the reason, and the user can figure out why based on `err`?

rust/lance-index/src/vector/bq/storage.rs
BubbleCal: I still don't get this...
anyway I just tested it out:
```
Method 1: round((x-min)*255/(max-min))
  Quantized Range: [0, 255]
  MSE: 0.000513
  MAE: 0.019597
  Max Error: 0.039213

Method 2: floor((x-min)*255.99/(max-min))
  Quantized Range: [0, 255]
  MSE: 0.002030
  MAE: 0.039004
  Max Error: 0.078118

Comparison:
  MSE: Method 2 is -295.84% worse than Method 1
  MAE: Method 2 is -99.03% worse than Method 1
```

rust/lance-index/src/vector/bq/storage.rs
BubbleCal: agree, added

rust/lance-linalg/build.rs
BubbleCal: fixed

jbapple: I'm suggesting that you remove ANY reference to compiler age and AVX-512 support.

rust/lance-index/src/vector/bq/storage.rs
jbapple: What program were you using to do this MSE and MAE test?

Can you say more concretely what you don't get about the distribution point I was making? Is there some specific word or concept you didn't follow?

BubbleCal: > while floor with 255.99 gets as close as possible.

just didn't get why this would get closest result?

rust/lance-linalg/build.rs
BubbleCal: kk, removed the guesses, still keep `Skipping build of AVX-512 dist_table` so that we can know what's failed to build

rust/lance-index/src/vector/bq/storage.rs
jbapple: It gets the closest result (for some definition of "closest") by mapping the range to the domain in as even a way as possible. Consider, for instance, if the range were 0 to 2, instead of 0 to 255. Presumably we would want a third of the values in the domain to map to 0, a third to 1, and a third to 2. If the domain is 0-99, inclusive, then round((x-min)*2.0/99) maps (since min = 0) fully *half* of the values in the domain to 1, rather than a third, and only a quarter of the values map to 0 or 2. Now the 0-2 range is just an example, but the same thing happens as the range scales up. Roughly speaking, rounding means that the sizes of the preimages of the endpoints of the range are half the sizes of the preimages of the values that are not endpoints.

Please also note that 255.99 is an example. We really need something like the following to handle large enough domains:

```
let result = (d - qmin) * (256.0 / (qmax - qmin));
return (result as i32).min(255) as u8;
```

BubbleCal: Oh I get the point now.
sorry for confusing, the target of scalar quantization isn't to perfectly evenly distribute the values, because most data follows normal distribution.

And yes, rust's `as u8` would map all values greater than 255 to 255

rust/lance-index/src/vector.rs
```diff
@@ -166,7 +176,7 @@ pub trait VectorIndex: Send + Sync + std::fmt::Debug + Index {
     /// partitions to the query vector).
     ///
     /// The results should be in sorted order from closest to farthest.
-    fn find_partitions(&self, query: &Query) -> Result<UInt32Array>;
+    fn find_partitions(&self, query: &Query) -> Result<(UInt32Array, Float32Array)>;
```
Xuanwo, line 180: Can you update the comments to explain what `Float32Array` is?

rust/lance-linalg/src/simd/dist_table.c
Xuanwo, line 1: We need this because Rust currently doesn't support AVX512?

BubbleCal: Right, will implement this in rust once we upgrade to 1.89.0

Xuanwo: Got it.

rust/lance-index/src/vector.rs
BubbleCal: added

jbapple (approved):

PR #4345, opened 2025-07-30
chore: return the distances when computing partitions

BubbleCal:
this is useful when we need to know the distance from each vector to the centroid it belongs to (the case of RabitQ)

### DIFF
--- a/rust/lance-index/src/vector/kmeans.rs
+++ b/rust/lance-index/src/vector/kmeans.rs
@@ -204,7 +204,25 @@ pub trait KMeansAlgo<T: Num> {
         dimension: usize,
         distance_type: DistanceType,
         index: Option<&SimpleIndex>,
-    ) -> (Vec<Option<u32>>, f64);
+    ) -> (Vec<Option<u32>>, f64) {
+        let (membership, dists) =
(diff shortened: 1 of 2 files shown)


rust/lance-index/src/vector/kmeans.rs
```diff
@@ -265,16 +283,10 @@ where
...
-                .map(|cd| cd.map(|(c, _)| c))
-                .collect::<Vec<_>>(),
-            cluster_and_dists
-                .iter()
-                .map(|cd| cd.map(|(_, d)| d as f64).unwrap_or_default())
-                .sum(),
-        )
+        cluster_and_dists
+            .iter()
+            .map(|cd| (cd.map(|(c, _)| c), cd.map(|(_, d)| d)))
```
Xuanwo: Hi, please give `.map(Option::unzip)` a try which can make `Option<(C,D)> → (Option<C>, Option<D>)`.

Xuanwo (commented):
Mostly LGTM, only a small nit.

rust/lance-index/src/vector/kmeans.rs
BubbleCal: It's what I'm looking for!

PR #4347, opened 2025-07-30
feat: add sdk_version in tracing events

ddupg:
closes: https://github.com/lancedb/lance/discussions/4234#discussioncomment-13860114

### DIFF
--- a/python/src/tracing.rs
+++ b/python/src/tracing.rs
@@ -15,6 +15,7 @@
 // specific language governing permissions and limitations
 // under the License.
 
+use crate::CLIENT_VERSION;
 use chrono::{SecondsFormat, Utc};
 use datafusion_common::HashMap;
 use pyo3::pyclass;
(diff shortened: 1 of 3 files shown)


yanghua (approved):
LGTM

python/src/tracing.rs
```diff
@@ -256,6 +258,8 @@ impl tracing_subscriber::Layer<Registry> for LoggingPassthroughRef {
 
             let now = Utc::now().to_rfc3339_opts(SecondsFormat::Secs, true);
             args.args.insert("timestamp".to_string(), now.to_string());
+            args.args
+                .insert("sdk_version".to_string(), SDK_VERSION.to_string());
```
Xuanwo: `SDK_VERSION` is a bit confusing to me; how about using `CLIENT_VERSION` instead?

python/src/tracing.rs
```diff
@@ -53,6 +53,8 @@ use tracing_subscriber::Registry;
 static SUBSCRIBER: LazyLock<Arc<RwLock<Option<LoggingPassthroughState>>>> =
     LazyLock::new(|| Arc::new(RwLock::new(None)));
 
+const SDK_VERSION: &str = env!("CARGO_PKG_VERSION");
```
Xuanwo: I also find it strange to have this const here; maybe we could move it to `lib.rs` instead.

Xuanwo (approved):
Thank you!

PR #4349, opened 2025-07-30
fix: remove the broken large binary test with dictionary encoding && fsst compression

HaochengLIU:
Remove a test that never triggers dict encoding and FSST at the same time.

The `check_round_trip_encoding_random(...)` will create data with high cardinality which will always be larger than the thread. The metadata DICT_DIVISOR_META_KEY is not good enough to trigger both FSST and dict.

### DIFF
--- a/rust/lance-encoding/src/encodings/physical/binary.rs
+++ b/rust/lance-encoding/src/encodings/physical/binary.rs
@@ -553,25 +553,6 @@ pub mod tests {
 
     #[rstest]
     #[test_log::test(tokio::test)]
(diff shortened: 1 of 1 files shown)


HaochengLIU:
@Xuanwo pls review ty

HaochengLIU:
The failing test is irrevalent

Xuanwo (approved):
LGTM, would you like to fix the conflicts?

HaochengLIU:
@Xuanwo rebased, ty

PR #4350, opened 2025-07-30
feat: modernize metadata handling with incremental updates and table metadata

wjones127:
## Summary

This PR modernizes Lance's metadata handling system with two major improvements:

### 1. Refactored Metadata Update Architecture
- **Before**: Separate fields for upserts/deletes with inconsistent handling across metadata types
- **After**: Unified `UpdateMap` structure supporting both full replacements and incremental updates

Key changes:
- Updated `Operation::UpdateConfig` enum with new field structure:
  - `config_updates`: Library configuration settings
  - `table_metadata_updates`: User-defined table metadata  
  - `schema_metadata_updates`: Schema-level metadata
  - `field_metadata_updates`: Per-field metadata
- Added automatic translation of old-style transactions for full backward compatibility
- Updated commit loop to handle incremental updates via `apply_update_map` helper
- Fixed protobuf conversion to properly separate `schema_metadata` (tag 5) and `table_metadata` (tag 18)

### 2. Table Metadata API
New clean API for managing table-level metadata:

```rust
// Get current metadata
dataset.metadata() -> &HashMap<String, String>

// Unified update method  
dataset.update_metadata(
    values: HashMap<String, Option<String>>, // None = delete key
    replace: bool                           // true = replace entire map
) -> Result<()>
```

**Features:**
- ✅ Incremental updates (`replace=false`)
- ✅ Full replacement (`replace=true`) 
- ✅ Key deletion (pass `None` as value)
- ✅ Persistence across dataset reloads
- ✅ Proper separation from config and schema metadata

## Test Plan

- [x] All existing functionality preserved with full backward compatibility
- [x] New table metadata API tested comprehensively:
  - [x] Incremental updates and key additions
  - [x] Key deletion via `None` values
  - [x] Full replacement with `replace=true`
  - [x] Persistence across dataset reloads
- [x] All update-related tests pass (33/33)
- [x] All metadata-related tests pass (4/4)
- [x] Transaction conflict resolution tests updated and passing (19/19)

## Breaking Changes

None - full backward compatibility maintained through automatic translation of old-style transactions.


java/core/src/test/java/com/lancedb/lance/DatasetTest.java
```diff
@@ -861,7 +861,7 @@ void testReplaceSchemaMetadata(@TempDir Path tempDir) {
       Map<String, String> replaceMetadata = new HashMap<>();
       replaceMetadata.put("key1", "value1");
       replaceMetadata.put("key2", "value2");
-      dataset.replaceSchemaMetadata(replaceMetadata);
+      dataset.updateSchemaMetadata(replaceMetadata, true);
```
wjones127: Why don't we keep one instance of `replaceSchemaMetadata` in the test to ensure it still works.

java/core/src/test/java/com/lancedb/lance/DatasetTest.java
```diff
@@ -777,7 +777,7 @@ void testUpdateConfig(@TempDir Path tempDir) {
       Map<String, String> updateConfig = new HashMap<>();
       updateConfig.put("key1", "value1");
       updateConfig.put("key2", "value2");
-      dataset.updateConfig(updateConfig);
+      dataset.updateConfig(updateConfig, false);
```
wjones127: Can we have an overload that doesn't pass the `replace` argument? That would be backwards compatible. Then we can keep this line the same.

java/core/src/main/java/com/lancedb/lance/Dataset.java
```diff
@@ -814,8 +820,11 @@ public void replaceSchemaMetadata(Map<String, String> metadata) {
   /**
    * Replace target field metadata of the dataset. This method won't affect fields not in the map
    *
+   * @deprecated Use {@link #updateFieldMetadata(Map, boolean)} with replace=true instead.
+   *             This method will be removed in a future version.
    * @param fieldMetadataMap field id to metadata map
    */
+  @Deprecated
   public void replaceFieldMetadata(Map<Integer, Map<String, String>> fieldMetadataMap) {
```
wjones127: Let's implement `replaceFieldMetadata` in Java as just dispatching to `updateFieldmetadata(fieldMetadatMap, true)`. Then we can cleanup `nativeReplaceFieldMetadata` immediately.

java/core/src/main/java/com/lancedb/lance/Dataset.java
```diff
@@ -800,8 +803,11 @@ public Tags tags() {
   /**
    * Replace the schema metadata of the dataset.
    *
+   * @deprecated Use {@link #updateSchemaMetadata(Map, boolean)} with replace=true instead.
+   *             This method will be removed in a future version.
    * @param metadata the new table metadata
    */
+  @Deprecated
   public void replaceSchemaMetadata(Map<String, String> metadata) {
```
wjones127: Let's implement `replaceSchemaMetadata` as `updateSchemaMetadata(metadata, true)`. Then we can cleanup `nativeReplaceSchemaMetadata` in this PR.

java/core/src/main/java/com/lancedb/lance/Dataset.java
```diff
@@ -720,8 +720,11 @@ public Map<String, String> getConfig() {
    * Update the config of the dataset. This operation will only overwrite and NOT delete the
    * existing config.
    *
+   * @deprecated Use {@link #updateConfig(Map, boolean)} instead for more flexible metadata updates.
+   *             This method will be removed in a future version.
    * @param tableConfig the config to update
    */
+  @Deprecated
```
wjones127: Honestly I think it's fine if we keep this as an overload.

However, we should just make the Java implementation call `updateConfig(tableConfig, false)`.

python/python/lance/dataset.py
```diff
@@ -784,26 +784,233 @@ def replace_schema_metadata(self, new_metadata: Dict[str, str]):
...
         new_metadata: dict
             The new metadata to set
         """
+        warnings.warn(
+            "replace_schema_metadata is deprecated. "
+            "Use update_schema_metadata(metadata, replace=True) instead.",
+            DeprecationWarning,
+            stacklevel=2,
+        )
         self._ds.replace_schema_metadata(new_metadata)
```
wjones127: Same as java: replace with `self.update_schema_metadata(new_metadata, replace=True)`

python/python/lance/dataset.py
```diff
@@ -784,26 +784,233 @@ def replace_schema_metadata(self, new_metadata: Dict[str, str]):
...
         new_metadata: dict
             The new metadata to set
         """
+        warnings.warn(
+            "replace_field_metadata is deprecated. "
+            "Use update_field_metadata with field paths and replace=True instead.",
+            DeprecationWarning,
+            stacklevel=2,
+        )
         self._ds.replace_field_metadata(field_name, new_metadata)
```
wjones127:
```suggestion
        self.update_field_metadata({field_name: new_metadata}, replace=True)
```

python/python/lance/dataset.py
```diff
@@ -784,26 +784,233 @@ def replace_schema_metadata(self, new_metadata: Dict[str, str]):
...
+        Get the table metadata of the dataset.
+
+        Returns
+        -------
+        Dict[str, str]
+            The table metadata as a dictionary of key-value pairs.
+        """
+        return self._ds.get_table_metadata()
+
+    def metadata(self) -> Dict[str, str]:
```
wjones127, line 1049: Make this a `@property`

python/python/lance/dataset.py
```diff
@@ -784,26 +784,233 @@ def replace_schema_metadata(self, new_metadata: Dict[str, str]):
...
+            "replace_field_metadata is deprecated. "
+            "Use update_field_metadata with field paths and replace=True instead.",
+            DeprecationWarning,
+            stacklevel=2,
+        )
         self._ds.replace_field_metadata(field_name, new_metadata)
 
+    # Unified metadata APIs
+
+    def get_table_metadata(self) -> Dict[str, str]:
```
wjones127: We don't need two aliases for this. Delete this one.

python/python/lance/dataset.py
```diff
@@ -784,26 +784,233 @@ def replace_schema_metadata(self, new_metadata: Dict[str, str]):
...
+        This is an alias for `get_table_metadata()`.
+
+        Returns
+        -------
+        Dict[str, str]
+            The table metadata as a dictionary of key-value pairs.
+        """
+        return self._ds.metadata()
+
+    def schema_metadata(self) -> Dict[str, str]:
```
wjones127, line 1061: This is redundant. users can just call `dataset.schema.metadata`, right?

java/core/src/main/java/com/lancedb/lance/Dataset.java
```diff
@@ -771,14 +771,9 @@ public Map<String, String> getConfig() {
    * @param tableConfig the config to update
    */
   public void updateConfig(Map<String, String> tableConfig) {
-    try (LockManager.WriteLock writeLock = lockManager.acquireWriteLock()) {
-      Preconditions.checkArgument(nativeDatasetHandle != 0, "Dataset is closed");
-      nativeUpdateConfig(tableConfig);
-    }
+    updateConfig(tableConfig, false);
```
jackye1995: Instead of adding a ton of new JNI methods for setting all these update maps, we can just mark `updateConfig` as `@Deprecated` and route it to call the new transaction builder instead, something like

```java
dataset
  .newTransactionBuilder()
  .operation(UpdateConfig.builder()
    .configUpdates(...)
    .build()
  )
  .commit()
``` 

and let users use the transaction builder API for performing other types of metadata updates.

jackye1995 (approved):
Read through all the changes, overall looks good to me! 1 comment regarding the java side change.

PR #4351, opened 2025-07-30
fix: fix bug in trim_ranges which could cause filtered read to miss rows

westonpace:
Minor logic bug in `trim_ranges` which is what applies the user's desired range (limit/offset) to the available (not deleted) rows in a fragment.  Without this fix the method would include some deleted rows in the output which would then get removed.  The end result was that a user would ask for something like `offset=20, limit=10` and get back less than 10 rows (even though 10 were available)


wkalt (approved):
:+1:  pending CI

PR #4352, opened 2025-07-30
refactor(rust): distinguish row id and row addr part-1

yanghua:


yanghua:
@jackye1995 The refactor has an extensive call chain in many files. And I try to choose a suite size for the pr changes. So I named the split PR by the part-n pattern. 

Let me know if it's OK with you. Or any suggestions?

yanghua:
@jackye1995 Please take a look, thanks!

rust/lance-index/src/scalar/inverted/index.rs
```diff
@@ -1830,7 +1830,7 @@ impl DocSet {
                 .iter()
                 .filter_map(|id| {
                     if let Some(frag_reuse_index_ref) = frag_reuse_index.as_ref() {
-                        frag_reuse_index_ref.remap_row_id(*id)
+                        frag_reuse_index_ref.remap_row_addr(*id)
```
wjones127: I don't understand some of these changes. Indices map queries to row ids, not row addresses, right?

yanghua: If we disable the `move stable row id` feature, we need remapping after compaction, here the row id is equal to the row address. 
If we enable the `move stable row id` feature, the row ID would not be changed after compaction, so we do not need to remap, right?

Please correct me if my understanding is incorrect.

This is what we want to distinguish after we realize the `move stable row id` feature. I saw some information about the original `remap_row_id`, it seems that it is used as`remap_row_addr`. For example, here:
https://github.com/lancedb/lance/blob/69f8c2adcf98c8d93932c999ebe6bb6643469544/rust/lance/src/index.rs#L232

And this description also describes the mapping is for address: https://github.com/lancedb/lance/blob/69f8c2adcf98c8d93932c999ebe6bb6643469544/rust/lance-index/src/frag_reuse.rs#L206

yanghua: After thinking, I think some renaming changes may cause a misunderstanding. IMO, a good solution is to introduce a struct like this in your draft PR:

```
pub enum CapturedRowIds {
    AddressStyle(RoaringTreemap),
    SequenceStyle(RowIdSequence),
}
```

yanghua: Ideally, if we own the row id, it would be better to get row address via:

https://github.com/lancedb/lance/blob/69f8c2adcf98c8d93932c999ebe6bb6643469544/rust/lance-table/src/rowids/index.rs#L55

But for now, when disabled `move stable row id` and if a context depends on row id, it seems the easiest way is let it go.

jackye1995:
We probably need to mark this as breaking change?

yanghua:
> We probably need to mark this as a breaking change?

I thought that if we touch the public API's name, and it causes compatibility issues, in that case we should mark a breaking change?

rust/lance-index/src/scalar/inverted/index.rs
wjones127: Okay that makes more sense to me.

wjones127:
Yeah I don't think this is a breaking change. As far as I can see, this doesn't affect the public API.

rust/lance-core/src/utils/mask.rs
```diff
@@ -652,14 +652,14 @@ impl RowIdTreeMap {
...
-    /// Convert the set into an iterator of row ids
+    /// Convert the set into an iterator of row addrs
     ///
     /// # Safety
     ///
     /// This is unsafe because if any of the inner RowIdSelection elements
     /// is not a Partial then the iterator will panic because we don't know
     /// the size of the bitmap.
-    pub unsafe fn into_id_iter(self) -> impl Iterator<Item = u64> {
+    pub unsafe fn into_addr_iter(self) -> impl Iterator<Item = u64> {
```
jackye1995, line 700: this one seems to be oriented for use for stable row ID? Because it does not really suit for cases like a Full selection.

rust/lance-index/src/scalar/lance_format.rs
```diff
@@ -1389,7 +1389,7 @@ pub mod tests {
                 let row_ids = result.row_ids();
 
                 let row_ids_set = row_ids
-                    .row_ids()
+                    .row_addrs()
```
jackye1995: I though you said all index related changes will be in a separated PR?

yanghua: Yes, this is due to `RowIdTreeMap#row_ids` -> `RowIdTreeMap#row_addrs`, which indirectly touched this file. Ok, let me revert these changes.

rust/lance-core/src/utils/mask.rs
yanghua: Not very sure, the main usage is `FilteredReadStream#plan_scan`, it seems it is used to precisely read based on the exact results of the index.

yanghua: The logic is used to build the row address. Do you suggest we keep the original name here?

yanghua:
@jackye1995 Have reverted all the changes related to the index.

yanghua:
@jackye1995 When you have time, PTAL.

rust/lance-core/src/utils/mask.rs
jackye1995: I see, I was confused by the usage in RowIdSelection

jackye1995 (approved):
looks good to me, sorry for the delay!

PR #4354, opened 2025-07-31
feat: allow configure encoding via field metadata

Xuanwo:
Part of https://github.com/lancedb/lance/issues/4302

This PR will allow users to configure encoding via field metadata

### DIFF
--- a/rust/lance-encoding/src/compression.rs
+++ b/rust/lance-encoding/src/compression.rs
@@ -16,13 +16,10 @@
 //! Fullzip compression is a per-value approach where we require that values are transparently
 //! compressed so that we can locate them later.
 
-/// Default threshold for RLE compression selection.
-/// RLE is chosen when the run count is less than this fraction of total values.
-const DEFAULT_RLE_COMPRESSION_THRESHOLD: f64 = 0.5;
-
 use crate::{
     buffer::LanceBuffer,
     compression_config::{CompressionFieldParams, CompressionParams},
+    constants::{COMPRESSION_LEVEL_META_KEY, COMPRESSION_META_KEY, RLE_THRESHOLD_META_KEY},
     data::{DataBlock, FixedWidthDataBlock, VariableWidthBlock},
     encodings::{
         logical::primitive::{fullzip::PerValueCompressor, miniblock::MiniBlockCompressor},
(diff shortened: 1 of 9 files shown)


rust/lance-encoding/src/compression.rs
```diff
@@ -54,7 +54,7 @@ use crate::{
 use arrow::{array::AsArray, datatypes::UInt64Type};
 use fsst::fsst::{FSST_LEAST_INPUT_MAX_LENGTH, FSST_LEAST_INPUT_SIZE};
 use lance_core::{
-    datatypes::{Field, COMPRESSION_META_KEY, RLE_THRESHOLD_META_KEY},
+    datatypes::{Field, COMPRESSION_LEVEL_META_KEY, COMPRESSION_META_KEY, RLE_THRESHOLD_META_KEY},
```
westonpace: Minor nit: I wonder if we should move these constants out of `lance_core` and into here?

rust/lance-encoding/src/compression.rs
```diff
@@ -131,11 +131,33 @@ impl DefaultCompressionStrategy {
...
+
+        params
+    }
+
     /// Build compressor based on parameters for fixed-width data
     fn build_fixed_width_compressor(
         &self,
         params: &CompressionFieldParams,
-        field: &Field,
+        _field: &Field,
```
westonpace: Should we just remove the parameter?  This is a private function

rust/lance-encoding/src/encodings/physical/bitpack.rs
```diff
@@ -461,20 +461,14 @@ mod test {
     async fn test_miniblock_bitpack() {
         let test_cases = TestCases::default().with_file_version(LanceFileVersion::V2_1);
 
-        let mut metadata = HashMap::new();
-        metadata.insert(
-            "lance-encoding:compression".to_string(),
-            "bitpacking".to_string(),
-        );
-
```
westonpace, line 469: Why this change?

rust/lance-encoding/src/encodings/physical/bitpack.rs
```diff
@@ -493,12 +487,7 @@ mod test {
                 arrays.push(arrow_cast::cast(&int64_array, &data_type).unwrap());
             }
 
-            let mut metadata = HashMap::new();
-            metadata.insert(
-                "lance-encoding:compression".to_string(),
-                "bitpacking".to_string(),
-            );
-            check_round_trip_encoding_of_data(arrays, &test_cases, metadata).await;
+            check_round_trip_encoding_of_data(arrays, &test_cases, HashMap::new()).await;
```
westonpace, line 490: Same question.  Why change this test in this way?

westonpace (approved):
Thanks, this is a nice cleanup to simplify the compression selection logic!

rust/lance-encoding/src/encodings/physical/bitpack.rs
Xuanwo: We added that metadata as a workaround to enforce the use of `bitpacking` during the implementation of RLE： https://github.com/lancedb/lance/pull/4177, but in the end, we no longer needed it, so we can simply revert those changes.

Those cod don't have any meaning after we remove the hack for `lance-encoding:compression`, but in this PR, we handled it correctly and exposed it again.

rust/lance-encoding/src/compression.rs
Xuanwo: Makes sense, let me do this change.

rust/lance-encoding/src/compression.rs
Xuanwo: Fixed

Issue #4355, opened 2025-07-31
Process hung when using tracing events


ddupg:
When using tracing events, there is a chance that the process becomes hung and fails to exit properly. As shown in the simple example below:
```python
import lance
import pyarrow as pa

from lance.tracing import capture_trace_events

def callback(evt):
    print(evt)

capture_trace_events(callback)

lance.write_dataset(pa.table({"x": range(100)}), "memory://test")
```

I obtained the thread stack dumps during the hang.
```
* thread #1, queue = 'com.apple.main-thread', stop reason = signal SIGSTOP
  * frame #0: 0x000000018cbaa9b8 libsystem_kernel.dylib`__ulock_wait + 8
    frame #1: 0x000000018cbed0b0 libsystem_pthread.dylib`_pthread_join + 608
    frame #2: 0x000000015ccfaba4 lance.abi3.so`std::sys::pal::unix::thread::Thread::join::hc92f1072ed8c169c at thread.rs:316:28 [opt]
    frame #3: 0x000000015500cd50 lance.abi3.so`std::thread::JoinInner$LT$T$GT$::join::h626ba985ca8a16f5(self=<unavailable>) at mod.rs:1741:9
    frame #4: 0x000000015500cec4 lance.abi3.so`std::thread::JoinHandle$LT$T$GT$::join::h9173357c1ab28df6(self=<unavailable>) at mod.rs:1887:9
    frame #5: 0x00000001549f45a8 lance.abi3.so`lance::tracing::LoggingPassthroughState::shutdown::h851386404889663a(self=0x0000600000a4cac0) at tracing.rs:172:13
    frame #6: 0x00000001549f6c14 lance.abi3.so`lance::tracing::shutdown_tracing::h6ffd10a7acc396c5((null)=(__0 = core::marker::PhantomData<(&pyo3::gil::GILGuard, pyo3::impl_::not_send::NotSend)> @ 0x000000016f00a06e)) at tracing.rs:379:5
    frame #7: 0x00000001549f6ca4 lance.abi3.so`lance::tracing::__pyfunction_shutdown_tracing::he099b456e4bd6f87((null)=(__0 = core::marker::PhantomData<(&pyo3::gil::GILGuard, pyo3::impl_::not_send::NotSend)> @ 0x000000016f00a18f), _slf=0x0000000000000000) at tracing.rs:376:1
[...]

PR #4356, opened 2025-07-31
fix: process hung when using tracing events

ddupg:
closes: https://github.com/lancedb/lance/issues/4355

### DIFF
--- a/python/src/tracing.rs
+++ b/python/src/tracing.rs
@@ -364,17 +364,20 @@ pub fn initialize_tracing(level: log::Level) {
 
 #[pyfunction]
 #[pyo3(signature=(callback))]
-pub fn capture_trace_events(callback: PyObject) {
+pub fn capture_trace_events(callback: PyObject, py: Python<'_>) {
     SUBSCRIBER
         .write()
         .unwrap()
         .as_mut()
         .unwrap()
-        .set_callback(callback);
+        .set_callback(callback.clone_ref(py));
 }
 
 #[pyfunction]
 #[pyo3(signature=())]
-pub fn shutdown_tracing() {
-    SUBSCRIBER.write().unwrap().as_mut().unwrap().shutdown();
+pub fn shutdown_tracing(py: Python<'_>) {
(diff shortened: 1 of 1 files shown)


ddupg:
I'm new to pyo3 and unsure if this is the proper fix, could @westonpace @Xuanwo please help review this?

python/src/tracing.rs
```diff
@@ -364,17 +364,20 @@ pub fn initialize_tracing(level: log::Level) {
...
 #[pyo3(signature=(callback))]
-pub fn capture_trace_events(callback: PyObject) {
+pub fn capture_trace_events(callback: PyObject, py: Python<'_>) {
     SUBSCRIBER
         .write()
         .unwrap()
         .as_mut()
         .unwrap()
-        .set_callback(callback);
+        .set_callback(callback.clone_ref(py));
```
westonpace, line 373: Is this change needed or just a performance optimization?

python/src/tracing.rs
```diff
@@ -364,17 +364,20 @@ pub fn initialize_tracing(level: log::Level) {
...
 #[pyfunction]
 #[pyo3(signature=())]
-pub fn shutdown_tracing() {
-    SUBSCRIBER.write().unwrap().as_mut().unwrap().shutdown();
+pub fn shutdown_tracing(py: Python<'_>) {
+    // Release Python GIL to avoid deadlock between current thread with the receiver thread.
+    py.allow_threads(|| {
+        SUBSCRIBER.write().unwrap().as_mut().unwrap().shutdown();
+    });
 }
```
westonpace, line 383: This seems fine to me.  There is no reason we would need the GIL for this call.

python/src/tracing.rs
ddupg: > Is this change needed or just a performance optimization?

This change is not close to the deadlock issue. This is to manually increase the reference count of python objects and avoid dangling references, as this `callback` will be held by the rust receiver thread for a long time.

PR #4362, opened 2025-08-01
fix: make sure cloud features aren't implicitly enabled (lancedb/lancedb#2567)

vlovich:
lance-io includes cloud features and critical types/methods that sibling libraries need without strictly needing cloud features. Make sure that lance-io is depended on by default without the default feature set except for lance which will pull in the requisite features as they're enabled.

This is so that building lancedb with default-features = false excludes the entire dependency chain for crates supporting the cloud features, shrinking the size and speeding up the build of the purely embedded version.

Both this and https://github.com/lancedb/lancedb/pull/2568 resolves https://github.com/lancedb/lancedb/issues/2567

Verified via:
```
cargo tree -p lancedb -e no-build -e no-dev --no-default-features -i aws-config | less
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -51,7 +51,7 @@ lance-datagen = { version = "=0.32.1", path = "./rust/lance-datagen" }
 lance-encoding = { version = "=0.32.1", path = "./rust/lance-encoding" }
 lance-file = { version = "=0.32.1", path = "./rust/lance-file" }
 lance-index = { version = "=0.32.1", path = "./rust/lance-index" }
-lance-io = { version = "=0.32.1", path = "./rust/lance-io" }
+lance-io = { version = "=0.32.1", path = "./rust/lance-io", default-features = false }
 lance-linalg = { version = "=0.32.1", path = "./rust/lance-linalg" }
 lance-table = { version = "=0.32.1", path = "./rust/lance-table" }
 lance-test-macros = { version = "=0.32.1", path = "./rust/lance-test-macros" }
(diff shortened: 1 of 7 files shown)


Xuanwo (approved):
Thank you, I love this change.

Xuanwo:
Many tests are failed after this change:

```
  $ /home/runner/.cargo/bin/cargo clippy --message-format json --locked --features aws,aws-sdk-dynamodb,azure,clap,cli,datafusion,datafusion-common,datafusion-sql,default,dynamodb,dynamodb_tests,fp16kernels,gcp,gcs-test,jieba-rs,lapack,lindera,lindera-tantivy,object_store_opendal,opendal,oss,prost_old,substrait,tensorflow,tfrecord,tokenizer-jieba,tokenizer-lindera --tests --benches --examples -- -Dwarnings
  error: Package `lance-table v0.32.1 (/home/runner/work/lance/lance/rust/lance-table)` does not have feature `aws-sdk-dynamodb`. It has an optional dependency with that name, but that dependency uses the "dep:" syntax in the features table, so it does not have an implicit feature with that name.
  Dependency `aws-sdk-dynamodb` would be enabled by these features:
  	- `dynamodb`
```

vlovich:
Thanks @Xuanwo - it looks like optional crates that are imported directly by `name` instead of by `dep:name` implicitly creates a `name` feature under your project which is why this is broken - lance has this "feature" but it was removed from "lance-table". I'll take a look and see if there's anywhere else this pops up & prepare an update to this PR.

vlovich:
Yup - there were a few other spots that dependent on a package by `name` instead of `dep:name` so I split out cleaning that up into a separate chore commit.

wjones127 (approved):

Issue #4181, opened 2025-07-09
[Feature] Lance's snapshot should support store summary property


xx789633:
While integrating Lance with Fluss, we want to store some summary property to record the current Fluss tiering status when Fluss commit a snapshot in Lance . But Lance does not support this feature, I think we can introduce a summary property into snapshot just like iceberg does.

paimon: https://github.com/apache/paimon/issues/5649
iceberg: https://github.com/apache/iceberg/pull/4956

jackye1995:
We discussed about this in https://github.com/lancedb/lance/pull/4078#discussion_r2165688242

should be a quick feature to add if you are interested, or I don't know if @chenghao-guo is planning to include that in his original PR

xx789633:
Cool. Please expose this feature in Java API.

chenghao-guo:
> We discussed about this in [#4078 (comment)](https://github.com/lancedb/lance/pull/4078#discussion_r2165688242)
> 
> should be a quick feature to add if you are interested, or I don't know if [@chenghao-guo](https://github.com/chenghao-guo) is planning to include that in his original PR

Hi @jackye1995 thanks for the suggestion! Sorry @cxxwang789 , I'm tied up with other priorities and won't be able to dedicate time to this feature immediately.  Assuming we can hold off until late July, I’ll have capacity to add this and progress the work.

Previously I was working on a simple design by setting a property in struct Transaction.
```
pub struct Transaction {
     ...............
    ///intended to add an extra_metadata to store the properties for a transaction
    **pub extra_metadata: Option<HashMap<String, String>>,**
}
```
Then bypass the write_dataset/add_columns optional parameter to transaction and stored in txn file. At that previous PR, I was only considering adding a simple commit_message, which is not general.

To meet your @cxxwang789 requirement, I believe the below works need to be done.
1. add the extra_metadata in Transaction and allows for parameter bypass as mentioned, currently a lot of rust methods has used Transaction, so there are still some works to be done to prevent making breaking changes.
2. expose in Java API. Regarding this , what kind of java API we need? for example, do we call  Dataset commit() in Java for this integration? If so, I can directly expose using existing commit() method.  Feel free to discuss it here if it doesn't meet the requirement.

I’ll circle back when bandwidth allows by the end of this month.  I hope that don’t let this block the progress of other PRs. Feel free to add it directly if it’s urgent!

leonardBang:
[...]

PR #4363, opened 2025-08-01
feat(java): support configure transaction properties

### LINKED ISSUE #4361: Java transaction supports transaction properties
This issue proposes enhancing the existing Java Transaction API to support setting and getting transaction properties. This involves adding the appropriate parameters to the Java Transaction class and updating the underlying rust commit() bindings to correctly pass transaction_properties to the TransactionBuilder introduced by pr #4078 . 
```
        TransactionBuilder::new(read_version, operation)
            .blobs_op(blobs_op)
            .transaction_properties(transaction_properties)
            .build()
```
This Java api is required for summary property  integration with Fluss, as detailed in issue #4181.

majin1102:
Close #4361

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -51,7 +51,7 @@ lance-datagen = { version = "=0.32.1", path = "./rust/lance-datagen" }
 lance-encoding = { version = "=0.32.1", path = "./rust/lance-encoding" }
 lance-file = { version = "=0.32.1", path = "./rust/lance-file" }
 lance-index = { version = "=0.32.1", path = "./rust/lance-index" }
-lance-io = { version = "=0.32.1", path = "./rust/lance-io" }
+lance-io = { version = "=0.32.1", path = "./rust/lance-io", default-features = false }
 lance-linalg = { version = "=0.32.1", path = "./rust/lance-linalg" }
 lance-table = { version = "=0.32.1", path = "./rust/lance-table" }
 lance-test-macros = { version = "=0.32.1", path = "./rust/lance-test-macros" }
--- a/rust/lance-core/Cargo.toml
+++ b/rust/lance-core/Cargo.toml
@@ -52,7 +52,7 @@ lance-testing.workspace = true
 proptest.workspace = true
 
 [features]
-datafusion = ["datafusion-common", "datafusion-sql"]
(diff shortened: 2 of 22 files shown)


majin1102:
@chenghao-guo @yanghua please help reviewing

java/core/src/test/java/com/lancedb/lance/TransactionTest.java
```diff
@@ -54,7 +54,12 @@ void testProjection(@TempDir Path tempDir) {
       assertEquals(testDataset.getSchema(), dataset.getSchema());
       List<Field> fieldList = new ArrayList<>(testDataset.getSchema().getFields());
       Collections.reverse(fieldList);
-      Transaction txn = dataset.newTransactionBuilder().project(new Schema(fieldList)).build();
+      Transaction txn =
+          dataset
+              .newTransactionBuilder()
+              .project(new Schema(fieldList))
+              .transactionProperties(Collections.singletonMap("op", "project"))
+              .build();
```
chenghao-guo, line 62: do we have a java get transaction  in UT, to validate if properties have been correctly set?

majin1102: For now I dont think so
That would need transaction reader api. would be another story(for Java binding，not sure this api is necessary)

I think when the reader api is ready，we could add tests for properties

chenghao-guo: Got it. This issue [#4181](https://github.com/lancedb/lance/issues/4181) may need the read api from the example, as the fluss may reuse to get the properties after set. We could probably add the reader api later.

For now, the PR looks great to me.

java/core/src/main/java/com/lancedb/lance/Transaction.java
```diff
@@ -109,6 +118,11 @@ public Builder readVersion(long readVersion) {
       return this;
     }
 
+    public Builder transactionProperties(Map<String, String> properties) {
+      this.writeParams = properties;
```
yanghua: Set the `properties` to `writeParams`? Is it correct?

majin1102: Yes, this is a mistake

yanghua (approved):
LGTM

PR #4366, opened 2025-08-01
ci: use cargo nextest to speed up our tests

Xuanwo:
This PR introduces `cargo-nextest`, splits our test cases into parts, and runs them in parallel. Let's see if it works for us.


Xuanwo:
We should split our CI workflow first.

PR #3813, opened 2025-05-13
feat!: use FilteredReadExec in the planner

westonpace:
Switches the planner to use the new FilteredReadExec node when not using legacy storage.

Starts to remove the storage_class prototype.  Keeping it up was too difficult in this change and the prototype has not seen use or active development into a proper feature.  That can still happen in the future.

Refactors most of the planner to use `Projection` instead of `Schema` to represent the user's projection.

BREAKING CHANGE: The names `_rowid` and `_rowaddr` are no longer valid column names (it's unclear they ever should have been and using these names would have likely led to errors)

PR #4375, opened 2025-08-02
ror_id

akhildevelops:


BubbleCal:
thanks for this work @akhildevelops 
I think this PR is ready to go, just need to format the code to get CI passed

Xuanwo:
I didn't get this. cc @BubbleCal would you want to take a look?

JaySon-Huang:
I meet the same issue when trying to run the full_text_search example
```
> cargo run --package lance-examples --example full_text_search
thread 'main' panicked at rust/examples/src/full_text_search.rs:56:14:
called `Result::unwrap()` on an `Err` value: InvalidInput { source: "The column _rowid is a reserved name and cannot be used in a Lance dataset", location: Location { file: "rust/lance/src/dataset/write/insert.rs", line: 355, column: 31 } }
```

I think the example is broken since https://github.com/lancedb/lance/pull/3813 which introduce a BREAKING CHANGE that I quote
> BREAKING CHANGE: The names `_rowid` and `_rowaddr` are no longer valid column names (it's unclear they ever should have been and using these names would have likely led to errors)

PR #4377, opened 2025-08-04
fix: Python-level strict overwrite

HubertY:
The "strict overwrite" commit type, which respects `read_version` and skips all retries for Overwrite operations when `num_retries=0`, was broken at the Python level by the automatic rebasing added to improve transaction performance.

The rebasing was checking out the most recent version of the dataset, causing `read_version` to be ignored.

This PR fixes the issue by skipping automatic rebasing when in strict overwrite mode, and adds a Python test for the behavior.


westonpace (approved):
This makes sense to me.  I'll rebase and merge on green.

PR #4380, opened 2025-08-04
feat: add verify for encoding roundtrip tests

### LINKED ISSUE #4249: Encoding fuzz tests should verify the encodings that are used
In our fuzz tests for encodings we generate some random data and then make sure we can encode it and decode it.  However, in some cases we have hit issues where the encoding we expected was not the encoding actually used.  For example, we want to test FSST so we generate some string data but the generation doesn't make enough unique values and so it ends up getting dictionary encoded instead and we never test FSST.

As part of the test we should force the test author to declare what encodings they are expecting will be used so we can verify that assumption.

Xuanwo:
close #4249

This PR introduces a `expected_encoding` for `TestCases`, developers can define the expected encoding for given tests at easy.

---

Another idea I have is to use the existing `EncodingVerificationFn`, but since it can be a bit difficult to use, perhaps we could provide a new utility function instead.

### DIFF
--- a/.github/workflows/nightly_run.yml
+++ b/.github/workflows/nightly_run.yml
@@ -2,7 +2,7 @@ name: Nightly Run Jobs
 
 on:
   schedule:
-    - cron: '0 0 * * *'  # Runs every day at midnight UTC
+    - cron: "0 0 * * *" # Runs every day at midnight UTC
   workflow_dispatch:
 
 jobs:
@@ -16,3 +16,26 @@ jobs:
           workflow: file_verification.yml
           ref: main
           token: ${{ secrets.GITHUB_TOKEN }}
+
+  jumbo-tests:
+    # jumbo tests need more resources
+    runs-on: warp-ubuntu-latest-x64-8x
+    if: github.repository == 'lancedb/lance'
+    timeout-minutes: 60
+    permissions:
+      contents: read
+    steps:
+      - uses: actions/checkout@v4
+      - name: Install dependencies
+        run: |
+          sudo apt update
+          sudo apt install -y protobuf-compiler libssl-dev
+      - name: Run Jumbo String/Binary Tests
+        run: |
+          echo "Running jumbo tests for Lance 2.0 and 2.1..."
+          echo "System memory info:"
+          free -h
+          cd rust
+          # Run only the jumbo tests in release mode with limited parallelism
+          # to avoid OOM issues (these tests use >5GiB memory each)
+          cargo test --release --package lance-encoding -- --ignored jumbo --test-threads=1
--- a/java/CLAUDE.md
+++ b/java/CLAUDE.md
@@ -1,0 +1,10 @@
+## Structure
+Java code: `core/src`
+Rust JNI bindings: `core/lance-jni`
+
+## Commands
+Use `./mvnw` instead of `mvn` to ensure the correct version of Maven is used.
+format: `./mvnw spotless:apply && cargo fmt --all`
+lint rust: `cargo clippy --tests`
+compile: `./mvnw compile`
+test: `./mvnw test`
--- a/protos/CLAUDE.md
+++ b/protos/CLAUDE.md
@@ -1,0 +1,2 @@
+All changes should be backwards compatible. Don't re-use field numbers of change
+field numbers of existing fields.
--- a/python/CLAUDE.md
+++ b/python/CLAUDE.md
@@ -1,0 +1,8 @@
+Use the makefile for most actions:
+
+* Build: `maturin develop`
+* Test: `make test`
+* Run single test: `pytest python/tests/<test_file>.py::<test_name>`
+* Doctest: `make doctest`
+* Lint: `make lint`
+* Format: `make format`
--- a/python/python/lance/dataset.py
+++ b/python/python/lance/dataset.py
@@ -4372,7 +4372,9 @@ def write_dataset(
     commit_lock: Optional[CommitLock] = None,
     progress: Optional[FragmentWriteProgress] = None,
     storage_options: Optional[Dict[str, str]] = None,
-    data_storage_version: Optional[str] = None,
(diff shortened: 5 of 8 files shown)


rust/lance-encoding/src/testing.rs
```diff
@@ -417,6 +420,159 @@ impl TestCases {
             verify_encoding(encoding);
         }
     }
+
+    pub fn with_expected_encoding(mut self, encoding: &str) -> Self {
+        self.expected_encoding = Some(vec![encoding.to_string()]);
+        self
+    }
```
westonpace: Super minor nit, could use `impl Into<String>` here.

rust/lance-encoding/src/testing.rs
```diff
@@ -417,6 +420,159 @@ impl TestCases {
...
+
+    pub fn with_expected_encoding_chain(mut self, encodings: Vec<&str>) -> Self {
```
westonpace: Super minor nit: could use `&[impl Into<String>]`

rust/lance-encoding/src/testing.rs
```diff
@@ -417,6 +420,159 @@ impl TestCases {
...
+                }
+            }
+            ArrayEncodingEnum::ByteStreamSplit(_) => {
+                chain.push("byte_stream_split".to_string());
+            }
+        }
+    }
+
+    chain
+}
```
westonpace: It would be nice if there was some way to generalize this but I don't think protobuf lends itself to that very nicely :frowning:

rust/lance-encoding/src/testing.rs
```diff
@@ -417,6 +420,159 @@ impl TestCases {
...
+                    Layout::AllNullLayout(_) => {
+                        // No value encoding for all null
+                    }
+                }
+            }
+        }
+        PageEncoding::Legacy(encoding) => {
+            let chain = extract_array_encoding_chain(encoding);
+            actual_chain.extend(chain);
+        }
```
westonpace: I wonder if we should even bother with legacy.

rust/lance-encoding/src/testing.rs
```diff
@@ -564,7 +739,9 @@ async fn check_round_trip_encoding_inner(
     }
     let mut column_infos = Vec::new();
     for (col_idx, encoded_column) in encoded_columns.into_iter().enumerate() {
-        for page in encoded_column.final_pages {
+        // Keep track of pages for encoding verification
+        let pages = encoded_column.final_pages;
+        for page in pages {
```
westonpace: This change doesn't seem needed?

rust/lance-encoding/src/encodings/physical/rle.rs
```diff
@@ -1016,4 +955,79 @@ mod tests {
...
+        }
+
+        fn data_type(&self) -> &arrow_schema::DataType {
+            &arrow_schema::DataType::Int32
+        }
+
+        fn element_size_bytes(&self) -> Option<lance_datagen::ByteCount> {
+            Some(lance_datagen::ByteCount::from(4))
+        }
+    }
```
westonpace, line 1032: Can we keep this in the `lance-datagen` package?  It seems generally useful.  Also, we might be able to implement it as a generic "repeat" generator that repeats elements from an underlying generator (either in fixed run lengths or random run lengths)

Can do in a follow-up too

rust/lance-encoding/src/encodings/physical/rle.rs
```diff
@@ -1016,4 +955,79 @@ mod tests {
             }
         }
     }
+
+    // ========== Encoding Verification Tests ==========
+
+    #[test_log::test(tokio::test)]
+    async fn test_rle_encoding_verification() {
```
westonpace, line 962: Do we need a new test or should we just add `expected_encoding` to the existing tests?

rust/lance-encoding/src/encodings/physical/rle.rs
```diff
@@ -790,156 +790,95 @@ mod tests {
...
+            ("bug_4092_pattern", {
+                // Test exact scenario that produces 4092 bytes instead of 4096
+                let mut data = Vec::new();
+                for i in 0..1022 {
+                    data.push(i);
+                }
+                data.push(999999i32);
+                data.push(999999i32);
+                data
+            }),
```
westonpace, line 844: Nice cleanup

westonpace (approved):
Thanks!  This is very much what I was hoping for.  Nicely done.  I have some suggestions but none of them major.

rust/lance-encoding/src/testing.rs
Xuanwo: reverted.

rust/lance-encoding/src/testing.rs
Xuanwo: Let me give this a try.

rust/lance-encoding/src/encodings/physical/rle.rs
Xuanwo: Most tests have been modified by adding `expected_encoding` to existing cases, but some tests were not previously covered.

rust/lance-encoding/src/encodings/physical/rle.rs
Xuanwo: Tracked in https://github.com/lancedb/lance/issues/4392

rust/lance-encoding/src/testing.rs
Xuanwo: fixed

rust/lance-encoding/src/testing.rs
Xuanwo: Implemented as `I: IntoIterator<Item = T>, T: Into<String>`

Xuanwo:
Thank you @westonpace for the nice review, merging

PR #4384, opened 2025-08-04
Convert filter on rowid to take

wkalt:


rust/lance/src/dataset/scanner.rs
```diff
@@ -374,6 +374,50 @@ fn escape_column_name(name: &str) -> String {
...
+                            return Some(*val);
+                        }
+                        // Try to convert from Int64
+                        if let ScalarValue::Int64(Some(val)) = lit {
+                            if *val >= 0 {
+                                return Some(*val as u64);
+                            }
+                        }
+                    }
+                }
```
westonpace, line 413: Technically one of these branches is not needed (this one IIRC).  This is because we run DF's optimizer on filter expressions and it normalizes to `column op literal` I think.

rust/lance/src/datafusion/dataframe.rs
```diff
@@ -43,22 +76,30 @@ impl LanceTableProvider {
         ordered: bool,
     ) -> Self {
         let mut full_schema = Schema::from(dataset.schema());
-        let mut row_id_idx = None;
```
westonpace, line 46: I'm not sure the changes in this file are needed.  It seems you are interpreting the `with_row_id` parameter here as "user wants row id in the output".  However, the parameter means "user wants row id to appear in the table schema".

This change makes it so rowid _always_ appears in the schema (which you could make an argument for).  However, if we are going to do that, we might as well get rid of the `with_row_id` parameter.

rust/lance/src/datafusion/dataframe.rs
```diff
@@ -85,11 +126,16 @@ impl TableProvider for LanceTableProvider {
         limit: Option<usize>,
     ) -> datafusion::common::Result<Arc<dyn ExecutionPlan>> {
         let mut scan = self.dataset.scan();
+
+        // Check if _rowid is needed either in projection or filters
+        let mut needs_row_id = self.explicit_row_id;
```
westonpace, line 131: We should only include the row id in the output if it is requested by `projection`.  I don't think we want an `explicit_row_id`.

rust/lance/src/datafusion/dataframe.rs
```diff
@@ -100,6 +146,20 @@ impl TableProvider for LanceTableProvider {
...
             }
         }
+
+        // Check if _rowid is referenced in filters
+        for filter in filters {
+            if expr_references_rowid(filter) {
+                needs_row_id = true;
+                break;
+            }
+        }
```
westonpace, line 156: This doesn't seem correct.  If a user does `SELECT x WHERE _rowid = 7` then I would not expect two columns in the output (`_rowid` will be loaded from disk but dropped before the plan finishes).

I would think you could skip this entire loop.

rust/lance/src/dataset/fragment.rs
```diff
@@ -1817,10 +1817,13 @@ impl FragmentReader {
...
-            .output_schema
-            .try_with_column(ROW_ID_FIELD.clone())
-            .expect("Table already has a column named _rowid");
+        // Only add _rowid if it's not already in the schema
+        if self.output_schema.column_with_name(ROW_ID).is_none() {
+            self.output_schema = self
+                .output_schema
+                .try_with_column(ROW_ID_FIELD.clone())
+                .expect("Failed to add _rowid column");
+        }
```
westonpace, line 1826: How does `_rowid` end up in `output_schema`?  That seems incorrect.

rust/lance/src/dataset/scanner.rs
```diff
@@ -1427,6 +1471,52 @@ impl Scanner {
...
+            if let Some(take_exec) =
+                TakeExec::try_new(self.dataset.clone(), row_id_plan, projection)?
+            {
+                let mut plan: Arc<dyn ExecutionPlan> = Arc::new(take_exec);
+
+                // Apply post processing (limit, projection, etc.)
+                if let Some(limit) = self.limit {
+                    use datafusion::physical_plan::limit::GlobalLimitExec;
+                    plan = Arc::new(GlobalLimitExec::new(plan, 0, Some(limit as usize)));
+                }
```
westonpace, line 1514: Instead of repeating this logic I wonder if we could instead treat these sorts of plans as an additional input type.  In other words, instead of...

```
if has_rowid_filter(...) {
  create_plan_a()
else {
  create_plan_b()
}
```

Maybe we could modify stage 1 of create plan to be something like...

```
let plan = if self.has_rowid_filter() {
  // create TakeExec for row-id-scan
  // set filter to None
} else match (..) {
  // vector search
  // fts search
  // filtered read
};

// create rest of plan (filter, limit, offset, final projection, etc.)
```

As it stands I'm not sure something like `SELECT 2 * x WHERE _rowid = 7` will work because the final dynamic projection is not being applied in this alternate path.

westonpace:
I picked this up some in https://github.com/lancedb/lance/pull/4394

wkalt:
superseded

PR #4386, opened 2025-08-04
chore: make explicit string literal for the data storage format

eddyxu:
Use typing Literal for storage format strings


Xuanwo (approved):
Thank you for this!

Issue #2802, opened 2024-08-28
Primary key: write path changes


wjones127:
* [x] Rename the feature flag from move-stable row ids to primary keys
* [x] Update queries with primary keys
   * Preserves row ids, separated indexed and unindexed data, updated frag bitmap
* [x] Merge-insert queries with primary keys
   * Preserves row ids, separated indexed and unindexed data, updated frag bitmap

yanghua:
Hi @wjones127 I am working on the update-stable rowid feature. I found the sub-task:

* Rename the feature flag from move-stable row ids to primary keys

Have been marked as `completed`, but I searched the code base and did not find that this work has been done.

Want to check if it's a wrong mark? cc @jackye1995

wjones127:
It’s not merged but I think I was referring to my draft changes here https://github.com/lancedb/lance/pull/2727

yanghua:
> It’s not merged but I think I was referring to my draft changes here [#2727](https://github.com/lancedb/lance/pull/2727)

@wjones127 Would you still want to take over this work? 

Recently, I and Jack have worked on finishing off the `move stable row id`(#4289 and #4307). If you are busy with other things. I can help and push this feature with Jack, WDYT?

jackye1995:
> Rename the feature flag from move-stable row ids to primary keys

do we really need this rename @yanghua? I think we can just proceed with implementing the remaining feature and reuse the same feature flag.

Also ideally I think we should not call this a primary key, because in most database systems primary key is a user-defined key with columns in the table, not a system-internal key. Using stable row ID is probably the least confusing anyway

yanghua:
> do we really need this rename @yanghua? 

Seems not, just want to follow the front task split.

> I think we can just proceed with implementing the remaining feature and reuse the same feature flag.

Yes, we can.

> Also ideally I think we should not call this a primary key, because in most database systems primary key is a user-defined key with columns in the table, not a system-internal key. Using stable row ID is probably the least confusing anyway

+1, it seems the draft from @wjones127 also did not rename it to `primary key`.

PR #4388, opened 2025-08-05
feat(rust): support update stable rowid without index

yanghua:


yanghua:
This PR is not ready for review. It's an early PR for verifying CI.

yanghua:
@jackye1995 @wjones127  I am planning to split into two sub-tasks:

1) this PR: only process and reuse the row id;
2) a follow-up PR: process index-related logic;

Can we do this split?

rust/lance-table/src/rowids/index.rs
```diff
@@ -29,17 +29,19 @@ pub struct RowIdIndex(RangeInclusiveMap<u64, (U64Segment, U64Segment)>);
...
             .collect::<Vec<_>>();
         pieces.sort_by_key(|(range, _)| *range.start());
 
-        // Check for overlapping ranges and if found, return a NotImplementedError.
-        // We don't expect this pattern yet until we make row ids stable after
-        // updates.
-        if pieces.windows(2).any(|w| w[0].0.end() >= w[1].0.start()) {
+        // If users stable row ids, check for overlapping ranges and if found,
+        // return a NotImplementedError.
+        if !uses_stable_row_ids && pieces.windows(2).any(|w| w[0].0.end() >= w[1].0.start()) {
```
yanghua, line 44: Currently(in this PR), I tend to avoid checking this for stable row IDs. Next PR will consider this.

rust/lance/src/dataset/rowids.rs
```diff
@@ -432,48 +432,48 @@ mod test {
...
         let index = get_row_id_index(&dataset).await.unwrap().unwrap();
         assert!(index.get(0).is_some());
-        // Old address is still there.
-        assert_eq!(index.get(3), Some(RowAddress::new_from_parts(0, 3)));
-        // New location is there.
-        assert_eq!(index.get(5), Some(RowAddress::new_from_parts(1, 0)));
+        // the updated row ids mapping to new address
+        assert_eq!(index.get(3), Some(RowAddress::new_from_parts(1, 0)));
+        // there is no new row id
+        assert_eq!(index.get(5), None);
```
yanghua, line 438: These assertions need to be adjusted.

wjones127:
> this PR: only process and reuse the row id;
> a follow-up PR: process index-related logic;
> Can we do this split?

Yeah, that sounds fine.

yanghua:
Hi @wjones127 and @jackye1995, this PR is ready for review. Please take a look.

rust/lance/src/dataset/rowids.rs
wjones127: So is your plan to adjust the current move-stable row ids to make them update-stable? I hope no one is using the move-stable ones because these changes would break it for them, right?

wjones127 (commented):
Just one question

rust/lance/src/dataset/rowids.rs
yanghua: > So is your plan to adjust the current move-stable row ids to make them update-stable?

Based on @jackye1995 's suggestion(https://github.com/lancedb/lance/issues/2802#issuecomment-3135041135), the answer is yes.

I also think this is the lowest cost plan. If we introduce a new flag named `enable_update_stable_row_ids`. That will put a heavier debt of cleanup on us soon.

In the following period, I will prioritize implementing update-stable row ID. After that, I think we can unify the flag and remove the "move" keyword.

> I hope no one is using the move-stable ones because these changes would break it for them, right?

Right, these changes will affect the original `move-stable` feature. Based on the bug(https://github.com/lancedb/lance/pull/4289) we have found recently, it seems that not many users are using this feature currently?

Or we can print a warning log, looks like `At present, this feature is unstable and is under refactoring. Please enable it with caution.` when it is detected that the user has turned on this switch. WDYT?

wjones127: I think it's already labelled as experiment. I'm fine moving forward if you are. We aren't using this flag ourselves right now.

wjones127 (approved):

rust/lance/src/dataset/rowids.rs
yanghua: Thanks! I will try my best to push the update-stable row ID landing as soon as possible.

Issue #4385, opened 2025-08-04
bug: support nullability in struct


eddyxu:
Currently, lance does not allow null in struct. Be able to serialize null structs on disk in format v2.1

```python
import lance
import pyarrow as pa

schema = pa.schema(
    [
        pa.field("id", pa.int64()),
        pa.field(
            "point",
            pa.struct(
                [
                    pa.field("x", pa.float64(), nullable=True),
                    pa.field("y", pa.float64(), nullable=True),
                    pa.field("z", pa.float64(), nullable=True),
                ],
            ),
            nullable=True,
        ),
    ]
)


tbl = pa.Table.from_pylist([{"id": 5, "point": None}], schema=schema)
print("origin: ", tbl.to_pydict())

lance.write_dataset(tbl, "empty.lance", schema=schema, data_storage_version="2.1")
ds = lance.dataset("empty.lance")
tbl = ds.to_table()
print(tbl.to_pydict())
```

Output

```
origin:  {'id': [5], 'point': [None]}
[2025-08-04T15:35:35Z WARN  lance_file::v2::writer] You have requested an unstable format version.  Files written with this format version may not be readable in the future!  This is a development feature and should only be used for experimentation and never for production data.
{'id': [5], 'point': [{'x': 0.0, 'y': 0.0, 'z': 0.0}]}
```

westonpace:
This must be getting added in the fragment readers somewhere.  If I just read the file directly I get null.  For example, adding this to the repr script:

```
frag = ds.get_fragment(0)
path = frag.data_files()[0].path

from lance.file import LanceFileReader

reader = LanceFileReader(os.path.abspath(f"empty.lance/data/{path}"))
tbl = reader.read_all().to_table()

print(tbl.to_pydict())
```

Xuanwo:
I've set up a test to reproduce the issue and will work on fixing it tomorrow.

Xuanwo:
Fixed:

```shell
origin:  {'id': [5], 'point': [None]}
[2025-08-05T08:48:14Z WARN  lance_file::v2::writer] You have requested an unstable format version.  Files written with this format version may not be readable in the future!  This is a development feature and should only be used for experimentation and never for production data.
{'id': [5], 'point': [None]}
```

PR is coming

PR #4391, opened 2025-08-05
feat: support nullability in struct correctly

Xuanwo:
This PR will add correct support of nullability in struct.

In general, we should pass validity correctly in dataset. I mapped the issue's repro cases into rust for easier verify.

close https://github.com/lancedb/lance/issues/4385


python/python/tests/test_blob.py
```diff
@@ -174,9 +174,7 @@ def test_null_blobs(tmp_path):
             py_blob = blob.as_py()
             # When we write blobs to a file we store the position as 1 and size as 0
             # to avoid needing a validity buffer.
-            #
-            # TODO: We should probably convert these to null on read.
-            assert py_blob == {"position": None, "size": None} or py_blob == {
```
Xuanwo, line 179: cc @westonpace, I changed this test case, please let me know if it's not you expected.

westonpace: Yep, this is fine.

rust/lance-arrow/src/lib.rs
```diff
@@ -783,7 +783,8 @@ fn project(struct_array: &StructArray, fields: &Fields) -> Result<StructArray> {
             )));
         }
     }
-    StructArray::try_new(fields.clone(), columns, None)
+    // Preserve the struct's validity when projecting
+    StructArray::try_new(fields.clone(), columns, struct_array.nulls().cloned())
```
westonpace, line 787: Nice catch.  Guess we haven't had to worry about struct nullability in the table for a while :smile:

rust/lance-arrow/src/lib.rs
```diff
@@ -1015,13 +1016,11 @@ fn merge(left_struct_array: &StructArray, right_struct_array: &StructArray) -> S
...
-        .cloned()
-        .map(Arc::new)
-        .zip(columns.iter().cloned())
-        .collect::<Vec<_>>();
-    StructArray::from(zipped)
+    // Preserve validity from the left struct array (if any)
+    // This assumes that both structs have the same validity pattern if they're being merged
+    let validity = left_struct_array.nulls().cloned();
+
+    StructArray::try_new(Fields::from(fields), columns, validity).unwrap()
```
westonpace: Hmm...I'm not actually sure what the correct thing to do here is.  I don't know if we can assume the validity pattern is identical.

This method is primarily used when we are combining data from two different data files in the same fragment.

For example, maybe one data file has...

```
[
  { "metadata": { "height": 500 } },
  { "metadata": NULL },
  { "metadata": { "height": 600 } }
  { "metadata": NULL },
]
```

Then another file has...

```
[
  { "metadata": { "width": 300 } },
  { "metadata": { "width": 200 } },
  { "metadata": NULL },
  { "metadata": NULL },
]
```

I think we actually want the intersection of the validity but in places where the validity differs we need to push it down in the children.

```
[
  { "metadata": { "width": 300, "height": 500 } },
  { "metadata": { "width": 200, "height": NULL } },
  { "metadata": { "width": NULL, "height": 600 } },
  { "metadata": NULL },
]
```

CC @wjones127 for third opinion

westonpace (commented):
These are some good catches.  I'm not entirely sure we have the merge behavior correct though.

rust/lance-arrow/src/lib.rs
Xuanwo: > I think we actually want the intersection of the validity but in places where the validity differs we need to push it down in the children.

Makes sense to me, but I suppose it could get complicated when handling nested data. Let me prepare a test for that.

Xuanwo: Hi, @westonpace, please check if this test what you want: https://github.com/lancedb/lance/pull/4391/commits/3d1d1eb20514cfeeb64687998195c639daf23941

rust/lance-arrow/src/lib.rs
```diff
@@ -930,12 +931,124 @@ fn merge_list_struct(left: &dyn Array, right: &dyn Array) -> Arc<dyn Array> {
...
+                Some(left.clone())
+            }
+        }
+        (None, Some(right)) => {
+            // Check if right is a placeholder (all null)
+            if right.null_count() == right.len() {
+                None
+            } else {
+                Some(right.clone())
+            }
```
Xuanwo: I'm not sure if I handled this correctly; it's quite tricky. It seems like lance will generate an all-null validity while the data doesn't exist on this side.

This trick is solely for making test `test_insert_nested_subschemas` pass.

Can you @westonpace double check this test to ensure I didn't understand it in wrong?

westonpace: I think the logic is correct but you could maybe simplify it by dealing with placeholders first.

```
fn normalize_validity(validity: Option<&NullBuffer>) -> Option<&NullBuffer> {
  validity.and_then(|v| {
    if v.null_count() == v.len() {
      None
    } else {
      v
    }
  }
}

fn merge_struct_validity(
    left_validity: Option<&arrow_buffer::NullBuffer>,
    right_validity: Option<&arrow_buffer::NullBuffer>,
) -> Option<arrow_buffer::NullBuffer> {
  left_validity = normalize_validity(left_validity);
  right_validity = normalize_validity(right_validity);
  ...
```

rust/lance-arrow/src/lib.rs
```diff
@@ -930,12 +931,124 @@ fn merge_list_struct(left: &dyn Array, right: &dyn Array) -> Arc<dyn Array> {
...
+        }
+    };
+
+    // Create new array with adjusted validity
+    arrow_array::make_array(
+        arrow_data::ArrayData::try_new(
+            child.data_type().clone(),
+            child.len(),
+            Some(new_validity.into_inner().into_inner()),
+            0,
```
westonpace: Is this assuming `child` has no offset?

rust/lance-arrow/src/lib.rs
```diff
@@ -930,12 +931,124 @@ fn merge_list_struct(left: &dyn Array, right: &dyn Array) -> Arc<dyn Array> {
...
+            // This preserves the correct semantics: 1 = valid, 0 = null
+            let merged_buffer = child_buffer & parent_buffer;
+
+            arrow_buffer::NullBuffer::from(merged_buffer)
+        }
+    };
+
+    // Create new array with adjusted validity
+    arrow_array::make_array(
+        arrow_data::ArrayData::try_new(
```
westonpace, line 1019: Hmm, I wonder if we should use `new_unchecked` here if all we are changing is the validity it should be safe?

rust/lance-arrow/src/lib.rs
Xuanwo: Nice catch!

rust/lance-arrow/src/lib.rs
Xuanwo: > Hmm, I wonder if we should use `new_unchecked` here if all we are changing is the validity it should be safe?

That makes sense to me, but is this change really worth it? I usually keep things as they are until we see a significant improvement.

I'm open to fill an issue to implement it.

PR #4393, opened 2025-08-05
feat: allow configuring bss for encoding

Xuanwo:
This PR enables users to configure BSS for encoding.

We have implemented an internal BSS sensitivity concept to measure how sensitive we want to be, based on our byte position entity. However, we will not expose these details directly to users. Instead, we provide a simple interface for users to choose between `on`, `off`, and `auto`. The default value is `auto`, which means BSS will be enabled based on the data's statistics.

We can introduce additional statistics functions in the future. I don't see any blockers or conflicts.

### DIFF
--- a/rust/lance-encoding/src/compression.rs
+++ b/rust/lance-encoding/src/compression.rs
@@ -18,8 +18,10 @@
 
 use crate::{
     buffer::LanceBuffer,
-    compression_config::{CompressionFieldParams, CompressionParams},
-    constants::{COMPRESSION_LEVEL_META_KEY, COMPRESSION_META_KEY, RLE_THRESHOLD_META_KEY},
+    compression_config::{BssMode, CompressionFieldParams, CompressionParams},
+    constants::{
+        BSS_META_KEY, COMPRESSION_LEVEL_META_KEY, COMPRESSION_META_KEY, RLE_THRESHOLD_META_KEY,
+    },
     data::{DataBlock, FixedWidthDataBlock, VariableWidthBlock},
     encodings::{
         logical::primitive::{fullzip::PerValueCompressor, miniblock::MiniBlockCompressor},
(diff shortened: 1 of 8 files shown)


rust/lance-encoding/src/compression.rs
```diff
@@ -147,64 +227,34 @@ impl DefaultCompressionStrategy {
...
-            if compression_scheme != "none" && compression_scheme != "fsst" {
-                let scheme: CompressionScheme = compression_scheme.parse()?;
-                let config = CompressionConfig::new(scheme, params.compression_level);
-                base_encoder = Box::new(GeneralMiniBlockCompressor::new(base_encoder, config));
-            }
-        }
+        let base = try_bss_for_mini_block(data, params)
+            .or_else(|| try_rle_for_mini_block(data, params))
+            .or_else(|| try_bitpack_for_mini_block(data))
+            .unwrap_or_else(|| Box::new(ValueEncoder::default()));
```
westonpace, line 257: Nice and compact :)

rust/lance-encoding/src/encodings/physical/byte_stream_split.rs
```diff
@@ -257,6 +262,55 @@ impl MiniBlockDecompressor for ByteStreamSplitDecompressor {
     }
 }
 
+/// Determine if BSS should be used based on mode and data characteristics
+pub fn should_use_bss(data: &FixedWidthDataBlock, mode: BssMode) -> bool {
+    // Only support byte-aligned data
+    let bytes_per_value = data.bits_per_value / 8;
+    if data.bits_per_value % 8 != 0 || bytes_per_value < 2 {
+        return false;
+    }
```
westonpace: I thought somewhere else you limited it to 32 bits or 64 bits but here you are allowing 16 bits so that seems inconsistent?

westonpace (approved):
Looks like some good and careful thought here, thank you.

Should we disable BSS if general compression is not used?  BSS offers no compression on its own.

rust/lance-encoding/src/compression.rs
Xuanwo: Thank you for this!

rust/lance-encoding/src/encodings/physical/byte_stream_split.rs
Xuanwo: You are right, let me fix this,

PR #4394, opened 2025-08-05
feat: allow row ids (offsets, addresses) in filters to be translated into an optimized take operation

westonpace:
This is useful for APIs which do not have a dedicated take operation such as lancedb or SQL


westonpace:
Leaving in draft since this is stacked on https://github.com/lancedb/lance/pull/4340

rust/lance/src/dataset/scanner.rs
```diff
@@ -7204,4 +7367,126 @@ mod test {
...
+            .try_into_stream()
+            .await;
+        match res {
+            Ok(_) => panic!("Expected error"),
+            Err(e) => {
+                assert!(e
+                    .to_string()
+                    .contains("One or more row offsets went past the end of the dataset"))
+            }
+        }
```
wjones127: I wonder if the correct behavior should actually be to return an empty result set. When I think about the meaning of a SQL filter, that's what I would expect. Though that's definitely different than what one would expect from a take operation.

rust/lance/src/dataset/scanner.rs
```diff
@@ -7204,4 +7367,126 @@ mod test {
...
+
+        // Taking _rowid / _rowaddr of deleted row
+
+        // TODO: Would be nice to either get empty batches or an error when we try and take
+        // an id / addr that doesn't exist.  Today we get a bizarre error:
+        // check(&format!("_rowid = {}", row_ids.value(190)), &[]).await;
+        // check(&format!("_rowaddr = {}", row_addrs.value(190)), &[]).await;
+        check("_rowoffset = 190", &[210]).await;
+
+        // Grabbing after the deleted rows
```
wjones127, line 7537: What happens if you request a row id / row address that is deleted?

rust/lance/src/dataset/scanner.rs
westonpace: Seems like take could be made configurable for this

westonpace: Ok, I ended up switching to `FilteredRead` instead of `TakeExec` and that gives the behavior we'd expect.  Deleted rows and past-the-end rows are simply not included in the results.

rust/lance/src/dataset/scanner.rs
westonpace: Before you'd get "cannot merge batch of size 1 with batch of size 0".  Now it just doesn't show up in the results and you can get an empty result if all requested row ids are deleted.

PR #4398, opened 2025-08-07
fix: make sure aws feature is applied to code correctly

vlovich:
Some code seems to not have been annotated with a feature check for AWS correctly resulting in code failing to compile without the AWS feature.


vlovich:
The failing breaking change PR check seems unrelated to this PR?

Found breaking change in PR #4340: feat: allow empty projection

westonpace (approved):

westonpace:
Thanks for this fix!

PR #4404, opened 2025-08-07
feat(rust): support diff_meta api

yanghua:


rust/lance/src/dataset/delta.rs
```diff
@@ -0,0 +1,90 @@
...
+    pub(crate) begin_version: u64,
+    /// The current version number.
+    pub(crate) end_version: u64,
+    /// Base path of the dataset.
+    pub(crate) base_dataset: Dataset,
+}
+
+impl DatasetDelta {
+    /// Listing the transactions between two versions.
+    pub async fn diff_meta(&self) -> Result<Vec<Transaction>> {
```
majin1102: I wonder why this is a diff_meta not list_transactions

yanghua: The discussion is [here](https://github.com/lancedb/lance/pull/3979#issuecomment-3012601460. IMO, `diff_transactions` also sounds good. But, `list_transactions ` may not be suitable? Because it needs a `compared_version`.

yanghua: hmm... If we stand on the point of view of `DatasetDelta`. The `list_transactions` seems good.

majin1102: Yeah, I think if we look back to the dicussion and context. seems reasonable.

In my mental model, format layer could provide metadata tables including transaction. The interfaces could be designed according to the format itself. I think diff things would be something like the upper layer of applications. The model is like the format provides details, applications use them for different scenarios like diff.

Only my opinions for reference.

yanghua: Good point, have adopted the opinion!

yanghua:
@jackye1995 please take a look, thanks~

rust/lance/src/dataset/delta.rs
```diff
@@ -0,0 +1,90 @@
...
+    /// Listing the transactions between two versions.
+    pub async fn list_transactions(&self) -> Result<Vec<Transaction>> {
+        let mut transactions = Vec::new();
+        for version in (self.begin_version + 1)..=self.end_version {
+            let current_ds = self.base_dataset.checkout_version(version).await?;
+            if let Some(tx) = current_ds.read_transaction().await? {
+                transactions.push(tx);
+            }
+        }
+        Ok(transactions)
```
wjones127: Maybe you should run these in parallel with `buffered()`?

rust/lance/src/dataset/delta.rs
```diff
@@ -0,0 +1,90 @@
...
+    pub(crate) begin_version: u64,
+    /// The current version number.
+    pub(crate) end_version: u64,
+    /// Base path of the dataset.
+    pub(crate) base_dataset: Dataset,
+}
+
+impl DatasetDelta {
+    /// Listing the transactions between two versions.
+    pub async fn list_transactions(&self) -> Result<Vec<Transaction>> {
```
wjones127, line 23: What happens if one or more of the versions are deleted?

yanghua: Yes, it will throw a dataset not found exception. I have caught it and rethrown a more friendly exception.

yanghua:
I have adapted the suggestions. Will merge it, if there is any improvement, will fix it in the follow-up PR.

Issue #3796, opened 2025-05-09
Auto conflict resolution for delete


wjones127:

wjones127:
Closed by #4407

PR #4407, opened 2025-08-07
feat: add automatic conflict resolution support for delete operations

wjones127:
## Summary

Add `affected_rows` tracking to delete operations to enable automatic conflict resolution similar to update operations. This allows concurrent delete operations to succeed with proper retry-based conflict resolution.

## Changes

- **Add affected_rows field to DeleteData struct** - Stores the `RowIdTreeMap` containing row addresses affected by the delete operation
- **Capture RowIdTreeMap during delete execution** - Convert `removed_row_addrs` to `RowIdTreeMap::from()` for conflict detection
- **Pass affected_rows to CommitBuilder.with_affected_rows()** - Enable automatic conflict resolution in commit step
- **Add test_delete_concurrency test** - Similar to `test_update_concurrency`, verifies concurrent delete operations work correctly
- **Handle edge cases** - Proper handling for delete-all and delete-nothing scenarios

## Test Plan

- [x] All existing delete tests pass
- [x] New `test_delete_concurrency` test passes with 3 concurrent delete operations
- [x] `test_concurrent_delete_with_retries` continues to work with overlapping operations
- [x] Verified compatibility with retry-based conflict resolution infrastructure

Closes https://github.com/lancedb/lance/issues/3796


rust/lance/src/dataset/write/delete.rs
```diff
@@ -183,7 +185,9 @@ impl RetryExecutor for DeleteJob {
                         .iter()
                         .map(|f| f.id() as u64)
                         .collect();
-                    (Vec::new(), deleted_fragment_ids)
+                    // When deleting everything, we don't have specific row addresses,
+                    // but we can create an empty RowIdTreeMap since all fragments are deleted
+                    (Vec::new(), deleted_fragment_ids, RowIdTreeMap::new())
```
jackye1995: I think the logic should be right here based on the `conflict_resolver` logic, but can we add a test to confirm?

jackye1995 (approved):
looks good to me, just a comment regarding testing

rust/lance/src/dataset/write/delete.rs
wjones127: Good catch. Added a test and modified this not to return an `affected_rows` here.

Issue #4332, opened 2025-07-29
Java transaction supports Update and Delete operation


majin1102:

majin1102:
Both completed

PR #4408, opened 2025-08-08
feat(java): introduce update in transaction

fangbo:
Issue: https://github.com/lancedb/lance/issues/4332


java/core/src/main/java/com/lancedb/lance/operation/Update.java
```diff
@@ -0,0 +1,88 @@
...
+  public String toString() {
+    return String.format(
+        "Update{removedFragmentIds=%s, updatedFragments=%s, newFragments=%s}",
+        removedFragmentIds, updatedFragments, newFragments);
+  }
+
+  public static class Builder {
+    private List<Long> removedFragmentIds = Collections.emptyList();
+    private List<FragmentMetadata> updatedFragments = Collections.emptyList();
+    private List<FragmentMetadata> newFragments = Collections.emptyList();
```
majin1102, line 76: I wonder why leave them not adapted:

```
        /// The fields that have been modified
        fields_modified: Vec<u32>,
        /// The MemWAL (pre-image) that should be marked as flushed after this transaction
        mem_wal_to_flush: Option<MemWal>,
```

java/core/lance-jni/src/transaction.rs
```diff
@@ -299,6 +301,53 @@ fn convert_to_rust_operation(
                 config_upsert_values,
             }
         }
+        "Update" => {
```
majin1102, line 838: We need to implement read transaction path as well(convert_to_java_operation)

java/core/src/main/java/com/lancedb/lance/operation/Update.java
fangbo: > I wonder why leave them:
> 
> ```
>         /// The fields that have been modified
>         fields_modified: Vec<u32>,
>         /// The MemWAL (pre-image) that should be marked as flushed after this transaction
>         mem_wal_to_flush: Option<MemWal>,
> ```

At https://github.com/lancedb/lance/blob/main/rust/lance/src/dataset/write/update.rs#L437 , these two properties is not setted.

majin1102: Yeah, I see

I encountered some similar situations. I think this could be challenged for the unalignment with rust format. Maybe we should adapt them and don't use them in that case neither.

What do you think @jackye1995

jackye1995: `mem_wal_to_flush ` we can ignore those for now, WAL is WIP feature. 

`fields_modified` I think we have that for future proof, for example add a new column while upserting, but that is also not implement for now.

As long as we use a builder style, we could be able to add those new fields in the future.

java/core/lance-jni/Cargo.lock
jackye1995, line 1: I think this should not be committed?

jackye1995 (approved):
Looks good to me, pending rebase

fangbo:
> Looks good to me, pending rebase

Thanks very much . The PR has already been rebased.

Issue #3983, opened 2025-06-12
Performance issue: Lance point lookup 13x slower than LMDB


ddupg:
Hi everyone, I conducted a performance test to compare point lookup performance between Lance and LMDB, and I'm actively seeking methods to optimize Lance's performance, especially when compression is enabled.

## Environment
| Category       | Detail                                                                                                           |
|--------------|---------------------------------------------------------------------|
| Hardware      | CPU: 16 cores, RAM 96GB, Storage: SSD                                               |
| Software.      | OS: CentOS 9, Python: 3.10, Lance: 0.29, LMDB: 1.6                             |
| Dataset         | 65M records (1183 records). key: 32-char string, value: 54KB pickle. |

## Methodology:
- Cleared OS page cache before each test: `echo 3 | sudo tee /proc/sys/vm/drop_caches`.
- Tested 1,000 point lookups; reported average latency.

## Results Summary
|                                          |  LMDB       |  Lance           |  Lance(zstd)    |
|-------------------------|-----------|--------------|---------------|
| Time for 1,000 lookups  | 10.116 ms  | 138.539 ms  | 30992.766 ms |
| Avg. latency per lookup | 0.010 ms   | 0.138 ms       | 30.993 ms       |

## Key Observations
1. Lance’s point lookup latency is 13× higher than LMDB.
2. Enabling zstd compression in Lance increased latency by 223× vs. uncompressed Lance.

## Attachments
- [Test Code & Data](https://github.com/ddupg/lance-perf-test)

I'm eager to hear your suggestions on performance optimization, particularly for scenarios with compression enabled. Please share any insights or recommendations.
Happy to provide clarifications or additional details!

westonpace:
> I'm eager to hear your suggestions on performance optimization, particularly for scenarios with compression enabled.

> Enabling zstd compression in Lance increased latency by 223× vs. uncompressed Lance.

The 2.0 compression was always experimental and not great for random access.  It's what we've been working on in 2.1.  So I made some tweaks to run your experiment with 2.1.  I had too change the data type to `binary` instead of `large_binary` as we're working on large offset support at the moment.

Initially the bottleneck is table format overhead.  You're making 1000 calls for 1 row each and it takes time to figure out which file you're after, etc.  If I change it to one call with 1000 indices then it drops it down to about 70ms (with compression).  Note, I get about 7ms from lmdb on my hardware.

At this point the performance breakdown is:

* I/O - 4ms
* Decompression - 32ms
[...]

PR #4410, opened 2025-08-08
feat: dataset reader

wojiaodoubao:
Optimize repeated take using dataset api. See https://github.com/lancedb/lance/issues/4090 for more details.

### DIFF
--- a/rust/lance/src/dataset/dataset_reader.rs
+++ b/rust/lance/src/dataset/dataset_reader.rs
@@ -1,0 +1,434 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright The Lance Authors
+
+use crate::dataset::fragment::session::FragmentSession;
+use crate::dataset::take::{
+    do_take_rows, row_offsets_to_row_addresses, TakeFunction, TakeRangeFunction,
+};
+use crate::dataset::{ProjectionRequest, TakeBuilder};
+use crate::Dataset;
+use arrow_array::RecordBatch;
+use futures::FutureExt;
+use lance_core::Result;
+use lance_datafusion::projection::ProjectionPlan;
+use std::collections::HashMap;
+use std::sync::Arc;
+
+/// A [`DatasetReader`] manages a short-lived context of [`Dataset`], allowing us to maintain
+/// internal states instead of creating new ones each time.
+///
+/// This API works well for users making repeated requests over the same projection.
+pub struct DatasetReader {
+    dataset: Arc<Dataset>,
+    fragment_sessions: Arc<HashMap<usize, FragmentSession>>,
+    fragment_sessions_with_row_addr: Arc<HashMap<usize, FragmentSession>>,
+    projection: Arc<ProjectionPlan>,
+}
+
+impl DatasetReader {
+    pub async fn new(dataset: &Dataset, projection: impl Into<ProjectionRequest>) -> Result<Self> {
+        let dataset = Arc::new(dataset.clone());
+        let projection = Arc::new(projection.into().into_projection_plan(dataset.clone())?);
+        let schema = projection.physical_projection.to_bare_schema();
+
+        let mut fragment_sessions = HashMap::new();
+        let mut fragment_sessions_with_row_addr = HashMap::new();
+
+        for ff in dataset.get_fragments().iter() {
+            let session = ff.open_session(&schema, false).await?;
+            fragment_sessions.insert(ff.id(), session);
+
+            let session = ff.open_session(&schema, true).await?;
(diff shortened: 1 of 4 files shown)


wojiaodoubao:
Hi @Xuanwo , could you help review this when you have time, thanks very much !

rust/lance/src/dataset/take.rs
```diff
@@ -126,9 +139,11 @@ pub async fn take(
 }
 
 /// Take rows by the internal ROW ids.
-async fn do_take_rows(
+pub async fn do_take_rows(
     mut builder: TakeBuilder,
     projection: Arc<ProjectionPlan>,
+    do_take: TakeFunction,
+    do_take_range: TakeRangeFunction,
```
Xuanwo, line 146: I think this is over-designed.

rust/lance/src/dataset/dataset_reader.rs
```diff
@@ -0,0 +1,434 @@
...
+/// This API works well for users making repeated requests over the same projection.
+pub struct DatasetReader {
+    dataset: Arc<Dataset>,
+    fragment_sessions: Arc<HashMap<usize, FragmentSession>>,
+    fragment_sessions_with_row_addr: Arc<HashMap<usize, FragmentSession>>,
+    projection: Arc<ProjectionPlan>,
+}
+
+impl DatasetReader {
+    pub async fn new(dataset: &Dataset, projection: impl Into<ProjectionRequest>) -> Result<Self> {
```
Xuanwo, line 29: What's the input we expect for `impl Into<ProjectionRequest>`?

rust/lance/src/dataset/dataset_reader.rs
```diff
@@ -0,0 +1,434 @@
...
+        let mut fragment_sessions = HashMap::new();
+        let mut fragment_sessions_with_row_addr = HashMap::new();
+
+        for ff in dataset.get_fragments().iter() {
+            let session = ff.open_session(&schema, false).await?;
+            fragment_sessions.insert(ff.id(), session);
+
+            let session = ff.open_session(&schema, true).await?;
+            fragment_sessions_with_row_addr.insert(ff.id(), session);
+        }
```
Xuanwo, line 43: Get all fragments in dataset and open them twice doesn't look like a good idea to me. Maybe we should start session only while needed?

wojiaodoubao:
> Thank you for working on this. I have a feeling that a new abstraction like `DatasetReader` might not be particularly useful.

@Xuanwo Thanks your comments !!! Your concern is that "single-row point lookup" might not be something users actually need, right? There was an earlier discussion(https://github.com/lancedb/lance/issues/3983) about Lance's single-row point lookup, during which Westone expressed a similar view: if the dataset is small, it should be queried in batches; if the dataset is large (and only one row can be queried at a time), then the overhead of data reading will dominate. Therefore, optimizing single-row point lookup might not seem important.
If I were designing my training task from scratch and chose Lance from the beginning, this perspective would indeed make sense. However, in reality, there are two scenarios that force me to rely on Lance's single-row point lookup.
1. My client has already built their training task (dataset & dataloader) based on some other format (usually LMDB). This is a complex system (including parallel execution, sample stitching, preloading, etc.) that cannot be restructured to support batch loading, leaving single-row queries as the only option for Lance.
2. In large language model training, the size of a sample and the number of tokens contained in a sample are both unpredictable. Batch loading could potentially cause memory overflow.

rust/lance/src/dataset/dataset_reader.rs
wojiaodoubao: This can either be a Lance schema or an SQL query (e.g., ```ProjectionRequest::from_sql(vec![("foo", "i"), ("bar", "i*2")])```). It is carried over from the dataset's take method.

```
    #[instrument(skip_all, fields(num_rows=row_indices.len()))]
    pub async fn take(
        &self,
        row_indices: &[u64],
        projection: impl Into<ProjectionRequest>,
    ) -> Result<RecordBatch> {
        take::take(self, row_indices, projection.into()).await
    }
```

rust/lance/src/dataset/dataset_reader.rs
wojiaodoubao: Your are right, let me try to fix it. I used to concern about potential locking issues, where a lock would be required each time a fragment session is accessed. Let me see if there's a way to resolve this.

wojiaodoubao:
The performance bench result is here: https://github.com/lancedb/lance/issues/4090#issuecomment-3166827283.

Comparing to take from dataset api, using dataset reader api benefits 70ms in 1000 single row takes (from 256ms to 186ms).

Issue #4412, opened 2025-08-08
Panic during BinaryMiniBlock decompression: TargetAlignmentGreaterAndInputNotAligned


Xuanwo:
Our users encountered a situation where they experience panic during BinaryMiniBlock decompression: `TargetAlignmentGreaterAndInputNotAligned` even in very simple cases.

I confirm that this can be reproduced on a GCS bucket, at least on main, 0.32.1.

This test is not reliably reproducible, failing approximately 40% of the time in the test loop.

```python
# SPDX-License-Identifier: Apache-2.0
# SPDX-FileCopyrightText: Copyright The Lance Authors
"""
Exact reproduction of user's test case for TargetAlignmentGreaterAndInputNotAligned issue
"""

import lance
import pyarrow as pa


def test_exact_reproduction():
    """
    Exact reproduction of user's test case
    """
    schema = pa.schema([
        pa.field("string_col", pa.string()),
    ])

    data = [
        {
            "string_col": "ABC",
        },
        {
            "string_col": "XYZ",
        }
    ]
    tbl = pa.Table.from_pylist(data, schema=schema)
    
    # Using the provided test credentials
    endpoint = "https://storage.googleapis.com"
    access_key = "x"
    secret_key = "x"
    region = "us-east1"
    
    # Use a fresh path for each test
    import uuid
    test_id = str(uuid.uuid4())[:8]
    uri = f"s3://x/test-alignment-{test_id}.lance"
    
    print(f"origin: {tbl.to_pydict()}")
    print(f"Attempting to write directly to {uri}")
    
    lance.write_dataset(
        tbl,
        uri,
        mode="overwrite",
        schema=schema,
        data_storage_version="2.1",
        storage_options={
            "endpoint_url": endpoint,
            "aws_access_key_id": access_key,
            "aws_secret_access_key": secret_key,
            "region": region,
            "aws_s3_addressing_style": "path"
        }
    )
    print(f"Successfully wrote dataset directly to bucket!")
    
    dataset = lance.dataset(
        uri,
        storage_options={
            "endpoint_url": endpoint,
            "aws_access_key_id": access_key,
            "aws_secret_access_key": secret_key,
            "region": region,
            "aws_s3_addressing_style": "path"
        }
    )
    tbl = dataset.to_table()
    print(f"Read back: {tbl.to_pydict()}")


if __name__ == "__main__":
    test_exact_reproduction()
```

Xuanwo:
[...]

PR #4413, opened 2025-08-08
fix: fix target alignment incorrect by using LanceBuffer borrow_to_typed_slice

Xuanwo:
Close https://github.com/lancedb/lance/issues/4412


rust/lance-encoding/src/encodings/physical/binary.rs
```diff
@@ -248,6 +248,8 @@ impl MiniBlockDecompressor for BinaryMiniBlockDecompressor {
             // offset and at least one value
             assert!(data.len() >= 16);
 
+            // Ensure buffer is aligned for u64 access
+            let data = data.align_with::<u64>();
             let offsets: &[u64] = try_cast_slice(&data)
                 .expect("casting buffer failed during BinaryMiniBlock decompression");
```
westonpace: Actually, we already have a method for this :smile: 

```
// You will need to make `data` mutable
let offsets = data.borrow_to_typed_slice::<u64>();
```

rust/lance-encoding/src/buffer.rs
```diff
@@ -164,6 +164,32 @@ impl LanceBuffer {
...
+        if ptr.align_offset(alignment) != 0 {
+            log::debug!(
+                "Buffer at {:p} is not {}-byte aligned for type {}, copying to aligned buffer",
+                ptr,
+                alignment,
+                std::any::type_name::<T>()
+            );
+            // Buffer is not aligned, need to copy
+            let aligned_vec = self.as_ref().to_vec();
+            Self::Owned(aligned_vec)
```
westonpace: In practice, this will probably work.  In theory, it doesn't have to.  You are creating `Vec<u8>` and so the alignment of the new vector is only _guaranteed_ to be wide enough to fit `u8`.  However, the underlying memory allocator typically has a pretty large default minimum alignment (e.g. malloc will always align to at least 8 or 16) so we'll get away with it.

rust/lance-encoding/src/encodings/physical/binary.rs
```diff
@@ -269,6 +271,8 @@ impl MiniBlockDecompressor for BinaryMiniBlockDecompressor {
             // offset and at least one value
             assert!(data.len() >= 8);
 
+            // Ensure buffer is aligned for u32 access
+            let data = data.align_with::<u32>();
```
westonpace: Same as above, let's use `borrow_to_typed_slice`.

westonpace (changes requested):
We can use `borrow_to_typed_slice`.  Though we can keep `align_with` if you'd like :+1:

rust/lance-encoding/src/encodings/physical/binary.rs
Xuanwo: Nice idea, changed!

westonpace (approved):

Issue #3855, opened 2025-05-19
Allow FTS indices to be used in filters


westonpace:
We currently allow NGRAM indices to be used to speed up `contains` filters.  NGRAM indices are limited to a very narrow range of tokenizers which should avoid false negatives from the index.

Inverted indices (FTS indices) on the other hand, could also be used to speed up `contains` filters, but might introduce false negatives (e.g. a query for `bakin` will not match a document with `baking` if stemming is turned on).  However, some users may know they are not going to experience these false negatives (e.g. query string is whole words) or just wants to accept the chance of false negatives to speed up performance.

We could introduce a UDF `contains_tokens` which is similar to `contains` but allows an FTS / Inverted index to be used.  This makes it possible for the user to opt-in to allowing these indices.

wojiaodoubao:
Hi @westonpace , I would like to know if you have any plans for this contains_tokens? If there isn't any plan recently, I can help try to implement it.

westonpace:
I am not planning on working on it anytime soon.  I would be happy for help!

A good first step might be to get a UDF named contains_tokens added to our session context.  It could fallback to the normal contains behavior.  Then we can work on the optimization step.

PR #4420, opened 2025-08-10
feat: add contains_tokens udf

wojiaodoubao:
First step of https://github.com/lancedb/lance/issues/3855

### DIFF
--- a/rust/lance-datafusion/src/udf.rs
+++ b/rust/lance-datafusion/src/udf.rs
@@ -1,0 +1,115 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright The Lance Authors
+
+//! Datafusion user defined functions
+
+use arrow_array::{ArrayRef, BooleanArray, StringArray};
+use arrow_schema::DataType;
+use datafusion::logical_expr::{create_udf, ScalarUDF, Volatility};
+use datafusion::prelude::SessionContext;
+use datafusion_functions::utils::make_scalar_function;
+use std::sync::{Arc, LazyLock};
+
+/// Register UDF functions to datafusion context.
+pub fn register_functions(ctx: &SessionContext) {
+    ctx.register_udf(CONTAINS_TOKENS_UDF.clone());
+}
+
+/// This method checks whether a string contains another string. It utilizes FTS (Full-Text Search)
+/// indexes, but due to the false negative characteristic of FTS, the results may have omissions.
+/// For example, "bakin" will not match documents containing "baking."
+/// If the query string is a whole word, or if you prioritize better performance, `contains_tokens`
+/// is the better choice. Otherwise, you can use the `contains` method to obtain accurate results.
+///
+/// Usage
+/// * Use `contains_tokens` in sql.
+/// ```rust,ignore
+/// let sql = "SELECT * FROM table WHERE contains_tokens(text_col, 'bakin')"
+/// let mut ds = Dataset::open(&ds_path).await?;
+/// let mut builder = ds.sql(&sql);
+/// let records = builder.clone().build().await?.into_batch_records().await?;
+/// ```
+fn contains_tokens() -> ScalarUDF {
+    let function = Arc::new(make_scalar_function(
+        |args: &[ArrayRef]| {
+            let column = args[0].as_any().downcast_ref::<StringArray>().ok_or(
+                datafusion::error::DataFusionError::Execution(
+                    "First argument of contains_tokens can't be cast to string".to_string(),
+                ),
+            )?;
+            let scalar_str = args[1].as_any().downcast_ref::<StringArray>().ok_or(
+                datafusion::error::DataFusionError::Execution(
+                    "Second argument of contains_tokens can't be cast to string".to_string(),
+                ),
+            )?;
+
+            let result = column
+                .iter()
+                .enumerate()
+                .map(|(i, column)| column.map(|value| value.contains(scalar_str.value(i))));
+
+            Ok(Arc::new(BooleanArray::from_iter(result)) as ArrayRef)
+        },
+        vec![],
+    ));
+
+    create_udf(
+        "contains_tokens",
+        vec![DataType::Utf8, DataType::Utf8],
+        DataType::Boolean,
+        Volatility::Immutable,
+        function,
+    )
+}
+
+static CONTAINS_TOKENS_UDF: LazyLock<ScalarUDF> = LazyLock::new(contains_tokens);
+
+#[cfg(test)]
+mod tests {
+    use crate::udf::CONTAINS_TOKENS_UDF;
+    use arrow_array::{Array, BooleanArray, StringArray};
+    use arrow_schema::{DataType, Field};
+    use datafusion::logical_expr::ScalarFunctionArgs;
+    use datafusion::physical_plan::ColumnarValue;
+    use std::sync::Arc;
+
+    #[tokio::test]
+    async fn test_contains_tokens() {
+        // Prepare arguments
+        let contains_tokens = CONTAINS_TOKENS_UDF.clone();
+        let text_col = Arc::new(StringArray::from(vec![
+            "a cat",
+            "lovely cat",
+            "white cat",
+            "catch up",
+            "fish",
+        ]));
+        let token = Arc::new(StringArray::from(vec!["cat", "cat", "cat", "cat", "cat"]));
+
+        let args = vec![ColumnarValue::Array(text_col), ColumnarValue::Array(token)];
+        let arg_fields = vec![
+            Arc::new(Field::new("text_col".to_string(), DataType::Utf8, false)),
+            Arc::new(Field::new("token".to_string(), DataType::Utf8, false)),
+        ];
+
+        let args = ScalarFunctionArgs {
+            args,
+            arg_fields,
+            number_rows: 5,
+            return_field: Arc::new(Field::new("res".to_string(), DataType::Boolean, false)),
+        };
+
+        // Invoke contains_tokens manually
+        let values = contains_tokens.invoke_with_args(args).unwrap();
+
+        if let ColumnarValue::Array(array) = values {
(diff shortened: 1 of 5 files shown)


wojiaodoubao:
Hi @westonpace , could you help review this when you have time, thanks very much !

rust/lance/src/dataset/sql.rs
```diff
@@ -74,9 +80,16 @@ impl SqlQueryBuilder {
                 row_addr,
             )),
         )?;
+
+        Self::register_functions(&ctx);
```
jackye1995: this only registers the function in a SQL query. I think we want to probably always register it in our session context.

rust/lance/src/dataset/sql.rs
```diff
@@ -131,13 +144,52 @@ impl SqlQuery {
     }
 }
 
+fn contains_tokens() -> ScalarUDF {
```
jackye1995: probably add some documentation explaining its usage and arguments?

rust/lance/src/dataset/sql.rs
```diff
@@ -131,13 +144,52 @@ impl SqlQuery {
...
+    let function = Arc::new(make_scalar_function(
+        |args: &[ArrayRef]| {
+            let column = args[0].as_any().downcast_ref::<StringArray>().ok_or(
+                datafusion::error::DataFusionError::Execution(
+                    "Function contains_tokens first argument must be a StringArray".to_string(),
```
jackye1995: saying it must be a string array does not really explain what is the actual input shape that should be provided.

rust/lance/src/dataset/sql.rs
westonpace: We have a process-wide shared context that gets initialized in `lance_datafusion::exec::get_session_context`.  We should probably move this registration logic there and use that context here.

rust/lance-index/src/scalar/inverted/index.rs
```diff
@@ -48,32 +74,6 @@ use snafu::location;
...
-    },
-    iter::PlainPostingListIterator,
-    query::*,
-    scorer::{idf, BM25Scorer, Scorer, B, K1},
-};
-use super::{
-    builder::{InnerBuilder, PositionRecorder},
-    encoding::compress_posting_list,
-    iter::CompressedPostingListIterator,
-};
```
westonpace: I'm going to assume these import changes are just moving things around (maybe some kind of automated formatting tool?)

rust/lance-index/src/scalar/inverted/index.rs
```diff
@@ -329,11 +329,10 @@ impl ScalarIndex for InvertedIndex {
...
     ) -> Result<SearchResult> {
-        let query = query.as_any().downcast_ref::<SargableQuery>().unwrap();
-        return Err(Error::invalid_input(
-            format!("unsupported query {:?} for inverted index", query),
-            location!(),
-        ));
+        // TODO: use fts for search
+        let _query = query.as_any().downcast_ref::<TextQuery>().unwrap();
+        let values: Vec<u64> = vec![];
+        Ok(SearchResult::AtLeast(RowIdTreeMap::from_iter(values)))
```
westonpace: This is maybe jumping ahead of ourselves.  Let's include this change in the future PR when we can actually test it?

westonpace (changes requested):
Some minor suggestions but the UDF part looks good.  I suppose we will need to document our UDFs somewhere as we start to build up a library but we can save that for a future PR.

Leaving in "request changes" until we can address the `get_session_context` suggestion.

westonpace:
Feel free to re-request review when ready, thanks!

rust/lance-index/src/scalar/inverted/index.rs
wojiaodoubao: Yes, they are automated formatted, let me fix them.

rust/lance-index/src/scalar/inverted/index.rs
wojiaodoubao: Ok, let me redo this.

I used to add a unit test in ```sql.rs test_sql_contains_tokens```, test using contains_tokens on a column with FTS index. Temporarily returning SearchResult::AtLeast allows the test to pass successfully.  The unit test is also jumping ahead, I'll redo it too.

rust/lance/src/dataset/sql.rs
wojiaodoubao: ok, let me fix it.

rust/lance/src/dataset/sql.rs
```diff
@@ -63,18 +63,20 @@ impl SqlQueryBuilder {
...
+            self.table_name.clone(),
             Arc::new(LanceTableProvider::new(
                 self.dataset.clone(),
                 row_id,
                 row_addr,
             )),
         )?;
         let df = ctx.sql(&self.sql).await?;
+
+        ctx.deregister_table(self.table_name)?;
```
wojiaodoubao, line 79: The session context got from ```get_session_context``` is cloned from a static session context, and they share  the registered tables. So we need to deregister it after query.

There is still a chance we might ran into 'register table' conflicts. Maybe we should use a new session context each time?

rust/lance/src/dataset/sql.rs
wojiaodoubao: Thanks ! The actual input should be string, so I change it to ```First argument of contains_tokens can't be cast to string```. What do you think?

rust/lance/src/dataset/sql.rs
```diff
@@ -63,18 +63,19 @@ impl SqlQueryBuilder {
     }
 
     pub async fn build(self) -> lance_core::Result<SqlQuery> {
-        let ctx = SessionContext::new();
+        let ctx = new_session_context(&LanceExecutionOptions::default());
```
wojiaodoubao: We need to use a new session context to avoid `register table conflicts', so here is ```new_session_context```, not ```get_session_context```.

wojiaodoubao:
Hi @westonpace @jackye1995 , this pr is ready for review now, please help when you have time, thanks very much !

rust/lance/src/dataset/sql.rs
wojiaodoubao: The ```get_session_context``` returns a clone of a static session context. They share the same registered tables. So if we register tables with the same name, it run into a 'register table conflicts error'.

westonpace: Ah, I see the problem.  Actually, it is a bit weird that we have to register the table on the session context (which is per-process).  We can get away with registering it on the state (which is per-query) if we do something like this...

```
    pub async fn build(self) -> lance_core::Result<SqlQuery> {
        let ctx = get_session_context(&LanceExecutionOptions::default());
        let row_id = self.with_row_id;
        let row_addr = self.with_row_addr;
        let state = ctx.state();
        let table_ref: TableReference = self.table_name.clone().into();
        let table = table_ref.table().to_owned();
        state.schema_for_ref(table_ref)?.register_table(
            table,
            Arc::new(LanceTableProvider::new(
                self.dataset.clone(),
                row_id,
                row_addr,
            )),
        )?;
        let plan = state.create_logical_plan(&self.sql).await?;
        SQLOptions::new().verify_plan(&plan)?;

        let df = DataFrame::new(state, plan);
        Ok(SqlQuery::new(df))
    }
```

This way we can use the default session context since we are no longer modifying it.

wojiaodoubao:
> One more suggestion. Sorry to be annoying but I think it's a good cleanup opportunity to keep a single session context. We have found that this helps quite a bit for performance (creating a session context can be costly) especially for small queries.

Thanks @westonpace for your comments! The single session context makes sense to me, your suggestion is definitely not annoying. I tried the 'state register table way' and still got the 'table already exists error'. I think it is because the 'registered tables' are still shared even using the session state.

I think the single session context can be considered as a separate issue. There are some potential solutions we can discuss in https://github.com/lancedb/lance/pull/4464#issue-3321474456. Once it is resolved, we can come back and continue pushing this PR forward.

rust/lance/src/dataset/sql.rs
jackye1995: @wojiaodoubao I was commenting in the other PR about we might want to rethink a bit about the whole SQL query feature in Lance given the issue you hit. But regardless I think the contains_tokens UDF is worth merging even outside the context of using it through the SQL interface. Shall we do that first in this PR?

rust/lance-datafusion/src/udf.rs
```diff
@@ -0,0 +1,115 @@
...
+///
+/// Usage
+/// * Use `contains_tokens` in sql.
+/// ```rust,ignore
+/// let sql = "SELECT * FROM table WHERE contains_tokens(text_col, 'bakin')"
+/// let mut ds = Dataset::open(&ds_path).await?;
+/// let mut builder = ds.sql(&sql);
+/// let records = builder.clone().build().await?.into_batch_records().await?;
+/// ```
+fn contains_tokens() -> ScalarUDF {
```
jackye1995, line 33: we should also talk about how to register the UDF in https://github.com/lancedb/lance/blob/main/docs/src/integrations/datafusion.md, so users can use that independently in DataFusion if they want.

rust/lance/src/dataset/sql.rs
wojiaodoubao: @jackye1995 Thanks, I change it back to ```new_session_context```.

jackye1995: oh I was saying that we can remove all the changes in `sql.rs`, so we can independently merge the addition of the UDF, and we use the other open PR to see what we can do for the dataset SQL experience.

wojiaodoubao: haha, I misunderstood. Fix it.

westonpace:
> I think the single session context can be considered as a separate issue. There are some potential solutions we can discuss in https://github.com/lancedb/lance/pull/4464#issue-3321474456. Once it is resolved, we can come back and continue pushing this PR forward.

Works for me

Issue #4423, opened 2025-08-11
OSS path prefix is duplicated when using OpenDAL-backed object store


qidian99:
### Summary
When using `oss://bucket/<prefix>/table.lance`, all object keys are written under `<prefix>/<prefix>/...` instead of a single `<prefix>/...`. 

The root cause that the same `<prefix>` is concatenated again has been located -- see below.

### Environment
- Scheme: `oss://`
- Example URI: `oss://bucket/user/test/lance_datasets/append.lance`
- Write mode: APPEND

### How to reproduce
1. Create a dataset at `oss://bucket/user/test/lance_datasets/append.lance`.
2. Inspect object keys written to the bucket.

- Expected: keys under `user/test/lance_datasets/append.lance/...`
- Actual: keys under `user/test/lance_datasets/append.lance/user/test/lance_datasets/append.lance/...`

### Root cause
- `ObjectStore::from_uri_and_params`
- `ObjectStoreRegistry::get_store` 
- The OSS provider configures OpenDAL with `root` equal to the URL path prefix:
```52:54:rust/lance-io/src/object_store/providers/oss.rs
if !prefix.is_empty() {
    config_map.insert("root".to_string(), format!("/{}", prefix));
}
```
- Later, `from_uri_and_params` derives the dataset base path from the URL again via `provider.extract_path(&url)`:
```341:347:rust/lance-io/src/object_store.rs
let url = uri_to_url(uri)?;
let store = registry.get_store(url.clone(), params).await?;
// We know the scheme is valid if we got a store back.
let provider = registry.get_provider(url.scheme()).expect_ok()?;
let path = provider.extract_path(&url);

Ok((store, path))
```
- in dataset/fragment/write.rs, the fragments are written using 
```rust
do_write_fragments(object_store, &base_path, ...)
```

Thus duplicating the prefix as in `oss://bucket/<prefix>/<prefix>/table.lance`

### Impact
All OSS URIs that include a non-empty path after the bucket will write/read under a doubled prefix, causing incorrect directory layout and lookup failures.

### Proposed fix
 - Do not set OpenDAL `root` in `OssStoreProvider`. Leave `root` empty so keys are built solely from the URL-derived base path (`extract_path(url)`)

Xuanwo:
Thank you for the report, I'm going to fix it.

PR #4424, opened 2025-08-11
fix: wrong OSS path append

adrian-wang:
When using OSS as output path, if we write to oss://bucket1/a/b/c, the data will be write to oss://bucket1/a/b/c/a/b/c, the root path should be `/` instead.

Close https://github.com/lancedb/lance/issues/4423

### DIFF
--- a/rust/lance-io/src/object_store/providers/oss.rs
+++ b/rust/lance-io/src/object_store/providers/oss.rs
@@ -50,7 +50,7 @@ impl ObjectStoreProvider for OssStoreProvider {
         config_map.insert("bucket".to_string(), bucket);
 
         if !prefix.is_empty() {
-            config_map.insert("root".to_string(), format!("/{}", prefix));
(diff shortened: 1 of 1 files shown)


rust/lance-io/src/object_store/providers/oss.rs
```diff
@@ -50,7 +50,7 @@ impl ObjectStoreProvider for OssStoreProvider {
         config_map.insert("bucket".to_string(), bucket);
 
         if !prefix.is_empty() {
-            config_map.insert("root".to_string(), format!("/{}", prefix));
+            config_map.insert("root".to_string(), "/".to_string());
```
chunshao90, line 53: May be just remove line 53 & line32 will be simple.

Xuanwo: Either way looks fine to me. The current approach works well because it ensures `root` is overwritten.

PR #4428, opened 2025-08-11
fix: remove explain in sql

wojiaodoubao:
We can use something like ```EXPLAIN SELECT * FROM dataset``` or ```EXPLAIN ANALYZE SELECT * FROM dataset``` instead. I think it is more natural. So I'd like to remove `explain_plan` function.


wojiaodoubao:
Hi @yanghua , could you help review this when you have time, thanks very much.

yanghua:
Can you add some tests to show the equivalent ability?

wojiaodoubao:
> Can you add some tests to show the equivalent ability?

Sure, let me fix it.

python/python/tests/test_dataset.py
```diff
@@ -3757,13 +3757,32 @@ def test_metadata_cache_size(tmp_path):
...
 
 
+def test_dataset_sql_explain_analyze(tmp_path: Path):
+    table = pa.table({"id": [1, 2, 3], "value": ["a", "b", "c"]})
+    ds = lance.write_dataset(table, tmp_path / "test")
+
+    query = ds.sql("EXPLAIN SELECT * FROM test WHERE id > 1").table_name("test").build()
+    batch_records = query.to_batch_records()
+    explain_plan = pa.Table.from_batches(batch_records).to_pandas().to_string()
+    assert any(k in explain_plan for k in ("Filter", "full_filter", "refine_filter")), (
```
jackye1995: can we just compare against the whole plan string? Something like https://github.com/lancedb/lance/blob/main/rust/lance/src/dataset/write/merge_insert.rs#L3432-L3446

python/python/tests/test_dataset.py
```diff
@@ -3757,13 +3757,32 @@ def test_metadata_cache_size(tmp_path):
...
+        explain_plan
+    )
+
+    query = (
```
jackye1995: I think let's have 2 separated tests for explain vs explain analyze in this case

wojiaodoubao:
Hi @jackye1995 @yanghua , I have updated this pr. Please review when you have time, thanks !!!

PR #4434, opened 2025-08-11
refactor: refactor encoding proto for 2.1 to cleanup and future-proof

westonpace:
This makes a full scan through the 2.1 encoding schemes and cleans up the protobuf.  This removes a number of schemes that were 2.0 only and helps cleanup the intent.  In addition, many schemes are hard-coding an inner encoding of flat today.  We might want to do something different in the future.  This changes the protos so the hard-coding is done in code and not protobuf.  This should help future-proof the encoding scheme for minor changes in 2.2 or 2.1.2 or whatever is next.


protos/encodings.proto
Xuanwo, line 1: Do you think it's a good idea to make this `encodings_v2_0.proto` and  `encodings_v2_1.proto`?

protos/encodings-df.proto
```diff
@@ -1,31 +0,0 @@
...
-
-import "encodings.proto";
-
-// ZoneMaps are a way to wrap any leaf array with a set of zone maps that
-// can be used to apply pushdown filtering.
-//
-// A "zone map" is the min/max/null_count of a set of rows.  This can be
-// used to quickly filter out zones which are not included in simple predicates
-// like 'x = 5' or 'x > 10'.
-message ZoneMaps {
```
Xuanwo, line 17: We will redesign this?

protos/encodings21.proto
```diff
@@ -0,0 +1,444 @@
...
+  CompressiveEncoding values = 1;
+}
+
+// An encoding that compresses a data block into buffers
+message CompressiveEncoding {
+    oneof compression {
+        Flat flat = 1;
+        Variable variable = 2;
+        Constant constant = 3;
+        OutOfLineBitpacking out_of_line_bitpacking = 4;
```
Xuanwo: Many of our code place are using `bitpack` instead of `bitpacking`. Maybe this is the time to unify them?

protos/encodings21.proto
```diff
@@ -0,0 +1,444 @@
...
+        InlineBitpacking inline_bitpacking = 5;
+        Fsst fsst = 6;
+        Dictionary dictionary = 7;
+        Rle rle = 8;
+        ByteStreamSplit byte_stream_split = 9;
+        WrappedGeneral wrapped = 10;
```
Xuanwo: I actually like the idea of `General`.

protos/encodings21.proto
```diff
@@ -0,0 +1,444 @@
...
+// Note: Wrapped compression is somewhat unique at the moment as it is applied to the
+// output of the inner encoding and not the input like all other compressive encodings.
+//
+// Note: General compression can usually be applied in two spots.  We can apply
+// it to individual buffers or we can apply it here, to the entire array.
+//
+// For example, let's say we are storing mini-blocks of strings and we are using
+// FSST and bitpacking the offsets.  We have something like this...
+//
+// WRAPPED(†3) -> FSST -> VARIABLE -(offsets)-> INLINE_BITPACKING -(data)-> FLAT -> BUFFER (†1)
```
Xuanwo: Cool!

Xuanwo (commented):
Looks good to me, thank you for this!

protos/encodings21.proto
```diff
@@ -0,0 +1,444 @@
...
+// In addition, leaf encodings may output "buffers".  These are fully compressed buffers of data that
+// are stored in the page and no longer compressed.
+
+// Compression applied to a single buffer of data
+//
+// A buffer is the leaf of the compression tree.  Unlike data blocks, which can
+// be further compressed with a variety of techniques, a buffer cannot be understood
+// in any particular way.
+//
+// A general compression scheme may be applied to a buffer.  This is something like
```
mapleFU: Why this is not something like a enum?

protos/encodings-df.proto
westonpace: Already in progress :smile: The new approach treats zone map as an index (same as our other indexes) and not as an artifact of the file format.

https://github.com/lancedb/lance/pull/4244

protos/encodings.proto
westonpace: Yes, I get nervous anytime I change 2.0 files (don't want to break the existing) but I think this change should be safe enough and I agree much clearer.

protos/encodings21.proto
westonpace: I'm on the fence, I think I could go either way.  I'll prototype an enum and see how I like it.

westonpace: Ok, I switched to enum in https://github.com/lancedb/lance/pull/4434/commits/e134f6c602a14276db7220f5d8e6ee8114d5eab8

protos/encodings21.proto
westonpace: Switched to `General`

protos/encodings21.proto
westonpace: I changed `bitpack.rs` into `bitpacking.rs`.  Is there another place you think should change?

protos/encodings.proto
westonpace: Ok, I've renamed the files

protos/encodings21.proto
Xuanwo: We have some structs and API like `BitpackParams` and `bitpack_params`. But I don't have other concerns to this.

Xuanwo:
Question: How can we ensure compatibility? For example, some users are writing tables in `2.1`, will they be affected?

westonpace:
> Question: How can we ensure compatibility? For example, some users are writing tables in 2.1, will they be affected?

Yes, definitely.  That's why 2.1 is unstable at the moment.  Hopefully no one is writing permanent data in 2.1.  We do emit a warning:

> You have requested an unstable format version.  Files written with this format version may not be readable in the future!  This is a development feature and should only be used for experimentation and never for production data.

Issue #4390, opened 2025-08-05
Refractor CreateIndex operation definition in python to be consistent with rust


chenghao-guo:
While rust is using the 
```
    CreateIndex {
        /// The new secondary indices that are being added
        new_indices: Vec<Index>,
        /// The indices that have been modified.
        removed_indices: Vec<Index>,
    },

```
as CreateIndex operation definition.

Python is using this definition
```
    class CreateIndex(BaseOperation):
        uuid: str
        name: str
        fields: List[int]
        dataset_version: int
        fragment_ids: Set[int]
        index_version: int
        created_at: Optional[datetime] = None
```

 I don't know why it's different from rust version. Is there any history background?

I am in the progress of another issue and will commit index manually from python.  Just notice the difference, so would like to check if any background.  Was this intentional or legacy context? Not sure if @BubbleCal  @yanghua know some background?

My plan is to refractor the python operation possibly like this
```
    class CreateIndex(BaseOperation):
        new_indices: List[Index]
        removed_indices: List[Index] 
```
and modify the extract_bound/into_pyobject in transaction.rs pyo3. So the python interface can be consistent with rust.

wkalt:
@chenghao-guo we just discussed your plan, it sounds fine to us. Thanks.

chenghao-guo:
> [@chenghao-guo](https://github.com/chenghao-guo) we just discussed your plan, it sounds fine to us. Thanks.

Hi @wkalt , thank you for your confirmation. I have opened PR https://github.com/lancedb/lance/pull/4438 for this. Would you mind reviewing it when convenient?

PR #4438, opened 2025-08-12
feat(python)!: unify Python CreateIndex api with Rust implementation

chenghao-guo:
BREAKING CHANGE: the `CreateIndex` operation in Python now takes a list of indices instead of representing a single index

closes https://github.com/lancedb/lance/issues/4390

This commit refactors the Python CreateIndex class to match the Rust implementation structure, enabling consistent API across languages.

- Updated CreateIndex class to use new_indices and removed_indices fields
- Enhanced Index class from TypedDict to dataclass
- Simplified pyo3 bindings
- Updated test cases to use new API structure

### DIFF
--- a/python/python/tests/test_commit_index.py
+++ b/python/python/tests/test_commit_index.py
@@ -41,7 +41,17 @@ def dataset_with_index(test_table, tmp_path):
     return dataset
 
 
+def _get_field_id_by_name(lance_schema, field_name):
+    fields = lance_schema.fields()
+    for field in fields:
+        if field.name() == field_name:
+            return field.id()
+    return None
+
+
 def test_commit_index(dataset_with_index, test_table, tmp_path):
+    from lance.dataset import Index
+
     index_id = dataset_with_index.list_indices()[0]["uuid"]
 
     # Create a new dataset without index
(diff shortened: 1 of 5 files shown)


chenghao-guo:
cc @wkalt @yanghua  for review

wjones127:
FYI @chenghao-guo . It's important we get the PR title correct here. This is a breaking change to the API, so it needs to be labelled like that. And anything labelled `refactor` should only move around internals, it shouldn't affect the API. This should be `feat!` instead.

python/python/tests/test_commit_index.py
```diff
@@ -56,13 +58,22 @@ def test_commit_index(dataset_with_index, test_table, tmp_path):
 
     # Commit the index to dataset_without_index
     field_idx = dataset_without_index.schema.get_field_index("meta")
```
wjones127: FYI this is wrong, we want the field id, not the field index. They are often the same, but not always (especially if you have nested data). Not something we need to fix now, but what you to be aware of this.

python/src/transaction.rs
```diff
@@ -19,6 +19,79 @@ use uuid::Uuid;
...
+        let index_version = ob.getattr("index_version")?.extract()?;
+        let fragment_ids = ob.getattr("fragment_ids")?;
+        let created_at = ob.getattr("created_at")?.extract()?;
+
+        let fragment_ids_ref: &Bound<'_, PySet> = fragment_ids.downcast()?;
+        let fragment_ids = fragment_ids_ref
+            .into_iter()
+            .map(|id| id.extract())
+            .collect::<PyResult<Vec<u32>>>()?;
+        let fragment_bitmap = Some(fragment_ids.into_iter().collect());
```
wjones127: There's no reason to allocate twice for this, right?
```suggestion
            .collect::<PyResult<RoaringBitmap>>()?;
        let fragment_bitmap = Some(fragment_ids);
```

chenghao-guo:
> FYI @chenghao-guo . It's important we get the PR title correct here. This is a breaking change to the API, so it needs to be labelled like that. And anything labelled `refactor` should only move around internals, it shouldn't affect the API. This should be `feat!` instead.

Hi Will, thank you so much for the review and kind reminder! I'll carefully review the guidelines regarding the breaking changes,  and  implement adjustments per your suggestions.

python/python/tests/test_commit_index.py
chenghao-guo: Got it, thanks.  Updated per suggestions.

python/src/transaction.rs
chenghao-guo: Got it, thanks. Updated per suggestions.

PR #4441, opened 2025-08-12
feat(java): introduce delete in transaction

fangbo:
Issue: https://github.com/lancedb/lance/issues/4332

### DIFF
--- a/java/core/src/main/java/com/lancedb/lance/operation/Delete.java
+++ b/java/core/src/main/java/com/lancedb/lance/operation/Delete.java
@@ -1,0 +1,102 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.lancedb.lance.operation;
+
+import com.lancedb.lance.FragmentMetadata;
+
+import java.util.Collections;
+import java.util.List;
+import java.util.Objects;
+import java.util.Optional;
+
+public class Delete implements Operation {
+  private List<FragmentMetadata> updatedFragments;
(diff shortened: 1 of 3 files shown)


java/core/src/main/java/com/lancedb/lance/operation/Delete.java
```diff
@@ -0,0 +1,102 @@
...
+
+  public Optional predicate() {
+    return Optional.ofNullable(predicate);
+  }
+
+  @Override
+  public String name() {
+    return "Delete";
+  }
+
```
majin1102, line 55: I think we can override equals() for readTransaction testcase

java/core/src/test/java/com/lancedb/lance/TransactionTest.java
```diff
@@ -136,6 +138,50 @@ void testAppend(@TempDir Path tempDir) {
...
+      Transaction delete =
+          dataset
+              .newTransactionBuilder()
+              .operation(
+                  Delete.builder().deletedFragmentIds(deletedFragmentIds).predicate("1=1").build())
+              .build();
+      try (Dataset dataset = delete.commit()) {
+        Transaction txn = dataset.readTransaction().get();
+        Delete execDelete = (Delete) txn.operation();
+        assertEquals(deletedFragmentIds, execDelete.deletedFragmentIds());
```
majin1102: I think we can just apply assertEquals() on Delete operation if we add equals() for Delete just like other operations

majin1102 (changes requested):
Thanks for this contribution. I left some comments.

Please take a look

java/core/src/main/java/com/lancedb/lance/operation/Delete.java
```diff
@@ -0,0 +1,102 @@
...
+  public String toString() {
+    return String.format(
+        "Delete{updatedFragments=%s, deletedFragmentIds=%s, predicate=%s}",
+        updatedFragments, deletedFragmentIds, predicate);
+  }
+
+  @Override
+  public boolean equals(Object o) {
```
fangbo, line 63: @majin1102 This is equals method

java/core/src/test/java/com/lancedb/lance/TransactionTest.java
fangbo: > I think we can just apply assertEquals() on Delete operation if we add equals() for Delete just like other operations

Fix it

fangbo:
@jackye1995 This PR is ready to be reviewed. Could you please review this ?

PR #4442, opened 2025-08-12
fix: none ASCII path issue

### LINKED ISSUE #4021: Object store path contains Chinese can not decode successfully
```rust
        let (store, path) = ObjectStore::from_uri("gs://bucket/你好/foo.lance")
            .await
            .unwrap();
        assert_eq!(store.scheme, "gs");
        println!("{}", path.to_string());
        println!("{}", Path::parse(path.to_string()).unwrap().as_ref());
        println!("{}", Path::from_url_path(path.to_string()).unwrap().as_ref());
        println!("{}", Path::from_url_path(Path::from_url_path(path.to_string()).unwrap().as_ref()).unwrap());
        assert_eq!(Path::from_url_path("%E4%BD%A0%E5%A5%BD/foo.lance").unwrap().as_ref(), "你好/foo.lance");

        let a = Path::from_url_path("%E4%BD%A0%E5%A5%BD/foo.lance").unwrap();
        assert_eq!(a.as_ref(), "你好/foo.lance");
```

The keyword `你好` is Chinese.

The four prints listed below:

```
%25E4%25BD%25A0%25E5%25A5%25BD/foo.lance
%25E4%25BD%25A0%25E5%25A5%25BD/foo.lance
%E4%BD%A0%E5%A5%BD/foo.lance
你好/foo.lance
```

It seems to need two layers, `Path::from_url_path()` can decode Chinese.

When an object store path contains Chinese, it throws an exception, and the uri looks like this: `%25E4%25BD%25A0%25E5%25A5%25BD/foo.lance`

majin1102:
Close #4021

I encountered this issue in another PR: #4257. In that PR I need to persist a dataset url as String and use the String to re-construct a dataset path. I've stuck in a '~' encoding issue for some time(different with the original Chinese character issue, but I think the same encoding problem), which drives me to figure out the url/path encoding things.

Here's the comparision:

- Path::from()will encode both the non ASCII strings like Chinese characters and the reserved ASCII characters like ~
- Url::parse() will encode only non ASCII strings like Chinese characters
- Path::parse() will encode **nothing**
- Path::from_url_path will decode from a path

I think the key is If we want to generate a path from a url, we should assume the encoding done or not(or let's say the url is invalid or not, not encoded means invalid). Most cases:
1. We get a url by Url::parse(), which has encoded the non ASCII strings(if the string has been encoded, it won't encode again)
2. If we use a local path to generate a url, the internal string has been encoded.

If we assume the url is trusted in encoding, then we should always use Path::parse(). I think this is what we should do. If the path is invalid, meaning the url or the original string should be encoded first. This will avoid double encoding and unnecessary decoding in the Path::from_url_path case.

Note:
1. Use Url::parse and Path::parse will **NOT** encode the reserved ASCII characters like ~,?,&, for now it works well. The object_store path could handle the reserved ASCII characters correctly(refer to aws.rs testsutes).
2. For % in original paths(like 'aaa/bbb/cc%cc'), the % will **NOT** be encoded. We need to use urlencoding or other libs encode the url/path string before it is used. （seems quite reasonable to me)
3. Local paths seem more complicated. Modern OS supports non ASCII paths, but Windows and Linux/Mac seems different on the path processing. I think keeping the principle would be fine.

Please take a look when you have time @jackye1995 @Xuanwo @ddupg  cc @wjones127 @yanghua

### DIFF
--- a/python/python/lance/dataset.py
+++ b/python/python/lance/dataset.py
@@ -577,7 +577,7 @@ class LanceDataset(pa.dataset.Dataset):
             All columns are fetched if None or unspecified.
         filter: pa.compute.Expression or str
             Expression or str that is a valid SQL where clause. See
-            `Lance filter pushdown <https://lancedb.github.io/lance/introduction/read_and_write.html#filter-push-down>`_
+            `Lance filter pushdown <https://lancedb.github.io/lance/guide/read_and_write/#filter-push-down>`_
             for valid SQL expressions.
         limit: int, default None
             Fetch up to this many rows. All rows if None or unspecified.
@@ -828,7 +828,7 @@ class LanceDataset(pa.dataset.Dataset):
             All columns are fetched if None or unspecified.
         filter : pa.compute.Expression or str
             Expression or str that is a valid SQL where clause. See
-            `Lance filter pushdown <https://lancedb.github.io/lance/introduction/read_and_write.html#filter-push-down>`_
+            `Lance filter pushdown <https://lancedb.github.io/lance/guide/read_and_write/#filter-push-down>`_
             for valid SQL expressions.
         limit: int, default None
             Fetch up to this many rows. All rows if None or unspecified.
@@ -1843,6 +1843,7 @@ class LanceDataset(pa.dataset.Dataset):
         name: Optional[str] = None,
         *,
         replace: bool = True,
+        train: bool = True,
         **kwargs,
     ):
         """Create a scalar index on a column.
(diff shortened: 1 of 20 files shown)


rust/lance-io/src/object_store/providers/local.rs
```diff
@@ -56,11 +65,15 @@ mod tests {
             ("file:///usr/local/bin", "usr/local/bin"),
             ("file-object-store:///path/to/file", "path/to/file"),
             ("file:///path/to/foo/../bar", "path/to/bar"),
+            // for no ASCII string tests
+            ("file:///path/to/foo测试/../bar", "path/to/bar"),
+            ("file:///path/to/foo~2/../bar", "path/to/bar"),
```
jackye1995: I think we should remove the .. in the test path, that is skipping the test of the exact path with non ASCII characters

majin1102: Yeah, thanks for reminding

PR #4445, opened 2025-08-12
feat: make _take_rows public

westonpace:


westonpace:
Previously we've waited for stable row ids but I wonder if there is any harm in just doing it now?  This is lance, not lancedb, so we have some expectation for advanced users.  Should we improve the documentation or make it so the method can work with row addresses or dataset offsets as well?

PR #4450, opened 2025-08-13
feat(rust): support update stable row id for merge insert without index

yanghua:
Part of #2802


yanghua:
This PR is not ready for review. Just for verifying CI.

yanghua:
Hi @wjones127 can you take a look when you have time?

rust/lance/src/dataset/write/merge_insert/exec/write.rs
```diff
@@ -167,8 +179,196 @@ impl FullSchemaMergeInsertExec {
...
+        Ok(Box::pin(RecordBatchStreamAdapter::new(
+            output_schema,
+            stream,
+        )))
+    }
+
+    /// Two-phase streaming implementation for stable row ID scenarios
+    ///
+    /// This method ensures that all update operations are processed before insert operations
+    /// while avoiding full buffering in memory. It works in two phases:
```
wjones127: > while avoiding full buffering in memory

I don't see how it avoids this. The code appears to buffer all data into memory.

I would instead re-implement this as:

```rust
fn split_updates_and_inserts(
    input_stream: SendableRecordBatchStream,
    merge_state: Arc<Mutex<MergeState>>,
) -> DFResult<(SendableRecordBatchStream, SendableRecordBatchStream)>;
```

For each batch you receive in the input stream, split it immediately and send each split batch to their respective stream (through a channel).

rust/lance/src/dataset/write/merge_insert/exec/write.rs
```diff
@@ -225,92 +424,79 @@ impl FullSchemaMergeInsertExec {
...
-            let filtered_batch = arrow_select::take::take_record_batch(&batch, &indices)?;
+        Ok((row_addr_array, action_array))
+    }
 
-            // Project only the data columns
-            let output_columns: Vec<_> = data_column_indices
+    /// Create filtered batch from selected rows
+    fn create_filtered_batch(
+        batch: &RecordBatch,
+        keep_rows: Vec<usize>,
```
wjones127: Maybe just make this `Vec<u32>` so you don't have to translate it later?

rust/lance/src/dataset/write/merge_insert.rs
```diff
@@ -2209,6 +2352,137 @@ mod tests {
...
+        #[values(true, false)] enable_move_stable_row_ids: bool,
+    ) {
+        test_stable_row_ids_helper(
+            version,
+            enable_move_stable_row_ids,
+            &[4, 5, 6],
+            &[],
+            &[4, 5, 6, 7, 8, 9],
+            &[3, 3, 3],
+            |ds| {
```
wjones127: This signature is impossible to read. Refactor as builder?

rust/lance/src/dataset/write/merge_insert.rs
```diff
@@ -1986,50 +2030,152 @@ mod tests {
...
+        expected_right_keys: &[u32],
+        expected_stats: &[u64],
+        job_builder: impl FnOnce(Arc<Dataset>) -> MergeInsertJob,
+    ) {
+        let schema = create_test_schema();
+        let batch = create_initial_batch(schema.clone());
+        let new_batch = create_new_batch(schema.clone());
+
+        let test_dir = tempdir().unwrap();
+        let test_uri = test_dir.path().to_str().unwrap();
```
wjones127: When possible, try to do tests in `memory://` rather than writing to disk.

rust/lance/src/dataset/write/merge_insert/exec/write.rs
wjones127: You can write the updated data and new data in parallel.

rust/lance-table/src/rowids.rs
```diff
@@ -570,6 +568,27 @@ impl RowIdSeqSlice<'_> {
 pub fn rechunk_sequences(
     sequences: impl IntoIterator<Item = RowIdSequence>,
     chunk_sizes: impl IntoIterator<Item = u64>,
+) -> Result<Vec<RowIdSequence>> {
```
jackye1995: why not just update this signature? I don't think it has any actual public dependency although this is a public function. Any concerns? @wjones127

rust/lance/src/dataset/write/merge_insert.rs
```diff
@@ -1986,50 +2030,152 @@ mod tests {
...
                 Arc::new(UInt32Array::from(vec![4, 5, 6, 7, 8, 9])),
                 Arc::new(UInt32Array::from(vec![2, 2, 2, 2, 2, 2])),
                 Arc::new(StringArray::from(vec!["A", "B", "C", "A", "B", "C"])),
             ],
         )
-        .unwrap();
+        .unwrap()
+    }
+
+    async fn create_test_dataset(
```
jackye1995, line 2071: I think we can just use `lance_datagen::gen_batch()`?

rust/lance/src/dataset/write/merge_insert/exec/write.rs
yanghua: > I don't see how it avoids this. The code appears to buffer all data into memory.

Yes, the comment is old and wrong. Have adopted your suggestion.

yanghua:
@wjones127 @jackye1995 have addressed your comments.

rust/lance/src/dataset/write/merge_insert/exec/write.rs
```diff
@@ -33,23 +34,30 @@ use crate::{
...
 /// Shared state for merge insert operations to simplify lock management
 struct MergeState {
     /// Row addresses that need to be deleted, due to a row update or delete action
     delete_row_addrs: RoaringTreemap,
+    /// Shared collection to capture row ids that need to be updated
+    updating_row_ids: Arc<Mutex<CapturedRowIds>>,
     /// Merge operation metrics
     metrics: MergeInsertMetrics,
+    /// Whether to enable stable row ids
+    enable_stable_row_ids: bool,
```
wjones127: Merge insert doesn't enable stable row ids. That's set when the dataset is created. So we should just call this `stable_row_ids`.
```suggestion
    /// Whether the dataset uses stable row ids.
    stable_row_ids: bool,
```

rust/lance/src/dataset/write/merge_insert/exec/write.rs
```diff
@@ -33,23 +34,30 @@ use crate::{
...
 }
 
 impl MergeState {
-    fn new(metrics: MergeInsertMetrics) -> Self {
+    fn new(metrics: MergeInsertMetrics, enable_stable_row_ids: bool) -> Self {
         Self {
             delete_row_addrs: RoaringTreemap::new(),
+            updating_row_ids: Arc::new(Mutex::new(CapturedRowIds::new(enable_stable_row_ids))),
             metrics,
+            enable_stable_row_ids,
```
wjones127:
```suggestion
    fn new(metrics: MergeInsertMetrics, stable_row_ids: bool) -> Self {
Self {
delete_row_addrs: RoaringTreemap::new(),
            updating_row_ids: Arc::new(Mutex::new(CapturedRowIds::new(stable_row_ids))),
metrics,
            stable_row_ids,
```

rust/lance/src/dataset/write/merge_insert/exec/write.rs
```diff
@@ -75,6 +83,10 @@ impl MergeState {
                 if !row_addr_array.is_null(row_idx) {
                     let row_addr = row_addr_array.value(row_idx);
                     self.delete_row_addrs.insert(row_addr);
+
+                    if self.enable_stable_row_ids {
```
wjones127:
```suggestion
                    if self.stable_row_ids {
```

rust/lance/src/dataset/write/merge_insert/exec/write.rs
```diff
@@ -75,6 +83,10 @@ impl MergeState {
...
+                        self.updating_row_ids.lock().unwrap().capture(&[row_addr])?;
```
wjones127: Shouldn't this be capturing row id, not row address?

rust/lance/src/dataset/write/merge_insert.rs
```diff
@@ -1984,52 +2029,198 @@ mod tests {
...
+            .col("key", array::step_custom::<UInt32Type>(1, 1))
+            .col("value", array::fill::<UInt32Type>(1u32))
+            .col(
+                "filterme",
+                array::cycle_utf8_literals(&["A", "B", "A", "A", "B", "A"]),
+            )
+            .into_dataset_with_params(
                 test_uri,
-                Some(WriteParams::with_storage_version(version)),
+                FragmentCount(1),
```
wjones127: In order to catch bugs with stable row id, it's important you test with multiple fragments. If you have only 1 fragment, then `_rowaddr == _rowid` and you won't detect bugs that result from confusing the two.

rust/lance/src/dataset/write/merge_insert/exec/write.rs
```diff
@@ -167,8 +179,118 @@ impl FullSchemaMergeInsertExec {
...
+            self.prepare_stream_schema(input_stream.schema())?;
 
+        let output_schema_clone = output_schema.clone();
+        let stream = input_stream.map(move |batch_result| -> DFResult<RecordBatch> {
+            let batch = batch_result?;
+            let (row_addr_array, action_array) =
+                Self::extract_control_arrays(&batch, rowaddr_idx, action_idx)?;
+
+            // Process each row using the shared state
+            let mut keep_rows: Vec<u32> = Vec::new();
```
wjones127: For the common case of upsert, I would pre-size this vec:


```suggestion
            let mut keep_rows: Vec<u32> = Vec::with_capacity(batch.num_rows());
```

rust/lance/src/dataset/write/merge_insert.rs
yanghua: Thanks for reminding me. Yes, you are right. It really has a bug. I have fixed and added two test cases:

* do multiple merge inserts on the same keys;
* call update and merge insert APIs on the same keys;

wjones127 (approved):
Looks good. Thanks for adding those new tests with multiple operations.

PR #4460, opened 2025-08-14
refactor: remove owned LanceBuffer to eliminate runtime try_clone checks

Xuanwo:
This PR removes the owned LanceBuffer to eliminate runtime `try_clone` checks.

We observed numerous instances in recent user environments where buffers were reused but not properly aligned, leading to frequent occurrences of owned LanceBuffers. Unfortunately, many `try_clone` checks went unverified, and we had been assuming they were borrowed buffers.

Rather than attempting to fix all the unchecked owned buffers, we reconsidered whether retaining these owned buffers is worthwhile. In practice, they have never provided tangible benefits in real-world use cases.

Therefore, we have decided to remove the owned LanceBuffer concept entirely, eliminating the need for runtime `try_clone` checks. This change simplifies the codebase and makes it significantly easier to maintain.

cc @westonpace, should this PR be marked as breaking change? `LanceBuffer` seems only used in `lance-encoding`.

### DIFF
--- a/rust/lance-encoding/src/buffer.rs
+++ b/rust/lance-encoding/src/buffer.rs
@@ -9,7 +9,7 @@ use arrow_buffer::{ArrowNativeType, Buffer, MutableBuffer, ScalarBuffer};
 use lance_core::{utils::bit::is_pwr_two, Error, Result};
 use snafu::location;
 
-/// A copy-on-write byte buffer
+/// A copy-on-write byte buffer.
 ///
 /// It wraps arrow_buffer::Buffer which provides:
 /// - Cheap cloning (reference counted)
--- a/rust/lance-encoding/src/buffer.rs
+++ b/rust/lance-encoding/src/buffer.rs
@@ -6,98 +6,45 @@
 use std::{ops::Deref, panic::RefUnwindSafe, ptr::NonNull, sync::Arc};
 
 use arrow_buffer::{ArrowNativeType, Buffer, MutableBuffer, ScalarBuffer};
-use itertools::Either;
-use snafu::location;
 
-use lance_core::{utils::bit::is_pwr_two, Error, Result};
+use lance_core::{utils::bit::is_pwr_two, Result};
 
-/// A copy-on-write byte buffer
+/// A copy-on-write byte buffer backed by Arrow Buffer
 ///
-/// It can be created from read-only buffers (e.g. bytes::Bytes or arrow_buffer::Buffer), e.g. "borrowed"
-/// or from writeable buffers (e.g. Vec<u8>), e.g. "owned"
+/// It wraps arrow_buffer::Buffer which provides:
+/// - Cheap cloning (reference counted)
+/// - Zero-copy slicing
+/// - Automatic memory alignment
 ///
-/// The buffer can switch to borrowed mode without a copy of the data
-///
-/// LanceBuffer does not implement Clone because doing could potentially silently trigger a copy of the data
-/// and we want to make sure that the user is aware of this operation.
-///
-/// If you need to clone a LanceBuffer you can use borrow_and_clone() which will make sure that the buffer
-/// is in borrowed mode before cloning.  This is a zero copy operation (but requires &mut self).
-pub enum LanceBuffer {
-    Borrowed(Buffer),
-    Owned(Vec<u8>),
-}
-
-// Compares equality of the buffers, ignoring owned / unowned status
-impl PartialEq for LanceBuffer {
-    fn eq(&self, other: &Self) -> bool {
-        match (self, other) {
-            (Self::Borrowed(l0), Self::Borrowed(r0)) => l0 == r0,
-            (Self::Owned(l0), Self::Owned(r0)) => l0 == r0,
-            (Self::Borrowed(l0), Self::Owned(r0)) => l0.as_slice() == r0.as_slice(),
-            (Self::Owned(l0), Self::Borrowed(r0)) => l0.as_slice() == r0.as_slice(),
-        }
-    }
-}
-
-impl Eq for LanceBuffer {}
-
-impl std::fmt::Debug for LanceBuffer {
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        let preview = if self.len() > 10 {
-            format!("0x{}...", hex::encode_upper(&self[..10]))
-        } else {
-            format!("0x{}", hex::encode_upper(self.as_ref()))
-        };
-        match self {
-            Self::Borrowed(buffer) => write!(
-                f,
-                "LanceBuffer::Borrowed(bytes={} #bytes={})",
-                preview,
-                buffer.len()
-            ),
-            Self::Owned(buffer) => {
-                write!(
-                    f,
-                    "LanceBuffer::Owned(bytes={} #bytes={})",
-                    preview,
-                    buffer.len()
-                )
-            }
-        }
-    }
-}
+/// LanceBuffer is designed to be used in situations where you might need to
+/// pass around byte buffers efficiently without worrying about ownership.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct LanceBuffer(Buffer);
 
 impl LanceBuffer {
-    /// Convert into a mutable buffer.  If this is a borrowed buffer, the data will be copied.
+    /// Convert into a Vec<u8>, will copy the data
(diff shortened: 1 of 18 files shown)


rust/lance-encoding/src/buffer.rs
```diff
@@ -6,98 +6,45 @@
...
 
 use arrow_buffer::{ArrowNativeType, Buffer, MutableBuffer, ScalarBuffer};
-use itertools::Either;
-use snafu::location;
 
-use lance_core::{utils::bit::is_pwr_two, Error, Result};
+use lance_core::{utils::bit::is_pwr_two, Result};
 
-/// A copy-on-write byte buffer
+/// A copy-on-write byte buffer backed by Arrow Buffer
```
westonpace: I think this is probably an implementation detail.  It's not backed by Arrow Buffer for any particular reason.  It could also be backed by Bytes.

rust/lance-encoding/src/buffer.rs
```diff
@@ -166,92 +113,51 @@ impl LanceBuffer {
 
     /// Convert a buffer into a bytes::Bytes object
     pub fn into_bytes(self) -> bytes::Bytes {
-        match self {
-            Self::Owned(buf) => buf.into(),
-            Self::Borrowed(buf) => buf.into_vec::<u8>().unwrap().into(),
-        }
+        self.0.into_vec::<u8>().unwrap().into()
```
westonpace, line 115: Can we add a comment to this method that this is a zero-copy operation

rust/lance-encoding/src/buffer.rs
```diff
@@ -166,92 +113,51 @@ impl LanceBuffer {
...
     }
 
-    /// Convert into a borrowed buffer, this is a zero-copy operation
-    ///
-    /// This is often called before cloning the buffer
+    /// Convert into a borrowed buffer, this is now a no-op since all buffers are borrowed
     pub fn into_borrowed(self) -> Self {
```
westonpace: Let's just get rid of this method

rust/lance-encoding/src/buffer.rs
```diff
@@ -6,98 +6,45 @@
...
-}
+/// LanceBuffer is designed to be used in situations where you might need to
+/// pass around byte buffers efficiently without worrying about ownership.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct LanceBuffer(Buffer);
 
 impl LanceBuffer {
-    /// Convert into a mutable buffer.  If this is a borrowed buffer, the data will be copied.
+    /// Convert into a Vec<u8>, will copy the data
     pub fn into_owned(self) -> Vec<u8> {
```
westonpace: Let's get rid of this method.  No one calls it.  We already have `to_owned` and there is no advantage to using this.

rust/lance-encoding/src/buffer.rs
```diff
@@ -166,92 +113,51 @@ impl LanceBuffer {
...
-            Self::Owned(buffer) => Self::Owned(buffer.clone()),
-        }
+        Self(Buffer::from_vec(self.0.to_vec()))
     }
 
-    /// Creates a clone of the buffer but also puts the buffer into borrowed mode
+    /// Creates a clone of the buffer (cheap, reference counted)
     ///
-    /// This is a zero-copy operation
+    /// This method exists for backwards compatibility but now just calls clone()
```
westonpace: No need to worry about backwards compatibility.  Let's get rid of the method.

rust/lance-encoding/src/buffer.rs
```diff
@@ -166,92 +113,51 @@ impl LanceBuffer {
...
-                Self::Borrowed(buffer)
-            }
-        }
+        self.clone()
     }
 
-    /// Clones the buffer but fails if the buffer is in owned mode
+    /// Clones the buffer (always cheap, reference counted)
+    ///
+    /// This method exists for backwards compatibility
```
westonpace: Let's get rid of the method.

rust/lance-encoding/src/buffer.rs
```diff
@@ -166,92 +113,51 @@ impl LanceBuffer {
...
-    /// > needs to be equal to satisfy the [`dealloc`] requirement that memory must be
-    /// > allocated and deallocated with the same layout.)
-    ///
-    /// However, we can safely reinterpret Vec<T> into &[u8] which is what happens here.
     pub fn reinterpret_vec<T: ArrowNativeType>(vec: Vec<T>) -> Self {
-        Self::Borrowed(Buffer::from_vec(vec))
+        Self(Buffer::from_vec(vec))
     }
 
     /// Reinterprets Arc<[T]> as a LanceBuffer
```
westonpace, line 142: Why remove the rest of this comment?  I think it is helpful.

rust/lance-encoding/src/buffer.rs
```diff
@@ -166,92 +113,51 @@ impl LanceBuffer {
...
     /// Make an owned copy of the buffer (always does a copy of the data)
     pub fn deep_copy(&self) -> Self {
-        match self {
-            Self::Borrowed(buffer) => Self::Owned(buffer.to_vec()),
-            Self::Owned(buffer) => Self::Owned(buffer.clone()),
-        }
+        Self(Buffer::from_vec(self.0.to_vec()))
     }
 
     /// Reinterprets a Vec<T> as a LanceBuffer
```
westonpace, line 128: I think most of this comment is helpful.  Maybe just remove the `Note that this creates a borrowed buffer.` part and keep the rest?

rust/lance-encoding/src/buffer.rs
```diff
@@ -298,13 +204,16 @@ impl LanceBuffer {
...
+        Self(Buffer::from_vec(data))
     }
 
     /// Zips multiple buffers into a single buffer, consuming the input buffers
     ///
     /// Unlike concat_into_one this "zips" the buffers, interleaving the values
     pub fn zip_into_one(buffers: Vec<(Self, u64)>, num_values: u64) -> Result<Self> {
+        use lance_core::Error;
+        use snafu::location;
+
```
westonpace: Can we just keep these `use` statements at the top?  This isn't in a `cfg` block so I'm not sure we're saving much.

rust/lance-encoding/src/buffer.rs
```diff
@@ -398,30 +300,26 @@ impl LanceBuffer {
...
-    ///
     /// This method performs the bit slice using the Arrow convention of *bitwise* little-endian
     ///
     /// This means, given the bit buffer 0bABCDEFGH_HIJKLMNOP and the slice starting at bit 3 and
     /// with length 8, the result will be 0bNOPABCDE
     pub fn bit_slice_le_with_length(&mut self, offset: usize, length: usize) -> Self {
-        let Self::Borrowed(borrowed) = self.borrow_and_clone() else {
-            unreachable!()
-        };
         // Use this and remove backport once we upgrade to arrow-rs 54
```
westonpace: I think we're on arrow-rs 54 now.  We can probably remove this backport (can be done in follow-up)

westonpace:
> cc @westonpace, should this PR be marked as breaking change? LanceBuffer seems only used in lance-encoding.

No need to mark as a breaking change.  `LanceBuffer` is very much internal.

Xuanwo:
> We can also get rid of all the `try_clone` methods (both here and in `DataBlock`) but if you want to save that for a follow-up that is fine.

Yep, let me finish them in the following PR. I want to get this merged first.

Xuanwo:
> Yep, let me finish them in the following PR. I want to get this merged first.

Oops. Forget to merge this PR yeaterday, let me finish the work directly.

rust/lance-encoding/src/buffer.rs
```diff
@@ -6,98 +6,45 @@
...
 
 use arrow_buffer::{ArrowNativeType, Buffer, MutableBuffer, ScalarBuffer};
-use itertools::Either;
-use snafu::location;
-
 use lance_core::{utils::bit::is_pwr_two, Error, Result};
+use snafu::location;
 
-/// A copy-on-write byte buffer
+/// A copy-on-write byte buffer backed by Arrow Buffer
```
Xuanwo:
```suggestion
/// A copy-on-write byte buffer
```

rust/lance-encoding/src/buffer.rs
Xuanwo: This seems unchanged?

Issue #3788, opened 2025-05-08
Split up large reads in the lance file reader


westonpace:
In the 2.0 format we split large blocks on write.  This has caused various issues (empty pages, very small pages, etc.)  It would be more effective (and not too difficult) to split large blocks on read.  I would like to take this approach for 2.1.

Xuanwo:
Let me take a look over this.

PR #4462, opened 2025-08-14
feat: implement large read io split in format 2.1

Xuanwo:
This PR will implement large read IO splitting in format 2.1. 

In the future, we can implement new improvements such as enhanced I/O split logic and transforming read I/O into a stream, as a follow-up enhancement.

---

closes https://github.com/lancedb/lance/issues/3788

### DIFF
--- a/rust/lance-file/src/v2/io.rs
+++ b/rust/lance-file/src/v2/io.rs
@@ -5,15 +5,91 @@ use futures::{future::BoxFuture, FutureExt};
 use lance_encoding::EncodingsIo;
 use lance_io::scheduler::FileScheduler;
 
+use super::reader::DEFAULT_READ_CHUNK_SIZE;
+
 #[derive(Debug)]
-pub struct LanceEncodingsIo(pub FileScheduler);
+pub struct LanceEncodingsIo {
+    scheduler: FileScheduler,
+    /// Size of chunks when reading large pages
+    read_chunk_size: u64,
+}
+
+impl LanceEncodingsIo {
+    pub fn new(scheduler: FileScheduler) -> Self {
+        Self {
+            scheduler,
+            read_chunk_size: DEFAULT_READ_CHUNK_SIZE,
+        }
+    }
+
+    pub fn with_read_chunk_size(mut self, read_chunk_size: u64) -> Self {
+        self.read_chunk_size = read_chunk_size;
+        self
+    }
+}
 
 impl EncodingsIo for LanceEncodingsIo {
     fn submit_request(
         &self,
-        range: Vec<std::ops::Range<u64>>,
+        ranges: Vec<std::ops::Range<u64>>,
         priority: u64,
     ) -> BoxFuture<'static, lance_core::Result<Vec<bytes::Bytes>>> {
-        self.0.submit_request(range, priority).boxed()
+        let mut split_ranges = Vec::new();
+        let mut split_indices = Vec::new(); // Track which original range each split came from
+
+        // Split large ranges into smaller chunks
+        //
+        // TODO: consider read_chunk_size before submitting requests.
+        for (idx, range) in ranges.iter().enumerate() {
+            let range_size = range.end - range.start;
+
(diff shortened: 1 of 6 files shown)


rust/lance-file/src/v2/reader.rs
```diff
@@ -51,6 +51,10 @@ use crate::{
 
 use super::io::LanceEncodingsIo;
 
+/// Default chunk size for reading large pages (8MiB)
+/// Pages larger than this will be split into multiple chunks during read
+pub const DEFAULT_READ_CHUNK_SIZE: u64 = 8 * 1024 * 1024;
```
Xuanwo, line 56: cc @westonpace, Let me know what do you think.

rust/lance-file/src/v2/writer.rs
```diff
@@ -42,7 +42,9 @@ use crate::format::MAGIC;
 /// Pages buffers are aligned to 64 bytes
 pub(crate) const PAGE_BUFFER_ALIGNMENT: usize = 64;
 const PAD_BUFFER: [u8; PAGE_BUFFER_ALIGNMENT] = [72; PAGE_BUFFER_ALIGNMENT];
-const MAX_PAGE_BYTES: usize = 32 * 1024 * 1024;
+// In 2.1+, we allow larger pages and split them on read instead of write
+// This avoids empty pages and small pages issues from write-time splitting
+const MAX_PAGE_BYTES: usize = 128 * 1024 * 1024; // 128MB default, was 32MB
```
Xuanwo: I want to increase this at the write, ok to revert it.

cc @westonpace

rust/lance-file/src/v2/io.rs
```diff
@@ -5,15 +5,91 @@ use futures::{future::BoxFuture, FutureExt};
...
+                split_ranges.push(range.clone());
+                split_indices.push(idx);
+            }
+        }
+
+        let fut = self.scheduler.submit_request(split_ranges, priority);
+
+        async move {
+            let split_results = fut.await?;
+            let mut results = vec![Vec::new(); ranges.len()];
```
westonpace, line 77: Maybe this is a premature optimization but I wonder if we should add...

```
if split_results.len() == ranges.len() {
  return Ok(split_results);
}
```

This way we take a shortcut for the happy (and probably very common) path?

rust/lance-file/src/v2/reader.rs
westonpace: 8MB is good.

rust/lance-file/src/v2/writer.rs
westonpace: Let's not change this.  I'm worried about creating a regression in 2.0.

Fortunately, I think the 2.1 encoder completely ignores this field :laughing: so maybe we should just document that this is a 2.0 only flag.

rust/lance-file/src/v2/io.rs
Xuanwo: Added

rust/lance-file/src/v2/writer.rs
```diff
@@ -42,6 +42,9 @@ use crate::format::MAGIC;
 /// Pages buffers are aligned to 64 bytes
 pub(crate) const PAGE_BUFFER_ALIGNMENT: usize = 64;
 const PAD_BUFFER: [u8; PAGE_BUFFER_ALIGNMENT] = [72; PAGE_BUFFER_ALIGNMENT];
+// In 2.1+, we split large pages on read instead of write to avoid empty pages
+// and small pages issues. However, we keep the write-time limit at 32MB to avoid
+// potential regressions in 2.0 format readers.
 const MAX_PAGE_BYTES: usize = 32 * 1024 * 1024;
```
westonpace:
```suggestion
// potential regressions in 2.0 format readers.
//
// This limit is not applied in the 2.1 writer
const MAX_PAGE_BYTES: usize = 32 * 1024 * 1024;
```

Minor suggestion.  The current wording isn't clear that it isn't used in 2.1.

PR #4464, opened 2025-08-14
feat: use single session context in dataset sql

wojiaodoubao:
This is related to https://github.com/lancedb/lance/pull/4420#pullrequestreview-3116559468. We should use single session context for better performance.

The single session context causes a register table conflicts error. Regardless of whether we use the session context or the session state to register tables, we encounter this error because registered tables are shared. To address this problem, I introduced a table placeholder {{DATASET}}. In most cases, users should not set the table name manually but instead use the placeholder in their SQL queries, something like 'SELECT * FROM {{DATASET}}'. The SqlQueryBuilder will generate a unique table name for each query, and deregister the table after query is done. However, if users can confidently ensure that the table name will not conflict, they can manually set the table name.

There is another approach to solve this issue, which allows users to freely set table names without requiring the use of placeholders. In this approach, we first parse the SQL into an AST, traverse the AST to find all user-defined table names, and replace them with table names that include a UUID suffix. Then, we generate a logical plan from the modified AST and execute it, followed by deregistering the table. What do you think?  @westonpace @jackye1995 @yanghua

### DIFF
--- a/java/core/src/main/java/com/lancedb/lance/SqlQuery.java
+++ b/java/core/src/main/java/com/lancedb/lance/SqlQuery.java
@@ -21,11 +21,9 @@ import java.io.IOException;
 import java.util.Optional;
 
 public class SqlQuery {
-  private static final String DEFAULT_TABLE_NAME = "dataset";
-
   private Dataset dataset;
   private String sql;
-  private String table = DEFAULT_TABLE_NAME;
+  private Optional<String> table = Optional.empty();
   private boolean withRowId = false;
   private boolean withRowAddr = false;
 
@@ -34,8 +32,22 @@ public class SqlQuery {
     this.sql = sql;
   }
 
+  /**
+   * Specify a "table name" for the dataset, so that you can run SQL queries against it. In most
+   * cases, we should not directly set the table_name. Instead, use {{DATASET}} as a placeholder for
+   * the table name.
+   *
+   * <p>Example
+   *
+   * <pre>
+   * String sql = "SELECT * FROM {{DATASET}} WHERE age > 20";
+   * </pre>
+   *
+   * If you must set a table name, try to use a name that is unlikely to conflict, otherwise we may
+   * encounter a 'table already exists' error.
+   */
   public SqlQuery tableName(String tableName) {
-    this.table = tableName;
+    this.table = Optional.ofNullable(tableName);
     return this;
   }
 
@@ -51,8 +63,7 @@ public class SqlQuery {
 
   public ArrowReader intoBatchRecords() throws IOException {
     try (ArrowArrayStream s = ArrowArrayStream.allocateNew(dataset.allocator())) {
-      intoBatchRecords(
-          dataset, sql, Optional.ofNullable(table), withRowId, withRowAddr, s.memoryAddress());
+      intoBatchRecords(dataset, sql, table, withRowId, withRowAddr, s.memoryAddress());
       return Data.importArrayStream(dataset.allocator(), s);
     }
   }
--- a/java/core/src/test/java/com/lancedb/lance/SqlQueryTest.java
+++ b/java/core/src/test/java/com/lancedb/lance/SqlQueryTest.java
@@ -55,7 +55,7 @@ public class SqlQueryTest {
   @Test
   public void testToRecordBatches() throws IOException {
     // Test normal query
-    ArrowReader reader = dataset.sql("select * from " + NAME).tableName(NAME).intoBatchRecords();
+    ArrowReader reader = dataset.sql("select * from {{DATASET}}").intoBatchRecords();
     Assertions.assertEquals(
         "Schema<id: Int(32, true), name: Utf8>",
         reader.getVectorSchemaRoot().getSchema().toString());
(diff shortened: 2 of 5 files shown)


java/core/src/main/java/com/lancedb/lance/SqlQuery.java
```diff
@@ -34,8 +32,22 @@ public SqlQuery(Dataset dataset, String sql) {
...
 
+  /**
+   * Specify a "table name" for the dataset, so that you can run SQL queries against it. In most
+   * cases, we should not directly set the table_name. Instead, use {{DATASET}} as a placeholder for
+   * the table name.
+   *
+   * <p>Example
+   *
+   * <pre>
+   * String sql = "SELECT * FROM {{DATASET}} WHERE age > 20";
```
jackye1995, line 43: hmm this experience seems very cumbersome... I am starting to wonder if this is not the right place to expose SQL. Maybe we should do that at namespace level, so we can use the actual table name of the dataset to run SQL.

wojiaodoubao: > Maybe we should do that at namespace level...

Do you mean https://lancedb.github.io/lance-namespace/spec/operations/query-table/ ? This interface makes sense to me. The only thing to note is that we cannot use table names like ns0.ns1.table, as the . will be parsed by DataFusion. We might need to use name like ns0_ns1_table, or force user quote "ns0.ns1.table" in their sql. This can be considered further when we actually start implementing it.

> I am starting to wonder if this is not the right place to expose SQL.

I am slightly inclined to keep the SQL interface. It has the advantage of making it easier to use SQL based on the dataset api. Directly using the dataset is still an important scenario.

Hi @yanghua do you have more input ?

yanghua: I also think using a placeholder as a table name is not experience-friendly.

@wojiaodoubao Can you make the table with this pattern `{user-specified}-{uuid}`? IMO, we can find a way to solve this.

> I am slightly inclined to keep the SQL interface. It has the advantage of making it easier to use SQL based on the dataset api. Directly using the dataset is still an important scenario.

+1, we have used this API in our production. The user experience and performance(for less than hundreds of millions) are both very good. And very light-weight (no other dependency on catalog or distributed sql engine).

jackye1995: > Do you mean https://lancedb.github.io/lance-namespace/spec/operations/query-table/ ?

No, what I am thinking is just to assign the table with actual name under the context that makes sense. If it is in a directory, then it can just use the table's folder name there is no need to have any placeholder or uuid.

For example, consider if you have folder:

```
- tmp
  - t1.lance
  - t2.lance
```

```
import lance_namespace as ln

ns = ln.connect("dir", root="/tmp")

ns.sql("select t1.* from t1, t2 where t1.c1 = t2.c2") 
```

So it's basically the same interface and almost the same implementation, just lifted to a higher level. You can do SQL with the actual table name, just try to register the table provider with the table's actual name/id, and if it already exists then it is guaranteed to be the same table. It also unlocks features like join. Seems pretty nice to me, and I don't think there is any perf tradeoff we are making here.

For other more complicated namespace implementations, I think DataFusion can support up to 3 levels of `TableReference`. It should work for most namespaces. If an implementation somehow has more levels (e.g. Gravitino), you can also always define it to connect to the third level from the bottom.

wojiaodoubao: @jackye1995 Thanks your explanation. Agree 'namespace level sql api' is a good idea, much better than placeholder.

jackye1995: > just try to register the table provider with the table's actual name/id, and if it already exists then it is guaranteed to be the same table

One thing I haven't found an ideal solution is, if for some query we want the `_rowid` or `_rowaddr` but for some we don't want, how do we deal with it. But I think that's a problem not unique to the newly proposed direction, we have to solve as long as we are using a global session. We can treat that as a separated problem for now.

westonpace: Another approach someone mentioned, instead of a bunch of `allow_row_id`, `allow_row_addr`, etc. flags is to just have a single `include_meta_cols` flag.  If that is set the the full schema (with all metadata columns) is made available to Datafusion.

I agree that the namespace API is a more natural fit for an SQL API.  However, that will push the feature into `lancedb`, not `lance` (not a problem, just calling attention to this)

westonpace: Another option, if there is some desire to keep an API at the lance level, is to add a custom schema provider that always returns a lance table provider for our dataset regardless of the name.  The provider could also fail if it was asked for tables with two different names.

westonpace: Although, of the available approaches, if we are going to stay in pylance, I like the {{DATASET}} approach best.

jackye1995: > that will push the feature into lancedb, not lance

We can add the impl in lance_namespace, and bring it to lancedb through that layer. I am doing some work right now regarding it.

> Another approach someone mentioned, instead of a bunch of allow_row_id, allow_row_addr, etc. flags is to just have a single include_meta_cols flag. If that is set the the full schema (with all metadata columns) is made available to Datafusion.

I think the problem is more about if we have registered a table without the metadata columns for one query, and then the same table is reused in another query that need metadata columns, how do we do that. Do we always have to check and re-register the table? I am not super familiar with all DF features, @westonpace is there any good way?

westonpace: > We can add the impl in lance_namespace, and bring it to lancedb through that layer. I am doing some work right now regarding it.

That works.

> I think the problem is more about if we have registered a table without the metadata columns for one query, and then the same table is reused in another query that need metadata columns, how do we do that. Do we always have to check and re-register the table? I am not super familiar with all DF features, @westonpace is there any good way?

It's a lance feature, DF has no real concept of metadata columns that I'm aware of.  The problem boils down to the fact that DF expects tables to return a schema with all available columns.  So these flags just control whether or not those columns show up.

So one simple approach is to just always include these special columns.  Users can always choose whether they want those columns included in the query or not with project (e.g. select).  In other words, turning metadata columns on does not guarantee they are on the output, it only guarantees they are in the schema.

jackye1995: > Users can always choose whether they want those columns included in the query or not with project (e.g. select)

If we do that, I think if user does a `SELECT *`, it will basically always expand to include the metadata columns? Maybe that is fine. Ideally I think we want some experience like PG `ctid` so that user have to explicitly write `SELECT ctid, *` to get that value.

westonpace: For another reference, in postgres these are "[system columns](https://www.postgresql.org/docs/8.2/ddl-system-columns.html)" and they don't show up in the schema but you can still select them.  I like this behavior but we'd need changes to Datafusion to enable it.

PR #4440, opened 2025-08-12
refactor: optimize MiniBlockRepIndex decode by decoding from bytes directly

Xuanwo:
`MiniBlockRepIndex::decode` used to be an allocation-heavy operation, creating numerous `Vec<Vec<u8>>` instances in memory only to drop them after parsing. The idea behind this PR is straightforward: decode the rep index directly from bytes without introducing any additional allocations.

## Before

```shell
Benchmarking V2_1 Single Random Take FileReader(1048577 file size, 1024 batches, 1000 rows per take)
Benchmarking V2_1 Single Random Take FileReader(1048577 file size, 1024 batches, 1000 rows per take): Warming up for 3.0000 s

Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 16.7s, or reduce sample count to 20.
Benchmarking V2_1 Single Random Take FileReader(1048577 file size, 1024 batches, 1000 rows per take): Collecting 100 samples in estimated 16.699 s (100 iterations)
Benchmarking V2_1 Single Random Take FileReader(1048577 file size, 1024 batches, 1000 rows per take): Analyzing
V2_1 Single Random Take FileReader(1048577 file size, 1024 batches, 1000 rows per take)
                        time:   [162.71 ms 163.23 ms 163.74 ms]
                        change: [+1.7266% +2.6730% +3.6496%] (p = 0.00 < 0.10)
                        Performance has regressed.
```

The `MiniBlockScheduler::initiate` only takes 45.3% of whole time.

image

## After

```shell
Benchmarking V2_1 Single Random Take FileReader(1048577 file size, 1024 batches, 1000 rows per take)
Benchmarking V2_1 Single Random Take FileReader(1048577 file size, 1024 batches, 1000 rows per take): Warming up for 3.0000 s

Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 10.0s, or reduce sample count to 40.

Benchmarking V2_1 Single Random Take FileReader(1048577 file size, 1024 batches, 1000 rows per take): Collecting 100 samples in estimated 10.034 s (100 iterations)
Benchmarking V2_1 Single Random Take FileReader(1048577 file size, 1024 batches, 1000 rows per take): Analyzing
V2_1 Single Random Take FileReader(1048577 file size, 1024 batches, 1000 rows per take)
                        time:   [92.341 ms 93.659 ms 94.998 ms]
                        change: [-43.446% -42.621% -41.718%] (p = 0.00 < 0.10)
                        Performance has improved.
[...]

PR #4467, opened 2025-08-14
fix: lance failed to encode 1025 rep/def values

### LINKED ISSUE #4466: fuzz: lance failed to encode 1025 rep/def 1025 values

Xuanwo:
fix #4466

---

Hi, @westonpace I'm not sure if this fix is correct. Maybe also related to my changes of https://github.com/lancedb/lance/pull/4440

### DIFF
--- a/rust/lance-encoding/src/encodings/logical/primitive.rs
+++ b/rust/lance-encoding/src/encodings/logical/primitive.rs
@@ -1358,8 +1358,27 @@ impl ChunkInstructions {
 
             while rows_needed > 0 || need_preamble {
                 let chunk = &rep_index.blocks[block_index];
-                let rows_avail = chunk.starts_including_trailer - to_skip;
-                debug_assert!(rows_avail > 0);
+                let rows_avail = chunk.starts_including_trailer.saturating_sub(to_skip);
+
+                // Handle blocks that are entirely preamble (rows_avail = 0)
+                // These blocks have no rows to take but may have a preamble we need
+                if rows_avail == 0 {
+                    // Only process if this chunk has a preamble we need
+                    if chunk.has_preamble && need_preamble {
+                        chunk_instructions.push(Self {
+                            chunk_idx: block_index,
+                            preamble: PreambleAction::Take,
+                            rows_to_skip: 0,
+                            rows_to_take: 0,
+                            take_trailer: false,
+                        });
+                        need_preamble = false;
+                    }
+                    // Move to next block
+                    block_index += 1;
+                    to_skip = 0;
+                    continue;
+                }
 
                 let rows_to_take = rows_avail.min(rows_needed);
                 rows_needed -= rows_to_take;
--- a/rust/lance-encoding/src/encodings/logical/primitive.rs
+++ b/rust/lance-encoding/src/encodings/logical/primitive.rs
@@ -1357,12 +1357,18 @@ impl ChunkInstructions {
             let mut to_skip = user_range.start - rep_index.blocks[block_index].first_row;
 
             while rows_needed > 0 || need_preamble {
+                // Check if we've gone past the last block
+                if block_index >= rep_index.blocks.len() {
+                    break;
+                }
+
                 let chunk = &rep_index.blocks[block_index];
                 let rows_avail = chunk.starts_including_trailer.saturating_sub(to_skip);
 
                 // Handle blocks that are entirely preamble (rows_avail = 0)
                 // These blocks have no rows to take but may have a preamble we need
-                if rows_avail == 0 {
+                // We only look for preamble if to_skip == 0 (we're not skipping rows)
+                if rows_avail == 0 && to_skip == 0 {
                     // Only process if this chunk has a preamble we need
                     if chunk.has_preamble && need_preamble {
                         chunk_instructions.push(Self {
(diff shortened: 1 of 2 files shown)


rust/lance-encoding/src/encodings/logical/primitive.rs
```diff
@@ -1358,8 +1358,27 @@ impl ChunkInstructions {
...
+                    // Only process if this chunk has a preamble we need
+                    if chunk.has_preamble && need_preamble {
+                        chunk_instructions.push(Self {
+                            chunk_idx: block_index,
+                            preamble: PreambleAction::Take,
+                            rows_to_skip: 0,
+                            rows_to_take: 0,
+                            take_trailer: false,
+                        });
+                        need_preamble = false;
```
westonpace: I think we can only set `need_preamble = false` if there is at least one row (`chunk.starts_including_trailer > 0`)  Otherwise, it is possible the chunk is entirely preamble in which case we need at least the entire chunk and the preamble from the next chunk.

rust/lance-encoding/src/encodings/logical/primitive.rs
```diff
@@ -1358,8 +1358,27 @@ impl ChunkInstructions {
 
             while rows_needed > 0 || need_preamble {
                 let chunk = &rep_index.blocks[block_index];
-                let rows_avail = chunk.starts_including_trailer - to_skip;
-                debug_assert!(rows_avail > 0);
+                let rows_avail = chunk.starts_including_trailer.saturating_sub(to_skip);
+
+                // Handle blocks that are entirely preamble (rows_avail = 0)
+                // These blocks have no rows to take but may have a preamble we need
+                if rows_avail == 0 {
```
westonpace: I think this needs to be `rows_avail == 0 && to_skip == 0`.  If we need to skip even 1 row then we are not looking for the preamble.  That being said...I'm not sure if it is possible that `rows_avail == 0 && to_skip != 0` because otherwise we would have skipped the block in the binary search.  Still, might not hurt to be explicit.

westonpace (changes requested):
Good find.  This chunk instruction code is always super confusing to me (and I wrote it :laughing:)

Marking `Request changes` since I think the comment about `need_preamble = false` will need to be addressed.

rust/lance-encoding/src/encodings/logical/primitive.rs
Xuanwo: You are right, let me handle this.

rust/lance-encoding/src/encodings/logical/primitive.rs
Xuanwo: Added an extra edge check.

rust/lance-encoding/src/encodings/logical/primitive.rs
```diff
@@ -1357,9 +1357,49 @@ impl ChunkInstructions {
             let mut to_skip = user_range.start - rep_index.blocks[block_index].first_row;
 
             while rows_needed > 0 || need_preamble {
+                // Check if we've gone past the last block
+                if block_index >= rep_index.blocks.len() {
+                    break;
+                }
```
westonpace: Did you actually run into this?  I thought rows_needed couldn't be larger than the number of rows available?

rust/lance-encoding/src/encodings/logical/primitive.rs
```diff
@@ -1357,9 +1357,49 @@ impl ChunkInstructions {
...
+                        });
+                        // Only set need_preamble = false if the chunk has at least one row
+                        // Otherwise, the chunk is entirely preamble and we need the next chunk's preamble too
+                        if chunk.starts_including_trailer > 0 {
+                            need_preamble = false;
+                        }
+                    }
+                    // Move to next block
+                    block_index += 1;
+                    to_skip = 0;
```
westonpace: This line is redundant (we have `to_skip == 0` above) but harmless.
```suggestion
```

westonpace (approved):
This logic is correct I believe

rust/lance-encoding/src/encodings/logical/primitive.rs
Xuanwo: Yep, in our test we can hit a case where all blocks are preamble, and the `block_index` can larger than `rep_index.blocks.len()` after `block_index += 1`. I can merge this change with `chunk.starts_including_trailer > 0` to make it more clear like:

```diff
if chunk.has_preamble && need_preamble {
    chunk_instructions.push(...);

-    if chunk.starts_including_trailer > 0 {
+    if chunk.starts_including_trailer > 0 || block_index == rep_index.blocks.len() - 1 {
        need_preamble = false;
    }
}
```

Xuanwo: The test `encodings::logical::list::tests::test_fuzz_issue_4466` exactly failed on this case.

```
---- encodings::logical::list::tests::test_fuzz_issue_4466 stdout ----

thread 'encodings::logical::list::tests::test_fuzz_issue_4466' panicked at rust/lance-encoding/src/encodings/logical/primitive.rs:1365:46:
index out of bounds: the len is 2 but the index is 2


failures:
    encodings::logical::list::tests::test_fuzz_issue_4466
```

rust/lance-encoding/src/encodings/logical/primitive.rs
```diff
@@ -1357,9 +1357,48 @@ impl ChunkInstructions {
...
+                            preamble: PreambleAction::Take,
+                            rows_to_skip: 0,
+                            rows_to_take: 0,
+                            take_trailer: false,
+                        });
+                        // Only set need_preamble = false if the chunk has at least one row
+                        // Otherwise, the chunk is entirely preamble and we need the next chunk's preamble too
+                        if chunk.starts_including_trailer > 0 {
+                            need_preamble = false;
+                        }
```
westonpace:
```suggestion
                        if chunk.starts_including_trailer > 0
                            || block_index == rep_index.blocks.len() - 1
                        {
                            need_preamble = false;
                        }
```

rust/lance-encoding/src/encodings/logical/primitive.rs
```diff
@@ -1357,9 +1357,48 @@ impl ChunkInstructions {
             let mut to_skip = user_range.start - rep_index.blocks[block_index].first_row;
 
             while rows_needed > 0 || need_preamble {
+                // Check if we've gone past the last block
+                if block_index >= rep_index.blocks.len() {
+                    break;
+                }
```
westonpace:
```suggestion
                // Check if we've gone past the last block (should not happen)
                if block_index >= rep_index.blocks.len() {
                    log::warn!("schedule_instructions inconsistency: block_index >= rep_index.blocks.len(), exiting early");
                    break;
                }
```

westonpace (approved):
Ok, you were right.  We need to set `needs_preamble = false` on the last block :laughing:

PR #4474, opened 2025-08-14
refactor: reduce lance-encoding compile times

westonpace:
I created new features for lz4, zstd, and bitpacking (all on by default) which can be disabled if someone is working on lance-encoding.  With these features disabled, incremental build times of lance-encoding can be under 1s.

In addition, this moves the bitpacking code into its own crate.  Partly to enable the above features, but also partly to isolate compile times.  There is a lot of code generation in the bitpacking logic, and that logic is almost never modified, and we don't want that to have to run during incremental compilations of the lance-encoding crate.  So there should be some speedup even with the default features turned on (incremental builds ~1.5s for me)


Xuanwo (approved):
I like this idea, ok with all CI passed.

Issue #4472, opened 2025-08-14
Python test on macOS arm failed for deadlock


Xuanwo:
```
python/tests/forward_compat/test_compat.py::test_scans SKIPPED (--run-
forward not specified)
python/tests/forward_compat/test_compat.py::test_pq_buffer SKIPPED
python/tests/test_arrow.py::test_bf16_value PASSED
python/tests/test_arrow.py::test_bf16_repr PASSED
python/tests/test_arrow.py::test_bf16_array_str PASSED
python/tests/test_arrow.py::test_bf16_pandas[True] PASSED
python/tests/test_arrow.py::test_bf16_pandas[False] PASSED
python/tests/test_arrow.py::test_bf16_numpy PASSED
python/tests/test_arrow.py::test_bf16_array_cast PASSED
python/tests/test_arrow.py::test_bf16_fixed_size_list_cast PASSED
python/tests/test_arrow.py::test_bf16_roundtrip PASSED
python/tests/test_arrow.py::test_roundtrip_take_ext_types PASSED
python/tests/test_arrow.py::test_image_uri_arrays PASSED
[mutex.cc : 452] RAW: Lock blocking 0x600001b190f8   @
```

For example: https://github.com/lancedb/lance/actions/runs/16961475659/job/48075118130

---

Maybe related to https://github.com/lancedb/lance/pull/4449, the failure first show up after this PR merged.

Xuanwo:
I can reproduce locally on macOS by:

```shell
> uv run pytest -vvv python/tests/test_arrow.py
...
python/tests/test_arrow.py::test_image_tensor_arrays
```

Xuanwo:
Python package changes:


Package | Before | After
-- | -- | --
numpy | 2.1.3 | 2.3.2
protobuf | 5.29.5 | 6.31.1
tensorboard | 2.19.0 | 2.20.0
tensorflow | 2.19.0 | 2.20.0

Xuanwo:
Related to https://github.com/tensorflow/tensorflow/issues/98563

Xuanwo:
Closing this as completed. #4476 restored the macOS ARM CI path, and #7685 later moved the TensorFlow integration and its tests out of this repository, removing the original trigger surface. If a current macOS ARM Python job still deadlocks, please reopen this issue with the new run URL.

PR #4476, opened 2025-08-14
ci: pin tersorflow version to bring macOS ARM CI back

Xuanwo:
Part of https://github.com/lancedb/lance/issues/4472

I will remove it after ternsorflow have a fix.

### DIFF
--- a/python/pyproject.toml
+++ b/python/pyproject.toml
@@ -55,7 +55,10 @@ tests = [
     "pandas",
     "polars[pyarrow,pandas]",
     "pytest",
-    "tensorflow",
+    # Workardound for https://github.com/lancedb/lance/issues/4472
+    #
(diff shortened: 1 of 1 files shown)


python/pyproject.toml
```diff
@@ -55,7 +55,10 @@ tests = [
     "pandas",
     "polars[pyarrow,pandas]",
     "pytest",
-    "tensorflow",
+    # Workardound for https://github.com/lancedb/lance/issues/4472
+    #
+    # tensorflow is broken on 2.20.0 with macOS ARM.
+    "tensorflow==2.19.0",
```
eddyxu: can we use `tensorflow<=2.19`

Xuanwo: fixed

PR #4478, opened 2025-08-14
feat: allow reading the _rowoffset

westonpace:
The row offset (_rowoffset) is similar to the row id and the row address but slightly different.  The row offset is always a number from 0 to len(dataset).

The row offset is highly unstable (it can change from rows being deleted or from compaction, updates, or anything else that modifies a dataset) but it has one useful property.  The user does not need to query the list of available row offset up front for sampling purposes.  This makes it useful in some cases for sampling a dataset.

### DIFF
--- a/python/python/lance/dataset.py
+++ b/python/python/lance/dataset.py
@@ -557,6 +557,7 @@ class LanceDataset(pa.dataset.Dataset):
         prefilter: Optional[bool] = None,
         with_row_id: Optional[bool] = None,
         with_row_address: Optional[bool] = None,
+        with_row_offset: Optional[bool] = None,
         use_stats: Optional[bool] = None,
         fast_search: Optional[bool] = None,
         io_buffer_size: Optional[int] = None,
@@ -729,6 +730,7 @@ class LanceDataset(pa.dataset.Dataset):
         setopt(builder.late_materialization, late_materialization)
         setopt(builder.with_row_id, with_row_id)
         setopt(builder.with_row_address, with_row_address)
+        setopt(builder.with_row_offset, with_row_offset)
         setopt(builder.use_stats, use_stats)
         setopt(builder.use_scalar_index, use_scalar_index)
         setopt(builder.fast_search, fast_search)
@@ -809,6 +811,7 @@ class LanceDataset(pa.dataset.Dataset):
         prefilter: Optional[bool] = None,
         with_row_id: Optional[bool] = None,
         with_row_address: Optional[bool] = None,
+        with_row_offset: Optional[bool] = None,
         use_stats: Optional[bool] = None,
         fast_search: Optional[bool] = None,
         full_text_query: Optional[Union[str, dict, FullTextQuery]] = None,
@@ -874,6 +877,8 @@ class LanceDataset(pa.dataset.Dataset):
             Return row ID.
         with_row_address: bool, optional, default False
             Return row address
+        with_row_offset: bool, optional, default False
+            Return row offset
         use_stats: bool, optional, default True
             Use stats pushdown during filters.
         fast_search: bool, optional, default False
@@ -924,6 +929,7 @@ class LanceDataset(pa.dataset.Dataset):
             prefilter=prefilter,
             with_row_id=with_row_id,
             with_row_address=with_row_address,
+            with_row_offset=with_row_offset,
             use_stats=use_stats,
             fast_search=fast_search,
             full_text_query=full_text_query,
@@ -1008,6 +1014,7 @@ class LanceDataset(pa.dataset.Dataset):
         prefilter: Optional[bool] = None,
         with_row_id: Optional[bool] = None,
         with_row_address: Optional[bool] = None,
+        with_row_offset: Optional[bool] = None,
         use_stats: Optional[bool] = None,
         full_text_query: Optional[Union[str, dict]] = None,
         io_buffer_size: Optional[int] = None,
@@ -1044,6 +1051,7 @@ class LanceDataset(pa.dataset.Dataset):
             prefilter=prefilter,
             with_row_id=with_row_id,
             with_row_address=with_row_address,
+            with_row_offset=with_row_offset,
             use_stats=use_stats,
             full_text_query=full_text_query,
             strict_batch_size=strict_batch_size,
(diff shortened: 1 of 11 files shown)


python/python/lance/dataset.py
```diff
@@ -900,6 +912,15 @@ def to_table(
             if scan_in_order is true. Otherwise it will fellow as a random order.
             If specified, the return rows will follow the orderings. If a string is
             specified, it will assume ascending and nulls last ordering.
+        disable_scoring_autoprojection: bool, default False
+            Currently, when a search (vector or full text) is performed, the scoring
+            column (_distance, _score) is added to the end of the output even when a
+            projection is specified.  In the future, this will change.  The columns will
```
wkalt, line 918: not sure who the consumer of these docs is, but it might be useful to explain why it's added (presumably to help with sort operations higher up in the exec plan?)

rust/lance-datafusion/src/projection.rs
```diff
@@ -15,85 +15,88 @@ use std::{
...
-    ///
-    /// Note: this is related, but slightly different, to physical_projection.with_row_addr
-    /// which only tracks if the row address is needed.
-    ///
-    /// desires_row_addr implies with_row_addr is true
-    /// However, it is possible to have deisres_row_addr=false and with_row_addr=true (e.g. during
-    /// a count query)
-    pub desires_row_addr: bool,
+    /// Needs the row address converted into a row offset
+    pub needs_row_offset: bool,
```
wkalt: is it right to understand this as "must project row offset"?

rust/lance/src/io/exec/rowids.rs
```diff
@@ -322,6 +331,233 @@ impl ExecutionPlan for AddRowAddrExec {
...
+        for frag in dataset.get_fragments() {
+            let deletion_vector = frag.get_deletion_vector().await?;
+            frag_id_to_offset.insert(
+                frag.id() as u32,
+                FragInfo {
+                    row_offset,
+                    deletion_vector,
+                },
+            );
+            // Should be sync in most cases
```
wkalt: elaborate on "most"?

rust/lance-datafusion/src/projection.rs
```diff
@@ -15,85 +15,88 @@ use std::{
...
-    pub fn new(base: Arc<dyn Projectable>) -> Self {
-        let physical_projection = Projection::full(base.clone());
-        Self {
-            base,
-            physical_projection,
-            requested_output_expr: None,
-            desires_row_addr: false,
-            desires_row_id: false,
-        }
+    fn add_system_columns(schema: &ArrowSchema) -> ArrowSchema {
```
Xuanwo, line 47: I find `Self::add_system_columns(&full_schema)` is a bit confusing. Is it a good idea to move it out of `ProjectionPlan` as a standalone funcation with new like `add_system_columns_in_schema`?

rust/lance/src/io/exec/rowids.rs
```diff
@@ -322,6 +331,233 @@ impl ExecutionPlan for AddRowAddrExec {
...
+            row_offset += frag.count_rows(None).await? as u64;
+        }
+
+        Self::internal_new(input, Arc::new(frag_id_to_offset))
+    }
+
+    // A dummy placeholder to use in emergency when we can't find the fragment info (better than panic)
```
wkalt: can you spell out the consequences/implications of this?

Xuanwo (approved):
Mostly LGTM, only have a naming thing.

python/python/lance/dataset.py
westonpace: The consumer is the user here.  This only affects the final projection to the user.  So if a sort exec needs the column then it will be there for the sort exec and then dropped later.

There is no real good reason why it is added except when we coded it up originally we thought it might be an interesting column for the user to see and didn't make a way for them to _not_ see it.

rust/lance-datafusion/src/projection.rs
westonpace: Yes...maybe?

The whole concept of "projection plan" is a little sketchy.  We have the physical projection, which is the columns that we need to grab from the fragment readers.  row id and row offset are considered part of the physical projection because they are delivered by the fragment reader.

Then we have the requested_output_expr which is the set of columns the user is actually asking for.  This includes columns that don't exist yet (because they need to be dynamically calculated for example)

Then we have the row offset which, currently, is kind of "in between".  It gets added by the scanner (not the fragment reader) but it isn't exactly a UDF (I suppose it could be but it needs the dataset metadata which would make it a bit weird for a UDF) so we can't just throw it into the requested_output_expr.

As a compromise, we have this field, whose whole purpose is to inform the scanner that it needs to add the "calculate row offset" exec node.

rust/lance/src/io/exec/rowids.rs
westonpace: I'll update the comment.  The only way it wouldn't be sync is if we have a _very_ old writer where we didn't record the counts in the dataset metadata (and so we had to open the file metadata and deletion vector).

rust/lance/src/io/exec/rowids.rs
westonpace: I'll update the comment.  The consequence will be a nonsense row offset that is really large.

rust/lance-datafusion/src/projection.rs
westonpace: Hmm...I'm not entirely sure what your concern is.  Can you elaborate?

This function is supposed to take a schema and then add a bunch of columns to the end of the schema.  So we start with a schema `[X, Y, Z]` and we end with the schema `[X, Y, Z, _rowid, _rowaddr, _rowoffset]`.

At the moment it is static (no `&self`) because the set of system columns is a constant.  In the future, maybe that won't be the case (maybe system columns can be determined by UDFs or some other pluggable mechanism?)  Once that happens this method would probably move over to whatever "system column registry" we have.

rust/lance-datafusion/src/projection.rs
westonpace: I've renamed the field `must_add_row_offset`

rust/lance/src/io/exec/rowids.rs
westonpace: Comment udpated

rust/lance/src/io/exec/rowids.rs
westonpace: I ended up getting rid of this and returning an error instead as that's probably better than a garbage value that might seem accurate (not sure why I thought panic was the only alternative when I wrote the comment)

PR #4481, opened 2025-08-15
refactor!: rename stable row id feature flag

yanghua:


yanghua:
@jackye1995 @wjones127 This PR is ready for review. Not sure if we need to block it until the whole feature is done. What's your opinion?

jackye1995:
I am fine either way, I am good with merging this first as well since the goal is to make it update-stable row ID anyway, and I don't think people are using the flag anywhere at this moment.

yanghua:
> I am fine either way, I am good with merging this first as well since the goal is to make it update-stable row ID anyway, and I don't think people are using the flag anywhere at this moment.

@jackye1995 Thanks for sharing your thoughts. Agree! Now, the PR is ready for review. Please take a look.

jackye1995 (approved):
looks good to me

PR #4483, opened 2025-08-15
test(java): re-construct tests for merge operation

majin1102:
I found merge operation test is not right. It didn't construct by merging a new column, but instead merged some new rows. Although it works, it's not what we wanted.

In the meanwhile the TransactionTest has grown a little tedious and not that friendly to construct multiple testcases on a single operation. So I did a little refactoring.

The work includes:
1. Move all operation testcases into package com/lancedb/lance/operation.
2. Extend some new testcase for Append and Merge operation. We could do similar work to other operations.
3. Original TransactionTest is only used to test transaction framework like readTransaction and transaction properties.


majin1102:
@jackye1995 PTAL when you have time

majin1102:
@yanghua Could you take a look when you have time?
This is only test releated but could block next transaction steps

java/core/src/test/java/com/lancedb/lance/operation/MergeTest.java
```diff
@@ -0,0 +1,290 @@
...
+            }
+          }
+        }
+      }
+    }
+  }
+
+  @Test
+  void testReplaceAsDiffColumns(@TempDir Path tempDir) throws Exception {
+    String datasetPath = tempDir.resolve("testMergeNewColumn").toString();
```
yanghua: Wrong dir name?

java/core/src/test/java/com/lancedb/lance/operation/OperationTestBase.java
```diff
@@ -0,0 +1,99 @@
...
+
+import java.io.File;
+import java.util.Collections;
+import java.util.UUID;
+
+public class OperationTestBase {
+
+  public static final int TEST_FILE_FORMAT_MAJOR_VERSION = 2;
+  public static final int TEST_FILE_FORMAT_MINOR_VERSION = 0;
+  protected static Dataset dataset;
```
yanghua: Must it be `static`?

java/core/src/test/java/com/lancedb/lance/operation/MergeTest.java
majin1102: Yes, thanks for this

java/core/src/test/java/com/lancedb/lance/operation/OperationTestBase.java
majin1102: This is only a movement.
We can make it better with @TestInstance(TestInstance.Lifecycle.PER_CLASS)
Please take another look

PR #4488, opened 2025-08-16
feat: introduce manifest naming scheme v3

jackye1995:


rust/lance-table/src/io/commit.rs
```diff
@@ -66,63 +65,100 @@ use crate::format::{is_detached_version, Index, Manifest};
...
-    /// `_manifests/{u64::MAX - version}.manifest`
+    /// `_versions/{u64::MAX - version}.manifest`
     ///
     /// Zero-padded and reversed for O(1) lookup of latest version on object stores.
     V2,
+    /// `_versions/{reversed_binary_version}.manifest` with `_latest_manifest_hint.json`
+    ///
+    /// Binary representation reversed for S3 throughput optimization.
+    /// E.g., version 5 becomes "1010000000000000000000000000000000000000000000000000000000000000.manifest"
+    V3,
```
wjones127, line 83: Question: the other implementation uses directories like:

```
s3://my-table-data-bucket/my_ns.db/my_table/0101/0110/1001/10110010/category=orders/00000-0-5affc076-96a4-48f2-9cd2-d5efbc9f0c94-00001.parquet
```

It's not clear to me whether this is necessary to get the scaling benefit, as I don't really know what defines a prefix in S3. From what you know, will we get this benefit if we don't have the path separators?

rust/lance-table/src/io/commit.rs
```diff
@@ -233,6 +411,44 @@ async fn current_manifest_path(
     object_store: &ObjectStore,
     base: &Path,
 ) -> Result<ManifestLocation> {
+    // Try to use V3 hint file if available
+    if let Some(latest_hint) = read_latest_manifest_hint_best_effort(object_store, base).await? {
```
wjones127, line 415: IIUC, now all tables on V2 will have to make this IO request to check, right?

rust/lance/src/dataset.rs
wjones127, line 1: Could you update this test:

https://github.com/lancedb/lance/blob/9ec9611f2d08cfa13d6eb48fba94619f9b4edaac/rust/lance/src/dataset.rs#L2489-L2490

It took me a moment realize why this test wasn't failing here:

https://github.com/lancedb/lance/blob/9ec9611f2d08cfa13d6eb48fba94619f9b4edaac/rust/lance/src/dataset.rs#L2540-L2544

It turns out, the IOTracking store isn't counting requests that respond with 404. But that's really important for testing your changes, because getting 404 back is part of the expected path. We should modify the tracking store here (and similar places within that file):

https://github.com/lancedb/lance/blob/9ec9611f2d08cfa13d6eb48fba94619f9b4edaac/rust/lance/src/utils/test.rs#L418-L421

**In addition, I think we need to make sure that test runs for all three manifest naming schemes.**

wjones127:
I have one crazy idea to consider:

1. Keep the V2 scheme
2. Add the `_latest_version_hint.json`
3. Do LIST and GET `_latest_version_hint.json` in parallel
4. If GET returns 404:
    a. Wait for LIST
5. If GET returns result
    a. Return early
    b. Make background task: If LIST gets newer version, write a new `_latest_version_hint.json`

Benefits of that would be:

1. Speed benefit of `_latest_version_hint.json`, even faster without needing HEAD
2. Guaranteed 2 IOPS max
3. No migration needed for existing V2 users
4. No performance penalty for existing V2 users

Disadvantages:

1. If `_latest_version_hint.json` is wildly out of date, user could get back out-of-date version

rust/lance-table/src/io/commit.rs
jackye1995: > /0101/0110/1001/10110010/

The path separator was added because that was used for all data files, and there could be a lot of data files, and HDFS does not have paginated directory listing so it might end up with a very long duration listing without additional directories. And the scale of the number of versions is much smaller, so I did not bother adding the path directories.

jackye1995:
> Make background task: If LIST gets newer version, write a new _latest_version_hint.json

I think that is probably not always feasible, since many reader likely only has read-only permission. But we can probably depend on the writer updating _latest_version_hint.json most of the time.

> If GET returns result
> a. Return early

I think the biggest point we should discuss is this. I also went back and worth on this. We can:
1. treat the `_latest_version_hint.json` as the actual latest version directly, or
2. always don't fully trust this file and try to see what is the actual latest version.

I basically chose the second, that's also why I call this file a "hint" rather than the actual "_latest_version", and I left a comment that if we want best perf, we can just read the hint and directly return without doing HEAD.

In your proposal, we will go with 1, and the inaccuracy is compensated by having a background list to update the hint if it is not actually the latest. My concerns are that
1. when we do that, the current read is already not the latest version
2. linking back to the last comment I made, I think having readers with read-only permission to the dataset is a pretty common setup, and in those case basically we could have a risk of actually reading an old version all the time.

What do you think? @wjones127

rust/lance-table/src/io/commit.rs
jackye1995: oh yes you are right. Even if we go with the new scheme, we probably need to do what you are suggesting for parallel GET and LIST in order to make sure we do not regress performance for v2.

wjones127:
> I think the biggest point we should discuss is this. I also went back and worth on this. We can:
> 
> 1. treat the `_latest_version_hint.json` as the actual latest version directly, or
> 2. always don't fully trust this file and try to see what is the actual latest version.

This is honestly a tough one.

On one hand, if we have continuous writes, the probability it's out of date isn't neglibile. For example, if there is a 20ms delay from writing the manifest to writing `_latest_version_hint.json` and we are committing once per second, that 2% chance of being out of date. If the delay is 100ms, that's 10%. And that's without any crashes or anything.

That's kind of on the border of whether we should care or not. In some use cases, like queries, that kind of lag is almost undetectable to the user. But in other, likes writes, I think that could have a meaningful impact, at least in some situations.

Seems like it would be worth doing some sort of benchmark to see what would be best.

> 3\. Is there any risk that `batch_head_manifests()` could get caught in a loop when another process is writing new versions continuously?

One of the things I like about `LIST` is it gives you the latest for a given snapshot.


One of the hard things about the manifest naming schemes is we are going in kind of blind for what to expect. We don't know which scheme is used by the dataset until we check. I think we might want to consider allowing some hint to be provided by the caller (or catalog?) as to which to expect, and be able to have a more optimal code path based on that hint.

PR #4489, opened 2025-08-16
feat: inverted index for contains_tokens

wojiaodoubao:
This is the second step of #3855


rust/lance-index/src/scalar/inverted/index.rs
```diff
@@ -329,11 +359,19 @@ impl ScalarIndex for InvertedIndex {
...
         _metrics: &dyn MetricsCollector,
     ) -> Result<SearchResult> {
-        let query = query.as_any().downcast_ref::<SargableQuery>().unwrap();
-        return Err(Error::invalid_input(
-            format!("unsupported query {:?} for inverted index", query),
-            location!(),
-        ));
+        let query = query.as_any().downcast_ref::<TextQuery>().unwrap();
+        match query {
+            TextQuery::StringContains(text) => {
```
BubbleCal: should we add a new variant (`TextQuery::TokenContains`) for this? cc @westonpace

rust/lance-index/src/scalar/expression.rs
```diff
@@ -464,9 +464,6 @@ impl ScalarQueryParser for FtsQueryParser {
         }
         let scalar = maybe_scalar(&args[1], data_type)?;
         if let ScalarValue::Utf8(Some(scalar_str)) = scalar {
-            // TODO(https://github.com/lancedb/lance/issues/3855):
-            //
-            // Create the contains_tokens UDF
             if func.name() == "contains_tokens" {
                 let query = TextQuery::StringContains(scalar_str);
                 Some(IndexedExpression::index_query(
```
BubbleCal: it would be cool if we can add more params for this UDF in the future (e.g. AND op, fuzziness, prefix-matching)

BubbleCal (commented):
LGTM, just a minor question

rust/lance-index/src/scalar/inverted/index.rs
```diff
@@ -329,11 +359,19 @@ impl ScalarIndex for InvertedIndex {
...
+                let records = self.do_search(text).await?;
+                let row_ids = records
+                    .column(0)
+                    .as_any()
+                    .downcast_ref::<UInt64Array>()
+                    .unwrap();
+                let row_ids = row_ids.iter().flatten().collect_vec();
+                Ok(SearchResult::AtMost(RowIdTreeMap::from_iter(row_ids)))
```
BubbleCal, line 386: prob we can do this better in the future, if the tokenizer wouldn't change the original texts, we can return `AtLeast` directly

rust/lance-index/src/scalar/expression.rs
wojiaodoubao: Sorry for my late response. I think supporting the AND operator, fuzziness, and prefix-matching is an excellent idea!!! I have a few questions about this idea:

1. I'm not sure if my understanding is correct. The ```contains_tokens``` in #3855 is functionally the same as ```contains```, the difference is it leverages FTS instead of n-grams. If we support `fuzziness...`, the query might look something like this:```SELECT * FROM dataset WHERE contains_tokens(text, "+cat~2 and -dog")``` 
I'm wondering if the semantics of ```contains_tokens``` would shift to "full-text search" rather than "whether it contains a specific substring." This makes me a bit confused.

2. Shall we consider adding a ```fts()``` UDF to facilitate full-text search through SQL? The ```fts()``` UDF would accept a JSON representation of a full_text_query and return a boolean indicating whether the record matches. The query might look like this.
```
let full_text_query = r#"
{
  "query": {
    "bool": {
      "must": [
        { "match": { "content": "asia" } },
        { "match": { "content": { "value": "Felidae", "fuzziness": 2 } } }
      ],
      "should": [
        { "phrase": { "title": "wild animal" } },
        { "phrase": { "title": "home pet" } },
      ]
    }
  }
}
"#;

let sql = format!("SELECT author FROM dataset WHERE fts({})", full_text_query);
```
If a column in the full_text_query does not have an inverted index created, the query will fail instead of attempting to read the record for matching.

There are other approaches for full_text_query, such as the MATCH AGAINST syntax used by MySQL or Lucene's syntax. However, JSON has stronger expressive capabilities and can fully represent FtsQuery, so I think using JSON to express full_text_query might be a good idea.

BubbleCal: 1. if I understand correctly, `contains_tokens` is some different from `contains`, the users should expect `contains_token` would ignore the punctuations and white spaces (depends on the FTS params).

e.g. the text is "hello the world", then `contains_tokens("text", "hello world")` should match it but `contains` not. On the other hand, `contains_tokens("text", "wor")` can't match anything but `contains` would. cc @westonpace 

2. yes we evaluated Lucene's syntax, the FtsQuery object is organized in a very similar way, i think it's not hard to convert a FtsQuery object into JSON. in fact I implemented this converting logic ago, we are not using it because most users are using lance via the SDK, and query via constructing FtsQuery object is easier and more efficient (see https://lancedb.com/docs/search/full-text-search/#boolean-queries).

But yes it's a different story for SQL, I think it's a good idea to have the JSON syntax for SQL, we just need to parse it into the FtsQuery object

wojiaodoubao: > contains_tokens is some different from contains, the users should expect contains_token would ignore the punctuations and white spaces (depends on the FTS params).

Hi @BubbleCal , can I understand it this way: The functionality of ```contains_tokens``` is equivalent to FTS MatchQuery (with fuzziness disabled, Operator::And, and using the simple tokenizer). This semantic makes sense to me. I previously implemented a version of the ```contains_tokens```(https://github.com/lancedb/lance/pull/4420), and if we ultimately decide on this semantic, the UDF will need some adjustments.

Another point is that if the functionality of contains_tokens is substring matching, then FTS might introduce false negatives. However, if it follows the version described above, there would be no ambiguity at the semantic level.

I'm also wondering: do we need to support configuring tokenizers, fuzziness, prefix-matching and operator. If we make them as params of contains_tokens, the UDF might become too bloated. I think we can consider supporting an fts() UDF in the future, where more advanced full-text search capabilities can be covered.

Hi @westonpace , there are some different understandings regarding the syntax of the ```contains_tokens``` method. Could you help take a look?

westonpace: I think there is agreement.

* `contains_tokens` can be used for "best effort filtering" and will be accelerated by an inverted (FTS) index
  * The exact definition of "best effort" is a little vague
* `contains` is for "exact filtering" and will be accelerated by ngram but not inverted index.
* FTS _search_ is a different thing (not filtering) and it would be nice to expose in the SQL but that is a different task entirely.

rust/lance-index/src/scalar/inverted/index.rs
westonpace: Yes, I think that would be a good idea.  Or `StringContains` could take a boolean parameter.

westonpace: Actually, we may need a separate query entirely :thinking: 

Right now the logic assumes "if scalar index supports query X and we can parse query X then scalar index can be used"

So if we re-use `TextQuery` that will mean that an inverted index will be chosen for both `StringContains` _and_ `TokenContains` which isn't quite right (we only want it used for `TokenContains`)

rust/lance-index/src/scalar/expression.rs
westonpace: I've opened https://github.com/lancedb/lance/issues/4520 for discussion on FTS in SQL

wojiaodoubao:
Hi @BubbleCal @westonpace , could you help review this when you have time, thanks very much !

westonpace (commented):
What is our goal?

If the goal is to approximate "string contains" with the knowledge that there may sometimes be false negatives then we should return `AtMost` (which this PR does) and use `value.contains(scalar_str.value)` (which this PR does not do).

If the goal is to create a "contains tokens" function then we shouldn't define "token" very precisely.  This PR defines it as "separated by punctuation and white space" but that is not how the tokenizer is neccesarily configured.  It might be ngrams or apply stemming, and stop words may be removed.  In this case I think we should return exact (which this PR does not do) and use the collect_tokens based filter (which this PR does do)

wojiaodoubao:
> What is our goal?

The current goal is to create a "contains tokens" function. 

> In this case I think we should return exact (which this PR does not do) and use the collect_tokens based filter (which this PR does do).

The current functionality of ```contains_tokens``` is equivalent to the following FTS bm25_search:
* Tokenizer Configuration
```
InvertedIndexParams {
  base_tokenizer: "simple",
  language: English,
  with_position: false,
  max_token_length: None,
  lower_case: false,
  stem: false,
  remove_stop_words: false,
  ascii_folding: false,
  min_ngram_length: MAX u32, // disable ngram
  max_ngram_length: MAX u32, // disable ngram
  prefix_only: false,
}
```
* fuzziness: None
* Operator: And

Before using InvertedIndex to optimize query, we must check the ```InvertedIndexParams```. If the FTS index's ```InvertedIndexParams``` can guarantee that "the query results are a superset of ```contains_tokens``` query results," then fts index is used and AtMost is returned. Otherwise, AtLeast with zero rowid is returned.

The prerequisite for returning Exact is that the FTS index's ```InvertedIndexParams``` configuration is identical to the ```contains_tokens```. I think this could be considered as an optimization.

westonpace:
> "The query results are a superset of contains_tokens" can be checked by condition below.

Ah, I missed this part, thank you!

rust/lance/src/dataset.rs
```diff
@@ -7269,4 +7272,108 @@ mod tests {
...
+            .await
+            .unwrap();
+
+        let results = execute_sql(
+            "select * from foo where contains_tokens(text, 'cat catch fish')",
+            "foo".to_string(),
+            Arc::new(dataset.clone()),
+        )
+        .await
+        .unwrap();
```
westonpace, line 7347: Can we verify with explain plan or analyze plan that it is actually using the index?

rust/lance-index/src/scalar.rs
```diff
@@ -554,6 +555,43 @@ impl AnyQuery for TextQuery {
     }
 }
 
+/// A query that a InvertedIndex can satisfy
+#[derive(Debug, Clone, PartialEq)]
+pub enum TokenQuery {
+    /// Retrieve all row ids where the text contains all tokens parsed from given string. The tokens
```
westonpace, line 561: There might be a few pieces of glue missing.  I think we need a `TokenQueryParser` (see https://github.com/lancedb/lance/blob/60711f360b7f8692df44a0e84c98c8fdff2897a3/rust/lance-index/src/scalar/expression.rs#L348 for the `TextQueryParser`).

We also need to register the token query parser here: https://github.com/lancedb/lance/blob/60711f360b7f8692df44a0e84c98c8fdff2897a3/rust/lance/src/index.rs#L1361

Can we call `is_query_allowed` in the registration function (`scalar_index_info`)?  This way we can skip the scalar index entirely if it is not eligible.  Returning `AtLeast` with zero rows might lead to bad performance (the planner will think we are doing a scalar index optimized search and make certain decisions based on that)

westonpace (commented):
A few more small tasks.  There is a bit of glue we need to connect the query to the index.  I think we will need a test verifying the index is actually used too.

wojiaodoubao:
Sorry, I accidentally deleted the comment. I'm reposting it here.

> What is our goal?

The current goal is to create a "contains tokens" function.

> In this case I think we should return exact (which this PR does not do) and use the collect_tokens based filter (which this PR does do).

The current functionality of contains_tokens is equivalent to the following FTS bm25_search:

* Tokenizer Configuration
```
InvertedIndexParams {
  base_tokenizer: "simple",
  language: English,
  with_position: false,
  max_token_length: None,
  lower_case: false,
  stem: false,
  remove_stop_words: false,
  ascii_folding: false,
  min_ngram_length: MAX u32, // disable ngram
  max_ngram_length: MAX u32, // disable ngram
  prefix_only: false,
}
```
* fuzziness: None
* Operator: And

Before using InvertedIndex to optimize query, we must check the InvertedIndexParams. If the FTS index's InvertedIndexParams can guarantee that "the query results are a superset of contains_tokens query results," then fts index is used and AtMost is returned. Otherwise, AtLeast with zero rowid is returned.

"The query results are a superset of contains_tokens" can be checked by condition below.
```
    /// Whether the query can use the current index.
    fn is_query_allowed(&self, query: &TokenQuery) -> bool {
        match query {
            TokenQuery::TokensContains(_) => {
                self.params.base_tokenizer == "simple"
                    && self.params.max_token_length.is_none()
                    && self.params.language == Language::English
                    && !self.params.stem
            }
        }
    }
```
The prerequisite for returning Exact is that the FTS index's InvertedIndexParams configuration is identical to the contains_tokens. I think this could be considered as an optimization.

rust/lance-index/src/scalar.rs
wojiaodoubao: > I think we need a TokenQueryParser

We  already have a FtsQueryParser which parses ```contains_tokens``` into TokenQuery::TokensContains. Actually you implemented it (^-^). Shall we can just rely on FtsQueryParser?

> Can we call is_query_allowed in the registration function ...

Thanks your nice suggestion, let me fix it.

rust/lance/src/dataset.rs
wojiaodoubao: Fix it.

wojiaodoubao:
Hi @westonpace , this pr is ready for review, please help when you have time ,thanks very much!

rust/lance/src/index.rs
```diff
@@ -1374,8 +1374,19 @@ impl DatasetIndexInternalExt for Dataset {
...
-                        as Box<dyn ScalarQueryParser>,
+                    ScalarIndexType::Inverted => {
+                        let fts_index =
+                            lance_index::scalar::expression::ScalarIndexLoader::load_index(
+                                self,
+                                &field.name,
+                                &index.name,
+                                &NoOpMetricsCollector,
+                            )
+                            .await?;
```
westonpace, line 1385: This is a bit unique but I think it'll be fine for now.  It means we will always load the FTS index on every query, even if it isn't used.  Maybe once https://github.com/lancedb/lance/pull/4584 we can find someway to handle is_query_allowed in the plugin (perhaps based on details) so we don't have to load an index unless we know we can use it.

westonpace (approved):
Thanks for bearing with the reviews!

Issue #3480, opened 2025-02-26
merge_insert is loading too many columns during indexed merge insert


westonpace:
We are loading all columns from the target table.  We should only need to load the join key columns and/or any columns needed for a filter.

## Background

We are migrating from an implementation of merge insert that manually manipulates streams to one that constructs a DataFusion plan and runs optimizers on that plan for optimize it.

The old code path is in `execute_uncommitted_impl`:

https://github.com/lancedb/lance/blob/94dcaf5e735480f9f61216b4e4f90358dbeac2ba/rust/lance/src/dataset/write/merge_insert.rs#L1337-L1340

While the new codepath is under `execute_uncommitted_v2`:

https://github.com/lancedb/lance/blob/94dcaf5e735480f9f61216b4e4f90358dbeac2ba/rust/lance/src/dataset/write/merge_insert.rs#L1352-L1371

## Task

**Goal**: use the new code path when `has_scalar_index=True`.

If the index is up-to-date (no unindexed fragments), the plan should look like:

```
MergeInsert: on=[key], ...
  ProjectionExec: <assign action>
    AddRowAddrExec: ...
      ScalarIndexJoinExec: on=[key] index=[key_idx]
        StreamingTableExec: ...
```

Basically, the input should "join" against the index to find which do and don't have `_rowids` (exist in the dataset). Then we can use `AddRowAddrExec` to get the addresses.

If the index is not up-to-date, then the plan should be a union of the existing plan plus the indexed one, with an additional filter to ensure we don't duplicate new rows:

```
MergeInsert: on=[key], ...
  ProjectionExec: <assign action>
    UnionExec:
      AddRowAddrExec: ...
        FilterExec: _rowid is not null
          ScalarIndexJoinExec: on=[key] index=[key_idx]
            ReplayExec: ...
      HashJoinExec: on=[key]
        LanceScan: projection=[key], row_addr=true, ...
        ReplayExec: ...
```

Note: we have an existing test for query plans here you can imitate: https://github.com/lancedb/lance/blob/94dcaf5e735480f9f61216b4e4f90358dbeac2ba/rust/lance/src/dataset/write/merge_insert.rs#L3001-L3002

## TODO

* [ ] Implement `ScalarIndexJoinExec`
* [ ] Implement `ScalarIndexJoinNode` (logical plan) with `UserDefinedLocalNodeCore` trait
[...]

PR #4500, opened 2025-08-18
feat: implement ScalarIndexJoin for merge insert optimization

wjones127:
## Summary

- Implements ScalarIndexJoinExec physical execution node that uses scalar indices (BTree/Bitmap) instead of full table scans for merge insert joins
- Adds ScalarIndexJoinNode logical plan node with DataFusion integration  
- Automatically detects when scalar indices are available and uses them to optimize merge insert operations
- Handles out-of-date index scenario by falling back to regular joins when unindexed fragments exist
- Falls back to regular joins when no suitable index is available

This addresses issue #3480 by reducing I/O overhead when performing merge insert operations on tables with scalar indices.

## Test plan

- [x] Added test_scalar_index_join_plan() with detailed plan structure validation using assert_plan_node_equals()
- [x] Added test_scalar_index_out_of_date_fallback() with comprehensive plan verification
- [x] All existing merge insert tests continue to pass (25/25) - ensuring backward compatibility
- [x] Clippy and formatting checks pass

## Key Implementation Details

**Up-to-date Index Scenario:**
- Uses ScalarIndexJoinExec for optimal performance  
- Plan shows: `ScalarIndexJoinExec: on=[key], index=[key_btree], type=[Inner]`
- Avoids loading unnecessary columns from target table
- Only uses join key columns via scalar index lookups

**Out-of-date Index Scenario:**
- Detects unindexed fragments using `dataset.unindexed_fragments(index_name)`
- Falls back to HashJoinExec with LanceRead for target table access
- Plan shows: `HashJoinExec: mode=CollectLeft, join_type=Inner` with `LanceRead: projection=[key]`
- Maintains correctness over performance when index is stale

## Plan Structure Validation

The tests now validate exact plan structures:

**Optimized Plan (with up-to-date index):**
```
MergeInsert: on=[key], when_matched=UpdateAll, when_not_matched=DoNothing, when_not_matched_by_source=Keep
  CoalescePartitionsExec
    ProjectionExec: expr=[CASE WHEN key@1 IS NOT NULL THEN 1 ELSE 0 END as action]
      RepartitionExec: partitioning=RoundRobinBatch(32), input_partitions=1
        ScalarIndexJoinExec: on=[key], index=[key_btree], type=[Inner]
          RepartitionExec: partitioning=RoundRobinBatch(32), input_partitions=1
            StreamingTableExec: partition_sizes=1, projection=[value, key]
```

**Fallback Plan (with out-of-date index):**
```
MergeInsert: on=[key], when_matched=UpdateAll, when_not_matched=DoNothing, when_not_matched_by_source=Keep
  CoalescePartitionsExec
    ProjectionExec: expr=[_rowaddr@0 as _rowaddr, value@1 as value, key@2 as key, CASE WHEN key@2 IS NOT NULL AND _rowaddr@0 IS NOT NULL THEN 1 ELSE 0 END as action]
      CoalesceBatchesExec: target_batch_size=8192
        HashJoinExec: mode=CollectLeft, join_type=Inner, on=[(key@0, key@1)], projection=[_rowaddr@1, value@2, key@3]
          LanceRead: uri=..., projection=[key], num_fragments=2, range_before=None, range_after=None, row_id=false, row_addr=true, full_filter=--, refine_filter=--
          RepartitionExec: partitioning=RoundRobinBatch(32), input_partitions=1
            StreamingTableExec: partition_sizes=1, projection=[value, key]
```

### DIFF
--- a/rust/lance/src/dataset/write/merge_insert.rs
+++ b/rust/lance/src/dataset/write/merge_insert.rs
@@ -70,7 +70,7 @@ use crate::{
     dataset::{
         fragment::{FileFragment, FragReadConfig},
         transaction::{Operation, Transaction},
-        write::{merge_insert::logical_plan::MergeInsertPlanner, open_writer},
+        write::open_writer,
     },
     index::DatasetIndexInternalExt,
     io::exec::{
@@ -1150,8 +1150,7 @@ impl MergeInsertJob {
         //       projection pushdown for us.
         // Goal: we shouldn't have to add new branches in this code to handle
         //       indexed vs non-indexed cases. That should be handled by optimizer rules.
-        let session_config = SessionConfig::default();
-        let session_ctx = SessionContext::new_with_config(session_config);
+        let session_ctx = get_session_context(&Default::default());
         let scan = session_ctx.read_lance_unordered(self.dataset.clone(), true, true)?;
         let on_cols = self
             .params
@@ -1161,19 +1160,94 @@ impl MergeInsertJob {
             .collect::<Vec<_>>();
         let source_df = session_ctx.read_one_shot(source)?;
         let source_df_aliased = source_df.alias("source")?;
-        let scan_aliased = scan.alias("target")?;
         let join_type = if self.params.insert_not_matched {
             JoinType::Right
         } else {
             JoinType::Inner
         };
         let dataset_schema: Schema = self.dataset.schema().into();
-        let df = scan_aliased
-            .join(source_df_aliased, join_type, &on_cols, &on_cols, None)?
-            .with_column(
-                MERGE_ACTION_COLUMN,
+
+        // Check if we can use a scalar index for the join
+        let df = if let Some(index) = self.join_key_as_scalar_index().await? {
+            // Use scalar index join - this avoids loading all columns from target table
+            use crate::dataset::write::merge_insert::logical_plan::ScalarIndexJoinNode;
+            use datafusion_expr::{Extension, LogicalPlan};
+
+            // Get the source logical plan
+            let (source_session_state, source_logical_plan) = source_df_aliased.into_parts();
+
+            // Create ScalarIndexJoinNode for indexed data
+            let scalar_index_join_node = ScalarIndexJoinNode::try_new(
+                source_logical_plan.clone(),
+                self.dataset.clone(),
+                self.params.on[0].clone(), // We know it's a single column from join_key_as_scalar_index
+                index.name.clone(),
+                join_type,
+                Some(datafusion::common::TableReference::bare("target")), // Qualify _rowid as target._rowid
+            )?;
+
+            let indexed_join_plan = LogicalPlan::Extension(Extension {
+                node: Arc::new(scalar_index_join_node),
+            });
+
+            // Add AddRowAddrNode to translate the _rowid to _rowaddr with proper table qualification
+            let with_rowaddr_node = logical_plan::AddRowAddrNode::try_new(
+                indexed_join_plan.clone(),
+                self.dataset.clone(),
+                indexed_join_plan.schema().fields().len(), // Add _rowaddr at the end
+            )?;
+            let with_rowaddr_plan = LogicalPlan::Extension(Extension {
+                node: Arc::new(with_rowaddr_node),
+            });
+
+            // Convert back to DataFrame for action column processing
+            let indexed_joined_df = datafusion::dataframe::DataFrame::new(
+                source_session_state.clone(),
+                with_rowaddr_plan,
+            );
+
+            // Add the action column
+            let indexed_joined_with_action_df = indexed_joined_df.with_column(
+                "action",
                 merge_insert_action(&self.params, Some(&dataset_schema))?,
             )?;
+
+            // Check for unindexed fragments (out-of-date index case)
+            let unindexed_fragments = self.dataset.unindexed_fragments(&index.name).await?;
+
+            if !unindexed_fragments.is_empty() {
+                // Handle out-of-date index: need to union indexed data with unindexed data
+                // This is more complex with ScalarIndexJoin, so for now, fall back to regular join
+                // when there are unindexed fragments (out-of-date index case)
+
+                // TODO: Implement proper union of ScalarIndexJoin with regular join for unindexed data
+                // For now, fall back to regular join when index is out of date
+                let scan_aliased = scan.alias("target")?;
+                let source_df_for_fallback = datafusion::dataframe::DataFrame::new(
+                    source_session_state,
+                    source_logical_plan,
+                )
+                .alias("source")?;
+                scan_aliased
+                    .join(source_df_for_fallback, join_type, &on_cols, &on_cols, None)?
+                    .with_column(
+                        "action",
+                        merge_insert_action(&self.params, Some(&dataset_schema))?,
+                    )?
+            } else {
+                // Index is up-to-date, use only indexed data
+                indexed_joined_with_action_df
+            }
+        } else {
+            // Fallback to regular join when no suitable scalar index exists
+            let scan_aliased = scan.alias("target")?;
+            scan_aliased
+                .join(source_df_aliased, join_type, &on_cols, &on_cols, None)?
+                .with_column(
+                    "action",
+                    merge_insert_action(&self.params, Some(&dataset_schema))?,
+                )?
+        };
 
         let (session_state, logical_plan) = df.into_parts();
 
@@ -1188,8 +1262,11 @@ impl MergeInsertJob {
 
         let logical_plan = session_state.optimize(&logical_plan)?;
 
-        let planner =
-            DefaultPhysicalPlanner::with_extension_planners(vec![Arc::new(MergeInsertPlanner {})]);
+        let planner = DefaultPhysicalPlanner::with_extension_planners(vec![
+            Arc::new(logical_plan::MergeInsertPlanner {}),
+            Arc::new(logical_plan::ScalarIndexJoinPlanner {}),
+            Arc::new(logical_plan::AddRowAddrPlanner {}),
+        ]);
         // This method already does the optimization for us.
         let physical_plan = planner
             .create_physical_plan(&logical_plan, &session_state)
@@ -1269,7 +1346,6 @@ impl MergeInsertJob {
     ///
     /// The fast path is only available for specific conditions:
     /// - when_matched is UpdateAll or UpdateIf
-    /// - No scalar index on join key
     /// - Source schema matches dataset schema exactly
     /// - when_not_matched_by_source is Keep
     async fn can_use_create_plan(&self, source_schema: &Schema) -> Result<bool> {
@@ -1284,13 +1360,10 @@ impl MergeInsertJob {
             },
         );
 
-        let has_scalar_index = self.join_key_as_scalar_index().await?.is_some();
-
         Ok(matches!(
             self.params.when_matched,
             WhenMatched::UpdateAll | WhenMatched::UpdateIf(_)
-        ) && !has_scalar_index
-            && is_full_schema
+        ) && is_full_schema
             && matches!(
                 self.params.delete_not_matched_by_source,
                 WhenNotMatchedBySource::Keep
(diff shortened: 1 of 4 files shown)


wjones127:
# Benchmark: Upsert 2GB into 10GB of data

<details><summary>Script</summary>
<p>

Create data:

```python
import time
import pyarrow as pa
import pyarrow.compute as pc
import lance

DIMS = 1024
BYTES_PER_ROW = 4 * DIMS + 8 + 8  # 4 bytes per float, 8 bytes for int64 id, 8 bytes for int64 value
TARGET_SIZE = 10 * 1024 * 1024 * 1024  # 10 GB
NROWS = TARGET_SIZE // BYTES_PER_ROW

data = pa.table({
    'id': range(NROWS),
    'vector': pa.FixedSizeListArray.from_arrays(
        pc.random(NROWS * DIMS).cast(pa.float32()), DIMS
    ),
    'value': [0] * NROWS
})

ds = lance.write_dataset(data, "test_lance", mode="overwrite")

ds.create_scalar_index("id", "BTREE")
```

benchmark:

```python
import time
import pyarrow as pa
import pyarrow.compute as pc
import lance

DIMS = 1024
BYTES_PER_ROW = 4 * DIMS + 8 + 8  # 4 bytes per float, 8 bytes for int64 id, 8 bytes for int64 value
TARGET_SIZE = 2 * 1024 * 1024 * 1024  # 2 GB
NROWS = TARGET_SIZE // BYTES_PER_ROW

BATCH_SIZE = 20 * 1024 * 1024  # 20 MB
NROWS_BATCH = BATCH_SIZE // BYTES_PER_ROW

ds = lance.dataset("test_lance")

# Half existing ids, half new ids
start_id = ds.count_rows() - (NROWS // 2)

def gen_data():
    rows_output = 0
    while True:
        nrows = min(NROWS_BATCH, NROWS - rows_output)
        if nrows <= 0:
            break
        rows_output += nrows

        yield from pa.table({
            'id': range(rows_output - nrows, rows_output),
            'vector': pa.FixedSizeListArray.from_arrays(
                pc.random(nrows * DIMS).cast(pa.float32()), DIMS
            ),
            'value': [1] * nrows
        }).to_batches()

schema = pa.schema({
    'id': pa.int64(),
    'vector': pa.list_(pa.float32(), DIMS),
    'value': pa.int64()
})

new_data = pa.RecordBatchReader.from_batches(
    schema,
    gen_data()
)

(
    ds.merge_insert('id')
        .when_matched_update_all()
        .execute(new_data)
)
```

Run:

```shell
maturin develop --profile release-with-debug
python make_dataset.py
export LANCE_MEM_POOL_SIZE=$((1024 * 1024 * 1024 * 20))
export MEMORY_PROFILER_LOG=info
MEMORY_PROFILER_USE_SHADOW_STACK=0 LD_PRELOAD=/usr/local/lib/libbytehound.so python bench_merge_insert.py
```

</p>
</details> 

### Before: 4.3 GB peak memory use, errors by default

```shell
time python bench_merge_insert.py
```

```
OSError: LanceError(IO): Resources exhausted: Failed to allocate additional 20.6 MB for HashJoinInput with 100.0 MB already allocated for this reservation - 1600.0 B remain available for the total pool, /home/will/.rustup/toolchains/1.89.0-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/task/poll.rs:290:44
```

```shell
export LANCE_MEM_POOL_SIZE=$((1024 * 1024 * 1024 * 20))
time python bench_merge_insert.py
```

```
real    0m7.557s
user    0m4.818s
sys     0m6.639s
```

Screenshot 2025-08-19 at 5 24 53 PM


### After: 2x faster, 2GB peak memory use, memory use *can* be bounded with fewer threads

```shell
time python bench_merge_insert.py
```

```
real    0m3.820s
user    0m2.511s
sys     0m4.767s
```

Memory use seems to be tied to number of cores used by DataFusion. If I hard-code `target_partitions(1)`, then I get bounded memory use:

Screenshot 2025-08-21 at 3 15 50 PM


If I use default partitions tied to number of cores (32 on my machine), I get:

Screenshot 2025-08-21 at 3 15 36 PM

wjones127:
Not a major blocker, but this is a performance bottleneck I noticed: https://github.com/lancedb/lance/issues/4523

wjones127:
TODO:

* [x] Figure out why memory use is high
* [x] Make sure we use multiple cores in the ScalarIndexJoin
* [x] Benchmark speed improvement.

rust/lance/src/dataset/write/merge_insert.rs
```diff
@@ -1154,20 +1154,95 @@ impl MergeInsertJob {
...
         let dataset_schema: Schema = self.dataset.schema().into();
-        let df = scan_aliased
-            .join(source_df_aliased, join_type, &on_cols, &on_cols, None)?
-            .with_column(
+
+        // Check if we can use a scalar index for the join
+        let df = if let Some(index) = self.join_key_as_scalar_index().await? {
+            // Use scalar index join - this avoids loading all columns from target table
+            use crate::dataset::write::merge_insert::logical_plan::ScalarIndexJoinNode;
+            use datafusion_expr::{Extension, LogicalPlan};
```
wkalt, line 1174: pull up to top level?

rust/lance/src/io/exec/scalar_index_join.rs
```diff
@@ -0,0 +1,458 @@
...
+                SearchResult::Exact(row_ids) => {
+                    if row_ids.is_empty() {
+                        None
+                    } else {
+                        // For merge insert, we expect at most one match per key
+                        // If there are multiple, take the first one
+                        row_ids
+                            .row_ids()
+                            .and_then(|mut iter| iter.next())
+                            .map(u64::from)
```
wkalt, line 224: I think we've discussed before but this seems like a horrible situation to me. I almost wonder if this should be an error. If this happens then the user's data is almost certainly corrupt and the merge insert will worsen the corruption.

wkalt:
@wjones127 I think the after chart is missing

rust/lance/src/dataset/write/merge_insert/logical_plan.rs
```diff
@@ -206,3 +210,379 @@ impl ExtensionPlanner for MergeInsertPlanner {
...
+            join_type,
+            table_reference,
+            schema,
+        })
+    }
+}
+
+impl UserDefinedLogicalNodeCore for ScalarIndexJoinNode {
+    fn name(&self) -> &str {
+        "ScalarIndexJoin"
```
wkalt, line 333: should we call this NestedLoopJoin?

Not a huge deal but I think "ScalarIndexJoin" might not stand up if we introduce other join strategies in the future.

wkalt: also if it were "NestedLoopJoin", would that have applicability outside of merge_insert?

rust/lance/src/io/exec/scalar_index_join.rs
```diff
@@ -0,0 +1,458 @@
...
+            index_name: self.index_name.clone(),
+            join_type: self.join_type,
+            scalar_index: OnceLock::new(), // Don't clone the cached index
+            output_schema: self.output_schema.clone(),
+            properties: self.properties.clone(),
+        }
+    }
+}
+
+// Tests will be added later once the core implementation is working
```
wkalt, line 458: remove unless planning to add more

rust/lance/src/io/exec/scalar_index_join.rs
```diff
@@ -0,0 +1,458 @@
...
+
+        Ok(Box::pin(RecordBatchStreamAdapter::new(
+            self.output_schema.clone(),
+            Box::pin(output_stream),
+        )))
+    }
+
+    fn statistics(&self) -> datafusion::error::Result<Statistics> {
+        // For now, return unknown statistics
+        // In the future, we could use index statistics to provide better estimates
```
wkalt, line 429: seems like we can provide some of these without index stats right? This feeds into analyze_plan I think.

rust/lance/src/io/exec/scalar_index_join.rs
```diff
@@ -0,0 +1,458 @@
...
+            .dataset
+            .open_scalar_index(
+                &self.join_column,
+                &idx.uuid.to_string(),
+                &NoOpMetricsCollector,
+            )
+            .await?;
+
+        // Cache it
+        let _ = self.scalar_index.set(index);
```
wkalt, line 188: are indexes cached elsewhere in lance? How big could this get if doing concurrent merge inserts?

wkalt:
sorry, I followed a link and started reading this patch. Just noticed it's still marked draft.

rust/lance/src/io/exec/scalar_index_join.rs
wjones127: That's reasonable. It does error by default in Databricks [^1] and Snowflake [^2].

So we can start to enforce that in this PR. We can enforce that in other code paths as a follow up:
* https://github.com/lancedb/lance/issues/4535

[^1]: https://docs.databricks.com/aws/en/sql/language-manual/delta-merge-into#:~:text=MERGE%20operations%20fail%20with%20a%20DELTA_MULTIPLE_SOURCE_ROW_MATCHING_TARGET_ROW_IN_MERGE
[^2]: https://docs.snowflake.com/en/sql-reference/sql/merge#nondeterministic-results-for-update-and-delete

rust/lance/src/dataset/write/merge_insert/logical_plan.rs
wjones127: Why makes you say this is a `NestedLoopJoin`? My mental model of nested loop join is we do `O(sizeof(left) * sizeof(right))` comparisons. Whereas this join does `O(sizeof(input))` index lookups. That seems fundamentally different, right?

And what other index-powered joins are there? I'm down to change the name if there are others we can think of. What does postgres have?

wjones127: I think part of the weirdness is that this join is really against the index, not the table it's on. It doesn't project anything from the left side except the `_rowid` in the index.

wkalt: I didn't look super close but it seemed like you were iterating over the input rows and probing the index? Nested loop join is performed with an index so it's O(M * log N), it looked like what you were doing to me.

Merge joins can also benefit from an index when the join key has an index on each input.

```
I think part of the weirdness is that this join is really against the index, not the table it's on.
```
I haven't really read closely enough to have an opinion here, you may be right about that.

wkalt: for the join against the index part, the way postgres represents it is like this:
```
[local]:wyatt=# create table foo as select i from generate_series(1, 100000) i;
SELECT 100000
Time: 29.771 ms
[local]:wyatt=# create table bar as select i from generate_series(1, 1000) i;
SELECT 1000
Time: 14.037 ms
[local]:wyatt=# create index on foo(i);
CREATE INDEX
Time: 22.590 ms
[local]:wyatt=# create index on bar(i);
CREATE INDEX
Time: 7.830 ms
[local]:wyatt=# explain analyze select * from foo inner join bar on foo.i = bar.i;
                                                         QUERY PLAN                                                          
-----------------------------------------------------------------------------------------------------------------------------
 Nested Loop  (cost=0.29..1439.00 rows=1000 width=8) (actual time=0.019..0.558 rows=1000 loops=1)
   ->  Seq Scan on bar  (cost=0.00..15.00 rows=1000 width=4) (actual time=0.004..0.038 rows=1000 loops=1)
   ->  Index Only Scan using foo_i_idx on foo  (cost=0.29..1.41 rows=1 width=4) (actual time=0.000..0.000 rows=1 loops=1000)
         Index Cond: (i = bar.i)
         Heap Fetches: 0
 Planning Time: 0.154 ms
 Execution Time: 0.588 ms
(7 rows)

Time: 0.950 ms

```

The index only scan can be used whenever the index contains all required columns. Multicolumn indexes can serve it too.

rust/lance/src/dataset/write/merge_insert.rs
```diff
@@ -1150,8 +1150,7 @@ impl MergeInsertJob {
         //       projection pushdown for us.
         // Goal: we shouldn't have to add new branches in this code to handle
         //       indexed vs non-indexed cases. That should be handled by optimizer rules.
-        let session_config = SessionConfig::default();
-        let session_ctx = SessionContext::new_with_config(session_config);
+        let session_ctx = get_session_context(&Default::default());
```
cmccabe, line 1153: Hmm, does this result in a different session context than before, given that we're using Default::default?

rust/lance/src/dataset/write/merge_insert.rs
```diff
@@ -1161,20 +1160,95 @@ impl MergeInsertJob {
...
+                // Index is up-to-date, use only indexed data
+                indexed_joined_with_action_df
+            }
+        } else {
+            // Fallback to regular join when no suitable scalar index exists
+            let scan_aliased = scan.alias("target")?;
+            scan_aliased
+                .join(source_df_aliased, join_type, &on_cols, &on_cols, None)?
+                .with_column(
+                    "action",
```
cmccabe, line 1247: Hmm... Why is this now `action` when it was previously `MERGE_ACTION_COLUMN` which maps to `__action`? It seems that we want the double underscores to avoid conflicts?

rust/lance/src/io/exec/scalar_index_join.rs
jbapple: I'm guessing `BTreeIndex::index_cache` would not count as an answer to your question?

PR #4501, opened 2025-08-18
[WIP] Feat: Add scan tracing events when run with datafusion ffi

jaystarshot:


jaystarshot:
cc: @westonpace
When I run with DataFusion now, I see logs showing metrics like bytes read and I/O. Just wanted to check if this is the right approach. (By wrapping lance plan and making sure that our wrapper can collect metrics)

rust/lance/src/datafusion/dataframe.rs
```diff
@@ -124,7 +137,10 @@ impl TableProvider for LanceTableProvider {
         scan.limit(limit.map(|l| l as i64), None)?;
         scan.scan_in_order(self.ordered);
 
-        scan.create_plan().await.map_err(DataFusionError::from)
+        let plan = scan.create_plan().await.map_err(DataFusionError::from)?;
+        
+        // Wrap the plan with tracing to emit lance::execution events
+        Ok(Arc::new(TracedLanceExec::new(plan)))
```
jackye1995: I think there should be a way to directly leverage `report_plan_summary_metrics`. Can you describe what the exact situation that is not emitting metrics? (maybe describe it as a test so I can try to reproduce and play around?)

jaystarshot: Thanks @jackye1995! What I was actually looking for were the metrics emitted by tracer **lance::execution::plan_stats** in non-DataFusion scenarios. 
These traces aren't available when using DataFusion. Also, EXPLAIN ANALYZE on DataFusion doesn't expose any metrics because they're hidden behind FFIScan we need to ensure metrics are properly reported back to DataFusion.

With this change, I’m now able to capture them and even report custom traces to Python, for example

**iops: 127, requests: 55, bytes_read: 1,390,916,548 (1326.5MB), rows_scanned: 3296**

for most cases

jaystarshot:
Updated my PR, now the ouput i get is below for eg: With traces i am unable to get this (via datafusion)

```
=== FilteredReadExec Analysis ===
FilteredReadExec Analysis:
  I/O Stats:
    - Data Read: 892.3 MB (935687724 bytes)
    - I/O Operations: 75 iops across 27 requests
    - Avg Request Size: 33.0 MB
    - Rows Scanned: 543 rows
  Scan Efficiency:
    - Fragments Scanned: 5 of total
    - Ranges Scanned: 1
    - Data per Row: 1682.8 KB
=================================
```

jaystarshot:
It would be great if we could somehow populate this to datafusion explain analyze/ new tracer

westonpace:
Does https://github.com/lancedb/lance/pull/4479 help?  I'll look closer this evening in a few hours.

jaystarshot:
Nope unfortunately ( i tested after cherry picking).  The problem is that datafusion is bypassing that exec() call.

> let plan = scan.create_plan().await.map_err(DataFusionError::from)?;
> 
>     // Wrap the plan with tracing to emit lance::execution events
>     Ok(Arc::new(TracedLanceExec::new(plan)))

The tracer added here is run after execution completes so in a way it is doing https://github.com/lancedb/lance/pull/4479 too

westonpace:
> Nope unfortunately ( i tested after cherry picking). The problem is that datafusion is bypassing that exec() call.

Got it, yes, if you're using DF directly and bypassing lance for the query execution then this will happen.  I actually started down this road and then scrapped it after I had written it.

The problem I ran into is that some query plans will scan more than one Lance table.  For example, a query plan might be issuing a join.  In that case we will get output metrics for each scan of a Lance table.  I really wanted to have a single report at the end of the plan.

That's when I realized that the only real root point for the query will be whatever code is actually executing the plan (which we don't have any control over)

I'm not terribly opposed to this approach as a compromise but it should probably be opt-in and wouldn't we want to use tracing instead of println?

jaystarshot:
Thanks a lot! Makes sense i will clean it up and add a tracer.

PR #4504, opened 2025-08-19
feat: introduce optimize columns for dataset

geruh:
As a part of discussion [#3995](https://github.com/lancedb/lance/discussions/3995)

This PR introduces a new optimize feature called `optimize_columns` that reorganizes data files within existing fragments to group columns together.  Unlike vertical compaction (compact_files) which merges multiple fragments together, this performs 1:1 fragment transformations where each fragment gets the data files reorganized by columns without changing the total number of fragments.

> Note: This is a prototype implementation. The final version will follow a pattern closer to merge operations to replace data files within fragments, avoiding complex index handling. Also will move all logic into it's own separate file. 

#### Multiple Group Merging
**Input**
- Fragment: `[file(c1), file(c2), file(c3), file(c4)]`
- Groups: `[group(c1, c2), group(c3, c4)]`

**Output**
- Fragment: `[file(c1, c2), file(c3, c4)]`

#### File Splitting
**Input**
- Fragment: `[file(c1), file(c2, c3), file(c4)]`
- Groups: `[group(c1, c2), group(c3, c4)]`

**Output**
- Fragment: `[file(c1, c2), file(c3, c4)]`

### API
```rust
pub async fn optimize_columns(dataset,  options,) -> Result<Metrics>
```

I added a new `optimize_columns` function that lets you specify how you want columns grouped together:

```rust
let options = OptimizeColumnOptions {
    column_groups: Some(vec![
        ColumnGroup::new(vec!["id".to_string(), "data".to_string()]),
    ]),
    num_threads: None,
    defer_index_remap: false,
};

  let metrics = optimize_columns(&mut dataset, options).await?;
```

## implementation

The implementation follows a similar approach to `compact_files` but we aren't compacting fragments vertically but just horizontally then doing a swap with the new fragment containing the reorganized files. 

``` rust
  // 1. Collect plan for what fragments need changing
  let plan = build_optimize_columns_plan(fragment, &column_groups, schema)?;

  // 2. Write out the new data files
  let result = execute_fragment_optimize_columns(dataset, fragment, plan).await?;

  // 3. Commit changes to dataset
  commit_optimize_columns(dataset, results).await?;
```

## How it actually works under the hood

The current approach scans each fragment once and loads the data, then projects different column subsets to write multiple data files in parallel.

```rust
  // for each column group, project the columns we need
  // and write multiple output files in parallel from the same data
  let write_futures: Vec<_> = plan.write_tasks.iter()
      .map(|task| {
          let projected_batches = batches.iter()
              .map(|batch| batch.project(&task.projected_field_indices))
              .collect();

          write_single_data_file(task, &projected_batches, dataset)
      })
      .collect();
```

Keeping this as a draft, while I add more correctness tests and for feedback on the direction!

### DIFF
--- a/rust/lance/src/dataset/optimize.rs
+++ b/rust/lance/src/dataset/optimize.rs
@@ -987,7 +987,7 @@ impl ColumnGroup {
     }
 }
 
-/// Configuration options for horizontal column compaction.
+/// Configuration options for Optimize columns.
 ///
 /// This struct controls how columns are regrouped within fragments.
 #[derive(Debug, Clone, Default)]
@@ -1448,7 +1448,7 @@ fn construct_new_fragment(
     }
 }
 
-/// Commit horizontal compaction results in batched transaction
+/// Commit optimize columns results in batched transaction
 async fn commit_optimize_columns(
     dataset: &mut Dataset,
     completed_results: Vec<OptimizeColumnsResult>,


rust/lance/src/dataset/optimize.rs
```diff
@@ -968,6 +975,567 @@ pub async fn commit_compaction(
...
+
+/// Internal plan for optimizing a single fragment.
+#[derive(Debug, Clone)]
+struct FragmentOptimizeColumnsPlan {
+    /// Existing data files that exactly match requested column groups.
+    kept_files: Vec<DataFile>,
+    /// New data files that need to be created
+    write_tasks: Vec<DataFileWriteTask>,
+    /// Whether to preserve deletion files and include deleted rows in scans.
+    preserve_deletions: bool,
```
geruh, line 1023: Here I'm preserving the scan deletions if we can skip files, but if we always preserve and persist the delete file we can reduce this complexity

rust/lance/src/dataset/optimize.rs
```diff
@@ -968,6 +975,567 @@ pub async fn commit_compaction(
...
+/// Sets up row ID capture stream if needed for legacy row IDs
+async fn setup_row_id_capture(
+    mut scanner: Scanner,
+    needs_row_id_capture: bool,
+    uses_stable_row_ids: bool,
+) -> Result<(
+    Option<std::sync::mpsc::Receiver<CapturedRowIds>>,
+    SendableRecordBatchStream,
+)> {
+    if needs_row_id_capture {
```
geruh, line 1348: We can avoid with in place fragment changes similar to merge operations

protos/transaction.proto
```diff
@@ -202,6 +202,26 @@ message Transaction {
     repeated MemWalIndexDetails.MemWal removed = 3;
   }
 
+  message OptimizeColumns {
```
jackye1995, line 205: this is basically the same as RewriteGroup, why not just reuse that?

rust/lance/src/dataset/optimize.rs
```diff
@@ -968,6 +975,567 @@ pub async fn commit_compaction(
     Ok(metrics)
 }
 
+/// A group of columns that should be stored together
+#[derive(Debug, Clone)]
+pub struct ColumnGroup {
```
jackye1995, line 980: nit: based on https://github.com/lancedb/lance/discussions/3995#discussioncomment-14038095, probably want to call this a ColumnFamily

jackye1995: actually, since this structure is just used for OptimizeColumnOptions, why not just make `column_groups` as `Option<Vec<Vec<String>>>`

rust/lance/src/dataset/optimize.rs
```diff
@@ -968,6 +975,567 @@ pub async fn commit_compaction(
...
+    /// Each [`ColumnGroup`] will result in one data file per fragment.
+    /// If `None`, all columns will be placed in a single file (the default).
+    pub column_groups: Option<Vec<ColumnGroup>>,
+    /// The number of threads to use (how many compaction tasks to run in parallel).
+    /// Defaults to the number of compute-intensive CPUs.
+    pub num_threads: Option<usize>,
+    /// Whether to defer remapping indices during operation. If true, indices will
+    /// not be remapped during this operation. Instead, the fragment reuse index
+    /// is updated and will be used to perform remapping later.
+    pub defer_index_remap: bool,
```
jackye1995, line 1005: I don't think this would remap index anyway since we are not going to change the fragment ID or row position in the fragment, an index should remain useable even after the optimize_columns optimization

rust/lance/src/dataset/optimize.rs
jackye1995: I think we require in-place fragment change for this feature, we don't want to create a new fragment which will make all related index entries invalid.

rust/lance/src/dataset/optimize.rs
```diff
@@ -3420,4 +3988,419 @@ mod tests {
...
+        };
+
+        optimize_columns(&mut dataset, options).await.unwrap();
+        let new_count = dataset.count_rows(None).await.unwrap();
+        assert_eq!(original_count, new_count);
+    }
+
+    #[tokio::test]
+    async fn test_guarantee_column_grouping_correctness() {
+        let test_dir = tempdir().unwrap();
```
jackye1995, line 4082: check usage of `lance_datagen::gen_batch()` for generating testing data

rust/lance/src/dataset/optimize.rs
```diff
@@ -968,6 +975,567 @@ pub async fn commit_compaction(
...
+
+/// Configuration options for Optimize columns.
+///
+/// This struct controls how columns are regrouped within fragments.
+#[derive(Debug, Clone, Default)]
+pub struct OptimizeColumnOptions {
+    /// Specifies how columns should be grouped into files.
+    /// Each [`ColumnGroup`] will result in one data file per fragment.
+    /// If `None`, all columns will be placed in a single file (the default).
+    pub column_groups: Option<Vec<ColumnGroup>>,
```
jackye1995, line 998: we should just call this something like `optimize_groups`, try to avoid confusion once we introduce ColumnFamily actually as a part of the table metadata manifest.

rust/lance/src/dataset/optimize.rs
jackye1995: looks like the current logic is mostly the same as compact_files, which we should not use, as discussed below we should try to do in-place change in the same fragment rather than creating new fragments. This should also go away after we go with that direction, since we will always preserve deletion as the deletion file is not rewritten in this horizontal compaction.

rust/lance/src/dataset/optimize.rs
```diff
@@ -968,6 +975,567 @@ pub async fn commit_compaction(
...
+    let data_storage_version = dataset
+        .manifest()
+        .data_storage_format
+        .lance_file_version()?;
+
+    // Project schema for this file
+    let projected_schema = dataset.schema().project_by_ids(&task.field_ids, true);
+
+    // Project batches to include only needed columns
+    let projected_batches: Vec<RecordBatch> = all_batches
```
geruh, line 1393: I think we should build a fan out streamer for n column groups. So if in the future we want to allow column families in the write groups the write generator can support this

PR #4505, opened 2025-08-19
perf: use heap to maintain topk results

BubbleCal:
for partition with ~8K rows, the ANN index takes only 39.356µs to calculate the distances, but sorting the candidates by distances takes 132.739µs, it's even worse with larger partition, so use heap to reduce the overhead, which takes only 10.018µs


Xuanwo (approved):
Nice

PR #4506, opened 2025-08-19
Upgrade rust 1.89

BubbleCal:


westonpace:
Previously our MSRV policy has been to follow Datafusion which has a 3ish-months-after-release policy (they are on 1.86)

This would be a pretty big departure (1.89 released less than 2 weeks ago).  Are we able to use unstable features in 1.86 or are the AVX512 features buggy in that release?

wjones127:
I'm good to bump the `rust-toolchain.toml` version, but I'd rather not bump the MSRV. I bet that will break several of our Rust users.

Could we put these features behind an `avx512` feature flag? Then at least we can keep the MSRV  working without that feature flag.

Or, even better, re-use the dynamic dispatch stuff we have with `fp16kernels` in `lance-linalg`. Would that work for this use case? I think it would be cool to support advanced SIMD without requiring special builds. https://github.com/lancedb/lance/blob/main/rust/lance-linalg/build.rs

BubbleCal:
> I'm good to bump the `rust-toolchain.toml` version, but I'd rather not bump the MSRV. I bet that will break several of our Rust users.
> 
> Could we put these features behind an `avx512` feature flag? Then at least we can keep the MSRV working without that feature flag.
> 
> Or, even better, re-use the dynamic dispatch stuff we have with `fp16kernels` in `lance-linalg`. Would that work for this use case? I think it would be cool to support advanced SIMD without requiring special builds. https://github.com/lancedb/lance/blob/main/rust/lance-linalg/build.rs

it looks doable, the AVX512 target feature is just [stablized in rust 1.89.0](https://blog.rust-lang.org/2025/08/07/Rust-1.89.0/#more-x86-target-features), but I think the RabitQ AVX512 version is simple enough so I can write it in C

Xuanwo (approved):

PR #4509, opened 2025-08-19
feat!: add storage options to wrapping object store

wkalt:
Adds storage options to the wrap method of the WrappingObjectStore trait. This information may be necessary for the configuration of some kinds of object stores - for instance caching stores may care about azure storage account names.

Also exposes the cache_key trait method of ObjectStorageProvider, to enable custom storage providers to define their own methods of constructing keys.

### DIFF
--- a/rust/lance-io/src/object_store/providers.rs
+++ b/rust/lance-io/src/object_store/providers.rs
@@ -40,6 +40,25 @@ pub trait ObjectStoreProvider: std::fmt::Debug + Sync + Send {
             Error::invalid_input(format!("Invalid path in URL: {}", url.path()), location!())
         })
     }
+
+    /// Generate a cache URL for this provider.
+    ///
+    /// Providers can override this to implement custom cache key generation
+    /// that takes into account provider-specific requirements like namespace
+    /// isolation.
+    fn cache_url(&self, url: &Url, _params: &ObjectStoreParams) -> String {
+        if ["file", "file-object-store", "memory"].contains(&url.scheme()) {
+            // For file URLs, cache the URL without the path.
+            // The path can be different for different object stores,
+            // but we want to cache the object store itself.
+            format!("{}://", url.scheme())
+        } else {
+            // Bucket is parsed as domain, so drop the path.
+            let mut url = url.clone();
+            url.set_path("");
+            url.to_string()
+        }
+    }
 }
 
 /// A registry of object store providers.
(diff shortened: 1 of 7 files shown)


rust/lance-io/src/object_store/providers.rs
```diff
@@ -40,6 +40,25 @@ pub trait ObjectStoreProvider: std::fmt::Debug + Sync + Send {
             Error::invalid_input(format!("Invalid path in URL: {}", url.path()), location!())
         })
     }
+
+    /// Generate a cache URL for this provider.
+    ///
+    /// Providers can override this to implement custom cache key generation
+    /// that takes into account provider-specific requirements like namespace
+    /// isolation.
+    fn cache_url(&self, url: &Url, _params: &ObjectStoreParams) -> String {
```
wjones127: I'm confused as to the intention. The cache we use in Lance is keyed on `(Url, &ObjectStoreParams)`.

Putting object store params into the "cache_url" duplicates information from params, right?

wkalt: I have a custom store that implements "namespacing" on local disk using a directory prefix. In order to mock Azure's behavior, I need to exclude that directory prefix from the cache key.

wjones127: Could you add some concrete example to the doc string? I think anyone who looks at these lines of code is going to have the same confusion I have:

```rust
        let cache_path = provider.cache_url(&base_path, params);
        let cache_key = (cache_path, params.clone());
```

IIUC you use case, you want to be able to do:

```rust
let params = ObjectStoreParams {
   storage_options: HashMap::from(["prefix": "/prefix/"]),
};
let url = MyStoreProvider.cache_url("file://prefix/path/", &params);
assert_eq!("url", "file://prefix");
```

Is that it, basically?

wkalt: that's actually not quite it, and I'm wondering now if there is a deeper issue with how these URLs are getting used/cached.

I have a custom storage provider that mock's Azure's behavior, which allows you to have identically-named databases and tables in two different storage accounts. I've implemented this with prefixes similar to what you describe here. What I described above is not actually accurate - the prefix is in the storage options and in the cache key.

My URL scheme looks like this:
```
mock-az://storage/<database>/... {"azure_storage_account_name": "xyz"}
```

which maps to a file like this:
```
<root>/xyz/<database>
```
root would generally be a tmpdir. Each database in my context gets a different object store.

When I run requests against multiple databases/namespaces, what happens is the cache_url turns into "mock-az://storage" and collides across all my instances. My customization is actually adding the database, not the storage account, to the cache key.

westonpace (approved):

PR #4513, opened 2025-08-20
fix: fix FTS index cache underutilization due to DeepSizeOf overestimation

### LINKED ISSUE #4512: FTS index cache severely underutilized due to CompressedPostingList DeepSizeOf overestimation
## Description

The Full-Text Search (FTS) index cache is severely underutilized due to the `CompressedPostingList::DeepSizeOf` implementation incorrectly reporting memory usage. This causes the cache to evict entries prematurely, resulting in only ~5% of the configured cache capacity being used.

## The Bug

The `DeepSizeOf` implementation for `CompressedPostingList` was using Arrow's `get_array_memory_size()` which returns the **allocated buffer size** rather than the **actual data size**. Arrow pre-allocates buffers in power-of-2 chunks for efficiency, which can be significantly larger than the actual data.

### Impact

- **Cache Capacity**: Configured for 300,000 entries but only caches ~14,000 entries (4.7% utilization)
- **Memory Overestimation**: Each posting list reported as 444KB instead of actual ~12 bytes
- **Overestimation Factor**: 37,920x larger than actual data
- **Performance**: Excessive cache misses leading to repeated I/O operations

## Root Cause

In `rust/lance-index/src/scalar/inverted/index.rs`, the `DeepSizeOf` implementation was:

```rust
impl DeepSizeOf for CompressedPostingList {
    fn deep_size_of_children(&self, _context: &mut deepsize::Context) -> usize {
        self.blocks.get_array_memory_size() + 
        self.positions.as_ref()
            .map(|positions| positions.get_array_memory_size())
            .unwrap_or(0)
    }
}
```

The `get_array_memory_size()` returns Arrow's internal buffer allocations, not the actual compressed data size.

## The Fix

Calculate the actual size of compressed data by iterating through blocks and summing their lengths:

```rust
impl DeepSizeOf for CompressedPostingList {
    fn deep_size_of_children(&self, _context: &mut deepsize::Context) -> usize {
        let mut blocks_actual_size = 0;
        for i in 0..self.blocks.len() {
            blocks_actual_size += self.blocks.value(i).len();
        }
        
        let positions_actual_size = self.positions.as_ref()
            .map(|positions| {
                positions.len() * positions.values().len() * std::mem::size_of::<i32>() 
                    / positions.len().max(1)
            })
            .unwrap_or(0);
        
        blocks_actual_size + positions_actual_size
    }
}
```

## Verification

After the fix:
- Cache now properly utilizes configured capacity
- Memory reporting is accurate (actual data size + struct overhead)
- No more premature evictions due to overestimated sizes

## Performance Impact

This fix only affects cache weight calculation which happens once during insertion. The calculation itself is minimal overhead compared to the I/O savings from proper cache utilization.

LuQQiu:
Fixes #4512 
In local testing, Arrow.get_array_memory_size() for LargeBinaryArray(MutableBuffer) will reserve large virtual memory space and report that space which is much bigger than actual memory usage, can result in tens to hundreds times bigger than actual memory usage.
This PR switch to calculate the actual data size + some estimated struct overhead. This PR may result in cache size underestimation.
Cache reporting: 57.67MB for 8343 entries (7KB per entry)
Process memory: ~200MB RSS (133MB for consisntent hashing ring)
Original reporting: Would have been 3.6GB (455KB per entry)

### DIFF
--- a/rust/lance-arrow/src/deepcopy.rs
+++ b/rust/lance-arrow/src/deepcopy.rs
@@ -97,7 +97,8 @@ pub fn deep_copy_batch_sliced(batch: &RecordBatch) -> crate::Result<RecordBatch>
 pub mod tests {
     use std::sync::Arc;
 
-    use arrow_array::{Array, Int32Array};
+    use arrow_array::{Array, Int32Array, RecordBatch, StringArray};
+    use arrow_schema::{DataType, Field, Schema};
 
     #[test]
     fn test_deep_copy_sliced_array_with_nulls() {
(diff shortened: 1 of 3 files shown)


wjones127:
I think this is only a partial solution.

I bet the underlying problem is this
1. The posting list is created by slicing Arrow RecordBatches (remember that slices create zero-copy views into original data):
    https://github.com/lancedb/lance/blob/204fb9f8d30c8c87dfb4da72acfc6ab23f006493/rust/lance-index/src/scalar/inverted/index.rs#L1016-L1022
2. The `get_array_memory_size` doesn't account for slicing.

This presents a bit of a conundrum for the index cache.

post list cache drawio

First, consider when we have all the post lists:

* Currently, each posting list reports the total memory used by the batch. So for all three posting lists, this reports `3 * sizeof(Arrow Data)`, even though we actually only have `sizeof(Arrow Data)` ❌ 
* After your change, each posting list will report the size of it's slice of the batch. So the total will be `sizeof(Arrow Data)`, which is correct 👍 

But now consider when we evict 2 / 3 of the posting lists from the cache:

* Currently, the posting list will report the total memory used by the batch, not accounting for the slicing. So this reports `sizeof(Arrow Data)`, which is is correct 👍 
* After your change, the posting list will report the size of just the slice, even though the rest of the batch is taking up memory. It will report `sizeof(Arrow Data) / 3`, which is incorrect ❌ 

So your change works when all of the FTS index is in cache, but once some posting lists start to get evicted, it will **underreport memory usage in the cache**.

I think the real solution should be:

1. Revert the change to the `DeepSizeOf` impl
2. Call [shrink_to_fit](https://docs.rs/arrow-array/56.0.0/arrow_array/array/trait.Array.html#method.shrink_to_fit) after calling slice() on the postings list. It's only implemented for arrays so we probably need a helper method for recordbatch just like we do with take: https://github.com/lancedb/lance/blob/eeea03c21e9c5082a1699c5eee3c14dd3d0ef60a/rust/lance-arrow/src/lib.rs#L752-L756

Xuanwo:
Hi, I believe this should be fixed in https://github.com/lancedb/lance/pull/4515

LuQQiu:
@wjones127 Thanks for the info. I tried the fix in https://github.com/lancedb/lance/pull/4515 and it reports the same over estimated size as before. Would take a deeper look into the issue

LuQQiu:
@wjones127  change to use deep copy, could you take another look?

LuQQiu:
Test with the cluster, the actual memory usage is much bigger than the index cache size
Node memory usage 192GiB, index cache is show as 65GiB
Around 20 million entries

PR #4515, opened 2025-08-20
fix: shrink_to_fit after from_batch to ensure memory size is correct

Xuanwo:
Fix https://github.com/lancedb/lance/issues/4512

We will alloc more spaces to avoid re-alloc during writing by `LargeBinaryBuilder::with_capacity(_, length * 3)` but this can result in our memory report in wrong. As a fix, we will call `shrink_to_fit` to ensure the memory size fits to the actual size instead.


Xuanwo:
shrint_to_fit can't fully address this issue, I will post my findings in the issue instead.

PR #4525, opened 2025-08-21
fix: full text search BM25 score is not accurate

BubbleCal:


rust/lance/src/io/exec/fts.rs
```diff
@@ -226,6 +226,7 @@ impl ExecutionPlan for MatchQueryExec {
                 ),
             };
             let tokens = collect_tokens(&query.terms, &mut tokenizer, None);
+            log::info!("tokens: {:?}", tokens);
```
Xuanwo: Is this for debug?

BubbleCal: oh... yes, removed

Xuanwo (approved):
Thank you for fixing this.

PR #4526, opened 2025-08-21
fix: use java/lang/String to get DeletionFileType value

fangbo:


fangbo:
@yanghua A small bug fix. Could you please review it ?

yanghua (approved):
LGTM

PR #4528, opened 2025-08-21
feat(java): add deleteRows for Fragment

fangbo:
This PR adds a method for Dataset to delete some rows (identified by _rowaddr list) from a specific Fragment.

This method will be used in Spark MOR update. The updated old rows are deleted from the Fragment using Deletion Vector. And updated new rows are written in new Fragments. In commit stage, Update transaction operation is committed on the Dataset to apply updation.

### DIFF
--- a/java/core/lance-jni/src/fragment.rs
+++ b/java/core/lance-jni/src/fragment.rs
@@ -24,7 +24,7 @@ use crate::{
     blocking_dataset::{BlockingDataset, NATIVE_DATASET},
     traits::FromJString,
     utils::extract_write_params,
-    RT,
+    JNIEnvExt, RT,
 };
 
 //////////////////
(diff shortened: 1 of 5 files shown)


fangbo:
@jackye1995 @majin1102 This PR is ready. Could you please review it ? Thank you .

majin1102:
Thanks for this contribution.

Could you elaborate more on the necessity of this interface?
IMO the current transaction interface should be enough to do Spark MOR update.

jackye1995:
> IMO the current transaction interface should be enough to do Spark MOR update.

My guess is that you want to just use the `extend_deletions` function so you don't need to update the deletion vector and then manually construct a fragment metadata? If that is the case it seems is fine, but it certainly should not be a dataset level API.

fangbo:
@jackye1995 @majin1102 Thanks for you comments. I am implementing update on spark using MOR. In the commit stage, the fragment's updated rows should be removed. Please see https://github.com/lancedb/lance-spark/pull/79/files#diff-b9cf282012c147e822114a41fbe6c6727c20ba13fb9e3838091af1a1e1a88072R168

fangbo:
> > IMO the current transaction interface should be enough to do Spark MOR update.
> 
> My guess is that you want to just use the `extend_deletions` function so you don't need to update the deletion vector and then manually construct a fragment metadata? If that is the case it seems is fine, but it certainly should not be a dataset level API.

Thanks for your suggestion. I moved the method deleteByAddrs to Fragment.

cc @majin1102

fangbo:
@jackye1995 This PR is ready. Could you please review it again? Thank you.

java/core/lance-jni/src/fragment.rs
```diff
@@ -212,6 +212,73 @@ fn create_fragment<'a>(
...
+    }
+
+    let res = RT.block_on(async move { fragment.extend_deletions(indexes).await });
+
+    let obj = match res {
+        Ok(Some(f)) => f.metadata().into_java(env)?,
+        Ok(None) => JObject::default(),
+        Err(e) => {
+            return Err(Error::runtime_error(format!(
+                "Error to delete fragment:{} with error message:{}",
```
jackye1995: the error should be something like `format!("Cannot delete rows in fragment {}: {:?}", fragment_id, e)`

java/core/src/main/java/com/lancedb/lance/Fragment.java
```diff
@@ -101,6 +101,20 @@ public LanceScanner newScan(ScanOptions options) {
...
   }
 
+  /**
+   * Delete rows by row addresses. All the rows belong to the Fragment.
+   *
+   * @param rowAddrs The row addresses to delete.
+   * @return The fragment metadata after deletion. If all rows are deleted, return Null. Otherwise,
+   *     returns a new fragment with the updated deletion vector.
+   */
+  public FragmentMetadata deleteByAddrs(List<Long> rowAddrs) {
```
jackye1995: I am starting to wonder if this binding is really needed, because there are many different ways to expose this, like why not convert addresses to indexes and then do JNI, or just only do JNI for writing the deletion vectors, or don't do anything in JNI. Do you have the corresponding PR for Spark Update/Merge? It probably makes more sense to review those together. What do you think?

fangbo: @jackye1995 As this code https://github.com/lancedb/lance-spark/pull/79/files#diff-8ff2b32a01a531cca5f46144d603cfa63c99bf836f208c0c03950c4980f132acR180 shows that in the task commit stage, the updated rows should be deleted from the old fragments. Currently, the FileFragment.extend_deletions can delete the specific row addresses.

jackye1995:
```
@Override
    public void delete(InternalRow metadata, InternalRow id) throws IOException {
      int fragmentId = metadata.getInt(0);
      deletedAddrs.compute(
          fragmentId,
          (k, v) -> {
            if (v == null) {
              v = new ArrayList<>();
            }
            v.add(id.getLong(0));
            return v;
          });
    }
```

is it possible for us to directly make `deletedAddrs` storing fragment ID -> list of row indexes in the fragment ID? Becuase here you can already derive the row index from the address, that should be able to save the step in the commit to do another cycle of deriving the row index at JNI layer.

fangbo: @jackye1995 Thanks for your suggestion. I also think using row index has more clearer semantics. So I change method to `deleteRows` and change the parameter to `List<Integer> rowIndexes`. 

And as code https://github.com/lancedb/lance-spark/pull/79/files#diff-8ff2b32a01a531cca5f46144d603cfa63c99bf836f208c0c03950c4980f132acR148  shows that the deleted row indexes are collected and deleted in commit stage.

java/core/src/main/java/com/lancedb/lance/util/RowAddress.java
jackye1995, line 1: `RowAddress` should be at the parent level instead of in `util`

PR #4530, opened 2025-08-21
Adding basic boilplate code for bloom filter

HaochengLIU:
Address #4162

### DIFF
--- a/rust/lance-index/src/lib.rs
+++ b/rust/lance-index/src/lib.rs
@@ -98,6 +98,8 @@ pub enum IndexType {
 
     ZoneMap = 8, // ZoneMap
 
+    BloomFilter = 9, // Bloom filter
+
     // 100+ and up for vector index.
     /// Flat vector index.
     Vector = 100, // Legacy vector index, alias to IvfPq
@@ -120,6 +122,7 @@ impl std::fmt::Display for IndexType {
             Self::FragmentReuse => write!(f, "FragmentReuse"),
             Self::MemWal => write!(f, "MemWal"),
             Self::ZoneMap => write!(f, "ZoneMap"),
+            Self::BloomFilter => write!(f, "BloomFilter"),
             Self::Vector | Self::IvfPq => write!(f, "IVF_PQ"),
             Self::IvfFlat => write!(f, "IVF_FLAT"),
             Self::IvfSq => write!(f, "IVF_SQ"),
@@ -144,6 +147,7 @@ impl TryFrom<i32> for IndexType {
             v if v == Self::FragmentReuse as i32 => Ok(Self::FragmentReuse),
             v if v == Self::MemWal as i32 => Ok(Self::MemWal),
             v if v == Self::ZoneMap as i32 => Ok(Self::ZoneMap),
+            v if v == Self::BloomFilter as i32 => Ok(Self::BloomFilter),
             v if v == Self::Vector as i32 => Ok(Self::Vector),
             v if v == Self::IvfFlat as i32 => Ok(Self::IvfFlat),
             v if v == Self::IvfSq as i32 => Ok(Self::IvfSq),
@@ -170,6 +174,7 @@ impl IndexType {
                 | Self::Inverted
                 | Self::NGram
                 | Self::ZoneMap
+                | Self::BloomFilter
         )
     }
 
@@ -206,7 +211,7 @@ impl IndexType {
             Self::FragmentReuse => 0,
             Self::MemWal => 0,
             Self::ZoneMap => 0,
-
+            Self::BloomFilter => 0,
             // for now all vector indices are built by the same builder,
             // so they share the same version.
             Self::Vector


jackye1995:
Have we decided which bloom filter library to use?

HaochengLIU:
> Have we decided which bloom filter library to use?

Nah, plan to take a look tmr,  today I was distracted from the family chore :(

rust/lance-index/Cargo.toml
```diff
@@ -61,6 +61,9 @@ tempfile.workspace = true
 crossbeam-queue.workspace = true
 bytes.workspace = true
 uuid.workspace = true
+fastbloom = "0.14.0"
+parquet = "55.1"
```
jackye1995: I think we should not bring in parquet dependency, looks like the only usage is just `parquet::data_type::AsBytes`, in that case it seems like we can easily replace that?

rust/lance-index/src/scalar/bloomfilter.rs
```diff
@@ -0,0 +1,1769 @@
...
+        })
+    }
+
+    fn evaluate_block_against_query(
+        &self,
+        block: &BloomFilterStatistics,
+        query: &SargableQuery,
+    ) -> Result<bool> {
+        // If no bloom filter data is available, be conservative and return true
+        if block.bloom_filter_data.is_empty() {
```
jackye1995: why could a block have no data? If there is no data the block shouldn't really exist I think

rust/lance-index/src/scalar/bloomfilter.rs
```diff
@@ -0,0 +1,1769 @@
...
+        match query {
+            // TODO: Shall we care about null here?
+            SargableQuery::IsNull() => {
+                // For null values, we could use a special marker in the bloom filter
+                // For now, conservatively assume blocks might contain nulls
+                Ok(true)
+            }
+            SargableQuery::Equals(target) => {
+                if target.is_null() {
+                    // Handle null values - conservatively return true for now
```
jackye1995: I think we can just add a `has_null` flag in the metadata to make it work

rust/lance-index/src/scalar/bloomfilter.rs
```diff
@@ -0,0 +1,1769 @@
...
+        })
+    }
+
+    async fn prewarm(&self) -> Result<()> {
+        Ok(())
+    }
+
+    fn statistics(&self) -> Result<serde_json::Value> {
+        Ok(serde_json::json!({
+            "type": "BloomFilter",
```
jackye1995, line 432: type is not necessary in stats

rust/lance-index/src/scalar/bloomfilter.rs
```diff
@@ -0,0 +1,1769 @@
...
+            number_of_items,
+            number_of_bytes,
+        }
+    }
+}
+
+pub struct BloomFilterIndexBuilder {
+    params: BloomFilterIndexBuilderParams,
+    value_type: DataType,
+    blocks: Vec<BloomFilterStatistics>,
```
jackye1995, line 608: why do we call it blocks? why not just zones?

rust/lance-index/src/scalar/bloomfilter.rs
```diff
@@ -0,0 +1,1769 @@
...
+        if self.cur_block_offset > 0 {
+            self.new_block(self.cur_fragment_id)?;
+        }
+
+        Ok(())
+    }
+
+    fn bloomfilter_stats_as_batch(&self) -> Result<RecordBatch> {
+        // Create placeholder array for the indexed data type with same length as other columns
+        let data_type_placeholder = if self.blocks.is_empty() {
```
jackye1995: why do we need to record data type? feels unnecssary

jackye1995 (commented):
Nice start! Added some initial comments

rust/lance-index/Cargo.toml
HaochengLIU: I need the AsBytes traits defined for all arrow types
if it's removed, i will hit a build error as
```
error[E0277]: the trait bound `<T as ArrowPrimitiveType>::Native: AsBytes` is not satisfied
   --> rust/lance-index/src/scalar/bloomfilter.rs:577:29
    |
577 |                 sbbf.insert(&array.value(i));
    |                      ------ ^^^^^^^^^^^^^^^ the trait `AsBytes` is not implemented for `<T as ArrowPrimitiveType>::Native`
    |                      |
    |                      required by a bound introduced by this call

```

rust/lance-index/src/scalar/bloomfilter.rs
HaochengLIU: agree, the block is removed

rust/lance-index/src/scalar/bloomfilter.rs
HaochengLIU: parquet call it blocks so i inherit that naming convention

rust/lance-index/src/scalar/bloomfilter.rs
HaochengLIU: removed

rust/lance-index/src/scalar/bloomfilter.rs
HaochengLIU: Removed

rust/lance-index/Cargo.toml
jackye1995: Oh sorry I completely skipped the `sbbf.rs` in review. I think I would personally prefer to just fork the sbbf implementation based on the current situation. Curious what you think @westonpace 

It seems strange to pull in all the Parquet dependencies and have `parquet::data_type::AsBytes` in Lance, which is for Parquet types. We are already having an arrow array here, we should not need all the baggage of handling all the Parquet logical types to bytes.

rust/lance-index/src/scalar/bloomfilter.rs
HaochengLIU: let me put more thoughts into it

rust/lance-index/Cargo.toml
HaochengLIU: I think this is the file
https://arrow.apache.org/rust/src/parquet/data_type.rs.html#665-669

rust/lance-index/src/scalar/bloomfilter.rs
HaochengLIU: updated

rust/lance-index/Cargo.toml
jackye1995: yeah, what I am saying is can we just move the necessary logic in that here so we don't depend on parquet? Actually there might be some similar utils directly for arrow, Weston might know better about that.

HaochengLIU: I'm working on some other errors from the CI bot, will revisit it

HaochengLIU: @jackye1995 ,Kudos to LLM, it's forked now 
🤖 😲

python/python/lance/dataset.py
```diff
@@ -1951,6 +1952,10 @@ def create_scalar_index(
         * ``FTS/INVERTED``. It is used to index document columns. This index
           can conduct full-text searches. For example, a column that contains any word
           of query string "hello world". The results will be ranked by BM25.
+        * ``BLOOMFILTER``. This inexact index is a probabilistic data structure
+          that answers "Is this item definitely NOT in the set?" with 100% accuracy,
+          and "Is this item MAYBE in the set?" with some false positives with minimal
+          memory usage.
```
westonpace: Let's focus on what this means for the user.
```suggestion
        * ``BLOOMFILTER``. This inexact index uses a bloom filter.  It is small
             but can only handle filters with equals and not equals and may require
             more I/O than a btree or bitmap index```

File: python/src/dataset.rs [hunk 8]
```diff
@@ -1524,6 +1525,10 @@ impl Dataset {
                 index_type: "label_list".to_string(),
                 params: None,
             }),
+            "BLOOMFILTER" => Box::new(ScalarIndexParams {
+                index_type: "bloomfilter".to_string(),
+                params: None,
+            }),
```
westonpace, line 1531: There is a lot of boiler plate to add new index to python.  We should simplify some of this (definitely in future PR).  Probably just need to add a "short name" parameter to the index plugin.

rust/lance-index/src/lib.rs
```diff
@@ -101,6 +101,8 @@ pub enum IndexType {
 
     ZoneMap = 8, // ZoneMap
 
+    BloomFilter = 9, // Bloom filter
+
```
westonpace, line 105: I would like to eventually move away from needing to add items to this enum every time we add an index.  I don't think it needs to be strictly required (e.g. we don't have JSON here).

rust/lance-index/src/scalar.rs
```diff
@@ -60,6 +61,7 @@ pub enum BuiltinIndexType {
     NGram,
     ZoneMap,
     Inverted,
+    BloomFilter,
```
westonpace, line 63: Let's not add this here.

rust/lance-index/src/scalar.rs
```diff
@@ -98,7 +102,7 @@ impl TryFrom<IndexType> for BuiltinIndexType {
 pub struct ScalarIndexParams {
     /// The type of index to create
     ///
-    /// Builtin indexes are "btree", "ngram", "bitmap", "inverted", "labellist", and "zonemap"
+    /// Builtin indexes are listed in BuiltinIndexType
```
westonpace: Let's just remove this whole section and leave it at just...

```
/// The type of index to create
```

rust/lance-index/src/scalar/bloomfilter/as_bytes.rs
```diff
@@ -0,0 +1,157 @@
...
+/// Reference: https://arrow.apache.org/rust/src/parquet/data_type.rs.html
+pub trait AsBytes {
+    /// Convert the value to a byte slice
+    fn as_bytes(&self) -> &[u8];
+}
+
+impl AsBytes for i32 {
+    fn as_bytes(&self) -> &[u8] {
+        unsafe {
+            std::slice::from_raw_parts(self as *const i32 as *const u8, std::mem::size_of::<i32>())
```
westonpace: Hmm, can we use methods like `i32::to_le_bytes`?  I don't think we'd work on a big-endian machine anyways but let's at least be more clear and help remove a potential hurdle in the future?

rust/lance-index/src/scalar/bloomfilter.rs
```diff
@@ -0,0 +1,1716 @@
...
+    blocks: Vec<BloomFilterStatistics>,
+    number_of_items: u64,
+    number_of_bytes: usize,
+
+    store: Arc<dyn IndexStore>,
+    fri: Option<Arc<FragReuseIndex>>,
+    index_cache: LanceCache,
+}
+
+impl std::fmt::Debug for BloomFilterIndex {
```
westonpace: This looks like the default debug impl.  Can you just derive debug?

rust/lance-index/src/scalar/bloomfilter.rs
```diff
@@ -0,0 +1,1716 @@
...
+    // Whether this block contains any null values
+    has_null: bool,
+    // Serialized bloom filter data (SBBF bytes)
+    bloom_filter_data: Vec<u8>,
+}
+
+pub struct BloomFilterIndex {
+    blocks: Vec<BloomFilterStatistics>,
+    number_of_items: u64,
+    number_of_bytes: usize,
```
westonpace: Can we comment what these represent?

rust/lance-index/src/scalar/bloomfilter.rs
```diff
@@ -0,0 +1,1716 @@
...
+                Error::invalid_input("BloomFilterIndex: missing 'has_null' column", location!())
+            })?
+            .as_any()
+            .downcast_ref::<arrow_array::BooleanArray>()
+            .ok_or_else(|| {
+                Error::invalid_input(
+                    "BloomFilterIndex: 'has_null' column is not Boolean",
+                    location!(),
+                )
+            })?;
```
westonpace, line 209: These long blocks show up all over the code base.  Very tempted to add a `column_as_type` method to record batch :laughing:

rust/lance-index/src/scalar/bloomfilter.rs
```diff
@@ -0,0 +1,1716 @@
...
+                Error::invalid_input("BloomFilterIndex: missing 'block_size' column", location!())
+            })?
+            .as_any()
+            .downcast_ref::<arrow_array::UInt64Array>()
+            .ok_or_else(|| {
+                Error::invalid_input(
+                    "BloomFilterIndex: 'block_size' column is not UInt64",
+                    location!(),
+                )
+            })?;
```
westonpace: Very minor nitpick but I like to use `size` for "bytes" and `length` for "rows" whenever possible.

rust/lance-index/src/scalar/bloomfilter.rs
```diff
@@ -0,0 +1,1716 @@
...
+        query: &SargableQuery,
+    ) -> Result<bool> {
+        // Try to reconstruct the bloom filter from the stored data
+        let sbbf = match Sbbf::new(&block.bloom_filter_data) {
+            Ok(sbbf) => sbbf,
+            Err(_) => {
+                // If we can't reconstruct the bloom filter, be conservative
+                return Ok(true);
+            }
+        };
```
westonpace: What's the cost of this?  Why don't we just store `Sbbf` instead of `Vec<u8>`?

rust/lance-index/src/scalar/bloomfilter/sbbf.rs
```diff
@@ -0,0 +1,532 @@
...
+        match self {
+            SbbfError::WriteError { source } => Some(source),
+            _ => None,
+        }
+    }
+}
+
+pub type Result<T> = std::result::Result<T, SbbfError>;
+
+// We reuse the AsBytes trait from parquet crate instead of reimplementing it
```
westonpace: Well...we don't anymore do we?

rust/lance-index/src/scalar/bloomfilter.rs
```diff
@@ -0,0 +1,1716 @@
...
+                        datafusion_common::ScalarValue::LargeUtf8(Some(val)) => {
+                            sbbf.check(val.as_str())
+                        }
+                        datafusion_common::ScalarValue::Binary(Some(val)) => {
+                            sbbf.check(val.as_slice())
+                        }
+                        datafusion_common::ScalarValue::LargeBinary(Some(val)) => {
+                            sbbf.check(val.as_slice())
+                        }
+                        _ => true, // Conservative for unsupported types
```
westonpace: Ideally we never get here?  Can we add a check in `new_training_request`?

rust/lance-index/src/scalar/bloomfilter.rs
westonpace: It's back :)

rust/lance-index/src/scalar/bloomfilter.rs
```diff
@@ -0,0 +1,1716 @@
...
+        .expect("failed to parse Lance_BLOOMFILTER_DEFAULT_NUMBER_OF_BYTES")
+});
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct BloomFilterIndexBuilderParams {
+    #[serde(default = "default_number_of_items")]
+    number_of_items: u64,
+    #[serde(default = "default_number_of_bytes")]
+    number_of_bytes: usize,
+}
```
westonpace: This is _ok_ but is there some way we can just take in the selectivity instead?

rust/lance-index/src/scalar/bloomfilter.rs
westonpace: I also like zones since we will have several "zone-like" indexes and it would be nice for them to all align on terminology but not enough to fight over it too much :stuck_out_tongue:

rust/lance-index/src/scalar/bloomfilter.rs
```diff
@@ -0,0 +1,1716 @@
...
+            } else {
+                has_null = true;
+            }
+        }
+        has_null
+    }
+
+    fn update_stats(&mut self, array: &ArrayRef) -> Result<()> {
+        if let Some(ref mut sbbf) = self.sbbf {
+            let has_null = match array.data_type() {
```
westonpace, line 703: This seems like a fairly small set of supported types.  No unsigned integers or 1 or 2 byte integers?

rust/lance-index/src/scalar/bloomfilter.rs
```diff
@@ -0,0 +1,1716 @@
...
+            index_version: BLOOMFILTER_INDEX_VERSION,
+        })
+    }
+
+    fn provides_exact_answer(&self) -> bool {
+        false
+    }
+
+    fn version(&self) -> u32 {
+        0
```
westonpace: Let's make this a constant for consistency with other indexes?

rust/lance-index/src/scalar/bloomfilter.rs
```diff
@@ -0,0 +1,1716 @@
...
+    }
+
+    fn new_query_parser(
+        &self,
+        index_name: String,
+        _index_details: &prost_types::Any,
+    ) -> Option<Box<dyn ScalarQueryParser>> {
+        // TODO: Should it be sargable query?
+        Some(Box::new(SargableQueryParser::new(index_name, true)))
```
westonpace: Can we handle range filters?

rust/lance-index/src/scalar/bloomfilter.rs
```diff
@@ -0,0 +1,1716 @@
...
+                    _ => {
+                        // For unsupported types, be conservative
+                        Ok(true)
+                    }
+                }
+            }
+            SargableQuery::Range(_, _) => Err(Error::NotSupported {
+                source: "range queries are not supported for bloom filter indexes".into(),
+                location: location!(),
+            }),
```
westonpace: If range queries are not supported then we either need a new query type or we need to parameterize `SargableQuery` further.  This spot is too late to realize we don't support range queries.  We need to avoid thinking we can use the index in the first place.  Otherwise it won't properly use a non-index filter for range queries.

rust/lance-index/src/scalar.rs
HaochengLIU: remove it

rust/lance-index/src/scalar/bloomfilter.rs
HaochengLIU: Move the construction to load time rather than search time

rust/lance-index/src/scalar/bloomfilter.rs
HaochengLIU: expose N and P

rust/lance-index/src/scalar/bloomfilter.rs
HaochengLIU: time type? or create an issue to this type

rust/lance-index/src/scalar/bloomfilter.rs
HaochengLIU: Query parser should return None rather than the query

HaochengLIU: Add a test case first before impl

rust/lance-index/src/scalar/bloomfilter.rs
HaochengLIU: it's 100% gone this time :)

rust/lance-index/src/scalar/bloomfilter.rs
HaochengLIU: renamed

python/python/lance/dataset.py
HaochengLIU: reworded

rust/lance-index/src/scalar/bloomfilter/as_bytes.rs
HaochengLIU: done with `cow`

rust/lance-index/src/scalar/bloomfilter.rs
HaochengLIU: deleted

rust/lance-index/src/scalar/bloomfilter.rs
HaochengLIU: done

rust/lance-index/src/scalar/bloomfilter.rs
HaochengLIU: renamed all in zonemap and bloomfilter

rust/lance-index/src/scalar/bloomfilter.rs
HaochengLIU: added

rust/lance-index/src/scalar/bloomfilter.rs
HaochengLIU: reworded all remaining scalar filters in lance

rust/lance-index/src/scalar.rs
HaochengLIU: Update: @westonpace it's needed for the pyo3 bindings, otherwise it wont work

westonpace:
(FYI, in case you're working on this right now)

There was a rebase issue causing the tests to fail.  I rebased, fixed the issue, and force pushed back up.  So you'll need to pull with rebase (but I'm hoping this gets the CI passing and we can just merge)

PR #4531, opened 2025-08-21
fix: incorrect logic for logical or of RowIdMask

westonpace:


rust/lance-core/src/utils/mask.rs
```diff
@@ -63,6 +63,23 @@ impl RowIdMask {
...
+        {
+            allow_list -= &block_list;
+            Self {
+                allow_list: Some(allow_list),
+                block_list: None,
+            }
+        } else {
+            self
+        }
+    }
```
westonpace, line 81: I suppose this begs the question why we allow there to be both anyways.  It would simplify a few paths if we collapse RowIdMask into an enum of `All, Allow, Block`.

wkalt: does that assume the BlockList is a subset of the AllowList?

wkalt: I guess the two lists are exhaustive for the fragments covered by the index, so we don't have an issue here.

westonpace: It's defined as...

```
/// If both the allow_list and the block_list are Some then the only selected
/// row ids are those that are in the allow_list but not in the block_list
```

So it doesn't matter if the block list includes things that are not in the allow list.

wkalt:
@westonpace looks fine to me. Is there a place we can stick a higher level test on filter with an OR condition?

rust/lance-core/src/utils/mask.rs
westonpace: I think I've convinced myself I do want to collapse this but I'll do it in a future PR

westonpace:
> @westonpace looks fine to me. Is there a place we can stick a higher level test on filter with an OR condition?

~~Sure, I'll add your python test as it's pretty compact~~

Added in https://github.com/lancedb/lance/pull/4531/commits/631ffbcad047d502c91f815779c07d8f7c88b4a3

Issue #4538, opened 2025-08-22
fix: blocking_take may hang on V2.0


ColdL:
When batch_size is larger than to_take, blocking_take will hang for V2.0.

By modifying the UT test_blocking_take in rust/lance-file/src/v2/reader.rs and changing LanceFileVersion from V2_1 to V2_0, this issue can be reproduced.

westonpace:
What is `to_take` in this situation?

ColdL:
The `to_take` I mentioned refers to the local variable `to_take` defined in the `next_batch_task` method within `BatchDecodeIterator` in file `lance-encoding/src/decoder.rs`.

This `next_batch_task` method internally calls the `wait_for_io` method.

Inside the `wait_for_io` method, the original implementation is:
```
let loaded_need = self.rows_drained + self.rows_per_batch as u64 - 1;
```

However, I found that when `to_take` is small, it should be written as:
```
let loaded_need = self.rows_drained + to_take.min(self.rows_per_batch as u64) - 1;
```

Otherwise, the next step `self.root_decoder.wait` will get stuck.

Xuanwo:
Closing as completed by #4539. The decoder wait condition was fixed, with blocking-take coverage for V2.0, V2.1, and V2.2.

PR #4433, opened 2025-08-11
chore(java): set the default file version to stable in java file api

frankliee:
Since v2.1 is still unstable for production, we should set the default file version to stable for java file api (v2.0).
By the way, in python file api, the default file version is v2.0.

| Version        | Minimal Lance Version | Maximum Lance Version | Description                                                                                                                                  |
| -------------- | --------------------- | --------------------- | -------------------------------------------------------------------------------------------------------------------------------------------- |
| 0.1            | Any                   | Any                   | This is the initial Lance format.                                                                                                            |
| 2.0            | 0.16.0                | Any                   | Rework of the Lance file format that removed row groups and introduced null support for lists, fixed size lists, and primitives              |
| 2.1 (unstable) | None                  | Any                   | Enhances integer and string compression, adds support for nulls in struct fields, and improves random access performance with nested fields. |
| legacy         | N/A                   | N/A                   | Alias for 0.1                                                                                                                                |
| stable         | N/A                   | N/A                   | Alias for the latest stable version (currently 2.0)                                                                                          |
| next           | N/A                   | N/A                   | Alias for the latest unstable version (currently 2.1)                                                                                        |

Issue #4422, opened 2025-08-10
Failed to read v2.0 lance file with java sdk


ChongWei905:
I tried to read V2.0 lance file with java sdk LanceFileReader.open and LanceFileReader.readAll (tag v0.32.0). But following error appears:

```text
thread '<unnamed>' panicked at /Users/weichong/Documents/working_area/lance/rust/lance-encoding/src/v2/encodings/physical/binary.rs:160:9:
there is no reactor running, must be called from the context of a Tokio 1.x runtime
stack backtrace:
   0: rust_begin_unwind
             at /rustc/05f9846f893b09a1be1fc8560e33fc3c815cfecb/library/std/src/panicking.rs:695:5
   1: core::panicking::panic_fmt
             at /rustc/05f9846f893b09a1be1fc8560e33fc3c815cfecb/library/core/src/panicking.rs:75:14
   2: core::panicking::panic_display
             at /Users/weichong/.rustup/toolchains/1.86.0-aarch64-apple-darwin/lib/rustlib/src/rust/library/core/src/panicking.rs:261:5
   3: tokio::task::spawn::spawn_inner::panic_cold_display
   4: tokio::task::spawn::spawn_inner
   5: tokio::task::spawn::spawn
   6: <lance_encoding::v2::encodings::physical::binary::BinaryPageScheduler as lance_encoding::decoder::PageScheduler>::schedule_ranges
   7: <lance_encoding::v2::encodings::logical::primitive::PrimitiveFieldSchedulingJob as lance_encoding::v2::decoder::SchedulingJob>::schedule_next
   8: <lance_encoding::v2::encodings::logical::struct::SimpleStructSchedulerJob as lance_encoding::v2::decoder::SchedulingJob>::schedule_next
   9: lance_encoding::decoder::DecodeBatchScheduler::do_schedule_ranges_legacy
  10: lance_encoding::decoder::DecodeBatchScheduler::do_schedule_ranges
  11: lance_encoding::decoder::DecodeBatchScheduler::schedule_ranges
  12: lance_encoding::decoder::schedule_and_decode_blocking
  13: lance_file::v2::reader::FileReader::read_range_blocking
  14: lance_file::v2::reader::FileReader::read_stream_projected_blocking
  15: lance_jni::file_reader::BlockingFileReader::open_stream
  16: lance_jni::file_reader::inner_read_all
  17: lance_jni::file_reader::Java_com_lancedb_lance_file_LanceFileReader_readAllNative::{{closure}}
[...]

PR #4539, opened 2025-08-22
fix: blocking_take may hang on V2.0

ColdL:
fix decoder wait_for_io when to_take is less than batch_size (#4538)

### DIFF
--- a/rust/lance-encoding/src/decoder.rs
+++ b/rust/lance-encoding/src/decoder.rs
@@ -1516,7 +1516,7 @@ impl<T: RootDecoderType> BatchDecodeIterator<T> {
     /// Note that `scheduled_need` is cumulative.  E.g. this method
     /// should be called with 5, 10, 15 and not 5, 5, 5
     #[instrument(skip_all)]
-    fn wait_for_io(&mut self, scheduled_need: u64) -> Result<u64> {
+    fn wait_for_io(&mut self, scheduled_need: u64, to_take: u64) -> Result<u64> {
         while self.rows_scheduled < scheduled_need && !self.messages.is_empty() {
             let message = self.messages.pop_front().unwrap()?;
             self.rows_scheduled = message.scheduled_so_far;
@@ -1538,7 +1538,7 @@ impl<T: RootDecoderType> BatchDecodeIterator<T> {
             }
         }
 
-        let loaded_need = self.rows_drained + self.rows_per_batch as u64 - 1;
+        let loaded_need = self.rows_drained + to_take.min(self.rows_per_batch as u64) - 1;
 
         self.root_decoder
             .wait(loaded_need, &self.wait_for_io_runtime)?;
@@ -1568,7 +1568,7 @@ impl<T: RootDecoderType> BatchDecodeIterator<T> {
                 "Draining from scheduler (desire at least {} scheduled rows)",
                 desired_scheduled
             );
-            let actually_scheduled = self.wait_for_io(desired_scheduled)?;
+            let actually_scheduled = self.wait_for_io(desired_scheduled, to_take)?;
             if actually_scheduled < desired_scheduled {
                 let under_scheduled = desired_scheduled - actually_scheduled;
                 to_take -= under_scheduled;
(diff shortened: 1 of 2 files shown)


ColdL:
In `lance-encoding/src/decoder.rs`, the `next_batch_task` calls the `wait_for_io method` to wait for the read operation to complete.

However, `wait_for_io` always waits for data of length `rows_per_batch`, which is actually `batch_size`, rather than the actual amount of data needed, `to_take`.

For some reason, under the `V2_0` version, this can cause the program to hang at `wait_for_io`. Changing the file version used in the UT `test_blocking_take` in `rust/lance-file/src/v2/reader.rs` from `V2_1` to `V2_0` can reproduce this problem. (However, tests show that the `V2_1` version does not have this issue, which I do not fully understand yet).

To fix this issue, we can change the wait target of `wait_for_io` to `min(rows_per_batch, to_take)`, that is, change this line:
```
let loaded_need = self.rows_drained + self.rows_per_batch as u64 - 1;
```
to this line:
```
let loaded_need = self.rows_drained + to_take.min(self.rows_per_batch as u64) - 1;
```
I have tested this and it solves the problem.

Although I haven't fully understood why `V2_1` doesn't have this issue, I believe this change itself should be safe and can fix the UT failure under `V2_0` file version (this commit also modifies the UT `test_blocking_take` in `rust/lance-file/src/v2/reader.rs` to run under both `V2_0` and `V2_1`). The last run seemed to fail due to code formatting issues. Hope you can help review this change again.

ColdL:
Screenshot 2025-09-04 at 02 42 24

The above error seems to be caused by CI environment instability. Could you please help re-run the jobs?

jackye1995:
> (However, tests show that the V2_1 version does not have this issue, which I do not fully understand yet).

want to make sure we are not hiding any potential issues in 2.1, @Xuanwo any thoughts on this bug?

ColdL:
I took a closer look at the current commit to understand why, without this commit, the unit test `test_blocking_take` hangs for `V2_0` but not for `V2_1`.

I think the immediate cause may be this: for `V2_0`, `wait_for_io` actually waits, while for `V2_1`, `wait_for_io` doesn't really "wait."

The key lies in `lance-encoding/src/decoder.rs`, in the `wait_for_io` method. On the line
```
let loaded_need = ...
```
the current commit changed
```
let loaded_need = self.rows_drained + self.rows_per_batch as u64 - 1;
```
to
```
let loaded_need = self.rows_drained + to_take.min(self.rows_per_batch as u64) - 1;
```
This `loaded_need` is then immediately passed to `self.root_decoder.wait`. This seems to be the crux.

1. For `V2_0`, `self.root_decoder.wait` goes into the `SimpleStructDecoder` implementation, which performs a real `runtime.block_on` wait;
2. However, for `V2_1`, `self.root_decoder.wait` goes into the `StructuralStructDecoder` implementation, whose wait function is effectively a no-op and immediately returns Ok ("Waiting happens elsewhere"). Therefore, whatever value is passed as `loaded_need` to wait has no effect.

This appears to be the direct reason for "Why the `V2_1` version does not have this issue."

This also indicates that the current change will only affect `SimpleStructDecoder` (mainly for `V2_0`) and will not affect `StructuralStructDecoder` (for `V2_1`).

Any thoughts on this issue? @jackye1995 @Xuanwo

ChongWei905:
Really looking forward this PR to be merged. This PR plus some changes in encoding could fix this issue: https://github.com/lancedb/lance/issues/4422 and finally set lance file version to v2.0 in java sdk file writer: https://github.com/lancedb/lance/pull/4433

jackye1995:
Sorry for the wait, I was hoping if Xuanwo can provide some insights because I am not super familiar with this code path and I don't want to merge something I don't fully understand the implication. If there is no further response I will take a stab at it tonight.

ChongWei905:
> Sorry for the wait, I was hoping if Xuanwo can provide some insights because I am not super familiar with this code path and I don't want to merge something I don't fully understand the implication. If there is no further response I will take a stab at it tonight.

thanks a lot!

PR #4545, opened 2025-08-22
feat: add lance-tools command

cmccabe:
Add a lance-tools command to look at data files.


rust/lance-tools/Cargo.toml
```diff
@@ -0,0 +1,28 @@
...
+authors.workspace = true
+license.workspace = true
+repository.workspace = true
+readme.workspace = true
+keywords.workspace = true
+categories.workspace = true
+rust-version.workspace = true
+
+[[bin]]
+name = "lance-tools"
```
jackye1995, line 15: should we just call the command `lance`?

rust/lance-tools/src/cli.rs
```diff
@@ -0,0 +1,39 @@
...
+
+use clap::{Parser, Subcommand};
+use std::io::Error;
+use std::io::ErrorKind;
+use lance_core::Result as LanceResult;
+
+#[derive(Parser, Debug)]
+#[command(
+    name = "lance-tools",
+    about = "Tools for working with Lance files",
```
jackye1995: I think it would be nice if the command can eventually cover both file and table, so we can have a subcommand domain like `lance file meta -s /path/to/file`, so that we can have another table domain like `lance table meta -s /path/to/table` in the future

rust/lance-tools/Cargo.toml
westonpace: I think the inspiration was to mirror https://github.com/hangxie/parquet-tools

rust/lance-tools/Cargo.toml
```diff
@@ -0,0 +1,28 @@
...
+path = "src/main.rs"
+
+[dependencies]
+clap = { workspace = true, features = ["derive"] }
+futures.workspace = true
+lance-core = { path = "../lance-core" }
+lance-encoding = { path = "../lance-encoding" }
+lance-file = { path = "../lance-file" }
+lance-io = { path = "../lance-io" }
+object_store = "0.12.2"
```
westonpace:
```suggestion
object_store.workspace = true
```

rust/lance-tools/Cargo.toml
```diff
@@ -0,0 +1,28 @@
+[package]
+name = "lance-tools"
+description = "Lance tools"
+version.workspace = true
+edition.workspace = true
+authors.workspace = true
+license.workspace = true
+repository.workspace = true
+readme.workspace = true
```
westonpace, line 9: It might not hurt for this crate to have its own readme

rust/lance-tools/Cargo.toml
```diff
@@ -0,0 +1,28 @@
+[package]
+name = "lance-tools"
+description = "Lance tools"
```
westonpace:
```suggestion
description = "Tools for working with Lance files"
```

(can match the about for the command)

rust/lance-tools/src/main.rs
```diff
@@ -0,0 +1,53 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright The Lance Authors
+
+use clap::Parser;
+use lance_tools::cli::{LanceToolsArgs, run};
+
+#[tokio::main]
+async fn main() -> Result<(), std::io::Error> {
```
westonpace: You could potentially just return `lance_core::Error` here instead of `std::io::Error` and save yourself the trouble of conversion.

Or, eventually, maybe it would make sense to define your own error type for the crate.  This way you can add errors (maybe needed for invalid command line argument combinations or something).  If you do that then you can also do `impl From` and wouldn't need the conversion.

rust/lance-tools/src/meta.rs
```diff
@@ -0,0 +1,67 @@
...
+            file_metadata,
+         };
+        return Ok(lance_tool_file_metadata);
+    }
+}
+
+pub async fn run(args: MetaArgs) -> Result<(), std::io::Error> {
+    let metadata =
+        crate::cli::lance_result_to_std_result(LanceToolFileMetadata::open(&args.source).await)?;
+    print!("{}", metadata.to_string());
```
westonpace: Why `print` and not `println`?

rust/lance-tools/src/cli.rs
```diff
@@ -0,0 +1,39 @@
...
+        },
+    };
+}
+
+pub fn lance_result_to_std_result<T>(lance_result: LanceResult<T>) -> Result<T, std::io::Error> {
+    return match lance_result {
+        Ok(t) => Result::Ok(t),
+        Err(e) => Result::Err(Error::new(ErrorKind::Other, e.to_string())),
+    };
+}
```
westonpace: This is a very minor and probably opinionated nitpick so don't feel required to act on it but for mapping errors where both error types are external (so you can't do `impl From...` I like it better when the end result looks like `do_something().map_my_error(...)` instead of `map_my_error(do_something())` (just because it reads nicer left to right) so two choices are...

* Change the function from `lance_result_to_std_result` to `lance_err_to_std_err` and then use `map_err` so you have `do_something().map_err(lance_err_to_std_err)`
* Create a trait `LanceResultExt` and add a `infer_std_err` method to the trait and then `impl` the trait for `LanceResult` so you can do `do_something().infer_std_err()`

That being said, it might be easiest to add your own error type (see next comment)

rust/lance-tools/src/cli.rs
westonpace: Agreed.  I often need to ask users to print out and send me their manifest when I am debugging issues and right now I have to do this by asking them to run a python script...

```
import lance
from lance.debug import format_manifest

ds = lance.dataset(...)
print(format_manifest(ds))
```

Having a CLI tool that can print the manifest would be pretty handy.

cmccabe: I added a subcommand system so that now we have:
```
cmccabe@Iris lance % ./target/debug/lance-tools
Tools for working with Lance

Usage: lance-tools <COMMAND>

Commands:
  file  Commands for interacting with Lance files
  help  Print this message or the help of the given subcommand(s)

Options:
  -h, --help     Print help
  -V, --version  Print version
cmccabe@Iris lance % ./target/debug/lance-tools file
Usage: lance-tools file <COMMAND>

Commands:
  meta  Display file metadata
  help  Print this message or the help of the given subcommand(s)

Options:
  -h, --help  Print help
cmccabe@Iris lance % ./target/debug/lance-tools file meta
error: the following required arguments were not provided:
  --source <source>

Usage: lance-tools file meta --source <source>

For more information, try '--help'.
cmccabe@Iris lance % ./target/debug/lance-tools file meta -s ~/test_data/lance-test-data/t.lance/data/c133c026-630d-424b-9e64-7848d5745549.lance```

File: rust/lance-tools/Cargo.toml (hunk 5 above)
Comment (cmccabe) [reply rust/lance-tools/Cargo.toml]: Changed to `Tools for interacting with Lance files and tables` to match the command's help.

File: rust/lance-tools/src/main.rs (hunk 6 above)
Comment (cmccabe) [reply rust/lance-tools/src/main.rs]: Hmm, interesting. For now, I changed it to use just use `lance_core::Error`. That's certainly cleaner than wrapping in a standard error with `ErrorKind::Other`.

File: rust/lance-tools/src/meta.rs (hunk 7 above)
Comment (cmccabe) [reply rust/lance-tools/src/meta.rs]: I changed it to `write!` to make it easier to unit test.

Looking at the output, it looks like `writeln` would add an extra newline that I don't want.

File: rust/lance-tools/src/cli.rs (hunk 8 above)
Comment (cmccabe) [reply rust/lance-tools/src/cli.rs]: I got rid of the error conversion as you suggested above.

**Comment by cmccabe** [comment]
Here is some example output:
```
cmccabe@Iris lance % ./target/debug/lance-tools file meta -s ~/test_data/lance-test-data/t.lance/data/c133c026-630d-424b-9e64-7848d5745549.lance
version: 2.0
num_rows: 30000
num_data_bytes: 270144
num_column_metadata_bytes: 298
num_footer_bytes: 443
schema:
Field(id=0, name=color, type=string)
Field(id=1, name=digit, type=int64)```

cmccabe:
some error output:

```cmccabe@Iris lance % ./target/debug/lance-tools file meta -s /tmp/nonexistent
Error: IO { source: NotFound { path: "/tmp/nonexistent", source: Os { code: 2, kind: NotFound, message: "No such file or directory" } }, location: Location { file: "/Users/cmccabe/src/lance/rust/lance-io/src/object_store.rs", line: 568, column: 12 } }```

Issue #4527, opened 2025-08-21
The example code for "Combine Features and Vectors" fails to execute successfully


LuciferYang:
I'm using Python 3.11.12, and the module versions in my environment are as follows:

```
Package         Version
--------------- -----------
duckdb          1.3.2
numpy           2.3.2
pandas          2.3.1
pip             25.2
pyarrow         21.0.0
pylance         0.33.0
python-dateutil 2.9.0.post0
pytz            2025.2
setuptools      80.7.1
six             1.17.0
tqdm            4.67.1
tzdata          2025.2
```

When I try to run the example code refer to  https://github.com/lancedb/lance/blob/main/docs/src/quickstart/vector-search.md#combine-features-and-vectors , I encounter the following error:

```
>>> sift1m.to_table(
...     columns=["item_id", "revenue"],
...     nearest={"column": "vector", "q": samples[0], "k": 10}
... )
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/xxx/Tools/lance-venv/lib/python3.11/site-packages/lance/dataset.py", line 932, in to_table
    ).to_table()
      ^^^^^^^^^^
  File "/Users/xxx/Tools/lance-venv/lib/python3.11/site-packages/lance/dataset.py", line 4136, in to_table
    return self.to_reader().read_all()
           ^^^^^^^^^^^^^^^^
  File "/Users/xxx/Tools/lance-venv/lib/python3.11/site-packages/lance/dataset.py", line 4139, in to_reader
    return self._scanner.to_pyarrow()
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
ValueError: LanceError(Schema): Schema error: No field named item_id. Valid fields are _distance, _rowid., /Users/runner/work/lance/lance/rust/lance-datafusion/src/projection.rs:159:25
```

It seems that adding columns to `tbl` does not affect `sift1m`.

jackye1995:
Yes nice catch

```
tbl = sift1m.to_table()
tbl = tbl.append_column("item_id", pa.array(range(len(tbl))))
tbl = tbl.append_column("revenue", pa.array((np.random.randn(len(tbl))+5)*1000))
```

is only adding columns in an Arrow table, not changing the underlying Lance dataset. It should directly do `sift1m.add_columns` instead. @davidmyriel we need to fix this

PR #4546, opened 2025-08-22
docs: fix vector search issue

davidmyriel:
solves https://github.com/lancedb/lance/issues/4527

### DIFF
--- a/docs/src/quickstart/vector-search.md
+++ b/docs/src/quickstart/vector-search.md
@@ -238,12 +238,13 @@ You can add metadata columns to your vector dataset and query both vectors and f
 
 In real-life situations, users have other feature or metadata columns that need to be stored and fetched together. If you're managing data and the index separately, you have to do a bunch of annoying plumbing to put stuff together. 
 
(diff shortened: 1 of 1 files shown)


docs/src/quickstart/vector-search.md
```diff
@@ -238,12 +238,13 @@ You can add metadata columns to your vector dataset and query both vectors and f
...
 
-With Lance it's a single call:
+With Lance, you can add columns directly to the dataset using `add_columns()`:
 
 ```python
-tbl = sift1m.to_table()
-tbl = tbl.append_column("item_id", pa.array(range(len(tbl))))
-tbl = tbl.append_column("revenue", pa.array((np.random.randn(len(tbl))+5)*1000))
+sift1m = sift1m.add_columns({
+    "item_id": pa.array(range(len(sift1m))),
```
jackye1995, line 245: hmm it does not look exactly right. You can consult examples in https://lancedb.github.io/lance/guide/data_evolution/#with-data-backfill, those are correct

PR #4547, opened 2025-08-22
feat: Provide interface to bulk query an index

### LINKED ISSUE #4523: Provide some way to build query a scalar index
For indexes that have pages / partitions, like BTree, each time we run a query we load the index part from cache.

This means if we have a tight loop, like during an indexed join, we keep reloading the same partition from the IndexCache. This is a lot of wasted allocations and CPU cycles.

https://github.com/lancedb/lance/blob/b3a41fdb38b0123c70256beb30ee2505acb2911f/rust/lance-index/src/scalar/btree.rs#L815-L834

jaystarshot:
feat(btree): add interface to bulk query an index

This needs to be used somewhere

Fixes #4523

### DIFF
--- a/rust/lance-index/src/scalar/btree.rs
+++ b/rust/lance-index/src/scalar/btree.rs
@@ -833,6 +833,73 @@ impl BTreeIndex {
         }
     }
 
+/// Bulk search interface that can run multiple queries efficiently by loading each page only once.
+pub async fn search_bulk(
+    &self,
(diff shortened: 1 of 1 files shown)


jaystarshot:
cc: @wjones127  (just getting started with lance + rust) and saw that adding this was easy

jaystarshot:
Doesn't seem to improve anything will fix and repopen

PR #4552, opened 2025-08-24
feat: provide feature flag for shallow cloning

majin1102:
Context:

https://github.com/lancedb/lance/pull/4257#issuecomment-3208930462

This is simpler than I thought...

### DIFF
--- a/rust/lance-table/src/feature_flags.rs
+++ b/rust/lance-table/src/feature_flags.rs
@@ -18,8 +18,10 @@ pub const FLAG_STABLE_ROW_IDS: u64 = 2;
 pub const FLAG_USE_V2_FORMAT_DEPRECATED: u64 = 4;
 /// Table config is present
 pub const FLAG_TABLE_CONFIG: u64 = 8;
+/// Dataset is a shallow clone with external base paths
+pub const FLAG_SHALLOW_CLONE: u64 = 16;
 /// The first bit that is unknown as a feature flag
-pub const FLAG_UNKNOWN: u64 = 16;
+pub const FLAG_UNKNOWN: u64 = 32;
 
 /// Set the reader and writer feature flags in the manifest based on the contents of the manifest.
 pub fn apply_feature_flags(manifest: &mut Manifest, enable_stable_row_id: bool) -> Result<()> {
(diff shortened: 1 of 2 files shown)


Xuanwo:
Is this related to our storage version? Do we need to add it in 2.2 instead of 2.1? 

I ask this question since we are very close to stabilize format 2.1 and we don't want to introduce any breaking changes after that.

Cc @jackye1995 @westonpace @wjones127

majin1102:
> Is this related to our storage version? Do we need to add it in 2.2 instead of 2.1?

From my perspective, I recommend including the shallow_clone feature in ​​v2.1​​ for the following reasons:

1. Schedule Uncertainty​​: The timeline for v2.2 is unclear, whereas v2.1 is already in progress.
2. Non-intrusive Design​​: This feature is purely a table-format enhancement, fully backward-compatible with existing datasets（if you don't use old versions read or write cloned dataset）.
3. No Conflict Risk​​: It does not interfere with the main futures intruduced by v2.1 (As I know, 2.1 mainly enhanced file format?).

However, if there are broader compatibility or dependency concerns I might have missed, I’m open to deferring it to v2.2. Would appreciate your thoughts!

majin1102:
@jackye1995  This PR is ready for review

westonpace:
> Is this related to our storage version? Do we need to add it in 2.2 instead of 2.1?
>
>I ask this question since we are very close to stabilize format 2.1 and we don't want to introduce any breaking changes after that.

I have not been following the shallow clone feature very closely but I think it is a table format feature and not a file format feature.  The 2.1 stabilization effort is for the file format.

We have not yet had a need to version the table format.  We require all changes are backwards compatible and have gone to some substantial effort to avoid any kind of breaking changes.  Not all changes are forwards compatible however.

westonpace (approved):
This looks good to me.  An attempt to open a dataset that has shallow clones should fail if the Lance version is too old to understand shallow clones.

I don't think we have any "forwards compatibility" testing framework (e.g. no way for a CI test to confirm that an old reader will fail to read a cloned dataset with the right error message).  Did you confirm this manually?

jackye1995 (approved):
Overall looks good to me.

> We require all changes are backwards compatible and have gone to some substantial effort to avoid any kind of breaking changes. Not all changes are forwards compatible however.

Yeah I think this is more of a forward compatibility problem of the table format.

PR #4553, opened 2025-08-24
feat!: shallow_clone supports index

majin1102:
Context:

https://github.com/lancedb/lance/pull/4257#issuecomment-3196495952


majin1102:
Hi, @jackye1995 I believe this is ready for review

rust/lance-table/src/format/manifest.rs
```diff
@@ -189,9 +189,9 @@ impl Manifest {
         &self,
         ref_name: Option<String>,
         ref_path: String,
+        ref_base_id: u32,
```
jackye1995, line 195: maybe we should call this `prepare_for_shallow_clone` since the result manifest is not complete yet.

rust/lance/src/index/vector/ivf.rs
```diff
@@ -2267,24 +2267,58 @@ mod tests {
         .await
         .unwrap();
 
-        let index = dataset
```
jackye1995, line 2270: are these changes related? Cannot connect why we need these changes

rust/lance/src/index/scalar.rs
```diff
@@ -445,7 +445,14 @@ pub async fn open_scalar_index(
     metrics: &dyn MetricsCollector,
 ) -> Result<Arc<dyn ScalarIndex>> {
     let uuid_str = index.uuid.to_string();
-    let index_store = Arc::new(LanceIndexStore::from_dataset(dataset, &uuid_str));
+    let index_dir = dataset.indice_files_dir(index)?.child(uuid_str.as_str());
+    let cache = dataset.metadata_cache.file_metadata_cache(&index_dir);
+    let index_store = Arc::new(LanceIndexStore::new(
```
jackye1995, line 450: if we initialize index_store like this, I think it will be used for both read and write, so if we train a new index, it will actually write to the old location?

rust/lance-table/src/format/manifest.rs
majin1102: I think the shallow_clone here means shallow_clone a manifest, not a dataset. Similar like new in Manifest it's a memory operation which generate a new manifest.

I think the trouble here is the confusion with Dataset.shallow_clone which return a dataset. Then how about new_ref or new_reference,  consistent with the params?

rust/lance/src/index/vector/ivf.rs
majin1102: These changes are introduced due to the dependency of open_index like this:

```
        let remapped = dataset_mut
            .open_vector_index(
                WellKnownIvfPqData::COLUMN,
                &new_uuid.to_string(),
                &NoOpMetricsCollector,
            )
            .await
            .unwrap();
```

The testcase would open index after building, now open_index demands index has been registered in dataset(The testcase worked well because open_index only needed uuid even though the index could not been seen in the dataset).

rust/lance/src/index/scalar.rs
majin1102: I don't think so. the index_store construction here is limited to opening indexes. Building indexes would write to the cloned dataset dir.

Scalar index building uses paths:
```
impl LanceIndexStoreExt for LanceIndexStore {
    fn from_dataset(dataset: &Dataset, uuid: &str) -> Self {
        let index_dir = dataset.indices_dir().child(uuid);
        let cache = dataset.metadata_cache.file_metadata_cache(&index_dir);
        Self::new(dataset.object_store.clone(), index_dir, Arc::new(cache))
    }
}
```

Vector index building uses  paths like this:

```
        IvfIndexBuilder::<FlatIndex, ScalarQuantizer>::new(
            dataset.clone(),
            column.to_owned(),
            dataset.indices_dir().child(uuid),
            params.metric_type,
            Box::new(shuffler),
            Some(ivf_params.clone()),
            Some(sq_params.clone()),
            (),
            frag_reuse_index,
        )?
        .build()
        .await?;
```

jackye1995: Ah I see, that makes sense, every time the merge creates a new dest_store. I modified your test and verified the behavior.

rust/lance-table/src/format/manifest.rs
jackye1995: I guess it's fine then, maybe just add some explanations the behavior with doc string for the function should be good.

majin1102:
Have added docs for Manifest.shallow_clone.

Thank you for supplementing the test cases!

jackye1995 (approved):
Looks good to me!

PR #4560, opened 2025-08-25
feat: add more metrics for FTS queries

BubbleCal:


rust/lance/src/io/exec/fts.rs
```diff
@@ -401,7 +437,22 @@ impl ExecutionPlan for FlatMatchQueryExec {
...
+            batch
+        })
+        .finally(move || {
+            metrics_clone2
+                .baseline_metrics
+                .elapsed_compute()
+                .add_elapsed(timer);
+            metrics_clone2.baseline_metrics.done();
+        });
         Ok(Box::pin(InstrumentedRecordBatchStreamAdapter::new(
```
westonpace, line 446: I think `InstrumentedRecordBatchStreamAdapter` will already add `baseline_metrics`?  Do you also need to record it here?

BubbleCal: removed

westonpace: Can we be consistent and use `InstrumentedRecordBatchStreamAdapter` everywhere?  For example: https://github.com/westonpace/lance/commit/fee17c8f56691492d59277104de087faddb573f2

LuQQiu (approved):

PR #4562, opened 2025-08-25
fix: restore behavior of projection plans from schema

westonpace:
In the past when we created a projection plan from a schema it was possible to mask out certain columns while still maintaining the overall structure.  This was only exposed via rust (no way to project with schema via python) but we (LanceDB) found at least one workflow relying on this behavior.

The behavior changed (unintentionally) in https://github.com/lancedb/lance/commit/eeea03c21e9c5082a1699c5eee3c14dd3d0ef60a and this PR restores the old behavior.  That being said, I don't particularly love that there are two different ways to specify a projection and it might be nice to update downstream workflows at some point or come up with some way to do masking via projection expressions.

### DIFF
--- a/rust/lance-datafusion/src/projection.rs
+++ b/rust/lance-datafusion/src/projection.rs
@@ -136,16 +136,66 @@ impl ProjectionPlan {
 
     /// Set the projection from a schema
     ///
-    /// This plan will have no complex expressions
+    /// This plan will have no complex expressions, the schema must be a subset of the dataset schema.
+    ///
+    /// With this approach it is possible to refer to portions of nested fields.
+    ///
+    /// For example, if the schema is:
+    ///
+    /// ```
+    /// {
+    ///   "metadata": {
+    ///     "location": {
+    ///       "x": f32,
+    ///       "y": f32,
+    ///     },
+    ///     "age": i32,
+    ///   }
+    /// }
+    /// ```
+    ///
+    /// It is possible to project a partial schema that drops `y` like:
+    ///
+    /// ```
+    /// {
+    ///   "metadata": {
+    ///     "location": {
+    ///       "x": f32,
+    ///     },
+    ///     "age": i32,
+    ///   }
+    /// }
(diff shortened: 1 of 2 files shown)


rust/lance-datafusion/src/projection.rs
```diff
@@ -136,16 +136,66 @@ impl ProjectionPlan {
...
+    ///
+    /// ```
+    /// {
+    ///   "metadata": {
+    ///     "location": {
+    ///       "x": f32,
+    ///     },
+    ///     "age": i32,
+    ///   }
+    /// }
```
wkalt, line 167: these seem to be getting executed or validated somehow

rust/lance-datafusion/src/projection.rs
```diff
@@ -136,16 +136,66 @@ impl ProjectionPlan {
...
-        )
+        // Calculate the physical projection directly from the schema
+        //
+        // The _rowid and _rowaddr columns will be recognized and added to the physical projection
+        //
+        // Any columns with an id of -1 (e.g. _rowoffset) will be ignored
+        let physical_projection = Projection::empty(base).union_schema(projection);
+        let mut must_add_row_offset = false;
+        // Now calculate the output expressions.  This will only reorder top-level columns.  We don't
+        // support reordering nested fields.
```
wkalt, line 180: what happens when a user tries to do this

rust/lance-datafusion/src/projection.rs
westonpace: Ah, yep.  Needed to add `ignore` or else it gets doctest'd but these are internalish methods so no need to be thorough on doc testing I think.

rust/lance-datafusion/src/projection.rs
westonpace: If would just be silently re-ordered.  E.g. they would ask for `metadata.location: {y, x}` and get back `metadata.location: {x, y}`

Issue #4516, opened 2025-08-20
feat: Support JSON as data type


eddyxu:
# Support JSON as data type

Users are expecting to store schema less json blobs in the dataset. In other systems, 

* Postgres  can use `jsonb` type
* Spark offers functions, such as `get_json_object` / `json_object_keys` function on json strings.
* Pyarrow has [JsonType](https://arrow.apache.org/docs/python/generated/pyarrow.JsonType.html) now

It'd be beneficial for Lance to have capability to store, access and search json objects.

- [x] Store JSON as one of the arrow field. 
- [x] Implement reading values by key or `json path`. 
- [ ] Create index on one of the scalar field.
- [ ] Search a scalar json field
- [ ] Create vector index on one of vector field.
- [ ] Search a vector json field
- [ ] Create full text index on one of the string field.
- [ ] Search a fulltext json field.

westonpace:
I think adding JSON to file format (e.g. variant or something, if needed) and adding JSON to the table format (JSON indexes) and adding JSON to the compute engine (JSON functions) can all be tackled independently.

The file format and compute engine changes are probably more well understood.

The table format changes are a little less understood to me.  There is maybe an easy path for some kind of read-only JSON column where a write means rewriting the entire JSON value.  Making a full schema-less table will be quite different.

Is a user going to want to change a JSON value for a row (e.g. an update)?  Will they want to add a new JSON column (to an existing JSON object)?

westonpace:
I assume if we are training a vector index on a vector field and that field isn't a vector type on some row (JSON does not enforce homogenous types) or if the vector has different dimensions then we just skip the field?

What should we use to determine the dimension of the JSON vector field?  Should it be the first value we encounter?

wjones127:
One good option is to join the effort in DataFusion to support the Parquet Variant data type. (It's just called Parquet Variant because that's where the standard was defined, not because it's tied to parquet.)

Similar to how we can add geospatial support by just including a few plugins (see demo from https://github.com/lancedb/lance/discussions/4482), we could probably add variant support with a few plugins.

Relevant packages:

* [parquet-variant](https://github.com/apache/arrow-rs/tree/main/parquet-variant)
* [parquet-variant-json](https://github.com/apache/arrow-rs/tree/main/parquet-variant-json)
* [parquet-variant-compute](https://github.com/apache/arrow-rs/tree/main/parquet-variant-compute)

wjones127:
If we want something more immediate, there is:

[...]

PR #4566, opened 2025-08-26
feat: add JSONB read/write support

Xuanwo:
Part of https://github.com/lancedb/lance/issues/4516

This PR will add basic JSONB read/write support so users can use them like an extension arrow type like `blob` and `bfloat16`.

```py
"""Test filtering and selecting with JSON columns."""

json_data = [
    {"user_id": 1, "preferences": {"theme": "dark"}},
    {"user_id": 2, "preferences": {"theme": "light"}},
    {"user_id": 3, "preferences": {"theme": "dark"}},
]

json_strings = [json.dumps(d) for d in json_data]
json_ext_array = JsonArray.from_json_strings(json_strings)

table = pa.table(
    {
        "id": pa.array([1, 2, 3]),
        "user_data": json_ext_array,
        "active": pa.array([True, False, True]),
    },
    schema=pa.schema(
        [
            pa.field("id", pa.int32()),
            json_field("user_data"),
            pa.field("active", pa.bool_()),
        ]
    ),
)

    dataset_path = Path(tmpdir) / "filter_test.lance"
    lance.write_dataset(table, dataset_path, data_storage_version="2.2")
    dataset = lance.dataset(dataset_path)

    # Test column selection
    selected = dataset.to_table(columns=["id", "user_data"])
    assert selected.num_columns == 2
    assert "active" not in selected.column_names
```

---


python/python/lance/_arrow/json.py
```diff
@@ -0,0 +1,103 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright The Lance Authors
+
+"""JSON support for Lance using JSONB binary format."""
+
+import pyarrow as pa
```
westonpace: Can we use https://arrow.apache.org/docs/python/generated/pyarrow.JsonType.html and skip this?

python/python/tests/test_json.py
```diff
@@ -0,0 +1,348 @@
...
+            ]
+        ),
+    )
+
+    # Write to Lance dataset
+    with tempfile.TemporaryDirectory() as tmpdir:
+        dataset_path = Path(tmpdir) / "json_test.lance"
+
+        # Write the dataset with version 2.2 (required for JSON support)
+        lance.write_dataset(table, dataset_path, data_storage_version="2.2")
```
westonpace, line 38: I don't think we actually need 2.2 if we're doing everything at the table level (originally I thought you were adding JSONB to file level)

python/python/tests/test_json.py
```diff
@@ -0,0 +1,348 @@
...
+        data_column = result_table.column("data")
+        for i, expected in enumerate(json_strings):
+            if expected is None:
+                assert data_column[i].as_py() is None
+            else:
+                # The data is stored as binary, need to decode
+                binary_data = data_column[i].as_py()
+                if binary_data is not None:
+                    # For now, we just verify the data is stored and retrieved
+                    assert binary_data is not None
```
westonpace: Is this actually binary data?  Ideally the user would get back strings since that is what they put in.  But we can do that in follow-up or later in this PR too?

python/python/tests/test_json.py
```diff
@@ -0,0 +1,348 @@
...
+        assert len(dataset.schema) == 5
+
+        result = dataset.to_table()
+        assert result.num_rows == 2
+        assert result.column("id").to_pylist() == [1, 2]
+        assert result.column("name").to_pylist() == ["Product A", "Product B"]
+        assert result.column("in_stock").to_pylist() == [True, False]
+
+
+def test_json_append_and_overwrite():
```
westonpace: Is there any particular reason we expect these things (json with other types or appending / overwriting json) to be different than any other data type?  I'm wondering if we need these tests?

python/python/tests/test_json.py
```diff
@@ -0,0 +1,348 @@
...
+        dataset = lance.write_dataset(overwrite_table, dataset_path, mode="overwrite")
+
+        # Verify overwrite
+        result = dataset.to_table()
+        assert result.num_rows == 1
+        assert result.column("id").to_pylist() == [3]
+
+
+def test_json_filter_and_select():
+    """Test filtering and selecting with JSON columns."""
```
westonpace: Unless this test case is actually testing filtering on the JSON column I'm not sure it's warranted.  We didn't make any changes to the scanner did we?

python/python/tests/test_json.py
```diff
@@ -0,0 +1,348 @@
...
+        data_column = result.column("optional_data")
+        assert data_column.null_count == 2
+        assert data_column.is_null().to_pylist() == [False, True, False, True, False]
+
+
+def test_json_large_documents():
+    """Test handling of large JSON documents."""
+
+    # Create a moderately large JSON document
+    large_doc = {
```
westonpace: Unless we are crossing the 2GB threshold or benchmarking I'm not sure this size is particularly interesting.

python/src/arrow.rs
```diff
@@ -88,3 +92,35 @@ pub fn bfloat16_array(values: Vec<Option<f32>>, py: Python<'_>) -> PyResult<PyOb
...
+        JSON_EXPORT_METADATA
+            .iter()
+            .map(|(k, v)| (k.to_string(), v.to_string()))
+            .collect(),
+    );
+
+    // Wrap JsonArray in a RecordBatch to preserve metadata through PyArrow conversion
+    let schema = Schema::new(vec![field]);
+    let batch = RecordBatch::try_new(Arc::new(schema), vec![Arc::new(json_array)])
+        .map_err(|err| PyValueError::new_err(format!("Failed to build array: {}", err)))?;
```
westonpace: Are we just adding field metadata to the array?  Why do we need to drop down to rust?

rust/lance-arrow/Cargo.toml
```diff
@@ -22,6 +22,7 @@ arrow-schema = { workspace = true }
 arrow-select = { workspace = true }
 bytes = { workspace = true }
 half = { workspace = true }
+jsonb = "0.5.3"
```
westonpace: Minor nit: Can we put this in the workspace cargo, even if it is only used in one place?  I find it is just simpler to have all versions in one spot.

rust/lance-arrow/src/json.rs
```diff
@@ -0,0 +1,365 @@
...
+
+impl TryFrom<LargeStringArray> for JsonArray {
+    type Error = ArrowError;
+
+    fn try_from(array: LargeStringArray) -> Result<Self, Self::Error> {
+        Self::try_from(&array)
+    }
+}
+
+impl TryFrom<&LargeStringArray> for JsonArray {
```
westonpace, line 236: Hmm, supporting both string and large string is ok, but not sure it is vital at the moment if it becomes an inconvenience.

rust/lance-core/src/datatypes.rs
```diff
@@ -202,6 +201,7 @@ impl TryFrom<&LogicalType> for DataType {
             "binary" => Some(Binary),
             "large_string" => Some(LargeUtf8),
             "large_binary" => Some(LargeBinary),
+            "json" => Some(LargeBinary),
```
westonpace, line 204: Does this mean JSON is always large binary?

rust/lance-datafusion/src/udf/json.rs
```diff
@@ -0,0 +1,144 @@
...
+}
+
+/// Extract value from JSONB using JSONPath
+fn extract_json_path(jsonb_bytes: &[u8], path: &str) -> Result<Option<String>> {
+    let json_path = jsonb::jsonpath::parse_json_path(path.as_bytes()).map_err(|e| {
+        datafusion::error::DataFusionError::Execution(format!("Invalid JSONPath: {}", e))
+    })?;
+
+    let raw_jsonb = jsonb::RawJsonb::new(jsonb_bytes);
+    match raw_jsonb.select_by_path(&json_path) {
```
westonpace, line 96: Cool :+1:

rust/lance/src/dataset.rs
```diff
@@ -552,6 +553,26 @@ impl Dataset {
...
+                    source: format!(
+                        "Cannot read dataset with JSON fields from file format version {}. \
+                         JSON fields require Lance file format version 2.2 or later.",
+                        storage_version
+                    )
+                    .into(),
+                    location: location!(),
+                });
+            }
+        }
```
westonpace, line 574: Can we remove this check for now since we aren't modifying file format?  Maybe once this feature stabilzes we can put it back (if we add variant type to file format) or handle it with a table feature flag.

rust/lance/src/dataset/write/insert.rs
```diff
@@ -363,6 +364,23 @@ impl<'a> InsertBuilder<'a> {
...
+                source: format!(
+                    "JSON fields require Lance file format version 2.2 or later. \
+                     Current version: {}. Please specify data_storage_version='2.2' in WriteParams",
+                    context.storage_version
+                )
+                .into(),
+                location: location!(),
+            });
+        }
+
```
westonpace, line 383: Same as above, let's remove this check for now

westonpace (approved):
I have some suggestions but overall I think this is a good starting point for moving forward on JSON.

Some of the test cases in test_json.py seem a little extraneous.

Also, someday, I'd love to get rid of logical type from the lance table format and just use Arrow there too but definitely we should save that for a future PR :laughing:

python/python/tests/test_json.py
Xuanwo: Let me remove this.

rust/lance-core/src/datatypes.rs
Xuanwo: Yes for now. I'm open to provide `Binary` support for small json. Maybe it's worth a seperate issue?

rust/lance/src/dataset.rs
Xuanwo: I'm hesitant about this because the binary format for jsonb is not quite stable and is still evolving. My current plan is to change into pg compatible jsonb format in the future.

python/python/lance/_arrow/json.py
Xuanwo: Yep! I have updated the tests to use `JsonType` directly.

python/python/tests/test_json.py
Xuanwo: Correct, let me just remove it.

python/src/arrow.rs
Xuanwo: Removed those functions entirely.

rust/lance-arrow/Cargo.toml
Xuanwo: Fixed.

python/python/tests/test_json.py
Xuanwo: Yep, that's incorrect. I have it fixed.

rust/lance/src/dataset.rs
Xuanwo: Tracked at https://github.com/lancedb/lance/issues/4575, I'm open to create a new PR to get it removed. And for now, I will merge this PR first.

PR #4567, opened 2025-08-26
fix: get rust-benchmark.yml working again

cmccabe:
Get rust-benchmark.yml working again by installing the correct Linux dependencies.


eddyxu (approved):
Pending CI

PR #4569, opened 2025-08-26
fix(doc): fix take_blobs python doc wrong parameters format

yanghua:


python/python/lance/dataset.py
```diff
@@ -1182,6 +1182,7 @@ def take_blobs(
         object. For more details, see :py:class:`lance.BlobFile`.
 
         Exactly one of ids, addresses, or indices must be specified.
+
```
yanghua, line 1185: The Python [doc](https://lancedb.github.io/lance-python-doc/all-modules.html#lance.LanceDataset.take_blobs) of `take_blobs` looks chaotic. 

```
Exactly one of ids, addresses, or indices must be specified. :param blob_column: The name of the blob column to select. :type blob_column: str :param ids: row IDs to select in the dataset. :type ids: Integer Array or array-like :param addresses: The (unstable) row addresses to select in the dataset. :type addresses: Integer Array or array-like :param indices: The offset / indices of the row in the dataset. :type indices: Integer Array or array-like
```

I guess it missed a blank line.

yanghua:
cc @Xuanwo

Xuanwo (approved):
Thank you for this fix!

PR #4571, opened 2025-08-26
feat: add min/max matched metric for scalar queries

westonpace:


rust/lance-core/src/utils/mask.rs
```diff
@@ -47,6 +47,45 @@ impl RowIdMask {
...
+                block_list_len = Some(len);
+            } else {
+                return None;
+            }
+        }
+
+        match (allow_list_len, block_list_len) {
+            (Some(allow_list_len), Some(block_list_len)) => Some(allow_list_len - block_list_len),
+            (Some(allow_list_len), None) => Some(allow_list_len),
+            (None, Some(_)) => Some(0),
```
BubbleCal: should this be None?
i think it's `num_rows - block_list_len` but `num_rows` is unknown right?

BubbleCal (approved):
just a question

rust/lance-core/src/utils/mask.rs
westonpace: Good thinking.  `None` might be clearer and `Some(0)` isn't giving us any useful information anyways.

westonpace: That made the logic much simpler too because now we can return None if we don't know the allow list len :slightly_smiling_face:

PR #4577, opened 2025-08-27
feat: add UDFs for json

Xuanwo:
Part of https://github.com/lancedb/lance/issues/4516

This PR will add UDFs for JSON so users can read JSON like:

```python
ds_path = Path(tmpdir) / "nested_json_test.lance"
lance.write_dataset(table, ds_path, data_storage_version="2.2")
dataset = lance.dataset(ds_path)

# Access nested fields using json_get recursively
# First get user, then profile, then name
result = dataset.to_table(
    filter="json_get_string(json_get(json_get(data, 'user'), 'profile'), 'name') = 'Alice'"
)
assert result.num_rows == 1
assert result["id"][0].as_py() == 1

# Or use JSONPath for deep access
result = dataset.to_table(
    filter="json_extract(data, '$.user.profile.settings.theme') = '\"dark\"'"
)
assert result.num_rows == 1
assert result["id"][0].as_py() == 1
```

---

### DIFF
--- a/rust/lance-arrow/src/json.rs
+++ b/rust/lance-arrow/src/json.rs
@@ -7,14 +7,17 @@ use std::convert::TryFrom;
 use std::sync::Arc;
 
 use arrow_array::builder::LargeBinaryBuilder;
-use arrow_array::{Array, ArrayRef, LargeBinaryArray, LargeStringArray, StringArray};
+use arrow_array::{Array, ArrayRef, LargeBinaryArray, LargeStringArray, RecordBatch, StringArray};
 use arrow_data::ArrayData;
-use arrow_schema::{ArrowError, DataType, Field as ArrowField};
+use arrow_schema::{ArrowError, DataType, Field as ArrowField, Schema};
 
 use crate::ARROW_EXT_NAME_KEY;
 
-/// Arrow extension type name for JSON data
+/// Arrow extension type name for JSON data (Lance internal)
 pub const JSON_EXT_NAME: &str = "lance.json";
+
+/// Arrow extension type name for JSON data (Arrow official)
+pub const ARROW_JSON_EXT_NAME: &str = "arrow.json";
 
 /// Check if a field is a JSON extension field (Lance internal JSONB storage)
 pub fn is_json_field(field: &ArrowField) -> bool {
@@ -23,6 +26,17 @@ pub fn is_json_field(field: &ArrowField) -> bool {
             .metadata()
             .get(ARROW_EXT_NAME_KEY)
             .map(|name| name == JSON_EXT_NAME)
+            .unwrap_or_default()
+}
+
+/// Check if a field is an Arrow JSON extension field (PyArrow pa.json() type)
+pub fn is_arrow_json_field(field: &ArrowField) -> bool {
+    // Arrow JSON extension type uses Utf8 or LargeUtf8 as storage type
+    (field.data_type() == &DataType::Utf8 || field.data_type() == &DataType::LargeUtf8)
+        && field
+            .metadata()
+            .get(ARROW_EXT_NAME_KEY)
+            .map(|name| name == ARROW_JSON_EXT_NAME)
             .unwrap_or_default()
 }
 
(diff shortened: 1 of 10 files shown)


Xuanwo:
> Quick initial question. How do these compare to https://github.com/datafusion-contrib/datafusion-functions-json ?

`datafusion-functions-json` has: 

```shell
json_get::json_get_udf(),
json_get_bool::json_get_bool_udf(),
json_get_float::json_get_float_udf(),
json_get_int::json_get_int_udf(),
json_get_json::json_get_json_udf(),
json_get_array::json_get_array_udf(),
json_as_text::json_as_text_udf(),
json_get_str::json_get_str_udf(),
json_contains::json_contains_udf(),
json_length::json_length_udf(),
json_object_keys::json_object_keys_udf(),
```

Perhaps it would be better to implement them directly on `jsonb` rather than converting to JSON and using those UDFs. What do you think?

rust/lance-arrow/src/json.rs
```diff
@@ -309,8 +309,9 @@ fn get_json_path(
 ) -> Result<Option<String>, Box<dyn std::error::Error>> {
     let json_path = jsonb::jsonpath::parse_json_path(path.as_bytes())?;
     let raw_jsonb = jsonb::RawJsonb::new(jsonb_bytes);
+    let mut selector = jsonb::jsonpath::Selector::new(raw_jsonb);
 
-    match raw_jsonb.select_by_path(&json_path) {
+    match selector.select_values(&json_path) {
```
westonpace, line 328: Just out of curiosity, why this change?  Is it for simplicity or does it actually affect semantics?

rust/lance-datafusion/src/udf.rs
```diff
@@ -15,7 +15,16 @@ mod json;
...
+    // JSON functions
     ctx.register_udf(json::json_extract_udf());
+    ctx.register_udf(json::json_exists_udf());
+    ctx.register_udf(json::json_get_udf());
+    ctx.register_udf(json::json_get_string_udf());
+    ctx.register_udf(json::json_get_int_udf());
+    ctx.register_udf(json::json_get_float_udf());
+    ctx.register_udf(json::json_get_bool_udf());
+    ctx.register_udf(json::json_array_contains_udf());
+    ctx.register_udf(json::json_array_length_udf());
```
westonpace, line 27: Let's very soon create a PR to document these UDFs somewhere?

rust/lance-datafusion/src/udf/json.rs
```diff
@@ -106,18 +107,1037 @@ fn extract_json_path(jsonb_bytes: &[u8], path: &str) -> Result<Option<String>> {
...
+    )
+}
+
+/// Implementation of json_exists function with ColumnarValue
+fn json_exists_columnar_impl(args: &[ColumnarValue]) -> Result<ColumnarValue> {
+    let arrays: Vec<ArrayRef> = args
+        .iter()
+        .map(|arg| match arg {
+            ColumnarValue::Array(arr) => arr.clone(),
+            ColumnarValue::Scalar(scalar) => scalar.to_array().unwrap(),
```
westonpace: It would be great if we could avoid broadcasting the scalars here.  Let's create a follow-up?

rust/lance-datafusion/src/udf/json.rs
```diff
@@ -106,18 +107,1037 @@ fn extract_json_path(jsonb_bytes: &[u8], path: &str) -> Result<Option<String>> {
...
+    let path_array = args[1]
+        .as_any()
+        .downcast_ref::<StringArray>()
+        .ok_or_else(|| {
+            datafusion::error::DataFusionError::Execution(
+                "Second argument must be String".to_string(),
+            )
+        })?;
+
+    let mut builder = BooleanBuilder::new();
```
westonpace: Can we initialize this with a starting capacity?

rust/lance-datafusion/src/udf/json.rs
```diff
@@ -106,18 +107,1037 @@ fn extract_json_path(jsonb_bytes: &[u8], path: &str) -> Result<Option<String>> {
...
+    }
+
+    Ok(Arc::new(builder.finish()))
+}
+
+/// Check if a JSONPath exists in JSONB
+fn check_json_path_exists(jsonb_bytes: &[u8], path: &str) -> Result<bool> {
+    let json_path = jsonb::jsonpath::parse_json_path(path.as_bytes()).map_err(|e| {
+        datafusion::error::DataFusionError::Execution(format!("Invalid JSONPath: {}", e))
+    })?;
```
westonpace: This is just a minor nit but it seems a little odd that we create an error in `check_json_path_exists` and then immediately re-wrap the error in `json_exists_impl`.  Could we just create the error with the message we want the first time and then use `?`?

rust/lance-datafusion/src/udf/json.rs
```diff
@@ -106,18 +107,1037 @@ fn extract_json_path(jsonb_bytes: &[u8], path: &str) -> Result<Option<String>> {
...
+
+    let raw_jsonb = jsonb::RawJsonb::new(jsonb_bytes);
+    let mut selector = jsonb::jsonpath::Selector::new(raw_jsonb);
```
westonpace, line 327: Would it be more efficient to create the selector once and reuse it instead of creating a selector for every value?

rust/lance-datafusion/src/udf/json.rs
```diff
@@ -106,18 +107,1037 @@ fn extract_json_path(jsonb_bytes: &[u8], path: &str) -> Result<Option<String>> {
...
+    Ok(Arc::new(builder.finish()))
+}
+
+/// Check if a JSONPath exists in JSONB
+fn check_json_path_exists(jsonb_bytes: &[u8], path: &str) -> Result<bool> {
+    let json_path = jsonb::jsonpath::parse_json_path(path.as_bytes()).map_err(|e| {
+        datafusion::error::DataFusionError::Execution(format!("Invalid JSONPath: {}", e))
+    })?;
+
+    let raw_jsonb = jsonb::RawJsonb::new(jsonb_bytes);
```
westonpace, line 326: Do we know if this a zero-copy operation?

rust/lance-datafusion/src/udf/json.rs
```diff
@@ -106,18 +107,1037 @@ fn extract_json_path(jsonb_bytes: &[u8], path: &str) -> Result<Option<String>> {
...
+    let mut selector = jsonb::jsonpath::Selector::new(raw_jsonb);
+    match selector.exists(&json_path) {
+        Ok(exists) => Ok(exists),
+        Err(_) => Ok(false),
```
westonpace: Should we return the error instead of false?  This indicates invalid JSON or an invalid path I think.

rust/lance-datafusion/src/udf/json.rs
```diff
@@ -106,18 +107,1037 @@ fn extract_json_path(jsonb_bytes: &[u8], path: &str) -> Result<Option<String>> {
...
+        })?;
+
+    let key_array = args[1]
+        .as_any()
+        .downcast_ref::<StringArray>()
+        .ok_or_else(|| {
+            datafusion::error::DataFusionError::Execution(
+                "Second argument must be String".to_string(),
+            )
+        })?;
```
westonpace: There's a lot of repetition among the UDFs here.  I wonder if this is something an LLM could help us avoid by refactoring into some helper methods?

rust/lance-datafusion/src/udf/json.rs
```diff
@@ -106,18 +107,1037 @@ fn extract_json_path(jsonb_bytes: &[u8], path: &str) -> Result<Option<String>> {
...
+        Err(e) => {
+            return Err(datafusion::error::DataFusionError::Execution(format!(
+                "Failed to get field: {}",
+                e
+            )))
+        }
+    }
+
+    // Try as array index
+    if let Ok(index) = key.parse::<usize>() {
```
westonpace: It might be more efficient to determine if the key is a field or an integer once instead of each value?  Although I guess it could be an integer-named field and so vary from value to value :stuck_out_tongue:

rust/lance-datafusion/src/udf/json.rs
```diff
@@ -106,18 +107,1037 @@ fn extract_json_path(jsonb_bytes: &[u8], path: &str) -> Result<Option<String>> {
...
+
+    Ok(Arc::new(builder.finish()))
+}
+
+/// Get a field value from JSONB (returns JSONB bytes)
+fn get_json_field(jsonb_bytes: &[u8], key: &str) -> Result<Option<Vec<u8>>> {
+    let raw_jsonb = jsonb::RawJsonb::new(jsonb_bytes);
+
+    // Try as object field first
+    match raw_jsonb.get_by_name(key, false) {
```
westonpace: What's this bool here, case sensitivity?

rust/lance-datafusion/src/udf/json.rs
```diff
@@ -106,18 +107,1037 @@ fn extract_json_path(jsonb_bytes: &[u8], path: &str) -> Result<Option<String>> {
...
+                ))
+            } else {
+                Err(datafusion::error::DataFusionError::Execution(format!(
+                    "Cannot convert string '{}' to boolean",
+                    s
+                )))
+            }
+        }
+    }
+}
```
westonpace: These parsing functions all seem to follow a general pattern too.  I wonder if there is some reuse here.

rust/lance-arrow/src/json.rs
Xuanwo: For better readability.

rust/lance-datafusion/src/udf.rs
Xuanwo: Thank you, will do!

Xuanwo: Tracked at https://github.com/lancedb/lance/issues/4590

rust/lance-datafusion/src/udf/json.rs
Xuanwo: DONE

rust/lance-datafusion/src/udf/json.rs
Xuanwo: Fixed.

rust/lance-datafusion/src/udf/json.rs
Xuanwo: Fixed.

rust/lance-datafusion/src/udf/json.rs
Xuanwo: Yes：

```rust
    pub fn new(data: &'a [u8]) -> Self {
        Self { data }
    }
```

rust/lance-datafusion/src/udf/json.rs
Xuanwo: Added.

rust/lance-datafusion/src/udf/json.rs
Xuanwo: Replaced by just using jsonb's API.

rust/lance-datafusion/src/udf/json.rs
Xuanwo: Yep, `ignore_case`.

rust/lance-datafusion/src/udf/json.rs
Xuanwo: No really, it's just a consutrcutor.

Xuanwo:
Catching some bugs from python side, working on it.

Xuanwo:
I really don't like how I'm handling JSON and JSONB conventions in the code. Maybe I over-optimized too early, adding unnecessary complexity just to allow we can perform UDFs on JSONB instead of the entire JSON.

I'm starting to reconsider @westonpace's suggestion to keep JSONB as part of the file format instead of treating it as a logical type. While it's working for now, we might as well keep moving forward and see how it holds up in the long run. We can always make adjustments later.

I'll merge this PR first and do anyother changes in follow-up PRs. This will also unlock us to implement index over JSON.

Xuanwo:
The follow-up tracked at https://github.com/lancedb/lance/issues/4595

PR #4578, opened 2025-08-27
feat: support build FTS index distributedly

### LINKED ISSUE #4514: Feature: Add create_fragment_inverted_index for distributed inverted index building
Related with https://github.com/lancedb/lance-ray/issues/12 ,  As the Inverted index/FTS building is quite slow in my large datasets (example dataset is 1B, around 7TB), I am seeking for build the inverted index using distributed way. 

To leveraging Ray for parallel building inverted index, I am going to add the below Interfaces in lance.  The core changes include new public interfaces and a distributed indexing workflow. 

### New Interfaces

Three new public interfaces will be added to support distributed index creation in lance dataset.  I will propose a PR within next week.

1.  **`create_fragment_index(...)`**: A lower-level API that creates an index on a specified subset of data fragments. This is used by the distributed workers to build indexes on their assigned data partitions. It accepts fragment IDs and index specifications.
2.  **`merge_inverted_index_metadata(...)`**: After individual index fragments are built by the workers, this function is called to merge the metadata from all the partial indexes into a single, cohesive index metadata file for the dataset.

The methods name can be changed if any better suggestions.

### Distributed  Inverted Index Indexing Process
The distributed index building process will be operated in three main stages:

1.  **Fragment level splitting**: The `build_distributed_fts_index` function (maybe put in the lance-ray project) begins by retrieving all fragment IDs from the dataset. It then divides these fragments into balanced batches based on the specified number of `num_workers`. Each batch is a partition that will be processed independently.

2.  **Distributed Task Execution**: For each partition, a Ray remote task (`_build_fragment_index_task`) is launched. These tasks run in parallel on the Ray cluster, with each worker calling `create_fragment_index` to build an index for its assigned fragments.

3.  **Index Consolidation**: Once all parallel tasks are complete, the main process (ray head) gathers the results and calls `merge_inverted_index_metadata` to consolidate the metadata.lance from the individual index parts. Finally, it commits the new index to the dataset, making it ready for use.

### Following Work
After these done, I will then add  **`build_distributed_fts_index(...)`** in lance-ray project,  The main entry point for building a distributed Full-Text Search (FTS) or Inverted index. It takes the dataset, column name, and number of workers as primary inputs and orchestrates the entire distributed process using Ray.

chenghao-guo:
Close #4514
Related with https://github.com/lancedb/lance-ray/issues/12

This PR introduces distributed fts index capabilities enabling parallel index creation across multiple fragments. 

### New added methods
- `execute_uncommitted()`: Creates index metadata without dataset commitment, returning `IndexMetadata` for distributed coordination. Chainable method specifying target fragment IDs for selective indexing as suggested by Will Jones in discussion https://github.com/lancedb/lance/issues/4514#issuecomment-3207183353
```
              let partial_index=CreateIndexBuilder::new(&mut dataset, &["text"], IndexType::Inverted, &params)
                    .name("distributed_index".to_string())
                    .fragments(vec![fragment_id])
                    .fragment_uuid(shared_uuid.clone())
                    .execute_uncommitted()
                    .await?;
```

- `merge_index_metadata()`: Merges distributed index metadata from multiple workers into consolidated final metadata

### Distributed Workflow
1. **Split and Parallel Phase**:  Ray header distributes the fragments to different workers. (We may distribute the fragments evenly to different workers by fragment statistics, which will be implemented in lance-ray connector)
Ray Workers call `execute_uncommitted()` on specific fragments using `fragments()` method. Shared UUID via `fragment_uuid()` ensures consistent index identity.
2. **Merge Phase**: `merge_index_metadata()` consolidates partition metadata files (`part_*_metadata.lance`)
3. **Commit Phase**: Final index commitment with unified metadata

**Example**
The workflow example can be found in `test_distribute_fts_index_build` in [test_scalar_index.py](https://github.com/lancedb/lance/pull/4578/files#diff-a95edaddaa3a260e498c04e10f073261bdc529cd4f47b928ad80274754af0548R1964-R2021).

**Following work after this PR:**
The distributed index building workflow PR will be proposed to lance-ray connector.
lance-ray draft PR https://github.com/lancedb/lance-ray/pull/45

**Other implementation details on fragment_mask**
 Optional mask with fragment_id in high 32 bits. When provided,
only partitions whose partition id matches this fragment will be included.
The fragment mask is constructed as `(fragment_id as u64) << 32`.


rust/lance-index/src/scalar/inverted/builder.rs
```diff
@@ -728,3 +791,99 @@ pub(crate) fn posting_file_path(partition_id: u64) -> String {
...
+        })?;
+
+        // Only process the new format (u64 array), properly handle JSON deserialization errors
+        let partition_ids: Vec<u64> =
+            serde_json::from_str(partitions_str).map_err(|e| Error::Index {
+                message: format!("Failed to parse partitions from file {}: {}", file_name, e),
+                location: location!(),
+            })?;
+
+        all_partitions.extend(partition_ids);
```
BubbleCal, line 815: I think each indexing worker/fragment would have partition IDs starting from 0, we need to map each new partition id to `all_partitions.len() + part_id`?

python/python/tests/test_scalar_index.py
```diff
@@ -1839,3 +1839,291 @@ def test_fts_backward_v0_27_0(tmp_path: Path):
...
+    # Verify we have the expected number of fragments
+    fragments = ds.get_fragments()
+    assert len(fragments) == num_fragments, (
+        f"Expected {num_fragments} fragments, got {len(fragments)}"
+    )
+
+    return ds
+
+
+def test_distribute_fts_index_build(tmp_path):
```
BubbleCal, line 2819: it would be nice if we can add a helper function that builds the FTS index in distributed way, then we can add it into all existing FTS tests, to make sure the 2 ways will always get the same result

BubbleCal (changes requested):
This is a great work!
Just some questions

python/python/tests/test_scalar_index.py
chenghao-guo: Got it, thank you for your great suggestions. I will add them later.

rust/lance-index/src/scalar/inverted/builder.rs
chenghao-guo: Thanks for the suggestion — that’s convincing. I’ll adjust the partition ID assignment logic and make the corresponding changes.

python/python/tests/test_scalar_index.py
chenghao-guo: Done. Added helper function `assert_distributed_fts_consistency` and applied it to all existing FTS tests. Validated that distributed execution produces results consistent with single-node FTS index searches.

rust/lance-index/src/scalar/inverted/builder.rs
chenghao-guo: > I think each indexing worker/fragment would have partition IDs starting from 0, we need to map each new partition id to `all_partitions.len() + part_id`?

Hi @BubbleCal , thanks for taking a look. I want to confirm that I’m interpreting the implementation correctly.

In the previous design, each partition ID was assigned using the mask `fragment_id << 32` to ensure isolation across fragments and avoid conflicts. As a result, partition IDs are not sequential (i.e., not 0, 1, 2, 3, …).

Based on the suggestions, in the final stage, I reorder and rename all fragment partitions and then reassign partition IDs starting from 0. While this entails some rename operations, I’m not sure this is the best approach. I can’t assign sequential partition IDs 0,1,2... at fragment-index creation time because that work happens on different machines (e.g., Ray workers). Therefore, the only practical option I see is to sort the IDs and remap them in the last stage in a coordinator (e.g. Ray header)

I’ve made some changes to the function accordingly—please see the diff here:
- https://github.com/lancedb/lance/pull/4578/files#diff-613f7dde20f16c5577e9f69e63d58c5fe3601f7c2053c831bc8d2b7562b56d3bR833-R908

I’d appreciate any feedback or alternative suggestions if you think there’s a better way to handle this.

chenghao-guo:
Rebased to resolve conflict.

rust/lance-index/src/scalar/inverted/builder.rs
BubbleCal: This makes sense to me, TBH I didn't realize that the partition IDs were encoded by `frag_id << 32 + i`, but I think the current reorder and re-assign design is still better, just in case we will probably build single index over multiple fragments (e.g. there are many small fragments?)

chenghao-guo: @BubbleCal Yes, I completely agree with your idea. Using “reorder” and “reassign” is definitely clearer, since the generated FTS index is almost the same as the single-node index.

My only concern is that the renaming and overall code structure may not be very elegant, which could make the code harder for others to understand. To mitigate this, I’ve added inline comments for clarity. I’ll likely refactor the code structure in the near future.

BTW, kindly let me know if any other suggestions, or if I could get your approval to merge the PR.

python/python/tests/test_scalar_index.py
```diff
@@ -721,6 +721,15 @@ def test_fts_score(tmp_path):
     assert results.num_rows == 3
     assert results["id"].to_pylist() == [3, 2, 1]
 
+    # Assert distributed FTS produces the same BM25 scores as single-machine
+    assert_distributed_fts_consistency(
+        data,
+        "text",
+        "lance search text",
+        tmp_path,
+        index_params={"with_position": False},
```
BubbleCal, line 730: NIT, we've changed the default value to `False` for `with_position`, so no need to specify it here. we can remove this in the future PR

BubbleCal (approved):
This work is amazing, thanks!

python/python/tests/test_scalar_index.py
chenghao-guo: Got it , thanks !

PR #4581, opened 2025-08-27
feat: expose open_session in python

wojiaodoubao:


rust/lance/src/dataset/fragment/session.rs
```diff
@@ -14,6 +14,7 @@ use std::sync::Arc;
 /// internal states instead of creating new ones each time.
 ///
 /// This API works well for users making repeated requests over the same projected schema.
+#[derive(Clone, Debug)]
```
westonpace: Minor nit: If there are a lot of deleted ids it might be a bit costly to clone.  Is the `Clone` needed since you have `Arc<FragmentSession>`?

westonpace (approved):
Neat feature!

One minor nit but will merge tomorrow (can always clean it up in follow-up).

rust/lance/src/dataset/fragment/session.rs
wojiaodoubao: You are right, there's no need to clone. Let me remove it, thank you for your suggestion!

PR #4589, opened 2025-08-28
feat(rust): support refresh frag bitmap for index after updating when…

yanghua:
… enable stable rowid

### DIFF
--- a/rust/lance/src/dataset/write/update.rs
+++ b/rust/lance/src/dataset/write/update.rs
@@ -539,7 +539,6 @@ mod tests {
     use lance_io::object_store::ObjectStoreParams;
     use lance_linalg::distance::MetricType;
     use object_store::throttle::ThrottleConfig;
-    use roaring::RoaringBitmap;
     use rstest::rstest;
     use tempfile::{tempdir, TempDir};
     use tokio::sync::Barrier;


rust/lance/src/dataset/transaction.rs
```diff
@@ -210,8 +211,12 @@ pub enum Operation {
         new_fragments: Vec<Fragment>,
         /// The fields that have been modified
         fields_modified: Vec<u32>,
```
yanghua, line 213: Would like to confirm if this field is used to track which fields' values have been updated. Or just the field itself has been modified(add, alter, delete). There are two different behavior: https://github.com/lancedb/lance/blob/998e4fe387893c79bad6d4ba9a1ba3e4bc1e0979/rust/lance/src/dataset/write/update.rs#L400 

and

https://github.com/lancedb/lance/blob/998e4fe387893c79bad6d4ba9a1ba3e4bc1e0979/rust/lance/src/dataset/write/merge_insert.rs#L1353

rust/lance/src/dataset/write/merge_insert.rs
```diff
@@ -1358,7 +1359,9 @@ impl MergeInsertJob {
                 updated_fragments,
                 new_fragments,
                 fields_modified,
+                value_updated_fields: vec![],
                 mem_wal_to_flush: self.params.mem_wal_to_flush,
+                update_mode: Some(VerticalFullSchema),
```
yanghua: Actually, it belongs `VerticalPartialSchema`

rust/lance/src/index/vector/ivf/v2.rs
```diff
@@ -980,16 +980,18 @@ mod tests {
...
-            .update_where(&format!("id >= {} and id<{}", half_rows, half_rows + 50))
-            .unwrap()
-            .set("id", &format!("{}+id", NUM_ROWS))
-            .unwrap()
-            .build()
-            .unwrap()
-            .execute()
-            .await
-            .unwrap();
+        let update_result = Box::pin(
```
yanghua: Change this to fix large future size overflow.

rust/lance/src/dataset/transaction.rs
yanghua: @wjones127 Could you please help answer this question?

rust/lance/src/dataset/write/merge_insert.rs
yanghua: @wjones127 It seems that even if it's a non-full-schema merge-insert, we still can not distinguish the updated fields(may contain the `on` field).

yanghua:
@wjones127 When you have time, I hope you'll be able to answer. Very appreciated!

jbapple:
@yanghua this is still marked as "Draft". Is that intentional?

yanghua:
> @yanghua this is still marked as "Draft". Is that intentional?

@jbapple Yeah, still need some questions, I am waiting for the answer from @wjones127

rust/lance/src/dataset/transaction.rs
wjones127: Really the relevant place to look is how it's handled in the commit:

https://github.com/lancedb/lance/blob/575272a0e0fb88f73bcc5b95647827acb5ebf350/rust/lance/src/dataset/transaction.rs#L1755-L1757

The first one up there is empty because the rows that are updated in the fragment are moved: deleted and appended. So it's okay if an index still keeps that fragment in it's fragment bitmap, since the old versions of the rows will be masked out.

The second case is different: we update the fragments in place. To do the update, we tombstone the existing columns in the original data files. Then we append new data files with the updated column to each fragment. So the same fragments have new values for the rows. If one of these fields that has been "modified in-place" has an index, we need to remove that field from the fragment bitmap. Or at least mask it out somehow.

protos/transaction.proto
```diff
@@ -194,6 +194,26 @@ message Transaction {
...
     /// The MemWAL (pre-image) that should be marked as merged after this transaction
     MemWalIndexDetails.MemWal mem_wal_to_merge = 5;
+    /// The field ids that their values have been updated
+    repeated uint32 value_updated_fields = 6;
+    // The mode of update
+    UpdateMode update_mode = 7;
+  }
+
+  // The mode of update operation
+  enum UpdateMode {
```
wjones127, line 205: I worry "vertical" and "horizontal" can be confusing. How about:

* `REWRITE_ROWS`: rows are deleted in current fragments and rewritten in new fragments. This is most optimal when the majority of columns are being rewritten or only a few rows are being updated.
* `REWRITE_COLUMNS`: within each fragment, columns are fully rewritten and inserted as new data files. Old versions of columns are tombstoned. This is most optimal when most rows are affected but a small subset of columns are affected.

wjones127: It seems like whether it's full or partial schema is a separate concern, no?

rust/lance/src/index/vector/ivf/v2.rs
wjones127: Hmm I wonder if `UpdateBuilder::execute()` should return a boxed future instead. If we are getting this issue in our unit test, then I'm sure callers of the API will get something similar.

rust/lance/src/dataset/transaction.rs
```diff
@@ -1752,6 +1796,33 @@ impl Transaction {
...
+            let index_covers_modified_field = index
+                .fields
+                .iter()
+                .any(|field_id| fields_modified_set.contains(&u32::try_from(*field_id).unwrap()));
+
+            if !index_covers_modified_field {
+                if let Some(fragment_bitmap) = &mut index.fragment_bitmap {
+                    for fragment_id in pure_update_frag_ids.iter().map(|f| *f as u32) {
+                        fragment_bitmap.insert(fragment_id);
+                    }
```
wjones127: Can't you only add these if the index covered the fragments where the updates came from?

For example, say I have fragments 1 and 2. And an index covers 1. Then I run an update, which updates some rows from frags 1 and 2, making fragment 3. You can't add 3, because it contains some rows from frag 2 which were never indexed. Right?

rust/lance/src/index/vector/ivf/v2.rs
wjones127: Yeah I would do it right around here:

https://github.com/lancedb/lance/blob/a72cc51ca047d687491ad2972fa6efba0ff43c98/rust/lance/src/dataset/write/update.rs#L278

rust/lance/src/dataset/transaction.rs
yanghua: thanks, I was confused by the naming, now I understand. Actually, the `fields_modified` field is only used to track fields for realizing the function of `prune_updated_fields_from_indices`.

protos/transaction.proto
yanghua: > I worry "vertical" and "horizontal" can be confusing. How about:
> 
> * `REWRITE_ROWS`: rows are deleted in current fragments and rewritten in new fragments. This is most optimal when the majority of columns are being rewritten or only a few rows are being updated.
> * `REWRITE_COLUMNS`: within each fragment, columns are fully rewritten and inserted as new data files. Old versions of columns are tombstoned. This is most optimal when most rows are affected but a small subset of columns are affected.

After thinking, it seems there is no better naming. Both LGTM.

rust/lance/src/dataset/transaction.rs
yanghua: Good catch, have fixed this issue and added a test case.

rust/lance/src/dataset/transaction.rs
```diff
@@ -212,6 +213,11 @@ pub enum Operation {
         fields_modified: Vec<u32>,
         /// The MemWAL (pre-image) that should be marked as merged after this transaction
         mem_wal_to_merge: Option<MemWal>,
+        /// The fields that used to judge whether to preserve the new frag's id into
+        /// the frag bitmap of the specified indices.
+        fields_for_preserving_frag_bitmap: Vec<u32>,
```
yanghua, line 218: To avoid confusion, name to this.

yanghua:
@wjones127 This PR is ready for review, PTAL.

yanghua:
@wjones127 Any input?

rust/lance/src/dataset/write/update.rs
```diff
@@ -1048,4 +1069,273 @@ mod tests {
...
+            .iter()
+            .find(|idx| idx.name == "vec_idx")
+            .unwrap();
+
+        assert_eq!(str_index.fragment_bitmap.as_ref().unwrap().len(), 2);
+        assert!(str_index.fragment_bitmap.as_ref().unwrap().contains(0));
+        assert!(str_index.fragment_bitmap.as_ref().unwrap().contains(1));
+        assert_eq!(vec_index.fragment_bitmap.as_ref().unwrap().len(), 2);
+        assert!(vec_index.fragment_bitmap.as_ref().unwrap().contains(0));
+        assert!(vec_index.fragment_bitmap.as_ref().unwrap().contains(1));
```
wjones127: Instead of having to make all these assertions, maybe it would just be easier to directly assert equality:
```suggestion
        assert_eq!(str_index.fragment_bitmap.as_ref(), &RoaringBitmap::from([0, 1]));
        assert_eq!(vec_index.fragment_bitmap.as_ref(), &RoaringBitmap::from([0, 1]));
```

rust/lance/src/dataset/write/merge_insert.rs
```diff
@@ -4520,4 +4535,256 @@ MergeInsert: on=[id], when_matched=UpdateAll, when_not_matched=InsertAll, when_n
...
+        // the frag bitmap of the index for 'value' field should not been updated
+        let value_bitmap = updated_value_index.fragment_bitmap.as_ref().unwrap();
+        assert_eq!(value_bitmap.len(), 2);
+        assert!(value_bitmap.contains(0));
+        assert!(value_bitmap.contains(1));
+
+        let vec_bitmap = updated_vec_index.fragment_bitmap.as_ref().unwrap();
+        assert_eq!(vec_bitmap.len(), 2);
+        assert!(vec_bitmap.contains(0));
+        assert!(vec_bitmap.contains(1));
```
wjones127: It seems like, as written, we'll never update these for upsert, right? Originally, I was thinking if stable row ids were enabled, then updated rows and new rows would be written to separate fragments so that we could do this.

There is a case for merge_insert where we only do updates. This could be just because there aren't any new rows. Or it could be we set `WhenNotMatched::DoNothing`. In either of these cases, do we never update the bitmap?

rust/lance/src/dataset/write/merge_insert.rs
```diff
@@ -4520,4 +4535,256 @@ MergeInsert: on=[id], when_matched=UpdateAll, when_not_matched=InsertAll, when_n
...
+    }
+
+    #[tokio::test]
+    async fn test_sub_schema_upsert_fragment_bitmap() {
```
wjones127, line 4664: Eventually, I'd like see a lot more tests cases for this. Unfortunately, right now the ratio of lines of code to useful tests cases is pretty high. I'd suggest thinking about how you can parameterize the tests to cover more useful situations. For example, you might use the same original test data, and very the merge insert parameters, input data, and index states.

wjones127 (approved):
This seems adequate for now. But I think before we say this is production ready, I'd recommend broader test coverage.

rust/lance/src/dataset/write/merge_insert.rs
yanghua: > It seems like, as written, we'll never update these for upsert, right? 

Yes, here(in this test) is because in full-schema mode, we did not figure out which columns are updated. So I set all fields to the `fields_for_preserving_frag_bitmap`, and it would cause the update for the `frag bitmap` not to happen.

> Originally, I was thinking if stable row ids were enabled, then updated rows and new rows would be written to separate fragments so that we could do this.

IMO, it depends on whether we can distinguish which fields to update for each row in each fragment, right? It's hard to do this for `RewriteRow` mode. Correct me if I am wrong.

> In either of these cases, do we never update the bitmap?

The current test cases are pure updates as you mentioned.

Currently, for merge insert, all the cases(update, upsert) would not update the frag bitmap. This is due to `ReWriteRow` mode, it's hard to split which fields would be updated. The detailed logic is here: 

```
let index_covers_modified_field = index.fields.iter().any(|field_id| {
          value_updated_field_set.contains(&u32::try_from(*field_id).unwrap())
});
```

rust/lance/src/dataset/write/merge_insert.rs
yanghua: > Eventually, I'd like see a lot more tests cases for this.

Currently, the whole sub-schema merge insert would not be affected by this PR. It belongs to the `RewriteColumn` mode. It would update in-place. For pure update, the `row id`, `fragment` would not be updated, right?

yanghua:
> But I think before we say this is production ready, I'd recommend broader test coverage.

Absolutely, we can move faster. I would add more test cases and refactor them. Some logic may need to refactor, e.g. #4695 .

yanghua:
Will merge it, if there are any issues, will fix them in the further PRs.

PR #4597, opened 2025-08-29
feat: allow using opendal to access s3, azblob and gcs

jackye1995:
User can set `use_opendal=true` storage option to enable this route. Any storage option will be passed to OpenDAL as is to initialize its operator.

Related to: https://github.com/apache/arrow-rs-object-store/issues/472


wjones127:
Is there any background on this? What's the benefit of using the opendal options?

wjones127 (approved):

PR #4593, opened 2025-08-28
refactor!: deprecate tensorflow support

jackye1995:
## Summary
This PR removes TensorFlow support from Lance to simplify dependencies and reduce maintenance burden.

## Breaking Changes
⚠️ **This is a breaking change** - TensorFlow support has been completely removed from Lance.
Users relying on `lance.tf` modules will need to migrate to alternative solutions.

## Changes Made
- 🗑️ Removed `lance.tf` Python module and all TensorFlow integrations
- 📦 Removed TensorFlow dependencies from `pyproject.toml` and `Cargo.toml`
- 🦀 Removed TFRecord support from Rust codebase
- 📚 Removed TensorFlow documentation and examples
- 🖼️ Updated image encoding/decoding to use only Pillow
- 🔧 Removed TensorFlow from CI/CD pipelines

## Test Plan
- [ ] Verify all tests pass without TensorFlow dependencies
- [ ] Confirm the build succeeds on all platforms
- [ ] Check that image operations still work with Pillow-only implementation

🤖 Generated with [Claude Code](https://claude.ai/code)

PR #4598, opened 2025-08-29
chore: update DataFusion, Arrow, Object Store

timsaucer:
This PR is to update the following dependencies:

- DataFusion to 49.0.2
- Arrow to 55.2
- Object Store to 0.12.3

The rest of the changes are just due to updates upstream, but none are breaking.

### DIFF
--- a/rust/lance-index/src/scalar/btree.rs
+++ b/rust/lance-index/src/scalar/btree.rs
@@ -33,7 +33,7 @@ use datafusion::physical_plan::{
     union::UnionExec, ExecutionPlan, SendableRecordBatchStream,
 };
 use datafusion_common::{DataFusionError, ScalarValue};
-use datafusion_physical_expr::{expressions::Column, LexOrdering, PhysicalSortExpr};
+use datafusion_physical_expr::{expressions::Column, PhysicalSortExpr};
 use deepsize::DeepSizeOf;
 use futures::{
     future::BoxFuture,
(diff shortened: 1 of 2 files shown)


jackye1995:
I am trying to upgrade and received the same error in #4593. I will wait for this then 😆

timsaucer:
Digging into this for this and also #4593 the problem I'm seeing in `lance_encoding::encodings::logical::struct::tests::test_empty_struct` is that the update of arrow from [55.1](https://github.com/apache/arrow-rs/blob/55.1.0/arrow-select/src/concat.rs#L315) to [55.2](https://github.com/apache/arrow-rs/blob/55.2.0/arrow-select/src/concat.rs#L278) now has a check for struct arrays having at least one field during the call to `concat_structs`. In arrow 56, this is corrected because it calls [StructArray::try_new_with_length](https://github.com/apache/arrow-rs/blob/56.1.0/arrow-select/src/concat.rs#L278) instead of `StructArray::try_new`.

I don't think we want to bump arrow to 56 with datafusion at 49 because datafusion is at arrow 55 and that would have duplicate dependencies.

I added a TODO in the code I pushed to work around this error.

timsaucer:
@westonpace I'm not familiar with the CI failure I'm seeing here - it looks like something related to file system errors on windows. Is that something you've seen in this repo?

I'm also not wild about the fix I put in for the test I describe in the comments above, but I believe it can be removed in the next version, DF50 which should bring Arrow 56

rust/lance-index/src/scalar/btree.rs
```diff
@@ -1705,7 +1706,11 @@ mod tests {
             .into_df_exec(RowCount::from(10), BatchCount::from(100));
         let schema = data.schema();
         let sort_expr = PhysicalSortExpr::new_default(col("value", schema.as_ref()).unwrap());
-        let plan = Arc::new(SortExec::new(LexOrdering::new(vec![sort_expr]), data));
+        let plan = Arc::new(SortExec::new(
```
ddupg: Nit: here can be simplified to look like this:
```rust
let plan = Arc::new(SortExec::new([sort_expr].into(), data));
```

timsaucer: Thanks. I applied the pattern throughout that file and Scanner.

timsaucer:
It looks like the CI failure I was getting earlier was resolved after rebase. I think this is ready to go.

PR #4600, opened 2025-08-29
fix: resolve deadlock when multiple threads access a single LanceFileWriter

lorinlee:
PR #4383  attempted to release the GIL inside `LanceFileWriter`, but it collided with the writer’s own inner lock and can deadlock when the same writer is used from multiple threads. The exact sequence is:

1. Hold the GIL  
2. Acquire the inner lock  
3. Release the GIL  
4. Finish work, re-acquire the GIL  
5. Release the inner lock  

With multiple threads sharing one `LanceFileWriter`, it’s possible for:

- Thread 1 to hold the inner lock while waiting for the GIL (step 4)  
- Thread 2 to hold the GIL while waiting for the inner lock (step 2)  

resulting in a deadlock.

Even though the inner lock serializes access and prevents true concurrency, hanging is still unexpected and must be fixed.  
Therefore I replaced the inner lock with a Tokio `Mutex` and now acquire it only after the GIL has been released.

### DIFF
--- a/python/python/tests/test_file.py
+++ b/python/python/tests/test_file.py
@@ -2,9 +2,11 @@
 # SPDX-FileCopyrightText: Copyright The Lance Authors
 
 import os
+import threading
 
 import numpy as np
 import pyarrow as pa
+import pyarrow.compute as pc
 import pyarrow.parquet as pq
 import pytest
 from lance.file import LanceFileReader, LanceFileWriter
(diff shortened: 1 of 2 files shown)


lorinlee:
@wjones127 Hi, would you kindly take a moment to review this PR? thank you so much!

lorinlee:
@wjones127 Hi, Multithreaded access to the same file might cause a deadlock—it's a bug. Could you kindly take a look?

lorinlee:
Hi @Xuanwo, sorry to bother you—do you know who else might be able to review this PR? Thanks!

lorinlee:
@wjones127 Hi, could you please help review PR? Or do I need to provide any other information? Thank you!

PR #4601, opened 2025-08-29
fix: disable default features from jsonb to avoid changing serde-json behavior

Xuanwo:
Fix failed tests in https://github.com/lancedb/lancedb/pull/2610


wkalt (approved):
looks good pending CI, thanks

PR #4602, opened 2025-08-29
[WIP] feat: support branch based on shallow_clone

majin1102:
WIP

### DIFF
--- a/rust/lance/src/dataset/refs.rs
+++ b/rust/lance/src/dataset/refs.rs
@@ -11,37 +11,64 @@ use object_store::path::Path;
 use serde::{Deserialize, Serialize};
 use std::sync::Arc;
 
+use crate::dataset::refs::Ref::{Branch, Tag, Version};
 use crate::{Error, Result};
+use serde::de::DeserializeOwned;
+use snafu::location;
 use std::cmp::Ordering;
 use std::collections::HashMap;
+use std::fmt;
+use std::fmt::Formatter;
 
 /// Lance Ref
 #[derive(Debug, Clone)]
 pub enum Ref {
+    // This is a version number of Main branch
     Version(u64),
+    // 0 is the optional branch name, if null it is the main branch
+    // 1 is the tag name
     Tag(String),
+    // 0 is the branch name
+    // 1 is the version number
+    Branch(String, u64),
 }
 
 impl From<u64> for Ref {
     fn from(ref_: u64) -> Self {
-        Self::Version(ref_)
+        Version(ref_)
     }
 }
 
 impl From<&str> for Ref {
     fn from(ref_: &str) -> Self {
-        Self::Tag(ref_.to_string())
+        Tag(ref_.to_string())
+    }
+}
+
+impl From<(&str, u64)> for Ref {
+    fn from(_ref: (&str, u64)) -> Self {
+        Branch(_ref.0.to_string(), _ref.1)
+    }
+}
+
+impl fmt::Display for Ref {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        match self {
+            Version(version_number) => write!(f, "Main:{}", version_number),
+            Tag(tag_name) => write!(f, "{}", tag_name),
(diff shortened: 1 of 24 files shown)


rust/lance/src/index/vector/ivf/v2.rs
```diff
@@ -1276,7 +1276,7 @@ mod tests {
             test_index_multivec(params.clone(), nlist, recall_requirement).await;
         }
         test_distance_range(Some(params.clone()), nlist).await;
-        test_remap(params.clone(), nlist).await;
+        Box::pin(test_remap(params.clone(), nlist)).await;
```
majin1102, line 1279: This is unrelated changes due to clippy issue

rust/lance/src/dataset/refs.rs
```diff
@@ -11,51 +11,197 @@ use object_store::path::Path;
...
+}
+
+impl Branches {
+    fn object_store(&self) -> &ObjectStore {
+        &self.refs.object_store
+    }
+}
+
+#[async_trait::async_trait]
+pub trait RefOperations<T> {
```
majin1102: The introduction of this trait might break the interface compatibility of Tags, which probably requires some discussion.

majin1102: I feel not right about this trait. It's more like a constaint than an abstraction cause we did not use it anyway. Besides it brought uncompatibility

protos/transaction.proto
```diff
@@ -179,6 +179,8 @@ message Transaction {
     uint64 ref_version = 3;
     // the absolute base path of the source dataset for cloning
     string ref_path = 4;
+    // if the cloned dataset is a branch, this is the branch name of the cloned dataset
```
jackye1995: try to see if I understand this correctly, so this is the branch name of the new/target dataset, and ref_name is the branch name of the source dataset?

rust/lance/src/dataset/refs.rs
```diff
@@ -11,51 +11,197 @@ use object_store::path::Path;
...
 use std::collections::HashMap;
+use std::fmt;
+use std::fmt::Formatter;
+use std::io::ErrorKind;
 
 /// Lance Ref
 #[derive(Debug, Clone)]
 pub enum Ref {
+    // This is a version number of Main branch, the global version identifier is Main::<version_number>
     Version(u64),
```
jackye1995: I am thinking if we should just make them:

```rust
pub enum Ref {
  Version(Option<String>, u64),
  Tag(Option<String, String)
}
```

because a tag can also be created on a branch recursively

rust/lance/src/dataset/refs.rs
jackye1995: yeah I think we probably want just customized operations for branch and tag. In Iceberg, both branch and tag are refs and thus share some common operations. Here a branch is technically not a ref because it describes where the branch starts and then you list the branch dataset for its latest version. I feel it is better to just let branch and tag expose operations that make sense to them. This is similar to you have `git tag` and `git branch` which have different operations under them.

protos/transaction.proto
majin1102: yes！

majin1102: let me improve the comments

rust/lance/src/dataset/refs.rs
majin1102: > Tag(Option<String>, String)

For tag, I don't think this make sense. Tag in a global and static reference and identified only by it's name. If we use for example ("branch1", "tag1") to identify a tag, that means ("branch2", "tag1") is a different tag which apparently not right.

> because a tag can also be created on a branch recursively

I think we should distinguish the content and the identifier here. The enum Ref here is declaring how to identify reference, not declaring the contents of them. We use TagContents and BranchContents for content instead. So we need to add branch field to TagContents to adress the demand you mentioned. We can see it below.

> Version(Option<String>, u64),

Yeah, I think this is better. I was concerning the compatibilitiy. But it is fine after I tries this way

PR #4609, opened 2025-08-31
chore(java): refactor toString and Optional fields

majin1102:
TL;DR

1. ​​**Unify toString() style**​​: The toString() implementations were a little bit messy, so I unified them in this PR Currently, I used the original ToStringBuilder(create a until class to configure global style) but aligned it with Guava's style to facilitate future replacement if needed.

2. **​​Eliminate warnings for optional fields​​**: Avoid using Optional for fields entirely, as it is not a recommended practice in Java (unlike Rust). Instead, use Optional only in getter methods to prevent potential null-related warnings and maintain code clarity. **Note: didn't change any native interface using Optional**

3. **​​No interface changes​​**: This refactor ​​did not​​ modify any existing interfaces.

4. **Open to better solutions**​​: Feel free to suggest improvements. I’ll follow up after further discussion.

### DIFF
--- a/java/core/src/main/java/com/lancedb/lance/FragmentMetadata.java
+++ b/java/core/src/main/java/com/lancedb/lance/FragmentMetadata.java
@@ -16,8 +16,7 @@ package com.lancedb.lance;
 import com.lancedb.lance.fragment.DataFile;
 import com.lancedb.lance.fragment.DeletionFile;
 import com.lancedb.lance.fragment.RowIdMeta;
-
-import org.apache.commons.lang3.builder.ToStringBuilder;
+import com.lancedb.lance.util.ToStringHelper;
 
 import java.io.Serializable;
 import java.util.List;
@@ -98,12 +97,12 @@ public class FragmentMetadata implements Serializable {
 
   @Override
   public String toString() {
-    return new ToStringBuilder(this)
-        .append("id", id)
-        .append("physicalRows", physicalRows)
-        .append("files", files)
-        .append("deletionFile", deletionFile)
-        .append("rowIdMeta", rowIdMeta)
+    return ToStringHelper.of(this)
+        .add("id", id)
+        .add("physicalRows", physicalRows)
+        .add("files", files)
+        .add("deletionFile", deletionFile)
+        .add("rowIdMeta", rowIdMeta)
         .toString();
   }
 }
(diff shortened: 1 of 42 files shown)


majin1102:
PLAL when you have time @jackye1995

java/core/src/main/java/com/lancedb/lance/ReadOptions.java
```diff
@@ -22,8 +22,8 @@
 /** Read options for reading from a dataset. */
 public class ReadOptions {
 
-  private final Optional<Integer> version;
```
jackye1995, line 25: I am in general neutral about using optional or not to store instance variables. I think it is not recommended but projects like Trino use it extensively and it does make things more clear. If we want to switch back to not store optional, I think we should clearly annotate which variables are nullable in inputs and actively checks for `Objects.requireNonNull` for non-null fields.

java/core/src/main/java/com/lancedb/lance/FragmentMetadata.java
```diff
@@ -98,12 +97,12 @@ public boolean equals(Object o) {
 
   @Override
   public String toString() {
-    return new ToStringBuilder(this)
```
jackye1995, line 101: thanks for fixing this, it has been quite inconsistent. I am wondering if ToStringHelper is over-complicating things. Because must Java people use intellij and it allows auto-generate this kind of things, it feels like we should just stick with one (either apache commons or guava) and go with it, and add checkstyle rules to disallow other ways. What do you think?

majin1102: Yeah, let's use guava then? I'm a liitle uncertained so I write like this first

java/core/src/main/java/com/lancedb/lance/ReadOptions.java
majin1102: I think the getter method clearify the nullable.

But I agree optional fields make things more clear. I guess I need to ajust my IDE warnning conditions then. I will allign the Optional style in operations since we agree this is long-term all right.

java/core/src/main/java/com/lancedb/lance/FragmentMetadata.java
jackye1995: Sure! Do you think we should to shade the Guava classes?

majin1102: Yes, let me deal with this

majin1102:
This PR has been splitted to #4662 and #4676

Issue #4036, opened 2025-06-19
Add more flexible autocleanup strategy


yanghua:
Currently, lance only supports auto cleanup based on time interval. It would be better to have a more flexible cleanup strategy, e.g. commit-num based.

wojiaodoubao:
It would be good to have the cleanup strategies below:
* Retain latest n versions.
* Cleanup orphan files.

majin1102:
> * ain latest n versions.
> * Cleanup orphan files.

Why cleanup orphan is a policy not an option?

I think most cases if I want to cleanup orphan files, we still need a policy logic to know snapshot retention policies

jackye1995:
I just merged the PR. Are we good with closing this?

Xuanwo:
Closing as completed by #4614. Cleanup supports `retain_versions` through the Rust and Python APIs, with policy coverage.

PR #4614, opened 2025-09-02
feat: add new cleanup policy to retain latest n versions

wojiaodoubao:
Related to https://github.com/lancedb/lance/issues/4036

### DIFF
--- a/rust/lance-io/src/object_store.rs
+++ b/rust/lance-io/src/object_store.rs
@@ -87,7 +87,7 @@ impl<O: OSObjectStore + ?Sized> ObjectStoreExt for O {
         let output = self.list(Some(dir_path.into())).map_err(|e| e.into());
         if let Some(unmodified_since_val) = unmodified_since {
             output
-                .try_filter(move |file| future::ready(file.last_modified < unmodified_since_val))
+                .try_filter(move |file| future::ready(file.last_modified <= unmodified_since_val))
                 .boxed()
         } else {
             output.boxed()
--- a/rust/lance/src/dataset/cleanup.rs
+++ b/rust/lance/src/dataset/cleanup.rs
@@ -480,11 +480,11 @@ impl Default for CleanupPolicy {
 }
 
 #[derive(Default)]
-pub struct CleanupBuilder {
+pub struct CleanupPolicyBuilder {
     policy: CleanupPolicy,
 }
 
-impl CleanupBuilder {
+impl CleanupPolicyBuilder {
     /// Cleanup all versions before the specified timestamp.
     pub fn before_timestamp(mut self, timestamp: DateTime<Utc>) -> Self {
         self.policy.before_timestamp = Some(timestamp);
@@ -575,7 +575,7 @@ pub async fn auto_cleanup_hook(
         return Ok(None);
     }
 
-    let mut builder = CleanupBuilder::default();
+    let mut builder = CleanupPolicyBuilder::default();
     if let Some(older_than) = manifest.config.get("lance.auto_cleanup.older_than") {
         let std_older_than = match parse_duration(older_than) {
             Ok(t) => t,
@@ -863,7 +863,9 @@ mod tests {
             let db = self.open().await?;
             cleanup_old_versions(
                 &db,
-                CleanupBuilder::default().before_timestamp(before).build(),
+                CleanupPolicyBuilder::default()
+                    .before_timestamp(before)
+                    .build(),
             )
             .await
         }
@@ -882,7 +884,7 @@ mod tests {
             let db = self.open().await?;
             cleanup_old_versions(
                 &db,
-                CleanupBuilder::default()
+                CleanupPolicyBuilder::default()
                     .before_timestamp(before)
                     .delete_unverified(delete_unverified.unwrap_or(false))
                     .error_if_tagged_old_versions(error_if_tagged_old_versions.unwrap_or(true))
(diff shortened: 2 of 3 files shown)


rust/lance/src/dataset/cleanup.rs
```diff
@@ -117,13 +120,13 @@ const UNVERIFIED_THRESHOLD_DAYS: i64 = 7;
 impl<'a> CleanupTask<'a> {
     fn new(
         dataset: &'a Dataset,
-        before: DateTime<Utc>,
+        policy: CleanupPolicy,
         delete_unverified: bool,
```
jackye1995: why `delete_unverified` and `error_if_old_versions_tagged` are not a part of the policy?

rust/lance/src/dataset/cleanup.rs
```diff
@@ -444,6 +459,32 @@ impl<'a> CleanupTask<'a> {
     }
 }
 
+#[derive(Debug, Clone)]
+pub enum CleanupPolicy {
```
jackye1995: I feel it makes more sense for this to be a struct, instead of an enum, so that people can express composite policies like "cleanup all versions before version X and also timestamp Y"

rust/lance/src/dataset/cleanup.rs
wojiaodoubao: My thoughts on this is: the CleanupPolicy only determines which manifests need to be cleaned. The ```delete_unverified``` and ```error_if_old_versions_tagged``` determines the behavior of cleanup task.

```
#[derive(Clone, Debug)]
struct CleanupTask<'a> {
    dataset: &'a Dataset,
    /// The policy determines which versions should be cleanup
    policy: CleanupPolicy,
    /// If true, delete unverified data files even if they are recent
    delete_unverified: bool,
    /// If true, return an Error if a tagged version is old
    error_if_old_versions_tagged: bool,
}
```

rust/lance/src/dataset/cleanup.rs
wojiaodoubao: "composite policies" is a great idea. Let me try this.

rust/lance/src/dataset/cleanup.rs
```diff
@@ -481,41 +583,73 @@ pub async fn auto_cleanup_hook(
...
+        };
+
+        if manifest.version % interval != 0 {
+            return Ok(None);
         }
+    } else {
+        return Ok(None);
+    }
+
+    // TODO: Support complex composite policies.
```
wojiaodoubao: We need to support configuring more complex composite policies. May be we can add a new key `lance.auto_cleanup.cleanup_policy`, and the value could be something like this.
```
{
  "must": [{"retian_versions": 100}],
  "should":[],
  "must_not": [{"older_than": "2025-03-10T12:00:00Z"}],
}
```

I think may be I can add this at another pr, so I can keep this one clean.

wojiaodoubao:
Hi @jackye1995 @wjones127 , could you help review this when you have time, thanks very much!

jackye1995:
@wojiaodoubao I am thinking if doing this composite policy is a bit too much of a complication. I like the fact that it is trying to make the framework extensible, but when I said "composite policy", I was just thinking about the case that user gives a condition both for time and number of versions, because in your original version, the code allows either specifying time or version but not both. 

When I look into other expiration APIs like Iceberg [ExpireSnapshots](https://github.com/apache/iceberg/blob/main/api/src/main/java/org/apache/iceberg/actions/ExpireSnapshots.java), allowing users to configure these 2 seems sufficient, and we can just allow the API (e.g. having a CleanupBuilder) to take both and then handle the case that user specify both conditions. What do you think?

wojiaodoubao:
> @wojiaodoubao I am thinking if doing this composite policy is a bit too much of a complication. I like the fact that it is trying to make the framework extensible, but when I said "composite policy", I was just thinking about the case that user gives a condition both for time and number of versions, because in your original version, the code allows either specifying time or version but not both.
> 
> When I look into other expiration APIs like Iceberg [ExpireSnapshots](https://github.com/apache/iceberg/blob/main/api/src/main/java/org/apache/iceberg/actions/ExpireSnapshots.java), allowing users to configure these 2 seems sufficient, and we can just allow the API (e.g. having a CleanupBuilder) to take both and then handle the case that user specify both conditions. What do you think?

Thanks @jackye1995  your detailed explanation, it makes sense to me. I think I also understand why you said "delete_unverified and error_if_old_versions_tagged part of policy", we can use a CleanupPolicy struct to wrap all. Let me fix it.

wojiaodoubao:
Hi @jackye1995 , I have updated the pr. Please review when you have time, thanks very much!

rust/lance/src/dataset/cleanup.rs
```diff
@@ -444,6 +443,90 @@ impl<'a> CleanupTask<'a> {
...
+        }
+    }
+}
+
+#[derive(Default)]
+pub struct CleanupBuilder {
+    policy: CleanupPolicy,
+}
+
+impl CleanupBuilder {
```
jackye1995: this should be CleanupPolicyBuilder? Since it's building a policy. 

When I say having a `CleanupBuilder`, I was more thinking about

```
CleanupBuilder::new(dataset)
  . before_timestamp(...)
  .execute()
```

kind of experience similar to `MergeInsertBuilder`, but the current one seems also fine.

rust/lance-io/src/object_store.rs
```diff
@@ -87,7 +87,7 @@ impl<O: OSObjectStore + ?Sized> ObjectStoreExt for O {
         let output = self.list(Some(dir_path.into())).map_err(|e| e.into());
         if let Some(unmodified_since_val) = unmodified_since {
             output
-                .try_filter(move |file| future::ready(file.last_modified < unmodified_since_val))
+                .try_filter(move |file| future::ready(file.last_modified <= unmodified_since_val))
```
jackye1995, line 90: ah nice catch! I think you are right

PR #4615, opened 2025-09-02
feat(rust): support diff data between versions

yanghua:


yanghua:
@jackye1995 Would like to align with you about the logic to compare two versions:
- The prerequisite is that the dataset must enable the stable row ID feature.
- the diff would be based on the metadata, here is RowLatestUpdateVersionSequence；
- Let’s suppose there are two versions that need to be diffed, named V(O, old) ~ V(N, new). In Lance V(N), all the latest update version seqs are <=  N. And in lanceV(O), all the latest update version seqs are <= O. So we need:
  - checkout lance V(N), fetch updated and inserted records via iterate latest update version seqs then filter which update version larger than O; (we may need to introduce some metadata(e.g. min_latest_update_version, max_latest_update_version) for fragment to do some skip)
  - Fetch deleted records via comparing the lance V(N) and lance V(O) and filter the records that exist in V(O) but not in V(N).
  
Does it match your thoughts?

jackye1995:
> Let’s suppose there are two versions that need to be diffed, named V(O, old) ~ V(N, new). In Lance V(N), all the latest update version seqs are <= N. And in lanceV(O), all the latest update version seqs are <= O. So we need:
checkout lance V(N), fetch updated and inserted records via iterate latest update version seqs then filter which update version larger than O; (we may need to introduce some metadata(e.g. min_latest_update_version, max_latest_update_version) for fragment to do some skip)


I think for this part we are thinking about the same thing: `SELECT * FROM table WHERE latest_update_version > O AND latest_update_version <= N`

> Fetch deleted records via comparing the lance V(N) and lance V(O) and filter the records that exist in V(O) but not in V(N).

For this part, I feel we could do better. Because every time some rows are deleted, we can efficiently store those deleted row IDs in a compact structure of RowIdSequence, it might be easier to just compute this information when performing the deletion and record it somewhere (e.g. add a new `cdc` system index) and we can quickly retrieve that information if we want to do a diff.

What do you think?

yanghua:
> For this part, I feel we could do better. Because every time some rows are deleted, we can efficiently store those deleted row IDs in a compact structure of RowIdSequence, it might be easier to just compute this information when performing the deletion and record it somewhere (e.g. add a new `cdc` system index) and we can quickly retrieve that information if we want to do a diff.

OK, let me think about it, if we can compute it from the deletion vector from the two versions.

jackye1995:
Technically data deletion only happens in 1.5 places:
1. Delete
2. MergeInsert when setting WhenNotMatchedBySource::Delete or WhenNotMatchedBySource::DeleteIf

Both cases we should get a stream of records to remove and build up the deletion information.

yanghua:
> I think for this part we are thinking about the same thing: SELECT * FROM table WHERE latest_update_version > O AND latest_update_version <= N

@jackye1995 Want to make sure, here the SQL is only for describing the logic. The final function would be a `diff` API. Right?

jackye1995:
> Want to make sure, here the SQL is only for describing the logic. The final function would be a diff API. Right?

Yes!

yanghua:
Done via #4741

PR #4616, opened 2025-09-02
feat!: add target_partition_size vector index param

BubbleCal:
this makes `num_partitions` optional, and encourage users to use `target_partition_size`.
the `target_partition_size` is optional too, each index type has a default `target_partition_size`.

if both `num_partitions` and `target_partition_size` are set, indexing would take `num_partitions`

### DIFF
--- a/python/python/tests/test_lance.py
+++ b/python/python/tests/test_lance.py
@@ -129,8 +129,6 @@ def test_create_index(tmp_path):
         dataset.create_index("emb", "IVF_PQ")
     with pytest.raises(ValueError):
         dataset.create_index("emb", "IVF_PQ", num_partitions=5)
-    with pytest.raises(ValueError):
-        dataset.create_index("emb", "IVF_PQ", num_sub_vectors=4)
     with pytest.raises(KeyError):
         dataset.create_index("foo", "IVF_PQ", num_partitions=5, num_sub_vectors=16)
     with pytest.raises(NotImplementedError):


rust/lance/src/index/vector/ivf.rs
```diff
@@ -1207,23 +1207,24 @@ pub async fn build_ivf_model(
     metric_type: MetricType,
     params: &IvfBuildParams,
 ) -> Result<IvfModel> {
+    let num_partitions = params.num_partitions.unwrap();
```
Xuanwo, line 1210: Is there a guarantee that `num_partitions` must be valid?

rust/lance-index/src/vector/ivf/builder.rs
```diff
@@ -18,8 +18,14 @@ use lance_io::stream::RecordBatchStream;
 /// Parameters to build IVF partitions
 #[derive(Debug, Clone)]
 pub struct IvfBuildParams {
+    /// Deprecated: use `target_partition_size` instead.
     /// Number of partitions to build.
-    pub num_partitions: usize,
+    pub num_partitions: Option<usize>,
```
Xuanwo, line 23: Can we use rust's `#[deprecate]` here?

rust/lance-index/src/lib.rs
```diff
@@ -192,6 +220,54 @@ impl IndexType {
...
+
+            // for now all vector indices are built by the same builder,
+            // so they share the same version.
+            Self::Vector
+            | Self::IvfFlat
+            | Self::IvfSq
+            | Self::IvfPq
+            | Self::IvfHnswSq
+            | Self::IvfHnswPq
+            | Self::IvfHnswFlat => 1,
```
Xuanwo, line 249: I have seen usage like `IndexFileVersion::V3`, what's the relation between them?

BubbleCal: that `IndexFileVersion::V3` indicates the index file is stored as lance format.
this version indicates the index format version, e.g. we changed the location we store PQ codebook, this introduces breaking change so we bumped the index version.

rust/lance-index/src/vector/ivf/builder.rs
BubbleCal: we still use it, just make it optional

rust/lance/src/index/vector/ivf.rs
BubbleCal: https://github.com/lancedb/lance/pull/4616/files#diff-8df08788f6e7e51e8f34db59a22e029e06b172865bbda785e10a58e18fdcafe8R302-R313
would set `num_partitions`, but yes agree it's hard to guarantee. going to make `unwrap_or(32)` because 32 is the default value before this PR

BubbleCal:
mark this breaking because it changes the default value, and the rust API breaking

PR #4617, opened 2025-09-02
fix: shallow_clone multiple times refering to wrong base path

majin1102:
The original testcase is:

```
async fn test_shallow_clone_multiple_times(
...
        // Generate clone paths
        let clone_paths = (1..=1)
            .map(|i| test_dir.path().join(format!("clone{}", i)))
            .collect::<Vec<_>>();
        let mut cloned_datasets = Vec::with_capacity(1);
...
```

It did't really test clone multiple times. And there is actually a bug here. I probably wanted to change this loop before committing but forgot it.  When I was debugging branch this bug took me quite a while. I will pay more attentions to cases like this.

Sorry for the miss @jackye1995


Xuanwo (approved):
Thank you for this fix!

PR #4621, opened 2025-09-03
feat: add a scalar index for JSON

westonpace:
Adds a scalar index for JSON columns by creating a scalar index on some sub-field in the JSON data.  Any scalar index type can be used as the underlying type.

This PR also introduces the ability to configure scalar indexes in a generic way.  It might be nice to eventually switch the Inverted index so that it can also be trained in this way (and then we don't need a special "create full text index" method) but that can be left for future work.

This PR also modifies the expression parsing slightly to introduce the idea that a query might match a direct column reference _or_ might match a projected column reference.  This could be useful in the future for things like indexes on struct fields, list items, image metadata, etc.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4559,6 +4559,7 @@ dependencies = [
  "half",
  "itertools 0.13.0",
  "jieba-rs",
+ "jsonb",
  "lance-arrow",
  "lance-core",
  "lance-datafusion",
--- a/protos/index.proto
+++ b/protos/index.proto
@@ -5,6 +5,8 @@ syntax = "proto3";
 
 package lance.index.pb;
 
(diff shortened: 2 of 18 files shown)


westonpace:
There's still a number of TODOs left in this PR.  Perhaps the most significant one is how we want to handle type specification and/or extraction.  Right now the index uses "extract" with a JSON path.  The resulting target index will then always be an index against string data.

We could also support type inference where the user provides a sub-type (e.g. number) and we create the underlying index against numeric data to slightly improve index search speeds.  I'm not sure if that will really matter too much to start so I think we can defer that particular TODO.

Xuanwo (approved):

Issue #4625, opened 2025-09-03
JSON's index extract should be type-aware instead of being treated as a string


Xuanwo:
We currently treat all JSON values as strings, which can lead to incorrect comparisons such as `> 10`. We need to build the JSON index with the correct data types, such as `float` or `int`.

PR #4626, opened 2025-09-03
feat: build type-aware index for JSON

Xuanwo:
This PR will fix https://github.com/lancedb/lance/issues/4625

To make the best use of the JSONB format, I added a new UDF that extracts JSONB as its raw value along with its type tag. This way, there's no need to first convert the JSONB value to a JSON string and then back to the desired type.

---

### DIFF
--- a/rust/lance-datafusion/src/udf/json.rs
+++ b/rust/lance-datafusion/src/udf/json.rs
@@ -12,6 +12,40 @@ use datafusion::physical_plan::ColumnarValue;
 use datafusion::prelude::create_udf;
 use std::sync::Arc;
 
+/// Represents the type of a JSONB value
+#[repr(u8)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum JsonbType {
+    Null = 0,
+    Boolean = 1,
+    Int64 = 2,
+    Float64 = 3,
+    String = 4,
+    Array = 5,
+    Object = 6,
+}
+
+impl JsonbType {
+    /// Convert from u8 value
+    pub fn from_u8(value: u8) -> Option<Self> {
+        match value {
+            0 => Some(Self::Null),
+            1 => Some(Self::Boolean),
+            2 => Some(Self::Int64),
+            3 => Some(Self::Float64),
+            4 => Some(Self::String),
+            5 => Some(Self::Array),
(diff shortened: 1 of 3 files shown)


rust/lance-index/src/scalar/json.rs
```diff
@@ -363,38 +367,322 @@ impl JsonIndexPlugin {
...
+                .as_any()
+                .downcast_ref::<LargeBinaryArray>()
+                .ok_or_else(|| Error::InvalidInput {
+                    source: "value is not LargeBinary".into(),
+                    location: location!(),
+                })?;
+
+            // Convert based on target type using serde deserialization
+            let converted_array: Arc<dyn Array> = match target_type {
+                DataType::Boolean => {
```
Xuanwo, line 548: I'm not sure if we have better ways to handle this.

jackye1995: can we just compare the raw_jsonb against the 2 possible values directly?

rust/lance-datafusion/src/udf/json.rs
```diff
@@ -12,6 +12,40 @@ use datafusion::physical_plan::ColumnarValue;
 use datafusion::prelude::create_udf;
 use std::sync::Arc;
 
+/// Represents the type of a JSONB value
+#[repr(u8)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum JsonbType {
+    Null = 0,
+    Boolean = 1,
+    Int64 = 2,
```
jackye1995, line 21: my understanding is that user will specify how to extract value, which means it can technically be more precise like int32 and float32, that could be beneficial for saving index size. is that right, or do we completely derive types based on jsonb encoding information? But anyway this is a good set of types.

rust/lance-index/src/scalar/json.rs
Xuanwo: Oh, I mean different types like `DataType::Utf8` and more.

rust/lance-datafusion/src/udf/json.rs
Xuanwo: That's a good question. Currently, we store the index in the physical type specified by JSONB. It might be worth exploring in the future.

PR #4633, opened 2025-09-03
chore: add test for index shallow_clone mulitple times

majin1102:
Based on the bugs discovered #4617 , supplemented the test case for the index scenario


majin1102:

jackye1995 (approved):
looks good to me!

PR #4634, opened 2025-09-04
feat: implement anti-join optimization with early stopping

LuQQiu:
Add EarlyStopAntiJoinExec and AntiJoinOptimizer to optimize anti-join queries with LIMIT clauses. This optimization converts patterns like `LIMIT -> HashJoinExec(Anti)` to use early-stopping execution, significantly improving performance for queries such as:

- SELECT * FROM table WHERE id NOT IN (subquery) LIMIT n
- SELECT * FROM table WHERE NOT EXISTS (subquery) LIMIT n

The optimizer creates a hash table for the exclusion list, then matches one by one in the scan table with early stop when limit is reached.

Originally:
SELECT * FROM large_600m_table lt WHERE NOT EXISTS (
    SELECT 1 FROM small_1m_table st WHERE st.id = lt.id
) LIMIT 50000;

Will scan 600 million rows of large_600m_table and 1 million rows of small_1m_table.

With optimization:
Will only scan at most 1 million + 50K limit rows.

### DIFF
--- a/rust/lance/src/io/exec/early_stop_anti_join.rs
+++ b/rust/lance/src/io/exec/early_stop_anti_join.rs
@@ -367,6 +367,113 @@ impl Stream for EarlyStopAntiJoinStream {
 pub struct AntiJoinOptimizer {}
 
 impl AntiJoinOptimizer {
+    /// Push down scan limit to LanceScanExec if possible
+    /// The probe side only needs to scan at most: limit + build_rows
+    /// (worst case: first build_rows all match exclusion list)
+    fn optimize_lance_scan(
+        &self,
+        scan_child: Arc<dyn ExecutionPlan>,
+        build_rows: usize,
+        limit: usize,
+    ) -> Arc<dyn ExecutionPlan> {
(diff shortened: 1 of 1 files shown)


rust/lance/src/io/exec/early_stop_anti_join.rs
```diff
@@ -0,0 +1,648 @@
...
+
+                Ok(())
+            }
+        }
+    }
+}
+
+impl std::fmt::Debug for EarlyStopAntiJoinExec {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("EarlyStopAntiJoinExec")
```
wkalt, line 102: I think we can probably avoid "EarlyStop" in the user-facing display. Users will probably expect all our operators to stop early in this way.

wkalt: even in the names, could just call this HashAntiJoinExec

LuQQiu:
change to https://github.com/lancedb/lance/pull/4653

PR #4639, opened 2025-09-04
feat: support to create FTS index on multiple columns

BubbleCal:


python/python/lance/dataset.py
```diff
@@ -1958,9 +1958,10 @@ def create_scalar_index(
 
         Parameters
         ----------
-        column : str
-            The column to be indexed.  Must be a boolean, integer, float,
-            or string column.
+        column : str, list[str]
```
LuQQiu, line 1961: any benefit to do it in this way other than user call create_scalar_index for multiple times?
As a user, would prefer this API to be the same for scalar indexes, basically support BTREE, BITMAP..... other indices with multi column support as well

wkalt: I agree - when we have multicolumn support for those other types, it'll be a single index spanning multiple columns. Users would probably expect this to behave the same way.

BubbleCal:
after discussed, we think this API may confuse users because we may use the same API to create compound index in the future, so close it.

PR #4642, opened 2025-09-04
docs: non-optional enum fields in proto3 do not have explicit presence

jbapple:
It is not possible to read DeletionFile.file_type and get back a value of "unspecified". For better or worse, an absent value will be read as ARROW_ARRAY: https://protobuf.dev/programming-guides/field_presence

### DIFF
--- a/protos/table.proto
+++ b/protos/table.proto
@@ -357,21 +357,20 @@ message DataFile {
 //
 // The path of the deletion file is constructed as:
 //   {root}/_deletions/{fragment_id}-{read_version}-{id}.{extension}
-// where {extension} is `.arrow` or `.bin` depending on the type of deletion.
+// where {extension} depends on DeletionFileType.
 message DeletionFile {
-  // Type of deletion file, which varies depending on what is the most efficient
-  // way to store the deleted row offsets. If none, then will be unspecified. If there are
-  // sparsely deleted rows, then ARROW_ARRAY is the most efficient. If there are
-  // densely deleted rows, then BIT_MAP is the most efficient.
+  // Type of deletion file, intended as a way to increase efficiency of the storage of deleted row
+  // offsets. If there are sparsely deleted rows, then ARROW_ARRAY is the most efficient. If there
+  // are densely deleted rows, then BITMAP is the most efficient.
   enum DeletionFileType {
-    // Deletion file is a single Int32Array of deleted row offsets. This is stored as
-    // an Arrow IPC file with one batch and one column. Has a .arrow extension.
+    // A single Int32Array of deleted row offsets, stored as an Arrow IPC file with one batch and
+    // one column. Has a .arrow extension.
     ARROW_ARRAY = 0;
-    // Deletion file is a Roaring Bitmap of deleted row offsets. Has a .bin extension.
+    // A Roaring Bitmap of deleted row offsets. Has a .bin extension.
     BITMAP = 1;
   }
 
-  // Type of deletion file. If it is unspecified, then the remaining fields will be missing.
+  // Type of deletion file.
   DeletionFileType file_type = 1;
   // The version of the dataset this deletion file was built from.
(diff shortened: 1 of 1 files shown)


protos/table.proto
```diff
@@ -357,21 +357,21 @@ message DataFile {
...
+    // one column. Has a .arrow extension.
     ARROW_ARRAY = 0;
-    // Deletion file is a Roaring Bitmap of deleted row offsets. Has a .bin extension.
+    // A Roaring Bitmap of deleted row offsets. Has a .bin extension.
     BITMAP = 1;
   }
 
-  // Type of deletion file. If it is unspecified, then the remaining fields will be missing.
+  // Type of deletion file. Because in Proto3 enums without an optional cardinality marker does not
+  // have "explicit presence", an absent value is actually a 0, which is ARROW_ARRAY.
```
jackye1995: I think we use the proto3 a bit differently, as you see we still put explicit optional annotation. Fields without optional is basically required. This translates to a field having Option or not in the rust generated code.

jbapple: Sorry, I'm confused. `file_type` does not have an explicit optional annotation. Are you saying that, despite that, Rust can somehow tell the difference between a `file_type` set to `ARROW_ARRAY` (`= 0`) and a `file_type` that has not been set at all? I think I must be misunderstanding you, because https://protobuf.dev/reference/rust/rust-generated/#implicit-presence

jackye1995: See https://github.com/lancedb/lance/blob/main/rust/lance-table/src/format/fragment.rs#L189-L226

You can also see the codegen:

```
#[derive(Clone, Copy, PartialEq, ::prost::Message)]
pub struct DeletionFile {
    /// Type of deletion file. If it is unspecified, then the remaining fields will be missing.
    #[prost(enumeration = "deletion_file::DeletionFileType", tag = "1")]
    pub file_type: i32,
    /// The version of the dataset this deletion file was built from.
    #[prost(uint64, tag = "2")]
    pub read_version: u64,
    /// An opaque id used to differentiate this file from others written by concurrent
    /// writers.
    #[prost(uint64, tag = "3")]
    pub id: u64,
    /// The number of rows that are marked as deleted.
    #[prost(uint64, tag = "4")]
    pub num_deleted_rows: u64,
    /// The base path index of the data file. Used when the file is imported or referred from another dataset.
    /// Lance use it as key of the base_paths field in Manifest to determine the actual base path of the data file.
    #[prost(uint32, optional, tag = "7")]
    pub base_id: ::core::option::Option<u32>,
}
```

where the `base_id` has an optional annotation and becomes `base_id: ::core::option::Option<u32>`. `file_type` is not.

jbapple: Yes, that is what I was trying to say. Because `file_type` does not have an optional annotation, it cannot be "none" the way line 363 hypothesizes. This is exactly why I am suggesting this change.

jackye1995: yes I understand that change is correct, but at this specific line, I think there will not really be "an absent value"?

jbapple: Ah, _now_ I understand! Thank you for your patience!!

I removed that reference to "an absent value".

protos/table.proto
```diff
@@ -380,8 +379,9 @@ message DeletionFile {
   uint64 id = 3;
   // The number of rows that are marked as deleted.
   uint64 num_deleted_rows = 4;
-  // The base path index of the data file. Used when the file is imported or referred from another dataset.
-  // Lance use it as key of the base_paths field in Manifest to determine the actual base path of the data file.
+  // The base path index of the data file. Used when the file is imported or referred from another
+  // dataset. Lance uses it as key of the base_paths field in Manifest to determine the actual base
+  // path of the data file.
```
jackye1995: nit: "of the deletion file" (it's a miss from the original author but would appreciate if you can fix along the way)

PR #4649, opened 2025-09-05
feat(java): supports fragment-level mergeColumn interface

steFaiz:
Support fragment level merge column api. See issue: 
[4650](https://github.com/lancedb/lance/issues/4650)

### DIFF
--- a/java/core/lance-jni/src/fragment.rs
+++ b/java/core/lance-jni/src/fragment.rs
@@ -311,7 +311,7 @@ fn inner_merge_column<'local>(
             unsafe { env.get_rust_field::<_, _, BlockingDataset>(jdataset, NATIVE_DATASET) }?;
         (
             dataset.inner.get_fragment(fragment_id as usize),
-            dataset.get_max_field_id(),
+            dataset.inner.manifest().max_field_id(),
         )
     };
     let mut fragment = match fragment_opt {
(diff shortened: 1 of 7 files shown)


steFaiz:
@jackye1995 PTAL when you have a chance. Thanks!

java/core/src/main/java/com/lancedb/lance/Dataset.java
```diff
@@ -675,6 +675,22 @@ public long calculateDataSize() {
...
   private native DataStatistics nativeGetDataStatistics();
 
+  /**
+   * Get the max field id of all fields. Note that it will traverse all DataFiles as well as current
+   * Manifest Schema. If some columns are dropped from this schema, we can restore the used field
+   * ids from DataFiles.
+   *
+   * @return maxFieldId
+   */
+  public int getMaxFieldId() {
```
jackye1995: it seems like we don't really need to expose this getMaxFieldId through a dedicated JNI method, because we can just compute that as a part of the nativeMergeColumns in JNI? It has access to the dataset, and we can just get the max field ID at that time.

steFaiz: Thanks for your reply! I've updated the code.

java/core/lance-jni/src/fragment.rs
```diff
@@ -266,6 +273,70 @@ fn inner_delete_rows<'local>(
...
+    arrow_array_stream_addr: jlong,
+    left_on: JString,
+    right_on: JString,
+) -> Result<JObject<'local>> {
+    let (fragment_opt, max_field_id) = {
+        let dataset =
+            unsafe { env.get_rust_field::<_, _, BlockingDataset>(jdataset, NATIVE_DATASET) }?;
+        (
+            dataset.inner.get_fragment(fragment_id as usize),
+            dataset.get_max_field_id(),
```
jackye1995: can just do `.inner.manifest().max_field_id()` do remove the get_max_field_id function

steFaiz: Thanks for your review again! I have removed the redundant method.

jackye1995 (approved):
looks good to me!

PR #4651, opened 2025-09-05
feat: expose lance-tools as python command

wojiaodoubao:
```
(base) lijinglun@n37-033-161:~/Project/lance-community/python$ pip install pylance-0.35.0-cp39-abi3-manylinux_2_31_x86_64.whl
(base) lijinglun@n37-033-161:~/Project/lance-community/python$ lance-tools
Tools for interacting with Lance files and tables

Usage: lance-tools <COMMAND>

Commands:
  file  Commands for interacting with Lance files
  help  Print this message or the help of the given subcommand(s)

Options:
  -h, --help     Print help
  -V, --version  Print version
(base) lijinglun@n37-033-161:~/Project/lance-community/python$ lance-tools file meta --source ~/tmp/alice1.lance/data/3980c788-3974-4ef0-80f4-dfbb34e80edc.lance
version: 2.0
num_rows: 6
num_data_bytes: 320
num_column_metadata_bytes: 402
num_footer_bytes: 584
schema:
Field(id=0, name=name, type=string)
Field(id=1, name=city, type=string)
Field(id=2, name=age, type=int64)
```


python/pyproject.toml
```diff
@@ -38,6 +38,9 @@ classifiers = [
     "Topic :: Scientific/Engineering",
 ]
 
+[project.scripts]
+lance-tools="lance.cli:lance_tools_cli"
```
Xuanwo, line 42: I'm not sure I like the idea of having lance_tools bundled directly in `pylance`. Maybe it should just be a separate binary instead?

wojiaodoubao: Hi @Xuanwo , thanks your comments! I think there are 2 main reasons why we need bundle lance-tools in pylance:

1. Ease of Use: The simplest and most common way to use Lance is by running ```pip install pylance```. Now, after installing pylance, we can directly use the lance-tools command without needing any additional installations.
2. Version Consistency: If lance-tools and pylance are installed separately, users need to ensure that the version of lance-tools is compatible with pylance.

PR #4653, opened 2025-09-05
feat: limit push down for anti hash join

LuQQiu:


test_physical_plan.rs
```diff
@@ -0,0 +1,72 @@
...
+    
+    // Check children
+    let children = physical_plan.children();
+    if !children.is_empty() {
+        println!("First child type: {:?}", children[0].name());
+        println!("First child has fetch(): {:?}", children[0].fetch());
+    }
+    
+    Ok(())
+}
```
wkalt, line 72: this test isn't making assertions. Maybe this was for debugging and included by accident?

rust/lance/src/io/exec/anti_join_limit_pushdown.rs
```diff
@@ -0,0 +1,249 @@
...
+    /// Push down limit to scan by using with_fetch or adding LocalLimitExec
+    fn push_limit_to_scan(
+        &self,
+        probe_side: Arc<dyn ExecutionPlan>,
+        limit: usize,
+        build_rows: usize,
+    ) -> Arc<dyn ExecutionPlan> {
+        // Calculate maximum rows we need to scan
+        // Worst case: first build_rows all match (excluded), then next limit rows don't match
+        let max_probe_rows = limit + build_rows;
```
wkalt, line 77: @LuQQiu I don't quite follow this. The comment seems incorrect to me - shouldn't the worst case be scanning all rows in each table? That would happen if there weren't enough matches to fill the limit. I don't see how the maximum probe rows would be dependent on the user's limit.

LuQQiu: You raised a good point. Is it based on the assumption that the values are unique. if it's not unique, the case would be very different.

jbapple: Would that hopefully mean that we can just use DataFusion's operator?

PR #4658, opened 2025-09-08
feat: create indexes from another source dataset

jackye1995:
## Summary

This PR adds a new `initialize_indices` method to the `DatasetIndexInternalExt` trait that enables copying index configurations from a source dataset to a target dataset.

## Motivation

When working with Lance datasets, there's often a need to create consistent index configurations across related datasets. This is particularly useful for:
- Creating development/test datasets with production-like indices
- Setting up delta indices for incremental data processing
- Maintaining consistent indexing across data pipeline stages

### New API Method
Added `initialize_indices` to `DatasetIndexInternalExt` trait:
```rust
async fn initialize_indices(&mut self, source_dataset: &Dataset) -> Result<()>
```

### Features
- **Index Type Support**:
  - Vector indices (IVF_FLAT, IVF_PQ, IVF_SQ)
  - Scalar indices (BTree, Bitmap, LabelList)
  - Full-text search indices (Inverted)
  
- **Validation**:
  - Checks field existence in target dataset
  - Validates data type compatibility
  - Filters out system indices (FragmentReuse, MemWal)
  
- **Parameter Extraction**:
  - Extracts index parameters from source statistics
  - Preserves index names from source dataset
  - Automatically adjusts fragment bitmaps for target dataset

### Tests
Added comprehensive tests:
- `test_initialize_indices`: Tests copying various index types
- `test_initialize_indices_with_missing_field`: Tests error handling

## Future Work
- For vector indices: Extract and reuse centroids from source dataset to create true delta indices (currently trains new centroids)

## Example Usage
```rust
let mut target_dataset = Dataset::open("path/to/target").await?;
let source_dataset = Dataset::open("path/to/source").await?;

// Initialize all indices from source dataset
target_dataset.initialize_indices(&source_dataset).await?;
```


rust/lance/src/index.rs
```diff
@@ -1461,6 +1468,136 @@ impl DatasetIndexInternalExt for Dataset {
...
+                    .ok_or_else(|| Error::Index {
+                        message: format!(
+                            "Field '{}' required by index '{}' not found in target dataset",
+                            source_field.name, index_name
+                        ),
+                        location: location!(),
+                    })?;
+
+            // Verify field types match
+            if source_field.data_type() != target_field.data_type() {
```
wkalt, line 1511: will this work for struct/nested types?

rust/lance/src/index.rs
```diff
@@ -1461,6 +1468,136 @@ impl DatasetIndexInternalExt for Dataset {
...
+        }
+
+        if field_names.is_empty() {
+            return Err(Error::Index {
+                message: format!("Index '{}' has no fields", index_name),
+                location: location!(),
+            });
+        }
+
+        let column_name = field_names[0].as_str();
```
wkalt: this is based on the assumption of no multicolumn indexes I think - that works for now but we'll need to be careful not to break it when those are introduced.

rust/lance-index/src/scalar/flat.rs
```diff
@@ -334,6 +334,11 @@ impl ScalarIndex for FlatIndex {
     fn update_criteria(&self) -> UpdateCriteria {
         unimplemented!()
     }
+
+    fn derive_index_params(&self) -> Result<super::ScalarIndexParams> {
+        // FlatIndex is used internally and doesn't have user-configurable parameters
+        unimplemented!("FlatIndex is an internal index type and cannot be recreated")
```
wkalt, line 340: should this return an error instead of panicking? Could an end user cause this to happen?

rust/lance/src/index.rs
wkalt: probably we should either pass a vec all the way through if we can (I know some of our methods work on vecs already for this reason), or error here if there's more than one column

rust/lance/src/index/vector.rs
```diff
@@ -758,3 +963,1021 @@ pub(crate) async fn open_vector_index_v2(
...
+            .get("sub_index")
+            .and_then(|v| v.as_object())
+            .expect("IVF_PQ index should have sub_index");
+        assert_eq!(
+            sub_index.get("index_type").and_then(|v| v.as_str()),
+            Some("PQ"),
+            "Sub-index type should be PQ"
+        );
+        assert_eq!(
+            sub_index.get("nbits").and_then(|v| v.as_u64()),
```
wkalt: should this be num_bits based on the code above? I think this is just getting 8 because the key nbits doesn't exist

rust/lance/src/index/vector.rs
```diff
@@ -758,3 +963,1021 @@ pub(crate) async fn open_vector_index_v2(
...
+                    .get("num_sub_vectors")
+                    .and_then(|v| v.as_u64())
+                    .unwrap_or(16) as usize;
+
+                VectorIndexParams::ivf_pq(
+                    num_partitions,
+                    num_bits,
+                    num_sub_vectors,
+                    metric_type,
+                    50,
```
wkalt: why 50?

rust/lance/src/index/vector.rs
```diff
@@ -758,3 +963,1021 @@ pub(crate) async fn open_vector_index_v2(
...
+            location: location!(),
+        })? as usize;
+
+        // Extract metric type
+        let metric_type_str = stats
+            .get("metric_type")
+            .and_then(|v| v.as_str())
+            .unwrap_or("L2");
+        let metric_type = match metric_type_str {
+            "L2" => lance_linalg::distance::MetricType::L2,
```
wkalt: should this be case-insensitive like the others?

rust/lance-index/src/scalar/inverted/index.rs
```diff
@@ -508,6 +508,20 @@ impl ScalarIndex for InvertedIndex {
             UpdateCriteria::only_new_data(criteria)
         }
     }
+
+    fn derive_index_params(&self) -> Result<ScalarIndexParams> {
+        let mut params = self.params.clone();
+        if params.base_tokenizer.is_empty() {
+            params.base_tokenizer = "simple".to_string();
```
wkalt, line 515: does this match some default we use elsewhere?

rust/lance/src/index/scalar.rs
```diff
@@ -465,6 +465,73 @@ pub fn index_matches_criteria(
...
 
+/// Initialize a scalar index from a source dataset
+pub async fn initialize_scalar_index(
+    target_dataset: &mut Dataset,
+    source_dataset: &Dataset,
+    source_index: &lance_table::format::Index,
+    column_name: &str,
+) -> Result<()> {
+    use lance_index::metrics::NoOpMetricsCollector;
+    use lance_index::scalar::inverted::tokenizer::InvertedIndexParams;
```
wkalt: I find it kind of weird that we classify inverted indexes as "scalar". I'm pretty sure this code is just mirroring existing stuff but I think from an inverted index POV the values are collections instead of scalars and it'll be more of a stretch than it is today (with strings) if we ever extend our inverted indexes to cover JSON docs or arrays.

Maybe something we can clean up later if it bugs others.

rust/lance/src/index/scalar.rs
```diff
@@ -465,6 +465,73 @@ pub fn index_matches_criteria(
...
+    use lance_index::{DatasetIndexExt, IndexType};
+
+    log::info!(
+        "Initializing scalar index '{}' on column '{}'",
```
wkalt: seems like it would be useful to log the type here if we have it

rust/lance-index/src/scalar/inverted/index.rs
```diff
@@ -508,6 +508,20 @@ impl ScalarIndex for InvertedIndex {
...
+        }
+
+        let params_json = serde_json::to_string(&params)?;
```
wkalt, line 518: it's a little weird that we need to roundtrip our own structured type through JSON. Might be worth cleaning up if it's easy.

rust/lance/src/index.rs
```diff
@@ -1461,6 +1468,136 @@ impl DatasetIndexInternalExt for Dataset {
...
+        }
+
+        // Group indices by name to avoid duplicate initialization
+        let mut unique_index_names = HashSet::new();
+        for index in non_system_indices.iter() {
+            unique_index_names.insert(index.name.clone());
+        }
+
+        log::info!(
+            "Initializing {} unique indices from source dataset",
```
wkalt: I think we can remove "unique" here, it might give the impression these are "unique indexes" which is something different.

rust/lance-index/src/scalar/inverted/index.rs
jackye1995: this is a part of the effort to make scalar index pluggable, so each index exposes parameters as JSON and can be serialized and transmitted easily through REST API in the future

rust/lance/src/index/scalar.rs
jackye1995: oh I need to clean up the logging I use for debug...

rust/lance/src/index/scalar.rs
jackye1995: I think so far the definition of "scalar" is that it is working against column(s) that is not a vector column, so inverted index also fits in the definition. But overall I agree we can clean this up a bit more.

rust/lance-index/src/scalar/inverted/index.rs
jackye1995: yes

```
impl Default for InvertedIndexParams {
    fn default() -> Self {
        Self::new("simple".to_owned(), tantivy::tokenizer::Language::English)
    }
}
```

rust/lance/src/index/vector.rs
jackye1995:
```
impl Default for PQBuildParams {
    fn default() -> Self {
        Self {
            num_sub_vectors: 16,
            num_bits: 8,
            max_iters: 50,
            kmeans_redos: 1,
            codebook: None,
            sample_rate: 256,
        }
    }
}
```

we have indeed a lot of default values just floating around, worth cleanup some time...

jackye1995: oh I get what you mean now, the max iteration can also be derived and not always the default

rust/lance-index/src/scalar/flat.rs
jackye1995: technically this would never be called by an end user

rust/lance/src/index/vector.rs
jackye1995: yeah overall we are missing some info in stats, let me fill them in as a part of this effort

rust/lance/src/index.rs
jackye1995: agree, updated

rust/lance/src/index/vector.rs
jackye1995: I have updated as much as possible, some info are lost after initial training. Looks like today if we retrain a vector index, we basically will completely fallback to the default regardless of what was the original setting. I added a TODO and will fix that separately.

rust/lance/src/index/vector.rs
jackye1995: see same comment as https://github.com/lancedb/lance/pull/4658#discussion_r2338203120

rust/lance/src/index/vector.rs
jackye1995: Moved to directly get metric_type as enum.

Issue #4663, opened 2025-09-08
perf: faster and more balanced clustering


BubbleCal:
- [ ] balanced clustering: this is needed because if some partitions go much larger than the others, any queries hitting such partitions would be with high tail latency
- [ ] faster clustering for large K (the number of partitions): to limit the size of single partition, we have to be able to train IVF with large K

PR #4664, opened 2025-09-08
feat!: use FCFC algo for training IVF

BubbleCal:
related to https://github.com/lancedb/lance/issues/4663
this algo takes cluster sizes into consideration, such that the cluster sizes can be closer, to reduce the search tail latency

the std dev of partition sizes goes down with this algo:
2388.0670596216514 -> 2021.9423056803575

mark this breaking because it deprecates the `retrain` flag for `optimize_indices`


BubbleCal:
included in https://github.com/lancedb/lance/pull/4726

PR #4667, opened 2025-09-08
feat: support btree distributely

### LINKED ISSUE #4665: feat: Support create btree index distributedly
After merging this PR (https://github.com/lancedb/lance/pull/4578), we can implement distributed construction of btree indexes, which can significantly shorten the index create time of very large datasets and make point query capabilities available on very large data sets.

xloya:
Closed #4665.

Overall Concept:
1. Create an ordered Btree sub-indexes at the fragment level based on Ray / Daft.
2. Sort and merge the sub-indexes using a multi-way merge algorithm, supporting pre-reading of sub-index files.
3. Commit the final index file.

Production Test Results:
In a production scenario, using Ray and 50 workers on a string ID field in a dataset of 700 million records, we achieved the following:
1. Btree index build time was reduced from 190 minutes to 19 minutes, a 10x increase in build speed.
2. Peak memory usage on the Ray head node was reduced from 90+ GB to 4+ GB, a 95%+ reduction.


xloya:
@jackye1995 @westonpace @BubbleCal @chenghao-guo  Please take a look when you have time, thx!

python/python/lance/dataset.py
```diff
@@ -2731,8 +2731,40 @@ def prewarm_index(self, name: str):
...
+    def merge_index_metadata(
+        self,
+        index_uuid: str,
+        index_type: Union[
+            Literal["BTREE"],
+            Literal["INVERTED"],
+        ],
+        prefetch_batch: Optional[int] = None,
+    ):
+        """
```
chenghao-guo, line 2771: Thank you for the refactor. My previous design wasn’t thoroughly considered; yours is a significant improvement

python/python/lance/dataset.py
```diff
@@ -2731,8 +2731,40 @@ def prewarm_index(self, name: str):
...
+        Merge an index which not commit at present.
```
jackye1995: nit: "which is not committed at present"

python/python/lance/dataset.py
```diff
@@ -2731,8 +2731,40 @@ def prewarm_index(self, name: str):
         """
         return self._ds.prewarm_index(name)
 
-    def merge_index_metadata(self, index_uuid: str):
-        return self._ds.merge_index_metadata(index_uuid)
+    def merge_index_metadata(
+        self,
+        index_uuid: str,
+        index_type: Union[
```
jackye1995: we are already checking type not in a list below dynamically, I think we can just say `index_type: str ` for type hint here.

python/python/tests/test_scalar_index.py
```diff
@@ -2983,3 +2983,411 @@ def test_backward_compatibility_no_fragment_ids(tmp_path):
...
+    print(f"Using index ID: {index_id}")
+    index_name = "btree_multiple_fragment_idx"
+
+    fragments = ds.get_fragments()
+    fragment_ids = [fragment.fragment_id for fragment in fragments]
+    print(f"Fragment IDs: {fragment_ids}")
+
+    for fragment in ds.get_fragments():
+        fragment_id = fragment.fragment_id
+        print(f"Creating B-tree index for fragment {fragment_id}")
```
jackye1995: nit: should remove prints in tests

python/python/tests/test_scalar_index.py
```diff
@@ -2983,3 +2983,411 @@ def test_backward_compatibility_no_fragment_ids(tmp_path):
...
+    complete_ds.create_scalar_index(
+        column="id",
+        index_type="BTREE",
+        name=complete_index_name,
+    )
+
+    print("Both indices created successfully")
+
+    # Detailed query comparison tests
+    test_cases = [
```
jackye1995: should use parameterized test

python/python/lance/dataset.py
```diff
@@ -2731,8 +2731,40 @@ def prewarm_index(self, name: str):
...
+        prefetch_batch: int, optional
+            The number of prefetch batches of sub-page files for merging.
+            Default 1.
+        """
+        index_type = index_type.upper()
+        if index_type not in [
+            "BTREE",
+            "INVERTED",
+        ]:
+            raise NotImplementedError(
```
jackye1995, line 2790: you are double checking this in both python and rust.

python/src/dataset.rs
```diff
@@ -1670,47 +1670,111 @@ impl Dataset {
...
+                        location: location!(),
+                    });
+                }
+            };
+
             let store = LanceIndexStore::from_dataset_for_new(self.ds.as_ref(), index_uuid)?;
             let index_dir = self.ds.indices_dir().child(index_uuid);
+            if idx_type == IndexType::Inverted {
+                // List all partition metadata files in the index directory
+                let mut part_metadata_files = Vec::new();
```
jackye1995: we should probably pushdown to the rust level, add functions like `lance_index::scalar::inverted::builder::list_metadata_files` and `lance_index::scalar::btree::list_page_lookup_files`

python/src/dataset.rs
```diff
@@ -1670,47 +1670,111 @@ impl Dataset {
...
+    #[pyo3(signature = (index_uuid, index_type, prefetch_batch))]
+    fn merge_index_metadata(
+        &self,
+        index_uuid: &str,
+        index_type: &str,
+        prefetch_batch: Option<usize>,
+    ) -> PyResult<()> {
         RT.block_on(None, async {
+            let index_type = index_type.to_uppercase();
+            let idx_type = match index_type.as_str() {
```
jackye1995: I think we don't need to first do a match to convert string to `IndexType` and do if else with index type. We can just do something like:

```rust
match idx_type.to_uppercase().as_str() {
  "BTREE" => {
       // do btree things
   },
   "INVERTED" => {
       // do fts things
   },
    _ => {  // fail with not supported }
}
```

rust/lance-index/src/scalar/btree.rs
```diff
@@ -54,16 +54,37 @@ use lance_datafusion::{
...
+use tokio::runtime::{Builder, Runtime};
 use tracing::info;
 
 const BTREE_LOOKUP_NAME: &str = "page_lookup.lance";
 const BTREE_PAGES_NAME: &str = "page_data.lance";
 pub const DEFAULT_BTREE_BATCH_SIZE: u64 = 4096;
 const BATCH_SIZE_META_KEY: &str = "batch_size";
+
+/// Global thread pool for B-tree prefetch operations
+static BTREE_PREFETCH_RUNTIME: LazyLock<Runtime> = LazyLock::new(|| {
```
jackye1995: I think in general we want to pass runtimes in from caller, instead of initializing static runtimes.

rust/lance-index/src/scalar/btree.rs
```diff
@@ -1393,385 +1438,2297 @@ pub async fn train_btree_index(
...
+                .await
+        }
+        .boxed();
+        std::task::Poll::Ready(Some(read_task))
+    }
+}
+
+/// Parameters for a btree index
+#[derive(Debug, Serialize, Deserialize)]
+pub struct BTreeParameters {
```
jackye1995: seems like the git diff is a bit messed up, these are shown as deleted above and added new here. Can you clean it up?

jackye1995 (commented):
this is amazing feature! I added some initial comments, will look more into the details tomorrow

xloya:
@jackye1995 Thanks for your review, address all comments.

python/python/lance/dataset.py
```diff
@@ -2731,8 +2731,38 @@ def prewarm_index(self, name: str):
...
+        """
+        Merge an index which is not commit at present.
+
+        Parameters
+        ----------
+        index_uuid: str
+            The uuid of the index which want to merge.
+        index_type: str
+            The type of the index.
+            Only "BTREE" and "INVERTED" are supported now.
```
westonpace, line 2780: Minor nit: can we load the index by uuid and use that to determine the index type?  That way we don't need to make the user provide it?

python/python/tests/test_scalar_index.py
```diff
@@ -92,6 +92,99 @@ def data_table(indexed_dataset: lance.LanceDataset):
...
+    fragment_index_name = "fragment_btree_precise_test"
+
+    fragments = fragment_ds.get_fragments()
+    fragment_ids = [fragment.fragment_id for fragment in fragments]
+    print(f"Fragment IDs: {fragment_ids}")
+
+    # Create fragment-level indices
+    for fragment in fragments:
+        fragment_id = fragment.fragment_id
+        print(f"Creating B-tree index for fragment {fragment_id}")
```
westonpace: Can we remove the prints?

python/python/tests/test_scalar_index.py
```diff
@@ -2983,3 +3076,272 @@ def test_backward_compatibility_no_fragment_ids(tmp_path):
...
+        # Use the same index_name for all fragments (like in FTS test)
+        ds.create_scalar_index(
+            column="id",  # Use integer column for B-tree
+            index_type="BTREE",
+            name=index_name,
+            replace=False,
+            fragment_uuid=index_id,
+            fragment_ids=[fragment_id],
+        )
+
```
westonpace, line 3142: Should we verify the dataset is still searchable in this state where the individual indexes have been created but not yet merged?

python/python/tests/test_scalar_index.py
```diff
@@ -2983,3 +3076,272 @@ def test_backward_compatibility_no_fragment_ids(tmp_path):
...
+    # Verify the index was created
+    indices = ds.list_indices()
+    assert len(indices) == 1
+    assert indices[0]["name"] == "full_dataset_btree_idx"
+    assert indices[0]["type"] == "BTree"
+
+    # Test that the index works
+    results = ds.scanner(filter="id = 50").to_table()
+    assert results.num_rows > 0
```
westonpace: This test is probably redundant, we have other tests that create btree indexes using the non-distributed path.

python/src/dataset.rs
```diff
@@ -1670,47 +1670,56 @@ impl Dataset {
...
+                        )
+                        .await?;
+
+                    // Call merge_metadata_files function for btree index
+                    lance_index::scalar::btree::merge_metadata_files(
+                        Arc::new(store),
+                        &part_page_files,
+                        &part_lookup_files,
+                        prefetch_batch,
                     )
```
westonpace: We can clean this up in a future PR but it seems like this should just be one method call.  In other words, we should just call `ds.merge_index_files(uuid)` and let the index figure out how to merge the files.  Otherwise this match statement is going to grow and it will be more difficult to add new indexes.

rust/lance-index/src/scalar/bitmap.rs
```diff
@@ -528,6 +528,7 @@ impl ScalarIndexPlugin for BitmapIndexPlugin {
         data: SendableRecordBatchStream,
         index_store: &dyn IndexStore,
         _request: Box<dyn TrainingRequest>,
+        _fragment_ids: Option<Vec<u32>>,
     ) -> Result<CreatedIndex> {
         Self::train_bitmap_index(data, index_store).await?;
         Ok(CreatedIndex {
```
westonpace: Let's verify that `fragment_ids` is empty and throw an error if it is not.  Or at the very least, let's do an `assert(fragment_ids.is_empty())`.

rust/lance-index/src/scalar/btree.rs
```diff
@@ -1366,10 +1369,33 @@ pub async fn train_btree_index(
     sub_index_trainer: &dyn BTreeSubIndex,
     index_store: &dyn IndexStore,
     batch_size: u64,
+    fragment_ids: Option<Vec<u32>>,
 ) -> Result<()> {
-    let mut sub_index_file = index_store
-        .new_index_file(BTREE_PAGES_NAME, sub_index_trainer.schema().clone())
-        .await?;
+    let fragment_mask = fragment_ids.as_ref().and_then(|frag_ids| {
```
westonpace, line 1386: Why is this called a mask?  Isn't it a partition id?  Also, I'm not sure this will work if multiple fragment ids are specified?

Not something we have to fix right now but maybe in the future we could just use a uuid for the partition id and then write the fragment ids into the lookup file's schema metadata?

rust/lance-index/src/scalar/btree.rs
```diff
@@ -1393,14 +1419,731 @@ pub async fn train_btree_index(
...
+    }
+
+    // Step 1: Create lookup map for page files by partition ID
+    let mut page_files_map = HashMap::new();
+    for page_file in part_page_files {
+        let partition_id = extract_partition_id(page_file)?;
+        page_files_map.insert(partition_id, page_file);
+    }
+
+    // Step 2: Validate that all lookup files have corresponding page files
```
westonpace, line 1539: Should we also verify that `part_page_files` and `part_lookup_files` have the same length?

rust/lance-index/src/scalar/btree.rs
```diff
@@ -1393,14 +1419,731 @@ pub async fn train_btree_index(
...
+    let first_lookup_reader = store.open_index_file(&part_lookup_files[0]).await?;
+    let batch_size = first_lookup_reader
+        .schema()
+        .metadata
+        .get(BATCH_SIZE_META_KEY)
+        .map(|bs| bs.parse().unwrap_or(DEFAULT_BTREE_BATCH_SIZE))
+        .unwrap_or(DEFAULT_BTREE_BATCH_SIZE);
+
+    // Get the value type from lookup schema (min column)
+    let lookup_batch = first_lookup_reader.read_range(0..1, None).await?;
```
westonpace: Can you use the schema from the file reader instead of reading one row?

rust/lance-index/src/scalar/btree.rs
```diff
@@ -1393,14 +1419,731 @@ pub async fn train_btree_index(
...
+        .new_index_file(BTREE_PAGES_NAME, arrow_schema.clone())
+        .await?;
+
+    let mut prefetch_config = PrefetchConfig::default();
+    if let Some(batch) = prefetch_batch {
+        prefetch_config = prefetch_config.with_prefetch_batch(batch);
+    }
+
+    // Step 4: Merge pages and create lookup entries
+    let lookup_entries = merge_page(
```
westonpace: Minor nit but maybe we can call this `merge_pages`?

rust/lance-index/src/scalar/btree.rs
```diff
@@ -1393,14 +1419,731 @@ pub async fn train_btree_index(
...
+            buffer_guard.remove(index)
+        } else {
+            None
+        }
+    }
+
+    #[allow(dead_code)]
+    fn get_reader(&self) -> Arc<dyn IndexReader> {
+        self.reader.clone()
+    }
```
westonpace: Let's drop this?

rust/lance-index/src/scalar/btree.rs
```diff
@@ -1393,14 +1419,731 @@ pub async fn train_btree_index(
...
+                                    break;
+                                }
+                            }
+
+                            current_pos = end_pos;
+                        }
+                        Err(_) => {
+                            // Read failed, wait before retry
+                            tokio::time::sleep(tokio::time::Duration::from_millis(10)).await;
+                        }
```
westonpace: Why retry?  Also, won't this retry indefinitely?

rust/lance-index/src/scalar/btree.rs
```diff
@@ -1393,14 +1419,731 @@ pub async fn train_btree_index(
...
+                    }
+
+                    // Add some delay to avoid excessive resource consumption
+                    tokio::time::sleep(tokio::time::Duration::from_millis(1)).await;
```
westonpace: Why delay here?  If we haven't hit prefetch limit we should keep reading?  What resource are we worried about overloading?

rust/lance-index/src/scalar/btree.rs
```diff
@@ -1393,14 +1419,731 @@ pub async fn train_btree_index(
...
+            // Get required data by slicing from prefetch buffer
+            let slice_start = self.rows_read - entry.start_row;
+            let slice_len = rows_to_read;
+            entry.batch.slice(slice_start, slice_len)
+        } else {
+            // Fallback to direct read
+            self.reader
+                .read_range(self.rows_read..end_row, None)
+                .await?
+        };
```
westonpace: This seems prone to race condition.  What if the main reader gets ahead of the prefetch reader.  For example, first call to `load_next_batch` we decide to fallback to direct read.  Meanwhile the prefetcher grabs the first batch.  Then on the next call to `load_next_batch` won't we return the first batch again?

rust/lance-index/src/scalar/btree.rs
```diff
@@ -1393,14 +1419,731 @@ pub async fn train_btree_index(
...
+/// Partition iterator for loading partition data in batches with integrated prefetching
+struct PartitionIterator {
+    reader: Arc<dyn IndexReader>,
+    current_batch: Option<RecordBatch>,
+    current_position: usize,
+    rows_read: usize,
+    batch_size: u64,
+    prefetch_buffer: Arc<tokio::sync::Mutex<VecDeque<BufferEntry>>>,
+    prefetch_task: Option<tokio::task::JoinHandle<()>>,
+}
```
westonpace: I don't think any of this prefetch is needed.  The `FileReader` already does its own prefetch.  Instead of making multiple calls to `read_range` just make a single call to `read_stream` and then use `chunk_concat_stream` to force the stream to emit batches of the correct size.  You will need to add `read_stream` to the `IndexReader` trait but I think that is ok.

The `read_stream` method takes in a `batch_readahead` parameter which controls how many batches it will decode in advance.  There is also a configuration which controls how much data we load in advance and this will default to 2GiB.  However, this is set at the index store level and so it should be a shared 2GiB limit no matter how many readers you have open (e.g. it will not load 2GiB per reader) so we should be fine.

python/python/lance/dataset.py
xloya: Hi, if the user does not input the index type, we may only be able to find the index directory by uuid and find the file name created in the distributed stage of each index by list directory to determine which index is used now because the index does not exist in the metadata yet . Do you think this is a good practice?

rust/lance-index/src/scalar/btree.rs
xloya: In fact, here I refer to the current inverted index distributed construction sub-index shard naming method: https://github.com/lancedb/lance/blob/main/rust/lance-index/src/scalar/inverted.rs#L51-L60.
For more details, please refer to: https://github.com/lancedb/lance/pull/4578#discussion_r2322160835.

rust/lance-index/src/scalar/btree.rs
xloya: In fact, when loading a batch from the prefetch buffer, it will choose whether to read the current batch based on the offset. In addition, the default value of the direct read batch is `batch_size`, while the prefetch batch is `prefetch * batch_size`, which is more useful for setting a large prefetch. However, you mentioned using streams directly, so I will see how to refactor this.

rust/lance-index/src/scalar/btree.rs
xloya: Let me see how to refactor this.

xloya:
> Also, does this merge the batches themselves? For example, if I sort fragment 1 and I get a batch with values 10-50 and then I sort fragment 2 and I get a batch with values 30-80 is there some code that is merging these two batches (maybe the output is a batch with 10-50 and 50-80)? I might be missing it but maybe it is just emitting 10-50 and then 30-80 in sequence?
> 
I might not fully understand what's going on here. Let me explain the logic using these two fragments:
1. Use `train_btree_index` to construct sub-indexes for fragment_1 and fragment_2. If they are grouped together, I believe this code will ensure strict ordering: https://github.com/lancedb/lance/blob/main/rust/lance/src/index/scalar.rs#L131-L135, which generates the sequence [10, 80]. If they are not grouped together, the sequences [10, 50] and [30, 80] are generated, respectively.
2. Use the k-way merge algorithm to merge them, ensuring the final order.

westonpace:
> I might not fully understand what's going on here. Let me explain the logic using these two fragments:
> 
>  1.  Use train_btree_index to construct sub-indexes for fragment_1 and fragment_2. If they are grouped together, I believe this code will ensure strict ordering: https://github.com/lancedb/lance/blob/main/rust/lance/src/index/scalar.rs#L131-L135, which generates the sequence [10, 80]. If they are not grouped together, the sequences [10, 50] and [30, 80] are generated, respectively.
>  2.  Use the k-way merge algorithm to merge them, ensuring the final order.

Ah, I see where I was confused.  I thought the partition iterator was yielding batches.  Instead it is yielding rows.  So the heap is built one row at a time and batches are merged that way.  You can ignore that particular comment.

python/python/lance/dataset.py
westonpace: Yes, I see now.  There are ways we can avoid this (e.g. write the index type into a file somewhere) and I think we will need to do something like that in the future.  But we can save that for a future PR.

xloya:
@jackye1995 @westonpace @BubbleCal  Hi, I've refactored the code based on your feedback. Please review it again when you have time, thanks! I've also noticed that using Datafusion's SortPreservingMerge has slowed down index creation compared to my previous version(the speed can be improved by adjusting the prefetch nums and the number of sub-indexes), but it's still significantly faster than the current single-node creation. Furthermore, the code has been significantly simplified. I think we can further optimize performance in future PRs.

xloya:
@jackye1995 @westonpace @BubbleCal Gentle pin for this

PR #4668, opened 2025-09-08
chore(java): adjust java module directory structure

majin1102:
When I wanted to add the Shade plugin and Guava dependency to the Java module, I realized the directory structure was somewhat suboptimal—since we no longer need submodules like Spark in the Java module, I removed the nested core directory and relocated the src and lance-jni directories to the Java root directory. 

I believe this will improve clarity and maintainability as more big data developers join this module. If any concern remains uncovered, please let me know.

This PR does not involve any changed in actual code. What do you think about this @jackye1995


jackye1995:
I think the key question is, do we expect other potential Java sub-modules here, or are we going to create sub-projects for all those use cases? I thought the Guava work would actually need sub-modules so we have something like

```
java
  - lance-shaded-guava
  - lance-core
```

without that, are you going to directly shade it in the `lance-core`?

majin1102:
> I think the key question is, do we expect other potential Java sub-modules here, or are we going to create sub-projects for all those use cases? I thought the Guava work would actually need sub-modules so we have something like
> 
> ```
> java
>   - lance-shaded-guava
>   - lance-core
> ```
> 
> without that, are you going to directly shade it in the `lance-core`?


I was preparing to use Maven Shade Plugin to dynamic relocate guava classes like Spark. I noted Iceberg used a different way. I did some search and mostly recommend Spark way. That's why I raised this PR to talk about it.

Anyway this was only an attempt. Could you elaborate more about your thoughts on the the comparison?

Except shade scenarios, I don't figure any other potential Java sub-modules(For now I think the sub-project way more suitable for Java eco-systems). I think the final look would be quite similar to Python module

jackye1995:
> I was preparing to use Maven Shade Plugin to dynamic relocate guava classes like Spark. I noted Iceberg used a different way. I did some search and mostly recommend Spark way. That's why I raised this PR to talk about it.

I see, that sounds fine to me, I know both ways exist but don't know their pros and cons.

> Anyway this was only an attempt. Could you elaborate more about your thoughts on the the comparison?
> Except shade scenarios, I don't figure any other potential Java sub-modules(For now I think the sub-project way more suitable for Java eco-systems). I think the final look would be quite similar to Python module

Cool then let's go with this. I cannot think of any major reason either. If things come up we can always add additional child modules later.

PR #4670, opened 2025-09-08
feat: add LanceFileSession for sharing object store amongst file readers

westonpace:


python/python/lance/file.py
```diff
@@ -184,6 +191,49 @@ def num_rows(self) -> int:
         return self._reader.num_rows()
 
 
+class LanceFileSession:
+    """
+    A file session for reading Lance files.
+
+    If you plan on opening many readers then creating a session first can be more
+    efficient as it will share the underlying object_store configuration with all
+    of the readers.
```
wkalt, line 200: Why wouldn't someone always want that?

westonpace: Just a less convenient API to have to make two calls instead of one.

```
reader = LanceFileReader("/my/dataset/foo.lance")
```

vs...

```
session = LanceFileSession("/my/dataset")
reader = session.open_reader("foo.lance")
```

westonpace:
> it looks ok but I'm still wondering if this should be an optional mechanism or default behavior

It'd be nice to make it default but then that means we are caching object_store instances process-wide and I'm not entirely sure the implications of that.

jmhsieh:
kewl.  lmk when I can get a beta release from fury.

Issue #4477, opened 2025-08-14
Index cache seems to double count some memory


wjones127:
When using the index cache, the reported weighted size of the cache (which should be the size in bytes), tends to be about double the observed memory use of our pods. This suggests that we are double counting some structures.

We want to figure out why this is and see if we can eliminate or at least mitigate this impact.

First, let's implement `DeepSizeOf` for `LanceCache` that does an exact computation. It should iterator over the keys and values and get their exact sizes. `DeepSizeOf` passes down a context that tracks `Arc` pointers that have already been seen, so it should avoid the double counting.

Then we can start comparing the true `DeepSizeOf` value against the weighted size of the cache to see when they are different. Strategically placed `println` in the `DeepSizeOf` impls might help us see where the duplication is happening.

Xuanwo:
Let me take a look. Is this an emergency?

wjones127:
Not a true emergency, but it is high priority.

Xuanwo:
Let me take a look over this.

wjones127:
## Test plan

There is one main critical property we need to test:

> If I free() `T`, the drop in memory allocated should be (approximately) equal to `DeepSizeOf::<T>::size_of()`.

<details><summary>Measuring memory allocated</summary>
<p>

```rust
use std::alloc::{GlobalAlloc, Layout, System};
use std::sync::atomic::{AtomicUsize, Ordering};

struct TrackingAllocator {
    inner: System,
    allocated: AtomicUsize,
    allocations: AtomicUsize,
}

unsafe impl GlobalAlloc for TrackingAllocator {
    unsafe fn alloc(&self, layout: Layout) -> *mut u8 {
        let ptr = self.inner.alloc(layout);
        if !ptr.is_null() {
            self.allocated.fetch_add(layout.size(), Ordering::Relaxed);
            self.allocations.fetch_add(1, Ordering::Relaxed);
        }
        ptr
    }

    unsafe fn dealloc(&self, ptr: *mut u8, layout: Layout) {
        self.allocated.fetch_sub(layout.size(), Ordering::Relaxed);
        self.allocations.fetch_sub(1, Ordering::Relaxed);
        self.inner.dealloc(ptr, layout);
    }
}

#[global_allocator]
static ALLOCATOR: TrackingAllocator = TrackingAllocator {
    inner: System,
    allocated: AtomicUsize::new(0),
    allocations: AtomicUsize::new(0),
};
```

</p>
</details> 

[...]

PR #4680, opened 2025-09-09
fix: circular reference in index caching leaks memory

wkalt:
Prior to this commit, when a dataset loaded an index into its index cache, it would attach a strong reference to its index index cache to the index.

This resulted in a circular reference: the dataset's cache contained self-referential values. This prevented the cache from getting deallocated when the dataset was dropped, resulting in a leak if many datasets were opened over time.

This commit puts weak references to the cache on the indexes so the cycle is broken.

### DIFF
--- a/python/python/tests/test_memory_leaks.py
+++ b/python/python/tests/test_memory_leaks.py
@@ -1,0 +1,96 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright The Lance Authors
+
+from __future__ import annotations
+
+import gc
+import os
+from typing import Callable
+
+import lance
+import psutil
+import pyarrow as pa
+
+MiB = 1024 * 1024
+
+
+def get_memory_usage() -> int:
+    return psutil.Process(os.getpid()).memory_info().rss
+
+
+def assert_noleaks(
+    operation: Callable[[], None],
+    *,
+    iterations: int = 100,
+    warmup_iterations: int = 5,
+    threshold_mb: float = 1.0,
+    check_interval: int = 10,
+    leeway_factor: float = 2.0,  # optional jitter cushion
+) -> None:
+    """Check if an operation retains memory across repeated executions.
+
+    Args:
+        operation: A callable that performs the operation to test
+        iterations: Number of times to run the operation
+        warmup_iterations: Number of warmup runs before measuring
+        threshold_mb: Maximum allowed memory growth in MB
+        check_interval: How often to check memory during iterations
+        leeway_factor: Factor to multiply threshold for early bailout
+
+    Raises:
+        AssertionError: If memory leak is detected
+    """
+    if iterations <= 0:
+        raise ValueError("iterations must be > 0")
+    if check_interval <= 0:
+        raise ValueError("check_interval must be > 0")
+
+    for _ in range(warmup_iterations):
+        operation()
+    gc.collect()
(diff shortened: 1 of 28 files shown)


rust/lance-core/src/cache.rs
```diff
@@ -107,7 +123,11 @@ impl LanceCache {
     pub fn with_key_prefix(&self, prefix: &str) -> Self {
         Self {
             cache: self.cache.clone(),
-            prefix: format!("{}{}/", self.prefix, prefix),
+            prefix: if self.prefix.is_empty() {
+                prefix.to_string()
+            } else {
+                format!("{}/{}", self.prefix, prefix)
+            },
```
wkalt: need to double check this. I didn't understand how with_key_prefix and get_key were consistent previously

rust/lance-core/src/cache.rs
```diff
@@ -324,6 +340,169 @@ impl LanceCache {
     }
 }
 
+/// A weak reference to a LanceCache, used by indices to avoid circular references.
+/// When the original cache is dropped, operations on this will gracefully no-op.
+#[derive(Clone, Debug)]
+pub struct WeakLanceCache {
+    inner: std::sync::Weak<Cache<(String, TypeId), SizedRecord>>,
+    prefix: String,
+}
```
wjones127: Did you consider making this not have any methods other than `fn upgrade(&self) -> Option<LanceCache>`?

rust/lance/src/session.rs
```diff
@@ -24,6 +26,10 @@ pub mod index_extension;
 /// A user session tracks the runtime state.
 #[derive(Clone)]
 pub struct Session {
+    /// Strong references to the actual caches - Session owns these
+    index_cache_arc: Arc<Cache<(String, TypeId), lance_core::cache::SizedRecord>>,
+    metadata_cache_arc: Arc<Cache<(String, TypeId), lance_core::cache::SizedRecord>>,
```
wjones127: Doesn't it seem smelly to duplicate these? Why aren't `GlobalIndexCache` and `GlobalMetadataCache` sufficient here?

rust/lance-core/src/cache.rs
wjones127: It feels unfortunate we would have to duplicate these methods, but it does seem like it adds meaningful convenience, so I'm good with it 👍

python/python/tests/test_memory_leaks.py
wjones127, line 1: This is cool!

I want to add more memory tests soon: https://github.com/lancedb/lance/issues/4477#issuecomment-3259487072

Maybe this can be of use :)

rust/lance/src/session.rs
```diff
@@ -94,9 +95,28 @@ impl Session {
...
+        use lance_core::cache::SizedRecord;
+        use moka::future::Cache;
+
+        let index_cache_arc = Arc::new(
+            Cache::builder()
+                .max_capacity(index_cache_size as u64)
+                .weigher(|_, v: &SizedRecord| v.deep_size_of().try_into().unwrap_or(u32::MAX))
+                .support_invalidation_closures()
+                .build(),
+        );
```
wjones127: Why was this brought out of the LanceCache constructor?

wkalt: ah, good catch, this fell out of some rearranging I had previously done. Will address.

PR #4682, opened 2025-09-09
fix: vector index on nested field

jackye1995:
## Summary

This PR fixes the optimization of vector indices on nested fields (struct sub-fields) in Lance. Previously, operations like index optimization would fail when working with nested fields like `embedding_data.vector` because the code was using only the leaf field name instead of the full field path.

## Problem

When creating indices on nested fields using dot notation (e.g., `embedding_data.vector`), the following issues occurred:
- Index optimization would fail with "Column vector does not exist" error
- Index statistics retrieval would fail after optimization
- Remapping operations would use incorrect field paths

## Solution

The fix uses `field_ancestry_by_id` method to retrieve the complete field path from root to the target field, ensuring all index operations use the full nested path.

### Key Changes

1. **`merge_indices_with_unindexed_frags` in append.rs**:
   - Now constructs full field path using `field_ancestry_by_id`
   - All operations use the complete path instead of just field name

2. **`remap_index` in index.rs**:
   - Fixed to use full field path for both scalar and vector indices

3. **`index_statistics` in index.rs**:
   - Properly constructs full field path for opening indices

## Testing

Added comprehensive test `test_vector_index_on_nested_field` that verifies:
- ✅ Creating vector indices on nested fields
- ✅ Querying with nested field indices
- ✅ **Optimizing indices on nested fields (previously failed)**
- ✅ Index statistics retrieval
- ✅ Combined queries with filters

All existing tests continue to pass with no regressions.

## Example

```rust
// Schema with nested field
let struct_field = Field::new(
    "embedding_data",
    DataType::Struct(vec![
        Field::new("vector", DataType::FixedSizeList(...), false),
    ].into()),
    false,
);

// Create index on nested field
dataset.create_index(
    &["embedding_data.vector"],  // Nested field path
    IndexType::Vector,
    Some("nested_vec_idx".to_string()),
    &params,
    true,
).await.unwrap();

// Now optimization works!
dataset.optimize_indices(&OptimizeOptions::append()).await.unwrap();
```

## Impact

This change enables full support for vector indices on nested fields within struct types, allowing users to:
- Organize embeddings and metadata in structured fields
- Optimize performance of nested field indices
- Use Lance with more complex data schemas


rust/lance-core/src/datatypes/schema.rs
```diff
@@ -221,20 +221,36 @@ impl Schema {
         self.do_project(columns, false)
     }
 
-    /// Check that the top level fields don't contain `.` in their names
-    /// to distinguish from nested fields.
+    /// Check that field names don't contain `.` to avoid ambiguity in field paths,
+    /// and validate other schema constraints.
```
jackye1995: Not sure if this is the right way. If we don't do this, then using field path would cause potential conflict (although the chance is very small, but it's a correctness issue). 

An alternative is to set a delimiter explicitly with default dot, so user with dot in field names can use an alternative delimiter.

Another way is to use field ID to identify a column, but that will touch almost all index APIs.

rust/lance-core/src/datatypes/schema.rs
```diff
@@ -1192,15 +1232,365 @@ impl Projection {
...
+/// Backticks within quoted fields must be escaped by doubling them.
+/// For example: "`field``with``backticks`" represents the field name "field`with`backticks"
+///
+/// Returns an error if:
+/// - The input path is empty
+/// - The path has malformed quotes (unclosed, misplaced, etc.)
+/// - The path has empty segments (e.g., "parent..child" or "parent.")
+///
+/// The result is guaranteed to contain at least one element.
+pub fn parse_field_path(path: &str) -> Result<Vec<String>> {
```
jackye1995, line 1300: Originally we discussed we want to use double quote, but I realized that double quoted identifier is parsed as string literal and backquotes are parsed as column identifiers in our implementation of `Planner::parse_sql_expr`, so backquote should be the right escape character.

rust/lance/src/index.rs
```diff
@@ -3983,4 +3990,545 @@ mod tests {
...
+
+    #[tokio::test]
+    async fn test_vector_index_on_nested_field_with_dots() {
+        let dimensions = 16;
+        let num_rows = 256;
+
+        // Create schema with nested field containing dots in the name
+        let struct_field = Field::new(
+            "embedding_data",
+            DataType::Struct(
```
jackye1995, line 4002: currently there is no nested field datagen util. I will put another PR to refactor these tests.

rust/lance/src/index.rs
```diff
@@ -3983,4 +3990,545 @@ mod tests {
...
+            .scan()
+            .filter("metadata.`status.code` = 5")
+            .unwrap()
+            .try_into_batch()
+            .await
+            .unwrap();
+
+        // Should have ~100 rows with status code 5
+        assert!(results.num_rows() > 0);
+        assert_eq!(results.num_rows(), 100);
```
wkalt, line 4395: these seem redundant, and the comment seems wrong because the assertion is exact

rust/lance/src/index.rs
```diff
@@ -3983,4 +3990,545 @@ mod tests {
...
+        for batch in results {
+            found_count += batch.num_rows();
+        }
+        // We expect to find approximately 1/3 of documents (those with i % 3 == 1)
+        assert!(
+            found_count > 0,
+            "Should find at least some documents with 'machine learning'"
+        );
+        assert!(found_count < num_rows, "Should not match all documents");
+    }
```
wkalt, line 4533: @jackye1995 do you have any tests that exercise the query execution on a deeper nesting than 2? Also interested in the behavior in cases like,
* what happens when you have a double-nested structure like {"a": {"b": {"c": 10 } } } and you project both "a.b" and "a.b.c".
* what happens when you have {"a": {"b": {"c": 1, "d": 2}}} and you have an index on a.b.c, and you project all of a where a.b.c = 10. i.e project the whole parent based on an indexed subfield filter. Also is the output sub-field ordering the same whether the indexed select is c or d.
* What happens when there are filters on both a.b.c and a.b.d, and either/both of them are indexed. Is the column ordering consistent, do we end up with any duplicate fields or similar in execution.

Also, the tests here have a lot of boilerplate. I wonder if we could structure them to be more table-driven and extract the boilerplate/setup from the assertion logic a bit.

wkalt: btw, sorry if some of that stuff is tested and I missed it

rust/lance-index/src/scalar/expression.rs
```diff
@@ -1096,21 +1105,73 @@ fn maybe_column(expr: &Expr) -> Option<&str> {
...
+    parts.reverse();
+
+    // Format the path correctly
+    let field_refs: Vec<&str> = parts.iter().map(|s| s.as_str()).collect();
+    Some(lance_core::datatypes::format_field_path(&field_refs))
+}
+
 // Extract a column from the expression, if it is a column, and we have an index for that column, or None
 //
 // There's two ways to get a column.  First, the obvious way, is a
```
westonpace, line 1152:
```suggestion
// There's several ways to get a column.  First, the obvious way, is a
```

westonpace (approved):

wkalt (approved):
@jackye1995 approved assuming we're comfortable with the test cases I mentioned

jackye1995:
thanks, I will add a few more tests described above

PR #4683, opened 2025-09-10
feat: add scalar index support in Java/JNI interface

beinan:
This PR will add basic scalar index support in Java/JNI interface


jackye1995:
looks like this is a duplicate of https://github.com/lancedb/lance/pull/4301. I have some comment there, and there are also some cleanup done in that PR that are good. Could you apply those changes here? We can mark both as co-authors

beinan:
> looks like this is a duplicate of #4301. I have some comment there, and there are also some cleanup done in that PR that are good. Could you apply those changes here? We can mark both as co-authors

is it possible to merge this pr first? and then we can make another PR just for the clean up?

jackye1995:
it's not a merge first and cleanup situation. Technically I still have mostly the same comment here, 

(1) IndexParams used to basically mean VectorIndexParams, with this change we should remove fields like `distanceType` within it to make it clear

(2) we just did a refactoring of ScalarIndexParams in rust and we should match it in Java JNI and Java interfaces level. By doing that, you also don't need the specific index params like `BTreeIndexParams` and `ZonemapIndexParams` because it's a generic ScalarIndexParams with index type string and JSON string parameters.

beinan:
> it's not a merge first and cleanup situation. Technically I still have mostly the same comment here,
> 
> (1) IndexParams used to basically mean VectorIndexParams, with this change we should remove fields like `distanceType` within it to make it clear
> 
> (2) we just did a refactoring of ScalarIndexParams in rust and we should match it in Java JNI and Java interfaces level. By doing that, you also don't need the specific index params like `BTreeIndexParams` and `ZonemapIndexParams` because it's a generic ScalarIndexParams with index type string and JSON string parameters.

got it, I will make the change as you said

beinan:
> it's not a merge first and cleanup situation. Technically I still have mostly the same comment here,
> 
> (1) IndexParams used to basically mean VectorIndexParams, with this change we should remove fields like `distanceType` within it to make it clear
> 
> (2) we just did a refactoring of ScalarIndexParams in rust and we should match it in Java JNI and Java interfaces level. By doing that, you also don't need the specific index params like `BTreeIndexParams` and `ZonemapIndexParams` because it's a generic ScalarIndexParams with index type string and JSON string parameters.

Hi @jackye1995 , I think I fixed the issue you mentioned, could you help take a look? Thanks!

java/lance-jni/Cargo.lock
jackye1995, line 1: there shouldn't be any change in the lock?

java/src/main/java/com/lancedb/lance/index/IndexParams.java
```diff
@@ -21,39 +22,39 @@
...
   private final Optional<VectorIndexParams> vectorIndexParams;
+  private final Optional<ScalarIndexParams> scalarIndexParams;
 
   private IndexParams(Builder builder) {
-    this.distanceType = builder.distanceType;
     this.vectorIndexParams = builder.vectorIndexParams;
+    this.scalarIndexParams = builder.scalarIndexParams;
   }
 
   public static class Builder {
```
jackye1995, line 42: we should add a `public static Builder builder() { ... }` method

java/src/main/java/com/lancedb/lance/index/IndexType.java
```diff
@@ -19,6 +19,7 @@ public enum IndexType {
   BITMAP(2),
   LABEL_LIST(3),
   INVERTED(4),
+  ZONEMAP(8),
```
jackye1995, line 25: also missing NGRAM, refer to the other PR

java/src/main/java/com/lancedb/lance/index/vector/VectorIndexParams.java
```diff
@@ -199,8 +199,8 @@ public VectorIndexParams build() {
     }
   }
 
-  public DistanceType getDistanceType() {
-    return distanceType;
+  public String getDistanceType() {
```
jackye1995: this change seems unnecessary?

java/src/test/java/com/lancedb/lance/ScalarIndexTest.java
```diff
@@ -0,0 +1,104 @@
...
+
+    try (BufferAllocator allocator = new RootAllocator()) {
+      try (Dataset dataset =
+          Dataset.create(allocator, datasetPath, schema, new WriteParams.Builder().build())) {
+
+        // Create Zonemap scalar index parameters (no additional params needed)
+        ScalarIndexParams scalarParams = ScalarIndexParams.create("zonemap");
+
+        IndexParams indexParams =
+            new IndexParams.Builder().setScalarIndexParams(scalarParams).build();
```
jackye1995: zone map index supports parameter `rows_per_zone`, we should try to test that and verify it has been set properly.

java/lance-jni/src/blocking_dataset.rs
```diff
@@ -631,12 +634,31 @@ fn inner_create_index(
     let columns = env.get_strings(&columns_jobj)?;
     let index_type = IndexType::try_from(index_type_code_jobj)?;
     let name = env.get_string_opt(&name_jobj)?;
-    let params = get_index_params(env, params_jobj)?;
     let replace = replace_jobj != 0;
     let columns_slice: Vec<&str> = columns.iter().map(AsRef::as_ref).collect();
+
+    // Handle scalar vs vector indices differently and get params before borrowing dataset
+    let params_result: Result<Box<dyn IndexParams>> = match index_type {
+        IndexType::BTree | IndexType::ZoneMap => {
```
jackye1995: we should list out all the options here

java/lance-jni/src/blocking_dataset.rs
```diff
@@ -631,12 +634,31 @@ fn inner_create_index(
...
+            let (index_type_str, params_opt) = get_scalar_index_params(env, params_jobj)?;
+            let scalar_params = lance_index::scalar::ScalarIndexParams {
+                index_type: index_type_str,
+                params: params_opt,
+            };
+            Ok(Box::new(scalar_params))
+        }
+        _ => {
+            // For vector indices, use the existing parameter handling
+            get_index_params(env, params_jobj)
```
jackye1995: should probably rename this to `get_vector_index_params`

java/src/test/java/com/lancedb/lance/ScalarIndexTest.java
```diff
@@ -0,0 +1,104 @@
...
+                Field.nullable("id", new ArrowType.Int(32, true)),
+                Field.nullable("name", new ArrowType.Utf8())),
+            null);
+
+    try (BufferAllocator allocator = new RootAllocator()) {
+      try (Dataset dataset =
+          Dataset.create(allocator, datasetPath, schema, new WriteParams.Builder().build())) {
+
+        // Create BTree scalar index parameters
+        ScalarIndexParams scalarParams = ScalarIndexParams.create("btree", "{\"zone_size\": 2048}");
```
jackye1995, line 54: should verify zone_size is set in created index

java/src/test/java/com/lancedb/lance/ScalarIndexTest.java
```diff
@@ -0,0 +1,104 @@
...
+        // Create BTree index on 'id' column
+        dataset.createIndex(
+            Collections.singletonList("id"),
+            IndexType.BTREE,
+            Optional.of("btree_id_index"),
+            indexParams,
+            false);
+
+        // Verify index was created
+        assert dataset.listIndexes().contains("btree_id_index");
```
jackye1995: nit: assert should use junit5 Assertions

java/src/test/java/com/lancedb/lance/ScalarIndexTest.java
```diff
@@ -0,0 +1,104 @@
...
+        // Create Zonemap index on 'value' column
+        dataset.createIndex(
+            Collections.singletonList("value"),
+            IndexType.ZONEMAP,
+            Optional.of("zonemap_value_index"),
+            indexParams,
+            false);
+
+        // Verify index was created
+        assert dataset.listIndexes().contains("zonemap_value_index");
```
jackye1995: nit: assert should use junit5 Assertions

java/src/main/java/com/lancedb/lance/index/vector/VectorIndexParams.java
jackye1995: oh I see, the JNI binding was trying to get a string. But I think returning DistanceType makes more sense at Java level? Can we just fix the JNI?

java/lance-jni/Cargo.lock
beinan: The jni module looks like a little bit behind other module,  I guess no one has been touch it for a while.  I looked into other lock files, the datafusion is 49.0.2.   Or maybe I should submit another PR to upgrade the datafusion in JNI?

java/src/main/java/com/lancedb/lance/index/vector/VectorIndexParams.java
beinan: If we fixed on jni side, there would be a `reverse` call from native to java, which might be much more expensive then calling from java to native.   How about we add another method getDistanceTypeString() to return the string value?

java/lance-jni/src/blocking_dataset.rs
beinan: done

java/lance-jni/src/blocking_dataset.rs
beinan: done

java/src/main/java/com/lancedb/lance/index/vector/VectorIndexParams.java
jackye1995: sounds good!

java/lance-jni/Cargo.lock
jackye1995: I see. then let's just keep it here

jackye1995: well, I guess it's up to you if you want to do a separate PR. I am fine either way

java/src/main/java/com/lancedb/lance/index/IndexType.java
beinan: added

java/src/test/java/com/lancedb/lance/ScalarIndexTest.java
beinan: it might be difficult for now,  so I added a TODO here,  probably we need implement a getIndexDetails method in java, what do you think?

java/src/test/java/com/lancedb/lance/ScalarIndexTest.java
beinan: good call, done

java/src/test/java/com/lancedb/lance/ScalarIndexTest.java
beinan: good call, I added rows_per_zone in the test parameters, but now I cannot verify it, so I just left a todo for now

java/src/test/java/com/lancedb/lance/ScalarIndexTest.java
beinan: done

java/src/main/java/com/lancedb/lance/index/vector/VectorIndexParams.java
beinan: done

java/lance-jni/Cargo.lock
beinan: pr for the lock file: https://github.com/lancedb/lance/pull/4693

java/src/main/java/com/lancedb/lance/index/IndexParams.java
beinan: yes, added

java/src/test/java/com/lancedb/lance/ScalarIndexTest.java
jackye1995: sounds good!

beinan:
> looks good to me, pending CI fixes

Hi @jackye1995  Thank you for the thoughtful review! All CI checks have passed.

Issue #4520, opened 2025-08-20
FTS in SQL


westonpace:
This is related to https://github.com/lancedb/lance/issues/1782 but while that issue mostly discusses vector search this issue can be for full text search.

There are a few broad approaches (more discussion in the linked issue):

* The FTS search could be recognized automatically from a standard SQL query (e.g. with a sort and order by)
* The FTS search could be recognized via a UDTF
* The FTS search could be recognized as a custom table provider

wojiaodoubao:
Hi @westonpace @BubbleCal , I'd like to help with this issue, I think may be I can start with the UDTF part.

PR #4684, opened 2025-09-10
feat: add udtf for fts query

wojiaodoubao:
This is the second step of https://github.com/lancedb/lance/issues/4520

### DIFF
--- a/rust/lance/src/dataset/udtf.rs
+++ b/rust/lance/src/dataset/udtf.rs
@@ -168,7 +168,7 @@ impl TableProvider for FtsTableProvider {
 ///  .unwrap();
 /// ```
 #[derive(Debug)]
-struct FtsQueryUDTF {
+pub struct FtsQueryUDTF {
     datasets: HashMap<String, Arc<Dataset>>,
 }
 
@@ -233,12 +233,10 @@ fn parse_query_options(options: &str) -> datafusion::common::Result<(bool, bool,
 }
 
 /// Builder of `FtsQueryUDTF`
-#[allow(dead_code)]
-struct FtsQueryUDTFBuilder {
+pub struct FtsQueryUDTFBuilder {
     datasets: HashMap<String, Arc<Dataset>>,
 }
 
-#[allow(dead_code)]
 impl FtsQueryUDTFBuilder {
     pub fn builder() -> Self {
         Self {
@@ -330,6 +328,7 @@ pub mod tests {
                 "with_row_id": true
             }
             "#;
+
         let df = ctx
             .sql(&format!(
                 "SELECT * FROM fts('foo', '{}', '{}') WHERE number > 1",
--- a/rust/lance/src/dataset/udtf.rs
+++ b/rust/lance/src/dataset/udtf.rs
@@ -260,8 +260,7 @@ impl FtsQueryUDTFBuilder {
 
 #[cfg(test)]
 pub mod tests {
-    use crate::datafusion::LanceTableProvider;
-    use crate::dataset::udtf::{FtsQueryUDTF, FtsQueryUDTFBuilder};
+    use crate::dataset::udtf::FtsQueryUDTFBuilder;
     use crate::Dataset;
     use arrow_array::{
         Array, Int32Array, RecordBatch, RecordBatchIterator, StringArray, UInt64Array,
--- a/rust/lance/src/dataset/udtf.rs
+++ b/rust/lance/src/dataset/udtf.rs
@@ -1,0 +1,356 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright The Lance Authors
+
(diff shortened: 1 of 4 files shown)


rust/lance/src/dataset/udtf.rs
```diff
@@ -0,0 +1,350 @@
...
+///  .unwrap();
+/// ```
+#[derive(Debug)]
+struct FtsQueryUDTF {
+    datasets: HashMap<String, Arc<Dataset>>,
+}
+
+impl FtsQueryUDTF {
+    /// Build a FtsQueryUDTF from session context. Tables already registered in session context are
+    /// allowed in FtsQueryUDTF. The tables registered after this function can't be recognized.
```
jackye1995: this seems like a strange experience. 

```
/// let ctx = SessionContext::new();
/// let fts_query_udtf = FtsQueryUDTFBuilder::builder()
///   .register_table("HarryPotter_Chapter1", data1.clone())
///   .register_table("HarryPotter_Chapter2", data2.clone())
///   .build();
/// ctx.register_udtf("fts", Arc::new(fts_query_udtf));
```

this is pretty straightforward, but I don't see it in the code

wojiaodoubao: Sorry @jackye1995 , I made it wrong. Originally I registered table through FtsQueryUDTFBuilder. Later, I realized that since we register tables to SessionContext, it might be a better approach to retrieve the registered tables from the SessionContext and register them to the UDTF. So, I made some modifications but missed the comments.

jackye1995: I see your point now... seems like there is no really ideal way around this... if there is no other options, I feel we should stick with manual registration, otherwise we could end up with a lot of tables that are registered with the UDTF but not really using it. What do you think?

wojiaodoubao: Hi @jackye1995 , sorry somehow I missed this comment. I agree with your opinion and have made the changes. 

I also tried other approaches, hoping to retrieve the dataset from the SessionContext during the UDTF execution. However, all attempts failed due to async and sync issues. So, for now, there really doesn't seem to be a better solution.

I know you are very busy recently, this pr is not hurry, you can review whenever you have time, thanks very much !

rust/lance/src/dataset/udtf.rs
```diff
@@ -0,0 +1,352 @@
...
+        .is_some_and(|v| v.as_bool().unwrap_or(false));
+    let ordered = value
+        .get("ordered")
+        .is_some_and(|v| v.as_bool().unwrap_or(false));
+    Ok((with_row_id, with_row_addr, ordered))
+}
+
+/// Builder of `FtsQueryUDTF`
+#[allow(dead_code)]
+struct FtsQueryUDTFBuilder {
```
jackye1995: I think we should make this struct and the module public so people can use it directly?

jackye1995 (commented):
mostly looks good to me, one minor comment

rust/lance/src/dataset/udtf.rs
wojiaodoubao: Thanks! Fixed it.

rust/lance/src/dataset/udtf.rs
```diff
@@ -0,0 +1,352 @@
...
+        .get("with_row_addr")
+        .is_some_and(|v| v.as_bool().unwrap_or(false));
+    let ordered = value
+        .get("ordered")
+        .is_some_and(|v| v.as_bool().unwrap_or(false));
+    Ok((with_row_id, with_row_addr, ordered))
+}
+
+/// Builder of `FtsQueryUDTF`
+#[allow(dead_code)]
```
jackye1995: I think we can remove this now given it's public? (Same for the impl)

wojiaodoubao: My bad, let me fix it.

LuQQiu:
@wojiaodoubao Wonderful feature! Thanks for adding it!
Do you know whether this fts UDTF is supported in table.count_rows(predicate <-- here) & table.search().where(predicate <-- here)?
or is it only supported in Flight SQL?

wojiaodoubao:
> @wojiaodoubao Wonderful feature! Thanks for adding it! Do you know whether this fts UDTF is supported in table.count_rows(predicate <-- here) & table.search().where(predicate <-- here)? or is it only supported in Flight SQL?

Hi @LuQQiu , nice to meet you~.  Currently it is only supported in Flight SQL. If we want count_rows of fts result, I think we need something like this.
```
    /// let dataset = Dataset::open(uri).await.unwrap();
    /// dataset.scan()
    ///    .full_text_search("col", "query").unwrap()
    ///    .count_rows();
```

LuQQiu:
@wojiaodoubao Thanks! Got it. looks like other python API already support those functionalities!

PR #4685, opened 2025-09-10
feat(java): expose merge_insert api

fangbo:
Related Issue https://github.com/lancedb/lance/issues/4050


fangbo:
@jackye1995 @majin1102 This PR is ready. Could you please review it. Thank you.

java/lance-jni/src/merge_insert.rs
```diff
@@ -0,0 +1,255 @@
...
+    MergeInsertBuilder, MergeStats, WhenMatched, WhenNotMatched, WhenNotMatchedBySource,
+};
+use lance_datafusion::planner::LanceContextProvider;
+use std::sync::Arc;
+use std::time::Duration;
+
+#[no_mangle]
+pub extern "system" fn Java_com_lancedb_lance_Dataset_nativeMergeInsert<'a>(
+    mut env: JNIEnv<'a>,
+    jdataset: JObject,
```
jackye1995: nit: we should always add comment about the exact type for any `JObject`

java/src/main/java/com/lancedb/lance/merge/MergeInsert.java
```diff
@@ -0,0 +1,268 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.lancedb.lance.merge;
+
+import java.util.List;
+
+public class MergeInsert {
```
jackye1995: I think this should better be called `MergeInsertParams`

java/lance-jni/src/merge_insert.rs
```diff
@@ -0,0 +1,255 @@
...
+        "Delete" => Ok(WhenNotMatchedBySource::Delete),
+        "DeleteIf" => {
+            let sql_expr = DFParserBuilder::new(when_not_matched_by_source_delete_expr)
+                .build()
+                .unwrap()
+                .parser
+                .parse_expr()
+                .unwrap();
+
+            let expr = SqlToRel::new(&LanceContextProvider::default())
```
jackye1995: I was thinking what's the best way to do this, this is what I am thinking:

1. in MergeInsertParams in Java, we should expose not just the String SQL expression API, but also a ByteBuffer API to accept Substrait expression, just like what we do in ScanOptions. So for example `withMatchedUpdateIf(String sqlExpr)` and `withMatchedUpdateIf(ByteBuffer substraitExpr)`
3. in JNI, we should just be able to call `LanceFilter::to_datafusion` for both cases

java/src/main/java/com/lancedb/lance/merge/MergeInsertStats.java
```diff
@@ -0,0 +1,80 @@
...
+  public long bytesWritten() {
+    return bytesWritten;
+  }
+
+  public long numFilesWritten() {
+    return numFilesWritten;
+  }
+
+  @Override
+  public String toString() {
```
majin1102, line 66: Please use guava MoreObjects.toStringHelper() for toString()

java/src/test/java/com/lancedb/lance/operation/MergeInsertTest.java
```diff
@@ -0,0 +1,244 @@
...
+
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Path;
+import java.util.Arrays;
+import java.util.List;
+import java.util.TreeMap;
+
+public class MergeInsertTest extends OperationTestBase {
```
majin1102: I'm not sure if this testcase should be put under operation package.
1. This is a data manipulate unlike other metadata operations.
2. This is more like a testcase of dataset.
I think put it under DatasetTest or under the same package would be more reasonable. What do you think

java/src/main/java/com/lancedb/lance/Dataset.java
```diff
@@ -961,6 +963,37 @@ public SqlQuery sql(String sql) {
...
+      MergeInsertResult result = nativeMergeInsert(mergeInsert, source.memoryAddress());
+
+      Dataset newDataset = result.dataset();
+      if (selfManagedAllocator) {
+        newDataset.allocator = new RootAllocator(Long.MAX_VALUE);
+      } else {
+        newDataset.allocator = allocator;
+      }
+
+      return new MergeInsertResult(newDataset, result.stats());
```
majin1102: Just a little curious.
Why return a new MergeInsertResult since the old one has got the right allocator.

java/lance-jni/src/merge_insert.rs
```diff
@@ -0,0 +1,255 @@
...
+    when_not_matched_by_source: &str,
+    when_not_matched_by_source_delete_expr: &str,
+) -> Result<WhenNotMatchedBySource> {
+    match when_not_matched_by_source {
+        "Keep" => Ok(WhenNotMatchedBySource::Keep),
+        "Delete" => Ok(WhenNotMatchedBySource::Delete),
+        "DeleteIf" => {
+            let sql_expr = DFParserBuilder::new(when_not_matched_by_source_delete_expr)
+                .build()
+                .unwrap()
```
majin1102: I don't know if we can use ? instead of unwrap() here. IMO unwrap() should only be used in testcases

fangbo:
@jackye1995 @majin1102 Greatly appreciate your suggestions. I have made some modifications according to your comments. Could you please review it again? Thank you.

rust/lance-datafusion/src/planner.rs
```diff
@@ -152,7 +152,7 @@ impl ScalarUDFImpl for CastListF16Udf {
 }
 
 // Adapter that instructs datafusion how lance expects expressions to be interpreted
-struct LanceContextProvider {
+pub struct LanceContextProvider {
```
jackye1995: this change is no longer needed?

fangbo: > this change is no longer needed?

Thanks, it's my mistake. Already reverted it.

PR #4687, opened 2025-09-10
fix(rust): fix duplicated source rows when merge insert

### LINKED ISSUE #4535: Error in merge insert when we detect non-deterministic merge
When doing merge insert, if we find multiple source rows that match to the same target row, we should consider the update ambiguous and therefore error. This will help prevent users from accidentally duplicating rows with merge insert.

This is default behavior for other systems like Databricks [^1] and Snowflake [^2].

[^1]: https://docs.databricks.com/aws/en/sql/language-manual/delta-merge-into#:~:text=MERGE%20operations%20fail%20with%20a%20DELTA_MULTIPLE_SOURCE_ROW_MATCHING_TARGET_ROW_IN_MERGE
[^2]: https://docs.snowflake.com/en/sql-reference/sql/merge#nondeterministic-results-for-update-and-delete

## Related

* https://github.com/lancedb/lance/issues/4087

yanghua:
close #4535

### DIFF
--- a/rust/lance/src/dataset/write/merge_insert.rs
+++ b/rust/lance/src/dataset/write/merge_insert.rs
@@ -44,6 +44,7 @@ use arrow_array::{
 };
 use arrow_schema::{DataType, Field, Schema};
 use datafusion::common::NullEquality;
+use datafusion::error::DataFusionError;
 use datafusion::{
     execution::{
         context::{SessionConfig, SessionContext},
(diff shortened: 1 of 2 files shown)


rust/lance/src/dataset/write/merge_insert.rs
```diff
@@ -1866,6 +1869,21 @@ impl Merger {
...
+
+                let mut processed_row_ids = self.processed_row_ids.lock().unwrap();
+                for &row_id in row_ids.values() {
+                    if !processed_row_ids.insert(row_id) {
+                        return Err(datafusion::error::DataFusionError::Execution(
+                            format!(
+                                "Ambiguous merge insert: multiple source rows match the same target row with _rowid {}. \
+                                This could lead to data corruption. Please ensure each target row is matched by at most one source row.",
+                                row_id
+                            )
```
wjones127: I think the user would find it very valuable if we could provide the `on` column value here that matched multiple source rows. I *think* we should have enough context for that, right?

rust/lance/src/dataset/write/merge_insert/exec/write.rs
```diff
@@ -49,6 +50,66 @@ struct MergeState {
...
     /// Whether the dataset uses stable row ids.
     stable_row_ids: bool,
+    /// Set to track processed row IDs to detect duplicates
+    processed_row_ids: HashSet<u64>,
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use arrow_array::UInt64Array;
```
wjones127: Why is the test up here and not at the bottom of the file?

rust/lance/src/dataset/write/merge_insert.rs
yanghua: Sounds reasonable.

rust/lance/src/dataset/write/merge_insert/exec/write.rs
yanghua: Oh... that's because I thought I only changed `MergeState`, so added the `mod tests` there. But yes, move to the bottom is better.

PR #4689, opened 2025-09-10
feat: implement lazy loading and prewarm for bitmap index

LuQQiu:
## Summary

This PR converts the bitmap index from eager loading (all bitmaps loaded into memory during index load) to lazy loading (bitmaps loaded on-demand and cached). This significantly reduces memory usage for large bitmap indices while maintaining performance through caching.

## Motivation

The current bitmap index implementation loads all bitmaps into memory during index load. For datasets with high cardinality columns, this can consume significant memory. This change implements lazy loading similar to BTree and FTS indices, loading bitmaps only when needed.

### 1. Lazy Loading Implementation
- Changed `BitmapIndex` structure to store `value_locations` (metadata) instead of full bitmaps
- Bitmaps are now loaded on-demand via `load_bitmap()` method
- Null bitmap is kept in memory as it's frequently accessed

### 2. Cache Integration
- Integrated with Lance's consolidated `LanceCache` system
- Each bitmap is cached with a unique key (`BitmapEntryKey`)
- Cache keys are automatically prefixed with dataset URI and index UUID to prevent collisions

### 3. Prewarm Support
- Implemented `prewarm()` method that reads the index file in chunks
- Loads all bitmaps into cache for predictable performance
- Idempotent - can be called multiple times safely

### 4. Remapping Support
- FragReuseIndex remapping is properly applied during both:
  - Initial load (for null bitmap)
  - Lazy load (for each bitmap loaded on-demand)
- Fixed and enabled the `test_remap_bitmap_with_null` test

## Performance Characteristics

| Operation | Before (Eager) | After (Lazy) |
|-----------|---------------|--------------|
| Index Load | O(n) - Load all bitmaps | O(n) - Scan keys only |
| Memory Usage | All bitmaps in RAM | Only cached bitmaps |
| First Query | Immediate | One I/O per unique value |
| Subsequent Query | Immediate | Cache hit (immediate) |
| Prewarm | N/A | Load all bitmaps in chunks |

## Testing

- Added comprehensive test suite for lazy loading functionality
- Added tests for prewarm with cache state tracking
- Fixed and enabled the previously broken remap test
- All existing bitmap tests continue to pass

## Backward Compatibility

This change maintains full backward compatibility:
- The index file format is unchanged
- Existing indices work without modification
- The public API remains the same

## Future Work

- Consider implementing batch loading for range queries
- Optimize metadata storage to avoid key scanning during load
- Add metrics for cache hit/miss rates

Co-Authored-By: Claude <noreply@anthropic.com>


LuQQiu:
https://github.com/lancedb/lance/pull/4699

PR #4695, opened 2025-09-11
fix(rust): fix rechunk sequences bug and refactor for better readability

yanghua:


yanghua:
Hi @wjones127 Please take a look when you have time.

rust/lance-table/src/rowids.rs
```diff
@@ -586,49 +586,64 @@ pub fn rechunk_sequences(
         .flat_map(|sequence| sequence.0.into_iter())
         .peekable();
 
+    let too_few_segments_error = || {
```
jackye1995: is there any way we can provide more info in the error message? like the chunk size

PR #4696, opened 2025-09-11
chore: remove function build_inverted_index

wojiaodoubao:
We should consolidate the logic related to building scalar indexes into the ScalarIndexPlugin. Use build_scalar_index rather than using build_inverted_index method.

### DIFF
--- a/rust/lance-index/src/scalar/btree.rs
+++ b/rust/lance-index/src/scalar/btree.rs
@@ -1534,8 +1534,15 @@ impl ScalarIndexPlugin for BTreeIndexPlugin {
         data: SendableRecordBatchStream,
         index_store: &dyn IndexStore,
         request: Box<dyn TrainingRequest>,
-        _fragment_ids: Option<Vec<u32>>,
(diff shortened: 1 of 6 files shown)


wojiaodoubao:
Hi @jackye1995 , could you help review this pr when you have time, thanks very much!

jackye1995:
will take a look tomorrow, have some work to finish right now

rust/lance-index/src/scalar/btree.rs
```diff
@@ -1534,6 +1534,7 @@ impl ScalarIndexPlugin for BTreeIndexPlugin {
         data: SendableRecordBatchStream,
         index_store: &dyn IndexStore,
         request: Box<dyn TrainingRequest>,
+        _fragment_ids: Option<Vec<u32>>,
```
jackye1995: I think for the other indexes, instead of just ignoring, we should fail with not supported if this is set

jackye1995 (commented):
overall I like the idea! some minor comments

PR #4697, opened 2025-09-11
feat: add traces for file level IO

jaystarshot:
A trace per file access will be very useful to analyze the access pattern (_indices vs _manifest vs _data)

### DIFF
--- a/rust/lance-io/src/scheduler.rs
+++ b/rust/lance-io/src/scheduler.rs
@@ -473,7 +473,7 @@ impl IoTask {
 
     async fn run(self) {
         let file_path = self.reader.path().to_string();
-        let num_bytes = self.num_bytes();    
+        let num_bytes = self.num_bytes();
         let bytes = if self.to_read.start == self.to_read.end {
             Ok(Bytes::new())
         } else {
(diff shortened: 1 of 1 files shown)


rust/lance-io/src/scheduler.rs
```diff
@@ -486,7 +488,15 @@ impl IoTask {
...
-        };
+        };    
+        // Emit per-file I/O trace event only when tracing is enabled    
+        tracing::trace!(
+            "File I/O completed (file={} bytes_read={} requests=1 range={}..{})",
+            file_path,
+            num_bytes,
+            self.to_read.start,
+            self.to_read.end
+        );
```
wjones127: Don't format trace events as strings like this. That makes it much harder to parse later. Instead use fields:

```suggestion
        tracing::trace!(
            file = file_path,
            bytes_read = bytes_read,
            requests = 1,
            range_start = self.to_read.start,
            range_end = self.to_read.end,
            "File I/O completed"
        );        
```

jaystarshot: Thanks! Didn't know this

PR #4699, opened 2025-09-11
feat: optimize bitmap index with lazy loading and column projection

LuQQiu:
Fixes the problem of loading bitmap index when creating it. The whole bitmap index was loaded during first search or first get index statistics. Get index statistics should not load and cache the bitmaps.

This PR converts the bitmap index from eager loading (all bitmaps loaded into memory during index load) to lazy loading (bitmaps loaded on-demand and cached). This significantly reduces memory usage for large bitmap indices while maintaining performance through caching.

The implementation adds lazy loading support where bitmaps are loaded on-demand via load_bitmap(), integrates with Lance's consolidated LanceCache system, and implements a prewarm() method that reads the index file in chunks to preload all bitmaps when predictable performance is needed. The null bitmap is kept in memory as it's frequently accessed, and FragReuseIndex remapping is properly applied during both initial load and lazy load operations.


rust/lance-index/src/scalar/bitmap.rs
```diff
@@ -39,117 +42,186 @@ use crate::{
...
                 } else {
-                    index_map.insert(key, bitmap);
+                    index_map.insert(key, global_row_idx);
                 }
+
+                global_row_idx += 1;
+            }
+        }
+
+        if let Some(null_loc) = null_location {
```
LuQQiu, line 203: may be null can also in cache?

rust/lance-index/src/scalar/bitmap.rs
```diff
@@ -39,117 +42,186 @@ use crate::{
...
 /// A scalar index that stores a bitmap for each possible value
 ///
 /// This index works best for low-cardinality columns, where the number of unique values is small.
 /// The bitmap stores a list of row ids where the value is present.
 #[derive(Clone, Debug)]
 pub struct BitmapIndex {
-    index_map: BTreeMap<OrderableScalarValue, RowIdTreeMap>,
-    // We put null in its own map to avoid it matching range queries (arrow-rs considers null to come before minval)
+    /// Maps each unique value to its bitmap location in the index file
+    /// The usize value is the global row index in the bitmap_page_lookup.lance file
```
westonpace:
```suggestion
    /// The usize value is the row offset in the bitmap_page_lookup.lance file
```

Minor nit: trying to use offset when relevant as it can be more clear than index

rust/lance-index/src/scalar/bitmap.rs
```diff
@@ -39,117 +42,186 @@ use crate::{
...
             )));
         }
 
-        let mut index_map: BTreeMap<OrderableScalarValue, RowIdTreeMap> = BTreeMap::new();
+        let mut index_map: BTreeMap<OrderableScalarValue, usize> = BTreeMap::new();
         let mut null_map = RowIdTreeMap::default();
         let mut value_type: Option<DataType> = None;
-        let mut index_map_size_bytes = 0;
+        let mut null_location: Option<usize> = None;
+        let mut global_row_idx = 0;
```
westonpace: Minor nit: `global_row_idx -> row_offset`?

rust/lance-index/src/scalar/bitmap.rs
westonpace: It could, but it probably won't be _that_ large either.  I think I could go either way on this.  I don't see any particular reason to eagerly load it like we are doing here but I don't see any strong reason not to.

rust/lance-index/src/scalar/bitmap.rs
```diff
@@ -158,20 +230,61 @@ impl BitmapIndex {
...
-            index_map_size_bytes,
             store,
+            WeakLanceCache::from(index_cache),
+            frag_reuse_index,
         )))
     }
+
+    async fn load_bitmap(&self, key: &OrderableScalarValue) -> Result<Arc<RowIdTreeMap>> {
+        if key.0.is_null() {
+            return Ok(Arc::new(self.null_map.clone()));
```
westonpace: Maybe store `null_map` as `Arc<RowIdTreeMap>` so this clone isn't expensive?

rust/lance-index/src/scalar/bitmap.rs
```diff
@@ -158,20 +230,61 @@ impl BitmapIndex {
...
+        }
+
+        let row_offset = match self.index_map.get(key) {
+            Some(loc) => *loc,
+            None => return Ok(Arc::new(RowIdTreeMap::default())),
+        };
+
+        let page_lookup_file = self.lazy_reader.get().await?;
+        let batch = page_lookup_file
+            .read_range(row_offset..row_offset + 1, None)
```
westonpace: Minor nit: should we only load the values column here?

rust/lance-index/src/scalar/bitmap.rs
```diff
@@ -158,20 +230,61 @@ impl BitmapIndex {
...
+            return Ok(Arc::new(self.null_map.clone()));
+        }
+
+        let cache_key = BitmapKey { value: key.clone() };
+
+        if let Some(cached) = self.index_cache.get_with_key(&cache_key).await {
+            return Ok(cached);
+        }
+
+        let row_offset = match self.index_map.get(key) {
```
westonpace, line 259: If we get to this point can we record a part load in the metrics collector?

rust/lance-index/src/scalar/bitmap.rs
```diff
@@ -259,14 +415,20 @@ impl ScalarIndex for BitmapIndex {
...
+                metrics.record_comparisons(offsets.len());
 
-                metrics.record_comparisons(maps.len());
-                RowIdTreeMap::union_all(&maps)
+                let mut union_bitmap = RowIdTreeMap::default();
+                for (key, _offset) in offsets {
+                    let bitmap = self.load_bitmap(&key).await?;
+                    union_bitmap |= (*bitmap).clone();
+                }
+                union_bitmap
```
westonpace: I believe `RowIdTreeMap::union_all` is more efficient than a series of `|=` operations.  Is it possible to still use that method?  Maybe we can collect all matching bitmaps and then use `union_all`?

rust/lance-index/src/scalar/bitmap.rs
```diff
@@ -349,16 +513,18 @@ impl ScalarIndex for BitmapIndex {
...
         dest_store: &dyn IndexStore,
     ) -> Result<CreatedIndex> {
-        let mut state = self
-            .index_map
-            .iter()
-            .map(|(key, bitmap)| (key.0.clone(), bitmap.clone()))
-            .collect::<HashMap<_, _>>();
+        let mut state = HashMap::new();
+
+        for key in self.index_map.keys() {
```
westonpace, line 548:
```suggestion
        // Initialize builder with existing data
        for key in self.index_map.keys() {
```

rust/lance-index/src/scalar/bitmap.rs
```diff
@@ -414,7 +580,6 @@ impl BitmapIndexPlugin {
             bitmap.serialize_into(&mut bytes).unwrap();
             let bitmap_size = bytes.len();
 
-            // If adding this bitmap would overflow, flush current batch
             if cur_bytes + bitmap_size > MAX_BITMAP_ARRAY_LENGTH {
```
westonpace, line 613: Why remove these comments?

LuQQiu: added back

rust/lance-index/src/scalar/bitmap.rs
LuQQiu: changed back to union_all

PR #4705, opened 2025-09-12
feat: add sdk version compatibility checking

### LINKED ISSUE #4704: Check SDK compatibility for read sdk and manifest writer version
The manifest has recorded the `WriterVersion`, like this:

```
writer_version: Some(
        WriterVersion {
            library: "lance",
            version: "0.31.2",
        },
),
```

I would like to suggest checking the SDK compatibility. There are two cases I can share:

* We met an issue with our customer(there are different teams writing and reading Lance data independently). Once, the writing team used a higher version, but the reading team used a lower version, which caused an error due to #4028 . 

* Another case, a customer's Python version is less than 3.9, and they ran `pip install pylance`, it would install `0.10.x`. But it is not easy to find. 

We can do two things:

* Check the reader SDK, if its version is less than the writer version recorded in the reading manifest. If yes, print a warning log.
* We could introduce a compatibility matrix if we allow a looser reader/writer compose strategy in the future. Considering an upgrade to the reader SDK is not a casual decision(it may bring new issues or bugs).

yanghua:
closes #4704

### DIFF
--- a/rust/lance-table/src/format/manifest.rs
+++ b/rust/lance-table/src/format/manifest.rs
@@ -537,6 +537,46 @@ impl WriterVersion {
             version: new_version,
         }
     }
+
+    /// Check if the current SDK version is older than the writer version
+    /// and log a warning if so. This helps users identify potential compatibility issues.
+    pub fn check_sdk_compatibility(&self) {
+        let current_sdk_version = env!("CARGO_PKG_VERSION");
+        let current_version = Self {
+            library: "lance".to_string(),
+            version: current_sdk_version.to_string(),
+        };
+
+        // Only check compatibility for lance library
+        if self.library == "lance" {
+            if let (Some(writer_parts), Some(current_parts)) =
+                (self.semver(), current_version.semver())
+            {
+                // Compare versions: if current SDK version < writer version, warn
+                if (current_parts.0, current_parts.1, current_parts.2)
+                    < (writer_parts.0, writer_parts.1, writer_parts.2)
+                {
+                    log::warn!(
+                        "Dataset was written with Lance SDK version {} but current SDK version is {}. \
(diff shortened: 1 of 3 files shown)


rust/lance-table/src/format/manifest.rs
```diff
@@ -537,6 +537,46 @@ impl WriterVersion {
...
+                (self.semver(), current_version.semver())
+            {
+                // Compare versions: if current SDK version < writer version, warn
+                if (current_parts.0, current_parts.1, current_parts.2)
+                    < (writer_parts.0, writer_parts.1, writer_parts.2)
+                {
+                    log::warn!(
+                        "Dataset was written with Lance SDK version {} but current SDK version is {}. \
+                         Some features may not be available or data may not be readable correctly. \
+                         Consider upgrading your Lance SDK to version {} or later.",
```
wjones127, line 562: We shouldn't be making breaking changes like this. We already have a mechanism for enforcement of breaking features, which are the feature flags:

https://github.com/lancedb/lance/blob/25be97b7e25f050b604b24738beb24d8db6951b9/protos/table.proto#L79-L100

yanghua:
Hi @wjones127 Did you see the two cases that I have mentioned [here](https://github.com/lancedb/lance/issues/4704#issue-3408626444)? They are the real cases we have met.

> We already have a mechanism for enforcement of breaking features

I think not all the break changes are tracked by the feature flag. For example, if a PR has [this change](https://github.com/lancedb/lance/pull/4029#pullrequestreview-2946159997) in the writer SDK, but not in the reader SDK(that means the reader's version is less than the writer's version). All the decoder behavior would fail.

wjones127:
See my reply here: https://github.com/lancedb/lance/issues/4704#issuecomment-3304910171

PR #4706, opened 2025-09-12
feat(java): supports do compaction distributedly

steFaiz:
This PR is about to introduce the distributed compaction mechanism in rust to java, which is important for bigdata ecosystem such as Flink as Spark.


steFaiz:
@jackye1995 Hi, PTAL if you have some time!
I notice that there's an existing PR to introduce compaction to the [Dataset level](https://github.com/lancedb/lance/pull/4703).
However, I think that these two PRs do not conflict with each other too much (except the CompactionOptions class). Please tell me which PR you prefer to merge first, and Im pleased to modify my PR continuously!

jackye1995:
sounds good, I just merged the other PR, could you rebase on top of it?

steFaiz:
> sounds good, I just merged the other PR, could you rebase on top of it?

@jackye1995 
I've rebased my code! Note that I follow the code routine as WriteParams, removing those default values in java side. I think that the rust-side default values should be the ground-truth, to avoid any potential conflicts.

java/src/main/java/com/lancedb/lance/compaction/CompactionMetrics.java
```diff
@@ -0,0 +1,60 @@
...
+
+  public long getFilesRemoved() {
+    return filesRemoved;
+  }
+
+  public long getFragmentsAdded() {
+    return fragmentsAdded;
+  }
+
+  public long getFragmentsRemoved() {
```
majin1102, line 47: I suggest we make it clear if we want more rust style(eliminate get) or use this more java way. They have quite seperated looks @jackye1995

steFaiz: @majin1102 
Hi, these `get` methods here provide the approach to expose these metrics to java caller e.g. Flink TaskManager.
I think these Java classes can exhibit a clear Java style (being friendly to Java programmers), as long as it doesn't affect the invocation of the underlying Rust code.  Can you tell me which class in current Java module is in 'Rust Style'? Im unfamiliar about that, Thank!

majin1102: I was saying the getxxx() or direct xxx naming issue.

There was a short discussion that should use which style. Any way this is not a blocking thing

java/src/main/java/com/lancedb/lance/compaction/IndexRemapperType.java
```diff
@@ -0,0 +1,22 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.lancedb.lance.compaction;
+
+/** The Index remapper type. */
+public enum IndexRemapperType {
```
jackye1995: I think we can just ignore the index remapper type in java binding. It's always DatasetIndexRemapper, Ignore is only used for testing but we won't really do equivalent testing in java.

java/src/main/java/com/lancedb/lance/compaction/CompactionOptions.java
```diff
@@ -0,0 +1,169 @@
...
+import com.google.common.base.MoreObjects;
+
+import java.io.Serializable;
+import java.util.Optional;
+
+/**
+ * Compaction options. All fields are optional in Java side. We do not set default value for them to
+ * avoid conflicting with those default value in Rust side.
+ */
+public class CompactionOptions implements Serializable {
```
jackye1995, line 27: yeah we have the same discussions a few times in python binding. If everything is pushed to rust then it's pretty hard for people to figure out what is the actual default. I think so far we have not set a norm yet in Java. If we want to go with this, we should at least add the link to find the default  in rust - https://docs.rs/lance/latest/src/lance/dataset/optimize.rs.html#118

java/src/main/java/com/lancedb/lance/compaction/IndexRemapperType.java
steFaiz: Done removing this enum class.

java/src/main/java/com/lancedb/lance/compaction/CompactionOptions.java
steFaiz: @jackye1995   Thanks for your suggestion! I've added a link to the rust-side default values in class comments.

steFaiz:
@jackye1995 
Just a friendly ping on this PR whenever you have a moment. Please let me know if you have any questions or if there's anything I can do to help with the review. This PR is a little large, but most of the content are just java classes mirroring rust structs as well as conversion between them. 

Thanks!

majin1102:
> looks good to me. @majin1102 any further comments?

No, thanks for this contribution @steFaiz 
Also useful in our product environments.

PR #4710, opened 2025-09-12
feat(rust)!: support branch based on shallow_clone

majin1102:
After some messy rebases that caused issues, I created this fresh PR for a clean slate @jackye1995 .

Design doc: https://github.com/lancedb/lance/discussions/4256#discussioncomment-14304177(completed feature for that part.)


jackye1995:
I have been playing around with this locally to understand it better, since this is a bit different from what I was imagining this to be.

Let me spell out my original thinking first:

1. create `BranchContents` definition, basically same as you described:

```rust
pub struct BranchContents {
    pub parent_branch: Option<String>,
    pub parent_version: u64,
    pub create_at: u64, // unix timestamp
    pub manifest_size: usize,
}
```

2. create `Branches` to manage branches, which have mostly similar functionalities as yours for `list`, `get`, `delete`, `create`, etc., but what I am thinking differently is that we just have a `checkout_branch` also here, so that:

```rust
impl Branches {

    pub async fn checkout_branch(branch_name: &str) -> Result<Dataset> {
      ...
    }
}
```

Basically we treat a branch at dataset location `dataset_location/tree/<branch_name>` as a new dataset location and directly load the dataset, instead of touching any existing checkout and commit behaviors of a dataset. In other words, for any given dataset, we have just exposed the ability to (1) examine what are the direct branches and manage their metadata definition, (2) load a specific branch as a new dataset. We have not really changed any existing functionalities.

I feel this is much more backwards compatible and a much simpler approach. Although we cannot do things like completely switching the current dataset object to point to a branch, but I feel that is probably better because it would totally mess up with things like caching.

What do you think? I read about the detailed design in https://github.com/lancedb/lance/discussions/4256#discussioncomment-14304177, and I think in this approach we will miss the ability to do globally unique tag, but that seems fine to me, as we have defined the "tag" to be local, and we can build global tag on top of it by making a "<branch_name>:<local_tag_name>" syntax. But I don't know if I am missing anything important in your design that I have overlooked.

majin1102:
Try to understand the difference you meant.

> Basically we treat a branch at dataset location dataset_location/tree/<branch_name> as a new dataset location and directly load the dataset, instead of touching any existing checkout and commit behaviors of a dataset

I think current codebase doesn't touch any existing checkout and commit behaviors of a dataset as well (If I understand right about the touching meaning).

I put the checkout_branch into Dataset so that we can use a root directory and branch name to checkout a branch latest version.  This procedure doesn't need to touch commit behaviors of the current or other datasets. If we move checkout_branches into Branches I don't see any difference case both ways return a new dataset. However if we don't implement checkout back and forth across branches, I think that could actually simplify the codebase by eliminating DatasetLocation. The trade off includes:

1. I'm now in Branch A, we can **NOT** use A.branches.checkout_branch(B). We only allow users use root.branches.checkout_branch(B) or load B from xxxxx/tree/B.
2. Global tag as you said.
3. I'm now in Branch A, I actually don't know any other branches or tags if we don't have a DatasetLocation to trace the root location. On the contrary, if we have DatasetLocation, we don't necessarily need introducing the trade offs. This I think is somehow related to the read authorization of a branch. I mean if I was in Branch A, I should read the whole branch list since A should not be allowed to have its own branch list (implemented in this codebase).

So for this part, IMO it's about the necessarily of DatasetLocation. The introduction of DatasetLocation mainly changed the way of getting url and base_path of datasets. I think the risk of this part is under control.

> I feel this is much more backwards compatible and a much simpler approach. Although we cannot do things like completely switching the current dataset object to point to a branch, but I feel that is probably better because it would totally mess up with things like caching.

I think this concern is solid. I didn't think too much about the cache backwards compatible. I agree we'd better leave cache key not changed. I think we could achieve this by using a different metadata cache object(Or let's just use a load new dataset code path)

In conclusion, I think there are two aspects we are talking about:
1. The introduction of DatasetLocation for global reference view in whatever situaions, global tags, checkout back and forth. Or don't use it and limit the global reference view.
2. Make branch and main datasets more independent and eliminate the compatibilty risks

What do you think?

majin1102:
I took a deep look into caches.

Now the manifest cache key has rollbacked to its original state. The key change is when we  check a manifest cache, we first use the uri to get the right dataset metadata cache:

https://github.com/majin1102/lance/blob/1a952b353ad4023fc07da3d999952323dea27c51/rust/lance/src/dataset.rs#L808

Now I think the cache backwards compatibility risk is under control since branches and the main dataset use different caches by uri.

I'm neutral on moving checkout_branch to Branches. Actually I tried and found some blocking point. checkout_branch mainly depends on the function:

```
    fn checkout_manifest(
        object_store: Arc<ObjectStore>,
        dataset_location: DatasetLocation,
        manifest: Arc<Manifest>,
        manifest_location: ManifestLocation,
        session: Arc<Session>,
        commit_handler: Arc<dyn CommitHandler>,
        file_reader_options: Option<FileReaderOptions>,
    ) -> Result<Self> {
    ........
    }
```

If we do this in refs，we need to introduce at least session and file_reader_options into refs.rs. I thought this was feasible but hesitated if we should make refs so rich and we can't ignore we need to add read_manifest logics to refs as well. These functions could be reused in dataset.

jackye1995:
> In conclusion, I think there are two aspects we are talking about:
> The introduction of DatasetLocation for global reference view in whatever situaions, global tags, checkout back and forth. Or don't use it and limit the global reference view.
> Make branch and main datasets more independent and eliminate the compatibilty risks

Yeah agree these are the 2 main points. I am wondering, can we first merge the parts without the DatasetLocation? Like the Branches management part those things.

I am just not really convinced that we need to introduce DatasetLocation, which is quite complicated to achieve the feature of switching branch and having global tags. But I am trying to see if we can move the progress forward a bit while I am thinking this through.

majin1102:
> Yeah agree these are the 2 main points. I am wondering, can we first merge the parts without the DatasetLocation? Like the Branches management part those things.

Sure, I originally planned to introduce DatasetLocation at the second step. It seems I can now adjust the tempo mainly based on DatasetLocation.

The actual essence of DatasetLocation is a util/tool for uri and path constructions of branches. So the usage of DatasetLocation is through each function of branching. Even though we don't use DatasetLocation, we still need the same functions for branching(maybe another name like BranchLocation, depends on you. Before that let me use DatasetLocation for now). So IMO maybe we could control the scope of DatasetLocation instead of elminating it or not.

- **Step1.** Support create/list/delete branches as you proposed. The usage of DatasetLocation is limited to generate a branch tree path. In this step we could load a branch by the branch uri. But won't allow branch operations on a branch as well as global tags.
- **Step2.** Support checkout_branches based on DatasetLocation. DatasetLocation will be used to switch branch, global branches and tags would be introduced here(Yet the DatasetLocation will be limited in branches functions).
- **Step3.** Use DatasetLocation to take place of uri and base in Dataset. This phase will impact the uri and base getting from dataset. We could evaluate the progress carefully.

I believe the seperation of 2 and 3 could help us see the issue more clearly. If feeling not right, we could take a hold at step 2 or 3. To me the difference from 2 to 3 is to only reuse DatasetLocation to encapsulate uri and base, which could impact more files but seems more elegent to me.

How do you feel?

> I am just not really convinced that we need to introduce DatasetLocation, which is quite complicated to achieve the feature of switching branch and having global tags. But I am trying to see if we can move the progress forward a bit while I am thinking this through.

I envision a version management system that operates in a more Git-like style. This is particularly compelling as Lance is being applied in more Python use cases, which inherently favors interactive execution scenarios compared to the Java ecosystem. The vision of simply combining object store and Python(or with a lightweight CLI tool)—to deliver a fully interactive, Git-style version management is quite attractive to me (I think hugging-face has proven the value of the combination of dataset and Git-style version management. Lance might take a furthur step?  lol).

But yes I agree we should avoid over-elaborate and unnecessary risks.

protos/transaction.proto
```diff
@@ -172,13 +172,14 @@ message Transaction {
...
-    // for branch scenario, this is the branch name in the source dataset
-    // for non branch scenario(normal cloning), this is an optional tag name in the source dataset
+    // in most cases it should be the the branch or tag name in the source dataset
     optional string ref_name = 2;
     // the version of the source dataset for cloning
     uint64 ref_version = 3;
     // the absolute base path of the source dataset for cloning
     string ref_path = 4;
+    // if the target dataset is a branch, this is the branch name of the target dataset
+    optional string branch_name = 5;
```
jackye1995, line 182: do we really need this branch name? This is duplicated information from the `optional string branch = 20;` in the manifest.

rust/lance/src/dataset.rs
```diff
@@ -401,10 +411,60 @@ impl Dataset {
         Ok(())
     }
 
-    fn already_checked_out(&self, location: &ManifestLocation) -> bool {
+    /// Check out the latest version of the branch
+    pub async fn checkout_branch(&mut self, branch: &str) -> Result<Self> {
+        self.checkout_by_ref(None, Some(branch.to_string())).await
+    }
+
+    pub async fn create_branch(
```
jackye1995, line 441: Should add some docstring explaining potential orphan branch since it's a 2 step process

rust/lance/src/dataset/scanner.rs
```diff
@@ -5943,7 +5947,7 @@ mod test {
     #[rstest]
     #[tokio::test]
     async fn test_index_take_batch_size() {
-        let fixture = ScalarIndexTestFixture::new(LanceFileVersion::Stable, false).await;
+        let fixture = Box::pin(ScalarIndexTestFixture::new(LanceFileVersion::Stable, false)).await;
```
majin1102, line 5951: Due to clippy issue

rust/lance/src/dataset/builder.rs
```diff
@@ -110,6 +112,13 @@ impl DatasetBuilder {
         self
     }
 
+    /// Sets `version` for the builder using a branch and optional version number
+    /// If version_number is null, checkout the latest version
+    pub fn with_branch(mut self, branch: &str, version_number: Option<u64>) -> Self {
```
majin1102, line 117: We might deside if we really need this option as well as with_tag.

This reference option might cause double loading if the uri is not the branch we want to checkout. But this is not a technical trouble but a design concern. I implemented the double loading below

rust/lance/src/dataset.rs
```diff
@@ -573,7 +654,8 @@ impl Dataset {
             manifest_location,
             commit_handler,
             session,
-            tags,
+            tags: refs.tags(),
```
jackye1995: should we just store a `refs` in dataset instead of separated tags and branches?

majin1102:
This is ready for another review@jackye1995

rust/lance/src/dataset.rs
```diff
@@ -1742,6 +1839,27 @@ impl Dataset {
         builder.execute(transaction).await
     }
 
+    async fn resolve_reference(&self, _ref: refs::Ref) -> Result<(Option<String>, u64)> {
```
jackye1995: nit: should not have underscore in used variable name

rust/lance/src/dataset/branch_location.rs
```diff
@@ -0,0 +1,191 @@
...
+    pub branch: Option<String>,
+}
+
+impl BranchLocation {
+    /// Find the root location
+    pub fn find_root(&self) -> Result<Self> {
+        if let Some(branch_name) = self.branch.as_ref() {
+            let mut path = self.path.clone();
+            let mut uri = self.uri.clone();
+            let segment_count = branch_name.split('/').count();
```
jackye1995: what if I have a branch with name `my_branch/`? Would that cause issue?

rust/lance/src/dataset/branch_location.rs
```diff
@@ -0,0 +1,191 @@
...
+                branch: Some(target_branch.clone()),
+            })
+        } else {
+            Ok(root_location)
+        }
+    }
+
+    fn parent_str(path_str: &str) -> Result<&str> {
+        let trimmed = path_str.trim_end_matches('/');
+        match trimmed.rfind('/') {
```
jackye1995: would this fail on windows?

rust/lance/src/dataset/builder.rs
```diff
@@ -300,31 +309,96 @@ impl DatasetBuilder {
...
-                Ref::Version(v) => Some(v),
-                Ref::Tag(t) => {
-                    let tags = Tags::new(
-                        object_store.clone(),
-                        commit_handler.clone(),
-                        base_path.clone(),
-                    );
-                    Some(tags.get_version(t.as_str()).await?)
+        // Two cases that need to check out after loading the manifest:
+        // 1. If the target is a configured as a branch, we need to check the branch field in the manifest
```
jackye1995: nit: "is configured as a"

rust/lance/src/dataset/builder.rs
```diff
@@ -300,31 +309,96 @@ impl DatasetBuilder {
...
+        // and reload the right branch in case the uri is not the right one.
```
jackye1995, line 323: I think we should not allow publicly load a URI that is not a root table URI, user must specify the dataset using its root URI, and then specify a branch or tag to checkout. That should also avoid the case of "we don't find the tag under the table_uri"

jackye1995 (commented):
Thanks for updating this, I think we are very close!

protos/transaction.proto
majin1102: 1. This helps to know if this clone operation is for branching
2. We need to pass the branch_name to do_commit_new_dataset. I don't figure out a better way if we don't use the operation struct to pass this field.

rust/lance/src/dataset/builder.rs
majin1102: I think it's hard to abandon users to use a branch uri since branch could be considered as an independent dataset.

I wonder if you mean we should not introduce a double loading if we can't find the TagContent directly from the uri? We directly throw an error here?

rust/lance/src/dataset/branch_location.rs
majin1102: why is that?

rust/lance/src/dataset/builder.rs
jackye1995: > I think it's hard to abandon users to use a branch uri since branch could be considered as an independent dataset.

I think that's something we should define as a part of the format - branches are meant to be used as a part of the source dataset, not independently. Otherwise users can just recursively create branches and tags that are within the branch. (this was what I was originally thinking about, but you convinced me that we should just use the git as the target experience). If user have a use case to load the branch by uri directly and use it like a normal dataset, they should just do a shallow clone instead. That seems to be a good way to distinguish these 2 features as well.

rust/lance/src/dataset/branch_location.rs
jackye1995: Do we in general support paths like `C:\Users\Username\Documents\file`? Maybe I am just over thinking.

rust/lance/src/dataset.rs
majin1102: Ok, This is caught in a dilemma​. The original tag interface is like dataset.tag.create...list...delete. If we use refs here I think we shall use dataset.tags(). But somehow it brings backward compatibility.

The second solution is to put branches and tags together into refs and keep original tag interfaces and add branch operations.

But, the original tag operations is like create, list, delete, fetch. This will be strange if we have a create_branch() to create branch and a create() to create tag.

That's why it looks like this now

rust/lance/src/dataset/builder.rs
majin1102: done

rust/lance/src/dataset/branch_location.rs
majin1102: Oh, this is an invalid branch_name:
https://github.com/majin1102/lance/blob/lance-branch2/rust/lance/src/dataset/refs.rs#L627

I don't think invalid branch name would enter here

rust/lance/src/dataset.rs
majin1102: done

rust/lance/src/dataset.rs
jackye1995: > But somehow it brings backward compatibility.
 
what do you mean by that?

rust/lance/src/dataset/branch_location.rs
majin1102: ​​Sorry, I didn't consider this.

The tests seem working fine on windows. I guess Path and Uri have handled the uri string to xxx/xxx/xx pattern.
Let me verify it

rust/lance/src/dataset/builder.rs
majin1102: Let's say we have dataset at 's3://bucket/main', we have branches 'feature/A' and 'feature/B'.

We can checkout branch 'feature/A' by(which is our common sense):
1. load dataset at  's3://bucket/main'
2. checkout_branch or checkout_version 'feature/A'

The other way is:
1. load branch at 's3://bucket/main/feature/A'

This approach will directly get the target branch and we could know this is a branch by the 'branch' field in manifest., which means:
1. There's no difference at format layer with the first loading dataset 
2. When we create branch on this loaded dataset, we will know the root location and create the branch at the right location. In the meanwhile, we could get the global reference view in this loaded dataset.
3. This approach won't need double loading and give user a reasonble response.

 When I said independent, I meant relativey independent on timeline(I think I might need to say 'relatively' afterward). I think the loading approach won't affect the global reference management on datasets.

jackye1995: > The other way is:
> load branch at 's3://bucket/main/feature/A'
> This approach will directly get the target branch and we could know this is a branch by the 'branch' field in manifest.,

oh I see. i guess I misunderstood. as long as we are in the same state if we load main and then checkout, vs we load the branch directly, I think we are good

rust/lance/src/dataset.rs
majin1102: I think I might misunderstand the meaning of backward compatibility.

What I want to say is the interface compatibility. Because we use dataset.tags for tag API, if we change that, the upper modules(java, python at least) or services like lancedb may need to adapt it too. What do you think of this?

jackye1995: I think that's fine. We make breaking changes all the time, this is already a breaking change anyway

rust/lance/src/dataset.rs
majin1102: I added some comments.
And I make delete_branch could delete zombie branch dataset. This may help the case we discussed.

rust/lance/src/dataset/refs.rs
```diff
@@ -216,43 +347,331 @@ impl Tags {
...
+
+    pub async fn delete(&mut self, branch: &str) -> Result<()> {
+        check_valid_branch(branch)?;
+
+        let root_location = self.refs.root()?;
+        let branch_file = branch_contents_path(&root_location.path, branch);
+        self.object_store().delete(&branch_file).await?;
+
+        // Clean up branch directories
+        self.cleanup_branch_directories(branch).await
```
jackye1995, line 486: this should log a warning if failed.

rust/lance/src/dataset/refs.rs
```diff
@@ -216,43 +347,331 @@ impl Tags {
...
+            .chars()
+            .all(|c| c.is_alphanumeric() || c == '.' || c == '-' || c == '_')
+        {
+            return Err(Error::InvalidRef {
+                message: format!("Branch segment '{}' contains invalid characters. Only alphanumeric, '.', '-', '_' are allowed.", segment),
+            });
+        }
+    }
+
+    if branch_name.ends_with(".lock") {
```
jackye1995, line 678: why do we need this restriction?

rust/lance/src/dataset/refs.rs
```diff
@@ -216,43 +347,331 @@ impl Tags {
...
+    }
+}
+
+impl BranchContents {
+    pub async fn from_path(path: &Path, object_store: &ObjectStore) -> Result<Self> {
+        from_path(path, object_store).await
+    }
+}
+
+pub fn check_valid_branch(branch_name: &str) -> Result<()> {
```
jackye1995, line 634: I am debating with myself, if we should just now allow the branch name to be `main`, since main should be the root dataset, it will be confusing to have another main branch.

rust/lance/src/dataset.rs
jackye1995: I think having that only in cleanup is fine. If we want to do it through delete_branch, we can maybe have a `delete(force=True)` for deleting zombie branch, but we should not do that by default because otherwise it would appear that user can delete a branch that does not exist.

jackye1995:
@majin1102 thank you so much for following through this, I have some final questions and comment on the delete branch thing. Once those are fixed I think we are good to go

rust/lance/src/dataset/refs.rs
majin1102: Yes! I thought of this but forgot to handle

rust/lance/src/dataset/refs.rs
majin1102: Oh, this is copied from check_valid_tag. Just for consistency. Do we need to delete it?

rust/lance/src/dataset.rs
majin1102: This is resolved. I think we are good to mark

rust/lance/src/dataset.rs
majin1102: Sorry, missed this comment.
I prefer `delete(force=True)` so that we could have a quick way to fix it. Don't need switch to a OSS tool if we are using Python for example.

majin1102:
> Do we in general support paths like C:\Users\Username\Documents\file? Maybe I am just over thinking.

I took a deep look into this.
First of all, branch name can not have '\\' and this has been ruled by check_valid_branch:

```
    // Validate if there are any dangerous characters in the branch name
    if branch_name.contains("..") || branch_name.contains('\\') {
        return Err(Error::InvalidRef {
            message: "Branch name cannot contain '..' or '\\'".to_string(),
        });
    }
```
Secondly, windows path accept both "C:/Users/Username/Documents/file" or "C:\\Users\\Username\\Documents\\file/tree/branchA". So I think we could just stick to '/' as the seperator.

Yet, we may encounter the situation that using "C:\\Users\\Username\\Documents\\file\\tree\\branchA" to load branch, I added logic to process this case reasonably and improve the tests.

PTAL @jackye1995

majin1102:
> Do we in general support paths like C:\Users\Username\Documents\file? Maybe I am just over thinking.

I took a deep look into this.
First of all, branch name can not have '\' and this has been ruled by check_valid_branch:

```
    // Validate if there are any dangerous characters in the branch name
    if branch_name.contains("..") || branch_name.contains('\\') {
        return Err(Error::InvalidRef {
            message: "Branch name cannot contain '..' or '\\'".to_string(),
        });
    }
```
Secondly, windows path accept both `"C:/Users/Username/Documents/file"` or `"C:\\Users\\Username\\Documents\\file/tree/branchA"`. So I think we could just stick to '/' as the seperator.

Yet, we may encounter the situation that using "C:\\Users\\Username\\Documents\\file\\tree\\branchA" to load branch. This case find_main would fail. So I added logic to process this case reasonably.

PTAL @jackye1995

rust/lance/src/dataset.rs
```diff
@@ -4477,39 +4622,39 @@ mod tests {
...
 
-        assert_eq!(dataset.tags.list().await.unwrap().len(), 0);
+        assert_eq!(dataset.tags().list().await.unwrap().len(), 0);
 
-        let bad_tag_creation = dataset.tags.create("tag1", 3).await;
+        let bad_tag_creation = dataset.tags().create("tag1", 3).await;
         assert_eq!(
             bad_tag_creation.err().unwrap().to_string(),
-            "Version not found error: version 3 does not exist"
+            "Version not found error: version Main::3 does not exist"
```
jackye1995, line 4630: should probably use lowercase `main`

rust/lance/src/dataset/refs.rs
```diff
@@ -216,43 +347,345 @@ impl Tags {
...
+        self.object_store()
+            .put(
+                &branch_file,
+                serde_json::to_string_pretty(&branch_contents)?.as_bytes(),
+            )
+            .await
+            .map(|_| ())
+    }
+
+    pub async fn delete(&mut self, branch: &str, forced: bool) -> Result<()> {
```
jackye1995: should add docstring explain the meaining of force. Also I think typically should use `force` instead of `forced` for the variable name.

rust/lance/src/dataset.rs
```diff
@@ -401,10 +418,80 @@ impl Dataset {
...
+        let transaction = Transaction::new(version_number, clone_op, None, None);
+
+        let builder = CommitBuilder::new(WriteDestination::Uri(branch_location.uri.as_str()))
+            .with_store_params(store_params.unwrap_or_default())
+            .with_object_store(Arc::new(self.object_store().clone()))
+            .with_commit_handler(self.commit_handler.clone())
+            .with_storage_format(self.manifest.data_storage_format.lance_file_version()?);
+        let dataset = builder.execute(transaction).await?;
+
+        // Create BranchContent after shallow_clone
```
jackye1995: in various places across the code, should use `BranchContents` instead of `BranchContent`

rust/lance/src/dataset/refs.rs
```diff
@@ -11,51 +11,138 @@ use object_store::path::Path;
...
+    fn from(_ref: (&str, Option<u64>)) -> Self {
+        Version(Some(_ref.0.to_string()), _ref.1)
+    }
+}
+
+impl fmt::Display for Ref {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        match self {
+            Version(branch, version_number) => {
+                let branch_name = branch.as_deref().unwrap_or("Main");
```
jackye1995: should use lowercase main

rust/lance/src/dataset/refs.rs
```diff
@@ -216,43 +347,345 @@ impl Tags {
...
+        }
+    }
+
+    if branch_name.ends_with(".lock") {
+        return Err(Error::InvalidRef {
+            message: "Branch name cannot end with '.lock'".to_string(),
+        });
     }
+
+    if branch_name.eq_ignore_ascii_case("main") {
```
jackye1995: I think branch is case sensitive, so we don't need to check ignore case

jackye1995 (approved):
final nit comments before merging

jackye1995:
looks like there are some failing tests with windows

majin1102:
> looks like there are some failing tests with windows

Taking a look

jackye1995:
Thanks for all the work!!

rust/lance/src/dataset.rs
```diff
@@ -4477,39 +4622,39 @@ mod tests {
...
-        let bad_tag_creation = dataset.tags.create("tag1", 3).await;
+        let bad_tag_creation = dataset.tags().create("tag1", 3).await;
         assert_eq!(
             bad_tag_creation.err().unwrap().to_string(),
-            "Version not found error: version 3 does not exist"
+            "Version not found error: version Main::3 does not exist"
         );
 
-        let bad_tag_deletion = dataset.tags.delete("tag1").await;
+        let bad_tag_deletion = dataset.tags().delete("tag1").await;
```
LuQQiu, line 4633: This is a BREAKING CHANGE

jackye1995: Yeah I think it's already marked as breaking in the title?

majin1102: https://github.com/lancedb/lance/pull/4710#discussion_r2377872888

Here is the context. We dicussed through this, considering the long-way benifits, we introduced this change. Sorry for the additional work.

Next time we could bring you in earlier.

Issue #4650, opened 2025-09-05
Fragment-Level Merge/Update columns interface for Java


steFaiz:
Distributed column merging is a crucial feature that enables flexible schema evolution for users. In big data engine scenarios, we need to update columns in a distributed manner, where specific concurrent tasks operate only on specific fragments.

This is a complex functionality. At the computation engine level, the most fundamental requirement is the ability to perform merge and update operations on columns at the fragment granularity.

The purpose of this issue is to add column-level merge and update APIs to the Java module. This will serve as the cornerstone for the next implementation of distributed column merging/updating in Spark and Flink.

This issue mainly contains two interfaces:
1. MergeColumn interface: The jave side `fragment#merge_columns` method. This method will merge new columns into target fragment. A new fragment with the same FragId as well as the merged Schema will be returned.
2. UpdateColumn interface:
In a distributed scenario, modifying the schema for each fragment individually is a risky operation. For this reason, we introduce an update column interface, which is semantically similar to Paimon's PartialUpdate mechanism. It builds upon the existing merge_column logic with the following key characteristics:
  a. The new batch updates an existing column rather than adding a new one.
  b. A join is still performed based on the values of a specified key column. For each matched row, the new value is determined by comparing the incoming value (`value_new`) with the current value (`value_old`):
     1. If `value_new` is present and not null, it is used as the updated value.
    2. If `value_new` is not present or null, but value_old is not null, the original `value_old` is kept.
    3. Otherwise (if both are null), the new value is null.

The UpdateColumn interface is based on the observation that in big data scenarios, users rarely add a complete set of new columns to a table all at once. Instead, they tend to add columns incrementally over the course of multiple jobs, as illustrated in the diagram below:

<img width="922" height="614" alt="Image" src="https://github.com/user-attachments/assets/e2923b32-8c7b-40df-b0d4-9ca6fcae08c5" />

Without UpdateColumn interface, the result might be:

<img width="706" height="882" alt="Image" src="https://github.com/user-attachments/assets/1542ce03-7270-41d1-b13a-441bf8665335" />

While with UpdateColumn, we can get the ideal result:

<img width="748" height="784" alt="Image" src="https://github.com/user-attachments/assets/f0eaa2e5-c751-4740-a1a7-47e5264fc30b" />

Xuanwo:
[...]

PR #4715, opened 2025-09-12
feat: support fragment-level update columns

wayneli-vt:
As described in https://github.com/lancedb/lance/issues/4650, this PR implements a fragment-level column update operation in Rust and exposes it through the Java API. This allows for efficient, partial updates of a dataset without copying irrelevant fields.

The update mechanism follows these rules:
*  **Update existing rows**: For any row identified by a given key in the target table, if a corresponding update value is present in the source data, the target row's value is replaced. This applies even if the new value is null.
*  **Preserve untouched rows**: If a row in the target table has no corresponding update in the source data, it remains unchanged.

The image below illustrates this process:
image

**Important:** It is the responsibility of the calling engine (e.g., Spark, Flink) to ensure that the update data (the source) is correctly partitioned and routed to the corresponding fragments of the target table. The core operation assumes it receives the correct data for the fragment it is operating on.


steFaiz:
@westonpace @jackye1995 
PTAL if you have some time! We introduce the fragment-level update interface just like the `merge_insert` of Dataset, except that deletion is not considered here.

java/src/main/java/com/lancedb/lance/Constants.java
```diff
@@ -0,0 +1,21 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.lancedb.lance;
+
+public class Constants {
```
jackye1995: I think we can call this MetadataColumns

java/src/test/java/com/lancedb/lance/operation/UpdateTest.java
```diff
@@ -90,4 +103,125 @@ void testUpdate(@TempDir Path tempDir) throws Exception {
...
     }
   }
+
+  @Test
+  void testUpdateColumns(@TempDir Path tempDir) throws Exception {
+    String datasetPath = tempDir.resolve("testUpdateColumns").toString();
+    try (RootAllocator allocator = new RootAllocator(Long.MAX_VALUE)) {
+      TestUtils.UpdateColumnTestDataset testDataset =
+          new TestUtils.UpdateColumnTestDataset(allocator, datasetPath);
+      WriteParams writeParams = new WriteParams.Builder().withEnableStableRowIds(true).build();
```
jackye1995: I don't think you need to enable stable row ID for your test?

java/src/test/java/com/lancedb/lance/operation/UpdateTest.java
```diff
@@ -90,4 +103,125 @@ void testUpdate(@TempDir Path tempDir) throws Exception {
...
+       *   0:   |    0   |  "Update 0"  |
+       *   1:   |    2   |  "Update 1"  |
+       *   2:   |    4   |  "Update 2"  |
+       *   3:   |    6   |  "Update 3"  |
+       *   4:   |    8   |  "Update 4"  |
+       *   5:   |   10   |  "Update 5"  |
+       *   6:   |   12   |  "Update 6"  |
+       *   7:   |   14   |  "Update 7"  |
+       *   8:   |   16   |  "Update 8"  |
+       *   9:   |   18   |  "Update 9"  |
```
jackye1995: this only tests about the value override case. We should also cover testing setting some columns to null, and also some null columns to value.

rust/lance/src/dataset/hash_joiner.rs
```diff
@@ -206,6 +206,88 @@ impl HashJoiner {
...
+                                location!(),
+                            ))
+                    })
+                    .await;
+                    match task_result {
+                        Ok(Ok(array)) => {
+                            if array.null_count() > 0 && !lance_supports_nulls(array.data_type()) {
+                                return Err(Error::invalid_input(format!(
+                                    "Found rows on LHS that do not match any rows on RHS. Lance would need to write \
+                                    nulls on the RHS, but Lance does not yet support nulls for type {:?}.",
```
jackye1995: I don't think there is any type we don't support null? What do you mean here?

jackye1995: This feels like just an old restriction. Is that still an issue today? @Xuanwo

westonpace: Yeah, this is the list for the 0.1 file format.  In 2.0 file format we support null on everything except struct.  In 2.1 we should support nulls for all data types.

Maybe we should change `lance_supports_nulls` to be a function on the dataset so we can refer to the format version.

java/src/main/java/com/lancedb/lance/Constants.java
wayneli-vt: Thanks for you review! I've renamed the class.

java/src/test/java/com/lancedb/lance/operation/UpdateTest.java
wayneli-vt: Thanks for your suggestion! I've reverted the WriteParams to its default settings, which disables the stable row ID.

java/src/test/java/com/lancedb/lance/operation/UpdateTest.java
wayneli-vt: Thanks for the great suggestion! I've restructured the test data to be more comprehensive.
It now covers:
* Overriding existing values.
* Changing a column's value to null.
* Updating a column from null to a new value.
* Ensuring that untouched rows retain their original state (for both null and non-null values).

rust/lance/src/dataset/hash_joiner.rs
wayneli-vt: Thanks for the valuable suggestions, @jackye1995 and @westonpace!

To give some context, this validation logic was part of `HashJoiner::collect()`, which is called by `FileFragment::merge_columns()`. (Link: [hash_joiner.rs#L130](https://github.com/lancedb/lance/blob/b4e3c68801fee7226f870b289b7adc7b267ddc68/rust/lance/src/dataset/hash_joiner.rs#L130))

Lance now supports nulls for more data types, but I wasn't sure about the specific details, so I kept the original code as a precaution. In hindsight, I probably should have raised this as a question from the start.

Following your advice, I've now implemented `lance_supports_nulls()` as a method on Dataset. It checks for nullability based on the LanceFileVersion.

I really appreciate you taking the time to explain this!

wayneli-vt:
@jackye1995 @westonpace  Hi, I've addressed all the comments. Could you please take another look? Thanks very much!

java/lance-jni/src/transaction.rs
```diff
@@ -881,11 +881,18 @@ fn convert_to_rust_operation(
...
+            let mut buffer: Vec<i64> = vec![0; len as usize];
+            env.get_long_array_region(&updated_field_ids_array, 0, &mut buffer)?;
+            let updated_field_ids_unsafe: Vec<u32> = buffer.into_iter().map(|val| val as u32).collect();
+
             Operation::Update {
                 removed_fragment_ids,
                 updated_fragments,
                 new_fragments,
-                fields_modified: vec![],
+                fields_modified: updated_field_ids_unsafe,
```
majin1102: There's context about this:
https://github.com/lancedb/lance/pull/4408#discussion_r2263380296

I checked the usage in UpdateTest. I don't know this should not be used in this update scenario. But I think it's a little confusing here. Let's say the normal update fields_modified is empty, why update a whole column(maybe the real change is limited) makes fields_modified not empty? The actual field list is actually not changed.

That comment said we should use fields_modified when we add a new column when upserting. That makes sense to me because the fields list is actually modified.

majin1102 (changes requested):
Hi, left one comment. Mainly my own confusing.  
PTAL

java/lance-jni/src/transaction.rs
majin1102: Yeah, the updater actually adds a new column and marks a tombstone.

I wonder if this should belong to the Update operation. Because the code path is more like  merging existed columns. The normal update in my mental mind is being splitted into Delete and Insert as a type of row-oriented operation, which is quite different from the column-oriented case metioned in this PR.

Or let's say if we want to define a SQL to do this DML, should it be a merge or update (or maybe merge and update)? From syntax perspective I think this case match "merge into .... when matched ...update". Paste a whole case:

```
MERGE INTO orders AS o
USING new_orders AS no
ON o.order_id = no.order_id
WHEN MATCHED THEN 
    UPDATE SET o.quantity = no.quantity
```

There also a issue raised to talk about the difference between Update and DataReplacement #4744 I think both point to the same operation categories issue.

In my mental mind, there could be three clear categories of data operations:

- row-oriented: append, update, delete
- column-oriented: merge, merge_update(for this case in my mind)/data_replacement
- row-column-mixed: merge_insert

I think the categories are necessary because column operations for now are the speciality of Lance. There's also a concern that the misalignment of Update may lead to higher educational and use costs like schema evolution.

WDYT? @westonpace @jackye1995

steFaiz: > The normal update in my mental mind is being splitted into Delete and Insert as row-oriented operations, which is quite different from the column-oriented case metioned in this PR.
> 
> I think we could ask this question as another way: if we want to define a SQL to do this DML, should it be a merge or update.

@majin1102  I think the key point is that row-oriented operations and column-oriented operations are fundamentally different in nature, so perhaps we don't need to unify the concepts between them.
The main motivation for proposing the UpdateColumn interface is to address the issue [4650](https://github.com/lancedb/lance/issues/4650) raised: MergeColumn can only merge one complete column at a time and brings some schema-level risks. This column-level update is more similar to Paimon's PartialUpdate mechanism — new data can fill in nulls in existing columns or overwrite specific values, rather than causing existing data to be lost.

majin1102: >  I think the key point is that row-oriented operations and column-oriented operations are fundamentally different in nature, so perhaps we don't need to unify the concepts between them

I think the concepts are meaningful for formats cause they should provide clear syntax for varities of engines. As you said, row-oriented operations and column-oriented operations are fundamentally different in nature, if we could clarify them I think will make something better:

1. For engines, they will have a clear choice to make. For example, the case has a merge syntax and I actually could use merge operation commit and get the same result.
2. When we list transactions, we clearly know what happend by operation type
3. Educational cost thing

> MergeColumn can only merge one complete column at a time and brings some schema-level risks

Indeed I agree this should be a different operation from MergeColumn. Just wonder if this should be Update

> This column-level update is more similar to Paimon's PartialUpdate mechanism 

Glad to see this introduction since Lance has fragments which makes this column-level update a more systematic ability. And I don't think Paimon has a systematic column operation which Lance has at least for now(I mean for Paimon the concept of PartialUpdate is somehow reasonable to me)

majin1102: I think I'm self-convinced about using Update

The column-oriented seems physical, not logical. So we could consider the final commit as row changes since the logical columns doesn't change.

Another issue is conflict resolving. It seems the Update operation has been designed as accepting physical fields changing and dealing with it.

Good job on this!

wayneli-vt:
Thanks for the reviews on this so far!

I'm moving this PR to draft status as it conflicts with the recently merged [#4589.](https://github.com/lancedb/lance/pull/4589).
For context, #4589 expanded `Operation::Update` with new fields:
```rust
    Update {
        // ... existing fields ...
        // 👇 New fields from #4589
        /// The fields that used to judge whether to preserve the new frag's id into
        /// the frag bitmap of the specified indices.
        fields_for_preserving_frag_bitmap: Vec<u32>,
        /// The mode of update
        update_mode: Option<UpdateMode>,
    },

```

I'm going to open a separate, smaller PR to expose these new fields to the Java layer first. 

And Once that compatibility PR is merged, I'll update this branch and ping here when this is ready for another look!

wayneli-vt:
@jackye1995 @westonpace @majin1102 Hi, this pr is ready for review! PTAL when you have a moment. Thanks!

rust/lance/src/dataset.rs
```diff
@@ -1748,6 +1749,30 @@ impl Dataset {
     pub fn sql(&mut self, sql: &str) -> SqlQueryBuilder {
         SqlQueryBuilder::new(self.clone(), sql)
     }
+
+    /// Returns true if Lance supports writing this datatype with nulls.
+    pub fn lance_supports_nulls(&self, datatype: &DataType) -> bool {
```
jackye1995: can we make this pub(crate)? Should not really expose this to the public.

jackye1995 (approved):
mostly looks good to me, just a nit

rust/lance/src/dataset.rs
wayneli-vt: Modified. Thanks for you review!

rust/lance/src/dataset/fragment.rs
```diff
@@ -1531,6 +1531,93 @@ impl FileFragment {
...
+            ));
+        };
+        for field in right_schema.fields() {
+            if ROW_ID.eq(field.name()) || ROW_ADDR.eq(field.name()) {
+                continue;
+            }
+            if self.schema().field(field.name()).is_none() {
+                return Err(Error::invalid_input(
+                    format!(
+                        "Column {} in left side fragment does not exist in right side fragment",
```
westonpace:
```suggestion
                        "Column {} in right side fragment does not exist in left side fragment",
```
Minor nit: I think left/right are backwards in this message?

rust/lance/src/dataset/fragment.rs
```diff
@@ -1531,6 +1531,93 @@ impl FileFragment {
...
+            return Err(Error::invalid_input(
+                format!(
+                    "Column {} does not exist in the right side fragment",
+                    right_on
+                ),
+                location!(),
+            ));
+        };
+        for field in right_schema.fields() {
+            if ROW_ID.eq(field.name()) || ROW_ADDR.eq(field.name()) {
```
westonpace, line 1562: Will this lead to an error below when we call `project_by_schema`?  Or do these columns get dropped somewhere?

westonpace (approved):
Looks good from my perspective (just looking at rust changes)

rust/lance/src/dataset/fragment.rs
wayneli-vt: Thank you for the correction! You're right, I had them backwards. This has been updated!

rust/lance/src/dataset/fragment.rs
wayneli-vt: That's a great question. You are correct—`ROW_ID` and `ROW_ADDR` are internal metadata columns and are in the dataset's main schema (`self.schema()`). My original `continue` was intended to skip them to prevent an error during the schema check.

However, your review made me realize a more fundamental issue: the `write_schema` (derived from `right_schema` by excluding the `right_on` column) should not contain `ROW_ID` or `ROW_ADDR` in the first place, and an `Error` should be raised to prevent callers from trying to update these reserved fields.

I've updated the validation logic accordingly. Thank you for the review!

jackye1995 (approved):
Looks good to me ！

fangbo:
@wayneli-vt There is an error when updating a Struct field using FileFragment.update_columns. This is the detail https://github.com/lance-format/lance/issues/5717 . Could you please take a look ?

PR #4721, opened 2025-09-13
feat: add LANCE_LOG_FILE environment variable for file logging

yanghua:


yanghua:
Hi @westonpace would you please review this PR?

java/lance-jni/src/lib.rs
```diff
@@ -95,13 +97,41 @@ fn set_timestamp_precision(builder: &mut env_logger::Builder) {
...
 }
 
+fn set_log_file_target(builder: &mut env_logger::Builder) {
+    if let Ok(log_file_path) = env::var("LANCE_LOG_FILE") {
+        let path = Path::new(&log_file_path);
+
+        // Create parent directories if they don't exist
+        if let Some(parent) = path.parent() {
+            if std::fs::create_dir_all(parent).is_err() {
+                // just fallback to stderr
```
westonpace: Should we print the error here too like we do below?

python/src/lib.rs
```diff
@@ -151,13 +153,41 @@ fn set_timestamp_precision(builder: &mut env_logger::Builder) {
...
 
+fn set_log_file_target(builder: &mut env_logger::Builder) {
+    if let Ok(log_file_path) = env::var("LANCE_LOG_FILE") {
+        let path = Path::new(&log_file_path);
+
+        // Create parent directories if they don't exist
+        if let Some(parent) = path.parent() {
+            if std::fs::create_dir_all(parent).is_err() {
+                // just fallback to stderr
+                return;
```
westonpace: Same question as above

PR #4725, opened 2025-09-15
fix: refine SIMD support detection for AArch64 on iOS/tvOS

felix-schultz:
This pull request updates the detection logic for FP16 SIMD support on AArch64 platforms in the `lance-core` crate. The main change is to ensure that on iOS and tvOS, NEON SIMD support is assumed for FP16 operations, while on other AArch64 platforms, runtime detection is still used.

Platform-specific SIMD detection:

* In `rust/lance-core/src/utils/cpu.rs`, the initialization of `FP16_SIMD_SUPPORT` now always assumes NEON support for FP16 on AArch64 iOS and tvOS platforms, while retaining runtime detection for other AArch64 targets.

### DIFF
--- a/rust/lance-core/src/utils/cpu.rs
+++ b/rust/lance-core/src/utils/cpu.rs
@@ -16,7 +16,12 @@ pub enum SimdSupport {
 
 /// Support for FP16 SIMD operations
 pub static FP16_SIMD_SUPPORT: LazyLock<SimdSupport> = LazyLock::new(|| {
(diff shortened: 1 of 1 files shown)


Xuanwo (approved):

Xuanwo:
Mostly LGTM, the only thing left is a code format.

Xuanwo:
Thank you @felix-schultz for working on this, merging.

PR #4726, opened 2025-09-15
perf: hierarchical clustering

### LINKED ISSUE #4663: perf: faster and more balanced clustering

BubbleCal:
resolve #4663

This boosts the KMeans algo by ~30x faster, and even much more for large `k` (`num_partitions`)

With this, it's possible to run KMeans with `k=16384` within 25s (which requires over 10mins before this).
And no obvious accuracy loss observed according to my benchmark.


rust/lance-index/src/vector/kmeans.rs
```diff
@@ -70,6 +73,16 @@ pub struct KMeansParams {
...
+
+    /// Balance factor for the kmeans clustering.
+    /// Higher value means more balanced clustering.
+    ///
+    /// Setting this value to 0 means no balance factor,
+    /// which is the same as normal kmeans clustering.
+    pub balance_factor: f32,
+
+    /// Whether to use hierarchical clustering.
+    pub hierarchical_k: usize,
```
wjones127: This comment makes it sound like an on/off switch, but the type is an integer. Could you update the description?

rust/lance-index/src/vector/kmeans.rs
```diff
@@ -103,6 +118,16 @@ impl KMeansParams {
             ..Default::default()
         }
     }
+
+    pub fn with_balance_factor(mut self, balance_factor: f32) -> Self {
+        self.balance_factor = balance_factor;
+        self
+    }
+
+    pub fn with_hierarchical_k(mut self, hierarchical_k: usize) -> Self {
```
wjones127: Could you add doc comments to these?

rust/lance-index/src/vector/kmeans.rs
```diff
@@ -598,21 +682,248 @@ impl KMeans {
...
+    where
+        T::Native: Clone,
+    {
+        let mut subset_data = Vec::with_capacity(indices.len() * dimension);
+        for &idx in indices {
+            let start = idx * dimension;
+            let end = start + dimension;
+            subset_data.extend_from_slice(&data_values[start..end]);
+        }
+        let array = PrimitiveArray::<T>::from_iter_values(subset_data);
```
wjones127: Does `From` not work here? That should be zero-copy, whereas I think `from_iter_values` will create a new buffer?
```suggestion
        let array = PrimitiveArray::<T>::from(subset_data);
```

rust/lance-index/src/vector/kmeans.rs
```diff
@@ -598,21 +682,248 @@ impl KMeans {
...
+        debug_assert_eq!(heap.len(), target_k);
+
+        // Construct final KMeans model with all centroids
+        let mut all_clusters: Vec<Cluster<T::Native>> = heap.into_vec();
+        // Sort by ID to ensure consistent ordering
+        all_clusters.sort_by_key(|c| c.id);
+
+        let flat_centroids: Vec<T::Native> =
+            all_clusters.into_iter().flat_map(|c| c.centroid).collect();
+        let centroids_array = PrimitiveArray::<T>::from_iter_values(flat_centroids);
```
wjones127: Same thought here:
```suggestion
        let centroids_array = PrimitiveArray::<T>::from(flat_centroids);
```

PR #4732, opened 2025-09-15
ci: fix up file integration test so it runs again.

westonpace:
We have a nightly file integration test that has been failing regularly.  It seems the failure was OOM.  This increases the timeout and runs on a bigger runner so we don't have to worry about that.


westonpace:
~~Example run with changes: https://github.com/lancedb/lance/actions/runs/17744993640/job/50428019340~~

Example run with changes: https://github.com/lancedb/lance/actions/runs/17745780167

Xuanwo (approved):

Issue #4262, opened 2025-07-19
Improve throttling handling for S3


jackye1995:
Some learnings that we could borrow from Iceberg: https://github.com/apache/iceberg/pull/11052

But maybe it belongs more to the arrow object_store layer?

Xuanwo:
It should be possible for us to develop our own retry policy. Let me take a look.

PR #4737, opened 2025-09-16
feat: optimize file name for s3 throughput

jackye1995:
Related to #4262

### DIFF
--- a/rust/lance/src/dataset/fragment/write.rs
+++ b/rust/lance/src/dataset/fragment/write.rs
@@ -23,6 +23,54 @@ use crate::dataset::write::do_write_fragments;
 use crate::dataset::{WriteMode, WriteParams, DATA_DIR};
 use crate::Result;
 
+/// Generates a filename optimized for S3 throughput using a UUID-based approach.
+///
+/// This approach follows Apache Iceberg's ObjectStoreLocationProvider pattern:
+/// - Takes a UUID (16 bytes total)
+/// - Uses first 3 bytes (24 bits) as binary string prefix for S3 distribution
+/// - Uses remaining 13 bytes as hex string for uniqueness
+///
+/// Format: `<24-bit-binary><remaining-hex>`
+/// Example: "101100101101010011010110a1b2c3d4e5f6g7h8i9j0"
+///
+/// The binary prefix ensures files are distributed evenly across S3 prefixes,
+/// minimizing throttling and maximizing throughput, while maintaining uniqueness.
+pub(crate) fn generate_random_filename() -> String {
+    // Generate a UUID (16 bytes)
+    let uuid = Uuid::new_v4();
+    let uuid_bytes = uuid.as_bytes();
+
+    // Use first 3 bytes (24 bits) for binary prefix
+    let binary_prefix = to_binary_string(&uuid_bytes[0..3]);
+
+    // Convert remaining 13 bytes to hex string
+    let hex_suffix: String = uuid_bytes[3..]
+        .iter()
+        .map(|byte| format!("{:02x}", byte))
+        .collect();
+
+    // Combine binary prefix with hex suffix
+    format!("{}{}", binary_prefix, hex_suffix)
+}
+
+/// Converts bytes to binary string representation (0s and 1s)
+/// This provides optimal distribution for S3 prefix sharding
+fn to_binary_string(bytes: &[u8]) -> String {
+    let mut result = String::with_capacity(bytes.len() * 8);
+
+    for &byte in bytes {
+        for i in (0..8).rev() {
+            if (byte >> i) & 1 == 1 {
+                result.push('1');
+            } else {
+                result.push('0');
+            }
+        }
+    }
+
+    result
+}
+
 /// Builder for writing a new fragment.
 ///
 /// This builder can be re-used to write multiple fragments.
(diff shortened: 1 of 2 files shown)


rust/lance/src/dataset/fragment/write.rs
```diff
@@ -23,6 +23,54 @@ use crate::dataset::write::do_write_fragments;
...
 
+/// Generates a filename optimized for S3 throughput using a UUID-based approach.
+///
+/// This approach follows Apache Iceberg's ObjectStoreLocationProvider pattern:
+/// - Takes a UUID (16 bytes total)
+/// - Uses first 3 bytes (24 bits) as binary string prefix for S3 distribution
+/// - Uses remaining 13 bytes as hex string for uniqueness
+///
+/// Format: `<24-bit-binary><remaining-hex>`
+/// Example: "101100101101010011010110a1b2c3d4e5f6g7h8i9j0"
```
jackye1995, line 34: Note that here we don't do the additional slashes (`1001/1001/1001/10011001`) like Iceberg, because that is only going to harm the performance (although the effect is relatively small). The main reason for doing the slashes is to be friendly to HDFS file listing because it does not paginate, but that is probably the platform we care the least about, so I just did not bother adding those.

rust/lance/src/dataset/fragment/write.rs
```diff
@@ -23,6 +23,59 @@ use crate::dataset::write::do_write_fragments;
...
+
+    // Convert remaining 13 bytes to hex string
+    let hex_suffix: String = uuid_bytes[3..]
+        .iter()
+        .map(|byte| format!("{:02x}", byte))
+        .collect();
+
+    // Combine binary prefix with hex suffix
+    format!("{}{}", binary_prefix, hex_suffix)
+}
```
Xuanwo: is it worth to avoid duplicated `format!()` call?

(generated by GPT-5)

```rust
pub(crate) fn generate_random_filename() -> String {
    let bytes = Uuid::new_v4().as_bytes();

    let mut out = String::with_capacity(50);

    for &b in &bytes[..3] {
        for i in (0..8).rev() {
            out.push(if (b >> i) & 1 == 1 { '1' } else { '0' });
        }
    }

    const HEX: &[u8; 16] = b"0123456789abcdef";
    for &b in &bytes[3..] {
        out.push(HEX[(b >> 4) as usize] as char);
        out.push(HEX[(b & 0xf) as usize] as char);
    }

    out
}
```

Xuanwo (approved):
LGMT overall, only a small nit

rust/lance/src/dataset/fragment/write.rs
```diff
@@ -23,6 +23,46 @@ use crate::dataset::write::do_write_fragments;
 use crate::dataset::{WriteMode, WriteParams, DATA_DIR};
 use crate::Result;
 
+/// Generates a filename optimized for S3 throughput using a UUID-based approach.
+///
+/// This approach follows Apache Iceberg's ObjectStoreLocationProvider pattern:
+/// - Takes a UUID (16 bytes total)
+/// - Uses first 3 bytes (24 bits) as binary string prefix for S3 distribution
+/// - Uses remaining 13 bytes as hex string for uniqueness
```
wjones127, line 31: I'm confused. I thought `ObjectStoreLocationProvider` inserted randomness at the **beginning** of the S3 path.

https://aws.amazon.com/blogs/big-data/improve-operational-efficiencies-of-apache-iceberg-tables-built-on-amazon-s3-data-lakes/

Like `s3://<random prefix here>/my_database/my_table/09be6e98-e228-48c9-a6c1-8728ebd21308.lance`.

This seems to just modify the filename, but doesn't do anything for the beginning of the path, which I would think is what matters, right?

~~It's also unclear to me how this is an improvement. Isn't the string, especially the prefix, already random? Why does is binary format any better?~~ I see your comments below.

jackye1995: Ah yeah that blog was in 2023 from my team, and the binary bits are introduced 2024. 

Basically, introducing random prefix at the beginning is going to be the most effective in theory, but no one ended up ever doing that because people still in general hold the assumption that all data of one table goes within one fixed prefix, and no multiple tables share the same prefix. It is tied to many things like access control, garbage collection, etc. that can hardly be changed just for the sake of fewer S3 throttling. The "dream" was that people in that mode would use S3 inventory report to do garbage collection, and use S3 tags to do access control. But inventory report did not offer Iceberg integration until very recently and is a S3 specific vendor feature, and tags are expensive and have fundamental limits in the max number of tags per account.

`ObjectStoreLocationProvider` itself always inject randomness after the data directory by default, it has been like that since the beginning. So what we could really do is to make whatever after the data directory easier to scale up. And based on all the experiments, it is the fastest to scale up with the binary scheme because it allows S3 to provision 2 partitions, then 4 partitions, then 8, ... Compared to hex scheme which you will directly provision 16 and then 16^2, etc. the binary is faster because there are fewer machines to provision (throttling continues until all machines are provisioned), and also you are more easily qualified for scale up based on throttling traffic.

But of course what i am taking about is the S3 in late 2024. There has been conversations to make scaling up hex prefixes experience better because everyone is storing UUID names in S3. But that has not landed yet as far as I know.

wjones127: Ah okay. Thanks for the background. Didn't realize you were on that same team!

PR #4739, opened 2025-09-16
refactor: remove fixed parameter constraints from count_rows

Jay-ju:
i hope
```
scanner = ds.scanner(fragments=fragments, columns=required_columns, filter=filter, limit=limit)
scanner.count_rows()
```

but currently need create cnt_scanner, like this
```
cnt_scanner = ds.scanner(fragments=fragments, columns=[], filter=filter, limit=limit, with_row_id=True)
if cnt_scanner.count_rows() == 0:
        return iter(())
```

So here remove mandatory column clearing and row_id requirement for scanner-based count_rows:
- Automatically handle column optimization internally
- Set with_row_id=True by default for scanner inputs
- Maintain existing behavior for non-scanner use cases
- Improve method documentation with usage examples

Before:
  scanner.count_rows(columns=[], with_row_id=True)  # required

After:
  scanner.count_rows()  # simplified usage

### DIFF
--- a/python/python/tests/test_dataset.py
+++ b/python/python/tests/test_dataset.py
@@ -885,17 +885,16 @@ def test_count_rows(tmp_path: Path):
 def test_count_rows_via_scanner(tmp_path: Path):
     ds = lance.write_dataset(pa.table({"a": range(100), "b": range(100)}), tmp_path)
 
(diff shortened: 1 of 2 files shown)


Jay-ju:
@jackye1995 @yanghua  When you have time, could you please take a look at this PR?

Issue #4728, opened 2025-09-15
Simpler and more efficient block bitpacking


westonpace:
The block compressor / decompressor for bitpacking could be simplified.  Right now we use inline bitpacking if there are fewer than 1024 values and out-of-line bitpacking if there are more than 1024 values.  However, we could change either one to work for both cases.

Inline bitpacking can't work today because it expects there to be at most 1024 values.  This is because it wants to read the compressed bit width from the start of the buffer and then apply that to the next 1024 values.  This works well for mini-block since we've created a mini-block for each 1024 values and we decompress a block at a time.  To fix inline bitpacking we would need to make an unpack method that handles more than 1024 values.  It would work something like this...

```
while not_at_end {
  let compressed_bit_width = read_from_buffer_start(buffer);
  unpack(buffer, output);
  buffer += bytes_in_chunk;
}
```

Out-of-line bitpacking works today but the last block may be inefficiently compressed.  This is because the last block might not have a full 1024 values.  Currently we always solve this by padding with zeros and then bitpacking.  For very small last blocks this is inefficient.  For example, if I have 1025 values then the last block of size 1 is probably not going to benefit from bitpacking.  Instead we should just encode the last block as-is.  Then, during decompression, we can look at the size of the last block.  If it equals T * num_values_in_last_block we can assume it is just uncompressed data and skip the unpacking.

Xuanwo:
Let me take a look on how to get Out-of-line bitpacking works for the last block first.

PR #4740, opened 2025-09-16
feat: allow out-of-line bitpacking supports tail chunk

Xuanwo:
Part of https://github.com/lancedb/lance/issues/4728

This PR will change the out-of-line bitpacking a bit to support tail chunk. My idea is simple: we just enable the tail chunk AS IS and checking it during decoding.

---


Xuanwo:
This change doesn’t seem to be working as I expected, I’ll keep working on it.

rust/lance-encoding/src/encodings/physical/bitpacking.rs
```diff
@@ -318,28 +318,22 @@ fn bitpack_out_of_line<T: ArrowNativeType + BitPacking>(
...
-            compressed_bits_per_value,
-            &last_chunk,
-            &mut output[output_start..],
-        );
-    }
+    output.truncate(num_whole_chunks * words_per_chunk);
+    let remaining_items = data_buffer.len() - last_chunk_start;
+    output.reserve(remaining_items);
+    // Append the raw tail so that we avoid padding zeros and keep decoding fast.
+    output.extend_from_slice(&data_buffer[last_chunk_start..]);
```
westonpace: This changes from "always pad and compress tail" to "never pad and compress tail".  However, I think the ideal is "sometimes pad and compress tail".

For example, if we are packing 16-bits into 2-bits and the tail has 1000 items then it makes sense to add 48-bits of padding to gain 14,000 bits of savings.  However, if we are packing 16-bits into 2-bits and the tail has 100 items then it doesn't make sense to add 1848 bits of padding to gain 1400 bits of savings.

I guess the two calculations are:

padding_cost = compressed_bit_width * (1024 - num_values_in_tail)
tail_pack_savings = (uncompressed_bit_width - compressed_bit_width) * num_values_in_tail

If `padding_cost` is greater than `tail_pack_savings` we don't pack.  If `padding_cost` is less then we do pack.

rust/lance-encoding/src/encodings/physical/bitpacking.rs
```diff
@@ -318,28 +318,22 @@ fn bitpack_out_of_line<T: ArrowNativeType + BitPacking>(
...
 
-/// Unpacks a FixedWidthDataBlock that has been bitpacked with a constant bit width
+/// Unpacks a FixedWidthDataBlock that has been bitpacked with a constant bit width.
 ///
-/// The compressed_bits_per_value is provided
-/// The uncompressed_bits_per_value is based on T's size
+/// The compressed bit width is provided while the uncompressed width comes from `T`.
+/// Buffers written by older versions always packed a padded 1024-value tail. New buffers
+/// append the raw remainder instead. We detect the format from the buffer length so both
+/// layouts remain compatible.
```
westonpace: I'm slightly torn on adding complexity to support backwards compatibility for an unreleased version.  Luckily, it's a non-issue :laughing: 

If we change the pack routine to "sometimes pack" then we need to support both routes anyways, just not for backwards compatibility reasons (so if we change that then let's update this comment)

rust/lance-encoding/src/encodings/physical/bitpacking.rs
Xuanwo: Yep, that makes sense. I will change into this way.

Xuanwo:
cc @westonpace, PTAL, thank you!

rust/lance-encoding/src/encodings/physical/bitpacking.rs
```diff
@@ -125,8 +125,10 @@ impl InlineBitpacking {
         }
 
         // Handle the last chunk
-        let last_chunk_elem_num = if data.num_values % ELEMS_PER_CHUNK == 0 {
-            1024
+        let last_chunk_elem_num = if data.num_values == 0 {
```
westonpace: This should never happen: https://github.com/lancedb/lance/pull/4775

rust/lance-encoding/src/encodings/physical/bitpacking.rs
```diff
@@ -318,28 +314,46 @@ fn bitpack_out_of_line<T: ArrowNativeType + BitPacking>(
...
-        BitPacking::unchecked_pack(
-            compressed_bits_per_value,
-            &last_chunk,
-            &mut output[output_start..],
-        );
+    let uncompressed_bits = data.bits_per_value as usize;
+    let tail_bit_savings = uncompressed_bits.saturating_sub(compressed_bits_per_value);
+    let padding_cost = compressed_bits_per_value * (ELEMS_PER_CHUNK as usize - remaining_items);
+    let tail_pack_savings = tail_bit_savings.saturating_mul(remaining_items);
+    let pad_tail = remaining_items > 0 && tail_bit_savings > 0 && padding_cost < tail_pack_savings;
```
westonpace: `remaining_items > 0` and `tail_bit_savings > 0` seems like something we can assert.  I don't think these can ever be zero.

rust/lance-encoding/src/encodings/physical/bitpacking.rs
```diff
@@ -318,28 +314,46 @@ fn bitpack_out_of_line<T: ArrowNativeType + BitPacking>(
         return LanceBuffer::reinterpret_vec(output);
     }
 
-    // If we get here then the last chunk needs to be padded with zeros
-    let remaining_items = data_buffer.len() % ELEMS_PER_CHUNK as usize;
     let last_chunk_start = num_whole_chunks * ELEMS_PER_CHUNK as usize;
+    output.truncate(num_whole_chunks * words_per_chunk);
```
westonpace: Minor nit: instead of truncating here maybe we just set the length with `num_whole_chunks` above?

rust/lance-encoding/src/encodings/physical/bitpacking.rs
```diff
@@ -318,28 +314,46 @@ fn bitpack_out_of_line<T: ArrowNativeType + BitPacking>(
...
+
+    if pad_tail {
+        // Padding buys us more than it costs: pad to 1024 values and pack them as a normal chunk.
+        let mut last_chunk: Vec<T> = vec![T::from_usize(0).unwrap(); ELEMS_PER_CHUNK as usize];
+        last_chunk[..remaining_items].copy_from_slice(&data_buffer[last_chunk_start..]);
+        let start = output.len();
+        output.extend(std::iter::repeat_n(
+            T::from_usize(0).unwrap(),
+            words_per_chunk,
+        ));
```
westonpace: Minor nit: you can (I think) avoid this extend call and just call `set_len` with the appropriate length.

rust/lance-encoding/src/encodings/physical/bitpacking.rs
```diff
@@ -318,28 +314,46 @@ fn bitpack_out_of_line<T: ArrowNativeType + BitPacking>(
...
+        unsafe {
+            BitPacking::unchecked_pack(
+                compressed_bits_per_value,
+                &last_chunk,
+                &mut output[start..start + words_per_chunk],
+            );
+        }
+    } else {
+        output.reserve(remaining_items);
```
westonpace: Minor nit: This `reserve` call is harmless but I don't think it is needed.  The capacity was set larger than this when we first created the vec.

rust/lance-encoding/src/encodings/physical/bitpacking.rs
```diff
@@ -349,29 +363,47 @@ fn unpack_out_of_line<T: ArrowNativeType + BitPacking>(
...
         unsafe {
             BitPacking::unchecked_unpack(
                 compressed_bits_per_value,
                 &compressed_words[input_start..input_end],
-                &mut decompressed[output_start..output_end],
+                &mut chunk_buf,
             );
         }
+        decompressed.extend_from_slice(&chunk_buf);
+    }
```
westonpace: Why this change?  Doesn't it introduce an extra copy?  I have to decompress into `chunk_buf` and then copy into `decompressed` instead of decompressing directly into `decompressed`?

rust/lance-encoding/src/encodings/physical/bitpacking.rs
```diff
@@ -349,29 +363,47 @@ fn unpack_out_of_line<T: ArrowNativeType + BitPacking>(
...
+
+    if tail_values > 0 {
+        // Older buffers always padded the tail to 1024 items and then packed it.
+        // New buffers append the raw tail values directly to the compressed blocks.
+        // We infer the layout from the buffer length so we can decode both versions.
```
westonpace:
```suggestion
        // The tail might be padded and bit packed or it might be appended raw.  We infer the
        // layout from the buffer length to decode appropriately.
```

rust/lance-encoding/src/encodings/physical/bitpacking.rs
Xuanwo: merged.

rust/lance-encoding/src/encodings/physical/bitpacking.rs
Xuanwo: Changed as `debug_assert`

rust/lance-encoding/src/encodings/physical/bitpacking.rs
Xuanwo: Yep, it should work.

Xuanwo:
> Do we have a bitpacking microbenchmark that we can compare against?

Not yet, I can build one for this.

rust/lance-encoding/src/encodings/physical/bitpacking.rs
Xuanwo: Fixed.

rust/lance-encoding/src/encodings/physical/bitpacking.rs
Xuanwo: Removed.

rust/lance-encoding/src/encodings/physical/bitpacking.rs
Xuanwo: I tried to refactor this part a bit to get rid of the unsafe. It's now reverted.

westonpace:
> Not yet, I can build one for this.

Now that you've reverted the refactor we can save it for a follow-up.  If performance is stable after the benchmark then I agree that it is better to remove the unsafe.

Xuanwo:
> Now that you've reverted the refactor we can save it for a follow-up. If performance is stable after the benchmark then I agree that it is better to remove the unsafe.

Haha, I have built one! Your steps is a bit after me. https://github.com/lancedb/lance/pull/4777

I agree that we can defer the refactor later. I only post the bench results before and after this PR.

PR #4741, opened 2025-09-16
feat: support tracking newly inserted and updated rows between versions

yanghua:


yanghua:
@jackye1995 When you have time, PTAL! thanks.

protos/row_versions.proto
jackye1995, line 1: given your rust file is just at `rowid/version.rs`, why not just put these in `rowid.proto`?

protos/row_versions.proto
```diff
@@ -0,0 +1,39 @@
...
+
+syntax = "proto3";
+
+package lance.table;
+
+/// A sequence of row latest update versions. Similar to RowIdSequence but tracks
+/// the latest update version for each row. This is optimized for different patterns
+/// of version assignment: uniform (all rows same version), incremental (consecutive
+/// versions), or sparse (irregular version patterns).
+message RowLatestUpdateVersionSequence {
```
jackye1995: Here the structure is technically a map of row_id -> version. But we already have all the row_ids in RowIdSequence. Why not just record another sequence here, and we can reconstruct the mapping by reading the RowIdSequence and the RowLatestUpdatedVersionSequence.

protos/row_versions.proto
```diff
@@ -0,0 +1,39 @@
...
+
+    /// A single (row_id, version) pair.
+    message RowVersionPair {
+        uint64 row_id = 1;
+        uint64 version = 2;
+    }
+
+    oneof segment {
+        /// When consecutive rows have the same version.
+        UniformRange uniform_range = 1;
```
jackye1995: this is missing the other option?

protos/row_versions.proto
```diff
@@ -0,0 +1,39 @@
...
+        /// The start row ID, inclusive.
+        uint64 start_row = 1;
+        /// The end row ID, exclusive.
+        uint64 end_row = 2;
+        /// The version number for all rows in this range.
+        uint64 version = 3;
+    }
+
+    /// A single (row_id, version) pair.
+    message RowVersionPair {
```
jackye1995: I think it's better to do things like U64Segment like `UniformRangeWithHoles`, because with that you can directly encode unique versions in a dedicated unit64 array, instead of having to mix that with other messages in `oneof` which increases the serialization size.

protos/row_versions.proto
yanghua: I thought it was an independent part, so I put it into a single file. But it's also ok to put in `rowid.proto`

protos/row_versions.proto
yanghua: You mean `RowIdSequence` for row_id, `RowLatestUpdatedVersionSequence` for `version`. And build the mapping based on these two sequences?

protos/row_versions.proto
yanghua: oh, currently this structure(`RowVersionPair`) may not be necessary. It was used here: https://github.com/lancedb/lance/pull/4741/commits/cdb96789ae419c7b69cc64596563c252cffe532f#diff-19f6b0b5021f4957fa3b17c1186ceb08832dcc1d8fc0997ef2336fd8c3e5d2c9R44

But I deleted the `SparseVersions `.

protos/row_versions.proto
yanghua: In the original implementation, I introduced another two structured `IncrementalRange ` and `SparseVersions `. But now, I feel that if we only have `UniformRange`, it has bug.

protos/row_versions.proto
jackye1995: yes

rust/lance/src/dataset/delta.rs
```diff
@@ -152,4 +555,747 @@ mod tests {
...
+
+        let dataset = Dataset::write(data, "memory://test_no_stable_ids", Some(write_params))
+            .await
+            .unwrap();
+
+        // Try to execute diff operation - should fail
+        let diff_builder = DatasetDiffBuilder::new(Arc::new(dataset.clone()), 1);
+        let result = diff_builder.execute().await;
+
+        assert!(result.is_err(), "Diff should fail without stable row IDs");
```
jackye1995: should check the error message is as expected

rust/lance/src/dataset/delta.rs
```diff
@@ -152,4 +555,747 @@ mod tests {
...
+        let result = diff_builder.execute().await;
+        assert!(
+            result.is_err(),
+            "Should fail when compared version is greater than current version"
+        );
+
+        // Test with version 0 (should fail)
+        let diff_builder = DatasetDiffBuilder::new(Arc::new(dataset.clone()), 0);
+        let result = diff_builder.execute().await;
+        assert!(result.is_err(), "Should fail when compared version is 0");
```
jackye1995: should check the error message is as expected

rust/lance/src/dataset/delta.rs
```diff
@@ -50,15 +57,411 @@ impl DatasetDelta {
...
+                new_data,
+                version,
+            });
+        }
+
+        Ok(records)
+    }
+
+    /// Get record data for a specific row ID from a dataset
+    async fn get_record_data(&self, row_id: u64, dataset: &Dataset) -> Result<RecordBatch> {
```
jackye1995: 2 questions:

(1) this would be quite inefficient to get a single row from a dataset, can we instead batch the operation?

(2) what if the dataset is already being cleaned up? I think getting the pre-image should be optional. what do you think?

rust/lance/src/dataset/delta.rs
```diff
@@ -50,15 +57,411 @@ impl DatasetDelta {
...
+            let old_record_result = self.get_record_data(row_id, &self.old_dataset).await;
+            let row_exists_in_old = old_record_result.is_ok();
+
+            let operation = if row_exists_in_old {
+                // If the row exists in both versions, we need to check if it actually changed
+                let old_record = old_record_result.unwrap();
+                let new_record = self.get_record_data(row_id, &self.current_dataset).await?;
+
+                // Compare the records to see if they're actually different
+                let is_actually_updated = self.records_are_different(&old_record, &new_record)?;
```
jackye1995: I think this should not be possible? Is this check necessary?

yanghua: hmm.. it's a defensive checks. But, yes, if there is any case about this, it would be a bug. For performance consideration, we can remove it.

rust/lance/src/dataset/transaction.rs
```diff
@@ -1626,7 +1631,21 @@ impl Transaction {
...
                 }
+                // Identify fragments that were updated or newly created in this update
+                let mut target_ids: HashSet<u64> = HashSet::new();
+                target_ids.extend(updated_fragments.iter().map(|f| f.id));
+                target_ids.extend(new_fragments.iter().map(|f| f.id));
+
                 final_fragments.extend(new_fragments);
+                if config.use_stable_row_ids {
+                    let current_version = self.read_version + 1;
+                    // Override version metadata for updated/new fragments to mark this change
```
jackye1995: One important thing I just realize I forgot to mention, I think for any new fragments, the `RowLatestUpdateVersionMeta` should remain null. This saves computational effort and also makes it more backwards compatible. Essentially, if the fragment does not contain this information, it is considered that all rows are added in the latest version.

yanghua: Sounds reasonable, will try it.

yanghua: Done.

jackye1995:
Saw quite a few new changes, let me know when this is ready to be reviewed again!

yanghua:
> Saw quite a few new changes, let me know when this is ready to be reviewed again!

@jackye1995 Addressed your comments and added new test cases. Please review again.

rust/lance/src/dataset/delta.rs
```diff
@@ -152,4 +373,1254 @@ mod tests {
...
+                assert_eq!(updated_at[i], 2);
+            }
+        }
+    }
+
+    #[tokio::test]
+    async fn test_row_version_metadata_after_delete() {
+        use arrow_array::cast::AsArray;
+        use arrow_array::types::UInt64Type;
+        use lance_core::{ROW_CREATED_AT_VERSION, ROW_LAST_UPDATED_AT_VERSION};
```
yanghua: We can extract these imports to be common imports.

rust/lance/src/dataset/delta.rs
```diff
@@ -152,4 +373,1254 @@ mod tests {
             _ => panic!("Expected VersionNotFound error."),
         }
     }
+
```
yanghua, line 381: We may need to add some tests about:

* diff after compaction;
* diff after merge insert(full schema, sub col);

to verify version rechunk.

It could be done in further work.

rust/lance/src/dataset/delta.rs
```diff
@@ -152,4 +373,1254 @@ mod tests {
...
+        let keys = result["key"].as_primitive::<Int32Type>().values();
+
+        for i in 0..result.num_rows() {
+            assert_eq!(created_at[i], 3);
+            assert!(keys[i] >= 80 && keys[i] < 100);
+        }
+    }
+
+    #[tokio::test]
+    async fn test_get_updated_rows() {
```
yanghua, line 1432: We can also extract some code or build a test harness to reduce the code.

jackye1995:
Looks like test have been failing due to https://github.com/lancedb/lance/issues/4926. Feel free to merge.

rust/lance/src/dataset/delta.rs
jackye1995: Oh great point, missed those cases

jackye1995 (approved):
thanks for working on this!

PR #4743, opened 2025-09-16
fix: speed up generating ProjectionPlan for full schema

timsaucer:
In https://github.com/lancedb/lance/pull/4478 a change was introduced that added the `ProjectionPlan::full()` function. This will get the entire schema for a dataset and turn every column into a SQL expression in order to call `from_expressions`. This is causing us a significant performance hit when we have datasets that have very large schemas. Since we know for certain that this function is getting all of the columns, we should be able to short circuit the work being done in parsing strings to columns against the schema and instead jump straight to evaluating each column in the schema as an `Expr`.

Running one of our benchmarks (time in seconds):

- On commit 2b774a47 (before PR 4478): 0.4338
- On commit eeea03c2 (after PR 4478): 0.7486
- eeea03c2 with my update in this PR: 0.4274

### DIFF
--- a/rust/lance-datafusion/src/projection.rs
+++ b/rust/lance-datafusion/src/projection.rs
@@ -199,13 +199,29 @@ impl ProjectionPlan {
     }
 
     pub fn full(base: Arc<dyn Projectable>) -> Result<Self> {
-        let projection = base
+        let physical_cols: Vec<&str> = base
             .schema()
             .fields
             .iter()
-            .map(|f| (f.name.as_str(), format!("`{}`", f.name.as_str())))
+            .map(|f| f.name.as_ref())
             .collect::<Vec<_>>();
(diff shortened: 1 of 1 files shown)


timsaucer:
The one failing test does not appear related to this PR, but I don't have the ability to rerun failing jobs.

westonpace (approved):
Nice catch.

We definitely need more benchmarks like this helping us regress PRs.  What was the benchmark doing?  Just a simple query?  Was the schema unusually wide?

timsaucer:
> Nice catch.
> 
> We definitely need more benchmarks like this helping us regress PRs. What was the benchmark doing? Just a simple query? Was the schema unusually wide?

Yes, this table was very wide and our benchmark is currently running 500-2000 queries, so the time spent on this became noticeable.

PR #4745, opened 2025-09-16
Expand testing of various queries

wjones127:


.github/workflows/rust.yml
```diff
@@ -127,6 +126,37 @@ jobs:
           files: coverage.codecov
           flags: unittests
           fail_ci_if_error: false
+
+  integration_test:
```
cmccabe: does it make sense to call this `query_integration_tests`? Since we may add other kinds of integration test in the future to sophon.

cmccabe:
sorry, this is probably a dumb question, but why do these take much more time than our other tests? is it because of the greater volume of data here?

wjones127:
> sorry, this is probably a dumb question, but why do these take much more time than our other tests? is it because of the greater volume of data here?

Two motivations for saying this:

1. We run each test for a combination of table states. So, right now, each test is run `3 x 2 x 2 = 12` ways, and that list of states might keep growing as we add more versions. That makes each individual test take longer.
2. I also guide people add compute-intensive tests (for example, ones building bigger ANN indices) here, rather than in the unit tests. I think those benefit the most from being run in release mode.

.github/workflows/rust.yml
wjones127: I'll add other integration tests for sure. But they will be in the same binary and I think run in the same CI job.

rust/lance/tests/query/mod.rs
```diff
@@ -0,0 +1,179 @@
...
+fn create_datafusion_context() -> SessionContext {
+    let ctx = SessionContext::new();
+    register_functions(&ctx);
+    ctx
+}
+
+mod primitives;
+mod vectors;
+
+/// Scanning an ordering by id should give same result as original.
```
wjones127:
```suggestion
/// Scanning and ordering by id should give same result as original.
```

rust/lance/tests/query/mod.rs
```diff
@@ -0,0 +1,179 @@
...
+        (0..num_rows.min(10)).collect(), // Sequential
+        vec![num_rows - 1, 0],           // Last and first
+    ];
+
+    for indices in cases {
+        // Skip cases with invalid indices
+        if indices.iter().any(|&i| i >= num_rows) {
+            continue;
+        }
+
```
wjones127: Shouldn't be necessary
```suggestion
```

rust/lance/tests/query/mod.rs
```diff
@@ -0,0 +1,179 @@
...
+        // Convert to u64 for Lance take
+        let indices_u64: Vec<u64> = indices.iter().map(|&i| i as u64).collect();
+
+        if indices_u64.is_empty() {
+            // Skip empty case as Lance may not handle it the same way
+            continue;
+        }
```
wjones127:
```suggestion
```

.github/workflows/rust.yml
```diff
@@ -136,14 +136,41 @@ jobs:
...
           cargo test --locked --features ${ALL_FEATURES} --no-run
       - name: Start DynamodDB and S3
         run: docker compose -f docker-compose.yml up -d --wait
       - name: Run tests
         run: |
-          ALL_FEATURES=`cargo metadata --format-version=1 --no-deps | jq -r '.packages[] | .features | keys | .[]' | grep -v protoc | sort | uniq | paste -s -d "," -`
+          ALL_FEATURES=`cargo metadata --format-version=1 --no-deps | jq -r '.packages[] | .features | keys | .[]' | grep -v -e protoc -e slow_tests | sort | uniq | paste -s -d "," -`
           cargo test --locked --features ${ALL_FEATURES}
+  query-integration-tests:
+    runs-on: ubuntu-2404-8x-arm64
```
Xuanwo:
```suggestion
    runs-on: ubuntu-24.04-arm64-8x
```

PR #4751, opened 2025-09-17
chore: upgrade datafusion to 50.0.0 and arrow to 56.1

ddupg:
This PR is to update the following dependencies:
- DataFusion to 50.0.0
- Arrow to 56.1

I am working on supporting GEO type #4678 , which is blocked by the DataFusion upgrade. Recently, DataFusion released 50 https://github.com/apache/datafusion/issues/16799 , so follow up on the upgrade

Closes https://github.com/lancedb/lance/issues/4753

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,18 +4,18 @@ version = 3
 
 [[package]]
 name = "addr2line"
-version = "0.24.2"
+version = "0.25.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "dfbe277e56a376000877090da837660b4427aad530e3028d44e0bffe4f89a1c1"
+checksum = "1b5d307320b3181d6d7954e663bd7c774a838b8220fe0593c86d9fb09f498b4b"
 dependencies = [
  "gimli",
 ]
 
 [[package]]
 name = "adler2"
-version = "2.0.0"
+version = "2.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "512761e0bb2578dd7380c6baaa0f4ce03e84f95e960231d1dec8bf4d7d6e2627"
+checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "adler32"
@@ -94,12 +94,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "683d7910e743518b0e34f1186f92494becacb047c7b6bf616c96772180fef923"
 
 [[package]]
-name = "android-tzdata"
-version = "0.1.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e999941b234f3131b00bc13c22d06e8c5ff726d1b6318ac7eb276997bbb4fef0"
-
-[[package]]
 name = "android_system_properties"
 version = "0.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -116,9 +110,9 @@ checksum = "4b46cbb362ab8752921c97e041f5e366ee6297bd428a31275b9fcf1e380f7299"
 
 [[package]]
 name = "anstream"
-version = "0.6.19"
+version = "0.6.21"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "301af1932e46185686725e0fad2f8f2aa7da69dd70bf6ecc44d6b703844a3933"
+checksum = "43d5b281e737544384e969a5ccad3f1cdd24b48086a0fc1b2a5262a26b8f4f4a"
 dependencies = [
  "anstyle",
  "anstyle-parse",
@@ -131,9 +125,9 @@ dependencies = [
 
 [[package]]
 name = "anstyle"
-version = "1.0.11"
+version = "1.0.13"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "862ed96ca487e809f1c8e5a8447f6ee2cf102f846893800b20cebdf541fc6bbd"
+checksum = "5192cca8006f1fd4f7237516f40fa183bb07f8fbdfedaa0036de5ea9b0b45e78"
 
 [[package]]
 name = "anstyle-parse"
@@ -146,29 +140,29 @@ dependencies = [
 
 [[package]]
 name = "anstyle-query"
-version = "1.1.3"
+version = "1.1.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6c8bdeb6047d8983be085bab0ba1472e6dc604e7041dbf6fcd5e71523014fae9"
+checksum = "9e231f6134f61b71076a3eab506c379d4f36122f2af15a9ff04415ea4c3339e2"
 dependencies = [
- "windows-sys 0.59.0",
+ "windows-sys 0.60.2",
 ]
 
 [[package]]
 name = "anstyle-wincon"
-version = "3.0.9"
+version = "3.0.10"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "403f75924867bb1033c59fbf0797484329750cfbe3c4325cd33127941fabc882"
+checksum = "3e0633414522a32ffaac8ac6cc8f748e090c5717661fddeea04219e2344f5f2a"
 dependencies = [
  "anstyle",
  "once_cell_polyfill",
- "windows-sys 0.59.0",
+ "windows-sys 0.60.2",
 ]
 
 [[package]]
 name = "anyhow"
-version = "1.0.98"
+version = "1.0.100"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e16d2d3311acee920a9eb8d33b8cbc1787ce4a264e85f964c2404b969bdcd487"
+checksum = "a23eb6b1614318a8071c9b2521f36b424b2c83db5eb3a0fead4a6c0809af6e61"
 
 [[package]]
 name = "approx"
@@ -378,7 +372,7 @@ version = "56.2.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b3aa9e59c611ebc291c28582077ef25c97f1975383f1479b12f3b9ffee2ffabe"
 dependencies = [
- "bitflags 2.9.1",
+ "bitflags 2.9.4",
  "serde",
  "serde_json",
 ]
@@ -427,9 +421,9 @@ dependencies = [
 
 [[package]]
 name = "async-channel"
-version = "2.3.1"
+version = "2.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "89b47800b0be77592da0afd425cc03468052844aff33b84e33cc696f64e77b6a"
+checksum = "924ed96dd52d1b75e9c1a3e6275715fd320f5f9439fb5a4a11fa51f4221158d2"
 dependencies = [
  "concurrent-queue",
  "event-listener-strategy",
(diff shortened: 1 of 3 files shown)


rust/lance-datafusion/src/planner.rs
```diff
@@ -448,6 +450,7 @@ impl Planner {
                 enable_options_value_normalization: false,
                 collect_spans: false,
                 map_string_types_to_utf8view: false,
+                default_null_ordering: NullOrdering::NullsMax,
```
westonpace, line 453: Do we know if this was the same as before?

LuQQiu:
@ddupg Could you help fix the build issues?

ddupg:
> @ddupg Could you help fix the build issues?

@LuQQiu Sure, now blocked in datafusion-python is not published to pypi, but it may take a few weeks https://github.com/apache/datafusion/issues/16799#issuecomment-3302501977

westonpace:
> Sure, now blocked in datafusion-python is not published to pypi

Hmm...we use the FFI for our boundary between datafusion-python and lance rust which in theory has a stable ABI.  Do we know why we have a hard pin on the datafusion-python version?  Do the tests pass if we loosen the pin?

rust/lance/src/io/exec/rowids.rs
```diff
@@ -303,7 +303,7 @@ impl ExecutionPlan for AddRowAddrExec {
         // is a minimum size of 64 bytes.
         let mut added_byte_size = stats
             .num_rows
-            .map(|n| (n * 8).max(64))
+            .map(|n| n * 8)
```
ddupg, line 306: Closes: https://github.com/apache/arrow-rs/issues/7383

ddupg:
> Do we know why we have a hard pin on the datafusion-python version? Do the tests pass if we loosen the pin?

Related to https://github.com/lancedb/lance/pull/4577, maybe Json type has a requirement for datafusion version.

> Do the tests pass if we loosen the pin?

The currently failing tests are not related to datafusion version, I'm trying to fix it.

ddupg:
The failing test python/tests/test_scalar_index.py::test_null_handling is seems to be caused by a bug of DataFusion rust 50.0.0, and I have raised an issue https://github.com/apache/datafusion/issues/17681 to the DataFusion

ddupg:
> The failing test python/tests/test_scalar_index.py::test_null_handling is seems to be caused by a bug of DataFusion rust 50.0.0, and I have raised an issue [apache/datafusion#17681](https://github.com/apache/datafusion/issues/17681) to the DataFusion

This is not a bug with DataFusion, it seems to be because Lance using indexes and DataFusion have different logic on how to handle NULL. After https://github.com/apache/datafusion/pull/17088, datafusion simplifies the handling of NULL expressions, which affects the results of the test python/tests/test_scalar_index.py::test_null_handling.

In rust/lance-index/src/scalar/expression.rs:create_filter_plan https://github.com/lancedb/lance/blob/5c60975b2c032314304ca1d38865d6eefde4d790/rust/lance-index/src/scalar/expression.rs#L1560, if the filter executed is `IN (1, NULL)`, then the passed `Expr` is looks like this:
```
BinaryExpr(
    BinaryExpr {
        left: BinaryExpr(
            BinaryExpr {
                left: Column(
                    Column {
                        relation: None,
                        name: "x",
                    },
                ),
                op: Eq,
                right: Literal(
                    Int32(1),
                    None,
                ),
            },
        ),
        op: Or,
        right: BinaryExpr(
            BinaryExpr {
                left: Column(
                    Column {
                        relation: None,
                        name: "x",
                    },
                ),
                op: Eq,
                right: Cast(
                    Cast {
                        expr: Literal(
                            NULL,
                            None,
                        ),
                        data_type: Int32,
                    },
                ),
            },
        ),
    },
)
```
but after `self.optimize_expr(filter)?`, the `Eq` `Expr` about NULL is simplified, it looks like this:
```
BinaryExpr(
    BinaryExpr {
        left: BinaryExpr(
            BinaryExpr {
                left: Column(
                    Column {
                        relation: None,
                        name: "x",
                    },
                ),
                op: Eq,
                right: Literal(
                    Int32(1),
                    None,
                ),
            },
        ),
        op: Or,
        right: Literal(
            Boolean(NULL),
            None,
        ),
    },
)
```
and then `Literal` sub `Expr` cannot apply indexes, resulting in the entire `Expr` not being able to use indexes, which eventually causes the test to fail.

Can @westonpace please give me some help on how to fix this test?

timsaucer:
This appears to be a duplicate of my PR: https://github.com/lancedb/lance/pull/4778 but since you're actively working on the issue I was just looking at with the null unit test failure, I can close my PR in favor of this one.

timsaucer:
A little more from my own investigation comparing DF 49 to 50:

Screenshot 2025-09-24 at 9 59 41 AM

Screenshot 2025-09-24 at 10 00 37 AM

~~I suspect the issue may not be in the change to the DF optimizer. I am not sure why the prior test did work right now.~~

Correction: I am beginning to suspect that because the new plan has a `Boolean(NULL)` that the index read filters are no longer passing these through. It also appears that the output now is closer to what DataFusion expects to give *by accident not intention*. By that I mean it the optimizer were enhanced to give `Int64(NULL)` we would go back to the prior result, which is not expected from a DF perspective.

jackye1995:
Thank you @timsaucer for sharing this. I also chatted with @ddupg offline a bit, sorry I forgot to share the analysis here.

I think we did not handle the case of a `Literal(Boolean(Null))` in [visit_node](https://github.com/lancedb/lance/blob/5c60975b2c032314304ca1d38865d6eefde4d790/rust/lance-index/src/scalar/expression.rs#L1428-L1455). 

In the past because the right hand side is a `X = NULL`, it is handled by `visit_comparison`, but now it directly returns None, indicating there is no index match which is where the wrong logic is.

timsaucer:
> Thank you @timsaucer for sharing this. I also chatted with @ddupg offline a bit, sorry I forgot to share the analysis here.
> 
> I think we did not handle the case of a `Literal(Boolean(Null))` in [visit_node](https://github.com/lancedb/lance/blob/5c60975b2c032314304ca1d38865d6eefde4d790/rust/lance-index/src/scalar/expression.rs#L1428-L1455).
> 
> In the past because the right hand side is a `X = NULL`, it is handled by `visit_comparison`, but now it directly returns None, indicating there is no index match which is where the wrong logic is.

I do want to point out that "fixing" the issue by handling the literal null boolean will still give different results than an equivalent SQL command. From a SQL perspective not returning the row with the null IS the correct behavior, even though I think it is counter intuitive.

ddupg:
Thank @jackye1995 @timsaucer both for sharing and thinking, which helped me a lot

> I do want to point out that "fixing" the issue by handling the literal null boolean will still give different results than an equivalent SQL command. From a SQL perspective not returning the row with the null IS the correct behavior, even though I think it is counter intuitive.

Agreed.

My personal opinion is that root cause stems from the different semantics of handling NULL when indexing is enabled or not. This is weird behavior, so I think we should make semantics consistent first. I proposed a draft #4812 to solve this problem. This will fix the currently failed test.

Secondly, the DataFusion upgrade exposed a situation where the index is not working as expected. DF 50 optimizes Expr of  NULL handling, for example, the Expr `x = 1 or x = NULL` is optimized to `x = 1 or NULL`, where the sub Expr `NULL` cannot infer indexes, resulting in the entire query being unable to apply indexes. This may be something we need to optimize in the future.

westonpace:
I agree that we should migrate to the SQL behavior and I agree with @ddupg's summary.  Shall I make a PR to fix the second issue?

westonpace:
Also, there is a third issue, reported [here](https://github.com/lancedb/lance/issues/4756), which will unfortunately be quite a bit trickier to fix.  I don't think that one needs to block the DF upgrade however.

westonpace:
I ended up creating https://github.com/lancedb/lance/pull/4815 which complements #4812 although maybe #4812 is no longer necessary.

jackye1995:
looks like there are some additional errors, any clue?

ddupg:
> looks like there are some additional errors, any clue?

```
called `Result::unwrap()` on an `Err` value: IO { source: Custom { kind: Other, error: Generic { store: "LocalFileSystem", source: InvalidUrl { url: Url { scheme: "file", cannot_be_a_base: false, username: "", password: None, host: None, port: None, path: "/C:%5CUsers%5CRUNNER~1%5CAppData%5CLocal%5CTemp%5C.tmpIGGVXH/some_file.lance", query: None, fragment: None } } } }, location: Location { file: "rust\\lance-file\\src\\v2\\writer.rs", line: 191, column: 9 } }
```

This unusual local path `"/C:%5CUsers%5CRUNNER~1%5CAppData%5CLocal%5CTemp%5C.tmpIGGVXH/some_file.lance"`  appears to have a parsing issue in Windows.

jackye1995:
hmm this is quite strange, I don't see any change that could affect the windows path handling... any clue? @westonpace @wjones127 @timsaucer

timsaucer:
> hmm this is quite strange, I don't see any change that could affect the windows path handling... any clue? @westonpace @wjones127 @timsaucer

I think I've seen this happen in CI in this repo before, but having trouble finding an exact example from my PRs. In the past rerunning CI has resolved it.

timsaucer:
These new failures I'm also seeing on PRs that are completely unrelated. example: https://github.com/lancedb/lance/actions/runs/18080818538/job/51443808660?pr=4834

westonpace:
> These new failures I'm also seeing on PRs that are completely unrelated. example: https://github.com/lancedb/lance/actions/runs/18080818538/job/51443808660?pr=4834

Agreed, the test_zonemap_zone_size test was fixed on main

westonpace:
I'll see if I can make sense of the Windows failures or at least see what changed

ddupg:
I rebased the main branch code and reran the tests multiple times. This error consistently occurs in Windows.
```
IO { source: Custom { kind: Other, error: Generic { store: "LocalFileSystem", source: InvalidUrl { url: Url { scheme: "file", cannot_be_a_base: false, username: "", password: None, host: None, port: None, path: "/C:%5CUsers%5CRUNNER~1%5CAppData%5CLocal%5CTemp%5C.tmpllPc01/some_file.lance", query: None, fragment: None } } } }, location: Location { file: "rust\\lance-file\\src\\v2\\writer.rs", line: 191, column: 9 } }
```

westonpace:
Tracked down the root cause at least.  Still working on fix.  The upgrade to DF caused a transitive dependency (`url`) to upgrade from 2.5.4 to 2.5.7 and this is the root of the failure.  My guess is we are hitting this change: https://github.com/servo/rust-url/pull/1060

Anyways, I'm sure there is just some magic incantation of path manipulation that has to happen.  My trial and error continues 🫠

Cargo.lock
```diff
@@ -8868,13 +8926,14 @@ dependencies = [
...
-version = "2.5.4"
+version = "2.5.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "32f8b686cadd1473f4bd0117a5d28d36b1ade384ea9b5069a1c40aefed7fda60"
+checksum = "08bc136a29a3d1758e07a9cca267be308aeebf5cfd5a10f3f67ab2097683ef5b"
 dependencies = [
  "form_urlencoded",
  "idna",
  "percent-encoding",
+ "serde",
```
timsaucer, line 8892: Since there are issues filed upstream in both object-store and url crates, can we simply revert these lines of the lock file? Or we can pin url in the cargo.toml to 2.5.4

westonpace: Datafusion asks for 2.5.7 specifically so we can't revert sadly :cry:

westonpace:
I've got a fix (more of a workaround) in progress that avoids these tricky paths by reworking how we do temporary directories: https://github.com/lancedb/lance/pull/4860

rust/lance-datafusion/src/planner.rs
ddupg: I saw the PR https://github.com/apache/datafusion/pull/16963/, `nulls_max` is the default behavior.

westonpace:
The Windows issue is fixed in https://github.com/lancedb/lance/pull/4860 if anyone is up for a review (I'm shamelessly :fishing_pole_and_fish: for reviewers)

westonpace:
Ok, merged.  If you rebase then hopefully you are good to go.

ddupg:
> Ok, merged. If you rebase then hopefully you are good to go.

Thank @westonpace very much for the fix, all the tests have passed after I rebased

ddupg:
Could @jackye1995 @westonpace @timsaucer please help review this?

jackye1995 (approved):
thank you for pushing this through!

Issue #4749, opened 2025-09-17
Add full text json index


wojiaodoubao:
Discussion could be found at https://github.com/lancedb/lance/issues/4516#issuecomment-3257233268

This is a sub-issue of https://github.com/lancedb/lance/issues/4516

wojiaodoubao:
## TODO
- [x] Basic framework of json inverted index. https://github.com/lancedb/lance/pull/4752
- [x] Support fuzziness for json inverted match query. https://github.com/lancedb/lance/pull/5048
- [x] Python create json fts index. https://github.com/lancedb/lance/pull/5020
- [x] Update document. https://github.com/lancedb/lance/pull/4865
- [ ] Support flat match query on json
- [ ] Encode path in triplet.
- ~~Support List(json) inverted index and search. https://github.com/lancedb/lance/pull/5029~~

PR #4752, opened 2025-09-17
feat: add full text json index

wojiaodoubao:
Close [#4749](https://github.com/lancedb/lance/issues/4749)


wojiaodoubao:
Hi @westonpace @BubbleCal , could you help review this when you have time, thanks very much !

rust/lance-index/src/scalar/inverted/tokenizer/json.rs
```diff
@@ -0,0 +1,326 @@
...
+        )]))
+    }
+}
+
+pub fn jsonb_to_json(col: &ArrayRef, col_name: &str) -> lance_core::Result<ArrayRef> {
+    let binary_array = col
+        .as_any()
+        .downcast_ref::<LargeBinaryArray>()
+        .unwrap_or_else(|| panic!("column {} is not a large binary array", col_name));
+    let mut builder = arrow_array::builder::StringBuilder::with_capacity(binary_array.len(), 1024);
```
BubbleCal: Should we use `LargeString`?
jsonb is stored as `LargeBinary` so the concern is we may be out of `i32` range with `String`

rust/lance-index/src/scalar/inverted/tokenizer/json.rs
```diff
@@ -0,0 +1,326 @@
...
+                out.push((prefix.to_string(), "str".to_string(), token.text.clone()));
+            }
+        }
+        _ => {
+            let value_type = match value {
+                Value::Null => "null",
+                Value::Bool(_) => "bool",
+                Value::Number(n) if n.is_i64() => "i64",
+                Value::Number(n) if n.is_f64() => "f64",
+                Value::Number(n) if n.is_u64() => "u64",
```
BubbleCal: how does these `is_xxx()` work?
it seems hard for users to know the type if they are storing some mixed data, can we map all of these to "number"?

rust/lance/src/io/exec/fts.rs
```diff
@@ -463,6 +430,37 @@ impl ExecutionPlan for FlatMatchQueryExec {
...
+    match field.data_type() {
+        DataType::Utf8 | DataType::LargeUtf8 => Ok(input),
+        DataType::List(field)
+            if matches!(field.data_type(), DataType::Utf8 | DataType::LargeUtf8) =>
+        {
+            Ok(input)
+        }
+        DataType::Binary | DataType::LargeBinary => {
+            match field.metadata().get(ARROW_EXT_NAME_KEY) {
+                Some(name) if name == &JSON_EXT_NAME.to_string() => {
```
BubbleCal: we can compare `name.as_str()` with `JSON_EXT_NAME` to avoid calling `to_string()`?

rust/lance-index/src/scalar/inverted/tokenizer/json.rs
wojiaodoubao: Yes, "number" is better. Let me fix it.

About the behavior, I test with an example, the parser rule seems to be:
1. If the number is float, use f64.
2. If the number is integer and within the range of i64, use i64.
3. If the number is integer and exceed the scope of i64 but within the range of u64, use u64.
4. If the number is integer and exceed the scope of both i64 and u64, use f64 and loss of precision.

rust/lance-index/src/scalar/inverted/tokenizer/json.rs
```diff
@@ -0,0 +1,325 @@
...
+/// Example:
+/// text: "title,str,harrypotter;title,str,chapter;title,str,one"
+/// triplets:
+///   - "title,str,harrypotter"
+///   - "title,str,chapter"
+///   - "title,str,one"
+#[derive(Clone)]
+pub struct TripletTokenizer;
+
+impl Tokenizer for TripletTokenizer {
```
BubbleCal: A question about this, can this work with the existing tokenizer params?
- would `max_token_length` count the field name and type? I think we won't expect that
- would `stemming` still work? say we expect "title,str,searches" -> "title,str,search"
- would `lower_case` work? say `Title,str,Search` -> `Title,str,search`, it shouldn't lowercase the field name
- would `remove_stop_words` work? say `title,str,the` should be removed

It would be nice if we can add more tests covering these cases

rust/lance/src/dataset.rs
```diff
@@ -7694,6 +7696,237 @@ mod tests {
...
+            .try_into_batch()
+            .await
+            .unwrap();
+        assert_eq!(2, batch.num_rows());
+
+        // Test phrase query
+        let query = FullTextSearchQuery {
+            query: FtsQuery::Phrase(
+                PhraseQuery::new(
+                    "title,str,harrypotter;title,str,one;title,str,chapter".to_string(),
```
BubbleCal: We can support syntax like "title,str,harrypotter one chapter" in the future

rust/lance/src/dataset.rs
```diff
@@ -7694,6 +7696,237 @@ mod tests {
...
+            wand_factor: None,
+        };
+        let batch = dataset
+            .scan()
+            .full_text_search(query)
+            .unwrap()
+            .try_into_batch()
+            .await
+            .unwrap();
+        assert_eq!(0, batch.num_rows());
```
BubbleCal, line 7935: i think there should be one row hit, because the index param specifies `lower_case: True`

rust/lance-index/src/scalar/inverted/tokenizer/json.rs
wojiaodoubao: These tokenize params can work with TripletTokenizer. That is because params like stemmer, lowercase, max_token_length and remove_stop_words are handled by ```TokenFilter```. ```Tokenizer``` only need focus on splitting tokens.

But the TokenFilter will take ```field_name```, ```type``` and ```value``` as a whole. For example:
* ```max_token_length``` would count the filed name and type.
* "title,str,searches" -> "title,str,search" won't work.
* ```lower_case``` will work.
* ```title,str,the``` won't be removed.

If we want the tokenizer parameters to take effect only in the value part, we need to implement a custom ```TokenFilter``` for triplet tokens.

rust/lance/src/dataset.rs
wojiaodoubao: This case is testing "wrong sequence should not hit" in phrase query. The text ```"Title": "HarryPotter Chapter One"``` has been transformed into ```"title,str,harrypotter;title,str,chapter;title,str,one"```. The first case is a wrong sequence ```"title,str,harrypotter;title,str,one;title,str,chapter"```, and expect an empty result. The second case is the right sequence and expect one row hit (https://github.com/wojiaodoubao/lance/blob/json-fts/rust/lance/src/dataset.rs#L7817).

rust/lance-index/src/scalar/inverted/tokenizer/json.rs
BubbleCal: could you check how tantivy handle this? we are using tantivy's tokenizer and they support JSON, so probably we can make it work without introducing too many changes.

I'm just thinking, is it possible to retrieve the original triplet by `offset_from` and `offset_to`, and set `text` to the text content excluding field name and type? Then it can work with the existing `TokenFilter`s

java/lance-jni/src/blocking_dataset.rs
```diff
@@ -635,7 +635,8 @@ fn inner_create_index(
         | IndexType::Inverted
         | IndexType::NGram
         | IndexType::ZoneMap
-        | IndexType::BloomFilter => {
+        | IndexType::BloomFilter
+        | IndexType::Json => {
```
westonpace: We already have a scalar index named json.  I wonder if we can name this one FtsJson or InvertedJson?

rust/lance-index/src/scalar/json.rs
```diff
@@ -779,40 +784,72 @@ impl ScalarIndexPlugin for JsonIndexPlugin {
...
-
-        let index_details = crate::pb::JsonIndexDetails {
-            path,
-            target_details: Some(target_index.index_details),
-        };
-        Ok(CreatedIndex {
-            index_details: prost_types::Any::from_msg(&index_details)?,
-            index_version: JSON_INDEX_VERSION,
-        })
+        match request.parameters.target_index_type.as_str() {
```
westonpace: Can you make a new plugin for this instead of sharing the plugin?  These are two rather different index types.

rust/lance/src/index/scalar.rs
```diff
@@ -256,6 +257,34 @@ impl IndexDetails {
             self.get_plugin().map(|p| p.version())
         }
     }
+
+    pub fn index_type(&self) -> IndexType {
```
westonpace: What is this method needed for?  I'd like to move away from these kinds of switch statements in indexes.

westonpace (commented):
This is a cool idea.  Can we make it so that it is either a separate plugin or a special kind of inverted index instead of being a special kind of JSON index?  The existing JSON index is for filtering and not search and I think it will get confusing having the two share the same plugin.

rust/lance-index/src/scalar/inverted/tokenizer/json.rs
wojiaodoubao: Sorry for my late response. I studied how Tantivy handles JSON index. In Tantivy, the JSON type does not require a special TokenFilter, it shares the same TokenFilter as other types. Tantivy achieves this simplicity by relying on its Term structure, which is the core structure of both the inverted index and full text search. The Term decouples tokenization and index matching.
A Term consists of metadata and data. For JSON types, the structure of a Term is: meta + path_id + type + token.text, where token.text is the tokenized result of the value part in a JSON triplet.
For example, given the JSON text {"a": "hello world"}, Tantivy only applies tokenization to the value part, i.e., "hello world", and then constructs two terms: <meta, id_of_path_a, str, hello> and <meta, id_of_path_a, str, world>.

In Lance, we build full-text indexes and queries based on tokens, so the current implementation fully encodes the triplet into the token. We can't only use the value part like in tantivy, because this way we will lose the path and type part in train and search. However, the path and type in the triplet prevent us from reusing the TokenFilter.

I think we don't need to introduce ```Term```, as that would involve significant changes. I think we could abstract a layer over ```tantivy::tokenizer::TextAnalyzer``` named ```LanceTokenizer``` in ```InvertedIndex```, and providing an implementation something like ```JsonLanceTokenizer``` to encapsulate the details of triplet processing. This way, the current ```InvertedIndex``` wouldn't require major modifications and could continue to train and search based on tokens. The ```LanceTokenizer``` would be responsible for correctly splitting tokens for different types.
Another insight from Tantivy is assigning IDs to paths in JSON full-text indexing, which can significantly reduce the storage overhead of redundant paths. We also need to implement this, but it would introduce a new structure in the InvertedIndex to store the mapping between paths and IDs. I think this can be addressed as the second part of work.

@BubbleCal @westonpace what do you think?

rust/lance-index/src/scalar/json.rs
wojiaodoubao: Your suggestion makes sense to me, let me fix it.

wojiaodoubao:
I updated the patch. In the new implementation, the tokenizer is used to handle the creation/querying of full-text for **different types** of documents. A new abstraction, LanceTokenizer, is introduced to address the tokenization different types, including JSON type. As a result, JSON full-text indexing can be treated as a use case of the Inverted Index, without requiring modifications to the JsonIndexPlugin or introducing a separate full-text index type specifically for JSON.

Hi @BubbleCal @westonpace , please help review when you have time, thanks very much!

The test failures are unrelated, caused by https://github.com/lancedb/lance/issues/4829. I'll re-trigger ci after it is fixed.

BubbleCal (approved):
LGTM, any concerns? cc @westonpace

jackye1995:
@wojiaodoubao could you also update the documentation https://github.com/lancedb/lance/blob/main/docs/src/format/table/index/scalar/fts.md accordingly later?

wojiaodoubao:
> @wojiaodoubao could you also update the documentation https://github.com/lancedb/lance/blob/main/docs/src/format/table/index/scalar/fts.md accordingly later?

Hi @jackye1995 , thanks your reminding!!! I'll update the documentation later. I thin maybe I can update the doc in a separate pr,  for a full dicsussion.

jackye1995:
Merging this since there are a few PRs depending on this, @westonpace let us know if you have any other comments we can address separately.

Issue #4337, opened 2025-07-29
How to get a snapshot list with some summary properties


majin1102:
The basic motivation is to integrate a lakehouse management platform like:

<img width="3260" height="1312" alt="Image" src="https://github.com/user-attachments/assets/3d739d16-a2a8-4032-a6f6-4706c7d5564b" />

The page above comes from Apache(incubating) Amoro, which is a multi-format lakehouse management system. I have raised a discussion to integrate lance format and got considerable feedback: https://github.com/apache/amoro/discussions/3668

In production scenarios, when users want to explore a table/dataset, Snapshot list with these basic summary would help them to:
1. Get the basic writing workloads on this table/dataset(frequency, volume, sizes .etc)
2. Quickly get the target tag/version to rollback
3. Know the trends
4. Know the health

I think snapshot/version list is a basic demand for lakehouse management.

### Iceberg
For Iceberg, the snapshot list is stored in metadata file with all necessary summary information in it. When requesting a page of snapshot list, that's O(1) cost for reading metadata file:

<img width="770" height="480" alt="Image" src="https://github.com/user-attachments/assets/376bafe6-4cec-4638-a6b1-b8c7cf8ebcf4" />

### Paimon
Paimon does't have a root metadata file and use a snapshot file to store summaries for each snapshot:
https://paimon.apache.org/docs/0.8/concepts/specification/#snapshot

The listing snapshot operation loads each snapshot file sequentially which would lead a bad experience for snapshot pages. Amoro optimized this process by using a parallelism way. 

### Delta

Delta desn't have a root metadata file too. It uses transaction json files and checkpoint files to store snapshots. Amoro doesn't have the plan to integrate Delta yet.

### Lance
For lance, I think it is more similar to Paimon. And there's a list version operation in dataset:
```
pub struct Version {
    /// version number
    pub version: u64,

    /// Timestamp of dataset creation in UTC.
    pub timestamp: DateTime<Utc>,

    /// Key-value pairs of metadata.
    pub metadata: BTreeMap<String, String>,
}

/// Convert Manifest to Data Version.
impl From<&Manifest> for Version {
    fn from(m: &Manifest) -> Self {
        Self {
            version: m.version,
            timestamp: m.timestamp(),
            metadata: BTreeMap::default(),
        }
    }
}
```

However, there are two concerns for this listing version operation:
[...]

Issue #4308, opened 2025-07-24
Inline Transaction File


jackye1995:
As a part of improving performance for small commits like WAL appends, writing the transaction file becomes an expensive operation, and could be inlined in the manifest file if the transaction content is small.

wjones127:
Duplicate of #3487

jackye1995:
Closing as duplicate

PR #4754, opened 2025-09-17
feat: support manifeset summary and get it from Version

majin1102:
Discussion: #4337 

Still we can't get transaction information directly from manifest. This will be enabled after feature #4308


rust/lance-table/src/format/manifest.rs
```diff
@@ -437,6 +446,73 @@ impl Manifest {
...
+    /// Get the summary information of a manifest.
+    ///
+    /// This function calculates various statistics about the manifest, including:
+    /// - total-records: Total number of records in the dataset
+    /// - total-files-size: Total size of all data files in bytes
+    /// - total-fragments: Number of fragments in the dataset
+    /// - total-data-files: Total number of data files across all fragments
+    /// - total-deletions: Total number of deleted records
+    /// - total-deletion-files: Number of fragments with deletion files
+    pub fn summary(&self) -> BTreeMap<String, String> {
```
wjones127: Why output as a map of strings? Then the caller has to parse the integers. Seems like it would be better to just expose the `ManifestStats` as public, right?

majin1102: I was planning to extend the summary to transaction as #4337 metioned. For that part we may need string. And after that I probobly use the stats in Version struct which already have a map of string（the final usage I think accepts string more friendly）.

I think you are right we could expose ManifestStats and then transform to map<string, string> for Version struct. That should be more flexible.

majin1102: @wjones127 PTAL when you have time

rust/lance-table/src/format/manifest.rs
```diff
@@ -117,6 +117,44 @@ fn compute_fragment_offsets(fragments: &[Fragment]) -> Vec<usize> {
...
+    total_files_size: u64,
+    total_deletion_files: u64,
+    total_deletions: u64,
+}
+
+impl From<ManifestStats> for BTreeMap<String, String> {
+    fn from(val: ManifestStats) -> Self {
+        let mut stats_map = Self::new();
+        stats_map.insert(
+            "total-fragments".to_string(),
```
wjones127: kebab case does seem like an odd choice. I'm guessing this is the match the iceberg statistics?

wjones127 (approved):
Seems good. I was thinking we'd want these kinds of stats easily available anyways. Thanks!

rust/lance-table/src/format/manifest.rs
majin1102: > kebab case does seem like an odd choice. I'm guessing this is the match the iceberg statistics?

The initial idea was inspired by Iceberg, and we utilized the summary for management functions. However, matching the naming convention wasn't intentional; I just didn't give it much thought. 

I've switched it to snake_case. Appreciate you catching that!

rust/lance-table/src/format/manifest.rs
```diff
@@ -437,6 +475,50 @@ impl Manifest {
...
+        // Calculate total fragments
+        let mut summary = self
+            .fragments
+            .iter()
+            .fold(ManifestSummary::default(), |mut summary, f| {
+                // Count data files in the current fragment
+                summary.total_data_files += f.files.len() as u64;
+                // Sum the number of rows for the current fragment (if available)
+                if let Some(num_rows) = f.num_rows() {
+                    summary.total_records += num_rows as u64;
```
jackye1995: I think it's better to say `total-data-file-rows`, `total-deletion-file-rows` instead, and we can have a `total-rows` that is `total-data-file-rows - total-deletion-file-rows`

majin1102: Oh, you prefer kebab case or snake case? I'm OK with both

jackye1995: Oh I just copied it from the docstring, looks like you need to update them to be consistent.

majin1102: It's done. 

Feel free to let me know if you have any other thoughts

rust/lance-table/src/format/manifest.rs
```diff
@@ -437,6 +480,51 @@ impl Manifest {
     pub fn should_use_legacy_format(&self) -> bool {
         self.data_storage_format.version == LEGACY_FORMAT_VERSION
     }
+
+    /// Get the summary information of a manifest.
+    ///
+    /// This function calculates various statistics about the manifest, including:
+    /// - total-records: Total number of records in the dataset
```
jackye1995: looks like these docs are still not updated to snake case?

majin1102: sorry，my miss

majin1102: PTAL and feel free to merge if no other comments

PR #4760, opened 2025-09-17
refactor: rename lance_table::format::Index to lance_table::format::IndexMetadata

westonpace:
The proto files have a message which is `IndexMetadata` which is stored in the manifest.  In the rust code this was represented by a struct `Index`.  Perhaps I am too simple but this has been very confusing for me as there is also a trait `Index` in `lance_index`.  In several places we had aliased `Index` to `IndexMetadata` already.  This PR goes ahead and performs the rename.  We now match the protobuf message (like we do for all of our other protobuf messages) and there is no confusion with the trait.

To avoid a hard breaking change I have left in the old name but marked it deprecated.


wjones127 (approved):
Thank you!

PR #4764, opened 2025-09-17
feat: exposing option to provide a serialized manifest to the Java SDK

morales-t-netflix:
Creates bindings for a user to provide a serialized manifest in the Java SDK (using the same methodology as the Python SDK)


java/lance-jni/Cargo.lock
jackye1995, line 1: this seems to be an unintentional change?

morales-t-netflix: Ah - yeah, this showed up when I built the java module. Just removed!

PR #4765, opened 2025-09-18
feat: add multi-path support for lance data paths

### LINKED ISSUE #4702: Support multi bucket layout
Currently entire lance dataset needs to be in one bucket. We should add a multi bucket layout support. That way indices, manifest and data can be stored in separate locations (maybe separate filesystems)

jaystarshot:
Fixes #4702 

Supports multi path support for lance datasets.
sample api shown
### How to Use It

**Creating a multi-path dataset:**
```python
# Set up multiple storage locations and pick where to write
dataset = lance.write_dataset(
    data,
    "s3://primary/dataset",
    mode="create",
    initial_data_paths=["s3://bucket1/data", "s3://bucket2/data"],
    target_data_paths=["s3://bucket1/data"] 
)
```

**Appending to a different location:**
```python
# Later, append new data to bucket2 instead
lance.write_dataset(
    new_data,
    dataset,
    mode="append",
    target_data_paths=["s3://bucket2/data"] 
)
```

**Overwriting with completely new paths:**
```python
# Replace everything - both the path registry and the data
lance.write_dataset(
    replacement_data,
    "s3://primary/dataset", 
    mode="overwrite",
    initial_data_paths=["s3://new-bucket1/data", "s3://new-bucket2/data"],
    target_data_paths=["s3://new-bucket1/data"]
)
```

Main Changes
1. Modified overwrite transaction to add the bases which will be written to the manifest.
2. Fixed handling around object stores when using multi bases. This was because previously we were using the primary dataset's object store to write / read from multiple bases.

WIP and unclear
1. Compaction (and possibly other) operation handling in a multi path scenario
2. Check if Conflict with git like operations
3. Adding new paths to a dataset possibly in new operations
4. Currently storage options only allow us to specify storage options for one cloud provider but we might need storage options per path especially if we want to support multi cloud setup etc

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -156,7 +156,7 @@ jobs:
           ALL_FEATURES=`cargo metadata --format-version=1 --no-deps | jq -r '.packages[] | .features | keys | .[]' | grep -v protoc | sort | uniq | paste -s -d "," -`
           cargo test --locked --features ${ALL_FEATURES}
   build-no-lock:
-    runs-on: warp-ubuntu-2404-x64-4x
+    runs-on: warp-ubuntu-2404-x64-8x
     timeout-minutes: 30
     env:
       # Need up-to-date compilers for kernels


python/python/lance/dataset.py
```diff
@@ -4774,6 +4775,15 @@ def write_dataset(
         and can be retrieved using read_transaction().
         If both `commit_message` and `properties` are provided, `commit_message` will
         override any "lance.commit.message" key in `properties`.
+    data_bucket_uris: list of str, optional
+        *Experimental*. Additional bucket URIs for data file distribution.
```
jackye1995: I think we don't need to say experimental here, but we might want to rename this feature flag to something more generic like `FLAG_BASE_PATHS`: https://github.com/lancedb/lance/pull/4552

python/python/lance/dataset.py
```diff
@@ -4696,6 +4696,7 @@ def write_dataset(
     auto_cleanup_options: Optional[AutoCleanupConfig] = None,
     commit_message: Optional[str] = None,
     transaction_properties: Optional[Dict[str, str]] = None,
+    data_bucket_uris: Optional[List[str]] = None,
```
jackye1995: I feel "bucket" is a term very specific to S3, can we use something more generic

python/src/dataset.rs
```diff
@@ -2507,6 +2507,14 @@ pub fn get_write_params(options: &Bound<'_, PyDict>) -> PyResult<Option<WritePar
 
         p.commit_handler = get_commit_handler(options)?;
 
+        // Handle data bucket URIs
+        if let Some(data_bucket_uris) = get_dict_opt::<Vec<String>>(options, "data_bucket_uris")? {
+            println!("🐍 Python -> Rust: data_bucket_uris = {:?}", data_bucket_uris);
```
jackye1995: remove prints

rust/lance-table/src/format/manifest.rs
```diff
@@ -944,3 +944,133 @@ mod tests {
...
+    fn hash_fragment_id(&self, fragment_id: u32) -> u64 {
+        // Use a simple multiplicative hash that works well for sequential IDs
+        // This provides better distribution than DefaultHasher for small sequential numbers
+        let hash = (fragment_id as u64).wrapping_mul(0x9e3779b97f4a7c15);
+        hash
+    }
+
+    /// Check if multi-bucket layout is enabled (more than one bucket)
+    pub fn is_multi_bucket_enabled(&self) -> bool {
+        self.base_paths.len() > 1
```
jackye1995: this won't be true if you are writing the data first time with these additional buckets. In that case, I feel we should not put this config at the time of `write_dataset`. I propose that we should expose an API for users to set the base paths, like `dataset.base_paths.add/delete` (similar to how we manage tags), and then we add an additional property of `write_data_files=true/false` to indicate if the path should be used for writing. What do you think?

test_multi_bucket_logging.py
jackye1995, line 1: let's turn this to proper tests in python and rust.

jackye1995 (commented):
This is exciting feature! Added some initial thoughts

jaystarshot:
Thanks @jackye1995 for the review! I will incorporate your suggestions and try to add tests in all current flows too!

rust/lance-table/src/format/manifest.rs
jaystarshot: Yeah makes sense, or we can do both?

jackye1995: If we set it as a property if the table already, why do you still need to set it again when writing?

jaystarshot: There may be scenarios where, once a bucket becomes overloaded, we need to add additional data requiring a new bucket and an appropriate routing policy. (cloud providers have bucket level SLAs/ QPS limits)

Similarly, if a bucket must be migrated across regions or between cloud providers, that would also require different APIs.

jackye1995: I see. In that case, I feel we should separate the concepts of available buckets vs buckets to use for write.

For available buckets, those should be registered in the table. The key reason of that is that write eventually translates to a commit, and if you want to both commit the write + add the new bases, it would require additional field in all the write transaction model, and also check conflicts with all other transaction types for this new field. It is doable, but feels like a lot of work to make this feature work.

Compared to that, it seems like there are 2 personas here, the admin and the writer. I imagine that the admin can register the available paths to the table and commit this as a simple operation:

```
dataset.add_base("s3://bucket/abc", name="primary")
dataset.add_base("s3://bucket3/abc", name="us-west-2")
dataset.add_base("s3://bucket3/abc", name="eu-west-2")
```

When writing, the writer can specify a subset of them for write to keep things flexible:

```
dataset.write_data(..., base_paths=["primary", "us-west-2"])
```

what do you think?

jackye1995:
posting some offline discussions here:

1. we should allow directly adding new bases as a part of the dataset creation process. Related fields should be added to the Overwrite transaction model. This should be a simple change since Overwrite conflicts with everything.
2. after dataset is created, we should allow new bases to be added through API like `dataset.add_base(...)`
3. for update/delete/insert, if user wants to add new bases, it can do 2 commits internally, 1 to add_base, 1 to do the actual write. This avoids the complexity to add base information to the other write transaction models and handle their concurrent issues. It is similar to how we reserve fragment IDs as a separated commit during compaction.

python/python/tests/test_multi_bucket.py
```diff
@@ -0,0 +1,313 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright The Lance Authors
+
+"""
+Tests for multi-bucket dataset functionality.
```
jackye1995: Can we change the general naming to "multi-base"? Because bucket is a very S3 specific thing and we want to use Lance terminologies to be generic

python/python/lance/dataset.py
```diff
@@ -4774,6 +4776,19 @@ def write_dataset(
         and can be retrieved using read_transaction().
         If both `commit_message` and `properties` are provided, `commit_message` will
         override any "lance.commit.message" key in `properties`.
+    data_path_uris: list of str, optional
+        *Experimental*. Additional bucket URIs for registering in the manifest
+        during dataset creation. When provided in CREATE mode, these bucket URIs
+        will be registered in the manifest as available buckets for future use.
+        Only used in CREATE mode for manifest registration.
+        Example: ["s3://bucket2/data", "s3://bucket3/data"]
+    target_path_uri: str, optional
```
jackye1995: I think this should also be a list instead of just a single value

jaystarshot: If its a list we need to add a target selector (hash etc) during write and i was thinking that could be an extension PR?

jaystarshot: or maybe its easy to add

python/python/lance/dataset.py
```diff
@@ -4774,6 +4776,19 @@ def write_dataset(
         and can be retrieved using read_transaction().
         If both `commit_message` and `properties` are provided, `commit_message` will
         override any "lance.commit.message" key in `properties`.
+    data_path_uris: list of str, optional
```
jackye1995: Can we use the term "data_file_bases" and "initial_data_file_bases"? I think in general we want to use the "base" terminology to be consistent, and "initial" makes it clear that the field is dedicated to create mode. 

Also we should define the input types a bit better. I think we should accept a list of string for the "data_file_bases" so that it means the list of base names. And we should accept a list of BasePath objects (name + path) for the "initial_data_file_bases" so those get proper names to be referenced later. What do you think?

python/python/lance/dataset.py
jackye1995: You mean do a hash distribution to write to each base? I think we can say we only support one item in the list initially, but the API should be set to list so we can extend in the future

python/src/dataset.rs
```diff
@@ -2507,6 +2507,21 @@ pub fn get_write_params(options: &Bound<'_, PyDict>) -> PyResult<Option<WritePar
...
+            p.data_path_uris = Some(data_path_uris);
+        } else {
+            println!("🐍 Python -> Rust: no data_path_uris provided");
+        }
+
+        if let Some(target_path_uri) = get_dict_opt::<String>(options, "target_path_uri")? {
+            println!("🐍 Python -> Rust: target_path_uri = {:?}", target_path_uri);
+            p.target_path_uri = Some(target_path_uri);
+        } else {
+            println!("🐍 Python -> Rust: no target_path_uri provided");
```
jackye1995: Nit: tests should not have prints

python/python/lance/dataset.py
```diff
@@ -4975,7 +4977,24 @@ def write_dataset(
         and can be retrieved using read_transaction().
         If both `commit_message` and `properties` are provided, `commit_message` will
         override any "lance.commit.message" key in `properties`.
+    initial_data_paths: list of str, optional
+        *Experimental*. Data file base URIs for registering in the manifest
```
jackye1995: nit: no need to mark as experimental

python/python/lance/dataset.py
```diff
@@ -4975,7 +4977,24 @@ def write_dataset(
...
+        during dataset creation. When provided in CREATE mode, these URIs
+        will be registered in the manifest as available data file bases for future use.
+        Only used in CREATE mode for manifest registration.
+        Example: ["s3://storage1/data", "s3://storage2/data"]
+    target_data_paths: list of str, optional
+        *Experimental*. Target URI for writing data files (array with exactly one element).
```
jackye1995: nit: no need to mark as experimental

rust/lance/src/dataset.rs
```diff
@@ -1376,6 +1389,37 @@ impl Dataset {
         }
     }
 
+    /// Get the ObjectStore for a specific path based on base_id
+    pub(crate) fn get_object_store_for_path(&self, base_id: Option<&u32>) -> Result<Arc<lance_io::object_store::ObjectStore>> {
```
jackye1995: I guess this would be necessary for multi-cloud use case, but apart from that, would the configuration within the same cloud be different? We cannot really pass different configurations for the same cloud, so I think we should cache object stores by the path scheme, instead of by the path. This will reduce the number of object stores we cache per dataset, especially if you need to set a lot of bases.

jaystarshot: Yes we need this for each bucket since object store has the bucket info in the constructor , but that could be changed i guess. 
Else we write to the same bucket as the primary. The cache key here is the bucket uri, so there will just be one entry per bucket.

jackye1995: The key thing here I think is, do you expect to pass in different object store configs per data path? Today these will just be initialized with whatever the credentials and settings is in the environment or you pass in when opening the dataset. They won't really be different object stores from configuration perspective. If you need to customize at that level, we need to probably find a way to pass that info along the way here

jaystarshot: Not for the current use case but i guess in general it could be nice to have in future.

jackye1995: Sounds good. But in general, we already have a cache in ObjectStoreRegistry, we should use that instead of creating another cache here. That cache already uses base path as the cache key as well.

jaystarshot: @jackye1995 we need the object store cache at the dataset level (like the primary object store is stored [here](https://github.com/lancedb/lance/blob/8027b5c9fece8dd078b842e6d70cf8f75184855a/rust/lance/src/dataset.rs#L141)

This is used by the reader to fetch the correct object store (using **get_object_store_for_path**)
Are you suggesting to store the ObjectStoreRegistry here?

jackye1995: Sorry let me be clear, I mean in the dataset builder, we should not just initialize these extra object stores. It is not always gonna be used, and it is not a cheap struct to initialize. We should try to initialize it on demand when we need to use it.

protos/transaction.proto
```diff
@@ -68,6 +68,8 @@ message Transaction {
     map<string, bytes> schema_metadata = 3;
     // Key-value pairs to merge with existing config.
     map<string, string> config_upsert_values = 4;
+    // Additional path URIs for data file distribution in multi-path layouts
+    repeated string initial_data_paths = 5;
```
jackye1995: Regardless of what we do in the Python API, at transaction level we should have a list of BasePath , not just string data paths. In BasePath we have important fields like is_dataset_root, as well as a user provided name of the path.

majin1102:
> 3\. It is similar to how we reserve fragment IDs as a separated commit during compaction.

OK, found the context!

I think there is difference with the two cases: reserving fragment IDs does not add content into manifest right? It's an updating field operation. The transactional issue would bring unused paths(might even exists). Even if we have multi-statement transaction, I think we might not garentee people use it correctly and might result in an bloated manifest file for a long lifetime dataset.

Maybe we could add cleaning unrelated base_paths in the cleanup procedure for this. Or make this a table config like cleaning unrelated paths during latest n snapshots. And of course this is optional in case users want to manage it manually. This could also guard the files deletion/compaction cases.

What do you think? @jackye1995 @dacort

jackye1995:
>  The transactional issue would bring unused paths(might even exists). Even if we have multi-statement transaction, I think we might not garentee people use it correctly and might result in an bloated manifest file for a long lifetime dataset.

can you elaborate a bit more? I did not get the concern here. 

I think it is important for users to assign a logical name for the path for deduping purpose, so that if you try to add paths of the same name, it should fail.

python/python/lance/dataset.py
```diff
@@ -4975,7 +4977,24 @@ def write_dataset(
         and can be retrieved using read_transaction().
         If both `commit_message` and `properties` are provided, `commit_message` will
         override any "lance.commit.message" key in `properties`.
+    initial_data_paths: list of str, optional
```
jackye1995: I was thinking about this last night, I feel there is an issue between simple user experience and flexibility and correctness here:

In the most precise way, what user should define is the full base path information, something like:

```python
write_dataset(data, dataset, mode="create", initial_bases=[
  {"name": "b1", "path": "s3://a/b1", "is_dataset_root": "true"},
{"name": "b2", "path": "s3://a/b2", "is_dataset_root": "false"},
])

# target existing bases
write_dataset(data2, dataset, mode="append", target_bases=[
  "b1", 
  "b2"
])

# target new bases
write_dataset(data2, dataset, mode="append", target_bases=[
  {"name": "b3", "path": "s3://a/b3", "is_dataset_root": "true"},
])
```

but I also understand this is quite cumbersome to write, comparing to directly write `initial_data_paths` and `target_data_paths`. One big concern I have around that is people write paths slightly differently all the time (e.g. `s3://a/b1` vs `s3://a/b1/`). It is hard to know if they mean the same one or they actually mean different paths.

We also have a similar problem about how expressive we should be in the Overwrite transaction, as I commented above.

So here is my latest thinking: at the user interface level, instead of having a dedicated field only for create mode, we just fully separate the concept of the new bases to register and the target bases to use. Regardless of create mode or not, user always register new bases in the field `new_bases`, and specify what to use in `target_bases`. Here is the updated example:

```python
write_dataset(data, dataset, mode="create", new_bases=[
  {"name": "b1", "path": "s3://a/b1", "is_dataset_root": "true"},
{"name": "b2", "path": "s3://a/b2", "is_dataset_root": "false"},
],
target_bases=["b1", "b2"])

# target existing bases
write_dataset(data2, dataset, mode="append", target_bases=[
  "b1", 
  "b2"
])

# target new bases
write_dataset(data2, dataset, mode="append", new_bases=[
  {"name": "b3", "path": "s3://a/b3", "is_dataset_root": "true"},
], target_bases=["b3"])
```

And at transaction level, the `new_bases` directly translate to the new bases we should add in the transaction model, for example in Overwrite:

```proto
  // Create or overwrite the entire dataset.
  message Overwrite {
    // The new fragments
    //
    // Fragment IDs are not yet assigned.
    repeated DataFragment fragments = 1;
    // The new schema
    repeated lance.file.Field schema = 2;
    // Schema metadata.
    map<string, bytes> schema_metadata = 3;
    // Key-value pairs to merge with existing config.
    map<string, string> config_upsert_values = 4;
    //  base paths for the whole dataset
    repeated BasePath initial_bases = 5;
}
```

@majin1102 @jaystarshot what do we think about this proposal?

jaystarshot: I think we can define the api for the **overwrite** behavior once we have concrete use cases. For us just having a separate API is fine but there could be other use cases.

rust/lance/src/dataset.rs
jaystarshot:
```
/// Additional path URIs for multi-path support (cached in ObjectStoreRegistry)
    pub(crate) extra_path_uris: Arc<HashMap<u32, String>>,
    
    /// Store parameters used to create object stores (needed for extra paths)
    pub(crate) store_params: ObjectStoreParams,
```

I will need to store these in dataset so that we can create the object store on demand

rust/lance/src/dataset.rs
```diff
@@ -1397,24 +1382,31 @@ impl Dataset {
...
                 // Get the object store from the registry (which caches it)
-                // Using the same params as the primary store for consistency
                 let (store, _) = lance_io::object_store::ObjectStore::from_uri_and_params(
                     self.session.store_registry(),
-                    path_uri,
-                    &self.extra_store_params,
-                ).await?;
-                
+                    &bucket_uri,
+                    &ObjectStoreParams::default(),
```
jaystarshot: The default params likely won’t work when storage params are specified ( gcs/oci).

majin1102:
> > The transactional issue would bring unused paths(might even exists). Even if we have multi-statement transaction, I think we might not garentee people use it correctly and might result in an bloated manifest file for a long lifetime dataset.
> 
> can you elaborate a bit more? I did not get the concern here.
> 
> I think it is important for users to assign a logical name for the path for deduping purpose, so that if you try to add paths of the same name, it should fail.

Sorry for the late reply.

I was thinking that if the scenario was writing files to a none added base. This should potentially be an ACID operation. But on second thought this was reasonable to split this into two operations. One to add base, one to write data, with a little limitation to the scenario if people want to dynamically manage bases according to data.

Two cases in my mental mind:
1. The base paths are static, just added once for all.
2. The base paths are added according to the written data, by some field value or even UUID. This is somehow like partition usage.

For the second scenario, I assumed we have to add the base before commit the data. Then we might encounter the case that data commit failed but base added. For this secnario we could expect a commit retry but without assurance. Then the base could be left there forever(if we use UUID then it is just a rubbish hole).

I think the multi-base management could be a complex issue and somehow I think is related to the scenario of partitions(like this multi-bucket case). Or let's say it might provide a basement that partitions aims for. Nowdays when we talk about partitions we usually thought of partition spec balabala in Iceberg, but if we looked back to Hive, the partitions are just managed paths. I was thinking if we could reuse this multi-base ability in partition solution.  What do you think of this? @dacort @jackye1995 

To be clear I'm on the side of not designing multi-base as heavy as partitions. But I think if we considered there are common things between them we could draw lessons from APIs and consider how the upper layers will evolve towards a partition-oriented solution based on the multi-base architecture in the future

majin1102:
> I think it is important for users to assign a logical name for the path for deduping purpose, so that if you try to add paths of the same name, it should fail.

I didn't get this. I think the path itself is identical and could be used deduping purpose?  if we force people to use a name as identifier  I might get something from the path which would be a little unnecessary.

python/python/lance/dataset.py
majin1102: Agree with the proposal of putting them together in one operation.

I was wondering if it is that necessary of registering unrelated base paths in write interface. If we make adding base path idempotent I think we could just:

```
write_dataset(data2, dataset, mode="append", bases=[{"path": "s3://a/b3"}])
```

This means name is none(if we don't force there must be one) and the is_dataset_root is default.

Or we could use:
```
write_dataset(data2, dataset, mode="append", bases=[{"name":"b3", "path": "s3://a/b3", "is_dataset_root":"false"}])
```

In the runtime we check if the base path exists and make sure it is added before commiting the data.

PR #4766, opened 2025-09-18
fix: fts boolean query and boost query do not support phrase

wojiaodoubao:


wojiaodoubao:
Hi @BubbleCal , could you help review this when you have time, thanks very much !

wojiaodoubao:
> can we just set `params = params.with_phrase_slop(Some(phrase.slop));` in https://github.com/lancedb/lance/blob/main/rust/lance/src/io/exec/fts.rs#L502 ?
> 
> i think the 2 solutions are the same but the above one can be simpler

Thanks, you are right ! Let me fix it.

BubbleCal (approved):
Thanks for fix!

PR #4769, opened 2025-09-18
feat(java): support blob api

### LINKED ISSUE #4707: Java: Support Blob API
With https://github.com/lancedb/lance-spark/pull/89 user will be able to read blob metadata using Spark, and typically the next step is to use the position and size to fetch the actual blobs. For this, Java is missing the blob API that Python has to complete the full integration.

majin1102:
Close #4707


majin1102:
ready for review @jackye1995 
PTAL when you have time

PR #4773, opened 2025-09-19
perf!: dynamic pruning for vector search

BubbleCal:
this avoids searching partitions that are too far away from the query, which could save computing resources to improve QPS and potentially reduce the latency (~5%)
image

it doesn't affect recall, because the dynamic pruning only excludes partitions that are too far

breaking because this changes `minimum_nprobes` default from 20 to 1, but this feature will search over `max(minimum_nprobes, pruned_nprobes)` in early_search phase

### DIFF
--- a/.bumpversion.toml
+++ b/.bumpversion.toml
@@ -1,5 +1,5 @@
 [tool.bumpversion]
-current_version = "0.40.0-beta.1"
+current_version = "0.40.0-beta.2"
 parse = "(?P<major>\\d+)\\.(?P<minor>\\d+)\\.(?P<patch>\\d+)(-(?P<prerelease>(beta|rc))\\.(?P<prerelease_num>\\d+))?"
 serialize = [
     "{major}.{minor}.{patch}-{prerelease}.{prerelease_num}",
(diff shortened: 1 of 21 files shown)


Xuanwo (approved):

rust/lance-index/src/vector.rs
```diff
@@ -87,6 +87,8 @@ pub struct Query {
 
     /// The minimum number of probes to load and search.  More partitions
     /// will only be loaded if we have not found k results.
+    ///
+    /// The planner will always search at least this many partitions. Defaults to 1.
```
wjones127, line 92: Why this change? It seems like it would change the recall, wouldn't it?

wjones127 (commented):
Could you explain more whether this change will affect recall? I wonder if we should have some recall benchmarks or something.

rust/lance-index/src/vector.rs
BubbleCal: No, it will set `min_nprobes = max(min_probes, pruned_nprobes)` where `pruned_nprobes` is a value that the number of partitions that are close to the closest one.

Will add more comments here.

The figure I posted in PR desc contains recalls, it doesn't affect recall, because the dynamic pruning only excludes partitions that are too far

Issue #3487, opened 2025-02-27
Write path optimization: Inline transaction files


wjones127:
Currently, transaction files are separate from the manifest. They are only locatable from the manifest itself. Thus, if you want to read the transaction data for a given version, it requires two hops: first, read the manifest to figure out the transaction file path, then read the transaction file itself. This also means the commit path requires at least two iops, since you need to write two different objects.

Instead, we should put the transaction file within the manifest. If the manifest is small enough, that means when we read the footer we will also get the transaction data for free. (We already read data with a minimal block size.) For larger manifests, we won't read this data, unless requested.

This will take writes from 3 write IOPS to 2 write IOPS, by combining the transaction file and manifest writes. When there are conflicts, and we need to read the conflicting transactions, this will often result in fewer IOPS as well,  since we can read the transaction information in one IOP instead of two.

westonpace:
This is just the one transaction that was used to create the particular manifest?  Or is it the history of transactions?

wjones127:
Just the one transaction

Xuanwo:
Closing as completed by #4774. Inline transactions are implemented in manifest read/write paths and covered by tests that verify no separate transaction file is required.

PR #4774, opened 2025-09-19
feat: provide inline_transaction model for IO optimizing

majin1102:


protos/table.proto
```diff
@@ -162,6 +162,22 @@ message Manifest {
   // base_paths[id = 0] + /data/ + file.path
   // the key(a.k.a index) starts from 0, increased by 1 for each new base path.
   repeated BasePath base_paths = 18;
+
+  // The transaction that created this version.
+  oneof inline_transaction {
+    // if < 200KB, store the transaction content inline
+    bytes transaction_content = 19;
+    // if >= 200KB, store the transaction content at the specified offset
+    TransactionSection transaction_section = 20;
```
jackye1995: I was thinking we should just keep it as a separated file which seems to be more backwards compatible. Is there any other benefit in storing it at offset of the manifest?

protos/table.proto
```diff
@@ -162,6 +162,22 @@ message Manifest {
...
+  }
```
wjones127: We should just unconditionally store the transaction at an offset, IMO. This makes this simpler, and it's still possible to read the transaction when it is small.

We currently store the index metadata at an offset, and here's how it works right now:

1. When we read the manifest file, we always read the first `block_size` bytes (4kb for local fs, 64kb for object storage)
2. We decode as much as we can from that last chunk. If the entire manifest file is < 64kb, this often means we get the entire content (including the index metadata) in 1 IOP.
3. If that last chunk wasn't sufficient, we read the remainder.

This basically gives the same effect as the optionally inline message: If it's small, we automatically read the index and transaction information in the first IO request. But if it's large, we do it in a separate request.

wjones127: Here's the place where we opportunistically read the index metadata from manifests:

https://github.com/lancedb/lance/blob/a3859653e29fc73399de49187956817c40f83a61/rust/lance/src/dataset.rs#L511-L536

protos/table.proto
wjones127: For backwards compatibility we should write both for a while. We can stop writing the external file after adding a feature flag 6+ months in the future. I've described the sequence of steps in https://github.com/lancedb/lance/issues/3487

wjones127 (commented):
I think putting inline in the protobuf is necessary. We can put it at an offset and use existing optimizations to read it without an additional IOP.

protos/table.proto
```diff
@@ -162,6 +162,22 @@ message Manifest {
...
+
+  message TransactionSection {
+    // The summary of the transaction including operation type and some kvs in transaction_properties
+    // This is used to get some information about the transaction without loading the whole transaction content
+    map<string, string> summary = 1;
```
majin1102: Hi, @wjones127 , What do you think of this part.

Basically it's redundent content for operation and transaction_properties. The case I want to solve now may encount listing historical manifest and transaction summaries. I'm not sure the opportunistically read could actually help this listing case if the manifest is large(actually the larger manifest we might need to accelerate it more).

On the other hand, some information may be useful if we could directly read from manifest like commit message. So I added this summary thing for directly reading from manifest without reading transaction part. 
cc @jackye1995

wjones127: For the list transactions use case, I can’t help like feeling the best thing would be to cache compacted summary data in some secondary file. Like we could generate some Lance file next to the manifest that contains all transaction summaries before it. Then to get most of the history you just need to query that. Then read the transactions of the next few uncached versions. What do you think of that?

jackye1995: How much performance do we want for this use case? I always thought reading information like transaction summary is more for non performance critical workloads, like displaying info to the user, triggerring some optimization jobs based on the action performed, etc. that can stand a few more IOPS.

majin1102: > Like we could generate some Lance file next to the manifest that contains all transaction summaries before it

I had the same ideas. I thought lance might provide a lazy summary file includes tags and branches, summaries and properties. And could use the index framework to maintain this file. This was only an early idea. I could raise a discussion for it. Then we should ignore the summary here.

> How much performance do we want for this use case? I always thought reading information like transaction summary is more for non performance critical workloads, like displaying info to the user, triggerring some optimization jobs based on the action performed, etc. that can stand a few more IOPS

For my case, one page displays at least 20 versions. each version should get a summary info includes manifest and transaction. If files seperated, the IO costs might have ​​a magnification of 20x. Let's say read a transaction file costs 100ms，that means 2 seconds magnification at least.

wjones127: Yeah, I'm inclined to not have the summary there. Just keep it in the transaction. Pulling it out into the manifest means that listing is still `O(num versions)` IO requests. To make that fast, better to create some mechanism to query it.

> non performance critical workloads, like displaying info to the user

I'd argue that displaying info to the user is somewhat performance critical, in that we'd like it to return fast enough to feel responsive.

jackye1995: > If files seperated, the IO costs might have ​​a magnification of 20x. Let's say read a transaction file costs 100ms，that means 2 seconds magnification at least.
> I'd argue that displaying info to the user is somewhat performance critical, in that we'd like it to return fast enough to feel responsive.

I was thinking we will just make the requests in parallel, so the latency won't really be 2 seconds, more like 200-300ms if anything got throttled and retried. But agree we need to make sure it is fast enough to feel responsive.

> I thought lance might provide a lazy summary file includes tags and branches, summaries and properties. And could use the index framework to maintain this file.

that sounds like a nice idea, I originally created the concept of "system index" basically for such use cases.

majin1102:
I believe this PR is ready for review @wjones127 @jackye1995

majin1102:
Ready for review @wjones127  cc @jackye1995

protos/table.proto
```diff
@@ -118,6 +118,10 @@ message Manifest {
   // and {uuid} is a hyphen-separated UUID.
   string transaction_file = 12;
 
+  // The file position of the transaction content.
+  // None if transaction is empty
+  optional uint64 transaction_section = 21;
```
wjones127: Could you describe here how to calculate where the message ends?

rust/lance/src/dataset/write/commit.rs
```diff
@@ -181,6 +183,13 @@ impl<'a> CommitBuilder<'a> {
         self
     }
 
+    /// Control whether to also write an external transaction file.
+    /// Default is true for backward compatibility. Set to false to write inline only.
+    pub fn with_write_transaction_file(mut self, write_transaction_file: bool) -> Self {
+        self.write_transaction_file = write_transaction_file;
+        self
+    }
```
wjones127: I don't think this should be a runtime setting. We should add a writer feature flag for `only writing inline transaction files`, make it off by default. Users can then enable it on table creation or with a later update. That will enforce older libraries don't try to write to this.

rust/lance/src/dataset.rs
```diff
@@ -2828,60 +2858,51 @@ mod tests {
...
-            }
+        // Delete external transaction file to force inline path usage
+        if let Some(tx_file) = ds2.manifest.transaction_file.as_ref() {
+            let tx_path = ds2.base.child("_transactions").child(tx_file.as_str());
+            let _ = ds2.object_store.inner.delete(&tx_path).await; // ignore errors
         }
+
+        // Read should still succeed via inline transaction
+        let read_tx = ds2.read_transaction().await.unwrap().unwrap();
+        assert_eq!(read_tx.read_version, tx.read_version);
```
wjones127: Nice test!

rust/lance/src/dataset.rs
```diff
@@ -643,6 +640,29 @@ impl Dataset {
             }
         }
 
+        // If transaction is also in the last block, we can take the opportunity to
+        // decode them now and cache them.
```
wjones127, line 706: I don't see an IO test for this. Could you add one as part of this test?

https://github.com/lancedb/lance/blob/2caeeaac6124a570b94be28bf515f1f8f9992c0e/rust/lance/src/dataset.rs#L2873-L2874

You can see an example of how we did this for indices here:

https://github.com/lancedb/lance/blob/2caeeaac6124a570b94be28bf515f1f8f9992c0e/rust/lance/src/index.rs#L2437-L2446

majin1102:
> Nice work. Excited to have this in soon!
> 
> I'd like to see a test on IO. And I think disabling writing separate transaction files should enable a writer feature flag, which will alert older libraries of the incompatibility.

Thanks for the guide and review so far.  I'm currently on a public holiday. I will adress your comments as soon as possible!

rust/lance/src/dataset/write/commit.rs
majin1102: > Users can then enable it on table creation or with a later update

I guess that means directly deleting the code of write_transaction_file. And for the six months as you said, we kept the write_transaction_file and don't control it by any setting. And we update the feature flags by hardcode in some version in the future. Do I understand right?

The reason I ask is feature flags applying is after write_transaction_file, besides, we don't have a initialized feature flag to read in `commit_new_dataset`.  I think if we want to dynamically control the behivour we still need some kind of setting. But I think this may be over thinking. Hardcode seems fine.

majin1102:
Ready for another look @wjones127

rust/lance-table/src/format/transaction.rs
```diff
@@ -0,0 +1,41 @@
...
+    /// Keeping this simple avoids ring dependencies while still enabling
+    /// Struct-first parameter passing in lance-table.
+    pub inner: pb::Transaction,
+}
+
+impl Transaction {
+    /// Construct a format-layer Transaction from a protobuf Transaction.
+    pub fn from_pb(pb_tx: pb::Transaction) -> Self {
+        Self { inner: pb_tx }
+    }
```
wjones127: This probably should just be a `From<pb::Transaction>` impl

rust/lance/src/dataset/write/commit.rs
wjones127: > And we update the feature flags by hardcode in some version in the future. Do I understand right?

Sorry, I think I wasn't clear. We should still have a setting.

The way you had it written before, users would choose on each commit whether to write the transaction file. I proposed instead they should decide that only once, when they create the dataset. Or they could decide later when they do a one-time migration of an existing dataset.

See how we do it for stable row id or v2 manifest paths, in `WriteParams`:

https://github.com/lancedb/lance/blob/ec2ac35b9a0bd644e222b09a2f19fe899e9b1ed8/rust/lance/src/dataset/write.rs#L196-L207

majin1102:
> Generally looks good. Do you think we should have a setting? It would be nice to be able to enable it?

Yeah, I did think we should have a setting. 

> See how we do it for stable row id or v2 manifest paths, in WriteParams:

I'm a little confused about this `WriteParams`. The parameters did't seem to store in any places. The only way we could use is to pass from the write interface. How does this implement the fact that 

> instead they should decide that only once, when they create the dataset. Or they could decide later when they do a one-time migration of an existing dataset

I'm not sure if you are saying just modify the default value of `WriteParams` to migrate this setting(If so I think CommitConfig might be suitable)? @wjones127

wjones127:
@majin1102 Did you look at how stable row ids work? I think that's a good guide here.

The user turns them on via `WriteParams`:

https://github.com/lancedb/lance/blob/8027b5c9fece8dd078b842e6d70cf8f75184855a/rust/lance/src/dataset/write.rs#L196-L200

This is passed down with `ManifestWriteConfig`:

https://github.com/lancedb/lance/blob/8027b5c9fece8dd078b842e6d70cf8f75184855a/rust/lance/src/dataset.rs#L2428-L2435

And then it is saved in the feature flags:

https://github.com/lancedb/lance/blob/8027b5c9fece8dd078b842e6d70cf8f75184855a/rust/lance-table/src/feature_flags.rs#L27

When we load the dataset, we call `Manifest::uses_stable_row_ids()`, which is just implemented as a check on the feature flag:

https://github.com/lancedb/lance/blob/8027b5c9fece8dd078b842e6d70cf8f75184855a/rust/lance-table/src/format/manifest.rs#L515-L518

majin1102:
> @majin1102 Did you look at how stable row ids work? I think that's a good guide here.
> 
> The user turns them on via `WriteParams`:
> 
> https://github.com/lancedb/lance/blob/8027b5c9fece8dd078b842e6d70cf8f75184855a/rust/lance/src/dataset/write.rs#L196-L200
> 
> This is passed down with `ManifestWriteConfig`:
> 
> https://github.com/lancedb/lance/blob/8027b5c9fece8dd078b842e6d70cf8f75184855a/rust/lance/src/dataset.rs#L2428-L2435
> 
> And then it is saved in the feature flags:
> 
> https://github.com/lancedb/lance/blob/8027b5c9fece8dd078b842e6d70cf8f75184855a/rust/lance-table/src/feature_flags.rs#L27
> 
> When we load the dataset, we call `Manifest::uses_stable_row_ids()`, which is just implemented as a check on the feature flag:
> 
> https://github.com/lancedb/lance/blob/8027b5c9fece8dd078b842e6d70cf8f75184855a/rust/lance-table/src/format/manifest.rs#L515-L518
> 
> 

Thank you so much for the details!

I was using open call in IDEA. Next time I will look more patiently and carefully(maybe with some simple tests)

wjones127:
@majin1102 did you still want to finish this? It seemed like it was almost done and then you moved it into draft.

majin1102:
> @majin1102 did you still want to finish this? It seemed like it was almost done and then you moved it into draft.

Sorry for delaying so far. I didn't mean to leave this unfinished. I was encountering an assertion failure on windows after some code rebasing(assert IO times). I had some clue and tried to fix, but didn't work. I was intending to use some more time to look through this. But yeah there's some other parallel work and eventually led to the PR being consistently delayed.

In the meanwhile I found load_new_transactions() was still using transaction files:
https://github.com/lancedb/lance/blob/2f95f341606220292343dfaa0fa64f5453357067/rust/lance/src/dataset.rs#L2078-L2120

I missed this part.  I will complete this asap . Thanks again for the review and guide @wjones127

majin1102:
> I was encountering an assert failure on windows after some code rebasing(assert IO times). I had some clue and tried to fix, but didn't work

After the latest rebase, all Windows tests pass. Looks like a recent commit fixed that windows issue. Feel free to merge if it looks good to you. Please let me know if you have any other comments!

PR #4778, opened 2025-09-19
chore: update DataFusion to 50, Arrow to 56

timsaucer:
Leaving as draft until datafusion-python 50.0.0 releases.

This is a dependency update for DataFusion and Arrow.

The only change that seems like it is impactful is that the arrow buffers no longer use a minimum capacity of 64 bytes. https://github.com/apache/arrow-rs/pull/7984 appears to be the originating source. This updates the statistics that are generated in the unit test.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -13,9 +13,9 @@ dependencies = [
 
 [[package]]
 name = "adler2"
-version = "2.0.0"
+version = "2.0.1"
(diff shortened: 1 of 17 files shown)


timsaucer:
I'll rebase it before moving to ready. The first two commits are already in main.

timsaucer:
Closing in favor of https://github.com/lancedb/lance/pull/4751

PR #4784, opened 2025-09-19
fix: fix decoder bugs in 2.1 list handling, expand test coverage

westonpace:
This fixes various issues in the 2.1 list decoder.  In addition, I've reworked the test utilities slightly as a large number of tests were not running against 2.1.  This exposed several issues and some tests I've had to defer fixing for future PRs and I've created TODO's for those.

### DIFF
--- a/rust/lance-encoding/src/encodings/logical/primitive.rs
+++ b/rust/lance-encoding/src/encodings/logical/primitive.rs
@@ -235,14 +235,14 @@ impl DecodeMiniBlockTask {
             let mut rep = rep.as_ref();
             // If there is a preamble and we need to skip it then do that first.  The work is the same
             // whether there is def information or not
-            let mut items_in_preamble = 0;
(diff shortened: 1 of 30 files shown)


rust/lance-encoding/src/encodings/logical/primitive.rs
```diff
@@ -342,19 +343,18 @@ impl DecodeMiniBlockTask {
 
                 // Adjust for any skipped preamble
                 if preamble_action == PreambleAction::Skip {
-                    // TODO: Should this be items_in_preamble?  If so, add a
-                    // unit test for this case
-                    new_start += first_row_start;
-                    new_end += first_row_start;
+                    new_start += items_in_preamble;
+                    new_end += items_in_preamble;
```
westonpace, line 347: If only past-Weston had been more diligent :sob: :sob: :sob:

PR #4786, opened 2025-09-20
fix: move indexes back to table.proto to avoid forward compat. issues

westonpace:
This also adds forward compatibility testing for all scalar indexes (not FTS)


.github/workflows/python.yml
```diff
@@ -119,18 +119,50 @@ jobs:
...
+          name: linux-wheels
+          path: python/target/wheels/pylance-*.whl
+
+  forward-compat:
+    needs: linux
+    runs-on: ubuntu-24.04
+    name: Forward Compatibility Tests (${{ matrix.lance-version }})
+    strategy:
+      matrix:
+        lance-version: ["0.16.0", "0.29.1.beta2", "0.36.0"]
```
wkalt: how will this be kept up to date?

protos/table.proto
```diff
@@ -527,3 +527,34 @@ message MemWalIndexDetails {
     uint64 last_updated_dataset_version = 7;
   }
 }
+
+// ----- Forward compatibility -----
+
+// NOTE: Do *NOT* add new index details here.  Add them to the index.proto file instead.
+// These are only here for forward compatibility.  If you need to modify these messages
```
wkalt: we should specify the forward compatibility scenario/concern here, makes it easier on readers

.github/workflows/python.yml
westonpace: :shrug: Right now it's just a smattering of versions where we've known to have issues.  In theory we want to support compatibility with all previous versions but that testing matrix is too large.  Maybe we could just randomly sample from releases in some way?

wkalt: I was wondering if maybe it would be an improvement to always test compatibility relationship with the previous release or n-2 or whatever. That also won't test the full matrix but it would stay up to date without maintenance. Do you think that would be any better?

protos/table.proto
westonpace: Added a bit more details.

.github/workflows/python.yml
westonpace: That works.  Perhaps we can keep the 0.16 / 0.29.1.beta2 though so we have "very old" and "relatively recent".  Mind if I save this for a follow-up?  This is blocking people and I'd rather get it through quickly.

wkalt: yeah, that's totally fine to handle it separate :+1:

wjones127: +1 on having a recent one.

wjones127 (approved):
Well, I guess now I understand why protobuf recomends [defining each message in a separate file](https://protobuf.dev/best-practices/dos-donts/#separate-files).

westonpace:
> Well, I guess now I understand why protobuf recomends [defining each message in a separate file](https://protobuf.dev/best-practices/dos-donts/#separate-files).

Actually, the true culprit is prost's [to_msg](https://docs.rs/prost-types/latest/prost_types/struct.Any.html#method.to_msg).  Typically protobuf ignores the type URL but the `to_msg` method goes out of its way to explicitly check it.  I get the spirit of the idea (provide more useful error feedback and avoid potential false positive when something just happens to decode into a different message type) but it locks you into the package name.

PR #4787, opened 2025-09-20
[WIP] feat: implement Bloom Filter concurrent conflict detection for merge insert operations

### LINKED ISSUE #4585: MergeInsert produces duplicated rows
When concurrent processes both write new rows of the same primary key through merge-insert, currently we do not have a way to detect it as conflict because it does not touch any existing fragments.

I think this is "technically" the right behavior under snapshot isolation, given we do not guarantee exactly one row per `on` column. But from the perspective of the people using merge insert as a way to ensure one row per primary key, this basically breaks the assumption.

I am able to verify with this test:

```rust
    #[tokio::test]
    async fn test_concurrent_merge_insert_duplicate_new_rows() {
        // This test verifies that when two writers both do merge insert to add a new row
        // with the same key, the result dataset will have duplicated rows because
        // new rows don't have addresses yet so conflict detection doesn't catch this.
        
        let test_dir = tempdir().unwrap();
        let test_uri = test_dir.path().to_str().unwrap();
        
        // Create initial dataset with some existing data
            Field::new("key", DataType::Int32, false),
            Field::new("value", DataType::Utf8, false),
            Field::new("writer", DataType::Utf8, false),
        ]));
        
        let initial_batch = RecordBatch::try_new(
            schema.clone(),
            vec![
                Arc::new(Int32Array::from(vec![1, 2, 3])),
                Arc::new(StringArray::from(vec!["value1", "value2", "value3"])),
                Arc::new(StringArray::from(vec!["initial", "initial", "initial"])),
            ],
        )
        .unwrap();
        
        let reader = Box::new(RecordBatchIterator::new(
            vec![Ok(initial_batch)],
            schema.clone(),
        ));
        
        let dataset = Dataset::write(reader, test_uri, Default::default())
            .await
            .unwrap();
        let initial_version = dataset.version().version;
        let dataset = Arc::new(dataset);
        
        // First writer: operates on current dataset version
        let writer1_batch = RecordBatch::try_new(
            schema.clone(),
            vec![
                Arc::new(Int32Array::from(vec![4])),
                Arc::new(StringArray::from(vec!["value_from_writer1"])),
                Arc::new(StringArray::from(vec!["writer1"])),
            ],
        )
        .unwrap();
        
        let job1 = MergeInsertBuilder::try_new(dataset.clone(), vec!["key".to_string()])
            .unwrap()
            .when_matched(WhenMatched::UpdateAll)
            .when_not_matched(WhenNotMatched::InsertAll)
            .try_build()
            .unwrap();
        
        let reader1 = Box::new(RecordBatchIterator::new(
            vec![Ok(writer1_batch)],
            schema.clone(),
        ));
        let stream1 = reader_to_stream(reader1);
        
        // Execute first merge insert
        let (dataset_after_writer1, stats1) = job1.execute(stream1).await.unwrap();
        assert_eq!(stats1.num_inserted_rows, 1, "Writer 1 should insert 1 row");
        
        // Second writer: simulates a concurrent writer by checking out the initial version
        // This mimics the scenario where writer2 started before writer1 committed
        let dataset_for_writer2 = dataset.checkout_version(initial_version).await.unwrap();
        
        let writer2_batch = RecordBatch::try_new(
            schema.clone(),
            vec![
                Arc::new(Int32Array::from(vec![4])),
                Arc::new(StringArray::from(vec!["value_from_writer2"])),
                Arc::new(StringArray::from(vec!["writer2"])),
            ],
        )
        .unwrap();
        
        let job2 = MergeInsertBuilder::try_new(Arc::new(dataset_for_writer2), vec!["key".to_string()])
            .unwrap()
            .when_matched(WhenMatched::UpdateAll)
            .when_not_matched(WhenNotMatched::InsertAll)
            .try_build()
            .unwrap();
        
        let reader2 = Box::new(RecordBatchIterator::new(
            vec![Ok(writer2_batch)],
            schema.clone(),
        ));
        let stream2 = reader_to_stream(reader2);
        
        // Execute second merge insert - this should succeed despite writer1 already adding key=4
        // because new rows don't have addresses for conflict detection
        let result2 = job2.execute(stream2).await;
        
        match result2 {
            Ok((dataset_after_writer2, stats2)) => {
                assert_eq!(stats2.num_inserted_rows, 1, "Writer 2 should also insert 1 row");
                
                // Use the most recent dataset version
                let final_dataset = if dataset_after_writer1.version().version > dataset_after_writer2.version().version {
                    dataset_after_writer1
                } else {
                    dataset_after_writer2
                };
                
                // Check if we have duplicate rows with key=4
                let result = final_dataset
                    .scan()
                    .filter("key = 4")
                    .unwrap()
                    .try_into_batch()
                    .await
                    .unwrap();
                
                let num_rows = result.num_rows();
                
                // This assertion demonstrates the bug: we expect 2 duplicate rows
                // because both writers succeeded in inserting the same key
                assert_eq!(
                    num_rows, 2,
                    "Expected 2 duplicate rows with key=4 due to concurrent insert of new rows \
                     (conflict detection doesn't work for new rows without addresses)"
                );
                
                // Verify both values are present
                let writer_col = result
                    .column_by_name("writer")
                    .unwrap()
                    .as_any()
                    .downcast_ref::<StringArray>()
                    .unwrap();
                
                let writers: Vec<&str> = (0..num_rows)
                    .map(|i| writer_col.value(i))
                    .collect();
                
                assert!(
                    writers.contains(&"writer1") && writers.contains(&"writer2"),
                    "Both writers should have successfully inserted their rows"
                );
            }
            Err(e) => {
                // If writer2 fails due to conflict detection, that would be the desired behavior
                // but currently this doesn't happen for new rows
                panic!(
                    "Writer 2 failed with error: {:?} - this shouldn't happen since new rows \
                     don't have addresses for conflict detection",
                    e
                );
            }
        }
    }
```

yanghua:
closes #4585


yanghua:
This PR is only for CI. Still WIP, not ready for review.

yanghua:
cc @jackye1995 please take a look when you have time.

protos/join_key.proto
```diff
@@ -0,0 +1,52 @@
...
+  uint32 bitmap_bits = 3;
+  // Reserved for future fields to avoid reuse.
+  reserved 4, 5;
+  reserved "hash_seed", "hash_algo";
+}
+
+// Join key metadata attached to a Transaction for conflict detection.
+message JoinKeyMetadata {
+  // Names of columns participating in the join key.
+  repeated string columns = 1;
```
jackye1995, line 45: this should be column field ID

protos/join_key.proto
```diff
@@ -0,0 +1,52 @@
...
+// Bloom filter data for join key set membership tests.
+message BloomFilterData {
+  // Bitset backing the bloom filter.
+  bytes bitmap = 1;
+  // Number of hash functions used.
+  uint32 num_hashes = 2;
+  // Total number of bits in the bitmap.
+  uint32 bitmap_bits = 3;
+  // Reserved for future fields to avoid reuse.
+  reserved 4, 5;
```
jackye1995, line 38: why have all these reserved fields? I think they are not really needed until we want to add them in the future?

protos/join_key.proto
```diff
@@ -0,0 +1,52 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright The Lance Authors
+
+syntax = "proto3";
+
+package lance.table;
+
+// Value of the join key representation (reserved for future use)
```
jackye1995, line 8: this is unused, should just remove

protos/join_key.proto
```diff
@@ -0,0 +1,52 @@
...
+  uint32 num_hashes = 2;
+  // Total number of bits in the bitmap.
+  uint32 bitmap_bits = 3;
+  // Reserved for future fields to avoid reuse.
+  reserved 4, 5;
+  reserved "hash_seed", "hash_algo";
+}
+
+// Join key metadata attached to a Transaction for conflict detection.
+message JoinKeyMetadata {
```
jackye1995, line 43: this name does not feel right to me. Consider we use this for an INSERT, then there is not really a join key. What about just keyMetadata or something like KeySet or RowKeySet?

protos/transaction.proto
```diff
@@ -33,6 +34,8 @@ message Transaction {
   // __lance_commit_message is a reserved key
   map<string, string> transaction_properties = 4;
 
+  // Join key metadata using typed protobuf message. This is the sole carrier.
+  optional JoinKeyMetadata join_key_metadata = 6;
```
jackye1995, line 38: I think this should not be in the transaction itself, it should be unique to only a few write transactions to do detection for operations like insert or merge_insert that could add new data to the table?

rust/lance/src/dataset/write/merge_insert.rs
```diff
@@ -1254,11 +1257,27 @@ impl MergeInsertJob {
...
+        // iterator to obtain two streams: one for join key computation and one for execution.
         let stream = source.into_stream();
-        self.execute_uncommitted_impl(stream).await
+        let mut iter = super::new_source_iter(stream, true).await?; // enable replay to duplicate
+        let first = iter
+            .next()
+            .expect("source stream exhausted while computing join key filter");
+
+        let join_key_metadata =
+            compute_join_key_metadata_from_stream(first, &self.params.on).await?;
```
jackye1995, line 1273: this is not what I was expecting to implement the feature, please let me know what you think. In my mind, the bloom filter or exact set should be computed similar to how we compute the `affected_rows`, it's just this is for newly added rows, and affected_rows are for rows where the join key (primary key) already exists in the dataset.

rust/lance/src/dataset/conflict_detection/conflict_detector.rs
jackye1995, line 1: I think instead of having a new conflict detector, we should just update the existing logic, so that instead of just `affected_rows`, we have a new field `new_rows` in the Update transaction model that tracks the bloom filter or exact set, and then the existing conflict resolution module should be able to just consume it like how it consumes affected_rows

PR #4793, opened 2025-09-23
test: add python test for configuring the zone size of a zone map index

westonpace:


Xuanwo (approved):
Thank you!

HaochengLIU:
out of curiosity, is it for catching a missing case or a bug? @westonpace

westonpace:
> out of curiosity, is it for catching a missing case or a bug?

No.  I thought we couldn't do this and I needed it.  So I wrote a test that I was expecting to fail.  Then it passed.  Then I figured I'd just leave it in for regression purposes.

Issue #4762, opened 2025-09-17
LanceDataset.take does not support fetching the _rowid


westonpace:
```
import lance
import pyarrow as pa
import random

sample_size = 10
ds = lance.write_dataset(pa.table({"x": range(1000)}), "memory://test")

total_rows = len(ds)
sampled_indices = random.sample(range(total_rows), min(sample_size, total_rows))

sample_dataset = ds.take(sampled_indices, columns=["_rowid"])
assert sample_dataset.num_rows == 10
assert sample_dataset.num_columns == 1
```

Pmathsun:
Good issue! I just ran into this problem too.

chunshao90:
I will try to support this.

Xuanwo:
Closing as completed by #4794. `take` supports `_rowid` and `_rowaddr`, including projection ordering and duplicate-column coverage.

PR #4794, opened 2025-09-23
feat: support fetching _rowid and _rowaddr in take operation

chunshao90:
refer to: https://github.com/lancedb/lance/issues/4762

### DIFF
--- a/rust/lance/src/dataset/take.rs
+++ b/rust/lance/src/dataset/take.rs
@@ -129,8 +129,8 @@ async fn do_take_rows(
     mut builder: TakeBuilder,
     projection: Arc<ProjectionPlan>,
 ) -> Result<RecordBatch> {
-    let projection_with_row_id = projection.physical_projection.with_row_id;
-    let projection_with_row_addr = projection.physical_projection.with_row_addr;
+    let with_row_id_in_projection = projection.physical_projection.with_row_id;
+    let with_row_addr_in_projection = projection.physical_projection.with_row_addr;
 
     let row_addrs = builder.get_row_addrs().await?.clone();
 
@@ -184,8 +184,8 @@ async fn do_take_rows(
         })?;
 
         let read_config = FragReadConfig::default()
-            .with_row_id(projection_with_row_id)
-            .with_row_address(projection_with_row_addr);
+            .with_row_id(with_row_id_in_projection)
+            .with_row_address(with_row_addr_in_projection);
         let reader = fragment.open(&physical_schema, read_config).await?;
         reader.legacy_read_range_as_batch(range).await
     } else if row_addr_stats.sorted {
@@ -231,8 +231,8 @@ async fn do_take_rows(
                 fragment,
                 row_offsets,
                 physical_schema.clone(),
-                projection_with_row_id,
-                projection_with_row_addr,
+                with_row_id_in_projection,
+                with_row_addr_in_projection,
             );
             batches.push(batch_fut);
         }
@@ -271,7 +271,7 @@ async fn do_take_rows(
                     fragment,
                     indices,
                     physical_schema.clone(),
-                    projection_with_row_id,
+                    with_row_id_in_projection,
                     true,
                 )
             })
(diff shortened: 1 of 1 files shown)


chunshao90:
While implementing this PR, I noticed that the current implementation has inconsistent usage of _rowid and _rowaddr, which is quite confusing. This seems to be due to historical implementation reasons.

For example, in the do_take_rows method, both TakeBuilder and ProjectionPlan have configurations for with_row_addr, and similar inconsistencies exist in the fragment reading pipeline (I don't recall the exact details, but they're spread across multiple layers).

steFaiz:
> While implementing this PR, I noticed that the current implementation has inconsistent usage of _rowid and _rowaddr, which is quite confusing. This seems to be due to historical implementation reasons.

Yes, the community is trying to distinguish them, see [4352](https://github.com/lancedb/lance/pull/4352)

chunshao90:
PTAL: PR review needed.  @jackye1995

jackye1995:
I thought we already have 

```rust
pub struct TakeBuilder {
    dataset: Arc<Dataset>,
    row_ids: Option<Vec<u64>>,
    row_addrs: Option<Vec<u64>>,
    projection: Arc<ProjectionPlan>,
    with_row_address: bool,
}
```

and you can do 

```rust
pub fn take_builder(
        self: &Arc<Self>,
        row_ids: &[u64],
        projection: impl Into<ProjectionRequest>,
    ) -> Result<TakeBuilder> {
```

Why do you want to change the `take::take`?

chunshao90:
> I thought we already have
> 
> ```rust
> pub struct TakeBuilder {
>     dataset: Arc<Dataset>,
>     row_ids: Option<Vec<u64>>,
>     row_addrs: Option<Vec<u64>>,
>     projection: Arc<ProjectionPlan>,
>     with_row_address: bool,
> }
> ```
> 
> and you can do
> 
> ```rust
> pub fn take_builder(
>         self: &Arc<Self>,
>         row_ids: &[u64],
>         projection: impl Into<ProjectionRequest>,
>     ) -> Result<TakeBuilder> {
> ```
> 
> Why do you want to change the `take::take`?

I previously replied that parameters like with_row_id exist in both TakeBuilder and Projection. When implementing take, different code paths retrieve these parameters from different places, and the logic is currently inconsistent.
As mentioned by @steFaiz, this is being addressed and improved in https://github.com/lancedb/lance/pull/4352.
Given that context, I'm a bit confused — could you clarify what I should do based on your comment?

jackye1995:
Ah I see. Sorry I think I totally misunderstood... Let me re-review

python/python/tests/test_take.py
```diff
@@ -0,0 +1,94 @@
...
+
+import random
+
+import lance
+import pyarrow as pa
+
+sample_size = 10
+
+
+def _gen_data():
```
jackye1995: This feels unnecessary to have a util for a 1-liner

python/python/tests/test_take.py
jackye1995, line 1: this can just be in `test_dataset.py` since all the other take cases are also there

rust/lance/src/dataset/take.rs
```diff
@@ -228,15 +242,15 @@ async fn do_take_rows(
...
-            physical_schema.as_ref(),
-            &ArrowSchema::new(vec![ArrowField::new(
-                ROW_ADDR,
-                arrow::datatypes::DataType::UInt64,
-                false,
-            )]),
-        )?;
-        let schema_with_row_addr = Arc::new(ArrowSchema::from(&projection_with_row_addr));
+        let schema = if !projection_with_row_addr {
+            let projection_with_row_addr = Schema::merge(
```
jackye1995: this is quite confusing since you are creating a variable with the same name here. Can we rename one of them?

rust/lance/src/dataset/take.rs
```diff
@@ -228,15 +242,15 @@ async fn do_take_rows(
...
-        let projection_with_row_addr = Schema::merge(
-            physical_schema.as_ref(),
-            &ArrowSchema::new(vec![ArrowField::new(
-                ROW_ADDR,
-                arrow::datatypes::DataType::UInt64,
-                false,
-            )]),
-        )?;
-        let schema_with_row_addr = Arc::new(ArrowSchema::from(&projection_with_row_addr));
+        let schema = if !projection_with_row_addr {
```
jackye1995: I think this should still be called `schema_with_row_addr` since at this point is it guaranteed to have it.

python/python/tests/test_take.py
```diff
@@ -0,0 +1,94 @@
...
+    ds = lance.write_dataset(pa.table({"x": range(1000)}), "memory://test")
+    return ds
```
artemru: can we consider a case with several fragments ?

python/python/tests/test_take.py
artemru: nit : this could be a fixture

rust/lance/src/dataset/take.rs
chunshao90: > I think this should still be called `schema_with_row_addr` since at this point is it guaranteed to have it.

I've removed this part because do_take already sets with_row_address to true. Also, there's no need to manually remove fields
https://github.com/lancedb/lance/blob/30dcb5b8eb01e2fd4d05824b908d4174bab549d6/rust/lance/src/dataset/take.rs#L303-L305

since 
```
let batch = projection.project_batch(batch).await?; 
```
correctly projects the desired columns.

Could you please confirm if this change looks good? Thanks!

chunshao90:
I've added some tests to cover multi-fragment scenarios, including both enable_stable_row_ids = true and enable_stable_row_ids = false cases.

PTAL @jackye1995 
Also, CI run would be appreciated. Thanks!

chunshao90:
I've fixed the CI issues.
PTAL @jackye1995, and restart the CI.

chunshao90:
PTAL @jackye1995

rust/lance/src/dataset/take.rs
```diff
@@ -707,6 +714,226 @@ mod test {
...
+                max_rows_per_file: 50,
+                ..Default::default()
+            };
+            let batches = RecordBatchIterator::new([Ok(data.clone())], data.schema());
+            let dataset = Dataset::write(batches, "memory://", Some(write_params))
+                .await
+                .unwrap();
+
+            assert_eq!(dataset.count_rows(None).await.unwrap(), 400);
+            // test projection with expression
```
jackye1995: nit: should separate into different tests.

rust/lance/src/dataset/take.rs
```diff
@@ -130,6 +129,9 @@ async fn do_take_rows(
     mut builder: TakeBuilder,
     projection: Arc<ProjectionPlan>,
 ) -> Result<RecordBatch> {
+    let projection_with_row_id = projection.physical_projection.with_row_id;
```
jackye1995: nit: the variable name is a bit confusing, what about `projection_has_row_id` and `projection_has_row_addr` to be more clear these are booleans

jackye1995 (approved):
just 2 nits, overall looks good to me! thanks for working on this!

rust/lance/src/dataset/take.rs
chunshao90: `with_row_id` is used as a boolean in many places throughout the project, so I think it's better to keep it consistent.

rust/lance/src/dataset/take.rs
chunshao90: Splitting this test into two cases depending on the value of enable_stable_row_ids?

jackye1995: No I mean split all the different test blocks

rust/lance/src/dataset/take.rs
jackye1995: What about with_row_id_in_projection

chunshao90: Done.

rust/lance/src/dataset/take.rs
chunshao90: Done.

jackye1995:
pending CI

chunshao90:
> pending CI

I have rebased main, please restart the CI.

jackye1995:
Thanks for working on this!

PR #4795, opened 2025-09-23
feat: add scan_range_after_filter pushdown optimization for filtered reads

LuQQiu:
Implements pushdown optimization for `scan_range_after_filter` (LIMIT/OFFSET after filtering) to reduce unnecessary I/O when reading filtered data with limits.
  - Push limits down to fragment level during planning when we have exact index
  results or sufficient AtLeast guarantees
  - Apply runtime soft limits per fragment and hard range enforcement globally
  - Multi-partition support not yet implemented


LuQQiu:
Adding more E2E test cases

rust/lance/src/io/exec/filtered_read.rs
```diff
@@ -520,18 +537,22 @@ impl FilteredReadStream {
...
+    // 2. The index result is AtLeast with guaranteed rows >= limit (we have enough guaranteed matches)
+    // Returns: (fragment reads, whether limit was pushed down to fragment ranges)
     #[instrument(name = "plan_scan", skip_all)]
     async fn plan_scan(
         dataset: &Dataset,
         fragments: Vec<LoadedFragment>,
         evaluated_index: &Option<Arc<EvaluatedIndex>>,
-        options: FilteredReadOptions,
+        options: &FilteredReadOptions,
+        _global_metrics: &FilteredReadGlobalMetrics,
```
cmccabe: why are we adding _global_metrics here if we don't use it?

rust/lance/src/io/exec/filtered_read.rs
```diff
@@ -659,7 +680,135 @@ impl FilteredReadStream {
...
+    /// Returns true if the limit was successfully pushed down to fragment ranges
+    fn apply_scan_range_after_filter(
+        scoped_fragments: &mut Vec<ScopedFragmentRead>,
+        range_after_filter: &Range<u64>,
+        refine_filter: &Option<Expr>,
+        evaluated_index: Option<&Arc<EvaluatedIndex>>,
+    ) -> bool {
+        if refine_filter.is_some() || evaluated_index.is_none() {
+            return false;
+        }
```
westonpace: What if evaluated_index is none but there is no filter at all?  I guess we just use scan range before filter in that case?

rust/lance/src/io/exec/filtered_read.rs
```diff
@@ -659,7 +680,135 @@ impl FilteredReadStream {
...
+                    .map(|frag| frag.ranges.iter().map(|r| r.end - r.start).sum::<u64>())
+                    .sum::<u64>();
+
+                if total_guaranteed >= range_after_filter.end {
+                    Self::trim_fragments_to_range(scoped_fragments, range_after_filter);
+                    true
+                } else {
+                    false
+                }
+            }
```
westonpace: I'm a little uncertain about this path.  It's possible all the guaranteed rows are at the end.  For example, what if we have 100 rows with 10 fragments and 10 rows in each fragment.  Then we get an AtLeast with the last 20 rows.  Then we have a limit of 5.  Wouldn't this throw away the last 9 fragments?  What if there are 0 matches in the first fragment?

I think you can apply a limit (but not skip) with AtMost however.

rust/lance/src/io/exec/filtered_read.rs
```diff
@@ -659,7 +680,135 @@ impl FilteredReadStream {
...
+                // Cannot optimize AtMost or other cases
+                false
+            }
+        }
+    }
+
+    fn trim_fragments_to_range(scoped_fragments: &mut Vec<ScopedFragmentRead>, range: &Range<u64>) {
+        let mut to_skip = range.start;
+        let mut to_take = range.end - range.start;
+        let mut fragments_to_keep = Vec::new();
```
westonpace:
```suggestion
        let mut fragments_to_keep = Vec::with_capacity(scoped_fragments.len());
```

Minor nit: this might be an over-estimate but better that than having to do several allocations

rust/lance/src/io/exec/filtered_read.rs
```diff
@@ -824,8 +973,18 @@ impl FilteredReadStream {
...
+                    });
+
+                let batch_stream = if let Some(ref range) = self.range_to_apply {
+                    Self::apply_hard_range(base_batch_stream, range.clone())
+                        .map_err(|e: lance_core::Error| DataFusionError::External(e.into()))
+                        .boxed()
+                } else {
+                    base_batch_stream
+                        .map_err(|e: lance_core::Error| DataFusionError::External(e.into()))
+                        .boxed()
```
westonpace, line 923: Minor nit: both paths have `map_err(...).boxed()`.  Maybe take it out of the if/else?

rust/lance/src/io/exec/filtered_read.rs
```diff
@@ -989,6 +1157,73 @@ impl FilteredReadStream {
...
+
+        stream.try_filter_map({
+            let rows_seen = rows_seen.clone();
+            move |batch| {
+                if batch.num_rows() == 0 {
+                    return future::ready(Ok(None));
+                }
+
+                let batch_rows = batch.num_rows();
+                let current_position = rows_seen.fetch_add(batch_rows, Ordering::SeqCst);
```
westonpace: Minor nit: could be relaxed ordering.

This is potentially troubling if the batches are being consumed with some kind of `buffered_unordered` although I don't know if it's wrong in that case.  If the user doesn't care about the order of the batches then a limit is kind of ill defined anyways.

rust/lance/src/io/exec/filtered_read.rs
```diff
@@ -824,8 +973,18 @@ impl FilteredReadStream {
                     })
                     .finally(move || {
                         partition_metrics.baseline_metrics.done();
-                    })
-                    .map_err(|e: lance_core::Error| DataFusionError::External(e.into()));
+                    });
+
+                let batch_stream = if let Some(ref range) = self.range_to_apply {
+                    Self::apply_hard_range(base_batch_stream, range.clone())
```
westonpace, line 917: What advantage is there to applying the filter here instead of using a dedicated `FilterExec`?  I guess it is just because we've decided to pushdown the limit and so we must apply it here?

rust/lance/src/io/exec/filtered_read.rs
```diff
@@ -1560,6 +1791,81 @@ impl ExecutionPlan for FilteredReadExec {
...
+        } else {
+            self.options
+                .scan_range_after_filter
+                .as_ref()
+                .map(|range| (range.end - range.start) as usize)
+        }
+    }
+
+    fn supports_limit_pushdown(&self) -> bool {
+        // This is to push limits down to downstream execs
```
westonpace:
```suggestion
        // This is to push the limit through the node and into an upstream node.
        // The only upstream node is the index search and we can't push the limit
        // to that node.
```

rust/lance/src/io/exec/filtered_read.rs
```diff
@@ -1560,6 +1791,81 @@ impl ExecutionPlan for FilteredReadExec {
...
+            self.options
+                .scan_range_before_filter
+                .as_ref()
+                .map(|range| (range.end - range.start) as usize)
+        } else {
+            self.options
+                .scan_range_after_filter
+                .as_ref()
+                .map(|range| (range.end - range.start) as usize)
+        }
```
westonpace, line 1753: What if both `scan_range_before_filter` AND `scan_range_after_filter` are set?

rust/lance/src/io/exec/filtered_read.rs
```diff
@@ -1560,6 +1791,81 @@ impl ExecutionPlan for FilteredReadExec {
...
+    }
+
+    fn supports_limit_pushdown(&self) -> bool {
+        // This is to push limits down to downstream execs
+        false
+    }
+
+    fn with_fetch(&self, limit: Option<usize>) -> Option<Arc<dyn ExecutionPlan>> {
+        limit?;
```
westonpace: Very strange that they pass in `Option<usize>` here.  Maybe they intend for this method to both _apply_ a fetch and **remove** a fetch?  If that is the case then we are not properly removing the fetch here.

westonpace:
```suggestion
        let limit = limit?;
```
Minor nit: then you don't need the `unwrap` below.

rust/lance/src/io/exec/filtered_read.rs
```diff
@@ -1560,6 +1791,81 @@ impl ExecutionPlan for FilteredReadExec {
...
+        // Support limit push down
+        let mut updated_options = self.options.clone();
+        let fetch = limit.unwrap();
+
+        if self.options.full_filter.is_none() {
+            let new_range = if let Some(existing_range) = &self.options.scan_range_before_filter {
+                let start = existing_range.start;
+                let end = existing_range.end.min(start + fetch as u64);
```
westonpace: This takes the lesser of the two fetches (existing and provided).  Is this method intended to replace the fetch entirely I wonder?  I feel like this case (`with_fetch` called when there is already a fetch) is somewhat ill defined.

rust/lance/src/io/exec/filtered_read.rs
```diff
@@ -1560,6 +1791,81 @@ impl ExecutionPlan for FilteredReadExec {
...
+            } else {
+                0..(fetch as u64)
+            };
+
+            match updated_options.with_scan_range_after_filter(new_range) {
+                Ok(options) => {
+                    updated_options = options;
+                }
+                Err(e) => {
+                    log::debug!("Cannot apply fetch limit to scan_range_after_filter: {}", e);
```
westonpace: I'm wondering if this should be `log::warn`.  Is this expected to happen?

rust/lance/src/io/exec/filtered_read.rs
```diff
@@ -964,7 +1124,15 @@ impl FilteredReadStream {
...
             )))
-            .map(|(batch_fut, args)| Self::wrap_with_filter(batch_fut, args.0, args.1)))
+            .map(|(batch_fut, args)| Self::wrap_with_filter(batch_fut, args.0, args.1));
+
+        if let Some(limit) = fragment_soft_limit {
+            Ok(Box::pin(
+                Self::apply_soft_limit(fragment_stream, limit).boxed(),
+            ))
+        } else {
+            Ok(Box::pin(fragment_stream.boxed()))
```
westonpace: `Box::pin(...boxed())` seems like we might be double boxing?  Do we need any boxing at all?

rust/lance/src/io/exec/filtered_read.rs
```diff
@@ -659,7 +680,135 @@ impl FilteredReadStream {
...
+                true
+            }
+            IndexExprResult::AtLeast(_) => {
+                let total_guaranteed = scoped_fragments
+                    .iter()
+                    .map(|frag| frag.ranges.iter().map(|r| r.end - r.start).sum::<u64>())
+                    .sum::<u64>();
+
+                if total_guaranteed >= range_after_filter.end {
+                    Self::trim_fragments_to_range(scoped_fragments, range_after_filter);
```
LuQQiu: This may not be correct, original fragment range may not consider the AtLeast filter results, double cehck

rust/lance/src/io/exec/optimizer.rs
```diff
@@ -172,5 +172,7 @@ pub fn get_physical_optimizer() -> PhysicalOptimizer {
     PhysicalOptimizer::with_rules(vec![
         Arc::new(crate::io::exec::optimizer::CoalesceTake),
         Arc::new(crate::io::exec::optimizer::SimplifyProjection),
+        // Push down limit into FilteredReadExec and other Execs via with_fetch()
+        Arc::new(datafusion::physical_optimizer::limit_pushdown::LimitPushdown::new()),
```
westonpace, line 176: I wonder if we need to revisit the `supports_limit_pushdown` method of our other exec nodes? (can be in follow-up)

rust/lance/src/io/exec/filtered_read.rs
LuQQiu: Yep, i will add a check in FilteredReadExec to prevent setting scan_range_after_filter & no filter combination. In this case, should just use scan_range_before_filter

rust/lance/src/io/exec/filtered_read.rs
LuQQiu: This is similar to the effect of the GlobalLimitExec, but the limit is already push down, so must apply it here

rust/lance/src/io/exec/filtered_read.rs
LuQQiu: The stream type is different, so cannot take it out the if/else

rust/lance/src/io/exec/filtered_read.rs
LuQQiu: Yep, None basically means remove any fetch limit. If limit is None, we just return None result (without pushing down).
It could be helpful during the plan optimization to redo the limit push down or compare plans

rust/lance/src/io/exec/filtered_read.rs
LuQQiu: Thanks, removed the global metrics

rust/lance/src/io/exec/filtered_read.rs
LuQQiu: This is incorrect, remove the AtLeast case for now

rust/lance/src/io/exec/filtered_read.rs
LuQQiu: Added the check to prevent scan_range_after_filter with no filter

rust/lance/src/io/exec/filtered_read.rs
LuQQiu: changed to relaxed

rust/lance/src/io/exec/filtered_read.rs
LuQQiu: the scan_range_before_filter current implementation is a little weird, for Exact and AtMost index result, it also applied after applying the filter, so i am not quite sure about the correct of that. 
I have touched the " scan_range_before_filter"  logics.
Changed to logics to with_filter -> go with "scan_range_after_filter", without_filter, go with "scan_range_before_filter"

rust/lance/src/io/exec/filtered_read.rs
LuQQiu: Yep agree, change to if existing fetch exist, then reject the push down.

rust/lance/src/io/exec/filtered_read.rs
LuQQiu: remove the double boxed()

rust/lance/src/io/exec/optimizer.rs
LuQQiu: Revisit all exec nodes and added the supports_limit_pushdown method, will work on lancedb side as well

LuQQiu:
Remaining question:
1. `scan_range_before_filter` with filters. I would expect it should limit what rows are scanned BEFORE any filtering, but the original implementation applies it AFTER index filtering https://github.com/lancedb/lance/blob/main/rust/lance/src/io/exec/filtered_read.rs#L626
2. push down `scan_range_after_filter` into "AtLeast" is "AtLeast" guaranteed matched rows > limit

LuQQiu:
@cmccabe @westonpace Addressed the comments, PTAL, thanks

rust/lance/src/io/exec/filtered_read.rs
```diff
@@ -520,158 +470,291 @@ impl FilteredReadStream {
...
-                            // we read, but we can't push a filtered scan range down
-                            let valid_ranges = row_id_sequence.mask_to_offset_ranges(row_id_mask);
-                            if let Some(deletion_vector) = &deletion_vector {
-                                Self::filter_deleted_rows(valid_ranges, deletion_vector)
-                            } else {
-                                valid_ranges
+            if to_take == 0 {
+                scan_planned_with_limit_pushed_down = true;
+                for (frag_id, ranges) in scan_push_down_fragments_to_read {
+                    fragments_to_read.insert(frag_id, ranges);
```
westonpace: Why do we need to insert here?  Don't we always insert in `fragments_to_read` whenever we insert into `scan_push_down_fragments_to_read`?

rust/lance/src/io/exec/filtered_read.rs
```diff
@@ -520,158 +470,291 @@ impl FilteredReadStream {
...
+            .map(|r| r.start)
+            .unwrap_or(0);
+        let mut to_take = options
+            .scan_range_after_filter
+            .as_ref()
+            .map(|r| r.end - r.start)
+            .unwrap_or(u64::MAX);
+
+        let mut fragments_to_read: HashMap<u32, Vec<Range<u64>>> = HashMap::new();
+        let mut scan_push_down_fragments_to_read: HashMap<u32, Vec<Range<u64>>> = HashMap::new();
```
westonpace, line 504: I'm not sure I understand the purpose of this variable.  Can you add a comment here explaining why we need two maps?

westonpace (approved):
I think this is correct, and the tests look right.  Thanks for taking on such a technically challenging task!

A few minor nits but let's go!

rust/lance/src/io/exec/filtered_read.rs
LuQQiu: Added the comments, one is for original result without scan_range_after_filter push down, one is for fragments with scan_range_after_filter push down

rust/lance/src/io/exec/filtered_read.rs
LuQQiu: when to_take is zero, we should use the fragments in `scan_push_down_fragments_to_read` only.

LuQQiu:
@westonpace 
The current approach for pushing down `scan_range_after_filter` has a critical limitation
 when the user cares about the result sequence.
 
Consider the scenario where fragments are ordered as [unindexed_fragment, indexed_fragment] (although unlikely, but we may have indexed_fragment-> unindexed -> indexed):

- With limit pushdown: The system may skip the unindexed fragment entirely and read only from the indexed fragment to satisfy the limit
- Without limit pushdown: The system reads from the unindexed fragment first (with filters and fetch all matched result), then the indexed fragment, preserving the original row ordering

This difference means that the same offset value produces different results depending on whether the limit was pushed down. The rows returned starting from offset N with pushdown may be completely different from the rows at offset N without pushdown, violating the expected sequence ordering that users rely on for consistent pagination and data retrieval.

PR #4796, opened 2025-09-23
test: stablize distance range test

BubbleCal:
the results may contain the same distances, this handles such case


java/lance-jni/Cargo.lock
jackye1995, line 1: this is unintended change?

BubbleCal: right, it's made by `cargo check`, should i restore it?

jackye1995: Oh okay then I guess let's just update it along the way

PR #4797, opened 2025-09-23
chore(rust): optimize shallow_clone parameter

majin1102:
I found that the interfaces using store_params as` Option<ObjectStoreParams>`

This will make this interace more friendly to use. This is an interface change and I thought it's better to make a seperated PR to track


majin1102:

PR #4811, opened 2025-09-25
feat(index/fts): build fst at write time instead of read time

Xuanwo:
We are now building the FST at write time instead of reading time, so the token set is constructed once during ingestion and only loaded during read operations. (Thank you @BubbleCal for this great idea!)

On a test with a 40M Wikipedia dataset, this change reduces the P95 latency by **32.3%**, from `5708.43 ms` to `3863.13 ms`.

The token set file's size reduced from `93.1 MB` to `46.2 MB`.

Before:

```shell
- open(ms): mean=84.96, p95=111.08, p99=113.98, min=69.28, max=114.70
- search(ms): mean=4993.26, p95=5708.43, p99=5778.55, min=4424.76, max=5796.08
- total(ms): mean=5078.22, p95=5814.05, p99=5891.44, min=4521.33, max=5910.78
- wall-clock(s): 25.98
```

image

After:

```shell
- open(ms): mean=79.45, p95=107.79, p99=113.32, min=63.61, max=114.70
- search(ms): mean=3386.18, p95=3863.13, p99=3962.93, min=3115.72, max=3987.87
- total(ms): mean=3465.63, p95=3970.92, p99=4076.24, min=3183.25, max=4102.57
- wall-clock(s): 18.00
```

image

---

### DIFF
--- a/rust/lance-index/src/scalar/inverted/index.rs
+++ b/rust/lance-index/src/scalar/inverted/index.rs
@@ -2576,6 +2576,10 @@ mod tests {
                 "params".to_owned(),
                 serde_json::to_string(&InvertedIndexParams::default()).unwrap(),
             ),
+            (
+                TOKEN_SET_VERSION_KEY.to_owned(),
+                serde_json::to_string(&CURRENT_TOKEN_SET_VERSION).unwrap(),
+            ),
         ]);
         let mut writer = store
             .new_index_file(METADATA_FILE, Arc::new(arrow_schema::Schema::empty()))
--- a/rust/lance-index/src/scalar/inverted/index.rs
+++ b/rust/lance-index/src/scalar/inverted/index.rs
@@ -950,10 +950,6 @@ impl TokenSet {
     }
 
     async fn load_v0(reader: Arc<dyn IndexReader>) -> Result<Self> {
-        let mut next_id = 0;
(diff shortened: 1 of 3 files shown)


rust/lance-index/src/scalar/inverted/index.rs
```diff
@@ -82,6 +82,10 @@ use crate::Index;
 use crate::{prefilter::PreFilter, scalar::inverted::iter::take_fst_keys};
 
 pub const INVERTED_INDEX_VERSION: u32 = 0;
+pub const TOKEN_SET_VERSION_V0: u32 = 0;
+pub const TOKEN_SET_VERSION_V1: u32 = 1;
```
BubbleCal: can we use an enum for this?
```
pub enum TokenSetFormat {
    Lance,
    FstBinary,
}
```

BubbleCal (approved):
impressive improvements!
just minor question

rust/lance-index/src/scalar/inverted/index.rs
Xuanwo: The format is still lance actually, but I think it’s good idea to change it to an enum instead.

Xuanwo: changed.

PR #4812, opened 2025-09-25
feat!: make handling of null semantics consistent in indexes

ddupg:
Closes: #4814

### DIFF
--- a/rust/lance-index/src/scalar/flat.rs
+++ b/rust/lance-index/src/scalar/flat.rs
@@ -216,13 +216,7 @@ impl ScalarIndex for FlatIndex {
         // Since we have all the values in memory we can use basic arrow-rs compute
         // functions to satisfy scalar queries.
         let mut predicate = match query {
-            SargableQuery::Equals(value) => {
-                if value.is_null() {
-                    arrow::compute::is_null(self.values())?
-                } else {
-                    arrow_ord::cmp::eq(self.values(), &value.to_scalar()?)?
-                }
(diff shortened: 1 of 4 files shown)


rust/lance-index/src/scalar/flat.rs
```diff
@@ -216,13 +216,7 @@ impl ScalarIndex for FlatIndex {
...
-                if value.is_null() {
-                    arrow::compute::is_null(self.values())?
-                } else {
-                    arrow_ord::cmp::eq(self.values(), &value.to_scalar()?)?
-                }
-            }
+            SargableQuery::Equals(value) => arrow_ord::cmp::eq(self.values(), &value.to_scalar()?)?,
             SargableQuery::IsNull() => arrow::compute::is_null(self.values())?,
             SargableQuery::IsIn(values) => {
                 let mut has_null = false;
```
westonpace, line 222: We should remove this flag as well?

westonpace (commented):
Thanks for looking into this.  I'm happy we are finally getting these consistent.  I agree that this way (aligning with SQL / DF) is the best approach.

westonpace:
I'm surprised that `BETWEEN NULL AND 3` doesn't optimize down into `NULL` in DF's optimizer.  I think that might be an eventual DF improvement but there is no harm handling it here too.

ddupg:
Thanks #4815 of @westonpace for fixing this issue, I closed this PR.

PR #4815, opened 2025-09-25
refactor: cowardly refuse to use scalar indexes on expressions with null literals

### LINKED ISSUE #4756: Scalar index queries handle NOT and NULL incorrectly
This happens with both BTREE and BITMAP:

```python
import pyarrow as pa
import lance

data = pa.table({"value": [True, False, None]})

ds.to_table(filter="NOT value")
```
```
value: int64
----
value: [[false]]
```

```python
ds.create_scalar_index("value", "BTREE")
ds.to_table(filter="NOT value")
```
```
pyarrow.Table
value: int64
----
value: [[false,null]]
```

```python
ds.create_scalar_index("value", "BITMAP", replace=True)
ds.to_table(filter="NOT value")
```
```
pyarrow.Table
value: int64
----
value: [[false,null]]
```

Similar things happen if you use integers and use condition like `NOT (value = 1)`

westonpace:
The current search logic with scalar indexes assumes that the list returned by a scalar index contains either:

* All the true rows and all other rows are false
* All the false rows and all other rows are true

This is a problem when dealing with nulls.  This is related to https://github.com/lancedb/lance/issues/4756

This PR doesn't fix #4756 however it does help avoid some areas where we are definitely likely to do the wrong thing.

### DIFF
--- a/rust/lance-index/src/scalar/expression.rs
+++ b/rust/lance-index/src/scalar/expression.rs
@@ -253,6 +253,16 @@ impl ScalarQueryParser for SargableQueryParser {
         low: &Bound<ScalarValue>,
         high: &Bound<ScalarValue>,
     ) -> Option<IndexedExpression> {
+        if let Bound::Included(val) | Bound::Excluded(val) = low {
+            if val.is_null() {
+                return None;
+            }
+        }
+        if let Bound::Included(val) | Bound::Excluded(val) = high {
+            if val.is_null() {
+                return None;
+            }
+        }
         let query = SargableQuery::Range(low.clone(), high.clone());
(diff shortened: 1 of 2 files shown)


rust/lance-index/src/scalar/expression.rs
```diff
@@ -1704,6 +1720,7 @@ mod tests {
         expected: Option<IndexedExpression>,
         optimize: bool,
     ) {
+        println!("Checking expression: {}", expr);
```
jackye1995: Should remove the println?

westonpace: Whoops.  Removed.

PR #4820, opened 2025-09-26
chore: add file format version to jni WriteParams

luohao:
This PR allows setting file format version via WriteParams from Java lib.

### DIFF
--- a/java/lance-jni/src/utils.rs
+++ b/java/lance-jni/src/utils.rs
@@ -35,6 +35,7 @@ pub fn extract_storage_options(
     Ok(storage_options)
 }
 
+#[allow(clippy::too_many_arguments)]
 pub fn extract_write_params(
     env: &mut JNIEnv,
     max_rows_per_file: &JObject,


jackye1995:
I think this is currently blocked by https://github.com/lancedb/lance/pull/4539, because 2.0 has some bugs in blocking take that prevents it from being used and we cannot just enable it in a buggy mode. Xuanwo said he will take a quick look, I can merge this once that issue is resolved.

luohao:
@westonpace Yes I think we should add this to both file and dataset API. ~I'll update the PR.~ The file writer doesn't take WriteParams as argument so looks like it's a bigger change. Let's merge this as is and I'll update file writer in a separate PR.

@jackye1995 I did some more tests and it triggers this bug when using fragment and transaction API. It works fine with the `Dataset.create()` tho, which is what I was using.

jackye1995:
Thanks @luohao and @ColdL for pushing this through!

PR #4825, opened 2025-09-26
feat: short circuit SQL parsing

timsaucer:
This PR avoids a problem encountered when creating a scalar index on a column that had a name that was not easily parseable by the SQL parser. If we are getting a raw expression that matches a column name in the schema, then return it directly. 

More generally I wonder if we should be doing this sql parsing at all when we know we have columns. Should `ProjectionPlan::from_expressions` actually take in instead `columns: &[(impl AsRef<str>, &Expr)]` ?

### DIFF
--- a/rust/lance-datafusion/src/planner.rs
+++ b/rust/lance-datafusion/src/planner.rs
@@ -839,6 +839,10 @@ impl Planner {
     /// Note: the returned expression must be passed through [optimize_filter()]
     /// before being passed to [create_physical_expr()].
     pub fn parse_expr(&self, expr: &str) -> Result<Expr> {
+        if let Ok(_) = self.schema.field_with_name(expr) {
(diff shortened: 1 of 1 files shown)


wjones127 (approved):
Seems reasonable. Thanks!

wjones127:
> More generally I wonder if we should be doing this sql parsing at all when we know we have columns. Should `ProjectionPlan::from_expressions` actually take in instead `columns: &[(impl AsRef<str>, &Expr)]` ?

Could you expand more on what you mean by that?

I think the intention of the expression was to allow indexing nested fields or fields within JSON blobs.

wjones127:
> This PR avoids a problem encountered when creating a scalar index on a column that had a name that was not easily parseable by the SQL parser.

Actually, before I merge, do you think this is worth a unit test?

PR #4832, opened 2025-09-28
refactor: rename Update operation fields

yanghua:


jackye1995:
I think a problem is that the train has already been released, it is hard to now change the name. What about we just add more documentation in the transaction protobuf definition and rust struct so people can better understand what those fields mean? I am working on fully refreshing the table spec, so once we update the proto the expectation is that it will be displayed in the spec doc in the future automatically.

yanghua:
> I think a problem is that the train has already been released, it is hard to now change the name.

What does the `train` mean? Do you mean the compatibility? 

It seems `protobuf ` is really an issue that would cause compatibility. Not sure if the community would introduce a fallback configuration mapping mechanism in the future. What about doing this:

* for fields in pb and rust struct, use the old names;
* for general variable naming them to be the new;

WDYT?

jackye1995:
> What does the train mean? Do you mean the compatibility?

yeah sorry I mean a [release train](https://en.wikipedia.org/wiki/Software_versioning#Release_train)

> for fields in pb and rust struct, use the old names;
> for general variable naming them to be the new;

I could see the point for `fields_modified`, but how is `bitmap_preserve_field_ids` better than `fields_for_preserving_frag_bitmap`? Aren't we just changing the position of the words?

yanghua:
> but how is bitmap_preserve_field_ids better than fields_for_preserving_frag_bitmap? Aren't we just changing the position of the words?

My original thought is that if `fields_modified` could be changed to be `bitmap_prune_field_ids`, as a similar purpose field, `fields_for_preserving_frag_bitmap` could be aligned to `bitmap_preserve_field_ids`. It also makes the naming more concise. But the main motivation is to align with `bitmap_prune_field_ids`.

yanghua:
@jackye1995 I did a trade-off: keep the old name in pb to keep compatibility, just renamed the other fields to make them follow the real semantics and better readability. WDYT?

PR #4833, opened 2025-09-28
fix: add serde feature flag dependency

timsaucer:
If you attempt to build with no default features for the `lance` crate (not just the workspace):

```
cargo build -p lance --no-default-features
```

Then you will get the following error:

```
error[E0277]: the trait bound `chrono::DateTime<chrono::Utc>: Deserialize<'_>` is not satisfied
   --> rust/lance/src/dataset.rs:175:10
    |
175 | #[derive(Deserialize, Serialize)]
    |          ^^^^^^^^^^^ the trait `Deserialize<'_>` is not implemented for `chrono::DateTime<chrono::Utc>`
    |
    = note: for local types consider adding `#[derive(serde::Deserialize)]` to your `chrono::DateTime<chrono::Utc>` type
    = note: for types from other crates check whether the crate offers a `serde` feature flag
```

This is a problem for projects that are using the lance crate and building with the `--no-default-features` flag. It can be overcome on the user's end by making this an explicit dependency during our builds, but it should probably be included here.


eddyxu:
Generally looks good. Pending CI

wjones127:
Fixed in https://github.com/lance-format/lance/pull/5110

Issue #4469, opened 2025-08-14
get_data_statistics brings TOO much overhead


qidian99:
## Problem

[#3337](https://github.com/lancedb/lance/pull/3337)

This calculation is blocking spark job from submitting way too long. E.g., the statistics calculation take over 80 minutes (in spark driver :() while the parallel data processing takes 60 minutes.

## Key code paths
- The JNI method calls a full dataset scan of v2 file metadata: lance/src/dataset/statistics.rs:36:69
```rust
impl DatasetStatisticsExt for Dataset {
    async fn calculate_data_stats(self: &Arc<Self>) -> Result<DataStatistics> {
        let field_ids = self.schema().field_ids();
        let mut field_stats: HashMap<u32, FieldStatistics> = /* init zeros */;
 
        if !self.is_legacy_storage() {
            let scan_scheduler = ScanScheduler::new(
                self.object_store.clone(),
                SchedulerConfig::max_bandwidth(self.object_store.as_ref()),
            );
            for fragment in self.fragments().as_ref() {
                let file_fragment = FileFragment::new(self.clone(), fragment.clone());
                file_fragment
                    .update_storage_stats(&mut field_stats, self.schema(), scan_scheduler.clone())
                    .await?;
            }
        }
        Ok(DataStatistics { fields: field_stats })
    }
}
```
- Each fragment opens v2 file readers to collect column-level sizes: lance/src/dataset/fragment.rs:768:783
```rust
pub(crate) async fn update_storage_stats(
    &self,
    field_stats: &mut HashMap<u32, FieldStatistics>,
    dataset_schema: &Schema,
    scan_scheduler: Arc<ScanScheduler>,
) -> Result<()> {
    for reader in self
        .open_readers(dataset_schema, &FragReadConfig::default().with_scan_scheduler(scan_scheduler))
        .await?
    {
        reader.update_storage_stats(field_stats);
    }
    Ok(())
}
```
- v2 reader summarizes per-column sizes by inspecting the column metadata (which requires opening the file, reading tail and column-metadata sections): lance-file/src/v2/reader.rs:369:390
```rust
pub fn file_statistics(&self) -> FileStatistics {
    let column_metadatas = &self.metadata().column_metadatas;
 
    let column_stats = column_metadatas
        .iter()
        .map(|col_metadata| {
            let num_pages = col_metadata.pages.len();
            let size_bytes = col_metadata
                .pages
                .iter()
[...]

Issue #4163, opened 2025-07-07
Column Statistics


westonpace:
We should expose column statistics for columns.  At a minimum this should include the minimum, maximum, and null count.  However, other statistics such as median, quantiles, histograms, lists of unique values, NaN count, may be included as well.

These will be useful both internally (for planning) as well as externally (for users)

westonpace:
@jackye1995 in response to https://github.com/lancedb/lance/issues/4162#issuecomment-3046265069

I think zone maps _could_ provide column statistics.  However, I'm uncertain if they are the best tool.  For example, we have generally tried to avoid any kind of potential I/O (even that likely to be cached) when running the planner.

If there were table level statistics (stored in the manifest or some dedicated statistics index) then they could be much smaller (one set of statistics per column) and used for high-level planning.  On the other hand, having too many versions of similar concepts could be confusing.

For now, I think it would be fine to add dataset-level APIs to get column-wide statistics (not broken out by zone) and implement those APIs using the zone index.  Then if we want to switch to some different approach later then we can.

jackye1995:
> If there were table level statistics (stored in the manifest or some dedicated statistics index) then they could be much smaller (one set of statistics per column) and used for high-level planning.

Totally agree. When I was thinking about zone map, in my mind it can be a generic structure to hold various column stats, and a zone can be at sub-fragment level, or per-fragment level or even more coarse grain, so that we can place coarse grain stats directly in the manifest (e.g. in index metadata) and the more fine-grained ones loaded on demand. But maybe that is too big of a leap. 

When you say a zone map, what is the granularity of a zone? is this every X number of rows in a file based on some writer configuration?

westonpace:
> When you say a zone map, what is the granularity of a zone? is this every X number of rows in a file based on some writer configuration?

My thinking was something on the order of 10000 rows but it probably should be configurable.  This parameter allows users to trade off RAM / index size for read amplification.  As the zone gets smaller (hundreds of rows or less) then a btree/bitmap index is probably a better choice.

> When I was thinking about zone map, in my mind it can be a generic structure to hold various column stats, and a zone can be at sub-fragment level, or per-fragment level or even more coarse grain, so that we can place coarse grain stats directly in the manifest (e.g. in index metadata) and the more fine-grained ones loaded on demand. But maybe that is too big of a leap.

Yes, coarse-grained stats in the manifest and fine-grained stats loaded on demand is what I was thinking.

jackye1995:
[...]

PR #4835, opened 2025-09-29
[WIP]: Create plan_splits method

jtuglu1:
WIP

Related to https://github.com/lancedb/lance/issues/4163, https://github.com/lancedb/lance/issues/4469

### DIFF
--- a/rust/lance-index/src/scalar/zonemap.rs
+++ b/rust/lance-index/src/scalar/zonemap.rs
@@ -34,7 +34,7 @@ use arrow_array::{new_empty_array, ArrayRef, RecordBatch, UInt32Array, UInt64Arr
 use arrow_schema::{DataType, Field};
 use datafusion::execution::SendableRecordBatchStream;
 use datafusion_common::ScalarValue;
-use std::{collections::HashMap, sync::Arc};
+use std::{collections::{HashMap, HashSet}, sync::Arc};
 
 use super::{AnyQuery, IndexStore, MetricsCollector, ScalarIndex, SearchResult};
 use crate::scalar::FragReuseIndex;
@@ -54,17 +54,17 @@ const ZONEMAP_INDEX_VERSION: u32 = 0;
 
 /// Basic stats about zonemap index
 #[derive(Debug, PartialEq, Clone)]
-struct ZoneMapStatistics {
-    min: ScalarValue,
-    max: ScalarValue,
-    null_count: u32,
+pub struct ZoneMapStatistics {
+    pub min: ScalarValue,
+    pub max: ScalarValue,
+    pub null_count: u32,
     // only apply to float type
-    nan_count: u32,
-    fragment_id: u64,
+    pub nan_count: u32,
+    pub fragment_id: u64,
     // zone_start is the start row of the zone in the fragment, also known
     // as local row offset
-    zone_start: u64,
-    zone_length: usize,
+    pub zone_start: u64,
+    pub zone_length: usize,
 }
 
 impl DeepSizeOf for ZoneMapStatistics {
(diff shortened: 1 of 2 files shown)


jackye1995:
Thanks for working on this! The current implementation feels tied to the zone map index, and I think it should be implemented more generically. 

I think we probably need to define the experience a bit more first. What I was thinking is that this can be a scanner level feature, so that similar to today users can do this to read data:

```rust

scanner.filter("i = 100").unwrap().project(&["i"]).unwrap();

scanner.try_into_batch().await.unwrap()
```

We will have something like

```rust

scanner.filter("i = 100").unwrap().project(&["i"]).unwrap();

scanner.plan_splits().await.unwrap()
```

This allows the query planning to go through the normal logical planning and optimization paths, and we can return a list of `Split`s based on some plan parameters.

The plan parameters could be things like the expected split size (so we can do the head tail bin pack to combine zones), if the split should contain complete fragments (more useful for writes), what indexes should the planning use (e.g. prefer zone map over btree if the column has both). 

With this, we can then expose APIs like `dataset.plan_splits()` in rust and python accordingly.

@westonpace I don't know if this aligns with what you think, let us know if you have better ideas.

Also cc @Jay-ju since this is related to https://github.com/lancedb/lance/pull/4735#discussion_r2353725197

Also cc @steFaiz since this is related to https://github.com/lancedb/lance/discussions/4000#discussioncomment-14518671

westonpace:
> @westonpace I don't know if this aligns with what you think, let us know if you have better ideas.

Yes, your description matches roughly how I would expect the process to go.  I think we could add this capability to the scanner.  We have `create_plan` but we could also have `create_estimation_plan` or something along those lines.  I think there are a few criteria:

If we create a `FilterPlan` then we should have a few different scenarios:

* No filter - split evenly based on fragment metadata
* Only index_query - Run the index query, split evenly based on the index search result
* Only non-index filter - Use column statistics (which we don't really have yet) combined with DF's cardinality estimation
* Both index query & non-index filter - Run index query and potentially refine with column statistics in the future

If we are going to assume we have zone maps then we can focus on the "only index_query" case.  So it should just be creating the initial filter plan, grabbing the index_query if it exists, evaluating it, and then splitting up the index search result.  The index search result is an allow list / block list of row addresses.

westonpace:
I also like the API you've provided I think that each "split" returned by `plan_splits` should (at a minimum) have a "rows to skip" and "rows to take" (e.g. limit/offset).

Probably another feature that is missing is the ability for the scanner to define a pre-filter limit/offset.  Right now we only allow the post-filter limit/offset to be specified.  A course pre-filter can be specified using fragment ids but it would be nice to have something more fine-grained.  The read node itself (filtered read) supports defining a pre-filter so this shouldn't be too big of a task (just plumbing through the scanner and API)

jtuglu1:
> If we are going to assume we have zone maps then we can focus on the "only index_query" case. So it should just be creating the initial filter plan, grabbing the index_query if it exists, evaluating it, and then splitting up the index search result. The index search result is an allow list / block list of row addresses.

@westonpace regarding this: did we still want to return zonemap statistic results to the caller?

Reason being is my understanding was we'd want to provide these stats, say, in the absence of a scalar index which could help filter out address ranges, etc. Or is the point that this isn't needed assuming the indices are evaluated during planning stage?

jackye1995:
> did we still want to return zonemap statistic results to the caller?

I think we can start with not returning that. We might want something like the residual (e.g. if there are some filters that were not fully applied in the planning phase and should be further evaluated), but we can always add that later.

Jay-ju:
A very useful idea; I am also facing the same problem.
Here I have a few questions I'd like to ask you.

1.  Why do we need to reconstruct the fragment? Is it because the original fragment object stores too much redundant data?

2.  why use ZoneMapStatistics here? How should this statistical information be used when returned to the computing engine? Can we directly filter the size of the data to be scanned according to the fragment, then pack it, and evenly form splits?

3. In addition, I would like to provide a logical basis for packing fragments, which is to calculate the approximate number of rows of the fragment through the index. Based on this, we can evenly distribute fragments of different sizes into groups. https://github.com/lancedb/lance/pull/4930

PR #4837, opened 2025-09-29
feat!: incremental indexing via SPFresh

BubbleCal:
This PR implements a new incremental indexing mechanism inspired by SPFresh, aiming to speed up vector index updates without requiring full reindexing.
It introduces dynamic partition split/join and reassignment to maintain index quality efficiently.

### Partition Split & Join
- Split triggered when `partition_len > max_part_length`
- Join triggered when `partition_len < min_part_length`
### Reassignment (LIRE protocol)
- Reassign vectors during split/join based on centroid distance.
### New Parameters
- max_part_length, 4x target_partition_size
- min_part_length, 25% target_partition_size
- reassign_range, 64 according to SPFresh paper
- max_delta_indices, TODO
### change `num_indices_to_merge` param of `OptimizeOptions`
- `num_indices_to_merge` is optional now, and the default is `None` (`1` before this)
- `num_indices_to_merge` is `None` indicates to use SPFresh LIRE protocol to automatically handle the delta indices


rust/lance-index/src/optimize.rs
```diff
@@ -2,21 +2,21 @@
...
+    /// If `num_indices_to_merge` is Some(N), the delta updates and latest N indices
     /// will be merged into one single index.
     ///
     /// It is up to the caller to decide how many indices to merge / keep. Callers can
     /// find out how many indices are there by calling [`Dataset::index_statistics`].
     ///
     /// A common usage pattern will be that, the caller can keep a large snapshot of the index of the base version,
     /// and accumulate a few delta indices, then merge them into the snapshot.
-    pub num_indices_to_merge: usize,
+    pub num_indices_to_merge: Option<usize>,
```
rpgreen, line 19: Is this a breaking change?

rpgreen (commented):
Does this impact all existing vector indices? Should we have a way to enable it / feature flag it?

rust/lance/src/index/vector/ivf.rs
```diff
@@ -609,7 +600,7 @@ async fn optimize_ivf_pq_indices(
 
     let start_pos = existing_indices
         .len()
-        .saturating_sub(options.num_indices_to_merge);
+        .saturating_sub(options.num_indices_to_merge.unwrap_or(1));
```
Xuanwo, line 603: The behavior here seems not align what we declared for `num_indices_to_merge: None`.

> If `num_indices_to_merge` is None, lance will create a new delta index if no partition is split, otherwise it will merge all delta indices.

Can you double check here?

rust/lance-index/src/optimize.rs
BubbleCal: Yes it is, say if no `num_indices_to_merge` provided, then the default behavior will be SPFresh (it's merge the new data with the last index before this)

BubbleCal:
> Does this impact all existing vector indices? Should we have a way to enable it / feature flag it?

It impacts most vector indices (except V1 indices, but I think today all indices are in v3 format)
It's fine for existing code if they specified `num_indices_to_merge`, as I know we do specify this param.

rust/lance/src/index/vector/ivf.rs
BubbleCal: Expected because this is the old v1 index, which doesn't support SPFresh, so unwrap this `Option` to be 1 (the default value before this PR)

BubbleCal:
Will merge this after benchmark

Issue #4826, opened 2025-09-26
Bug: Schema mismatch for filter with Nullable column


LuQQiu:
Script
```
#!/usr/bin/env python3
"""
Test script using Lance API directly with scanner function.
"""

import lance
import pyarrow as pa
import shutil
import os

def main():
    """Main function using Lance API and scanner."""
    
    # Clean up any existing dataset
    dataset_path = "./test_lance_dataset"
    if os.path.exists(dataset_path):
        shutil.rmtree(dataset_path)
    
    # Define schema with nullable struct fields
    struct_schema = pa.struct([
        pa.field("company_id", pa.int64(), nullable=True),
        pa.field("company_name", pa.string(), nullable=True),
        pa.field("count", pa.int64(), nullable=True),
    ])

    table_schema = pa.schema([
        ("cid", pa.string()),
        ("companies", pa.list_(struct_schema)) 
    ])

    # Test data with mixed None values
    test_data = {
        "cid": ["1", "2", "3"],
        "companies": [
            # Row 1: company_name has values, other fields are None
            [
                {"company_id": None, "company_name": "Google", "count": None},
                {"company_id": None, "company_name": "Microsoft", "count": None}
            ],
            # Row 2: All None
            [
                {"company_id": None, "company_name": None, "count": None},
                {"company_id": None, "company_name": None, "count": None}
            ],
            # Row 3: All have values
            [
                {"company_id": 100, "company_name": "Apple", "count": 50},
                {"company_id": 101, "company_name": "Amazon", "count": 60}
            ]
        ]
    }

    test_table = pa.table(test_data, schema=table_schema)
    
    try:
        # Create Lance dataset
        dataset = lance.write_dataset(test_table, dataset_path)
        print(f"✅ Successfully created Lance dataset at '{dataset_path}' with {dataset.count_rows()} rows")
        
        # Query 2: Scan with filter using Lance scanner
        try:
            print("\n2. Scan with filter cid='1' using Lance scanner:")
            # Create scanner with filter expression
            scanner = dataset.scanner(filter="cid='1'")
            results = scanner.to_table()
            print(f"   ✅ Success - Retrieved {len(results)} rows")
            for row in results.to_pylist():
                print(f"   Row: cid={row['cid']}, companies={row['companies']}")
        except Exception as e:
            print(f"   ❌ Failed with error: {e}")
            print("   This reproduces the schema mismatch error!")
            
            
    except Exception as e:
[...]

PR #4838, opened 2025-09-29
fix: schema mismatch when filtering nullable List<Struct> columns

Xuanwo:
This PR will fix https://github.com/lancedb/lance/issues/4826 where the schema mismatch which introduced by subtract_predicate. We should not remove columns needed for the final projection.

---

### DIFF
--- a/rust/lance/src/dataset/scanner.rs
+++ b/rust/lance/src/dataset/scanner.rs
@@ -5972,7 +5972,11 @@ mod test {
         data_storage_version: LanceFileVersion,
         #[values(false, true)] use_stable_row_ids: bool,
     ) {
-        let fixture = ScalarIndexTestFixture::new(data_storage_version, use_stable_row_ids).await;
+        let fixture = Box::pin(ScalarIndexTestFixture::new(
+            data_storage_version,
+            use_stable_row_ids,
+        ))
+        .await;
 
         for use_index in [false, true] {
             for use_projection in [false, true] {
@@ -6037,7 +6041,7 @@ mod test {
     #[rstest]
     #[tokio::test]
     async fn test_index_take_batch_size() {
-        let fixture = ScalarIndexTestFixture::new(LanceFileVersion::Stable, false).await;
+        let fixture = Box::pin(ScalarIndexTestFixture::new(LanceFileVersion::Stable, false)).await;
         let stream = fixture
             .dataset
             .scan()
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -496,15 +496,6 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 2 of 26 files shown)


rust/lance/src/dataset/scanner.rs
```diff
@@ -1609,11 +1609,19 @@ impl Scanner {
...
             .clone()
-            // Subtract columns that are expensive
-            .subtract_predicate(|f| !self.is_early_field(f))
+            // Subtract columns that are expensive, but never remove columns needed for the final projection
+            .subtract_predicate(|f| {
+                if protected_ids.contains(&f.id) {
+                    false
+                } else {
+                    !self.is_early_field(f)
+                }
```
westonpace: This seems to defeat the purpose of late materialization?  The idea is that we don't read those columns in the first read but instead read them after the filtering has been applied.

Xuanwo: Yeah, I found it breaks all our late materialization tests. I'm working on a better fix.

Xuanwo: cc @westonpace, PTAL for the new changes. I don’t actually like this solution, but I couldn’t find a better approach. Suggestions welcome.

westonpace (commented):
The test is good but I don't think the fix is targeting the right spot.  I think the bug is here: https://github.com/lancedb/lance/blob/28468c2f7ac99f3254fd0f32ad8a52f4b861f17e/rust/lance/src/io/exec/take.rs#L257

We should be taking the scanned batch (which has the company id and count) and the taken batch (which has the company_name) and merging them together with `merge_with_schema`.

The inputs to `merge_with_schema` are correct.  `self` has the id and count and `new_data` has the company_name.  However, the output is not (it only has id and count, e.g. the batch is unchanged).  So the batch output from the take exec does not match the take exec's output schema and we get the mismatch error reported.

Xuanwo:
> The test is good but I don't think the fix is targeting the right spot. I think the bug is here:
> 
> https://github.com/lancedb/lance/blob/28468c2f7ac99f3254fd0f32ad8a52f4b861f17e/rust/lance/src/io/exec/take.rs#L257
> 
> We should be taking the scanned batch (which has the company id and count) and the taken batch (which has the company_name) and merging them together with `merge_with_schema`.
> 
> The inputs to `merge_with_schema` are correct. `self` has the id and count and `new_data` has the company_name. However, the output is not (it only has id and count, e.g. the batch is unchanged). So the batch output from the take exec does not match the take exec's output schema and we get the mismatch error reported.

Thank for pointing out, I will take this route a look!

Xuanwo:
Got it, tracked in https://github.com/lancedb/lance/issues/4867

PR #4839, opened 2025-09-29
fix: correct null_count aggregation in boolean statistics collection

YinZheng-Sun:
The `get_boolean_statistics` function in `statistics.rs` had a bug where it would incorrectly aggregate `null_count` when processing multiple arrays. The function would break early when both `true` and `false` values were found in the first array, causing subsequent arrays' null counts to be skipped.


YinZheng-Sun:
> Thank you for working on this and sorry for missing this PR.
> 
> Can you fix clippy and format? This PR is almost ready to go.

done @Xuanwo

PR #4840, opened 2025-09-29
feat: allow reading blob descs in 2.1, fix bugs in FSL decoder when items are nullable

westonpace:
This also fixes a number of unit tests today that hard-code the version and adds a python utility to get the current stable version as a string (useful for testing to determine the expected value of certain fields)


protos/encodings_v2_1.proto
```diff
@@ -273,7 +273,7 @@ message Variable {
 // There is no output.
 message Constant {
   // The value (TODO: define encoding for literals?)
-  bytes value = 1;
+  optional bytes value = 1;
```
Xuanwo, line 276: What's this change for?

westonpace: So we can distinguish between NULL and certain empty literals (e.g. string / binary).  We could also invent a special encoding for string/binary literals but this seemed simplest.

westonpace: Note: at the moment the constant encoding is only used in a few very narrow places so it cannot be used today for empty strings / bytes.  This was just future-proofing the message to try and avoid a change here if we start using this message more often.

Xuanwo: Got it, thank you!

PR #4846, opened 2025-09-30
Add a chunk cache to avoid decoding duplicated miniblock chunks

niyue:
# Description
When miniblock encoding is used in a Lance file, reading the file with the v2 `FileReader` via the `read_stream_projected` API can become inefficient if the provided `ReadBatchParams::Indices` contains many nearby but non-contiguous row indices.
For example:
```
29, 168, 180, 194, 376, 559, 574, 665, 666, 667, ..., 968, 969, 970, 973, 975, ...
```
This kind of access pattern causes the same chunk to be decoded repeatedly, resulting in slow performance and high CPU usage.

# Solution
This PR introduces a lightweight single-entry cache in `DecodePageTask`. While it only helps when chunks are accessed in a somewhat sequential manner, row indices are typically sorted in ascending order, so the cache strikes a balance between saving memory and improving performance.

# Test
On a local setup with a Lance file containing 100k rows (each row with a text column of 200+ bytes):
* Reading 1700+ nearby but non-contiguous rows at random
* `zstd` is used for general compression
* With this change, performance improved by 3x–5x, depending on the dataset.


rust/lance-encoding/src/encodings/logical/primitive.rs
```diff
@@ -527,13 +527,42 @@ impl DecodePageTask for DecodeMiniBlockTask {
...
+            .chain(std::iter::once(false)) // the last one never needs caching
+            .collect();
+
+        // Cache for storing decoded chunks when beneficial
+        let mut chunk_cache: Option<(usize, DecodedMiniBlockChunk)> = None;
+
         // Now we iterate through each instruction and process it
-        for (instructions, chunk) in self.instructions.iter() {
-            // TODO: It's very possible that we have duplicate `buf` in self.instructions and we
-            // don't want to decode the buf again and again on the same thread.
```
niyue, line 536: This PR partially addresses this TODO. It improves performance unless chunks are accessed in a fully random pattern, which would require a HashMap-based cache at the cost of higher memory usage.

niyue:
This PR is ready for review. 

The CI still reports one test failure on `mac-build` (`index::vector::ivf::v2::tests::test_build_ivf_pq_4bit::case_3`), I tried rebased onto the latest of `main` branch, and it is still failing. But I’m unable to reproduce it locally on my MacBook. I’m not very familiar with the failing test, it passed in one of my previous pushes, and it doesn’t appear to touch the code paths I modified at all. Please let me know if you think it’s related — I’ll be happy to investigate further. Thanks!

niyue:
I confirmed that the test case `index::vector::ivf::v2::tests::test_build_ivf_pq_4bit::case_3` is flaky — when running it on the main branch (commit `7e65e8b0`) on my MacBook locally, 4 out of 50 runs failed.

niyue:
I rebased onto the latest main branch and encountered another flaky test `dataset::optimize::tests::test_read_btree_index_with_defer_index_remap` (as reported [here](https://github.com/lancedb/lance/issues/4925)). I’ll leave the code unchanged for now.

niyue:
@westonpace could you please help review this PR when you have a moment? I believe it partially addresses one of the comments you made in the code you previously wrote for this part

niyue:
Hi @westonpace, just wanted to gently check in to see if you’ve had a chance to take a look at this PR.
No rush at all — just making sure it didn’t slip through the cracks 😊

rust/lance-encoding/src/encodings/logical/primitive.rs
westonpace: Chunks _should_ always be accessed sequentially I believe.  We have a requirement at some point in the decoding process for offsets / ranges to be in sorted order.

westonpace:
Rebased and will merge on green

westonpace:
> Hi @westonpace, just wanted to gently check in to see if you’ve had a chance to take a look at this PR.
No rush at all — just making sure it didn’t slip through the cracks 😊

Sorry, this took way longer to get to than it should have.

niyue:
> Do you have a test case / benchmark of any kind that you've been running to verify performance?

I tested it within my application, as described in the `Test` section of this PR description. The workload involves a sequential but non-contiguous row access pattern, retrieving about 2% of the data. With this enhancement, I observed a 3x–5x performance improvement. I expect block-based compression schemes such as Zstd and LZ4 will show similar performance improvements.

I can try to add a benchmark within Lance itself to verify this improvement more systematically later. Thanks.

Issue #4780, opened 2025-09-19
Allow querying empty FTS index


wjones127:
I was thinking we couldn't support empty FTS indexes. But I was playing with Lance, and it seems we can get FTS results even for tokens that aren't in the index yet. So it's really just the case when we have an empty index with new rows that we aren't getting results.

```python
import pyarrow as pa
import lance

schema = pa.schema({'text': pa.string()})
ds = lance.write_dataset(pa.Table.from_pylist([], schema=schema), 'memory://')
ds.create_scalar_index('text', "INVERTED")

# Empty query on empty table
ds.to_table(full_text_query='hello')
```
```
pyarrow.Table
text: string
_score: float
----
text: []
_score: []
```

```python
# Empty query on non-empty table
ds.insert(pa.Table.from_pylist([{'text': 'hello!'}], schema=schema))
ds.to_table(full_text_query='hello')
```
```
pyarrow.Table
text: string
_score: float
----
text: []
_score: []
```

```python
# Optimized index, can now get result
ds.optimize.optimize_indices()
ds.to_table(full_text_query='hello')
```
```
pyarrow.Table
text: string
_score: float
----
text: [["hello!"]]
_score: [[0.28768212]]
```

```python
# New data, can get new results!
ds.insert(pa.Table.from_pylist([{'text': 'hello world'}], schema=schema))
ds.to_table(full_text_query='hello')
```
```
pyarrow.Table
text: string
_score: float
----
text: [["hello!","hello world"]]
_score: [[0.28768212,0.28768212]]
```

```python
# New data, can get new results even with new tokens!
ds.to_table(full_text_query='world')
```
```
pyarrow.Table
text: string
_score: float
----
text: [["hello world"]]
_score: [[0.28768212]]
```

wjones127:
@BubbleCal does this make sense? We can support this, right?

wojiaodoubao:
I did some research and found the reason. For newly added data that is not in the index, calculating the BM25 score requires using ```num_docs``` and ```average_doc_length``` from the index. However, for empty documents, num_docs is 0 and average_doc_length is undefined, which causes all scores to be undefined.

I think for scenarios where num_docs is zero or the column has not created a full-text index, we can first iterate through the input data to calculate num_docs and average_doc_length, and then calculate the BM25 score. Hi @BubbleCal , is this approach acceptable? I will try to implement it.

BubbleCal:
[...]

PR #4859, opened 2025-09-30
fix: fts match query on column without inverted index

wojiaodoubao:
Close https://github.com/lancedb/lance/issues/4780


wojiaodoubao:
I'll rebase this pr after https://github.com/lancedb/lance/pull/4752 is merged.

wojiaodoubao:
Hi @jackye1995 @BubbleCal , could you help review this pr when you have time, thank very much !

rust/lance/src/dataset.rs
```diff
@@ -5768,6 +5768,189 @@ mod tests {
...
+            .into(),
+            vec![
+                Arc::new(title_col) as ArrayRef,
+                Arc::new(content_col) as ArrayRef,
+            ],
+        )
+        .unwrap();
+        let schema = batch.schema();
+        let batches = RecordBatchIterator::new(vec![batch].into_iter().map(Ok), schema);
+        let mut dataset = Dataset::write(batches, &test_uri, None).await.unwrap();
```
wjones127: I'd recommend for most tests to just use the in-memory store:
```suggestion
        let mut dataset = Dataset::write(batches, "memory://", None).await.unwrap();
```

rust/lance/src/dataset.rs
```diff
@@ -5768,6 +5768,189 @@ mod tests {
...
     }
 
+    #[tokio::test]
+    async fn test_fts_unindexed_data_on_empty_index() {
+        let test_uri = TempStrDir::default();
+
+        // Empty dataset with fts index
+        let params = InvertedIndexParams::default();
+        let title_col = GenericStringArray::<i32>::from(Vec::<&str>::new());
+        let content_col = GenericStringArray::<i32>::from(Vec::<&str>::new());
```
wjones127: This could just be `StringArray`: 
```suggestion
        let title_col = StringArray::from(Vec::<&str>::new());
        let content_col = StringArray::from(Vec::<&str>::new());
```

rust/lance/src/io/exec/fts.rs
```diff
@@ -1122,7 +1136,8 @@ pub mod tests {
                 "text",
                 lance_datagen::array::rand_utf8(ByteCount::from(10), false),
             )
-            .into_df_exec(RowCount::from(15), BatchCount::from(2));
+            .into_df_repeat_exec(RowCount::from(15), BatchCount::from(2))
+            .unwrap();
```
wjones127: Why was this change necessary?

I don't see it being re-used below. Unless it means the query itself has to scan the input data twice. If that's the case, it seems worrying to me.

rust/lance-index/src/scalar/inverted/scorer.rs
```diff
@@ -22,13 +22,106 @@ pub trait Scorer: Send + Sync {
...
+impl<'a> BM25Scorer<'a> {
+    pub fn num_docs(&self) -> usize {
+        match self {
+            BM25Scorer::Index(scorer) => scorer.num_docs(),
+            BM25Scorer::Memory(scorer) => scorer.num_docs(),
+            BM25Scorer::Union(left, right) => left.num_docs() + right.num_docs(),
+        }
+    }
+
+    pub fn avgdl(&self) -> f32 {
```
wjones127: This method name is very hard to read. Not even sure if I got the name right? Could you make it full words?
```suggestion
    pub fn avg_doc_length(&self) -> f32 {
```

rust/lance-index/src/scalar/inverted/scorer.rs
```diff
@@ -22,13 +22,106 @@ pub trait Scorer: Send + Sync {
...
+            BM25Scorer::Memory(scorer) => scorer.avgdl(),
+            BM25Scorer::Union(left, right) => {
+                (left.avgdl() * left.num_docs() as f32 + right.avgdl() * right.num_docs() as f32)
+                    / (left.num_docs() + right.num_docs()) as f32
+            }
+        }
+    }
+
+    // the number of documents that contain the token
+    pub fn nq(&self, token: &str) -> usize {
```
wjones127: Same thing here about the method name. I like that you have a comment describing it though.

rust/lance-index/src/scalar/inverted/index.rs
```diff
@@ -2390,17 +2446,42 @@ pub fn flat_bm25_search_stream(
...
     doc_col: String,
     query: String,
-    index: &InvertedIndex,
+    index: &Option<InvertedIndex>,
+    unindexed_scorer: MemoryBM25Scorer,
 ) -> SendableRecordBatchStream {
-    let mut tokenizer = Box::new(index.tokenizer.clone());
+    let mut tokenizer = match index {
+        Some(index) => index.tokenizer(),
+        // TODO: allow users to specify a tokenizer when querying columns without an inverted index.
```
wjones127: I think creating an index is how they should specify the tokenizer. Even if they just pass `train=False`, we at least get the configuration.

wjones127 (commented):
It's exciting that we'll have this soon.

My main concern is understanding if and why `into_df_repeat_exec` is necessary.

rust/lance/src/io/exec/fts.rs
wojiaodoubao: Yes, the query itself has to scan the input data twice. When we compute score for document in fts, we need some statistics of all documents: 
* number of documents
* average document length
* number of documents containing the query term.

The statistics data are saved at BM25Scorer(https://github.com/lancedb/lance/pull/4859/files#diff-60d4f0739e08b11e907ca92afb4b5308809e2a4f71f53418074de3b5f3dbda46R25).

Since we want to support full text search when there is no index or the index is empty, we can't rely on the BM25Scorer based on index. We have to scan first to obtain the BM25Scorer, and then scan a second time to score each document.

Thanks @wjones127 for pointing out the scan twice issue. Scanning unindexed data twice is indeed an issue that cannot be ignored, so I came up with an alternative solution: 
we can incrementally calculate the BM25Scorer while scanning the unindexed data. When scoring, we can combine the incremental BM25Scorer with the index BM25Scorer. By sacrificing some score accuracy, we can only scanning unindexed data once. In our original implementation, we use the index BM25Scorer for scoring unindexed data, which actually already sacrifices some accuracy. Therefore, this solution might be feasible.

@wjones127 what do you think?

wjones127: If we only have to scan twice when there is not index data, then I think that's acceptable.

wojiaodoubao: Hi @wjones127 , I updated this pr with the new approach which only scan once. Please help review when you have time, thanks very much!

PR #4860, opened 2025-09-30
feat: upgrade url crate, rework temporary directory utils

westonpace:
There is a behavior change in url 2.5.7 that is blocking a datafusion upgrade.  This PR attempts to just solve that behavior change.  Unfortunately, the fix here is to just avoid nasty Windows paths while we wait for guidance from upstream.


cmccabe:
What is the current state of using slashes in paths on Windows? It used to be that you were supposed to use https://doc.rust-lang.org/std/path/constant.MAIN_SEPARATOR.html but I don't know if we still want to do that.

westonpace:
> What is the current state of using slashes in paths on Windows? It used to be that you were supposed to use https://doc.rust-lang.org/std/path/constant.MAIN_SEPARATOR.html but I don't know if we still want to do that.

The following should work (even on Windows):

* `C:/foo/bar`
* `/foo/bar` (not sure if this uses a "default disk" or if it is always C:)
* `file:///C:/foo/bar`
* `foo\bar`

The following will fail:
* `C:\foo\bar`
* `file:///C:\foo\bar`

So my recommendation would be to always use `/`.

python/python/lance/vector.py
```diff
@@ -734,6 +734,11 @@ def _partition_and_pq_codes_assignment() -> Iterable[pa.RecordBatch]:
     )
     if dst_dataset_uri is None:
         dst_dataset_uri = tempfile.mkdtemp()
+        if re.search(r".:\\", dst_dataset_uri) is not None:
+            # Hack for Windows due to
+            # https://github.com/apache/arrow-rs-object-store/issues/499
+            dst_dataset_uri = dst_dataset_uri.replace("\\", "/", 1)
+        print(f"dst_dataset_uri={dst_dataset_uri}")
```
jackye1995: nit: remove print

rust/lance-core/src/utils/tempfile.rs
```diff
@@ -0,0 +1,340 @@
...
+    }
+}
+
+/// A temporary file that is exposed as a standard library path
+///
+/// This is a wrapper around [`TempFile`] that exposes the path as a standard library path.
+/// It is useful when you need to create a temporary file that is only used as a standard library path.
+#[derive(Default)]
+pub struct TempStdFile {
+    _tempfile: TempFile,
```
jackye1995: nit: since this variable is used, this should not have underscore prefix? (similar comment for other `_tempfile` and `_tempdir`

python/python/lance/vector.py
westonpace: Oops, done.

rust/lance-core/src/utils/tempfile.rs
westonpace: Some of them really are use but `TempStdFile` and `TempStdPath` I modified to remove the underscore.

PR #4866, opened 2025-10-01
feat: fts supports custom stop words

wojiaodoubao:
Currently only languages built-in tantivy supports stop words. We can let fts support custom stop words, this is useful for other languages like chinese (https://github.com/goto456/stopwords/).


rust/lance-index/src/scalar/inverted/tokenizer.rs
```diff
@@ -203,6 +211,11 @@ impl InvertedIndexParams {
         self
     }
 
+    pub fn custom_stop_words_file(mut self, custom_stop_words_file: Option<String>) -> Self {
+        self.custom_stop_words_file = custom_stop_words_file;
+        self
+    }
```
wjones127: You should add a docstring here, since this is the public method. Could you make sure to describe the format the file should be in?

wjones127 (commented):
Having a separate file is interesting. However, what if we just passed a list/array of strings instead. This would be beneficial in that:

1. The user can choose how the stop words are serialized and where they are stored.
2. The API is clearer, users don't have to worry whether their file is in the right format.

wojiaodoubao:
> Having a separate file is interesting. However, what if we just passed a list/array of strings instead. This would be beneficial in that:
> 
> 1. The user can choose how the stop words are serialized and where they are stored.
> 2. The API is clearer, users don't have to worry whether their file is in the right format.

Hi @wjones127 , thanks you comments and nice suggestions. It makes sense to me. I was initially torn between directly passing in stop words or using a file. The advantage of using a file is that, when performing full-text search in Chinese, the stop words are usually the same and can be reused instead of being written repeatedly every time an index is created. However, it is  inconvenient, as users need to ensure that the stop word file is in the correct location; otherwise, the full-text search will fail.

Users can read the stop-words file themselves to get the stop words list. So Lance doesn't need to provide the ability to read stop-word files. I think directly passing in stop words is a better solution. Let me make the changes.

> ... passed a list/array of strings instead ...
> The user can choose how the stop words are serialized and where they are stored.

I have a small question I'd like to ask. I want to save the stop words directly passed in by the user as a list(string) in InvertedIndexParams. When you mentioned that "users can choose how the stop words are serialized and where they are stored," how can we allow users to make this choice?

wjones127:
> When you mentioned that "users can choose how the stop words are serialized and where they are stored," how can we allow users to make this choice?

In you original API, user would have to save the stop words in a particular format (that we choose), and then supply the file like:

```python
ds = lance.write_dataset(data, tmp_path, mode="overwrite")
ds.create_scalar_index(
    "text",
    "INVERTED",
    base_tokenizer="jieba/default",
    remove_stop_words=True,
    custom_stop_words_file=str(stop_words_path),
)
```

But with the new API, the user can still read from a file with their own code. That means they can save them however they want, and read it with their own code:

```python
with open("my_tokens.json) as f:
    stop_words = json.load(f)['stop_word']

ds.create_scalar_index(
    "text",
    "INVERTED",
    base_tokenizer="jieba/default",
    remove_stop_words=True,
    custom_stop_words=stop_words,
)
```

wojiaodoubao:
Hi @wjones127 , sorry to bother. Could you help merge this when you have time, thanks very much!

PR #4871, opened 2025-10-01
fix: ensure fragment IDs are sorted in load_training_data

jtuglu1:
This is a fix in response to https://github.com/lancedb/lance-ray/issues/48.
Also, the dedup() operation which doesn't do a full de-dup unless the container is sorted.


rust/lance/src/index/create.rs
```diff
@@ -455,11 +455,17 @@ mod tests {
         );
 
         // Test fragments() method with specific fragment IDs
-        let selected_fragments = vec![fragment_ids[0], fragment_ids[1]];
+        let mut selected_fragments = vec![fragment_ids[0], fragment_ids[1]];
```
jackye1995: Can we also add a few duplicates to ensure the dedupe logic is correct?

rust/lance/src/index/create.rs
```diff
@@ -454,12 +454,21 @@ mod tests {
...
-        // Test fragments() method with specific fragment IDs
-        let selected_fragments = vec![fragment_ids[0], fragment_ids[1]];
+        // Test fragments() method with specific fragment IDs and ensure duplicate/out-of-order fragments are handled properly
+        let selected_fragments = vec![
+            fragment_ids[2],
+            fragment_ids[1],
+            fragment_ids[0],
+            fragment_ids[1],
+        ];
+        let mut selected_fragments_expected = selected_fragments.clone();
```
jackye1995: I think the expected should be spelled out as

```rust
vec![
            fragment_ids[0],
            fragment_ids[1],
            fragment_ids[2],
        ];
```

so that we are sure sort and dedupe are behaving correctly

rust/lance/src/index/scalar.rs
```diff
@@ -201,7 +201,12 @@ pub(crate) async fn load_training_data(
     if train {
         // Use the training request to scan data with fragment filtering
         if let Some(ref fragment_ids) = training_request.fragment_ids {
-            let fragment_ids = fragment_ids.clone().into_iter().dedup().collect_vec();
+            let fragment_ids = fragment_ids
+                .clone()
+                .into_iter()
+                .sorted()
```
jackye1995, line 207: actually, should we just use `unique()`? https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.unique

jtuglu1: Yes, I saw this, but that method allocates a hash set in the iterator to track seen values. Since we already need to sort the vector, we can use sort()/dedup() to avoid the O(n) auxiliary memory.

jackye1995: oh I see I was thinking if we can just get fragments based on the unordered, but it will be slower than `get_frags_from_ordered_ids` so sorting is still better.

jackye1995:
Looks like Mac runner got stuck, merging.

PR #4875, opened 2025-10-01
fix: don't retain indices that have no rows

### LINKED ISSUE #4892: Vector indices that cover no active fragments aren't removed
I noticed in https://github.com/lancedb/lance/issues/4842#issuecomment-3357706910 that some users had vector indices that were no longer covering any rows. You could see this in the `index_statistics()` output:

```
'num_indexed_rows_per_delta': [..., 0]
```

This is a problem in the method `Transaction::retain_relevant_indices()`.

wjones127:
Fixes #4892

Adding a bunch more tests since I relize this logic is complex and want to make sure there aren't additional bugs.

### DIFF
--- a/rust/lance/src/dataset.rs
+++ b/rust/lance/src/dataset.rs
@@ -4832,17 +4832,19 @@ mod tests {
 
         dataset.delete("true").await.unwrap();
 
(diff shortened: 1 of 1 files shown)


jackye1995 (approved):
looks good to me

Issue #4831, opened 2025-09-28
Can't use JSON fields with merge_insert


ztorchan:
Hi.

https://github.com/lancedb/lance/blob/30dcb5b8eb01e2fd4d05824b908d4174bab549d6/rust/lance/src/dataset/write/merge_insert/exec/write.rs#L361-L373

I found that the metadata of data column field was not copied into the output schema in `FullSchemaMergeInsertExec::prepare_stream_schema`, resulting in metadata loss. 
This results in an error when performing mergeinsert with a json column in the dataset. like:
`Query Execution error: Execution error: LanceError(Schema): Attempt to project field by different types: LargeBinary and Utf8`

wjones127:
Normally, we don't care about the field metadata of data you are inserting. We will just throw it away in the end and keep the metadata that's already in the table schema.

But JSON columns are special, because we detect based on the metadata that we should convert them to a JSONB encoding. When we stripped that metadata away, we dropped the conversion from string JSON (Utf8) to JSONB (LargeBinary), and that caused the type mismatch.

westonpace:
> But JSON columns are special, because we detect based on the metadata that we should convert them to a JSONB encoding. When we stripped that metadata away, we dropped the conversion from string JSON (Utf8) to JSONB (LargeBinary), and that caused the type mismatch.

Just to be clear you are stating that the conversion should be based on the field metadata of the dataset schema right?  We should always ignore the field metadata of the data itself?

wjones127:
> Just to be clear you are stating that the conversion should be based on the field metadata of the dataset schema right? We should always ignore the field metadata of the data itself?

Oh that's a good point. The consequence of this is a user should be able to do:


```python
# Create dataset using JSON type
schema = pa.schema({'data': pa.json()})
data = pa.table({'data': [json.dumps(dict(x=3))], schema=schema)
ds = lance.write_dataset(data, "memory://")

# Insert plain JSON strings -> will be automatically converted
new_data = pa.table({'data': [json.dumps(dict(x=5))]})
ds.merge_insert(on=[...]).execute(new_data)
```

Maxwell-Guo:
let me take a try ? @wjones127 @westonpace

mackrorysd:
Looks like we can close this issue, right?

westonpace:
> Looks like we can close this issue, right?

Maybe...I'm not certain all merge-insert paths work (there are several of them now) but I am not aware of any that fail.

[...]

PR #4877, opened 2025-10-02
fix: field metadata lost in full schema mergeinsert

ztorchan:
https://github.com/lancedb/lance/issues/4831

### DIFF
--- a/rust/lance/src/dataset/write/merge_insert.rs
+++ b/rust/lance/src/dataset/write/merge_insert.rs
@@ -2059,7 +2059,8 @@ mod tests {
     use datafusion::common::Column;
     use datafusion_physical_plan::stream::RecordBatchStreamAdapter;
     use futures::{future::try_join_all, FutureExt, StreamExt, TryStreamExt};
-    use lance_arrow::FixedSizeListArrayExt;
+    use lance_arrow::json::ARROW_JSON_EXT_NAME;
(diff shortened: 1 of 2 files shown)


wjones127:
Hi @ztorchan. In the issue you said you ran into an error `Query Execution error: Execution error: LanceError(Schema): Attempt to project field by different types: LargeBinary and Utf8`. Could you add a unit test that would produce this error without this change?

ztorchan:
Hi @wjones127, I have added a merge insert unit test for json in `python/python/tests/test_json.py`.

Without this change, it will run into an error like this:

screen_shoot

And with this change, it runs correctly.

ztorchan:
Is there any progress on this problem?

wjones127:
Thanks for adding the test. We were confused as to what bug this was exactly fixing. Now that I see it's a JSON column it makes a lot more sense. Normally we don't care about the field metadata of the source data, but in the case of JSON we do.

wjones127:
I added some more tests. I think you also need to get the case of the subschema merge insert working.

Also, Weston had a good point about where the metadata should come from: https://github.com/lancedb/lance/issues/4831#issuecomment-3377752002. Instead of preserving the input metadata, we should probably match and pull the table schema metadata.

There is a tricky bit I found, where the input schema metadata has the `arrow.json` extension type, but we translate it to the `lance.json` extension type, so they aren't exactly equal.

wjones127:
Closing this for now. If you want to keep working on this, feel free to re-open it.

PR #4879, opened 2025-10-02
feat: change one-pass GPU ivf-pq indexing as default

eddyxu:


westonpace (approved):
Seems fine once tests pass

Issue #4842, opened 2025-09-29
Vector index failed with panicked at 'called `Option::unwrap()` on a `None` value': /src/lance/rust/lance-index/src/vector/pq.rs:162


LuQQiu:
```
thread 'tokio-runtime-worker' panicked at 'called `Option::unwrap()` on a `None` value': /src/lance/rust/lance-index/src/vector/pq.rs:162
   0: log_panics::Config::install_panic_hook::{{closure}}
   1: std::panicking::rust_panic_with_hook
   2: std::panicking::begin_panic_handler::{{closure}}
   3: std::sys::backtrace::__rust_end_short_backtrace
   4: __rustc::rust_begin_unwind
   5: core::panicking::panic_fmt
   6: core::panicking::panic
   7: core::option::unwrap_failed
   8: <core::iter::adapters::map::Map<I,F> as core::iter::traits::iterator::Iterator>::fold
   9: <alloc::vec::Vec<T> as alloc::vec::spec_from_iter::SpecFromIter<T,I>>::from_iter
  10: <lance_table::utils::ExactSize<I> as core::iter::traits::iterator::Iterator>::next
  11: <alloc::vec::Vec<T> as alloc::vec::spec_from_iter::SpecFromIter<T,I>>::from_iter
  12: lance_index::vector::pq::ProductQuantizer::transform_impl
  13: <lance_index::vector::pq::ProductQuantizer as lance_index::vector::quantizer::Quantization>::quantize
  14: <lance_index::vector::pq::transform::PQTransformer as lance_index::vector::transform::Transformer>::transform
  15: <lance_index::vector::ivf::IvfTransformer as lance_index::vector::transform::Transformer>::transform
  16: lance::index::vector::builder::IvfIndexBuilder<S,Q>::shuffle_data::{{closure}}::{{closure}}::{{closure}}
  17: tokio::runtime::task::core::Core<T,S>::poll
  18: tokio::runtime::task::raw::poll
  19: tokio::runtime::scheduler::multi_thread::worker::Context::run_task
  20: tokio::runtime::scheduler::multi_thread::worker::Context::run
  21: tokio::runtime::context::scoped::Scoped<T>::set
  22: tokio::runtime::context::runtime::enter_runtime
  23: tokio::runtime::scheduler::multi_thread::worker::run
  24: <tokio::runtime::blocking::task::BlockingTask<T> as core::future::future::Future>::poll
  25: tokio::runtime::task::core::Core<T,S>::poll
[...]

PR #4890, opened 2025-10-02
fix: filter out vectors that could produce infinite l2 distances

wjones127:
This is a quick fix for #4842.

I think a long-term fix will need to make this aware of the distance type. Or maybe we filter when actually computing distances later, so we don't have false-positives. I know this mostly applies for l2. Cosine (IIRC) normalizes the vectors and doesn't have this problem.

### DIFF
--- a/rust/lance-index/src/vector/transform.rs
+++ b/rust/lance-index/src/vector/transform.rs
@@ -97,7 +97,15 @@ impl KeepFiniteVectors {
     }
 }
 
-fn is_all_finite<T: ArrowPrimitiveType>(arr: &dyn Array) -> bool
+// We compute L2 distance for f16 in f32 space, so f16 vectors will not overflow.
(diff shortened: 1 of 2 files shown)


BubbleCal (commented):
Thanks for finding this!

I think we'd still run into this issue with cosine/dot because normalization needs to calculate the length of the vectors which is essentially the same as calculating the distance to zero vector.

PQ is trained on sample data, if the user was running into this consistently, that means most of their vectors are too long, the solution should be to switch data type to f64 vector.

And in the case of that all vectors are big but close, the L2 distance won't be INF, we should keep these vectors, so I don't think we should filter out vectors by abs values.

This issue causes unwrap on None, because argmin implementation sets the initial min value to INF and tries to update the min value and get an index, maybe we can just set the initial value to the first value to avoid the case of all INFs?

PR #4895, opened 2025-10-03
feat: support tracking newly inserted and updated rows between versions

jackye1995:
Based on #4741 

Related discussion: https://github.com/lancedb/lance/discussions/4935

This PR introduces 2 metadata columns: `_row_created_at_version` and `_row_updated_at_version`. This allows us to track and get newly inserted and updated rows between 2 versions. We also update the `DatasetDelta` API to support this functionality accordingly.


jackye1995:
Close as merged to #4741

Issue #4896, opened 2025-10-03
Add general compression for large block arrays


westonpace:
Currently block compression is limited to bitpacking.  This is mainly because the only things we block compress are dictionaries (which are usually small-ish) and repdef levels (which are usually highly bitpackable).

Adding general compression should be a quick win to shore up our support when we have large dictionary arrays.

lyang24:
give my attempt - in my mind the hardest part is how to integrate with current `create_block_compressor` call.

currently i had if compression scheme is lz4 or zstd we compress the non bitpackable data - not sure if its ideal.

Xuanwo:
Has this issue been resolved?

westonpace:
Yes, I think we have general block compression for dictionaries now.

PR #4900, opened 2025-10-05
feat: support general compression zstd/lz4 in blocks

lyang24:
touches https://github.com/lancedb/lance/issues/4896

### DIFF
--- a/rust/lance-encoding/src/compression.rs
+++ b/rust/lance-encoding/src/compression.rs
@@ -54,6 +54,7 @@ use crate::{
         ProtobufUtils21,
     },
     statistics::{GetStat, Stat},
+    version::LanceFileVersion,
 };
 
 use arrow_array::{cast::AsArray, types::UInt64Type};
@@ -65,6 +66,9 @@ use std::str::FromStr;
 /// Default threshold for RLE compression selection.
 /// RLE is chosen when the run count is less than this fraction of total values.
 const DEFAULT_RLE_COMPRESSION_THRESHOLD: f64 = 0.5;
+
+// Minimum block size (32kb) to trigger general block compression
+const MIN_BLOCK_SIZE_FOR_GENERAL_COMPRESSION: u64 = 32 * 1024;
 
 /// Trait for compression algorithms that compress an entire block of data into one opaque
 /// and self-described chunk.
@@ -125,6 +129,8 @@ pub trait CompressionStrategy: Send + Sync + std::fmt::Debug {
 pub struct DefaultCompressionStrategy {
     /// User-configured compression parameters
     params: CompressionParams,
+    /// The lance file version for compatibilities.
+    version: LanceFileVersion,
 }
 
 fn try_bss_for_mini_block(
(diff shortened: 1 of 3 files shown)


rust/lance-encoding/src/compression.rs
```diff
@@ -475,7 +478,9 @@ impl CompressionStrategy for DefaultCompressionStrategy {
         field: &Field,
         data: &DataBlock,
     ) -> Result<(Box<dyn BlockCompressor>, CompressiveEncoding)> {
-        match data {
+        let field_params = Self::parse_field_metadata(field);
```
westonpace: We should probably be following the same pattern we follow in miniblock (in fact, this should probably be a helper function we use in all thread `create_...` functions:

```
        let mut field_params = self
            .params
            .get_field_params(&field.name, &field.data_type());
        // Override with field metadata if present (highest priority)
        let metadata_params = Self::parse_field_metadata(field);
        field_params.merge(&metadata_params);
```

rust/lance-encoding/src/encodings/physical/general.rs
```diff
@@ -134,14 +134,66 @@ impl MiniBlockDecompressor for GeneralMiniBlockDecompressor {
...
+
+impl BlockDecompressor for GeneralBlockDecompressor {
+    fn decompress(&self, data: LanceBuffer, num_values: u64) -> Result<DataBlock> {
+        let decompressor = GeneralBufferCompressor::get_compressor(self.compression)?;
+        let mut buf = vec![];
+        decompressor.decompress(&data, &mut buf)?;
+
+        self.inner.decompress(LanceBuffer::from(buf), num_values)
+    }
+}
```
westonpace: Minor nit: I could go either way on this but this could probably be a single `GeneralBlock` struct that impls both traits.

Or perhaps, even better, we could just add these traits to `CompressedBufferEncoder` instead of making any new structs?

rust/lance-encoding/src/encodings/physical/general.rs
```diff
@@ -660,4 +712,114 @@ mod tests {
...
+            Box::new(VariableEncoder::default()),
+            CompressionConfig {
+                scheme: CompressionScheme::Zstd,
+                level: Some(3),
+            },
+        );
+
+        let compressed = compressor.compress(block.clone()).unwrap();
+
+        // Test decompression - FIXED: Use BinaryBlockDecompressor instead of VariableDecoder
```
westonpace: What does FIXED mean here?

westonpace (changes requested):
This looks good!

I think there are two general issues we will need to tackle before we merge.

First, we need to add a check that the version is >= 2.2.  Now that 2.1 is stable we can't make forwards-compatibility breaking changes.  In other words, we can't have a 2.1 writer start writing with general block compression because the oldest 2.1 readers won't be able to reader it.

To do this we can add `version: LanceFileVersion` to `DefaultCompressionStrategy`.

Second, and a bit more subtle, is that we can't only rely on field metadata to decide we should use general compression.  It's a bit subtle but we don't actually use block compression at the field level.  This is because it is unsuitable for random access.  Instead, we use block compression for individual parts that we will need to decompress in their entirety.

There are two parts we compress with block compression today:

1. RepDef levels for mini-block encoding.  We can see that [here](https://github.com/lancedb/lance/blob/main/rust/lance-encoding/src/encodings/logical/primitive.rs#L3566-L3570)

Note that we create a dummy field (and so it will never have field configuration).  This particular usage will probably never be a good candidate for general compression anyways (if we apply bitpacking and RLE to these levels that is almost certainly good enough).

2. Dictionaries.  We can see that [here](github.com/lancedb/lance/blob/main/rust/lance-encoding/src/encodings/logical/primitive.rs#L3787-L3791)

Here I think we _do_ want to apply block compression, since dictionaries can potentially be large and, being strings, can also potentially be good candidates for general compression.  Unfortunately, as you can see from that link, we once again use a dummy field, and so there will never be any `field_params` set on the field.

I think the easiest thing to do would be to put a size criteria in here.  If the data block is large (32KiB or more) then we apply general compression, even if there are field_params.

So something like...

```
// User-requested compression (unused today but perhaps still used
// in the future someday)
if let Some(compression) = field_params.compression... {
  ...
}
// Automatic compression because data is large enough to justify it
if data.data_size() > 32 * 1024 {
  // Note: there is no CompressionScheme::default but there is a
  // CompressedBufferEncoder::default (see earlier comment)
  let scheme = CompressionScheme::default();
  ...
}
```

rust/lance-encoding/src/compression.rs
lyang24: ty fixed

rust/lance-encoding/src/encodings/physical/general.rs
lyang24: yep give my attempt impl both traits in `CompressedBufferEncoder` 
in compression variable length it bind the valueEncoder to encode
in decom it uses BinaryBlockDecompressor to decode

rust/lance-encoding/src/encodings/physical/general.rs
lyang24: its a mistake i wrote test with cursor i think it responded to my adjustment prompt :( i wrote a new e2e test in rust/lance-encoding/src/previous/encodings/logical/primitive.rs

PR #4903, opened 2025-10-06
feat: handle forking better

cmccabe:
When a process which is making use of lance calls fork(), threads in that process will not carry over into
the child process. This PR tries to handle the situation slightly better by re-creating thread executors
in the child process. Note that we still do not want to officially support fork. For one thing, tokio has
some issues with it. Therefore, we will still print a warning and recommend using a non-forking mode
for multiprocessing.

### DIFF
--- a/python/src/lib.rs
+++ b/python/src/lib.rs
@@ -134,6 +134,7 @@ pub fn rt() -> &'static mut BackgroundExecutor {
         if !EXECUTOR_INSTALLED.fetch_or(true, Ordering::SeqCst) {
             break;
         }
+        std::thread::yield_now();
     }
     if !ATFORK_INSTALLED.fetch_or(true, Ordering::SeqCst) {
         install_atfork();
@@ -147,6 +148,7 @@ pub fn rt() -> &'static mut BackgroundExecutor {
 /// runs in "async-signal context" which means that we can't (safely) do much here.
 extern "C" fn atfork_child() {
     BACKGROUND_EXECUTOR.store(std::ptr::null_mut(), Ordering::SeqCst);
+    EXECUTOR_INSTALLED.store(false, Ordering::SeqCst);
 }
 
 #[cfg(not(windows))]
--- a/rust/lance-core/src/utils/tokio.rs
+++ b/rust/lance-core/src/utils/tokio.rs
@@ -52,7 +52,6 @@ pub static IO_CORE_RESERVATION: LazyLock<usize> = LazyLock::new(|| {
 });
 
 fn create_runtime() -> Runtime {
-    // TODO: make this runtime configurable (e.g. num threads)
     Builder::new_multi_thread()
         .thread_name("lance-cpu")
         .max_blocking_threads(get_num_compute_intensive_cpus())
@@ -78,6 +77,7 @@ fn global_cpu_runtime() -> &'static mut Runtime {
         if !RUNTIME_INSTALLED.fetch_or(true, Ordering::SeqCst) {
             break;
         }
+        std::thread::yield_now();
     }
     if !ATFORK_INSTALLED.fetch_or(true, Ordering::SeqCst) {
         install_atfork();
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,13 @@
 # Prerequisites
 *.d
 
+# vim swap files
+*.swp
+
+# cargo config files
+.cargo/config
+python/.cargo/config
+
 # Compiled Object files
 *.slo
 *.lo


docs/src/integrations/pytorch.md
```diff
@@ -23,12 +23,12 @@ Then, you can use the Lance dataset in PyTorch training and inference loops.
 
 Note:
 
-1. the PyTorch dataset automatically convert data into `torch.Tensor`
+1. The PyTorch dataset automatically convert data into `torch.Tensor`.
 
-2. lance is not fork-safe. If you are using multiprocessing, use spawn instead.
-The safe dataloader uses the spawning method.
+2. If you are using multiprocessing, it is preferrable to use spawn rather than fork.
```
westonpace:
```suggestion
2. If you are using multiprocessing, it is preferrable to use forkserver or spawn rather than fork.
```

spawn is noticeably slower

python/Cargo.toml
```diff
@@ -47,6 +47,7 @@ lance-io = { path = "../rust/lance-io" }
 lance-linalg = { path = "../rust/lance-linalg" }
 lance-table = { path = "../rust/lance-table" }
 lance-datafusion = { path = "../rust/lance-datafusion" }
+libc = "0.2.176"
```
westonpace, line 50: Is this going to be a problem on Windows?

python/python/tests/test_fork.py
```diff
@@ -0,0 +1,38 @@
...
+        }
+    )
+
+def test_table_roundtrip(tmp_path: Path):
+    uri = tmp_path
+
+    tbl = create_table(100)
+    lance.write_dataset(tbl, uri)
+
+    os.fork()
```
westonpace, line 34: How does fork work with pytest?

python/src/lib.rs
```diff
@@ -113,7 +115,54 @@ fn register_datagen(py: Python, m: &Bound<'_, PyModule>) -> PyResult<()> {
...
+extern "C" fn atfork_child() {
+    RT.store(std::ptr::null_mut(), Ordering::SeqCst);
+}
+
+fn install_atfork() {
+    let result = unsafe {
+        libc::pthread_atfork(None, None, Some(atfork_child))
+    };
+    assert_eq!(result, 0, "pthread_atfork failed");
+}
```
westonpace: Might need to wrap this for Windows?

westonpace (commented):
There is one more runtime here (the CPU runtime): https://github.com/lancedb/lance/blob/ec2ac35b9a0bd644e222b09a2f19fe899e9b1ed8/rust/lance-core/src/utils/tokio.rs#L53

It's currently only used for encoding (not decoding) and vector search.  As a result, you might not encounter it in your test (since you don't do these things after the fork)

Xuanwo:
Thank you for this contribution but I think this fix can be wrong, more details could be found here https://github.com/tokio-rs/tokio/issues/4301

I suggest we just make it clear that we don't support fork and forbid users to use fork.

python/Cargo.toml
cmccabe: Based on what I've found online, it is possible to link against the `libc` crate on Windows.

python/src/lib.rs
cmccabe: Yeah, good point. I will stub this out on Windows.

cmccabe:
> Thank you for this contribution but I think this fix can be wrong, more details could be found here https://github.com/tokio-rs/tokio/issues/4301 I suggest we just make it clear that we don't support fork and forbid users to use fork.

Hmm, this is a tough one. I do think we should discourage people from using fork but I wonder what the lesser evil is here.

westonpace:
> Thank you for this contribution but I think this fix can be wrong, more details could be found here https://github.com/tokio-rs/tokio/issues/4301
> 
> I suggest we just make it clear that we don't support fork and forbid users to use fork.

My understanding is that "tokio doesn't support fork" means that "tokio's runtime will not be usable after a fork".  As long as we replace the runtimes before they are ever used then I think it should be safe enough.

westonpace:
Ah, nevermind, I should have read more closely

> The process / signal modules define some global things that break on fork. For example, I think we sometimes have a global file descriptor for sending some notifications, and it doesn't survive the fork. There's also the signal handler, which may or may not behave well across a fork.

python/python/tests/test_fork.py
cmccabe: pytest doesn't seem to mind forking. Since multiprocessing is supported in pytest I expect there are others using fork here.

westonpace:
Hmm, it does seem that signal handling is limited to libraries that interact with `tokio::signal` (I'm fairly sure we don't).  I think there's enough unknowns that if testing seems to work we could potentially expose with a "use at your own risk" log warning.

python/src/lib.rs
```diff
@@ -112,8 +113,49 @@ fn register_datagen(py: Python, m: &Bound<'_, PyModule>) -> PyResult<()> {
...
+
+pub fn rt() -> &'static mut BackgroundExecutor {
+    loop {
+        let ptr = BACKGROUND_EXECUTOR.load(Ordering::SeqCst);
+        if !ptr.is_null() {
+            return unsafe { &mut *ptr };
+        }
+        if !EXECUTOR_INSTALLED.fetch_or(true, Ordering::SeqCst) {
+            break;
+        }
```
westonpace, line 136: Is this to ensure only one thread actually creates the executor?  Any other simultaneous threads will just busy loop?

If so, it might be polite to call `std::thread::yield_now` before the break.  That being said, I don't expect it to matter all that much.

python/src/lib.rs
```diff
@@ -112,8 +113,49 @@ fn register_datagen(py: Python, m: &Bound<'_, PyModule>) -> PyResult<()> {
...
+    }
+    if !ATFORK_INSTALLED.fetch_or(true, Ordering::SeqCst) {
```
westonpace, line 139: Is this variable needed because a forked process will not have to reinstall the fork handler (because it inherits it from the parent)?

rust/lance-core/src/utils/tokio.rs
```diff
@@ -50,7 +51,8 @@ pub static IO_CORE_RESERVATION: LazyLock<usize> = LazyLock::new(|| {
         .unwrap()
 });
 
-pub static CPU_RUNTIME: LazyLock<Runtime> = LazyLock::new(|| {
+fn create_runtime() -> Runtime {
+    // TODO: make this runtime configurable (e.g. num threads)
```
westonpace: Isn't it already configurable via `LANCE_CPU_THREADS` variable?

rust/lance-core/src/utils/tokio.rs
```diff
@@ -59,7 +61,46 @@ pub static CPU_RUNTIME: LazyLock<Runtime> = LazyLock::new(|| {
...
+static ATFORK_INSTALLED: atomic::AtomicBool = atomic::AtomicBool::new(false);
+
+fn global_cpu_runtime() -> &'static mut Runtime {
+    loop {
+        let ptr = CPU_RUNTIME.load(Ordering::SeqCst);
+        if !ptr.is_null() {
+            return unsafe { &mut *ptr };
+        }
+        if !RUNTIME_INSTALLED.fetch_or(true, Ordering::SeqCst) {
+            break;
```
westonpace, line 78: Same note as above re: yield

python/src/lib.rs
```diff
@@ -112,8 +113,49 @@ fn register_datagen(py: Python, m: &Bound<'_, PyModule>) -> PyResult<()> {
...
+    }
+    let new_ptr = Box::into_raw(Box::new(create_background_executor()));
+    BACKGROUND_EXECUTOR.store(new_ptr, Ordering::SeqCst);
+    unsafe { &mut *new_ptr }
+}
+
+/// After a fork() operation, force re-creation of the BackgroundExecutor. Note: this function
+/// runs in "async-signal context" which means that we can't (safely) do much here.
+extern "C" fn atfork_child() {
+    BACKGROUND_EXECUTOR.store(std::ptr::null_mut(), Ordering::SeqCst);
```
westonpace, line 150: Do you need to also set `EXECUTOR_INSTALLED` to false here?

python/src/lib.rs
cmccabe: Right.

python/src/lib.rs
cmccabe: good catch.

PR #4905, opened 2025-10-07
feat: support dynamic storage options provider with AWS credentials vending

jackye1995:
This PR introduces a new dynamic storage options provider interface in Lance dataset. The main idea is that the provider describes the storage options to use for a given dataset, and when these options will expire. Lance is responsible for fetching another new set of storage options to re-initialize the object store when expiration happens.

This is mainly useful for cases where the dataset's access credentials are temporary, and the user would like to invoke a specific credentials endpoint to fetch a new set of credentials. Currently we have only added support for AWS by implementing a credentials provider.

The PR also provides an implementation of the `StorageOptionsProvider` with Lance Namespace, because Lance Namespace provides a DescribeTable endpoint which returns the storage options that should be used at the given time. Based on the namespace spec, a `expires_at_millis` key can be added to the storage options to indicate the expiration time of those options.

Because Lance Namespace provides native implementations in python and Java, we also provides binding interfaces of of `PyStorageOptionsProvider` and `JavaStorageOptionsProvider`, which then further integrated with Lance Namespace implemented in those specific languages.

### DIFF
--- a/java/src/main/java/com/lancedb/lance/Dataset.java
+++ b/java/src/main/java/com/lancedb/lance/Dataset.java
@@ -260,7 +260,6 @@ public class Dataset implements Closeable {
    * Dataset dataset = Dataset.open()
    *     .namespace(myNamespace)
    *     .tableId(Arrays.asList("my_table"))
-   *     .refreshStorageOptions(true)
    *     .build();
    * }</pre>
    *
--- a/java/src/main/java/com/lancedb/lance/OpenDatasetBuilder.java
+++ b/java/src/main/java/com/lancedb/lance/OpenDatasetBuilder.java
@@ -48,7 +48,6 @@ import java.util.Map;
  * Dataset dataset = Dataset.open()
  *     .namespace(myNamespace)
  *     .tableId(Arrays.asList("my_table"))
- *     .refreshStorageOptions(true)
  *     .build();
  * }</pre>
  */
@@ -59,7 +58,7 @@ public class OpenDatasetBuilder {
   private LanceNamespace namespace;
   private List<String> tableId;
   private ReadOptions options = new ReadOptions.Builder().build();
-  private boolean refreshStorageOptions = false;
+  private boolean ignoreNamespaceTableStorageOptions = false;
 
   /** Creates a new builder instance. Package-private, use Dataset.open() instead. */
   OpenDatasetBuilder() {}
@@ -130,18 +129,15 @@ public class OpenDatasetBuilder {
   }
 
   /**
-   * Sets whether storage options should be automatically refreshed before they expire.
+   * Sets whether to ignore storage options from the namespace's describe_table().
    *
-   * <p>This is only applicable when using namespace-based opening. It is currently only used for
-   * refreshing AWS temporary access credentials. When enabled, the namespace will be queried
-   * periodically to fetch new temporary credentials before the current ones expire. The new storage
-   * options will contain updated AWS access credentials with a new expiration time.
-   *
-   * @param refreshStorageOptions If true, storage options will be automatically refreshed
+   * @param ignoreNamespaceTableStorageOptions If true, storage options returned from
+   *     describe_table() will be ignored (treated as null)
    * @return this builder instance
    */
-  public OpenDatasetBuilder refreshStorageOptions(boolean refreshStorageOptions) {
-    this.refreshStorageOptions = refreshStorageOptions;
+  public OpenDatasetBuilder ignoreNamespaceTableStorageOptions(
+      boolean ignoreNamespaceTableStorageOptions) {
+    this.ignoreNamespaceTableStorageOptions = ignoreNamespaceTableStorageOptions;
     return this;
   }
 
@@ -203,26 +199,25 @@ public class OpenDatasetBuilder {
 
     DescribeTableResponse response = namespace.describeTable(request);
 
-    // Extract location
     String location = response.getLocation();
     if (location == null || location.isEmpty()) {
       throw new IllegalArgumentException("Namespace did not return a table location");
     }
 
-    // Build new ReadOptions with initial storage options
+    Map<String, String> namespaceStorageOptions =
+        ignoreNamespaceTableStorageOptions ? null : response.getStorageOptions();
+
     ReadOptions.Builder optionsBuilder =
         new ReadOptions.Builder()
             .setIndexCacheSizeBytes(options.getIndexCacheSizeBytes())
             .setMetadataCacheSizeBytes(options.getMetadataCacheSizeBytes());
 
-    // Only set storage options provider if refresh is enabled
-    if (refreshStorageOptions) {
+    if (namespaceStorageOptions != null && !namespaceStorageOptions.isEmpty()) {
       LanceNamespaceStorageOptionsProvider storageOptionsProvider =
           new LanceNamespaceStorageOptionsProvider(namespace, tableId);
       optionsBuilder.setStorageOptionsProvider(storageOptionsProvider);
     }
 
-    // Set optional fields only if present
     options.getVersion().ifPresent(optionsBuilder::setVersion);
     options.getBlockSize().ifPresent(optionsBuilder::setBlockSize);
     options.getSerializedManifest().ifPresent(optionsBuilder::setSerializedManifest);
@@ -230,10 +225,9 @@ public class OpenDatasetBuilder {
         .getS3CredentialsRefreshOffsetSeconds()
         .ifPresent(optionsBuilder::setS3CredentialsRefreshOffsetSeconds);
 
-    // Add initial storage options from describe_table response if present
     Map<String, String> storageOptions = new HashMap<>(options.getStorageOptions());
-    if (response.getStorageOptions() != null) {
-      storageOptions.putAll(response.getStorageOptions());
+    if (namespaceStorageOptions != null) {
+      storageOptions.putAll(namespaceStorageOptions);
     }
     optionsBuilder.setStorageOptions(storageOptions);
 
--- a/java/src/test/java/com/lancedb/lance/NamespaceIntegrationTest.java
+++ b/java/src/test/java/com/lancedb/lance/NamespaceIntegrationTest.java
@@ -252,7 +252,6 @@ public class NamespaceIntegrationTest {
               .allocator(allocator)
               .namespace(namespace)
               .tableId(Arrays.asList(tableName))
-              .refreshStorageOptions(true) // Refresh enabled
               .readOptions(readOptions)
               .build()) {
         // With the fix, describeTable should only be called once during open
@@ -357,7 +356,6 @@ public class NamespaceIntegrationTest {
               .allocator(allocator)
               .namespace(namespace)
               .tableId(Arrays.asList(tableName))
-              .refreshStorageOptions(true) // Enable automatic refresh
               .readOptions(readOptions)
               .build()) {
         // With the fix, describeTable should only be called once during open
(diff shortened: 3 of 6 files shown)


rust/lance-io/src/object_store/credential_vending.rs
```diff
@@ -0,0 +1,878 @@
...
+/// Implementations can fetch credentials from various sources (namespace servers,
+/// credential vending machines, etc.) and are usable from Python/Java via FFI.
+#[async_trait]
+pub trait CredentialVendor: Send + Sync {
+    /// Fetch fresh credentials
+    ///
+    /// Returns a tuple of (storage_options, expires_at_millis)
+    /// where storage_options is a map of credential key-value pairs
+    /// and expires_at_millis is the epoch time in milliseconds when credentials expire
+    async fn get_credentials(&self) -> Result<(HashMap<String, String>, u64)>;
```
wjones127: Could we use a proper time type here, like `std::time::Instant`?
```suggestion
    /// and expires_at_millis is the epoch time in milliseconds when credentials expire
    async fn get_credentials(&self) -> Result<(HashMap<String, String>, Instant)>;
```

rust/lance-io/src/object_store/storage_options.rs
```diff
@@ -0,0 +1,878 @@
...
+        Ok((storage_options, expires_at_millis))
+    }
+}
+
+/// Configuration parameters for credential vending
+#[derive(Debug, Clone)]
+pub struct CredentialVendingParams {
+    /// How early to refresh credentials before expiration (in milliseconds)
+    /// Default: 300,000 (5 minutes)
+    pub refresh_lead_time_ms: u64,
```
wjones127: Similarly, can we use a proper type here like `std::time::Duration`?
```suggestion
    /// How early to refresh credentials before expiration (in milliseconds)
    /// Default: 300,000 (5 minutes)
    pub refresh_lead_time_ms: Duration,
```

rust/lance-io/src/object_store/storage_options.rs
```diff
@@ -0,0 +1,878 @@
...
+            inner: original,
+        })
+    }
+}
+
+/// Delegating ObjectStore that auto-refreshes credentials
+pub struct DelegatingObjectStore {
+    wrapper: Arc<CredentialVendingObjectStoreWrapper>,
+    inner: Arc<dyn OSObjectStore>,
+}
```
wjones127: How are the credentials passed to the object store?

rust/lance-io/src/object_store/storage_options.rs
wjones127, line 1: I've been looking through this code, and I haven't been able to figure out how the credentials are passed to the actual object store instance? I see them being refreshed to the side, but I don't see how the object store is supposed to pick up the new credentials.

What I would expect is somewhere you build a [CredentialProvider](https://docs.rs/object_store/latest/object_store/client/trait.CredentialProvider.html) and then you pass it into [ObjectStore::with_credentials()](https://docs.rs/object_store/latest/object_store/aws/type.AwsCredentialProvider.html) when constructing the object store. For example, that's how we have credential refresh for AWS working right now:

https://github.com/lancedb/lance/blob/d342d5db2da90edcbcf4bd88aa38985bcb111aa8/rust/lance-io/src/object_store/providers/aws.rs#L58-L64
https://github.com/lancedb/lance/blob/d342d5db2da90edcbcf4bd88aa38985bcb111aa8/rust/lance-io/src/object_store/providers/aws.rs#L80-L82

wjones127 (commented):
Got stuck trying to understand the basics of how this works.

jackye1995:
Actually it might be better to call this feature more generically, like `StorageOptionProvider`. Let me do another round of refactoring.

jackye1995:
> The new dynamic credential provider is carried on ObjectStoreParams.storage_options_provider, but the Hash/PartialEq implementations used by the object‑store registry still ignore this field. 

Good point. Added the rust one locally, pending merging python and java one.

Xuanwo (approved):

rust/lance-io/src/object_store/storage_options.rs
```diff
@@ -0,0 +1,119 @@
...
+/// Implementations can fetch storage options from various sources (namespace servers,
+/// secret managers, etc.) and are usable from Python/Java.
+///
+/// # Equality and Hashing
+///
+/// Implementations must provide `provider_id()` which returns a unique identifier for
+/// equality and hashing purposes. Two providers with the same ID are considered equal
+/// and will share the same cached ObjectStore in the registry.
+#[async_trait]
+pub trait StorageOptionsProvider: Send + Sync + fmt::Debug {
```
wjones127, line 56: Is there a reason to make this generic storage options? If not, I think we should narrow it to just be credentials.

rust/lance-io/src/object_store/providers/aws.rs
```diff
@@ -269,6 +283,58 @@ pub async fn build_aws_credential(
...
+async fn build_aws_credential_with_storage_options_provider(
+    storage_options_provider: Arc<dyn StorageOptionsProvider>,
+    credentials_refresh_offset: Duration,
+    credentials: Option<AwsCredentialProvider>,
+    storage_options_credentials: Option<StaticCredentialProvider<ObjectStoreAwsCredential>>,
+    expires_at_millis: Option<u64>,
+) -> Result<AwsCredentialProvider> {
+    match (expires_at_millis, credentials, storage_options_credentials) {
+        // Case 1: provider + credentials + expiration time
+        (Some(expires_at), Some(cred), _) => {
```
wjones127: Do we need to handle expiration?

rust/lance-io/src/object_store/providers/aws.rs
```diff
@@ -269,6 +283,58 @@ pub async fn build_aws_credential(
...
+            )),
+            location: location!(),
+        }),
+        // Case 4: provider + credentials without expiration - FAIL
+        (None, Some(_), _) => Err(Error::IO {
+            source: Box::new(std::io::Error::other(
+                "expires_at_millis is required when using storage_options_provider with credentials",
+            )),
+            location: location!(),
+        }),
```
wjones127, line 338: I think we can trust if they pass a credential provider they already handled expiration, right?

rust/lance-io/src/object_store/providers/aws.rs
```diff
@@ -420,13 +486,180 @@ impl ObjectStoreParams {
...
+/// AWS Credential Provider that uses StorageOptionsProvider
+///
+/// This adapter converts our generic StorageOptionsProvider trait into
+/// AWS-specific credentials that can be used with S3. It caches credentials
+/// and automatically refreshes them before they expire.
+pub struct DynamicStorageOptionsCredentialProvider {
+    provider: Arc<dyn StorageOptionsProvider>,
+    cache: Arc<RwLock<Option<CachedCredential>>>,
+    refresh_offset: Duration,
+}
```
wjones127, line 515: If I wanted to be fancy, I would do something like:

```rust
pub struct DynamicCredentials(pub Arc<HashMap<String, String>>);

pub struct NamespaceCredentialsProvider {
    provider: Arc<dyn StorageOptionsProvider>,
    cache: Arc<RwLock<Option<CachedCredential>>>,
    refresh_offset: Duration,
}

impl CredentialProvider for NamespaceCredentialsProvider<AwsCredentials> {
    type Credential = AwsCredentials;

    async fn get_credential(&self) -> ObjectStoreResult<Arc<Self::Credential>> {
        self.fetch_credential::<Self::Credential>().await
    }
}

impl TryFrom<DynamicCredentials> for AwsCredentials {}
impl TryFrom<DynamicCredentials> for AzureCredentials {}
impl TryFrom<DynamicCredentials> for GcpCredentials {}

impl<T: TryFrom<DynamicCredentials>> NamespaceCredentialsProvider {
    async fn fetch_credential<T>(&self) -> Result<T> {
        let credential = todo!();
         credential.try_into()
    }
}
```

Then you don't need to repeat this adapter for all three clouds. The only thing that's cloud specific is how to map from `DynamicCredentials` to the particular credentials type.

rust/lance-io/src/object_store/providers/aws.rs
```diff
@@ -225,17 +225,20 @@ async fn resolve_s3_region(
...
+/// 4. The default credential provider chain from AWS SDK.
 ///
 /// `credentials_refresh_offset` is the amount of time before expiry to refresh credentials.
 pub async fn build_aws_credential(
     credentials_refresh_offset: Duration,
     credentials: Option<AwsCredentialProvider>,
     storage_options: Option<&HashMap<AmazonS3ConfigKey, String>>,
     region: Option<String>,
+    storage_options_provider: Option<Arc<dyn StorageOptionsProvider>>,
+    expires_at_millis: Option<u64>,
```
wjones127, line 250: 1. It seems like `storage_options_provider` and `credentials` should be mutually exclusive? Either I want the credentials to be controlled by `storage_options_provider` or I want to pass my own customer `AwsCredentialProvider`.
2. I'm not sure why we need `expires_at_millis`. My understanding is that the final `AwsCredentialProvider::get_credentials()` is called on every request, so the expiration could be handled internally, right?

rust/lance-io/src/object_store/storage_options.rs
jackye1995: In near term, might not be. 

If I think from what was possible in similar features in Iceberg, there are things like 
- bucket alias, that user can say "for this bucket name, swap to use another one temporarily"
- s3 tags, user can configure some dynamic tags to be assigned to the files written by specific sessions

That could be enabled.

Basically everything that is controllable in the Iceberg FileIO can be controlled dynamically in such way, so I was thinking about making equivalent changes in Lance for ObjectStore.

But I don't have a strong opinion, it's probably fine to also just do the features case by case.

rust/lance-io/src/object_store/providers/aws.rs
jackye1995: yeah I actually originally implemented it that way. But I tried on GCP and it did not really work out, because GCP does not really provide remote temporary credentials and the generation of temporary token and signing is done all locally. I need to look more into it so I ended up just do it for AWS for now. Let me know if you prefer to first set it up this way.

rust/lance-io/src/object_store/providers/aws.rs
jackye1995: This is actually intentionally not mutually exclusive. What I am trying to achieve here is that this is the initial value. Because initially when we try to load a dataset, we already want to call `namespace.describeTable` once, which gives the table location + storage options that contains the initial credentials and expiration time. I am basically leveraging the existing fields to pass those in, so that I don't need to do another duplicated `namespace.describeTable` call in order to just fetch another new set of credential and expiration time. Once the initial usage is done, these fields are not respected.

wjones127: Ah okay. Could you explain that in a comment?

rust/lance-io/src/object_store/providers/aws.rs
wjones127: Ah okay. That's fine if it's hard to do. Can also be a future refactor.

ObjectStore has this `TokenProvider` internal trait they use. Not sure if you saw that already:

https://github.com/apache/arrow-rs-object-store/blob/b72c00fb6d9ec30355fea7085556e4df41646ff5/src/client/mod.rs#L847-L895

rust/lance-io/src/object_store/storage_options.rs
wjones127: Okay I'm more open to this. The tags idea seems cool. I also could see a future where you might want the catalog to tell you which KMS keys to use for an S3 bucket. 👍

beinan (approved):
We're looking forward to this feature, thanks!

rust/lance-io/src/object_store/providers/aws.rs
jackye1995: Yes, this is more for completeness

PR #4906, opened 2025-10-07
fix: forward compatibility of FTS index

jackye1995:
When creating using 0.38 and reading using 0.36 for `list_indices` the operation fails with 

```
thread '<unnamed>' panicked at /home/runner/work/lance/lance/rust/lance-index/src/scalar/inverted/index.rs:846:30:
called `Option::unwrap()` on a `None` value
```

### DIFF
--- a/python/python/tests/forward_compat/test_compat.py
+++ b/python/python/tests/forward_compat/test_compat.py
@@ -109,6 +109,8 @@ def test_pq_buffer():
     reason="FTS token set format was introduced in 0.36.0",
 )
 def test_list_indices_ignores_new_fts_index_version():
+    # Dataset::load_manifest does not do retain_supported_indices
+    # so this can only work with no cache
     session = lance.Session(index_cache_size_bytes=0, metadata_cache_size_bytes=0)
     ds = lance.dataset(get_path("fts_index"), session=session)
     indices = ds.list_indices()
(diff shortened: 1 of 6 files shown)


python/python/tests/forward_compat/test_compat.py
```diff
@@ -101,3 +101,16 @@ def test_pq_buffer():
...
     )
+
+
+@pytest.mark.forward
+@pytest.mark.skipif(
+    Version(lance.__version__) < Version("0.36.0"),
+    reason="FTS token set format was introduced in 0.36.0",
+)
+def test_list_indices_ignores_new_fts_index_version():
+    session = lance.Session(index_cache_size_bytes=0, metadata_cache_size_bytes=0)
```
cmccabe, line 114: wait, are we supposed to set index_cache_size_bytes=0 and metadata_cache_size_bytes=0?

rust/lance-index/src/scalar/inverted/index.rs
```diff
@@ -81,7 +81,9 @@ use crate::Index;
 use crate::{prefilter::PreFilter, scalar::inverted::iter::take_fst_keys};
 use std::str::FromStr;
 
-pub const INVERTED_INDEX_VERSION: u32 = 0;
+// Version 0: Arrow TokenSetFormat (legacy)
+// Version 1: Fst TokenSetFormat (new default, incompatible with old clients)
```
cmccabe: mention lance version where version 1 was added?

PR #4914, opened 2025-10-09
chore: keep the large vectors

BubbleCal:
related #4842 
In https://github.com/lancedb/lance/pull/4890, we filter out the large vectors, this PR reverts that and try best to retrieve the large vectors:
- if 2 vectors are both large but close to each other, the distance can be finite, so we can retrieve them
- if 2 vectors are both large and the distance is infinite, we just return them in random order


Baibaiwuguo:
@BubbleCal  
hello. To avoid the situation where the denominator of the cosine is 0, could we consider adding a small positive number through dynamic setting？

PR #4915, opened 2025-10-09
chore: clean up the temporary directory after the example is completed

zhangyue19921010:
chore for `rust/examples/src/write_read_ds.rs`
Clean up the temporary directory after the example is completed

Before
截屏2025-10-09 12 43 06

After
截屏2025-10-09 12 43 45


zhangyue19921010:
> This will only clean resources if the program is successful. If it's not it will still leave them. Maybe we should write to an actual temporary directory instead?

Hi @wjones127, thanks a lot for your input. Modify the write directory to the system temporary directory. 

Also, maybe Lance should be responsible for cleaning up tmp data, no matter where it is written. :) 

So that I modified the codo that Lance will automatically clean up the history files regardless of success or error scenarios. PTAL~

zhangyue19921010:
Thanks @wjones127 for your help :)

PR #4921, opened 2025-10-09
test: create framework for testing memory

### LINKED ISSUE #4761: Add utility to test memory usage in integration tests
Acceptance criteria:

* Can make assertions about: `num_allocations`, `total_bytes_allocated`, `num_deallocations`, `peak_bytes_allocated`
    * No leaks: `assert_eq!(num_allocations, num_deallocations)`
* Can have this scoped per operation / test
* Bonus: still supports running tests in parallel

This is going to be critical for solving: https://github.com/lancedb/lance/issues/4477

We can probably use the approach used in [hotpath](https://github.com/pawurb/hotpath) and [allocation-counter](https://github.com/fornwall/allocation-counter), where a custom allocator is specified in the binary (could be the integration test binary) and trace events are emitted on allocations. Note: it seems like `hotpath` only works in single threaded mode. I think if we pass down some sort of trace id in the context, then we should be able to track allocations per test independently, even if there are multiple tests running in parallel.

wjones127:
Adds a new integration test `resource_test` which contains a utility for making assertions about memory use with a span. This needs to be in a separate integration tests because it hooks into the global memory allocator, and doing that for all tests would probably make them quite slow.

Here's an example of how this can be used:

```rust
let alloc_tracker = AllocTracker::new();
{
    let _guard = alloc_tracker.enter();
    // Do something that allocates, even in async tasks
}
let stats = alloc_tracker.stats();
assert_lt!(stats.max_bytes_allocated, 100_000_000, "Used more than 100MB");
assert_eq!(stats.total_bytes_allocated, stats.total_bytes_deallocated, "No memory leaked");
```

Closes #4761

### DIFF
--- a/rust/lance/tests/resource_tests.rs
+++ b/rust/lance/tests/resource_tests.rs
@@ -3,5 +3,5 @@
 
 // The memory tests don't work currently on MacOS because they rely on thread
 // local storage in the allocator, which seems to have some issues on MacOS.
-#[cfg(not(target_os = "macos"))]
+#[cfg(target_os = "linux")]
 mod resource_test;


rust/lance-index/src/scalar/bitmap.rs
```diff
@@ -104,15 +114,21 @@ pub struct BitmapIndex {
 
     value_type: DataType,
 
-    store: Arc<dyn IndexStore>,
-
+    // store: Arc<dyn IndexStore>,
```
cmccabe: why is this not needed any more?

rust/lance-index/src/scalar/bitmap.rs
```diff
@@ -297,7 +313,9 @@ impl DeepSizeOf for BitmapIndex {
         let mut total_size = 0;
 
         total_size += self.index_map.deep_size_of_children(context);
-        total_size += self.store.deep_size_of_children(context);
+        total_size += self.null_map.deep_size_of_children(context);
+        total_size += self.lazy_reader.deep_size_of_children(context);
+        // total_size += self.store.deep_size_of_children(context);
```
cmccabe: should we remove self.store if it's not needed?

wjones127: yeah, need to figure that out. Remember this PR is a draft.

rust/lance/tests/resource_test/write.rs
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    let stats = alloc_tracker.stats();
+    // Allow for 2x the batch size to account for overheads.
+    // The key test is that we don't load all 100MB into memory at once
+    assert_le!(
+        stats.max_bytes_allocated,
+        (batch_size * 2) as isize,
+        "Max memory usage exceeded"
+    );
```
wkalt, line 44: I wonder if it would make sense to wrap this up in a macro. Would that promote more widespread usage?
```
#[maxallocs=10*1024*1024]
{
        let tmp_dir = tempfile::tempdir().unwrap();
        let tmp_path = tmp_dir.path().to_str().unwrap();
        let _dataset = InsertBuilder::new(tmp_path)
            .execute_stream(data)
            .await
            .unwrap();
}
```

wkalt (approved):

rust/lance/tests/resource_test/write.rs
wjones127: We can add that later if we find it simplifies things. Will be easier to see once we have more tests in here and see what patterns are repeating.

wjones127:
Had to disable this for MacOS. We can implement a version that is cross platform in a future PR.

PR #4923, opened 2025-10-09
test: standardize IO tests and improve assertions

wjones127:
* `IOTracker` is now public in `lance-io`, so it can be re-used throughout the codebase.
* Added new assertions `assert_io_eq!()`, `assert_io_lt!()` and `assert_io_gt!()` which will print out the list of requests in case of failure:

```rust
thread 'dataset::tests::test_load_manifest_iops' panicked at rust/lance/src/dataset.rs:2882:9:
assertion failed: `(left == right)`: Expected read_iops to be 3, got 2. Requests: [
    IORequest(method=list, path="test/_versions"),
    IORequest(method=get_opts, path="test/_versions/1.manifest"),
]

Diff < left / right > :
<2
>3
```

### DIFF
--- a/rust/lance-io/src/utils/tracking_store.rs
+++ b/rust/lance-io/src/utils/tracking_store.rs
@@ -134,11 +134,27 @@ macro_rules! assert_io_lt {
 // These fields are "dead code" because we just use them right now to display
 // in test failure messages through Debug. (The lint ignores Debug impls.)
 #[allow(dead_code)]
-#[derive(Debug, Clone)]
+#[derive(Clone)]
 pub struct IoRequestRecord {
     pub method: &'static str,
     pub path: Path,
     pub range: Option<Range<u64>>,
+}
+
+impl std::fmt::Debug for IoRequestRecord {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        // For example: "put /path/to/file range: 0-100"
+        write!(
+            f,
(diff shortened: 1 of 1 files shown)


rust/lance-io/src/utils/tracking_store.rs
```diff
@@ -0,0 +1,459 @@
...
+    async fn list_with_delimiter(&self, prefix: Option<&Path>) -> OSResult<ListResult> {
+        let _guard = self.hop_guard();
+        self.record_read(
+            "list_with_delimiter",
+            prefix.cloned().unwrap_or_default(),
+            0,
+            None,
+        );
+        self.target.list_with_delimiter(prefix).await
+    }
```
westonpace, line 372: Doesn't have to be this PR but I wonder if we should start recording lists as a separate operation entirely (e.g. not reads)

wjones127:
> I think the only thing I still find annoying is the need to setup the `object_store_wrapper`.

It's not just that. You also *have to* to use the `memory://` or `file-object-store://` URI, since we bypass object store APIs often when you just use file paths. So it's a bit annoying.

> Thoughts for future PR: I think (could be wrong) that we wrap all object stores today with the tracing object store anyways. I wonder if it would be possible to just put IOP tracking inside the tracing object store.

Hmm, now that's an interesting idea. Yeah we could have some sort of test feature that enables collection of IO requests. And then it wouldn't require much setup. I'll think more about that.

Issue #4927, opened 2025-10-10
Using FTS as a filter in vector search


wojiaodoubao:
A common use case is: user stores tags in a text column. Among all records that contain the specified tag(e.g. city=shanghai), he/she needs to search for the records nearest to the specified vector(e.g. embedding of traffic light image).

Currently, `Scanner` does not support setting both FTS query and vector query. I think we can support full text search as a filter in vector scan, for better performance and usability.

Since FTS is a filter of vector scan, the behavior is also affected by prefilter.
1. If prefilter is true, we perform vector search based on the results of FTS.
2. If prefilter is false, we perform vector search, then filter out the results that do not meet the FTS conditions.

In lancedb, we have supported hybrid search, which is a powerful feature that allows arbitrary sorting of query results through a Reranker. However, it requires executing two separate queries, even if we just want to perform a vector search based on the filtering results of FTS. I think in lance, we can handle this case better.

BubbleCal:
Cool idea!
I think today many users are using FTS as filters, this could be very useful.

Will take a look at the PR

ztorchan:
Hi, is there any progress on this issue?
I am very interested in this feature!

westonpace:
Can we expand on what it means for FTS to be "a filter"?  FTS returns a score for each row, not a boolean, so it doesn't quite act like a filter.  Is the thought that FTS as a filter is "all results with score/distance in some threshold"?

wojiaodoubao:
Hi @westonpace , when using FTS as a filter, we only focus on whether a document matches the FTS query; in other words, we filter for documents with a BM25 score greater than 0. 

This solves the problem of filtering specific tags when tags are stored in text or JSON format. We only care about the presence of the tags—we do not concern with the frequency of tag occurrences, the total number of tags, or similar metrics.

Issue #1782, opened 2024-01-04
Scanner simplification


wjones127:
The code in [dataset/scanner.rs](https://github.com/lancedb/lance/blob/74b743aebebd024f4bce76024b260b521fdc45ed/rust/lance/src/dataset/scanner.rs) has gotten extremely complicated, to a point where it is hard to test. Before we make any improvements, we need to refactor this to be easier to test and extend.

## Tasks

* [ ] Create a [LogicalPlan](https://docs.rs/datafusion-expr/latest/datafusion_expr/logical_plan/enum.LogicalPlan.html) for KNN search
* [ ] Write an extension for [LogicalPlanBuilder](https://docs.rs/datafusion-expr/latest/datafusion_expr/logical_plan/builder/struct.LogicalPlanBuilder.html) and rewrite `Scanner::create_plan()` in terms of that.
* [ ] Create a custom [PhysicalPlanner](https://docs.rs/datafusion/latest/datafusion/physical_planner/trait.PhysicalPlanner.html) to convert out logical plans into relevant nodes
* [ ] Implement a [PhysicalOptimizerRule](https://docs.rs/datafusion/latest/datafusion/physical_optimizer/optimizer/trait.PhysicalOptimizerRule.html) for each of the optimizations we make:
   * [ ] Using vector indices (ANN)
   * [ ] Using scalar indices (ANN and scan)

jayzhan211:
I'm interesting in improve datafusion extensibility for lance

wjones127:
I think a good approach to this would be starting to design some logical plan nodes implementing [UserDefinedLogicalNode](https://docs.rs/datafusion-expr/latest/datafusion_expr/logical_plan/trait.UserDefinedLogicalNode.html).

From there, we can create a `create_plan_v2()` method that creates a logical plan instead of the `ExecutionPlan`. Later then we can physical planner and other things. This would keep the existing planning intact until we are ready to switch things over.

> I'm interesting in improve datafusion extensibility for lance

Is there a particular goal you have in mind that you want to work towards?

jayzhan211:
> Is there a particular goal you have in mind that you want to work towards?

None yet. I would like to know how lance built on datafusion and what are the room for improvement

> I think a good approach to this would be starting to design some logical plan nodes implementing [UserDefinedLogicalNode](https://docs.rs/datafusion-expr/latest/datafusion_expr/logical_plan/trait.UserDefinedLogicalNode.html).

Probably I could start from this!

[...]

PR #4928, opened 2025-10-10
feat: support using FTS as a filter in vector search

wojiaodoubao:
Close https://github.com/lancedb/lance/issues/4927

### DIFF
--- a/java/lance-jni/src/merge_insert.rs
+++ b/java/lance-jni/src/merge_insert.rs
@@ -9,7 +9,7 @@ use arrow::ffi_stream::{ArrowArrayStreamReader, FFI_ArrowArrayStream};
 use jni::objects::{JObject, JString, JValueGen};
 use jni::sys::jlong;
 use jni::JNIEnv;
-use lance::dataset::scanner::LanceFilter;
+use lance::dataset::scanner::ExprFilter;
 use lance::dataset::{
     MergeInsertBuilder, MergeStats, WhenMatched, WhenNotMatched, WhenNotMatchedBySource,
 };
@@ -158,7 +158,7 @@ fn extract_when_not_matched_by_source_str<'local>(
 fn extract_when_not_matched_by_source_delete_expr<'local>(
     env: &mut JNIEnv<'local>,
     jparam: &JObject,
-) -> Result<Option<LanceFilter>> {
+) -> Result<Option<ExprFilter>> {
     let when_not_matched_by_source_delete_expr = env
         .call_method(
             jparam,
@@ -169,7 +169,7 @@ fn extract_when_not_matched_by_source_delete_expr<'local>(
         .l()?;
 
     if let Some(expr) = env.get_string_opt(&when_not_matched_by_source_delete_expr)? {
-        return Ok(Some(LanceFilter::Sql(expr)));
+        return Ok(Some(ExprFilter::Sql(expr)));
     }
 
     let when_not_matched_by_source_delete_substrait_expr = env
@@ -182,7 +182,7 @@ fn extract_when_not_matched_by_source_delete_expr<'local>(
         .l()?;
 
     match env.get_bytes_opt(&when_not_matched_by_source_delete_substrait_expr)? {
-        Some(expr) => Ok(Some(LanceFilter::Substrait(expr.to_vec()))),
+        Some(expr) => Ok(Some(ExprFilter::Substrait(expr.to_vec()))),
         None => Ok(None),
     }
 }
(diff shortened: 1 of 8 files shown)


wojiaodoubao:
Hi @BubbleCal , sorry to bother. Could you help review when you have time, thanks very much!

rust/lance/src/dataset.rs
```diff
@@ -8476,6 +8478,142 @@ mod tests {
...
+                IndexType::Inverted,
+                None,
+                &InvertedIndexParams::default().with_position(true),
+                true,
+            )
+            .await
+            .unwrap();
+
+        // Case 1: search with prefilter=true
+        let query_vector = Float32Array::from(vec![300f32, 300f32, 300f32, 300f32]);
```
BubbleCal: How would this new feature work if the query contains filter, FTS and vector query all?

rust/lance/src/dataset.rs
```diff
@@ -8476,6 +8478,142 @@ mod tests {
...
+            .unwrap()
+            .prefilter(false)
+            .full_text_search(FullTextSearchQuery::new("text".to_string()))
+            .unwrap()
+            .try_into_stream()
+            .await
+            .unwrap();
+
+        let results = stream.try_collect::<Vec<_>>().await.unwrap();
+        let batch = concat_batches(&results[0].schema(), &results).unwrap();
```
BubbleCal: I'd expect if the FTS is a post-filter, then we should get vector search results (mean there should be '_distance' column), does it work like this?

wojiaodoubao:
> This feature is cool, but I think we still need to figure out the API and semantic things:
> this PR assumes if both FTS and vector query are provided, then one of them would be as the filter of the other one (depends on prefilter param), I'd think it would be better to always apply filter by filter() method, we can change this method to take an enum param which can be one of string, FTSQuery, VectorQuery.

@BubbleCal  Thanks for your nice suggestion! The current implementation does have some semantic confusion. I previously thought "using FTS as a filter for vector search" and "using vector search as a filter for FTS" were equivalent. Specifically, vector search with prefilter=true and FTS as filter is equivalent to full text search with prefilter=false and vector search as filter. 

However, they are actually different because the returned schema is not the same. The results of an FTS query should include _score instead of _distance, while the results of a vector search should include _distance instead of _score.

I like your idea "apply filter by filter() method, we can change this method to take an enum param which can be one of string, FTSQuery, VectorQuery". This approach allows us to distinguish between filter and search. Additionally, I think the enum param can be slightly modified to support cases where both string & fts or string & vector are used as filters simultaneously.
```
enum FilterParam {
    Sql(String),
    Fts(FullTextSearchQuery),
    Vector(Query),
    SqlAndFts(String, FullTextSearchQuery),
    SqlAndVector(String, Query)
}
```

Let me fix it.

rust/lance/src/dataset.rs
wojiaodoubao: For all 3 cases in the unit tests, the result schema is (id, vector, text, _distance, _score). So yes it contains '_distance', but it shouldn't contain '_score' since fts is used as filter. This is related to the semantic and api issue above, let me fix them together.

rust/lance/src/dataset.rs
wojiaodoubao: If prefilter=true, scanner will first perform the filter, then perform fts based on filter result, then perform vector search based on fts result.

If prefilter=false, scanner will first perform vector search, then filter the result with fts, then filter the result with filter.

BubbleCal:
@wjones127 I'm under the impression that we plan to make some changes to the query builder like:
`ds.filter("x>0").filter("x<10").nearest(vec)` is equal to `ds.filter("x>0 and x<10").nearest(vec)`

Could you confirm? If so I think it's a good time to implement this, and then we can have the filter enum simpler:
```
enum Filter {
    Sql(String),
    Fts(FullTextSearchQuery),
    Vector(Query),
}
```
cc @wojiaodoubao

wojiaodoubao:
Hi @BubbleCal @wjones127 , please help review when you have time, thanks very much!

wjones127:
> @wjones127 I'm under the impression that we plan to make some changes to the query builder like: `ds.filter("x>0").filter("x<10").nearest(vec)` is equal to `ds.filter("x>0 and x<10").nearest(vec)`
> 
> Could you confirm?

Yeah I think we are getting to the place where we need to replace the Scanner implementation entirely. What we really need is a logical plan builder, so users can compose arbitrary plans together.

That's a larger project. I'm personally okay doing something basic here to support this, with the idea in mind that we will soon replace the scanner API entirely. But that's just my opinion, I don't know what others think.

wojiaodoubao:
> ... we are getting to the place where we need to replace the Scanner implementation entirely. What we really need is a logical plan builder, so users can compose arbitrary plans together.

Hi @wjones127 , great to see that lance is going to work on the logical plan! I remembered there was an issue about this: https://github.com/lancedb/lance/issues/1782. I'd be happy to help if there is anything I can do.

Using fts/vector as a filter is useful, so I'd like to have a basic impl if it is possible.

wojiaodoubao:
Hi @wjones127 @westonpace @BubbleCal , sorry to bother you. Could you help review this when you have time, thanks very much!

BubbleCal (approved):
Great work!
Let's create a follow-up issue for adding this to python API!

Issue #4929, opened 2025-10-10
bug: vector delta index is deleted after optimizing multiple indices with `num_indices_to_merge=0`


BubbleCal:
If a dataset has multiple indices, and one is scalar index or FTS index, which doesn't have delta index, and there's a vector index, we expect optimize_indices with `num_indices_to_merge=0` should create a delta index for the vector index, and merge the unindexed data into scalar index. But what we got is:
1. unindexed data is merged into scalar index
2. new delta vector index is created
3. but the old delta vector index is deleted

This is caused by a bug in transaction commit, what removes the indices with the same name in the new indices, intend to handle the case of creating new index with `replace=True`.

This could result in unexpected performance downgrade

reproduce:
```
import lance
import numpy as np
import pyarrow as pa

# Create random data
num_rows = 10000
ids = np.arange(num_rows, dtype=np.int32)
vectors = np.random.randn(num_rows, 16).astype(np.float32)

# Create PyArrow table
schema = pa.schema(
    [pa.field("id", pa.int32()), pa.field("vector", pa.list_(pa.float32(), 16))]
)

table = pa.Table.from_arrays(
    [ids, pa.FixedSizeListArray.from_arrays(vectors.ravel(), 16)], schema=schema
)

# Write to Lance dataset
uri = "data/random_dataset.lance"
dataset = lance.write_dataset(table, uri, mode="overwrite")

print(f"Created dataset at {uri}")
print(f"Schema: {dataset.schema}")
print(f"Number of rows: {dataset.count_rows()}")

dataset.create_index(
    "vector", index_type="IVF_PQ", num_partitions=4, num_sub_vectors=16
)
dataset.create_scalar_index("id", index_type="BTREE")

# Write more data to the existing dataset

# Generate additional random data
num_new_rows = 5000
new_ids = np.arange(num_rows, num_rows + num_new_rows, dtype=np.int32)
new_vectors = np.random.randn(num_new_rows, 16).astype(np.float32)

# Create new PyArrow table
new_table = pa.Table.from_arrays(
    [new_ids, pa.FixedSizeListArray.from_arrays(new_vectors.ravel(), 16)], schema=schema
)

# Append new data to the dataset
dataset.insert(new_table)

print(f"Appended {num_new_rows} new rows to dataset at {uri}")
print(f"New row count: {lance.dataset(uri).count_rows()}")

indices = dataset.list_indices()
print(f"Indices before optimization: {indices}")
# optimize with num_indices_to_merge=0
[...]

PR #4931, opened 2025-10-10
fix: optimize_indices may unexpectly delete delta indices

BubbleCal:
fix https://github.com/lancedb/lance/issues/4929


BubbleCal:
it's strange the test fails occasionally, will continue investigating

Issue #4897, opened 2025-10-03
Add RLE for block compression


westonpace:
Currently we only have bitpacking in the block compressor.  Block compression is used for compressing the repdef levels in the miniblock encoding.  We should also add RLE so we can mirror parquet (as repdef levels are often highly RLE compressible)

yingjianwu98:
I have some bandwidth to take on this, thanks!

PR #4937, opened 2025-10-13
feat: add RLE support for block

yingjianwu98:
resolve https://github.com/lancedb/lance/issues/4897

### DIFF
--- a/python/python/tests/test_memory_leaks.py
+++ b/python/python/tests/test_memory_leaks.py
@@ -5,7 +5,6 @@ from __future__ import annotations
 
 import gc
 import os
-import sys
 from typing import Callable
 
 import lance
@@ -17,23 +16,6 @@ MiB = 1024 * 1024
 
 def get_memory_usage() -> int:
     return psutil.Process(os.getpid()).memory_info().rss
-
-
-def maybe_trim_allocator() -> None:
-    # The leak checks in this file use RSS, which can include allocator arena growth.
-    # On Linux/glibc, malloc_trim helps return freed heap memory back to the OS and
-    # reduces test flakiness while still catching true leaks.
-    if sys.platform != "linux":
-        return
-    try:
-        import ctypes
-
-        libc = ctypes.CDLL("libc.so.6")
-        trim = getattr(libc, "malloc_trim", None)
-        if trim is not None:
-            trim(0)
-    except Exception:
-        return
 
 
 def assert_noleaks(
@@ -66,7 +48,6 @@ def assert_noleaks(
     for _ in range(warmup_iterations):
         operation()
     gc.collect()
-    maybe_trim_allocator()
 
     baseline = get_memory_usage()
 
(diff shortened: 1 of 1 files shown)


rust/lance-encoding/src/compression.rs
Xuanwo, line 1036: We need to support `RLE` here.

rust/lance-encoding/src/compression.rs
```diff
@@ -873,7 +906,7 @@ mod tests {
 
         // The compressor should be RLE wrapped in general compression
         assert!(debug_str.contains("GeneralMiniBlockCompressor"));
-        assert!(debug_str.contains("RleMiniBlockEncoder"));
+        assert!(debug_str.contains("RleEncoder"));
     }
```
Xuanwo, line 1262: Maybe it's worth to add an end-to-end test for RLE on block.

rust/lance-encoding/src/compression.rs
```diff
@@ -756,16 +771,43 @@ impl DecompressionStrategy for DefaultDecompressionStrategy {
...
+            Compression::Rle(rle) => {
+                let bits_per_value = validate_rle_compression(rle);
+                Ok(Box::new(RleDecompressor::new(bits_per_value)))
+            }
             _ => todo!(),
         }
     }
 }
+/// Validates RLE compression format and extracts bits_per_value
+fn validate_rle_compression(rle: &crate::format::pb21::Rle) -> u64 {
```
Xuanwo: Hi, could you explain more about this check?

rust/lance-encoding/src/encodings/physical/rle.rs
```diff
@@ -524,30 +550,60 @@ impl RleMiniBlockDecompressor {
...
+        // fetch the values_size
+        if data.len() < 8 {
+            return Err(Error::InvalidInput {
+                location: location!(),
+                source: format!("Insufficient data size: {}", data.len()).into(),
+            });
+        }
+
+        let values_size_bytes: [u8; 8] = data[..8].try_into().unwrap();
+        let values_size: usize = usize::from_le_bytes(values_size_bytes);
```
Xuanwo: We encoded `values_size` as `u64` and it's better to use the same value.

rust/lance-encoding/src/encodings/physical/rle.rs
```diff
@@ -385,13 +385,39 @@ impl MiniBlockCompressor for RleMiniBlockEncoder {
     }
 }
 
+impl BlockCompressor for RleEncoder {
```
Xuanwo, line 393: Maybe it's worth a comment here to declare that we are concating values & run_lengths buffers together.

rust/lance-encoding/src/compression.rs
yingjianwu98: ah, this block is from the original mini_block decompression validation.

I move it to a single function so it can be reused across mini_block and block

yingjianwu98:
@Xuanwo 
I have updated the PR based on comments! Let me know what you think, thanks!

PR #4938, opened 2025-10-13
chore: add index types to the name of the default index

Jay-ju:


jackye1995:
Mostly looks good to me, pending rebase

Jay-ju:
> Mostly looks good to me, pending rebase

@jackye1995  Thank you for your patience. However, we have fixed many historical compatibility issues here, which took some time.

wkalt:
This change would be breaking for users who rely on the default names for checking the existence or readiness of an index and could also result in a mix of schemes across a user's infrastructure which may be confusing - users might wrongly assume they can infer a particular scheme based on browsing a few tables.

Can we figure out a solution that does not requiring changing the default?

Another consideration is it may make sense in the future to support multiple indexes of the same type on a table. For example a user might be trying to pick between two parameterizations of an index or we might add partial index support. That makes me think we will need to extend the scheme again with a number or something. Maybe going straight to the number would make sense (meta_idx, meta_idx2, etc). @jackye1995 @Jay-ju thoughts?

If not for the existing default I would be +1 on the proposed scheme.

jackye1995:
What about we check for the current indexes of the column, and then decide what name to use? If there is no index, we use current name. If there is index but of a different type, we use this new name. I'd there is index but the same name, we add a random hash in the end.

wkalt:
I think if you are already checking index existence, a number conveys more information than a hash. At least you know it's the second one you created and you can rely on the default behavior if you really want to.

I think the proposal of including the type name only if an index exists already is an option,  and does solve the compatibility issue, but it still has the potential to be more confusing than a single rule. If a user sees meta_idx and meta_btree_idx, then they will wonder what the rule is and might conclude it's an implementation gap for bitmap indexes rather than an issue with the first created index.

jackye1995:
Actually, I was suggesting random hash trying to reduce possibility of creating new indexes of the same name. But technically we cannot really fully prevent it today because user might create index of the same name for Delta indexes so we cannot check for index of the same name and just fail the request during conflict resolution. 

Fundamentally we are lacking a flag to know if the index creation is for new index or delta index, and we might want to add that. 

And with that, we can just check current index names and increment a counter and create indexes with names like col1_idx2. What do we think?

jackye1995:
@wkalt looks like we replied at the same time lol

Jay-ju:
@wkalt @jackye1995 Thank you for your patience and comments
> If a user sees meta_idx and meta_btree_idx, then they will wonder what the rule is and might conclude it's an implementation gap for bitmap indexes rather than an issue with the first created index.

* Here I want to confirm `meta_idx` and `meta_btree_idx`. When was `meta_idx` created? Was it before this change? If so, I understand that there would be this confusion. The reason why we don't want to use "number" here will be explained below.

> we can just check current index names and increment a counter and create indexes with names like col1_idx2

* Here I would like to explain the reasons behind my current approach
  * The naming rule here is `{column_name}_{index_type}_id`. With this rule, there is an additional logic: if an index with the same name or type is to be created, a `replace` option needs to be supported. That is, only one index of the same type can be retained for the same column. Using just a number seems unable to convey this meaning. If the name is defined as `{column_name}_id_{int}`, then it would be impossible to replace the existing names.

What do you all think about this?

jackye1995:
> When was meta_idx created? Was it before this change? If so, I understand that there would be this confusion

Yes. 

I feel there are technically 2 problems, (1) the naming, (2) the conflict resolution logic. Today we don't distinguish creating a brand new index and creating a delta index, so we cannot fail concurrent creation of index of the same name. I think we should fix that. Once we fix that, the naming can be relatively flexible. If @wkalt is saying this has some backwards compatibility requirements, I think keep doing `col_idx` as the first name, and `col_idx_{number}` for more indexes on the same column seems fine to me, and having a middle `col_{type}_idx_{number}` also seems beneficial for user to more easily know what index type it is from name.

Issue #2271, opened 2024-04-29
Add when_matched_delete to merge_insert


westonpace:
The merge insert operation in other databases supports a "when matched delete" action.  This can be very useful for operations like "bulk delete given uuid".

PR #4939, opened 2025-10-13
feat: support when_matched_delete in merge_insert

jtuglu1:
Closes https://github.com/lancedb/lance/issues/2271

### DIFF
--- a/rust/lance/src/dataset/write/merge_insert/exec/delete.rs
+++ b/rust/lance/src/dataset/write/merge_insert/exec/delete.rs
@@ -74,22 +74,28 @@ impl DeleteOnlyMergeInsertExec {
         })
     }
 
-    /// Returns the merge statistics if the execution has completed.
+    /// Takes the merge statistics if the execution has completed.
     pub fn merge_stats(&self) -> Option<MergeStats> {
-        self.merge_stats.lock().ok().and_then(|guard| guard.clone())
+        self.merge_stats
+            .lock()
+            .ok()
+            .and_then(|mut guard| guard.take())
     }
 
-    /// Returns the transaction if the execution has completed.
+    /// Takes the transaction if the execution has completed.
     pub fn transaction(&self) -> Option<Transaction> {
-        self.transaction.lock().ok().and_then(|guard| guard.clone())
+        self.transaction
+            .lock()
+            .ok()
+            .and_then(|mut guard| guard.take())
     }
 
-    /// Returns the affected rows (deleted row addresses) if the execution has completed.
+    /// Takes the affected rows (deleted row addresses) if the execution has completed.
     pub fn affected_rows(&self) -> Option<RoaringTreemap> {
         self.affected_rows
             .lock()
             .ok()
-            .and_then(|guard| guard.clone())
+            .and_then(|mut guard| guard.take())
     }
 
     async fn collect_deletions(
--- a/rust/lance/src/dataset/write/merge_insert/exec/write.rs
+++ b/rust/lance/src/dataset/write/merge_insert/exec/write.rs
@@ -179,25 +179,31 @@ impl FullSchemaMergeInsertExec {
         })
     }
 
-    /// Returns the merge statistics if the execution has completed.
+    /// Takes the merge statistics if the execution has completed.
     /// Returns `None` if the execution is still in progress or hasn't started.
     pub fn merge_stats(&self) -> Option<MergeStats> {
-        self.merge_stats.lock().ok().and_then(|guard| guard.clone())
+        self.merge_stats
+            .lock()
+            .ok()
+            .and_then(|mut guard| guard.take())
     }
 
-    /// Returns the transaction if the execution has completed.
+    /// Takes the transaction if the execution has completed.
     /// Returns `None` if the execution is still in progress or hasn't started.
     pub fn transaction(&self) -> Option<Transaction> {
-        self.transaction.lock().ok().and_then(|guard| guard.clone())
+        self.transaction
+            .lock()
+            .ok()
+            .and_then(|mut guard| guard.take())
     }
 
-    /// Returns the affected rows (deleted/updated row addresses) if the execution has completed.
+    /// Takes the affected rows (deleted/updated row addresses) if the execution has completed.
     /// Returns `None` if the execution is still in progress or hasn't started.
     pub fn affected_rows(&self) -> Option<RoaringTreemap> {
         self.affected_rows
             .lock()
             .ok()
-            .and_then(|guard| guard.clone())
+            .and_then(|mut guard| guard.take())
     }
 
     /// Creates a filtered stream that captures row addresses for deletion and returns


jtuglu1:
@wjones127 I plan to push some changes to bump the coverage a bit. Lack of coverage is from explain/analyze plan methods. Something I think would be a good idea to future-proof this file is starting to standardize how these various combinations of configs should be unit tested together. We can always attempt to test all relevant combinations with targeted tests, but not sure how scalable that is. Maybe there's a better approach with parameterized tests, etc.?

rust/lance/src/dataset/write/merge_insert.rs
```diff
@@ -4790,4 +4807,192 @@ MergeInsert: on=[id], when_matched=UpdateAll, when_not_matched=InsertAll, when_n
...
+                    key,
+                    value_col.value(i)
+                );
+            }
+        }
+    }
+
+    #[rstest::rstest]
+    #[tokio::test]
+    async fn test_when_matched_delete_no_matches(
```
wjones127: One case you are missing that is likely to be important: Doing  `WhenMatched::Delete` with just a list of ids (omitting all other columns). Could you add that?

I think when you do that, you'll find we are missing some implementation on one of the code paths.

wjones127:
> Maybe there's a better approach with parameterized tests, etc.?

There's probably a better parametrized test. I'm working on adding a larger test suite, which can cover more write cases.

rust/lance/src/dataset/write/merge_insert.rs
jtuglu1: Yeah, @wjones127 non-full schema matching merges will break with this. From what I can see, there's 3 codepaths (fast path conditions, full_schema=true, full_schema=false). For my own understanding, I'm wondering why these codepaths need to be different? Is there a way we can condense these somehow? At the very least it might be good to document this for future changes (I can do this).

wjones127: Yes. We are in the middle of a refactor to consolidate these code paths. That refactor is tracked in this milestone: https://github.com/lancedb/lance/milestone/7

wjones127: In the end all of them should go through the fast path. If you want to just say that non-full schema isn't supported for now, that's okay.

jtuglu1: Gotcha. What do you recommend? I think the primary use-case of this feature might be for non-full schema batch deletes (like the example you gave above), but I don't want to create more work for you folks if the impl will be converted anyways to use datafusion.

wjones127: `WhenMatched::Delete` will probably only be used in cases where the only resulting action is `Delete`. In which case, the current fast path physical plan probably won't be used in the end anyways. I think it would be worth creating a specific fast code path for it, which generates a specialized physical plan for bulk deletion. The first part of the plan can be the same, but the write node at the end should be specialized for deletion.

This is because in the delete case, the final write step doesn't actually require reading all the columns even of the source data. They should be projected away before the join, since all you need to do procedurally is:

1. Join the id columns to find the matching row ids
2. Apply the row id deletions

So I would look into creating a specific physical write node that handles that step (2). 

In particular, the `necessary_children_exprs()` implementation can just say it needs `_rowaddr`:

https://github.com/lancedb/lance/blob/89c266f5ab442c778d37a82b01a09a6f912ce812/rust/lance/src/dataset/write/merge_insert/logical_plan.rs#L145

And then you should create a new physical node like `FullSchemaMergeInsertExec` that uses `apply_deletions`:

https://github.com/lancedb/lance/blob/89c266f5ab442c778d37a82b01a09a6f912ce812/rust/lance/src/dataset/write/merge_insert/exec/write.rs#L479-L482

python/python/lance/dataset.py
```diff
@@ -163,6 +163,16 @@ def when_matched_update_all(
         """
         return super(MergeInsertBuilder, self).when_matched_update_all(condition)
 
+    def when_matched_delete(self) -> "MergeInsertBuilder":
```
jackye1995, line 167: can we add a corresponding test in `test_dataset.py`

rust/lance/src/dataset/write/merge_insert.rs
```diff
@@ -1437,11 +1442,31 @@ impl MergeInsertJob {
 
         let has_scalar_index = self.join_key_as_scalar_index().await?.is_some();
 
+        // Check if this is a delete-only operation (no update/insert writes needed from source)
+        // For delete-only, we don't need the full source schema, just key columns for matching
+        let is_delete_only = matches!(
```
jackye1995: this name seems misleading, since it can be both delete and do nothing, it's more like `no_upsert`?

rust/lance/src/dataset/write/merge_insert.rs
```diff
@@ -1437,11 +1442,31 @@ impl MergeInsertJob {
...
+        let schema_ok = is_full_schema || (is_delete_only && source_has_key_columns);
+
         Ok(matches!(
             self.params.when_matched,
-            WhenMatched::UpdateAll | WhenMatched::UpdateIf(_) | WhenMatched::Fail
+            WhenMatched::UpdateAll
+                | WhenMatched::UpdateIf(_)
+                | WhenMatched::Fail
+                | WhenMatched::Delete
+                | WhenMatched::DoNothing
```
jackye1995: we should also add tests related to DoNothing

jtuglu1: Considering this is not related to this change, can I add in follow-up?

rust/lance/src/dataset/write/merge_insert.rs
```diff
@@ -1437,11 +1442,31 @@ impl MergeInsertJob {
...
+                .fields()
+                .iter()
+                .any(|f| f.name() == key.as_str())
+        });
+        let schema_ok = is_full_schema || (is_delete_only && source_has_key_columns);
+
         Ok(matches!(
             self.params.when_matched,
-            WhenMatched::UpdateAll | WhenMatched::UpdateIf(_) | WhenMatched::Fail
+            WhenMatched::UpdateAll
```
jackye1995, line 1474: since we have made these now to be `can_use_create_plan`, we should also add tests checking the query plan is as expected, see how we use `assert_plan_node_equals` to do that in other similar tests.

rust/lance/src/dataset/write/merge_insert.rs
jackye1995: Sure, in that case let's just remove DoNothing related changes here

rust/lance/src/dataset/write/merge_insert.rs
jtuglu1: As I only added `WhenMatched::Delete` here(removing `WhenMatched::DoNothing`), so I'll assert the plans are ok as well for that one only.

jtuglu1:
@wjones127 @jackye1995 thoughts here?

jackye1995:
@jtuglu1 I thought about this a bit more, and also saw Will's old comment (sorry totally missed that previously) I think what Will suggested makes sense, I pushed a commit to add DeleteOnlyMergeInsertExec with some additional refactoring, let me know if this looks good to you

jtuglu1:
Yeah I considered this (and Will's comment), but I didn't really like the idea of adding physical plans for every single type of enum (.*MergeInsertExec). It seemed easier (to me) to handle things all in the same codepath (the projection optimizations, etc.). I haven't really looked through your changes here yet, but are there any performance benefits as compared to the current implementation?

Edit: looks like this is skipping write step which is I guess some time saved.

jackye1995:
> Edit: looks like this is skipping write step which is I guess some time saved.

yes that's the main goal

rust/lance/src/dataset/write/merge_insert/exec/delete.rs
```diff
@@ -0,0 +1,315 @@
...
+    pub fn transaction(&self) -> Option<Transaction> {
+        self.transaction.lock().ok().and_then(|guard| guard.clone())
+    }
+
+    /// Returns the affected rows (deleted row addresses) if the execution has completed.
+    pub fn affected_rows(&self) -> Option<RoaringTreemap> {
+        self.affected_rows
+            .lock()
+            .ok()
+            .and_then(|guard| guard.clone())
```
jtuglu1: Potentially naive question: @jackye1995 is this a common pattern in Rust? If I'm reading this right, I see it used elsewhere in the codebase (in write.rs), but I wonder if it's really necessary to do a copy of the rows.

jackye1995: the clone is necessary here because the MutexGuard from lock() only lives for the duration of the method call. We can't return a reference to data inside the mutex since the guard gets dropped at function return, releasing the lock.

It does return a copy of the affected_rows, but I thought it should be fine since it is only accessed after execution has completed to perform the actual transactional commit and use for conflict resolution. If we do an Arc to avoid copying, I think it would make things a lot more complicated, and the copy is likely inevitable in the transactional commit code path.

jtuglu1: (I have not profiled this yet, but I'm assuming it's not a problem as existing write path does this same thing). Maybe since this code is doing some more I/O-intensive logic (the transaction itself), it doesn't show up in fg or similar profiles since most time might be spent writing/deleting the data. Given this is being called only once in both paths, I think this should be fine for now.

PR #4941, opened 2025-10-13
fix: rebuild HNSW graph while remapping it

BubbleCal:
HNSW can't be simply remapped because the graph structure also needs to be rebuilt after the vectors changed,
we didn't allow to remap HNSW before this, this PR allows remapping HNSW by rebuilding the graph.

Before this, compaction on a table with HNSW index would raise error "remap HNSW is not supported yet". Users would be able to do remap (triggered by compaction) after this.

### DIFF
--- a/rust/lance-index/src/vector/hnsw/builder.rs
+++ b/rust/lance-index/src/vector/hnsw/builder.rs
@@ -781,7 +781,7 @@ impl IvfSubIndex for HNSW {
     ) -> Result<Self> {
         // We can't simply remap the row ids in the graph because the vectors are changed,
         // so the graph needs to be rebuilt.
-        HNSW::index_vectors(store, self.inner.params.clone())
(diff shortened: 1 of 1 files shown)


wkalt:
What is the scope of this bug/fix? Can you elaborate in the PR description?

wkalt:
@BubbleCal can you also add impact? Like what does a user see before this, when does it cause a problem? We need to be able to understand whether this is a critical user-facing issue and if so when it would kick in.

Does it affect only HNSW users or IVF-PQ users too?

BubbleCal:
> @BubbleCal can you also add impact? Like what does a user see before this, when does it cause a problem? We need to be able to understand whether this is a critical user-facing issue and if so when it would kick in.
> 
> Does it affect only HNSW users or IVF-PQ users too?

KK added, the changes only impact HNSW users, even IVF_PQ now would use HNSW, but it would never call remap

PR #4945, opened 2025-10-13
feat: implement add_bases api to add bases to lance dataset

jaystarshot:
Follow up of [commit](https://github.com/lancedb/lance/commit/dba1e0f2c70c8b41dcf7dc29db37422022c861d9#diff-e053fd3f17169afa9b8ec27bc109b55a1fbae5442d6d0818fc15baef76286b6f).

Sample API

```
 dataset.add_bases([
            DatasetBasePath(test2_base, name="test2_base"),
            DatasetBasePath(test3_base, name="test3_base"),
        ])
```

### DIFF
--- a/python/python/lance/dataset.py
+++ b/python/python/lance/dataset.py
@@ -2056,7 +2056,9 @@ class LanceDataset(pa.dataset.Dataset):
         """
         self._ds.restore()
 
-    def add_bases(self, new_bases: list, transaction_properties: Optional[Dict[str, str]] = None):
+    def add_bases(
+        self, new_bases: list, transaction_properties: Optional[Dict[str, str]] = None
+    ):
         """
         Add new base paths to the dataset for multi-base storage.
 


protos/transaction.proto
```diff
@@ -273,6 +273,12 @@ message Transaction {
     repeated MemWalIndexDetails.MemWal removed = 3;
   }
 
+  // An operation that adds new base paths to the dataset.
+  message AddBases {
```
jackye1995: I think we should probably make this `UpdateBases`. Now we will only do add_base, but it allows us to do other things later.

rust/lance/src/dataset.rs
```diff
@@ -1341,6 +1341,44 @@ impl Dataset {
...
+    ///
+    /// # Returns
+    ///
+    /// Returns a new `Dataset` instance with the updated manifest containing the
+    /// new base paths.
+    pub async fn add_bases(
+        self: &Arc<Self>,
+        new_bases: Vec<lance_table::format::BasePath>,
+    ) -> Result<Self> {
+        use crate::dataset::transaction::{Operation, Transaction};
```
jackye1995: nit: prefer imports at top

rust/lance/src/dataset.rs
```diff
@@ -1341,6 +1341,44 @@ impl Dataset {
...
+        use crate::dataset::write::CommitBuilder;
+
+        let operation = Operation::AddBases { new_bases };
+
+        let transaction = Transaction::new(
+            self.manifest.version,
+            operation,
+            None, // No blob operation for AddBases
+            None, // No transaction properties
```
jackye1995: might want to consider supporting this, it allows you to add some audit info like who added this, and when this is added, for what purpose, etc.

rust/lance/src/dataset.rs
jaystarshot: Done thanks

protos/transaction.proto
jaystarshot: Done thanks

rust/lance/src/dataset.rs
jaystarshot: Done thanks

python/python/lance/dataset.py
```diff
@@ -2056,6 +2056,30 @@ def restore(self):
         """
         self._ds.restore()
 
+    def add_bases(self, new_bases: list, transaction_properties: dict = None):
```
jackye1995: the type hints are not consistent with the doc: `list` -> `list[DatasetBasePath]`, `dict` -> `Optional[Dict[str, str]]`

python/python/lance/dataset.py
```diff
@@ -2056,6 +2056,30 @@ def restore(self):
...
+        """
+        Add new base paths to the dataset for multi-bucket storage.
```
jackye1995: nit: multi-base storage

python/python/tests/test_multi_base.py
```diff
@@ -486,3 +486,563 @@ def test_fragment_distribution(self):
...
+        assert len(result) == 40
+
+        # Verify schema is preserved
+        assert list(result.columns) == ["id", "name", "age", "score", "active"]
+
+        # Test filtering on different columns
+        filtered = dataset.scanner(filter="age > 30").to_table().to_pandas()
+        assert all(filtered["age"] > 30)
+
+    def test_add_bases_method_chaining(self):
```
jackye1995: I don't think we really need to test this one, because when we chain it, it's just doing multiple commits, it's not batching all the add calls to make a single commit.

python/python/tests/test_multi_base.py
```diff
@@ -486,3 +486,563 @@ def test_fragment_distribution(self):
...
+        # Verify distribution
+        assert len(result[result["source"] == "initial"]) == 100
+        assert len(result[result["source"] == "new_base1"]) == 200
+        assert len(result[result["source"] == "new_base2"]) == 200
+        assert len(result[result["source"] == "new_base3"]) == 200
+
+        # Verify all IDs are present
+        assert set(result["id"]) == set(range(700))
+
+    def test_add_bases_with_complex_schema(self):
```
jackye1995: I don't think we need to test this, the feature has nothing to do with schema

rust/lance/src/dataset.rs
```diff
@@ -9015,4 +9048,155 @@ mod tests {
             .unwrap();
         assert!(branches.is_empty());
     }
+
+    #[tokio::test]
+    async fn test_add_bases() {
```
jackye1995, line 9053: we should probably add a few tests for commit conflict. There are some good examples of such tests in for example https://github.com/lancedb/lance/blob/main/rust/lance/src/dataset/optimize.rs#L2219. The idea is that you can create a dataset, clone it so that it can act as if it is running and committing concurrently, so you can test out how the conflict resolver is actually behaving.

jackye1995 (approved):
looks good to me! pending CI

PR #4949, opened 2025-10-14
feat(python): support shallow_clone

### LINKED ISSUE #4856: Support shallow_clone in Python

majin1102:
Close #4856

### DIFF
--- a/rust/lance-index/src/scalar/inverted/index.rs
+++ b/rust/lance-index/src/scalar/inverted/index.rs
@@ -1401,7 +1401,7 @@ impl PostingListReader {
     }
 
     async fn read_positions(&self, token_id: u32) -> Result<ListArray> {
-        Ok(self.index_cache.get_or_insert_with_key(PositionKey { token_id }, || async move {
+        let positions = self.index_cache.get_or_insert_with_key(PositionKey { token_id }, || async move {
             let batch = self
                 .reader
                 .read_range(self.posting_list_range(token_id), Some(&[POSITION_COL]))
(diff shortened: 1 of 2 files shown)


majin1102:
Ready for review @jackye1995

python/python/lance/dataset.py
```diff
@@ -3224,6 +3224,49 @@ def validate(self):
...
+        Shallow clone the specified version into a new dataset at target_path.
+
+        This operation copies only the dataset metadata and references, without
+        rewriting data files.
+
+        Parameters
+        ----------
+        target_path : str or Path
+            The URI or filesystem path to clone the dataset into.
+        version : int or str
```
jackye1995: this is not up to date with the `Union[int, str, Tuple[int, str]]`

python/python/lance/dataset.py
```diff
@@ -3224,6 +3224,49 @@ def validate(self):
...
+            A new LanceDataset representing the shallow-cloned dataset.
+        """
+        if isinstance(target_path, Path):
+            target_uri = os.fspath(target_path)
+        else:
+            target_uri = target_path
+
+        new_inner = self._ds.shallow_clone(target_uri, version, storage_options)
+
+        ds = LanceDataset.__new__(LanceDataset)
```
jackye1995: can we just directly load a new dataset, in stead of overriding these parameters? It feels fragile if we add anything new to LanceDataset that needs override.

majin1102: Sounds reasonable.

So we just ignore the shallow_clone returned dataset right? the manifest is cached anyway. Just make sure this is noticed

jackye1995: We could have some static method like `LanceDataset.clone_from` to do it, but I feel it is not worth the complexity, but let me know if you disagree. 

But this reminds me that, we should add a kwarg so it can pass in whatever other parameters (e.g. `read_params`, `default_scan_options`) that the user wants to open the cloned dataset.

majin1102: Done, PTAL when you have time

jackye1995 (approved):

jackye1995:
Error unrelated to this PR and exists on main branch, merging

Issue #2862, opened 2024-09-11
Allow packing of variable length columns


westonpace:
We added a packed struct encoding in #2601 but it does not support variable-width columns.  We should extend that encoding or add a new encoding which is capable of packing any kind of column (fixed or variable).  The resulting encoding will necessarily be a variable width encoding.

Xuanwo:
Just hitting this while working on https://github.com/lancedb/lance/pull/4948, will take a look first.

PR #4950, opened 2025-10-14
feat: add VariablePackedStruct defination

Xuanwo:
Part of https://github.com/lancedb/lance/issues/2862

This PR will add VariablePackedStruct defination in lance first.

### DIFF
--- a/.bumpversion.toml
+++ b/.bumpversion.toml
@@ -75,6 +75,16 @@ replace = 'lance-linalg = {{ version = "={new_version}"'
 
 [[tool.bumpversion.files]]
 filename = "Cargo.toml"
+search = 'lance-namespace = {{ version = "={current_version}"'
+replace = 'lance-namespace = {{ version = "={new_version}"'
+
+[[tool.bumpversion.files]]
+filename = "Cargo.toml"
+search = 'lance-namespace-impls = {{ version = "={current_version}"'
+replace = 'lance-namespace-impls = {{ version = "={new_version}"'
+
+[[tool.bumpversion.files]]
+filename = "Cargo.toml"
 search = 'lance-table = {{ version = "={current_version}"'
 replace = 'lance-table = {{ version = "={new_version}"'
 
@@ -146,6 +156,16 @@ replace = 'version = "{new_version}"'
 
 [[tool.bumpversion.files]]
 filename = "rust/lance-linalg/Cargo.toml"
+search = 'version = "{current_version}"'
+replace = 'version = "{new_version}"'
+
+[[tool.bumpversion.files]]
+filename = "rust/lance-namespace/Cargo.toml"
+search = 'version = "{current_version}"'
+replace = 'version = "{new_version}"'
+
+[[tool.bumpversion.files]]
+filename = "rust/lance-namespace-impls/Cargo.toml"
 search = 'version = "{current_version}"'
 replace = 'version = "{new_version}"'
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -409,6 +409,16 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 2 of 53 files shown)


protos/encodings_v2_2.proto
Xuanwo, line 1: I don't know how we add new defs for `v2.2`, is this the correct way?

cc @westonpace

westonpace: I think we should maybe just put them in the 2_1 proto file but make sure the changes are backwards compatible.  What do you think?  I worry it might get to be too much copy/paste overhead if we keep making new proto files.

protos/encodings_v2_2.proto
```diff
@@ -0,0 +1,550 @@
...
+  // Layout for a single child field inside each packed row
+  message FieldLayout {
+    oneof details {
+      // Bit width of a fixed-width child value within the row buffer
+      uint64 bits_per_value = 1;
+      // Bit width of the length prefix for a variable-width child value
+      uint64 bits_per_length = 2;
+    }
+  }
+}
```
westonpace: Can you describe a little more how you expect this to work?

Will the children be compressed?  Do they need to be compressed with per-value compressors or can they use mini-block compressors?  From the message it seems almost like the children are zipped together and then compressed as rows.  Is this intentional?  Do we think row compression will be better than column compression for some reason?

protos/encodings_v2_2.proto
Xuanwo: I’m fine with adding them to the `2.1` proto in a compatible way. My intention is to ensure no changes are made to the `2.1` proto so we can more easily detect unexpected proto changes.

protos/encodings_v2_2.proto
Xuanwo: Sure.

> Will the children be compressed?

Good question. Yes, the children are zipped together and then compressed into rows. I designed it this way to ensure we can extract the structure as a whole. My expectation is that the packed struct won't be overly large. What do you think?

The design is mainly served the need from blob v2.

protos/encodings_v2_2.proto
Xuanwo: Fixed.

Xuanwo:
After our offline discussion, I redesigned VariablePackedStruct to compress first and then transpose. Please review it again, @westonpace

protos/encodings_v2_1.proto
```diff
@@ -379,23 +379,40 @@ message FixedSizeList {
...
 message PackedStruct {
-  // The number of bits per value for each field
+  // The number of bits contributed by each child field in the packed row
   repeated uint64 bits_per_value = 1;
-  // The compression used to store the values data block
+  // The compression used to store the packed fixed-width values
   CompressiveEncoding values = 2;
 }
 
+// Variable-width packed struct encoding (2.1 extension)
```
westonpace:
```suggestion
// Variable-width packed struct encoding (2.2 extension)
```

protos/encodings_v2_1.proto
```diff
@@ -379,23 +379,40 @@ message FixedSizeList {
...
+//
+// Each child value is compressed independently before being transposed into
+// a row-major layout. This preserves per-field compression boundaries at the
+// cost of disabling mini-block compression. Readers must prefer this field
+// when present and fall back to the legacy encoding otherwise.
```
westonpace, line 398: I'm not sure I understand this last sentence.  Readers should just read whatever was written.  Writers should always choose the newer one.  Right?

westonpace (approved):
Yes, this definition seems correct.

protos/encodings_v2_1.proto
Xuanwo: Yep!

PR #4956, opened 2025-10-15
feat: support WhenMatched::Fail for MergeInsert

### LINKED ISSUE #4944: support WhenMatched::Fail for MergeInsert
This would be useful when the system wants to ensure that one record is not updated after inserted (e.g. atomically create an item or fail if it exists, where an item is a row in the Lance table)

ddupg:
Closes #4944

### DIFF
--- a/java/lance-jni/src/merge_insert.rs
+++ b/java/lance-jni/src/merge_insert.rs
@@ -113,6 +113,7 @@ fn extract_when_matched<'local>(env: &mut JNIEnv<'local>, jparam: &JObject) -> R
             Some(expr) => Ok(WhenMatched::UpdateIf(expr)),
             None => Err(Error::input_error("No matched updated expr".to_string())),
         },
+        "Fail" => Ok(WhenMatched::Fail),
         _ => Err(Error::input_error(format!(
             "Illegal when_matched: {when_matched}",
         ))),
--- a/java/src/main/java/com/lancedb/lance/merge/MergeInsertParams.java
+++ b/java/src/main/java/com/lancedb/lance/merge/MergeInsertParams.java
@@ -88,6 +88,19 @@ public class MergeInsertParams {
   }
 
   /**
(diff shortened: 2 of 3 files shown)


jackye1995 (approved):
looks good to me! @ddupg do you plan to also do this for java? If not please help create an issue tracking that

ddupg:
> looks good to me! @ddupg do you plan to also do this for java? If not please help create an issue tracking that

@jackye1995 Sorry I missed the java binding, I've implemented this for java now

jtuglu1:
Is this purposefully only handling the case where the entire source schema is passed in? Have we tested whether this will work if a subset of the schema is passed in, i.e if we pass in just an `id` column to check the presence of in the target?

I think that case might be missing.

jackye1995:
@jtuglu1 I think the expectation of this is typically that you have data of the entire row that you only want to insert but not update. For your case, why would you use this instead of just count rows that matches the given ids?

But from completeness perspective, you are right, @ddupg I agree we should add a check for partial schema input

jtuglu1:
> For your case, why would you use this instead of just count rows that matches the given is?

If you want to do an atomic batch check of any rows which match a set of ids from an id column, failing if any exist, for example.

I agree you could likely do this behavior with a count() but given it's being provided as an option, I think it's worth to consider whether we want to implement that case.

jackye1995:
> I agree you could likely do this behavior with a count() but given it's being provided as an option, I think it's worth to consider whether we want to implement that case.

I think the baseline is that the operation needs to have to result in a possible atomic write to the table. In the example you describe, it's basically a pure check/read operation, there is no possible outcome that could be a write, that's where I feel it is a bit strange to use it through `MergeInsert` which is after all a write operation.

I think if match then fail with a partial schema might make sense when we do if not matched then delete, then it makes sense to just pass in a partial schema for the `on` column.

PR #4959, opened 2025-10-15
feat: add support for large minichunk size (u32) in format v2.2

niyue:
## Overview

Following the discussion in [this thread](https://github.com/lancedb/lance/discussions/4848)(https://github.com/lancedb/lance/discussions/4848), this PR expands the `MiniBlockChunk` chunk size by changing its type from `u16` to `u32`. This allows the encoder such as binary encoding to use larger chunk sizes, which can improve compression ratios for general compression and potentially other compression schemes as well.

## Description

This PR is still experimental and open to feedback. The main changes are as follows:

* Added a new protobuf property `has_large_chunk` for `MiniBlockLayout`
* Introduced a new metadata key: `lance-encoding:minichunk-size`, allowing users to specify the mini chunk size for binary columns (or any other columns that respect this parameter).
* Updated `MiniBlockChunk`:

  * The `buffer_sizes` field is now stored as `u32` in memory.
  * **For Lance format v2.1:** Chunk buffer sizes are still stored as `u16` on disk, but are converted to `u32` after being loaded into memory as `MiniBlockChunk`.
  * **For Lance format v2.2:** The default binary mini chunk size remains 4KB. However, users can override this by setting the new metadata key mentioned above for specific fields, enabling chunk sizes of 64KB or larger.

## TODO

* The `has_large_chunk` protobuf property was temporarily added to v2.1 because a v2.2 protobuf definition does not yet exist. This may need to be adjusted later once v2.2 is introduced.


cmccabe:
cc @westonpace

rust/lance-encoding/benches/decoder.rs
```diff
@@ -45,12 +45,15 @@ const PRIMITIVE_TYPES: &[DataType] = &[
...
-    keep_original_array: true,
-    buffer_alignment: 64,
-};
+fn default_encoding_options(version: LanceFileVersion) -> EncodingOptions {
+    EncodingOptions {
+        cache_bytes_per_column: 8 * 1024 * 1024,
+        max_page_bytes: 32 * 1024 * 1024,
+        keep_original_array: true,
+        buffer_alignment: 64,
+        support_large_chunk: version >= LanceFileVersion::V2_2,
```
westonpace: I like putting the setting here.  We have a lot of these checks just scattered throughout the code and I think it is cleaner here.  Maybe we could include the version in the options and do something like...

```
    EncodingOptions {
        ...,
        version
    }
}

impl EncodingOptions {
    pub fn support_large_chunk(&self) -> bool {
        self.version >= LanceFileVersion::V2_2
    }
}
```

This way we can add new version-dependent configs without adding new fields.  We could then add them as fields only if we have a need for users to disable them (even on a newer version).

protos/encodings_v2_1.proto
```diff
@@ -112,6 +112,9 @@ message MiniBlockLayout {
   // The page already records how many rows are in the page.  For mini-block we also need to know how
   // many "items" are in the page.  A row and an item are the same thing unless the page has lists.
   uint64 num_items = 9;
+
+  // Since Lance 2.2, we support miniblocks with larger chunk size (>= 64KB)
+  bool support_large_chunk = 10;
```
westonpace: Nit: Instead of `support_large_chunk` can we call this `has_large_chunk` or `has_large_chunks` (here and pretty much everywhere else)?

Before we write the data I think `support_large_chunk` (can we?) is ok.  After we write the data I think we should use `has_large_chunk` (did we?)

rust/lance-encoding/src/compression_config.rs
```diff
@@ -67,6 +67,9 @@ pub struct CompressionFieldParams {
 
     /// Byte stream split mode for floating point data
     pub bss: Option<BssMode>,
+
+    /// Minichunk size threshold for binary encoding
+    pub binary_minichunk_size: Option<i64>,
```
westonpace: Nit: let's just call it `minichunk_size`?

rust/lance-encoding/src/constants.rs
```diff
@@ -13,6 +13,8 @@ pub const COMPRESSION_META_KEY: &str = "lance-encoding:compression";
 pub const COMPRESSION_LEVEL_META_KEY: &str = "lance-encoding:compression-level";
 /// Metadata key for specifying RLE (Run-Length Encoding) threshold
 pub const RLE_THRESHOLD_META_KEY: &str = "lance-encoding:rle-threshold";
+/// Metadata key for specifying binary minichunk size
+pub const BINARY_MINICHUNK_SIZE_META_KEY: &str = "lance-encoding:binary-minichunk-size";
```
westonpace: Same here, can we just call it `lance-encoding:minichunk-size`?  I'm worried users will think this setting is specific to binary types.

rust/lance-encoding/src/encoder.rs
```diff
@@ -233,6 +233,10 @@ pub struct EncodingOptions {
     /// The encoder needs to know this so it figures the position of out-of-line
     /// buffers correctly
     pub buffer_alignment: u64,
+    /// If true (the default in Lance 2.2+), miniblock buffer sizes are u32  
+    /// to allow storing bigger chunk size for better compression.
+    /// For Lance 2.1, this is false, and miniblock buffer sizes are u16.
+    pub support_large_chunk: bool,
```
westonpace: This one we can leave as `support_large_chunk`.

rust/lance-encoding/src/encodings/logical/primitive.rs
```diff
@@ -446,10 +448,18 @@ impl DecodeMiniBlockTask {
...
+            |bytes| u32::from_le_bytes([bytes[0], bytes[1], bytes[2], bytes[3]])
+        } else {
+            |bytes| u16::from_le_bytes([bytes[0], bytes[1]]) as u32
+        };
+
         let buffer_sizes = (0..self.num_buffers)
             .map(|_| {
-                let size = u16::from_le_bytes([buf[offset], buf[offset + 1]]);
-                offset += 2;
+                let size = read_fn(&buf[offset..offset + read_size]);
```
westonpace: Does this introduce a virtual function call or does rust optimize it out you think?  Since this is on a per-chunk basis it probably doesn't matter too much.

rust/lance-encoding/src/encodings/logical/primitive.rs
```diff
@@ -631,6 +641,7 @@ struct MiniBlockDecoder {
     num_rows: u64,
     num_buffers: u64,
     dictionary: Option<Arc<DataBlock>>,
+    support_large_chunk: bool,
```
westonpace: `support_large_chunk -> has_large_chunk`

rust/lance-encoding/src/encodings/logical/primitive.rs
```diff
@@ -3510,17 +3580,28 @@ impl PrimitiveStructuralEncoder {
             }
 
             let chunk_bytes = data_buffer.len() - start_pos;
-            assert!(chunk_bytes <= 32 * 1024);
+            let max_chunk_size = if support_large_chunk {
+                2 * 1024 * 1024 * 1024 // 2GB limit with u32 metadata
```
westonpace: Nit: 2GiB should be more than enough but isn't the limit technically 4GiB?

rust/lance-encoding/src/encodings/logical/primitive.rs
```diff
@@ -3476,9 +3541,14 @@ impl PrimitiveStructuralEncoder {
                 data_buffer.extend_from_slice(&bytes_def.to_le_bytes());
             }
 
-            for buffer_size in &chunk.buffer_sizes {
-                let bytes = *buffer_size;
-                data_buffer.extend_from_slice(&bytes.to_le_bytes());
+            if support_large_chunk {
```
westonpace, line 3612: Do we need to adjust the padding logic below to make sure data starts on 8-byte boundary?  I don't think so (since we use `data_buffer.len`) but just want to double-check.

rust/lance-encoding/benches/decoder.rs
niyue: That's a great suggestion. I will change the code accordingly

protos/encodings_v2_1.proto
niyue: Sure. Now I rename them, and it is called `support_large_chunk` on the encoding path, and it is called `has_large_chunk` on the decoding path.

rust/lance-encoding/src/compression_config.rs
niyue: Sure

rust/lance-encoding/src/constants.rs
niyue: Done

rust/lance-encoding/src/encodings/logical/primitive.rs
niyue: No, there are still some runtime function calls, but they are not virtual. After a closer look, this can indeed be improved. 

I’ve now moved the `has_large_chunk` check outside the loop since it remains constant during iteration, and I’m using a const generics function to eliminate the need for checking this condition for every item in the loop.

rust/lance-encoding/src/encodings/logical/primitive.rs
niyue: You’re right — I confused it with the i32 limit.

rust/lance-encoding/src/encodings/logical/primitive.rs
niyue: I don’t think we need to adjust the padding logic. It looks good to me.

niyue:
@westonpace I’ve updated the code based on your feedback, and please take a look. Currently, only the `binary` and `fsst` encoders (which internally use `binary`) make use of the newly added parameter. I’m not sure whether this parameter would also benefit the other encoders, so I haven’t modified them yet. Please let me know if you’d prefer to update all of them together in this PR. Thanks!

rust/lance-encoding/benches/decoder.rs
```diff
@@ -45,12 +45,15 @@ const PRIMITIVE_TYPES: &[DataType] = &[
...
-};
+fn default_encoding_options() -> EncodingOptions {
+    EncodingOptions {
+        cache_bytes_per_column: 8 * 1024 * 1024,
+        max_page_bytes: 32 * 1024 * 1024,
+        keep_original_array: true,
+        buffer_alignment: 64,
+        version: LanceFileVersion::default(),
+    }
+}
```
westonpace: Should we do `impl Default for EncodingOptions` instead of adding this method?  That would be more rust-friendly I think.

rust/lance-encoding/src/compression.rs
```diff
@@ -335,6 +335,27 @@ impl DefaultCompressionStrategy {
             }
         }
 
+        // Parse minichunk size
+        if let Some(minichunk_size_str) = field
```
westonpace, line 339: We should update `docs/src/format/file/encoding.md`.  Can you make a follow-up issue for that?

rust/lance-encoding/src/encodings/logical/primitive/miniblock.rs
```diff
@@ -44,8 +44,9 @@ pub struct MiniBlockCompressed {
 pub struct MiniBlockChunk {
     // The size in bytes of each buffer in the chunk.
     //
-    // The total size must be less than or equal to 8Ki - 6 (8188)
-    pub buffer_sizes: Vec<u16>,
+    // For Lance 2.1, the chunk size is limited to 32KiB, so only 16-bit is stored here and used.
```
westonpace:
```suggestion
    // In Lance 2.1, the chunk size is limited to 32KiB, so only 16-bits are used.
```

We are still storing them as u32 in the vec .

rust/lance-encoding/benches/decoder.rs
niyue: I noticed that the `Default` trait had already been implemented for `EncodingOptions`, so we can simply reuse it with a small adjustment. Therefore, I removed the custom `default_encoding_options` helper and replaced it with `EncodingOptions::default()`.

rust/lance-encoding/src/encodings/logical/primitive/miniblock.rs
niyue: You're right, it's stored as `u32`. My wording was not very accurate. English is not my first language, so please excuse the confusion. I've updated the comment here.

rust/lance-encoding/src/compression.rs
niyue: Sure. I log a new issue https://github.com/lancedb/lance/issues/5240 for tracking this doc update

niyue:
@westonpace I've updated the code based on your feedback, and a new issue https://github.com/lancedb/lance/issues/5240 was logged for tracking the encoding doc update. Please take a look when you have time. Thanks!

westonpace:
😆 it seems the move has left me without write permissions.  I'll merge after I figure that out.

niyue:
Hi @westonpace 

Just a gentle ping on this PR, please let me know if you need anything from me. 

No rush at all, just keeping it on your radar :) 

Thanks!

PR #4960, opened 2025-10-15
chore: add github workflow to publish protobuf to BSR

### LINKED ISSUE #4197: Publish Lance protobuf definitions to Buf
Based on some asks from Discord.

The user would like to consume the raw protobuf definition to build a native Lance reader and writer in Java, and thus would like to consume the protobuf definitions directly and use its Java generated client.

One way, similar to what Substrait did, is to publish the definitions to https://buf.build. 

Create this issue to track it.

luohao:
Closes #4197

### DIFF
--- a/.github/workflows/buf-publish.yml
+++ b/.github/workflows/buf-publish.yml
@@ -1,16 +1,26 @@
 name: Publish protobuf to BSR
 
 on:
-  push:
-    paths:
-      - 'buf.yaml'
-      - 'protos/**'
+  release:
+    types: [published]
+  workflow_dispatch:
+    inputs:
+      ref:
+        description: 'Git ref to checkout (branch, tag, or SHA)'
+        required: false
+        default: ''
+        type: string
+
 jobs:
   push-module:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
-      - uses: bufbuild/buf-setup-action@v1
-      - uses: bufbuild/buf-push-action@v1
-        with:
-          buf_token: ${{ secrets.BUF_TOKEN }}
+    - uses: actions/checkout@v4
+      with:
+        ref: ${{ inputs.ref || github.ref }}
+        fetch-depth: 0
+        lfs: true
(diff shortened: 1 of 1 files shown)


.github/workflows/buf-publish.yml
```diff
@@ -0,0 +1,19 @@
+name: Publish protobuf to BSR
+
+on:
+  pull_request:
```
jackye1995: I think this should only be run on main?

.github/workflows/buf-publish.yml
```diff
@@ -0,0 +1,19 @@
...
+jobs:
+  push-module:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: bufbuild/buf-setup-action@v1
+      - uses: bufbuild/buf-push-action@v1
+        with:
+          buf_token: ${{ secrets.BUF_TOKEN }}
+          create_visibility: private
```
jackye1995: this should be public

.github/workflows/buf-publish.yml
luohao: Yes this is jus to trigger a run in my PR to test it.

Looks like the BUF_TOKEN secret doesnt exist or is empty string, can you take a look?

jackye1995: actually this should not really be triggering on main branch push, but rather on release or workflow dispatch makes more sense:

```yml
on:
  release:
    types: [published]
  workflow_dispatch:
    inputs:
      ref:
        description: 'Git ref to checkout (branch, tag, or SHA)'
        required: false
        default: ''
        type: string
```

and then later

```yml
    - uses: actions/checkout@v4
      with:
        ref: ${{ inputs.ref || github.ref }}
        fetch-depth: 0
        lfs: true
```

jackye1995: > Looks like the BUF_TOKEN secret doesnt exist or is empty string, can you take a look?

I think the issue is you don't have access to the BUF_TOKEN in the main repo, but after we merge it this should work.

PR #4961, opened 2025-10-15
feat(java): supports building scalar indices distributedly in java module

steFaiz:
This PR is about to bring the distributed index creation functionality (see https://github.com/lancedb/lance/pull/4667, https://github.com/lancedb/lance/pull/4578) to java module, which is aligned with the python implementation.


steFaiz:
This PR is ready for review, PTAL, thanks! @jackye1995

java/src/main/java/com/lancedb/lance/Dataset.java
```diff
@@ -621,17 +622,17 @@ public void restore() {
...
-   * @param replace whether to replace the existing index
+   * @param options options for building index
    */
   public void createIndex(
       List<String> columns,
       IndexType indexType,
       Optional<String> name,
       IndexParams params,
-      boolean replace) {
+      IndexOptions options) {
```
jackye1995: I feel this has become quite a long list of things that we might want to just have a `CreateIndexBuilder` which has all the available options including columns, index type, name, params. That seems like the best way to ensure we don't keep changing our public APIs again and again, what do you think?

steFaiz: Thanks for your advise! I think you are right, this interface might be evolving. I've wrapped all params into IndexOptions class with required options and optional options. I think `CreateIndexBuilder` might be more rust style, so I kept the `IndexOptions` class. The original interface is also kept for backwards compatibility.

java/lance-jni/src/blocking_dataset.rs
```diff
@@ -678,11 +681,113 @@ fn inner_create_index(
...
+        index_builder = index_builder.name(name);
+    }
+
+    let has_fragment_ids = fragment_ids.is_some();
+
+    if let Some(fragment_ids) = fragment_ids {
+        index_builder = index_builder.fragments(fragment_ids);
+    }
+
+    if let Some(fragment_uuid) = fragment_uuid {
```
jackye1995: not a problem for you, but more for @chenghao-guo for this `fragment_uuid` thing. I get this is shared by the fragment level index builder, but isn't this just an index UUID? Why do we call it fragment UUID? It is confusing to me.

java/src/main/java/com/lancedb/lance/index/Index.java
```diff
@@ -31,7 +31,7 @@ public class Index {
   private final List<Integer> fields;
   private final String name;
   private final long datasetVersion;
-  private final byte[] fragmentBitmap;
```
jackye1995, line 34: I think we probably want to keep this, since for a created index we would prefer to return the bitmap. Also we should make sure bitmap and list of fragments should not be set at the same time

java/lance-jni/src/blocking_dataset.rs
chenghao-guo: Your understanding is correct. I agree that using `index_uuid` improves clarity and reduces ambiguity. Once this get refracted, I will update that in lance-ray project.

chenghao-guo: Looks like java api already used the index_uuid in the method, I think we can go ahead and refractor it in this PR. I will update lance-ray/daft method parameter to uniform it as well.

java/src/main/java/com/lancedb/lance/index/Index.java
steFaiz: @jackye1995  Thanks for your review! I followed the python api, see [python index](https://lancedb.github.io/lance-python-doc/all-modules.html#lance.Index.base_id). Python Index is also directly converted from (and would be also directly converted to) Rust IndexMetadata struct, see [python conversion](https://github.com/lancedb/lance/blob/main/python/src/transaction.rs#L25). 
Current java api stores serialized rust RoaringBitmap in a byte array, which is absolutely opaque to java user.  A List makes fragments info visible, which might be useful in some situations. For example, in distributed scan scenerios, we could configure more resources for unindexed fragments.
Do you think we need to have both a List and a byte[] in java side ? Or just follow the Python implementation. And we can also introduce RoaringBitmap dependency if you think a List object is too large.

java/lance-jni/src/blocking_dataset.rs
steFaiz: @chenghao-guo ok, thanks for your reply. I'll rename it in this pr.

steFaiz: @jackye1995 I've refactored this. PTAL, thanks!

steFaiz:
@jackye1995  Friendly reminder to review this when convenient!

java/src/main/java/com/lancedb/lance/index/Index.java
jackye1995: Sounds good I think we can have it as a materialized for now, but in long term I think we should introduce roaring bitmap to java sdk but that can be a separated task.

steFaiz:
> looks good to me, sorry I missed reviewing this after the update. Let me know when it is rebased!

@jackye1995 Thanks for your review! I've rebased onto the main branch and all stable tests are passed.

jackye1995 (approved):

PR #4962, opened 2025-10-15
feat(java): support shallow_clone

### LINKED ISSUE #4854: Support shallow_clone in Java

majin1102:
Close #4854

### DIFF
--- a/java/lance-jni/src/ffi.rs
+++ b/java/lance-jni/src/ffi.rs
@@ -395,7 +395,7 @@ impl JNIEnvExt for JNIEnv<'_> {
                 .call_method(&java_object, "get", "()Ljava/lang/Object;", &[])?
                 .l()?;
             let inner_value = self
-                .call_method(&inner_jobj, "longValue", "()I", &[])?
+                .call_method(&inner_jobj, "longValue", "()J", &[])?
                 .j()?;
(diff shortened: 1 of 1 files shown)


majin1102:
Ready for review @jackye1995

java/src/main/java/com/lancedb/lance/Dataset.java
```diff
@@ -1044,4 +1044,71 @@ private native MergeInsertResult nativeMergeInsert(
...
+   * copying them. Only metadata is written at the destination.
+   *
+   * @param targetPath the URI to clone the dataset into
+   * @param version the source version to clone from
+   * @param storageOptions Optional object store options for the destination dataset; empty uses
+   *     default store parameters
+   * @return a new Dataset instance at the target path
+   */
+  public Dataset shallowClone(
+      String targetPath, long version, Optional<Map<String, String>> storageOptions) {
```
jackye1995: can we have a `ShallowCloneOptions` that contains version, tag and storage options? So it's easier to extend other options in the future. I think likely we also need a `branch` option later

majin1102: I'm thinking of adding a Reference for extension. It could also be used in other cases like checkout.
Please take a look.

Issue #4955, opened 2025-10-14
Trying to filter with `< current_date()` gives wrong result


wjones127:
```python
import pyarrow as pa
import lance
from datetime import date

data = pa.table({'date': [date(2020, 1, 1), date(2020, 1, 2)]})
ds = lance.write_dataset(data, "memory://")

ds.to_table(filter="date <= current_date()")
```

```
pyarrow.Table
date: date32[day]
----
date: []
```

I expect both dates to be returned.

PR #4963, opened 2025-10-15
fix: filter with < current_date() should expand with correct time

Xuanwo:
Fix https://github.com/lancedb/lance/issues/4955

`current_date()` will be evaluated as `1970-01-01` during constant folding, so we should expand it to the correct time before running.

---


wjones127:
> `current_date()` will be evaluated as `1970-01-01` during constant folding, so we should expand it to the correct time before running.

🤣 🤦

wjones127 (approved):
Thanks for working on this.

PR #4966, opened 2025-10-15
fix: use correct logic to detect old/new scheme in binary block decoder

westonpace:


Xuanwo (approved):
Thank you, we paid a lot for forward compatibility.

rust/lance-encoding/src/encodings/physical/binary.rs
```diff
@@ -415,7 +415,13 @@ impl BlockDecompressor for BinaryBlockDecompressor {
         // In the standard scheme we use 4 bytes for the bits per offset and 4 bytes for the bytes_start_offset and we
         // rely on the passed in num_values to be correct.
 
-        let (bits_per_offset, bytes_start_offset, offset_start) = if data[0] == 0 {
+        // This isn't perfect but it's probably good enough and the best I think we can do.  The bits per offset will
+        // never be more than 255 and it's little endian so the last 3 bytes will always be 0.  These will be the least
+        // significant 3 bytes of the number of values in the old scheme.  It's pretty unlikely these are all 0 (that would
+        // mean there are at least 16M values in a single page) so we'll use this to determine if the old scheme is used.
```
luohao, line 421: I am a little concerned this may turn into a timed bomb as it can potentially silently corrupt the data if this really happens.

For this particular case I think it's reasonable to either
1. give up on backward compatibility since v2.1 is considered experimental
2. introduce a binary v2 encoding for the new layout

westonpace: > I am a little concerned this may turn into a timed bomb as it can potentially silently corrupt the data if this really happens.

Well, it wouldn't really corrupt the data as this is read path and not the write path.  So we could always cross that bridge when we come to it.  At that point it might be less painful to go with approach 1 since there will hopefully be less data written by these unstable writers.

> give up on backward compatibility since v2.1 is considered experimental

Since we have no way of detecting this case anyways I'm not sure there is a "give up" path that is more clear?  Mostly I think I just have scars from the 2.0 rollout where I made a few cleanups at the end said "that was experimental anyways" and then I find out a bunch of people were relying on those unstable versions (was rather annoying) and I needed to support it anyways.  2.1 has been experimental for a while so I do have some fear.

> introduce a binary v2 encoding for the new layout

This might have been a better idea but now that 2.1 has been declared stable I think it is too late.  If we did this then supposedly stable readers would no longer be able to read 2.1.

Issue #4898, opened 2025-10-03
Better determination of when to apply dictionary encoding


westonpace:
Currently we use a divisor configuration variable.  This is a little confusing.  At the very least we should convert this to a 0-1 parameter like we have for RLE or BSS.  However, maybe we should just look to see if the estimated size of the dictionary will be significantly less than the size of the raw data.  For example, if we have large strings, we might want to apply dictionary encoding even if it only reduces the cardinality by a small amount.

Also, this is anecdotal, but HLL seems to be a pretty wild estimator in my debugging.  It might be simpler to treat HLL as a first-pass and then, if HLL suggests dictionary encoding might be useful, we calculate the actual cardinality (as we are encoding the data) and if the actual cardinality is significantly off from our estimate we abandon the encoding.

yingjianwu98:
I can take a stab at this, thanks!

yingjianwu98:
I have a first PR ready for review that introduces the ratio config: https://github.com/lancedb/lance/pull/4971

I will have two more to come up one related to looking at the size and another on HLL, which I also see has high error rate based on my unit test (eg: I created a block with 1000 values, and targeting 400 cardinality but the compute stats return the cardinality = 527)

yingjianwu98:
Second PR for review that introduces should dictionary encoded based on size:
https://github.com/lancedb/lance/pull/4972/commits/2ed6b634222d182c728b93e6bdae587af9bcb8ed

yingjianwu98:
@westonpace 
A follow up question here:  is it worth to have both?

The main advantage of dictionary encoding is for storage saving so I am thinking size_based should be sufficient?  Let me know what you think, thanks!

westonpace:
I think we can close this out with the size-based heuristic that was added.

PR #4971, opened 2025-10-15
feat: introduce ratio config for dictionary encoding

yingjianwu98:
Solve this https://github.com/lancedb/lance/issues/4898.

More specifically, this introduce a new dictionary cardinality ratio config


yingjianwu98:
closing this for now, as we are going to use size based approach

PR #4972, opened 2025-10-15
feat: should dictionary encode based on size

yingjianwu98:
This PR enables check on dictionaryEncoded based on size. resolve https://github.com/lancedb/lance/issues/4898

**Dictionary encoding splits data into two parts:**
1. Dictionary: stores unique values
2. Indices: maps each value to a dictionary entry

**For FixedWidth:**
1. Dictionary: cardinality × 16 bytes (128 bits per value)
2. Indices: num_values × 4 bytes (32-bit i32)

**For VariableWidth (strings/binary):**
1. Dictionary values: cardinality × avg_value_size (actual data)
2. Dictionary offsets: cardinality × offset_size (32 or 64 bits)
4. Indices: num_values × offset_size (same as dictionary offsets)

### DIFF
--- a/rust/lance-encoding/src/encodings/logical/primitive.rs
+++ b/rust/lance-encoding/src/encodings/logical/primitive.rs
@@ -14,8 +14,7 @@ use std::{
 
 use crate::{
     constants::{
-        DICT_DIVISOR_META_KEY, STRUCTURAL_ENCODING_FULLZIP, STRUCTURAL_ENCODING_META_KEY,
-        STRUCTURAL_ENCODING_MINIBLOCK,
+        STRUCTURAL_ENCODING_FULLZIP, STRUCTURAL_ENCODING_META_KEY, STRUCTURAL_ENCODING_MINIBLOCK,
     },
     data::DictionaryDataBlock,
     encodings::logical::primitive::blob::{BlobDescriptionPageScheduler, BlobPageScheduler},
@@ -65,6 +64,10 @@ use crate::{
 };
 use lance_core::{datatypes::Field, utils::tokio::spawn_cpu, Result};
 
+use crate::constants::DICT_SIZE_RATIO_META_KEY;
+use crate::encodings::logical::primitive::dict::{
+    DICT_FIXED_WIDTH_BITS_PER_VALUE, DICT_INDICES_BITS_PER_VALUE,
+};
 use crate::{
     buffer::LanceBuffer,
     data::{BlockInfo, DataBlockBuilder, FixedWidthDataBlock},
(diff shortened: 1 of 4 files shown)


rust/lance-encoding/src/encodings/logical/primitive.rs
```diff
@@ -4117,31 +4121,150 @@ impl PrimitiveStructuralEncoder {
...
+    /// - Dictionary offsets: cardinality × offset_size (32 or 64 bits)
+    /// - Indices: num_values × offset_size (same as dictionary offsets)
+    fn estimate_dict_size(data_block: &DataBlock) -> Option<u64> {
+        let cardinality = data_block
+            .get_stat(Stat::Cardinality)?
+            .as_primitive::<UInt64Type>()
+            .value(0);
+
+        let num_values = data_block.num_values();
+
```
westonpace, line 4145: Can we skip the match block and just do something like...

```
let data_size = data_block.data_size();
let avg_value_size = data_size / num_values;
let dict_value_size = cardinality * avg_value_size;
Some(dict_values_size + indices_size);
```

This way we don't have to keep adding to this method if we start dictionary encoding other types?

rust/lance-encoding/src/encodings/logical/primitive.rs
```diff
@@ -4117,31 +4121,150 @@ impl PrimitiveStructuralEncoder {
...
 
-        // Somewhat arbitrary threshold rule.  Apply dictionary encoding if the number of unique
-        // values is less than 1/2 the total number of values.
-        let divisor: u64 = field
+        // Get raw data size
+        let data_size = if let Some(data_size_array) = data_block.get_stat(Stat::DataSize) {
+            data_size_array.as_primitive::<UInt64Type>().value(0)
+        } else {
+            return false;
+        };
```
westonpace: I didn't actually remember that we have a stat named `Stat::DataSize`.  Can we just use `data_block.data_size()`?  That is simpler, though maybe it's not exactly the same thing?

rust/lance-encoding/src/encodings/logical/primitive.rs
```diff
@@ -4117,31 +4121,150 @@ impl PrimitiveStructuralEncoder {
...
                 env::var("LANCE_ENCODING_DICT_DIVISOR")
                     .ok()
                     .and_then(|val| val.parse().ok())
-            })
-            .unwrap_or(2);
+            });
+
+        // If both are configured, panic
+        if ratio_optional.is_some() && divisor_optional.is_some() {
+            panic!("Both dict-ratio and dict-divisor specified, and only one should be set.",);
```
westonpace: Maybe let's just pick one to be the primary and then log a warning?

rust/lance-encoding/src/encodings/logical/primitive.rs
```diff
@@ -4158,6 +4281,20 @@ impl PrimitiveStructuralEncoder {
...
+        let too_small = env::var("LANCE_ENCODING_DICT_TOO_SMALL")
+            .ok()
+            .and_then(|val| val.parse().ok())
+            .unwrap_or(100);
+        if data_block.num_values() < too_small {
+            return false;
+        }
+
+        Self::should_dictionary_encode_based_on_cardinality(data_block, field)
+            || Self::should_dictionary_encode_based_on_size(data_block, field)
```
westonpace: I guess my original thinking was that, if we introduced a size-based method, we would always use that.  Is there some reason we still need `should_dictionary_encode_based_on_cardinality`?

westonpace:
Oh, and thanks for taking a look at this :smile:

rust/lance-encoding/src/encodings/logical/primitive.rs
yingjianwu98: ah, initially I thought we might want both and the more I think about it, the more I realize that size based is probably better and should be sufficient.

rust/lance-encoding/src/encodings/logical/primitive.rs
yingjianwu98: Thanks for the suggestion! I looked into this and I don't think we can simplify it that much because the current dictionary encoding has different behaviors for FixedWidth vs VariableWidth:

  Issues with the simplified approach:
  1. FixedWidth dictionaries always use 128-bit values, regardless of the input data's bits_per_value. So using data_size() would give us the wrong size - if we have i32 data (4 bytes), the dictionary still stores
  each unique value as 16 bytes (128 bits).
  2. VariableWidth dictionaries need offsets - the simplified version is missing dict_offsets_size = (cardinality + 1) * (bits_per_offset / 8). Without this, we'd significantly underestimate the dictionary size.
  3. Indices have different sizes: FixedWidth uses 32-bit indices always, but VariableWidth uses matching bits_per_offset (32 or 64 bits).

yingjianwu98:
Also I think if we go with size based approach, it might still be beneficial to abort when the actual compressed size is larger than the data size, with two reason:

1. We are still replying on hll cardinality computation.
2. We currently rely on the heuristic of avg_value_size = data_size / num_values, which could be inaccurate.

yingjianwu98:
@westonpace 

I have updated based on comments and respond to 1 of your comments regarding the estimate_dict_size function.

Additionally, during test, I found a bug in decode_helper for dictionary encoding and have a PR fix for it here: https://github.com/lancedb/lance/pull/4995

Also, I think the reason hll gives really bad estimation is because the current precision is set too low, which is 4 and I have some benchmark stats for different level of precision here if you are interested:
https://github.com/lancedb/lance/pull/4996

I am not a parquet expert and based on my high level understanding of the code, parquet doesn't have an early estimation before starting the actual encoding work but uses similar idea you bring up that aborts the dictionary encoding if it does not find it useful during the computation.

rust/lance-encoding/src/encodings/logical/primitive.rs
westonpace: > FixedWidth dictionaries always use 128-bit values, regardless of the input data's bits_per_value. So using data_size() would give us the wrong size - if we have i32 data (4 bytes), the dictionary still stores
each unique value as 16 bytes (128 bits).

I think what is happening is that we only consider dictionary encoding on fixed-width data if the data is 128 bits for some historical reasons.  Smaller data types were less likely to benefit from dictionary encoding because at the time (and maybe still) we did not bit-pack the dictionary indices.  So converting a 4-byte value to 4-byte dictionary indices could not possibly help.

We should not be promoting (and I'm pretty sure we are not) 4 byte values into 16 byte values.

> VariableWidth dictionaries need offsets - the simplified version is missing dict_offsets_size = (cardinality + 1) * (bits_per_offset / 8). Without this, we'd significantly underestimate the dictionary size.

Agreed, but offsets should be part of `data_block.data_size()` so we should be ok.

> Indices have different sizes: FixedWidth uses 32-bit indices always, but VariableWidth uses matching bits_per_offset (32 or 64 bits).

We shouldn't do this.  There is no reason for dictionary indices to match the bits_per_offset of the data type.  We should just always use 32-bit indices and rely on bitpacking of indices.  We could even just always use 64-bit indices and rely on bitpacking of indices.

To be clear, we don't need to fix this as part of this PR, so let's worry about it later.

westonpace:
> Also, I think the reason hll gives really bad estimation is because the current precision is set too low, which is 4 and I have some benchmark stats for different level of precision here if you are interested:

Those are some very useful stats, thank you!  I think HLL is good as a first-pass filter which helps us identify which columns we should consider for dictionary encoding.  Then we can remove the errors in HLL by aborting.  I've created #5006 to track this.

yingjianwu98:
@westonpace Thanks for the review and I have rebased the PR!

Also thanks for the great detail explanation, and they make sense.

westonpace:
Pretty sure the test failures were unrelated but I've rebased real quick just to make sure.

PR #4973, opened 2025-10-15
chore: fix preview release bump logic and always publish java

jackye1995:
Ensure the preview release also bump version to preview version number instead of staying at the stable release version number.

Example situation today: 0.38.3-beta.1 has version 0.38.2 on all files: https://github.com/lancedb/lance/blob/v0.38.3-beta.1/Cargo.toml#L25

This should make the version bump experience consistent between lance and lancedb

### DIFF
--- a/.github/workflows/java-publish.yml
+++ b/.github/workflows/java-publish.yml
@@ -9,7 +9,7 @@ on:
   workflow_dispatch:
     inputs:
       mode:
-        description: 'Release mode'
+        description: "Release mode"
         required: true
         type: choice
         default: dry_run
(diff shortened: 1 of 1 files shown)


.github/workflows/java-publish.yml
```diff
@@ -1,9 +1,8 @@
 name: Build and publish Java packages
 on:
   release:
-    # Use released instead of published, since we don't publish preview/beta
-    # versions
-    types: [released]
+    # Trigger on published to include both stable and preview/beta releases
+    types: [published]
```
jackye1995, line 5: since Maven Central does not have hard limit, and we bump version also for preview release now, we can publish java all the time

Xuanwo (approved):

PR #4976, opened 2025-10-16
docs: add docs for newly added options

Xuanwo:
This PR will add docs for newly added options about encoding added in 2.1 format.

---

**This PR was primarily authored with Amp using Claude-Sonnet-4.5 and then hand-reviewed by me. I AM responsible for every change made in this PR. I aimed to keep it aligned with our goals, though I may have missed minor issues. Please flag anything that feels off, I'll fix it quickly.**


docs/src/format/file/encoding.md
```diff
@@ -580,8 +625,282 @@ threshold. If the number of unique values is less than the threshold then we app
...
 
-Packed struct encoding is a semi-structural transformation described above.
+Packed struct encoding is a semi-structural transformation described above. When enabled, struct values are stored
+in row-major format rather than the default columnar format. This reduces the number of I/O operations needed for
+random access but prevents reading individual fields independently.
+
+This is always opt-in and should only be used when all struct fields are typically accessed together.
+
+## Usage Examples
+
```
jackye1995: I think in general we should keep the format doc with text only, and have code specific guides in user guide, what do you think?

Xuanwo: Makes sense to me. Will remove.

PR #4977, opened 2025-10-16
fix: the KMeans may result in all zeros centroids

BubbleCal:
when we compute a centroid:
1. we first sum up all vectors belong to it
2. and then divide all elements by the number of vectors
step2 would make all elements be 0s if all the 3 conditions met:
1. there are many vectors used for training kmeans (happens if num_partitions is large)
2. there are only a few partitions (kmeans k param, happens if hierarchical kmeans, which train with k'=16 each iteration)
3. the data type precision is not high enough (happens on f16)

this PR fixes this by sampling data in hierarchical kmeans, up to `k' * 512` vectors


rust/lance-index/src/vector/kmeans.rs
```diff
@@ -611,6 +615,14 @@ impl KMeans {
     where
         T::Native: Num,
     {
+        // we have to limit the number of data to avoid division underflow,
+        // the threshold 512 is chosen because the minimal normal f16 value will be 0 if divided by 1024.
+        let data = if data.len() >= k * 512 {
+            data.slice(0, k * 512)
+        } else {
+            data.clone()
+        };
```
westonpace, line 627: Why is the sample rate greater than 512 here?  In the issue I originally reported it was unmodified.  Is that because we use all the data for the k=16 initial kmeans so the sample rate for that kmeans will be (num_partitions / 16 * user_provided_sample_rate)?

I guess it just seems a little strange to silently ignore data at this point.  If anything maybe we should return an error.  Then we can fix at a higher level.

For example, when we start the k=16 kmeans then instead of using all the data we can slice it to either the sample_rate * 16 or 512 * 16?

Also, if a user supplies a sample rate greater than 512 we can return an error immediately saying "sample_rate too large"?

BubbleCal: The sample_rate is 256 by default 

We sample num_partitions * 256 rows for training IVF, but before this PR, hierarchical kmeans would set k=16 and train over all sampled data, so it's training 16 clusters over num_partitions * 256 vectors

This PR samples the sampled data again if it's too much

BubbleCal: Added a warning log if sample_rate>=1024 and the data type is float16

BubbleCal: Didn't return an error directly because it's hard to say whether it will really get 0s centroid with that large sample rate, 1024 is the min possible value, but it doesn't mean users will always get 0s centroid is sample_rate >= 1024

westonpace (approved):

PR #4980, opened 2025-10-17
fix: populate all fields from ExecutionSummaryCounts to  ScannerStats

jaystarshot:
Python API ScannerStats was missing some fields from ExecutionSummaryCounts, This PR populate them for the scan fallback

`scan_stats_callback (Callable[[ScanStatistics], None], default None) – A callback function that will be called with the scan statistics after the scan is complete. Errors raised by the callback will be logged but not re-raised.`

https://lancedb.github.io/lance-python-doc/all-modules.html#lance.dataset.LanceDataset.scanner

For the added UT i can see

`ScanStatistics(iops=3, requests=3, bytes_read=3649, indices_loaded=0, parts_loaded=0, index_comparisons=0, all_counts={"ranges_scanned": 1, "rows_scanned": 100, "fragments_scanned": 1})`

### DIFF
--- a/python/python/lance/lance/__init__.pyi
+++ b/python/python/lance/lance/__init__.pyi
@@ -527,10 +527,17 @@ class PyFullTextQuery:
     ) -> PyFullTextQuery: ...
 
 class ScanStatistics:
+    """Statistics about a scan operation."""
+
     iops: int
+    requests: int
     bytes_read: int
     indices_loaded: int
     parts_loaded: int
+    index_comparisons: int
+    all_counts: Dict[
+        str, int
+    ]  # Additional metrics for debugging purposes. Subject to change.
 
 __version__: str
 language_model_home: Callable[[], str]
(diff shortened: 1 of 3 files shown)


jaystarshot:
I want to add file level or bucket level IO stats here somehow in future (guarded by an env variable) but i guess its a bit tricky

westonpace:
> I want to add file level or bucket level IO stats here somehow in future (guarded by an env variable) but i guess its a bit tricky

Hmm, I think we could eventually do this.  Why the env variable?  Is that because the stats might be quite large if there are many files?

jaystarshot:
Thank you! Yeah, does lance have a in built metric system like prometheus client etc? Maybe that could be useful in that case

westonpace:
> Thank you! Yeah, does lance have a in built metric system like prometheus client etc? Maybe that could be useful in that case

Lance metrics are emitted to rust's `tracing` library.  By default, these are just dumped into the void.  However, you can configure the "tracing subscriber" to capture these and do something with them.  For example, this library (https://docs.rs/telemetry-subscribers/latest/telemetry_subscribers/) makes it possible to convert span latencies into prometheus metrics.  I think you can also forward the tracing events to open telemetry collectors and then calculate the metrics at that point.

For the non-rust clients (e.g. java / python) we have to code up integrations.  For example, pylance has `capture_trace_events` which allows for a rust-to-python conversion of tracing events.

jaystarshot:
Thanks a lot! checking

PR #4981, opened 2025-10-17
feat: support mTLS in REST namespace

jackye1995:
Expected use pattern:

```rust

let mut properties = HashMap::new();
properties.insert("uri".to_string(), "https://api.example.com".to_string());
properties.insert("tls.cert_file".to_string(), "/path/to/client-cert.pem".to_string());
properties.insert("tls.key_file".to_string(), "/path/to/client-key.pem".to_string());
properties.insert("tls.ssl_ca_cert".to_string(), "/path/to/ca-cert.pem".to_string());
properties.insert("tls.assert_hostname".to_string(), "true".to_string());

let namespace = RestNamespace::new(properties);
```


Xuanwo (approved):
I'm ok with this after all CI passed.

PR #4983, opened 2025-10-17
fix: correct read column ordering in Fragment::update_columns

wayneli-vt:
This pr fixes bugs in `Fragment::update_columns` that could occur when user specifies join keys:
1. **Column Not Found**: When `left_on` and `right_on` had different names, the code would incorrectly try to read the `right_on` column from the `left_fragment`, which may does not exist, causing a runtime error.
2. **Column Misalignment**: When the join keys had the same name, an incorrect projection order could lead to column misalignment during the update.
For example, with a `right_stream` of `(key, A, B)`, the buggy logic caused `left_fragment` would also read `(key, A, B)`, and the `HashJoiner` would incorrectly map the `right_stream`'s column `A` to the `left_fragment`'s `key` column.

This fix addresses these issues by constructing a correct read order. It reads the columns from `write_schema` first and then appends the `left_on` join key at the end. This ensures correct read order to prevent mapping errors.


wayneli-vt:
@wjones127 @westonpace @jackye1995
Hi, this pr is ready for review! Please take a look when you have a moment. Thanks!

wayneli-vt:
@wjones127 @westonpace @jackye1995

Just a gentle ping on this PR. All checks have passed, and it's ready for review when you have a moment! Thanks a lot!

Issue #4772, opened 2025-09-19
Should we provide a query interface from rowId to rowAddr to the non-Rust side?


steFaiz:
The stable rowId is a crucial feature of Lance, giving it usable primary key functionality. Currently, the rowIdIndex is a completely internal data structure that plays a vital role in various internal operations such as update, take, and indexed queries. However, these operations are at the Dataset level and cannot be directly utilized by big data engines like Flink, Spark, or Ray.

For example, in the column addition scenario, Lance needs to align new data with existing rows based on rowId, as illustrated below:

<img width="1552" height="830" alt="Image" src="https://github.com/user-attachments/assets/3a4227ba-953b-447a-9658-303b1daec647" />

 In distributed computing, we need to horizontally partition the dataset for computation by rows (i.e., Fragments), which requires shuffling the data.

<img width="1094" height="712" alt="Image" src="https://github.com/user-attachments/assets/301e8705-8030-4fb6-8186-9a8fd8ed30fd" />

Therefore, we need the ability to convert rowId to rowAddress at the computing engine level, so that we can further obtain the FragmentId. Currently, there is no ready-to-use interface for this purpose. As a workaround, I have to SELECT all `<RowId, RowAddress>` pairs from the source table and perform a join with the new data to obtain the corresponding RowAddresses. However, this join incurs significant overhead, and even if the column addition involves only a small subset of rows, it still requires scanning the entire source table, which is highly inefficient.

Therefore, I'm wondering if we could expose an interface in the Dataset to query the mapping from rowId to rowAddr:
*** 
    impl Dataset {
      pub async fn get(&self, row_id: u64) -> Option<RowAddress> {
          ...
      }
      // get batch to reduce jni cost
      pub async fn getBatch(&self, row_ids: &[u64]) -> &[RowAddress] {
          ...
      }
    }
***

Or alternatively, implement a delegated row ID index at the Java/Python FFI (Rust) boundary to enable efficient rowId-to-rowAddr lookups from non-Rust environments.
***
    Struct LanceRowIdIndex {
        inner: RowIdIndex
    }
  
    impl LanceRowIdIndex {
        pub async fn open(&self, dataset: &Dataset) -> LanceRowIdIndex {
            ...
        }
      
        pub async fn get(&self, row_id: u64) -> Option<RowAddress> {
            ...
        }
    
        // get batch to reduce jni cost
        pub async fn getBatch(&self, row_ids: &[u64]) -> &[RowAddress] {
             ...
        }
    }
[...]

PR #4985, opened 2025-10-17
refactor: add shortcut for querying row address by row id in take op

steFaiz:
This PR mainly includes:
1. add shortcut for rowid -> rowaddr conversion. see [comment](https://github.com/lancedb/lance/issues/4772#issuecomment-3311848472)
2. remove redundant with_row_address option of TakeBuilder after [issue](https://github.com/lancedb/lance/pull/4794)


rust/lance/src/dataset/take.rs
```diff
@@ -131,6 +129,36 @@ async fn do_take_rows(
 ) -> Result<RecordBatch> {
     let with_row_id_in_projection = projection.physical_projection.with_row_id;
     let with_row_addr_in_projection = projection.physical_projection.with_row_addr;
+    let physical_schema = Arc::new(projection.physical_projection.to_bare_schema());
+
+    // short cut: if projection do not contain any physical column, directly returns id & addr columns
+    if physical_schema.fields.is_empty()
+        && builder.row_ids.is_some()
+        && (with_row_id_in_projection || with_row_addr_in_projection)
```
westonpace, line 144: Can we test what happens in these cases?

```
        let projection = ProjectionRequest::from_columns(["_rowid"], dataset.schema());
        let projection = ProjectionRequest::from_columns(["_rowaddr"], dataset.schema());
```

The first one is admittedly, a little silly, the user is asking for _rowid and passing in _rowid but users do silly things sometimes :shrug:.  I am worried we might be returning `_rowaddr` even though the user didn't ask for it.

In the second one maybe the user is just trying to convert one set of `_rowid` values into a set of `_rowaddr` values and doesn't care about the ordering / mapping.  Again, I am worried we might be returning `_rowid` even though the user didn't ask for it.

rust/lance/src/dataset/take.rs
```diff
@@ -449,16 +466,9 @@ impl TakeBuilder {
...
             dataset,
-            with_row_address: false,
         })
     }
 
-    /// Adds row addresses to the output
-    pub fn with_row_address(mut self, with_row_address: bool) -> Self {
-        self.with_row_address = with_row_address;
-        self
-    }
```
westonpace, line 521: This is part of the public API.  I don't think we can remove this without breaking backwards compatibility.  Do you want to mark it deprecated instead?

In `do_take_rows` we can add something like...

```
async fn do_take_rows(
    mut builder: TakeBuilder,
    mut projection: Arc<ProjectionPlan>,
) -> Result<RecordBatch> {
    if self.with_row_address {
        let mut proj_with_row_addr = projection.as_ref().clone();
        proj_with_row_addr.include_row_address();
        projection = Arc::new(proj_with_row_addr);
    }
    let with_row_id_in_projection = projection.physical_projection.with_row_id;
    ...
```

This way we can still get rid of the old `with_row_address` paths.

westonpace (changes requested):
One request for backwards compatibility purposes but this is cool, thanks for looking into it

rust/lance/src/dataset/take.rs
steFaiz: I have added the test cases for only returning _rowid or _rowaddr column. Actually my original code applies the projection through `projection.project_batch(batch)`. But during the test, I found a weird bug that if the projection only contains `_rowid` column, then it's ok. However, if the projection only contains `_rowaddr` column, there would be some unexpected panic:
```
IO { source: Internal("Input field name _rowid does not match with the projection expression _rowaddr"), 
location: Location { file: "/Users/XXX/lance/rust/lance-datafusion/src/projection.rs", line: 365, column: 35 } }
```
Note that projection with <'_rowaddr', '_rowid'> (disordered with original batch) also won't cause any panic.
This panic originates from datafusion lib:
```
    pub async fn project_batch(&self, batch: RecordBatch) -> Result<RecordBatch> {
        ...
        // panic here
        let projection = Arc::new(ProjectionExec::try_new(physical_exprs, src)?);
        ...
    }
```
I'm not familiar about this lib, so I directly use the `batch.project()` method, which is also more lightweight.

rust/lance/src/dataset/take.rs
steFaiz: Thanks for your review! I've rolled back this part and also added some test cases for backwards compatibility.

steFaiz:
@westonpace PTAL if you have some time! Thanks!

Issue #4957, opened 2025-10-15
lance FTS returns incomplete suboptimal results


artemru:
First of all, thanks for building a new generation binary format!

I've started to use FTS feature from lance and found that in many cases it's not able to retrieve the optimal results for simple text matching query.
Below I provided a synthetic data reproducer for 2 issues I discovered (my real case dataset is much more complicated and there're potentially more issues). 

1. When query contains several words, lance FTS seems to miss the result for the one with the highest score. It appears only in a combination with a prefilter=True, when prefilter=False and a large limit, the query returns a correct result

<img width="394" height="427" alt="Image" src="https://github.com/user-attachments/assets/9b535efb-0bbc-4bd9-9d51-1bee79d1c71c" />

2. For some text from a dataset, lance FTS is incapable to find an optimal result (the exact match in that case) with a small limit (no filter involved here). The results becomes correct depending of limit value (it's always correct starting from some large value but it can be correct or incorrect for small values).

<img width="749" height="822" alt="Image" src="https://github.com/user-attachments/assets/d73972f9-bc18-4169-b200-b07774c551ca" />

## A reproducer script 
(lance version: '0.38.0'):
```python

import random

import lance
import numpy as np
import polars as pl
import pyarrow as pa
from lance.query import MatchQuery


# ============================================================================
# Data Generation
# ============================================================================

def generate_random_word(min_length: int = 3, max_length: int = 10) -> str:
    """Generate pronounceable random word using consonant-vowel alternation."""
    vowels = "aeiou"
    consonants = "bcdfghjklmnpqrstvwxyz"

    length = random.randint(min_length, max_length)
    use_consonant = random.choice([True, False])

    word = []
    for _ in range(length):
        word.append(random.choice(consonants if use_consonant else vowels))
        use_consonant = not use_consonant

    return "".join(word)


def generate_vocabulary(vocab_size: int, seed: int = 42) -> list[str]:
    """Generate unique vocabulary of random words."""
    random.seed(seed)
    vocab = set()
    while len(vocab) < vocab_size:
        vocab.add(generate_random_word())
    return sorted(list(vocab))


def create_zipf_distribution(vocab_size: int, alpha: float = 1.5) -> np.ndarray:
    """Create Zipf (power-law) distribution for word frequencies."""
    ranks = np.arange(1, vocab_size + 1)
    probabilities = 1.0 / np.power(ranks, alpha)
    return probabilities / probabilities.sum()


def generate_sentence(
    vocab: list[str],
[...]

PR #4986, opened 2025-10-17
fix: full text search may miss some results

BubbleCal:
fix https://github.com/lancedb/lance/issues/4957

there are 2 issues:
- flat search may miss some results because it advances posting iterators more than expected
- we don't multiply the (K1+1.0) for block_max_score, then the score is underestimated

### DIFF
--- a/python/python/tests/test_scalar_index.py
+++ b/python/python/tests/test_scalar_index.py
@@ -3650,3 +3650,43 @@ def test_fts_flat_fallback_matches_wand(tmp_path):
         f"Limited WAND scores differ from full WAND scores.\n"
         f"limited scores={limited_wand_scores}\nfull scores={full_wand_scores}"
     )
+
+
+def test_scan_statistics_callback(tmp_path):
+    """Test that scan_stats_callback receives all expected fields."""
+    # Create a simple dataset
+    table = pa.table(
+        {
+            "id": range(100),
+            "value": np.random.randn(100),
+        }
+    )
+
+    dataset = lance.write_dataset(table, tmp_path / "test_stats.lance")
+
+    scan_stats = None
+
+    def scan_stats_callback(stats: lance.ScanStatistics):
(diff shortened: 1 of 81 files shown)


artemru:
Thanks for the fix !
do I get it right that we dont need to recompute FST index after that PR since two bugs were only in the retrieval phase ?

python/python/tests/test_scalar_index.py
```diff
@@ -3503,3 +3503,150 @@ def test_btree_query_comparison_parametrized(
...
+    tbl = pa.table(
+        {
+            "id": np.arange(n, dtype=np.int64),
+            "category": pa.array(categories.astype(str)),
+            "text": pa.array(texts, type=pa.large_string()),
+        }
+    )
+
+    ds_path = tmp_path / "fts_fallback.lance"
+    ds = lance.write_dataset(tbl, ds_path)
```
artemru, line 3565: I wonder if it would be useful to consider a case of multiple fragments dataset as well in that test ?

BubbleCal:
> Thanks for the fix ! do I get it right that we dont need to recompute FST index after that PR since two bugs were only in the retrieval phase ?

yes, this requires to rebuild the index

python/python/tests/test_scalar_index.py
BubbleCal: that won't impact how the index works

artemru:
dummy question : what is the release frequence ? (just want to get that fix into a wheel)

BubbleCal:
> dummy question : what is the release frequence ? (just want to get that fix into a wheel)

Lance would make a stable release per 2 weeks, we just made one last week so will make a new one next week.
preview release is made frequently, you can use [v0.38.3-beta.5](https://github.com/lancedb/lance/releases/tag/v0.38.3-beta.5), which includes this fix

PR #4988, opened 2025-10-17
feat: log version on dataset load event

wjones127:
For debugging it's critical to know what version you are operating with. This adds that information to the loading trace event.

### DIFF
--- a/rust/lance/src/dataset/builder.rs
+++ b/rust/lance/src/dataset/builder.rs
@@ -10,6 +10,7 @@ use crate::{
     session::Session,
     Dataset,
 };
+use futures::FutureExt;
 use lance_core::utils::tracing::{DATASET_LOADING_EVENT, TRACE_DATASET_EVENTS};
 use lance_file::datatypes::populate_schema_dictionary;
 use lance_file::v2::reader::FileReaderOptions;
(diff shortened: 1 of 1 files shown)


rust/lance/src/dataset/builder.rs
```diff
@@ -298,8 +299,22 @@ impl DatasetBuilder {
...
 
     #[instrument(skip_all)]
-    pub async fn load(mut self) -> Result<Dataset> {
-        info!(target: TRACE_DATASET_EVENTS, event=DATASET_LOADING_EVENT, uri=self.table_uri);
+    pub async fn load(self) -> Result<Dataset> {
+        let uri = self.table_uri.clone();
+        let target_ref = self.version.clone();
+        match self.load_impl().boxed().await {
+            Ok(dataset) => {
+                info!(target: TRACE_DATASET_EVENTS, event=DATASET_LOADING_EVENT, uri=uri, target_ref = ?target_ref, version=dataset.manifest.version, status="success");
```
LuQQiu, line 307: Would it be better to be debug?? looks like this may generate a lot of logs?

wjones127: The existing event is `info`. I don't want to break existing users that rely on this.

Issue #3864, opened 2025-05-22
What things should be in the lance public API?


westonpace:
Clearly defining a public API for the `lance` crate can make it easier for us to have stable releases because we can know what to count as a breaking change.  It will also make lance easier to use as I think it is awkward at the moment that users need to add imports to `lance_io`, `lance_encoding`, etc.

In my head, I think the way that the `lance` library should work, is that everything that can be done with a `Dataset` can be done with `lance` imports alone.

Currently, this means we would need to re-export a few things:

* There are a lot of types we'd need to re-export to support training and creating indices.  Should we re-export these or should `lance_index` be considered a "top-level" package that adds capabilities onto `lance`?
* I think we should probably re-export arrow_array and arrow_schema as we use those types throughout our public API
* Should we re-export datafusion?  There are a few public APIs that create plans or consume datafusion expressions that could benefit from this..
* Should we re-export `object_store`?
* I think we should examine dataset's API and see if there is anywhere we are depending on minor crates like `lance_table`, `lance_io`, `lance_core`, `lance_encoding`, `lance_linalg`, etc.  We should re-export rather than consider these top-level crates

HaochengLIU:
```
lance
 ├── lance-arrow
 ├── lance-core
 ├── lance-datafusion
 ├── lance-encoding
 ├── lance-file
 ├── lance-io
 ├── lance-linalg
 ├── lance-index
 └── lance-table
...
lance-index
 ├── lance-arrow
 ├── lance-core
 ├── lance-datafusion
 ├── lance-encoding
 ├── lance-file
 ├── lance-io
 ├── lance-linalg
 └── lance-table
```
> - There are a lot of types we'd need to re-export to support training and creating indices. Should we re-export these or should lance_index be considered a "top-level" package that adds capabilities onto lance?

The dependencies of `Lance-index` is a subset of `lance` and it's a direct dependency. I dont have enough context here to decide the best solution here. If we do want to support file formats other than lance in the near future, then  `index` should be its own top crate?

PR #4991, opened 2025-10-17
refactor!: cleanup public API, remove lance_arrow re-exports

westonpace:
I'd like to start addressing https://github.com/lancedb/lance/issues/3864 so that we can reduce the amount of breaking changes we introduce by minimizing the public API surface that we commit to.  Regrettably, doing so involves creating many breaking changes :face_exhaling: as we have too many exports today as well as duplicate exports.

This PR is the first in a series of PRs to cleanup the public API.  All of these changes are breaking changes.

* Reduces the scope of lance_arrow re-exports to just the bfloat16 array.  The rest of the lance_arrow consists of convenience utilities, which are handy, but not something we want to commit to supporting indefinitely, and things like FloatType, which we might re-export as part of lance-index, but is not generally applicable for lance users.
* Removes the re-export of lance_core::error.  We already re-export `Error` and `Result` and the rest of the stuff in there is not meant for public consumption
* ~~Removes the `crate::arrow::json` utilities.  These converted data types and schemas to/from JSON.  This is neat, but only used in lancedb today.  It should be included there, or put in its own crate.  We don't want to commit to maintaining this.~~
* Change `crate::session::index_extension` to be `pub(crate)`
* Removes lance_datafusion::sql re-export.  This was a convenience export from an earlier refactor that is no longer needed.
* Removes lance_index::vector::kmeans for the same reason

### DIFF
--- a/python/pyproject.toml
+++ b/python/pyproject.toml
@@ -60,7 +60,7 @@ tests = [
     # Only test tensorflow on linux for now. We will deprecate tensorflow soon.
     "tensorflow; sys_platform == 'linux'",
     "tqdm",
-    "datafusion==50",
+    "datafusion>=50.1",
 ]
 dev = ["ruff==0.4.1", "pyright"]
 benchmarks = ["pytest-benchmark"]


westonpace:
I could not remove the JSON serialization routines.  It appears there are public python bindings for these and have been for some time.  We consider the python API to be somewhat stable and/or public so I guess we are committed to maintaining these for now at least.  We could maybe drop them in a deprecation cycle in the future.  However, this serialization format is also used by lancedb in the REST API so that may be tricky :frowning:

PR #4993, opened 2025-10-18
feat: add reserved field IDs for metadata columns

jackye1995:
After https://github.com/lancedb/lance/pull/4794, a schema can now include _rowid and _rowaddr fields with id=-1, this is unexpected for processes that tries to get the field IDs of the schema. Also different metadata columns share the same field ID which is prune to future errors.

This PR reserves field IDs to these metadata columns. This is similar to how Iceberg does it as a reference: https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/MetadataColumns.java

### DIFF
--- a/rust/lance-core/src/datatypes/schema.rs
+++ b/rust/lance-core/src/datatypes/schema.rs
@@ -497,11 +497,26 @@ impl Schema {
     }
 
     // Recursively collect all the field IDs, in pre-order traversal order.
-    // Only includes fields with non-negative IDs (i.e., actual data fields).
-    // Metadata fields like _rowid and _rowaddr have id=-1 and are excluded.
+    // Includes metadata fields with their dedicated IDs (i32::MAX, i32::MAX-1, etc.)
     // TODO: pub(crate)
     pub fn field_ids(&self) -> Vec<i32> {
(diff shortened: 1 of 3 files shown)


rust/lance-core/src/datatypes/schema.rs
```diff
@@ -497,11 +497,28 @@ impl Schema {
...
     }
 
+    /// Check if a field ID corresponds to a metadata field.
+    ///
+    /// Metadata fields have dedicated field IDs in the range i32::MAX down to i32::MAX-4:
+    /// - ROW_ID_FIELD_ID (i32::MAX)
+    /// - ROW_ADDR_FIELD_ID (i32::MAX - 1)
+    /// - ROW_OFFSET_FIELD_ID (i32::MAX - 2)
+    /// - ROW_LAST_UPDATED_AT_VERSION_FIELD_ID (i32::MAX - 3)
+    /// - ROW_CREATED_AT_VERSION_FIELD_ID (i32::MAX - 4)
```
Xuanwo, line 513: Can we reserve the lowest dedicated field IDs like `i32::MAX - 1024` so we don’t have to check them one by one?

Xuanwo (commented):
Other LGTM

Issue #3634, opened 2025-04-02
Merge_insert schema error message confusing when inserting subschemas


wjones127:
This logic here does not provide helpful error message:

https://github.com/lancedb/lance/blob/e0d3179bcc6e6ce5f9c5fcfeb4398789a7005467/rust/lance/src/dataset/write/merge_insert.rs#L387-L397

When checking for subschema, it tries to detect it, but if there is a mismatch, it returns the error message assuming it is the full schema.

Instead, we should be returning an error message from a check for subschema.

PR #4994, opened 2025-10-18
fix: improve schema validation for nullability and subschemas

### LINKED ISSUE #4518: Be more permissive with nullability in merge_insert
If a dataset has non-nullable columns, and we pass data with nullable fields but no actual nulls, we *should* be able to write without error. We support this in `insert()`, but it doesn't not work with `merge_insert()` right now.

```python
import pyarrow as pa
import lance

schema = pa.schema(
    [
        pa.field("id", pa.int64(), nullable=False),
        pa.field("value", pa.int64(), nullable=False),
    ]
)
data = pa.table({"id": [1, 2, 3], "value": [0, 0, 0]}, schema=schema)

# Nullable, but no actual nulls
new_schema = pa.schema(
    [
        pa.field("id", pa.int64(), nullable=True),
        pa.field("value", pa.int64(), nullable=True),
    ]
)
new_data = pa.table({"id": [4, 5, 6], "value": [10, 20, 30]}, schema=new_schema)

ds.insert(new_data)  # Works

# Fails:
(
    ds.merge_insert("id")
    .when_matched_update_all()
    .execute(new_data)
)

```

fenfeng9:
Closes #4518 and #3634

- **More Permissive Nullability:** `check_compatible` now ignores nullability differences. This fixes panics when merging a source with nullable fields into a non-nullable target, as long as no actual nulls are present. The behavior now correctly matches `insert()`. 

- **Better Error Messages:** Subschema validation is now integrated directly into `check_compatible`. This provides accurate error messages for type mismatches in subschemas, instead of misleading "missing fields" errors.


fenfeng9:
Hi @wjones127, I've submitted this PR to address "Be more permissive with nullability in merge_insert" (#4518) and "Merge_insert schema error message confusing when inserting subschemas"(#3634).
Would you be able to take a look when you have a moment? Looking forward to your feedback!

PR #4997, opened 2025-10-19
fix: do not modify Lance schema when projecting system columns

jackye1995:
Because `Schema::project_by_schema` is public API, adding system columns to it and use it in projection in #4794 was breaking some use cases.

### DIFF
--- a/rust/lance-datafusion/src/projection.rs
+++ b/rust/lance-datafusion/src/projection.rs
@@ -175,33 +175,59 @@ impl ProjectionPlan {
     ///
     /// This is something that cannot be done easily using expressions.
     pub fn from_schema(base: Arc<dyn Projectable>, projection: &Schema) -> Result<Self> {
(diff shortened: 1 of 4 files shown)


Xuanwo (approved):
Thank you for catching up.

rust/lance-datafusion/src/projection.rs
```diff
@@ -175,33 +175,59 @@ impl ProjectionPlan {
...
+                } else if field.name == ROW_ADDR {
+                    with_row_addr = true;
                     must_add_row_offset = true;
-                    physical_projection.with_row_id = true;
                 }
-                OutputColumn {
-                    expr: Expr::Column(Column::from_name(&f.name)),
-                    name: f.name.clone(),
+                // Note: Other system columns like _rowoffset are computed differently
+                // and shouldn't appear in the schema at this point
```
westonpace, line 197: Why wouldn't they?

PR #5002, opened 2025-10-20
feat: support tracking deleted rows between versions

yanghua:


jackye1995:
Just a thought, not sure if it's better or worse, what if we just reuse the last_updated_at column for deletes? So if the row is deleted, last_updated_at means the version the row was deleted at, and when the row is not deleted it means the version the row was updated at. That could save us from adding yet another metadata column, but I don't know if that would make things confusing. What do you think?

yanghua:
> So if the row is deleted, last_updated_at means the version the row was deleted at, and when the row is not deleted it means the version the row was updated at.

We can do this. It's a performance thinking. But I prefer to use `deleted_at_version`. It provides better readability.

I have an idea, but still have no time to try it.

We may support getting:

* deletion diff via `ds.sql(select * from dataset where _row_deleted_at_version > {} AND _row_deleted_at_version <= {})`
* updation diff via `ds.sql(select * from dataset where _row_created_at_version <= {} AND _row_last_updated_at_version > {} AND _row_last_updated_at_version <= {})`

It would be clearer to do this via semantic independence metadata fields.

About your concern, this feature(and metadata) only works on enabling stable row ID feature. We still have no plan to force switch to stable row ID mode and give up unstable, right? So I think a new metadata cost could be accepted?

PR #5003, opened 2025-10-20
feat: add variable packed struct support

Xuanwo:
Close https://github.com/lancedb/lance/issues/2862

This PR will add variable packed struct support in file version 2.2

---

### DIFF
--- a/rust/lance-encoding/src/encodings/logical/primitive.rs
+++ b/rust/lance-encoding/src/encodings/logical/primitive.rs
@@ -14,8 +14,7 @@ use std::{
 
 use crate::{
     constants::{
-        DICT_DIVISOR_META_KEY, STRUCTURAL_ENCODING_FULLZIP, STRUCTURAL_ENCODING_META_KEY,
-        STRUCTURAL_ENCODING_MINIBLOCK,
+        STRUCTURAL_ENCODING_FULLZIP, STRUCTURAL_ENCODING_META_KEY, STRUCTURAL_ENCODING_MINIBLOCK,
     },
     data::DictionaryDataBlock,
     encodings::logical::primitive::blob::{BlobDescriptionPageScheduler, BlobPageScheduler},
@@ -65,6 +64,10 @@ use crate::{
 };
 use lance_core::{datatypes::Field, utils::tokio::spawn_cpu, Result};
 
+use crate::constants::DICT_SIZE_RATIO_META_KEY;
+use crate::encodings::logical::primitive::dict::{
+    DICT_FIXED_WIDTH_BITS_PER_VALUE, DICT_INDICES_BITS_PER_VALUE,
+};
 use crate::{
     buffer::LanceBuffer,
     data::{BlockInfo, DataBlockBuilder, FixedWidthDataBlock},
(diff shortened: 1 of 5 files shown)


Xuanwo:
@codex review

rust/lance-encoding/src/compression.rs
```diff
@@ -468,6 +480,35 @@ impl CompressionStrategy for DefaultCompressionStrategy {
...
+                    }
+                    Ok(Box::new(PackedStructVariablePerValueEncoder::new(
+                        self.clone(),
+                        field.children.clone(),
+                    )))
+                } else {
+                    Err(Error::invalid_input(
+                        "Packed struct per-value compression should not be used for fixed-width-only structs",
+                        location!(),
+                    ))
```
westonpace, line 505: This can be in a follow-up but we probably will want some kind of per-value solution for fixed-width data types.  The current mini-block solution is somewhat limited and not great if there are many fields.

rust/lance-encoding/src/compression.rs
```diff
@@ -781,6 +822,52 @@ impl DecompressionStrategy for DefaultDecompressionStrategy {
                     general.compression.as_ref().expect_ok()?.scheme(),
                 )?))
             }
+            Compression::VariablePackedStruct(description) => {
```
westonpace, line 821: Why do we need a block compressor?

rust/lance-encoding/src/encodings/logical/primitive.rs
```diff
@@ -4220,15 +4220,31 @@ impl PrimitiveStructuralEncoder {
...
-                    .any(|child| !matches!(child, DataBlock::FixedWidth(_)))
-                {
-                    panic!("packed struct encoding currently only supports fixed-width fields.")
-                }
+            let requires_full_zip_packed_struct =
+                if let DataBlock::Struct(ref struct_data_block) = data_block {
+                    struct_data_block
+                        .children
+                        .iter()
+                        .any(|child| !matches!(child, DataBlock::FixedWidth(_)))
```
westonpace: We do this "if any match" statement in a few spots.  We could maybe add a helper method to StructDataBlock to see if it has any variable-width children.

rust/lance-encoding/src/encodings/logical/primitive.rs
```diff
@@ -4220,15 +4220,31 @@ impl PrimitiveStructuralEncoder {
...
-            // if the `data_block` is a `StructDataBlock`, then this is a struct with packed struct encoding.
-            if let DataBlock::Struct(ref struct_data_block) = data_block {
-                if struct_data_block
-                    .children
-                    .iter()
-                    .any(|child| !matches!(child, DataBlock::FixedWidth(_)))
-                {
-                    panic!("packed struct encoding currently only supports fixed-width fields.")
-                }
+            let requires_full_zip_packed_struct =
```
westonpace, line 4302: Super minor nit but instead of a check here I wonder if we should adjust the checks in `prefers_miniblock` and `prefers_fullzip`.  This way there are fewer "special cases" in this method.  Might be easiest to just wait until we have more special cases and it gets complicated though.

rust/lance-encoding/src/encodings/physical/packed.rs
```diff
@@ -165,3 +176,904 @@ impl MiniBlockDecompressor for PackedStructFixedWidthMiniBlockDecompressor {
...
+    fn append_row_bytes(&self, row_idx: usize, output: &mut Vec<u8>) -> Result<()> {
+        match self {
+            Self::Fixed { block } => {
+                let bits_per_value = block.bits_per_value;
+                if bits_per_value % 8 != 0 {
+                    return Err(Error::invalid_input(
+                        "Packed struct variable encoding requires byte-aligned fixed-width children",
+                        location!(),
+                    ));
+                }
```
westonpace, line 201: We will need a follow up task for packed structs with boolean fields

rust/lance-encoding/src/encodings/physical/packed.rs
```diff
@@ -165,3 +176,904 @@ impl MiniBlockDecompressor for PackedStructFixedWidthMiniBlockDecompressor {
...
+                let prefix_bytes = (*bits_per_length / 8) as usize;
+                if !(prefix_bytes == 4 || prefix_bytes == 8) {
+                    return Err(Error::invalid_input(
+                        "Packed struct variable children must use 32 or 64-bit length prefixes",
+                        location!(),
+                    ));
+                }
```
westonpace, line 233: Lets make a follow-up task to investigate variable-width integer encodings for length prefixes

rust/lance-encoding/src/encodings/physical/packed.rs
```diff
@@ -165,3 +176,904 @@ impl MiniBlockDecompressor for PackedStructFixedWidthMiniBlockDecompressor {
...
+                    ));
+                    field_data.push(VariablePackedFieldData::Variable {
+                        block,
+                        bits_per_length,
+                    });
+                }
+            }
+        }
+
+        let mut row_buffers: Vec<Vec<u8>> = Vec::with_capacity(num_values as usize);
```
westonpace: Let's make a follow-up to add a benchmark for this encoding.  I suspect there are two things that will make this compression a little slower than idea.  First, we are doing a lot of allocation to use `Vec<Vec<u8>>`.  It would probably be ideal to build a single buffer.

rust/lance-encoding/src/encodings/physical/packed.rs
```diff
@@ -165,3 +176,904 @@ impl MiniBlockDecompressor for PackedStructFixedWidthMiniBlockDecompressor {
...
+            max_row_len = max_row_len.max(row_bytes.len());
+            total_bytes = total_bytes.checked_add(row_bytes.len()).ok_or_else(|| {
+                Error::invalid_input("Packed struct row data size overflow", location!())
+            })?;
+            row_buffers.push(row_bytes);
+        }
+
+        let use_u32_offsets = total_bytes <= u32::MAX as usize && max_row_len <= u32::MAX as usize;
+        let bits_per_offset = if use_u32_offsets { 32 } else { 64 };
+        let mut row_builder = DataBlockBuilder::with_capacity_estimate(total_bytes as u64);
```
westonpace: Second performance concern is that `DataBlockBuilder` is maybe overkill here. We can just build the buffers directly.  This would save a lot of virtual calls, not sure how much of a perf diff it will make

westonpace (approved):
This is great!  This is exactly what I had in mind for per-value packed struct :+1:

rust/lance-encoding/src/compression.rs
Xuanwo: Yep, I have the same feeling, created https://github.com/lancedb/lance/issues/5021 to follow.

rust/lance-encoding/src/encodings/logical/primitive.rs
Xuanwo: Good idea, will add it.

rust/lance-encoding/src/encodings/logical/primitive.rs
Xuanwo: I can take a look and give this idea a try.

rust/lance-encoding/src/encodings/physical/packed.rs
Xuanwo: tracked in https://github.com/lancedb/lance/issues/5022

rust/lance-encoding/src/encodings/physical/packed.rs
Xuanwo: Nice catch. Didn't take a deep thinking here, will take a look.

rust/lance-encoding/src/compression.rs
Xuanwo: Sorry, I didn’t understand this comment. What does “block compressor” refer to here?

rust/lance-encoding/src/encodings/logical/primitive.rs
Xuanwo: Added: `has_variable_width_child`

rust/lance-encoding/src/encodings/physical/packed.rs
Xuanwo: Tracked in https://github.com/lancedb/lance/issues/5023

Xuanwo:
Test failure fixed by https://github.com/lancedb/lance/pull/5025, will continue after #5025 been merged.

rust/lance-encoding/src/compression.rs
westonpace: Nevermind, poor reading comprehension on my part.  For some reason I thought this diff was in the `create_block_decompressor` method below.

rust/lance-encoding/src/encodings/physical/packed.rs
Xuanwo: Should be fixed in latest commit.

rust/lance-encoding/src/encodings/physical/packed.rs
Xuanwo: Fixed at the same time

rust/lance-encoding/src/encodings/logical/primitive.rs
Xuanwo: I weighed in here a bit and developed two versions, but neither looks good to me. For now, I’ll keep it as is until it becomes more complex.

PR #5007, opened 2025-10-20
refactor: don't add new with_... columns to scanner

westonpace:
We recently added `with_row_last_updated_at_version` and `with_row_created_at_version` methods to `Scanner`.  I'd like to avoid that, preferring to use `project` to ask for those columns.  This is because we may have quite a few system columns over time and I don't want the scanner to have lots of `with_` methods.  These methods need to be added to a lot of places (python bindings, take methods, java bindings, lancedb, etc.)  In addition, the `with_` methods make it unclear the expected output order of the columns (e.g. if I specify `with_row_id` _and_ `with_row_addr` then which comes first).

This refactor changes things so we just use `project`.

However, this introduced one small wrinkle, which is that we need to be able to say "all columns plus these system columns".  I introduced the concept of `*` to allow this so you can no do something like `project(&["*", "_rowid"])`.

I need to add more test cases for the wildcard but wanted to verify that this is an approach everyone is happy with first.

### DIFF
--- a/rust/lance/src/dataset/scanner.rs
+++ b/rust/lance/src/dataset/scanner.rs
@@ -4863,7 +4863,6 @@ mod test {
         check_exprs(&["_rowid", "*", "x * 2"], &["_rowid", "x", "y", "x * 2"]).await;
 
         let check_fails = |projection: &[&str]| {
-            println!("Checking fails: {:?}", projection);
             let mut scan = data.scan();
             assert!(scan.project(projection).is_err());
         };


jackye1995:
> We recently added with_row_last_updated_at_version and with_row_created_at_version methods to Scanner. I'd like to avoid that, preferring to use project to ask for those columns. This is because we may have quite a few system columns over time and I don't want the scanner to have lots of with_ methods.

Yeah definitely +1 on this, I was actually planning to do this today after the quick fix yesterday, you beat me 😆 

> This refactor changes things so we just use project.

I was thinking if we should completely remove all the with_... and have from scanner to projection to fragment reader all just take the system columns like normal columns, and we assign reserved field IDs to those system columns so that they are uniquely identified. Otherwise in the fragment reader we are still just appending system columns in the end of the record batch, instead of preserving the requested projection order.

westonpace:
> Otherwise in the fragment reader we are still just appending system columns in the end of the record batch, instead of preserving the requested projection order.

This is ok I think.  We do one last step at the end to reorder the columns to the user's desired order.  This is really cheap and easier than trying to make sure the order is correct all the way through.

> I was thinking if we should completely remove all the with_... and have from scanner to projection to fragment reader all just take the system columns like normal columns, and we assign reserved field IDs to those system columns so that they are uniquely identified.

That seems fine too.  I don't know how much it simplifies things but at least we won't have a bunch of fields on `Projection`.

jackye1995:
> We do one last step at the end to reorder the columns to the user's desired order. This is really cheap and easier than trying to make sure the order is correct all the way through.

Putting the simplifying things aspect aside, I think the main problem is that there are increasing number of use cases that directly uses the fragment level APIs, and they are public APIs today. That will have the same issues as scanner today. Do we envison the fragment-level read APIs to be public APIs, or do we envision that users should always use it through scanner with `with_fragments`?

westonpace:
> Putting the simplifying things aspect aside, I think the main problem is that there are increasing number of use cases that directly uses the fragment level APIs, and they are public APIs today. That will have the same issues as scanner today. Do we envison the fragment-level read APIs to be public APIs, or do we envision that users should always use it through scanner with with_fragments?

I think, at the python / java level, sure, we can have fragment level APIs.  At the rust level, I think the fragment stuff probably needs to be split.  Some of it is "fragment API" which is like the `Dataset` class but just implicitly operates on one fragment.  All of the "real work" should be in scanner / updater / insert builder / etc.  The rest of the stuff in fragment.rs today is more like a "fragment reader" which should be more internal.

image

westonpace:
> That will have the same issues as scanner today. Do we envison the fragment-level read APIs to be public APIs, or do we envision that users should always use it through scanner with with_fragments?

So, today, if we can make the fragment implementation redirect to the scanner, that is something we should be doing.

Issue #5004, opened 2025-10-20
Allow dictionary encoding to support more fixed width types


westonpace:
Currently the dictionary encoding in 2.1 only applies to some 128-bit fixed-width types.  I think this was a combination of a few things:

* At the time, we were not bit-packing indices, and so applying dictionary encoding to smaller types was unlikely to be beneficial
* Estimating cardinality on all columns was expensive
* TPC-H data has a lot of 128-bit decimals that benefit from dictionary encoding so there was an outsized (and unrealistic) benchmarking gain to focusing on 128-bit types.

However, we should at least be considering dictionary encoding for 64-bit types and maybe even 32-bit types.

westonpace:
There are multiple PRs I think that added (or will add) support for 64bit types.  However, I think we still don't support 128bit types.  I'm going to move this issue to 2.3 for 128bit types.

I'm not really worried about 128bit types.  These seem more specific to TPC-H and maybe not that common in real-world use cases.

lixmgl:
hi @westonpace I am interested in picking this up. Planning to scope this PR to 32-bit fixed-width types (u32/i32/f32/Date32), mirroring the existing 64-bit implementation in dict.rs. Willing to follow up with a genericized refactor if that's preferred. Does that match what you had in mind, or did you want broader coverage (16-bit, 256-bit)?

PR #5008, opened 2025-10-20
feat: add support for 64bits fixed width dictionary encoding

yingjianwu98:
resolve https://github.com/lancedb/lance/issues/5004

Please look at only the second commits change as it is a dependent work on the first commit which will be merged first into main.

Also I didn't include 32 bits change here because the current estimate function doesn't consider bit-piacking compression, and the estimation will always evaluate 32 bits as not worth for compression.


yingjianwu98:
@westonpace 

I have rebased the PR and it should be good to merge, thanks!

yingjianwu98:
Looks like 640bit dictionary encoding is supported now, so closing this PR

PR #5009, opened 2025-10-20
feat: support cleanup across branches

### LINKED ISSUE #4858: Support file lifetime management across branches

majin1102:
Close #4858

### DIFF
--- a/rust/lance/src/dataset/cleanup.rs
+++ b/rust/lance/src/dataset/cleanup.rs
@@ -610,7 +610,7 @@ impl<'a> CleanupTask<'a> {
                             referenced_branches.insert(branch_name.clone());
                         }
                     }
-                    Ok::<(), Box<dyn std::error::Error + Send + Sync>>(())
+                    Ok::<(), Error>(())
                 }
             })
             .collect();
@@ -664,7 +664,7 @@ impl<'a> CleanupTask<'a> {
                             stats_guard.old_versions += stats.old_versions;
                         }
                     }
-                    Ok::<(), Box<dyn std::error::Error + Send + Sync>>(())
+                    Ok::<(), Error>(())
                 }
             })
             .collect();


majin1102:
Something goes wrong after rebasing. I will fix it tomorrow

majin1102:
Add more tests for cornor cases.

rust/lance/src/dataset/cleanup.rs
```diff
@@ -453,6 +686,8 @@ pub struct CleanupPolicy {
     pub delete_unverified: bool,
     /// If true, return an Error if a tagged version is old
     pub error_if_tagged_old_versions: bool,
+    /// If clean the referenced branches
+    pub clean_referenced_branches: bool,
```
jackye1995, line 798: my understanding is that we always want to cleanup? Because branch is a part of the complete dataset. If there is not a real use case, I think we should always do that, but if there is a valid use case I guess we could have this boolean but is should be true by default.

rust/lance/src/dataset/cleanup.rs
```diff
@@ -441,6 +469,211 @@ impl<'a> CleanupTask<'a> {
...
+                    // we should check the referenced branch to make sure the referenced files could be retained
+                    if let Ok(manifest) = manifest {
+                        if self.policy.should_clean(&manifest) {
+                            referenced_branches.push((branch.clone(), *ref_version));
+                        }
+                    }
+                } else {
+                    return Err(Error::Internal {
+                        message: format!(
+                            "Branch {} is not referenced by any version from {}",
```
jackye1995: in that case,  I think we should cleanup the branch because it is orphan/invalid? Since having a reference from a source branch is required. If that source version is already deleted, this branch should be deleted as well.

rust/lance/src/dataset/cleanup.rs
```diff
@@ -282,9 +310,9 @@ impl<'a> CleanupTask<'a> {
         let old_manifests = inspection.old_manifests.clone();
         let num_old_manifests = old_manifests.len();
 
-        // Ideally this collect shouldn't be needed here but it seems necessary
+        // Ideally this collect shouldn't be needed here but it sseems necessary
```
jackye1995: nit: typo sseems

rust/lance/src/dataset/branch_lineage.rs
```diff
@@ -0,0 +1,516 @@
...
+/// - `children`: ordered set for deterministic traversal.
+#[derive(Debug)]
+pub struct BranchLineage {
+    pub deleted: bool,
+    pub branch: Option<String>,
+    pub parent_version_number: Option<u64>,
+    pub children: BTreeSet<BranchLineage>,
+}
+
+impl BranchLineage {
```
jackye1995: I am wondering if this is an over-complication to solve the problem. in my mind, there are 2 things we need:
1. get a post-order for the branches to cleanup
2. then for each branch, cleanup like a normal dataset. But also we need to pass in the tagged and branch head versions in this dataset.

I don't think we need a full BranchLineage concept to do this. This is an example I requested claude to write, just for reference, but I think it should work:

```rust
  /// Get cleanup order (post-order) for branches descending from `root`
  fn get_descendants_post_order(
      branches: &HashMap<String, BranchContents>,
      root: Option<&str>,
  ) -> Vec<String> {
      let mut result = Vec::new();
      let mut stack = vec![root.map(String::from)];

      while let Some(current) = stack.pop() {
          // Find children of current
          let children: Vec<_> = branches.iter()
              .filter(|(_, c)| c.parent_branch == current)
              .map(|(name, _)| name.clone())
              .collect();

          if children.is_empty() {
              if let Some(name) = current {
                  result.push(name);
              }
          } else {
              // Re-push current, then push children
              stack.push(current);
              for child in children {
                  stack.push(Some(child));
              }
          }
      }
      result
  }
```

```rust
  /// Get versions pinned by child branches
  fn get_pinned_versions(
      branches: &HashMap<String, BranchContents>,
      branch: Option<&str>,
  ) -> HashSet<u64> {
      branches.values()
          .filter(|c| c.parent_branch.as_deref() == branch)
          .map(|c| c.parent_version)
          .collect()
  }
```

What do you think?

rust/lance/src/dataset/branch_lineage.rs
```diff
@@ -0,0 +1,516 @@
...
+///
+/// Semantics:
+/// - `branch`: node name (None denotes the virtual main root).
+/// - `parent_version_number`: the version of the parent at which this child branch was created.
+/// - `deleted`: indicates the node was inferred from a preserved snapshot (parent deleted in contents),
+///   still kept to preserve lineage connectivity for inspection.
+/// - `children`: ordered set for deterministic traversal.
+#[derive(Debug)]
+pub struct BranchLineage {
+    pub deleted: bool,
```
jackye1995: should we really support this? I think because we track branch root instead of branch head, we should enforce that child branch is deleted before parent branch. I understand we want the experience to be similar to git, but git tracks branch by head not by root, and there is no data dependency between child and parent, so it is much more feasible in git than in our case. Even if we want to support this, I feel it should be a much later iteration of the feature after we make the basic cases working.

rust/lance/src/dataset/branch_lineage.rs
majin1102: > I don't think we need a full BranchLineage concept to do this. This is an example I requested claude to write, just for reference, but I think it should work

After re-thinking this through, I use a global branch identifier to make things work. 
1. Branch Identifier is global identical even if it's deleted unlike branch name. This could avoid id traveling if we delete a branch and create a new one using the same name.
2. Use sort or reverse_sort to pre-order or post-order traverse child branches, making traversing simple.
3. Potentially we could use branch_id and version_number to globally identifier a snapshot.
4. Potentially if we use branch_id to generate the branch path(just use the last uuid, like dataset/tree/uuid), we could safely delete branch and create a new one with the same name(might introduce break change, not implemented yet).

rust/lance/src/dataset/cleanup.rs
majin1102: Each branch would call cleanup hook when they have new versions come in. This is a lighter way for cleanup. I think If we set this true by default there would be a lot of unnecessary manifest scanning especially for the main branch. Unlike Iceberg has a metadata file, I have some performance concerns in commit phases(if we have a very write-frequent branch, that might impact all branches commit). There are scenarios I think need this to be true: 
1. We use a time window policy in cleanup. And some branches were inactive for a very long time. We don't expect a write on each branch to free the files in the root.
2. We could call a cleanup on the main branch to cleanup the whole dataset.

What do you think?

rust/lance/src/dataset/branch_lineage.rs
majin1102: I agree that we should as least provide the enforcement in `delete_branch`.  I reveal my thoughts as below.

>  I understand we want the experience to be similar to git, but git tracks branch by head not by root, and there is no data dependency between child and parent

This is a good top thinking. Let me try to elaborate further on this.

For Git:

- Git uses a `head` to track a branch
- Potentially the branch lineage relies on the `parent pointer` in each commit version
- All commit version metadata would not delete(as far as I know). The branch lineage would not lose when we delete a branch.

Iceberg uses similar mechnisim to manage branches. Let's say we have main, branch1 from main, branch2 from branch1. When we delete branch1, the snapshots in branch1 which are not inherited by branch2 would become unreference and be cleaned up in  `ExpireSnapshots`. That says Iceberg supports delete a middle branch and could deal with cleanup correctly.

For Lance:

- Lance uses branch_name:version_number to track a branch
- Potentially the branch lineage relies on the branch root mapping and version sequencial. This drives to Branch Identifier
- Delete branch would delete both branch data and metadata

I think the reasonable action for deleting a branch to make deletion work and keep data available is:
1. delete the `BranchContents`  file at first (this is the last operation for now)
2. Prepare a particular CleanupPolicy for the deleted branch dataset. The policy would only retain the referenced files
3. Run the cleanup.
4. If the cleanup failed. It could be re-run by a parent branch cleanup by setting clean_referenced_branch true

For this PR, I think it's fine to just throw error when deleting branches with children.  I think the deletion action above could be resolved after(or if) we use UUID to generate the branch path to avoid id traveling and  name occuping issue(honestly let me look back I wouldn't prefer the branch name to build path although it seems more friendly to oss users and could reuse dataset commit to provide atomicity). On the other hand, introducing BranchIdentifier could be a good chance to add backward and forward compatibility—if we decide to go this route. I’d love to hear your thoughts on this @jackye1995 

This will introduce format change. We should remain cautious. I’ll organize my thoughts if necessary, but overall, after this PR, the feature has reached a production-ready state—except for some limitations around branch deletion.

majin1102:
Some discussions above. Please take a look with the newest code.

jackye1995:
Sorry I missed this PR lately, will take another pass tomorrow

majin1102:
I've pushed some updates today:

- Parallelized the branch cleaning process for better performance.
- Polished comments and naming for improved readability.

Apologies for the multiple pushes, I hope it didn't interrupt any work. @jackye1995  

The code is truely ready for review. I know this PR has accumulated quite a bit of context over time, please let me know if anything is unclear or if you'd like me to clarify any part of the changes

jackye1995:
I checked out our branch locally, the commit history seems a bit messed up? 

I see this:

```
0d49b66e2 (HEAD -> branch_clean_up) fix clippy after rebasing
f16c54510 Merge branch 'main' into branch_clean_up
772c6d87a ci: run workflows also on release branch (#5398)
052e01885 refactor: split dataset tests in a tests mod (#5387)
14087e7fd feat(blob_v2): add external blob support (#5385)
58fd8fcff squash
c71607eb8 chore: remove lancedb in github discussion links and java pom file (#5394)
d9dcd8a5c fix: respect index metric when user overrides (#5395)
926156226 fix: don't allow change blob version during update (#5386)
72075b84b refactor: rename RowIdTreeMap to RowAddrTreeMap (#5266)
6c50d977b chore: bump main to 1.1.0-beta.0
88e44b1ab refactor!: deprecate mac x86 support (#5391)
6381f7c65 fix: add graceful shutdown and start for rest namespace adapter (#5325)
e7379a0fa chore: polish agents.md for better behavior (#5383)
6e458989a perf: avoid allocating filtered nodes on HNSW search path (#5377)
b7024fa37 chore: update lock file for python binding (#5376)
c3600d489 feat(java): support writing schema metadata through java LanceFileWriter API (#5310)
ff89675eb (tag: v1.0.0-beta.16) chore: release beta version 1.0.0-beta.16
```

looks like somehow there is a squash in the middle, and some new commits after the head. Could you double check it?

majin1102:
Sorry about the confusion. After that squash commit, I performed a lengthy rebase and had already resolved the conflicts beforehand.
To avoid inconsistencies, I recommend either:

1. Deleting your local branch and checking out again from the updated remote, or
2. Resetting your HEAD to the commit just before the squash, then pulling the latest changes. (not sure if there are any stale commit before that `squash` commit)

The issue you're seeing is likely because your local branch was based on the old version of the squash commit. If you do a fresh pull from a clean state, you should see the rebased version line(**NOTE: the `squash` commit has a different revision from your local**).

image

rust/lance/src/dataset/cleanup.rs
```diff
@@ -542,6 +572,216 @@ impl<'a> CleanupTask<'a> {
...
+            .cloned()
+            .collect();
+        Ok(referenced_branches)
+    }
+
+    async fn clean_referenced_branches(
+        &self,
+        referenced_branches: &[(String, u64)],
+    ) -> Result<RemovalStats> {
+        let final_stats = Mutex::new(RemovalStats::default());
```
jackye1995: this is currently using a sync mutex, but it's used in `async move`, I think we need to use `tokio::sync::Mutex`?

rust/lance/src/dataset/cleanup.rs
```diff
@@ -542,6 +572,216 @@ impl<'a> CleanupTask<'a> {
...
+                            branch_dataset.manifest.as_ref(),
+                        )
+                        .await?
+                        {
+                            let mut stats_guard = final_stats.lock().unwrap();
+                            stats_guard.bytes_removed += stats.bytes_removed;
+                            stats_guard.old_versions += stats.old_versions;
+                        }
+                    }
+                    Ok::<(), Box<dyn std::error::Error + Send + Sync>>(())
```
jackye1995: this can just be `Ok::<(), lance_core::Error>(())` I think?

PR #5012, opened 2025-10-20
feat(java): support branch operations

### LINKED ISSUE #4855: Support branch in Java

majin1102:
Close #4855

### DIFF
--- a/java/lance-jni/src/blocking_dataset.rs
+++ b/java/lance-jni/src/blocking_dataset.rs
@@ -210,11 +210,6 @@ impl BlockingDataset {
         Ok(())
     }
 
-    pub fn checkout_branch(&mut self, branch: &str) -> Result<Self> {
-        let inner = RT.block_on(self.inner.checkout_branch(branch))?;
-        Ok(Self { inner })
-    }
-
     pub fn checkout_reference(
         &mut self,
         branch: Option<String>,


majin1102:
Ready for review @jackye1995

java/lance-jni/src/blocking_dataset.rs
```diff
@@ -186,8 +186,61 @@ impl BlockingDataset {
...
+        Ok(branches)
+    }
+
+    pub fn create_branch(
+        &mut self,
+        branch: &str,
+        version: u64,
+        source_branch: Option<&str>,
+    ) -> Result<Self> {
+        let ref_ = match source_branch {
```
jackye1995: why this strange `ref_` variable name?

java/src/main/java/com/lancedb/lance/Reference.java
```diff
@@ -31,20 +30,32 @@ public Reference(
...
   }
 
-  public Optional<String> getTagName() {
+  public Optional<String> tagName() {
     return tagName;
   }
 
-  public static Builder builder() {
-    return new Builder();
+  public static Reference ofMain(long versionNumber) {
```
jackye1995: I guess it's missing a `public static Reference ofMain()` that takes nothing?

java/src/main/java/com/lancedb/lance/Branch.java
```diff
@@ -0,0 +1,89 @@
...
+
+/**
+ * Branch metadata aligned with Rust's BranchContents. name is the branch name, parentBranch may be
+ * null (indicating main), parentVersion is the version on which the branch was created, createAt is
+ * the unix timestamp (seconds), and manifestSize is the size of the referenced manifest file in
+ * bytes.
+ */
+public class Branch {
+  private final String name;
+  private final String parentBranch; // nullable
```
jackye1995: can we use `Optional<String>` for consistency?

java/src/main/java/com/lancedb/lance/Reference.java
```diff
@@ -31,20 +30,32 @@ public Reference(
     this.tagName = tagName;
   }
 
-  public Optional<Long> getVersionNumber() {
+  public Optional<Long> versionNumber() {
```
jackye1995: why change this? The ones in branch are using `getXXX`. I think we should stick with one thing and maybe add a cehckstyle rule to enforce it

java/src/main/java/com/lancedb/lance/Dataset.java
```diff
@@ -951,6 +951,39 @@ public Tags tags() {
...
+    }
+  }
+
+  /**
+   * Checkout the latest version of a branch.
+   *
+   * @param branch branch name
+   * @return a new Dataset instance checked out to the branch's latest version
+   */
+  public Dataset checkoutBranch(String branch) {
```
jackye1995: this seems redundant?

java/src/main/java/com/lancedb/lance/Reference.java
majin1102: I remember a PR said this style is prefered(first one or two PR of mine I guess). There are many places using this rust nameing style in Java module now. I could raise another PR to align them.  Actually I prefer this getXXX way more because our agent loves it

majin1102:
Ready for another look @jackye1995

java/src/main/java/com/lancedb/lance/Reference.java
jackye1995: Sure. My point is more on having ONE style instead of mixed, and having rules enforcing it. I am good with either way.

jackye1995 (approved):

Issue #4925, opened 2025-10-10
flaky test_read_btree_index_with_defer_index_remap


jackye1995:
https://github.com/lancedb/lance/actions/runs/18398027809/job/52421101916?pr=4765

I will take a look later

PR #5014, opened 2025-10-20
fix: update to respect file version from write params when writing fragments in java

morales-t-netflix:
As part of the Lance 2.1 release, [Java began exposing format version in the write params](https://github.com/lancedb/lance/pull/4820). 

However, the write_fragments method used by the Java SDK _doesn't_ currently respect this param, and overwrites to always match stable.

This change now makes it so the file version is respected


morales-t-netflix:
Looks like failing test is just flakey: https://github.com/lancedb/lance/issues/4925

jackye1995 (approved):
looks good to me!

PR #4865, opened 2025-10-01
docs: add json to full text search document

wojiaodoubao:

Issue #4524, opened 2025-08-21
feat: ability to get tokenized query in SDK for full text search


BubbleCal:
If we can know what the tokenized query is, it would be useful for debugging for both the dev and the users. I think we just need to expose the tokenizer, then we can achieve this even without creating the FTS index.

- [ ] show tokenized query in `analyze_plan`
- [ ] expose tokenizer to Pylance
- [ ] expose tokenizer to LanceDB SDKs

PR #5020, opened 2025-10-21
feat: add json fts in python

wojiaodoubao:
Related to https://github.com/lancedb/lance/issues/4749

### DIFF
--- a/python/python/tests/test_scalar_index.py
+++ b/python/python/tests/test_scalar_index.py
@@ -3977,7 +3977,7 @@ def test_json_inverted_match_query(tmp_path):
     # Convert to JSON strings
     json_strings = pa.array([json.dumps(doc) for doc in json_data], type=pa.json_())
     table = pa.table({"json_col": json_strings, "id": range(len(json_data))})
-    dataset = lance.write_dataset(table, tmp_path, data_storage_version="2.2")
+    dataset = lance.write_dataset(table, tmp_path)
 
     # Create inverted index with JSON tokenizer
     dataset.create_scalar_index(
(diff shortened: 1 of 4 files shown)


wojiaodoubao:
> Do we have a plan to add some docs somewhere explaining how this works? Just from looking at the API I (knowing little about what changes we've done here) am a little confused by the difference between `lance_tokenizer` and `base_tokenizer`.

Hi @westonpace , yes we have a document for this: https://github.com/lancedb/lance/pull/4865

BubbleCal:
related to https://github.com/lancedb/lance/issues/4524

wojiaodoubao:
Hi @BubbleCal @westonpace , could you help review this when you have time, thanks very much!

BubbleCal:
This seems not what I expected, why do we need the new param `lance_tokenizer`? my understanding is we can just use json tokenizer if the column is JSON, is it because we can't infer this from the column data type?

If so, can we just rename this param to `content_type`?

wojiaodoubao:
Hi @BubbleCal , thanks your suggestion!  Inferring lance_tokenizer based on the storage type is a good idea, as it simplifies things at the API level.

Exposing lance_tokenizer (or content_type) would provide more flexibility. This means that if a column of data is stored as string instead of pa.json_(), but its content is JSON, you can still create a JSON full-text index by configuring lance_tokenizer.

Ultimately, I agree with the approach of using type inference. If we truly need flexibility in the future, we can revisit the idea of exposing content_type.

wojiaodoubao:
I updated the patch following suggestion. Failed ut is unrelated. Hi @BubbleCal,  please let me know your thoughts, thanks very much!

PR #5025, opened 2025-10-21
fix: general block decompression mismatch for Lance 2.2 dictionaries

Xuanwo:
PR #4900 enabled writing dictionary pages through `Compression::General`, but the read path `DefaultDecompressionStrategy::create_block_decompressor` still routed those pages directly into `BinaryBlockDecompressor`. When the compressed payload actually held fixed-width dictionary values (e.g., MiniBlockSchedulerDictionary with fixed Int8 lists), the first byte was reinterpreted as `bits_per_offset`, panicking inside `BinaryBlockDecompressor`.

We didn't detect this during PR because we didn't add the test for `2.2` version.

This PR fixed this issue and added a new test to cover.

---

### DIFF
--- a/rust/lance-encoding/src/compression.rs
+++ b/rust/lance-encoding/src/compression.rs
@@ -1282,50 +1282,42 @@ mod tests {
     #[test]
     #[cfg(any(feature = "lz4", feature = "zstd"))]
     fn test_general_block_decompression_fixed_width_v2_2() {
-        use crate::compression::BlockCompressor;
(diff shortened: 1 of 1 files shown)


westonpace (approved):
Do we still need `impl BlockDecompressor for CompressedBufferEncoder`?

Xuanwo:
> Do we still need `impl BlockDecompressor for CompressedBufferEncoder`?

I think they don't need anymore, let me create a new PR to remove them.

PR #5034, opened 2025-10-22
feat: support GEO RTree index

ddupg:


rust/lance-index/src/scalar/rtree.rs
```diff
@@ -0,0 +1,1224 @@
...
+
+    fn statistics(&self) -> Result<serde_json::Value> {
+        serde_json::to_value(self.metadata.clone()).map_err(|e| Error::Internal {
+            message: format!("Error serializing statistics: {}", e),
+            location: location!(),
+        })
+    }
+
+    async fn prewarm(&self) -> Result<()> {
+        if self.metadata.num_pages == 0 {
```
Kontinuation: It probably should be `if self.metadata.num_pages > 0`.

rust/lance-index/src/scalar/rtree.rs
```diff
@@ -0,0 +1,1224 @@
...
+
+        Ok(())
+    }
+
+    pub async fn write_nulls(store: &dyn IndexStore, null_map: RowIdTreeMap) -> Result<()> {
+        let mut writer = store
+            .new_index_file(RTREE_NULLS_NAME, RTREE_NULLS_SCHEMA.clone())
+            .await?;
+        let mut bytes = Vec::new();
+        null_map.serialize_into(&mut bytes)?;
```
Kontinuation, line 829: The serialization format of `RowIdTreeMap` is not stable according to the documentation of [`RowIdTreeMap.serialize_into`](https://docs.rs/lance-core/0.38.2/lance_core/utils/mask/struct.RowIdTreeMap.html#method.serialize_into). Persisting it in a file may introduce problems when the serialization format is changed.

Kontinuation (commented):
I'm very interested in your work. After a preliminary review of the code, I have a few suggestions/comments

rust/lance-index/src/scalar/rtree.rs
ddupg: Fixed.

rust/lance-index/src/scalar/rtree.rs
ddupg: > The serialization format of `RowIdTreeMap` is not stable according to the documentation of [`RowIdTreeMap.serialize_into`](https://docs.rs/lance-core/0.38.2/lance_core/utils/mask/struct.RowIdTreeMap.html#method.serialize_into). Persisting it in a file may introduce problems when the serialization format is changed.

Thank you @Kontinuation for reviewing this. I'm glad to find that `RowIdTreeMap.serialize_into` is not stable. I drew inspiration from the bitmap index by directly serializing the RowIdTreeMap, which reduces code complexity and storage overhead. Perhaps should consider alternative approaches.

ddupg:
The python example is currently blocked at #5144

rust/lance-index/src/scalar/rtree.rs
ddupg: > The serialization format of RowIdTreeMap is not stable

Have make `RowIdTreeMap.serialize_into` stable in #5105

Cargo.toml
```diff
@@ -8,6 +8,7 @@ members = [
     "rust/lance-encoding",
     "rust/lance-file",
     "rust/lance-index",
+    "rust/lance-geo",
```
jackye1995, line 10: this is a pretty big change, I would suggest we have a PR to get the `lance-geo` module in first, with the tests for geo type and functions.

Cargo.toml
```diff
@@ -105,7 +107,7 @@ criterion = { version = "0.5", features = [
     "html_reports",
 ] }
 crossbeam-queue = "0.3"
-datafusion = { version = "50.0.0", default-features = false, features = [
+datafusion = { version = "50.3.0", default-features = false, features = [
```
jackye1995: I think let's bump datafusion in a separate PR?

rust/lance-geo/src/bbox.rs
```diff
@@ -0,0 +1,340 @@
...
+    for item in arr.iter().flatten() {
+        bbox.add_geometry(&item?);
+    }
+
+    Ok(bbox)
+}
+
+/// Convert a length-1 GeoArrowArray to a geo::Geometry scalar.
+pub fn bounding_box_single_scalar(arr: &dyn GeoArrowArray) -> ArrowResult<BoundingBox> {
+    assert_eq!(arr.len(), 1);
```
jackye1995: prefer error not panic

rust/lance-index/src/scalar/rtree/sort/hilbert_sort.rs
```diff
@@ -0,0 +1,257 @@
...
+                "hilbert only supports array arguments".to_owned(),
+            )),
+        }?;
+
+        let rect_array = bounding_box(value.as_ref()).map_err(DataFusionError::from)?;
+
+        let hilbert_max = ((1 << 16) - 1) as f64;
+        let len = rect_array.len();
+        let width = self.bbox.maxx() - self.bbox.minx();
+        let height = self.bbox.maxy() - self.bbox.miny();
```
jackye1995, line 183: what if the points are in the same x or y axis? seems like we need a minimum for width and height that is not zero, since I remember hilbert curve does not divide by 0 (if it does then never mind)

ddupg: Good point. In fact, division-by-zero will not occur. If all x are 0, in the subsequent expression, `(bbox.minx() + bbox.maxx()) / 2. - self.bbox.minx()` will yield 0, resulting in the entire expression being 0.
```
let x = (hilbert_max * ((bbox.minx() + bbox.maxx()) / 2. - self.bbox.minx()) / width)
                .floor() as u32;
```
To be safe, I've also updated the logic to guard against division-by-zero by setting the `width`/`height` to 1.0 if max equals min. This approach aligns the one used in flatbush ([lines 158-171](https://github.com/mourner/flatbush/blob/1e333b0369c19135fa5d89b1f8faab82432c11e8/index.js#L158-L171)) which effectively ensures the normalization accuracy.

rust/lance-geo/src/bbox.rs
ddupg: The method is unnecessary, removed it.

Cargo.toml
ddupg: This upgrade may be unnecessary, I just wanted to keep it consistent with the lock file. I noticed that bump datafusion 51 is in progress, so I've rolled back this bump.

Cargo.toml
ddupg: OK，I separate to https://github.com/lance-format/lance/pull/4678/, please help review it.

rust/lance-index/src/scalar/rtree.rs
jackye1995, line 1: And one other thing is that we should raise a separated PR + vote to add RTree to the table index spec, treat it as a spec change. This follows the new community governance rule, and also allows a pre-review of the general design. We did not do that for zone map and it caused quite a few issues, so I think let's follow the new process here.

ddupg: Thanks for the suggestion! I’ve raised a PR #5360 outlining the RTree index spec.

ddupg:
Could @jackye1995 please review this again when you have time?

jackye1995 (approved):
looks good to me! I think all the old comments are addressed, and I am good with the current implementation. Thanks for pushing this through!

PR #5043, opened 2025-10-22
chore: use latest mock_instant for testing

jackye1995:
mock_instant latest version has now a thread local clock, so we don't need a custom mutex anymore


Xuanwo (approved):
LGTM once CI passing.

Issue #5042, opened 2025-10-22
Drop default features on lance workspace dependency


valkum:
Now that `lance-namespace` moved from [lancedb/lance-namespace](https://github.com/lancedb/lance-namespace) in here, https://github.com/lancedb/lance-namespace/pull/242 no longer applies.
The latest beta still pulls in aws-* crates due to lancedb depending on `lance-namespace-impls` with default features enabled.
`lance-namespace-impls` will pull in `lance` with default features enabled, which in fact will enable the `aws` feature.

# Details
When using `lance` in one of the subcrates such as `lance-namespace-impls` this will pull in all the dependencies that are enabled by the default features.

Dropping the default features avoids having `aws-*` crates as an implicit hard dependency.

Here is a cargo tree output of the current beta. I have `v0.38.3-beta.9` checked out locally and tested that disabling the default features on `lance` will in fact, remove the dependency on aws-lc-rs (and the other aws-* crates)

```
aws-lc-rs v1.14.1
├── rustls v0.23.34
│   ├── aws-smithy-http-client v1.1.3
│   │   ├── aws-smithy-http-client feature "default"
│   │   │   └── aws-smithy-runtime v1.9.3
│   │   │       ├── aws-smithy-runtime feature "client"
│   │   │       │   ├── aws-config v1.8.8
│   │   │       │   │   ├── aws-config feature "credentials-process"
│   │   │       │   │   │   └── aws-config feature "default"
│   │   │       │   │   │       └── lance-io v0.38.3-beta.9 (/Users/valkum/Developer/lance/rust/lance-io)
│   │   │       │   │   │           ├── lance v0.38.3-beta.9 (/Users/valkum/Developer/lance/rust/lance)
│   │   │       │   │   │           │   ├── lance feature "aws"
│   │   │       │   │   │           │   │   └── lance feature "default" << This is causing the inclusion of `aws-*`
│   │   │       │   │   │           │   │       └── lance-namespace-impls v0.38.3-beta.9 (/Users/valkum/Developer/lance/rust/lance-namespace-impls)
│   │   │       │   │   │           │   │           ├── lance-namespace-impls feature "default"
│   │   │       │   │   │           │   │           │   └── lancedb v0.22.3-beta.2 (https://github.com/lancedb/lancedb.git?tag=v0.22.3-beta.2#451411b1) (*)
```

<details><summary>More cargo tree output</summary>
<p>

```
aws-lc-rs v1.14.1
├── rustls v0.23.34
[...]

PR #5045, opened 2025-10-22
refactor: use lance-io object store for dir namespace and improve builder

jackye1995:
I ended up doing these in https://github.com/lancedb/lance/pull/4984 and https://github.com/lancedb/lance/pull/4905 so I decided to pull it out and get it cleaned up first.

This PR moves the directory namespace from using OpenDAL directly to using Lance ObjectStore. This avoids the inconsistency between the dir namespace and the underlying lance table storage configurations. User can still use OpenDAL, and if we fully migrate Lance to OpenDAL it will be applied to both layers at the same time as well.

The PR also improves the builder of the namespaces with builder style and allow supplying a Lance session. Since we have not published a stable version yet, we do not care about backwards compatibility.

This PR also ensures the lance-namespace-impls features are consistent with lance-io features. Related to https://github.com/lancedb/lance/issues/5042

### DIFF
--- a/rust/lance-namespace-impls/src/rest.rs
+++ b/rust/lance-namespace-impls/src/rest.rs
@@ -703,10 +703,9 @@ mod tests {
 
     /// Create a test REST namespace instance
     fn create_test_namespace() -> RestNamespace {
-        let mut properties = HashMap::new();
-        properties.insert("uri".to_string(), "http://localhost:8080".to_string());
-        properties.insert("delimiter".to_string(), ".".to_string());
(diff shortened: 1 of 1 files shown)


Xuanwo (approved):
LGTM while all CI passing. But I think the `aws-*` crates is not introduced by opendal.

jackye1995:
Oh yeah that's a separated issue, not introduced by Opendal, let me clarify in the description

jackye1995:
Known flaky test https://github.com/lancedb/lance/issues/4925

PR #5071, opened 2025-10-26
feat: introduce optimize columns operation

xerryd:
## Summary
This continues work originally started by @geruh in #4504. Thanks for letting me pick this up and see it through! The original implementation had the right idea but needed some refinement around transaction handling, conflict resolution, and the overall architecture. This PR addresses all of that, and should be in a good state. 

`optimize_columns` reorganizes how columns are stored within data files without changing fragment structure. Think of it as horizontal compaction where you're not merging fragments together like `compact_files`) but you're just rearranging which columns live in which files within each fragment.

The key constraint in this impl is that fragment IDs and row counts stay the same. This means all your existing indices continue to work without any remapping needed.

## Usage
```
use lance::dataset::optimize_columns::{optimize_columns, OptimizeColumnsOptions, OptimizeGroups};

let options = OptimizeColumnsOptions {
    optimize_groups: Some(vec![
        OptimizeGroups::new(vec!["vector".to_string()]),
        OptimizeGroups::new(vec!["id".to_string(), "metadata".to_string()]),
    ]),
    num_threads: Some(4),
};

let metrics = optimize_columns(&mut dataset, options).await?;
println!("Reorganized into {} new files, removed {} old files", 
         metrics.files_added, metrics.files_removed);
```
If you don't specify groups, it puts everything in one file (basically a no-op if you're already at one file per fragment).

### Workflow
Similar to what existed in the previous PR.

- Validation: Make sure column groups are valid (no duplicates, all columns covered, etc)
- Planning: For each fragment, figure out which files need rewriting vs which already match the target layout
- Execution: Stream through fragments, fan out each batch to multiple writers in parallel
- Commit: Replace old fragments with new ones in a transaction

TODO: add workflow diagram

### Transaction Handling

I Added a new `Operation::OptimizeColumns` that tracks old and new fragments. The validation enforces that this is truly a 1:1 replacement with the same fragment IDs, same row counts, just different file layouts. As for conflict resolution, this operation is compatible with operations that don't touch fragments (`Append`, `CreateIndex`, etc), conflicts with operations that modify the same fragments (`Delete`, `Update`, `Rewrite` on overlapping fragments),
always conflicts with schema evol operations (`Merge`, `DataReplacement`), and 
Incompatible with operations that invalidate dataset state like `Overwrite`.

### DIFF
--- a/protos/transaction.proto
+++ b/protos/transaction.proto
@@ -262,6 +262,14 @@ message Transaction {
     repeated DataReplacementGroup replacements = 1;
   }
 
+  // An operation that rewrites column groups within fragments while preserving row content.
+  message OptimizeColumns {
+    // The fragments replaced by this operation.
+    repeated DataFragment old_fragments = 1;
+    // The replacement fragments with reorganized optimize groups.
+    repeated DataFragment new_fragments = 2;
+  }
+
   // Update the state of the MemWal index
   message UpdateMemWalState {
 
@@ -296,6 +304,7 @@ message Transaction {
     UpdateMemWalState update_mem_wal_state = 112;
     Clone clone = 113;
     UpdateBases update_bases = 114;
+    OptimizeColumns optimize_columns = 115;
   }
 
   // An operation to apply to the blob dataset
--- a/rust/lance-core/src/utils/tracing.rs
+++ b/rust/lance-core/src/utils/tracing.rs
@@ -84,3 +84,4 @@ pub const DATASET_DELETING_EVENT: &str = "deleting";
 pub const DATASET_COMPACTING_EVENT: &str = "compacting";
 pub const DATASET_CLEANING_EVENT: &str = "cleaning";
 pub const DATASET_LOADING_EVENT: &str = "loading";
+pub const DATASET_OPTIMIZING_COLUMNS_EVENT: &str = "optimizing_columns";
--- a/rust/lance/src/dataset.rs
+++ b/rust/lance/src/dataset.rs
@@ -69,6 +69,7 @@ mod hash_joiner;
 pub mod index;
 mod metadata;
 pub mod optimize;
+pub mod optimize_columns;
 pub mod progress;
 pub mod refs;
 pub(crate) mod rowids;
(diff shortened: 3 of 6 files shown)


steFaiz:
Thank you for this valuable work! I have a small curiosity—this is not a blocking issue, though.

Would it be possible to directly reuse the Compaction code module? It seems your implementation follows the same workflow as Compaction: Plan + Execution + Commit. My main concern is that Compaction already exposes interfaces at both the Java and Python layers and supports distributed execution. Moreover, engines like Spark and Ray are expected to support distributed compaction in the future. If we could add some parameters to CompactOptions to control column merging, it would minimize changes in the upper layers and keep the codebase cleaner and more consistent.

Also, I think it's reasonable to keep this separate if we think ColumnOptimization is essentially different from Compaction.

xerryd:
Hey @steFaiz, I really appreciate you taking a look! After taking a look at the initial implementation from @geruh I noticed, there was a bunch of overlap between the operations. However, the more I looked at it the more I wanted this to stay a single-purpose operation. Column layout optimization is strictly about reorganizing files horizontally inside a fragment and leaving row ids and indexing untouched keeping things as light weight as can be. 

 Since compaction, on the other hand, rewrites whole fragments and toggles a bunch of other knobs (index remap, row-id, deletion cleanup). Mixing the two felt like we’d be adding side effects to compaction and losing the “do one thing well” property here, so I kept them split. 
 
 I'd be happy to hear more thoughts on this.

xerryd:
I thought about this a bit more, and one thing `compact_columns` really shows is that there’s now a clear capability to group or rewrite files around column families. I think it makes sense to view this as a standalone maintenance operation, since it lays the groundwork for extending writers to support column families down the line.

For example, if we eventually introduce a column family feature for Lance tables, any write operation could leverage that configuration. In that world, compaction could actually respect column families and rewrite table state accordingly. So there’s a pretty clean line between the two operations: this one focuses on lightweight reorganization, while compaction could evolve to build on top of that in the future.

PR #5080, opened 2025-10-28
fix: vector search should respect target fragment

yingjianwu98:
Vector search currently does not respect the target fragment list specified via Scanner.with_fragments().

  Given a dataset with:
  - indexed_fragments = [1, 2, 3]
  - unindexed_fragments = [4, 5]

  Scenario 1: Query targets only unindexed fragments [4, 5]
  - Bug: Still scans and returns results from indexed fragments [1, 2, 3]
  - Expected: Should only scan fragments [4, 5] using flat search

  Scenario 2: Query targets mixed fragments [3, 4]
  - Bug: Scans indexed fragment 3 correctly, but also scans unindexed fragment 5 even though it's not in the target list
  - Expected: Should only scan fragments [3, 4]

### DIFF
--- a/rust/lance/src/dataset/scanner.rs
+++ b/rust/lance/src/dataset/scanner.rs
@@ -2836,7 +2836,23 @@ impl Scanner {
         } else {
             Arc::new(vec![])
         };
-        if let Some(index) = indices.iter().find(|i| i.fields.contains(&column_id)) {
+
+        // TODO: refactor the code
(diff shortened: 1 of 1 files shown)


yingjianwu98:
@westonpace @wjones127 

Sorry for the direct ping, but I believe this bug could potentially have Lance to return results from unintended fragments. This might have broader implications so I want to bring this to your attention, thanks.

yingjianwu98:
closing this in favor of https://github.com/lance-format/lance/pull/5924

Issue #5063, opened 2025-10-24
scanner with full_text_query does not respect fragments restriction


artemru:
a simple repro that produces simple lance dataset with two fragments 
```python
import pyarrow as pa
import lance
from lance.query import MatchQuery
import polars as pl

table1 = pa.Table.from_pydict({"text": list("bcdfghjkl")})
table2 = pa.Table.from_pydict({"text": list("123456")})
lance_ds = lance.write_dataset(table1, "/tmp/tt.lance", mode="overwrite")
lance_ds.insert(table2)
assert len(lance_ds.get_fragments()) == 2

lance_ds.create_scalar_index(
    index_type="INVERTED",
    column="text",
    ascii_folding=False,
    stem=False,
    max_token_length=30,
    base_tokenizer="simple",
    with_position=False,
    remove_stop_words=False,
    replace=True
)

print(pl.from_arrow(lance_ds.scanner(full_text_query=MatchQuery("c", "text"),
                                     fragments=[lance_ds.get_fragments()[1]],
                                     scan_in_order=True,
                                     limit=2, with_row_id=True).to_table()))
```
produces 
<img width="246" height="140" alt="Image" src="https://github.com/user-attachments/assets/772b8c21-54b7-466e-90de-063566d57905" />

however,  `lance_ds.get_fragments()[1]` does not contain `c` !

artemru:
I was trying to do some simple (fragment) distributed queries logic on top of `.scanner(... fragments=...)` API and I'm currently blocked because of that bug.

artemru:
it's fixed as now with lance>=4.0.0 ! 
Closing!

PR #5081, opened 2025-10-28
fix: fts search should respect target fragment for scan

yingjianwu98:
FTS search currently does not respect the target fragment list specified via Scanner.with_fragments().

  Given a dataset with:
  - indexed_fragments = [1, 2, 3]
  - unindexed_fragments = [4, 5]

  Scenario 1: Query targets only unindexed fragments [4, 5]
  - Bug: Still scans and returns results from indexed fragments [1, 2, 3]
  - Expected: Should only scan fragments [4, 5] using flat search

  Scenario 2: Query targets mixed fragments [3, 4]
  - Bug: Scans indexed fragment 3 correctly, but also scans unindexed fragment 5 even though it's not in the target list
  - Expected: Should only scan fragments [3, 4]

### DIFF
--- a/rust/lance/src/dataset/scanner.rs
+++ b/rust/lance/src/dataset/scanner.rs
@@ -2717,6 +2717,28 @@ impl Scanner {
 
         let (match_plan, flat_match_plan) = match &index {
             Some(index) => {
+                // Check if any of the scanner's fragments are covered by the FTS index
+                let indexed_frags_bitmap = index
+                    .fragment_bitmap
+                    .as_ref()
+                    .map(|bitmap| bitmap & &self.get_fragments_as_bitmap())
+                    .unwrap_or_else(|| self.get_fragments_as_bitmap());
+
+                let has_indexed_fragments = !indexed_frags_bitmap.is_empty();
+
+                if !has_indexed_fragments {
+                    // No indexed fragments in the scan - do flat search only on scanner's fragments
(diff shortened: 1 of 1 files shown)


artemru:
fixes https://github.com/lancedb/lance/issues/5063

yingjianwu98:
@westonpace @wjones127 

Sorry for the direct ping, but I believe this bug could potentially have Lance to return results from unintended fragments. This might have broader implications so I want to bring this to your attention, thanks.

rust/lance/src/dataset/scanner.rs
```diff
@@ -2708,14 +2717,44 @@ impl Scanner {
...
                     params.clone(),
                     prefilter_source.clone(),
                 ));
 
-                let unindexed_fragments = self.dataset.unindexed_fragments(&index.name).await?;
+                let mut unindexed_fragments = self.dataset.unindexed_fragments(&index.name).await?;
+
+                // Filter by scanner's fragment list to respect fragment restrictions
+                if let Some(scanner_frags) = &self.fragments {
+                    let scanner_ids: std::collections::HashSet<_> =
```
jackye1995, line 2753: we can use `self.get_fragments_as_bitmap();` in order to perform the same operation instead of using a hashset.

yingjianwu98:
closing this in favor of https://github.com/lance-format/lance/pull/5924

PR #5083, opened 2025-10-28
refactor(python): remove explain_plan api from sql

yanghua:
Since we have removed this api from rust via #4428 . We also need to cleanup the same api in python.


yanghua:
@jackye1995 please take a look, thanks.

yanghua:
duplicated with #5085

Issue #5084, opened 2025-10-28
Missing `explain_plan` pymethod in SqlQuery


ddupg:
Execute the following code will result in an error
```
ds.sql("SELECT * FROM test WHERE id > 1").table_name("test").build().explain_plan(True)
```

```
self = <lance.dataset.SqlQuery object at 0x1122158e0>, verbose = True

    def explain_plan(self, verbose: bool = False) -> str:
        """
        Explain the query plan.
    
        Parameters
        ----------
        verbose: bool, default False
            If True, print the verbose plan.
    
        Returns
        -------
        str
            The query plan.
        """
>       return self._query.explain_plan(verbose)
               ^^^^^^^^^^^^^^^^^^^^^^^^
E       AttributeError: '_lib.SqlQuery' object has no attribute 'explain_plan'

../lance/dataset.py:3592: AttributeError

```

westonpace:
You can also do `ds.sql("EXPLAIN SELECT * FROM test WHERE id > 1")` but the returned explain will be a slightly different format.

ddupg:
> You can also do `ds.sql("EXPLAIN SELECT * FROM test WHERE id > 1")` but the returned explain will be a slightly different format.

@westonpace Yes, I got this approach. The current bug is that the Python method exists, but the corresponding Rust method is missing.

PR #5085, opened 2025-10-28
fix: remove remainder explain_plan method in Python

ddupg:
Closes: #5084

### DIFF
--- a/python/python/lance/dataset.py
+++ b/python/python/lance/dataset.py
@@ -3575,22 +3575,6 @@ class SqlQuery:
         """
         return self._query.to_stream_reader()
 
-    def explain_plan(self, verbose: bool = False) -> str:
-        """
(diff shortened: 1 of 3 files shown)


westonpace:
It looks like this was intentionally removed in https://github.com/lancedb/lance/pull/4428

I'm not sure I have a strong opinion either way.  CC @wojiaodoubao @jackye1995 @wjones127 from the other PR.

ddupg:
> It looks like this was intentionally removed in #4428
> 
> I'm not sure I have a strong opinion either way. CC @wojiaodoubao @jackye1995 @wjones127 from the other PR.

So we should also remove the remainder Python code, right? I'll make the changes.

yanghua:
duplicated with #5083 , but never mind, it's yours.

Issue #5073, opened 2025-10-27
Add docs for branch


majin1102:
Sections should be updated:

1. README.md
2. docs/src/quickstart/versioning.md
3. docs/src/guide

PR #5088, opened 2025-10-28
feat: support global tag retrieval and improve tag api

majin1102:
When I was planning the work #5073 . I found some uncompleted work:

1. Tag retrieval doesn't contain branch field both in Java and Python.
2. The reference parameter is not well used in scenarios including createTag, updateTag. And createBranch both in Java and Python.
3. For python, Tuple[str, int] is not the best type and lead to the api of checkout_branch. Use Tuple[Optional[str], Optional[int]] instead for more syntax expressions and remove checkout_branch.
4. Some necessary refactoring like removing all ref_

### DIFF
--- a/python/python/tests/test_dataset.py
+++ b/python/python/tests/test_dataset.py
@@ -4821,7 +4821,7 @@ def test_shallow_clone(tmp_path: Path):
 
     table_v3 = pa.table({"a": [7, 8, 9], "b": [40, 50, 60]})
     branch = ds.create_branch("branch", 2)
-    lance.write_dataset(table_v3, branch.uri(), mode="overwrite")
+    lance.write_dataset(table_v3, branch.uri, mode="overwrite")
     clone_branch_v3 = tmp_path / "clone_branch_v3"
     cloned_by_branch = branch.shallow_clone(clone_branch_v3, 3)
     assert cloned_by_branch.to_table() == table_v3


majin1102:
There's another issue that I don't feel right but might need some discussion.

For now we use None to present the main branch. For example, we are in branch A:10,  if we want to checkout A:9, we need to use `checkout_version(('A', 9))`. This is all right if user knows he is on branch A, however, if he load the dataset by a branch uri, he just treats the uri as an independent dataset(at least the experience). He might just use checkout_version(9) or checkout((None, 9)). But the actual result is he checked out the main.

I think most cases when people see None, maybe they are expecting no changes. So I think we might should to present None as the current branch instead of the main. On the other hand, we could use a constant string 'main' to present the main branch since we have reserved it. Although We store None for the main in the spec,  we could provide better API for this.

How do you feel about this? @jackye1995  If you agree I could do this in this PR

jackye1995:
Looks like there are 2 things here, 

(1) feels like from what you are describing that putting "None" is confusing, maybe we should just reserve the name `main` to mean the main branch, so user can use that to reference the main branch. (they can continue to use None to also mean main branch that is fine)

(2) I agree that checkout_version(version_number) should better be against the current branch, we can improve that experience.

majin1102:
> Looks like there are 2 things here,
> 
> (1) feels like from what you are describing that putting "None" is confusing, maybe we should just reserve the name `main` to mean the main branch, so user can use that to reference the main branch. (they can continue to use None to also mean main branch that is fine)
> 
> (2) I agree that checkout_version(version_number) should better be against the current branch, we can improve that experience.

Sorry I think I didn't make myself clear enough.

If we use `checkout_version(version_number)` the version_number would be a `impl Into<Ref>` and that would be (None, Some(version_number)) pointing to the main branch. So I guess 1) and 2) are somehow conflicted:

```
    pub async fn checkout_version(&self, version: impl Into<refs::Ref>) -> Result<Self> {
        let reference: refs::Ref = version.into()
        ............
```

We can transform an u64 to (current_branch, Some(version_number)), but will make the interface complicated(the Into trait doesn't have current branch, so we might need to split the api). If we make (None, Some(version_number)) just point to the current branch, it will make sense to the `impl Into<Ref>`

This is just an impl issue, but I think we better make these all matched and clear. WDYT? @jackye1995

java/src/main/java/com/lancedb/lance/Dataset.java
```diff
@@ -1028,29 +1066,29 @@ public void delete(String tag) {
...
      * @param tag the tag name
-     * @param versionNumber the version number to tag
+     * @param versionNumber the versionNumber on main.
      */
-    public void update(String tag, long versionNumber) {
-      try (LockManager.WriteLock writeLock = lockManager.acquireWriteLock()) {
-        Preconditions.checkArgument(nativeDatasetHandle != 0, "Dataset is closed");
-        nativeUpdateTag(tag, versionNumber);
-      }
+    public void update(String tag, int versionNumber) {
```
jackye1995: should be long

python/python/lance/dataset.py
```diff
@@ -2116,7 +2094,9 @@ def latest_version(self) -> int:
         """
         return self._ds.latest_version()
 
-    def checkout_version(self, version: int | str | Tuple[str, int]) -> "LanceDataset":
+    def checkout_version(
+        self, version: int | str | Tuple[Optional[str], [int]]
```
jackye1995: Optional[int]

python/src/dataset.rs
```diff
@@ -2684,21 +2677,10 @@ impl Dataset {
...
-                } else if len == 2 {
-                    let (branch_name, version_number) = tuple.extract::<(String, u64)>()?;
-                    Ok(Ref::Version(Some(branch_name), Some(version_number)))
+                if tuple.len() == 2 {
+                    let (branch_name, version_number) =
+                        tuple.extract::<(Option<String>, Option<u64>)>()?;
+                    Ok(Ref::Version(branch_name, version_number))
                 } else {
                     Err(PyValueError::new_err(
                         "Version tuple must have 1 or 2 elements",
```
jackye1995: this should be always exactly 2?

java/src/main/java/org/lance/Dataset.java
```diff
@@ -989,29 +1027,29 @@ public Map<String, String> getTableMetadata() {
...
     }
 
     /**
      * Create a new tag on a specified branch.
      *
      * @param tag the tag name
-     * @param versionNumber the version number to tag
+     * @param ref the referenced version to tag
      */
-    public void create(String tag, long versionNumber, String targetBranch) {
```
jackye1995, line 1225: should start to add deprecation notes like @Deprecated("will be removed in 2.0.0, please use xxx instead")

majin1102:
Hi,  @jackye1995 

I think there's a bug for current branch datasets. As we know we now use (str, u64) to identify a global version. if we are working on a branch and use interfaces like this:

```
    pub async fn read_transaction_by_version(&self, version: u64) -> Result<Option<Transaction>> {
        let dataset_version = self.checkout_version(version).await?;
        dataset_version.read_transaction().await
    }
```

We will redirect to the main branch and get the transaction in that version line. On the other hand, everytime we use checkout_verson by a version_number, we are checking out the main branch which might be probably unexpected.

I'm solving this issue by introducing a `VersionNumber` enumation which is binded to the current branch context. In the meantime, we keep `None` as the main branch.

Please take a look on the newsest code when you have time

python/src/dataset.rs
```diff
@@ -2860,25 +2841,12 @@ impl Dataset {
...
-                    Ok(Ref::Version(Some(branch_name), Some(version_number)))
+                if tuple.len() == 2 {
+                    let (branch_name, version_number) =
+                        tuple.extract::<(Option<String>, Option<u64>)>()?;
+                    Ok(Ref::Version(branch_name, version_number))
                 } else {
-                    Err(PyValueError::new_err(
-                        "Version tuple must have 1 or 2 elements",
-                    ))
+                    Err(PyValueError::new_err("Version must be a (int, str) tuple"))
```
yanghua: (str, int) mapping to `<branch, version>` Looks better?

python/src/dataset.rs
```diff
@@ -2860,25 +2841,12 @@ impl Dataset {
                 let tag: &str = &tag_name.to_string_lossy();
                 Ok(Ref::from(tag))
             } else if let Ok(tuple) = reference.downcast_bound::<PyTuple>(py) {
-                let len = tuple.len();
-                if len == 1 {
```
yanghua, line 2864: Why delete the condition when `len == 1`? Can we keep one value to make it more scalable? For example:

If only `version num`, we suppose it is on the  `main` branch.

python/python/lance/dataset.py
```diff
@@ -588,7 +588,7 @@ def branches(self) -> "Branches":
     def create_branch(
```
yanghua, line 588: What about moving this method into the `Branch` class to be a `create` method?

python/python/tests/test_dataset.py
```diff
@@ -4778,17 +4805,17 @@ def test_shallow_clone(tmp_path: Path):
     ds = lance.write_dataset(table_v2, src_dir, mode="overwrite")
 
     # Create a tag pointing to version 1
-    ds.tags.create("v1", 1, None)
+    ds.tags.create("v1", (None, 1))
```
yanghua: Add a case to verify if we are not in the `main` branch, but create a tag point to the main branch?

python/python/tests/test_dataset.py
```diff
@@ -4846,7 +4873,7 @@ def test_branches(tmp_path: Path):
     assert "branch1" not in branches_after
     assert "branch2" in branches_after
 
-    branch2 = ds_main.checkout_branch("branch2")
+    branch2 = ds_main.checkout_version(("branch2", None))
```
yanghua: Add some cases to verify checkout not-existed branch/tag/version?

python/python/lance/dataset.py
```diff
@@ -588,7 +588,7 @@ def branches(self) -> "Branches":
...
         self,
         branch: str,
-        reference: Optional[int | str | Tuple[str, int]] = None,
+        reference: Optional[int | str | Tuple[Optional[str], Optional[int]]] = None,
```
yanghua, line 591: Should the doc also need to be updated?

python/python/lance/dataset.py
```diff
@@ -2223,7 +2201,9 @@ def latest_version(self) -> int:
         """
         return self._ds.latest_version()
 
-    def checkout_version(self, version: int | str | Tuple[str, int]) -> "LanceDataset":
+    def checkout_version(
+        self, version: int | str | Tuple[Optional[str], Optional[int]]
```
yanghua, line 2206: ditto, update the doc about the arg type?

python/python/lance/dataset.py
majin1102: That's because create_branch includes two steps:

1. shallow_clone from the source branch
2. create branch metadata a.k.a BranchContents

The refs.rs API is only response for metadata and the shallow_clone operation needs to be done within datase(with a potential storage_options). That's why we provide a create_branch in dataset instead of Branches. And I mark the related create in Branches as `crate` as it shouldn't be used by users.

python/src/dataset.rs
majin1102: I guess you are saying the first if?

```
         if let Some(reference) = reference {
            if let Ok(i) = reference.downcast_bound::<PyInt>(py) {
                let version_number: u64 = i.extract()?;
                Ok(Ref::from(version_number))
            } else if let Ok(tag_name) = reference.downcast_bound::<PyString>(py) {
                ......
            } else if let Ok(tuple) = reference.downcast_bound::<PyTuple>(py) {
                ........
            } else {
                Err(PyValueError::new_err(
                    "Version must be an int, a str or a (int, str) tuple.",
                ))
            }
```

I delete this because I original thought we might have some case like (version_number) or (branch_name). But this is over complicated and not necessary.

> If only version num, we suppose it is on the main branch.

We can already do that by now.

python/python/tests/test_dataset.py
majin1102: This is the shallow_clone testcase.
I think we can add a testcase to shallow_clone a branch.

> Add a case to verify if we are not in the main branch, but create a tag point to the main branch?

Let me add your case to test_branches

majin1102:
Comments addressed, please take a look @yanghua

yanghua (approved):
+1

majin1102:
@jackye1995 Please let me know if you have any further comments

PR #5089, opened 2025-10-28
ci!: move to semantic versioning release mechanism

jackye1995:
This PR moves the release mechanism to follow the semantic versioning standard with a streamlined RC voting process and automated breaking change detection. See `release_process.md` for the updated workflow and examples.

### DIFF
--- a/release_process.md
+++ b/release_process.md
@@ -94,9 +94,9 @@ Lance uses semantic versioning with prerelease identifiers:
 flowchart LR
     subgraph main["Main Branch"]
         direction LR
-        M0["1.3.0-beta.2<br/>📍 release-root/1.3.0-beta.N<br/>📍 release-root/1.4.0-beta.N"] --> M1["1.4.0-beta.0"]
-        M1 --> M2["1.4.0-beta.1<br/>🏷️ v1.4.0-beta.1<br/>📍 release-root/1.5.0-beta.N"]
-        M2 --> M3["1.5.0-beta.0"]
+        M0["1.3.0-beta.2<br/>📍 release-root/1.4.0-beta.N<br/>📍 release-root/2.0.0-beta.N"] --> M1["1.4.0-beta.0"]
+        M1 --> M2["1.4.0-beta.1<br/>🏷️ v1.4.0-beta.1"]
+        M2 --> M3["2.0.0-beta.1<br/>🏷️ v2.0.0-beta.1"]
     end
 
     subgraph release["Release Branch: release/v1.3"]
@@ -112,10 +112,10 @@ flowchart LR
 ```
 
 **Flow explanation:**
-- **Main branch**: Commit M0 at `1.3.0-beta.2` has two release-root tags: `release-root/1.3.0-beta.N` (from previous RC) and `release-root/1.4.0-beta.N` (created when cutting v1.3.0-rc.1) → M1 bumps to `1.4.0-beta.0` (unreleased) → M2 publishes `1.4.0-beta.1` (preview, tagged) and gets `release-root/1.5.0-beta.N` tag when cutting v1.4.0-rc.1 → M3 bumps to `1.5.0-beta.0`
+- **Main branch**: Commit M0 at `1.3.0-beta.2` has `release-root/1.4.0-beta.N` (created when cutting v1.3.0-rc.1, pointing to this commit) and `release-root/2.0.0-beta.N` (created when breaking changes bumped major version, pointing to same commit) → M1 bumps to `1.4.0-beta.0` (unreleased) → M2 publishes `1.4.0-beta.1` (preview, tagged) → M3 publishes `2.0.0-beta.1` after detecting breaking changes (tagged)
 - **Release branch** `release/v1.3` created from M0, starts at `1.3.0-rc.1` (tagged) → `1.3.0` (stable, tagged) → `1.3.1-beta.0` → `1.3.1-rc.1` (tagged) → `1.3.1` (stable, tagged) → `1.3.2-beta.0`
 - **Tags**: 🏷️ = version tag (points to tagged commit), 📍 = release-root tag (points to commit before RC was created, used for breaking change detection)
-- **Breaking changes**: If breaking changes detected when cutting RC from M0, the RC would be `2.0.0-rc.1` and main would bump to `2.1.0-beta.0`. M0 would then have `release-root/2.0.0-beta.N` instead of `release-root/1.4.0-beta.N`
+- **Breaking changes**: Both `release-root/1.4.0-beta.N` and `release-root/2.0.0-beta.N` point to M0 (same commit), showing that 2.0.0 is a major version bump from the 1.3.0-rc.1 baseline
 
 **Note**: All commits are linear on their respective branches. `beta.0` = unreleased, `beta.1+` = published previews.
 


release_process.md
```diff
@@ -1,162 +1,453 @@
...
+                       ↓                      |
+release/v1.3:    1.3.0 ────────────────────────
+                   |
+                   ↓
+              1.3.1-beta.1 → ... (for patches)
+
+Note: RC commits (shown in [brackets]) are detached commits tagged for voting.
+      Stable releases are derived from the original beta commits, not RC commits.
+      This ensures clean git history with stable releases as direct descendants of development.
+```
```
wjones127: Would be nice to have this as a mermaid diagram.
````suggestion
```mermaid
gitGraph
    commit id: "Beta release 1" tag: "1.3.0-beta.1"
    commit
    commit id: "RC release 1" tag: "1.3.0-rc.1"
    commit id: "RC release 2" tag: "1.3.0-rc.2"
    branch "release/v1.3"
    checkout "release/v1.3"
    commit id: "Stable release 1" tag: "1.3.0" type: HIGHLIGHT
    checkout main
    commit id: "critical fix"
    checkout "release/v1.3"
    cherry-pick id: "critical fix"
    commit id: "Hotfix release" tag: "1.3.1" type: HIGHLIGHT
    checkout main
    commit
    commit id: "Beta release 3" tag: "1.4.0-beta.1"
    commit id: "RC release 3" tag: "1.4.0-rc.1"
    branch "release/v1.4"
    checkout "release/v1.4"
    commit id: "Stable release 2" tag: "1.4.0" type: HIGHLIGHT
```

> [!NOTE]
> RC commits (shown in [brackets]) are detached commits tagged for voting.
> Stable releases are derived from the original beta commits, not RC commits.
> This ensures clean git history with stable releases as direct descendants of development.
````

File: release_process.md [hunk 2]
```diff
@@ -1,162 +1,453 @@
...
+              (detached commit)               |
+                       |                      |
+                       | (promote: derive from original beta commit)
+                       ↓                      |
+release/v1.3:    1.3.0 ────────────────────────
+                   |
+                   ↓
+              1.3.1-beta.1 → ... (for patches)
+
+Note: RC commits (shown in [brackets]) are detached commits tagged for voting.
```
wjones127: Why not have the RC commits live on the release branch? That's where I would expect them to exist.

release_process.md
```diff
@@ -1,162 +1,453 @@
...
+- `.bumpversion.toml` - Version management configuration
+- Supports beta, rc, and stable version formats
+- Synchronizes versions across 50+ files
+- **Version Numbering**: Beta and RC versions start from 1 (e.g., `1.3.0-beta.1`, `1.3.0-rc.1`)
+
+## Example Workflows
+
+### Standard Major/Minor Release
+
+```bash
```
wjones127, line 337: Maybe?
````suggestion
```markdown
````

File: ci/bump_main_after_major_minor_release.sh
Comment (wjones127) [inline ci/bump_main_after_major_minor_release.sh:1]: Overall, this seems like too many scripts and workflows. Can we consolidate this into a single script?

```shell
python ci/bump_version.py beta
python ci/bump_version.py rc
python ci/bump_version.py stable
```

ci/promote_major_minor_rc.sh
```diff
@@ -0,0 +1,81 @@
...
+RC_VERSION=$(echo "${RC_TAG}" | sed "s/^${TAG_PREFIX}//")
+STABLE_VERSION=$(echo "${RC_VERSION}" | sed 's/-rc\.[0-9]*$//')
+
+echo "Stable version will be: ${STABLE_VERSION}"
+
+# Get the original commit that was used to create the RC
+# The RC tag points to a commit with RC version
+# Its parent is the original commit with beta version
+RC_COMMIT=$(git rev-list -n 1 "${RC_TAG}")
+ORIGINAL_COMMIT=$(git rev-list -n 1 "${RC_COMMIT}^")
```
wjones127: Why not just create the release off of the RC commit? That seems like it would be more direct.

ci/bump_main_after_major_minor_release.sh
Xuanwo: Yep, I shared the same concern.

jackye1995: The issue is that we have now a few different workflows that do some overlapping but overall quite different steps, if you read the release process doc for different types of releases. I can definitely squeeze everything to one or two scripts, but I feel having a 1-1 relationship with the GitHub workflow and match what user would eventually do seems to make more sense.

When Will commented on this I was trying to just split all steps to different scripts and that was definitely too much, but this current state is already a version that tries to reuse things as much as possible.

ci/promote_major_minor_rc.sh
jackye1995: Updated to do that

release_process.md
jackye1995: Updated to do that

release_process.md
jackye1995: Updated to use mermaid diagram

release_process.md
```diff
@@ -1,162 +1,550 @@
...
+Lance uses a semantic versioning release workflow with Release Candidates (RC) for voting before stable releases. The process features linear commit history on branches.
+
+## Overview
+
+The release process follows these stages:
+1. **Development** on `main` branch with beta versions
+2. **Preview Releases** published from beta versions (beta.1+)
+3. **Release Candidates** created for voting (rc.1, rc.2, etc.)
+4. **Stable Releases** promoted from approved RCs
+5. **Patch Releases** managed through release branches
```
westonpace: It's not clear here (and maybe it doesn't need to be) if the release branches are introduced at the release candidates stage or the patch releases stage.

wjones127: Here's a suggested summary:
````suggestion
Lance uses [semantic versioning](https://semver.org/) and maintains a linear commit history.

* All pull requests are merged into the `main` branch.
* Beta releases (or preview releases) are created on-demand from the `main` branch.
* Stable releases (non-prereleases) are created only after a voting process and come from a release branch `release/vX.Y`. These are typically created once every two weeks.
* Release Candidates (RC) are created from release branches prior to voting.
* Patch releases are created by cherry-picking fixes from `main` into the release branch, voting on a new RC, and releasing.

```mermaid
gitGraph
    commit
    branch feature
    checkout feature
    commit
    checkout main
    merge feature
    branch bugfix
    checkout bugfix
    commit id: "bugfix"
    checkout main
    branch "release/v1.4"
    checkout "release/v1.4"
    commit tag: "1.4.0-rc.1" tag: "1.4.0"
    checkout main
    merge bugfix
    commit id: "merged"
    checkout "release/v1.4"
    cherry-pick id: "merged"
    commit tag: "1.4.1-rc.1" tag: "1.4.1"
    checkout main
    commit tag: "1.5.0-beta.1"
    
```
````

File: release_process.md [hunk 6]
```diff
@@ -1,162 +1,550 @@
...
+- **Rust**: Not published (use git tag)
+- **Python**: fury.io
+- **Java**: Maven Central (pre-release)
+- **Protobuf**: Buf Schema Registry (pre-release)
+
+**Beta Releases (`-beta.N`)**:
+- **Rust**: Not published (use git tag)
+- **Python**: fury.io
+- **Java**: Maven Central (pre-release)
+- **Protobuf**: Buf Schema Registry (pre-release)
```
wjones127: Let's make this a markdown table:


```suggestion

| Language | Stable release | RC release | Beta release |
|----------|----------------|------------|--------------|
| Rust     | [crates.io](https://crates.io/crates/lance) | Tag on GitHub [^1] | Tag on GitHub [^1] |
| Python   | [PyPI](https://pypi.org/project/pylance/) [^2] | fury.io [^3] | fury.io [^3] |
| Java | [Maven Central](https://mvnrepository.com/artifact/com.lancedb/lance-core) | [Maven Central](https://mvnrepository.com/artifact/com.lancedb/lance-core) | [Maven Central](https://mvnrepository.com/artifact/com.lancedb/lance-core) |
| Protobuf | [Buf Schema Registry](https://buf.build/lancedb/lance) | [Buf Schema Registry](https://buf.build/lancedb/lance) | [Buf Schema Registry](https://buf.build/lancedb/lance) |

[^1]: Since Rust is a source-only release, with no built artifacts, we just publish non-stable releases as source code on GitHub.
[^2]: Note: there is another project called "lance" on PyPI, but it is not affiliated with Lance. https://pypi.org/project/lance/
[^3]: To install from fury.io, pass `--pre --extra-index-url https://pypi.fury.io/lancedb/`
```

release_process.md
```diff
@@ -1,162 +1,550 @@
...
+
+### GitHub Releases and Release Notes
+
+**RC Releases (`-rc.N`)**:
+- Created as GitHub Pre-Releases with custom-generated release notes
```
wjones127: I find "custom-generated" release notes unclear. Does that mean a person wrote them? Or a script?

I think the goal of this section should be to make clear how and when release notes are created. If there is a script that generates them, we should link to that.

release_process.md
```diff
@@ -1,162 +1,550 @@
...
+
+    M0 -.->|"create RC<br/>(no breaking changes)"| R1
+```
 
-We create a full release of Lance up to every 2 weeks. In between full releases,
-we make preview releases of the latest features and bug fixes, which are hosted
-on fury.io. This allows us to release frequently and get feedback on new features
-while keeping under the PyPI project size limits.
+**Flow explanation:**
+- **Main branch** at `1.3.0-beta.2` (has release-root tag) → create RC checks for breaking changes → if none, RC is `1.3.0-rc.1`, main bumps to `1.4.0-beta.0` with new release-root tag → `1.4.0-beta.1` (preview, tagged) → create RC for v1.4.0 → main bumps to `1.5.0-beta.0`
```
wjones127: What did you think of my earlier idea of making RC's just tags, but not reflected in the `version` field?

So instead, the release branch immediately changes version to stable `1.3.0`, but tags the commit with `v1.3.0-rc.1`. Once the vote passes, then add another tag `v1.3.0` and trigger the final release.

Is that just hard to do with the current release artifacts?

release_process.md
```diff
@@ -1,162 +1,550 @@
...
-manual version updates. The workflow handles version bumping, breaking change 
-validation, and release creation automatically.
+## Workflows
 
-## Choosing a full versus preview release
+### User-Facing Workflows
 
-There are three conditions that can trigger a full release:
+1. **publish-beta.yml** - Publish beta preview releases from any branch
+2. **create-major-minor-rc.yml** - Create initial RC for new major/minor version
```
wjones127: Could we instead call this `create-release-branch`? I think that's more the action I would think of wanting to do. "create-major-minor-rc" feels more wordy and unclear.

release_process.md
```diff
@@ -1,162 +1,550 @@
...
+3. **create-rc.yml** - Create RC on existing release branch (for new patch release RC or iterations of an existing RC)
+4. **promote-rc-to-stable.yml** - Promote any RC to stable (works for all release types)
```
wjones127: Could be renamed to "approve RC".
```suggestion
4. **approve-rc.yml** - Promote any RC to stable (works for all release types)
```

release_process.md
```diff
@@ -1,162 +1,550 @@
...
-2. We want to make a release of LanceDB that requires new features from Lance
-   (LanceDB can't depend on a preview release of Lance)
-3. It's been two weeks since we last released a full release.
+## Beta Preview Releases
 
-Otherwise, we should make a preview release.
+To publish preview releases for testing before creating RCs:
 
-## Make a preview release
+### Publish Beta Preview from Main
```
wjones127: The following sections are really wordy. I think we should try to consolidate into three brief step-by-step guides:

1. Create a beta / preview release
2. Create a major / minor release
3. Create a patch / bugfix release

If we can, let's put unnecessary detail in a `<details>` block that's collapsed by default.

release_process.md
jackye1995: > What did you think of my earlier idea of making RC's just tags, but not reflected in the version field?

Yeah I was implementing that but I had to revert to have dedicated RC commits. The main reason is basically the release artifacts. I want to be able to actually publish the python/java artifacts to fury and MavenCentral that corresponds to the version, and then we can ask codex to create a branch for testing that rc in lancedb, as well as internally. If I just tag it, then the artifact will have name `1.3.0` directly, it would cause potential issue with resolution if someone configures both the public and fury repo (like geneva), and also if we have multiple RCs then they will end up publishing to the same version, which I think would fail? Even if it can be overwritten it might cause confusion down the road if some artifacts in some platforms fail and the artifacts are in a mixed state.

I also experimented with making the branch just holding beta releases, so each RC just bumps up the beta release version, and we just tag a specific beta release is for RC.N (e.g. 1.3.0-beta.2 is for rc.1). That would solve the issue above, but I ended up feeling if we are doing that mapping of which RC maps to which beta release, it's probably cleaner to just have a dedicated rc commit.

Let me know what you think about it and if there are better ways

release_process.md
```diff
@@ -1,162 +1,550 @@
...
+Lance uses semantic versioning with prerelease identifiers:
+- **Stable**: `X.Y.Z` (e.g., `1.3.0`)
+- **Beta**: `X.Y.Z-beta.N` (e.g., `1.3.0-beta.1`, `1.3.0-beta.2`)
+- **RC**: `X.Y.Z-rc.N` (e.g., `1.3.0-rc.1`, `1.3.0-rc.2`)
+
+### Beta Version States
+
+- **beta.0**: Unreleased version (exists on branch but not published)
+  - Created after cutting an RC to mark the next unreleased version
+  - Indicates no preview has been published yet
```
westonpace, line 51: Do we need a beta.0?  Is this an actual tag or branch?  Or is it just a placeholder in Cargo.toml when something hasn't been released?

release_process.md
```diff
@@ -1,162 +1,550 @@
...
+5. **Patch Releases** managed through release branches
+
+## Version Semantics
+
+### Version Format
+
+Lance uses semantic versioning with prerelease identifiers:
+- **Stable**: `X.Y.Z` (e.g., `1.3.0`)
+- **Beta**: `X.Y.Z-beta.N` (e.g., `1.3.0-beta.1`, `1.3.0-beta.2`)
+- **RC**: `X.Y.Z-rc.N` (e.g., `1.3.0-rc.1`, `1.3.0-rc.2`)
```
westonpace, line 45: What does it mean if there is both a `1.3.0-beta.2` and a `1.3.0-rc.2`?  Is it safe to assume the `rc` comes after the `beta`?

release_process.md
```diff
@@ -1,162 +1,550 @@
...
+- Notes comparison:
+  - **Main branch**: Compares against `release-root/X.Y.Z-beta.N` tag (shows all changes on main since the RC was cut)
+  - **Release branch**: Compares against previous stable tag `vX.Y.(Z-1)` (shows changes since last stable)
+
+## Branching Strategy
+
+### Main Branch
+- Always contains the latest development work
+- Version format: `X.Y.Z-beta.N`
+- After RC creation, bumped to next minor version with `-beta.0` (unreleased)
```
westonpace, line 81: If I look in `main` and I look at the `Cargo.toml` do I see "the version most recently released"?  Or do I see "the next version to be released?"  Or do I see something else?

westonpace (approved):
Approve of `release_process.md`

release_process.md
jackye1995: Do we want to make "These are typically created once every two weeks." more specific like minor version typically every two weeks, and major version typically every 2 months?

release_process.md
jackye1995: The key is just from which commit to which commit should the release note cover. I changed it to a markdown table as well, let me know if that looks more clear

release_process.md
wjones127: I think this seems fine. We might find a way later to simplify in practice. I agree that the artifacts make it necessary to do the RCs for now.

release_process.md
wjones127: Honestly the major version might be more often, as I realized DataFusion often does major releases once a month. I say let's wait on giving specifics for major releases until we see how often it becomes in practice.

release_process.md
jackye1995: yeah it's just a placeholder for unreleased content since the last release branch cut. We always first commit the version bump, and then tag on that bumped version, and then trigger all the publish workflows with it, so there won't really be beta.0 release version, the first beta version will always be x.y.z-beta.1.

An alternative could be to reverse the order so we tag the latest main first and then bump version. But that created other complications in detecting breaking changes because you could be in x.y.z-beta.1 and have to tag it with a (x+1).0.0-beta.1 because it actually contains breaking changes. So I ended up using this beta.0 approach that feels the most logical when user sees the code at any commit version.

release_process.md
jackye1995: Also updated the patch release case as "Patch releases are created by committing fixes directly to the release branch, voting on a new RC, and releasing." since there could be cases that we commit a simpler fix in the release branch, and do a larger refactoring in the main branch.

jackye1995: superseded by https://github.com/lancedb/lance/pull/5089#discussion_r2495914866 which mentions the release branch

release_process.md
jackye1995: Yes, because when there is `1.3.0-rc.2` it means we already have cut a release branch `release/1.3` and the main will be bumped to 1.4.0-beta.0 or 2.0.0-beta.0.

release_process.md
jackye1995: If you see `-beta.0` that means there was just a release branch cut, and this is there is no any new beta releases or new branch cuts yet.

After a first beta release, you see the version of the beta release that was released previously.

The main reasons I did not end up using the released version but use the next version + `-beta.0` are:
1. for beta releases, you see the version and you know this is that release + some new commits. But for the released version, because of the divergence after the branch cut, the content in main is no longer exactly the released version + some new commits, so I wanted to make that distinction. 
2. if I use the released version on main as the new version after cutting the branch, then I need to handle one more case both when I cut the next release branch and when I do a beta release, which becomes quite complicated compared to the current setup.
3. in this setup basically when people build any commit that is not tagged as stable, it will be shown as some sort of beta or rc. When it is recorded in the writer version we will have an idea that it is building some version off the main branch instead of using the released version.

release_process.md
```diff
@@ -1,162 +1,475 @@
...
-
-### Automated Version Bumping
+# Release Process
+
+Lance uses [semantic versioning](https://semver.org/) and maintains a linear commit history.
+* All pull requests are merged into the `main` branch.
+* Beta releases (or preview releases) are created on-demand from the `main` branch.
+* Stable releases (non-prereleases) are created only after a voting process and come from a release branch `release/vX.Y`. These are typically created once every two weeks.
+* Release Candidates (RC) are created from release branches prior to voting.
+* Patch releases are created by committing fixes directly to the release branch, voting on a new RC, and releasing.
```
Xuanwo, line 8: Should we always require a commit on main first, then backport to the release branch?

release_process.md
Xuanwo: I'm thinking of always use `1.3.0-dev` on main and only update the version while doing release. So

- `dev` version never been released.
- we don't need to have a `beta.0`.

release_process.md
```diff
@@ -1,162 +1,475 @@
...
+  - Created after cutting an RC to mark the next unreleased version
+  - Indicates no preview has been published yet
+- **beta.1+**: Published preview releases
+  - Created when publishing beta preview artifacts
+
+### Publishing Channels
+
+| Language     | Stable release      | RC release                        | Beta release                      |
+|--------------|---------------------|-----------------------------------|-----------------------------------|
+| **Rust**     | crates.io           | Not published (use git tag)       | Not published (use git tag)       |
```
Xuanwo: I'm thinking of always release them so downstream can just use version instead of playing with git tag.

release_process.md
```diff
@@ -1,162 +1,475 @@
...
+
+### GitHub Releases and Release Notes
+
+| Release Type              | GitHub Release Type | Start Commit (exclusive)    | End Commit (inclusive) | Explanation                                                          |
+|---------------------------|---------------------|-----------------------------|------------------------|----------------------------------------------------------------------|
+| **RC (Major/Minor)**      | Pre-Release         | `release-root/X.Y.0-beta.N` | `vX.Y.0-rc.N`          | All changes for the release                                          |
+| **RC (Patch)**            | Pre-Release         | `vX.Y.(Z-1)`                | `vX.Y.Z-rc.N`          | Only changes in this patch release                                   |
+| **RC (Iterations)**       | Pre-Release         | `vX.Y.(Z-1)`                | `vX.Y.Z-rc.N`          | Only changes in this patch release (not changes against previous RC) |
+| **Stable (Major/Minor)**  | Release             | `release-root/X.Y.0-beta.N` | `vX.Y.0`               | All changes from main + RC fixes                                     |
```
Xuanwo: Let's place stable release first: stable -> rc -> beta

release_process.md
```diff
@@ -1,162 +1,475 @@
...
+- Always contains the latest development work
+- Version format: `X.Y.Z-beta.N`
+- After RC creation, bumped to next minor version with `-beta.0` (unreleased)
+  - Beta previews published by bumping to `-beta.1+`
+
+### Release Branches
+- Format: `release/v{major}.{minor}` (e.g., `release/v1.3`)
+- Created when cutting initial RC for major/minor release
+- Maintained for patch releases
+- Version progression: `rc.1` → `rc.2` → stable → `beta.0` → `rc.1` (for patches)
```
Xuanwo, line 88: The same: if we are using `v1.3.0-dev` always, we don't need to commit `rc.1`, `beta.0` in repo.

release_process.md
jackye1995: Okay I think based on the latest discussion "* Patch releases are created by cherry-picking fixes from `main` into the release branch, voting on a new RC, and releasing." is the expected behavior, changed back to that, directly patching release is unrecommended except for emergent cases, but we don't need to write about exceptions

release_process.md
jackye1995: Discussed a bit, landed with let's use git for now, publishing everything could make the page look messy, also we don't want to make beta releases too easily consumable that people start to use that in prod unintentionally.

release_process.md
jackye1995: Updated with that

release_process.md
jackye1995: this is basically the same issue with https://github.com/lancedb/lance/pull/5089#discussion_r2496268515, that it would mess up with release artifacts.

jackye1995:
Thanks everyone for the review!

PR #5091, opened 2025-10-29
feat(python): expose DatasetDeltaBuilder and relevant apis

yanghua:


python/python/lance/dataset.py
```diff
@@ -3585,6 +3585,41 @@ def sql(self, sql: str) -> "SqlQueryBuilder":
         """
         return SqlQueryBuilder(self._ds.sql(sql))
 
+    def delta(self) -> "DatasetDeltaBuilder":
```
jackye1995: I think python don't really have builders, what about we just have all the builder inputs as function arguments?

yanghua: After thinking about this suggestion. There are two opinions:

* There are some config options that have some check logic in `build` fn:

https://github.com/lance-format/lance/blob/b21872585560cc4538a1f1977d5ac89f799b75c2/rust/lance/src/dataset/delta.rs#L91

Reusing the `DatasetDeltaBuilder` so that we can reuse the check logic?

* The build pattern could keep the scalarity for the future.

WDYT?

jackye1995: what about having the inputs in the function, so it's functional style, and then we internally call into the binded builder within the function? So the DatasetDeltaBuilder is just internal usage and can be `_DatasetDeltaBuilder`

python/python/lance/dataset.py
```diff
@@ -3585,6 +3585,104 @@ def sql(self, sql: str) -> "SqlQueryBuilder":
...
+
+        builder = _DatasetDeltaBuilder(self._ds.delta())
+
+        if has_compared_against:
+            builder = builder.compared_against_version(compared_against)
+        else:
+            builder = builder.with_begin_version(begin_version)
+            builder = builder.with_end_version(end_version)
+
+        return builder.build()
```
yanghua, line 3669: Agree `DatasetDeltaBuilder` could be `_DatasetDeltaBuilder`. Can we reuse the arg check in the `builder.build()` function?

jackye1995: What do you mean? It should already use it?

python/python/lance/dataset.py
```diff
@@ -3586,6 +3586,104 @@ def sql(self, sql: str) -> "SqlQueryBuilder":
...
+            raise ValueError(
+                "Must specify either 'compared_against' or both "
+                "'begin_version' and 'end_version'."
+            )
+
+        if has_range:
+            if begin_version is None or end_version is None:
+                raise ValueError(
+                    "Both 'begin_version' and 'end_version' must be specified together."
+                )
```
yanghua: I mean, this block is not necessary, because it has been checked in Rust:

https://github.com/lance-format/lance/blob/b21872585560cc4538a1f1977d5ac89f799b75c2/rust/lance/src/dataset/delta.rs#L105

And it could be reused for Python and Java?

python/python/lance/dataset.py
jackye1995: oh I see what you mean, you mean we don't need to duplicate the checks, agree

jackye1995 (approved):

Issue #4258, opened 2025-07-18
Implement ByteStreamSplit as a layer of encoding


Xuanwo:
So this works, but there might be a simpler way to implement this.  I often try to think of these encodings running layers (or maybe the right term is cascading).  For example, here is the diagram from the BtrBlocks paper:

<img width="910" height="299" alt="image" src="https://github.com/user-attachments/assets/28fbb441-de88-4edf-a4df-ac09aa9bf396" />

What you've done here is you've fused the byte stream split encoder and the `ValueEncoder` (which has the chunking logic) into a single encoding (reimplementing the chunking logic).  You could simplify it by allowing the downstream encoder to be provided by the caller.  For example:

```
#[derive(Debug, Clone)]
pub struct ByteStreamSplitEncoderMiniBlock {
  bits_per_value: usize,
  inner: Box<dyn MiniBlockCompressor>,
}

...

impl MiniBlockCompressor for ByteStreamSplitEncoderMiniBlock {
  fn compress(&self, page: DataBlock) -> Result<(MiniBlockCompressed, pb::ArrayEncoding)> {
    if DataBlock::FixedWidth(data) = page {
      // Do the actual byte stream split on the data buffer and make a new data buffer (no chunking)
      let split_buffer: Buffer = split_values(data.data);
      let split_data = DataBlock::FixedWidth(FixedWidthDataBlock {
        data: split_buffer,
        bits_per_value: data.bits_per_value,
        num_values: data.num_values,
        block_info: BlockInfo::new()
      };
      let (compressed, descr) = self.inner.compress(split_data);
      let wrapped_descr = ProtobufUtils::byte_stream_split(descr);
      Ok((compressed, wrapped_descr))
    } else {
      // error
    }
  }
}
```

This also makes it possible to do things like BSS -> Bitpacking + delta or some other layers of encoding (I don't know if that ever makes sense or we would need to do that).

_Originally posted by @westonpace in https://github.com/lancedb/lance/pull/4236#discussion_r2215998795_

Xuanwo:
See our current progress at https://github.com/lancedb/lance/issues/5135

PR #5095, opened 2025-10-29
refactor: make byte_stream_split a layer instead

Xuanwo:
Close https://github.com/lancedb/lance/issues/4258

This PR will make byte_stream_split a layer so we can finally chain them all together.

This PR is compatible with the 2.1 format, old files can still be read by new code. But the old code doesn't support reading new layered BSS generated in the new file, so we added a version 2.2 check for that path.

---


Xuanwo:
This design is much more complex than we thought. We (@westonpace and I) decided to hold off on this work for now until we figure out what’s best for us. The details are recorded at https://github.com/lancedb/lance/issues/5135

PR #5103, opened 2025-10-30
feat: add public accessors for count plan construction

wkalt:
This patch changes the count plan accessor to public, changes the get_deletion_vector method on fragments to public, and adds a couple public accessors on EvaluatedIndex.

The purpose of this patch is to enable systems to construct and optimize the count plan using datafusion optimizer rules.

### DIFF
--- a/rust/lance/src/io/exec/filtered_read.rs
+++ b/rust/lance/src/io/exec/filtered_read.rs
@@ -71,7 +71,8 @@ impl EvaluatedIndex {
         &self.index_result
     }
 
-    /// Get a reference to the applicable fragments bitmap
+    /// Get a reference to the applicable fragments bitmap, containing the set of fragment IDs
+    /// implicated by the filter.
     pub fn applicable_fragments(&self) -> &RoaringBitmap {
         &self.applicable_fragments
     }


rust/lance/src/io/exec/filtered_read.rs
```diff
@@ -65,6 +65,16 @@ pub struct EvaluatedIndex {
 }
 
 impl EvaluatedIndex {
+    /// Get a reference to the index result
```
wjones127: It would be nice for the comment to state what the data returned represents and not just parrot the type signature.
```suggestion
    /// Get the row id mask representing which rows matched the index filter.
```

rust/lance/src/io/exec/filtered_read.rs
```diff
@@ -65,6 +65,16 @@ pub struct EvaluatedIndex {
...
+    pub fn index_result(&self) -> &IndexExprResult {
+        &self.index_result
+    }
+
+    /// Get a reference to the applicable fragments bitmap
```
wjones127: Same here. I don't think the reader will immediately understand what "applicable fragments bitmap".

IIUC, it's the set of fragment ids that may contain matching rows.

wkalt:
@wjones127 thanks, updated

Issue #5093, opened 2025-10-29
The serialization method for BitmapIndex is not stable


ddupg:
In BitmapIndex, `RowIdTreeMap` is serialized into bytes using the [RowIdTreeMap.serialize_into](https://github.com/lancedb/lance/blob/afc0f9832cf11d0bf74381c2b63fd37de1c5f415/rust/lance-core/src/utils/mask.rs#L626) method, which is not stable. Does this bring a risk?

https://github.com/lancedb/lance/blob/afc0f9832cf11d0bf74381c2b63fd37de1c5f415/rust/lance-core/src/utils/mask.rs#L613-L627

Should we adopt alternative stable serialization methods or ensure `RowIdTreeMap.serialize_into` becomes stable?

@wjones127 @HaochengLIU @raunaks13 @westonpace Looking forward to your thoughts.

Thank @Kontinuation for helping find this.

westonpace:
This is a good point.  It is dangerous if this is not stable.  Any change to serialization will break forward and backward compatibility.

Easiest would be to just declare `serialize_into` stable I think.  This would probably involve creating a few backwards / forwards compatibility tests.  We might be able to piggy-back onto the existing backward / forward compatibility testing of indexes.

The backward compatibility tests should ensure we can read bitmap indexes written by old versions.  I'm not sure if any of our backwards compatibility tests cover this case.

The forward compatibility tests should ensure we can read bitmap indexes written by newer versions.  I think `test_index_search` in `forward_compat/test_compat.py` gives us some basic coverage here but maybe it could be made more extensive to cover other variations in the row id tree map (e.g. make sure we have multiple fragments, some fragments are fully covered by mask, etc.)

ddupg:
> This is a good point. It is dangerous if this is not stable. Any change to serialization will break forward and backward compatibility.
> 
> Easiest would be to just declare `serialize_into` stable I think. This would probably involve creating a few backwards / forwards compatibility tests. We might be able to piggy-back onto the existing backward / forward compatibility testing of indexes.

Agreed.

> The backward compatibility tests should ensure we can read bitmap indexes written by old versions. I'm not sure if any of our backwards compatibility tests cover this case.

It appears that backward compatibility tests has not yet been added. Could we provide it in the same manner as forward compatibility tests?

[...]

PR #4584, opened 2025-08-27
refactor: rework scalar index loading, training, and parsing into a plugin trait

westonpace:
These operations were previously all done via match statements.  This made it difficult to create new index types.

PR #5105, opened 2025-10-30
ci: generate multiple fragments in forward compatibility tests and doc RowIdTreeMap.serialize…

ddupg:
…_into stable

Closes: #5093

This PR made the following changes:
1. doc RowIdTreeMap.serialize_info as stable, which was used in bitmap index
2. make sure that the forward compatibility tests datasets have multiple fragments to guarantee older versions Lance can correctly load bitmap index.
3. make sure backward compatibility test `test_old_btree_bitmap_indices` use bitmap index
4. fix fails introduced by #4906

### DIFF
--- a/rust/lance-core/src/utils/mask.rs
+++ b/rust/lance-core/src/utils/mask.rs
@@ -612,7 +612,7 @@ impl RowIdTreeMap {
 
     /// Serialize the set into the given buffer
     ///
-    /// Keep the serialization format stable, it is used for serializing indexes data.
+    /// The serialization format is stable and used for index serialization
     ///
     /// The serialization format is:
     /// * u32: num_entries


ddupg:
Forward compatibility tests `test_write_scalar_index` and `test_write_fts` introduced by #5116 failed after generated dataset contained multiple fragments. When `ds.optimize.compact_files()`, trigger indexes remap, different indexes fail for different reasons:
- fts: same reason with #4906 
- btree: #4584 changes the schema of btree index, rename the `row_ids` column to `ids`, causes 
```
OSError: Invalid user input: Cannot write batch.  The batch was missing the column `row_ids`, /Users/runner/work/lance/lance/rust/lance-file/src/v2/writer.rs:360:35
```
- zonemap and bloomfilter: was introduced in 0.36.0, older versions cannot parse it.

I've also attempted to fix above problems here. Please @jackye1995 help review this.

ddupg:
The failed tests `index::vector::ivf::v2::tests::*` appears unrelated to this PR.

ddupg:
@wjones127 @westonpace could please help review this?

python/python/tests/forward_compat/test_compat.py
```diff
@@ -39,39 +39,48 @@ def test_json_index():
...
 
 
-@pytest.mark.forward
+def query_seven(ds, filt: str):
+    table = ds.to_table(filter=filt)
+    assert table.num_rows == 1
+    assert table.column("idx").to_pylist() == [7]
+
+    explain = ds.scanner(filter=filt).explain_plan()
+    print(explain)
```
wjones127: `pytest` should already print out this in case of failure, right?
```suggestion
```

rust/lance-core/src/utils/mask.rs
```diff
@@ -612,7 +612,7 @@ impl RowIdTreeMap {
 
     /// Serialize the set into the given buffer
     ///
-    /// The serialization format is not stable.
+    /// Keep the serialization format stable, it is used for serializing indexes data.
```
wjones127: This is public facing doc string.
```suggestion
    /// The serialization format is stable and used for index serialization
```

python/python/tests/forward_compat/test_compat.py
```diff
@@ -157,7 +229,10 @@ def test_write_fts(tmp_path: str):
     # copy to tmp path to avoid modifying original
     shutil.copytree(path, tmp_path, dirs_exist_ok=True)
 
-    ds = lance.dataset(tmp_path)
+    # Dataset::load_manifest does not do retain_supported_indices
+    # so this can only work with no cache
+    session = lance.Session(index_cache_size_bytes=0, metadata_cache_size_bytes=0)
+    ds = lance.dataset(tmp_path, session=session)
```
wjones127, line 234: 🤦 We should really fix this bug.

wjones127 (approved):
Looks good. I have some minor comments I can change before I merge.

PR #5106, opened 2025-10-30
fix: merge struct array use wrong child values

wojiaodoubao:
When I work with a dataset using struct column below, I got error 'Incorrect array length for StructArray field type, expected 5 got 3019'.
```
Total rows: 1000
Total items of 'type': 3019

Schema: map_data: list<item: struct<lane_dir: large_string, type: int64, xyz: list<item: list<item: double>>>>
  child 0, item: struct<lane_dir: large_string, type: int64, xyz: list<item: list<item: double>>>
      child 0, lane_dir: large_string
      child 1, type: int64
      child 2, xyz: list<item: list<item: double>>
          child 0, item: list<item: double>
              child 0, item: double
```

It is caused by values not correctly read when processing the child list.

### DIFF
--- a/rust/lance-arrow/src/lib.rs
+++ b/rust/lance-arrow/src/lib.rs
@@ -27,7 +27,9 @@ pub mod schema;
 pub use schema::*;
 pub mod bfloat16;
 pub mod floats;
+use crate::list::ListArrayExt;
 pub use floats::*;
+
 pub mod cast;
 pub mod json;
 pub mod list;
@@ -1308,8 +1310,8 @@ fn merge_with_schema(
                             .unwrap();
                         let merged_values = merge_list_child_values(
                             child_field.as_ref(),
-                            get_list_values(left_list).clone(),
-                            get_list_values(right_list).clone(),
+                            left_list.trimmed_values(),
+                            right_list.trimmed_values(),
                         );
                         let merged_validity =
                             merge_struct_validity(left_list.nulls(), right_list.nulls());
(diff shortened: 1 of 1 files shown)


wojiaodoubao:
Hi @westonpace , @Xuanwo , could you help review when you have time, thanks very much~

rust/lance-arrow/src/lib.rs
```diff
@@ -1390,6 +1390,15 @@ fn merge_with_schema(
...
 }
 
+fn get_list_values<O: OffsetSizeTrait>(list_array: &GenericListArray<O>) -> ArrayRef {
+    let offsets = list_array.value_offsets();
+    let start = offsets[0].to_usize().expect("offset overflow");
+    let end = offsets[list_array.len()]
+        .to_usize()
+        .expect("offset overflow");
+    list_array.values().slice(start, end - start)
+}
```
westonpace: There is already a function `trimmed_values` that you can use in `rust/lance-arrow/src/list.rs` in `ListArrayExt`.

rust/lance-arrow/src/lib.rs
```diff
@@ -1308,8 +1308,8 @@ fn merge_with_schema(
                             .unwrap();
                         let merged_values = merge_list_child_values(
                             child_field.as_ref(),
-                            left_list.values().clone(),
```
westonpace, line 1311: This change is definitely needed, good catch.

wojiaodoubao:
Hi @westonpace , thanks your suggestion, I've updated the pr, please review when you have time, thanks very much!

PR #5111, opened 2025-10-30
fix: no panic on unknown version

wjones127:
For more context, see: https://github.com/lancedb/lance/discussions/5099#discussioncomment-14819897


rust/lance-table/src/format/manifest.rs
```diff
@@ -619,6 +619,14 @@ pub struct WriterVersion {
     pub version: String,
 }
 
+/// A parsed Lance library version with major, minor, and patch components.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+pub struct LanceVersion {
```
jackye1995: why do we want a new LanceVersion instead of just the current `WriterVersion`? If we are doing this, I think we should just use https://docs.rs/semver/latest/semver/

wjones127: `WriterVersion` is the version of some arbitrary library. It might be `lance`, it might be another library. And if it's another library, it might not use a semantic versioning scheme.

`LanceVersion` is the version of the Lance Rust library. We guarantee it's semantically versioned. We could pull in `semver` for this, but it's also not clear we really care to add a dependency for this stuff. If you think it's worth it, feel free.

jackye1995: I think the main benefit is that it handles all sorts of pre release string formats so we don't need a custom parser anymore. Originally Xuanwo proposed it but I didn't add because it breaks backwards compatibility, but since we are already refactoring it here we could just do it along the way. But I can put another PR for it after this along they way to make the pre release forward compatible. I will just hold off publishing it until this is merged.

rust/lance-table/src/format/manifest.rs
```diff
@@ -684,15 +692,35 @@ impl WriterVersion {
...
             .unwrap_or_else(|| panic!("Invalid writer version: {}", self.version))
     }
 
-    /// Return true if self is older than the given major/minor/patch
-    pub fn older_than(&self, major: u32, minor: u32, patch: u32) -> bool {
-        let version = self.semver_or_panic();
-        (version.0, version.1, version.2) < (major, minor, patch)
+    /// Check if this is a Lance library version older than the given major/minor/patch.
+    /// Returns None if the library is not "lance" or the version cannot be parsed.
+    pub fn older_than(&self, major: u32, minor: u32, patch: u32) -> Option<bool> {
```
jackye1995: With this setup, I think it makes more sense to say self.lance_version()?.older_than(...) ?

rust/lance-table/src/format/manifest.rs
```diff
@@ -684,15 +692,35 @@ impl WriterVersion {
         Some((major, minor, patch, tag))
     }
 
+    /// Parse the version as a Lance library version.
+    /// Returns None if the library is not "lance" or the version cannot be parsed as semver.
+    pub fn lance_version(&self) -> Option<LanceVersion> {
```
jackye1995: can we call this `LanceLibVersion`? So it's not confused with the actual format version.

rust/lance-table/src/format/manifest.rs
```diff
@@ -619,6 +619,26 @@ pub struct WriterVersion {
...
 }
 
+/// A parsed Lance library version.
+///
+/// This is a wrapper around semver::Version that provides Lance-specific
+/// version comparison methods.
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord)]
+pub struct LanceVersion(semver::Version);
+
+impl LanceVersion {
```
jackye1995: this should be `LanceLibVersion` to match `lance_lib_version()`?

rust/lance-table/src/format/manifest.rs
```diff
@@ -662,9 +662,27 @@ pub enum VersionPart {
...
+        VersionPart::Patch => {
+            version.patch += 1;
+        }
+    }
+}
+
 impl WriterVersion {
     /// Try to parse the version string as a semver string. Returns None if
     /// not successful.
+    #[deprecated(note = "Use `lance_lib_version()` instead")]
```
jackye1995, line 685: I was thinking this could still be useful for non Lance library version as long as they are using semver. But this should be fine to deprecate for now.

jackye1995 (approved):
Looks good to me!

rust/lance-table/src/format/manifest.rs
wjones127: Yeah if we start doing carveouts for other libraries, we can add that back.

Issue #5044, opened 2025-10-22
Java lock is not updated after rust dependency update


jackye1995:
Need to update CI to detect this

Xuanwo:
Closing as completed by #7425. Java CI uses `--locked` for Rust dependency changes, and the repository also has the `cargo-lock-sync` hook as a second guard.

PR #5116, opened 2025-10-31
fix: forward incompatibility of prerelease in writer version

jackye1995:
WIth previous CI change, we now use the actual prerelease version as writer version. However, old version cannot parse such version string and cause panic. 

This PR makes sure that the version in WriterVersion is always just major.minor.patch. Any prerelease and build metadata are stored separately and not visible to old clients.


java/lance-jni/Cargo.lock
jackye1995, line 1: this is the old problem of #5044

protos/table.proto
```diff
@@ -58,7 +58,33 @@ message Manifest {
...
+    // This field stores the prerelease portion of a semantic version separately
+    // from the core version number. For example, if the full version is "2.0.0-rc.1",
+    // the version field would contain "2.0.0" and prerelease would contain "rc.1".
+    //
+    // This separation ensures forward compatibility: older readers can parse the
+    // clean version field without errors, while newer readers can reconstruct the
+    // full semantic version by combining version, prerelease, and build_metadata.
+    //
+    // If absent, the version field is used as-is.
+    optional string prerelease = 3;
```
jackye1995, line 78: the original proposal was to just use a single classifier string, but that makes it hard to leverage the semver parser, so I made it aligned with the semver spec

protos/table.proto
wjones127, line 1: Are we baking in semver as a requirement for writers? Seems unnecessary for the format to be opinionated about that?

jackye1995: It still works with arbitrary string. But in general it feels like recording the full semver is beneficial that we can know if the writer is a specific version, if it is the main release version or a specific beta version.

wjones127: Yeah I don't disagree with recording the full semver. More concerned that providing specific `prerelease` and `build_metadata` then we baking in semver concepts into the format. It seems like we could just have a field, `version_extra` or something like that where we put that segment of the version.

protos/table.proto
```diff
@@ -58,7 +58,33 @@ message Manifest {
     // that the library is semantically versioned, this is a string. However, if it
     // is semantically versioned, it should be a valid semver string without any 'v'
     // prefix. For example: `2.0.0`, `2.0.0-rc.1`.
+    //
+    // For forward compatibility with older readers, when writing new manifests this
+    // field should contain only the core version (major.minor.patch) without any
+    // prerelease or build metadata. The prerelease/build info should be stored in
+    // the separate prerelease and build_metadata fields instead.
```
wjones127, line 65: Let's make it clear this is a compatibility issue specific to Lance Rust library and only for certain old versions. That makes it clearer to new libraries what the impact would be of deviating from this.

```suggestion
    // Compatibility note: The Lance Rust library in versions <=0.38.3 would
    // panic if it encountered a non-semver version string here or a version string
    // with anything more than major, minor, patch version.
    // For forward compatibility with older readers, when writing new manifests this
    // field should contain only the core version (major.minor.patch) without any
    // prerelease or build metadata. The prerelease/build info should be stored in
    // the separate prerelease and build_metadata fields instead.
```

protos/table.proto
jackye1995: yeah I went back and forth with it tbh. 

Originally I just have a classifier field (that was originally what was proposed on Slack), then my internal debate is that for example I have a `1.2.3-beta.2` string, I can choose to store `beta.2` in classifier, and I just split by `-` string to get that split. But what if there is a `1.2.3+build.abcde` string which is still semver in the future, then it does not work.

I also thought about storing `-beta.2` in the classifier, but then we need a parser to seek to the first non-version position of the string and then split there. It feels like a bit of an overkill, given for the Lance library it is basically always semver. Even if we have some internal versions in the future, it probably still makes sense to follow the semver part and just have `-internal.2` or leverage build metadata to store internal info, so we can easily infer the ordering of different versions.

So that was my thoughts to arrive at this state.

I guess having a single classifier/version_extra field would also work, let me know which you prefer.

wjones127: Yeah I was thinking you could just parse semver as `str_concat(version, version_extra)`.

jackye1995: The problem is I think the reverse, how do you move from a version string to `version` and `version_extra`. It is not clear to me if for example if I have `1.0-beta.1`, this is not a standard semver, does it mean I need to put everything to version, or split it to `1.0` and `-beta.1`.

The current approach provides a clear rule that (1) if it is semver, then can leverage those additional fields, (2) if it is not semver, everything is still stored in the version string.

If we want to move to just a version_extra, looks like we will define something like if it starts with 3 numbers connected with 2 dots, those go to version, and the rest go to the extra. Does that sound good?

wjones127: > The current approach provides a clear rule that (1) if it is semver, then can leverage those additional fields, (2) if it is not semver, everything is still stored in the version string.

Okay, I guess that is fine.

jackye1995:
Failure due to flaky tests, merging

PR #5117, opened 2025-10-31
feat: support create vector index distributedly

### LINKED ISSUE #4723: support build IVF_FLAT/PQ/SQ vector index distributedly
Sub-issue of: https://github.com/lancedb/lance/issues/4155
- We recently added support for FTS indexes in:
  - Core: https://github.com/lancedb/lance/pull/4578
  - Ray connector: https://github.com/lancedb/lance-ray/pull/45
- It’s also great to see the distributed B-tree index PR under review by @xloya:
  - https://github.com/lancedb/lance/pull/4667
- Given this momentum, additional scalar index types may follow soon.
### Proposal: Distributed Vector Index (CPU-based IVF)
- Question: Is there ongoing or existing work on a distributed vector index building?
- Context:
  - Current vector index builds rely on CUDA for acceleration.
  - Single-node index builds still have high resource requirements (e.g., transient disk usage during build), which may also impact distributed scenarios.
- Plan:
  - I intend to begin with a distributed IVF index implementation targeted at CPU-only nodes.
  - The design will draw on approaches from OpenSearch’s KNN and other disk-based ANN systems for building and managing IVF in a distributed setting.
### Next steps
- @yanghua and I will conduct research and begin a prototype over the next few months.
- We will share design notes and open a PR once we have a concrete proposal or initial implementation for community feedback.

chenghao-guo:
close #4723   

**Key Changes:**        
  The distributed index creation leverages the existing IVF framework while adding coordination mechanisms for multi-node execution. The index merger component now handles distributed fragment consolidation and metadata  synchronization.      This work enables scalable vector index creation for large-scale datasets, significantly reducing index build time.                                                                                                                                                                

-   Implemented distributed IVF index building infrastructure                                                                                                                                                                 for parallel index construction across multiple nodes                                                                                                                                                         
-   Enhanced the index merger component for distributed operations         
-  For IVF_HNSW part, the HNSW graph is built locally within the shard as a sub-index of the partition; there is no cross-shard graph merging and no cross-shard edges.   These are supported but distribution only happens in IVF.   
-  CPU only, torch accelerator will not be supported and fall back to single node IVF index creation.                                                                                                                                        

                                                                                                                                                                                                                                
  **Current Status in this PR:**                                                                                                                                                                                                                                                                                                                                                                                                                          
  • FLAT/SQ: Should work now, it is under active testing phase, validating distributed performance and accuracy       
 • PQ (Product Quantization): Currently depends on global training codebook, requiring centralized training before distributed deployment.                                                                                                                                                                                                                   
  • RQ (Residual Quantization): I didn't consider this when I design this PR. Not yet supported in distributed mode maybe, planned for future implementation   

Once I finish all the testing phase in my side on performance and recall accuracy, I will mark it ready to review.

### DIFF
--- a/python/python/tests/test_vector_index.py
+++ b/python/python/tests/test_vector_index.py
@@ -186,14 +186,14 @@ def test_ann(indexed_dataset):
 @pytest.mark.parametrize(
     "fixture_name,index_type,index_params,similarity_threshold",
     [
-        ("dataset", "IVF_FLAT", {"num_partitions": 4}, 0.95),
+        ("dataset", "IVF_FLAT", {"num_partitions": 4}, 0.80),
         (
             "indexed_dataset",
             "IVF_PQ",
             {"num_partitions": 4, "num_sub_vectors": 16},
-            0.90,
+            0.80,
         ),
-        ("dataset", "IVF_SQ", {"num_partitions": 4}, 0.90),
+        ("dataset", "IVF_SQ", {"num_partitions": 4}, 0.80),
     ],
 )
 def test_distributed_vector(
@@ -210,7 +210,7 @@ def test_distributed_vector(
         topk=10,
         world=2,
         similarity_metric="recall",
-        similarity_threshold=0.8,
+        similarity_threshold=similarity_threshold,
     )
 
 


yanghua:
Hi @BubbleCal, this PR is ready for review. Please take a look, thanks.

python/python/lance/indices/builder.py
```diff
@@ -203,6 +203,79 @@ def train_pq(
         )
         return PqModel(num_subvectors, pq_codebook)
 
+    def prepare_global_ivfpq(
+        self,
+        num_partitions: int,
```
BubbleCal: this should be optional, lance would calculate a default value for this param now

rust/lance/src/index/vector/builder.rs
```diff
@@ -120,6 +120,9 @@ pub struct IvfIndexBuilder<S: IvfSubIndex, Q: Quantization> {
 
     frag_reuse_index: Option<Arc<FragReuseIndex>>,
 
+    // fields for distributed indexing
```
BubbleCal: typo: fields -> fragments

rust/lance-index/src/vector/distributed/index_merger.rs
```diff
@@ -0,0 +1,1818 @@
...
+    }
+}
+
+/// Merge all partial_* vector index auxiliary files under `index_dir/{uuid}/partial_*/auxiliary.idx`
+/// into `index_dir/{uuid}/auxiliary.idx`.
+///
+/// Supports IVF_FLAT, IVF_PQ, IVF_SQ, IVF_HNSW_FLAT, IVF_HNSW_PQ, IVF_HNSW_SQ storage types.
+/// For PQ and SQ, this assumes all partial indices share the same quantizer/codebook
+/// and distance type; it will reuse the first encountered metadata.
+pub async fn merge_vector_index_files(
```
BubbleCal: I think we should just reuse `IvfIndexBuilder::merge_partitions`, otherwise this is too hard to maintain

rust/lance-index/src/vector/hnsw/builder.rs
```diff
@@ -713,7 +713,41 @@ impl IvfSubIndex for HNSW {
 
         let schema = VECTOR_RESULT_SCHEMA.clone();
         if self.is_empty() {
-            return Ok(RecordBatch::new_empty(schema));
+            // Fallback: perform flat search over storage when HNSW graph is empty
```
BubbleCal: why this change? here `self.is_empty()` means no data in this partition (storage)

rust/lance-index/src/vector/ivf/storage.rs
```diff
@@ -110,12 +110,19 @@ impl IvfModel {
...
-            distance_type,
-            vec![],
-        );
-        internal.find_partitions(query, nprobes)
+        if let Some(centroids) = self.centroids.clone() {
+            let internal =
+                crate::vector::ivf::new_ivf_transformer(centroids, distance_type, vec![]);
+            internal.find_partitions(query, nprobes)
+        } else {
+            // Fallback: if centroids are not available (e.g., distributed IVF_FLAT shards without pretrained centroids),
```
BubbleCal: how could this happen? could you add more explain for this case?

rust/lance/src/index/vector/ivf/v2.rs
```diff
@@ -261,27 +261,31 @@ impl<S: IvfSubIndex + 'static, Q: Quantization> IVFIndex<S, Q> {
...
-                                    1,
-                                    FilterExpression::no_filter(),
-                                )?
-                                .try_collect::<Vec<_>>()
-                                .await?;
-                            concat_batches(&schema, batches.iter())?
-                        }
+                let batch = {
+                    let num_rows_meta = self.reader.metadata().num_rows;
+                    let num_rows_reader = self.reader.num_rows();
```
BubbleCal: what's the diff between `self.reader.metadata().num_rows` and `self.reader.num_rows()`?

chenghao-guo:
> I don't think I've clearly understood this PR, but some questions so far:
> 
> * how do we divide the index job into many tasks that are sent to the nodes? I think it should be the per IVF partition by the design, but it looks like the implementation is sending fragment. If I'm reading wrong, could you point the code position?
> * I think there are many lines that just repeating what `IvfIndexBuilder` does, prefer to reuse `IvfIndexBuilder`'s ability so that we don't need to maintain 2 copies, esp the merge partitions logic.

Thanks a lot for looking into this PR and for raising these points — they are very helpful.

On how the index job is divided and sent to nodes:  In the distributed path we first run the same “global” create_index flow (training and partitioning) and then dispatched fragments to workers. As each fragment corresponds to a concrete set of IVF partitions, we assign different rows in fragment to the nearest IVF partition. So the actual work that each node performs is still partition‑based. the fragment is just our transport/execution unit for the distributed scheduler.

Regarding reusing IvfIndexBuilder: I fully agree with you. The intention is to keep the high‑level index pipeline in one place and only make the execution strategy (local vs distributed) pluggable. In this PR the distributed builder reuses the same training and partitioning logic as create_index, but some parts of the partition‑merge flow are still mirroring what IvfIndexBuilder does. I plan to refactor this so that the distributed path delegates more directly to IvfIndexBuilder for merge and partition handling, and the distributed code focuses only on job splitting and task orchestration.

rust/lance-index/src/vector/hnsw/builder.rs
yanghua: This change tries to give a fallback for this logic: if there is no graph node, but there may be vectors in the storage.

Seems not necessary. Will revert.

yanghua:
Hi @BubbleCal, after doing some analysis about the merger component, I'd like to share some thoughts:

* The partition sources of the single-machine and the distributed mode are different: 
  * for the single machine: the data comes from the memory;
  * for distributed: the data is persisted into the storage;
* About the **file writer initatization** they are also different:
  * for the single machine, it is initialized in the `merge_partitions` method;
  * for the distributed: we implement via `init_writer_for_flat/pq/sq` for different vector index type;
* About the logic of the merger, they are also different due to different partition data sources.

For different partition sources, it would be better to abstract a `PartitionSource` trait. After that, we can introduce a `UnifiedPartitionMerger` to do a general merger. Introducing a `StorageWriterFactory` to create different writers.

For a merger, the generic logic can be split into four common steps:

* create merger;
* instantiate merger;
* merger#merge();
* write final metadata;

But these changes are too extensive to make this PR bigger and hard to review. I would suggest moving these changes to another review PR. WDYT? If you agree, I could file an issue ticket.

chenghao-guo:
To evaluate the performance, I did some test based on lance-ray connector with this preview PR https://github.com/lance-format/lance-ray/pull/67 and below is the result

### Testing Environment

We conducted tests using the `imagenet_train` dataset, which consists of **1,281,167 images**. These images were embedded using a **2048-dimension** `doubao-embedding` model, resulting in a dataset size of approximately **140GB** after converting to **lance**.

- **Single-machine index building**
  - Machine: **4 cores, 16GB RAM**
  - Storage: **object-store** for original dataset

- **Distributed setup**
  - Topology: **1 head node + 4 worker nodes**
  - Each node: **4 cores, 16GB RAM**
  - Purpose: use very cheap machines to extend overall runtime and evaluate performance improvements of distributed index building on low-config machines

- **Index building parameters**
  - `num_partitions = 1000`
  - sub_vectors=128 for PQ
  - Data split into **253 fragments**
  - Distributed processing: **4 workers** in parallel

---

### Testing Results

- **Unit:** `m` = minute

#### IVF_SQ Group (Tested on S3)

| Configuration | Global Train IVF Time | Index Builder Time | Total Time |
|---|---:|---:|---:|
| Distributed `(1 + 4) * 4c-16GB` | 11.25m | 25.7m | 36.95m |
| Single Machine `4C-16GB` | 11.5m | 137.5m | 149.0m |

#### IVF_FLAT Group

| Configuration | Global Train IVF | Index Builder Time | Total Time |
|---|---:|---:|---:|
| Distributed `(1 + 4) * 4c-16GB` | 10.68m | 25.82m | 36.5m |
| Single Machine `4C-16GB` | 11.8m | OOM Killed | OOM Killed |

#### IVF_PQ Group

| Configuration | Global Train IVF Time | Global PQ Training Time | Index Builder Time | Total Time |
|---|---:|---:|---:|---:|
| Distributed `(1 + 4) * 4c-16GB` | 10.53m | 180.4m | 48m | 238.9m |
| Single Machine `4C-16GB` | 11.48m | 171m | 460.5m | 642.98m |

---

### Overall Observation

Since it works on **S3**, **network speed** and **S3 I/O** may vary during the process. Sometimes it can take hours, but mainly due to the reasons below.

The distributed index was built using **4-core, 16GB** machines (**4 workers** in total). We also checked if performance could scale linearly. The main reasons for the performance acceleration are:

1. **Avoiding backpressure throttle exceeded in IO** on a single machine.  
   We observed that when building an index on a low-performance machine, the lance process kept reporting:  
   > `"backpressure throttle exceeded in IO"`
2. **Distributed assignment** allows using CPU resources from worker nodes to parallelize the process.

---

### Conclusion

The test was carried out with a **4-core, 16GB** machine and an **S3 object-store**. The results clearly show that the distributed setup significantly outperforms the single-machine setup in terms of index building time, especially for the `Distribute Builder Time` metric, for **scaling linearly** or better.

---

### Limitation

Since **PQ (Product Quantization)** depends on:

- global training of the **Inverted File (IVF)**, and  
- **PQ codebook training**,

both of which are carried out on a single machine, if the training of the PQ codebook takes a long time, the acceleration effect of **IVF-PQ** may not be satisfactory. However, it is more suitable for a larger number of tables.

The most well-balanced and superior method appears to be **IVF_SQ**, as this method can generally guarantee good recall and support distributed parallelism.

rust/lance/src/index/vector/ivf.rs
```diff
@@ -1847,6 +1856,186 @@ async fn write_ivf_hnsw_file(
...
+                distance_type: dt,
+            })?
+        };
+
+    // Write root index.idx via V2 writer so downstream opens through v2 path.
+    let index_path = index_dir.child(INDEX_FILE_NAME);
+    let obj_writer = object_store.create(&index_path).await?;
+
+    // Schema for HNSW sub-index: include neighbors/dist fields; empty batch is fine.
+    let arrow_schema = Arc::new(arrow_schema::Schema::new(vec![
```
BubbleCal: use correct schema by `XxxIndex::schema()` (e.g. `FlatIndex::schema()`)

python/python/tests/test_vector_index.py
```diff
@@ -178,6 +183,37 @@ def test_ann(indexed_dataset):
...
+        (
+            "indexed_dataset",
+            "IVF_PQ",
+            {"num_partitions": 4, "num_sub_vectors": 16},
+            0.90,
+        ),
+        ("dataset", "IVF_SQ", {"num_partitions": 4}, 0.90),
+    ],
+)
+def test_distributed_vector(
```
BubbleCal, line 199: can we test all index types supported?

python/python/tests/test_vector_index.py
```diff
@@ -1912,3 +1932,779 @@ def func(rs: pa.Table):
...
+       preprocessed artifacts
+    5) For each query, compute ground-truth TopK IDs using exact search
+       (use_index=False), then compute TopK using single index and the distributed
+       index with consistent nearest settings (refine_factor=1; IVF uses nprobes)
+    6) Compute recall for single and distributed using the provided formula and
+       assert the absolute difference is <= 0.10. Also print the recalls.
+    """
+    # Keep signature compatibility but ignore similarity_metric/threshold
+    _ = similarity_metric
+    _ = similarity_threshold
```
BubbleCal: we should check this, but it's acceptable to set a low threshold (0.5~0.8), so that it can find potential bugs. see tests in `rust/lance/src/index/vector/ivf/v2.rs` for recall requirement numbers

python/python/tests/test_vector_index.py
yanghua: Currently, for controlling the scale of this PR. This PR focuses on `IVF_FLAT_` indices. I will start a new issue to track `IVF_HNSW_` indices.

yanghua:
@BubbleCal Thanks for reviewing! I have filed some tickets to track further work: #5621 and #5622.

Issue #5119, opened 2025-10-31
Panic: unwrap on None in lance-encoding/src/decoder.rs:791


wjones127:
`thread 'tokio-runtime-worker' panicked at 'called `Option::unwrap()` on a `None` value': /src/lance/rust/lance-encoding/src/decoder.rs:791\n   0: log_panics::Config::install_panic_hook::{{closure}}\n   1: std::panicking::rust_panic_with_hook\n   2: std::panicking::begin_panic_handler::{{closure}}\n   3: std::sys::backtrace::__rust_end_short_backtrace\n   4: __rustc::rust_begin_unwind\n   5: core::panicking::panic_fmt\n   6: core::panicking::panic\n   7: core::option::unwrap_failed\n   8: lance_encoding::decoder::CoreFieldDecoderStrategy::create_legacy_field_scheduler\n   9: lance_encoding::decoder::CoreFieldDecoderStrategy::create_legacy_field_scheduler\n  10: lance_encoding::decoder::CoreFieldDecoderStrategy::create_legacy_field_scheduler\n  11: lance_encoding::decoder::create_scheduler_decoder::{{closure}}\n  12: tokio::runtime::task::core::Core<T,S>::poll\n  13: tokio::runtime::task::raw::poll\n  14: tokio::runtime::scheduler::multi_thread::worker::Context::run_task\n  15: tokio::runtime::scheduler::multi_thread::worker::Context::run\n  16: tokio::runtime::context::scoped::Scoped<T>::set\n  17: tokio::runtime::context::runtime::enter_runtime\n  18: tokio::runtime::scheduler::multi_thread::worker::run\n  19: <tokio::runtime::blocking::task::BlockingTask<T> as core::future::future::Future>::poll\n  20: tokio::runtime::task::core::Core<T,S>::poll\n  21: tokio::runtime::task::harness::Harness<T,S>::poll\n  22: tokio::runtime::blocking::pool::Inner::run\n  23: std::sys::backtrace::__rust_begin_short_backtrace\n  24: core::ops::function::FnOnce::call_once{{vtable.shim}}\n  25: std::sys::pal::unix::thread::Thread::new::thread_start\n  26: <unknown>\n  27: <unknown>\n`

camilesing:
If there are no repeated steps and expected behaviors, then I think you hope it doesn't throw a panic

Xuanwo:
[...]

PR #5121, opened 2025-11-02
fix: panic unwrap on None in decoder.rs

camilesing:
for: https://github.com/lancedb/lance/issues/5119

### DIFF
--- a/.bumpversion.toml
+++ b/.bumpversion.toml
@@ -1,5 +1,5 @@
 [tool.bumpversion]
-current_version = "1.0.0-beta.15"
+current_version = "1.0.0-beta.16"
 parse = "(?P<major>\\d+)\\.(?P<minor>\\d+)\\.(?P<patch>\\d+)(-(?P<prerelease>(beta|rc))\\.(?P<prerelease_num>\\d+))?"
 serialize = [
     "{major}.{minor}.{patch}-{prerelease}.{prerelease_num}",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3215,7 +3215,7 @@ checksum = "42703706b716c37f96a77aea830392ad231f44c9e9a67872fa5548707e11b11c"
 
 [[package]]
 name = "fsst"
-version = "1.0.0-beta.15"
+version = "1.0.0-beta.16"
 dependencies = [
  "arrow-array",
  "lance-datagen",
@@ -3388,9 +3388,9 @@ dependencies = [
 
 [[package]]
 name = "geo-types"
-version = "0.7.17"
+version = "0.7.18"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "75a4dcd69d35b2c87a7c83bce9af69fd65c9d68d3833a0ded568983928f3fc99"
+checksum = "24f8647af4005fa11da47cd56252c6ef030be8fa97bdbf355e7dfb6348f0a82c"
 dependencies = [
  "approx",
  "num-traits",
@@ -4398,7 +4398,7 @@ dependencies = [
 
 [[package]]
 name = "lance"
-version = "1.0.0-beta.15"
+version = "1.0.0-beta.16"
 dependencies = [
  "all_asserts",
  "approx",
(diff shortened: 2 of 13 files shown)


rust/lance-encoding/src/decoder.rs
```diff
@@ -788,6 +788,12 @@ impl CoreFieldDecoderStrategy {
             let scheduler = self.create_primitive_scheduler(field, column_info, buffers)?;
             return Ok(scheduler);
         } else if data_type.is_binary_like() {
+            if column_infos.next().is_none() {
+                return Err(Error::InvalidInput {
+                    source: "column_info cannot be none".into(),
+                    location: location!(),
+                });
+            }
             let column_info = column_infos.next().unwrap().clone();
```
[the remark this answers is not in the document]
wjones127: The AI is right here: it seems evident with all the test failures.

wjones127: Here's how you could avoid calling `.next()` twice:
```suggestion
            let Some(column_info) = column_infos.next() else {
                return Err(Error::InvalidInput {
                    source: "column_info cannot be none".into(),
                    location: location!(),
                });
            };
            let column_info = column_info.clone()
```

rust/lance-encoding/src/decoder.rs
```diff
@@ -788,6 +788,12 @@ impl CoreFieldDecoderStrategy {
             let scheduler = self.create_primitive_scheduler(field, column_info, buffers)?;
             return Ok(scheduler);
         } else if data_type.is_binary_like() {
+            if column_infos.next().is_none() {
+                return Err(Error::InvalidInput {
+                    source: "column_info cannot be none".into(),
```
wjones127: This error is very unexpected, so we should make sure to add as much context as we can.

I would recommend adding the `field` data type and name as part of the error message.

rust/lance-encoding/src/decoder.rs
wjones127: Looking at the code above, I actually think we might instead be able to re-use this utility:
```suggestion
            let column_info = column_infos.expect_next()?;
```

camilesing:
Do we have a requirement for code coverage? @wjones127

rust/lance-encoding/src/decoder.rs
```diff
@@ -789,7 +789,19 @@ impl CoreFieldDecoderStrategy {
...
+                    source: format!(
+                        "column_info cannot be none. the field type is {}, name is {}",
+                        field.data_type(),
+                        field.name
+                    )
+                    .into(),
+                    location: location!(),
+                });
+            }
+            let column_info = column_info_next.unwrap().clone();
```
wjones127: Please avoid using unwrap unnecessarily. Even if you know it won't panic, placing an unwrap there adds additional code size because now it needs to add a panic message for that line into the binary.

Better to use statements like:
```suggestion
            let Some(column_info) = column_infos.next() else {
                return Err(Error::InvalidInput {
                    source: format!(
                        "column_info cannot be none. the field type is {}, name is {}",
                        field.data_type(),
                        field.name
                    )
                    .into(),
                    location: location!(),
                });
            };
            let column_info = column_info.clone();
```

rust/lance-encoding/src/decoder.rs
```diff
@@ -813,18 +825,30 @@ impl CoreFieldDecoderStrategy {
...
-                    ))
-                    .unwrap();
-                    let list_scheduler = self.create_list_scheduler(
+                    ));
+
+                    if list_field_result.is_err() {
+                        return Err(list_field_result.err().unwrap());
+                    }
+
+                    let list_field = list_field_result.unwrap();
```
wjones127: Why not just use `?` here?
```suggestion
                    let list_field = Field::try_from(ArrowField::new(
                        field.name.clone(),
                        list_type,
                        field.nullable,
                    ))?;
```

wjones127 (changes requested):

rust/lance-encoding/src/decoder.rs
camilesing: Thx u for the guidance, I have learned

camilesing:
@wjones127 It looks like everything is ready

camilesing:
@wjones127 hello?

PR #5125, opened 2025-11-03
perf: optimize metadata read for file reader

yanghua:
Currently, if we want to perform a projected read of a file, we load all of the column metadata, and then only read the column data that is requested.  This is fine for most cases. However, if there are many columns then loading all of the column metadata may be expensive.  We should support a mode where we only load the column metadata for the columns that are requested (the file format supports this).

There is a benchmark in the [F3 paper](https://dl.acm.org/doi/pdf/10.1145/3749163)'s 7.2 section, it also describes this issue as below:

> For example, Lance currently has to read and deserialize all the metadata, despite using
FlatBuffer, making it the slowest among the FlatBuffer-based ones [38].

### DIFF
--- a/rust/lance-file/src/v2/reader.rs
+++ b/rust/lance-file/src/v2/reader.rs
@@ -27,7 +27,7 @@ use lance_encoding::{
 };
 use log::debug;
 use object_store::path::Path;
-use prost::{Message, Name};
+use prost::Message;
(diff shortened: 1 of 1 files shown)


westonpace:
I think this has been superseded by https://github.com/lance-format/lance/pull/7375

PR #5126, opened 2025-11-03
feat: support add sub-column to struct col

wojiaodoubao:
Adding sub-columns to a struct data type column.

### DIFF
--- a/rust/lance/src/dataset/schema_evolution.rs
+++ b/rust/lance/src/dataset/schema_evolution.rs
@@ -187,6 +187,10 @@ fn check_field_conflict(
         (DataType::FixedSizeList(fl, _), DataType::FixedSizeList(fr, _)) => {
             check_field_conflict(fl, fr, version)
         }
+        (l_type, r_type) if l_type == r_type => Err(Error::invalid_input(
+            format!("Column {} already exists in the dataset", left.name()),
+            location!(),
+        )),
         (_, _) => Err(Error::invalid_input(
             format!(
                 "Type conflicts between {}({}) and {}({})",


wojiaodoubao:
In this pr, adding sub-columns to struct column only supports versions 2.1 and later. Considering that the v2.0 format will gradually be phased out, I think it might be unnecessary to support adding sub-columns for v2.0.

If we need to support adding sub-columns for v2.0, the main challenge is: In v2.0, a header column is stored for nested columns (struct, list), and it assumes the same header column will not exist multiple times within a single fragment.

There are two possible approaches:
* Approach 1: Keep the encoding and decoding unchanged. New files will include the new leaf columns and redundant header columns, and when reading the header column, always select the first header column found.
* Approach 2: Modify the encoding and decoding. New files will only include the new leaf columns and will not include redundant header columns (this requires significant changes).

wojiaodoubao:
The failures are caused by https://github.com/lancedb/lance/pull/5206.

wojiaodoubao:
Hi @jackye1995 , could you help review when you have time, looking forward to your suggestions!

wojiaodoubao:
This is a prerequisite for the variant (https://github.com/lance-format/lance/discussions/5238). Hi @jackye1995 , could you help review this when you have time, thanks very much~

rust/lance/src/dataset/schema_evolution.rs
```diff
@@ -132,6 +133,63 @@ fn is_upcast_downcast(from_type: &DataType, to_type: &DataType) -> bool {
...
+            }
+
+            for l_field in fl.iter() {
+                if let Some((_, r_field)) = fr.find(l_field.name()) {
+                    check_field_conflict(l_field, r_field, version)?;
+                }
+            }
+            Ok(())
+        }
+        (DataType::List(fl), DataType::List(fr)) => check_field_conflict(fl, fr, version),
```
jackye1995, line 185: I think we also need fixed size list and large list?

rust/lance/src/dataset/schema_evolution.rs
```diff
@@ -132,6 +133,63 @@ fn is_upcast_downcast(from_type: &DataType, to_type: &DataType) -> bool {
...
+        (_, _) => Err(Error::invalid_input(
+            format!("Column {} already exists in the dataset", left.name()),
```
jackye1995: this error message should more be like "type mismatch between xxx and yyy"?

rust/lance/src/dataset.rs
```diff
@@ -10115,4 +10115,221 @@ mod tests {
...
+                None,
+            )
+            .await
+            .unwrap_err();
+        assert!(err.to_string().contains(
+            "Column animal is a struct col, add sub column is not supported in Lance file version"
+        ));
+    }
+
+    #[tokio::test]
```
jackye1995: this should be parameterized test?

rust/lance/src/dataset.rs
```diff
@@ -10115,4 +10115,221 @@ mod tests {
...
+    async fn test_add_sub_column_to_struct_col_v2_2() {
+        test_add_sub_column_to_struct_col(LanceFileVersion::V2_2).await;
+    }
+
+    async fn test_add_sub_column_to_struct_col(version: LanceFileVersion) {
+        let mut dataset = prepare_initial_dataset_with_struct_col(version, HashMap::new()).await;
+
+        // add 2 sub-columns of animal
```
jackye1995: can we test a few different and more complicated cases here, like fixed size list and deeply nested struct?

Xuanwo:
File format v2.1 has been stabilized, and we should not add any more features to it. Instead, please target file format 2.2.

rust/lance-encoding/src/version.rs
```diff
@@ -81,6 +81,10 @@ impl LanceFileVersion {
 
         Self::iter().filter(|&v| v != Self::Stable && v != Self::Next && v != Self::Legacy)
     }
+
+    pub fn support_add_sub_column(&self) -> bool {
```
Xuanwo: I’m a bit concerned about adding the API this way.

wojiaodoubao:
> File format v2.1 has been stabilized, and we should not add any more features to it. Instead, please target file format 2.2.

@Xuanwo Thanks your reminding! Marked v2.1 as unsupported in adding sub-column.

wojiaodoubao:
Fixed all comments. Hi @jackye1995 , please help review when you have time, thanks very much~

wojiaodoubao:
I rebased this pr and it is ready for review now. Hi @jackye1995 , please help when you have time, thanks very much~.

rust/lance/src/dataset/schema_evolution.rs
```diff
@@ -132,6 +133,73 @@ fn is_upcast_downcast(from_type: &DataType, to_type: &DataType) -> bool {
...
+                }
+            }
+            Ok(())
+        }
+        (DataType::List(fl), DataType::List(fr)) => check_field_conflict(fl, fr, version),
+        (DataType::LargeList(fl), DataType::LargeList(fr)) => check_field_conflict(fl, fr, version),
+        (DataType::FixedSizeList(fl, _), DataType::FixedSizeList(fr, _)) => {
+            check_field_conflict(fl, fr, version)
+        }
+        (_, _) => Err(Error::invalid_input(
```
jackye1995, line 194: sorry I guess this actually need 2 different error messages, one is this one and one is the original one, because if the type is the same then this message is not really right. could you update that?

PR #5131, opened 2025-11-04
refactor: remove not used storage class and blob dataset

Xuanwo:
This PR removes unused storage class and blob dataset to make our logic cleaner and easier to follow. That’s especially important as we work on support for blob v2 now.

This PR includes breaking changes at the API level, but since we never used a blob dataset before, it won’t break users’ existing datasets.

---

### DIFF
--- a/rust/lance-core/src/datatypes/field.rs
+++ b/rust/lance-core/src/datatypes/field.rs
@@ -995,8 +995,6 @@ mod tests {
 
     use arrow_array::{DictionaryArray, StringArray, UInt32Array};
     use arrow_schema::{Fields, TimeUnit};
-    use std::collections::HashMap;
-
     #[test]
     fn arrow_field_to_field() {
         for (name, data_type) in [
--- a/rust/lance-core/src/datatypes/schema.rs
+++ b/rust/lance-core/src/datatypes/schema.rs
@@ -146,56 +146,6 @@ impl Schema {
         }
     }
 
-    pub fn retain_blob_fields(&self) -> Self {
-        self.retain_by(|f| f.is_blob())
-    }
-
-    pub fn retain_non_blob_fields(&self) -> Self {
-        self.retain_by(|f| !f.is_blob())
-    }
-
-    fn retain_by<F: Fn(&Field) -> bool>(&self, predicate: F) -> Self {
-        let fields = self
-            .fields
-            .iter()
-            .filter(|f| predicate(f))
-            .cloned()
-            .collect();
-        Self {
-            fields,
-            metadata: self.metadata.clone(),
-        }
-    }
-
-    /// Splits the schema into two schemas, one with non-blob fields and the other with blob fields.
-    /// If there are no blob fields, the second schema will be `None`.
-    /// The order of fields is preserved.
-    pub fn partition_by_blob_columns(&self) -> (Self, Option<Self>) {
-        let mut non_blob_fields = Vec::with_capacity(self.fields.len());
-        let mut blob_fields = Vec::with_capacity(self.fields.len());
-        for field in self.fields.iter() {
-            if field.is_blob() {
-                blob_fields.push(field.clone());
-            } else {
-                non_blob_fields.push(field.clone());
-            }
-        }
-        (
-            Self {
-                fields: non_blob_fields,
-                metadata: self.metadata.clone(),
-            },
-            if blob_fields.is_empty() {
-                None
-            } else {
-                Some(Self {
-                    fields: blob_fields,
-                    metadata: self.metadata.clone(),
-                })
-            },
-        )
-    }
-
     pub fn has_dictionary_types(&self) -> bool {
         self.fields.iter().any(|f| f.has_dictionary_types())
     }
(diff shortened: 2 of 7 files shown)


rust/lance-table/src/format/manifest.rs
```diff
@@ -99,9 +99,6 @@ pub struct Manifest {
     /// is used to tell libraries how to read, write, or manage the table.
     pub table_metadata: HashMap<String, String>,
 
-    /// Blob dataset version
-    pub blob_dataset_version: Option<u64>,
-
     /* external base paths */
     pub base_paths: HashMap<u32, BasePath>,
 }
```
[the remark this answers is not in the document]
Xuanwo: That's expected.

protos/file.proto
```diff
@@ -168,24 +169,18 @@ message Field {
...
-  /// "blob" - The field is compacted into fewer rows per fragment.
-  ///
-  /// Fields that have non-default storage classes are stored in different
-  /// datasets (e.g. blob fields are stored in the nested "_blobs" dataset)
-  string storage_class = 11;
+  // Field number 11 was previously `string storage_class`.
+  // Keep it reserved so older manifests remain compatible while new writers
+  // avoid reusing the slot.
+  reserved 11;
+  reserved "storage_class";
```
westonpace, line 183: :+1:

rust/lance-core/src/datatypes/field.rs
```diff
@@ -32,7 +31,7 @@ use super::{
 };
 use crate::{datatypes::BLOB_DESC_LANCE_FIELD, Error, Result};
 
-pub const LANCE_STORAGE_CLASS_SCHEMA_META_KEY: &str = "lance-schema:storage-class";
+const LEGACY_STORAGE_CLASS_METADATA_KEY: &str = "lance-schema:storage-class";
```
westonpace: Can we get rid of it instead of just making it private?

rust/lance-core/src/datatypes/field.rs
```diff
@@ -978,18 +931,24 @@ impl TryFrom<&ArrowField> for Field {
...
+            .map(|v| v.to_ascii_lowercase())
+        {
+            if value == "blob" {
+                metadata
+                    .entry(lance_arrow::BLOB_META_KEY.to_string())
+                    .or_insert_with(|| "true".to_string());
+            }
+            metadata.remove(LEGACY_STORAGE_CLASS_METADATA_KEY);
+        }
+
```
westonpace: Why not just ignore?  No need for this complexity.

rust/lance-core/src/datatypes/field.rs
```diff
@@ -1032,6 +990,7 @@ impl From<&Field> for ArrowField {
     fn from(field: &Field) -> Self {
         let out = Self::new(&field.name, field.data_type(), field.nullable);
         let mut metadata = field.metadata.clone();
+        metadata.remove(LEGACY_STORAGE_CLASS_METADATA_KEY);
```
westonpace: Again, no need to remove, field metadata is harmless and no one should be adding it anyways.

rust/lance-core/src/datatypes/schema.rs
```diff
@@ -159,28 +167,29 @@ impl Schema {
...
 
-    /// Splits the schema into two schemas, one with default storage class fields and the other with blob storage class fields.
-    /// If there are no blob storage class fields, the second schema will be `None`.
+    /// Splits the schema into two schemas, one with non-blob fields and the other with blob fields.
+    /// If there are no blob fields, the second schema will be `None`.
     /// The order of fields is preserved.
-    pub fn partition_by_storage_class(&self) -> (Self, Option<Self>) {
-        let mut local_fields = Vec::with_capacity(self.fields.len());
-        let mut sibling_fields = Vec::with_capacity(self.fields.len());
+    pub fn partition_by_blob_columns(&self) -> (Self, Option<Self>) {
```
westonpace: Can we just get rid of this method instead of renaming it?

rust/lance-file/src/datatypes.rs
```diff
@@ -29,6 +31,19 @@ impl From<&pb::Field> for Field {
...
+            .unwrap_or(false);
+        if legacy_blob_marker {
+            lance_metadata
+                .entry(lance_arrow::BLOB_META_KEY.to_string())
+                .or_insert_with(|| "true".to_string());
+        }
+        // Drop the legacy storage class metadata key after translating it to blob semantics.
+        if legacy_blob_marker {
+            lance_metadata.remove(LEGACY_STORAGE_CLASS_METADATA_KEY);
+        }
```
westonpace: I don't think we need to translate or migrate.

rust/lance-core/src/datatypes/field.rs
Xuanwo: updated

rust/lance-core/src/datatypes/field.rs
Xuanwo: Removed

rust/lance-file/src/datatypes.rs
Xuanwo: Yep, let's just remove it.

rust/lance-core/src/datatypes/schema.rs
Xuanwo: Yep, let's remove it.

PR #5134, opened 2025-11-04
feat: add `Scanner::projection_from_schema`

teh-cmc:
Just exposing `ProjectionPlan::from_schema` all the way through.


jackye1995:
we typically don't just expose something and don't use it, what is the use case here?

PR #5137, opened 2025-11-04
[WIP] refactor: distinguish row id and row addr part-2

yanghua:


python/python/tests/forward_compat/test_compat.py
```diff
@@ -100,6 +100,7 @@ def test_scans():
     Version(lance.__version__) < Version("0.29.1.beta2"),  # at least 0.29.1-beta.2
     reason="Lance 0.29.1-beta.2 would ignore indices too new",
 )
+@pytest.mark.skip
```
yanghua: Need to remove this annotation after the next release and add a new conditional skip.

PR #5140, opened 2025-11-04
fix: handle logical rows deletion properly for zonemap and bloomfilter

### LINKED ISSUE #4758: Zone map and bloom filter don't seem to handle deletions correctly
I am getting fewer results than expected from queries when there is a zone map filter and deletions.

## Reproduction

```python
import pyarrow as pa
import lance

data = pa.table({
    "id": range(10),
    "value": [True, False] * 5,
})
ds.delete("NOT value") # Works if I comment this out

ds.to_table(filter="value")
```
```
pyarrow.Table
id: int64
value: bool
----
id: [[0,2,4,6,8]]
value: [[true,true,true,true,true]]
```

But if I add a zone map index, I get only the first three results:
```python
ds.create_scalar_index("value", "ZONEMAP")
ds.to_table(filter="value")
```
```
pyarrow.Table
id: int64
value: bool
----
id: [[0,2,4]]
value: [[true,true,true]]
```

HaochengLIU:
The old zonemap && bloomfilter use the logical row concept to define zones. It breaks when the rows are not contiguous, e.g. deletion. This PR now ensures that the fragment offset is taken into consideration so that the new "physical row  address" handles deletion scenario.

Both Rust and Python tests are added.

close #4758

### DIFF
--- a/rust/lance/src/index/scalar.rs
+++ b/rust/lance/src/index/scalar.rs
@@ -546,8 +546,8 @@ mod tests {
     use lance_core::utils::tempfile::TempStrDir;
     use lance_core::{datatypes::Field, utils::address::RowAddress};
     use lance_datagen::array;
-    use lance_index::pbold::NGramIndexDetails;
     use lance_index::IndexType;
+    use lance_index::{pbold::NGramIndexDetails, scalar::BuiltinIndexType};
     use lance_table::format::pb::VectorIndexDetails;
 
     fn make_index_metadata(
@@ -1186,6 +1186,84 @@ mod tests {
     }
 
     #[tokio::test]
+    async fn test_zonemap_with_deletions() {
+        let deletion_predicates = [
+            "NOT value",            // every other row
+            "id > 8191 or id < 10", // Second zone of each fragment
+            "id < 9190 ",           // Most of first zone
+        ];
+        let query_predicates = ["value", "id <= 8191", "id >= 1"];
+
+        async fn filter_query(ds: &Dataset, query_pred: &str) -> arrow_array::RecordBatch {
+            ds.scan()
+                .filter(query_pred)
+                .unwrap()
+                .try_into_batch()
+                .await
+                .unwrap()
+        }
+
+        for del_pred in &deletion_predicates {
+            // We use 2 * 8192 so each fragment has two zones.
+            let mut ds = lance_datagen::gen_batch()
+                .col("id", array::step::<UInt64Type>())
+                .col("value", array::cycle_bool(vec![true, false]))
+                .into_ram_dataset(FragmentCount::from(2), FragmentRowCount::from(2 * 8192))
+                .await
+                .unwrap();
+
+            // Create zonemap index on "value" column
+            let params = ScalarIndexParams::for_builtin(BuiltinIndexType::ZoneMap);
+            ds.create_index_builder(&["value"], IndexType::Scalar, &params)
+                .name("value_zone_map".into())
+                .await
+                .unwrap();
+            ds.create_index_builder(&["id"], IndexType::Scalar, &params)
+                .name("id_zone_map".into())
+                .await
+                .unwrap();
+
+            ds.delete(del_pred).await.unwrap();
+            let mut result_before = Vec::new();
+            for query_pred in &query_predicates {
+                let batch = filter_query(&ds, query_pred).await;
+                result_before.push(batch);
+            }
+            ds.drop_index("value_zone_map").await.unwrap();
+            ds.drop_index("id_zone_map").await.unwrap();
+
+            let mut expected = Vec::new();
+            for query_pred in &query_predicates {
+                let batch = filter_query(&ds, query_pred).await;
+                expected.push(batch);
+            }
+
+            for (before, expected) in result_before.iter().zip(expected.iter()) {
+                assert_eq!(before, expected, "Zonemap index with deletions returned wrong results for deletion predicate '{}'", del_pred);
+            }
+
+            // Now recreate the indexes for the next iteration
+            ds.create_index_builder(&["value"], IndexType::Scalar, &params)
+                .name("value_zone_map".into())
+                .await
+                .unwrap();
+            ds.create_index_builder(&["id"], IndexType::Scalar, &params)
+                .name("id_zone_map".into())
+                .await
+                .unwrap();
+            let mut result_after = Vec::new();
+            for query_pred in &query_predicates {
+                let batch = filter_query(&ds, query_pred).await;
+                result_after.push(batch);
+            }
+
+            for (after, expected) in result_after.iter().zip(expected.iter()) {
+                assert_eq!(after, expected, "Zonemap index with deletions returned wrong results for deletion predicate '{}' after re-creating the index", del_pred);
+            }
+        }
+    }
+
+    #[tokio::test]
     async fn test_zonemap_deletion_then_index() {
         use arrow::datatypes::UInt64Type;
         use lance_datagen::array;
@@ -1239,14 +1317,11 @@ mod tests {
             .await
             .unwrap();
 
-        let after_ids: Vec<u64> = after_index["id"]
+        let after_ids = after_index["id"]
             .as_any()
             .downcast_ref::<arrow_array::UInt64Array>()
             .unwrap()
-            .values()
-            .iter()
-            .copied()
-            .collect();
+            .values();
 
         // This assertion will FAIL if bug #4758 is present
         assert_eq!(
@@ -1258,7 +1333,7 @@ mod tests {
         );
         assert_eq!(
             after_ids,
-            vec![0, 2, 4, 6, 8],
+            &[0, 2, 4, 6, 8],
             "Zonemap index with deletions returns wrong results"
         );
     }


rust/lance/src/utils.rs
```diff
@@ -5,7 +5,8 @@
 
 pub(crate) mod future;
 pub(crate) mod temporal;
+// Public test utilities module - only available during testing
 #[cfg(test)]
-pub(crate) mod test;
+pub mod test;
```
HaochengLIU: so it can be used by other modules

rust/lance/src/utils/test.rs
```diff
@@ -363,6 +363,12 @@ pub struct NoContextTestFixture {
     pub dataset: Dataset,
 }
 
+impl Default for NoContextTestFixture {
```
HaochengLIU, line 366: To suppress clippy warnings

rust/lance/src/index/scalar.rs
```diff
@@ -1184,4 +1184,527 @@ mod tests {
...
+        let after_ids: Vec<u64> = after_index[0]
+            .column_by_name("id")
+            .unwrap()
+            .as_any()
+            .downcast_ref::<arrow_array::UInt64Array>()
+            .unwrap()
+            .values()
+            .iter()
+            .copied()
+            .collect();
```
wjones127: You can simplify this a lot:

1. Use `try_into_batch()` to just get one `RecordBatch` from the scan
2. Since we don't mind panicking in tests, you can directly index into the column with `after_index["id"]`
3. Instead of collecting the values to a vec, just grab the `ScalarBuffer`. This implements `AsRef<[T]>`, so you can compare it to a slice in the `assert_eq!()` call.
```suggestion
        let after_index = ds
            .scan()
            .filter("value")
            .unwrap()
            .try_into_batch()
            .await
            .unwrap();

        let after_ids = after_index["id"]
            .as_any()
            .downcast_ref::<arrow_array::UInt64Array>()
            .unwrap()
            .values();
```

rust/lance/src/index/scalar.rs
```diff
@@ -1184,4 +1184,527 @@ mod tests {
...
+        assert_eq!(
+            after_ids.len(),
+            5,
+            "Expected 5 rows after index creation, got {}. Only {:?} returned instead of [0, 2, 4, 6, 8]",
+            after_ids.len(),
+            after_ids
+        );
+        assert_eq!(
+            after_ids,
+            vec![0, 2, 4, 6, 8],
```
wjones127:
```suggestion
            &[0, 2, 4, 6, 8],
```

rust/lance/src/utils.rs
wjones127: What modules are using this?

rust/lance-index/src/scalar/bloomfilter.rs
```diff
@@ -52,10 +52,11 @@ const BLOOMFILTER_INDEX_VERSION: u32 = 0;
 #[derive(Debug, Clone)]
 struct BloomFilterStatistics {
     fragment_id: u64,
-    // zone_start is the start row of the zone in the fragment, also known
-    // as local row offset
+    // zone_start is the actual first row address (local offset within fragment)
```
wjones127: Local offset and row address are **not** the same thing.
```suggestion
    // zone_start is the start row of the zone in the fragment, also known
    // as local row offset. To get the first row address, you can do 
    // `fragment_id << 32 + zone_start`.
```

rust/lance-index/src/scalar/bloomfilter.rs
```diff
@@ -52,10 +52,11 @@ const BLOOMFILTER_INDEX_VERSION: u32 = 0;
...
     zone_start: u64,
-    zone_length: usize,
+    // zone_length is the address span: (last_row_addr - first_row_addr + 1)
+    // AKA offset in the fragment, which allows handling non-contiguous addresses after deletions
```
wjones127: I don't understand this comment. What does it mean? How are deletions handled with respect to this?

rust/lance-index/src/scalar/zonemap.rs
```diff
@@ -61,10 +61,11 @@ struct ZoneMapStatistics {
     // only apply to float type
     nan_count: u32,
     fragment_id: u64,
-    // zone_start is the start row of the zone in the fragment, also known
-    // as local row offset
+    // zone_start is the actual first row address (local offset within fragment)
```
wjones127: Same comment applies here

rust/lance-index/src/scalar/zonemap.rs
```diff
@@ -807,11 +820,30 @@ impl ZoneMapIndexBuilder {
                 if desired > remaining {
                     // Not enough data to fill a map, just increment counts
                     self.update_stats(&data_array.slice(array_offset, remaining))?;
+
+                    // Track first and last row addresses (local offsets within fragment)
+                    let first_addr = row_addrs_array.value(array_offset) & 0xFFFFFFFF;
```
wjones127: What does this do?

For clarity, could you use the `RowAddress` struct to manipulate row ids?

https://github.com/lancedb/lance/blob/5ba47fbd30c03f7787c4a43a3a7cad6012dcfad8/rust/lance-core/src/utils/address.rs#L7

rust/lance/src/index/scalar.rs
```diff
@@ -1184,4 +1184,527 @@ mod tests {
...
+        use lance_datagen::array;
+        use lance_index::scalar::{BuiltinIndexType, ScalarIndexParams};
+        use lance_index::IndexType;
+
+        // Create dataset with 10 rows: alternating boolean values
+        // Rows 0,2,4,6,8 have value=true, rows 1,3,5,7,9 have value=false
+        let mut ds = lance_datagen::gen_batch()
+            .col("id", array::step::<UInt64Type>())
+            .col("value", array::cycle_bool(vec![true, false]))
+            .into_ram_dataset(FragmentCount::from(1), FragmentRowCount::from(10))
```
wjones127: Whenever you are testing something involving row addresses, it's really important you test with multiple fragments. You'll miss a lot of bugs if you don't.


```suggestion
            .into_ram_dataset(FragmentCount::from(2), FragmentRowCount::from(10))
```

wjones127 (changes requested):

rust/lance/src/utils.rs
HaochengLIU: dead code. Will remove.

rust/lance/src/index/scalar.rs
HaochengLIU: It's cool... will rewrite

rust/lance-index/src/scalar/bloomfilter.rs
HaochengLIU: true, will fix the wrong wording

rust/lance-index/src/scalar/bloomfilter.rs
HaochengLIU: Rewrote the whole part and added two examples

rust/lance/src/index/scalar.rs
HaochengLIU: updated

rust/lance-index/src/scalar/zonemap.rs
HaochengLIU: updated

rust/lance-index/src/scalar/zonemap.rs
HaochengLIU: updated

rust/lance-index/src/scalar/bloomfilter.rs
```diff
@@ -49,12 +50,29 @@ const BLOOMFILTER_ITEM_META_KEY: &str = "bloomfilter_item";
...
+// Example: Suppose we have two fragments, each with 4 rows. Deletion is 0 index based.
+// We delete the 0th and 1st row in fragment 0, and the 1st and 2nd row in fragment 1,
+// now building the bloom filter index:
+// Fragment 0: zone_start = 2, zone_length = 2 // covers rows 2, 3 in fragment 0
+// The row addresses for fragment 0 are: 2, 3
+// Fragment 1: zone_start = 0, zone_length = 4  // covers rows 0, 3 in fragment 1
+// The row addresses for fragment 1 are: 32>>1, 32>>1 + 3
 #[derive(Debug, Clone)]
 struct BloomFilterStatistics {
     fragment_id: u64,
```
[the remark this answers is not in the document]
HaochengLIU: @wjones127 mind reviewing again? I think I've addressed all the comments from the first round review.
Above is one thing I really dont like my init implementation... The type of the annotation was really confusing

rust/debug.output
```diff
@@ -0,0 +1,43 @@
+diff --git a/Cargo.lock b/Cargo.lock
```
westonpace: Drop this file?

rust/lance-index/src/scalar/bloomfilter.rs
```diff
@@ -49,12 +50,28 @@ const BLOOMFILTER_ITEM_META_KEY: &str = "bloomfilter_item";
...
+// The row addresses for fragment 0 are: 0, 1, 2, 3
+// Fragment 1: zone_start = 0, zone_length = 4  // covers rows 0, 1, 2, 3 in fragment 1
+// The row addresses for fragment 1 are: 32>>1, 32>>1 + 1, 32>>1 + 2, 32>>1 + 3
+//
+// Deletion is 0 index based. We delete the 0th and 1st row in fragment 0,
+// and the 1st and 2nd row in fragment 1,
+// Fragment 0: zone_start = 2, zone_length = 2 // covers rows 2, 3 in fragment 0
+// The row addresses for fragment 0 are: 2, 3
+// Fragment 1: zone_start = 0, zone_length = 4  // covers rows 0, 3 in fragment 1
+// The row addresses for fragment 1 are: 32>>1, 32>>1 + 3
```
westonpace, line 65: Is this example describing `BloomFilterStatistics`?

HaochengLIU:
> This seems correct but there is a lot of similarity between zone map and bloom filter still. Do you want to create an issue to reduce the code by creating some common abstractions? That way we can remember to tackle this at some point.

https://github.com/lancedb/lance/issues/5230

HaochengLIU:
@wjones127 gentle ping, I will beout starting from next Tuesday for a trip, if possible would like to merge the fix before I'm gone

wjones127:
Test failures unrelated.

PR #5145, opened 2025-11-05
fix: remove unnecessary object store initialization

zhangyue19921010:
When `self.object_store` exists, there is no need to perform the initialization operation of object_store again.


zhangyue19921010:
Update: Failed Unit tests are all passed on my local env

Xuanwo (approved):
Makes sense to me, thank you for working on this!

PR #5151, opened 2025-11-05
fix: use typed LanceNamespace for python storage options provider

jackye1995:
Fix the error that we did not use proper request response style calls to perform describe_table in python.

### DIFF
--- a/python/python/lance/dataset.py
+++ b/python/python/lance/dataset.py
@@ -34,6 +34,7 @@ from typing import (
 
 import pyarrow as pa
 import pyarrow.dataset
+from lance_namespace import DescribeTableRequest, LanceNamespace
 from pyarrow import RecordBatch, Schema
 
 from lance.log import LOGGER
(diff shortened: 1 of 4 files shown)


python/python/lance/dataset.py
```diff
@@ -5230,22 +5231,21 @@ def write_dataset(
...
 
-        # Merge initial storage options from describe_table with user-provided options
-        namespace_storage_options = table_info.get("storage_options", {})
-        if storage_options:
-            # User-provided options take precedence
-            merged_storage_options = {**namespace_storage_options, **storage_options}
-        else:
-            merged_storage_options = namespace_storage_options
-        storage_options = merged_storage_options
+        namespace_storage_options = response.storage_options
```
jackye1995, line 5240: make it consistent with how `lance.dataset` handles the storage options

python/python/lance/dataset.py
```diff
@@ -5230,22 +5231,21 @@ def write_dataset(
...
+        if namespace_storage_options:
+            # TODO: support dynamic storage options provider
```
jackye1995, line 5242: this I will put a separated PR, need to properly handle CreateEmptyTable case

bryanck (approved):

LuQQiu (approved):

PR #5159, opened 2025-11-06
fix: prevent panic with zero dimension vectors

### LINKED ISSUE #5102: Zero dimension vectors cause Lance to panic
Not that this is useful, but it should return an error instead of panicking.

```python
import pyarrow as pa
import lance
schema = pa.schema([
    pa.field("id", pa.int64()),
    pa.field("vec", pa.list_(pa.float32(), 0)),  # Fixed size = 0
])
data = pa.table({
    "id": [1, 2, 3],
    "vec": [[], [], []],  # Empty arrays matching the 0 dimension
}, schema=schema)

ds.to_table()
```

```
thread 'lance_background_thread' panicked at /home/will/Documents/lance/rust/lance-encoding/src/data.rs:403:9:
attempt to divide by zero

thread 'lance_background_thread' panicked at /home/will/Documents/lance/rust/lance-table/src/utils/stream.rs:390:46:
called `Result::unwrap()` on an `Err` value: JoinError::Panic(Id(46), "attempt to divide by zero", ...)
  File "<stdin>", line 1, in <module>
  File "/home/will/Documents/lance/python/python/lance/dataset.py", line 1073, in to_table
    ).to_table()
  File "/home/will/Documents/lance/python/python/lance/dataset.py", line 4631, in to_table
  File "pyarrow/ipc.pxi", line 762, in pyarrow.lib.RecordBatchReader.read_all
pyarrow.lib.ArrowInvalid: External error: RuntimeError: Task was aborted
```

Maxwell-Guo:
Fix the problem that zero dimension vectors cause Lance to panic.

Fixes #5102

### DIFF
--- a/rust/lance-encoding/src/data.rs
+++ b/rust/lance-encoding/src/data.rs
@@ -399,8 +399,14 @@ pub struct FixedSizeListBlock {
 }
 
 impl FixedSizeListBlock {
-    pub fn num_values(&self) -> u64 {
-        self.child.num_values() / self.dimension
+    pub fn num_values(&self) -> Result<u64> {
+        let child_values = self.child.num_values()?;
+        child_values
+            .checked_div(self.dimension)
+            .ok_or(Error::InvalidInput {
+                source: "Division by zero: dimension cannot be zero".into(),
+                location: location!(),
+            })
     }
 
     /// Try to flatten a FixedSizeListBlock into a FixedWidthDataBlock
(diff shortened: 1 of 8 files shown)


wjones127 (commented):
Seeing all the places we need to change to return a result, it does make me wonder if it would have been simpler to just return zero if dividing by zero, and handle zero-dimensional lists elsewhere. What do you think?

BTW, here's a tests we should change to validate this:

https://github.com/lancedb/lance/blob/5ba47fbd30c03f7787c4a43a3a7cad6012dcfad8/rust/lance-encoding/src/encodings/fuzz_tests.rs#L366

It should become

```rust
        list_size in 0..=100i32,
```

Maxwell-Guo:
Hey @wjones127 , what about just panic here and we handle zero-dimensional lists elsewhere ?

wjones127:
> Hey @wjones127 , what about just panic here and we handle zero-dimensional lists elsewhere ?

Could also do that. Most important thing we don't panic for that input overall.

Maxwell-Guo:
@wjones127 let me fix this issue in this way.

PR #5162, opened 2025-11-06
fix: unify max_rows_per_group calibration

zhangyue19921010:
unify max_rows_per_group param calibration

image

### DIFF
--- a/rust/lance/src/dataset/write.rs
+++ b/rust/lance/src/dataset/write.rs
@@ -7,7 +7,7 @@ use datafusion::physical_plan::stream::RecordBatchStreamAdapter;
 use datafusion::physical_plan::SendableRecordBatchStream;
 use futures::{Stream, StreamExt, TryStreamExt};
 use lance_core::datatypes::{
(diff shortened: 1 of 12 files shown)


rust/lance/src/dataset/write.rs
```diff
@@ -577,9 +580,6 @@ pub async fn write_fragments_internal(
         (data, schema)
     };
 
-    // Make sure the max rows per group is not larger than the max rows per file
-    params.max_rows_per_group = std::cmp::min(params.max_rows_per_group, params.max_rows_per_file);
```
yanghua, line 589: Did you check whether the `max_rows_per_group` only works under `LanceFileVersion::Legacy` in the whole codebase?

zhangyue19921010: I checked the code `max_rows_per_group` will work under: 
1. Current code location
2. https://github.com/lancedb/lance/blob/1f38a7fbc0d27ea9234aed8f1bcf5964f361c13b/rust/lance/src/dataset/fragment/write.rs#L257  The code changes here will not have any impact for case2

PR #5175, opened 2025-11-06
fix: memory-limited string BTREE index building

### LINKED ISSUE #4881: BTree index training on string column uses lots of memory
AC: have a test that shows that BTree index can run while using a limited amount of memory

wjones127:
Closes #4881

### DIFF
--- a/rust/lance-index/src/scalar/btree.rs
+++ b/rust/lance-index/src/scalar/btree.rs
@@ -982,12 +982,6 @@ impl BTreeIndex {
         new_data: SendableRecordBatchStream,
         chunk_size: u64,
     ) -> Result<SendableRecordBatchStream> {
-        let data_type = new_data.schema().field(0).data_type().clone();
-        // Datafusion currently has bugs with spilling on string columns
-        // See https://github.com/apache/datafusion/issues/10073
-        //
-        // One we upgrade we can remove this
-        let use_spilling = !matches!(data_type, DataType::Utf8 | DataType::LargeUtf8);
         let value_column_index = new_data.schema().index_of(VALUE_COLUMN_NAME)?;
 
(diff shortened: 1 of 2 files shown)


wjones127:
## Memory benchmark

Creating a BTree index on a 1GB string column with `LANCE_MEM_POOL_SIZE` set to 20MB.

<details><summary>Script</summary>
<p>

```python
# LD_PRELOAD=/usr/local/lib/libbytehound.so python test_btree.py
# bytehound server memory-profiling_*.dat
import os

os.environ['LANCE_MEM_POOL_SIZE'] = str(20 * 1024 * 1024)  # 20 MB

import pyarrow as pa
import lance
from datetime import datetime

def gen_data():
    # 1GB of strings
    for i in range(100):
        # 10MB of strings
        array_len = 10 * 1024 * 1024 // 8
        arr = pa.array([f'{j:>8}' for j in range(i * array_len, (i + 1) * array_len)])
        yield pa.RecordBatch.from_arrays(
            [arr],
            ['strings']
        )

reader = pa.RecordBatchReader.from_batches(
    schema=pa.schema([pa.field('strings', pa.string())]),
    batches=gen_data()
)

ds = lance.write_dataset(
    reader,
    './tmp',
    mode='overwrite'
)

print(datetime.now(), "Creating index")

ds.create_scalar_index('strings', index_type="BTREE")
```

</p>
</details> 


### Before: 2.7GB peak

Screenshot 2025-11-14 at 4 01 30 PM


### After: 240MB peak


Screenshot 2025-11-14 at 3 55 53 PM

I believe the memory use is from the scanner buffering IO and reading batches ahead. After the initial scan, it does a pretty good job of keeping with the memory pool amount.

PR #5176, opened 2025-11-06
fix: skip elapsed_compute timing in parent execution nodes to avoid double-counting

### LINKED ISSUE #5155: The `InstrumentedRecordBatchStreamAdapter` is often misused and double counts CPU time
The `InstrumentedRecordBatchStreamAdapter` measures CPU time by measuring the time it takes to run `poll_next()`:

https://github.com/lancedb/lance/blob/35a0547426c47dd94425619c442dcd03ffd9860c/rust/lance/src/io/exec/utils.rs#L291-L293

However, if the stream for a given execution plan call into it's children, then this will double count the time in the child nodes.

Nodes that have children generally should use this. This includes:

* `FlatMatchQueryExec`
* `KNNVectorDistanceExec`
* `AddRowAddrExec`

It's okay for things like `ScalarIndexExec` and `LancePushdownScanExec` which have no children.

We found this bug by noticing an analyze_plan output showed `KNNVectorDistanceExec` taking a lot of CPU time even though it had zero rows as input:

```
KNNVectorDistance: metric=l2, metrics=[output_rows=0, elapsed_compute=1.52917033s, output_batches=0], cumulative_cpu=2.891887865s
    FilterExec: scalar_id@1 = 100, metrics=[output_rows=0, elapsed_compute=8.036778ms], cumulative_cpu=1.362717535s
```

Also, the nodes that spawn CPU tasks should also capture their CPU usage. `KNNVectorDistanceExec` is an example of a node that spawns CPU tasks.

BorenTsai:
Fixes: https://github.com/lancedb/lance/issues/5155

## Summary
1. Adds elapsed_compute timing to the relevant parent executors
2. Adds a MetricsMode enum which controls if the `InstrumentedRecordBatchStreamAdapter` tracks the elapsed_compute time or not. 

## Tests
1. Added unittest to test metrics mode handling
2. Created sample case to capture before/after timing behavior
```bash
// Before
KNNVectorDistance: metric=l2, metrics=[output_rows=0, elapsed_compute=78.574917ms, output_batches=0], cumulative_cpu=141.68833ms
    FilterExec: id@1 > 1000000, metrics=[output_rows=0, elapsed_compute=1ns], cumulative_cpu=63.113413ms
                
// After
KNNVectorDistance: metric=l2, metrics=[output_rows=0, elapsed_compute=2ns, output_batches=0], cumulative_cpu=65.048006ms
    FilterExec: id@1 > 1000000, metrics=[output_rows=0, elapsed_compute=1ns], cumulative_cpu=65.048004ms
```

Thanks @wjones127 for the clarifications and guidance


rust/lance/src/io/exec/knn.rs
```diff
@@ -221,6 +221,7 @@ impl ExecutionPlan for KNNVectorDistanceExec {
         context: Arc<datafusion::execution::context::TaskContext>,
     ) -> DataFusionResult<SendableRecordBatchStream> {
         let input_stream = self.input.execute(partition, context)?;
+        let baseline_metrics = BaselineMetrics::new(&self.metrics, partition);
```
thexiay, line 224: `BaselineMetrics` this class is a bit redundant. It can use like `MetricBuilder::new(&self.metrics).elapsed_compute(partition);` to only calculate the elapsed time here

rust/lance/src/io/exec/knn.rs
```diff
@@ -229,7 +230,9 @@ impl ExecutionPlan for KNNVectorDistanceExec {
             .map(move |batch| {
                 let key = key.clone();
                 let column = column.clone();
+                let baseline_metrics = baseline_metrics.clone();
                 async move {
```
thexiay, line 234: Maybe wrapping this future to a struct like `InstrumentedRecordBatchFutureAdapter` would be more general

PR #5178, opened 2025-11-07
ci: create a new library compatibility test suite

### LINKED ISSUE #4416: Broad backwards- & forwards-compatability test
We want broader testing that datasets written with old version of libraries are readable with new libraries.

# Coverage

For wide coverage, we can have one set of data that we write. It should have most possible data types (all primitives plus common nested combinations) and all index types (with some parameter variations).

We should also vary:

* Lance file version (don't need to care about unstable versions)
* Lance index versions
* Field compression settings

## Forwards compat testing

See: 

* [datagen.py](https://github.com/lancedb/lance/blob/main/python/python/tests/forward_compat/datagen.py)
* [test_compat.py](https://github.com/lancedb/lance/blob/main/python/python/tests/forward_compat/test_compat.py)

* [ ] Scalar index types
    * [x] JSON index
    * [x] N Gram
    * [x] Zone map
    * [x] BTree
    * [x] Bitmap
    * [ ] IVF-PQ
    * [ ] IVF_HNSW_PQ
    * [ ] IVF HNSW SQ
    * [x] FTS
    * [x] Label list 
    * [x] Bloom filter

## Backwards compat

* [ ] Scalar index types
    * [ ] JSON index
    * [ ] N Gram
    * [ ] Zone map
    * [ ] BTree
    * [ ] Bitmap
    * [ ] IVF-PQ
      * [ ] Old index versions
    * [ ] IVF_HNSW_PQ
    * [ ] IVF HNSW SQ
    * [ ] FTS
    * [ ] Label list 
    * [ ] Bloom filter

# Tests

## Forwards compat

* [x] Full scan
* [x] Query the index
* [ ] Test optimize indices
* [ ] Test compaction
* [ ] Test appending

## Backwards compat

* [ ] Full scan
* [ ] Query the index
* [ ] Test optimize indices
* [ ] Test compaction
* [ ] Test appending

wjones127:
This adds a new `compat` test suite that replaces the `forwards_compat` suite. Key changes:

1. It tests both forwards **and** backwards compatibility.
2. It adds the most recent stable and beta release as comparison targets. Individual tests can also choose which versions to test against.
3. It can be run easily locally via a single command `make compattest`

Closes #4416

## How it works

Tests are written like:

```python
@compat_test()
class MyTest(UpgradeDowngradeTest):
    def __init__(self, path: Path):
        self.path = path

    def create(self):
        # write initial data

    def check_read(self):
        # test reading data

    def check_write(self):
        # test writing data
```

Then two tests will be generated:

1. **downgrade**: We call `create` in current version, then `check_read` and `check_write` in old version.
2. **upgrade_downgrade**: We call `create` in old version, then `check_read` and `check_write` in current version, then go back and call `check_read` and `check_write` in old version again.

The way we execute on old versions of the library is using a persistent executor process within special virtual environment. We create one virtual env per old version, and one executor process per virtual env. The executor process receives a tuple `(TestObject, method_name)` (for example, `(MyTest, 'create')`), which is serialized via pickle. The executor process runs the method, and returns back a status indicating if there were any errors.

### DIFF
--- a/python/python/tests/compat/test_vector_indices.py
+++ b/python/python/tests/compat/test_vector_indices.py
@@ -16,22 +16,12 @@ import pyarrow as pa
 import pyarrow.compute as pc
 
 from .compat_decorator import (
-    LAST_BETA_RELEASE,
-    LAST_STABLE_RELEASE,
     UpgradeDowngradeTest,
     compat_test,
 )
 
 
-@compat_test(
-    versions=[
-        "0.29.1.beta2",
-        "0.30.0",
-        "0.36.0",
-        LAST_STABLE_RELEASE,
-        LAST_BETA_RELEASE,
-    ]
-)
+@compat_test(min_version="0.29.1.beta2")
 class PqVectorIndex(UpgradeDowngradeTest):
     """Test PQ (Product Quantization) vector index compatibility."""
 
@@ -92,13 +82,7 @@ class PqVectorIndex(UpgradeDowngradeTest):
         ds.optimize.compact_files()
 
 
-@compat_test(
-    versions=[
-        "0.39.0",
-        LAST_STABLE_RELEASE,
-        LAST_BETA_RELEASE,
-    ]
-)
+@compat_test(min_version="0.39.0")
 class HnswPqVectorIndex(UpgradeDowngradeTest):
     """Test IVF_HNSW_PQ vector index compatibility.
 
@@ -163,13 +147,7 @@ class HnswPqVectorIndex(UpgradeDowngradeTest):
         ds.optimize.compact_files()
 
 
-@compat_test(
-    versions=[
-        "0.39.0",
-        LAST_STABLE_RELEASE,
-        LAST_BETA_RELEASE,
-    ]
-)
+@compat_test(min_version="0.39.0")
 class HnswSqVectorIndex(UpgradeDowngradeTest):
     """Test IVF_HNSW_SQ vector index compatibility.
 
--- a/python/python/tests/compat/compat_decorator.py
+++ b/python/python/tests/compat/compat_decorator.py
@@ -12,34 +12,57 @@ import inspect
 import json
 import subprocess
 import sys
+import urllib.request
 from functools import lru_cache
+from typing import List
 
 import pytest
+from packaging.version import Version
 
 
 @lru_cache(maxsize=1)
-def last_stable_release():
-    """Returns the latest stable version available on PyPI.
-
-    Queries the PyPI JSON API to get the latest stable release of pylance.
-    Results are cached to avoid repeated network calls.
-    """
+def pylance_stable_versions() -> List[Version]:
+    """Fetches and returns a sorted list of stable pylance versions from PyPI."""
     try:
-        import urllib.request
-
         with urllib.request.urlopen(
             "https://pypi.org/pypi/pylance/json", timeout=5
         ) as response:
             data = json.loads(response.read())
-            version = data["info"]["version"]
-            return version
+            releases = data["releases"].keys()
+            stable_versions = [
+                Version(v)
+                for v in releases
+                if not any(c in v for c in ["a", "b", "rc"])
+            ]
+            stable_versions.sort()
+            return stable_versions
     except Exception as e:
-        # If we can't fetch, return None which will be filtered out
         print(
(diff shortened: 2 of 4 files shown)


python/python/tests/compat/venv_runner.py
```diff
@@ -0,0 +1,106 @@
...
+            obj, method_name = request
+
+            # Execute method with timing
+            start_time = time.time()
+            if DEBUG:
+                print(
+                    f"[VENV TIMING] Executing {method_name}...",
+                    file=sys.stderr,
+                    flush=True,
+                )
```
westonpace, line 77: Is this timing still useful?  Or was it just for developing this framework?

python/python/tests/compat/venv_runner.py
```diff
@@ -0,0 +1,106 @@
...
+
+            method = getattr(obj, method_name)
```
westonpace, line 79: Why wrap the method in an object?  Why not just pickle the method itself?  Then you could use `__name__` to get the method name for logging.

python/python/tests/compat/venv_manager.py
```diff
@@ -0,0 +1,354 @@
...
+        if DEBUG:
+            print(
+                f"[TIMING]   venv creation: {time.time() - venv_start:.2f}s", flush=True
+            )
+
+        # Determine python path in venv
+        if sys.platform == "win32":
+            self.python_path = self.venv_path / "Scripts" / "python.exe"
+        else:
+            self.python_path = self.venv_path / "bin" / "python"
```
westonpace: This is repeated 3 times. Maybe just put in a helper method?

python/python/tests/compat/test_vector_indices.py
```diff
@@ -0,0 +1,234 @@
...
+    UpgradeDowngradeTest,
+    compat_test,
+)
+
+
+@compat_test(
+    versions=[
+        "0.29.1.beta2",
+        "0.30.0",
+        "0.36.0",
```
westonpace: Looks like we still have a hard-coded list of versions.  At one point we had discussed using "the latest X major releases"  Is that still a follow-up or do we want this approach for some reason?

westonpace (approved):
A few minor nits.  I suppose one question would be why is this approach better than the old CI-based one?

wjones127:
> I suppose one question would be why is this approach better than the old CI-based one?

Main advantage is that anyone can run this locally with one command: `make compattest`. Before they had to run a sequence of steps.

python/python/tests/compat/test_vector_indices.py
wjones127: We do the last stable and last beta release. We could expand that further.

Separately, there's the issue of whether we want to hard code versions. Some of the hardcoded ones here are just representing the first time we have compatibility working for that feature. I think at least that version is worth hard-coding in.

We might be able to take some of these versions out in the middle. I think some are there for no particular reason. But it's possible we might want to record some specific version in there if there was a risky compatibility change that happened in that version.

python/python/tests/compat/venv_runner.py
wjones127: Ah I can delete that.

python/python/tests/compat/venv_manager.py
wjones127: Yeah that can be simplified.

python/python/tests/compat/venv_runner.py
wjones127: I suppose there's two ways you could go here:

1. (What I did): Put all the parameters (such as the temp directory) in the class attributes. Then you can just pickle the object and call it's method without having to pass them.
2. Pickle the function and the parameters separately, and pass the parameters in on the other side.

It doesn't seem like pickle works with closures. So I couldn't just create a closure with the parameters passed in, pickle it, and run it on the other side.

I felt like option (1) was a simpler way to implement this. Plus I kind of liked using the class to organize the methods into a logical grouping.

westonpace: Makes sense to me

python/python/tests/compat/test_vector_indices.py
wjones127: Thinking about it, I think what I'd like to do is for all test, have it test against:

* Last stable release
* Last preview release
* Last 2 major releases

I can make that change.

PR #5189, opened 2025-11-07
feat: add blob compaction support

Xuanwo:
This is just the first step: the next steps include streaming blobs and adding a maximum limit to prevent OOMs.

---


python/python/tests/test_optimize.py
```diff
@@ -36,6 +37,41 @@ def test_dataset_optimize(tmp_path: Path):
...
+    )
+
+    dataset = lance.write_dataset(
+        table,
+        base_dir,
+        schema=schema,
+        max_rows_per_file=1,
+        data_storage_version="stable",
+    )
+    assert len(dataset.get_fragments()) == 2
```
wjones127: Does this measure anything meaningful? Aren't the blob fragments different than regular fragments?

rust/lance/src/dataset/optimize.rs
```diff
@@ -155,6 +156,11 @@ pub struct CompactionOptions {
     /// not be remapped during this compaction operation. Instead, the fragment reuse index
     /// is updated and will be used to perform remapping later.
     pub defer_index_remap: bool,
+    /// Whether to rewrite blob columns during compaction.
+    ///
+    /// When enabled, blob columns are streamed as binary data so they can be
+    /// rewritten safely. Datasets containing blob columns must set this flag.
+    pub compact_blobs: bool,
```
wjones127: > Datasets containing blob columns must set this flag.

What happens if they don't set this? If it's required, why can't we detect this automatically from the schema?

wjones127 (commented):
It's unclear to me whether we need this flag. Was that discussed somewhere?

rust/lance/src/dataset/optimize.rs
Xuanwo: Sorry for not bringing up this discussion earlier. Compact blobs lead to high memory usage, which can easily cause OOM. I think it's better not to enable compact blobs by default.

If users didn't set this, they will meet the same error like before.

python/python/lance/dataset.py
```diff
@@ -4809,6 +4810,9 @@ def compact_files(
             to reduce this if you are running out of memory during compaction.
 
             The default will use the same default from ``scanner``.
+        compact_blobs: bool, default False
+            When True, allows datasets containing blob columns to be compacted.
+            Required if any field has ``metadata={"lance-encoding:blob": "true"}``.
```
westonpace: Why would I ever set this to false?

Xuanwo:
Thank you @wjones127 and @westonpace for the review, I have removed the `compact_blobs` flag, PTAL.

rust/lance/src/dataset/scanner.rs
```diff
@@ -627,7 +629,24 @@ impl Scanner {
...
+    }
+
+    fn apply_blob_handling(&mut self) {
+        let projection = self
+            .projection_plan
+            .physical_projection
+            .clone()
+            .with_blob_handling(self.blob_handling.clone());
+        self.projection_plan.physical_projection = projection;
+    }
```
westonpace, line 644: Do we also need to call `apply_blob_handling` in `Scanner::project_with_transform`?

Xuanwo: Yep, we already did that 😆 

https://github.com/lancedb/lance/pull/5189/files#diff-c25e7065f1e84cc2421b1d8b920d96038c14557fb3daadd07932f719098bc81bR732

PR #5192, opened 2025-11-07
fix: ensure recheck for IsNotNull in bloom filter

### LINKED ISSUE #5171: Bloom filter does not handle `IS NOT NULL` correctly
```python
import pyarrow as pa
import lance

data = pa.table({"value": [1.0, 2.0, None]})

ds.to_table(filter="value is not null")
```
```
pyarrow.Table
value: double
----
value: [[1,2]]
```

```python
ds.create_scalar_index("value", "BLOOMFILTER")
ds.to_table(filter="value is not null")
```
```
pyarrow.Table
value: double
----
value: []
```

Xuanwo:
This PR will fix #5171.

BloomFilter only provides `AtMost` for `IS NULL`. Our planner implements `IS NOT NULL` as `NOT (IS NULL)`, which causes the BloomFilter result to be treated as `AtLeast`. But this is incorrect for BloomFilter. Additionally, FilteredReadExec treats `AtLeast` as exact, which can drop many ranges that might contain `NOT NULL`.

We fixed this by adding a recheck before applying `NOT` to BloomFilter.

---

### DIFF
--- a/rust/lance/src/index/vector/ivf/v2.rs
+++ b/rust/lance/src/index/vector/ivf/v2.rs
@@ -670,6 +670,7 @@ mod tests {
 
     const NUM_ROWS: usize = 512;
     const DIM: usize = 32;
+    const PARTITION_SPLIT_APPEND_ROWS: usize = 50_000;
 
     async fn generate_test_dataset<T: ArrowPrimitiveType>(
         test_uri: &str,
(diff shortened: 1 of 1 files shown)


wjones127 (approved):
Thanks for fixing this!

Issue #5167, opened 2025-11-06
Blob Encoding not working with Lance Java SDK


rahil-c:
## Issue Summary

Hi all, recently I have been working on adding lance file format support within Apache Hudi: https://github.com/apache/hudi/issues/14127 and wanted to raise a potential issue(hopefully my understanding is correct of this feature)

During the integration, I wanted to test out storing binary content and leverage the `blob encoding` feature that Lance mentions here: https://lancedb.github.io/lance/guide/blob/?h=blob, to avoid materializing these blobs on each read and just get back a struct of the position and size.

However during my testing with the `LanceFileReader` via the Java sdk I would find that the blob contents are always materialized despite passing the property `"lance-encoding:blob": "true` during the intial write using the LanceFileWriter.

To isolate the issue to only lance, I have the following unit test which is just using the LanceFileWriter and LanceFileReader to show the reproduction. Should be easy to copy into an IDE and run and i have attached results as well.

## Reproduction

```
package org.apache.hudi.io.storage;

import com.lancedb.lance.file.LanceFileReader;
import com.lancedb.lance.file.LanceFileWriter;
import org.apache.arrow.memory.BufferAllocator;
import org.apache.arrow.memory.RootAllocator;
import org.apache.arrow.vector.LargeVarBinaryVector;
import org.apache.arrow.vector.VectorSchemaRoot;
import org.apache.arrow.vector.ipc.ArrowReader;
import org.apache.arrow.vector.types.pojo.ArrowType;
import org.apache.arrow.vector.types.pojo.Field;
import org.apache.arrow.vector.types.pojo.FieldType;
import org.apache.arrow.vector.types.pojo.Schema;
import org.junit.jupiter.api.Test;
import org.junit.jupiter.api.io.TempDir;

import java.nio.file.Path;
import java.util.*;

import static org.junit.jupiter.api.Assertions.*;

public class BlobEncodingTest {

    @Test
    public void testBlobEncodingReturnsDescriptors(@TempDir Path tempDir) throws Exception {
        String filePath = tempDir.resolve("test_blob.lance").toString();
        BufferAllocator allocator = new RootAllocator();

        // Step 1: Write blob-encoded data
        Map<String, String> blobMetadata = new HashMap<>();
        blobMetadata.put("lance-encoding:blob", "true");

        Field blobField = new Field(
                "blob_data",
                new FieldType(true, ArrowType.LargeBinary.INSTANCE, null, blobMetadata),
                Collections.emptyList()
[...]

PR #5193, opened 2025-11-07
fix: ensure blob encoding work when using file reader directly

rahil-c:
## Issue summary 

Raised the following issue here which provides a reproduction of using the java file reader api which is not returning back the blob encoding when setting the `"lance-encoding:blob": "true"` for a given column. https://github.com/lancedb/lance/issues/5167

## With Patch
When using the following change, and after building the java bindings locallly, I am able to run the same repro as whats attached in the original issue and see the blob encoding returned.
Screenshot 2025-11-07 at 6 37 58 AM

## Special Note
I am not super deep with lance file format and rust yet, so please feel free to correct this patch.

### DIFF
--- a/.bumpversion.toml
+++ b/.bumpversion.toml
@@ -1,5 +1,5 @@
 [tool.bumpversion]
-current_version = "3.0.0-beta.2"
+current_version = "3.0.0-beta.3"
 parse = "(?P<major>\\d+)\\.(?P<minor>\\d+)\\.(?P<patch>\\d+)(-(?P<prerelease>(beta|rc))\\.(?P<prerelease_num>\\d+))?"
 serialize = [
     "{major}.{minor}.{patch}-{prerelease}.{prerelease_num}",
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -252,7 +252,7 @@ jobs:
     runs-on: ubuntu-24.04
     strategy:
       matrix:
-        msrv: ["1.88.0"] # This should match up with rust-version in Cargo.toml
+        msrv: ["1.91.0"] # This should match up with rust-version in Cargo.toml
     env:
       # Need up-to-date compilers for kernels
       CC: clang
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -383,6 +383,22 @@ dependencies = [
 ]
 
 [[package]]
+name = "arrow-scalar"
+version = "57.0.0"
+dependencies = [
+ "arrow-array",
+ "arrow-buffer",
+ "arrow-cast",
+ "arrow-data",
+ "arrow-ord",
+ "arrow-row",
+ "arrow-schema",
+ "half",
+ "proptest",
+ "rstest 0.23.0",
+]
+
+[[package]]
 name = "arrow-schema"
 version = "57.3.0"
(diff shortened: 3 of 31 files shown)


rahil-c:
cc @jackye1995 @westonpace @eddyxu

rahil-c:
Was wondering if I can have a maintainer kick off the github ci for this?

rahil-c:
> Ah, interesting use of projection, but that definitely works. However, this will force it to _always_ return blobs as descriptors. I thought the goal was to allow the user of the `LanceFileReader` to choose?

Hi @westonpace thanks for the help. I think i was running into issues changing the LanceFileReader so ended up making this change. Was wondering if you had a reference on how i change it in order to achieve the above?

I think when i was working backwards i saw the following logic https://github.com/lancedb/lance/blob/f084677fa629c45f2e757cf90fa434746dfe74a8/rust/lance-core/src/datatypes/schema.rs#L975 for how it determines to either return back the blob encoding or materialize blob, and saw that the caller of this logic was `apply_projection`https://github.com/lancedb/lance/blob/f084677fa629c45f2e757cf90fa434746dfe74a8/rust/lance-core/src/datatypes/field.rs#L264, which came from, which eventually was invoked from `to_bare_schema`https://github.com/lancedb/lance/blob/f084677fa629c45f2e757cf90fa434746dfe74a8/rust/lance-core/src/datatypes/schema.rs#L1319

Is there an easier way to ensure the blob handling is correctly passed via file reader?

rahil-c:
Hi @westonpace @eddyxu @jackye1995 , Let me know if the recent patch looks directionally better based on the discussion in the original github thread.

java/src/main/java/org/lance/file/LanceFileReader.java
```diff
@@ -126,16 +127,43 @@ private Schema load_schema() throws IOException {
...
+   * @param blobReadMode how to return blob-encoded columns: as materialized binary content ({@link
+   *     BlobReadMode#CONTENT}) or as descriptors with position and size ({@link
+   *     BlobReadMode#DESCRIPTOR})
+   * @return an ArrowReader for the Lance file
+   */
+  public ArrowReader readAll(
+      @Nullable List<String> projectedNames,
+      @Nullable List<Range> ranges,
+      int batchSize,
+      BlobReadMode blobReadMode)
```
westonpace: My only nit is that the number of options here might grow over time (e.g. JSON read mode, etc.) and so maybe instead of `BlobReadMode` you might want some kind of options object that can have new fields added to it without causing API changes.  That being said, let's do this in a follow-up if you think it is needed.

westonpace:
Looks like there are some legitimate CI failures.  Ping me when those are passing and we can merge this.

java/src/main/java/org/lance/file/LanceFileReader.java
rahil-c: @westonpace Thanks for the feedback, i tried to follow a similar pattern as for what is done in the `ReadOptions` (dataset level) and `WriteParams` by introducing a `FileReadOptions` that the api will take in.

rahil-c:
Thanks @westonpace for the assistance, ci looks green now. If there are other issues though let me know.

Issue #5115, opened 2025-10-31
Compact Operation Fails with Schema Incompatibility: Blob Field Incorrectly Interpreted as Struct Type


zszheng:
# Step 1：Import Mp4 To Lance，with blob type

```python
import lance
import pyarrow as pa
import pandas as pd
from pathlib import Path

schema = pa.schema([
    pa.field("id", pa.int64()),
    pa.field("video_name", pa.string()),
    pa.field("video_data", 
             pa.large_binary(),
             metadata={"lance-encoding:blob": "true"}
            ),
])
mp4_files_dir="./HDFS"
mp4_dir = Path(mp4_files_dir)
mp4_files = list(mp4_dir.glob("*.mp4"))

if not mp4_files:
    print(f"在目录 {mp4_files_dir} 中未找到 MP4 文件")

print(f"找到 {len(mp4_files)} 个 MP4 文件")

ids = []
video_names = []
file_sizes = []
video_data_list = []

for i, mp4_file in enumerate(mp4_files):
    try:
        with open(mp4_file, 'rb') as f:
            video_bytes = f.read()

        ids.append(i + 1)
        video_names.append(mp4_file.name)
        file_sizes.append(mp4_file.stat().st_size)
        video_data_list.append(video_bytes)

        print(f"已加载: {mp4_file.name} ({len(video_bytes)} 字节)")

    except Exception as e:
        print(f"读取文件 {mp4_file} 时出错: {e}")
        continue

if not ids:
    print("没有成功读取任何 MP4 文件")

table = pa.table({
    "id": ids,
    "video_name": video_names,
    "video_data": video_data_list
}, schema=schema)
print(f"type {type(table)}， {type(video_data_list[0])}, schema {schema}")
dataset_uri = "./example_video_dataset1.lance"
lance.write_dataset(table, dataset_uri, mode="append")
```

# Step2 ：Compact Files
```
OSError: LanceError(Schema): Attempt to project incompatible fields: Field(id=2, name=video_data, type=large_binary) and 
Field(id=2, name=video_data, type=struct, children=[Field(id=3, name=position, type=uint64), Field(id=4, name=size, 
type=uint64), ]), [/home/runner/work/lance/lance/rust/lance-core/src/datatypes/field.rs:687:31]
(http://proxy.ml.bilibili.co/home/runner/work/lance/lance/rust/lance-core/src/datatypes/field.rs#line=686)
```

# Step3：Throw Exception

```
[...]

PR #5199, opened 2025-11-09
Feature/compact blobs

zszheng:
# Issue
#5115 

# Improvement Points
The current implementation has not yet considered the reuse of Fragment processing for multiple blob Columns, which can reduce IO operations and improve Compact performance

# Verification
1. Write a batch of data to generate multiple *lance files
2. Trigger compact_files to merge data files
3. Read the actual data to check if the content matches the expected results
4. Perform cleanup, check the merging of *lance files, and query blob data


Xuanwo:
Should be supported by https://github.com/lancedb/lance/pull/5189

Issue #5164, opened 2025-11-06
Supports range-based distributed BTree index


steFaiz:
# Introduction
The BTree index is a crucial scalar index in Lance. Many thanks to @westonpace for the implementation! Its main structure is shown below:

<img width="600" height="455" alt="Image" src="https://github.com/user-attachments/assets/f0977fc0-bab4-40b1-bd05-20c2cd71530c" />

In summary, all `<value, rowAddress>` pairs are stored in a single page_data.lance in object storage. A B-tree structure from page_lookup.lance is kept in memory to enable fast access to a relatively small data block, and the final results are obtained through sequential scanning within that block. 
To accelerate index building, @xloya  contributed the distributed version (as illustrated below). The key points are: 
1. Fragments are horizentally divided. Each subtask only build BTree index for a subset of the full-list fragments.
2. A dedicated merger is responsible for k-ways merging all page files into one final page file as well as the lookup file. Since all page files are already sorted, the merge process can be relativly faster.

<img width="1502" height="448" alt="Image" src="https://github.com/user-attachments/assets/c5d4fee3-6a67-4f09-8868-e575fdc4d308" />
However, inevitably, a single point (the Merger) has to process all the data, which becomes unacceptable at very large scales (tens of billions of records or more).

# Resolution
Existing big data computing engines (such as Spark and Flink) provide mature range-shuffle capabilities, enabling efficient global sorting of datasets at the scale of hundreds of billions of records. Leveraging this, we have implemented a **range-based BTree index** using Lance-Spark and Lance, with the following key features:
1. **Minimal changes to the existing Lance BTree index** — only a few hundred lines of core code were added.
2.  **Full backward compatibility**— completely compatible with the current implementation.
3.  **Dramatic performance improvement** — for datasets with hundreds of millions of records, the merge phase is <font color="red">**reduced from tens of minutes to under a few seconds**</font>, and end-to-end index construction completes in <font color="red">**just a few minutes**</font>.

The range-based BTree index has a structure nearly identical to the current implementation, with the only addition being multiple distinct page files—each covering a non-overlapping data range:

<img width="750" height="500" alt="Image" src="https://github.com/user-attachments/assets/e1bcbf88-efd3-4dce-b8e4-b382e88a5835" />

## Workflow

<img width="1496" height="478" alt="Image" src="https://github.com/user-attachments/assets/a5897342-7c87-455b-b8b9-0cc0b711e0d2" />
The overall workflow is as follows:

[...]

PR #5202, opened 2025-11-10
feat: distributed range-based BTree index

steFaiz:
This PR is about to introduce range-based BTree index.
Some test results:
| num of rows | num of ranges | execution time | merge time|
|:---------:|:-------:|:-------:|:----------:|
|130 million|3|23 min|1 s|
|130 million|50|3 min|3 s|
|10 billion|1000|15 min| 46 s|

Please refer https://github.com/lancedb/lance/issues/5164 for more details.


steFaiz:
This PR is ready for review, PTAL if you have some time! @westonpace @wjones127 @xloya

rust/lance/src/index/create.rs
```diff
@@ -38,6 +39,7 @@ pub struct CreateIndexBuilder<'a> {
     train: bool,
     fragments: Option<Vec<u32>>,
     index_uuid: Option<String>,
+    preprocessed_data: Option<Box<dyn RecordBatchReader + Send + 'static>>,
```
westonpace, line 42: This is very vague as an API.  It might be an ok intermediate step but I'd like to transition to something better.  I can't remember where else we had this discussion but I think eventually we should probably not have a single `CreateIndexBuilder` that works for all scenarios.  We can leave this as a "simple nothing fancy" builder that gets invoked via SQL (e.g. on a `CREATE INDEX` command).

However, for things like distributed build or where we are using external compute I think we should have the entire build be outside of lance and just use the commit.  Or we should have a dedicated API for distributed btree indexing (similar to the API for compaction or the fragments API)

rust/lance-index/src/scalar/btree.rs
```diff
@@ -66,6 +69,9 @@ const BTREE_LOOKUP_NAME: &str = "page_lookup.lance";
 const BTREE_PAGES_NAME: &str = "page_data.lance";
 pub const DEFAULT_BTREE_BATCH_SIZE: u64 = 4096;
 const BATCH_SIZE_META_KEY: &str = "batch_size";
+pub const DEFAULT_RANGE_PARTITIONED: bool = false;
```
westonpace: Does this need to be pub?

rust/lance-index/src/scalar/btree.rs
```diff
@@ -690,26 +696,124 @@ impl BTreeLookup {
...
     }
 }
 
+/// Index reader to dispatch page query to corresponding ranged page-files.
+struct LazyRangedIndexReader {
+    #[allow(clippy::type_complexity)]
+    readers: Arc<DashMap<String, Arc<tokio::sync::OnceCell<Arc<dyn IndexReader>>>>>,
+    store: Arc<dyn IndexStore>,
+    // for each range, we store the corresponding file name and start offset
+    ranges_to_files: Arc<RangeInclusiveMap<u32, (String, u32)>>,
```
westonpace, line 839: Can we document what the key (u32) represents here?  Or maybe change the name?  If I understand the code below the key is the page number.  So maybe this is `page_number_to_files`?

rust/lance-index/src/scalar/btree.rs
```diff
@@ -1212,10 +1355,31 @@ impl ScalarIndex for BTreeIndex {
...
-        self.store
-            .copy_index_file(BTREE_LOOKUP_NAME, dest_store)
-            .await?;
+        // Copy the lookup file as-is if not range-partitioned
+        if self.ranges_to_files.is_none() {
+            self.store
+                .copy_index_file(BTREE_LOOKUP_NAME, dest_store)
+                .await?;
+        } else {
+            // else we should remove range-partition configuration options
```
westonpace: Is this falling back to the old non-range impl when there is a remap?

westonpace (commented):
Initial review of the rust.  I'm still a little confused what the "range" is here?  This seems to be a u32 that replaces the page id but I'm not entirely sure what it represents?

I think some more documentation would be welcome but that can be done in follow-ups (to be fair, old btree didn't have enough documentation either).

I'm going to try soon and play with adding some python bindings.

Initially I was a little uncertain about the two indexes sharing the same struct / file but there is a lot of code shared between the two paths so maybe it is ok.

westonpace:
Also, forgot to mention, this is a very cool idea!

rust/lance-index/src/scalar/btree.rs
wayneli-vt: You're right, it doesn't need to be public. Thank you for pointing that out!

rust/lance-index/src/scalar/btree.rs
wayneli-vt: Thanks for your feedback!
You're right to question the key's meaning. It's actually `range_id` rather than the page number. 
In short, a `range_id` is the ordinal ID for a globally-sorted data partition used in a distributed build. The key idea is that if you concatenate the data from all partitions in range_id order, the result is a globally sorted dataset.
To make this all clearer in the code, I've made a few documentation updates:

1. Added a detailed doc comment for [range_id](https://github.com/steFaiz/lance/blob/2e38bdced59eca33de2ebfe895ec48c7be5f3d5e/rust/lance-index/src/scalar/btree.rs#L866) to explain what it is and its requirements.

2. Documented the purpose of [`ranges_to_files`](https://github.com/steFaiz/lance/blob/2e38bdced59eca33de2ebfe895ec48c7be5f3d5e/rust/lance-index/src/scalar/btree.rs#L866) and how it's [constructed](https://github.com/steFaiz/lance/blob/2e38bdced59eca33de2ebfe895ec48c7be5f3d5e/rust/lance-index/src/scalar/btree.rs#L1085).

Hope this helps clarify things!

rust/lance/src/index/create.rs
wayneli-vt: That's a great point, and I completely agree. A dedicated API for distributed builds is the right long-term direction. And I've opened https://github.com/lance-format/lance/issues/5359 to start designing this new, unified API.

For the scope of this current PR, I'm happy to treat this implementation as the "intermediate step". I will then refactor it to use the new API once it's finalized in the follow-up issue.

rust/lance-index/src/scalar/btree.rs
wayneli-vt: Yes, that's right. Previously, calling `remap` would fall back to the non-range implementation and create a single, unified `page_data.lance`.

Now it remaps each `part_i_page_data.lance` individually. This generates a new series of page files and preserves the range-based structure.

wayneli-vt:
@westonpace Hi, I've addressed all the comments. Could you please take another look? Thanks!

rust/lance-index/Cargo.toml
```diff
@@ -69,6 +69,8 @@ async-channel = "2.3.1"
 bitpacking = { version = "0.9.2", features = ["bitpacker4x"] }
 rand_distr.workspace = true
 lance-datagen.workspace = true
+rangemap.workspace = true
+dashmap = "6"
```
westonpace: Is there any particular reason we need dashmap here?  Just wondering if we can get away without adding a new dependency.

rust/lance-index/src/scalar/btree.rs
```diff
@@ -759,6 +862,36 @@ pub struct BTreeIndex {
...
+    ///
+    /// # Example
+    ///
+    /// If the map contains an entry `(100..=199) => ("part_2_page_file.lance", 100)`, and we
+    /// need to find `global_page_idx = 142`:
+    ///
+    /// - The map finds that 142 falls within the range `100..=199`, and it returns
+    ///   `("part_2_page_file.lance", 100)`.
+    /// - The local page_idx is calculated: `142 - 100 = 42`.
+    /// - The system now knows to read page `42` from the file `part_2_page_file.lance`.
```
westonpace, line 999: This is very helpful for understanding, thank you!

rust/lance-index/src/scalar/btree.rs
```diff
@@ -690,26 +696,123 @@ impl BTreeLookup {
...
+            ranges_to_files,
+        }
+    }
+
+    async fn get_reader(&self, file_name: &str) -> Result<Arc<dyn IndexReader>> {
+        let reader_cell = {
+            let guard = self
+                .readers
+                .entry(file_name.to_string())
+                .or_insert_with(|| Arc::new(tokio::sync::OnceCell::new()));
```
westonpace: Ok, so I guess this is why we need the concurrent hash map?  We could probably just put a regular hash map in a mutex?  The critical section would be pretty brief (just cloning the `Arc` holding the `OnceCell`) and this wouldn't be an area of super-high contention.

rust/lance-index/src/scalar/btree.rs
```diff
@@ -690,26 +696,123 @@ impl BTreeLookup {
...
+            .await?;
+        Ok(reader.clone())
+    }
+
+    async fn get_reader_and_local_page_idx(
+        &self,
+        page_idx: u32,
+    ) -> Result<(Arc<dyn IndexReader>, u32)> {
+        let (page_file_name, offset) = self.ranges_to_files.get(&page_idx).unwrap_or_else(|| {
+            panic!("Unexpected page index, index {} is out of range.", page_idx)
```
westonpace: Can we make this an error instead of a panic?

rust/lance-index/src/scalar/btree.rs
```diff
@@ -937,18 +1076,52 @@ impl BTreeIndex {
...
+
+        let range_partitioned = file_schema
+            .metadata
+            .get(RANGE_PARTITIONED_META_KEY)
+            .map(|bs| bs.parse().unwrap_or(DEFAULT_RANGE_PARTITIONED))
+            .unwrap_or(DEFAULT_RANGE_PARTITIONED);
+        // For range-partitioned indices, construct the `ranges_to_files` map.
+        // This converts the list of (partition ID, page count) from metadata into a map
+        // from a global page range to its corresponding file and starting offset.
+        let ranges_to_files = if range_partitioned {
```
westonpace, line 1199: This works but would it be simpler to just add a column to the lookup file containing the starting offset?  Will there be the same number of entries in the map that there are rows in the lookup?

rust/lance-index/src/scalar/btree.rs
```diff
@@ -2872,4 +3280,502 @@ mod tests {
...
+        let remapped_data = remapped_pages
+            .read_record_batch(0, remapped_pages.num_rows() as u64)
+            .await
+            .unwrap();
+
+        assert_eq!(original_data, remapped_data);
+    }
+
+    #[tokio::test]
+    async fn test_update_ranged_index() {
```
westonpace, line 3836: Thanks for all the unit tests

westonpace (changes requested):
A few minor requests and then I think this is ready to merge.  Thanks for continuing to work on this!

rust/lance-index/Cargo.toml
steFaiz: Sorry for this! I agree with you that the range-file-reader scenario won't incur severe concurrency. I've remove this dashmap dependency and use Mutex + Hashmap to deal with it.

rust/lance-index/src/scalar/btree.rs
steFaiz: Of course! I've refactored it.

rust/lance-index/src/scalar/btree.rs
steFaiz: > Will there be the same number of entries in the map that there are rows in the lookup?

The answer is NO. The number of entries in the range map is equal to the number of page files, which would be much smaller than the row number in lookup files. Actually, we only store start index for each page file, and when loading into memory, we use a `RangeMap` to store the start index of each page index range. 
If we add a column in lookup file, the memory occupied by lookup file will increase by 1/4, and we will store so many duplicated `start_index`

steFaiz:
resolving conflicts

PR #5203, opened 2025-11-10
ci: use in-memory storage for unit blob tests

Xuanwo:
This PR will address flaky tests in CI.

---

### DIFF
--- a/rust/lance/src/dataset/blob.rs
+++ b/rust/lance/src/dataset/blob.rs
@@ -220,8 +220,6 @@ pub(super) async fn take_blobs(
 mod tests {
     use std::sync::Arc;
 
-    use std::sync::atomic::{AtomicU64, Ordering};
-
     use arrow::{array::AsArray, datatypes::UInt64Type};
     use arrow_array::RecordBatch;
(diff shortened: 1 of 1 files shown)


rust/lance/src/dataset/blob.rs
```diff
@@ -220,27 +220,34 @@ pub(super) async fn take_blobs(
...
-        _test_dir: TempStrDir,
         dataset: Arc<Dataset>,
         data: Vec<RecordBatch>,
     }
 
+    fn next_memory_uri() -> String {
+        static COUNTER: AtomicU64 = AtomicU64::new(0);
+        let id = COUNTER.fetch_add(1, Ordering::Relaxed);
+        format!("memory://blob-tests/{id}")
+    }
```
wjones127: Memory uris don't need to be distinct unless you are re-using the same `Session` between tests, which I don't think you are.

Xuanwo: Removed `COUNTER` logic.

Xuanwo:
In memory storage seems to have sessions issues while in unit test, need more research.

PR #5217, opened 2025-11-12
refactor!: move all previous code into `previous` mod

Xuanwo:
This PR will move all previous code into `previous` mod to make our code more clean for both human and AI.

This PR will also add clear alias of previous code like `PreviousReader`.

---

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -72,7 +72,8 @@ jobs:
           command: check
 
   linux-build:
-    runs-on: "ubuntu-24.04"
+    # Make our CI happy until https://github.com/lancedb/lance/issues/5218 addressed.
+    runs-on: warp-ubuntu-2404-x64-8x
     timeout-minutes: 60
     strategy:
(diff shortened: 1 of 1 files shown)


java/lance-jni/src/file_reader.rs
```diff
@@ -22,7 +22,7 @@ use lance::io::ObjectStore;
 use lance_core::cache::LanceCache;
 use lance_core::datatypes::Schema;
 use lance_encoding::decoder::{DecoderPlugins, FilterExpression};
-use lance_file::v2::reader::{FileReader, FileReaderOptions, ReaderProjection};
+use lance_file::reader::{FileReader, FileReaderOptions, ReaderProjection};
```
westonpace, line 25: This is a breaking rust-level change.  There are potentially a few (maybe not very many) Rust users who use `FileReader` directly.  I think we can still proceed with the change (we haven't ever claimed the Rust API is stable) but we should maybe mark the PR as breaking.

Issue #5163, opened 2025-11-06
Naming discussion around blob v2


Xuanwo:
Minor naming change.  I think (but may be understanding wrong) that "blob version 1" is "inline only" and "blob version 2" is "inline, packed, and dedicated".

If this is correct then I think we should call it `BLOBFILE_DESC_FIELD` (or `PACKED_DESC_FIELD`).  We are not replacing the existing inline approach, we are adding a new packed approach which will utilize blob-files.  The inline approach will still use the 2-field description.  The packed approach will use the 5-field description.

_Originally posted by @westonpace in https://github.com/lancedb/lance/pull/4948#pullrequestreview-3422052320_

Xuanwo:
cc @westonpace 

My current idea is to have a blob v2 concept that includes all supported blob types like 'inline', 'packed', or 'dedicated'. This means blob v2 will cover all the uses of blob v1, which is just inline. I think this makes compatibility easier without changing any logic on the inline side.

We’ll only reuse the file encoding from blob v1 (inline), but all the table schema will be in blob v2. With this change, we can do the version check early at the table level instead of at the time of writing data.

Btw, I’m not strong on this. If you prefer to just keep `packed/dedicated` blob types as an extension to `inline`, I’m fine with making this change.

westonpace:
I think we need a table level setting which controls whether we use blob files or not.  If this is enabled then users don't have to select which columns are blobs.  In other words, if we are doing table blobs, we shouldn't require the user to specify field metadata.

The field metadata for inline blobs can either stick around or be silently deprecated.

I'm not sure if this all makes sense.

Xuanwo:
> I think we need a table level setting which controls whether we use blob files or not. If this is enabled then users don't have to select which columns are blobs. In other words, if we are doing table blobs, we shouldn't require the user to specify field metadata.
> 
> The field metadata for inline blobs can either stick around or be silently deprecated.
> 
> I'm not sure if this all makes sense.

Yep, that's exactly what I want. The current implementation is a hard switch: once users are on file format 2.2, they must use blob v2. Now I understand what you mean: we can just add a table setting for blob v2 instead of tying it to file format 2.2.

I will implement this change later.

PR #5220, opened 2025-11-12
refactor: move blob version as a table level config

Xuanwo:
Close https://github.com/lancedb/lance/issues/5163

This PR will move blob version as a table level config instead.

---


rust/lance/src/dataset/blob.rs
```diff
@@ -190,9 +201,25 @@ pub(super) async fn take_blobs(
...
+const INLINE_BLOB_KIND: u8 = 0;
+
+fn collect_blob_files_v1(
+    dataset: &Arc<Dataset>,
+    blob_field_id: u32,
+    descriptions: &StructArray,
+    row_addrs: &arrow::array::PrimitiveArray<UInt64Type>,
+) -> Result<Vec<BlobFile>> {
     let positions = descriptions.column(0).as_primitive::<UInt64Type>();
     let sizes = descriptions.column(1).as_primitive::<UInt64Type>();
```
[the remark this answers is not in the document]
Xuanwo: We don't need this.

rust/lance-core/src/datatypes/field.rs
```diff
@@ -488,36 +495,13 @@ impl Field {
...
-            }
-        }
-    }
-
     /// If the field is a blob, return a new field with the same name and id
     /// but with the data type set to a struct of the blob description fields.
     ///
     /// If the field is not a blob, return the field itself.
-    pub fn into_unloaded(mut self) -> Self {
+    pub fn into_unloaded_with_version(mut self, version: BlobVersion) -> Self {
```
westonpace, line 502: What is some columns are inline (just have description/size) and some columns are packed/dedicated (have description/size/id/uri/kind)?  Or is the intention that even inline blobs will have description/size/id/uri/kind for v2 blobs?

Xuanwo: > even inline blobs will have description/size/id/uri/kind for v2 blobs?

Yes, I hope so, although most of them will just be `NULL`. In my next PR at https://github.com/lancedb/lance/pull/5239, I plan to return them as an Arrow extension type called `Blob`, so users don’t need to parse them themselves.

PR #5221, opened 2025-11-12
feat: add describe_indices function

westonpace:
The list_indices function relies on APIs from the index objects themselves.  This means we need to load the indices to populate the information.  In addition, the python function uses the index statistics which can be slow.

Rather than modify the existing method (which may introduce a breaking change) this creates a new method `describe_indices`.  This method only uses information available in the dataset manifest.  This ensures that minimal I/O will be required (loading the manifest if it hasn't been loaded) and the call shouldn't be slow.

### DIFF
--- a/python/python/tests/test_vector_index.py
+++ b/python/python/tests/test_vector_index.py
@@ -1402,7 +1402,7 @@ def test_describe_vector_index(indexed_dataset: LanceDataset):
     assert info.fields == [0]
     assert info.field_names == ["vector"]
     assert len(info.segments) == 1
-    assert info.segments[0].fragment_ids == [0]
+    assert info.segments[0].fragment_ids == {0}
     assert info.segments[0].dataset_version_at_last_update == 1
     assert info.segments[0].index_version == 1
     assert info.segments[0].created_at is not None


python/python/lance/lance/__init__.pyi
```diff
@@ -213,6 +214,7 @@ class _Dataset:
     def index_statistics(self, index_name: str) -> str: ...
     def serialized_manifest(self) -> bytes: ...
     def load_indices(self) -> List[Index]: ...
+    def describe_indexes(self) -> List[IndexDescription]: ...
```
rpgreen: Naming nit: is there a reason we're going with "indexes" vs "indices"?

We probably should choose a plural form and be consistent everywhere. My preference would be "indices"

westonpace: Agree we should pick one and be consistent.  I was in team "indices" for a while but was persuaded a while back that "indices" is math/stats and "indexes" is more db historical.

I don't have strong feelings here

I'll make an issue / discussion for this.

rust/lance/src/index.rs
```diff
@@ -476,6 +536,42 @@ impl DatasetIndexExt for Dataset {
         Ok(())
     }
 
+    async fn describe_indexes<'a, 'b>(
+        &'a self,
+        criteria: Option<IndexCriteria<'b>>,
+    ) -> Result<Vec<Arc<dyn IndexDescription>>> {
```
wjones127: One thing I wanted to do with a new list indexes API is aggregate the index deltas, so we return just one entry per index name. What would you think of doing that here?

westonpace: Yes, I think that is a good idea.  I'll do that.

rust/lance/src/index.rs
```diff
@@ -368,6 +370,64 @@ fn vector_index_details() -> prost_types::Any {
...
+                location: location!(),
+            })?;
+        let details = IndexDetails(index_details.clone());
+        let mut rows_indexed = 0;
+
+        for fragment in dataset.get_fragments() {
+            if fragment_bitmap.contains(fragment.id() as u32) {
+                rows_indexed += fragment.fast_physical_rows()? as u64;
+            }
+        }
```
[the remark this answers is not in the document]
westonpace: This was intentional.  I consider this to be a new feature and want to keep the API as synchronous as possible to avoid accidentally introducing slow I/O.

rust/lance/src/index.rs
wjones127: This would change the `IndexDescription` to something like:

```rust
struct IndexDescriptionImpl {
    name: String,
    fragments: Vec<IndexMetadata>,
    details: IndexDetails,
    rows_indexed: u64,
}
```

python/python/lance/lance/__init__.pyi
westonpace: I've gone back to `describe_indices` as I think we use `indices` in too many places to try and change this.

rust/lance/src/index.rs
westonpace: Updated.  Now we have:

```
struct IndexDescriptionImpl {
    name: String,
    field_ids: Vec<u32>,
    segments: Vec<IndexMetadata>,
    index_type: String,
    details: IndexDetails,
    rows_indexed: u64,
}
```

westonpace: I'm assuming that the `field_ids` and `index_type` will be consistent across all ~~shards~~ segments.

westonpace: I'm also assuming the details will be consistent across all segments.  This may not be true at some point in the future.  However, when we get there, I think we will have "common details" (for the whole index) and segment details (for individual segments) so I think it's ok to still have details at the index-level.

python/python/lance/lance/indices/__init__.pyi
```diff
@@ -47,3 +50,24 @@ def transform_vectors(
     pq_codebook: pa.Array,
     dst_uri: str,
 ): ...
+
+class IndexSegmentDescription:
+    uuid: str
+    dataset_version: int
```
wjones127: Do we want a better name that this? As-is, I think it's unclear which version this is supposed to be.

python/python/lance/lance/indices/__init__.pyi
```diff
@@ -47,3 +50,24 @@ def transform_vectors(
...
+    fragment_ids: list[int]
```
wjones127: nit: should this be a set? I say this because I made the fragment bitmap a set over here:

https://github.com/lancedb/lance/blob/f48bbd9cd0885b9b96b578eb967cc7eaa270b409/python/python/lance/dataset.py#L3698

python/python/lance/lance/indices/__init__.pyi
```diff
@@ -47,3 +50,24 @@ def transform_vectors(
...
+
+class IndexDescription:
+    name: str
+    type_url: str
+    index_type: str
+    num_rows_indexed: int
+    fields: list[int]
+    field_names: list[str]
+    segments: list[IndexSegmentDescription]
+    details: str
```
wjones127: Are details a string? Or `bytes`?

wjones127: Oh it's JSON. I wonder if we should parse it eagerly for them or if that's a bad idea.

python/python/tests/test_vector_index.py
```diff
@@ -1391,6 +1391,23 @@ def test_load_indices(dataset):
     assert len(indices) == 1
 
 
+def test_describe_vector_index(indexed_dataset: LanceDataset):
+    info = indexed_dataset.describe_indices()[0]
+
+    assert info.name == "vector_idx"
+    assert info.type_url == "/lance.table.VectorIndexDetails"
+    # This is currently Unknown because vector indices are not yet handled by plugins
+    assert info.index_type == "Unknown"
```
wjones127, line 1400: This seems unfortunate. Hopefully we can fix this very soon!

rust/lance-index/src/scalar/json.rs
```diff
@@ -832,6 +834,19 @@ impl ScalarIndexPlugin for JsonIndexPlugin {
...
+    fn details_as_json(&self, details: &prost_types::Any) -> Result<String> {
+        let registry = self.registry().unwrap();
+        let json_details = crate::pb::JsonIndexDetails::decode(details.value.as_slice())?;
+        let target_details = json_details.target_details.as_ref().expect_ok()?;
+        let target_plugin = registry.get_plugin_by_details(target_details).unwrap();
+        let target_details_json = target_plugin.details_as_json(target_details)?;
+        Ok(serde_json::json!({
+            "path": json_details.path,
+            "target_details": target_details_json,
+        })
```
wjones127: The nested details means that just one `json.loads()` later won't be enough, which is annoying.

Could avoid this by parsing the `target_details_json` into `serde_json::Value` before putting it in the `target_details` field.

rust/lance/src/index.rs
```diff
@@ -367,6 +369,132 @@ fn vector_index_details() -> prost_types::Any {
...
+            .fragment_bitmap
+            .as_ref()
+            .ok_or_else(|| Error::Index {
+                message: "Fragment bitmap is required for index description.  This index must be retrained to support this method.".to_string(),
+                location: location!(),
+            })?;
+
+            for fragment in dataset.get_fragments() {
+                if fragment_bitmap.contains(fragment.id() as u32) {
+                    rows_indexed += fragment.fast_physical_rows()? as u64;
```
wjones127: Should we subtract the number of deleted rows in that fragment? Otherwise I worry this metric will be confusing for users.

Imagine:

* Write 1000 rows
* Create index
* Delete every other row

`ds.count_rows()` will report 500 rows. `index.rows_indexed()` will report 1000 rows.

python/python/lance/lance/indices/__init__.pyi
westonpace: Probably a good idea since I don't even know the answer.  I _think_ it's the version the index was trained against?  I'll look it up.

python/python/lance/lance/indices/__init__.pyi
westonpace: I don't see any harm in it but I also don't see the reasoning?  Is it because we assume users will want to do set-like operations on this?

python/python/lance/lance/indices/__init__.pyi
westonpace: Parse it into what?  Python dicts?

python/python/tests/test_vector_index.py
westonpace: Yeah, the current implementation (`list_indices`) can actually determine the type for vector indexes so this is a bit of a regression but I think it involves opening the index and I'd like to be able to do it from the details / manifest only.

rust/lance-index/src/scalar/json.rs
westonpace: That's a good suggestion.  Will do.

rust/lance/src/index.rs
westonpace: We do document that this overcounts.  Do we record the number of deleted rows per fragment in the manifest?  I didn't want to have to load the deletion vector.

westonpace: Ah, yep, it is in the deletion file metadata.  I can fix this up.

python/python/lance/lance/indices/__init__.pyi
wjones127: Yeah I think so too.

python/python/lance/lance/indices/__init__.pyi
wjones127: Yeah, that's my thinking. I don't feel strongly about this.

python/python/lance/lance/indices/__init__.pyi
wjones127: Yeah Python dicts. Maybe a bad idea, but could be nice to do for them. Or at least specify in the docstring or something it is JSON.

python/python/lance/lance/indices/__init__.pyi
westonpace: Ok, switched to set

python/python/lance/lance/indices/__init__.pyi
westonpace: I went ahead and converted to python dict

rust/lance-index/src/scalar/json.rs
westonpace: Done

rust/lance/src/index.rs
westonpace: Now includes deletion count in the row count calculation

wjones127 (approved):
Excellent work! Thank you.

PR #5228, opened 2025-11-12
test: simplify IO tests

wjones127:
This PR makes it easier to make assertions about IO

* Make IO statistics on by default.
* Make IO statistics tracked even for local object reader (which previously bypassed statistics)
* Expose IO stats in Python

### DIFF
--- a/python/src/dataset/io_stats.rs
+++ b/python/src/dataset/io_stats.rs
@@ -29,8 +29,8 @@ pub struct IoStats {
 impl IoStats {
     fn __repr__(&self) -> String {
         format!(
-            "IOStats(read_iops={}, read_bytes={}, write_iops={}, write_bytes={}, num_hops={})",
-            self.read_iops, self.read_bytes, self.write_iops, self.written_bytes, self.num_hops
+            "IOStats(read_iops={}, read_bytes={}, write_iops={}, write_bytes={})",
+            self.read_iops, self.read_bytes, self.write_iops, self.written_bytes
         )
     }
 }
@@ -42,7 +42,7 @@ impl IoStats {
             read_iops: stats.read_iops,
             read_bytes: stats.read_bytes,
             write_iops: stats.write_iops,
-            written_bytes: stats.write_bytes,
+            written_bytes: stats.written_bytes,
         }
     }
 }


wjones127:
@westonpace this is meant to address your comment in https://github.com/lancedb/lance/pull/4923#pullrequestreview-3325837100

python/src/dataset/io_stats.rs
```diff
@@ -0,0 +1,54 @@
...
+    pub read_bytes: u64,
+    /// Number of write IO operations performed
+    pub write_iops: u64,
+    /// Total bytes written to storage
+    pub write_bytes: u64,
+    /// Number of disjoint periods where at least one IO is in-flight
+    ///
+    /// This metric helps understand IO parallelism. A lower number indicates
+    /// more parallel IO operations.
+    pub num_hops: u64,
```
wkalt: calling these hops/network hops isn't clear to me. Would this be network hops behind the S3 endpoint? That's how I would interpret it but I don't think we have that info

wkalt: is this just network requests?

rust/lance-io/src/local.rs
```diff
@@ -203,7 +236,14 @@ impl Reader for LocalObjectReader {
         .map_err(|err: std::io::Error| object_store::Error::Generic {
             store: "LocalFileSystem",
             source: err.into(),
-        })
+        });
+
+        // Record the read operation if tracking is enabled
```
wkalt: is the optionality required? Elsewhere in the PR it seems to indicate it'll always be enabled

python/src/dataset/io_stats.rs
wjones127: Maybe I should give an example in the comment.

Imagine a process:

1. Call list to get 10 files
2. In parallel, call head on 10 files
3. Read the largest file

That's a total of 12 requests (list, 10 heads, 1 get). But we do them in 3 "hops". Maybe that's not the best term. Could do "stages" or something else.

wkalt: I see. Makes me think of dependency chains - not sure if that's a helpful term.

python/src/dataset.rs
```diff
@@ -2045,6 +2047,68 @@ impl Dataset {
...
+    /// Examples
+    /// --------
+    /// >>> import lance
+    /// >>> dataset = lance.dataset("my_dataset")
+    /// >>> # Perform some operations...
+    /// >>> stats = dataset.io_stats_incremental()
+    /// >>> print(f"Read {stats.read_bytes} bytes in {stats.read_iops} operations")
+    /// >>> # Next call returns only new stats since last call
+    /// >>> more_data = dataset.to_table()
+    /// >>> stats2 = dataset.io_stats_incremental()
```
westonpace:
```suggestion
    /// Get incremental IO statistics for this dataset
```

Ditto

python/src/dataset.rs
```diff
@@ -2045,6 +2047,68 @@ impl Dataset {
...
+    /// --------
+    /// >>> import lance
+    /// >>> dataset = lance.dataset("my_dataset")
+    /// >>> result = dataset.to_table()
+    /// >>> # Check stats without resetting
+    /// >>> stats = dataset.io_stats_snapshot()
+    /// >>> print(f"Read {stats.read_bytes} bytes in {stats.read_iops} operations")
+    /// >>> # Can check again and see the same values
+    /// >>> stats2 = dataset.io_stats_snapshot()
+    /// >>> assert stats.read_bytes == stats2.read_bytes
```
westonpace:
```suggestion
    /// Get a snapshot of current IO statistics without resetting counters
```

python/src/dataset/io_stats.rs
```diff
@@ -0,0 +1,54 @@
...
+/// IO statistics for dataset operations
+///
+/// This tracks the number of IO operations and bytes transferred for read and write
+/// operations performed on the dataset's object store.
+///
+/// Note: Calling `io_stats()` returns the statistics accumulated since the last call
+/// and resets the internal counters (incremental stats pattern).
+#[pyclass(name = "IOStats", module = "_lib", get_all)]
+#[derive(Clone, Debug)]
+pub struct IoStats {
```
westonpace, line 17: It would be nice if we could include these docs in mkdocs.  I'll make an issue to figure that out.  Then we wouldn't need the lengthy `Returns` block on the python.

python/src/dataset/io_stats.rs
westonpace: I agree that "hops" is not quite the correct term (though I am used to our usage of it in this way from the Rust unit tests).

I also am not aware of any standard term.

Is this something we want to expose?  Should we mention it is likely meaningless if there are concurrent operations in flight?  Or that it can be a somewhat noisy metric?

rust/lance-io/src/local.rs
westonpace: Same question.

rust/lance-io/src/utils/tracking_store.rs
```diff
@@ -191,22 +229,28 @@ impl IoTrackingStore {
...
         stats.write_iops += 1;
         stats.write_bytes += num_bytes;
+        #[cfg(feature = "test-util")]
         stats.requests.push(IoRequestRecord {
             method,
             path,
             range: None,
         });
+        #[cfg(not(feature = "test-util"))]
+        let _ = (method, path); // Suppress unused variable warnings
```
westonpace, line 259: What are we feature gating here?  Tracking of every request's path / method / range vs. just tracking the counts?

westonpace (approved):
Are these counters process-wide or dataset-wide?

A few questions but no real concerns.

rust/lance-io/src/utils/tracking_store.rs
wjones127: Yeah with `test-util` enabled, we will track a list of all requests made in the IO stats. This makes it much easier to debug a failing test.

But for normal usage, keeping track of those will just accumulate a lot of data for now reason.

This particular gate is just because `method` and `path` are only used for the request tracking, so added a line to suppress the unused variable warning.

rust/lance-io/src/local.rs
wjones127: I can remove it. There is a constructor that doesn't pass a tracker down (used for tests I think). But I can just make it create an empty stats instance and record unconditionally. That can simplify some code.

python/src/dataset/io_stats.rs
wjones127: I think I'll rename it for `num_stages`. You're also right it's mostly useful for testing. So I'll gate it under `test-utils`.

Issue #5186, opened 2025-11-07
Strategized Compaction Plan


zhangyue19921010:
# Background
Currently, Lance allows users to manually trigger a standalone job to perform Compaction, to achieve the following:
1. Merging small files.
2. Materializing deletions.

However, in practice, there are some limitations:
1. The current Compaction plan phase only supports simple grouping strategies(as shown below), which are not sufficiently comprehensive. Adding new grouping logic is tightly coupled with existing logic.
    - Grouping based on target_rows_per_fragment.
    - Grouping based on whether a fragment contains deleted rows.
2. The current Compaction is a full compaction, with no limit on the total target I/O. This leads to high pressure on the Compaction Job and makes it difficult to guarantee efficiency.

# Design
## New CompactionPlanner Trait
```
[async_trait::async_trait]
pub trait CompactionPlanner: Send + Sync {
    async fn plan(
        &self,
        dataset: &Dataset,
        options: &CompactionOptions,
    ) -> Result<CompactionPlan>;
}
```

A new CompactionPlanner trait is introduced with a core plan method. This allows users to define custom compaction plan strategies, offering greater flexibility and decoupling the logic of different strategies.

## Default Implementation
```
#[derive(Debug, Clone, Default)]
pub struct DefaultCompactionPlanner;

#[async_trait::async_trait]
impl CompactionPlanner for DefaultCompactionPlanner {
    async fn plan(
        &self,
        dataset: &Dataset,
        options: &CompactionOptions,
    ) -> Result<CompactionPlan> {
        // ..... Same as the current implementation
    }
}
```
Provide a default implementation to ensure that the current behavior remains identical to the historical behavior.

# Plan
1. Complete the foundational work for the Strategized Compaction Plan and provide a default implementation to ensure that the current behavior remains identical to the historical behavior.
2. Introduce new compaction plan strategies from different perspectives, for example:
    - **IO-bounded Compaction Plan:** Limits the total I/O of each compaction to control the load of a single merge job.
    - **Deletion-Only Compaction Plan**: Only merges fragments with deleted rows.
    - **Incremental Compaction Plan**: Only performs compaction on newly changed fragments.
    - etc.
# Core API Compatibility
**Two core APIs have been retained, and default implementations are provided within them to ensure API compatibility.**
## compact_files
```
pub async fn compact_files(
    dataset: &mut Dataset,
    options: CompactionOptions,
[...]

PR #5233, opened 2025-11-13
feat: strategized plan compaction

zhangyue19921010:
Close https://github.com/lancedb/lance/issues/5186

### DIFF
--- a/rust/lance/src/dataset/optimize.rs
+++ b/rust/lance/src/dataset/optimize.rs
@@ -3668,6 +3668,6 @@ mod tests {
         assert!(!plan.tasks.is_empty());
         assert_eq!(plan.read_version, dataset.manifest.version);
         // make sure options.validate() worked
-        assert_eq!(plan.options.materialize_deletions, false);
+        assert!(!plan.options.materialize_deletions);
     }
 }


rust/lance/src/dataset/optimize.rs
```diff
@@ -205,9 +205,157 @@ impl AddAssign for CompactionMetrics {
...
+
+    // no filter by default
+    async fn filter_fragments(
+        &self,
+        _dataset: &Dataset,
+        fragments: Vec<FileFragment>,
+        _options: &CompactionOptions,
+    ) -> Result<Vec<FileFragment>> {
+        Ok(fragments)
+    }
```
westonpace: Do these really need to be trait methods?  I think we can probably leave them out and just let individual implementations use them if they want to.  It will keep the trait simpler.

rust/lance/src/dataset/optimize.rs
```diff
@@ -205,9 +205,157 @@ impl AddAssign for CompactionMetrics {
...
+
+    async fn plan(&self, dataset: &Dataset, options: &CompactionOptions) -> Result<CompactionPlan>;
```
westonpace: Let's document this method.

westonpace: Will `CompactionOptions` be flexible enough for all possible strategies?  Should we maybe accept options as a JSON string or a `Map<String, String>`?  This way different strategies can expose their own custom options.  That would leave the API a little less defined but it would be more flexible.

rust/lance/src/dataset/optimize.rs
```diff
@@ -205,9 +205,157 @@ impl AddAssign for CompactionMetrics {
...
+        compaction_plan.extend_tasks(final_bins);
+
+        Ok(compaction_plan)
+    }
+}
+
 /// Compacts the files in the dataset without reordering them.
 ///
-/// This does a few things:
+/// By default, his does a few things:
```
westonpace:
```suggestion
/// By default, this does a few things:
```

rust/lance/src/dataset/optimize.rs
```diff
@@ -216,14 +364,24 @@ impl AddAssign for CompactionMetrics {
 ///
 /// If no compaction is needed, this method will not make a new version of the table.
 pub async fn compact_files(
+    dataset: &mut Dataset,
+    options: CompactionOptions,
+    remap_options: Option<Arc<dyn IndexRemapperOptions>>, // These will be deprecated later
```
westonpace, line 374: :+1:

westonpace (commented):
A few (mostly minor) suggestions

rust/lance/src/dataset/optimize.rs
wjones127: Do we even need to take `CompactionOptions`? Maybe it should be a argument to the constructor of the individual structs. That way each could have their own arguments but also be strongly typed.

rust/lance/src/dataset/optimize.rs
zhangyue19921010: removed.

rust/lance/src/dataset/optimize.rs
zhangyue19921010: added docs

rust/lance/src/dataset/optimize.rs
zhangyue19921010: changed.

rust/lance/src/dataset/optimize.rs
zhangyue19921010: > Will CompactionOptions be flexible enough for all possible strategies? Should we maybe accept options as a JSON string or a Map<String, String>? This way different strategies can expose their own custom options. That would leave the API a little less defined but it would be more flexible.

Hi @westonpace Thanks a lot for your review. Added `Map<String, String>` for flexible.

> Do we even need to take CompactionOptions? Maybe it should be a argument to the constructor of the individual structs. That way each could have their own arguments but also be strongly typed.

Hi @wjones127 Thanks a lot for your review.

I have tried several ways to eliminate the CompactionOptions parameter in the plan method, but none of them are perfect :( The main contradiction is that during users start planning compaction based on the built planner, they may dynamically adjust the options parameters on certain conditions, such as

https://github.com/lancedb/lance/blob/254a8217ac26666585983aa7ec8c4234f4c3f99f/rust/lance/src/dataset/optimize.rs#L225

If options are passed in when building the planner, then after modifying the options subsequently, it must also be ensured that the options in the planner can be seen. Therefore, we need Arc + mutex and cannot use clone.

On the contrary, it might be simpler and more flexible to pass in the desired options each time the plan method is called here.

rust/lance/src/dataset/optimize.rs
```diff
@@ -205,9 +205,167 @@ impl AddAssign for CompactionMetrics {
...
+/// This trait allows users to define their own compaction strategies by implementing
+/// the `plan` method. The default implementation is provided by [`DefaultCompactionPlanner`].
+#[async_trait::async_trait]
+pub trait CompactionPlanner: Send + Sync {
+    // get all fragments by default
+    fn get_fragments(&self, dataset: &Dataset, _options: &CompactionOptions) -> Vec<FileFragment> {
+        // get_fragments should be returning fragments in sorted order (by id)
+        // and fragment ids should be unique
+        dataset.get_fragments()
+    }
```
wjones127: Was already commented on, but do we need this? It seems like individual implementations can just call `dataset.get_fragments()` and then do whatever filtering they would like.

rust/lance/src/dataset/optimize.rs
wjones127: > If options are passed in when building the planner, then after modifying the options subsequently, it must also be ensured that the options in the planner can be seen. Therefore, we need Arc + mutex and cannot use clone.

I don't understand this. The logic of `validate()` can live in the planner and be internal.

If I were to rewrite `compact_files`, I would do:

```rust
pub async fn compact_files(
    dataset: &mut Dataset,
    mut options: CompactionOptions,
    remap_options: Option<Arc<dyn IndexRemapperOptions>>, // These will be deprecated later
) -> Result<CompactionMetrics> {
    info!(target: TRACE_DATASET_EVENTS, event=DATASET_COMPACTING_EVENT, uri = &dataset.uri);
    // .validate() now happens inside of `from_options`
    let planner = DefaultCompactionPlanner::from_options(options);

    compact_files_with_planner(dataset, &planner, remap_options).await
}
```

zhangyue19921010:
Hi @westonpace and @wjones127 Thanks a lot for your review. All comments are addressed. PTAL :)

wjones127 (commented):
Sorry for the delay, I accidentally left my comments pending. I'm still not sure about the design. It seems like it could be simplified further.

rust/lance/src/dataset/optimize.rs
zhangyue19921010: Removed.

rust/lance/src/dataset/optimize.rs
zhangyue19921010: changed as you suggested ~

zhangyue19921010:
> Sorry for the delay, I accidentally left my comments pending. I'm still not sure about the design. It seems like it could be simplified further.

No worried @westonpace . Really appreciate for your response. Also all comments are addressed. PTAL~

rust/lance/src/dataset/optimize.rs
```diff
@@ -206,9 +206,177 @@ impl AddAssign for CompactionMetrics {
...
+    /// * `dataset` - Reference to the dataset to be compacted
+    /// * `options` - Compaction options including target row count, deletion thresholds, etc.
+    /// * `configs` - Additional configuration parameters as key-value pairs
+    async fn plan(
+        &self,
+        dataset: &Dataset,
+        configs: HashMap<String, String>,
+    ) -> Result<CompactionPlan>;
+
+    fn get_compaction_options(&self) -> &CompactionOptions;
```
wjones127: Why is this necessary? The `CompactionPlan` already contains the `CompactionOptions`, so isn't the return value of `plan` sufficient?

rust/lance/src/dataset/optimize.rs
```diff
@@ -206,9 +206,177 @@ impl AddAssign for CompactionMetrics {
...
+    ///
+    /// # Arguments
+    ///
+    /// * `dataset` - Reference to the dataset to be compacted
+    /// * `options` - Compaction options including target row count, deletion thresholds, etc.
+    /// * `configs` - Additional configuration parameters as key-value pairs
+    async fn plan(
+        &self,
+        dataset: &Dataset,
+        configs: HashMap<String, String>,
```
wjones127: Why do we need `configs`? The point of my earlier suggestion was to get rid of string typed configurations in favor of proper constructors. For example:

```rust
pub trait CompactionPlanner: Send + Sync {
    async fn plan(&self, dataset: &Dataset) -> Result<CompactionPlan>;
}

pub struct DefaultCompactionPlanner {
    options: CompactionOptions,
}

impl CompactionPlanner for DefaultCompactionPlanner {
    async fn plan(&self, dataset: &Dataset) -> Result<CompactionPlan> {
        let tasks = todo!();
        Ok(CompactionPlan {
            tasks,
            read_version: dataset.manifest.version,
            options: self.options.clone()
        })
    }
}

pub struct CustomPlanner {
    config_a: i32,
    config_b: Duration,
    options: CompactionOptions
}

impl CustomPlanner {
    fn new(config_a: i32, config_b: Duration, options: CompactionOptions) -> {
        todo!()
    }
}

impl CompactionPlanner for CustomPlanner {
    async fn plan(&self, dataset: &Dataset) -> Result<CompactionPlan> {
        let tasks = todo!("Use config_a and config_b to plan tasks");
        Ok(CompactionPlan {
            tasks,
            read_version: dataset.manifest.version,
            options: self.options.clone()
        })
    }
}
```

Notice how `CustomPlanner:new()` takes strongly-typed parameters. We no long have to validate string inputs, which makes the API safer and easier to use.

wjones127: BTW we should probably distinguish from configurations for *planning* and configurations for *executing*. The reason we need `CompactionOptions` later is they are needed for *execution*. But the options that will vary for planners are the *planning* options.

wjones127 (commented):
Just got back from vacation, so sorry for the delay.

This is headed in the right direction. I still have some comments on the API.

zhangyue19921010:
> Just got back from vacation, so sorry for the delay.
> 
> This is headed in the right direction. I still have some comments on the API.

Hi @wjones127. Thanks for your response and wish u have a fantastic vacation! I will address the comments asap :)

rust/lance/src/dataset/optimize.rs
zhangyue19921010: > BTW we should probably distinguish from configurations for _planning_ and configurations for _executing_. The reason we need `CompactionOptions` later is they are needed for _execution_. But the options that will vary for planners are the _planning_ options.

I quickly glanced through it, and there are quite a few parts that need to be modified. I will try to create a new PR to solve this problem ：）

zhangyue19921010:
Hi @wjones127 Really appreciate for your help. Comments are all addressed . PTAL ~

wjones127 (approved):
This looks great. Thanks for working with me on this.

zhangyue19921010:
> This looks great. Thanks for working with me on this.

Thanks @wjones127 :) Will optimize the compaction plan strategy, based on the horizontal and tiered vertical strategies we discuss before as next step ~

PR #5234, opened 2025-11-13
feat: add target_bases extension to python write_fragments API

jaystarshot:
The lance.write_dataset() function already supports writing to multiple storage buckets via the target_bases parameter in https://github.com/lancedb/lance/pull/4765

However, write_fragments()  did not expose this capability, even though the underlying Rust implementation 
 _write_fragments, _write_fragments_transaction  already supported it

### DIFF
--- a/python/src/transaction.rs
+++ b/python/src/transaction.rs
@@ -362,7 +362,6 @@ impl<'py> IntoPyObject<'py> for PyLance<&Operation> {
                     .getattr("Overwrite")
                     .expect("Failed to get Overwrite class");
 
-                // Convert initial_bases to Python if present.
                 let initial_bases_py = if let Some(bases) = initial_bases {
                     use crate::dataset::DatasetBasePath;
                     // Convert each Rust BasePath to a Python DatasetBasePath object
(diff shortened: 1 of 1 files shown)


python/python/lance/fragment.py
```diff
@@ -954,6 +957,19 @@ def write_fragments(
...
         secondary indices need to be updated to point to new row ids.
+    target_bases : list of str, optional
+        References to base paths where data should be written. Can be
+        specified in all modes.
+
+        Each string is resolved by trying to match:
+        1. Base name (e.g., "primary", "archive") from registered bases
+        2. Base path URI (e.g., "s3://bucket1/data")
+
+        **CREATE mode**: References must match bases in `initial_bases`
```
jackye1995, line 972: can we also add tests for create and overwrite modes?

python/src/transaction.rs
```diff
@@ -361,7 +362,26 @@ impl<'py> IntoPyObject<'py> for PyLance<&Operation> {
...
+                } else {
+                    None
+                };
+
+                // Call Python Overwrite constructor with or without initial_bases
+                // to maintain backward compatibility with existing code
+                if let Some(bases_list) = initial_bases_py {
+                    cls.call1((schema_py, fragments_py, bases_list))
+                } else {
+                    cls.call1((schema_py, fragments_py))
```
jackye1995: I think once you added `initial_bases` to the Overwrite model, since it's a data class, you won't really be able to invoke this constructor?

jaystarshot: `initial_bases` in the overwrite has a default value of None:

jackye1995: but in that case bases_list will also be None right? so `cls.call1((schema_py, fragments_py, bases_list))` should still work?

jaystarshot: Ahh yes thanks, i explicitly set to only none now

jaystarshot: i think previously it was a rust none and just need to convert to python none

PR #5241, opened 2025-11-14
fix: panic if only one partition and split is triggered

BubbleCal:
- handle the case of single partition is split
- this also fixes that single partition won't count the new data while checking split threshold
- rewrite the tests to cover more cases

### DIFF
--- a/rust/lance/src/index/vector/ivf/v2.rs
+++ b/rust/lance/src/index/vector/ivf/v2.rs
@@ -672,7 +672,6 @@ mod tests {
 
     const NUM_ROWS: usize = 512;
     const DIM: usize = 32;
-    const PARTITION_SPLIT_APPEND_ROWS: usize = 50_000;
 
     async fn generate_test_dataset<T: ArrowPrimitiveType>(
         test_uri: &str,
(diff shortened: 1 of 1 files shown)


rust/lance-index/src/vector/v3/shuffler.rs
```diff
@@ -287,46 +281,6 @@ impl ShuffleReader for IvfShufflerReader {
     }
 }
 
-pub struct SinglePartitionReader {
```
BubbleCal, line 290: We can't know how many rows in the reader, but SPFresh needs `num_rows` to determine whether to trigger a split job, so remove this

rust/lance/src/index/vector/builder.rs
```diff
@@ -1760,17 +1770,15 @@ impl<S: IvfSubIndex + 'static, Q: Quantization + 'static> IvfIndexBuilder<S, Q>
...
+        let min_dist_idx = dists.values().iter().position_min_by(|a, b| a.total_cmp(b));
+        let min_dist = min_dist_idx
+            .map(|idx| dists.value(idx))
+            .unwrap_or(f32::INFINITY);
         match split_centroids_dists {
             Some((d1, d2)) => {
                 if min_dist <= d1 && min_dist <= d2 {
                     Ok(ReassignPartition::ReassignCandidate(
-                        reassign_candidate_ids.value(min_dist_idx),
+                        reassign_candidate_ids.value(min_dist_idx.unwrap()),
```
BubbleCal, line 1781: here `min_dist` is not `f32::INFINITY` so `min_dist_idx` must be `Some(...)`

rust/lance/src/index/vector/builder.rs
```diff
@@ -1779,7 +1787,7 @@ impl<S: IvfSubIndex + 'static, Q: Quantization + 'static> IvfIndexBuilder<S, Q>
                 }
             }
             None => Ok(ReassignPartition::ReassignCandidate(
-                reassign_candidate_ids.value(min_dist_idx),
+                reassign_candidate_ids.value(min_dist_idx.unwrap()),
```
BubbleCal, line 1790: This is a join job, which means there are at least 2 partitions, so `min_dist_idx` won't be None

rust/lance-index/src/vector/v3/shuffler.rs
BubbleCal: This would slow down indexing if `num_partitions=1`, but that means the dataset is small (otherwise users should indexing with more partitions), so it's fine

PR #5244, opened 2025-11-14
feat: add inline optimization for dir namespace

jackye1995:
Additional remaining work in https://github.com/lancedb/lancedb/pull/2708

### DIFF
--- a/rust/lance-namespace-impls/src/dir/manifest.rs
+++ b/rust/lance-namespace-impls/src/dir/manifest.rs
@@ -372,13 +372,7 @@ impl ManifestNamespace {
         let dataset: &mut Dataset = &mut dataset_guard;
 
         // Step 1: Create indexes if they don't already exist
-        let indices = dataset.load_indices().await.map_err(|e| Error::IO {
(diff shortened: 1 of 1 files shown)


rust/lance-namespace-impls/src/dir/manifest.rs
```diff
@@ -333,18 +351,160 @@ impl ManifestNamespace {
...
+                    true,
+                )
+                .await
+            {
+                log::warn!("Failed to create BTREE index on object_id for __manifest table: {:?}. Query performance may be impacted.", e);
+            } else {
+                log::info!(
+                    "Created BTREE index '{}' on object_id for __manifest table",
+                    OBJECT_ID_INDEX_NAME
+                );
```
wjones127, line 406: Should this be info level? Or maybe debug level. Seems like it would be noisy at info level.

wjones127 (approved):

rust/lance-namespace-impls/src/dir/manifest.rs
jackye1995: I felt info was better because this is a one time creation

PR #5250, opened 2025-11-17
feat: dataset supports deep_clone

### LINKED ISSUE #5249: Support deep_clone like Deta lake
Now Lance has supported shallow_clone and branch (The epic tracks them: https://github.com/lancedb/lance/issues/4853). 

There is one more thing to do is deep_clone, which could be useful when people want to get a totally indepedent dataset from some snapshot of the current dataset/branch.

Delta lake both support shallow_clone and deep_clone: https://docs.databricks.com/aws/en/delta/clone

majin1102:
Close #5249

### DIFF
--- a/rust/lance-table/src/io/deletion.rs
+++ b/rust/lance-table/src/io/deletion.rs
@@ -54,7 +54,7 @@ pub fn relative_deletion_file_path(fragment_id: u64, deletion_file: &DeletionFil
         ..
     } = deletion_file;
     let suffix = file_type.suffix();
-    format!("{DELETION_DIRS}/{fragment_id}-{read_version}-{id}.{suffix}")
+    format!("{DELETIONS_DIR}/{fragment_id}-{read_version}-{id}.{suffix}")
 }
 
 /// Write a deletion file for a fragment for a given deletion vector.
--- a/rust/lance/src/dataset/tests/dataset_io.rs
+++ b/rust/lance/src/dataset/tests/dataset_io.rs
@@ -541,7 +541,7 @@ async fn test_deep_clone(
     let original_version = branch.version().version;
     branch
         .tags()
-        .create_on_branch("tag", original_version, Some("branch"))
+        .create("tag", ("branch", original_version))
         .await
         .unwrap();
 


majin1102:
This PR is ready for review @jackye1995 

Please take a look if you have time. Thanks

rust/lance/src/dataset.rs
```diff
@@ -2141,6 +2166,86 @@ impl Dataset {
...
+        &mut self,
+        target_path: &str,
+        version: impl Into<refs::Ref>,
+        store_params: Option<ObjectStoreParams>,
+    ) -> Result<Self> {
+        use futures::StreamExt;
+
+        // Resolve source dataset and its manifest using checkout_version
+        let src_ds = self.checkout_version(version).await?;
+        let path_specs = self.collect_paths().await?;
```
jackye1995: this should use src_ds?

rust/lance/src/dataset.rs
```diff
@@ -207,6 +208,30 @@ impl From<&Manifest> for Version {
     }
 }
 
+/// A file path wrapper that can be used to represent a file in a dataset.
+/// This wrapper is used for changing the base_path like deep_clone
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct FilePath {
```
jackye1995: can we just not have this for now? I agree we probably need some refactoring to have this, but it should be in an independent PR that covers all cases.

rust/lance/src/io/commit.rs
```diff
@@ -138,37 +139,74 @@ async fn do_commit_new_dataset(
...
+                    .collect::<Result<Vec<_>>>()?
+            } else {
+                vec![]
+            };
+            (new_manifest, updated_indices)
         } else {
-            vec![]
-        };
-        (new_manifest, updated_indices)
+            // Deep clone: build a manifest that references local files (no external bases)
```
jackye1995, line 176: I just realized one thing, we need to also set base_id for ExternalFile, so those files can be handled for shallow clone case, and also not inherit base_id for deep clone case.

rust/lance/src/dataset.rs
```diff
@@ -2141,6 +2166,86 @@ impl Dataset {
...
+            .is_ok()
+        {
+            return Err(Error::DatasetAlreadyExists {
+                uri: target_path.to_string(),
+                location: location!(),
+            });
+        }
+
+        let io_parallelism = self.object_store.io_parallelism();
+        let copy_futures = path_specs
```
jackye1995: I guess this is the best we can do now, but all cloud storage have batch copy, we should ideally leverage that, but that requires upstream support. Maybe create a github issue to track that and add a TODO here.

majin1102:
> thanks for working on this, added some comments, and can we also add some tests?

Oops!
The dataset tests were relocated and I forgot to merge this test_deep_clone when I rebased. Sorry for that!

rust/lance/src/io/commit.rs
majin1102: I did realize that. But I searched the code and asked @yanghua he told me this has not been used in any write paths yet. And I don't know what the path looks like and how it shoud be copied.  So I wrote a check:

https://github.com/majin1102/lance/blob/3823f466a5daa6d9e4eed1ddaa00bb4cc95bd705/rust/lance/src/dataset.rs#L2273

Maybe you have more context to give me?

majin1102:
Ready for another look @jackye1995

rust/lance/src/io/commit.rs
jackye1995: Another place I can think of is https://github.com/lance-format/lance/blob/main/rust/lance/src/index/frag_reuse.rs#L162

For stable tow id, we are not doing it today but from spec correctness perspective I think we should probably handle it here. Can be a separated PR though I'm fine with that

yanghua: Yeah, I think we can handle `external file` in the further PRs.

Issue #4620, opened 2025-09-02
index_statistics on LABEL_LIST index is very slow


wjones127:
We see 30s load times on S3 express for 25 million rows.

Looking at the code, it appears that we need to load the whole index data just to get the statistics.

https://github.com/lancedb/lance/blob/ca1d9a61ab2e5c6577e16c66a98ef7ef233688e0/rust/lance-index/src/scalar/bitmap.rs#L111-L114

We should instead change this so we just need to read some metadata off of the footer of the file.

We should assert this doesn't perform many IO requests.

Xuanwo:
Let me take a look over this.

PR #5251, opened 2025-11-17
refactor: write bitmap index statistics in file instead

Xuanwo:
Close https://github.com/lancedb/lance/issues/4620

This PR will write bitmap index statistics in file instead so we don't need to load the entire index file to calculate it.

---

### DIFF
--- a/rust/lance/src/index.rs
+++ b/rust/lance/src/index.rs
@@ -208,7 +208,16 @@ fn type_name_from_uri(index_uri: &str) -> String {
 }
 
 /// Legacy mapping from type URL to the old IndexType string for backwards compatibility.
-fn legacy_type_name(index_uri: &str) -> String {
+/// Legacy mapping from type URL to the old IndexType string for backwards compatibility.
+///
+/// If `index_type_hint` is provided (e.g. parsed from the index statistics of a concrete
+/// index instance), it takes precedence so callers can surface the exact index type even
+/// when the type URL alone is too generic (such as VectorIndexDetails).
+fn legacy_type_name(index_uri: &str, index_type_hint: Option<&str>) -> String {
+    if let Some(hint) = index_type_hint {
+        return hint.to_string();
+    }
+
     let base = type_name_from_uri(index_uri);
 
     match base.as_str() {
@@ -976,7 +985,11 @@ impl DatasetIndexExt for Dataset {
         }
 
         let index_uri = index_uri.unwrap_or_else(|| "unknown".to_string());
-        let index_type = legacy_type_name(&index_uri);
+        let index_type_hint = indices_stats
+            .first()
+            .and_then(|stats| stats.get("index_type"))
+            .and_then(|v| v.as_str());
+        let index_type = legacy_type_name(&index_uri, index_type_hint);
 
         let indexed_fragments_per_delta = self.indexed_fragments(index_name).await?;
 


rust/lance-index/src/scalar/bitmap.rs
```diff
@@ -715,6 +726,10 @@ impl ScalarIndexPlugin for BitmapIndexPlugin {
         true
     }
 
+    fn index_type(&self) -> IndexType {
+        IndexType::Bitmap
+    }
```
wjones127: Why is this method necessary?

wjones127: This might interact with what was done here: https://github.com/lancedb/lance/pull/5221

rust/lance-index/src/scalar/bitmap.rs
```diff
@@ -759,19 +774,169 @@ impl ScalarIndexPlugin for BitmapIndexPlugin {
...
+            panic!("metadata store does not support rename")
+        }
+
+        async fn delete_index_file(&self, _name: &str) -> Result<()> {
+            panic!("metadata store does not support delete")
+        }
+    }
+
+    #[tokio::test]
+    async fn test_bitmap_metadata_statistics_minimal_io() {
```
wjones127: The tests I'd really like to see is:

1. Write a 1MB bitmap index
2. Use `list_indices()` or `index_statistics()`
3. Assert `io_stats.bytes_read` < 1KB or something.

See IO stats: https://github.com/lancedb/lance/pull/5228 and https://github.com/lancedb/lance/pull/4923

wjones127 (commented):
I like the approach. I think we should figure out whether or not we want `index_types()` method or if we want something else. Weston is working on something in parallel in https://github.com/lancedb/lance/pull/5221

rust/lance-index/src/scalar/bitmap.rs
Xuanwo: Yep, I'm thinking about this too, but we need a way to tell which index type it is without loading the index.

rust/lance-index/src/scalar/bitmap.rs
Xuanwo: I tried this before, but it looks like our IO stats are tracked globally, and when different tests run at the same time (on different threads), the differences could be wrong. Am I misunderstanding it?

wjones127: > I tried this before, but it looks like our IO stats are tracked globally, and when different tests run at the same time (on different threads), the differences could be wrong. Am I misunderstanding it?

They are not tracked globally. Each ObjectStore instance get it's own io statistics.

Xuanwo: > They are not tracked globally. Each ObjectStore instance get it's own io statistics.

Thank you for this, let me try again.

rust/lance-index/src/scalar/registry.rs
```diff
@@ -129,6 +130,9 @@ pub trait ScalarIndexPlugin: Send + Sync + std::fmt::Debug {
     /// Returns true if the index returns an exact answer (e.g. not AtMost)
     fn provides_exact_answer(&self) -> bool;
 
+    /// Returns the index type for this plugin
+    fn index_type(&self) -> IndexType;
```
wjones127: Say someone creates a new custom index plugin, that doesn't fit any of the enum variant we have built into our library. What do they return here?

rust/lance/src/index.rs
```diff
@@ -1873,6 +1890,92 @@ mod tests {
...
+        let stats = io_tracker.incremental_stats();
+        assert!(
+            stats.read_bytes < 1024,
+            "index_statistics should only read metadata, read {} bytes",
+            stats.read_bytes
+        );
+        assert_eq!(
+            stats.written_bytes, 0,
+            "index_statistics should not perform writes"
+        );
```
wjones127: If you use the `assert_io_eq` macro, then on failure it will print out the IO requests that happened, which is very useful for debugging.
```suggestion
        assert_io_eq!(
            stats,
            written_bytes,
            0,
            "index_statistics should not perform writes"
        );
```

rust/lance/src/index.rs
```diff
@@ -1873,6 +1890,92 @@ mod tests {
...
+        io_tracker.incremental_stats();
+
+        dataset.index_statistics("status_idx").await.unwrap();
+
+        let stats = io_tracker.incremental_stats();
+        assert!(
+            stats.read_bytes < 1024,
+            "index_statistics should only read metadata, read {} bytes",
+            stats.read_bytes
+        );
```
wjones127: Similar here:
```suggestion
        assert_io_lt!(
            stats,
            read_bytes,
            1024,
            "index_statistics should only read metadata, read {} bytes",
            stats.read_bytes
        );
```

rust/lance/src/index.rs
```diff
@@ -1873,6 +1890,92 @@ mod tests {
...
+                ..Default::default()
+            }),
+            ..ReadParams::default()
+        };
+
+        let mut dataset = DatasetBuilder::from_uri(&test_dir)
+            .with_read_params(read_params)
+            .load()
+            .await
+            .unwrap();
```
wjones127: I recently made it so you get an IO tracker on every object store by default, so you can simplify this to:
```suggestion
        let mut dataset = Dataset::write(reader, &test_dir, Some(write_params))
            .await
            .unwrap();
        let io_tracker = dataset.object_store().io_tracker().clone();
```

wjones127 (commented):
Some suggestions to modernize io tracking

rust/lance-index/src/scalar/registry.rs
wjones127: Instead of this, could we have something like:

```rust
impl IndexType {
    fn try_from_pb_name(pb_name: &str) -> Option<Self> { ... }
}
```

Seems like we should be able to derive from the name, right?

wjones127: cc @westonpace in case you have any thoughts on this

rust/lance/src/index.rs
Xuanwo: Perfect!

rust/lance-index/src/scalar/registry.rs
westonpace: Agreed.  We should be trying to move away from `IndexType` enums (because they cannot work with plugins).  We should not add this method.

With the introduction of `describe_indices` every index has two names.  The fully qualified type URL (unique, but not friendly, this is the name of the protobuf details message), for example `/lance.index.pb.JsonIndexDetails` and the short name (friendly, but not unique). The short name is provided by the `ScalarIndexPlugin::name` method.

The index statistics currently has a `index_type`.  This should be updated to have both `index_uri` and `index_typename`.  We can leave `index_type` for legacy / backwards compatibility.  We can add a method near `async fn index_statistics` which maps from `index_uri` to `index_type` for the old indexes...

```
fn legacy_type_name(index_uri: &str) -> String {
  match index_uri {
    BTreeIndexDetails::name => IndexType::BTree.to_string(),
    ...
    _ => "N/A".to_string() // This is a new index type, we don't populate this field any longer
  }
}
```

Xuanwo: Added, PTAL.

rust/lance/src/index.rs
```diff
@@ -196,6 +196,33 @@ fn auto_migrate_corruption() -> bool {
...
 }
 
+/// Derive a friendly (but not necessarily unique) type name from a type URL.
+fn friendly_type_name_from_uri(index_uri: &str) -> String {
+    let type_name = index_uri.rsplit('/').next().unwrap_or(index_uri);
+    type_name
+        .strip_suffix("IndexDetails")
+        .unwrap_or(type_name)
+        .to_string()
+}
```
westonpace, line 208: Now that you have this method you no longer call `index_type`.  Can you get rid of the `index_type` method you added to `ScalarIndexPlugin`?  Then this PR is good to go I think.

Xuanwo: done! PTAL

Xuanwo:
> I think we're good here on backwards / forwards compatibility?
> 
> If new code is reading an old index then `if let Some(stats) = plugin.load_statistics(index_store.clone(), index_details.as_ref()).await?` fails and we fallback to the old approach?
> 
> If old code is reading a new index then the old approach should still work since we didn't get rid of any information.

Yes, we should be good on this, either way works.

> We should probably add "get statistics" to the forwards / backwards compatibility test suites at some point but perhaps can be in a follow-up?

Makes sense to me.

PR #5253, opened 2025-11-17
chore: respect num_indices_to_merge param

BubbleCal:
if `num_indices_to_merge` is passed, don't trigger SPFresh split

### DIFF
--- a/rust/lance/src/index/vector/builder.rs
+++ b/rust/lance/src/index/vector/builder.rs
@@ -715,55 +715,51 @@ impl<S: IvfSubIndex + 'static, Q: Quantization + 'static> IvfIndexBuilder<S, Q> 
             .optimize_options
             .as_ref()
             .and_then(|opt| opt.num_indices_to_merge);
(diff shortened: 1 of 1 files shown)


rust/lance/src/index/vector/builder.rs
```diff
@@ -1163,7 +1167,12 @@ impl<S: IvfSubIndex + 'static, Q: Quantization + 'static> IvfIndexBuilder<S, Q>
         ivf: &IvfModel,
         reader: &dyn ShuffleReader,
         existing_indices: &[Arc<dyn VectorIndex>],
+        num_indices_to_merge: Option<usize>,
```
Xuanwo: This API change really surprised me. If we just input an `Option` and return `Ok(None)` if it's some, how about we do this check outside?

BubbleCal: updated

Xuanwo (approved):

PR #5256, opened 2025-11-17
feat: support credentials vending for file reader and session

jackye1995:
Allow user to directly create a LanceFileReader or LanceFileSession with storage options provider to refresh vended credentials.

We also add a few methods in LanceFileSession for generic object store usage, these are more as workarounds because there is no practical solutions today for an IO library with pluggable credentials provider. The long term plan is to use OpenDAL for related workloads.

### DIFF
--- a/python/src/file.rs
+++ b/python/src/file.rs
@@ -42,6 +42,7 @@ use pyo3::{
 use serde::Serialize;
 use std::collections::HashMap;
 use std::{pin::Pin, sync::Arc};
+use tokio::io::AsyncWriteExt;
 use tokio::sync::Mutex;
 #[pyclass(get_all)]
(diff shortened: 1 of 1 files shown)


rust/lance-io/src/object_store/providers/aws.rs
```diff
@@ -611,6 +611,11 @@ impl DynamicStorageOptionsCredentialProvider {
             }
         }
 
+        log::debug!(
```
jackye1995, line 614: add some additional logging since customers could plugin custom logic here and might be hard to debug

Xuanwo (approved):
approved as a workaround, we'll finally remove them.

jackye1995:
Failing CI fixed in newer PRs, merging

PR #5263, opened 2025-11-18
refactor: rename RowIdSelection to RowAddrSelection

yanghua:
As a step to distinguish the row ID and the row address, we should rename `RowIdSelection` to `RowAddrSelection`. To show its effect more clearly.

### DIFF
--- a/rust/lance-core/src/utils/mask.rs
+++ b/rust/lance-core/src/utils/mask.rs
@@ -433,7 +433,7 @@ impl RowIdTreeMap {
     pub fn len(&self) -> Option<u64> {
         self.inner
             .values()
-            .map(|row_id_selection| match row_id_selection {
+            .map(|row_addr_selection| match row_addr_selection {
                 RowAddrSelection::Full => None,
                 RowAddrSelection::Partial(indices) => Some(indices.len()),
             })
@@ -448,7 +448,7 @@ impl RowIdTreeMap {
         let inner_iters = self
             .inner
             .iter()
-            .filter_map(|(frag_id, row_id_selection)| match row_id_selection {
+            .filter_map(|(frag_id, row_addr_selection)| match row_addr_selection {
                 RowAddrSelection::Full => None,
                 RowAddrSelection::Partial(bitmap) => Some(
                     bitmap


jackye1995:
Try reopen PR to fix CI

rust/lance-core/src/utils/mask.rs
```diff
@@ -374,16 +374,16 @@ pub struct RowIdTreeMap {
     /// The contents of the set. If there is a pair (k, Full) then the entire
     /// fragment k is selected. If there is a pair (k, Partial(v)) then the
     /// fragment k has the selected rows in v.
-    inner: BTreeMap<u32, RowIdSelection>,
+    inner: BTreeMap<u32, RowAddrSelection>,
```
wjones127, line 377: Why is this appropriate? The docs above say that it can be row ids. Do stable row ids not use this structure at all?

https://github.com/lance-format/lance/blob/cb9e613c6ec3deae22d2e9a92c574eabdfa9ba93/rust/lance-core/src/utils/mask.rs#L365-L368

jackye1995, line 377: I think for this one we should use the term row ID? Because this is more used to accommodate stable row ID, and for row address technically directly use RoaringTreemap?

yanghua: It seems no matter `RowIdSelection` nor `RowAddrSelection` can cover the two semantics of row ID and row address. There is a full plan in my mind(sorry, did not write it down before):

It would be better to split the responsibility of the struct for distinguishing row ID(stable) and row address.

* for row address:
  - `RowIdTreeMap` -> `RowAddrMap`（BTreeMap<u32, RowAddrSelection>）
  - `RowIdSelection` → `RowAddrSelection`（Full | Partial(RoaringBitmap)）
  - `SearchResult`(scalar.rs) → `RowAddrSearchResult` (Exact(RowAddrMap) | AtMost(RowAddrMap))
  - `RowIdMask` → `RowAddrMask`

* for (stable) row id:
  - **add** rowid::`RowIdSet` (RoaringTreemap，64-bit set)
  - **add** rowid::`RowIdMask`
  - **add** `RowIdResolver`(row_id -> row_addr)

* for both:
  - add `RowSetOps` for `RowIdSet` and `RowAddrMap`

WDYT? cc @wjones127

wjones127: That's okay. Could you put this design in an issue?

Remember that you're collaborating with other developers and they need to be able to follow what you are doing.

yanghua: Thanks for the suggestion. Have filed a ticket: #5326

yanghua: Will change related comments when doing RowIdTreeMap -> RowAddrMap in #5266 .

rust/lance-core/src/utils/mask.rs
```diff
@@ -434,8 +434,8 @@ impl RowIdTreeMap {
         self.inner
             .values()
             .map(|row_id_selection| match row_id_selection {
```
jackye1995: this one needs to be renamed?

rust/lance-core/src/utils/mask.rs
```diff
@@ -449,8 +449,8 @@ impl RowIdTreeMap {
             .inner
             .iter()
             .filter_map(|(frag_id, row_id_selection)| match row_id_selection {
```
jackye1995: this one needs to be renamed?

rust/lance-core/src/utils/mask.rs
yanghua: yep, have addressed!

PR #5266, opened 2025-11-18
refactor: rename RowIdTreeMap to RowAddrTreeMap

yanghua:


rust/lance-core/src/utils/mask.rs
```diff
@@ -360,7 +360,8 @@ impl std::ops::BitOr for RowIdMask {
     }
 }
 
-/// A collection of row ids.
+/// A collection of row addresses and row ids(for stable row id mode,
+/// and would split it into another structure).
```
yanghua: Added some comments to describe a temporary solution. Will split the responsibility of the row ID.

yanghua:
Renaming some variables may not be thorough. Will recheck them when introducing a new collection for row ID.

rust/lance-core/src/utils/mask.rs
```diff
@@ -417,7 +418,7 @@ impl RowAddrSelection {
     }
 }
 
-impl RowIdTreeMap {
+impl RowAddrTreeMap {
```
jackye1995, line 422: this still contains a `row_ids()` method, should be renamed

rust/lance-core/src/utils/mask.rs
jackye1995: this sounds a bit confusing to me, what about:

```
/// A collection of row addresses.
///
/// Note: For stable row id mode, this may be split into a separate structure in the future.
```

yanghua:
@jackye1995 addressed

PR #5270, opened 2025-11-18
fix!: null handling when using `NOT` with scalar indices

### LINKED ISSUE #4756: Scalar index queries handle NOT and NULL incorrectly

wjones127:
BREAKING CHANGE: The `SearchResult` struct returned by `ScalarIndex::search()` now wraps a `NullableRowIdSet` instead of a `RowIdTreeMap`. Scalar indices must now provide the set of row ids where the expression value is null instead of just where it is true. Additionally, the `RowIdMask` is now an enum instead of a struct.

This PR fixes correctness bugs that show up when (a) running a filter with `NOT`, (b) the column you are filtering on contains nulls, and (c) we are using a scalar index (such as btree, or bitmap). Previously, this would give the wrong answer:

```python
import pyarrow as pa
import lance

data = pa.table({"value": [1, 5, None]})
ds.create_scalar_index("value", "BTREE")
ds.to_table(filter="NOT (value < 2)")
```
```
pyarrow.Table
value: int64
----
value: [[5,null]]
```

It should not include `null`. The reason it did is that our `RowIdMask` (which is output by a scalar index query) was not aware of nulls. So when it processed `value < 2`, it would select just row index `0`. Then `NOT` would invert that to `[1, 2]`, selecting both `[false, null]`.

This PR makes `RowIdMask` aware of nulls. When it processes `value < 2`, it records `selected: [0]` and `nulls: [2]`. Then, when you invert that and then drop, you get `selected: [1]`, giving the correct final answer of just `[5]`.

As part of this, we also refactor RowIdMask to make allow list and deny list mutually exclusive, which simplifies some of the logic.

Fixes #4756

### DIFF
--- a/rust/lance-core/src/utils/mask/nullable.rs
+++ b/rust/lance-core/src/utils/mask/nullable.rs
@@ -472,7 +472,7 @@ mod tests {
         let set1 = nullable_set(&[1, 2, 3], &[]);
         let set2 = nullable_set(&[2, 3, 4], &[]);
 
-        let mut result = set1.clone();
+        let mut result = set1;
         result &= &set2;
 
         // Intersection: [2, 3]
@@ -487,7 +487,7 @@ mod tests {
         let set1 = nullable_set(&[1, 2], &[]);
         let set2 = nullable_set(&[3, 4], &[]);
 
-        let mut result = set1.clone();
+        let mut result = set1;
         result |= &set2;
 
         // Union: [1, 2, 3, 4]
--- a/rust/lance-index/src/scalar/zoned.rs
+++ b/rust/lance-index/src/scalar/zoned.rs
@@ -270,7 +270,7 @@ where
         }
     }
 
-    Ok(crate::scalar::SearchResult::AtMost(row_addr_tree_map))
+    Ok(crate::scalar::SearchResult::at_most(row_addr_tree_map))
 }
 
 /// Helper that retrains zones from `stream` and appends them to the existing
@@ -745,14 +745,14 @@ mod tests {
         };
 
         // Fragment 0, offsets 0 and 1
-        assert!(map.contains(0));
-        assert!(map.contains(1));
+        assert!(map.selected(0));
+        assert!(map.selected(1));
         // Fragment 1 should be skipped entirely
-        assert!(!map.contains((1_u64 << 32) + 5));
-        assert!(!map.contains((1_u64 << 32) + 7));
+        assert!(!map.selected((1_u64 << 32) + 5));
+        assert!(!map.selected((1_u64 << 32) + 7));
         // Fragment 2 includes only the single offset 10
-        assert!(map.contains((2_u64 << 32) + 10));
-        assert!(!map.contains((2_u64 << 32) + 11));
+        assert!(map.selected((2_u64 << 32) + 10));
+        assert!(!map.selected((2_u64 << 32) + 11));
     }
 
     #[test]


docs/src/guide/migration.md
```diff
@@ -6,6 +6,13 @@ stable and breaking changes should generally be communicated (via warnings) for
 give users a chance to migrate.  This page documents the breaking changes between releases and gives advice on how to
 migrate.
 
+## 1.0.0
+
+* The `SearchResult` returned by scalar indices must now output information about null values.
+  Instead of containing a `RowIdTreeMap`, it now contains a `NullableRowIdSet`. Expressions that
+  resolve to null values must be included in search results in the null set. This ensures that
+  `NOT` can be applied to index search results correctly.
```
westonpace, line 14: It's not clear why this is in the migration guide?  Are we considering `SearchResult` and `RowIdTreeMap` to be part of our public API?

Or are you trying to explain that filter results may have changed?  If it's the latter I think this needs to be reworded to be more clear that this means query results will be different.

rust/lance-core/src/utils/mask/nullable.rs
```diff
@@ -0,0 +1,457 @@
...
+
+    /// Check if a row_id is selected (TRUE)
+    pub fn selected(&self, row_id: u64) -> bool {
+        self.selected.contains(row_id) && !self.nulls.contains(row_id)
+    }
+
+    /// Get the selected rows (TRUE or NULL)
+    pub fn selected_rows(&self) -> &RowIdTreeMap {
+        &self.selected
+    }
```
westonpace: It's a little unexpected that `selected` checks for `TRUE` but `selected_rows` is both `TRUE` and `NULL`.  Maybe we could use `not_false_rows` instead of `selected_rows`?

rust/lance-core/src/utils/mask/nullable.rs
```diff
@@ -0,0 +1,457 @@
...
+///
+/// This is often a result of a filter, where `selected` represents the rows that
+/// passed the filter, and `nulls` represents the rows where the filter evaluated
+/// to null. For example, in SQL `NULL > 5` evaluates to null. This is distinct
+/// from being deselected to support proper three-valued logic for NOT.
+/// (`NOT FALSE` is TRUE, `NOT TRUE` is FALSE, but `NOT NULL` is NULL.)
+#[derive(Clone, Debug, Default, DeepSizeOf)]
+pub struct NullableRowIdSet {
+    selected: RowIdTreeMap,
+    nulls: RowIdTreeMap,
```
westonpace: Can a row id be in both sets?  What does that signify?

rust/lance-core/src/utils/mask/nullable.rs
```diff
@@ -0,0 +1,457 @@
...
+    pub fn null_rows(&self) -> &RowIdTreeMap {
+        &self.nulls
+    }
+
+    /// Get the TRUE rows (selected but not null)
+    pub fn true_rows(&self) -> RowIdTreeMap {
+        self.selected.clone() - self.nulls.clone()
+    }
+
+    pub fn union_all(selections: &[Self]) -> Self {
```
westonpace, line 70: What happens if a row id is null in one set and true in another?  I guess that is a repeat of the "can a row id be in both sets" question?

rust/lance-core/src/utils/mask/nullable.rs
```diff
@@ -0,0 +1,457 @@
...
+
+impl std::ops::BitOrAssign<&Self> for NullableRowIdSet {
+    fn bitor_assign(&mut self, rhs: &Self) {
+        self.nulls = if self.nulls.is_empty() && rhs.nulls.is_empty() {
+            RowIdTreeMap::new() // Fast path
+        } else {
+            // null or null -> null (excluding rows that are true in either)
+            let true_rows =
+                (self.selected.clone() - &self.nulls) | (rhs.selected.clone() - &rhs.nulls);
+            (self.nulls.clone() | &rhs.nulls) - true_rows
```
westonpace: And this means that `TRUE | NULL = TRUE`?

rust/lance-core/src/utils/mask/nullable.rs
```diff
@@ -0,0 +1,457 @@
...
+                } else {
+                    (a.nulls.clone() & &b.nulls) // null and null -> null
+                    | (a.nulls & &b.selected) // null and true -> null
+                    | (b.nulls & &a.selected) // true and null -> null
+                };
+                let selected = a.selected & b.selected;
+                Self::AllowList(NullableRowIdSet { selected, nulls })
+            }
+            (Self::AllowList(allow), Self::BlockList(block))
+            | (Self::BlockList(block), Self::AllowList(allow)) => {
```
westonpace, line 191: I'm still a fan of simplifying things so that this case isn't allowed but we can investigate in future issue.

rust/lance-core/src/utils/mask/nullable.rs
```diff
@@ -0,0 +1,457 @@
...
+use deepsize::DeepSizeOf;
+
+use super::{RowIdMask, RowIdTreeMap};
+
+/// A set of row ids, with optional set of nulls.
+///
+/// This is often a result of a filter, where `selected` represents the rows that
+/// passed the filter, and `nulls` represents the rows where the filter evaluated
+/// to null. For example, in SQL `NULL > 5` evaluates to null. This is distinct
+/// from being deselected to support proper three-valued logic for NOT.
```
westonpace, line 13: Can we go ahead and state here that `NULL | TRUE = TRUE` and that `FALSE & NULL = FALSE`?

westonpace (commented):
This is great work!

Here is a first pass of questions before looking at the other files.

docs/src/guide/migration.md
wjones127: > Are we considering SearchResult and RowIdTreeMap to be part of our public API?

Yeah, I'm addressing this to authors of scalar indices outside of the main package.

rust/lance-core/src/utils/mask/nullable.rs
wjones127: Yes. Just signifies that it's null.

I was trying to imitate the Arrow approach, where you can ignore the actual value if it's null. This makes some computations easier, since you can handle null and selected separately.

rust/lance-core/src/utils/mask/nullable.rs
wjones127: Oh this is a good question. I think we should treat `union-all` as basically `reduce(selections, (a, b) => a | b)`, in which case I think it should mean `NULL | TRUE => TRUE`. That means there is a bug here. I will fix this.

rust/lance-core/src/utils/mask/nullable.rs
wjones127: I'm not sure how that would be possible, unless we got rid of BlockList as an option. How do you imagine this getting simpler?

rust/lance-core/src/utils/mask/nullable.rs
wjones127: I'm just going to remove this, as it's not that useful. The places where we use it are just in tests and for those we really want `.true_rows()`.

rust/lance-core/src/utils/mask/nullable.rs
westonpace: Nevermind.  I was thinking of something else entirely.  There is no simplification here.

wjones127:
I've added a bunch of test to give us 100% test coverage for the mask utilities.

PR #5281, opened 2025-11-19
refactor: rename RowIdMask to RowAddrMask

yanghua:
sub-task of #5326

### DIFF
--- a/rust/lance-core/src/utils/mask.rs
+++ b/rust/lance-core/src/utils/mask.rs
@@ -18,7 +18,7 @@ use super::address::RowAddress;
 
 mod nullable;
 
-pub use nullable::{NullableRowAddrSet, NullableRowIdMask};
+pub use nullable::{NullableRowAddrMask, NullableRowAddrSet};
 
 /// A mask that selects or deselects rows based on an allow-list or block-list.
 #[derive(Clone, Debug, DeepSizeOf, PartialEq)]
@@ -885,7 +885,7 @@ mod tests {
     }
 
     #[test]
-    fn test_row_id_mask_construction() {
+    fn test_row_addr_mask_construction() {
         let full_mask = RowAddrMask::all_rows();
         assert_eq!(full_mask.max_len(), None);
         assert_mask_selects(&full_mask, &[0, 1, 4 << 32 | 3], &[]);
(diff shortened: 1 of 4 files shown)


yanghua:
This work is a subtask of this issue. #5326

yanghua:
@jackye1995, please take a look, thanks.

yanghua:
Hi @jackye1995 @wjones127 Please take a look, when you have time.

majin1102 (approved):
Seems a light name changing from row_id to row_addr. Looks good to me

rust/lance/src/session/caches.rs
```diff
@@ -119,15 +119,15 @@ impl CacheKey for DeletionFileKey<'_> {
...
 }
 
-impl CacheKey for RowIdMaskKey {
-    type ValueType = RowIdMask;
+impl CacheKey for RowAddrMaskKey {
+    type ValueType = RowAddrMask;
 
     fn key(&self) -> Cow<'_, str> {
-        Cow::Owned(format!("row_id_mask/{}", self.version))
+        Cow::Owned(format!("row_addr_mask/{}", self.version))
```
yanghua, line 130: This refactor touched the cache key. Does it cause compatibility issues? @jackye1995 @wjones127

rust/lance-core/src/utils/mask/nullable.rs
jackye1995, line 1: looks like NullableRowIdMask was not renamed?

rust/lance-core/src/utils/mask.rs
```diff
@@ -880,144 +880,145 @@ mod tests {
         }
     }
 
-    fn selected_in_range(mask: &RowIdMask, range: std::ops::Range<u64>) -> Vec<u64> {
+    fn selected_in_range(mask: &RowAddrMask, range: std::ops::Range<u64>) -> Vec<u64> {
         range.filter(|val| mask.selected(*val)).collect()
     }
 
     #[test]
     fn test_row_id_mask_construction() {
```
jackye1995: should also update test names like this

rust/lance/src/io/exec/knn.rs
```diff
@@ -759,17 +759,17 @@ impl ANNIvfSubIndexExec {
                             .is_ok()
                         {
                             let initial_ids = state.initial_ids.lock().unwrap();
```
jackye1995: this variable should also be renamed

rust/lance/src/session/caches.rs
jackye1995: session cache is fully in memory, so the time user upgrade the cache will go away anyway, so I think it should be fine

Issue #5246, opened 2025-11-16
[bug] PanicException when calling compaction - AttributeError: module 'lance' has no attribute 'LanceTransaction'


zszheng:
## Description
I encountered a panic error when attempting to compact files in a Lance dataset using `optimize.compact_files()`. The error occurs after writing data to the dataset and then trying to compact it. The issue seems to be related to transaction handling during the compaction process.

## Steps to Reproduce
1. Create a Lance dataset with a schema containing a binary field (with blob encoding metadata)
2. Write data to the dataset using Ray Data's `write_lance()` method
3. Read the dataset using `lance.dataset()`
4. Call `dataset.optimize.compact_files()` with parameters
5. The panic occurs with the error message below

## Code to Reproduce

`import lance
import pyarrow as pa
import ray.data
import pandas as pd
from pathlib import Path
import shutil
import random

schema = pa.schema([
    pa.field("id", pa.int64()),
    pa.field("video_name", pa.string()),
    pa.field("video_size", pa.int64()),
    pa.field("video_data", pa.large_binary(), 
             metadata={"lance-encoding:blob": "true"}),
])

LANCE_PATH = "/tmp/test/lance"

def write(path: str):
    try:
        shutil.rmtree(LANCE_PATH)
    except Exception as ex:
        print(ex)
    
    ray.data.range(100).map(lambda x: {
        "id": x, 
        "video_name": f"video-{x}", 
        "video_size": 100, 
        "video_data": b"sample_data"
    }).write_lance(path, mode="create", schema=schema)

def read(path: str):
    return lance.dataset(path)

def single_compact(ds: lance.LanceDataset):
    ds.optimize.compact_files(
        target_rows_per_fragment=2000,
        materialize_deletions=False,
        batch_size=1024,
        num_threads=20,
    )

write(LANCE_PATH)
ds = read(LANCE_PATH)
single_compact(ds)
print(ds.get_transactions())`

## Error Message
PanicException: Failed to get RewriteGroup class: PyErr { type: <class 'AttributeError'>, value: AttributeError("module 'lance' has no attribute 'LanceTransaction'"), traceback: None }

## Expected Behavior
The compaction operation should complete successfully without panicking.

## Actual Behavior
The compaction operation fails with a panic exception related to a missing `LanceTransaction` attribute.

yanghua:
I am trying to fix this issue.

PR #5282, opened 2025-11-19
fix: panicException when calling compaction

yanghua:
Fixes: #5246


yanghua:
Hi @jackye1995, please take a look, thanks.

jackye1995 (approved):
looks good to me!

PR #5289, opened 2025-11-19
ci: move to lance-format pypi fury index

jackye1995:
also set up new FURY_TOKEN as org secret


westonpace (approved):

jackye1995:
hmm the compat test keeps failing but I was not able to reproduce after a few tries. My suspicion right now is that the index is not used for publication and is private, but I cannot change the visibility until one package is published there. So I will merge this first, and do a release, and then see if I can change the access setting and then this would be fixed.

wjones127:
@jackye1995 check the permissions on the new fury repo.

When I got to https://pypi.fury.io/lancedb/ I can see packages
But when I go to https://pypi.fury.io/lance-format/ I get unauthorized

PR #5298, opened 2025-11-20
refactor: add helper functions to delta.rs tests

yanghua:


jackye1995 (approved):
I think there is always a debate of do we want to create more util methods for testing, but those might become limiting factors in writing tests. But having these at least limited in the specific test mod for delta.rs seems good to me

PR #5301, opened 2025-11-20
fix: update btree index with its own zone size instead of DEFAULT_BTR…

ztorchan:
https://github.com/lance-format/lance/issues/5300


ztorchan:
> This should work in theory. Could you add a unit test?

I will add a unit test to verify it.

ztorchan:
Hi, I have added a unit test in `rust/lance/src/index/scalar.rs` to verify the BTree parameters are the same after optimization. With this change, it runs correctly.

PR #5305, opened 2025-11-20
ci: fix fury publish target to lance-format

jackye1995:


jackye1995:
Will merge to test with fury token

PR #5309, opened 2025-11-21
feat(java): support credential vending at write time

jackye1995:


java/src/main/java/com/lancedb/lance/WriteDatasetBuilder.java
```diff
@@ -0,0 +1,443 @@
...
+   * @param storageOptions Storage configuration options
+   * @return this builder instance
+   */
+  public WriteDatasetBuilder storageOptions(Map<String, String> storageOptions) {
+    this.storageOptions = new HashMap<>(storageOptions);
+    return this;
+  }
+
+  /**
+   * Sets whether to ignore storage options from the namespace's describe_table() or
```
majin1102: Should we use DescribeTable/describeTable and CreateEmptyTable/createEmptyTable?

java/src/main/java/com/lancedb/lance/WriteDatasetBuilder.java
```diff
@@ -0,0 +1,443 @@
...
+  public WriteDatasetBuilder mode(WriteParams.WriteMode mode) {
+    Preconditions.checkNotNull(mode);
+    this.mode = mode;
+    return this;
+  }
+
+  /**
+   * Sets the schema for the dataset.
+   *
+   * <p>If not provided, the schema will be inferred from the reader.
```
majin1102: I guess the comment should be  'If the reader and stream not provided, use the schema to create an empty dataset'? (get from code below)

java/src/main/java/com/lancedb/lance/WriteFragmentBuilder.java
```diff
@@ -0,0 +1,272 @@
...
+      return writeParamsBuilder.build();
+    } else {
+      return new WriteParams.Builder().build();
+    }
+  }
+
+  private void validate() {
+    Preconditions.checkNotNull(datasetUri, "datasetUri is required");
+    Preconditions.checkState(
+        vectorSchemaRoot != null || arrowArrayStream != null,
```
majin1102, line 263: If vectorSchemaRoot != null && arrowArrayStream != null, it seems arrowArrayStream would be ignored. Shall we validate it?

majin1102 (changes requested):
Left a few commens, PTAL

java/lance-jni/src/utils.rs
```diff
@@ -71,8 +75,43 @@ pub fn extract_write_params(
     let storage_options: HashMap<String, String> =
         extract_storage_options(env, storage_options_obj)?;
 
+    // Extract storage options provider if present
+    let storage_options_provider = if !storage_options_provider_obj.is_null() {
```
majin1102: nit: It's fine to extract optional struct like this. I think this util method could make it simpler:
https://github.com/lance-format/lance/blob/2e21599fd6887b1649e0359f92b12362ce04031f/java/lance-jni/src/ffi.rs#L139-L141

PR #5310, opened 2025-11-21
feat(java): support writing schema metadata through java LanceFileWriter API

steFaiz:
This PR supports writing some metadata into datafile schema. See https://github.com/lance-format/lance/issues/5254

### DIFF
--- a/java/lance-jni/src/file_writer.rs
+++ b/java/lance-jni/src/file_writer.rs
@@ -150,6 +150,35 @@ pub extern "system" fn Java_org_lance_file_LanceFileWriter_closeNative<'local>(
 }
 
 #[no_mangle]
+pub extern "system" fn Java_org_lance_file_LanceFileWriter_nativeAddSchemaMetadata<'local>(
+    mut env: JNIEnv<'local>,
+    writer: JObject,
+    schema_metadata: JObject, // Map<String, String>
+) -> JObject<'local> {
+    if let Err(e) = inner_add_schema_metadata(&mut env, writer, schema_metadata) {
+        e.throw(&mut env);
+        return JObject::null();
+    }
+    JObject::null()
+}
+
+fn inner_add_schema_metadata(
+    env: &mut JNIEnv<'_>,
+    writer: JObject,
+    schema_metadata: JObject, // Map<String, String>
+) -> Result<()> {
+    let metadata_map = JMap::from_env(env, &schema_metadata)?;
+    let metadata = to_rust_map(env, &metadata_map)?;
+    let writer_guard =
+        unsafe { env.get_rust_field::<_, _, BlockingFileWriter>(writer, NATIVE_WRITER) }?;
+    let mut writer = writer_guard.inner.lock().unwrap();
+    metadata.into_iter().for_each(|(k, v)| {
+        writer.add_schema_metadata(k, v);
+    });
+    Ok(())
+}
+
+#[no_mangle]
 pub extern "system" fn Java_org_lance_file_LanceFileWriter_writeNative<'local>(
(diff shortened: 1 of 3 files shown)


steFaiz:
@majin1102  @westonpace This PR is ready for review, PTAL if you have some time!

java/src/main/java/com/lancedb/lance/file/LanceFileWriter.java
```diff
@@ -120,6 +120,19 @@ public void write(VectorSchemaRoot batch) throws IOException {
...
   }
 
+  /**
+   * Write a schema metadata map to underlying file. User can retrieve those values from {@link
+   * LanceFileReader#schema() reader schema}.
+   *
+   * @param metadata metadata
+   * @throws IOException IOException
+   */
+  public void writeSchemaMetadata(Map<String, String> metadata) throws IOException {
```
majin1102: For field_ids, I wonder if we can just use:
```
  public void write(VectorSchemaRoot batch) throws IOException {
    try (ArrowArray ffiArrowArray = ArrowArray.allocateNew(allocator);
        ArrowSchema ffiArrowSchema = ArrowSchema.allocateNew(allocator)) {
      Data.exportVectorSchemaRoot(
          allocator, batch, dictionaryProvider, ffiArrowArray, ffiArrowSchema);
      writeNative(ffiArrowArray.memoryAddress(), ffiArrowSchema.memoryAddress());
    }
  }
```

to pass the metadata map?  If we already know the metadata, I think we could eliminate this unnecessary double write. What do you think?

But yeah, this interface might be useful for adding comments and some other metadata after data was written. Could you elaberate more on the usage of this in your mind?

steFaiz: @majin1102 Thanks for you advise! I think the key point is that computing engines and lakehouse formats will call `write_data` multiple times. They might only want to write some metadata e.g. fieldId on openning or closing file writers. I think keeping this method separate is alo feasible. Otherwise user might have to call `write(null, metadata)` if they want to set some metadata only, which is a little bit weird.

majin1102: I suggest to add some comments to tell the parameter `metadata` would override the existed one.

java/lance-jni/src/file_writer.rs
```diff
@@ -149,6 +149,37 @@ pub extern "system" fn Java_com_lancedb_lance_file_LanceFileWriter_closeNative<'
...
+    writer: JObject,
+    schema_metadata: JObject, // Map<String, String>
+) -> Result<()> {
+    let metadata_map = JMap::from_env(env, &schema_metadata)?;
+    let metadata = to_rust_map(env, &metadata_map)?;
+    let writer_guard =
+        unsafe { env.get_rust_field::<_, _, BlockingFileWriter>(writer, NATIVE_WRITER) }?;
+    let mut writer = writer_guard.inner.lock().unwrap();
+    metadata.into_iter().for_each(|(k, v)| {
+        writer.add_schema_metadata(k, v);
```
majin1102, line 176: nit: I think it might be better to align the interface with rust api 'add_schema_metadata' cause the underlying action didn't really write anything. This is different from writing data because we only flush the final metadata.

steFaiz:
@majin1102 Thanks for your constructive advise. I've modified the code.

majin1102:
Hello，@steFaiz 
I guess you need to resolve the conflicts before merging

steFaiz:
> Hello，@steFaiz I guess you need to resolve the conflicts before merging

@majin1102 Thanks for your remind! I've not noticed that we have changed our repository name.

PR #5315, opened 2025-11-21
fix: stop documenting FTS index type, standardize on INVERTED

mackrorysd:
For this issue: https://github.com/lance-format/lance/issues/5314.

No test included because this is removing behavior that currently doesn't work and is only partially documented (unless we decide we want to fully implement an FTS alias for INVERTED indices. INVERTED index tests should continue to work.


mackrorysd:
@westonpace Thanks - fixed the commit message and CI passed on the rerun.

PR #5317, opened 2025-11-21
docs: fix incorrect URLs and cleanup

prrao87:
- Remove termynal dependencies as we don't need it any more (leaner codebase)
- Update lakehouse stack diagram
- Fix incorrect URL slug

### DIFF
--- a/docs/mkdocs.yml
+++ b/docs/mkdocs.yml
@@ -1,5 +1,5 @@
 site_name: Lance
-site_description: "Documentation for Lance, an open lakehouse file and table format for multimodal AI"
+site_description: "Documentation for Lance, an open lakehouse format for Multimodal AI"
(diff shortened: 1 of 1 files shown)


prrao87:
Oops, slipped through the cracks. It's good to go, I think @Xuanwo. I just updated and tested locally, it works.

docs/mkdocs.yml
```diff
@@ -1,5 +1,5 @@
 site_name: Lance
-site_description: Open Lakehouse Format for Multimodal AI
+site_description: "Documentation for Lance, an open lakehouse file and table format for multimodal AI"
```
prrao87:
```suggestion
site_description: "Documentation for Lance, an open lakehouse format for Multimodal AI"
```

PR #5328, opened 2025-11-24
fix: join job may cause inconsistent delta indices

### LINKED ISSUE #5312: Panic while optimizing vector indices
Getting large number of panics when optimizing vector indices:

```
thread 'main' panicked at 'index out of bounds: the len is 117 but the index is 117': /src/lance/rust/lance-index/src/vector/ivf/storage.rs:95
   0: log_panics::Config::install_panic_hook::{{closure}}
   1: indexer::install_panic_handler::{{closure}}
   2: std::panicking::panic_with_hook
   3: std::panicking::panic_handler::{{closure}}
   4: std::sys::backtrace::__rust_end_short_backtrace
   5: __rustc::rust_begin_unwind
   6: core::panicking::panic_fmt
   7: core::panicking::panic_bounds_check
   8: <lance::index::vector::ivf::v2::IVFIndex<S,Q> as lance_index::vector::VectorIndex>::partition_size
   9: lance::index::vector::builder::IvfIndexBuilder<S,Q>::should_split
  10: lance::index::vector::builder::IvfIndexBuilder<S,Q>::build_partitions::{{closure}}
  11: lance::index::vector::builder::IvfIndexBuilder<S,Q>::build::{{closure}}
  12: lance::index::vector::ivf::optimize_vector_indices_v2::{{closure}}
  13: lance::index::vector::ivf::optimize_vector_indices::{{closure}}
  14: lance::index::<impl lance_index::traits::DatasetIndexExt for lance::dataset::Dataset>::optimize_indices::{{closure}}::{{closure}}
```

BubbleCal:
fix #5312 

- handle the case of delta indices have diff number of partitions, split/join will merge all delta indices into one
- trigger join on incremental indexing, not remap

### DIFF
--- a/rust/lance/src/index/vector/builder.rs
+++ b/rust/lance/src/index/vector/builder.rs
@@ -268,6 +268,7 @@ impl<S: IvfSubIndex + 'static, Q: Quantization + 'static> IvfIndexBuilder<S, Q> 
             ));
         };
 
+        log::info!("remap {} partitions", ivf.num_partitions());
         let existing_index = self.existing_indices[0].clone();
         let mapping = Arc::new(mapping.clone());
         let build_iter = (0..ivf.num_partitions()).map(move |part_id| {
@@ -299,7 +300,7 @@ impl<S: IvfSubIndex + 'static, Q: Quantization + 'static> IvfIndexBuilder<S, Q> 
 
         self.merge_partitions(
             stream::iter(build_iter)
-                .buffer_unordered(get_num_compute_intensive_cpus())
+                .buffered(get_num_compute_intensive_cpus())
                 .boxed(),
         )
         .await?;


rust/lance/src/index/vector.rs
```diff
@@ -633,7 +633,7 @@ pub(crate) async fn build_vector_index_incremental(
                     shuffler,
                     (),
                     frag_reuse_index,
-                    OptimizeOptions::new(),
+                    OptimizeOptions::append(),
```
westonpace: Why is this change needed?

rust/lance-index/src/vector/ivf/storage.rs
```diff
@@ -92,7 +92,7 @@ impl IvfModel {
     }
 
     pub fn partition_size(&self, part: usize) -> usize {
-        self.lengths[part] as usize
+        self.lengths.get(part).copied().unwrap_or_default() as usize
```
wkalt, line 95: this will return a zero in the previous case that panicked. Is that a correct fix? Seems like zero would be an incorrect or unexpected result. What's the consequence of a zero return value?

rust/lance-index/src/vector/v3/shuffler.rs
```diff
@@ -248,6 +248,10 @@ impl ShuffleReader for IvfShufflerReader {
         &self,
         partition_id: usize,
     ) -> Result<Option<Box<dyn RecordBatchStream + Unpin + 'static>>> {
+        if partition_id >= self.partition_sizes.len() {
+            return Ok(None);
+        }
```
wkalt, line 253: similar question here - is it a cause for concern that this function would be asked for an out of bounds partition? If so should this be an error?

rust/lance/src/index/vector/builder.rs
```diff
@@ -1694,6 +1647,18 @@ impl<S: IvfSubIndex + 'static, Q: Quantization + 'static> IvfIndexBuilder<S, Q>
...
+            if part_idx >= index.ivf_model().num_partitions() {
+                // there was a bug that may cause delta indices have different number of partitions,
+                // it's safe to skip loading the extra partition, and split/join the existing partitions,
+                // split/join would merge all delta indices into one so it would fix the issue
+                // see https://github.com/lance-format/lance/issues/5312
+                log::warn!(
+                    "partition index is {} but the number of partitions is {}, skip loading it",
+                    part_idx,
+                    index.ivf_model().num_partitions()
+                );
```
wkalt, line 1661: is it correct to think we'll see this in some cases initially but should not see it in the long term?

rust/lance-index/src/vector/ivf/storage.rs
BubbleCal: this gets the size of the given partition, say delta index A has 3 partitions and delta index B has 4 partitions, when we calculate the total partition size of 4-th partition, index A just returns 0, and this is correct because we can consider that index A has 4 partitions but the 4-th is empty.

and after the fix, we will merge all delta indices into one, so the `unwrap_or_default()` path will never be reached

rust/lance/src/index/vector/builder.rs
BubbleCal: Yes, this is just for fixing the corrupted delta indices, after merging them, we will never reach these branches

rust/lance/src/index/vector/builder.rs
```diff
@@ -1979,6 +1944,12 @@ enum ReassignPartition {
     ReassignCandidate(u32),
 }
 
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+enum PartitionMaintain {
```
wjones127: Maybe a better name?
```suggestion
enum PartitionAdjustment {
```

rust/lance/src/index/vector/builder.rs
```diff
@@ -1979,6 +1944,12 @@ enum ReassignPartition {
...
+    Split(usize),
```
wjones127, line 1952: It wasn't obvious at first glance if this mean "split into N" partition, or "split partition n". I think it's the second one.
```suggestion
    /// Split partition at given id
    Split(usize),
```

PR #5338, opened 2025-11-25
feat: cleanup only scan managed files

### LINKED ISSUE #5294: Cleanup scans unrelated directories and might clean other data.
We have a scenario that stores big video/audio files outside the format(there's a url field) but under the dataset root. The cleanup procedure would scan this data and might clean them unexpected:

https://github.com/lance-format/lance/blob/737c394ca9368bf1d350bf4a2018be1d858fff54/rust/lance/src/dataset/cleanup.rs#L261-L280

The default behivious is only clean files that is verified expired/old by set this parameter: `delete_unverified` false. But I don't think this parameter could expect to clean data unmanaged by table format when it is set true.

The other issue is now we have branches stored under the tree directory. I think we should exclude this directory from cleanup scanning cause each branch would have its own cleanup.

So I suggest to control the scanning scope within dirs including `_version, data, _index, _transaction` for safety and performance.

majin1102:
Close #5294

### DIFF
--- a/rust/lance/src/dataset.rs
+++ b/rust/lance/src/dataset.rs
@@ -129,9 +129,9 @@ pub use write::{
     WriteDestination, WriteMode, WriteParams,
 };
 
-pub const INDICES_DIR: &str = "_indices";
-pub const DATA_DIR: &str = "data";
-pub const TRANSACTIONS_DIR: &str = "_transactions";
+pub(crate) const INDICES_DIR: &str = "_indices";
+pub(crate) const DATA_DIR: &str = "data";
+pub(crate) const TRANSACTIONS_DIR: &str = "_transactions";
 
 // We default to 6GB for the index cache, since indices are often large but
 // worth caching.


majin1102:
Hi @westonpace, this PR is ready for review. 

Please take a look if you have time.

rust/lance/src/dataset.rs
```diff
@@ -128,9 +129,10 @@ pub use write::{
     WriteDestination, WriteMode, WriteParams,
 };
 
-const INDICES_DIR: &str = "_indices";
-
+pub const INDICES_DIR: &str = "_indices";
 pub const DATA_DIR: &str = "data";
+pub const TRANSACTIONS_DIR: &str = "_transactions";
```
yanghua: If we keep the visibility to be `pub(crate) `, is there any problem?

majin1102: OK, let's use them that way

majin1102:
Thanks for reviewing, I will wait 2 days to see if there are any other comments.
cc @westonpace @jackye1995

rust/lance-table/src/io/commit.rs
```diff
@@ -67,7 +67,7 @@ use {
     std::time::{Duration, SystemTime},
 };
 
-const VERSIONS_DIR: &str = "_versions";
+pub const VERSIONS_DIR: &str = "_versions";
```
westonpace, line 70: There's some inconsistency with all these dir constants.  Some are here, others are in dataset.rs.  I don't think it really matters but it might be nice sometime to just put them all in datasets.rs and either make them all pub (which I think we've done here) or refer to them all from the `Dataset` accessor functions.

westonpace (commented):
Overdue review but wanted to thank you for the contribution and good idea.

rust/lance-table/src/io/commit.rs
majin1102: Thanks for the suggestion.

I was planning to go with the latter approach, but I didn’t want to be too aggressive. I will carry this forward in other PRs

PR #5339, opened 2025-11-25
refactor!: use org.lance namespace for java package

jackye1995:
Looks like lance-namespace needs to be renamed first, and also I might need to separate lance-namespace integrations to a separated lance-namespace-impls for all the other non-rest non-dir implementations. Publish the draft first while thinking about it.

### DIFF
--- a/java/lance-jni/src/blocking_blob.rs
+++ b/java/lance-jni/src/blocking_blob.rs
@@ -35,7 +35,7 @@ impl From<BlobFile> for BlockingBlobFile {
 }
 
 #[no_mangle]
-pub extern "system" fn Java_com_lancedb_lance_Dataset_nativeTakeBlobs<'local>(
+pub extern "system" fn Java_org_lance_Dataset_nativeTakeBlobs<'local>(
     mut env: JNIEnv<'local>,
     jdataset: JObject,
     row_ids_obj: JObject, // List<Long>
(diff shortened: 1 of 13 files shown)


jackye1995:
I think given we have moved the main LanceNamespace trait in rust to core repo, we should also move the python and java one to the core repo. This will make lance-namespace repo responsible for 2 aspects: (1) codegen of the request response models in different languages, (2) all language specific namespace implementations except the directory and rest namespaces, which we will maintain in the main repo through rust binding.

Let me know what you think @majin1102

jackye1995:
This means we can merge this repo as is, and I will do another PR to do the migration of the LanceNamespace interfaces in python and java

majin1102:
> I think given we have moved the main LanceNamespace trait in rust to core repo, we should also move the python and java one to the core repo. This will make lance-namespace repo responsible for 2 aspects: (1) codegen of the request response models in different languages, (2) all language specific namespace implementations except the directory and rest namespaces, which we will maintain in the main repo through rust binding.
> 
> Let me know what you think @majin1102 

Sorry for being late，I've been on a business trip these past two days. I'm wondering if doing this would undermine the neutrality of the namespace and its ability to adapt to multiple formats?

In fact, we are considering replacing the catalog with a namespace at the product layer to adapt to more AI-native scenarios.

PR #5342, opened 2025-11-25
chore: don't print warning if fp16 kernel not build

Xuanwo:
This PR will suppress the warning for fp16 kernel not being built

I personally think this warning isn't helpful, maybe we can just remove it?

---


westonpace:
> I personally think this warning isn't helpful, maybe we can just remove it?

The purpose of the warning is to ensure that our production builds (which we deploy to real world workloads) have the correct features enabled.

Perhaps a better way to solve this same problem would be to add a `features` method which, at runtime, tells which features were compiled into the binary.  This could then be used in a CI or pre/post-deploy smoke test of some kind to ensure that we are running builds in production that have all features builtin?

Xuanwo:
> Perhaps a better way to solve this same problem would be to add a `features` method which, at runtime, tells which features were compiled into the binary.

Seems nice to me. Are you expecting to add this `features` method and include a dedicated test for it in the smoke tests?

BubbleCal:
how can we know if fp16 kernel is supported after this?

Xuanwo:
> how can we know if fp16 kernel is supported after this?

check if `fp16` included in `pylance.enabled_features()`;

BubbleCal (approved):

Xuanwo:
Not that meaningful, let's just close.

PR #5347, opened 2025-11-26
feat: return Unprocessable error while expected error happened

Xuanwo:
This PR intends to return Unprocessable error while expected error happened. In this way, our users can print helpful message to users instead of just return the Index error.

---

### DIFF
--- a/.github/workflows/rust-benchmark.yml
+++ b/.github/workflows/rust-benchmark.yml
@@ -49,6 +49,7 @@ jobs:
           # TODO: a few benchmarks are failing. Re-enable everything once they are fixed.
           cargo bench --bench sq --bench hnsw --bench inverted --bench pq_dist_table --bench pq_assignment -- --output-format bencher | tee -a ../../output.txt
       - name: Store benchmark result
+        if: github.event_name != 'pull_request'
         uses: benchmark-action/github-action-benchmark@v1
         with:
           name: Lance Rust Benchmarks


rust/lance-core/src/error.rs
```diff
@@ -63,6 +63,8 @@ pub enum Error {
     Internal { message: String, location: Location },
     #[snafu(display("A prerequisite task failed: {message}, {location}"))]
     PrerequisiteFailed { message: String, location: Location },
+    #[snafu(display("Precondition failed: {message}, {location}"))]
+    Precondition { message: String, location: Location },
```
westonpace: Is Precondition meaningfully different from InvalidInput?

Xuanwo:
cc @BubbleCal, maybe you will be interested in this PR.

rust/lance-core/src/error.rs
Xuanwo: Yes, `Precondition` here follows the same pattern as HTTP Status 412 `Precondition`, used by many services to indicate that the request itself is valid but the condition is not met. In this case, the dataset doesn’t have enough data. I wish this error could be used by users to identify these cases instead of bypassing real invalid input errors.

jackye1995: What about just saying it is unsupported? since this after all is us not able to support building/refreshing the index with this amount of data.

BubbleCal (approved):
I think we do error mapping somewhere else (e.g. lancedb), we may need to add a new map for `Precondition`

Xuanwo:
> My only nit with 412 is that it is normally used where the caller provides the preconditions (e.g. if-... headers). I think HTTP 422 (unprocessable) might be slightly more accurate. However, we are well into bike-shedding at this point so I'll trust your judgement.

Oh, I learned a new HTTP status code. After some thought, I think `unprocessable` better matches our semantics here, will add a patch for this.

westonpace:
> Oh, I learned a new HTTP status code.

To be fair, I had to look it up :laughing: I thought I remembered something like this but just couldn't remember the actual words.

PR #5348, opened 2025-11-26
refactor: allow datafiles to contain columns without field id

Xuanwo:
This PR removes the check that requires all columns in data files to have a field ID. The check was only for safety and doesn’t enforce any real restriction in our logic. I’m removing it so we can keep building Blob V2.

---

### DIFF
--- a/rust/lance/src/dataset/fragment.rs
+++ b/rust/lance/src/dataset/fragment.rs
@@ -1194,8 +1194,6 @@ impl FileFragment {
     /// Verifies:
     /// * All field ids in the fragment are distinct
     /// * Within each data file, field ids are in increasing order
-    /// * All fields in the schema have a corresponding field in one of the data
-    ///   files (extra columns without field ids are allowed and ignored)
     /// * All data files exist and have the same length
     /// * Field ids are distinct between data files.
     /// * Deletion file exists and has rowids in the correct range


rust/lance-table/src/format/fragment.rs
```diff
@@ -33,6 +33,9 @@ pub struct DataFile {
     ///
     /// Note that -1 is a possibility and it indices that the field has
     /// no top-level column in the file.
+    ///
+    /// Columns that lack a field id may still exist as extra entries in
+    /// `column_indices`; such columns are ignored by field-id–based projection.
```
westonpace, line 38:
```suggestion
    ///
    /// Columns that lack a field id may still exist as extra entries in
    /// `column_indices`; such columns are ignored by field-id–based projection.
    /// For example, some fields, such as blob fields, occupy multiple
    /// columns in the file but only have a single field id.
```

rust/lance/src/dataset/fragment.rs
```diff
@@ -1195,7 +1195,7 @@ impl FileFragment {
     /// * All field ids in the fragment are distinct
     /// * Within each data file, field ids are in increasing order
     /// * All fields in the schema have a corresponding field in one of the data
-    ///   files
+    ///   files (extra columns without field ids are allowed and ignored)
```
westonpace: I'm not sure this change is helpful because I don't think that is what we were checking.

However, I think this comment might have been wrong even before your change.  You could probably just remove the whole bullet point.  We allow schema fields to be missing from a fragment (in which case all values for that field are null)

The only thing we check in this method is that there is at least one schema field represented by the fragment (e.g. we make sure the fragment hasn't become entirely obsolete).

Xuanwo: updated.

PR #5349, opened 2025-11-26
feat: support map data type in lance format version 2.2

xloya:
Close https://github.com/lance-format/lance/issues/3620.
Currently, Lance does not support the `Map` data type. Importing Lance from a data source that supports `Map` data type requires special handling, which incurs significant processing costs, even though this type is very common in other data sources.

This PR aligns with the Map data type in Arrow, implementing the Map logical data type. In actual encoder, it uses an `Offsets Array + List<Struct<key, value>>` approach, which is similar to `List` data type .
And in actual decoder, it will decode the struct array and offset infos to restruct to the Arrow `MapArray`.

### DIFF
--- a/rust/lance-encoding/src/encodings/logical/map.rs
+++ b/rust/lance-encoding/src/encodings/logical/map.rs
@@ -216,7 +216,10 @@ impl StructuralDecodeArrayTask for StructuralMapDecodeTask {
         let entries = array
             .as_any()
             .downcast_ref::<arrow_array::StructArray>()
-            .expect("Map entries should be a StructArray")
+            .ok_or_else(|| Error::Schema {
+                message: "Map entries should be a StructArray".to_string(),
+                location: location!(),
+            })?
             .clone();
 
         // Build the MapArray from offsets, entries, validity, and keys_sorted


xloya:
@westonpace @jackye1995 @Xuanwo @eddyxu PTAL when you have time, thanks!

rust/lance-encoding/src/encodings/logical/map.rs
```diff
@@ -0,0 +1,607 @@
...
+
+        // Convert the decoded array to StructArray
+        let entries = array
+            .as_any()
+            .downcast_ref::<arrow_array::StructArray>()
+            .expect("Map entries should be a StructArray")
+            .clone();
+
+        // Build the MapArray from offsets, entries, and validity
+        let map_array = MapArray::new(entries_field, offsets, entries, validity, false);
```
Xuanwo: Do we need to find a way to preserve the `is_sorted` value?

xloya:
> Thank you for working on this!
> 
> Lance format 2.1 has been stabilized, so we can't add new data type support in its current form. Otherise older versions of Lance wouldn’t be able to read new Lance files.
> 
> We should add this feature in format version 2.2 instead.

@Xuanwo Thanks for your review! Currently, most of the logic for adding the Map data type is located in the Field, Encoder, and Decoder modules. These modules generally don't contain format version information. I'm unsure if it's appropriate to limit support for the Map data type to version 2.2+, since version 2.1+ uses the same logic for reading, writing, and encoding/decoding (different from version 2.0 and below). Do you have any suggestions on this?

rust/lance-encoding/src/encodings/logical/map.rs
xloya: I'll give it a try, perhaps I'll need to add a new `sorted_map` logical type.

xloya: @Xuanwo I tried building a separate logical type, `sorted_map`, and successfully restored the `keys_sorted` attribute to the Rust arrow schema. However, I found that the `keys_sorted` attribute is lost when converting from the Rust arrow schema to the Pyarrow schema using the following code in python binding sdk `dataset.rs`:
```
#[getter(schema)]
fn schema(self_: PyRef<'_, Self>) -> PyResult<PyObject> {
    let logical_schema = logical_schema_from_lance(self_.ds.schema());
    # will loss the `keys_sorted` property in MapType when returning the PyResult.
    logical_schema.to_pyarrow(self_.py())
}
``` 
I don't have much experience with cross-language arrow conversion and wonder if there are other ways to restore it. If we keep it as it is, the only way I can think of is to record it in the metadata and then reset the schema containing the map type data in the Python binding SDK based on the metadata. What do you think of this?

xloya: Another option is that we only support the Map type with keys sorted as false this pr and consider supporting sorted maps when there is a better solution in the future.

Xuanwo:
> @Xuanwo Thanks for your review! Currently, most of the logic for adding the Map data type is located in the Field, Encoder, and Decoder modules. These modules generally don't contain format version information. I'm unsure if it's appropriate to limit support for the Map data type to version 2.2+, since version 2.1+ uses the same logic for reading, writing, and encoding/decoding (different from version 2.0 and below). Do you have any suggestions on this?

Take Lance v1.0.0 as an example. If we add the `Map` type, users might find that data written in Lance v1.0.0 is not readable by Lance v0.39.0, and that's what we want to avoid.

rust/lance-core/src/datatypes/field.rs
```diff
@@ -980,16 +992,42 @@ impl TryFrom<&ArrowField> for Field {
...
                 .iter()
                 .map(|f| Self::try_from(f.as_ref()))
                 .collect::<Result<_>>()?,
             DataType::List(item) => vec![Self::try_from(item.as_ref())?],
             DataType::LargeList(item) => vec![Self::try_from(item.as_ref())?],
+            DataType::Map(entries, keys_sorted) => {
+                // TODO: We only support keys_sorted=true for now,
+                //  because converting a rust arrow map field to the python arrow field will
+                //  lose the keys_sorted property.
+                if *keys_sorted {
```
xloya, line 1050: @Xuanwo I tried to restore the `keys_sorted` parameter in the Map type, which worked fine in Rust, but the parameter was lost once it converted to Python Arrow Schema using `PyArrowType`. I think this might be a bug with the Arrow code itself. Therefore, I want to only support Map type with `keys_sorted=false` in this PR, which is also the default value of Arrow MapType, and continue to investigate this issue. I will consider supporting Map type with `keys_sorted=true` in a future PR. But to be honest, this parameter does not actually sort the Map in Arrow; it is only used as metadata, so I don't think it will make a significant difference to users.

rust/lance-encoding/src/encoder.rs
```diff
@@ -422,6 +423,42 @@ impl StructuralEncodingStrategy {
...
+                    // TODO: We only support keys_sorted=true for now,
+                    //  because converting a rust arrow map field to the python arrow field will
+                    //  lose the keys_sorted property.
+                    if keys_sorted {
+                        return Err(Error::NotSupported {
+                            source: format!("Map data type is only supported with keys_sorted=true now, current value is {}", keys_sorted).into(),
+                            location: location!(),
+                        });
+                    }
+                    if self.version < LanceFileVersion::V2_2 {
```
xloya, line 450: Versions 2.0 and below use a separate encoder / decoder, which inherently does not support Map types. Newer versions of the encoder / decoder disable the Map types use of version 2.1 because it is now stable.

rust/lance-core/src/datatypes/field.rs
Xuanwo: Got it, thank you for working on this

rust/lance-encoding/src/encodings/logical/struct.rs
```diff
@@ -272,9 +274,18 @@ impl StructuralStructDecoder {
                     field.data_type().clone(),
                 ))
             }
+            DataType::Map(entries_field, keys_sorted) => {
+                if *keys_sorted {
+                    todo!("Maps are not supported keys_sorted=true yet")
```
Xuanwo: It's better to return an unsupported error instead of panic.

rust/lance-encoding/src/decoder.rs
```diff
@@ -501,6 +504,20 @@ impl CoreFieldDecoderStrategy {
             validate_data: config.validate_on_decode,
             decompressor_strategy: Arc::new(DefaultDecompressionStrategy {}),
             cache_repetition_index: config.cache_repetition_index,
+            file_version: LanceFileVersion::default(),
+        }
+    }
+
+    /// Create a new strategy from decoder config with file version
+    pub fn from_decoder_config_with_version(
```
Xuanwo: Can we have an `with_file_version` API instead?

rust/lance-encoding/src/testing.rs
```diff
@@ -186,6 +194,7 @@ async fn test_decode(
         cache,
         &FilterExpression::no_filter(),
         &DecoderConfig::default(),
+        LanceFileVersion::V2_2, // Default to V2_2 for testing
```
Xuanwo: This change can introduce other 2.2 issues affecting our testing. I suggest configuring it separately for 2.2-related tests.

rust/lance-file/src/reader.rs
```diff
@@ -933,6 +936,7 @@ impl FileReader {
         projection: ReaderProjection,
         filter: FilterExpression,
         decoder_config: DecoderConfig,
+        file_version: LanceFileVersion,
```
Xuanwo: We’ve introduced `file_version` in too many places, can we avoid it? In fact we just need to avoid users trying to write map data under 2.2 version.

xloya:
> Here some nits about this PR. Maybe @westonpace wanna take an other look.

@Xuanwo Addressed all the comments, PTAL again when you have time, thanks!

xloya:
@Xuanwo @westonpace @jackye1995 Gentle pin for this, thanks!

rust/lance-encoding/src/decoder.rs
```diff
@@ -773,6 +774,25 @@ impl CoreFieldDecoderStrategy {
                 Ok(Box::new(StructuralListScheduler::new(child_scheduler))
                     as Box<dyn StructuralFieldScheduler>)
             }
+            DataType::Map(_, keys_sorted) => {
+                // TODO: We only support keys_sorted=false for now,
+                //  because converting a rust arrow map field to the python arrow field will
+                //  lose the keys_sorted property.
+                if *keys_sorted {
+                    return Err(Error::NotSupported {
+                        source: format!("Map data type is only supported with keys_sorted=true now, current value is {}", *keys_sorted).into(),
```
Xuanwo: This error message seems to be wrong: we didn't support `keys_sorted=true`

rust/lance-encoding/src/encoder.rs
```diff
@@ -436,6 +437,42 @@ impl StructuralEncodingStrategy {
                         child_encoder,
                     )))
                 }
+                DataType::Map(_, keys_sorted) => {
+                    // TODO: We only support keys_sorted=false for now,
+                    //  because converting a rust arrow map field to the python arrow field will
+                    //  lose the keys_sorted property.
+                    if keys_sorted {
+                        return Err(Error::NotSupported {
+                            source: format!("Map data type is only supported with keys_sorted=true now, current value is {}", keys_sorted).into(),
```
Xuanwo: The same.

rust/lance-encoding/src/encodings/logical/map.rs
```diff
@@ -0,0 +1,657 @@
...
+
+        // Extract the entries field and keys_sorted from the map data type
+        let (entries_field, keys_sorted) = match &self.data_type {
+            DataType::Map(field, keys_sorted) => (field.clone(), *keys_sorted),
+            _ => panic!("Map decoder did not have a map field"),
+        };
+
+        if keys_sorted {
+            panic!("Map type decoder does not support keys_sorted=true now")
+        }
```
Xuanwo: I suggest we return errors instead of panic here.

rust/lance-core/src/datatypes/field.rs
```diff
@@ -981,16 +993,42 @@ impl TryFrom<&ArrowField> for Field {
...
+                    return Err(Error::Schema {
+                        message: "Unsupported map field with keys_sorted=true".to_string(),
+                        location: location!(),
+                    });
+                }
+                // Validate Map key field is non-nullable (Arrow Map specification)
+                if let DataType::Struct(fields) = entries.data_type() {
```
Xuanwo: I'm not sure how strict the checks need to be here. For example:

- If `entries` is not a struct, should we just return an error instead of bypassing it?
- Do we need to check if `value_field` exists?

python/python/tests/test_map_type.py
```diff
@@ -0,0 +1,856 @@
...
+        lance.write_dataset(data, tmp_path / "v21", data_storage_version="2.1")
+    # Verify error message
+    error_msg = str(exc_info.value)
+    assert (
+        "Map data type" in error_msg
+        or "not yet implemented" in error_msg.lower()
+        or "not supported" in error_msg.lower()
+    )
+
+    # should raise a panic for v2.0 (not yet implemented)
```
Xuanwo: Do we really need to test this?

rust/lance-encoding/src/encodings/logical/map.rs
xloya: ok

rust/lance-core/src/datatypes/field.rs
xloya: Yes, you're right, we can add some more checks here.

python/python/tests/test_map_type.py
xloya: Will remove this.

xloya:
@Xuanwo Address the comments, sorry to bother you again, please take another look when you have time, thanks!

Xuanwo (approved):
Thank you for working on this! Can you help update our docs as a follow-up?

xloya:
> Thank you for working on this! Can you help update our docs as a follow-up?

Sure, I’ll open a PR to update docs soon.

westonpace (commented):
This is really good!  Sorry I took a while look at it.  Thanks for the addition :rocket:

westonpace:
I have one small suggestion I have made here: https://github.com/lance-format/lance/pull/5513

PR #5355, opened 2025-11-27
fix: parallelize bitmap partition loading in IsIn expressions

wkalt:
Prior to this commit an IsIn query on a bitmap index would load partitions in serial order. On high latency connections, queries with large IsIn lists can get very slow. This commit changes the partition loading to be parallelized over the number of CPUs, similar to btree.

### DIFF
--- a/rust/lance-index/src/scalar/bitmap.rs
+++ b/rust/lance-index/src/scalar/bitmap.rs
@@ -442,7 +442,7 @@ impl ScalarIndex for BitmapIndex {
                         let this = self.clone();
                         async move { this.load_bitmap(&key, None).await }
                     }))
-                    .buffered(get_num_compute_intensive_cpus())
+                    .buffer_unordered(get_num_compute_intensive_cpus())
                     .try_collect()
                     .await?;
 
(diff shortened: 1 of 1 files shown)


wkalt:
Here is some performance data
benchmark_plot

rust/lance-index/src/scalar/bitmap.rs
```diff
@@ -438,11 +438,13 @@ impl ScalarIndex for BitmapIndex {
...
-                    let mut bitmaps = Vec::new();
-                    for key in keys {
-                        let bitmap = self.load_bitmap(&key, Some(metrics)).await?;
-                        bitmaps.push(bitmap);
-                    }
+                    let bitmaps: Vec<_> = stream::iter(keys.into_iter().map(|key| {
+                        let this = self.clone();
+                        async move { this.load_bitmap(&key, None).await }
+                    }))
+                    .buffered(get_num_compute_intensive_cpus())
```
BubbleCal: it looks like we don't care the order, `buffered_unordered` may offer better performance

BubbleCal (approved):
Good catch!

rust/lance-index/src/scalar/bitmap.rs
wkalt: thanks, updated.

PR #5363, opened 2025-11-27
perf: use CPU pool to run WAND algo

BubbleCal:
this reduces 10%~20% cold latency for full text search

### DIFF
--- a/rust/lance-index/src/scalar/inverted/index.rs
+++ b/rust/lance-index/src/scalar/inverted/index.rs
@@ -9,7 +9,7 @@ use std::{
 };
 use std::{collections::HashMap, ops::Range};
 
-use crate::metrics::{LocalMetricsCollector, NoOpMetricsCollector};
+use crate::metrics::NoOpMetricsCollector;
 use crate::prefilter::NoFilter;
 use crate::scalar::registry::{TrainingCriteria, TrainingOrdering};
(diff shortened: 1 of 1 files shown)


rust/lance-index/src/scalar/inverted/index.rs
```diff
@@ -770,9 +791,20 @@ impl InvertedPartition {
...
             .try_collect::<Vec<_>>()
             .await?;
-        let scorer = IndexBM25Scorer::new(std::iter::once(self));
-        let mut wand = Wand::new(operator, postings.into_iter(), &self.docs, scorer);
-        wand.search(params, mask, metrics)
+        let params = params.clone();
+        let partition_ptr = PartitionPtr::new(self);
+        let (candidates, local_metrics) = spawn_cpu(move || {
+            let local_metrics = LocalMetricsCollector::default();
+            // SAFETY: `partition_ptr` points to `self`, which outlives this task because we await it.
```
Xuanwo: Can we avoid this? This can't be guaranteed since the future itself could be dropped at any time.

Xuanwo: I was wondering if you could refactor this function.

Move the postings calculation out and await it separately, since `bm25_search` is just a pure blocking function.

Xuanwo (approved):
Nice, I love it

PR #5372, opened 2025-11-28
feat: disable default features on internal use

valkum:
Part of https://github.com/lancedb/lancedb/issues/2771, continuing https://github.com/lancedb/lancedb/pull/2568.
`lance-namespace-impls` uses `lance` which pulled in default features (aws,gcp,azure,oss), which pulled `aws-*` crates into `lancedb` even when `lancedb` was used with `default-features = false`.

I don't have tests at hand, but it would probably be good to run `cargo tree -p lance -e no-build -e no-dev --no-default-features -i aws-config` in CI (maybe for each crate?) to avoid adding any dependency back by default in a future PR.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -62,7 +62,7 @@ lance-index = { version = "=1.1.0-beta.2", path = "./rust/lance-index" }
 lance-io = { version = "=1.1.0-beta.2", path = "./rust/lance-io", default-features = false }
 lance-linalg = { version = "=1.1.0-beta.2", path = "./rust/lance-linalg" }
 lance-namespace = { version = "=1.1.0-beta.2", path = "./rust/lance-namespace" }
-lance-namespace-impls = { version = "=1.1.0-beta.2", path = "./rust/lance-namespace-impls", default-features = false }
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -50,7 +50,7 @@ rust-version = "1.82.0"
 
 [workspace.dependencies]
 libc = "0.2.176"
-lance = { version = "=1.1.0-beta.2", path = "./rust/lance" }
+lance = { version = "=1.1.0-beta.2", path = "./rust/lance", default-features = false }
```
jackye1995, line 53: Actually I don't think we need this?

valkum: I am not at my computer right now, but this is the core issue. All crates that use lance inside this workspace will enable the default features of lance.

jackye1995: Can we move this `default-features = false` to only `lance-namespace-impls`? Because examples also depend on lance, and should have the default features enabled so people can try the example in cloud

jackye1995: I added a patch to make examples still have all features, let me know if this looks good

PR #5374, opened 2025-12-01
refactor: consolidate logic between zonemap and bloomfilter indexes

### LINKED ISSUE #5230: Consolidate the logic between ZoneMap and BloomFilters indexes
These two are zone based scalar indexes and they share a lot of similarity. Work needs to be done to reduce code duplication and simplify the logic.

fenfeng9:
Closes #5230

#### Background
ZoneMap and BloomFilter indexes both needed zone training and search capabilities, so this PR factors those shared pieces into a common module that the two indexes now reuse.

#### Changes
- add `scalar::zoned` with a shared trainer (`ZoneTrainer`/`ZoneProcessor`), `ZoneBound`, `search_zones`, and `rebuild_zones`
- migrate the zonemap index to the shared infrastructure so training, updating, and searching all use the common logic
- migrate the bloomfilter index the same way to remove duplicated state and ensure consistent row-range handling

### DIFF
--- a/rust/lance-index/src/scalar/bloomfilter.rs
+++ b/rust/lance-index/src/scalar/bloomfilter.rs
@@ -1160,6 +1160,14 @@ impl ScalarIndexPlugin for BloomFilterIndexPlugin {
                 as Arc<dyn ScalarIndex>,
         )
     }
+
+    async fn load_statistics(
+        &self,
+        _index_store: Arc<dyn IndexStore>,
+        _index_details: &prost_types::Any,
+    ) -> Result<Option<serde_json::Value>> {
+        Ok(None)
+    }
 }
 
 #[derive(Debug)]
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -9,6 +9,11 @@ debug = true
 codegen-units = 16
 lto = "thin"
 
+[profile.bench]
+inherits = "release"
+lto = "thin"
+codegen-units = 16
+
 [target.x86_64-unknown-linux-gnu]
 rustflags = ["-C", "target-cpu=haswell", "-C", "target-feature=+avx2,+fma,+f16c"]
 
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -37,6 +37,19 @@ jobs:
           components: rustfmt
       - name: Check formatting
         run: cargo fmt -- --check
+
+  rustdoc:
+    runs-on: ubuntu-24.04
+    timeout-minutes: 30
+    steps:
+      - uses: actions/checkout@v4
+      - name: Install dependencies
+        run: |
+          sudo apt update
+          sudo apt install -y protobuf-compiler libssl-dev
+      - name: Check documentation
+        run: RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps
+
   clippy:
     permissions:
       checks: write
--- a/deny.toml
+++ b/deny.toml
@@ -84,7 +84,8 @@ ignore = [
     { id = "RUSTSEC-2024-0370", reason = "`proc-macro-error` is used by jieba-rs via include-flate" },
     { id = "RUSTSEC-2024-0436", reason = "`paste` is used by datafusion" },
     { id = "RUSTSEC-2023-0071", reason = "`rsa` is used by opendal via reqsign" },
-    { id = "RUSTSEC-2025-0119", reason = "`number_prefix` used by hf-hub in examples" }
+    { id = "RUSTSEC-2025-0119", reason = "`number_prefix` used by hf-hub in examples" },
+    { id = "RUSTSEC-2025-0134", reason = "`rustls-pemfile` unmaintained; awaiting upstream object_store/hyper-rustls migration to rustls-pki-types" }
 ]
 # If this is true, then cargo deny will use the git executable to fetch advisory database.
 # If this is false, then it uses a built-in git library.
--- a/protos/encodings_v2_1.proto
+++ b/protos/encodings_v2_1.proto
@@ -102,7 +102,7 @@ message MiniBlockLayout {
   // If there is repetition then the depth must be at least 1.  If there are many layers
   // of repetition then deeper repetition indices will support deeper nested random access.  For
   // example, given 5 layers of repetition then the repetition index depth must be at least
-  // 3 to support access like rows[50][17][3].
+  // 3 to support access like `rows[50][17][3]`.
   //
   // We require `repetition_index_depth + 1` u64 values per mini-block to store the repetition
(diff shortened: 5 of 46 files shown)


fenfeng9:
Hi @HaochengLIU, sorry for the delay—could you take a look at the PR when you have a chance? It consolidates the shared zone logic between ZoneMap and BloomFilter. Happy to address any feedback, thanks!

rust/lance-index/src/scalar/bloomfilter.rs
```diff
@@ -45,34 +40,19 @@ use lance_core::Result;
...
-    // zone_start is start row of the zone in the fragment, also known
-    // as the local offset. To get the actual first row address,
-    // you can do `fragment_id << 32 + zone_start`
-    zone_start: u64,
-    // zone_length is the `row offset span` between the first and the last row in the current SBBF block
-    // calculated as: (last_row_offset - first_row_offset + 1)
-    zone_length: usize,
+    // Bound of this zone within the fragment. This combines the persisted
+    // `(fragment_id, zone_start, zone_length)` columns that are written to
+    // `bloomfilter.lance`, matching the previous `zone_*` fields.
```
HaochengLIU: we can remove the ", matching the previous `zone_*` fields." part

rust/lance-index/src/scalar/zonemap.rs
```diff
@@ -73,14 +59,10 @@ struct ZoneMapStatistics {
...
-    // zone_start is start row of the zone in the fragment, also known
-    // as the local offset. To get the actual first row address,
-    // you can do `fragment_id << 32 + zone_start`
-    zone_start: u64,
-    // zone_length is the `row offset span` between the first and the last row in the zone
-    // calculated as: (last_row_offset - first_row_offset + 1)
-    zone_length: usize,
+    /// Bound of this zone within the fragment. It combines the persisted
+    /// `(fragment_id, zone_start, zone_length)` columns that are written to
+    /// `zonemap.lance`, matching the former `zone_*` fields.
```
HaochengLIU: ditto here on the `zone_*` wording

rust/lance-index/src/scalar/zoned.rs
```diff
@@ -0,0 +1,890 @@
...
+
+        while let Some(batch) = batches.try_next().await? {
+            if batch.num_rows() == 0 {
+                continue;
+            }
+
+            let values = batch.column(0);
+            let row_addr_col = batch
+                .column_by_name(ROW_ADDR)
+                .ok_or_else(|| Error::invalid_input("missing _rowaddr column", location!()))?
```
HaochengLIU: Thse AI generated safe guards are unnecessary. You should revert it to the [original form](https://github.com/lance-format/lance/blob/main/rust/lance-index/src/scalar/bloomfilter.rs) for clarity.

rust/lance-index/src/scalar/zoned.rs
```diff
@@ -0,0 +1,890 @@
...
+                    row_addr_col,
+                    batch_offset,
+                    batch.num_rows(),
+                    fragment_id,
+                );
+                let capacity = zone_size - current_zone_len;
+                let take = run_len.min(capacity);
+
+                if take == 0 {
+                    // Should not happen because flush occurs before reaching here,
```
HaochengLIU: Is it trying to cover the [desired ==0](https://github.com/lance-format/lance/blob/main/rust/lance-index/src/scalar/bloomfilter.rs#L1079-L1087) logic? That's a valid condition. If not then delete the guard here as the possible infinite loop condition should be fixed in its root cause.

rust/lance-index/src/scalar/zoned.rs
```diff
@@ -0,0 +1,890 @@
...
+
+    fn flush_zone(
+        processor: &mut P,
+        zones: &mut Vec<P::ZoneStatistics>,
+        fragment_id: u64,
+        current_zone_len: &mut usize,
+        zone_start_offset: &mut Option<u64>,
+        zone_end_offset: &mut Option<u64>,
+    ) -> Result<()> {
+        if *current_zone_len == 0 {
```
HaochengLIU: Is it really needed?

rust/lance-index/src/scalar/zoned.rs
```diff
@@ -0,0 +1,890 @@
...
+
+    /// Count how many consecutive rows belong to the same fragment starting from `start`.
+    /// Returns the number of rows before encountering a different fragment or reaching the end.
+    fn count_run_length(
+        row_addrs: &UInt64Array,
+        start: usize,
+        total: usize,
+        fragment_id: u64,
+    ) -> usize {
+        let mut idx = start;
```
HaochengLIU: This function is uncessary. It could be replaced with something like
```rust
        (start..total)
            .take_while(|&idx| (row_addrs.value(idx) >> 32) == fragment_id)
            .count()
```

HaochengLIU:
@fenfeng9 finish the first round of review

fenfeng9:
Thanks for the feedback.I'll address these comments and fix the conflicts tomorrow.

rust/lance-index/src/scalar/bloomfilter.rs
fenfeng9: Updated

rust/lance-index/src/scalar/zonemap.rs
fenfeng9: Updated

rust/lance-index/src/scalar/zoned.rs
fenfeng9: I've reverted it to `unwrap()`.

rust/lance-index/src/scalar/zoned.rs
fenfeng9: I've removed it because take == 0 is unreachable here.
take is min(run_len, capacity). run_len is at least 1 (the current row always matches itself), and capacity is guaranteed to be positive because we flush_zone immediately whenever the zone becomes full or the fragment changes.

rust/lance-index/src/scalar/zoned.rs
fenfeng9: You are right. I missed the check in `train()` earlier—it is indeed possible for len to be 0 there if a zone fills up right before a fragment switch.

I've fixed it by adding `if current_zone_len > 0` at the `train()` and removed the defensive check inside `flush_zone`

rust/lance-index/src/scalar/zoned.rs
fenfeng9: Done. Updated to use take_while as suggested

HaochengLIU:
The PR lgtm. Once all tests pass I'll merge it. Thanks for the contribution! @fenfeng9

PR #5378, opened 2025-12-01
feat(cdf): support set start/end timestamp in cdf

zhangyue19921010:
closes: https://github.com/lance-format/lance/issues/5445

Iterate through versions, find the first version that is greater than or equal to it based on the begin date, and find the last version that is less than it based on the end_date

```
        let delta = ds
            .delta()
            .with_begin_date(begin_ts)
            .with_end_date(end_ts)
            .build()
            .unwrap();
```

### DIFF
--- a/rust/lance/src/dataset/delta.rs
+++ b/rust/lance/src/dataset/delta.rs
@@ -165,6 +165,12 @@ impl DatasetDeltaBuilder {
                     location!(),
                 ));
             }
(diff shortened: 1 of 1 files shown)


yanghua (approved):
LGTM. It would be better to update the related documentation: https://lance.org/format/table/row_id_lineage/#change-data-feed

PR #5379, opened 2025-12-01
fix: avx512 related symbol not found in mac x86

jackye1995:
related failure in lancedb: https://github.com/lancedb/lancedb/actions/runs/19769925006/job/56653636802?pr=2826

There are two independent AVX-512 builds that both set the same cfg flag kernel_support="avx512":
  1. f16 kernels (line 45-56): Uses -march=sapphirerapids - a fixed cross-compilation target
  2. dist_table (line 58-64): Uses -march=native - uses the build machine's CPU features

Then in dist_table.rs:
  - Lines 38-48: The AVX-512 code path is compiled (because cfg is set)
  - Lines 164-171: The extern "C" declaration for sum_4bit_dist_table_32bytes_batch_avx512 is compiled
  - But the C code was never built, so the symbol doesn't exist

Separate cfg flags should be used for f16 and dist_table AVX-512 support

### DIFF
--- a/python/Cargo.lock
+++ b/python/Cargo.lock
@@ -5083,12 +5083,13 @@ dependencies = [
 
 [[package]]
 name = "object_store_opendal"
-version = "0.54.1"
+version = "0.55.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c0b88fc0e0c4890c1d99e2b8c519c5db40f7d9b69a0f562ff1ad4967a4c8bbc6"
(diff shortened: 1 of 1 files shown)


westonpace (approved):
Approve the change but I'm not sure I understand the situation that caused the error so I'm not sure I understand if it would be feasible, practical, or useful to ask for some kind of test.

jackye1995:
Tried to verify in lancedb but can't do that without a pypi release. Will merge this first and see if it works.

PR #5380, opened 2025-12-01
ci: fix flaky test_read_btree_index_with_defer_index_remap

jackye1995:
Closes https://github.com/lance-format/lance/issues/4925

The compaction number of rows was set too small, resulting in the process issuing too many concurrent fragment reservation commits at the same time and sometimes fail the test.


westonpace (approved):
Nice debugging and thanks for investigating the flaky test!

jackye1995:
ran CI 4 times all passing, should be good now

PR #5385, opened 2025-12-02
feat(blob_v2): add external blob support

Xuanwo:
This PR will add external blob support: users can now write an external blob in lance and read it back.

In this PR, I changed the blob struct's fields to be `nullable: false` because our packed struct doesn't allow nullable fields.

---

### DIFF
--- a/rust/lance/src/dataset/blob.rs
+++ b/rust/lance/src/dataset/blob.rs
@@ -11,7 +11,7 @@ use snafu::location;
 use tokio::sync::Mutex;
 
 use super::Dataset;
-use arrow_array::{Array, StructArray};
+use arrow_array::StructArray;
 use lance_core::datatypes::{BlobKind, BlobVersion};
 use lance_core::{utils::address::RowAddress, Error, Result};
 use lance_io::traits::Reader;
@@ -76,15 +76,16 @@ impl BlobFile {
 
     pub async fn new_external(
         uri: String,
-        size: Option<u64>,
+        size: u64,
         registry: Arc<ObjectStoreRegistry>,
         params: Arc<ObjectStoreParams>,
     ) -> Result<Self> {
         let (object_store, path) =
             ObjectStore::from_uri_and_params(registry, &uri, &params).await?;
-        let size = match size {
-            Some(sz) => sz,
-            None => object_store.size(&path).await?,
+        let size = if size > 0 {
+            size
+        } else {
+            object_store.size(&path).await?
         };
         Ok(Self {
             object_store,
@@ -290,20 +291,19 @@ async fn collect_blob_files_v2(
     let positions = descriptions.column(1).as_primitive::<UInt64Type>();
     let sizes = descriptions.column(2).as_primitive::<UInt64Type>();
     let _blob_ids = descriptions.column(3).as_primitive::<UInt32Type>();
-    let _uris = descriptions.column(4).as_string::<i32>();
+    let blob_uris = descriptions.column(4).as_string::<i32>();
 
     let mut files = Vec::with_capacity(row_addrs.len());
     for (idx, row_addr) in row_addrs.values().iter().enumerate() {
-        if kinds.is_null(idx) {
-            // Null row
+        let kind = BlobKind::try_from(kinds.value(idx))?;
+
+        // Struct is non-nullable; null rows are encoded as inline with zero position/size and empty uri
+        if matches!(kind, BlobKind::Inline) && positions.value(idx) == 0 && sizes.value(idx) == 0 {
             continue;
         }
-        let kind = BlobKind::try_from(kinds.value(idx))?;
+
         match kind {
             BlobKind::Inline => {
-                if positions.is_null(idx) || sizes.is_null(idx) {
-                    continue;
-                }
                 let position = positions.value(idx);
                 let size = sizes.value(idx);
                 files.push(BlobFile::new_inline(
@@ -315,17 +315,8 @@ async fn collect_blob_files_v2(
                 ));
             }
             BlobKind::External => {
-                let uri = _uris.value(idx).to_string();
-                let size = if sizes.is_null(idx) {
-                    None
-                } else {
-                    let value = sizes.value(idx);
-                    if value == 0 {
-                        None
-                    } else {
-                        Some(value)
-                    }
-                };
+                let uri = blob_uris.value(idx).to_string();
+                let size = sizes.value(idx);
                 let registry = dataset.session.store_registry();
                 let params = dataset
                     .store_params


rust/lance-encoding/src/encodings/logical/blob.rs
```diff
@@ -58,7 +58,7 @@ impl BlobStructuralEncoder {
 
         // Use the original field's name for the descriptor
         let descriptor_field = Field::try_from(
-            ArrowField::new(&field.name, descriptor_data_type, field.nullable)
+            ArrowField::new(&field.name, descriptor_data_type, false)
```
westonpace: Can you add a comment explaining why this will never be null?

rust/lance-core/src/datatypes/field.rs
```diff
@@ -1004,7 +1006,7 @@ impl TryFrom<&ArrowField> for Field {
         // Check for JSON extension types (both Arrow and Lance)
         let logical_type = if is_arrow_json_field(field) || is_json_field(field) {
             LogicalType::from("json")
-        } else if is_blob_v2 {
+        } else if is_blob_v2 && !matches!(field.data_type(), DataType::Struct(_)) {
```
westonpace: Why is this !matches needed?  What happens if a struct is labeled as blob v2?

rust/lance-encoding/src/encodings/logical/blob.rs
```diff
@@ -327,34 +326,41 @@ impl FieldEncoder for BlobV2StructuralEncoder {
...
-            };
-            if is_null_row {
-                kind_builder.append_null();
-                position_builder.append_null();
-                size_builder.append_null();
-                blob_id_builder.append_null();
-                uri_builder.append_null();
+        for i in 0..struct_arr.len() {
+            if struct_arr.is_null(i) {
+                // Packed struct does not support nullable fields; use empty/default values and rely on rep/def.
```
westonpace: Where are the repdef placed?  If I recall correctly we used to place them in the position or size?  Here it looks like you are setting both to zero?

rust/lance-encoding/src/encodings/logical/blob.rs
```diff
@@ -327,34 +326,41 @@ impl FieldEncoder for BlobV2StructuralEncoder {
...
-            let is_null_row = match array.data_type() {
-                DataType::Struct(_) => array.is_null(i),
-                _ => binary_array.is_null(i),
-            };
-            if is_null_row {
-                kind_builder.append_null();
-                position_builder.append_null();
-                size_builder.append_null();
-                blob_id_builder.append_null();
-                uri_builder.append_null();
```
westonpace, line 340: Are these changes forwards / backwards compatible?

rust/lance-encoding/src/encodings/logical/blob.rs
```diff
@@ -327,34 +326,41 @@ impl FieldEncoder for BlobV2StructuralEncoder {
...
             }
 
-            let value = binary_array.value(i);
-            kind_builder.append_value(0);
-
-            if value.is_empty() {
+            let data_is_set = !data_col.is_null(i);
+            if data_is_set {
+                let value = binary_array.value(i);
+                kind_builder.append_value(0);
```
westonpace: Can we use constants for kinds?  or an enum that converts to/from an integer?

rust/lance-encoding/src/encodings/logical/blob.rs
```diff
@@ -327,34 +326,41 @@ impl FieldEncoder for BlobV2StructuralEncoder {
...
+                    let position =
+                        external_buffers.add_buffer(LanceBuffer::from(Buffer::from(value)));
+                    position_builder.append_value(position);
+                    size_builder.append_value(value.len() as u64);
+                }
+                blob_id_builder.append_value(0);
+                uri_builder.append_value("");
+            } else {
+                // external uri
+                let uri = uri_col.value(i);
```
westonpace, line 358: Do we need to check if the URI is empty?

rust/lance-encoding/src/encodings/logical/blob.rs
```diff
@@ -327,34 +326,41 @@ impl FieldEncoder for BlobV2StructuralEncoder {
...
-                size_builder.append_null();
-                blob_id_builder.append_null();
-                uri_builder.append_null();
+        for i in 0..struct_arr.len() {
+            if struct_arr.is_null(i) {
+                // Packed struct does not support nullable fields; use empty/default values and rely on rep/def.
+                kind_builder.append_value(0);
+                position_builder.append_value(0);
+                size_builder.append_value(0);
+                blob_id_builder.append_value(0);
```
westonpace, line 336: It looks like blob_id is always set to zero.  Can you remind me what this is for again?

westonpace (commented):
Some initial questions

rust/lance-encoding/src/encodings/logical/blob.rs
Xuanwo: Actually, it's not, I shouldn't touch it.

rust/lance-encoding/src/encodings/logical/blob.rs
Xuanwo: The repdef here refers to the packed struct's own repdef. We won't maintain repdef for blob v2 structs to simplify. Might it's better to update the comments here to avoid confusion.

rust/lance-encoding/src/encodings/logical/blob.rs
Xuanwo: `blob_id == 0` means it's empty (the valid value starts from `1`).

And blob_id reprents the id used by packed & dedicated blobs which are both not implemented yet.

rust/lance-encoding/src/encodings/logical/blob.rs
Xuanwo: Yes, I believe so. No blob v2 files have been written so far. It's part of file format 2.2

rust/lance-encoding/src/encodings/logical/blob.rs
Xuanwo: Makes sense to me, will update.

rust/lance-encoding/src/encodings/logical/blob.rs
Xuanwo: Good idea, we can check before writing.

rust/lance-core/src/datatypes/field.rs
Xuanwo: I'm too defensive here, removed.

rust/lance-encoding/src/encodings/logical/blob.rs
westonpace: Ah, I had forgotten we added a dedicated encoded / decoder for v2.

rust/lance-encoding/src/encodings/logical/blob.rs
westonpace: That makes sense.  My brain was still thinking of the 2.1 encoder.

rust/lance-encoding/src/encodings/logical/blob.rs
westonpace: What is the `!data_is_set` branch then?  I thought an inline blob was "data column is not null and uri is null" and a dedicated / packed blob is "data column is null and uri is not null"?

westonpace: Ah, I see the enum now.  So the `!data_is_set` approach is external.

PR #5392, opened 2025-12-02
fix: take_blobs_by_indices fails with stable row IDs on fragment 1+

jmhsieh:
When `enable_stable_row_ids=true`, `take_blobs_by_indices` was failing with "index out of bounds" for rows in fragment 1+. The bug was caused by passing row addresses to `blob::take_blobs` which expected row IDs.

Root cause:
- `take_blobs_by_indices` converts indices to row addresses
- It passed these addresses to `take_blobs` which calls `take_builder`
- `TakeBuilder.get_row_addrs()` looked up the values in the row ID index
- For fragment 0: addresses (0,1,2) matched row IDs (0,1,2) by accident
- For fragment 1+: addresses (4294967296+) didn't match any row IDs
- This caused empty results and missing `_rowaddr` column → panic

Fix:
- Add `take_blobs_by_addresses()` that uses `TakeBuilder::try_new_from_addresses` to bypass the row ID index lookup
- Update `take_blobs_by_indices` to call the new function
- Add defensive fix in `do_take_rows` to include `_rowaddr` column in empty batches

### DIFF
--- a/rust/lance/src/dataset/blob.rs
+++ b/rust/lance/src/dataset/blob.rs
@@ -306,7 +306,6 @@ pub async fn take_blobs_by_addresses(
     }
 }
 
-
 fn collect_blob_files_v1(
     dataset: &Arc<Dataset>,
     blob_field_id: u32,


rust/lance/src/dataset.rs
```diff
@@ -1494,13 +1494,17 @@ impl Dataset {
 
     /// Take [BlobFile] by row indices.
     ///
+    /// This method converts row indices (offsets in the dataset) to row addresses,
```
jackye1995: these are internal details, we don't need these in public doc

rust/lance/src/dataset/blob.rs
```diff
@@ -209,6 +210,54 @@ pub(super) async fn take_blobs(
     }
 }
 
+/// Internal function for taking blobs by row addresses (not row IDs).
+///
+/// This is used by `take_blobs_by_indices` which already has row addresses
+/// computed from row indices. Using this function bypasses the row ID index
+/// lookup, which is necessary when `enable_stable_row_ids=true` because
+/// row addresses are not the same as row IDs in that case.
+pub(super) async fn take_blobs_by_addresses(
```
jackye1995: I think we should make this a public method which can be invoked by callers directly with python and java bindings. Originally we had take_blobs which takes row_ids and row_ids can be row address, or stable row ID if it is enabled. This assumes stable row ID is the next generation of row address that replaces its use cases completely. However, I think the situation has changed now, that we do want to use both stable row ID and row address depending on different use cases, so it makes sense now that we expose APIs for them respectively.

jackye1995:
@jmhsieh looks like there is a merge conflict, could you resolve?

jmhsieh:
> @jmhsieh looks like there is a merge conflict, could you resolve?

updated and giving a go

rust/lance/src/dataset.rs
jmhsieh: updated

PR #5406, opened 2025-12-04
feat(blob_v2): add dedicated blob support

Xuanwo:
This PR will add dedicated blob support in lance.

---

### DIFF
--- a/.github/workflows/build_linux_wheel/action.yml
+++ b/.github/workflows/build_linux_wheel/action.yml
@@ -25,6 +25,11 @@ runs:
       shell: bash
       run: |
         echo "ARM BUILD: ${{ inputs.arm-build }}"
+    - name: Clean old wheels
+      shell: bash
+      run: |
+        # Ensure no cached pylance wheels linger across cache restores
+        rm -f python/target/wheels/pylance-*.whl || true
     - name: Build x86_64 Manylinux2014 wheel
       if: ${{ inputs.arm-build == 'false' && inputs.manylinux == '2_17' }}
       uses: PyO3/maturin-action@v1
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,7 +50,7 @@ rust-version = "1.82.0"
 
 [workspace.dependencies]
 libc = "0.2.176"
-lance = { version = "=1.1.0-beta.2", path = "./rust/lance" }
+lance = { version = "=1.1.0-beta.2", path = "./rust/lance", default-features = false }
 lance-arrow = { version = "=1.1.0-beta.2", path = "./rust/lance-arrow" }
 lance-core = { version = "=1.1.0-beta.2", path = "./rust/lance-core" }
 lance-datafusion = { version = "=1.1.0-beta.2", path = "./rust/lance-datafusion" }
@@ -141,7 +141,7 @@ humantime = "2.2.0"
 hyperloglogplus = { version = "0.4.1", features = ["const-loop"] }
 itertools = "0.13"
 jieba-rs = { version = "0.8.1", default-features = false }
-jsonb = { version = "0.5.3", default-features = false, features = ["databend"]}
+jsonb = { version = "0.5.3", default-features = false, features = ["databend"] }
 libm = "0.2.15"
 log = "0.4"
 mockall = { version = "0.13.1" }
--- a/docs/src/guide/performance.md
+++ b/docs/src/guide/performance.md
@@ -165,3 +165,74 @@ across the entire process. This limit is specified by the `LANCE_PROCESS_IO_THRE
 The default is 128 which is more than enough for most workloads. You can increase this limit if you are working
 with a high-throughput workload. You can even disable this limit entirely by setting it to zero. Note that this
 can often lead to issues with excessive retries and timeouts from the object store.
+
+## Indexes
+
+Training and searching indexes can have unique requirements for compute and memory. This section provides some
+guidance on what can be expected for different index types.
+
+### BTree Index
+
+The BTree index is a two-level structure that provides efficient range queries and sorted access.
+It strikes a balance between an expensive memory structure containing all values and an expensive disk
+structure that can't be efficiently searched.
+
+Training a BTree index is done by sorting the column. This is done using an [external sort](https://en.wikipedia.org/wiki/External_sorting) to constrain the total memory usage to a reasonable amount. Updating a BTree index does not
+require re-sorting the entire column. The new values are sorted and the existing values are merged into the new sorted
+values in linear time.
+
+#### Storage Requirements
+
+The BTree index is essentially a sorted copy of a column. The storage requirements are therefore the same as the column
+but an additional 4 bytes per value is required to store the row ID and there is a small lookup structure which
+should be roughly 0.001% of the size of the column.
+
+#### Memory Requirements
+
+Training a BTree index requires some RAM but the current implementation spills to disk rather aggressively and so the
+total memory usage is fairly low.
+
+When searching a BTree index, the index is loaded into the index cache in pages. Each page contains 4096 values.
+
+#### Performance
+
+The sort stage is the most expensive step in training a BTree index. The time complexity is O(n log n) where n is the number of rows in the column. At very large scales this can be a bottleneck and a distributed sort may be necessary. Lance currently does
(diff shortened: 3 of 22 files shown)


rust/lance-core/src/utils/blob.rs
```diff
@@ -0,0 +1,33 @@
...
+// SPDX-FileCopyrightText: Copyright The Lance Authors
+
+use object_store::path::Path;
+
+/// Directory name for blob sidecar files.
+pub const BLOB_SIDECAR_DIR: &str = "_blob";
+
+/// Format a dedicated blob sidecar path.
+///
+/// Layout: `_blob/<fragment_id>/<field_id>/<blob_id>.raw`
```
Xuanwo: The detailed layout isn’t decided yet. I plan to use some bits from the low end of the blob ID to structure it like `01001010/<fragment_id>/<field_id>/<blob_id>.raw`

rust/lance/src/dataset/blob.rs
```diff
@@ -26,6 +32,214 @@ pub fn blob_version_from_config(config: &HashMap<String, String>) -> BlobVersion
         .unwrap_or(BlobVersion::V1)
 }
 
+const DEDICATED_THRESHOLD: usize = 4 * 1024 * 1024;
```
Xuanwo, line 36: random picked. Will update after packed blob been implemented.

rust/lance-core/src/utils/blob.rs
Xuanwo: Updated: now I change to use the data file key along with the datafile itself.

rust/lance/src/dataset/blob.rs
westonpace: It's a good pick :)

rust/lance/src/dataset/blob.rs
```diff
@@ -26,6 +32,197 @@ pub fn blob_version_from_config(config: &HashMap<String, String>) -> BlobVersion
...
+
+pub struct BlobPreprocessor {
```
westonpace, line 43: Document what this is?

rust/lance/src/dataset/blob.rs
```diff
@@ -26,6 +32,197 @@ pub fn blob_version_from_config(config: &HashMap<String, String>) -> BlobVersion
...
+            data_dir,
+            data_file_key,
+            // Start at 1 to avoid a potential all-zero blob_id value.
+            local_counter: 1,
+        }
+    }
+
+    fn next_blob_id(&mut self) -> u32 {
+        let id = self.local_counter;
+        self.local_counter = self.local_counter.wrapping_add(1);
```
westonpace: Instead of wrapping_add maybe checked_add and return an error?  Also, why not use u64 for counter?  Unlikely we will ever hit u32 limit but it is just barely conceivable and better to just not worry about it.

rust/lance-core/src/utils/blob.rs
```diff
@@ -0,0 +1,26 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright The Lance Authors
+
+use object_store::path::Path;
+
+/// Format a dedicated blob sidecar path for a data file.
+///
+/// Layout: `<base>/<data_file_key>/<blob_id>.raw`
+/// - `base` is typically the dataset's data directory.
+/// - `data_file_key` is the stem of the data file (without extension).
```
westonpace, line 10: Interesting.  I hadn't expected path to include `data_file_key` but I don't see any harm in it and it could maybe be useful for things like cleanup?

rust/lance/src/dataset/blob.rs
```diff
@@ -26,6 +32,197 @@ pub fn blob_version_from_config(config: &HashMap<String, String>) -> BlobVersion
...
+    pub(crate) async fn preprocess_batch(&mut self, batch: &RecordBatch) -> Result<RecordBatch> {
+        let mut new_columns = Vec::with_capacity(batch.num_columns());
+        let mut new_fields = Vec::with_capacity(batch.num_columns());
+
+        for (array, field) in batch.columns().iter().zip(batch.schema().fields()) {
+            let is_blob_struct = field
+                .metadata()
+                .get(lance_arrow::ARROW_EXT_NAME_KEY)
+                .map(|v| v == lance_arrow::BLOB_V2_EXT_NAME)
+                .unwrap_or(false);
```
westonpace: Can this be `field.is_blob_v2()`?

rust/lance/src/dataset/blob.rs
```diff
@@ -26,6 +32,197 @@ pub fn blob_version_from_config(config: &HashMap<String, String>) -> BlobVersion
...
+
+    async fn write_blob(&self, blob_id: u32, data: &[u8]) -> Result<Path> {
+        let path = blob_path(&self.data_dir, &self.data_file_key, blob_id);
+        let mut writer = self.object_store.create(&path).await?;
+        writer.write_all(data).await?;
+        writer.shutdown().await?;
+        Ok(path)
+    }
+
+    pub(crate) async fn preprocess_batch(&mut self, batch: &RecordBatch) -> Result<RecordBatch> {
```
westonpace, line 75: This requires blobs to be top-level fields.  E.g. we cannot have Struct<blob> or List<blob>.  I think this is fine for now but we should make sure we eventually document this somewhere.

rust/lance/src/dataset/blob.rs
```diff
@@ -26,6 +32,197 @@ pub fn blob_version_from_config(config: &HashMap<String, String>) -> BlobVersion
...
+            )?;
+
+            new_columns.push(Arc::new(struct_array));
+            new_fields.push(Arc::new(
+                arrow_schema::Field::new(
+                    field.name(),
+                    ArrowDataType::Struct(child_fields.into()),
+                    field.is_nullable(),
+                )
+                .with_metadata(field.metadata().clone()),
```
westonpace, line 194: Do we need to mark the field as packed?

rust/lance-encoding/src/encodings/logical/blob.rs
```diff
@@ -272,98 +272,92 @@ impl FieldEncoder for BlobV2StructuralEncoder {
...
-                continue;
-            }
 
-            let data_is_set = !data_col.is_null(i);
-            if data_is_set {
-                let value = binary_array.value(i);
-                kind_builder.append_value(BlobKind::Inline as u8);
-                if value.is_empty() {
-                    position_builder.append_value(0);
-                    size_builder.append_value(0);
```
westonpace, line 347: Very happy to see this logic move into table level :smile:

rust/lance-encoding/src/encodings/logical/blob.rs
```diff
@@ -272,98 +272,92 @@ impl FieldEncoder for BlobV2StructuralEncoder {
...
+                                .add_buffer(LanceBuffer::from(Buffer::from(data_val)));
+
+                            (
+                                BlobKind::Inline as u8,
+                                position,
+                                blob_len,
+                                0,
+                                "".to_string(),
+                            )
+                        }
```
westonpace, line 345: Awesome, so this method is now focused on the inline case?  Makes a lot of sense.

rust/lance/src/dataset/blob.rs
Xuanwo: In our current design, the blob ID is unique per data file (fragement + field), so it shares the same upper limit as the data file rows: u32::MAX. It should be safe to use u32 here.

I will change this place to just `+=1`.

rust/lance-core/src/utils/blob.rs
Xuanwo: Yes! I picked up this design to make the GC much easier: we can just remove all blob files under the same `data_file_key` once we decide to remove that data file.

rust/lance/src/dataset/blob.rs
Xuanwo: Oh yeah, nice catch. I have to admit I didn’t take enough consideration into this and I actually didn’t expect this. I’ll create a follow up.

rust/lance/src/dataset/blob.rs
Xuanwo: No, this record batch only exists in memory and we won’t really encode it.

PR #5408, opened 2025-12-04
docs: provide more specifics on critical fix definition

wjones127:
Added examples of critical and non-critical bug fixes to the CONTRIBUTING.md file.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -75,9 +75,15 @@ There are two important labels to apply to relevant issues and PRs:
   after merging a PR.
 2. `critical-fix`: Any PR that fixes a critical bug (e.g., security issue, data
   corruption, crash) should be labelled as such. These are bugs that users might
-  have without realizing. Fixes that aren't critical include bugs that return
-  an error message. These labels are used to determine whether a patch release
-  is needed.
+  have without realizing. Examples:
+
+    1. Security fix: a statically linked crypto library had a security vulnerability.
+    2. Incorrect data: a query would return 5 for 2 + 2.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -75,9 +75,15 @@ There are two important labels to apply to relevant issues and PRs:
   after merging a PR.
 2. `critical-fix`: Any PR that fixes a critical bug (e.g., security issue, data
```
jackye1995, line 76: One thing I have been thinking is that, we want this to be used by some code agent to automate things, so this technically should also be in AGENTS.md.

But after all, what is really the difference between a developer that is contributing, vs a code agent? Should we just make AGENTS.md and CLAUDE.md aliases of CONTRIBUTING.md, and have all the information in the same file?

CONTRIBUTING.md
```diff
@@ -75,9 +75,15 @@ There are two important labels to apply to relevant issues and PRs:
...
   corruption, crash) should be labelled as such. These are bugs that users might
-  have without realizing. Fixes that aren't critical include bugs that return
-  an error message. These labels are used to determine whether a patch release
-  is needed.
+  have without realizing. Examples:
+
+    1. Security fix: a statically linked crypto library had a security vulnerability.
+    2. Incorrect data: a query would return 5 for 2 + 2.
```
jackye1995, line 81: I think there is a difference between "incorrect query result" which is what is written here, vs "incorrect data" which is more like data loss or data corruption on disk, e.g. the writer wrote bad lance files that cannot be read back for some data type

PR #5409, opened 2025-12-04
fix: remove expensive clone in bitmap search

westonpace:
https://github.com/lance-format/lance/commit/82df34342938a0ce36b08c1412b2fdef0e604009 parallelizes bitmap loading but it introduced a clone of the bitmap index.  Since this was a clone of the index itself and not just an arc clone it ended up being a very expensive deep clone, especially at scale.  It also led to unbounded memory usage since all clones needed to live at the same time and this could cause the search to crash.

Fortunately, these clones are not needed, so we can simply remove them.


westonpace:
Tested with the microbenchmark `bitmap_in_5/string_unique/cached` which went from 1.3294s per iteration to 77.812µs per iteration (this was in debug mode).

PR #5410, opened 2025-12-05
feat(blob): add external blob support

shiyajuan123:


yanghua:
It seems the `BlobFile` has been refactored, and we may not need to do this work...

Xuanwo:
Closing as blob v2 is working on, read https://github.com/lance-format/lance/issues/4947 for more context.

PR #5412, opened 2025-12-05
fix: fix vector index prewarm index

### LINKED ISSUE #5411: Could not prewarm a vector index when using a custom index name
After creating a vector index with custom name which is `vector_index`, the following error message will be returned when using `dataset.prewarm_index('vector_index')`:
```
RuntimeError: LanceError(Index): Column vector_index does not exist in the schema, /Users/runner/work/lance/lance/rust/lance/src/index.rs:1155:31
```

This is my reproduce code:
```
import random
import string

import lance
import numpy as np
from lance.vector import vec_to_table
import pyarrow as pa


def create_table(nvec=1000, ndim=128, nans=0, nullify=False, dtype=np.float32):
    mat = np.random.randn(nvec, ndim)
    if nans > 0:
        nans_mat = np.empty((nans, ndim))
        nans_mat[:] = np.nan
        mat = np.concatenate((mat, nans_mat), axis=0)
    mat = mat.astype(dtype)
    price = np.random.rand(nvec + nans) * 100

    def gen_str(n):
        return "".join(random.choices(string.ascii_letters + string.digits, k=n))

    meta = np.array([gen_str(100) for _ in range(nvec + nans)])
    tbl = (
        vec_to_table(data=mat)
        .append_column("price", pa.array(price))
        .append_column("meta", pa.array(meta))
        .append_column("id", pa.array(range(nvec + nans)))
    )
    if nullify:
        idx = tbl.schema.get_field_index("vector")
        vecs = tbl[idx].to_pylist()
        nullified = [vec if i % 2 == 0 else None for i, vec in enumerate(vecs)]
        field = tbl.schema.field(idx)
        vecs = pa.array(nullified, field.type)
        tbl = tbl.set_column(idx, field, vecs)
    return tbl


if __name__ == "__main__":
    tbl = create_table()
    tmp_path = "/tmp/test_dataset.lance"
    dataset = lance.write_dataset(tbl, tmp_path, data_storage_version="2.1")
    dataset = dataset.create_index(
        "vector", name="vector_index", index_type="IVF_PQ", num_partitions=4, num_sub_vectors=16
    )
    # Prewarm the index
    dataset.prewarm_index("vector_index")
```

xloya:
Close #5411. First, try to retrieve the field from the schema. If it cannot be retrieved, then retrieve it directly from the fields of `index_meta`.

### DIFF
--- a/.bumpversion.toml
+++ b/.bumpversion.toml
@@ -1,5 +1,5 @@
 [tool.bumpversion]
-current_version = "1.1.0-beta.0"
+current_version = "1.1.0-beta.2"
 parse = "(?P<major>\\d+)\\.(?P<minor>\\d+)\\.(?P<patch>\\d+)(-(?P<prerelease>(beta|rc))\\.(?P<prerelease_num>\\d+))?"
 serialize = [
     "{major}.{minor}.{patch}-{prerelease}.{prerelease_num}",
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -9,6 +9,11 @@ debug = true
 codegen-units = 16
 lto = "thin"
 
+[profile.bench]
+inherits = "release"
+lto = "thin"
+codegen-units = 16
+
 [target.x86_64-unknown-linux-gnu]
 rustflags = ["-C", "target-cpu=haswell", "-C", "target-feature=+avx2,+fma,+f16c"]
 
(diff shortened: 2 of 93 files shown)


xloya:
> Thank you for working on this!
> 
> I took a quick look and the fix seems correct, but there’s an issue with index access inside nested structs. I wonder if we can improve this.
> 
> Let’s build an API called `resolve_index_column(idx_meta, column_arg, schema) -> Result<(String, FieldRef)>`, where the `String` is the full path of the given index.
> 
> The full path can be used in all following API calls like `get_vector_type`.
> 
> This API could be used across all vector, scalar, and generic open/prewarm operations.
> 
> What do you think?

That sounds like a good idea, let me try to improve it.

xloya:
@Xuanwo Address the comment, please take another look when you have time, thanks!

PR #5413, opened 2025-12-05
feat(blob_v2): add packed blob support

Xuanwo:
This PR will add packed blob support for lance.

This PR is based on https://github.com/lance-format/lance/pull/5406, please review that first.

---

### DIFF
--- a/rust/lance/src/dataset/blob.rs
+++ b/rust/lance/src/dataset/blob.rs
@@ -169,7 +169,7 @@ impl BlobPreprocessor {
             .write_with_allocator(
                 &mut || {
                     let id = *counter;
-                    *counter = counter.wrapping_add(1);
+                    *counter += 1;
                     id
                 },
                 data,
@@ -243,7 +243,7 @@ impl BlobPreprocessor {
                     uri_builder.append_null();
                     blob_id_builder.append_value(blob_id);
                     blob_size_builder.append_value(data_len as u64);
-                    position_builder.append_value(0);
+                    position_builder.append_null();
                     continue;
                 }
 


rust/lance/src/dataset/blob.rs
```diff
@@ -33,7 +33,83 @@ pub fn blob_version_from_config(config: &HashMap<String, String>) -> BlobVersion
...
+        self.current_blob_id = Some(blob_id);
+        self.current_size = 0;
+        Ok(())
+    }
+
+    async fn write_with_allocator<F>(
+        &mut self,
+        alloc_blob_id: &mut F,
+        data: &[u8],
+    ) -> Result<(u32, u64)>
```
westonpace, line 90: Can we put at least a `//` comment on this method that explains what is getting returned here?  Looks like the blob_id and position of the data in the pack?

rust/lance/src/dataset/blob.rs
```diff
@@ -64,14 +147,27 @@ impl BlobPreprocessor {
         id
     }
 
-    async fn write_blob(&self, blob_id: u32, data: &[u8]) -> Result<Path> {
-        let path = blob_path(&self.data_dir, &self.data_file_key, blob_id);
+    async fn write_blob(&mut self, blob_id: u32, data: &[u8]) -> Result<Path> {
```
westonpace: Should we rename this method to `write_dedicated`?

rust/lance/src/dataset/blob.rs
```diff
@@ -64,14 +147,27 @@ impl BlobPreprocessor {
...
         Ok(path)
     }
 
+    async fn write_packed(&mut self, data: &[u8]) -> Result<(u32, u64)> {
+        let (counter, pack_writer) = (&mut self.local_counter, &mut self.pack_writer);
+        pack_writer
+            .write_with_allocator(
+                &mut || {
+                    let id = *counter;
+                    *counter = counter.wrapping_add(1);
```
westonpace: Again, not likely, but let's not wrap here (better to panic / error)

rust/lance/src/dataset/blob.rs
```diff
@@ -120,21 +218,36 @@ impl BlobPreprocessor {
...
                     let blob_id = self.next_blob_id();
                     self.write_blob(blob_id, data_col.value(i)).await?;
 
                     kind_builder.append_value(BlobKind::Dedicated as u8);
                     data_builder.append_null();
                     uri_builder.append_null();
                     blob_id_builder.append_value(blob_id);
-                    blob_size_builder.append_value(data_col.value(i).len() as u64);
+                    blob_size_builder.append_value(data_len as u64);
+                    position_builder.append_value(0);
```
westonpace: Not a big deal, but curious.  Why not append_null here?

westonpace (approved):
Great work as always!

rust/lance/src/dataset/blob.rs
Xuanwo: Great idea

rust/lance/src/dataset/blob.rs
Xuanwo: Aha, packed struct doesn't support null yet.

rust/lance/src/dataset/blob.rs
Xuanwo: Will do

rust/lance/src/dataset/blob.rs
westonpace: Aren't you appending null in other branches?

westonpace: https://github.com/lance-format/lance/pull/5413/files#diff-2c7a237ff0cf0b81c15bf58fa9e0a0cf89ca55cbebbcd36d80879bd1a9c63fc7R272

rust/lance/src/dataset/blob.rs
Xuanwo: added.

rust/lance/src/dataset/blob.rs
Xuanwo: Oh, sorry for the mistake. I misunderstood the place this comment referred to. Yes, it's fine to just use append_null for the position of the dedicated blob.

Xuanwo: Will update

PR #5415, opened 2025-12-05
feat: allow python tracing / logging to be independently configured

westonpace:


westonpace:
CC @yanghua as I believe you have used this python tracing feature in the past.  I don't know if this changes your workflow much (I think you were primarily interested in tracing events and since the default tracing level is now info in this PR it shouldn't hide any tracing events) but I wanted to double check.

PR #5426, opened 2025-12-07
chore: show agents how to view test code coverage

wjones127:
Providing some scripts and instructions for how to view coverage. This is usable by coding agents, which should enable them to help us improve our test coverage.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -92,7 +92,7 @@ jobs:
       - name: Run tests
         run: |
           ALL_FEATURES=`cargo metadata --format-version=1 --no-deps | jq -r '.packages[] | .features | keys | .[]' | grep -v protoc | sort | uniq | paste -s -d "," -`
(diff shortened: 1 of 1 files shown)


AGENTS.md
```diff
@@ -135,6 +139,8 @@ Tests:
   /// # }
   /// ```
   ```
+* Code coverage can be skipped for test utilities and non-critical paths using
```
westonpace, line 142: Is there any way to just skip code coverage for anything in the test cfg?  Or is that already done and this is for test utilities that are in main scope for some reason?

wjones127: Doesn't seem like there's a particularly easy way. https://github.com/taiki-e/cargo-llvm-cov/issues/123

But most of the time tests have their lines covered. It's just some edge cases like branches that only happen on test failures that get missed. I think that's okay.

westonpace (approved):

PR #5434, opened 2025-12-08
feat(compaction): binary copy capability for compaction

zhangyue19921010:
Closes: https://github.com/lance-format/lance/issues/5433

### DIFF
--- a/rust/lance/src/dataset/optimize/tests/binary_copy.rs
+++ b/rust/lance/src/dataset/optimize/tests/binary_copy.rs
@@ -693,7 +693,7 @@ async fn do_test_binary_copy_compaction_with_complex_schema(version: LanceFileVe
         .col("i64", array::step::<Int64Type>())
         .col("f32", array::rand::<Float32Type>())
         .col("f64", array::rand::<Float64Type>())
-        .col("bool", array::rand_boolean())
+        .col("bool", array::cycle_bool(vec![false, true]))
         .col("date32", array::rand_date32())
         .col("date64", array::rand_date64())
         .col(


rust/lance-datagen/src/generator.rs
```diff
@@ -2518,7 +2519,7 @@ pub mod array {
                 ))
             }
             DataType::Timestamp(TimeUnit::Millisecond, _) => {
-                Box::new(FnGen::<i64, TimestampMicrosecondArray, _>::new_known_size(
+                Box::new(FnGen::<i64, TimestampMillisecondArray, _>::new_known_size(
```
westonpace: Good catch :smile:

rust/lance-file/src/writer.rs
```diff
@@ -485,6 +485,18 @@ impl FileWriter {
...
+        &mut self,
+        schema: lance_core::datatypes::Schema,
+        column_metadata: Vec<pbfile::ColumnMetadata>,
+        rows_written: u64,
+    ) {
+        self.schema = Some(schema);
+        self.num_columns = column_metadata.len() as u32;
+        self.column_metadata = column_metadata;
+        self.rows_written = rows_written;
+    }
```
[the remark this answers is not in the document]
westonpace: I agree with this comment, it would be nice to have some docs on this method.

rust/lance-file/src/writer.rs
```diff
@@ -585,7 +597,9 @@ impl FileWriter {
             .collect::<FuturesOrdered<_>>();
         self.write_pages(encoding_tasks).await?;
 
-        self.finish_writers().await?;
+        if !self.column_writers.is_empty() {
+            self.finish_writers().await?;
+        }
```
westonpace, line 620: Why is this change needed?

rust/lance-io/src/object_store/providers/aws.rs
```diff
@@ -214,10 +214,18 @@ async fn resolve_s3_region(
...
+            Err(e) => {
+                log::debug!(
+                    "Failed to resolve S3 bucket region for '{}': {:?}; defaulting to provider chain",
+                    bucket,
+                    e
+                );
+                // Fallback to region provider chain; let downstream choose a default
+                Ok(None)
+            }
+        }
```
westonpace: This change seems unrelated (but is fairly minor).  Will let @jackye1995 or @wjones127 weigh in on whether this is the best thing to do here or not.

rust/lance/src/dataset/optimize.rs
```diff
@@ -183,6 +208,95 @@ impl CompactionOptions {
...
+        return false;
+    }
+
+    // Establish version baseline from first data file
+    let first_data_file_version = LanceFileVersion::try_from_major_minor(
+        fragments[0].files[0].file_major_version,
+        fragments[0].files[0].file_minor_version,
+    )
+    .map(|v| v.resolve())
+    .unwrap();
```
westonpace: I think you can just use `dataset.manifest.data_storage_format` like you check above.

rust/lance/src/dataset/optimize.rs
```diff
@@ -226,6 +340,13 @@ pub async fn compact_files(
 
     let compaction_plan: CompactionPlan = plan_compaction(dataset, &options).await?;
 
+    if compaction_plan.tasks().is_empty() && options.enable_binary_copy_force {
```
westonpace: Why is this an error instead of a no-op?

rust/lance/src/dataset/optimize.rs
```diff
@@ -672,38 +851,43 @@ async fn rewrite_files(
...
+            &fragments,
+            options.batch_size,
+            true,
+            needs_remapping,
+        )
+        .await?
     } else {
-        let data = SendableRecordBatchStream::from(scanner.try_into_stream().await?);
-        (None, data)
+        prepare_reader(
```
westonpace: Why are you preparing a reader in this case?  If we can use binary copy we shouldn't need a scan right?

rust/lance/src/dataset/optimize.rs
```diff
@@ -732,16 +916,73 @@ async fn rewrite_files(
...
+        .await?;
+
+        if new_fragments.is_empty() && options.enable_binary_copy_force {
+            return Err(Error::NotSupported {
+                source: format!("compaction task {}: binary copy is not supported", task_id).into(),
+                location: location!(),
+            });
+        }
+
+        if new_fragments.is_empty() {
```
westonpace: Why would this be empty?  Didn't we already verify that binary copy is supported if we reach this point?

rust/lance/src/dataset/optimize.rs
```diff
@@ -973,6 +1214,515 @@ async fn recalc_versions_for_rewritten_fragments(
     Ok(())
 }
 
+async fn rewrite_files_binary_copy(
```
westonpace: This (and `flush_footer`) are big methods. We might want some kind of `BinaryCopier` utility struct instead?  It could be in its own sub-module `e.g. lance::dataset::optimize::binary_copy`

rust/lance-file/src/writer.rs
zhangyue19921010: In Binary Copy, we pre-write data and metadata in a copy manner without calling FileWriter. However, when flushing the footer, in order to reuse existing code as much as possible, we will try to mock a file writer and call its finish method to trigger the writing of the footer. Therefore, a check is needed here to skip the execution of finish_writers in scenarios similar to Binary copy (where column_writers is empty).

rust/lance-io/src/object_store/providers/aws.rs
zhangyue19921010: Not this PR related. Removed.

rust/lance/src/dataset/optimize.rs
zhangyue19921010: Actually, it's not needed. This logic has already been removed.

rust/lance/src/dataset/optimize.rs
zhangyue19921010: This reader is indeed not needed during the Binary copy process. In the design of binary copy, there is a feature: if any problem causes a panic during the rewrite phase of binary copy, it will roll back to the normal compaction logic. Therefore, this object was initialized in advance. However, after careful consideration, this design is somewhat cumbersome and has been simplified.

The logic for initializing the reader in the scenario of Binary copy has been removed.

rust/lance/src/dataset/optimize.rs
zhangyue19921010: As mentioned before. Removed this logic

rust/lance/src/dataset/optimize.rs
zhangyue19921010: changed

rust/lance/src/dataset/optimize.rs
zhangyue19921010: done

zhangyue19921010:
Hi @westonpace Thanks a lot for your review. All comments are addressed. PTAL :)

zhangyue19921010:
Hi @westonpace  — hope you’re doing well ! 

Just a gentle ping on this PR. Your review would be really helpful. When you have time, could you take another look? 

Appreciate your help!

zhangyue19921010:
Hi @jackye1995 would u mind to take a look at your convenience? Really appreciate if there are any feedback. Thanks in advance!

rust/lance/src/dataset/optimize/binary_copy.rs
```diff
@@ -0,0 +1,559 @@
...
+            current_filename = Some(filename);
+        }
+        let writer = current_writer.take().unwrap();
+        flush_footer(writer, &schema, &final_cols, total_rows_in_current, version).await?;
+        // Register the final file
+        let (maj, min) = version.to_numbers();
+        let mut frag = Fragment::new(0);
+        let mut df = DataFile::new_unstarted(current_filename.take().unwrap(), maj, min);
+        // v2_0 vs v2_1+ field-to-column index mapping for the final file
+        let is_structural = version >= LanceFileVersion::V2_1;
```
jackye1995: this is duplicated logic with L378

rust/lance/src/dataset/optimize/binary_copy.rs
```diff
@@ -0,0 +1,559 @@
...
+                }
+                out.push(fragment_out);
+                total_rows_in_current = 0;
+            }
+        }
+    } // Finished writing all fragments; any remaining data in memory will be flushed below
+
+    if total_rows_in_current > 0 {
+        // Flush remaining rows as a final output file
+        // v2_0 compatibility: same single-page enforcement applies for the final file close
```
jackye1995: this is duplicate logic with L346

rust/lance/src/dataset/optimize.rs
```diff
@@ -785,16 +947,51 @@ async fn rewrite_files(
...
+        if needs_remapping {
+            let (tx, rx) = std::sync::mpsc::channel();
+            let mut addrs = RoaringTreemap::new();
+            for frag in &fragments {
+                let frag_id = frag.id as u32;
+                let count = frag.physical_rows.unwrap_or(0);
+                for i in 0..count {
+                    let addr =
+                        lance_core::utils::address::RowAddress::new_from_parts(frag_id, i as u32);
+                    addrs.insert(u64::from(addr));
```
jackye1995: this would be pretty inefficient, can we use `insert_range`

rust/lance/src/dataset/optimize/binary_copy.rs
```diff
@@ -0,0 +1,559 @@
...
+///   only performs consistent finalization.
+async fn flush_footer(
+    mut writer: ObjectWriter,
+    schema: &Schema,
+    final_cols: &[Arc<ColumnInfo>],
+    total_rows_in_current: u64,
+    version: LanceFileVersion,
+) -> Result<()> {
+    let pos = writer.tell().await? as u64;
+    let _new_pos = apply_alignment_padding(&mut writer, pos, version).await?;
```
jackye1995, line 511: why result is discarded?

rust/lance/src/dataset/optimize.rs
```diff
@@ -183,6 +200,93 @@ impl CompactionOptions {
...
+        return false;
+    }
+
+    // Establish version baseline from the dataset manifest
+    let storage_file_version = match dataset.manifest.data_storage_format.lance_file_version() {
+        Ok(version) => version.resolve(),
+        Err(_) => return false,
+    };
+    // Capture schema mapping baseline from first data file
+    let ref_fields = &fragments[0].files[0].fields;
```
jackye1995, line 285: edge case: to be safe, should check `!fragments[0].files.is_empty()`

rust/lance/src/dataset/optimize/binary_copy.rs
```diff
@@ -0,0 +1,559 @@
...
+/// - Enforces 64-byte alignment for page and buffer writes in V2.1+ files (V2.0 does not require alignment).
+/// - For v2.0, preserves single-page structural headers and normalizes their row counts/priority.
+/// - Flushes an output file once `max_rows_per_file` rows are accumulated, then repeats.
+///
+/// Parameters:
+/// - `dataset`: target dataset (for storage/config and schema).
+/// - `fragments`: fragments to merge via binary copy (assumed consistent versions).
+/// - `params`: write parameters (uses `max_rows_per_file`).
+/// - `read_batch_bytes_opt`: optional I/O batch size when coalescing page reads.
+pub async fn rewrite_files_binary_copy(
```
jackye1995, line 173: as we discussed in original github issue, we should reject the binary copy if the Lance file has global buffer. we can do that check after reading file_meta.

zhangyue19921010: Nice Catch. Add Global buffer checking in `can_use_binary_copy` function.

rust/lance/src/dataset/optimize/binary_copy.rs
zhangyue19921010: 1. `apply_alignment_padding` may write pad zero bytes via `writer.write_all(...)` and thereby advances the writer’s internal position
2. The u64 (new_pos) it returns is just `the theoretically new position after writing the padding`. **This value is not needed for subsequent calculations in flush_footer** (FileWriter::finish() will directly start writing the footer from the current position of the writer), so the return value is unused.
https://github.com/lance-format/lance/blob/e7540d7ce489ac2bb5585ea37bda3b0b312b3509/rust/lance-file/src/writer.rs#L605

zhangyue19921010:
Hi @jackye1995 .  Just a gentle reminder — could you please take another look when it’s convenient? Thanks!

rust/lance/src/dataset/optimize.rs
```diff
@@ -156,6 +159,17 @@ pub struct CompactionOptions {
     /// not be remapped during this compaction operation. Instead, the fragment reuse index
     /// is updated and will be used to perform remapping later.
     pub defer_index_remap: bool,
+    /// Whether to enable binary copy optimization when eligible.
+    /// Defaults to false.
+    pub enable_binary_copy: bool,
```
westonpace, line 170:
```suggestion
    /// Whether to enable binary copy optimization when eligible.
    ///
    /// This skips re-encoding the data and can lead to faster compaction
    /// times.  However, it cannot merge pages together and should not be
    /// used when compacting small files together because the pages in the
    /// compacted file will be too small and this could lead to poor I/O patterns.
    ///
    /// Defaults to false.
    pub enable_binary_copy: bool,
```

rust/lance/src/dataset/optimize.rs
```diff
@@ -183,6 +200,132 @@ impl CompactionOptions {
...
+    if fragments[0].files.is_empty() {
+        return Ok(false);
+    }
+    let ref_fields = &fragments[0].files[0].fields;
+    let ref_cols = &fragments[0].files[0].column_indices;
+    let mut is_same_version = true;
+
+    for fragment in fragments {
+        if fragment.deletion_file.is_some() {
+            return Ok(false);
```
westonpace, line 295: Whenever we return false it might be nice to log a debug message explaining why

rust/lance/src/dataset/optimize.rs
```diff
@@ -183,6 +200,132 @@ impl CompactionOptions {
...
+                .await?;
+            let file_meta = LFReader::read_all_metadata(&file_scheduler).await?;
+            // Binary copy only preserves page and column-buffer bytes. The output file's footer
+            // (including global buffers) is re-generated, not copied from inputs.
+            //
+            // Therefore, we reject input files that contain any additional global buffers beyond
+            // the required schema / file descriptor global buffer (global buffer index 0).
+            if file_meta.file_buffers.len() > 1 {
+                return Ok(false);
+            }
```
westonpace, line 340: At some point I think we are going to start writing file stats in the footer which might interfere with this check.

westonpace (approved):
I'm fine moving forwards with this.  I'll give a bit for @jackye1995 to weigh in or any comments to be addressed.

I think this is something of a niche use case because we want the underlying data pages to be large but there could be some cases where it is useful (compacting large files into huge files, or search-only use cases where everything is random access and we don't care as much about page size).

It doesn't add too much complexity (mainly some if/else checks in the compaction code) and the rest is hidden in a dedicated module.

zhangyue19921010:
Hi @westonpace Thanks a lot for your help. All comments are all addressed and CI passed :) cc @jackye1995

yanghua:
Hi @jackye1995 , any concerns?

PR #5439, opened 2025-12-08
docs: fix Append call in distributed write guide

rongou:
The `Append` operation no longer needs a schema.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -63,7 +63,7 @@ jobs:
           sudo apt install -y protobuf-compiler libssl-dev
       - name: Get features
         run: |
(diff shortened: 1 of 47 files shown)


docs/src/guide/distributed_write.md
```diff
@@ -104,7 +104,7 @@ import lance
 ds = lance.dataset(data_uri)
 read_version = ds.version # record the read version
 
-op = lance.LanceOperation.Append(schema, all_fragments)
+op = lance.LanceOperation.Append(all_fragments)
```
HaochengLIU, line 107: Confirmed it's an oversight when the API is updated
https://lance-format.github.io/lance-python-doc/all-modules.html#lance.LanceOperation.Append

HaochengLIU:
@rongou LGTM. Thanks for the contribution!

PR #5444, opened 2025-12-09
feat: order by column should be eagerly materialized

kaori-seasons:
Related to [issue-5290](https://github.com/lance-format/lance/issues/5290)


kaori-seasons:
@westonpace Hello, I apologize for the late reply.  Could you please take a look at this pull request when you have a chance?

PR #5446, opened 2025-12-09
perf: various btree performance improvements

westonpace:
The main improvements are:

* Sort page by row ids when loading
* Don't search page if the entire page matches, instead use a precomputed "all_ids" bitmap

### DIFF
--- a/rust/lance-index/src/scalar/btree.rs
+++ b/rust/lance-index/src/scalar/btree.rs
@@ -943,7 +943,11 @@ impl BTreeIndex {
                 // to search for X IN [5, 3]
                 subindex.search(query, metrics)
             }
-            Matches::All(_) => Ok(subindex.all()),
+            Matches::All(_) => Ok(match query {
+                // This means we hit an all-null page so just grab all row ids as true
+                SargableQuery::IsNull() => subindex.all_ignore_nulls(),
+                _ => subindex.all(),
+            }),
         }
     }
 
--- a/rust/lance-index/src/scalar/btree/flat.rs
+++ b/rust/lance-index/src/scalar/btree/flat.rs
@@ -4,22 +4,27 @@
 use std::collections::HashMap;
 use std::{ops::Bound, sync::Arc};
 
+use arrow_array::Array;
 use arrow_array::{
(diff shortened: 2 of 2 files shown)


westonpace:
Other potential improvements:

* Sort page by row ids on write (avoid the minor cold search penalty but make a backwards incompatible change to the index)
* Calculate pre-computed ids for ranges of pages

westonpace:
This drops the `btree_range_most/int_unique/cached` benchmark on my system from 19.5ms to 1.5ms

rust/lance-index/src/scalar/btree.rs
```diff
@@ -662,25 +687,86 @@ impl BTreeLookup {
...
+            }
         }
+
+        matches
     }
 
-    fn pages_null(&self) -> Vec<u32> {
-        self.null_pages.clone()
+    fn pages_null(&self) -> Vec<Matches> {
+        // TODO: We could keep track of all-null pages and return Matches::All for those.
```
BubbleCal: why we can't do this for now?

rust/lance-index/src/scalar/btree/flat.rs
```diff
@@ -46,167 +40,69 @@ impl DeepSizeOf for FlatIndex {
 }
 
 impl FlatIndex {
+    pub fn try_new(data: RecordBatch) -> Result<Self> {
+        // Sort by row id to make bitmap construction more efficient
+        let data = data.sort_by_column(1, None)?;
```
BubbleCal: isn't this already sorted?

rust/lance-index/src/scalar/btree.rs
westonpace: We can, I just didn't want to :laughing:.  Let me add that.

rust/lance-index/src/scalar/btree/flat.rs
westonpace: It's sorted by value.  This sorts it by row id.

rust/lance-index/src/scalar/btree.rs
westonpace: Done.

PR #5447, opened 2025-12-09
feat: add additional index APIs to support count rows split plan

jackye1995:
Related to https://github.com/lance-format/lance-spark/pull/140


java/src/main/java/org/lance/Dataset.java
```diff
@@ -928,6 +953,20 @@ public List<String> listIndexes() {
 
   private native List<String> nativeListIndexes();
 
+  /**
+   * Get all indexes with full metadata.
+   *
+   * @return list of Index objects with complete metadata including index type and fragment coverage
+   */
+  public List<org.lance.index.Index> getIndexes() {
```
majin1102: Maybe we could use the short name and import it

java/src/main/java/org/lance/Dataset.java
```diff
@@ -928,6 +953,20 @@ public List<String> listIndexes() {
...
+    try (LockManager.ReadLock readLock = lockManager.acquireReadLock()) {
+      Preconditions.checkArgument(nativeDatasetHandle != 0, "Dataset is closed");
+      return nativeGetIndexes();
+    }
+  }
+
+  private native List<org.lance.index.Index> nativeGetIndexes();
```
majin1102: same as above

java/lance-jni/src/blocking_dataset.rs
```diff
@@ -2445,3 +2445,313 @@ fn inner_cleanup_with_policy<'local>(
...
+            "add",
+            "(Ljava/lang/Object;)Z",
+            &[JValue::Object(&java_index)],
+        )?;
+    }
+
+    Ok(array_list)
+}
+
+fn create_java_index<'local>(
```
majin1102: I guess this is duplicated with https://github.com/majin1102/lance/blob/772c6d87a5b1d9530b5f3772fba7c33cf5c55b3a/java/lance-jni/src/transaction.rs#L81-L170

Maybe we should move that code here and add this index_type

java/src/main/java/org/lance/index/Index.java
```diff
@@ -166,6 +188,35 @@ public static Builder builder() {
     return new Builder();
   }
 
+  /**
+   * Create an Index instance directly. This is primarily for JNI use.
+   *
+   * @return a new Index instance
+   */
+  public static Index create(
```
majin1102: I wonder this should be a "create". It doesn't look like not a create index but only generate an object.

I think we could just use the private constructor in JNI?

PR #5449, opened 2025-12-10
refactor: use the same path for dedicated and packed blob

Xuanwo:
I made this change to make it easier for us to perform compaction or GC, as all blob IDs will now refer to the same blob paths. This means that as long as we know the largest blob IDs, we can simply remove them all at once.

---

### DIFF
--- a/rust/lance-core/src/utils/blob.rs
+++ b/rust/lance-core/src/utils/blob.rs
@@ -3,38 +3,25 @@
 
 use object_store::path::Path;
 
-/// Format a dedicated blob sidecar path for a data file.
+/// Format a blob sidecar path for a data file.
 ///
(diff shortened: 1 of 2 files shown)


Xuanwo:
> How do you envision compaction affecting blob files? Would small packed blob files (e.g. a few rows of 1MB per row so each file is only a few MBs) be compacted into a single packed file? Or would we never compact blob flies?

Yes, we will combine small packed files into a larger one when appropriate.

> I ask because compacting packed blob files seems like it would be an ok thing to do but packing dedicated blob files would make them no longer dedicated (not sure if that would be a problem or not)

Yeah, that's actually a beautiful thing about the current design. That is, we CAN compact dedicated blobs into packed blobs too if we want!

But we don't necessarily need to do this since we can expect that dedicated blobs are always large enough that they don't need to be packed.

PR #5450, opened 2025-12-10
feat(java): support multi-bases for writing database

ddupg:


ddupg:
Please @jaystarshot @jackye1995 help review this

java/lance-jni/Cargo.lock
```diff
@@ -3417,6 +3417,7 @@ dependencies = [
  "arrow-buffer",
  "arrow-cast",
  "arrow-data",
+ "arrow-ord",
```
majin1102, line 3420: Is this unnecessary?

java/lance-jni/src/utils.rs
```diff
@@ -92,6 +95,35 @@ pub fn extract_write_params(
         .map(|v| std::time::Duration::from_secs(v as u64))
         .unwrap_or_else(|| std::time::Duration::from_secs(10));
 
+    env.get_optional(initial_bases, |env, opt_obj| {
```
majin1102: I think this block works well.
I wonder if we could use import_vec and make an FromJObjectWithEnv impl for BasePath which is easy to reuse in other cases. there's also a IntoJava for transforming. But I guess we don't need that yet.

java/src/test/java/org/lance/MultiBaseTest.java
```diff
@@ -0,0 +1,256 @@
...
+        writer.start();
+        writer.writeBatch();
+        writer.end();
+      }
+      return new ArrowStreamReader(new ByteArrayInputStream(out.toByteArray()), allocator);
+    }
+  }
+
+  @Test
+  public void testCreateAndRead() throws Exception {
```
majin1102: Wondering if `testCreateMode` is better

java/src/test/java/org/lance/MultiBaseTest.java
```diff
@@ -0,0 +1,256 @@
...
+            .mode(WriteParams.WriteMode.OVERWRITE)
+            .targetBases(Arrays.asList("base2"))
+            .maxRowsPerFile(75)
+            .execute();
+
+    assertEquals(150, updated.countRows());
+  }
+
+  @Test
+  public void testIsDatasetRootFlag() throws Exception {
```
majin1102: Wondering if we could merge testIsDatasetRootFlag and testTargetByPathUri into the testOperation pattern tests?

java/lance-jni/Cargo.lock
ddupg: The previous PR updated the Cargo.lock files in rust and python, but omitted java. When I executed the `cargo build` command, it triggered an automatic update.

Java Cargo.lock in main branch has now been updated, and I have rebased to solve this.

java/lance-jni/src/utils.rs
ddupg: Good suggestion

java/src/test/java/org/lance/MultiBaseTest.java
ddupg: The content of these two tests overlaps, I have merged them.

majin1102:
> looks good to me, @majin1102 any further comments?

good to go

PR #5451, opened 2025-12-10
feat(blob_v2): add BlobAray API for user input

Xuanwo:
This PR will add BlobAray API for user input so that they don't need to play with metadata by themselves.

---


rust/lance-core/src/datatypes/field.rs
```diff
@@ -516,6 +516,23 @@ impl Field {
             .unwrap_or(false)
     }
 
+    /// If the field is a blob, return a new field with the same name and id
```
westonpace: This comment says "return a new field" but it is modifying the existing field.  Also, if it is going to modify the existing field, does it need to return self?  That introduces an awkward `let _ = ...` below.

rust/lance-core/src/datatypes/field.rs
```diff
@@ -744,6 +761,13 @@ impl Field {
             (&self_type, &other_type),
             (DataType::Struct(_), DataType::Struct(_)) | (DataType::List(_), DataType::List(_))
         ) {
+            // Blob v2 uses a struct logical type for descriptors, which differs from the logical
+            // input struct (data/uri). When intersecting schemas for projection we want to keep
+            // the projected blob layout instead of intersecting by child names.
+            if ignore_types && self.is_blob() && other.is_blob() {
```
westonpace: Why does `ignore_types` need to be in this if statement?

westonpace: What if `self.is_blob` is true but `other.is_blob` is false?  Would you still allow the intersection?

rust/lance/src/blob.rs
```diff
@@ -0,0 +1,238 @@
...
+        self.inner.len()
+    }
+
+    pub fn is_empty(&self) -> bool {
+        self.inner.is_empty()
+    }
+
+    pub fn null_count(&self) -> usize {
+        self.inner.null_count()
+    }
```
westonpace: The builder seems very useful.  This type `BlobArray` is a bit more confusing.  It seems almost like something that implements the `Array` trait (but it doesn't).  I think it's ok but just slightly curious where this is going.

Xuanwo: That's a good question. My intention here is to create a clear API for Blob. This defines the layout for the `Blob` array, so the Builder's `finish` API doesn't just return a `StructArray` simply.

What do you think? Should we just remove this API to make our API surface cleaner? Or I could implement `Array` for it instead?

rust/lance-core/src/datatypes/field.rs
Xuanwo: Oh yeah, I copied the comments for `into_unloaded_with_version` directly here. Let's polish them a bit.

rust/lance-core/src/datatypes/field.rs
Xuanwo: > Why does ignore_types need to be in this if statement?

That's not quiet needed, removing.

> What if self.is_blob is true but other.is_blob is false? Would you still allow the intersection?

I'm not sure how this could happen. It seems they will return an error anyway. So your idea is we can make it more lenient?

westonpace: No, an error is good.  I was worried that if one was a blob struct and the other was a non-blob struct then it would just silently return the non-blob intersection

rust/lance/src/blob.rs
westonpace: Either keep it as-is or just return a `StructArray` from the builder.

Xuanwo: Ok, let's starts with simple API. I will just return a `StructArray` from the builder for now.

Xuanwo: DONE

rust/lance-core/src/datatypes/field.rs
Xuanwo: Got it, added a check

westonpace (approved):

PR #5452, opened 2025-12-10
chore: use uv to run dbpedia benchmark

eddyxu:
Use `uv` to run `dbpedia-openai` benchmarks


eddyxu:
>How much resources are required to run the dbpedia benchmark?

It takes 10-20 minutes on a local xeon machine to run over all the confs.  it might take more than 30 minutes on those GHA machines. We can probably discuss this later?

PR #5458, opened 2025-12-11
feat(python): support cleanup_with_policy

ddupg:


python/python/lance/lance/__init__.pyi
```diff
@@ -298,6 +298,13 @@ class _Dataset:
         delete_unverified: Optional[bool] = None,
         error_if_tagged_old_versions: Optional[bool] = None,
     ) -> CleanupStats: ...
+    def cleanup_with_policy(
```
majin1102: I think we could abstract a CleanPolicy like rust since we said with_policy.

What do you think

ddupg: Thanks for reviewing.

> I think we could abstract a CleanPolicy like rust since we said with_policy.

That make sence, changes updated.

python/python/lance/dataset.py
```diff
@@ -2330,6 +2331,12 @@ def cleanup_old_versions(
             td_to_micros(older_than), delete_unverified, error_if_tagged_old_versions
         )
 
+    def cleanup_with_policy(self, policy: CleanupPolicy) -> CleanupStats:
```
wjones127: Why not just add the new `retain_versions` parameter to `cleanup_old_versions`, why a whole new method?

wjones127 (commented):
Got a question, as this seems like it might be overcomplicating things. There's only one new parameter, right?

python/python/lance/dataset.py
ddupg: Thank @wjones127 for your attention.

> Why not just add the new retain_versions parameter to cleanup_old_versions, why a whole new method?

Just want to align with [rust cleanup_with_policy](https://github.com/lance-format/lance/blob/b478d3b0aed361fca724f1f91d4004786b958804/rust/lance/src/dataset.rs#L1226), I can add new parameter `retain_versions` to `cleanup_old_versions`, but is it okay if they differ from the Rust method?

Or should we first merge rust [cleanup_old_versions](https://github.com/lance-format/lance/blob/b478d3b0aed361fca724f1f91d4004786b958804/rust/lance/src/dataset.rs#L1192) and [cleanup_with_policy](https://github.com/lance-format/lance/blob/b478d3b0aed361fca724f1f91d4004786b958804/rust/lance/src/dataset.rs#L1226) into one?

wjones127: I think it’s fine if the API design differs between Python and Rust. Python has named arguments, which is often a natural API design for Python users. Rust doesn’t, and instead needs to use option structs and builder patterns to make configuration readable.

It seems like this should just be another named argument in Python.

ddupg: > It seems like this should just be another named argument in Python.

Thank @wjones127 for your suggestion. I added `retain_versions` named argument in python cleanup_old_versions

python/python/tests/test_dataset.py
```diff
@@ -1181,6 +1181,45 @@ def test_cleanup_around_tagged_old_versions(tmp_path):
...
+def test_cleanup_with_older_than_and_retain_versions(tmp_path: Path):
+    base_dir = tmp_path / "cleanup_policy"
+    table = pa.Table.from_pydict({"a": range(100), "b": range(100)})
+    lance.write_dataset(table, base_dir, mode="create")
+    time.sleep(0.05)
+    lance.write_dataset(table, base_dir, mode="overwrite")
+    time.sleep(0.05)
+    lance.write_dataset(table, base_dir, mode="overwrite")
+    moment = datetime.now()
+    time.sleep(0.05)
```
wjones127, line 1211: Why are the sleep calls necessary?

python/python/tests/test_dataset.py
```diff
@@ -1181,6 +1181,45 @@ def test_cleanup_around_tagged_old_versions(tmp_path):
...
+    ds = lance.write_dataset(table, base_dir, mode="append")
+
+    stats = ds.cleanup_old_versions(
+        older_than=datetime.now() - moment, retain_versions=2
+    )
+    print(stats)
```
wjones127:
```suggestion
```

python/python/tests/test_dataset.py
ddupg: This test is to verify the combined effect of `older_than` and `retain_versions`, the expected outcome should satisfy both constraints simultaneously. To avoid timing skew affecting the `older_than` evaluation and causing flakiness, I add sleep calls to stabilize the boundary conditions.

PR #5465, opened 2025-12-12
fix: make column name lookups case-insensitive

### LINKED ISSUE #3424: Make column references in SQL strings and column lists case insensitive
When coming from languages that use camel case, such as TypeScript, it's easy to create datasets with mixed-case column names. However, it seems that our SQL parser works only with lower case, and to match any mixed case requires wrapper in backticks.

We should also handle column names with other characters, such as `-` and `:`.

This also is an issue with column lists, such as the `on` parameter in merge insert / `merge_insert`. It's current impossible to use a mixed-case column as the join key in merge insert queries.

Also affects creating scalar indices: https://github.com/lancedb/lancedb/issues/2635

## Fixes

Test that for a column with mixed case and special characters (`-`, `:`), we can:

* [ ] Create a table
* [ ] Select the column in a scan
* [ ] Order by that column in a scan
* [ ] Create a scalar index on that column
* [ ] Alter and drop the column
* [ ] Do merge insert with that column as a key

wjones127:
## Summary

- SQL parsers lowercase unquoted identifiers, but Arrow schema lookups are case-sensitive
- This caused mixed-case column names (e.g., `userId`) to fail in filter expressions, scalar index creation, and merge insert
- Added case-insensitive field lookup that tries exact match first, then falls back to case-insensitive match

## Test plan

- [x] Added Python tests for mixed-case column names (`python/tests/test_column_names.py`)
- [x] All 14 new tests pass
- [x] Existing Rust tests pass (68 merge_insert tests, 45 lance-datafusion tests)

Fixes #3424

### DIFF
--- a/python/python/tests/test_column_names.py
+++ b/python/python/tests/test_column_names.py
@@ -145,12 +145,19 @@ class TestCaseOnlyDifferentColumnNames:
 
     @pytest.fixture
     def case_variant_table(self):
-        """Create a table with columns that differ only in case."""
+        """Create a table with columns that differ only in case.
+
+        Values are deliberately non-correlated to ensure tests catch
+        incorrect column resolution:
+        - camelCase: 0, 1, 2, ... (ascending)
+        - CamelCase: 99, 98, 97, ... (descending)
+        - CAMELCASE: 50, 51, 52, ..., 99, 0, 1, ... (rotated)
+        """
         return pa.table(
             {
-                "camelCase": range(100),
-                "CamelCase": range(100, 200),
-                "CAMELCASE": range(200, 300),
+                "camelCase": list(range(100)),
+                "CamelCase": list(range(99, -1, -1)),  # reversed
+                "CAMELCASE": list(range(50, 100)) + list(range(50)),  # rotated
             }
         )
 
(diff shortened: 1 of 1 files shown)


python/python/tests/test_column_names.py
```diff
@@ -0,0 +1,379 @@
...
+    without requiring backtick escaping.
+
+    Current behavior:
+    - Table creation: WORKS
+    - Filter with SQL string: FAILS (column names are lowercased)
+    - Order by: WORKS (uses column name directly, not SQL parsing)
+    - Scalar index: FAILS (column name is lowercased during validation)
+    - Alter column: WORKS
+    - Drop column: WORKS
+    - Merge insert: FAILS (column names lowercased in join expressions)
```
wjones127:
```suggestion
```

python/python/tests/test_column_names.py
```diff
@@ -0,0 +1,379 @@
...
+    used for nested field paths. This test uses `-` and `:` instead.
+
+    Current behavior:
+    - Table creation: WORKS
+    - Filter with backticks: WORKS
+    - Order by: WORKS (uses column name directly, not SQL parsing)
+    - Scalar index: WORKS
+    - Alter column: WORKS
+    - Drop column: WORKS
+    - Merge insert: FAILS (column name causes parsing issues in join expressions)
```
wjones127:
```suggestion
```

python/python/tests/test_column_names.py
```diff
@@ -0,0 +1,381 @@
...
+    """
+
+    @pytest.fixture
+    def special_char_table(self):
+        """Create a table with special character column names."""
+        return pa.table(
+            {
+                "user-id": range(100),
+                "order:id": range(100, 200),
+                "item_name": [f"item_{i}" for i in range(100)],
```
westonpace, line 298: This one is scary :laughing: I wonder if we should limit the available character set for column names at all?

rust/lance-datafusion/src/planner.rs
```diff
@@ -842,10 +861,18 @@ impl Planner {
...
     /// before being passed to `create_physical_expr()`.
     pub fn parse_expr(&self, expr: &str) -> Result<Expr> {
+        // First check if it's a simple column reference (no operators, functions, etc.)
+        // Try exact match
         if self.schema.field_with_name(expr).is_ok() {
-            return Ok(col(expr));
+            return Ok(Expr::Column(Column::from_name(expr)));
+        }
+        // Try case-insensitive match for simple column reference
+        let resolved_name = self.resolve_column_name(expr);
```
westonpace: Is this trying the exact match twice?  Once explicitly and once implicitly in resolve_column_name?

rust/lance/src/dataset/write/merge_insert.rs
```diff
@@ -374,10 +374,29 @@ impl MergeInsertBuilder {
                 location!(),
             ));
         }
+
+        // Resolve column names using case-insensitive matching to handle
```
westonpace, line 378: This doesn't seem to have consistent "try exact first then fallback to inexact" semantics?

rust/lance/src/dataset/write/merge_insert/assign_action.rs
```diff
@@ -59,17 +59,19 @@ pub fn merge_insert_action(
 ) -> Result<Expr> {
     // Check that at least one key column is non-null in the source
     // This ensures we only process rows that have valid join keys
+    // Note: Column names are wrapped in double quotes to preserve case
+    // (DataFusion's col() function lowercases unquoted identifiers)
     let source_has_key: Expr = if params.on.len() == 1 {
         // Single key column case - check if the source key column is not null
         // Need to qualify the column to avoid ambiguity between target.key and source.key
-        col(format!("source.{}", &params.on[0])).is_not_null()
+        col(format!("source.\"{}\"", &params.on[0])).is_not_null()
```
westonpace, line 67: This seems like a bug fix.  I wonder if we can have a regression test for this?

rust/lance/src/index/create.rs
```diff
@@ -105,11 +105,26 @@ impl<'a> CreateIndexBuilder<'a> {
             });
         }
         let column = &self.columns[0];
-        let Some(field) = self.dataset.schema().field(column) else {
-            return Err(Error::Index {
-                message: format!("CreateIndex: column '{column}' does not exist"),
-                location: location!(),
-            });
+        // Use case-insensitive lookup to handle lowercased column names
+        // For nested paths, use exact match; for simple columns, try case-insensitive
```
westonpace: Why are nested paths using an exact match?  Do simple paths to exact first then case-insensitive like we do elsewhere?

westonpace (approved):
I'm unsure if this gives us consistent "try exact first and the case insensitive" behavior everywhere.  Also, it might be nice to have test cases for multiple columns with different cases.

I think case-insensitive but inconsistent if multiple columns with same name is better than what we have today but having consistent behavior everywhere would be idea.

python/python/tests/test_column_names.py
wjones127: I actually did have a user who wanted this. I don't think there's any reason we shouldn't support it.

rust/lance-datafusion/src/planner.rs
wjones127: I think you are right. Will fix this.

rust/lance/src/dataset/write/merge_insert/assign_action.rs
wjones127: I thought this was already covered by the `merge_insert` test cases in `test_column_names.py`. You think I should also add a Rust test?

wjones127: I can add one.

rust/lance/src/index/create.rs
wjones127: That's an oversight. I'll fix that as well.

rust/lance/src/dataset/write/merge_insert.rs
wjones127: That's implemented in `Schema.field_case_insensitive()`

rust/lance/src/dataset/write/merge_insert/assign_action.rs
westonpace: Ah, if it's covered we are good.

PR #5468, opened 2025-12-13
refactor: add store_prefix to lance-io's ObjectStore

cmccabe:
Add the store_prefix to lance-io's ObjectStore. This is helpful when caching ObjectStore objects, so that we can tell which ones map to the same data.

### DIFF
--- a/.github/workflows/cargo-publish.yml
+++ b/.github/workflows/cargo-publish.yml
@@ -22,7 +22,7 @@ env:
 jobs:
   build:
     # Needs additional disk space for the full build.
-    runs-on: ubuntu-2404-8x-x64
+    runs-on: warp-ubuntu-latest-x64-8x
     permissions:
       id-token: write
     timeout-minutes: 60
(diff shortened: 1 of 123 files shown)


westonpace (approved):
This seems reasonable to me.  This object is somewhat indecisive between using public properties (`list_is_lexically_ordered`, `use_constant_size_upload_parts`) and accessors (`block_size`, `max_iop_size`) so I guess both approaches are fine :laughing:

Xuanwo:
Seems tests case around the following failed, maybe worth another look, cc @cmccabe 

```

failures:
    scheduler::tests::test_backpressure
    scheduler::tests::test_priority
```

cmccabe:
Should be fixed now. I added a fallback path for object stores that aren't in the registry but that's we're "wrapping" via the deprecated API. For those we have to guess the storage prefix, unfortunately, since nothing can tell us about it. It should be OK to just use scheme+authority there.

PR #5469, opened 2025-12-14
feat: support dropping sub-column of list(struct)

wojiaodoubao:
Drop sub-column `people.item.city` from dataset.

```
Table Schema
-id: int32
-people: struct
  -item: list(struct)
    - name: utf8
    - age: int32
    - city: utf8
```

### DIFF
--- a/rust/lance-core/src/datatypes/schema.rs
+++ b/rust/lance-core/src/datatypes/schema.rs
@@ -403,6 +403,30 @@ impl Schema {
         })
     }
 
+    /// Exclude the fields from `other` Schema, and returns a new Schema.
+    pub fn exclude<T: TryInto<Self> + Debug>(&self, schema: T) -> Result<Self> {
+        let other = schema.try_into().map_err(|_| Error::Schema {
+            message: "The other schema is not compatible with this schema".to_string(),
+            location: location!(),
+        })?;
+        let mut fields = vec![];
+        for field in self.fields.iter() {
+            if let Some(other_field) = other.field(&field.name) {
+                if field.data_type().is_nested() {
+                    if let Some(f) = field.exclude(other_field) {
+                        fields.push(f)
+                    }
+                }
+            } else {
+                fields.push(field.clone());
+            }
+        }
+        Ok(Self {
+            fields,
+            metadata: self.metadata.clone(),
+        })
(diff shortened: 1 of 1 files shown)


wojiaodoubao:
Hi @jackye1995 , could you help review this when you have time, thanks very much!

wojiaodoubao:
Hi @jackye1995 , sorry to bother, could you help review when you have time, thanks a lot~

wojiaodoubao:
Hi @jackye1995 , could you help review this when you have time, thanks very much~

rust/lance-core/src/datatypes/schema.rs
```diff
@@ -403,30 +403,6 @@ impl Schema {
         })
     }
 
-    /// Exclude the fields from `other` Schema, and returns a new Schema.
-    pub fn exclude<T: TryInto<Self> + Debug>(&self, schema: T) -> Result<Self> {
```
yanghua, line 407: It seems removing this public API would break compatibility? Must we remove it?

wojiaodoubao: Thanks your reminding, let me revert the change.

rust/lance/src/dataset/schema_evolution.rs
```diff
@@ -702,8 +702,9 @@ pub(super) async fn drop_columns(dataset: &mut Dataset, columns: &[&str]) -> Res
         }
     }
 
+    let version = dataset.manifest.data_storage_format.lance_file_version()?;
     let columns_to_remove = dataset.manifest.schema.project(columns)?;
-    let new_schema = dataset.manifest.schema.exclude(columns_to_remove)?;
+    let new_schema = exclude(&dataset.manifest.schema, &columns_to_remove, &version)?;
```
yanghua, line 707: I mean, if we can keep the old callable style, like line number 706. And define a new exclude method with a `version` parameter. And make the old method with `#[deprecated]`.

wojiaodoubao: Thanks your suggestion! Do you mean adding a `exclude_with_version` function to Schema? The challenge is LanceVersion can't be reached in schema.rs because they are in different crates.

PR #5471, opened 2025-12-15
refactor(java): arrow schema should be returned by Fragment.mergeColumns

fangbo:
1. The Fragment.mergeColumns's returned schema will be used in org.lance.operation.Merge which needs a Arrow Schema. So, Fragment.mergeColumn returns a Arrow Schema can avoid the convert from Lance Schema to Arrow Schema in Java. Like Dataset.getSchema, the convert is implemented in Rust.

2. Fix the issue https://github.com/lance-format/lance-spark/issues/138

### DIFF
--- a/.github/workflows/ci-benchmarks.yml
+++ b/.github/workflows/ci-benchmarks.yml
@@ -46,15 +46,31 @@ jobs:
           source venv/bin/activate
           pip install maturin duckdb requests pytest pytest-benchmark
           maturin develop --locked --release
+      - name: Build memtest
+        run: |
+          source venv/bin/activate
+          cd memtest
+          make build-release
       - name: Generate datasets
         run: |
-          python -m venv venv
           source venv/bin/activate
           python python/ci_benchmarks/datagen/gen_all.py
       - name: Run benchmarks
         run: |
-          python -m venv venv
           source venv/bin/activate
           bencher run --project weston-lancedb --token ${{ secrets.LANCE_BENCHER_TOKEN }} --adapter python_pytest \
                       --branch main --testbed google-genoa --err --file results.json "python -mpytest --benchmark-json \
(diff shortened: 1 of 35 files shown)


fangbo:
@jackye1995  This PR is ready, could you please review it? Thank you.

java/src/main/java/org/lance/fragment/FragmentMergeResult.java
```diff
@@ -14,25 +14,25 @@
...
+import org.apache.arrow.vector.types.pojo.Schema;
 
 /**
  * Result of {@link org.lance.Fragment#mergeColumns(ArrowArrayStream, String, String)
  * Fragment.mergeColumns()}.
  */
 public class FragmentMergeResult {
   private final FragmentMetadata fragmentMetadata;
-  private final LanceSchema schema;
+  private final Schema schema;
```
majin1102, line 28: Thanks for catching this.

I wonder if we should fix the LanceSchema converting issue since the problem occurs there. Is there any blocking pointing that we could not use LanceSchema here?

fangbo: Thank you for comments.

I think LanceSchema should be fixed. It is another thing, we could submit another PR to fix it.

For Fragment.mergeColumns, it is sensible to return an Arrow schema, for three reasons:

1. The conversion from Lance schema to Arrow schema is already correctly implemented in Rust, so we can just reuse it like the Dataset.getSchema.
2. The Merge transaction itself expects an Arrow schema.
3. Dataset's public methods use Arrow  for data operation (read/write) . So Arrow schema can keep consistency with other methods.

fangbo:
This PR is duplicated with https://github.com/lance-format/lance/pull/5509 , close it.

PR #5472, opened 2025-12-15
feat: support HDFS object store

hfutatzhanghb:
## What changed

- add an optional `hdfs` feature for `lance-io` backed by OpenDAL
- register `hdfs://` with the object store registry
- map HDFS URLs, storage options, and supported environment variables into OpenDAL configuration
- add unit coverage for HDFS path and configuration handling

## Why

This allows Lance datasets to be accessed through HDFS URLs, including deployments using HDFS nameservices.

## Validation

- `cargo fmt --all -- --check`
- `cargo clippy -p lance-io --all-targets --features hdfs --locked -- -D warnings`
- `cargo test -p lance-io --features hdfs --locked object_store::providers::hdfs::tests --no-fail-fast`
- `cargo +1.91.0 check -p lance-io --features hdfs --locked`


hfutatzhanghb:
Hi, @jackye1995 @wojiaodoubao @majin1102 . Could you please help review this PR when you have free time? Thanks a lot.

rust/lance-io/Cargo.toml
```diff
@@ -63,13 +64,14 @@ name = "scheduler"
 harness = false
 
 [features]
-default = ["aws", "azure", "gcp"]
+default = ["aws", "azure", "gcp", "hdfs"]
```
Xuanwo: I think we shouldn't enable HDFS by default. It introduces many new dependencies in Lance and also requires users to have a Java setup. Without it, Lance will fail to start.

rust/lance-io/src/object_store/providers/hdfs.rs
```diff
@@ -0,0 +1,318 @@
...
+                if !hadoop_user.is_empty() {
+                    config_map.insert("user".to_string(), hadoop_user);
+                }
+            } else if let Some(hdfs_user) = std::env::var("HDFS_USER").ok() {
+                if !hdfs_user.is_empty() {
+                    config_map.insert("user".to_string(), hdfs_user);
+                }
+            } else {
+                // Fall back to system username
+                config_map.insert("user".to_string(), whoami::username());
```
Xuanwo: I prefer having users fill this out instead of using `whoami`.

java/pom.xml
```diff
@@ -385,6 +385,9 @@
                                     <!-- Copy native libraries to target/classes for runtime access -->
                                     <copyTo>${project.build.directory}/classes/nativelib</copyTo>
                                     <copyWithPlatformDir>true</copyWithPlatformDir>
+                                    <environmentVariables>
+                                        <CARGO_FEATURES>hdfs</CARGO_FEATURES>
```
Xuanwo: I feel the same. It doesn’t seem like a good idea to enable hdfs by default.

hfutatzhanghb: @Xuanwo Thanks for reviewing.  Got it, will push an new version based on your advice laterly.

hfutatzhanghb:
@Xuanwo @BubbleCal @jiaoew1991 Hi， could you please review this PR when free? Thanks very much. We have use HDFS as backend storage for a long time.

hfutatzhanghb:
@yanghua Hi， could you please help review this PR when have free time? Thanks very much!

yanghua:
@claude review

hfutatzhanghb:
> @claude review
😂， No output ? @claude

yanghua:
It seems the bot does not work. I could take a look at this PR later.

hfutatzhanghb:
> It seems the bot does not work. I could take a look at this PR later.

Thanks！

rust/lance-io/src/object_store/providers/hdfs.rs
```diff
@@ -0,0 +1,268 @@
...
+            .finish();
+        let opendal_store = Arc::new(OpendalStore::new(operator));
+
+        Ok(ObjectStore {
+            scheme: "hdfs".to_string(),
+            inner: opendal_store,
+            block_size,
+            max_iop_size: *DEFAULT_MAX_IOP_SIZE,
+            use_constant_size_upload_parts: params.use_constant_size_upload_parts,
+            list_is_lexically_ordered: params.list_is_lexically_ordered.unwrap_or(true),
```
yanghua: Should it be set to `false`?

rust/lance-io/src/object_store/providers/hdfs.rs
```diff
@@ -0,0 +1,268 @@
...
+
+        let user = storage_options
+            .0
+            .get("hdfs_user")
+            .cloned()
+            .or_else(|| env_vars.get("HADOOP_USER_NAME").cloned())
+            .or_else(|| env_vars.get("HDFS_USER").cloned());
+        if let Some(user) = user {
+            config.insert("user".to_string(), user);
+        }
```
yanghua, line 111: Shall we also filter empty string values?

rust/lance-io/tests/hdfs_integration.rs
```diff
@@ -0,0 +1,167 @@
...
+
+    fn hdfs_available() -> bool {
+        std::env::var("HDFS_NAME_NODE").is_ok() || std::env::var("HDFS_TEST_ENABLED").is_ok()
+    }
+
+    #[tokio::test]
+    async fn test_hdfs_store_creation() {
+        if !hdfs_available() {
+            return;
+        }
```
yanghua: These conditions would make these test cases hard to find real problems.

rust/lance-io/tests/hdfs_integration.rs
```diff
@@ -0,0 +1,167 @@
...
+        let (store, path) = ObjectStore::from_uri_and_params(registry, url, &params)
+            .await
+            .unwrap();
+
+        assert_eq!(store.scheme(), "hdfs");
+        assert_eq!(path, Path::from("user/test"));
+    }
+
+    #[tokio::test]
+    async fn test_hdfs_url_parsing() {
```
yanghua: Is it duplicated with `test_hdfs_store_paths`?

rust/lance-io/tests/hdfs_integration.rs
```diff
@@ -0,0 +1,167 @@
...
+        let params = ObjectStoreParams {
+            storage_options_accessor: Some(Arc::new(StorageOptionsAccessor::with_static_options(
+                storage_options,
+            ))),
+            ..Default::default()
+        };
+
+        // This test verifies the configuration validation logic without actually connecting
+        // The actual connection would fail, but we're testing the config validation
+        assert!(params.storage_options().is_some());
```
yanghua: It's too simple, we need to check `HdfsStoreProvider::build_config`.

rust/lance-io/tests/hdfs_integration.rs
```diff
@@ -0,0 +1,167 @@
...
+
+    use lance_io::object_store::{
+        ObjectStore, ObjectStoreParams, ObjectStoreRegistry, StorageOptionsAccessor,
+    };
+    use object_store::ObjectStoreExt;
+    use object_store::path::Path;
+    use url::Url;
+
+    fn hdfs_available() -> bool {
+        std::env::var("HDFS_NAME_NODE").is_ok() || std::env::var("HDFS_TEST_ENABLED").is_ok()
```
yanghua: Can we add some description on the doc for `HDFS_TEST_ENABLED `?

yanghua (commented):
Left some comments.

rust/lance-io/src/object_store/providers/hdfs.rs
hfutatzhanghb: I think set to true here is ok.  Because the  output of list rpc in hdfs is lexically_ordered.

rust/lance-io/src/object_store/providers/hdfs.rs
hfutatzhanghb: Thanks for reminding, Have fixed.

rust/lance-io/tests/hdfs_integration.rs
hfutatzhanghb: Yes, have removed test_hdfs_url_parsing. Thanks a lot!

rust/lance-io/tests/hdfs_integration.rs
hfutatzhanghb: Have added. Thanks.

rust/lance-io/tests/hdfs_integration.rs
hfutatzhanghb: Removed the ineffective integration test, which only asserted that `storage_options` was present and did not exercise HDFS configuration parsing. `HdfsStoreProvider::build_config` is private and is tested directly in `rust/lance-io/src/object_store/providers/hdfs.rs`, covering URI fallback, storage option precedence over environment/URI, environment configuration, optional Kerberos/atomic-write options, and rejection of URIs without a host. The targeted HDFS provider unit tests pass (6 tests).

rust/lance-io/tests/hdfs_integration.rs
hfutatzhanghb: Updated the HDFS integration tests to follow the existing object-store integration test pattern used by GooseFS and GCS. Each test is now marked with `#[ignore = "Requires HDFS cluster"]` instead of silently returning based on environment variables. The tests are explicitly run with `--ignored`, and once selected they execute fully so connection or behavior failures are surfaced. The default targeted run now reports `0 passed, 4 ignored` rather than treating unexecuted tests as passing.

hfutatzhanghb:
@yanghua Thanks very much for valuable review opinions. Have uploaded the latest codes.

yanghua:
@hfutatzhanghb, fix the conflicts?

hfutatzhanghb:
> @hfutatzhanghb, fix the conflicts?

have fixed and pushed. Let's wait pipeline ok. Thanks for your reviewing!!!

yanghua:
Will take a look today.

yanghua (approved):
+1 cc @Xuanwo

Xuanwo (changes requested):
Thanks for adding HDFS support. I think this needs a few correctness fixes before merge.

The main blocker is that this PR registers `hdfs://` as writable, but commit routing still falls through to `UnsafeCommitHandler`. That means concurrent HDFS writers can overwrite manifest versions without real create-if-absent / conflict detection semantics.

There are also two HDFS identity/listing issues:
- HDFS is advertised as `list_is_lexically_ordered=true`, but the backend does not guarantee sorted listings before Lance uses listing order for manifest discovery.
- `store_prefix` is derived from the URI authority even when `hdfs_name_node` / `HDFS_NAME_NODE` overrides the actual NameNode, so different effective clusters can be treated as the same datastore.

Please route HDFS to a safe commit handler or explicitly reject HDFS writes without one, default HDFS listings to unordered unless sorted in the provider, and derive `store_prefix` from the resolved effective NameNode.

hfutatzhanghb:
> Thanks for adding HDFS support. I think this needs a few correctness fixes before merge.
> 
> The main blocker is that this PR registers `hdfs://` as writable, but commit routing still falls through to `UnsafeCommitHandler`. That means concurrent HDFS writers can overwrite manifest versions without real create-if-absent / conflict detection semantics.
> 
> There are also two HDFS identity/listing issues:
> 
> * HDFS is advertised as `list_is_lexically_ordered=true`, but the backend does not guarantee sorted listings before Lance uses listing order for manifest discovery.
> * `store_prefix` is derived from the URI authority even when `hdfs_name_node` / `HDFS_NAME_NODE` overrides the actual NameNode, so different effective clusters can be treated as the same datastore.
> 
> Please route HDFS to a safe commit handler or explicitly reject HDFS writes without one, default HDFS listings to unordered unless sorted in the provider, and derive `store_prefix` from the resolved effective NameNode.

@Xuanwo Thanks for reviewing.
1. For UnsafeCommitHandler,  we have already implements a HdfsRenameCommitHandler and planed to push in another PR after this mergeed.  Now have pushed in this PR. 
2. For indentity/listing issues: Will fix immediately.

hfutatzhanghb:
@Xuanwo  @yanghua  To provide better rename semantics for the Lance HDFS feature, we also optimized the libhdfs rename operation in https://github.com/apache/hadoop/pull/8411.

Cargo.toml
```diff
@@ -104,6 +104,7 @@ half = { "version" = "2.1", default-features = false, features = [
     "num-traits",
     "std",
 ] }
+hdrs = "0.3.2"
```
Xuanwo: We should not depend on this

hfutatzhanghb: Hi, @Xuanwo . Do you mean we should avoid using hdrs directly and implement the HDFS commit rename through OpenDAL instead?  My concern is preserving the rename-if-destination-does-not-exist semantics required for commit conflict detection.

hfutatzhanghb: OpenDAL’s HDFS rename currently overwrites the destination by deleting it first. The manifest commit requires an atomic rename-if-destination-does-not-exist operation for conflict detection. Would you prefer adding conditional rename support to OpenDAL first, or is there another OpenDAL API intended for this semantic?

wjones127: See me comment here: https://github.com/lance-format/lance/issues/7584#issuecomment-4898152245

hfutatzhanghb: Hi, @wjones127 . Got it. That a good plan. This PR is not the final version currently. It depends on the rename_with capabilty of OpenDAL which was introduced by https://github.com/apache/opendal/pull/7815 . We should release a new version for OpenDAL firstly.   BTW, in our cluster. We use patch.crates-io to patch the OpenDAL code and HDRS code.

PR #5473, opened 2025-12-15
feat(blob_v2): add GC support

Xuanwo:
This PR adds GC support for blob v2.

Part of https://github.com/lance-format/lance/issues/4947

All blob files (whether packed or dedicated) follow the same naming pattern: `data/{data_file_key}/{blob_id:08x}.blob`. These files are only referenced by `data/{data_file_key}.lance`. Therefore, we can implement GC support by checking whether the related data file is still in use.

External blobs are not being considered at this stage.

---

### DIFF
--- a/rust/lance/src/dataset/cleanup.rs
+++ b/rust/lance/src/dataset/cleanup.rs
@@ -58,7 +58,7 @@ use std::{
     future,
     sync::{Mutex, MutexGuard},
 };
-use tracing::{info, instrument, Span};
+use tracing::{debug, info, instrument, Span};
 
 use super::refs::TagContents;
 use crate::{utils::temporal::utc_now, Dataset};
@@ -397,6 +397,10 @@ impl<'a> CleanupTask<'a> {
                 // These files are not referenced directly by the manifest.  Instead, treat them
                 // as referenced if their parent data file is referenced.
                 if !relative_path.as_ref().starts_with("data") {
+                    debug!(
+                        path = relative_path.as_ref(),
+                        "Will not garbage collect blob file because it does not follow convention"
+                    );
                     return Ok(None);
                 }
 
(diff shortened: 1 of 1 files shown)


rust/lance/src/dataset/cleanup.rs
```diff
@@ -390,6 +390,55 @@ impl<'a> CleanupTask<'a> {
...
                 }
             }
+            Some("blob") => {
+                // Blob v2 sidecar files are keyed by the data file stem:
+                //   data/{data_file_key}/{blob_id:08x}.blob
+                //
+                // These files are not referenced directly by the manifest.  Instead, treat them
+                // as referenced if their parent data file is referenced.
+                if !relative_path.as_ref().starts_with("data") {
+                    return Ok(None);
```
westonpace, line 404: Can we log a message at debug level here?  Something like "will not garbage collect blob file {relative_path} because it does follow convention".

rust/lance/src/dataset/cleanup.rs
```diff
@@ -390,6 +390,55 @@ impl<'a> CleanupTask<'a> {
...
+                }
+
+                let mut parts = relative_path.parts();
+                let data_dir = parts.next();
+                let data_file_key = parts.next();
+                let blob_file = parts.next();
+                // Be conservative: only handle the expected 3-part layout.
+                if data_dir.is_none() || data_file_key.is_none() || blob_file.is_none() {
+                    return Ok(None);
```
westonpace, line 417: Same, let's log here

rust/lance/src/dataset/cleanup.rs
```diff
@@ -390,6 +390,55 @@ impl<'a> CleanupTask<'a> {
...
+                }
+                if parts.next().is_some() {
+                    return Ok(None);
```
westonpace, line 424: And here

rust/lance/src/dataset/cleanup.rs
```diff
@@ -390,6 +390,55 @@ impl<'a> CleanupTask<'a> {
...
+                }
+
+                let data_file_key = data_file_key.expect("checked is_some");
+                let Ok(parent_data_path) =
+                    Path::parse(format!("data/{}.lance", data_file_key.as_ref()))
+                else {
+                    return Ok(None);
```
westonpace, line 436: And here (maybe with the error)

rust/lance/src/dataset/cleanup.rs
Xuanwo: Yep, makes sense.

PR #5474, opened 2025-12-15
refactor: expose take_blobs_by_addresses to python

Xuanwo:
This PR will expose take_blobs_by_addresses to python so that our python users won't depend on `take_blobs` anymore.

---

### DIFF
--- a/docs/src/format/table/row_id_lineage.md
+++ b/docs/src/format/table/row_id_lineage.md
@@ -7,9 +7,13 @@ Row addressing enables efficient random access to rows within the table through 
 Stable row IDs provide persistent identifiers that remain constant throughout a row's lifetime, even as its physical location changes.
 Row version tracking records when rows were created and last modified, enabling incremental processing, change data capture, and time-travel queries.
 
-## Row ID Styles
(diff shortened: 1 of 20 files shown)


python/python/tests/test_blob.py
```diff
@@ -110,6 +110,49 @@ def test_blob_files_by_address(dataset_with_blobs):
...
+                "blobs": pa.array([b"bar"], pa.large_binary()),
+                "idx": pa.array([1], pa.uint64()),
+            },
+            schema=table.schema,
+        )
+    )
+
+    t = ds.to_table(columns=["idx"], with_row_address=True)
+    # Pick a row from the second fragment. With stable row ids enabled, the row address
+    # is not a valid row id, so calling the row-id based API would fail.
```
yanghua: > so calling the row-id based API would fail

About this judgment description. IMO, we can either remove it or verify it.

yanghua (approved):
LGTM.

python/python/tests/test_blob.py
Xuanwo: Removed.

PR #5475, opened 2025-12-15
feat: add IVF_HNSW_RQ index support

BubbleCal:
Adds the IVF_HNSW_RQ vector index type and enables HNSW + RabitQ end-to-end. RabitQ codes remain unpacked for this combination and RabitQ now implements dist_calculator_from_id for HNSW. Exposes IVF_HNSW_RQ in the Python API and adds a basic Python test.


BubbleCal:
the RQ requires to rotate the residual query vector which takes `O(d^2)` time and it's too expensive, going to close this.

PR #5487, opened 2025-12-16
feat: expose `use_scalar_index` param in Java scanner

xloya:
This parameter is already exposed in the current Python SDK. When using the Java SDK, some query scenarios require disabling scalar indexes for faster query speeds. Therefore, exposing this parameter in the Java SDK helps users customize whether to enable or disable scalar indexes.


Xuanwo (approved):

PR #5488, opened 2025-12-16
feat(java): simplify the use of optional in jni

ddupg:
This PR made the following changes:
- Unwrap java `Optional` inside `get_optional` , so the callback receives the unwrapped value instead of Optional<T> ; removes repeated get() boilerplate
- Fix `inner_shallow_clone` to correctly treat the closure input
- Replace `Optional` handling with `get_optional` to reduce duplication and improve readability


ddupg:
Could please @LuQQiu @jackye1995 @majin1102 help review this?

jackye1995 (approved):
looks good to me!

PR #5490, opened 2025-12-16
refactor: clean code for decoder.rs

camilesing:
for issue: https://github.com/lance-format/lance/issues/5489


Xuanwo:
Thank you for working on this. However, I don't think refactoring the legacy code is worth the effort. Closing this now.

camilesing:
Oh no, I think this matter is as important as adding test cases to cover old code

PR #5491, opened 2025-12-16
feat(blob_v2): add Python API for Blob v2

Xuanwo:
This PR will expose blob v2 to python API, allow users to write blob v2 data.

Well, I don't have much experience in designing Python APIs. @westonpace, could you please take a look at this shape?

---

### DIFF
--- a/rust/lance-core/src/datatypes/field.rs
+++ b/rust/lance-core/src/datatypes/field.rs
@@ -683,8 +683,8 @@ impl Field {
             }
             (DataType::Struct(_), DataType::Struct(_)) => {
                 // Blob v2 columns are special: they can have different struct layouts
-                // (logical input vs. descriptor struct). We treat blob v2 structs as opaque
-                // during schema set operations (union/subtract).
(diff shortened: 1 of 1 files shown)


rust/lance-core/src/datatypes/field.rs
```diff
@@ -682,6 +682,12 @@ impl Field {
                 Ok(self.clone())
             }
             (DataType::Struct(_), DataType::Struct(_)) => {
+                // Blob v2 columns are special: they can have different struct layouts
+                // (logical input vs. descriptor struct). We treat blob v2 structs as opaque
+                // during schema set operations (union/subtract).
```
westonpace: I'm not sure I'd use the word "opaque" here.  Maybe "we treat blob v2 structs as primitive fields (like a binary column) during schema set operations?

Xuanwo:
> So what if we add to the blob array position and length (with length=-1 meaning the whole file) and so the input and output are the same? The only difference is that the input might be a mix of data and uri but the output would always have data be null and uri be set?

Seems like a nice idea to me, will create a follow up issue for this idea.

PR #5504, opened 2025-12-17
docs: add the supported encoding Arrow data types docs

xloya:
According to https://github.com/lance-format/lance/pull/5349#pullrequestreview-3582353553, we should add documentation for the Arrow data type that currently supports encoding.


xloya:
cc @Xuanwo

PR #5506, opened 2025-12-17
feat(python): expose search_filter in scanner

wojiaodoubao:
Adding python api to support using fts as a filter for vector search, or using vector_query as a filter for fts search. Related to https://github.com/lance-format/lance/pull/4928.

### DIFF
--- a/python/python/lance/dataset.py
+++ b/python/python/lance/dataset.py
@@ -55,6 +55,7 @@ from .lance import (
     DatasetBasePath,
     IOStats,
     LanceSchema,
+    PySearchFilter,
     ScanStatistics,
     _Dataset,
     _MergeInsertBuilder,
@@ -679,7 +680,9 @@ class LanceDataset(pa.dataset.Dataset):
     def scanner(
         self,
         columns: Optional[Union[List[str], Dict[str, str]]] = None,
-        filter: Optional[Union[str, pa.compute.Expression]] = None,
+        filter: Optional[
+            Union[str, pa.compute.Expression, FullTextQuery, VectorSearchQuery, dict]
+        ] = None,
         limit: Optional[int] = None,
         offset: Optional[int] = None,
         nearest: Optional[dict] = None,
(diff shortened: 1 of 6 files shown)


wojiaodoubao:
hi @BubbleCal , could you help review this when you have time, thanks very much~

python/python/lance/dataset.py
```diff
@@ -4576,8 +4578,19 @@ def columns(
             )
         return self
 
-    def filter(self, filter: Union[str, pa.compute.Expression]) -> ScannerBuilder:
-        if isinstance(filter, pa.compute.Expression):
+    def filter(
+        self,
+        filter: Union[
+            str, pa.compute.Expression, FullTextQuery, VectorSearchQuery, dict
```
BubbleCal, line 4797: let's add some doc strings for this param, esp for `dict` case

BubbleCal (approved):
Great work!
just need to add some doc strings then feel free to merge

python/python/lance/dataset.py
```diff
@@ -722,10 +725,19 @@ def scanner(
...
+            - Expression filter is pa.compute.Expression or str that is a valid SQL
+             where clause. See `Lance filter pushdown
+             <https://lance.org/guide/read_and_write/#filter-push-down>`_
+             for valid SQL expressions.
+
+            - Query filter is VectorSearchQuery which is only applied to full text
+             search, or FullTextQuery which is only applied to vector query.
+
+            - Dictionary is a combined filter containing both expression filter and
+             query filter.
```
westonpace: What are the keys supposed to be in this case?  We should document that, and ideally provide an example.  Looking at the tests it appears they are `expr_filter` and `query_filter`?

python/python/lance/dataset.py
```diff
@@ -722,10 +725,19 @@ def scanner(
...
+        filter: pa.compute.Expression, str, VectorSearchQuery, FullTextQuery or dict
+            Lance supports 2 kinds of filters: expression filter and query filter.
+
+            - Expression filter is pa.compute.Expression or str that is a valid SQL
+             where clause. See `Lance filter pushdown
+             <https://lance.org/guide/read_and_write/#filter-push-down>`_
+             for valid SQL expressions.
+
+            - Query filter is VectorSearchQuery which is only applied to full text
+             search, or FullTextQuery which is only applied to vector query.
```
westonpace: We need to document the behavior here.  Also, can we treat these as two different things.

```
- Query filter can be a FullTextQuery if the scan is a vector search.  In this case the search will only consider rows which contain at least one token matching the input query.  An error will be raised if the scan is not a vector search.

- Query filter can be a VectorSearchQuery if the scan is a full text search.  In this case the search will ... (I don't actually know what this means)
```

westonpace (commented):
Can we call this a "search filter" instead of a "query filter"?  I think of queries as both "filtered scans" and "searches".  This filter both:

 * Only applies to searches (can't use this on a filtered scan)
 * Is a search itself (e.g. a VectorSearchQuery or FullTextQuery)

python/python/lance/dataset.py
wojiaodoubao: Yes, let me fix it.

python/python/lance/dataset.py
wojiaodoubao: ok

wojiaodoubao:
Hi @westonpace , pr has been updated, please review when you have time, thanks very much!

python/python/lance/dataset.py
```diff
@@ -722,10 +725,50 @@ def scanner(
...
+             text search. Example:
+           .. code-block:: python
+
+             filter=VectorSearchQuery(
+                        "vector",
+                        np.array([12, 17, 300, 10], dtype=np.float32),
+                        5,
+                        20,
+                        True,
+                    )
```
westonpace, line 740: I still don't actually know what happens in this case.  Is the full text search applied after the vector search?  Basically just re-ranking the vector search results?

Or is there some equivalent to "at least one token matches" that is being used here?

wojiaodoubao: > Is the full text search applied after the vector search? Basically just re-ranking the vector search results?

Yes, in this case the vector search is used as a filter for full text search. But it doesn't mean the full text search is always applied after the vector search. The `Scanner::prefilter` determines whether the vector search filter is used before/after full text search.

If `prefiter=false`, we first do a full text search and get rows with score > 0, then we do refine filter using flat knn to get the top 5 nearest rows.

If `prefilter=true`, we first do a vector search and get 5 rows, then do a flat full text search based on the 5 rows. The full text search will drop the unmatched rows and re-rank the results by score.

wojiaodoubao:
Hi @westonpace, @BubbleCal, shall we move forward?

BubbleCal:
LGTM, thanks for adding this!

PR #5509, opened 2025-12-17
fix(java): support FixedSizeList for java LanceField

fangbo:
# Problem Description

The java method [Fragment.mergeColumns](https://github.com/lance-format/lance/blob/main/java/src/main/java/org/lance/Fragment.java#L152) returns a [LanceSchema](https://github.com/lance-format/lance/blob/main/java/src/main/java/org/lance/schema/LanceSchema.java).

The java LanceField is converted from a rust lance Field. If a field is a vector which arrow type is FixedSizeList with element's type is Float, the java LanceField missed the element's type when converted from rust. But the Merge transaction needs a arrow Schema. So, when java LanceSchema is converted to arrow Schema a [error occurs](https://github.com/lance-format/lance-spark/issues/138).

For vector field, the FixedSizeList's element type is only contained in logicalType. This PR use the logicalType to correctly build a arrow Field when converting java LanceSchema to arrow Schema.

@westonpace @jackye1995 @majin1102 Could you give some suggestion about this modification? Thank you very much.

### DIFF
--- a/java/src/main/java/org/lance/schema/LanceField.java
+++ b/java/src/main/java/org/lance/schema/LanceField.java
@@ -14,11 +14,17 @@
 package org.lance.schema;
 
 import com.google.common.base.MoreObjects;
+import com.google.common.collect.ImmutableMap;
+import org.apache.arrow.vector.types.DateUnit;
+import org.apache.arrow.vector.types.FloatingPointPrecision;
+import org.apache.arrow.vector.types.TimeUnit;
 import org.apache.arrow.vector.types.pojo.ArrowType;
 import org.apache.arrow.vector.types.pojo.DictionaryEncoding;
 import org.apache.arrow.vector.types.pojo.Field;
 import org.apache.arrow.vector.types.pojo.FieldType;
 
+import java.util.Arrays;
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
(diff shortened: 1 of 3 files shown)


java/src/main/java/org/lance/schema/LanceField.java
```diff
@@ -29,6 +35,7 @@ public class LanceField {
   private final int parentId;
   private final String name;
   private final boolean nullable;
+  private final String logicalType;
```
majin1102, line 39: https://github.com/lance-format/lance/pull/4207#discussion_r2201397463

Just a little context

majin1102 (commented):
Provide some context, please take a look

fangbo:
> Provide some context, please take a look

Thanks for your comment.

When Lance Schema(Rust) is converted to LanceSchema(Java) and then LanceSchema(java) is converted to Arrow's Schema using LanceSchema.asArrowSchema the FixedSizeList's children is missed.

The original issue is https://github.com/lance-format/lance-spark/issues/138 and related PR is https://github.com/lance-format/lance/pull/5471

java/src/main/java/org/lance/schema/LanceField.java
fangbo: For lance vector data type, only the logicalType has FixedSizeList's element type, such as `fixed_size_list:float:3`.

So, when LanceField is converted to arrow Field , the FixedSizeList's children should be constructed from logicalType.

fangbo:
@westonpace @jackye1995 This PR is ready. I look forward to your feedback. Thank you.

PR #5513, opened 2025-12-17
refactor: use list encoder / scheduler directly instead of replicating behavior in map encoder / scheduler

westonpace:
This is a minor simplification of the recently added map support.  It does not add any functionality, only simplifies the code slightly.  The current implementation creates a `StructuralMapEncoder` which mimics the `StructuralListEncoder`.  This change instead wraps the `StructuralListEncoder` (composition instead of duplication).  As a result we can get rid of the `StructuralMapScheduler`.  The decoder then simply casts from a list array to a map array.

Since the list encoder is so simple this isn't really saving us a ton of work.  However, if we choose to experiment or add new complicated features to the list encoder in the future then this will help us avoid doing the work twice or missing the changes in the map encoder.

### DIFF
--- a/rust/lance-encoding/src/decoder.rs
+++ b/rust/lance-encoding/src/decoder.rs
@@ -239,7 +239,6 @@ use crate::compression::{DecompressionStrategy, DefaultDecompressionStrategy};
 use crate::data::DataBlock;
 use crate::encoder::EncodedBatch;
 use crate::encodings::logical::list::StructuralListScheduler;
-use crate::encodings::logical::map::StructuralMapScheduler;
 use crate::encodings::logical::primitive::StructuralPrimitiveFieldScheduler;
 use crate::encodings::logical::r#struct::{StructuralStructDecoder, StructuralStructScheduler};
 use crate::format::pb::{self, column_encoding};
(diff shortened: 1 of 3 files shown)


westonpace:
CC @xloya do you mind reviewing?  I don't feel super strongly about this so if you prefer the old approach that is ok too.  Still, I think if we can avoid the map encoder having its own logic that will be helpful.

xloya:
I think this optimization is great! There's only one issue regarding the list field naming.

rust/lance-encoding/src/decoder.rs
```diff
@@ -784,14 +783,14 @@ impl CoreFieldDecoderStrategy {
...
-                    .children
-                    .first()
-                    .expect("Map field must have an entries child");
-                let child_scheduler =
-                    self.create_structural_field_scheduler(entries_child, column_infos)?;
-                Ok(Box::new(StructuralMapScheduler::new(child_scheduler))
-                    as Box<dyn StructuralFieldScheduler>)
+
+                let list_field = Field::try_from(ArrowField::new(
+                    field.name.clone(),
```
xloya, line 788: Should we use the name of `entries_field` here?

rust/lance-encoding/src/encodings/logical/struct.rs
```diff
@@ -284,9 +284,14 @@ impl StructuralStructDecoder {
                         location: location!(),
                     });
                 }
-                let child_decoder = Self::field_to_decoder(entries_field, should_validate)?;
+                let list_field = Arc::new(arrow_schema::Field::new(
+                    field.name().clone(),
```
xloya, line 288: Same question about name

westonpace:
Closing due to age

PR #5524, opened 2025-12-18
chore: clean up the residual FTS hardcode for index

shiyajuan123:
Before,  there were two names that mapped to the FTX index they are `INVERTED`, `FTS`.  At some time, the code has been refactored only to use `INVERTED`. But there are still some residual `FTS` hardcodes that need to be cleaned.


yanghua:
The failed CI may be caused by #5525.

PR #5526, opened 2025-12-18
feat: introduce memtrace to track the memory usage about lance

Xuanwo:
This PR provides an alternative implementation for `memtest`. In this version, we introduce `memtrace` as a new feature in Pylance. Users can enable it via the `memtrace` feature flag. It offers a similar API to `memtest`, but it eliminates the need for users to hook or inject dynamic libraries, making it easier to use and test.

Before:

```shell
maturin develop
make -C ../memtest build-release
LIB_PATH=$(lance-memtest)
LD_PRELOAD=$LIB_PATH pytest python/ci_benchmarks
```

After:

```shell
maturin develop --features memtrace
pytest python/ci_benchmarks
```

---

### DIFF
--- a/python/pyproject.toml
+++ b/python/pyproject.toml
@@ -115,4 +115,7 @@ filterwarnings = [
     # Pytorch 2.2 on Python 2.12
     'ignore:.*is deprecated and will be removed in Python 3\.14.*:DeprecationWarning',
     'ignore:.*The distutils package is deprecated.*:DeprecationWarning',
+    # TensorFlow/Keras import can emit NumPy deprecation FutureWarnings in some environments.
+    # Keep FutureWarnings as errors generally, but ignore this known-noisy import-time warning.
+    'ignore:.*np\.object.*:FutureWarning',
 ]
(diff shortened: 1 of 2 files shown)


Xuanwo:
> The `memtest` one is the fourth approach I tried, so I might share the four approaches I tried and why I went with this:

Thank you very much for this! I initially assumed that most of our workload is handled within the Rust core, so the Python part wouldn't require much attention. However, it seems my assumption was incorrect. 

Could you elaborate on why we need to consider Python's memory usage as well? From my current understanding, if we're building an online service around lancedb, operations like building the index should be handled server-side in Rust, while users would primarily use Python on the client side.

wjones127:
> Could you elaborate on why we need to consider Python's memory usage as well? From my current understanding, if we're building an online service around lancedb, operations like building the index should be handled server-side in Rust, while users would primarily use Python on the client side.

This is the Lance library, not LanceDB.

Some operations are handled entirely in Rust, like indexing and queries. But for writes, the data comes from outside. Testing that we can stream writes properly is one of the main things we want to test. That's why I gave the example of insert earlier: it tests that we can take a stream of data, and write it out without collecting it all into memory.

Xuanwo:
Thanks! The great answer addressed my questions. I'm now going to close this PR as I do think the current way seems to be the only way.

> But for writes, the data comes from outside.

One last question: is it a good idea to measure that lance memory usage didn't grow up during writing? So we can still measure the rust side instead.

Xuanwo:
Let's close.

PR #5527, opened 2025-12-18
fix: json's arrow extension metadata missing

Xuanwo:
This PR fixed the issue that json's arrow extension metadata missing between the convert of arrow field and lance field.

---


Xuanwo:
> Nice catch. Do we want at least one test at the python level?

on my way

PR #5541, opened 2025-12-18
fix: merge_insert uses full schema path for reordered columns

### LINKED ISSUE #5323: When all columns provided out-of-order, we should use full-schema merge_insert
Right now if you provide columns out-of-order, it ends up using the partial schema path. We should re-order the columns and use the full schema path.

wjones127:
## Summary

- When `merge_insert` received columns in a different order than the dataset schema, it fell back to the slower partial schema path
- Added `ignore_field_order: true` to the schema comparison so reordered columns can use the efficient full schema path
- Added test verifying both the fast path eligibility and data correctness with reordered columns

Fixes #5323

### DIFF
--- a/rust/lance/src/dataset/write/merge_insert.rs
+++ b/rust/lance/src/dataset/write/merge_insert.rs
@@ -1429,6 +1429,8 @@ impl MergeInsertJob {
                 compare_metadata: false,
                 // Allow nullable source fields for non-nullable targets.
                 compare_nullability: NullabilityComparison::Ignore,
+                // Allow columns to be in a different order; they will be matched by name.
+                ignore_field_order: true,
                 ..Default::default()
             },
         );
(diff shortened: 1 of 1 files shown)


rust/lance/src/dataset/write/merge_insert.rs
```diff
@@ -1402,6 +1402,8 @@ impl MergeInsertJob {
                 compare_metadata: false,
                 // Allow nullable source fields for non-nullable targets.
                 compare_nullability: NullabilityComparison::Ignore,
+                // Allow columns to be in a different order; they will be matched by name.
+                ignore_field_order: true,
```
jackye1995, line 1433: in what case would we not want this behavior?

wjones127: It doesn't seem like we assert field order anywhere right now, but it seems like a useful capability. `ArrowSchema` `PartialEq` implementation is strict on field order. There are places like in query plans where field order matters.

PR #5545, opened 2025-12-18
feat(format): support recording list of index files and sizes

wjones127:
Add file sizes to `IndexMetadata` so we can optimize cold reads of indexes (avoid HEAD calls) and report index on-disk size without extra IO.

For discussion, see: https://github.com/lance-format/lance/discussions/5456
For implementation, see: https://github.com/lance-format/lance/pull/5497

**This will require a vote before it can be merged.**


protos/table.proto
```diff
@@ -280,6 +280,20 @@ message IndexMetadata {
...
+  // If this is empty, the index files sizes are unknown.
+  repeated IndexFile files = 10;
+}
+
+// Metadata about a single file within an index segment.
+message IndexFile {
+  // Path relative to the index directory (e.g., "index.idx", "auxiliary.idx")
+  string path = 1;
+  // Size of the file in bytes
+  uint64 size_bytes = 2;
```
westonpace, line 296: Should this be relative to the dataset URI and include a base_id?  Otherwise how will this work if there are multiple bases?

westonpace: Oh wait, I guess we are just saying that "a single index cannot be broken across multiple bases" which seems like a reasonable constraint?

wjones127:
Voted passed in https://github.com/lance-format/lance/discussions/5757

PR #5561, opened 2025-12-22
feat: add describeIndex method for detailed index metadata retrieval

### LINKED ISSUE #5553: Support `describeIndex` in Dataset API
Since the Lance REST spec supports describing the details of an index for Lance, we'd better try to support this feature in the Dataset API.

yuqi1129:
This pull request introduces a new "describe index" feature across the Rust, Java, and Python APIs, allowing users to retrieve detailed metadata and statistics about a specific index by name. The change includes the addition of a new `IndexDescription` class in Java, updates to the JNI and Rust layers to support the new API, and corresponding tests and documentation in all three languages.

The most important changes are:

**API Additions:**

* Added a `describeIndex(String indexName)` method to the Java `Dataset` class, which returns a new `IndexDescription` object containing metadata and statistics for a specific index. This is supported by a new native JNI method and Rust FFI implementation. [[1]](diffhunk://#diff-04a9577f49fc3c004aa755fb15f03deb8865ce55e997a64a7a27e03a9ea86e07R985-R1005) [[2]](diffhunk://#diff-b3c4207806ed8561d35f7940d9ad9d4535be8861b774165ff2bb69b7a54b1f79R2505-R2649)
* Introduced a `describe_index(self, index_name: str)` method to the Python `LanceDataset` class, providing similar functionality for Python users. [[1]](diffhunk://#diff-3c0e591b6a64801b9e3f92380ab99c1546e03faaa769b6de4b37af58d95c588cR672-R714) [[2]](diffhunk://#diff-31492cb9a22ac40a820049c18a44ac88f651a5d0d7115fc09386899ab95f0c10R2653-R2668)
* Added a `describe_index` method to the Rust `DatasetIndexExt` trait, enabling retrieval of a single index's metadata without loading the full index.

**New Data Structures:**

* Implemented a new `IndexDescription` class in Java to encapsulate index metadata (type, distance metric, indexed/unindexed row counts) with a builder pattern for construction.

**Testing and Validation:**

* Added comprehensive tests for the new describe index functionality in Rust, Java, and Python to ensure correct behavior and error handling when describing existing and non-existent indices. [[1]](diffhunk://#diff-5377e74fdf73c8319af865fbcccd11b1276644c6afde7e1853e9a0995c943edfR2467-R2546) [[2]](diffhunk://#diff-9ea36460f24d3b368f9b6ea9ae43f9b79515089fb911438720b17309922f1e51R367-R402) [[3]](diffhunk://#diff-a95edaddaa3a260e498c04e10f073261bdc529cd4f47b928ad80274754af0548R4197-R4239)

**Documentation and Imports:**

* Updated imports and documentation to reflect the new API and data structures in both Java and Python. [[1]](diffhunk://#diff-04a9577f49fc3c004aa755fb15f03deb8865ce55e997a64a7a27e03a9ea86e07R21) [[2]](diffhunk://#diff-9ea36460f24d3b368f9b6ea9ae43f9b79515089fb911438720b17309922f1e51R17)

These changes make it much easier for users to programmatically inspect the properties and coverage of individual indices in a Lance dataset.

Fixed: #5553

### DIFF
--- a/rust/lance-index/src/traits.rs
+++ b/rust/lance-index/src/traits.rs
@@ -255,6 +255,47 @@ pub trait DatasetIndexExt {
         criteria: Option<IndexCriteria<'b>>,
     ) -> Result<Vec<Arc<dyn IndexDescription>>>;
 
+    /// Describes a single index by name
+    ///
+    /// This is a convenience method that filters `describe_indices` to a single index.
+    ///
+    /// This method should only access the index metadata and should not load the index into memory.
+    ///
+    /// # Parameters
+    /// - `index_name`: the name of the index to describe
+    ///
+    /// # Returns
+    /// - `Ok(Some(description))`: if the index exists
+    /// - `Ok(None)`: if no index with the given name exists
+    /// - `Err(e)`: if there is an error loading index metadata
+    ///
+    /// # Examples
+    ///
+    /// ```rust,ignore
+    /// use lance::dataset::Dataset;
+    /// use lance_index::DatasetIndexExt;
+    ///
+    /// # async fn example() -> lance_core::Result<()> {
+    /// let dataset = Dataset::open("path/to/dataset").await?;
+    /// 
+    /// if let Some(desc) = dataset.describe_index("my_index").await? {
+    ///     println!("Index name: {}", desc.name());
+    ///     println!("Index type: {}", desc.index_type());
+    ///     println!("Rows indexed: {}", desc.rows_indexed());
+    /// } else {
+    ///     println!("Index not found");
+    /// }
+    /// # Ok(())
+    /// # }
+    /// ```
+    async fn describe_index(
+        &self,
+        index_name: &str,
+    ) -> Result<Option<Arc<dyn IndexDescription>>> {
+        let indices = self.describe_indices(None).await?;
+        Ok(indices.into_iter().find(|idx| idx.name() == index_name))
+    }
+
     /// Loads a specific index with the given index name.
     async fn load_scalar_index<'a, 'b>(
         &'a self,
--- a/rust/lance-index/src/traits.rs
+++ b/rust/lance-index/src/traits.rs
@@ -291,10 +291,7 @@ pub trait DatasetIndexExt {
     async fn describe_index(
         &self,
         index_name: &str,
-    ) -> Result<Option<Arc<dyn IndexDescription>>> {
-        let indices = self.describe_indices(None).await?;
-        Ok(indices.into_iter().find(|idx| idx.name() == index_name))
-    }
+    ) -> Result<Option<Arc<dyn IndexDescription>>>;
 
     /// Loads a specific index with the given index name.
     async fn load_scalar_index<'a, 'b>(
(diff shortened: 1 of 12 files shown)


yuqi1129:
Can anyone help with what the problem is?
```text
Auto-detected mode: agent for event: pull_request
Using provided GITHUB_TOKEN for authentication
Checking permissions for actor: yuqi1129
Permission level retrieved: read
Warning: Actor has insufficient permissions: read
Error: Prepare step failed with error: Actor does not have write permissions to the repository
Error: Process completed with exit code 1.
``` 

What should I do to solve the problem?

yuqi1129:
@majin1102 @yanghua could you please help review this one?

rust/lance-index/src/traits.rs
```diff
@@ -255,6 +255,47 @@ pub trait DatasetIndexExt {
...
+    /// }
+    /// # Ok(())
+    /// # }
+    /// ```
+    async fn describe_index(
+        &self,
+        index_name: &str,
+    ) -> Result<Option<Arc<dyn IndexDescription>>> {
+        let indices = self.describe_indices(None).await?;
+        Ok(indices.into_iter().find(|idx| idx.name() == index_name))
```
majin1102: It seems we still need to load the full index metadata — likely because the underlying protobuf spec doesn’t support partial loading. I’m not sure this API is worth exposing unless we have solid use cases that `describe_indices` not suitable

majin1102 (commented):
Thanks for working on this. Left one comment

rust/lance-index/src/traits.rs
yuqi1129: We are going to push Lance's REST APIs, see: https://lance.org/format/namespace/integrations/gravitino/#option-1-native-lance-rest-support into practice. Indexes are also key components for the Lance table, so I raise it to support more metadata operations.

About `It seems we still need to load the full index metadata`, let me check whether we can optimize it.

majin1102: > We are going to push Lance's REST APIs, see: https://lance.org/format/namespace/integrations/gravitino/#option-1-native-lance-rest-support

That's good to hear. 

I’m not sure but I’m not opposed to it.
Thanks for this contribution

wjones127:
It's not obvious there's much benefit to a separate implementation from `describe_indexes()`. It doesn't do any less IO, unless I'm misunderstanding. So it's likely not worth the extra complexity.

If you really want this API, could we just make it do filter the output of `describe_indexes()`?

rust/lance/src/session/index_caches.rs
```diff
@@ -103,6 +103,20 @@ impl CacheKey for IndexMetadataKey {
...
+}
+
+impl CacheKey for IndexMetadataByNameKey<'_> {
+    type ValueType = Vec<IndexMetadata>;
+
+    fn key(&self) -> Cow<'_, str> {
+        Cow::Owned(format!("{}:{}", self.version, self.name))
+    }
+}
+
```
wjones127, line 119: We already cache the full list of index metadata, and often do so opportunistically when we open the dataset. This is one of the reasons I think you should just re-use the `describe_indexes()` function for this.

wjones127: To be specific, here is an example tests showing that we cache the full list of index metadata:

https://github.com/lance-format/lance/blob/fab656ab6dbcc48cffcfe1cc3b612e5057a63c73/rust/lance/src/index.rs#L2756-L2760

PR #5565, opened 2025-12-23
refactor: support java 21, drop java 8

cmccabe:
Drop the Java 8 build from CI. Java 8 is more than 10 years old, and has been end-of-life (EOL) since 2019. It is no longer receiving security updates and other fixes upstream.

Change the release to build with Java 11, since this is now the lowest Java version we support.

Add a Java 21 build to CI. Java 21 was released more than 2 years ago and is an LTS release that will probably be used for some time to come.

In the Java code, upgrade Apache Arrow from version 15.0.0 to version 18.3.0. The new version supports Java 21, and does not support Java 8.

### DIFF
--- a/java/AGENTS.md
+++ b/java/AGENTS.md
@@ -10,4 +10,4 @@ lint rust: `cargo clippy --tests --manifest-path ./lance-jni/Cargo.toml`
 compile: `./mvnw compile`
 test: `./mvnw test`
 
-JDK: pom.xml targets Java 8 (`maven.compiler.release` 8); align Rust toolchain with repository `rust-toolchain.toml`.
+JDK: pom.xml targets Java 11 (`maven.compiler.release` 11); align Rust toolchain with repository `rust-toolchain.toml`.
--- a/java/pom.xml
+++ b/java/pom.xml
@@ -419,8 +419,8 @@
                 <jdk>[11,)</jdk>
             </activation>
             <properties>
-                <!-- Ping release target to JDK8 to link only against Java 8 APIs -->
-                <maven.compiler.release>8</maven.compiler.release>
+                <!-- Ping release target to JDK11 to link only against Java 11 APIs -->
+                <maven.compiler.release>11</maven.compiler.release>
             </properties>
             <build>
                 <plugins>


eddyxu (approved):

cmccabe:
Hmm, it looks like the new Apache Arrow is not compatible with jdk8. I don't think there is a single Apache Arrow version that is compatible with both JDK8 and JDK21. Since the former is EOL now according to Oracle, I think we should just drop it going forward. (It has actually been EOL since 2019, except for people with support contracts with Oracle.)

What do you think @westonpace @jackye1995 @mackrorysd @eddyxu ?

yanghua:
Hi @cmccabe, since there may be conflicts between different JDK versions. WDYT about introducing the maven **profile** as a solution to solve this problem?

cmccabe:
> Hi @cmccabe, since there may be conflicts between different JDK versions. WDYT about introducing the maven profile as a solution to solve this problem?

I don't think Maven profiles solve the problem, because the issue is that the java ecosystem is leaving java 8 behind. There isn't ever going to be a new version of Apache Arrow that supports java 8, and the same is true for many other libraries. Meanwhile, jdk11, jdk17, and jdk21 don't require profiles since the same code can work for all of them.

Java 11 is already pretty old, and to be honest I think it would not be a good idea to deploy it in 2025. But it's at least receiving security updates. Java 8 is not. It was discontinued before the lance project even had its first release.

yanghua:
@cmccabe Want to make sure if you are looking forward to having this feature? https://github.com/lance-format/lance-spark/pull/160

yanghua:
> I don't think Maven profiles solve the problem, because the issue is that the java ecosystem is leaving java 8 behind. There isn't ever going to be a new version of Apache Arrow that supports java 8.

Let's focus on this issue. Currently, the Java module works fine for Java 8, before this PR, right?

mackrorysd:
> What do you think @westonpace @jackye1995 @mackrorysd @eddyxu ?

I think this seems very reasonable. We should definitely communicate this quickly in case there are people that will be affected by the change so they can plan for the impact.

cmccabe:
> @cmccabe Want to make sure if you are looking forward to having this feature? https://github.com/lance-format/lance-spark/pull/160

Yes, that is related. However, we must commit this PR before we can commit https://github.com/lance-format/lance-spark/pull/160, because otherwise it will not be possible to use JDK21 with Lance. There is a bug in older versions of Apache Arrow that blocks it. See https://github.com/apache/arrow/issues/35053

(There may also be other bugs in older versions of Apache Arrow as well, of course. But this is the really bad one we know about.)

> Let's focus on this issue. Currently, the Java module works fine for Java 8, before this PR, right?

Probably. But the newest versions of Apache software have dropped support for jdk8. For example, you cannot run jdk8 with the latest Apache Spark or Apache Kafka. So I don't think we're getting very much testing on this anyway.

In general if you want to argue for supporting JDK6 and JDK8, I think that argument would have to be made in the Apache Arrow project. Since they're our biggest dependency and freezing that at a very old version is not really an option for us.

> I think this seems very reasonable. We should definitely communicate this quickly in case there are people that will be affected by the change so they can plan for the impact.

Yeah, it should definitely be in the release notes.

In general, if you're using historical versions of everything else (Spark, Java, etc.), using an older version of Lance too is probably fine.

jackye1995:
@yanghua My understanding is that this is safe for java8, because we set https://github.com/lance-format/lance/blob/main/java/pom.xml#L423

So the Java syntax still needs to be java8 compatible, and the released binary is still showing java8 as the version, just we won't build it in ci. 

Let me know if my understanding is wrong.

cmccabe:
> @yanghua My understanding is that this is safe for java8, because we set https://github.com/lance-format/lance/blob/main/java/pom.xml#L423
>
> So the Java syntax still needs to be java8 compatible, and the released binary is still showing java8 as the version, just we won't build it in ci.
>
> Let me know if my understanding is wrong.

Unfortunately, newer versions of the Apache Arrow library that we depend on are built with JDK11 as the minimum version. So you will not be able to use JDK8 once this PR is merged.

We don't have much choice unless we want to stay on an old version of Arrow. But people want the features that newer versions of Apache Arrow have, such as JDK21 support. (Smart operators have already moved to JDK21 to save money, since there are some efficiency improvements in newer versions.)

cmccabe:
> we set https://github.com/lance-format/lance/blob/main/java/pom.xml#L423

As long as we're moving to jdk11 for the release, let me bump this to 11 too. We might as well get the new features in 11 if we're losing support for 8.

jackye1995:
I see. In that case I agree we should bump to 11. It would be nice if we can still keep java8 syntax so that people who need it can build a jar using java8 on their own, and also manually override the arrow version.

PR #5566, opened 2025-12-23
feat: support credentials vending in directory namespace

jackye1995:
This PR introduces the credentials vending feature to the namespace impl, allowing us to vend credentials if we run directory namespace, or run it as backend for rest namespace. This would allow us to fully test the credentials vending code path end to end.

The actual vending logic mainly consults the same feature implemented in Apache Polaris. The support covers aws, gcp and azure.

### DIFF
--- a/rust/lance-namespace-impls/src/credentials/azure.rs
+++ b/rust/lance-namespace-impls/src/credentials/azure.rs
@@ -1,0 +1,335 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright The Lance Authors
+
+//! Azure credential vending using SAS tokens.
+//!
+//! This module provides credential vending for Azure Blob Storage by generating
+//! SAS (Shared Access Signature) tokens with user delegation keys.
+
+use std::collections::HashMap;
+use std::sync::Arc;
+
+use async_trait::async_trait;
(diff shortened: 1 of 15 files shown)


Xuanwo (approved):

rust/lance-namespace-impls/src/credentials/azure.rs
```diff
@@ -0,0 +1,335 @@
...
+
+//! Azure credential vending using SAS tokens.
+//!
+//! This module provides credential vending for Azure Blob Storage by generating
+//! SAS (Shared Access Signature) tokens with user delegation keys.
+
+use std::collections::HashMap;
+use std::sync::Arc;
+
+use async_trait::async_trait;
```
Xuanwo, line 12: can we use `reqsign` instead of all SDKs from aws, azure and gcp?

jackye1995: let me try

jackye1995: looks like reqsign currently cannot perform scoped credentials generation, so cannot really use it at this moment

Xuanwo: Got it, will take a look

PR #5567, opened 2025-12-23
feat: improve error type: part 1

camilesing:
for issue: https://github.com/lance-format/lance/issues/4603


rust/lance-table/src/io/deletion.rs
```diff
@@ -155,11 +155,7 @@ pub async fn read_deletion_file(
             let data = std::io::Cursor::new(data);
             let mut batches: Vec<RecordBatch> = ArrowFileReader::try_new(data, None)?
                 .collect::<std::result::Result<_, ArrowError>>()
-                .map_err(box_error)
-                .context(CorruptFileSnafu {
-                    path: path.clone(),
-                    location: location!(),
-                })?;
+                .map_err(|e| Error::corrupt_file(path.clone(), e.to_string(), location!()))?;
```
niebayes: Can we define Error enums and make it derive Snafu for implicitly conveying location?

wjones127 (commented):
@camilesing See my comment in https://github.com/lance-format/lance/issues/4603#issuecomment-3700920169

camilesing:
@wjones127 ok

PR #5571, opened 2025-12-24
fix: panic when lance.auto_cleanup.interval is set to 0

majin1102:
For now, when lance.auto_cleanup.interval is set to 0, dataset commits panic with a "division by zero" error:

```
thread 'dataset::cleanup::tests::test_auto_cleanup_interval_zero' panicked at rust/lance/src/dataset/cleanup.rs:672:12:
attempt to calculate the remainder with a divisor of zero
```

To fix the panic, we can interpret `lance.auto_cleanup.interval = 0` as triggering cleanup after each commit, which is equivalent to `lance.auto_cleanup.interval = 1`

### DIFF
--- a/rust/lance/src/dataset/cleanup.rs
+++ b/rust/lance/src/dataset/cleanup.rs
@@ -1460,7 +1460,11 @@ mod tests {
             "lance.auto_cleanup.retain_versions".to_string(),
             Some("1".to_string()),
         );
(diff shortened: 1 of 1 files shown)


majin1102:
Hi, @westonpace
Please let me know what do you think about the semantics of `lance.auto_cleanup.interval = 0`

PR #5572, opened 2025-12-24
perf: materialize the tokens after WAND done

BubbleCal:
this avoids copying tokens for each candidate during WAND search, these small strings could lead high memory usage if we have many partitions and the queries are with large limit.

verified on a 1M dataset
before:
Peak RSS delta: 497.09 MiB

after:
Peak RSS delta: 222.34 MiB

this also improves the FTS performance by ~10% for such queries

### DIFF
--- a/rust/lance-index/src/scalar/inverted/index.rs
+++ b/rust/lance-index/src/scalar/inverted/index.rs
@@ -276,9 +276,7 @@ impl InvertedIndex {
                     let mut tokens_by_position = vec![String::new(); postings.len()];
                     for posting in &postings {
                         let idx = posting.term_index() as usize;
-                        if idx < tokens_by_position.len() {
-                            tokens_by_position[idx] = posting.token().to_owned();
-                        }
+                        tokens_by_position[idx] = posting.token().to_owned();
                     }
                     let params = params.clone();
                     let mask = mask.clone();
--- a/rust/lance-index/src/scalar/inverted/wand.rs
+++ b/rust/lance-index/src/scalar/inverted/wand.rs
@@ -371,11 +371,7 @@ impl<'a, S: Scorer> Wand<'a, S> {
             _ => {}
         }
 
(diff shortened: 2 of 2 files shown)


rust/lance-index/src/scalar/inverted/index.rs
```diff
@@ -256,19 +271,30 @@ impl InvertedIndex {
...
                     if postings.is_empty() {
-                        return Ok(Vec::new());
+                        return Ok(PartitionCandidates::empty());
+                    }
+                    let mut tokens_by_position = vec![String::new(); postings.len()];
+                    for posting in &postings {
+                        let idx = posting.term_index() as usize;
+                        if idx < tokens_by_position.len() {
+                            tokens_by_position[idx] = posting.token().to_owned();
+                        }
```
Xuanwo: It's better to avoid writing this boundary check manually:

```rust
let mut tokens_by_position = vec![String::new(); postings.len()];
for p in &postings {
    tokens_by_position
        .get_mut(p.term_index() as usize)
        .inspect(|slot| **slot = p.token().to_owned());
}
```

rust/lance-index/src/scalar/inverted/wand.rs
```diff
@@ -359,7 +371,11 @@ impl<'a, S: Scorer> Wand<'a, S> {
             _ => {}
         }
 
-        let mut candidates = BinaryHeap::new();
+        let mut candidates = if limit <= BLOCK_SIZE * 10 {
+            BinaryHeap::with_capacity(limit)
+        } else {
+            BinaryHeap::new()
```
Xuanwo: Even if the limit is quite large, starting with `BinaryHeap::with_capacity(BLOCK_SIZE * 10)` is still a good approach.

rust/lance-index/src/scalar/inverted/index.rs
BubbleCal: it won't be out of bound here, removed it

PR #5577, opened 2025-12-24
fix(python): correct type hint for to_tensor_fn parameter

### LINKED ISSUE #3129: lance.torch.data.LanceDataset(to_tensor_fn=...) typehint inconsistent with usage.
Hi, thanks for creating this amazing library!

There seems to be a small issue with the `lance.torch.data.LanceDataset` in the `to_tensor_fn` kwarg of the initializer. Its type hint is:
```python
to_tensor_fn: Optional[
    callable[[pa.RecordBatch], Union[dict[str, torch.Tensor], torch.Tensor]]
] = None,
```
However, it is called [here](https://github.com/lancedb/lance/blob/f257489894b158ae29686f53b9778d3441bebeba/python/python/lance/torch/data.py#L313) with an additional kwarg:
```python
batch = self._to_tensor_fn(batch, hf_converter=self._hf_converter)
```

AndreaBozzo:
Fixes #3129

### DIFF
--- a/python/python/lance/torch/data.py
+++ b/python/python/lance/torch/data.py
@@ -11,7 +11,16 @@ import logging
 import math
 import warnings
 from pathlib import Path
-from typing import Any, Callable, Dict, Iterable, List, Literal, Optional, Union
+from typing import (
+    Any,
+    Dict,
+    Iterable,
+    List,
+    Literal,
+    Optional,
+    Protocol,
+    Union,
+)
 
 import pyarrow as pa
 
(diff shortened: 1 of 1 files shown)


python/python/lance/torch/data.py
```diff
@@ -193,7 +193,9 @@ def __init__(
         shard_granularity: Optional[Literal["fragment", "batch"]] = None,
         batch_readahead: int = 16,
         to_tensor_fn: Optional[
-            Callable[[pa.RecordBatch], Union[dict[str, torch.Tensor], torch.Tensor]]
+            Callable[
+                ..., Union[dict[str, torch.Tensor], torch.Tensor]
+            ]
```
wjones127: It seems wrong to just make this more generic. How about this?

```python
Callable[[pa.RecordBatch, ...], Union[dict[str, torch.Tensor], torch.Tensor]]
```

Alternatively, can be fancier and do:

```python
from typing import Protocol

class ToTensorFn(Protocol):
    def __call__(self, batch: pa.RecordBatch, hf_converter, use_blob_api: bool) -> Union[dict[str, torch.Tensor], torch.Tensor]: ...

...
to_tensor_fn: ToTensorFn
```

wjones127 (changes requested):
Can we fix the type hint by making it correct *and* specific?

python/python/lance/torch/data.py
AndreaBozzo: Thanks for pointing this out, i agree about the hint being too generic. 
I modified it with the ' less fancy' approach + ran lint in the last commit

AndreaBozzo: I thought about your comment and this is probably the cleanest way i can come up with, what do you think?

PR #5581, opened 2025-12-25
feat(java): add builder-style scalar index params

wojiaodoubao:
Introduce some builder classes to make creating scalar index easy. For example.

```java
  ScalarIndexParams params = BTreeIndexParams.builder()
      .zoneSize(2048)
      .build();

  IndexParams indexParams = IndexParams.builder().setScalarIndexParams(scalarParams).build();

   // Create BTree index on 'id' column
   dataset.createIndex(
       Collections.singletonList("id"),
       IndexType.BTREE,
       Optional.of("btree_id_index"),
       indexParams,
       true);
```

### DIFF
--- a/java/src/main/java/org/lance/index/scalar/InvertedIndexParams.java
+++ b/java/src/main/java/org/lance/index/scalar/InvertedIndexParams.java
@@ -13,8 +13,12 @@
  */
 package org.lance.index.scalar;
 
+import org.lance.util.JsonUtils;
+
+import com.google.common.base.Preconditions;
+
 import java.util.ArrayList;
-import java.util.LinkedHashMap;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
@@ -234,7 +238,7 @@ public final class InvertedIndexParams {
 
     /** Build a {@link ScalarIndexParams} instance for an inverted index. */
     public ScalarIndexParams build() {
-      Map<String, Object> params = new LinkedHashMap<>();
+      Map<String, Object> params = new HashMap<>();
       if (baseTokenizer != null) {
         params.put("base_tokenizer", baseTokenizer);
       }
@@ -266,6 +270,11 @@ public final class InvertedIndexParams {
         params.put("min_ngram_length", minNgramLength);
       }
       if (maxNgramLength != null) {
+        Preconditions.checkArgument(
+            minNgramLength == null || maxNgramLength >= minNgramLength,
+            "maxNgramLength {} shouldn't less than minNgramLength {}",
+            maxNgramLength,
+            minNgramLength);
         params.put("max_ngram_length", maxNgramLength);
       }
       if (prefixOnly != null) {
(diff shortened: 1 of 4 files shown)


wojiaodoubao:
Hi @BubbleCal , @majin1102  could you help review this when you have time, thanks very much~

java/src/main/java/org/lance/index/scalar/JsonUtils.java
```diff
@@ -0,0 +1,33 @@
...
+ * limitations under the License.
+ */
+package org.lance.index.scalar;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+import java.util.Map;
+
+final class JsonUtils {
```
majin1102: I think we could put this into util package. WDYT

java/src/main/java/org/lance/index/scalar/ZoneMapIndexParams.java
```diff
@@ -0,0 +1,66 @@
...
+   *
+   * @return a new {@link Builder}
+   */
+  public static Builder builder() {
+    return new Builder();
+  }
+
+  public static final class Builder {
+    private Long rowsPerZone;
+    private final Map<String, Object> extraParams = new LinkedHashMap<>();
```
majin1102: 1. Do we need a set for this parameter?
2. Why not Map<String, String>?

java/src/main/java/org/lance/index/scalar/InvertedIndexParams.java
```diff
@@ -0,0 +1,282 @@
...
+      }
+      if (asciiFolding != null) {
+        params.put("ascii_folding", asciiFolding);
+      }
+      if (minNgramLength != null) {
+        params.put("min_ngram_length", minNgramLength);
+      }
+      if (maxNgramLength != null) {
+        params.put("max_ngram_length", maxNgramLength);
+      }
```
majin1102, line 279: I think we should probably verify minNgramLength <= maxNgramLength in the beginning

java/src/main/java/org/lance/index/scalar/InvertedIndexParams.java
```diff
@@ -263,9 +267,17 @@ public ScalarIndexParams build() {
         params.put("ascii_folding", asciiFolding);
       }
       if (minNgramLength != null) {
+        Preconditions.checkArgument(minNgramLength >= 0, "minNgramLength shouldn't be negative");
         params.put("min_ngram_length", minNgramLength);
       }
       if (maxNgramLength != null) {
+        Preconditions.checkArgument(maxNgramLength >= 0, "maxNgramLength shouldn't be negative");
+        Preconditions.checkArgument(
+            maxNgramLength >= minNgramLength,
```
majin1102: This is not safe since minNgramLengh could be null

java/src/main/java/org/lance/index/scalar/InvertedIndexParams.java
```diff
@@ -263,9 +267,17 @@ public ScalarIndexParams build() {
...
+            "maxNgramLength {} shouldn't less than minNgramLength {}",
+            maxNgramLength,
+            minNgramLength,
+            minNgramLength);
```
majin1102, line 277: nit: This minNgramLength is redundant

majin1102 (approved):

PR #5582, opened 2025-12-26
feat: support merge insert with duplicated rows in source

jackye1995:
So far merge_insert requires source table to have no duplicated rows, because if 2 rows match a row in target, it is ambiguous which one to merge into target.

This PR introduces 2 concepts: `dedupe_by` which allows user to specify a column (must be primitive type) in the source to dedupe duplicate rows, and `dedupe_sort_options` which allows user to specify which rows take precedence. The ordering can be increasing or decreasing, null first or last (similar to arrow `SortOptions`), the actual ordering of the data is based on DataFusion semantics.

When running merge insert with these 2 parameters set, during the merge phase, we decide which row to merge using the value of the dedupe column, and based on the dedupe order, decide which one to keep. If there are multiple rows of the same dedupe value, we will fail the operation since again it is ambiguous which one to merge into target.

For the actual execution, I chose to do a running dedupe during merge, rather than doing an aggregation to get the row to merge during the join phase, to avoid multiple passes against the source dataset.

The implementation also works for nested field as dedupe column, also works if the dedupe column only exists in source not in target (which is the actual main use case for merging MemTable to underlying table)

Note: this feature is for merge_insert v2 only.


wjones127:
This adds a lot of complexity to merge_insert. Have you considered instead providing a `dedup_use_first` option, where you just use the first value seen? Then I think we can re-use the existing row id capture logic (currently used to raise an error) and instead have it ignore rows if we've already seen the id.

I think this could serve the same use cases, just need to make the user sort their data in order they want before inserting. Using DataFusion, you can easily do sorting with spilling, making it efficient and the RAM use bound. And in some case, I bet you could cheaply provide data sorted in the way you want. For example, if you had data already in a LSM structure, you could provide data in reverse order of LSM segments.

PR #5583, opened 2025-12-26
feat: implemented tokenizer plugins in rust

KazuhitoT:
related to https://github.com/lance-format/lance/issues/3222

Key changes are as bellows:

* `include/lance_tokenizer_plugin.h`: add C API for the tokenizer plugin
* `protos/index_old.proto`: add two fields to restore the plugin tokenizer configuration
* `rust/lance-index/src/scalar/inverted/tokenizer.rs` and `rust/lance-index/src/scalar/inverted/plugin/*`: implement tokenizer loading
* `rust/lance-index/examples/`: add an example usage

During the PR creation process, I had two questions and left comments in the PR.


rust/lance-index/src/scalar/inverted/tokenizer/plugin/tokenizer.rs
```diff
@@ -0,0 +1,270 @@
...
+    }
+
+    fn create_stream<'a>(&'a mut self, text: &'a str) -> BoxTokenStream<'a> {
+        // Note: This is not the most efficient approach for repeated tokenization,
+        //       but it ensures thread safety and simplifies lifetime management.
+        //       For production use, consider caching the factory/tokenizer.
+        let stream = PluginTokenStreamAdapter::new(Arc::clone(&self.library), &self.config, text);
+        BoxTokenStream::new(stream)
+    }
+}
```
KazuhitoT, line 54: Creating the `factory` is expensive because it loads a tokenizer plugin (~*MB).
Should we cache the `factory` only, or cache both the `factory` and the `tokenizer`?

rust/lance-index/src/scalar/inverted/tokenizer.rs
```diff
@@ -282,7 +299,22 @@ impl InvertedIndexParams {
...
+        self.tokenizer_plugin_config = config;
+        self
+    }
+
     pub fn build(&self) -> Result<Box<dyn LanceTokenizer>> {
+        // Plugin tokenizer is handled separately as it returns LanceTokenizer directly
+        if self.base_tokenizer == "plugin" {
+            return self.build_plugin_tokenizer();
+        }
+
```
KazuhitoT: Since `build_plugin_tokenizer()` returns a `LanceTokenizer`, I added an early return instead of adding a new match condition in `build_base_tokenizer()`, which returns a `TextAnalyzerBuilder`.
Should I unify this branching logic?

KazuhitoT:
@wjones127 I have opened a draft PR https://github.com/lance-format/lance/pull/5583 mentioned in https://github.com/lancedb/lancedb/pull/2855#issuecomment-3682427161.
Could you take a look whenever you have time?

include/lance_tokenizer_plugin.h
```diff
@@ -0,0 +1,113 @@
...
+typedef struct LanceToken {
+    /// Start and end byte offsets in the original text (UTF-8)
+    uint32_t offset_from;
+    uint32_t offset_to;
+
+    /// Position of this token in the sequence (0-indexed)
+    uint32_t position;
+    uint32_t position_length;
+
+    /// Pointer to the token text (null-terminated UTF-8)
```
wjones127: Do we want null terminated?

include/lance_tokenizer_plugin.h
wjones127, line 1: This is going to mostly be used with Rust implementations, so using null-terminated strings doesn't seem like the right choice to me. Rust will have to do a memcpy to convert `&str` / `String` into null terminated strings.

I think it might be worth just defining a struct for a string reference like:

```c
typedef struct StringReference {
    size_t start;
    size_t length;
}
```

Then you can make a trivial conversion between that and Rust's `&str`.

include/lance_tokenizer_plugin.h
```diff
@@ -0,0 +1,113 @@
...
+
+typedef struct LanceTokenizerPlugin {
+    uint32_t (*api_version)(void);
+
+    /// Create a tokenizer factory with the given JSON configuration.
+    ///
+    /// @param config_json JSON configuration string (UTF-8, null-terminated)
+    /// @param config_len Length of config_json in bytes (not including null terminator)
+    /// @return Factory handle, or NULL on error (call get_error for details)
+    LanceTokenizerFactory* (*create_factory)(const char* config_json, uint32_t config_len);
```
wjones127: 1. If you are going to parse config here, we should have some facility to pass back an error.
2. I'd honestly prefer we didn't use JSON here, and instead just had the caller make multiple calls to `set_config(const StringReference key, const StringReference value)`. That way the plugin doesn't require a JSON parsing library.

Maybe the API should be like:

```c
typedef struct Error {
    StringReference message;
} Error;

typedef struct LanceTokenizerPlugin {
    void (*create_factory)(LanceTokenizerFactory* out);
    void (*set_config)(const StringReference key, const StringReference value, Error* err);
    ...
} LanceTokenizerPlugin;
```

include/lance_tokenizer_plugin.h
```diff
@@ -0,0 +1,113 @@
...
+    /// @param token Output parameter - filled with token data if a token is available
+    /// @return 1 if a token was produced, 0 if no more tokens, negative on error
+    int32_t (*next_token)(LanceTokenStream* stream, LanceToken* token);
+
+    /// Get the last error message.
+    ///
+    /// @param factory Factory handle (can be NULL to get global/loading errors)
+    /// @return Error message (null-terminated), or NULL if no error
+    ///         The returned string is valid until the next error-generating call
+    const char* (*get_error)(LanceTokenizerFactory* factory);
```
wjones127: I think we should return errors as an `out` parameter. It seems like the `get_error` style doesn't work well with multiple threads, and we definitely want to run the tokenizers in multiple threads. They are the most expensive part of creating an FTS index.

rust/lance-index/src/scalar/inverted/tokenizer.rs
wjones127: I think so. One of the goals should be to be able to replace the Lindera and Jieba feature flags with plugins, without changing the output. So it seems like you'll want to place the base tokenizer inside of `build_base_tokenizer`.

rust/lance-index/src/scalar/inverted/tokenizer/plugin/tokenizer.rs
wjones127: Probably just cache the factory.

protos/index_old.proto
```diff
@@ -39,4 +39,6 @@ message InvertedIndexDetails {
   uint32 min_ngram_length = 9;
   uint32 max_ngram_length = 10;
   bool prefix_only = 11;
+  optional string tokenizer_plugin_library = 12;
+  optional string tokenizer_plugin_config = 13;
```
wjones127, line 43: This is sort of a change to the format, so we should probably open a design discussion on this. I can do that.

wjones127 (commented):
Thanks for working on this. This is close to what I had in mind. I have various minor suggestions.

protos/index_old.proto
wjones127: Could you add some comments to this part of the protobuf file? It should be clear to an implementor of a Lance library how these variables should be handled. How do we resolve the name of the dynamic library? What should the library do if the library is not found?

wjones127, line 43: Maybe make config a map? Or do we think we'll have nested structures in there?
```suggestion
  optional string tokenizer_plugin_library = 12;
  optional map<string, string> tokenizer_plugin_config = 13;
```

wjones127: Started a discussion here: https://github.com/lance-format/lance/discussions/5607

include/lance_tokenizer_plugin.h
KazuhitoT: Not anymore. As mentioned later, `CStringRef` allows zero-copy with Rust's `&str`.

include/lance_tokenizer_plugin.h
KazuhitoT: I've implemented your definition (named `CStringRef`) for zero-copy string passing between Rust and the plugin.

include/lance_tokenizer_plugin.h
KazuhitoT: At the initial implementation, I assumed the configuration format would be JSON.
However, I later realized that different plugins may require different formats—for example, jieba uses JSON, while lindera uses YAML.
Therefore, I decided that the configuration format should be defined by each plugin, and I left the configuration file loading behavior unchanged.
I also felt it was natural for clients to take responsibility if an error occurs while loading the tokenizer configuration.
What do you think about this approach?

include/lance_tokenizer_plugin.h
KazuhitoT: Thanks, I added `Error* error` arg.

protos/index_old.proto
KazuhitoT: Thanks for opening a discussion.
I added some comments in proto.

KazuhitoT: As mentioned in https://github.com/lance-format/lance/pull/5583#discussion_r2657694982, I kept `tokenizer_plugin_config` as a `string`.

KazuhitoT: I found your comment in discussion: 

> If the plugin dynamic library is not found, the index should be ignored for queries and not updated on writes. Users should get an error message when attempting to update the index that they need to install the correct plugin.

As commented in proto, the current implementation will throw an erorr if the dynamic library is not found.
However, this is a provisional implementation and will be revised according to the results of the discussion.

include/lance_tokenizer_plugin.h
wjones127: Good point about Jieba and Lindera. If they both already take string options and each in a different format, then just having a string configuration input makes sense. 👍

jackye1995:
@KazuhitoT sorry I think this PR fell off our radar for quite some time, would you still be up for finishing this up? Or we can also take over and complete it if you are busy

jackye1995:
@claude review

KazuhitoT:
@jackye1995 Thanks for checking in.

I’d like to continue working on this PR.
I’ll address the feedback from the Claude review and rebase it as soon as possible.

KazuhitoT:
@jackye1995 I've updated my code as follows.
Please take a look at your convenience.

* Reworked plugin tokenization to flow through the normal TextAnalyzer path (related to https://github.com/lance-format/lance/pull/5583#pullrequestreview-3613513294)
* Added validation for plugin metadata and absolute plugin library paths before build/persistence, rejecting relative paths.
* Improved robustness at the C ABI boundary by validating the API version, rejecting NULL callbacks, checking string length conversion, and propagating plugin errors consistently.
* Reworked factory/tokenizer/stream ownership to serialize access to C handles and enforce one active stream per tokenizer.
* Moved the example plugin from `rust/lance-index/examples` into test support.

The last item has changed from the summary at the top of my PR comment.

I think the comments look appropriate to me, but they may still be a bit verbose.
Please let me know if you think they should be shorter, and I can trim or remove unnecessary comments.

rust/lance-index/Cargo.toml
```diff
@@ -86,6 +87,7 @@ lance-testing.workspace = true
 test-log.workspace = true
 rstest.workspace = true
 chrono.workspace = true
+serial_test.workspace = true
```
KazuhitoT, line 90: for factory cache test in https://github.com/lance-format/lance/pull/5583/changes#diff-eec8222bf06324ad720a6f756ed90875b446e34026a436c18b097edabaaafe30

PR #5584, opened 2025-12-28
feat: parse Ref from version and version numbers

Xuanwo:
This PR allows users to parse a str into `Ref`.

For example:

```rust
("0", Ref::VersionNumber(0)),
("42", Ref::VersionNumber(42)),
("main", Ref::Version(None, None)),
("main:", Ref::Version(None, None)),
("main:latest", Ref::Version(None, None)),
("main:10", Ref::Version(None, Some(10))),
("feature/a", Ref::Version(Some("feature/a".to_string()), None)),
("feature/a:", Ref::Version(Some("feature/a".to_string()), None)),
("feature/a:latest", Ref::Version(Some("feature/a".to_string()), None)),
("feature/a:10", Ref::Version(Some("feature/a".to_string()), Some(10))),
("tag1", Ref::Tag("tag1".to_string()))
```

---

### DIFF
--- a/rust/lance/src/dataset/refs.rs
+++ b/rust/lance/src/dataset/refs.rs
@@ -280,6 +280,13 @@ impl Tags<'_> {
                 message: format!("tag {} already exists", tag),
             });
         }
+
+        let branch_file = branch_contents_path(&root_location.path, tag);
+        if self.object_store().exists(&branch_file).await? {
+            return Err(Error::RefConflict {
+                message: format!("tag {} conflicts with existing branch", tag),
+            });
+        }
         let tag_contents = self.build_tag_content_by_ref(reference).await?;
 
         self.object_store()
@@ -314,6 +321,13 @@ impl Tags<'_> {
         if !self.object_store().exists(&tag_file).await? {
             return Err(Error::RefNotFound {
                 message: format!("tag {} does not exist", tag),
+            });
+        }
+
+        let branch_file = branch_contents_path(&root_location.path, tag);
+        if self.object_store().exists(&branch_file).await? {
+            return Err(Error::RefConflict {
+                message: format!("tag {} conflicts with existing branch", tag),
             });
         }
         let tag_contents = self.build_tag_content_by_ref(reference).await?;
@@ -452,6 +466,13 @@ impl Branches<'_> {
 
         let source_branch = source_branch.and_then(standardize_branch);
         let root_location = self.refs.root()?;
+
+        let tag_file = tag_path(&root_location.path, branch_name);
+        if self.object_store().exists(&tag_file).await? {
+            return Err(Error::RefConflict {
+                message: format!("branch {} conflicts with existing tag", branch_name),
+            });
+        }
         let branch_file = branch_contents_path(&root_location.path, branch_name);
         if self.object_store().exists(&branch_file).await? {
             return Err(Error::RefConflict {
--- a/rust/lance/src/dataset/refs.rs
+++ b/rust/lance/src/dataset/refs.rs
@@ -25,7 +25,7 @@ use std::io::ErrorKind;
 pub const MAIN_BRANCH: &str = "main";
 
 /// Lance Ref
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, PartialEq, Eq)]
 pub enum Ref {
     // Version number points of the current branch
     VersionNumber(u64),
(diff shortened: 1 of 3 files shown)


wjones127:
I'd +1 the automated code review concern here: how do we differentiate from branches? We can maybe add some additional rules for what is a valid tag. It seems like this might be worth designing the API in an issue first before going into implementation.

There any background on this feature? Why did you decide to implement it?

Xuanwo:
> I'd +1 the automated code review concern here: how do we differentiate from branches? We can maybe add some additional rules for what is a valid tag. 

My current idea is to use logic similar to Git: we do not allow the same name for refs, including branches and tags. Therefore, for any given string, we only need to check whether it is a tag or a branch.

> It seems like this might be worth designing the API in an issue first before going into implementation.

I'm willing to start one if needed

> There any background on this feature? Why did you decide to implement it?

I'm working on duckdb's versioning feature for lance: https://github.com/lance-format/lance-duckdb/issues/88, for example:

```sql
-- latest version on branch
SELECT * FROM my_lance.default.t AT (VERSION => 'feat')
-- branch with version
SELECT * FROM my_lance.default.t AT (VERSION => 'feat:45')
-- tag
SELECT * FROM my_lance.default.t AT (VERSION => 'v1.0')

SELECT * FROM my_lance.default.t AT (TIMESTAMP => TIMESTAMP '2025-09-22 12:32:43.217')
```

Note that DuckDB does not allow extending the `AT` clause to support `BRANCH` or `TAG`. Therefore, I plan to accept a `Ref` here instead. Users can pass any valid Ref to read as a version.

hamersaw:
[Linking a similar Spark PR](https://github.com/lance-format/lance-spark/pull/198) as related discussions.

IMO I think requiring the implementation to discern whether a string passed as version is a branch or a tag presents some ambiguity and an unnecessary restriction that you can not have a branch and a tag with the same name -- something we don't enforce today IIUC.

The proposal on Spark-side is to extend the table identifier to include a branch. So from your examples, something like:

```
-- branch 'feat' with version '45'
SELECT * FROM my_lance.default.t.b_feat AT (VERSION => '45')
```

Looking into [duckdb statments](https://duckdb.org/docs/stable/sql/introduction) I would make a similar argument here that our solution to support branches should span further than `SELECT`. Users will want to perform all operations on specific branches (ex. `INSERT`, `CREATE INDEX`, `UPDATED`, etc), so we would be building different branch support for those operations later anyways.

rust/lance/src/dataset/refs.rs
```diff
@@ -793,6 +855,41 @@ mod tests {
 
     use rstest::rstest;
 
+    #[rstest]
+    fn test_parse_ref_ok(
+        #[values(
+            ("0", Ref::VersionNumber(0)),
```
jackye1995, line 861: Could we add a `ref/` prefix before the branch or tag, so that we can differentiate a version `100` vs a branch/tag `"100"`?

jackye1995: Another thought I proposed in Spark is that `ref/` could indicate the delimiter is `/`. That means if we do `ref/branch/tag` then we know to use `/` to split branch and tag.

If a branch has a `/` (which is common), we an use a different delimiter, e.g. `ref:feature/abc:100`.

Basically the first 3 characters equaling `ref` indicates its not a version number, and the forth character is used as the delimiter.

What do we think? @Xuanwo @wjones127 @hamersaw

jackye1995:
> The proposal on Spark-side is to extend the table identifier to include a branch.

I think we definitely don't want to do another level of nesting like `my_lance.default.t.b_feat`, anything that is not 3 level is just causing additional incompatibility.

If we use the `ref` for tags that I proposed, one way is to do `my_lance.default.t__ref/main/10`, so we can match the keyword `__ref` to know if a table is referencing a specific branch or tag.

With that, I think it would be convenient to also reserve `main` to always mean the main branch. I remember we have thought about that but have not fully decided if we want to go with it or not. @majin1102 curious what you think.

rust/lance/src/dataset/refs.rs
hamersaw: Sounds great to me! Just a few clarifying questions:
- Do we need to prepend all tags with branches when using `ref` then? ex. with `ref/foo` is "foo" a branch or a tag (ie. tag "foo" on branch "main")?
- How do we handle ambiguity if a branch / version / tag is specified in both the table ID and in the `VERSION AS OF` clause? Just error out?

I've been trying to convince myself that it would be more simple (parsing + API) to only support branches in table IDs and versions / tags in `VERSION AS OF` clauses (within `SELECT` and `CREATE TAG` statements), but that's probably a bad idea if we parse things into our existing `Ref` type (which contains branch, version, and tag already).

majin1102:
> With that, I think it would be convenient to also reserve main to always mean the main branch. I remember we have thought about that but have not fully decided if we want to go with it or not. @majin1102 curious what you think.

The "main" has been already reversed and it always points to the actual main branch. This behavior is guaranteed by `check_valid_branch()` and `standardize_branch()`.

> If we use the ref for tags that I proposed, one way is to do my_lance.default.t__ref/main/10, so we can match the keyword __ref to know if a table is referencing a specific branch or tag.

Totally agree with this approach about the syntax handling.

Back to this PR. I'm wondering if introducing complex string conversion rules at this low level is necessary. I've considered this before and have a few concerns:

- Flexibility: Different engines should be allowed to have their own syntax rules. Enforcing uniformity at the low level might sacrifice flexibility, even if we have a unified top-level design.
- API Complexity: This adds cognitive load (or understanding cost) to the ref API

There are alternative approaches, such as adding dedicated APIs or providing utility functions. But the conversion logic in this PR now looks fine to me.

rust/lance/src/dataset/refs.rs
```diff
@@ -45,10 +45,51 @@ impl From<u64> for Ref {
 
 impl From<&str> for Ref {
     fn from(reference: &str) -> Self {
+        let reference = reference.trim();
+        if let Some((branch, version)) = reference.split_once(':') {
+            if branch.is_empty() {
+                return Tag(reference.to_string());
+            }
+
+            let branch = if branch == MAIN_BRANCH {
```
majin1102, line 54: nit: `standardize_branch` was designed as the entry point for this scenario

hpvd:
this pull request was closed automatically by github action, while lance duckdb extension is waiting for it to get merged
https://github.com/lance-format/lance-duckdb/issues/88

PR #5587, opened 2025-12-29
feat: add classic alp encoding support

Xuanwo:
This PR will add classic alp encoding support

---

### DIFF
--- a/rust/lance-encoding/src/encodings/physical/alp.rs
+++ b/rust/lance-encoding/src/encodings/physical/alp.rs
@@ -131,10 +131,7 @@ fn sample_positions(num_values: usize, sample_size: usize) -> Vec<usize> {
 /// stable and fast while tracking the same storage layout used by actual chunk encoding.
 fn find_best_exponents_f32(values: &[f32]) -> Exponents {
     let positions = sample_positions(values.len(), 32);
-    let sample = positions
-        .into_iter()
-        .map(|i| values[i])
-        .collect::<Vec<_>>();
+    let sample = positions.into_iter().map(|i| values[i]).collect::<Vec<_>>();
 
     let mut best = Exponents { e: 0, f: 0 };
     let mut best_bytes = usize::MAX;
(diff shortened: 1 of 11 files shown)


rust/lance-encoding/src/encodings/physical/alp.rs
```diff
@@ -0,0 +1,1228 @@
...
+}
+
+/// Pick sample indices for exponent search.
+///
+/// ALP exponent selection is relatively expensive. For large pages we use a small deterministic
+/// sample (up to 32 values) to estimate the best `(e, f)` without scanning the whole input.
+fn sample_positions(num_values: usize, sample_size: usize) -> Vec<usize> {
+    if num_values <= sample_size {
+        return (0..num_values).collect();
+    }
```
wjones127, line 121: I wonder if you should have some value like `None` or empty `Vec` that should be interpreted as "sample all rows", that way you can avoid the allocation in this common case.

rust/lance-encoding/src/encodings/physical/alp.rs
```diff
@@ -0,0 +1,1228 @@
...
+        }
+
+        let n = usize::try_from(num_values)
+            .map_err(|_| Error::invalid_input("ALP chunk too large for usize", location!()))?;
+
+        let mut iter = data.into_iter();
+        let packed_deltas = iter.next().unwrap();
+        let header = iter.next().unwrap();
+        let exception_positions = iter.next().unwrap();
+        let exception_bits = iter.next().unwrap();
```
wjones127, line 935: nit: you should be able to do the length check and unpack the array in one go. The upshot of this is that the compile doesn't need to put a panic message in the binary for each call to `.next().unwrap()`.
```suggestion
        let [packed_deltas, header, exception_positions, exception_bits] =
            data.try_into().map_err(|_| Error::invalid_input(
                format!("ALP decompression expects 4 buffers, got {}", data.len()),
                location!(),
            ))?;

        let n = usize::try_from(num_values)
            .map_err(|_| Error::invalid_input("ALP chunk too large for usize", location!()))?;
```

wjones127 (commented):
This is exciting!

Have you benchmarked the throughput of this and done any profiling? The sampling of `e` and `f` seems pretty heavy, so maybe worth spending some time optimizing.

Also, could you run the `fuzz_test.rs` on this, and do it for a longer period to check correctness? I think that's just running against 2.1 right now.

Also, I'm not sure it's from this PR, but `cargo check -p lance-encoding --no-default-features` does currently fail due to some bitpacking code not being gated properly.

jja725:
@Xuanwo is there any progress regarding this? love to see this in lance

PR #5588, opened 2025-12-29
fix: avoid panic while hitting non-null empty multi-vector

Xuanwo:
This PR will avoid panic while hitting non-null empty multi-vector

---

### DIFF
--- a/rust/lance-linalg/src/distance.rs
+++ b/rust/lance-linalg/src/distance.rs
@@ -129,15 +129,14 @@ pub fn multivec_distance(
     }
 
     let mut dists = Vec::with_capacity(vectors.len());
-    for (row_idx, v) in vectors.iter().enumerate() {
+    for v in vectors.iter() {
         match v {
             None => dists.push(f32::NAN),
             Some(v) => {
                 let multivector = v.as_fixed_size_list();
                 if multivector.len() == 0 {
-                    return Err(ArrowError::InvalidArgumentError(format!(
-                        "vectors row {row_idx} is an empty list; multi-vector distances require at least one sub-vector"
-                    )));
+                    dists.push(f32::NAN);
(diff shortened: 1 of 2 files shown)


rust/lance-linalg/src/distance.rs
```diff
@@ -128,12 +128,19 @@ pub fn multivec_distance(
...
-            v.map(|v| {
+    let mut dists = Vec::with_capacity(vectors.len());
+    for (row_idx, v) in vectors.iter().enumerate() {
+        match v {
+            None => dists.push(f32::NAN),
+            Some(v) => {
                 let multivector = v.as_fixed_size_list();
-                match distance_type {
+                if multivector.len() == 0 {
+                    return Err(ArrowError::InvalidArgumentError(format!(
```
BubbleCal: can we just assume the distance to a empty multivector is NaN and then exclude it from the results?

BubbleCal (commented):
Where does it panic before this change?

Xuanwo:
> Where does it panic before this change?

Our tokio runtime worker panic at

https://github.com/lance-format/lance/blob/543eb8605bf2e258765e90aed51950fba53de385/rust/lance-index/src/vector/flat.rs#L147

Resulted by 

https://github.com/lance-format/lance/blob/543eb8605bf2e258765e90aed51950fba53de385/rust/lance-linalg/src/distance.rs#L142-L149

rust/lance-linalg/src/distance.rs
Xuanwo: Let me know if https://github.com/lance-format/lance/pull/5588/commits/0a617380e86b44b6e09c57a06bfb25f2e91a195c is what you want

BubbleCal (approved):

PR #5589, opened 2025-12-29
feat: add alter column nullable to non-nullable support

Xuanwo:
This PR will add alter column nullable to non-nullable support

---

### DIFF
--- a/rust/lance/src/dataset/schema_evolution.rs
+++ b/rust/lance/src/dataset/schema_evolution.rs
@@ -586,8 +586,6 @@ pub(super) async fn alter_columns(
         })?;
 
         if let Some(nullable) = alteration.nullable {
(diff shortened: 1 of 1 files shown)


wjones127:
+1 on the review comment. Would like to have this supported in nested columns.

wjones127 (changes requested):
Add a test for nested columns for this feature.

rust/lance/src/dataset/schema_evolution.rs
```diff
@@ -546,13 +589,7 @@ pub(super) async fn alter_columns(
             // TODO: in the future, we could check the values of the column to see if
             //       they are all non-null and thus the column could be made non-nullable.
```
wjones127: Might want to remove this TODO before merging though, as it looks to be complete.

PR #5591, opened 2025-12-29
fix: filter garbage entries from null maps during encoding

wkalt:
Null map entries can have non-zero length with garbage values that should be ignored. MapStructuralEncoder was passing all entries to the child encoder, but repdef only counted valid entries, which caused errors to occur when encoding Structs with Map values.

Add MapArrayExt trait (mirroring ListArrayExt) with filter_garbage_nulls() and trimmed_entries() methods, and use them in MapStructuralEncoder.


rust/lance-arrow/src/map.rs
```diff
@@ -0,0 +1,144 @@
...
+    fn filter_garbage_nulls(&self) -> Self;
+
+    /// Returns a copy of the map's entries array that has been sliced to size
+    ///
+    /// Similar to ListArrayExt::trimmed_values, but for Map arrays.
+    fn trimmed_entries(&self) -> Arc<dyn Array>;
+}
+
+impl MapArrayExt for MapArray {
+    fn filter_garbage_nulls(&self) -> Self {
```
wjones127: Physically, `MapArray` is a `ListArray`. Part of my wonders if this would be simpler to just do `let trimmed = ListArray::from(self.clone()).filter_garbage_nulls()` and the convert back into a `MapArray`.

wkalt: Thanks for the suggestion. Just updated.

PR #5592, opened 2025-12-29
feat: support indexed v2 merge insert

jackye1995:
Support leveraging an index for v2 merge insert, the index optimization is added in `MergeInsertPlanner` to transform a normal hash join into optimized join. With this change, we can properly display query plan. A few cases:

`on` column fully covered by index in target:

```
MergeInsert: on=[id], when_matched=UpdateAll, when_not_matched=InsertAll, when_not_matched_by_source=Keep...
  CoalescePartitionsExec...
    IndexedLookup: key=id, index=id_idx
      Replay...
        StreamingTableExec: partition_sizes=1, projection=[id, value]
```

Index only cover some fragments in target, run a hybrid plan that unions results and then hash join:

```
MergeInsert: on=[id], when_matched=UpdateAll, when_not_matched=InsertAll, when_not_matched_by_source=Keep...
  CoalescePartitionsExec...
    HashJoinExec...join_type=Left...
      Replay...
        StreamingTableExec: partition_sizes=1, projection=[id, value]
      ...UnionExec...
        IndexedLookup: key=id, index=id_idx...
          Replay...
            StreamingTableExec: partition_sizes=1, projection=[id, value]
        ...LanceScan...range=None
```


rust/lance/src/dataset/write/merge_insert/exec/indexed_lookup.rs
```diff
@@ -0,0 +1,470 @@
...
+/// Lookup only mode:
+/// - `projection_columns` from target
+/// - `_rowid`
+/// - `_rowaddr` (if with_row_addr)
+///
+/// Join mode:
+/// - Source columns
+/// - Target columns (including `_rowid`, `_rowaddr`)
+#[derive(Debug)]
+pub struct IndexedLookupExec {
```
jackye1995, line 70: I ended up wrapping this pipeline of source -> project -> MapIndexExec -> AddRowAddrExec -> TakeExec -> project as this execution node, since otherwise DF optimizer keeps adding additional repartitioning steps in between. Not sure if there is any better way.

rust/lance/src/dataset/write/merge_insert/exec/indexed_lookup.rs
```diff
@@ -0,0 +1,470 @@
...
+    /// Pipeline structure:
+    /// ```text
+    /// HashJoinExec (Left or Inner join, CollectLeft mode)
+    ///   source (ReplayExec) <- LEFT, collected into hash table
+    ///   target (lookup result) <- RIGHT, streams and probes
+    /// ```
+    ///
+    /// The source_replay must be a ReplayExec created externally. This ensures
+    /// the same Arc is used for both the exposed child and internal DAG.
+    fn build_join_pipeline_with_replay(
```
jackye1995, line 315: for fully indexed case, I ended up doing the join within this exec node so that it can more easily pass around the replay node.

rust/lance/src/dataset/write/merge_insert/exec/indexed_lookup.rs
wjones127: > otherwise DF optimizer keeps adding additional repartitioning steps in between

I think the DataFusion optimizer does this based on [ExecutionPlan::benefits_from_input_partitioning](https://docs.rs/datafusion/latest/datafusion/physical_plan/trait.ExecutionPlan.html#method.benefits_from_input_partitioning) and [ExecutionPlan::required_input_distribution](https://docs.rs/datafusion/latest/datafusion/physical_plan/trait.ExecutionPlan.html#method.required_input_distribution). It compares those settings with the children nodes input partitioning to decide whether to add repartitioning. So also worth making sure you are stating the partitioning correctly.

PR #5593, opened 2025-12-29
feat: support FixedSizeList<Struct>

wkalt:
Prior to this commit, FixedSizeList was only supported with primitive element types (e.g., `FSL<Float32>` for vectors). This adds structural encoding support for `FSL<Struct>`, enabling use cases like fixed-size arrays of bounding boxes, coordinate tuples, or other structured data.

Key changes:
- New `FixedSizeListStructuralEncoder` that encodes FSL validity to rep/def and delegates child encoding to the struct encoder
- New `StructuralFixedSizeListScheduler` that scales row ranges by the FSL dimension when scheduling reads
- New `StructuralFixedSizeListDecoder` that reconstructs FSL arrays from child data and rep/def validity

A key challenge is "garbage filtering": unlike variable-length lists which can omit children under null entries, FSL children always exist. When an FSL row is null, any nested list-like types within its children contain undefined "garbage" data. The encoder normalizes these to empty null lists before encoding.

### DIFF
--- a/rust/lance-encoding/src/encoder.rs
+++ b/rust/lance-encoding/src/encoder.rs
@@ -20,8 +20,8 @@ use arrow_schema::DataType;
 use bytes::{Bytes, BytesMut};
 use futures::future::BoxFuture;
 use lance_core::datatypes::{Field, Schema};
-use lance_core::utils::bit::{is_pwr_two, pad_bytes_to};
 use lance_core::error::LanceOptionExt;
+use lance_core::utils::bit::{is_pwr_two, pad_bytes_to};
 use lance_core::{Error, Result};
 use snafu::location;
 


wkalt:
sorry, mislabeled this breaking. Should not be breaking.

wkalt:
I will update this patch once https://github.com/lance-format/lance/pull/5591 is merged - there are some related tests I want to add here that currently fail.

westonpace:
Should this encoder require 2.2?  Our general rule of thumb has been "any file written by any version of 2.1 should be readable by any version of 2.1" which is perhaps stricter than normal backwards compatibility rules.  Since this allows you to create files that old 2.1 readers will not be able to read I think this will need to be a 2.2 feature.

Or maybe you have that check and I just missed it.

Also, does this handle `FixedSizeList<List<...>>` or is that a todo still?

wkalt:
`FixedSizeList<List>` is still unimplemented - there is some existing todo note about `List<List>` somewhere I believe, and the same constraints still apply.

Sounds like this should require 2.2, thanks for catching that. I'll push an update shortly.

wkalt:
@westonpace thanks, this is updated, I think the failures are unrelated.

wjones127:
> I think the failures are unrelated.

👍  https://github.com/lance-format/lance/issues/5646

rust/lance-core/src/datatypes/field.rs
```diff
@@ -165,6 +165,16 @@ impl Field {
...
                 DataType::LargeList(Arc::new(ArrowField::from(&self.children[0])))
             }
+            lt if lt.is_fixed_size_list_struct() => {
+                // Parse size from "fixed_size_list:struct:N"
+                let size: i32 =
+                    lt.0.split(':')
+                        .next_back()
+                        .expect("fixed_size_list:struct logical type missing size suffix")
+                        .parse()
+                        .expect("fixed_size_list:struct logical type has invalid size");
```
wjones127, line 175: Does this mean we can panic if we read a dataset that has a corrupt schema?

wjones127 (commented):
Have one concern about handle corrupt schemas, but otherwise looks good.

rust/lance-core/src/datatypes/field.rs
westonpace: This is the lance schema.  I'm not entirely sure users are able to create their own lance schema.  So I think the only way this could happen is if there was some kind of corrupt protobuf.  Also, there is a significant panic potential down below at `lt => DataType::try_from(lt).unwrap()`.

I suppose it is technically a valid concern but this method has many many callsites and changing it to result returning should probably be a PR on its own or else this one is going to get real confusing.

rust/lance-datagen/src/generator.rs
```diff
@@ -1712,6 +1712,84 @@ impl ArrayGenerator for RandomListGenerator {
     }
 }
 
+#[derive(Debug)]
+struct RandomMapGenerator {
```
westonpace, line 1717: Can you document this (or the `rand_map` method).  Just a brief comment somewhere to describe that it will randomly generate maps with 0-4 items.

rust/lance-datagen/src/generator.rs
```diff
@@ -1712,6 +1712,84 @@ impl ArrayGenerator for RandomListGenerator {
...
+        &mut self,
+        length: RowCount,
+        rng: &mut rand_xoshiro::Xoshiro256PlusPlus,
+    ) -> Result<Arc<dyn Array>, ArrowError> {
+        let lengths = self.lengths_gen.generate(length, rng)?;
+        let lengths = lengths.as_primitive::<Int32Type>();
+        let total_entries = lengths.values().iter().sum::<i32>() as u64;
+        let offsets = OffsetBuffer::from_lengths(lengths.values().iter().map(|v| *v as usize));
+
+        let keys = self.keys_gen.generate(RowCount::from(total_entries), rng)?;
```
westonpace, line 1762: Not really relevant but I wonder if keys need to be unique within a map?  I guess not.

rust/lance-encoding/src/decoder.rs
```diff
@@ -774,6 +775,20 @@ impl CoreFieldDecoderStrategy {
                 Ok(Box::new(StructuralListScheduler::new(child_scheduler))
                     as Box<dyn StructuralFieldScheduler>)
             }
+            DataType::FixedSizeList(inner, dimension)
+                if matches!(inner.data_type(), DataType::Struct(_)) =>
+            {
+                let child = field
+                    .children
+                    .first()
+                    .expect("FixedSizeList field must have a child");
```
westonpace: Can you use `expect_ok` here so we get a result and not a panic?

rust/lance-encoding/src/encoder.rs
```diff
@@ -450,6 +453,36 @@ impl StructuralEncodingStrategy {
...
+                            )
+                            .into(),
+                            location: location!(),
+                        });
+                    }
+                    // Complex FixedSizeList needs structural encoding
+                    let child = field
+                        .children
+                        .first()
+                        .expect("FixedSizeList should have a child");
```
westonpace: expect_ok

rust/lance-encoding/src/encodings/logical/fixed_size_list.rs
```diff
@@ -0,0 +1,732 @@
...
+            new_entries.as_struct().clone(),
+            Some(new_validity),
+            keys_sorted,
+        )
+        .expect("MapArray construction should succeed"),
+    )
+}
+
+/// Filters garbage from nested FSL arrays that contain list-like children.
+fn filter_nested_fsl_garbage(
```
westonpace, line 492: Nice work here.

rust/lance-encoding/src/encodings/logical/fixed_size_list.rs
```diff
@@ -0,0 +1,732 @@
...
+        let data_type = make_fsl_struct_type(struct_fields, dimension);
+        let mut field_metadata = HashMap::new();
+        field_metadata.insert(
+            STRUCTURAL_ENCODING_META_KEY.to_string(),
+            structural_encoding.into(),
+        );
+        let field = Field::new("", data_type, true).with_metadata(field_metadata);
+        let test_cases = TestCases::basic().with_min_file_version(min_version);
+        check_specific_random(field, test_cases).await;
+    }
```
westonpace, line 651: Nice compact and comprehensive test

rust/lance-encoding/src/encodings/logical/fixed_size_list.rs
```diff
@@ -0,0 +1,732 @@
...
+            true,
+        );
+        let data_type = DataType::FixedSizeList(Arc::new(inner), 2);
+        let arrow_field = Field::new("test", data_type, true);
+        let lance_field = lance_core::datatypes::Field::try_from(&arrow_field).unwrap();
+        let _ = lance_field.data_type();
+    }
+
+    #[test]
+    #[should_panic(expected = "Unsupported logical type: map")]
```
westonpace, line 669: Would be nice if unsupported type changed to err instead of panic.  I'll make a ticket and we can handle in a follow-up.

wkalt:
@westonpace thanks, comments addressed

wkalt:
someone might also want to remove the breaking change label in case that feeds in to anything. It's an additive change.

PR #5597, opened 2025-12-30
ci: fix release notes for release branches

wjones127:
When we cherry-pick bug fixes onto release branches, we sometimes need to modify the commits to work on a new base. When we do that, we change the commit so it's not the same as the original merge commit from the PR. The GitHub release-notes API excludes these commits from the notes it generates. This PR reproduces the same change notes format, but uses the commit messages to grab the PRs, even if they have been rebased.


jackye1995 (approved):
thanks for the fix!

PR #5603, opened 2025-12-31
feat: support merge_insert with source dedupe on first seen value

jackye1995:
Based on feedback in https://github.com/lance-format/lance/pull/5582

Simplified implementation that just keep the first value in case of duplicated `on` rows in source during merge insert. Users are expected to sort the source properly before using merge-insert.

### DIFF
--- a/rust/lance/src/dataset/write/merge_insert.rs
+++ b/rust/lance/src/dataset/write/merge_insert.rs
@@ -44,9 +44,10 @@ use crate::{
 };
 use arrow_array::{
     cast::AsArray, types::UInt64Type, BooleanArray, RecordBatch, RecordBatchIterator, StructArray,
-    UInt64Array,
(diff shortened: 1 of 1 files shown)


rust/lance/src/dataset/write/merge_insert.rs
```diff
@@ -2079,44 +2109,70 @@ impl Merger {
                 let row_ids = matched.column(row_id_col).as_primitive::<UInt64Type>();
 
                 let mut processed_row_ids = self.processed_row_ids.lock().unwrap();
+                let mut duplicate_indices = Vec::new();
```
wjones127: It would be more efficient to just track `keep_indices`, then later you can directly call `take` on the record batch with those indices. Then you don't need to construct the hashmap or boolean mask.

PR #5604, opened 2025-12-31
feat: support truncate table api

zhangyue19921010:
Semantics
- Introduces **truncate table** to create a new snapshot version whose manifest has no fragments while preserving the existing schema and table config.
- Acts as an overwrite-style commit for the latest version; does not physically delete historical data files.
- After truncation, new data can be appended or the table can be restored to previous versions.

Differences vs Delete Rows

- Delete Rows:
  - Soft-deletes rows by writing deletion files/vectors; existing fragments remain
- Truncate Table:
  - Replaces the current version with an “empty fragments” version; no deletion files are produced and current-version fragments are not carried forward.

```
Rust:
ds.truncate_table().await.unwrap();

Python:
ds.truncate_table()

Java:
ds1.truncateTable()

```


wjones127:
> * Soft-deletes rows by writing deletion files/vectors; existing fragments remain

This isn't true; if you delete an entire fragment it is removed from the list. If you delete all rows, all fragments are removed. So this is just equivalent to `dataset.delete("true")`. We even have a fast path for that [^1]. Is there any other reason to make a special operation for this?

[^1]: https://github.com/lance-format/lance/blob/0a8d618b5ec6ce6ac978aadced88a0061487f9ec/rust/lance/src/dataset/write/delete.rs#L182-L192

zhangyue19921010:
Hi, thanks a lot for your attention @wjones127 ! The addition of support for `Truncate table` here is mainly intended to supplement the `TRUNCATE TABLE` related to standard SQL ([SQL:2008](https://en.wikipedia.org/wiki/Truncate_%28SQL%29?utm_source=chatgpt.com)) at the syntax level, making it more user-friendly for those analysts accustomed to using TRUNCATE TABLE command.

Of course, As u mentioned, Lance users can also use `delete("true")` and complete it quickly based on the fast path in the code.

wjones127:
Okay, I get that it's a more familiar API. Could you change the implementation to just be a simple wrapper around `delete("true")` then?

zhangyue19921010:
> Okay, I get that it's a more familiar API. Could you change the implementation to just be a simple wrapper around `delete("true")` then?

Sure thing. Changed. PTAL ~

zhangyue19921010:
Hi  @wjones127 would u mind to take another look? Thanks :) I have finished that `a simple wrapper around delete("true")`

wjones127 (approved):
Looks good!

PR #5606, opened 2025-12-31
feat: add Error::External variant for preserving user errors

wjones127:
## Summary

- Add `Error::External` variant with `#[snafu(transparent)]` to preserve user errors passed through streams
- Update `From<ArrowError>` to extract inner error from `ExternalError` variant
- Update `From<DataFusionError>` to handle `External` and nested `ArrowError::ExternalError`
- Add helper methods: `external()`, `external_source()`, `into_external()`

## Test plan

- [x] Unit tests for error conversions in `lance-core`
- [x] Integration tests for `InsertBuilder::execute_stream` and `MergeInsertJob::execute_reader`

### DIFF
--- a/rust/lance-core/src/error.rs
+++ b/rust/lance-core/src/error.rs
@@ -429,22 +429,7 @@ impl From<datafusion_common::DataFusionError> for Error {
                 message: e.to_string(),
                 location,
             },
-            datafusion_common::DataFusionError::ArrowError(arrow_err, _) => {
-                // Check if the ArrowError wraps an external error and extract it
-                match *arrow_err {
-                    ArrowError::ExternalError(source) => {
-                        // Try to downcast to lance_core::Error first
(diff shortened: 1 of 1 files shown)


rust/lance-core/src/error.rs
```diff
@@ -373,10 +409,22 @@ impl From<datafusion_common::DataFusionError> for Error {
...
+                        match source.downcast::<Self>() {
+                            Ok(lance_err) => *lance_err,
+                            Err(source) => Self::External { source },
+                        }
+                    }
+                    other => Self::Arrow {
+                        message: other.to_string(),
+                        location,
+                    },
+                }
```
westonpace: Would we get this automatically if we returned `LanceError::from(arrow_err)`?

wjones127: I think so. That's a good idea for simplification.

PR #5612, opened 2026-01-02
feat: refactor use of Error::io

### LINKED ISSUE #5601: Review use of `Error::io()` to find cases where that error should be a different type
Early in the codebase, we were sloppy about error types. Many errors were just defaulted to using IO errors, even if they had nothing to do with IO failures.

https://github.com/lance-format/lance/issues/3159 is an example of this, but there are many more obvious ones throughout the codebase.

We need to review these instances and map them to a more appropriate error type. Pay special attention to ones that need to be changed to `Error::InvalidInput`.

lichuang:
fix #5601


rust/lance-core/src/error.rs
```diff
@@ -164,6 +164,27 @@ impl Error {
             location,
         }
     }
+
+    pub fn not_found(message: impl Into<String>, location: Location) -> Self {
+        let message: String = message.into();
+        Self::NotFound {
+            uri: message,
+            location,
+        }
```
wjones127: We could probably simplify this a lot by just using `std::panic::Location::caller()`. This can even be applied to existing variants.
```suggestion
    pub fn not_found(uri: impl Into<String>) -> Self {
        Self::NotFound {
            uri: uri.into(),
            location: std::panic::Location::caller().to_snafu_location(),
        }
```

rust/lance-datafusion/src/projection.rs
```diff
@@ -280,7 +280,7 @@ impl ProjectionPlan {
             } else {
                 // Regular data column - validate it exists in base schema
                 if base.schema().field(&field.name).is_none() {
-                    return Err(Error::io(
+                    return Err(Error::not_found(
```
wjones127: The `not_found` error is really a `FileNotFound` error, so doesn't apply here.
```suggestion
                    return Err(Error::invalid_input(
```

rust/lance-file/src/previous/writer/mod.rs
```diff
@@ -204,8 +204,8 @@ impl<M: ManifestProvider + Send + Sync> FileWriter<M> {
                 .iter()
                 .map(|batch| {
                     batch.column_by_name(&field.name).ok_or_else(|| {
-                        Error::io(
-                            format!("FileWriter::write: Field '{}' not found", field.name),
+                        Error::not_found(
```
wjones127:
```suggestion
                        Error::invalid_input(
```

rust/lance-file/src/previous/writer/mod.rs
```diff
@@ -641,7 +641,7 @@ impl<M: ManifestProvider + Send + Sync> FileWriter<M> {
                     let value_arr = dict_info.values.as_ref().ok_or_else(|| {
                         Error::io(
```
wjones127: Missed one here:
```suggestion
                        Error::invalid_input(
```

rust/lance-file/src/previous/writer/mod.rs
```diff
@@ -641,7 +641,7 @@ impl<M: ManifestProvider + Send + Sync> FileWriter<M> {
...
                             format!(
-                        "Lance field {} is dictionary type, but misses the dictionary value array", 
+                        "Lance field {} is dictionary type, but misses the dictionary value array",
```
wjones127, line 644: We should probably make it standard that we wrap the field names in quotes:
```suggestion
                        "Lance field '{}' is dictionary type, but misses the dictionary value array",
```

rust/lance-file/src/previous/reader.rs
```diff
@@ -516,7 +516,7 @@ fn get_page_info<'a>(
     batch_id: i32,
 ) -> Result<&'a PageInfo> {
     page_table.get(field.id, batch_id).ok_or_else(|| {
-        Error::io(
+        Error::not_found(
```
wjones127:
```suggestion
        Error::invalid_input(
```

rust/lance-table/src/io/manifest.rs
```diff
@@ -57,13 +57,13 @@ pub async fn read_manifest(
     }
 
     if buf.len() < 16 {
-        return Err(Error::io(
+        return Err(Error::invalid_input(
```
wjones127: This seems like it might be corrupt file to me.
```suggestion
        return Err(Error::corrupt_file(
```

rust/lance-table/src/io/manifest.rs
```diff
@@ -57,13 +57,13 @@ pub async fn read_manifest(
...
             "Invalid format: file size is smaller than 16 bytes".to_string(),
             location!(),
         ));
     }
     if !buf.ends_with(MAGIC) {
-        return Err(Error::io(
+        return Err(Error::invalid_input(
```
wjones127: Same here:
```suggestion
        return Err(Error::corrupt_file(
```

lichuang:
@wjones127 i have resolved your suggestions, please review this pr again:)

wjones127:
@lichuang Generally looks good. Can you get the code compiling and tests passing? Then I can do a final review.

lichuang:
@wjones127 I have fixed compiling error, please run workflow, and review it again.

lichuang:
@wjones127 I have fixed the test case, let's run the workflows again.

lichuang:
@wjones127 I have fixed the test case, let's run the workflows again.

PR #5613, opened 2026-01-02
docs: update Lance-DuckDB docs to latest version 0.4.1

prrao87:
This PR updates the Lance docs for the DuckDB extension per the latest fixes [here](https://github.com/lance-format/lance-duckdb/pull/119). The examples show the latest public-facing API for vector search, FTS and hybrid search. For other functionality (including the full SQL reference and cloud reference), we point users to the source repo files.


prrao87:
@Xuanwo we'd need to expose the latest fix to the community extension so that users don't face the error I did - let me know if there's anything else we're missing before announcing its release!

jackye1995:
Is it possible that we make this consistent with the other connector docs, so the doc stays in the connector repo and we build it into the main website only during deployment time?

prrao87:
@jackye1995 yes, that could be done, but it'll require the lance-duckdb repo to have mkdocs set up. In the interest of time, and in having something up and running, I chose this faster approach. Could we merge this first, and create a separate issue to improve the process going forward?

PR #5624, opened 2026-01-05
refactor: introduce RowSetOps and refactor RowAddrTreeMap

yanghua:


yanghua:
cc @wjones127 @jackye1995

rust/lance-core/src/utils/mask.rs
```diff
@@ -243,6 +243,39 @@ impl std::ops::BitOr for RowAddrMask {
...
+    fn is_empty(&self) -> bool;
+
+    /// Returns the number of rows in the set, if it is known.
+    ///
+    /// Implementations that cannot always compute an exact size (for example
+    /// because of "full fragment" markers) should return `None`.
+    fn len(&self) -> Option<u64>;
+
+    /// Remove a value from the row set.
+    fn remove(&mut self, value: u64) -> bool;
```
majin1102: Is this value should be type of Row?

majin1102 (commented):
This refactor is cool. Just leave one comment

rust/lance-core/src/utils/mask.rs
yanghua: Good suggestion! Have addressed.

PR #5625, opened 2026-01-05
feat: prune fragments using scalar indices for a given filter

fangbo:
# Problem Description

We have 10,000 fragments for a dataset. When Spark reads data with filter from the dataset, there will be 10,000 scan tasks, even though the filter can prune the vast majority of fragments. Majority of scan tasks will not read any data. But too many scan tasks degrade the execution efficiency of Spark jobs.

# Purpose of this PR

The filter contain's a column which has index. Using the index filter can prune the fragments which should not be scanned.

This PR provides a method `prune_fragments` for Dataset. This method prunes some fragments depending on filter and index and returns the fragments need to be scanned.


fangbo:
@jackye1995 @majin1102 @yanghua @wojiaodoubao @Jay-ju This PR is ready. Could you please review it? Thank you.

majin1102:
It seems we already attempted similar work in [PR #4835](https://github.com/lance-format/lance/pull/4835). Maybe we could revisit it.

PR #5628, opened 2026-01-05
feat: improve the random access file benchmark

westonpace:
I've been using this benchmark for a number of potential reader improvements and made it more representative of real world use cases by adding multiple threads, more depth per thread, and a no-cache mode to force the data to be read from disk.


westonpace:
I've addressed P0 from claudes review.  I think P1 is fine.  500MB is needed to get 100,000 spaced out rows.  We need that many to get a good sense of the timing.

PR #5633, opened 2026-01-06
feat: merge-insert with primary key dedupe

jackye1995:
Based on https://github.com/lance-format/lance/pull/4787

Co-authored-by: vinoyang <vinoyang@apache.org>

### DIFF
--- a/rust/lance/src/dataset/write/merge_insert.rs
+++ b/rust/lance/src/dataset/write/merge_insert.rs
@@ -1639,7 +1639,6 @@ impl MergeInsertJob {
                     .map(|f| f.id as u32)
                     .collect(),
                 update_mode: Some(RewriteRows),
-                // Slow path doesn't track inserted keys - conflict detection not available
                 inserted_rows_filter: None, // not implemented for v1
             };
 
@@ -4357,6 +4356,9 @@ mod tests {
         }
     }
 
+    /// Test that two merge insert operations on the same existing key conflict.
+    /// First merge insert commits successfully, second one fails with conflict error
+    /// because both operations updated the same key (detected via bloom filter).
     #[tokio::test]
     async fn test_inserted_rows_filter_bloom_conflict_detection_concurrent() {
         // Create schema with unenforced primary key on "id" column
@@ -4380,24 +4382,7 @@ mod tests {
         )
         .unwrap();
 
-        // Throttle to increase contention
-        let throttled = Arc::new(ThrottledStoreWrapper {
-            config: ThrottleConfig {
-                wait_put_per_call: Duration::from_millis(5),
-                wait_get_per_call: Duration::from_millis(5),
-                wait_list_per_call: Duration::from_millis(5),
-                ..Default::default()
-            },
-        });
-
         let dataset = InsertBuilder::new("memory://")
-            .with_params(&WriteParams {
-                store_params: Some(ObjectStoreParams {
-                    object_store_wrapper: Some(throttled.clone()),
-                    ..Default::default()
-                }),
-                ..Default::default()
-            })
             .execute(vec![initial])
             .await
             .unwrap();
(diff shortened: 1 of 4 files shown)


protos/transaction.proto
```diff
@@ -184,6 +184,36 @@ message Transaction {
     optional string branch_name = 5;
   }
   
+  // Exact set of key hashes for conflict detection.
```
jackye1995, line 187: @yanghua I put up this draft, which has the following major changes:

1. some renames, I renamed the join key as just KeyExistenceFilter, so it is not really tied to a join or merge. We are just having a serialized data structure to store the filter.
2. instead of recording all keys, we only record the inserted keys. If there are 2 conflicting merge-inserts, the check for existing rows is already done through affected_rows. We just need to check any newly inserted rows do not conflict.
3. instead of processing the stream twice, we record the inserted rows filter in write exec node, which should be more efficient.
4. removed unnecessary conflict detection logic, leveraging existing conflict resolver.

rust/lance/src/dataset/write/merge_insert.rs
jackye1995, line 1: @yanghua One discussion which I have not finalized my thoughts yet, is if we should check against the merge column, or  the primary key.

I understand the argument that checking against merge column is more flexible, but I don't really know any case that it would actually make sense, that someone wants to merge against one column that is not the primary key and dedupe against it. If you have a good use case, let me know!

As we discussed before, the merge_insert semantics today is technically correct, because there is no primary key and having duplicated inserted rows in concurrent merge-inserts is valid snapshot-isolation behavior.

The behavior only breaks if we actually want to enforce primary key. So I think what makes more sense here is to expose a new config in the job builder: `enforce_primary_key`. Today we already allow users to define unenforced primary key in lance schema, and I think this config makes sense that when disabled (default), primary key is unenforced and the behavior today still holds. Only when we enforce primary key, this new behavior should be used.

Let me know what you think!

rust/lance/src/io/commit/conflict_resolver.rs
```diff
@@ -343,17 +343,71 @@ impl<'a> TransactionRebase<'a> {
...
         } = &self.transaction.operation
         {
+            if let Operation::Update {
+                inserted_rows_filter: other_inserted_rows_filter,
+                ..
+            } = &other_transaction.operation
+            {
+                match (self_inserted_rows_filter, other_inserted_rows_filter) {
+                    (Some(self_keys), Some(other_keys)) => {
+                        if self_keys.field_ids == other_keys.field_ids {
```
jackye1995: @yanghua Another point for using primary key is that, I don't really see what we can do when 2 merge-inserts have different on columns and need to enforce the keys are unique. We cannot even compare them, and we can basically only fail it to be safe.

rust/lance/src/dataset/write/merge_insert.rs
yanghua: > job builder 

You mean the merge insert builder?

> So I think what makes more sense here is to expose a new config in the job builder: enforce_primary_key.

In my mind, the deduplicate ability is combined into two parts:

* upcoming concurrent source table, as you suggested, based on one primary key;
* upcoming source table and existing target table, a thought is that the ability to deduplicate the source table and the target table is based on `merge insert` join. However, this case can not be controlled by the `enforce_primary_key`, right?

rust/lance/src/io/commit/conflict_resolver.rs
yanghua: > I don't really see what we can do when 2 merge-inserts have different on columns and need to enforce the keys are unique. We cannot even compare them, and we can basically only fail it to be safe.

Considering there are two concurrent commits with different join keys: `a`, `b`, there are four combined cases:

* a, a: can compare and judge;
* b,b: can compare and judge;
* a,b and b,a: can not compare, let it go;

But,  yes, I do not have a good case for multiple bloom-filters that base on join keys. This idea is just an thought when implementing.

rust/lance/src/dataset/write/merge_insert.rs
jackye1995: > You mean the merge insert builder?

Yes

PR #5636, opened 2026-01-06
perf: use binary search to skip documents

BubbleCal:
searching gains ~7% faster


rust/lance-index/src/scalar/inverted/wand.rs
```diff
@@ -242,9 +242,30 @@ impl PostingIterator {
...
                 }
                 self.index = self.index.max(block_idx * BLOCK_SIZE);
-                let length = self.list.len();
-                while self.index < length && (self.doc().unwrap().doc_id() as u32) < least_id {
-                    self.index += 1;
+                let length = list.length as usize;
+                while self.index < length {
+                    let block_idx = self.index / BLOCK_SIZE;
+                    let block_offset = self.index % BLOCK_SIZE;
+                    let compressed = unsafe {
```
Xuanwo: what's this for?

rust/lance-index/src/scalar/inverted/wand.rs
```diff
@@ -242,9 +242,30 @@ impl PostingIterator {
                     block_idx += 1;
                 }
                 self.index = self.index.max(block_idx * BLOCK_SIZE);
-                let length = self.list.len();
-                while self.index < length && (self.doc().unwrap().doc_id() as u32) < least_id {
```
Xuanwo, line 246: Why is our new PR faster? Is `self.doc()` a heavy operation? Would it be better to provide an API like `compressed_doc(doc_index)` instead of maintaining complex logic inside `next`?

BubbleCal: The idea is to use binary search to avoid scanning the entire block.

`doc()` is the most costly operation during FTS search because it decompresses the block if needed

rust/lance-index/src/scalar/inverted/wand.rs
BubbleCal: the posintg lists are compressed in memory and we do decompress the needed blocks while searching

PR #5639, opened 2026-01-06
feat: add a MVP for column statistics at dataset level on Rust and Python side

HaochengLIU:
[design doc](https://docs.google.com/document/d/1QbulTzBpPMwxlRPScTlDFBh2pZi3gmAyZaW7RJkxQt0/edit?usp=sharing) ,Related to #4540 

## Overview
This PR implements dataset-level column statistics for Lance, enabling query optimization through min/max/null/nan statistics stored at the dataset level. The implementation follows a 7-phase design with ~4,200 lines of code.

Default to on as it's additive.

## Executive Summary
The new column stats have a two-level lifecycle.

  How they are produced: when a Lance data file is written, FileWriter creates one ColumnStatisticsProcessor per supported column unless disable_column_stats
  is true. Each processor uses lance_arrow_stats::StatisticsAccumulator to track min, max, null_count, and nan_count in 1M-row zones. At file finalization,
  those zone stats are serialized as Arrow IPC into a file global buffer, and the schema metadata records the global-buffer index plus stats version. See rust/
  lance-file/src/writer.rs:546 and rust/lance-file/src/writer/column_stats.rs:1.

  How they become dataset-level stats: compaction/optimize can consolidate all per-file stats into _stats/column_stats.lance. The consolidated file has one row
  and one column per dataset column; each column is a list of zone structs: fragment_id, zone_start, zone_length, null_count, nan_count, min_value, max_value.
  Then an UpdateConfig transaction stores { path, version } in manifest.column_stats. See rust/lance/src/dataset/column_stats_consolidator.rs:145 and rust/
  lance/src/dataset/optimize.rs:2191.

  How they will be used: a future scan/filter path can read manifest.column_stats, open _stats/column_stats.lance, and use ColumnStatsReader to get typed min/
  max and zone bounds for a column. That enables zone/fragment pruning: if a predicate cannot match a zone based on min/max/null/nan stats, the scanner can
  skip that zone’s row range. The reader abstraction exists in rust/lance/src/dataset/column_stats_reader.rs:25, but the general scan/query path is not broadly
  wired to consume it yet.

  Important caveat: normal appends create per-file stats, but they do not currently recompute the consolidated dataset-level stats. The manifest-level pointer
  is refreshed during optimize/compaction, all-or-nothing only when every fragment has stats.
---

## 🎯 What This PR Does

**Goal**: When using Dataset write API, collect per-fragment column statistics at writer level and consolidate them into a single dataset-level file at compaction stage for query optimization.

**Key Features**:
- Per-fragment statistics collection (min, max, null_count, nan_count)
- Dataset-level consolidation into a single Lance file
- Policy enforcement (all-or-nothing: all fragments must have stats)
- Compaction integration (automatic consolidation)
- Backward/forward compatible format

---

## 📊 Architecture Summary
image

```
┌─────────────────────────────────────────────────────────────┐
│                     Write Operation                          │
│  ┌──────────────┐         ┌──────────────┐                  │
│  │ Data Chunks  │────────▶│ FileWriter   │                  │
│  └──────────────┘         │ + ZoneStats  │                  │
│                            └──────┬───────┘                  │
│                                   │                          │
│                                   ▼                          │
│                         ┌─────────────────────┐              │
│                         │  Fragment.lance     │              │
│                         │  + column_stats     │              │
│                         │    (Arrow IPC)      │              │
│                         └─────────────────────┘              │
└─────────────────────────────────────────────────────────────┘

┌─────────────────────────────────────────────────────────────┐
│                    Compaction Operation                      │
│                                                              │
│  ┌────────────┐   ┌────────────┐   ┌────────────┐          │
│  │ Frag 0     │   │ Frag 1     │   │ Frag 2     │          │
│  │ stats      │   │ stats      │   │ stats      │          │
│  └─────┬──────┘   └─────┬──────┘   └─────┬──────┘          │
│        │                │                │                  │
│        └────────────────┼────────────────┘                  │
│                         │                                   │
│                         ▼                                   │
│              ┌──────────────────────┐                       │
│              │  Consolidate Stats   │                       │
│              │  (Global Offsets)    │                       │
│              └──────────┬───────────┘                       │
│                         │                                   │
│                         ▼                                   │
│              ┌──────────────────────┐                       │
│              │ _stats/              │                       │
│              │  column_stats_v1_... │                       │
│              │  (Lance file)        │                       │
│              └──────────────────────┘                       │
└─────────────────────────────────────────────────────────────┘
```

---

### Phase 0: Code refactor between index and column stats (Foundation)
1. **`rust/lance-core/src/utils/zone.rs`** (NEW, 212 lines)
   - `ZoneBound`, `ZoneProcessor`, `FileZoneBuilder`
   - Reusable zone utilities for both stats and indices

### Phase 1: Policy Enforcement at writer level (Consistency)
2. **`rust/lance/src/dataset/write.rs`** (MODIFIED, +50 lines)
   - Added `enable_column_stats` field to `WriteParams`
   - `validate_column_stats_policy()`: Errors on mismatch

3. **`rust/lance/src/dataset/write/insert.rs`** (MODIFIED, +185 lines)
   - Sets `lance.column_stats.enabled` in manifest on dataset creation
   - 4 tests for policy enforcement

### Phase 2: Per-Fragment Writer (Collection)
4. **`rust/lance-file/src/writer.rs`** (MODIFIED, +407 lines)
   - `build_column_statistics()`: Collects stats using DataFusion accumulators
   - Stores as Arrow IPC in global buffer
   - Use a version number to control upgrade

### Phase 3: Per-Fragment Reader (Retrieval)
5. **`rust/lance-file/src/reader.rs`** (MODIFIED, +305 lines)
   - `has_column_stats()`: Quick metadata check
   - `read_column_stats()`: Deserialize Arrow IPC

### Phase 4: Consolidation Core (Aggregation)
6. **`rust/lance/src/dataset/column_stats.rs`** (NEW, 1,049 lines)
   - `consolidate_column_stats()`: Main consolidation logic
   - All-or-nothing policy check
   - Global offset calculation

#### **Manifest Reference**:
- **Key**: `lance.column_stats.file`
- **Value**: `_stats/column_stats.lance` (always the same path)
- **Version**: Stored as file metadata (`lance:dataset:version` key in Lance file)

### Phase 5: High-Level API (Type Dispatch)
7. **`rust/lance/src/dataset/column_stats_reader.rs`** (NEW, 397 lines)
   - `ColumnStatsReader`: Read with automatic type conversion
   - `parse_scalar_value()`: String → ScalarValue dispatch
   - Supports Int8-64, UInt8-64, Float32/64, Utf8, LargeUtf8

### Phase 6: Compaction Integration (Automation)
8. **`rust/lance/src/dataset/optimize.rs`** (MODIFIED, +630 lines)
   - Added `consolidate_column_stats` to `CompactionOptions` (default: `true`)
   - Automatic consolidation after compaction
   - 6 comprehensive tests covering all scenarios

---

### Backward Compatibility
- Old datasets without stats: ✅ Work normally (consolidation skipped)
- Old readers: ✅ Ignore new metadata (no breaking changes)

### Forward Compatibility
- New columns can be added: ✅ Schema is extensible
- New stat types: ✅ Can add columns to consolidated file
- Format versioning: ✅ `lance:column_stats:version` in metadata

---

## 🚀 Next Steps

This PR is the **foundation**. Future work:
1. **Query Optimizer Integration**: Use stats for predicate pushdown
2. **Python Bindings**: Expose stats to Python API
3. **Performance benchmarks**
4. **Additional Statistics**: Distinct count, histograms
5. **Incremental Updates**: Update stats without full consolidation
6. **Stats-based Query Planning**: Cardinality estimation

---

## 📝 Questions for Reviewers
1. **Policy Enforcement**: Is the error message clear when policy mismatches?
2. **Schema Evolution**: Any concerns about adding new stat types later?
3. **Compaction Integration**: Does it cover all compaction code paths?
4. **API Design**: Is `ColumnStatsReader` intuitive enough?
5. **Test Coverage**: Any missing edge cases?

## Summary by CodeRabbit

* **New Features**
  * Added optional dataset column statistics generation and storage (per-zone min/max, null and NaN counts), plus consolidated dataset-level artifacts.
  * Exposed APIs to read file-level and consolidated column statistics.
  * Added Python write support to enable/disable column statistics via `disable_column_stats`.
* **Bug Fixes**
  * Improved compatibility when appending datasets with differing column-statistics policy settings.
  * Updated column-statistics handling across projections and compaction to keep dataset versioning and metadata consistent.
* **Tests**
  * Extended schema/metadata validation and added regressions for legacy behavior around `disable_column_stats`.


HaochengLIU:
@westonpace @jackye1995 @wjones127 
Here is the first iteration of the PR. Feel free to add related parties and any advices are more than welcomed!

rust/lance-core/src/utils/zone.rs
```diff
@@ -0,0 +1,595 @@
...
+    pub fragment_id: u64,
+    /// Start row offset within the fragment (local offset)
+    ///
+    /// To get the actual first row address, use `(fragment_id << 32) | start`.
+    pub start: u64,
+    /// Span of row offsets between the first and last row in the zone
+    ///
+    /// Calculated as (last_row_offset - first_row_offset + 1). This is not
+    /// the count of physical rows, since deletions may create gaps within
+    /// the span.
```
westonpace: This comment is confusing.

`last_row_offset - first_row_offset + 1` **is** the count of physical rows.

Physical row count (includes deleted rows)
Logical row count (does not include deleted rows)

Which is it?  Based on the example in the struct comment I think this is a physical row count.

rust/lance-core/src/utils/zone.rs
```diff
@@ -0,0 +1,595 @@
...
+    /// This method is called repeatedly with chunks of data. Implementations
+    /// should accumulate statistics incrementally.
+    fn process_chunk(&mut self, values: &ArrayRef) -> Result<()>;
+
+    /// Emit statistics when the zone is full or the fragment changes.
+    ///
+    /// The provided `bound` describes the row range covered by this zone.
+    /// After calling this method, the processor should be ready to start
+    /// accumulating statistics for the next zone (via `reset()`).
+    fn finish_zone(&mut self, bound: ZoneBound) -> Result<Self::ZoneStatistics>;
```
westonpace, line 59: Minor nit: why doesn't `finish_zone` automatically do a reset?  Why do I have to manually call it?

rust/lance-core/src/utils/zone.rs
```diff
@@ -0,0 +1,595 @@
...
+        self.flush_zone()?;
+        Ok(self.zones)
+    }
+
+    /// Returns a reference to the collected zone statistics so far.
+    ///
+    /// Note: This does not include the current partial zone being accumulated.
+    pub fn zones(&self) -> &[P::ZoneStatistics] {
+        &self.zones
+    }
```
westonpace: Maybe this should be a method that consumes `self` and returns the zones?  Or do we really peek at the zones partway through building?

rust/lance-file/src/reader.rs
```diff
@@ -1400,6 +1408,100 @@ impl FileReader {
...
+    ///
+    /// Column statistics are stored in the schema metadata under the key
+    /// `lance:column_stats:buffer_index`. If this key exists, the file
+    /// has column statistics that can be read with `read_column_stats()`.
+    ///
+    pub fn has_column_stats(&self) -> bool {
+        self.metadata
+            .file_schema
+            .metadata
+            .contains_key("lance:column_stats:buffer_index")
```
westonpace: Make this key a constant somewhere?

rust/lance-file/src/reader.rs
```diff
@@ -1400,6 +1408,100 @@ impl FileReader {
...
+            .submit_request(
+                vec![
+                    buffer_descriptor.position..buffer_descriptor.position + buffer_descriptor.size,
+                ],
+                0,
+            )
+            .await?;
+
+        // TODO: Is it needed?
+        // Combine all bytes into a single buffer (usually should be just one chunk)
```
westonpace: The number of buffers returned by `submit_request` will exactly match the number of ranges provided.  Since you are providing one range you will get back one buffer.  So the `usually` here is `always` and the `Is it needed?` is `I don't think so`.

rust/lance-file/src/reader.rs
```diff
@@ -1400,6 +1408,100 @@ impl FileReader {
...
+            // Concatenate multiple chunks
+            let total_size: usize = stats_bytes_vec.iter().map(|b| b.len()).sum();
+            let mut combined = BytesMut::with_capacity(total_size);
+            for chunk in stats_bytes_vec {
+                combined.extend_from_slice(&chunk);
+            }
+            combined.freeze()
+        };
+
+        // Decode Arrow IPC format
```
westonpace, line 2557: Why Arrow IPC format?  Why not another lance file?

rust/lance-file/src/reader.rs
westonpace: Ah, you have one in the writer you can use

rust/lance-file/src/reader.rs
```diff
@@ -1400,6 +1408,100 @@ impl FileReader {
...
+    pub async fn read_column_stats(&self) -> Result<Option<arrow_array::RecordBatch>> {
+        // Check if column stats exist
+        let Some(buffer_index_str) = self
+            .metadata
+            .file_schema
+            .metadata
+            .get(COLUMN_STATS_BUFFER_INDEX_KEY)
+        else {
+            return Ok(None);
+        };
```
westonpace, line 2524: You should also check the column stats version here.  If it is greater than `COLUMN_STATS_VERSION` (e.g. 1) then you should pretend the file has no column stats.  This way you will support forwards compatibility in the future.

rust/lance-file/src/writer.rs
```diff
@@ -98,8 +110,151 @@ pub struct FileWriterOptions {
     /// versions may have more efficient encodings.  However, newer format versions will
     /// require more up-to-date readers to read the data.
     pub format_version: Option<LanceFileVersion>,
+
+    /// If true, enable column statistics generation when writing data files.
+    /// Column statistics can be used for planning optimization and filtering.
+    pub enable_column_stats: bool,
```
westonpace: I think this should probably be `disable_column_stats` so that the default is to record column stats?  Or is there some reason we don't want that to be the default?

rust/lance-file/src/writer.rs
```diff
@@ -98,8 +110,151 @@ pub struct FileWriterOptions {
...
+    min: ScalarValue,
+    max: ScalarValue,
+    null_count: u32,
+    nan_count: u32,
+    // TODO: add more stats like mean, avg_len and dist_cnt
+    bound: ZoneBound,
+}
+
+/// Statistics processor for a single column that implements ZoneProcessor trait
+struct ColumnStatisticsProcessor {
```
westonpace: Can you move this struct into a submodule?

rust/lance-file/Cargo.toml
```diff
@@ -20,13 +20,16 @@ arrow-arith.workspace = true
...
+arrow-ipc.workspace = true
 arrow-schema.workspace = true
 arrow-select.workspace = true
 async-recursion.workspace = true
 async-trait.workspace = true
 byteorder.workspace = true
 bytes.workspace = true
 datafusion-common.workspace = true
+datafusion-expr.workspace = true
+datafusion.workspace = true
```
westonpace: I don't love the idea of adding datafusion to lance-file :frowning: 

I'm guessing this is for `ScalarValue`?

rust/lance-file/src/writer.rs
```diff
@@ -98,8 +110,151 @@ pub struct FileWriterOptions {
...
+    // Int32(42) -> "42", Float64(3.14) -> "3.14"
+    if let Some(start) = debug_str.find('(') {
+        if let Some(end) = debug_str.rfind(')') {
+            return debug_str[start + 1..end].to_string();
+        }
+    }
+
+    // Fallback: return the whole debug string (shouldn't happen for supported types)
+    debug_str
 }
```
westonpace: This method makes me pretty uncomfortable.  We also need a binary representation for scalar values in https://github.com/lance-format/lance/pull/5641

What I proposed there (and have seen done elsewhere) is that the binary representation should be to take an arrow array of size 1 and then dump all the buffers.

Ideally we can have one utility for scalars somewhere.

rust/lance-file/src/writer.rs
```diff
@@ -639,6 +838,155 @@ impl FileWriter {
...
+        //
+        // Example layout for a dataset with 2 columns ("id", "price") and 2 zones:
+        // ┌─────────────┬─────────┬────────────┬─────────────┬────────────┬───────────┬───────────┬───────────┐
+        // │ column_name │ zone_id │ zone_start │ zone_length │ null_count │ nan_count │ min_value │ max_value │
+        // ├─────────────┼─────────┼────────────┼─────────────┼────────────┼───────────┼───────────┼───────────┤
+        // │ "id"        │ 0       │ 0          │ 1000000     │ 0          │ 0         │ "1"       │ "1000000" │
+        // │ "id"        │ 1       │ 1000000    │ 500000      │ 0          │ 0         │ "1000001" │ "1500000" │
+        // │ "price"     │ 0       │ 0          │ 1000000     │ 0          │ 0         │ "9.99"    │ "99.99"   │
+        // │ "price"     │ 1       │ 1000000    │ 500000      │ 5          │ 0         │ "10.50"   │ "100.50"  │
+        // └─────────────┴─────────┴────────────┴─────────────┴────────────┴───────────┴───────────┴───────────┘
```
westonpace: I haven't gotten to the table-wide stats yet (maybe it is different) but this format makes me a little uneasy.  I think I want to be able to grab the statistics for a single column without having to grab all the data for unrelated columns.

Then again, maybe this is just unexpected.  I'll think it over a bit.

westonpace (commented):
Some initial comments for lance-file and below.  Will look at table-level stuff next.

rust/lance-core/src/utils/zone.rs
HaochengLIU: You are def right. 
Let me fix this mistype

rust/lance-core/src/utils/zone.rs
HaochengLIU: This function is used for testing. Will refactor the tests so that it's verified at the very end.

rust/lance-core/src/utils/zone.rs
HaochengLIU: good catch
consolidated for better readability

rust/lance-file/src/reader.rs
HaochengLIU: ah, i missed it here
will check again

rust/lance-file/src/reader.rs
HaochengLIU: Here is my original thoughts:
* Per each fragment, the column stats is typically small. 1/1M -> 0.0001% of the original data.
* arrow IPC is the most straight-forward solution here since the data is record batch and they are stored in global buffer already

My knowledge leads me to use Arrow IPC here, if lance has a better strength/ the pattern here, happy to change

rust/lance-file/src/reader.rs
HaochengLIU: added

rust/lance-file/src/writer.rs
HaochengLIU: My original plan is to leave it as disabled by default, then make it default once people feel comfortable.

Humm, if the Lance HEAD is experimental, let's make it ON by default..?

HaochengLIU: note: TBD during the next sync

rust/lance-file/src/writer.rs
HaochengLIU: done

rust/lance-file/Cargo.toml
HaochengLIU: yes... Open to all other options

rust/lance-file/src/writer.rs
HaochengLIU: I dont like it at all as well :(
Do you have any suggestions?

rust/lance-file/src/writer.rs
HaochengLIU: I was back and forth on this topic as well. Given the fact that for each file,  the column stats is minimal compared to the actual  length of the data, I choose this representation over the binary presentation which compress stats into a single binary array for each column.

rust/lance/src/dataset/column_stats_consolidator.rs
```diff
@@ -0,0 +1,1604 @@
...
+pub struct ZoneStats {
+    /// Zone boundary information (fragment_id, start offset, length)
+    pub bound: ZoneBound,
+    pub null_count: u32,
+    pub nan_count: u32,
+    pub min: String, // ScalarValue as string (no type prefix)
+    pub max: String, // ScalarValue as string (no type prefix)
+}
+
+/// Consolidate column statistics from all fragments into a single file.
```
HaochengLIU, line 65: table stats

rust/lance-file/src/reader.rs
westonpace: Nah, IPC is fine, I was just curious about the logic.

rust/lance-file/src/writer.rs
westonpace: I don't recall what the consensus was.  Buf if we want it to be disabled by default in the short term but the goal is enabled by default in the long term, then let's name the variable `disable_column_stats` and write a custom `Default` impl.  This way the custom impl is short term and we can have the simple approach long term.

rust/lance/src/dataset/column_stats.rs
```diff
@@ -0,0 +1,1595 @@
...
+use lance_io::scheduler::{ScanScheduler, SchedulerConfig};
+use lance_io::utils::CachedFileSize;
+use object_store::path::Path;
+use snafu::location;
+
+use crate::dataset::fragment::FileFragment;
+use crate::{Dataset, Error};
+
+// Schema field definitions for consolidated statistics
+// Re-export from lance-file for consistency (these are used in the consolidated list-based layout)
```
westonpace: This isn't really a re-export.  Can it be?

rust/lance/src/dataset/column_stats.rs
```diff
@@ -0,0 +1,1595 @@
...
+            }
+        }
+    }
+
+    // If no statistics were collected, return early
+    if stats_by_column.is_empty() {
+        return Ok(None);
+    }
+
+    // Step 4: Build consolidated batch (column-oriented)
```
westonpace: What does `column-oriented` mean here?  Both the fragment format and the consolidated format are row-oriented right?

rust/lance/src/dataset/column_stats_consolidator.rs
```diff
@@ -0,0 +1,1595 @@
...
+/// │             │              │                     │  15.00,20.00] │  150.00,200.00]    │
+/// └─────────────┴──────────────┴─────────────────────┴───────────────┴────────────────────┘
+/// ```
+///
+/// **Key transformations**:
+/// - Fragment 0 local offset 0 → Global offset 0
+/// - Fragment 0 local offset 1M → Global offset 1M
+/// - Fragment 1 local offset 0 → Global offset 2M (base_offset = 2M)
+/// - Fragment 1 local offset 1M → Global offset 3M (base_offset + 1M)
+///
```
westonpace: Why transform the offsets into global offsets?

rust/lance/src/dataset/column_stats.rs
```diff
@@ -0,0 +1,1595 @@
...
+            Arc::new(self.null_counts.finish()) as ArrayRef,
+            Arc::new(self.nan_counts.finish()) as ArrayRef,
+            Arc::new(self.mins.finish()) as ArrayRef,
+            Arc::new(self.maxs.finish()) as ArrayRef,
+        ]
+    }
+}
+
+/// Create the Arrow schema for consolidated statistics
+pub(crate) fn create_consolidated_stats_schema() -> Arc<ArrowSchema> {
```
westonpace: Why `pub(crate)`?

westonpace: If this is constant you could just store it in a lazy static constant?

rust/lance/src/dataset/column_stats.rs
```diff
@@ -0,0 +1,1595 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright The Lance Authors
+
+//! Column statistics consolidation and reading utilities.
+//!
+//! This module provides functionality for:
+//! 1. Consolidating per-fragment column statistics into a single file
+//! 2. Reading consolidated statistics with automatic type dispatching
```
westonpace: Is that in this module?  Isn't that `column_stats_reader`?

rust/lance/src/dataset/optimize.rs
```diff
@@ -156,6 +157,14 @@ pub struct CompactionOptions {
     /// not be remapped during this compaction operation. Instead, the fragment reuse index
     /// is updated and will be used to perform remapping later.
     pub defer_index_remap: bool,
+    /// Whether to consolidate column statistics during compaction.
+    ///
+    /// When enabled, per-fragment column statistics are merged into a single
+    /// consolidated stats file. This only happens if ALL fragments have statistics
+    /// (all-or-nothing policy).
```
westonpace, line 280: Why is it all-or-nothing?  I think this is a tough limitation, would it mean that datasets have to be rewritten to take advantage of column stats?

rust/lance/src/dataset/optimize.rs
```diff
@@ -1126,6 +1145,36 @@ pub async fn commit_compaction(
...
+                    config_updates: Some(crate::dataset::transaction::translate_config_updates(
+                        &upsert_values,
+                        &[],
+                    )),
+                    table_metadata_updates: None,
+                    schema_metadata_updates: None,
+                    field_metadata_updates: HashMap::new(),
+                },
+                None,
+            );
```
westonpace: This is ok for the short term.  Long term I think we want this to be more of a first-class citizen in the manifest?

rust/lance/src/dataset/optimize.rs
```diff
@@ -1126,6 +1145,36 @@ pub async fn commit_compaction(
...
 
+    // Consolidate column statistics if enabled (after the commit)
+    if options.consolidate_column_stats {
+        let new_version = dataset.manifest.version;
+        if let Some(stats_path) =
+            crate::dataset::column_stats::consolidate_column_stats(dataset, new_version).await?
+        {
+            // Update manifest config with stats file path
+            let mut upsert_values = HashMap::new();
+            upsert_values.insert("lance.column_stats.file".to_string(), stats_path);
```
westonpace: Let's also insert a column stats version here.  This way, if we do something radically different in the future (maybe `lance.column_stats.directory` or something crazy) we can key off the version for evolution.

westonpace (changes requested):
I think my biggest concern is the scalar serialization.  That being said, I don't want to hold this work up.  I'll try and work something up next week and we can adapt to using anything I come up with after this PR merges.

This is a good start and we can start to iron out more details once we start adding features that make use of column stats.

HaochengLIU:
> I think my biggest concern is the scalar serialization

I second that. I dont have enough experience on the user experience from  Lance point of view and best practise in Rust, let's go over this tomorrow

rust/lance/src/dataset/column_stats.rs
HaochengLIU: good catch
removed confusing comment

rust/lance/src/dataset/column_stats.rs
HaochengLIU: reworked the comment  in both column_stats.rs and column_stats_reader.rs
`columns_stats.rs` is renamed to `column_stats_consolidator.rs`

rust/lance/src/dataset/column_stats.rs
HaochengLIU: removed the confusing comment

rust/lance/src/dataset/column_stats_consolidator.rs
HaochengLIU: Humm, i'm expecting  the users to ask `WHERE id BETWEEN 1,500,000 AND 2,500,000` thus doing the consolidation here

HaochengLIU: so the upstream can do filtering without knowing the underlying fragments

rust/lance/src/dataset/column_stats.rs
HaochengLIU: "create_consolidated_stats_schema" is used in both column_stats.rs(renamed to column_stats_consolidator.rs) and column_stats_reader.rs

HaochengLIU: I like the idea. Will refactor it

HaochengLIU: newbie q: do you mean to use lazyLock here?

rust/lance/src/dataset/optimize.rs
HaochengLIU: Here is my thoughts:
First, no data is better than incomplete/wrong data.
For the mix of fragments with and w/ col_stats, I take the lazy and easiest path here. 

An alternative way is to calculate the column stats for these old files on the fly and aggregate them to create the final consolidated file at the cost of extra file I/O?

rust/lance/src/dataset/optimize.rs
HaochengLIU: I'm a novice to this part of transaction code. Any suggestions on the `first-class citizen` here?

rust/lance/src/dataset/optimize.rs
HaochengLIU: added

rust/lance/src/dataset/column_stats_consolidator.rs
HaochengLIU: TODO: revert the global offset to local

rust/lance/src/dataset/column_stats.rs
HaochengLIU: done

rust/lance/src/dataset/optimize.rs
HaochengLIU: TODO: log a warning in the current version

rust/lance/src/dataset/column_stats_consolidator.rs
```diff
@@ -0,0 +1,1614 @@
...
+    let mut fragments_with_stats = 0;
+
+    for fragment in &fragments {
+        if fragment_has_stats(dataset, fragment).await? {
+            fragments_with_stats += 1;
+        }
+    }
+
+    if fragments_with_stats < total_fragments {
+        log::info!(
```
westonpace: Let's upgrade this to a warning.  The user asked for something and we could not deliver it.

rust/lance/src/dataset/optimize.rs
HaochengLIU: TODO: protobuf

HaochengLIU: Add a new message with 2 fields to protobuf

HaochengLIU: Handle the case  when it's none

rust/lance/src/dataset/column_stats_consolidator.rs
```diff
@@ -0,0 +1,1614 @@
...
+///
+/// **Consolidated stats** (one row per column, across all fragments):
+/// ```text
+/// ┌─────────────┬──────────────┬─────────────────────┬───────────────┬────────────────────┐
+/// │ column_name │ fragment_ids │ zone_starts         │ min_values    │ max_values         │
+/// │ (string)    │ (list<u64>)  │ (list<u64>)         │ (list<str>)   │ (list<str>)        │
+/// ├─────────────┼──────────────┼─────────────────────┼───────────────┼────────────────────┤
+/// │ "id"        │ [0,0,1,1]    │ [0,1M,2M,3M] ←GLOBAL│ [1,1M,2M,3M]  │ [1M,2M,3M,4M]      │
+/// │ "price"     │ [0,0,1,1]    │ [0,1M,2M,3M] ←GLOBAL│ [9.99,10.50,  │ [99.99,100.50,     │
+/// │             │              │                     │  15.00,20.00] │  150.00,200.00]    │
```
westonpace: If you have a `List<struct<fragment_id, zone_start, min_value, max_value>>` then it might be nice because you don't have to worry about the columns having different size lists.

rust/lance/src/dataset/column_stats_consolidator.rs
```diff
@@ -0,0 +1,1614 @@
...
+/// │ column_name │ zone_id │ zone_start │ zone_length │ min_value  │ max_value │
+/// ├─────────────┼─────────┼────────────┼─────────────┼────────────┼───────────┤
+/// │ "id"        │ 0       │ 0          │ 1000000     │ "2000001"  │ "3000000" │
+/// │ "id"        │ 1       │ 1000000    │ 1000000     │ "3000001"  │ "4000000" │
+/// │ "price"     │ 0       │ 0          │ 1000000     │ "15.00"    │ "150.00"  │
+/// │ "price"     │ 1       │ 1000000    │ 1000000     │ "20.00"    │ "200.00"  │
+/// └─────────────┴─────────┴────────────┴─────────────┴────────────┴───────────┘
+/// ```
+///
+/// This function **consolidates** them into a **list-based layout** with global offsets:
```
HaochengLIU: TODO: make both columnar

HaochengLIU: | id <struct??>|price<struct??>|

struct data type in Arrow

westonpace:
I was thinking on this yesterday and realized another advantage of using a columnar format for stats.  If you do that then you do not need the string representation for scalars.  For example, with the row based approach your schema is:

```
column_name: string
fragment_ids: List<u64>
zone_starts: List<u64>
min_values: List<string>
max_values: List<string>
```

Since `min_values` and `max_values` are `string` you need some way to convert arrow scalars to `string`.  Also, this type must be `string` (or `binary`) because the `min_values` column will need to contain different types.

However, if you have a columnar representation, then the schema (assuming no structs for simplicity) is something like:

```
id_fragment_id: u64
id_zone_start: u64
id_min_value: id-data-type
id_max_value: id-data-type
```

In this representation, since `id_min_values` and `id_max_values` only contains min/max for the `id` column it can be typed with the same type as `id`.  You don't need some way to convert scalars to / from string.

And then with a struct we have...

```
id: struct<
    fragment_id: u64,
    zone_start: u64,
    min_value: id-data-type,
    max_value: id-data-type,
>
```

rust/lance-file/benches/reader.rs
```diff
@@ -120,8 +120,7 @@ fn bench_reader(c: &mut Criterion) {
 }
 
 #[cfg(not(target_os = "linux"))]
-pub fn drop_file_from_cache(path: impl AsRef<std::path::Path>) -> std::io::Result<()> {
-    println!("drop_file_from_cache: not implemented on this platform");
+pub fn drop_file_from_cache(_path: impl AsRef<std::path::Path>) -> std::io::Result<()> {
```
HaochengLIU, line 130: to silence clippy warning

HaochengLIU:
@westonpace Mind reviewing again?

protobuf + columnar layout for both each data file and the final consolidated files are in

HaochengLIU:
There are some old tests are still failing as I enable the column stats to be ON by default. 

I will try my best to fix them before I will be OOO, if not can someone lend me a hand please? @jackye1995 @westonpace ? tyvm

ulgut:
@jackye1995 can I help push this through? Not sure what the status is here

HaochengLIU:
I believe its priority is downgraded due to other more important things poping up recently.  I will defer to Jacky and Weston to comment on the latest status of the PR

westonpace:
> I believe its priority is downgraded due to other more important things poping up recently. I will defer to Jacky and Weston to comment on the latest status of the PR

Yes, sorry, it's been on my back-burner for a while.  Last I was working on this I was trying to extract the scalar handling and stats code into its own modules because we need these in various other places and it will be easier if it is contained.  If anyone wants to help then a review of https://github.com/lance-format/lance/pull/5967 would be useful.  Once that is merged then I wanted to rework this PR to use arrow-stats instead of making its own stats.

Then I can focus a bit more on the transaction management.

HaochengLIU:
> > I believe its priority is downgraded due to other more important things poping up recently. I will defer to Jacky and Weston to comment on the latest status of the PR
> 
> 
> 
> Yes, sorry, it's been on my back-burner for a while.  Last I was working on this I was trying to extract the scalar handling and stats code into its own modules because we need these in various other places and it will be easier if it is contained.  If anyone wants to help then a review of https://github.com/lance-format/lance/pull/5967 would be useful.  Once that is merged then I wanted to rework this PR to use arrow-stats instead of making its own stats.
> 
> 
> 
> Then I can focus a bit more on the transaction management.

No worries at all! I'm semi settled in right now, I will take a review at the PR this weekend

HaochengLIU:
the python/window failing check looks red herring

westonpace:
I've recently been working on a related task and I'm interested in getting your opinion about the possible strategy.

In https://github.com/lance-format/lance/pull/7427 I capture zone-map information at write time and record it in the file footers (similar to this PR).  In that PR I call these "seeds".

However, it does not introduce a new column stats block in the table metadata.  Instead, we continue to treat the column's index as a basic zone map index.  Then, when the next call to update indexes happens, we read the seeds from the files (instead of re-reading the column) and use that to update the zone map for that column.

I think there are some advantages and disadvantages to the two approaches.

Advantages for seeds approach:

* No formal "column stats" concept needed, these are just indexes that are cheap to update
* Very flexible, any index can opt in to inspect the data at write time and leave behind whatever it wants as seeds.

Advantages for this approach:

* All stats combined into one file (the seeds approach has one index folder per column)
* No need to call optimize_indices, column stats calculated at compaction time.
* Does not require a create_index call

I think we could add on to the seeds approach with some kind of multi-index concept and automatic index creation to bring it closer to this approach.  I like the idea that we don't have to introduce a formal "column stats" object in the manifest.  I'm curious to hear your opinions.

HaochengLIU:
@westonpace Good evening , this makes sense to me. I think the key distinction is that seeds are an index-construction optimization, while column stats are a dataset-level metadata feature.

I see these as complementary pieces. At the file level, I agree the seed approach is more generic and looks like the better primitive. It gives different index types a way to collect write-time information while data is already flowing through the writer, and that seems like a good default direction.

I also think we should continue the dataset-level column statistics work. Column stats are foundational metadata requested across different compute engines and frameworks, and they are useful even when a user has not explicitly created or updated an index. That said, I have less context on the most common user patterns here, so I’m happy to defer on exactly how we prioritize the surface area.

If we want to support both, I think it would be better to merge your PR first and I can rebase this work on top of it. That should let us use seeds as the shared file-level mechanism and keep dataset-level column stats as the consolidated metadata layer for query engines and external integrations.

HaochengLIU:
Recently I had a thorough conversation with Weston on this topic and we've decided to close this PR for a better approach, the logistics are here:

1. The current approach does not provide a way to consistently keep the column index always up to date.
2. It's specifically designed for column stats with a new bespoke solution.
3. It touches the table level manifest with limited improvement.

What we want to propose instead is:
1. First and foremost, we will still support the notion of column stats.  The offering will not change.
2. We will extend the zone map index so that it supports column stats at file creation time. See details in this [PR](https://github.com/lance-format/lance/pull/7427).
3. We will support Muti-column index starting from zone map, meaning that an index may be built from multiple columns.
3.1. It's debatable on whether we should do `one column stats per file` or `multiple columns stats sharing the same file`, right now I prefer the latter one.
4. Automatically creation of index
5. Provide a better API to fetch column stats information than https://github.com/lance-format/lance/pull/7463

HaochengLIU:
I will provide more info in the discussion thread.

PR #5641, opened 2026-01-06
feat: evolute all_null_layout to constant layout

Xuanwo:
This PR will evolute all_null_layout to constant layout as mentioned in https://github.com/lance-format/lance/discussions/5631

The logic for checking validity and constant values is a bit clumsy, but I haven't found a better approach yet.

---


protos/encodings_v2_1.proto
```diff
@@ -147,13 +147,25 @@ message FullZipLayout {
   repeated RepDefLayer layers = 8;
 }
 
-// A layout used for pages where all values are null
+// A layout used for pages where all (visible) values are the same scalar value.
```
westonpace, line 150: What does "visible" mean here?

protos/encodings_v2_1.proto
```diff
@@ -147,13 +147,25 @@ message FullZipLayout {
...
+// There may be buffers of repetition and definition information if required in order
+// to interpret what kind of nulls are present / which items are visible.
+message ConstantLayout {
   // The meaning of each repdef layer, used to interpret repdef buffers correctly
   repeated RepDefLayer layers = 5;
+
+  // Inline fixed-width scalar value bytes.
+  //
+  // This MUST only be used for types where a single non-null element is represented by a single
+  // fixed-width Arrow value buffer (i.e. no offsets buffer, no child data).
```
westonpace, line 163: If we have a single value that is represented by multiple buffers couldn't we concatenate them with a header that gives us information on how to disassemble them?  Or maybe we have a different approach for that case?

protos/encodings_v2_1.proto
```diff
@@ -147,13 +147,25 @@ message FullZipLayout {
...
+  //
+  // Constraints:
+  // - MUST be absent for an all-null page
+  // - MUST be <= 32 bytes if present
```
westonpace, line 167: Why?  If it is larger than 32 bytes do we put it elsewhere?

rust/lance-encoding/src/encodings/logical/primitive.rs
```diff
@@ -3855,20 +3866,259 @@ impl PrimitiveStructuralEncoder {
         })
     }
 
+    fn extract_scalar_value(array: &ArrayRef, idx: usize) -> Result<ArrayRef> {
+        use arrow_data::transform::MutableArrayData;
+
+        let data = array.to_data();
+        let mut mutable = MutableArrayData::new(vec![&data], /*use_nulls=*/ true, 1);
+        mutable.extend(0, idx, idx + 1);
+        Ok(make_array(mutable.freeze()))
```
westonpace: `MutableArrayData` is pretty cool.  I didn't know it existed.  I wonder if we could someday replace some of our data builders with it in the future :thinking:

rust/lance-encoding/src/encodings/logical/primitive.rs
```diff
@@ -3855,20 +3866,259 @@ impl PrimitiveStructuralEncoder {
         })
     }
 
+    fn extract_scalar_value(array: &ArrayRef, idx: usize) -> Result<ArrayRef> {
+        use arrow_data::transform::MutableArrayData;
+
+        let data = array.to_data();
+        let mut mutable = MutableArrayData::new(vec![&data], /*use_nulls=*/ true, 1);
+        mutable.extend(0, idx, idx + 1);
```
westonpace: Can we move this into `lance-arrow` and add some testing?  In particular, can we test a few different types and verify that it works both when the item is null and when the item is not null?

rust/lance-encoding/src/encodings/logical/primitive.rs
```diff
@@ -4424,29 +4673,33 @@ impl PrimitiveStructuralEncoder {
...
-                .map(|arr| arr.logical_nulls().map(|n| n.null_count()).unwrap_or(0) as u64)
-                .sum::<u64>();
 
-            if num_values == num_nulls {
-                return if repdefs.iter().all(|rd| rd.is_simple_validity()) {
+            let leaf_validity = Self::leaf_validity(&repdef, num_values as usize)?;
+            let all_null = leaf_validity
+                .as_ref()
+                .map(|validity| validity.count_set_bits() == 0)
+                .unwrap_or(false);
```
westonpace, line 4638: Why is this better than the "is all null" check we had before?

rust/lance-encoding/src/encodings/logical/primitive.rs
```diff
@@ -4524,12 +4795,39 @@ impl PrimitiveStructuralEncoder {
                 );
                 let (indices_data_block, dictionary_data_block) =
                     dict::dictionary_encode(data_block);
+
+                if version.resolve() >= LanceFileVersion::V2_2
```
westonpace: Is this path reachable?  This is the "data was not dictionary encoded but cardinality was low enough that we decided to use dictionary encoding" path.  If the value is constant doesn't that mean we would catch it above when we call `find_constant_scalar`?

rust/lance-encoding/src/encodings/logical/primitive.rs
```diff
@@ -5980,4 +6278,292 @@ mod tests {
...
+        let page = encode_first_page(field, arr.clone(), LanceFileVersion::V2_2).await;
+
+        let PageEncoding::Structural(layout) = &page.description else {
+            panic!("Expected structural encoding");
+        };
+        let Layout::ConstantLayout(layout) = layout.layout.as_ref().unwrap() else {
+            panic!("Expected constant layout in slot 2");
+        };
+        assert!(layout.inline_value.is_some());
+        assert_eq!(page.data.len(), 0);
```
westonpace: Doesn't the `TestCases` struct already have the ability to verify which encodings / layouts are used?  Why do we need this custom stuff?

rust/lance-encoding/src/encodings/logical/primitive.rs
```diff
@@ -3855,20 +3866,259 @@ impl PrimitiveStructuralEncoder {
...
+        if scalar.null_count() != 0 {
+            return Ok(None);
+        }
+        if !Self::is_constant_values(arrays, &scalar, validity)? {
+            return Ok(None);
+        }
+        Ok(Some(scalar))
+    }
+
+    fn encode_constant_page(
```
westonpace: Could this go in the submodule?

rust/lance-encoding/src/encodings/logical/primitive/constant.rs
```diff
@@ -0,0 +1,709 @@
...
+
+fn write_u32(out: &mut Vec<u8>, v: u32) {
+    out.extend_from_slice(&v.to_le_bytes());
+}
+
+fn write_bytes(out: &mut Vec<u8>, bytes: &[u8]) {
+    out.extend_from_slice(bytes);
+}
+
+pub fn encode_scalar_value_buffer(scalar: &ArrayRef) -> Result<LanceBuffer> {
```
westonpace: Can we put the "encode/decode to buffer/inline" code in lance-arrow?  There will be other places we want a general purpose binary scalar representation of Arrow data (column statistics also needs this).

rust/lance-encoding/src/encodings/logical/primitive/constant.rs
```diff
@@ -0,0 +1,709 @@
...
+                        "Invalid constant layout: missing scalar source",
+                        location!(),
+                    ))
+                }
+                (None, 2) => {
+                    return Err(Error::invalid_input(
+                        "Invalid constant layout: ambiguous (2 buffers and no inline_value)",
+                        location!(),
+                    ))
+                }
```
westonpace, line 164: I think we could have two buffers if everything was null?  Though I guess that will be handled in a different decoder?

protos/encodings_v2_1.proto
Xuanwo: Except `NULL` which is invisible, maybe I should just use `non-null values`?

protos/encodings_v2_1.proto
Xuanwo: I did not support that in this PR. One reason is that it would make the encode/decode logic quite complex. Another reason is that I suspect such data is unlikely to share the same constant value.

protos/encodings_v2_1.proto
Xuanwo: Yes, if it exceeds 32B, we will place it in a dedicated buffer instead of in the metadata. 

The intention here is to avoid bloating our metadata too much. The size is set to the largest fixed data type we support (256B), though I am open to adjusting it.

rust/lance-encoding/src/encodings/logical/primitive.rs
Xuanwo: I didn't know about it either! I learned this from GPT-5.2 😆

rust/lance-encoding/src/encodings/logical/primitive.rs
Xuanwo: Makese sense, will add it.

rust/lance-encoding/src/encodings/logical/primitive.rs
Xuanwo: The `leaf_validity` is now loaded from our calculated repdef, which no longer depends on the array buffer itself. I believe this change will make the logic here more consistent.

rust/lance-encoding/src/encodings/logical/primitive.rs
Xuanwo: Yep, you're right, gonna to remove this part.

rust/lance-encoding/src/encodings/logical/primitive.rs
Xuanwo: Yep, will do.

rust/lance-encoding/src/encodings/logical/primitive.rs
Xuanwo: removed.

rust/lance-encoding/src/encodings/logical/primitive/constant.rs
Xuanwo: working on this.

rust/lance-encoding/src/encodings/logical/primitive/constant.rs
Xuanwo: Yes, all null values are handled separately and will not reach this point.

PR #5650, opened 2026-01-07
perf: improve FTS indexing perf and reduce memory footprint

BubbleCal:
- Reuse hashmap
- avoid frequent allocating by SmallVec 

20% faster indexing


artemru:
nice progress, cannot wait to test it!

PR #5655, opened 2026-01-07
feat(java): expose index description and statistics

majin1102:
This PR is to provide the ability to query index information in the Java ecosystem.


majin1102:
This PR is ready for review.
Please take a look when you have time @yanghua @jackye1995

java/src/main/java/org/lance/index/IndexCriteria.java
```diff
@@ -0,0 +1,98 @@
...
+public final class IndexCriteria {
+
+  private final Optional<String> forColumn;
+  private final Optional<String> hasName;
+  private final boolean mustSupportFts;
+  private final boolean mustSupportExactEquality;
+
+  private IndexCriteria(Builder builder) {
+    this.forColumn = Optional.ofNullable(builder.forColumn);
+    this.hasName = Optional.of(builder.hasName);
```
yanghua: Should it be `Optional.ofNullable` to avoid NPE? If users do not call `builder.hasName(...)`

java/src/main/java/org/lance/Dataset.java
```diff
@@ -1002,6 +1004,53 @@ public List<Index> getIndexes() {
...
 
+  /**
+   * Get statistics for a specific index in JSON form.
+   *
+   * <p>The JSON structure matches the Rust/Python index_statistics API.
+   *
+   * @param indexName the name of the index
+   * @return JSON string with index statistics
+   */
+  public String getIndexStatistics(String indexName) {
```
yanghua: Based on the Python API `def index_statistics(self, index_name: str) -> Dict[str, Any]`. What about replacing the result type with `Map<String, Object>` or an Entity like `IndexStatistics`?

yanghua (commented):
Left some comments.

majin1102:
Ready for another look @yanghua

PR #5656, opened 2026-01-07
feat!: make v2 manifest default

wojiaodoubao:
BREAKING CHANGE: defaults new datasets to use V2 manifest path naming scheme. This makes these datasets unreadable for versions of Lance library prior to v0.17.0 (released September 2024).

This default improves performance on object storage. See the original PR (https://github.com/lance-format/lance/pull/2798) for details.

Close https://github.com/lance-format/lance/issues/5634

### DIFF
--- a/java/lance-jni/src/blocking_dataset.rs
+++ b/java/lance-jni/src/blocking_dataset.rs
@@ -415,6 +415,24 @@ pub extern "system" fn Java_org_lance_Dataset_drop<'local>(
 }
 
 #[no_mangle]
+pub extern "system" fn Java_org_lance_Dataset_nativeMigrateManifestPathsV2(
+    mut env: JNIEnv,
+    java_dataset: JObject,
+) {
+    ok_or_throw_without_return!(
+        env,
+        inner_native_migrate_manifest_paths_v2(&mut env, java_dataset)
+    )
+}
+
+fn inner_native_migrate_manifest_paths_v2(env: &mut JNIEnv, java_dataset: JObject) -> Result<()> {
+    let mut dataset_guard =
+        unsafe { env.get_rust_field::<_, _, BlockingDataset>(java_dataset, NATIVE_DATASET) }?;
+    RT.block_on(dataset_guard.inner.migrate_manifest_paths_v2())?;
+    Ok(())
+}
+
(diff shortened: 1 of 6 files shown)


wojiaodoubao:
Hi @westonpace @wjones127 , could you help review this when you have time, thanks very much~

python/python/lance/dataset.py
wjones127, line 1: Need to update here as well:

https://github.com/wojiaodoubao/lance/blob/d20720fda47527a018379d298df4612eb66c7fbf/python/python/lance/dataset.py#L3298-L3305

java/src/main/java/org/lance/Dataset.java
```diff
@@ -464,12 +464,14 @@ public Dataset commitTransaction(Transaction transaction) {
    *
    * @param transaction The transaction to commit
    * @param detached If true, the commit will not be part of the main dataset lineage.
+   * @param enableV2ManifestPaths If true, use v2 manifest, default is true.
```
wjones127: Might want to use something similar to this string here:

https://github.com/wojiaodoubao/lance/blob/d20720fda47527a018379d298df4612eb66c7fbf/python/python/lance/dataset.py#L3298-L3305

In particular, I would let the user know it has no effect if the dataset already exists; it's just for creating new datasets.

java/lance-jni/src/blocking_dataset.rs
wjones127, line 1: There is a migration function https://docs.rs/lance/latest/lance/dataset/struct.Dataset.html#method.migrate_manifest_paths_v2, do you want to expose that in Java too?

wjones127 (changes requested):
I think you missed one or two places.

Also, I've labelled this PR as a breaking change, since changed in default settings are considered breaking changes.

wojiaodoubao:
This pr is updated, please help review when you have time, thanks very much~ @wjones127

wojiaodoubao:
Hi @wjones127 , sorry to bother, could you help review this when you have time, thanks very much~

wojiaodoubao:
Hi @wjones127 , @yanghua , could you help review this, thanks very much!

PR #5664, opened 2026-01-08
feat(java): support building vector index distributively

majin1102:
Note:
1. Reposition merge_index_metadata so that Java dataset could directly call mergeIndexMetadata for vector indexes.
2. Add `centroids` to IvfBuilderParams and `codebook` to PqBuilderParams to make distrubitively creating vector indexes work
3. Add a new index package under tests to have index tests.
4. This PR update Cargo.lock due to the intruduction of #5650.

### DIFF
--- a/java/src/main/java/org/lance/index/vector/VectorTrainer.java
+++ b/java/src/main/java/org/lance/index/vector/VectorTrainer.java
@@ -70,7 +70,4 @@ public final class VectorTrainer {
 
   private static native float[] nativeTrainPqCodebook(
       Dataset dataset, String column, PQBuildParams params);
-
-  private static native SQBuildParams nativeTrainSq(
-      Dataset dataset, String column, SQBuildParams params);
 }


java/src/test/java/org/lance/index/VectorIndexTest.java
```diff
@@ -0,0 +1,255 @@
...
+        for (int d = 0; d < dimension; d++) {
+          centroids[d] = 0.0f;
+          centroids[dimension + d] = 1.0f;
+        }
+
+        float[] codebook = new float[numCentroids * dimension];
+        for (int c = 0; c < numCentroids; c++) {
+          for (int d = 0; d < dimension; d++) {
+            codebook[c * dimension + d] = (float) (c + d);
+          }
```
chenghao-guo: maybe we can use train_ivf / train_pq to generate

majin1102: done

rust/lance-index/src/lib.rs
```diff
@@ -191,22 +191,22 @@ impl TryFrom<&str> for IndexType {
...
+            "Vector" | "VECTOR" => Ok(Self::Vector),
             "IVF_FLAT" => Ok(Self::IvfFlat),
             "IVF_SQ" => Ok(Self::IvfSq),
             "IVF_PQ" => Ok(Self::IvfPq),
             "IVF_RQ" => Ok(Self::IvfRq),
             "IVF_HNSW_FLAT" => Ok(Self::IvfHnswFlat),
             "IVF_HNSW_SQ" => Ok(Self::IvfHnswSq),
             "IVF_HNSW_PQ" => Ok(Self::IvfHnswPq),
+            "FragmentReuse" => Ok(Self::FragmentReuse),
+            "MemWal" => Ok(Self::MemWal),
```
yanghua, line 209: It would be better to avoid unnecessary changes in this PR?

rust/lance/src/dataset.rs
```diff
@@ -2748,6 +2750,59 @@ impl Dataset {
...
+    pub async fn merge_index_metadata(
+        &self,
+        index_uuid: &str,
+        index_type: IndexType,
+        batch_readhead: Option<usize>,
+    ) -> Result<()> {
+        let store = LanceIndexStore::from_dataset_for_new(self, index_uuid)?;
+        let index_dir = self.indices_dir().child(index_uuid);
+        log::info!(
+            "merge_index_metadata called with index_type={})",
```
yanghua: If we want to add some logging, make it more readable?

java/lance-jni/src/vector_trainer.rs
```diff
@@ -0,0 +1,234 @@
...
+    {
+        let dataset_guard =
+            unsafe { env.get_rust_field::<_, _, BlockingDataset>(dataset_obj, NATIVE_DATASET) }?;
+        let dataset = &dataset_guard.inner;
+
+        // Ensure the column exists and is a vector column. This will return a
+        // descriptive error if the column is invalid.
+        let _dim = get_vector_dim(dataset.schema(), &column)?;
+
+        // Drop dataset_guard at the end of this block before reusing env.
```
yanghua: useless?

java/lance-jni/src/vector_trainer.rs
```diff
@@ -0,0 +1,234 @@
...
+    let _sq_params = build_sq_params_from_java(env, &sq_params_obj)?;
+
+    {
+        let dataset_guard =
+            unsafe { env.get_rust_field::<_, _, BlockingDataset>(dataset_obj, NATIVE_DATASET) }?;
+        let dataset = &dataset_guard.inner;
+
+        // Ensure the column exists and is a vector column. This will return a
+        // descriptive error if the column is invalid.
+        let _dim = get_vector_dim(dataset.schema(), &column)?;
```
yanghua: What's the reason to define this var?

java/src/main/java/org/lance/index/vector/VectorTrainer.java
```diff
@@ -0,0 +1,95 @@
...
+   *
+   * <p>SQ does not require a pre-trained artifact. This entry point exists to mirror the IVF/PQ
+   * training APIs and to perform any parameter validation on the Rust side.
+   *
+   * @param dataset the dataset to inspect
+   * @param column the vector column name
+   * @param params SQ build parameters
+   * @return a (potentially validated / normalized) SQBuildParams instance
+   */
+  public static SQBuildParams trainSq(Dataset dataset, String column, SQBuildParams params) {
```
yanghua: IMO, we do not need to train for SQ.

rust/lance-index/src/lib.rs
majin1102: This is not newly added.
Just moved system indexes to the bottom so it’s easier to read.

rust/lance/src/dataset.rs
majin1102: Sorry.
This is just copied from original python dataset.rs. I agree to delete this info. WDYT

majin1102:
1. remove unnecessary info code
2. remove unnecessary trainSq

PTAL when you have time @yanghua

java/src/main/java/org/lance/index/vector/VectorTrainer.java
```diff
@@ -0,0 +1,76 @@
...
+    return nativeTrainPqCodebook(dataset, column, params);
+  }
+
+  private static native float[] nativeTrainIvfCentroids(
+      Dataset dataset, String column, IvfBuildParams params);
+
+  private static native float[] nativeTrainPqCodebook(
+      Dataset dataset, String column, PQBuildParams params);
+
+  private static native SQBuildParams nativeTrainSq(
```
yanghua: Delete it?

java/src/test/java/org/lance/index/VectorIndexTest.java
```diff
@@ -0,0 +1,363 @@
...
+import java.util.List;
+import java.util.Optional;
+import java.util.UUID;
+import java.util.stream.Collectors;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+public class VectorIndexTest {
```
yanghua, line 41: Maybe we can compare the recall performance between a single-machine built index and a distributedly built index? Just like Python. Currently, all the test cases do not measure the metrics for the recall.

majin1102: I think we could verify the accuracy of the recall implementation just once — since both Java and Python call the same underlying rust code. I‘d like to keep Java tests more functional-only

What do you think? @yanghua

yanghua: > since both Java and Python call the same underlying rust code

Maybe we can choose one case to verify? Although the underlying code is the same, the upper layer still needs to do some data type change or parameters pass. They may cause some issues?

Issue #5652, opened 2026-01-07
Make the on parameter of merge insert optional


yanghua:
In this PR #4585 , I discussed with @jackye1995 offline. We can make the `on` parameter of the merge insert optional. If it is none, we use the primary key.

PR #5667, opened 2026-01-09
feat: make on arg optional for merge insert api

yanghua:
closes: #5652

### DIFF
--- a/python/src/dataset.rs
+++ b/python/src/dataset.rs
@@ -171,7 +171,10 @@ impl MergeInsertBuilder {
             .when_matched(WhenMatched::DoNothing)
             .when_not_matched(WhenNotMatched::DoNothing);
 
-        Ok(Self { builder, dataset: dataset_py })
+        Ok(Self {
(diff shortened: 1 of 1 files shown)


python/python/lance/dataset.py
```diff
@@ -2026,7 +2026,7 @@ def insert(
 
     def merge_insert(
         self,
-        on: Union[str, Iterable[str]],
+        on: Optional[Union[str, Iterable[str]]] = None,
```
majin1102, line 2029: Wondering if the new table doesn't have the same column name of source table pk. How do we deal with such case.

This isn’t about the current PR. Just want to know if this is a reasonable constrait.

majin1102 (approved):
looks good to me

python/python/lance/dataset.py
yanghua: If the source table can not find the same column as the PK. When doing the `join` action in Data Fusion, it would throw an exception.

PR #5669, opened 2026-01-09
ci: fix cargo lock not updated

Xuanwo:
This PR will fix CI.


majin1102:
Thanks for this fix.
I think java/lance-jni/Cargo.lock needs to be updated as well.

PR #5679, opened 2026-01-10
faet: enhance not found errors with suggestions

camilesing:
for issue: https://github.com/lance-format/lance/issues/5642


wjones127 (commented):
Nice start on this. One note is I still think we should put the similarity computation in the `Display` implementation of the error. You can implement a custom error struct:

```rust
struct MyError {
   ...
}


impl std::fmt::Display for MyError {
    ...
}
```
And then use the `#[snafu(transparent)]` macro in Snafu to make it one of the variants:

```rust
enum Error {
    ...
    #[snafu(transparent)]
    MyError { source: MyError },
    ...
}
```

PR #5680, opened 2026-01-10
chore: add FTS alias for inverted index

BubbleCal:


python/python/lance/dataset.py
```diff
@@ -2548,6 +2550,7 @@ def create_scalar_index(
                 "ZONEMAP",
                 "LABEL_LIST",
                 "INVERTED",
+                "FTS",
```
westonpace, line 2553: I wonder if it would be easier to just do something like...

```
if index_type.upper() == "FTS":
  index_type = "INVERTED"
```

At the top of this method.  That way, if we add other checks for `INVERTED` in the future, we won't have to worry about maintaining the `FTS` alias?

BubbleCal: Tried this but that may be confusing if people switch back to rust lib and it doesn't work. intend to make sure both `FTS` and INVERTED` can work in all places

PR #5681, opened 2026-01-10
feat: support array_contains in LabelList scalar index

### LINKED ISSUE #3687: Support `array_contains` in `LABEL_LIST` index
The `LABEL_LIST` index works with `array_has_all` and `array_has_any`.  However, it does not currently support `array_contains` which may be surprising to many users.  It should be straightforward to add support for `array_contains` the the query parser for `LABEL_LIST`.

fenfeng9:
closes #3687

Changes:
 - Make array_contains(list_col, value) use the existing LABEL_LIST scalar index.
 - DataFusion treats array_contains as an alias of array_has (often wrapped in an alias expr), so we unwrap that and map it to the LabelList index query.
 - Add a Python test and update the LabelList docs to mention array_has / array_contains.

### DIFF
--- a/.bumpversion.toml
+++ b/.bumpversion.toml
@@ -1,5 +1,5 @@
 [tool.bumpversion]
-current_version = "2.0.0-beta.8"
+current_version = "2.0.0-beta.9"
 parse = "(?P<major>\\d+)\\.(?P<minor>\\d+)\\.(?P<patch>\\d+)(-(?P<prerelease>(beta|rc))\\.(?P<prerelease_num>\\d+))?"
 serialize = [
     "{major}.{minor}.{patch}-{prerelease}.{prerelease_num}",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -636,9 +636,9 @@ dependencies = [
 
 [[package]]
 name = "aws-lc-rs"
-version = "1.15.2"
+version = "1.15.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6a88aab2464f1f25453baa7a07c84c5b7684e274054ba06817f382357f77a288"
+checksum = "e84ce723ab67259cfeb9877c6a639ee9eb7a27b28123abd71db7f0d5d0cc9d86"
 dependencies = [
  "aws-lc-sys",
  "zeroize",
@@ -646,9 +646,9 @@ dependencies = [
 
 [[package]]
 name = "aws-lc-sys"
-version = "0.35.0"
+version = "0.36.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b45afffdee1e7c9126814751f88dddc747f41d91da16c9551a0f1e8a11e788a1"
+checksum = "43a442ece363113bd4bd4c8b18977a7798dd4d3c3383f34fb61936960e8f4ad8"
 dependencies = [
  "cc",
  "cmake",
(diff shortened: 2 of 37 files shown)


fenfeng9:
PTAL @westonpace.

fenfeng9:
Discovered a corner data correctness bug in the LABEL_LIST index and filed #5682 for follow-up.

python/python/tests/test_scalar_index.py
```diff
@@ -1972,6 +1972,39 @@ def test_label_list_index(tmp_path: Path):
     assert indices[0]["type"] == "LabelList"
 
 
+def test_label_list_index_array_contains(tmp_path: Path):
+    # Include lists with NULL items to ensure NULL needle behavior matches
+    # non-index execution.
+    tbl = pa.table(
+        {"labels": [["foo", "bar"], ["bar"], ["baz"], ["qux", None], [None], []]}
```
westonpace: Maybe also include an entry where the entire list is None?

rust/lance-index/src/scalar/expression.rs
```diff
@@ -490,6 +490,24 @@ impl ScalarQueryParser for LabelListQueryParser {
...
+        if func.name() == "array_has" {
+            let inner_type = match data_type {
+                DataType::List(field) | DataType::LargeList(field) => field.data_type(),
+                _ => return None,
+            };
+            let scalar = maybe_scalar(&args[1], inner_type)?;
+            // Do not push down NULL needles.
+            if scalar.is_null() {
+                return None;
+            }
```
westonpace: Is this because of https://github.com/lance-format/lance/issues/5682 ?

rust/lance-index/src/scalar/expression.rs
```diff
@@ -1651,6 +1669,7 @@ fn visit_node(
     }
     match expr {
         Expr::Between(between) => Ok(visit_between(between, index_info)),
+        Expr::Alias(alias) => visit_node(alias.expr.as_ref(), index_info, depth),
```
westonpace, line 1674: Nice catch

rust/lance-index/src/scalar/expression.rs
```diff
@@ -490,6 +490,24 @@ impl ScalarQueryParser for LabelListQueryParser {
         if args.len() != 2 {
             return None;
         }
+        if func.name() == "array_has" {
```
westonpace, line 494: How does this work for `array_contains`?  Is Datafusion mapping that to `array_has` already?

If so, can we add a comment here mentioning that this branch is also going to be hit for `array_contains`?

westonpace:
The build-no-lock failure should have been fixed by https://github.com/lance-format/lance/pull/5690

rust/lance-index/src/scalar/expression.rs
fenfeng9: Add comment

rust/lance-index/src/scalar/expression.rs
fenfeng9: Actually, this is independent of #5682. #5682 is about the index missing rows where a valid element and a NULL co-exists (e.g., searching for 'foo' misses ['foo', None]). I will provide more details in #5682 later.
I've added a comment to explain the check. And array_has_any and array_has_all don't have this semantic mismatch as they natively follow membership logic.

python/python/tests/test_scalar_index.py
fenfeng9: updated

fenfeng9:
PTAL @westonpace.

westonpace:
Thanks for picking this up!

PR #5683, opened 2026-01-11
feat: add order to primary key

touch-of-grey:
This PR adds ordering to primary key so that it can be used for btree index. This is a part of the work to make the MemTable proposal work after discussion with @jackye1995, and for my use case the primary key has to be multiple. @jackye1995 could you take a look?


docs/src/format/table/index.md
```diff
@@ -42,6 +42,30 @@ See [File Format Encoding Specification](../file/encoding.md) for details on ava
 
 </details>
 
+### Unenforced Primary Key
```
jackye1995, line 45: thanks for adding this! I missed it in the doc refresh

protos/file.proto
```diff
@@ -166,6 +166,11 @@ message Field {
 
   bool unenforced_primary_key = 12;
 
+  // Position of this field in the primary key (1-based).
+  // 0 means the field is part of the primary key but no explicit position is set.
+  // When set to a positive value, primary key fields are ordered by this position.
+  uint32 unenforced_primary_key_position = 13;
```
jackye1995, line 172: I think we should use the term `field_id` rather than `position`, considering we are introducing field ID concepts for other specs like partitioning spec and region spec

rust/lance-core/src/datatypes/field.rs
```diff
@@ -42,6 +42,13 @@ use crate::{
 /// (3) The field must not be within a list type.
 pub const LANCE_UNENFORCED_PRIMARY_KEY: &str = "lance-schema:unenforced-primary-key";
 
+/// Use this config key in Arrow field metadata to specify the position of a primary key column.
+/// The value is a 1-based integer indicating the order within the composite primary key.
+/// When specified, primary key fields are ordered by this position.
+/// When not specified, primary key fields are ordered by their lance schema field id.
+pub const LANCE_UNENFORCED_PRIMARY_KEY_POSITION: &str =
+    "lance-schema:unenforced-primary-key:position";
```
jackye1995, line 50: so this should be "lance-schema:unenforced-primary-key:field-id"

docs/src/format/table/index.md
```diff
@@ -42,6 +42,30 @@ See [File Format Encoding Specification](../file/encoding.md) for details on ava
...
+- The field, and all its ancestors, must not be nullable.
+- The field must be a leaf field (primitive data type without children).
+- The field must not be within a list or map type.
+
+To mark a field as part of the primary key, add the following metadata to the Arrow field:
+
+- `lance-schema:unenforced-primary-key`: Set to `true`, `1`, or `yes` (case-insensitive) to indicate the field is part of the primary key.
+- `lance-schema:unenforced-primary-key:field-id` (optional): A 1-based integer specifying the field's ID within a composite primary key.
+
+For composite primary keys with multiple columns, the field ID determines the column ordering:
```
jackye1995: nit: "determines the column ordering" should be "determines the primary key field ordering"

docs/src/format/table/index.md
```diff
@@ -42,6 +42,30 @@ See [File Format Encoding Specification](../file/encoding.md) for details on ava
...
 </details>
 
+### Unenforced Primary Key
+
+Lance supports defining an unenforced primary key through field metadata.
+This is useful for deduplication during merge-insert operations and other use cases that benefit from logical row identity.
+The primary key is "unenforced" meaning Lance does not always validate uniqueness constraints.
+Users can use specific workloads like merge-insert to enforce it if necessary.
+
+A primary key field must satisfy:
```
jackye1995, line 53: should also specify that primary key is fixed after initial setting, must not be updated or removed.

rust/lance-core/src/datatypes/field.rs
```diff
@@ -1127,10 +1130,18 @@ impl TryFrom<&ArrowField> for Field {
             }
             _ => vec![],
         };
-        let unenforced_primary_key = metadata
-            .get(LANCE_UNENFORCED_PRIMARY_KEY)
-            .map(|s| matches!(s.to_lowercase().as_str(), "true" | "1" | "yes"))
-            .unwrap_or(false);
+        // Parse primary key position: first try explicit position, then fall back to boolean flag
```
jackye1995: we don't need too many comments, this seems to be generated through AI, simplify all the inline comments.

rust/lance-core/src/datatypes/schema.rs
```diff
@@ -2599,4 +2619,111 @@ mod tests {
                 .contains(error_message_contains[idx]));
         }
     }
+
+    #[test]
+    fn test_schema_unenforced_primary_key_ordering() {
```
jackye1995, line 2620: we should have ordering tests for all 3 cases you listed:

- When field IDs are specified, fields are ordered by their field ID values (1, 2, 3, ...).
- When field IDs are not specified, fields are ordered by their lance schema field id.
- Fields with explicit field IDs are ordered before fields without explicit field IDs.

rust/lance-core/src/datatypes/schema.rs
```diff
@@ -111,11 +111,27 @@ impl<'a> Iterator for SchemaFieldIterPreOrder<'a> {
 }
 
 impl Schema {
-    /// The unenforced primary key fields in the schema
+    /// The unenforced primary key fields in the schema, ordered by field ID.
```
jackye1995: nit: "ordered by field ID." this should be "ordered by primary key field id". Overall make sure we clearly distinguish between schema field id vs primary key field id.

jackye1995 (approved):
looks good to me! pending CI

docs/src/format/table/index.md
```diff
@@ -42,6 +42,31 @@ See [File Format Encoding Specification](../file/encoding.md) for details on ava
...
+
+- The field, and all its ancestors, must not be nullable.
+- The field must be a leaf field (primitive data type without children).
+- The field must not be within a list or map type.
+
+To mark a field as part of the primary key, add the following metadata to the Arrow field:
+
+- `lance-schema:unenforced-primary-key`: Set to `true`, `1`, or `yes` (case-insensitive) to indicate the field is part of the primary key.
+- `lance-schema:unenforced-primary-key:field-id` (optional): A 1-based integer specifying the primary key field id within a composite primary key.
```
jackye1995: Sorry for the back and forth, but I think I agree now that `position` is better than `field-id`. For partition and region spec, they represent new fields in the spec so spec makes sense. But here it is pretty confusing to have field ID and primary key field ID in the same place. Let's move back to that.

PR #5686, opened 2026-01-11
feat(rust): add datafusion catalog_provider through namespace

### LINKED ISSUE #5712: Expose Lance tables in DataFusion via namespace
Integrate datafusion with lance namespace to support multi-table query

How Iceberg integrate datafusion through CatalogProvider:
https://github.com/apache/iceberg-rust/tree/main/crates/integrations/datafusion
How delta integrate datafusion through TableProvider:
https://github.com/delta-io/delta-rs/blob/main/crates/core/src/delta_datafusion/table_provider.rs

majin1102:
This is the first step of #4779 

Close #5712

### DIFF
--- a/rust/lance-namespace-datafusion/src/namespace_level.rs
+++ b/rust/lance-namespace-datafusion/src/namespace_level.rs
@@ -58,7 +58,7 @@ impl NamespaceLevel {
         self.namespace_id.clone().unwrap_or_default()
     }
 
-    /// Human readable name for this namespace (last component or default).
+    /// Name for this namespace (last component or default).
     pub fn name(&self) -> &str {
         self.namespace_id
             .as_deref()
@@ -119,7 +119,6 @@ impl NamespaceLevel {
         DatasetBuilder::from_namespace(
             Arc::clone(&self.root),
             self.child_id(table_name.to_string()),
-            false,
         )
         .await?
         .load()


majin1102:
Hi @wjones127 @jackye1995 @yanghua @westonpace @Xuanwo ,

This PR is driven by #4779. In our production scenario, we’d like to leverage DataFusion to deliver a better SQL experience that supports search, analytics, and AI-related operators (e.g., via UDFs/UDTFs).

I’ve already discussed this with @jackye1995 offline, and we agreed to try this with a dedicated crate for the DataFusion namespace integration. This PR introduces the foundational approach.

I’ll follow up with additional issues covering next steps—such as Python bindings, URL-based tables, SQL extensions, etc.

Please take a look when you have time!

rust/lance-namespace-datafusion/src/error.rs
```diff
@@ -0,0 +1,16 @@
...
+// SPDX-FileCopyrightText: Copyright The Lance Authors
+
+use datafusion::error::{DataFusionError, Result};
+
+/// Convert a Lance error into a DataFusion error.
+///
+/// This keeps all Lance-specific error formatting in a single place.
+pub fn to_datafusion_error<E: std::fmt::Display>(err: E) -> DataFusionError {
+    DataFusionError::Execution(err.to_string())
+}
```
wjones127: I think we should preserve the original error using `DataFusionError::External` instead of converting to a string. This lets the caller turn the DataFusion error back into the original Lance error.

You can see how this is already partially done in https://github.com/lance-format/lance/pull/5606

rust/lance-namespace-datafusion/src/catalog.rs
```diff
@@ -0,0 +1,142 @@
...
+use datafusion::error::Result;
+
+use crate::namespace_level::NamespaceLevel;
+use crate::schema::LanceSchemaProvider;
+#[allow(unused_imports)]
+use crate::SessionBuilder;
+
+/// A dynamic [`CatalogProviderList`] that maps Lance namespaces to catalogs.
+///
+/// The underlying namespace must be a four-level namespace. It is explicitly configured
```
wjones127, line 19: > must be a four-level namespace

Why is this necessary? I wonder if we can work around that.

For example, it's a three level namespace, it could be:

```
DEFAULT > LVL1 > LVL2 > LVL3
```

And then if it's a two level namespace, it could be

```
DEFAULT > DEFAULT > LVL1 > LVL2
```

There might be some other standard name besides default that would make more sense (maybe what other DataFusion plugins do), but you get the idea.

rust/lance-namespace-datafusion/tests/sql.rs
```diff
@@ -0,0 +1,418 @@
...
+
+    let customer_ids = Int32Array::from(vec![1, 2, 3]);
+    let names = StringArray::from(vec!["Alice", "Bob", "Carol"]);
+    let cities = StringArray::from(vec!["NY", "SF", "LA"]);
+
+    let batch = RecordBatch::try_new(
+        schema.clone(),
+        vec![Arc::new(customer_ids), Arc::new(names), Arc::new(cities)],
+    )
+    .unwrap();
```
wjones127: FYI, you could rewrite this and other batch constructions like this:

```suggestion
    let batch = record_batch!(
        ("customer_id", Int32, [1, 2, 3]),
        ("name", Utf8, ["Alice", "Bob", "Carol"]),
        ("city", Utf8, ["NY", "SF", "LA"])
    ).unwrap();
    let schema = batch.schema();
```

wjones127 (commented):
This is cool. I have a few suggestions.

rust/lance-namespace-datafusion/src/catalog.rs
majin1102: I think the docs might be somewhat misleading(have updated it). Let me clarify:

1. First, DataFusion only has three-level metadata. The CatalogProviderList is optional. As mentioned, we use `SessionBuilder::with_root` to configure a four-level namespace (let’s call this the root). In this case, all three-level child namespaces under the root are automatically registered as `LanceCatalogProviders`. This four-level namespace acts like a “catalog of catalogs” and is purely for convenience—instead of calling `add_catalog()` for each catalog individually.

2. We can always use S`essionBuilder::add_catalog` to manually register a catalog provider, regardless of whether the four-level namespace is configured.

3. I think what you’re really concerned about might be whether we can write queries against two- or three-level tables (e.g., `SELECT * FROM db.tb or SELECT * FROM tb`) when the four-level namespace is in use. **The answer is yes**. To do this, we need to configure the default catalog and schema names. If we don’t set them explicitly, they default to "`datafusion`" (for catalog) and "`public`" (for schema). That means `db.tb `would be interpreted as `datafusion.db.tb`, and `tb` as datafusion.public.`tb`.

4. I agree this is an important point. I’ve added methods to configure the default catalog and schema, and included examples in the tests.

majin1102:
Ready for another review. @wjones127 

Please take a look when you have time

wjones127 (approved):
Thanks for addressing my concerns. This looks good!

PR #5689, opened 2026-01-12
feat: use independent region manifest for MemWAL

touch-of-grey:
Based on discussion with @jackye1995 past few days, update the current MemWAL structure to introduce region specific manifest for independent update so that it can more efficiently run updates in streaming engines like Flink. This mostly comes from draft from @jackye1995, but separate out the MemWAL index part first with some my custom edits.

### DIFF
--- a/docs/src/format/table/mem_wal.md
+++ b/docs/src/format/table/mem_wal.md
@@ -623,7 +623,7 @@ Readers **MUST** merge results from multiple data sources (base table, flushed M
 
 When the same primary key exists in multiple sources, the reader must keep only the newest version based on:
 
-1. **Generation number** (`_gen`): Higher generation wins. The base table has generation -1, flushed MemTables have positive integers starting from 1.
+1. **Generation number** (`_gen`): Higher generation wins. The base table has generation -1, MemTables have positive integers starting from 1.
 2. **Row address** (`_rowaddr`): Within the same generation, higher row address wins (later writes within a batch overwrite earlier ones).
 
 The ordering for "newest" is: highest `_gen` first, then highest `_rowaddr`.
@@ -690,67 +690,27 @@ For example, with a region spec using `bucket(user_id, 10)` and a filter `user_i
 
 Region pruning applies to both scan queries and prefilters in search queries.
 
-#### Indexed Read
+#### Indexed Read Plan
 
 When data is merged from a flushed MemTable to the base table, the base table's indexes are rebuilt asynchronously by the [base table index builders](#base-table-index-builder).
 During this window, the merged data exists in the base table but is not yet covered by the base table's indexes.
 
 Without special handling, indexed queries would fall back to expensive full scans for the unindexed part of the base table.
-To maintain indexed read performance, the query planner should use `index_catchup` progress to determine the optimal data source for each query:
+To maintain indexed read performance, the query planner should use `index_catchup` progress to determine the optimal data source for each query.
 
-```
-For indexed query on region R using index I:
-  merged_gen = merged_generations[R]
-  index_gen = index_catchup[I].caught_up_generations[R]  # defaults to merged_gen if absent
-
-  if index_gen >= merged_gen:
-    # Base table index is caught up - optimal path
-    base_table.indexed_query(I)
-    memtable.query(gen > merged_gen)
-  else:
-    # Base table index is behind - use flushed MemTable indexes for the gap
-    base_table.indexed_query(I)                           # covers data up to index_gen
-    flushed_memtable.indexed_query(gen in (index_gen, merged_gen])  # uses pre-built indexes
-    memtable.query(gen > merged_gen)
-```
-
+The key insight is that flushed MemTables serve as a bridge between the base table's index catchup and the current merged state.
+For a query that requires a specific index for acceleration, when `index_gen < merged_gen`, 
+the generations in the gap `(index_gen, merged_gen]` have data already merged in the base table but are not covered by the base table's index.
 Since flushed MemTables contain pre-built indexes (created during [MemTable flush](#memtable-flush)), queries can use these indexes instead of scanning unindexed data in the base table.
 This ensures all reads remain indexed regardless of how far behind the async index builder is.
 
-The key insight is that flushed MemTables serve as a bridge between the base table's index catchup and the current merged state.
-When `index_gen < merged_gen`, the generations in the gap `(index_gen, merged_gen]` have data in the base table but are not covered by the base table's indexes.
-Instead of falling back to a full scan for this gap, the query planner uses the flushed MemTable indexes which were pre-built during flush.
-This maintains indexed query performance even during periods of heavy write activity where async index rebuilding falls behind.
-
 See [Appendix 4: Index Catchup Example](#appendix-4-index-catchup-example) for a detailed timeline showing how this works in practice.
-
 
 ### Query Execution
 
-Query execution unions datasets within each region and deduplicates by primary key.
-Deduplication uses two virtual columns: `_gen` (generation number, same value for all rows in a generation) and `_rowaddr` (row address within the dataset).
-The ordering for "newest" is: highest `_gen` first, then highest `_rowaddr` (within the same generation, later rows win).
+Query execution unions datasets within each region and deduplicates by primary key according to [LSM tree merge read](#lsm-tree-merge-read).
 
-A single write batch may contain duplicate primary keys. Query execution must deduplicate, keeping only the newest row for each key.
-
-For scan queries with filters, the plan unions datasets from relevant regions (after pruning), deduplicates within each region, then applies any remaining filters.
-Early termination is possible with a streaming deduplicate operator.
-
-For vector search queries, each generation's index is searched independently.
-Results from base table and MemTable generations use the same distance metric since they share identical index configuration.
-The bloom filter is used to detect stale results: for each candidate from generation G, check if the primary key exists in any bloom filter from generations > G.
-If found, the candidate is filtered out because a newer version exists that did not match the query as well.
-
-For full-text search queries, corpus statistics (document count, term frequencies) are aggregated across all generations to compute global BM25 parameters.
-Each generation's FTS index is then searched with the global parameters, producing comparable scores.
-Staleness filtering uses bloom filters similar to vector search.
-
-For point lookups, the planner can short-circuit by checking newest generations first and returning immediately when the key is found.
-
-### Expected Query Plans
-
-This section provides query plan expectations using custom execution nodes optimized for MemWAL's data model.
-See [Appendix 3: Execution Nodes](#appendix-3-execution-nodes) for detailed node descriptions.
+The next few subsections go through the query plan expectations using custom execution nodes optimized for MemWAL's data model.
 
 All query plans assume the following MemWAL setup:
 
@@ -766,6 +726,9 @@ region_B:
   gen 2: flushed_gen_2
   gen 3: in_memory_memtable
 ```
+
+Existing Lance index optimizations (scalar indexes, fragment pruning, etc.) continue to apply within each scan and is omitted.
+See [Appendix 3: Execution Nodes](#appendix-3-execution-nodes) for uncommon execution nodes we use here for optimized performance.
 
 #### Scan Queries
 
@@ -785,8 +748,6 @@ DeduplicateExec: partition_by=[pk], order_by=[_gen DESC, _rowaddr DESC]
     ScanExec: region_B[gen=2], filter=[pushed_down]
     ScanExec: region_B[gen=1], filter=[pushed_down]
 ```
-
-Existing Lance index optimizations (scalar indexes, fragment pruning, etc.) continue to apply within each scan.
 
 #### Point Lookups
 
@@ -809,8 +770,6 @@ CoalesceFirstExec: return_first_non_null
     TakeLastExec: region_A[gen=1], filter=[pk = target]
   TakeLastExec: base_table[gen=-1], filter=[pk = target]
 ```
-
-Existing Lance index optimizations (scalar indexes, fragment pruning, etc.) continue to apply within each lookup.
 
 #### Vector Search Queries
 
@@ -1017,11 +976,11 @@ This example demonstrates how `index_catchup` enables indexed reads during async
 Generation:         1       2       3       4       5       6
                     |       |       |       |       |       |
 State:           merged  merged  merged  merged flushed  active
-                    |       |       |       |       |       |
-Base IVF index:  [-- covers 1-3 --]
-                                    ↑               ↑       ↑
-                              index_gen=3    merged_gen=4   |
-                                                    current_gen=6
+                    |       |       |       |               |
+Base IVF index:  [--   covers 1-3   --]     |               |
+                            ↑               ↑               ↑
+                        index_gen=3    merged_gen=4         |
+                                                      current_gen=6
 ```
 
 In this example:
@@ -1032,7 +991,7 @@ In this example:
 - **Generation 5** is flushed to disk (not yet merged to base table)
 - **Generation 6** is the active in-memory MemTable
 
-#### Read Strategy for Vector Search
+#### Example Read Strategy for Vector Search
 
 Without `index_catchup` tracking, the query planner would need to perform an expensive full scan on the base table for generation 4.
 With `index_catchup`, the planner knows exactly which data is indexed and can use flushed MemTable indexes for the gap:
@@ -1045,12 +1004,3 @@ With `index_catchup`, the planner knows exactly which data is indexed and can us
 | Active MemTable | 6 | Use in-memory IVF index |
 
 All data sources provide indexed access, maintaining query performance during async index rebuild.
-
-#### Why Flushed MemTables Are Retained
-
-Flushed MemTable files are not garbage collected until:
-
-1. Their data has been merged to the base table (`gen <= merged_gen`)
-2. All maintained indexes have caught up (`gen <= index_gen` for all indexes)
-
-This ensures flushed MemTable indexes remain available to bridge the gap between `index_gen` and `merged_gen`.


jackye1995:
Thanks for pushing progress on this, I will take a look tomorrow morning!

docs/src/format/table/mem_wal.md
```diff
@@ -0,0 +1,1349 @@
...
+Lance MemTable & WAL (MemWAL) specification describes an Log-Structured-Merge (LSM) tree architecture for Lance tables, enabling high-performance streaming write workloads.
+
+## Prerequisites
+
+### Unenforced Primary Key
+
+MemWAL only works for Lance tables with unenforced primary key defined,
+also the unenforced primary key:
+
+- Must have a [btree index](./index/scalar/btree.md)
```
jackye1995: based on our latest discussion, I think we can remove this restriction. Point lookup is not a required feature, we mainly just need to check point existence. So we should instead just build a table-level bloom filter.

docs/src/format/table/mem_wal.md
```diff
@@ -0,0 +1,1349 @@
...
+- Must be included in the region spec's `source_ids` if a region spec is specified (see [Region Spec](#region-sepc) for more details)
```
jackye1995: I think we can remove the prerequisite section, because this can go to the Region Spec section. And the "IVF Vector Index" should come in the section that we describe the "MemWAL Index Details"

docs/src/format/table/mem_wal.md
```diff
@@ -0,0 +1,1349 @@
+# MemTable & WAL Specification
```
jackye1995: let's mark this as experimental so we can vote for that later. But I think we can let this PR go through first since in general we already know we want this feature.

docs/src/format/table/mem_wal.md
```diff
@@ -0,0 +1,1349 @@
...
+
+## Overall Architecture
+
+![MemWAL Overview](../../images/mem_wal_overview.png)
+
+### Base Table
+
+Under the MemWAL setup, the Lance table is called the **base table**.
+
+### MemWAL Region
```
jackye1995: just Region is fine, no need for "MemWAL Region" (sorry my bad)

docs/src/format/table/mem_wal.md
```diff
@@ -0,0 +1,1349 @@
...
+
+A **MemWAL Index** is the centralized structure for all MemWAL metadata for a base table.
+A table has at most one MemWAL index.
+It stores:
+
+- **Configuration**: Region specs defining how rows map to regions, and which indexes to maintain
+- **Merge progress**: Last generation merged to base table for each region
+- **Region snapshots**: Point-in-time snapshot of all region states for read optimization
+
+The index is the source of truth for **configuration** and **merge progress**, but region state snapshots are read-only (each region's manifest is authoritative for its own state).
```
jackye1995: "are for read-only"

docs/src/format/table/mem_wal.md
```diff
@@ -0,0 +1,1349 @@
...
+
+Writers read the MemWAL index to get configuration (region specs, maintained indexes) before writing.
+Readers use the index to get a snapshot of all region states, then query each region's data alongside the base table and merge results at runtime.
```
jackye1995: this should link to reader consistency guarantee section

docs/src/format/table/mem_wal.md
```diff
@@ -0,0 +1,1349 @@
...
+
+A background process periodically updates region snapshots by listing regions and loading their manifests.
```
jackye1995: this should link to the index builder process.

docs/src/format/table/mem_wal.md
```diff
@@ -0,0 +1,1349 @@
...
+See [MemWAL Index Details](#memwal-index-details) for the complete structure.
+
+## Region Architecture
```
jackye1995, line 56: It feels like there are too many layers of nesting. Can we do the overall architecture and region architecture , introduce basic terms first, and then have a level 2 sections for each of the important items (memWAL Index, MemTable, WAL, Flushed MemTable, Region Manifest, ...)

docs/src/format/table/mem_wal.md
```diff
@@ -0,0 +1,1349 @@
...
+Under the MemWAL setup, the Lance table is called the **base table**.
+
+### MemWAL Region
+
+**MemWAL Region** is the main unit to horizontally scale out writes.
+Each region has exactly one active writer at any time, using **epoch-based fencing** to guarantee single-writer semantics without distributed coordination.
+Writers claim a region by incrementing the writer epoch, then write data to that region.
+Data in each region is merged into the base table gradually in the background.
+
+#### Region Identifier
```
jackye1995: these details (identifier, region spec, region expression, region transform) should be a part of the MemWAL index since that holds the source of truth of these info.

docs/src/format/table/mem_wal.md
```diff
@@ -0,0 +1,1349 @@
...
+1. Compute the next version number
+2. Write the manifest to `{bit_reversed_version}.binpb` using put-if-not-exists
+3. In parallel best-effort write to `version_hint.json` with `{"version": <new_version>}` (failure is acceptable)
+
+To read the latest manifest version:
+
+1. Read `version_hint.json` to get the latest version hint. If not found, start from version 1
+2. Check existence for subsequent versions from the starting version
+3. Continue until a version is not found
+4. The latest version is the last found version
```
jackye1995, line 153: One thing I forgot to mention is that this works only when the write rate is significantly lower than the read rate, so you don't have a situation where HEAD keeps getting new versions and cannot terminate. This works in this case because region manifest is only updated by the cases when region metadata has changed, not on every write.

docs/src/format/table/mem_wal.md
```diff
@@ -0,0 +1,1349 @@
...
+
+#### Index Details Schema
+
+The `index_details` field in `IndexMetadata` contains a `MemWalIndexDetails` protobuf message with:
+
+| Field | Type | Description |
+|-------|------|-------------|
+| `snapshot_timestamp` | int64 | When the index was built (Unix timestamp in seconds) |
+| `num_regions` | uint32 | Number of regions in the snapshot |
+| `inline_snapshots` | bytes | Inline snapshot data for small region counts (optional) |
```
jackye1995: should specify that this should only be read when num_regions > 0. When there is no region, `inline_snapshots` can be None or an empty Lance file with 0 rows but proper schema, it does not matter.

docs/src/format/table/mem_wal.md
```diff
@@ -0,0 +1,1349 @@
...
+**Inline storage**: For small region counts, snapshots are serialized as a Lance file and stored in the `inline_snapshots` field.
+This keeps the index metadata compact while avoiding an additional file read for common cases.
+
+**External Lance file**: For large region counts, snapshots are stored as a Lance file at `_indices/{UUID}/index.lance`.
+This file uses standard Lance format with the region snapshot schema, enabling efficient columnar access and compression.
+
+#### Region Snapshot Schema
+
+Region snapshots are stored as a Lance file with one row per region.
+The schema has one column per `RegionManifest` field, with region fields as columns:
```
jackye1995: this is missing the columns in the region specs

docs/src/format/table/mem_wal.md
```diff
@@ -0,0 +1,1349 @@
...
+A writer must claim a region before performing any write operations:
+
+1. Load the latest region manifest
+2. Increment `writer_epoch` by one
+3. Atomically write a new manifest
+4. If the write fails (another writer claimed the epoch), reload the manifest and retry with a higher epoch
+5. Read WAL entries sequentially from `replay_after_wal_id + 1` until not found
+6. Replay valid WAL entries (those with `writer_epoch` ≤ current epoch) to reconstruct the MemTable with 1:1 fragment mapping (each WAL entry becomes one MemTable fragment)
+
+After recovery, the writer tracks subsequent fragment mappings as new WAL flushes occur (see [WAL Flush](#wal-flush)).
```
jackye1995: after initialization

docs/src/format/table/mem_wal.md
```diff
@@ -0,0 +1,1349 @@
...
+
+### Writer Fencing
```
jackye1995: writer fencing should be after MemTable Flush section otherwise its missing the right context

docs/src/format/table/mem_wal.md
```diff
@@ -0,0 +1,1349 @@
...
+
+Each write operation follows this sequence:
+
+1. Validate incoming records
+2. Insert records into the MemTable, creating an in-memory fragment (immediately queryable via full scan)
+3. Track the Lance data file in the new fragment for pending WAL flush
+4. Optionally trigger WAL flush based on size, count, or time thresholds
+5. For [durable writes](#writer-configuration), wait for WAL flush to complete before returning
+6. For [indexed writes](#writer-configuration), update MemTable indexes before returning:
+    - Insert primary keys into the btree index
```
jackye1995: this should be updated with our bloom filter discussion

docs/src/format/table/mem_wal.md
```diff
@@ -0,0 +1,1349 @@
...
+    - For each vector column with a base table index: encode and insert into HNSW graph
```
jackye1995: this is no longer true, vector index should be created in the same config as the base table vector index

docs/src/format/table/mem_wal.md
```diff
@@ -0,0 +1,1349 @@
...
+
+!!!note
+    The region manifest is **not** updated on every WAL flush. The `wal_id_last_seen` field is a hint that can be updated:
+    
+    1. **During MemTable flush** - when the region manifest is updated anyway
+    2. **By a background index builder** - which scans WAL entries and updates each region's `wal_id_last_seen`
+
+    This keeps the hot write path fast. On recovery, the writer reads WAL entries sequentially starting from `wal_id_last_seen + 1` to discover any WAL entries beyond what the manifest indicates.
+
+The [durable write](#writer-configuration) option also impacts flush behavior:
```
jackye1995: WAL flush behavior

docs/src/format/table/mem_wal.md
```diff
@@ -0,0 +1,1349 @@
...
+| Non-durable write | Buffer until threshold, return immediately | Batched Lance files (fewer S3 operations) |
+
+### MemTable Indexing
+
+MemTable indexing differs from base table indexing to balance write performance with query capability.
+Rather than maintaining all base table indexes, MemTables maintain a subset specified in the [MemWAL Index](#memwal-index).
+
+#### Primary Key Index
+
+The **primary key btree index** is always maintained for every MemTable, regardless of other index configuration.
```
jackye1995: this should be updated with bloom filter

docs/src/format/table/mem_wal.md
```diff
@@ -0,0 +1,1349 @@
...
+
+The primary key index is implemented as an in-memory `BTreeMap<OrderableScalarValue, RowId>` where `OrderableScalarValue` wraps Arrow's `ScalarValue` with `Ord` implementation (see `lance-index::scalar::btree::OrderableScalarValue`).
+For multi-column primary keys, the map key is a tuple of `OrderableScalarValue` for each column.
+
+#### Vector Indexes
+
+MemTables **automatically inherit** vector indexing from base table indexes.
+This inheritance is critical for **distance comparability** across generations.
+
+**Why inheritance is required:**
```
jackye1995: remove this line

docs/src/format/table/mem_wal.md
```diff
@@ -0,0 +1,1349 @@
...
+
+When ranking search results across generations, distances must be comparable:
+
+| Component | Must Match Base Table? | Reason |
+|-----------|------------------------|--------|
+| Quantization (PQ codebook, SQ params) | **Yes** | Determines distance calculation |
+| IVF centroids | No | Only affects partition assignment |
```
jackye1995: we should still match this because it impacts recall

docs/src/format/table/mem_wal.md
```diff
@@ -0,0 +1,1349 @@
...
+| Search structure (HNSW vs flat) | No | Only affects search efficiency |
```
jackye1995: should mention that distance is precise so comparable

docs/src/format/table/mem_wal.md
```diff
@@ -0,0 +1,1349 @@
...
+
+If MemTable used independent quantization, distances from MemTable and base table would not be comparable, leading to incorrect ranking.
+
+**Inheritance behavior:**
+
+For each vector index on the base table, MemTable automatically:
+
+1. **Inherits quantization parameters**: PQ codebook, SQ min/max, or no quantization (flat)
+2. **Encodes vectors** using inherited quantization during writes
+3. **Stores both** raw vectors (for potential refinement) and quantized codes
```
jackye1995: should only store quantized

docs/src/format/table/mem_wal.md
```diff
@@ -0,0 +1,1349 @@
...
+4. **Uses HNSW** as the search structure (optimal for small MemTable scale)
```
jackye1995: this is wrong, remvove it. It should be just using the same as base table

docs/src/format/table/mem_wal.md
```diff
@@ -0,0 +1,1349 @@
...
+
+| Base Table Index | MemTable Inherits | MemTable Search |
+|------------------|-------------------|-----------------|
+| IVF-PQ | PQ codebook | HNSW on PQ codes |
```
jackye1995: this is completely wrong. the MemTable should also use IVF-PQ with the same centroid and PQ codebook. Similarly the other ones are also wrong.

docs/src/format/table/mem_wal.md
```diff
@@ -0,0 +1,1349 @@
...
+  1. Search MemTable HNSW → candidates with quantized distances
+  2. Search base table IVF-PQ → candidates with quantized distances
+  3. Distances are COMPARABLE (same quantization)
+  4. Sort by distance directly
+  5. Apply staleness filtering
+```
+
+!!!warning "PQ Codebook Migration"
+    When the base table's PQ codebook is retrained, MemTable must switch to the new codebook.
+    During migration, maintain compatibility by:
```
jackye1995: should mention that this means the base table vector index should not change the codebook. To switch, a migration is required to create another vector index with the new codebook, MemTable will be configred to have both vector indexes, and eventually the old index will be dropped.

docs/src/format/table/mem_wal.md
```diff
@@ -0,0 +1,1349 @@
...
+**Inheritance behavior:**
+
+| Inherited | Not Inherited |
+|-----------|---------------|
+| Tokenizer type (simple, ngram, jieba, etc.) | Corpus statistics (IDF, avgdl) |
+| Language settings | Document frequencies |
+| Token filters (lowercase, stemming, etc.) | Posting lists |
+| Position storage setting | |
+
+**Why corpus statistics are NOT inherited:**
```
jackye1995: remove this subheader

docs/src/format/table/mem_wal.md
```diff
@@ -0,0 +1,1349 @@
...
+4. **Uses HNSW** as the search structure (optimal for small MemTable scale)
+
+| Base Table Index | MemTable Inherits | MemTable Search |
+|------------------|-------------------|-----------------|
+| IVF-PQ | PQ codebook | HNSW on PQ codes |
+| IVF-SQ | SQ parameters | HNSW on SQ codes |
+| IVF-FLAT | Nothing (no quantization) | HNSW on raw vectors |
+| IVF-HNSW-PQ | PQ codebook | HNSW on PQ codes |
+
+**Write path with inheritance:**
```
jackye1995: remove sub-header

docs/src/format/table/mem_wal.md
```diff
@@ -0,0 +1,1349 @@
...
+
+```
```
jackye1995: no need to surround with code block, also needs to be updated with latest info

docs/src/format/table/mem_wal.md
```diff
@@ -0,0 +1,1349 @@
...
+Write batch with vectors:
+  1. Load PQ/SQ codebook from base table index (cached)
+  2. Encode vectors using inherited quantization
+  3. Store raw vectors + quantized codes in MemTable
+  4. Insert into HNSW graph for search
+  5. On MemTable flush, serialize both raw vectors and codes
+```
+
+**Query path with comparable distances:**
```
jackye1995: remove sub-header

docs/src/format/table/mem_wal.md
```diff
@@ -0,0 +1,1349 @@
...
+
+```
```
jackye1995: no need to surround with code block, also needs to be updated with latest info

docs/src/format/table/mem_wal.md
```diff
@@ -0,0 +1,1349 @@
...
+
+These statistics are specific to each corpus (generation). If MemTable used base table's statistics, scores would be incorrect because:
+- IDF would be wrong (term rarity differs between 10K MemTable vs 10M base table)
+- avgdl would be wrong (document length distribution may differ)
+
+**Global BM25 scoring (Lucene-style):**
+
+At query time, statistics are **aggregated across all generations** for globally-comparable BM25 scores:
+
+```
```
jackye1995: no need to be in code block

docs/src/format/table/mem_wal.md
```diff
@@ -0,0 +1,1349 @@
...
+  IDF("machine") = log(1 + (N_global - df_global) / (df_global + 0.5))
+
+Step 4: Search each generation with global stats
+  Each FTS index returns candidates scored with global IDF and avgdl
+  Scores are now COMPARABLE across generations
+
+Step 5: Merge and rank globally
+```
+
+This follows the same pattern as [Apache Lucene's multi-segment BM25 scoring](https://github.com/apache/lucene), where:
```
jackye1995: no need to mention Lucene

docs/src/format/table/mem_wal.md
```diff
@@ -0,0 +1,1349 @@
...
+
+| Statistic | Description | Used For |
+|-----------|-------------|----------|
+| `doc_count` | Documents in this index | Global N |
+| `sum_total_term_freq` | Sum of all document lengths | Global avgdl |
+| `df(term)` | Documents containing term | Global IDF |
+
+These are summed at query time to compute global BM25 parameters.
+
+#### In-Memory Index Structure
```
jackye1995: this section is not needed

docs/src/format/table/mem_wal.md
```diff
@@ -0,0 +1,1349 @@
...
+
+| Component | Size | Notes |
+|-----------|------|-------|
+| HNSW graph structure | ~5-10MB | Neighbors + distances |
+| Thread-safe overhead | ~1MB | `RwLock` per node |
+| PQ codebook (cached) | ~1MB | Shared across MemTables |
+| PQ codes storage | ~1-2MB | 64 bytes/vector typical |
+| **Total** | ~10-15% of MemTable size |
+
+#### Index Update Timing
```
jackye1995: this section is not needed

docs/src/format/table/mem_wal.md
```diff
@@ -0,0 +1,1349 @@
...
+3. **Other indexes**: Update according to index-specific logic
+
+Entries reference MemTable fragment IDs and row offsets.
+
+#### Flushed MemTable Index Caching
+
+When a MemTable is flushed to storage:
+
+1. In-memory indexes are serialized to disk in the flushed MemTable's `_indices/` directory:
+    - **Primary key btree**: Written as Lance btree index format
```
jackye1995: no need to list, just based on lance table index format is sufficient. The only exception is the bloom filter. It should be serialized as a single bloom filter file in the generation directory

docs/src/format/table/mem_wal.md
```diff
@@ -0,0 +1,1349 @@
...
+
+- A flushed MemTable listed in `flushed_generations` may have been merged and garbage collected
+- New flushed MemTables may exist that are not yet in `flushed_generations`
+- WAL entries may have advanced beyond what the index shows
+
+The `snapshot_timestamp` field indicates when the index was built; readers can use this to estimate staleness and decide whether to refresh.
+
+For authoritative state, readers may load individual region manifests directly from `_memwal/{region_uuid}/manifest/`.
+
+## Writer Expectations
```
jackye1995, line 366: within the writer expectation section, we need to indicate which process happens synchronously and which happen asynchronously. Add a summary about it after Writer Configuration section, before having all the other sub-sections.

docs/src/format/table/mem_wal.md
```diff
@@ -0,0 +1,1349 @@
...
+| **Manifest source** | Reads region manifests directly, or uses MemWAL Index |
+
+**Strong consistency** requires both:
+
+1. Access to in-memory MemTables for **all** regions involved in the query
+2. Reading region manifests directly (not via MemWAL Index)
+
+Otherwise, the query is **eventually consistent**.
+
+#### Consistency Matrix
```
jackye1995: remove this section

docs/src/format/table/mem_wal.md
```diff
@@ -0,0 +1,1349 @@
...
+
+| MemTable Access | Manifest Source | Consistency |
+|-----------------|-----------------|-------------|
+| All regions | Region manifest | **Strong** |
+| All regions | MemWAL Index | Eventually consistent (index may be stale) |
+| Partial/None | Region manifest | Eventually consistent (missing unflushed data) |
+| Partial/None | MemWAL Index | Eventually consistent (both sources of staleness) |
+
+#### Sources of Staleness
```
jackye1995: remove this section

docs/src/format/table/mem_wal.md
```diff
@@ -0,0 +1,1349 @@
...
+
+- **Missing MemTable access**: Unflushed data in a writer's in-memory MemTable is not visible
+- **Stale MemWAL Index**: Newly flushed MemTables are not visible until the index is rebuilt
+- **Stale region manifest cache**: If readers cache region manifests, newly flushed MemTables may not be visible
+
+### Query Planning
```
jackye1995, line 656: describe this section using human words, not a code block with bullet points.

docs/src/format/table/mem_wal.md
```diff
@@ -0,0 +1,1349 @@
...
+2. **Flushed MemTables**: generations in range `(merged_generation, current_generation)` from region manifest or MemWAL Index
+3. **In-memory MemTable**: generation = `current_generation` (if accessible)
+
+### Query Execution
+
+Query execution unions all datasets and deduplicates by primary key.
+
+**Deduplication ranking** uses two virtual columns:
+
+- `_gen`: Generation number (-1 for base, 1+ for MemTables)
```
jackye1995: same value for all rows in the same generation

docs/src/format/table/mem_wal.md
```diff
@@ -0,0 +1,1349 @@
...
+- `_rowaddr`: Row address within the dataset
+
+The ordering for "newest" is: highest `_gen` first, then highest `_rowaddr` (within the same generation, later rows win).
+
+A single write batch may contain duplicate primary keys. Query execution must deduplicate, keeping only the newest row for each key.
+
+For detailed query plans by query type, see [Appendix 3: Query Execution Examples](#appendix-3-query-execution-examples).
+
+## Durability Guarantees
```
jackye1995: remove this section, duplicate of writer expectation.

docs/src/format/table/mem_wal.md
```diff
@@ -0,0 +1,1349 @@
...
+|------|-----------|---------|
+| Durable write | Data persisted to object storage before return | Higher (S3 PUT latency) |
+| Non-durable write | Data in memory only until next flush | Lower (memory write only) |
+
+Writers using non-durable writes accept potential data loss between the last flush and a crash.
+The flush interval and buffer size thresholds control the maximum data at risk.
+
+For configuration details, see [Writer Configuration](#writer-configuration).
+
+## MemWAL Optimizations
```
jackye1995: remove this section, optimizations have been described in details in previous sections.

docs/src/format/table/mem_wal.md
```diff
@@ -0,0 +1,1349 @@
...
+
+Vector search requires special handling for staleness detection. Consider this scenario:
+
+1. Base table has `pk=123` with vector `v1` that matches the query (distance = 0.1)
+2. MemTable has `pk=123` with updated vector `v2` that doesn't match (distance = 0.9)
+3. KNN search on base table returns `pk=123` (good score)
+4. KNN search on MemTable does NOT return `pk=123` (v2 is far from query)
+5. Without staleness detection, the old version from base table would be incorrectly returned
+
+The solution uses the **primary key btree index** to filter out stale results:
```
jackye1995: uses bloom filter.

docs/src/format/table/mem_wal.md
```diff
@@ -0,0 +1,1349 @@
...
+  SortExec: order_by=[_dist ASC]
+    FilterStaleExec: pk_indexes=[btree[gen=3], btree[gen=2], btree[gen=1]]
+      UnionExec
+        KNNExec: dataset[gen=3], k=k*overfetch  -- highest gen first
+        KNNExec: dataset[gen=2], k=k*overfetch
+        KNNExec: dataset[gen=1], k=k*overfetch
+        KNNExec: dataset[gen=-1], k=k*overfetch
+```
+
+For each candidate from generation G, `FilterStaleExec` checks if the primary key exists in btree indexes of generations > G. If found, the candidate is filtered out. The newer version doesn't participate in ranking since it didn't match the query.
```
jackye1995: "since it didn't match the query." should be more clear, "since it is not as good compared to other rows in the newer generations" or something similar.

docs/src/format/table/mem_wal.md
```diff
@@ -0,0 +1,1349 @@
...
+
+#### Full-Text Search Queries
+
+Full-text search has two challenges across generations:
+
+1. **Staleness**: A document may match a query in an older generation but not in a newer generation after the text was updated
+2. **Score comparability**: BM25 scores depend on corpus statistics which differ per generation
+
+**Solution:** Use global BM25 scoring (Lucene-style) with staleness filtering.
```
jackye1995: do not mention Lucene. Also use full sentence.

docs/src/format/table/mem_wal.md
```diff
@@ -0,0 +1,1349 @@
...
+
+2. **Region manifest is eventually consistent**: It may lag behind MemWAL Index after crashes, but is eventually updated by subsequent mergers.
+
+3. **Conflict resolution uses MemWAL Index**: When a commit conflicts, the merger checks the conflicting commit's MemWAL Index, not the region manifest.
+
+4. **No progress regression**: Because MemWAL Index is updated atomically with data, concurrent mergers cannot regress the merge progress.
+
+5. **Crash recovery is safe**: If a merger crashes after table commit but before region manifest update, subsequent mergers use MemWAL Index to skip already-merged generations.
+
+### Appendix 3: Query Execution Examples
```
jackye1995: this should not be in appendix, this should be a part of the reader expectations section. We just need to mention appendix 4 for all the new execution nodes we have introduced.

docs/src/format/table/mem_wal.md
```diff
@@ -0,0 +1,1349 @@
...
+
+This appendix provides query plan examples. All examples assume the planner has collected datasets as:
+
+```
+datasets = {
```
jackye1995: this example does not take into consideration of multiple regions. Update all the query plans to take into account of multiple regions. This should impact some queries like scan since rows across regions do not need to be deduped.

docs/src/format/table/mem_wal.md
```diff
@@ -0,0 +1,1349 @@
...
+}
+```
+
+The core pattern for all queries:
+
+1. **Union** all datasets with their generation number
+2. **Deduplicate** by primary key, ranking by `(_gen DESC, _rowaddr DESC)`
+3. **Apply** query-specific operators (filter, sort, limit)
+
+#### Scan Queries
```
jackye1995: For any filter, it should do pruning against region specs first to get the necessary regions to scan. This applies to both scan queries, also prefitlers in search queries. This is a step prior to physical execution, we should just have a dedicated section in reader expectation section for Region Pruning to talk about it as a part of planning.

docs/src/format/table/mem_wal.md
```diff
@@ -0,0 +1,1349 @@
...
+
+Here is a recap of the storage layout with all the files and concepts defined so far:
+
+```
+{table_path}/
+├── _indices/
+│   └── {index_uuid}/                    # MemWAL Index (uses standard index storage)
+│       └── regions.binpb                # Serialized region snapshots (protobuf binary)
+│
+└── _memwal/
```
jackye1995: should use `_mem_wal` to be consistent with file names.

jackye1995:
I updated the spec to also talk about indexed read and index builder, let me know if it looks good to you!

jackye1995:
Thanks for going through my tedious comments! Let me know when you get the next PRs up I will review asap!

PR #5691, opened 2026-01-12
feat(compaction): add Python config for defer_index_remap

zhangyue19921010:
add Python support for defer_index_remap


zhangyue19921010:
Hi @wjones127 . Would u mind to take a look at your convenient ？Thanks a lot!

zhangyue19921010:
Hi @jackye1995   Just a follow up work for your `defer_index_remap` feature. Would u mind to take a look at your convenient? Thanks :)

Issue #3685, opened 2025-04-14
Cache DataFusion SessionContext in Session


wjones127:
We often construct a new `SessionContext` in `merge_insert`, which can be expensive. Instead, we should re-use the same one.

In fact, we can probably just have one of them, and set the target partitions, and spill configuration to new values when we convert to `SessionState`.

wjones127:
This now happens every time we do `LanceFilter::to_datafusion`. We should cache the SessionState for that too.

<img width="1007" height="739" alt="Image" src="https://github.com/user-attachments/assets/040b37c8-373a-4e98-a0c9-0c746c23261b" />

PR #5696, opened 2026-01-12
perf: reuse session context

jackye1995:
Fixes [#3685](https://github.com/lance-format/lance/issues/3685)

BTree index search went from over 100k allocations to just 3k allocations per query.


jackye1995:
Failing CIs are all expected errors in main CI, merging

PR #5703, opened 2026-01-12
fix: project_by_schema now reorders fields inside List<Struct> types

### LINKED ISSUE #5702: `project_by_schema` does not reorder fields inside `List<Struct>` types
## Summary

When reading fragments where fields are stored out of order (scrambled `fields` array in DataFile metadata), the `project_by_schema` function fails to reorder fields inside `List<Struct>` columns. This causes Arrow validation errors when constructing the final `StructArray`.

## Error Message

```
Invalid argument error: Incorrect datatype for StructArray field "list",
expected List(Field { name: "item", data_type: Struct([
  Field { name: "field_a", ... },
  Field { name: "field_b", ... }
]), ... })
got List(Field { name: "item", data_type: Struct([
  Field { name: "field_b", ... },  // <-- order is swapped
  Field { name: "field_a", ... }
]), ... })
```

## Root Cause

The `project` function in `rust/lance-arrow/src/lib.rs:798-827` recursively handles `Struct` fields but not `List<Struct>`:

```rust
fn project(struct_array: &StructArray, fields: &Fields) -> Result<StructArray> {
    for field in fields.iter() {
        if let Some(col) = struct_array.column_by_name(field.name()) {
            match field.data_type() {
                // TODO handle list-of-struct   <-- acknowledged but not implemented
                DataType::Struct(subfields) => {
                    let projected = project(col.as_struct(), subfields)?;
                    columns.push(Arc::new(projected));
                }
                _ => {
                    columns.push(col.clone());  // List<Struct> falls through here
                }
            }
        }
    }
    // ...
}
```

## Conditions to Trigger

The bug requires all of the following:

1. **Out-of-order field storage**: A fragment where `DataFile.fields` is not in sequential order (e.g., `[2, 8, 1, 5, ...]` instead of `[0, 1, 2, 3, ...]`)

2. **Schema with `List<Struct>`**: A column with nested structure like `struct<list: list<struct<...>>>`

3. **Schema evolution (optional but common)**: Missing fields that require null-filling, triggering the merge + project code path

## How It Happens

1. Fragment is written with fields stored in non-sequential order (this can happen legitimately)
2. When reading, the file reader returns data with inner struct fields in file order
3. `project_by_schema` is called to reorder columns to match the output schema
4. Top-level and direct `Struct` fields are reordered correctly
5. Fields inside `List<Struct>` are NOT reordered (bug)
6. Arrow's `StructArray::new()` validation fails due to field/column order mismatch

## Reproduction

Fragment metadata showing the issue:

```python
# Good fragment - fields in order
>>> frags[0].metadata.files[0]
DataFile(fields=[0, 1, 2, 3, 4, ...], column_indices=[0, 1, 2, 3, 4, ...], ...)

# Bad fragment - fields out of order
>>> frags[3].metadata.files[0]
DataFile(fields=[2, 8, 29, 1, 5, 7, ...], column_indices=[0, 1, 2, 3, 4, 5, ...], ...)
```

The scrambled `fields` array means field ID 2 is stored in column 0, field ID 8 in column 1, etc. This is valid Lance format, but the reader fails to properly reorder nested `List<Struct>` fields when reconstructing the output.

## Suggested Fix

Extend the `project` function to handle `List`, `LargeList`, and `FixedSizeList` types recursively:

```rust
fn project(struct_array: &StructArray, fields: &Fields) -> Result<StructArray> {
    for field in fields.iter() {
        if let Some(col) = struct_array.column_by_name(field.name()) {
            match field.data_type() {
                DataType::Struct(subfields) => {
                    let projected = project(col.as_struct(), subfields)?;
                    columns.push(Arc::new(projected));
                }
                DataType::List(inner_field) => {
                    let list_arr = col.as_list::<i32>();
                    let projected_values = project_list_values(list_arr.values(), inner_field)?;
                    let projected_list = ListArray::new(
                        inner_field.clone(),
                        list_arr.offsets().clone(),
                        projected_values,
                        list_arr.nulls().cloned(),
                    );
                    columns.push(Arc::new(projected_list));
                }
                // Similar for LargeList, FixedSizeList
                _ => {
                    columns.push(col.clone());
                }
            }
        }
    }
    // ...
}
```

## Environment

- Lance version: 1.0.0-beta.8 (commit 1329bf440128047c279079ab7b0ff89a75d23c23)
- File version: 2.0

wjones127:
## Summary

- `project_by_schema` previously only handled field reordering for top-level columns and direct `Struct` fields, but not for fields nested inside `List<Struct>`, `LargeList<Struct>`, or `FixedSizeList<Struct>` types
- Added `project_array` helper that recursively handles these nested list types
- Added unit tests and integration test with checked-in test data that reproduces the original issue

## Test plan

- [x] Unit tests in `lance-arrow` verify `project_by_schema` correctly reorders fields inside `List<Struct>`
- [x] Integration test in `dataset_migrations.rs` reads test data that triggers the bug scenario
- [x] Test data created with fragment 0 having `List<Struct<a,b,c>>` and fragment 1 having reordered/missing inner struct fields
- [x] Verified that without the fix, reading fails with "Incorrect datatype for StructArray field expected List(Struct(...)) got List(Struct(...))"

Fixes #5702


test_data/list_struct_field_reorder/datagen.py
```diff
@@ -0,0 +1,108 @@
...
+# Append second fragment with reordered and missing inner struct fields
+lance.write_dataset(fragment1_data, dataset_path, mode="append")
+print("Appended fragment 1 with reordered inner struct")
+
+# Verify the test data was created correctly
+ds = lance.dataset(dataset_path)
+print(f"\nDataset created with {len(ds.get_fragments())} fragments")
+for i, frag in enumerate(ds.get_fragments()):
+    for df in frag.metadata.data_files():
+        print(f"Fragment {i}: fields={df.fields}")
```
wjones127: Instead of printing, can we make these assertions? Then it's clearer to the reader what structure we created.

Also, maybe we should assert there is a failure if we scan the data as part of this script.

test_data/v1.0.1/datagen.py
wjones127, line 1: Please move this to a test directory that is specific to the version of Lance it corresponds to. We can run this under 1.0.1, since that's the version that originally generated the data that caused this error.

rust/lance/src/dataset/tests/dataset_schema_evolution.rs
```diff
@@ -546,3 +548,104 @@ async fn prepare_initial_dataset_with_list_struct_col(version: LanceFileVersion)
...
     dataset
 }
+
+/// Regression test for issue #5702: project_by_schema should reorder fields inside List<Struct>.
+///
+/// This test simulates the scenario where a fragment's data file has fields stored in a
+/// different order than the schema expects. When reading such fragments, project_by_schema
+/// is called to reorder the columns, and it must handle nested List<Struct> types correctly.
+#[test]
+fn test_project_by_schema_list_struct_field_reorder_regression() {
```
wjones127: Since we have the test that uses test_data, I think we can drop this test.

PR #5709, opened 2026-01-14
feat: introduce MemWAL regional writer and MemTable reader

touch-of-grey:
Based on draft shared from @jackye1995 , cleanup and publish for review.

Currently regional writer can reach 300MB/s performance on Amazon S3, around 50MB/s under high backpressure based on the benchmark.

Memtable read performance is inline with Lance table with a single fragment in memory, and much faster than using Lance table and append 1 fragment per batch.

### DIFF
--- a/rust/lance/src/dataset/mem_wal/memtable/flush.rs
+++ b/rust/lance/src/dataset/mem_wal/memtable/flush.rs
@@ -96,6 +96,7 @@ impl MemTableFlusher {
 
         let random_hash = generate_random_hash();
         let generation = memtable.generation();
+        let gen_folder_name = format!("{}_gen_{}", random_hash, generation);
         let gen_path =
             flushed_memtable_path(&self.base_path, &self.region_id, &random_hash, generation);
 
@@ -115,12 +116,7 @@ impl MemTableFlusher {
 
         let last_wal_entry_position = memtable.last_flushed_wal_entry_position();
         let new_manifest = self
-            .update_manifest(
-                epoch,
-                generation,
-                gen_path.as_ref(),
-                last_wal_entry_position,
-            )
+            .update_manifest(epoch, generation, &gen_folder_name, last_wal_entry_position)
             .await?;
 
         info!(
@@ -131,7 +127,7 @@ impl MemTableFlusher {
         Ok(FlushResult {
             generation: FlushedGeneration {
                 generation,
-                path: gen_path.to_string(),
+                path: gen_folder_name,
             },
             rows_flushed: memtable.row_count(),
             covered_wal_entry_position: last_wal_entry_position,
@@ -205,6 +201,7 @@ impl MemTableFlusher {
 
         let random_hash = generate_random_hash();
         let generation = memtable.generation();
+        let gen_folder_name = format!("{}_gen_{}", random_hash, generation);
         let gen_path =
             flushed_memtable_path(&self.base_path, &self.region_id, &random_hash, generation);
 
@@ -290,12 +287,7 @@ impl MemTableFlusher {
 
         let last_wal_entry_position = memtable.last_flushed_wal_entry_position();
         let new_manifest = self
-            .update_manifest(
-                epoch,
-                generation,
-                gen_path.as_ref(),
-                last_wal_entry_position,
-            )
+            .update_manifest(epoch, generation, &gen_folder_name, last_wal_entry_position)
             .await?;
 
         info!(
@@ -306,7 +298,7 @@ impl MemTableFlusher {
         Ok(FlushResult {
             generation: FlushedGeneration {
                 generation,
-                path: gen_path.to_string(),
+                path: gen_folder_name,
             },
             rows_flushed: memtable.row_count(),
             covered_wal_entry_position: last_wal_entry_position,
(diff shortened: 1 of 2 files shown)


jackye1995:
Thanks for the fast turnaround! I will take a look tonight. Meanwhile, I think the code path deserves some benchmark, can you add that?

rust/lance-io/src/object_store.rs
```diff
@@ -647,6 +647,28 @@ impl ObjectStore {
         Ok(())
     }
 
+    /// Rename a file atomically, failing if the destination already exists.
+    ///
+    /// This is used for atomic writes where we want to detect conflicts
+    /// (e.g., another writer already wrote the same WAL entry).
+    ///
+    /// Returns `Err` with `AlreadyExists` if the destination file exists.
+    pub async fn rename_if_not_exists(&self, from: &Path, to: &Path) -> Result<()> {
```
jackye1995: why do we need rename_if_not_exists and copy_if_not_exists? We can fulfill rename operation with existing object_store APIs, see how we do that in commit.rs to do atomic manifest commit for local storage.

rust/lance/src/dataset/mem_wal/api.rs
```diff
@@ -0,0 +1,320 @@
...
+
+#[async_trait]
+impl DatasetMemWalExt for Dataset {
+    async fn initialize_mem_wal(&mut self, config: MemWalConfig) -> Result<()> {
+        // Validate that the dataset has a primary key (required for MemWAL)
+        let pk_fields = self.schema().unenforced_primary_key();
+        if pk_fields.is_empty() {
+            return Err(Error::invalid_input(
+                "MemWAL requires a primary key on the dataset. \
+                 Define a primary key using the 'lance-schema:unenforced-primary-key' field metadata.",
```
jackye1995: nit: Arrow field metadata

rust/lance/src/dataset/mem_wal/api.rs
```diff
@@ -0,0 +1,320 @@
...
+
+use super::write::MemIndexConfig;
+use super::write::RegionWriter;
+use super::RegionWriterConfig;
+
+/// Configuration for initializing MemWAL on a Dataset.
+#[derive(Debug, Clone, Default)]
+pub struct MemWalConfig {
+    /// Region specification for partitioning writes.
+    pub region_specs: Vec<RegionSpec>,
```
jackye1995: this should just take a single RegionSpec to begin with, and it should be optional (you can create a MemWAL index without region spec).

We can in the future add APIs like `add_region_spec` (we can add a TODO for those, don't need to add now)

rust/lance/src/dataset/mem_wal/api.rs
```diff
@@ -0,0 +1,320 @@
...
+        .await
+    }
+}
+
+impl Dataset {
+    /// Load vector index configuration with IVF centroids and PQ codebook.
+    ///
+    /// This opens the vector index and extracts the IVF model and product
+    /// quantizer needed for in-memory index maintenance.
+    async fn load_vector_index_config(
```
jackye1995: can these just be functions, not methods within Dataset?

rust/lance/src/dataset/mem_wal/config.rs
```diff
@@ -0,0 +1,123 @@
...
+    /// Whether to update indexes synchronously on each write.
+    ///
+    /// When true:
+    /// - Newly written data is immediately searchable via indexes
+    /// - Higher latency due to index update overhead
+    ///
+    /// When false:
+    /// - Index updates are deferred
+    /// - New data may not appear in index-accelerated queries immediately
+    pub indexed_writes: bool,
```
jackye1995: this name is confusing, because it can mean indexed vs not indexed. What about `sync_indexed_write`

rust/lance/src/dataset/mem_wal/manifest.rs
```diff
@@ -0,0 +1,421 @@
...
+                            "Failed to write manifest version {} for region {}: {}",
+                            version, self.region_id, e
+                        ),
+                        location!(),
+                    )
+                }
+            })?;
+
+        // Best-effort update version hint
+        self.write_version_hint(version).await;
```
jackye1995, line 231: this should log a warning if failed

rust/lance/src/dataset/mem_wal/manifest.rs
```diff
@@ -0,0 +1,421 @@
...
+
+        let pb_manifest = pb::RegionManifest::from(manifest);
+        let bytes = pb_manifest.encode_to_vec();
+
+        let put_opts = PutOptions {
+            mode: PutMode::Create, // PUT-IF-NOT-EXISTS
+            ..Default::default()
+        };
+
+        self.object_store
```
jackye1995: this should also differentiate between local and cloud, since local will use a rename to ensure atomicity?

rust/lance/src/dataset/mem_wal/manifest.rs
```diff
@@ -0,0 +1,421 @@
...
+            latest_found = hint;
+        } else if hint > 1 {
+            // Hint might be stale, scan from beginning
+            if self.version_exists(1).await? {
+                latest_found = 1;
+            }
+        }
+
+        // Parallel scan forward with batches of HEAD requests
+        let batch_size = 8;
```
jackye1995: this should be a config in region writer config

jackye1995: batch size can just be 2 as default

touch-of-grey:
All review comments have been addressed in commit cda38c01:

1. ✅ Removed `rename_if_not_exists`/`copy_if_not_exists` wrappers - using existing object_store APIs directly
2. ✅ Fixed "Arrow field metadata" typo
3. ✅ Changed `region_specs: Vec<RegionSpec>` to `region_spec: Option<RegionSpec>`, added TODO for `add_region_spec()` API
4. ✅ Moved `load_vector_index_config` and `load_ivf_pq_components` to standalone functions
5. ✅ Renamed `indexed_writes` to `sync_indexed_write` for clarity
6. ✅ Added comment clarifying that version hint failures are logged as warnings
7. ✅ Differentiated local vs cloud manifest writes (local uses temp file + atomic rename, cloud uses PUT-IF-NOT-EXISTS)
8. ✅ Made `manifest_scan_batch_size` configurable in `RegionWriterConfig` with default of 2

jackye1995:
Known CI failure, merging

PR #5713, opened 2026-01-14
feat: support List and Struct type for KeyValue in inserted_rows.rs

wojiaodoubao:
Using List and Struct type as primary key.

### DIFF
--- a/rust/lance/src/dataset/write/merge_insert.rs
+++ b/rust/lance/src/dataset/write/merge_insert.rs
@@ -2198,6 +2198,9 @@ impl Merger {
 mod tests {
     use super::*;
     use crate::dataset::scanner::ColumnOrdering;
+    use crate::dataset::write::merge_insert::inserted_rows::{
+        extract_key_value_from_batch, KeyExistenceFilter, KeyExistenceFilterBuilder,
+    };
     use crate::index::vector::VectorIndexParams;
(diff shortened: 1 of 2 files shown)


wojiaodoubao:
Hi @yanghua , could you help review this when you have time, thanks very much~

rust/lance/src/dataset/write/merge_insert.rs
```diff
@@ -4553,6 +4558,376 @@ mod tests {
...
+        let struct_array2 = StructArray::from(vec![
+            (
+                Arc::new(Field::new("first", DataType::Utf8, false)),
+                Arc::new(first_names2) as Arc<dyn Array>,
+            ),
+            (
+                Arc::new(Field::new("last", DataType::Utf8, false)),
+                Arc::new(last_names2) as Arc<dyn Array>,
+            ),
+        ]);
```
yanghua: This code snippet is duplicated many times. We can find a way to reduce the test code, for example, by defining some utility methods.

yanghua (commented):
For the feature, I have no concern. Two suggestions:

* try to reduce the test codes, there are many duplicated snippets we can reduce;
* can we add at least one test case to do an e2e test for the merge insert api(define a pk by using the supported types in this PR that try to support)

yanghua (approved):
+1

PR #5719, opened 2026-01-15
feat: make blob v2 dedicated threshold configurable

yanghua:


yanghua:
> Thank you for working on this, I hope this can be a column metadata like `blob_dedicated_size_threshold` instead.

OK, got it.

rust/lance-arrow/src/lib.rs
```diff
@@ -47,6 +47,9 @@ pub const ARROW_EXT_META_KEY: &str = "ARROW:extension:metadata";
 pub const BLOB_META_KEY: &str = "lance-encoding:blob";
 /// Arrow extension type name for Lance blob v2 columns
 pub const BLOB_V2_EXT_NAME: &str = "lance.blob.v2";
+/// Metadata key for overriding the dedicated blob size threshold (in bytes)
+pub const BLOB_DEDICATED_SIZE_THRESHOLD_META_KEY: &str =
```
Xuanwo, line 51: This key should be renamed to `lance-encoding:blob-dedicated-size-threshold`.

Please note that `lance.blob.v2` follows the naming convention of array extension types. It does not follow the style of Lance metadata keys.

rust/lance/src/dataset/blob.rs
```diff
@@ -187,6 +187,13 @@ impl BlobPreprocessor {
                 continue;
             }
 
+            let dedicated_threshold = field
+                .metadata()
+                .get(BLOB_DEDICATED_SIZE_THRESHOLD_META_KEY)
+                .and_then(|value| value.parse::<usize>().ok())
+                .filter(|&value| value > 0)
```
Xuanwo: I think we should make it at least larger than `DEDICATED_THRESHOLD`.

yanghua:
@xuanwo I have addressed your comments.

PR #5721, opened 2026-01-15
chore: change the FTS benchmark data distribution

BubbleCal:
switch to zipf distribution to match real world dataset distribution


wjones127:
@BubbleCal I already changed `random_sentence` to use zipf distributed words here: https://github.com/lance-format/lance/commit/bd18c036546bd273d5cd1f20efd1262e8ba52790#diff-05a6cb2a51514a3883006392e5a3b6990e908be79d42414de787ba5c5b623c87

I think it's also better because it uses real words and has the stop words ranked high in distribution.

PR #5722, opened 2026-01-15
fix: support system columns in dataset.take* operations

hamersaw:
Previously, "take*" operations did not support `_rowid`, `_rowoffset`, `_row_created_at_version`, and `_row_last_updated_at_version`. In this PR we add support for all of these columns.

We preserve these system columns through the initial schema projection so that they can be used to populate the correct flags when building the `ProjectionPlan` and `PhysicalProjection` structs.
- `_rowid` / `_rowaddr`: persisting these through to `ProjectionPlan` fields was enough to make them work
- `_rowoffset`: required additionally (1) stripping `ROW_OFFSET` field from `ProjectionPlan` `requested_output_expr` and (2) manually injecting column using `AddRowOffsetExec` (after exposing some methods publicly)
- `_row_created_at_version` / `_row_last_updated_at_version`: required piping through flags to `Fragment` readers.

Closes https://github.com/lance-format/lance/issues/5615.

### DIFF
--- a/python/python/tests/test_dataset.py
+++ b/python/python/tests/test_dataset.py
@@ -800,7 +800,7 @@ def test_take_version_system_columns(tmp_path: Path):
     """Test _row_created_at_version and _row_last_updated_at_version columns."""
     table = pa.table({"a": range(50)})
     base_dir = tmp_path / "test_take_version_columns"
-    lance.write_dataset(table, base_dir)
+    lance.write_dataset(table, base_dir, enable_stable_row_ids=True)
     dataset = lance.dataset(base_dir)
 
     # Initial version is 1


rust/lance-core/src/datatypes/schema.rs
```diff
@@ -255,12 +270,17 @@ impl Schema {
...
     }
 
     /// Project the columns over the schema, dropping unrecognized columns
     pub fn project_or_drop<T: AsRef<str>>(&self, columns: &[T]) -> Result<Self> {
-        self.do_project(columns, false)
+        self.do_project(columns, false, false)
+    }
+
+    /// Project the columns over the schema, preserving system columns.
+    pub fn project_preserve_system_columns<T: AsRef<str>>(&self, columns: &[T]) -> Result<Self> {
```
Xuanwo, line 299: The naming here implies that we support all system columns. Do we intend to support all of them? Or should we align with the list in `is_system_column`?

Xuanwo (commented):
Thank you @hamersaw for working on this! Only have a question.

rust/lance/src/dataset/take.rs
```diff
@@ -329,8 +344,19 @@ async fn do_take_rows(
...
-    let batch = projection.project_batch(batch).await?;
-    if builder.with_row_address {
+    // strip `ROW_OFFSET` field from projection (if exists)
+    let mut stripped_projection = projection.as_ref().clone();
+    stripped_projection.requested_output_expr = stripped_projection
+        .requested_output_expr
+        .clone()
+        .into_iter()
+        .filter(|e| e.name != ROW_OFFSET)
+        .collect();
```
hamersaw: I'm pretty unsure about this (feels hacky). I first attempted to inject the `ROW_OFFSET` field into the schema within [project_batch](https://github.com/hamersaw/lance/blob/bug%2Fsupport-system-columns-in-take-scan/rust/lance-datafusion/src/projection.rs#L440-L440) and inject the [AddRowOffsetExec](https://github.com/hamersaw/lance/blob/bug%2Fsupport-system-columns-in-take-scan/rust/lance/src/io/exec/rowids.rs#L346-L351) into the `ExecutionPlan` but was getting "index out of bounds" with len 2. I think this may be because we expect the `ROW_ADDR` column (required for `ROW_OFFSET` compute) but we manually inject it for `take*` operations.

My other idea is to strip this field from the `requested_output_expr` in the [`TakeBuilder` when we convert into the `ProjectionPlan`](https://github.com/hamersaw/lance/blob/bug%2Fsupport-system-columns-in-take-scan/rust/lance/src/dataset.rs#L361-L361). I'm unsure of the implications there.

rust/lance-core/src/datatypes/schema.rs
hamersaw: Thanks, this previous iteration only supported "_rowid" and "_rowaddr". Updated this to support all system columns.

rust/lance/src/dataset/take.rs
```diff
@@ -342,12 +368,26 @@ async fn do_take_rows(
...
         }
 
-        let row_addr_col = Arc::new(UInt64Array::from(row_addrs));
-        let row_addr_field = ArrowField::new(ROW_ADDR, arrow::datatypes::DataType::UInt64, false);
-        Ok(batch.try_with_column(row_addr_field, row_addr_col)?)
-    } else {
-        Ok(batch)
+        let row_addr_col: ArrayRef = Arc::new(UInt64Array::from(row_addrs));
+
+        if projection.must_add_row_offset {
```
jackye1995, line 384: I think there is an ordering problem here, if usr requests for example `dataset.take(columns=["a, "_rowoffset", "b"]`, then it will append the offset column in the end making it `dataset.take(columns=["a,  "b", "_rowoffset"]`.

I think instead of stripping it from `requested_output_expr`, we can keep it, inject the system column into the batch before calling `project_batch`, so that projection will handle the order of the columns.

hamersaw: Thanks! This worked perfectly - just needed to inject `ROW_OFFSET` into the `physical_projection` schema the same way this is done in the scanner. Did a ton of tests all permutations and orderings of system / non-system columns and I feel quite confident.

jackye1995:
looks like there are quite a few CI failures, could you fix those?

rust/lance/src/dataset/take.rs
westonpace: Yeah, the `project_batch` method at the bottom will correct any misordering

westonpace:
I've added https://github.com/lance-format/lance/issues/5823 which describes some of the stuff we talked about externally.

westonpace:
Looks like some legitimate test failures in the new python tests

jackye1995 (approved):
Looks good to me!

PR #5728, opened 2026-01-16
refactor!: introduce storage options accessor

jackye1995:
Same as https://github.com/lance-format/lance/pull/5547, but that one is too hard to rebase at this moment.

For easier integration with distributed engines, we want a way to get `dataset.latest_storage_options()` through the initial static storage options + the dynamic storage options provider.

However, current implementation pushes this logic down to AWS. This PR lifts the logic up into a component `StorageOptionsAccessor`, so that we can invoke it at dataset level, and also just wrap it as a CredentialsProvider for AWS. We also make `ObjectStoreParams` now hold a `StorageOptionsAccessor` instead of just the `storage_options` map.

For user API, we continue to let them input storage options and storage options provider, `StorageOptionsAccessor` is used only internally.

We also remove a few features added that turned out to be unnecessary in the whole credentials vending code path:
1. ignore_namespace_storage_options: this just never turns out to be possible to be set to true, I was overthinking in the beginning, so remove related code.
2. s3_credentials_refresh_offset_seconds: we overloaded this feature to set credentials refresh lead time for the storage options provider-based AWS credentials provider. But this is now not fitting the framework since the feature will be applicable to gcp and azure as well. Note that I removed all the related code in python and java. Although s3_credentials_refresh_offset exists for a long time, there was not really a way to set it in python and java. I added it for storage options provider only, but now it is no longer needed.

Note that recently I added credentials vending server side support in DirectoryNamespace for all 3 clouds. So we can now provide a generic solution to vend credentials and test everything end to end for GCP and Azure. However this PR still keeps the feature in AWS, we will do another PR to add GCP and Azure support officially

### DIFF
--- a/rust/lance-io/src/object_store/storage_options.rs
+++ b/rust/lance-io/src/object_store/storage_options.rs
@@ -218,7 +218,7 @@ impl StorageOptionsAccessor {
     ///
     /// The returned accessor will always return the provided options.
     /// This is useful when credentials don't expire or are managed externally.
-    pub fn static_options(options: HashMap<String, String>) -> Self {
+    pub fn with_static_options(options: HashMap<String, String>) -> Self {
         let expires_at_millis = options
             .get(EXPIRES_AT_MILLIS_KEY)
             .and_then(|s| s.parse::<u64>().ok());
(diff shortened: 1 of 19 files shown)


rust/lance-io/src/object_store/storage_options.rs
```diff
@@ -139,3 +156,558 @@ impl StorageOptionsProvider for LanceNamespaceStorageOptionsProvider {
...
+            .and_then(|s| s.parse::<u64>().ok())
+            .map(Duration::from_millis)
+            .unwrap_or(Duration::from_millis(DEFAULT_REFRESH_OFFSET_MILLIS))
+    }
+
+    /// Create an accessor with only static options (no refresh capability)
+    ///
+    /// The returned accessor will always return the provided options.
+    /// This is useful when credentials don't expire or are managed externally.
+    pub fn static_options(options: HashMap<String, String>) -> Self {
```
hamersaw: does `with_static_options` give more uniformity re: naming?

PR #5736, opened 2026-01-16
perf: use LRU cache for session contexts in get_session_context

wjones127:
## Motivation

We have a performance issue when `LANCE_MEM_POOL_SIZE`: when this is set (to something different than the default), the current cache always misses here:

https://github.com/lance-format/lance/blob/445dd5bfe7b0111fd8e146163393e825e5995679/rust/lance-datafusion/src/exec.rs#L398-L410

## Summary
- Replace static `LazyLock` session contexts with LRU cache (size 4)
- Cache key uses resolved configuration values (after env var lookup)
- Fixes cache misses when `LANCE_MEM_POOL_SIZE` env var is set

## Test plan
- [x] `cargo test -p lance-datafusion` passes
- [x] `cargo clippy -p lance-datafusion` passes

### DIFF
--- a/rust/lance-datafusion/src/exec.rs
+++ b/rust/lance-datafusion/src/exec.rs
@@ -946,8 +946,12 @@ impl ExecutionPlan for StrictBatchSizeExec {
 mod tests {
     use super::*;
 
+    // Serialize cache tests since they share global state
+    static CACHE_TEST_LOCK: std::sync::Mutex<()> = std::sync::Mutex::new(());
+
     #[test]
     fn test_session_context_cache() {
+        let _lock = CACHE_TEST_LOCK.lock().unwrap();
         let cache = get_session_cache();
 
         // Clear any existing entries from other tests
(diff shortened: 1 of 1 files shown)


rust/lance-datafusion/src/exec.rs
```diff
@@ -385,28 +385,70 @@ pub fn new_session_context(options: &LanceExecutionOptions) -> SessionContext {
...
+
+    // If key exists, update access time and return
+    if let Some(entry) = cache.get_mut(&key) {
+        entry.last_access = std::time::Instant::now();
+        return entry.context.clone();
     }
-    new_session_context(options)
+
+    // Evict least recently used entry if cache is full
+    const MAX_CACHE_SIZE: usize = 4;
```
cmccabe: can we have an environment variable for this or something?

wjones127: I guess so

PR #5737, opened 2026-01-17
fix: remove credential vending features from python and java bindings

jackye1995:
Fixing hanging Spark test: https://github.com/lance-format/lance-spark/pull/177

Without this fix, the additioanl azure and gcp dependencies in credentials vending featrues pulled in native-tls, which requires installing openssl 1.1. When not present, Spark hangs because rust execution fails in a separated thread and it fails silently, and then a semaphore keeps waiting for result.

For why it fails silently rather than propagating back, that's another fix needed in Spark, but should not block the release of 2.0.0

### DIFF
--- a/java/lance-jni/Cargo.lock
+++ b/java/lance-jni/Cargo.lock
@@ -3,12 +3,6 @@
 version = 3
 
 [[package]]
-name = "RustyXML"
-version = "0.3.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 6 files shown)


wjones127 (approved):

bryanck:
This also fixes an issue we are running into involving `native-tls`, where CA certs aren't loaded.

jtuglu1 (approved):
This fixes rest namespace for us

PR #5739, opened 2026-01-17
fix: skip missing indices in compaction rewrite

### LINKED ISSUE #5730: Error: Invalid compaction plan refers to index *** which does not exist
I think we can instead ignore indexes that no longer exist, so we shouldn't need this error.

https://github.com/lance-format/lance/blob/b1cd4237d50109c8b51bddbc2c065d6652d542ce/rust/lance/src/dataset/transaction.rs#L2203-L2211

AndreaBozzo:
Fixes #5730


AndreaBozzo:
> This for getting this done!


No problem! Have a nice day

PR #5740, opened 2026-01-17
feat: support tencent cos

ztorchan:
close https://github.com/lance-format/lance/issues/5726

### DIFF
--- a/rust/lance/Cargo.toml
+++ b/rust/lance/Cargo.toml
@@ -116,7 +116,7 @@ geo-types = { workspace = true }
 
 
 [features]
-default = ["aws", "azure", "gcp", "oss", "huggingface"]
+default = ["aws", "azure", "gcp", "oss", "huggingface", "tencent"]
 fp16kernels = ["lance-linalg/fp16kernels"]
 # Prevent dynamic linking of lzma, which comes from datafusion
 cli = ["dep:clap", "lzma-sys/static"]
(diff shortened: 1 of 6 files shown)


ztorchan:
Current support for COS still has room for improvement.

The options for `CosConfig` in opendal are very limited, so we need to set the `disable_config_load` option to false to load more configurations from environment variables, such as `TENCENTCLOUD_SECURITY_TOKEN` and `TENCENTCLOUD_REGION`.

Next, we need to improve opendal's support for COS configuration and improve lance accordingly.

ztorchan:
Hi, @Xuanwo @wjones127 @yanghua , please review this PR when you have time, thanks a lot

rust/lance/Cargo.toml
```diff
@@ -116,7 +116,7 @@ geo-types = { workspace = true }
 
 
 [features]
-default = ["aws", "azure", "gcp", "oss", "huggingface"]
+default = ["aws", "azure", "cos", "gcp", "oss", "huggingface"]
```
wjones127: It might be better to call it `tencent` instead of `cos`, for two reasons:

1. It's clearer that it's a Cloud vendor. Most people don't know what `cos` is but have heard of `tencent`.
2. It's more in line with `aws`, `azure` and `gcp`. (`oss` is a bad example here, and a confusing name for a feature, IMO)

```suggestion
default = ["aws", "azure", "tencent", "gcp", "oss", "huggingface"]
```

ztorchan: Makes sense. I have replaced all "cos" with "tencent", except for schema and corresponding storage_options key.

Issue #5746, opened 2026-01-18
iOS build fails: fp16kernels feature check in build.rs doesn't work for unsupported targets


durch:
## Description

Building for iOS target (`aarch64-apple-ios` or `aarch64-apple-ios-sim`) fails with:

```
error: Unable to build f16 kernels on given target_arch. Please use x86_64 or aarch64 or remove the fp16kernels feature
```

This happens even when `fp16kernels` feature is **not enabled**.

## Root Cause

In `rust/lance-linalg/build.rs`, there are two bugs:

### Bug 1: Feature check uses wrong method

```rust
fn build_f16_with_flags(suffix: &str, flags: &[&str]) -> Result<(), cc::Error> {
    if cfg!(not(feature = "fp16kernels")) {  // ❌ This checks build.rs features, not the crate's
        return Ok(());
    }
    // ...
}
```

`cfg!(feature = "fp16kernels")` is a compile-time check on the **build script itself**, not the library being built. The correct check is:

```rust
if env::var("CARGO_FEATURE_FP16KERNELS").is_err() {
    return Ok(());
}
```

### Bug 2: Error returned before feature check for unsupported targets

```rust
fn main() -> Result<(), String> {
    // ... arch/os checks ...
    
    if target_arch == "aarch64" && target_os == "macos" {
        build_f16_with_flags(...)?;  // Feature check happens inside
    } else if target_arch == "aarch64" && target_os == "linux" {
        build_f16_with_flags(...)?;
    } else if target_arch == "x86_64" {
        // ...
    } else {
        return Err("Unable to build f16 kernels...");  // ❌ Error BEFORE any feature check!
    }
}
```

For iOS (`target_os = "ios"`), the code falls through to the `else` branch and returns an error **before** `build_f16_with_flags` is ever called.

## Proposed Fix

Add the feature check at the **top of main()**, before any architecture branching:

```rust
fn main() -> Result<(), String> {
    // ... nightly check, rerun-if-changed ...

    // Check if fp16kernels feature is enabled - if not, skip everything
    if env::var("CARGO_FEATURE_FP16KERNELS").is_err() {
        return Ok(());
    }

    // ... rest of the code ...
}
```

This way, when `fp16kernels` is disabled (which is the default), the build script exits cleanly on all platforms including iOS, Android, and any future targets.

## Environment

- lancedb v0.23.1
- lance-linalg v1.0.1
- Target: `aarch64-apple-ios-sim`
[...]

PR #5747, opened 2026-01-18
fix(lance-linalg): check fp16kernels feature before arch-specific code

durch:
## Summary

The build script was checking target architecture BEFORE verifying if the `fp16kernels` feature was enabled. This caused builds to fail on platforms like iOS/Android when falling through to the error branch, even when the feature was disabled.

## The Problem

On line 78-79 of `rust/lance-linalg/build.rs`:
```rust
} else {
    return Err("Unable to build f16 kernels on given target_arch.  Please use x86_64 or aarch64 or remove the fp16kernels feature".to_string());
}
```

This error is returned for any target_arch not in the supported list (x86_64, aarch64 macos/linux, loongarch64), **before** the feature check in `build_f16_with_flags()` is ever called.

Additionally, `cfg!(not(feature = "fp16kernels"))` inside `build_f16_with_flags()` checks the **build script's** features, not the library's features (build scripts are compiled for the host, not the target).

## The Fix

Add an early exit using `CARGO_FEATURE_FP16KERNELS` env var check (which correctly reflects the library's features, not the build script's) **before** the target_arch branching logic.

## Testing

Verified fix enables successful iOS builds with:
- lancedb v0.23.1
- Tauri 2.x iOS target
- `fp16kernels` feature disabled (default)

Fixes: #5746

### DIFF
--- a/.bumpversion.toml
+++ b/.bumpversion.toml
@@ -1,5 +1,5 @@
 [tool.bumpversion]
-current_version = "2.0.0-beta.9"
+current_version = "2.0.0-beta.10"
 parse = "(?P<major>\\d+)\\.(?P<minor>\\d+)\\.(?P<patch>\\d+)(-(?P<prerelease>(beta|rc))\\.(?P<prerelease_num>\\d+))?"
 serialize = [
     "{major}.{minor}.{patch}-{prerelease}.{prerelease_num}",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -658,9 +658,9 @@ dependencies = [
 
 [[package]]
 name = "aws-runtime"
-version = "1.5.17"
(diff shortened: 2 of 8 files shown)


wjones127 (changes requested):
@durch Thanks for creating a PR.

I think we still want to execute `build_dist_table_with_flags` if fp16kernels is not enabled, but this short-circuits before that. Could you address that?

durch:
> @durch Thanks for creating a PR.
> 
> I think we still want to execute `build_dist_table_with_flags` if fp16kernels is not enabled, but this short-circuits before that. Could you address that?

@wjones127, thanks for the speedy review, this should do it

wjones127 (approved):
This looks good. Thanks!

PR #5748, opened 2026-01-19
feat: add Lance-HF docs to lance.org/integrations/huggingface/

prrao87:
Add Lance-Hugging Face docs to the docs deploy workflow by checking out the external repo and copying its [`index.md`](https://github.com/lance-format/lance-huggingface/blob/main/docs/src/index.md) into `huggingface.md` during CI, keeping Hugging Face docs sourced from the integration repo without changing local docs.


prrao87:
It's okay to overwrite the existing `docs/integrations/huggingface.md` file because it's a stub. The [live docs](https://github.com/lance-format/lance-huggingface/blob/main/docs/src/index.md) page in the external repo for the HF integration is where the docs will ultimately live.

.github/workflows/docs-deploy.yml
```diff
@@ -84,6 +89,9 @@ jobs:
           cat >> docs/src/integrations/.pages << 'EOF'
             - Ray: ray
           EOF
+      - name: Copy lance-huggingface docs
+        run: |
+          cp lance-huggingface/docs/src/index.md docs/src/integrations/huggingface.md
```
jackye1995: instead of just copying the index.md, we can directly remove `huggingface.md`, and copy the entire `lance-huggingface/docs/src` directory as `docs/src/integrations/huggingface` directory. The `.pages` inside `lance-huggingface/docs/src` will automatically make the index.md file show as the main page.

prrao87: Got it, that makes sense! I've update it to exactly follow the `lance-ray` template.

prrao87: Just one hitch with this: it's not a single page, but a new section in the sidebar. Is that okay?

jackye1995:
and we can remove https://github.com/lance-format/lance/blob/main/docs/src/integrations/huggingface.md now

prrao87:
Done!

PR #5750, opened 2026-01-19
feat(python): support namespace for tensorflow

yuqi1129:


python/python/lance/tf/data.py
```diff
@@ -132,20 +134,24 @@ def column_to_tensor(array: pa.Array, tensor_spec: tf.TensorSpec) -> tf.Tensor:
 
 
 def from_lance(
-    dataset: Union[str, Path, LanceDataset],
+    dataset: Optional[Union[str, Path, LanceDataset]] = None,
```
wjones127, line 137: I find the new API a little clunky, but I recognize you are just matching what is is in `lance.dataset`. I created an issue for improving this API: https://github.com/lance-format/lance/issues/5809

wjones127:
Test failures are unrelated.

PR #5752, opened 2026-01-19
refactor: align blob behavior that write via file format version, read via layout

Xuanwo:
This PR fixes a bug where users could encode blob v1 even when blob v2 was enabled. However, our decoder only reads the dataset's configuration `lance.blob.version`, which can lead to decoding issues.

In this PR, we changed the following:

- During write operations, we determine which blob layout to use based on file format version.
- During read operations, we decide which blob layout to decode based on its physical layout.

---

### DIFF
--- a/.github/workflows/cargo-publish.yml
+++ b/.github/workflows/cargo-publish.yml
@@ -11,6 +11,11 @@ on:
         description: "Tag to publish (e.g., v1.0.0)"
         required: true
         type: string
+      skip_check_repo:
+        description: "Skip checking if packages have been modified (useful for backfilling missed releases)"
+        required: false
+        type: boolean
+        default: false
 
 env:
   # This env var is used by Swatinem/rust-cache@v2 for the cache
@@ -87,6 +92,7 @@ jobs:
           registry-token: ${{ secrets.CARGO_REGISTRY_TOKEN }}
           args: "--all-features"
           path: .
+          check-repo: ${{ github.event_name != 'workflow_dispatch' || inputs.skip_check_repo != true }}
   report-failure:
     name: Report Workflow Failure
     runs-on: ubuntu-latest
--- a/.github/workflows/java-publish.yml
+++ b/.github/workflows/java-publish.yml
@@ -33,7 +33,7 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v4
         with:
-          ref: ${{ inputs.ref }}
+          ref: ${{ inputs.ref || github.ref }}
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v3
       - name: Check glibc version outside docker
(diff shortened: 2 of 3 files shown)


python/python/lance/dataset.py
```diff
@@ -5540,6 +5540,7 @@ def write_dataset(
     data_storage_version: Optional[
         Literal["stable", "2.0", "2.1", "2.2", "next", "legacy", "0.1"]
     ] = None,
+    blob_version: Optional[Literal["v1", "v2"]] = None,
```
westonpace: We should document this parameter.  Make sure to clearly specify when a user would choose `v1`.  Is it only if they need backwards compatibility with a legacy software?  Are there any good reasons a user would choose v1?

rust/lance-arrow/src/lib.rs
```diff
@@ -842,6 +842,13 @@ fn project_array(array: &ArrayRef, target_field: &Field) -> Result<ArrayRef> {
 }
 
 fn project(struct_array: &StructArray, fields: &Fields) -> Result<StructArray> {
+    if struct_array.fields().len() != struct_array.columns().len() {
```
westonpace: How do we get an Arrow array that is invalid like this?

rust/lance-encoding/src/encodings/logical/blob.rs
```diff
@@ -402,6 +422,218 @@ impl FieldEncoder for BlobV2StructuralEncoder {
     }
 }
 
+fn normalize_blob_v2_input(array: ArrayRef) -> Result<StructArray> {
```
westonpace: Can you drop a short comment explaining what this function is doing?

Xuanwo:
> I think my initial question is why a user will specify v1 or v2?

The original idea is to allow users to continue using blob v1 even after a new file version has been released. What do you think? Perhaps it would be better to require users to enable blob v2 on new datasets created by newer versions?

rust/lance-arrow/src/lib.rs
Xuanwo: removed.

rust/lance-core/src/datatypes/field.rs
```diff
@@ -1806,14 +1761,34 @@ mod tests {
...
+                    ArrowField::new("uri", DataType::Utf8, true),
+                ]
+                .into(),
+            ),
+            true,
+        )
+        .with_metadata(metadata)
+        .try_into()
+        .unwrap();
+        field.unloaded_mut();
```
westonpace, line 1790: Let's fix in a follow-up but this function should probably be called `unload` instead of `unloaded_mut`?  If I see `unloaded_mut` I think of an accessor function returning `&mut Unloaded` or something like that.

Xuanwo: Yep!

Xuanwo:
"index::vector::ivf::v2::tests::test_build_ivf_pq_4bit::case_3" is known to be a flaky test, just skip.

PR #5754, opened 2026-01-20
perf: merge partitions in stream style

BubbleCal:
this avoids reading all token sets and doc sets into memory, reduces the peak memory footprint by 25%

related: https://github.com/lance-format/lance/pull/5754


Xuanwo (approved):
Thank you!

PR #5760, opened 2026-01-20
fix: fix deletion when using file-object-store://

cmccabe:
When using `file-object-store://`, we currently do not delete directories that are empty. This is because the other object stores (such as s3) don't need this. This PR fixes this to explicitly delete the empty directories when using `file-object-store://`.

This fixes a bug where tables appear to be undeletable when using `file-object-store://`

```
>>> db = lancedb.connect("file-object-store:///tmp/foo")
>>> table = db.create_table("my_table", schema=schema)
>>> db.table_names()
['my_table']
>>> db.drop_table("my_table")
>>> db.table_names()
['my_table']
```

### DIFF
--- a/rust/lance-io/src/object_store.rs
+++ b/rust/lance-io/src/object_store.rs
@@ -1134,7 +1134,7 @@ mod tests {
             url.set_path(path.to_str().unwrap());
             url
         };
-        let (store, base) = ObjectStore::from_uri(&url.to_string()).await.unwrap();
+        let (store, base) = ObjectStore::from_uri(url.as_ref()).await.unwrap();
         store.remove_dir_all(base.child("foo")).await.unwrap();
 
         assert!(!path.join("foo").exists());


wjones127 (approved):

rust/lance-io/src/object_store.rs
```diff
@@ -1099,11 +1099,19 @@ mod tests {
...
+    #[tokio::test]
+    async fn test_delete_directory_file_object_store() {
+        test_delete_directory("file-object-store://").await;
+    }
+
+    async fn test_delete_directory(prefix: &str) {
         let path = TempStdDir::default();
         create_dir_all(path.join("foo").join("bar")).unwrap();
         create_dir_all(path.join("foo").join("zoo")).unwrap();
-        create_dir_all(path.join("foo").join("zoo").join("abc")).unwrap();
```
westonpace, line 1106: Why remove this case?

cmccabe: oh, this case should stay in. good catch.

cmccabe:
> I guess I'm not sure what I'd expect as a user. Should it behave exactly as LocalObjectStore (in which case this PR is diverging from expectations) or just behave how we want it to?

The big issue with the current behavior is that it breaks lancedb when using `file-object-store://` 
```
>>> db = lancedb.connect("file-object-store:///tmp/foo")
>>> table = db.create_table("my_table", schema=schema)
>>> db.table_names()
['my_table']
>>> db.drop_table("my_table")
>>> db.table_names()
['my_table']
```

If you want to use `file-object-store://` in unit testing, this is unfortunate.

PR #5764, opened 2026-01-21
feat: add ConvertScanFilterToFilteredRead PhysicalOptimizerRule

### LINKED ISSUE #5154: Add optimizer rule to change `LanceScanExec -> FilterExec` into `FilteredReadExec`
There are query plans like this one that could benefit from pushing down the filter:

https://github.com/lancedb/lance/blob/35a0547426c47dd94425619c442dcd03ffd9860c/rust/lance/src/dataset/scanner.rs#L7157-L7158

Note: this will become obsolete if we change `Scanner` to construct a logical plan using the `LanceTableProvider` since pushdown can be handled at that layer. However, that's a ways off.

https://github.com/lancedb/lance/blob/35a0547426c47dd94425619c442dcd03ffd9860c/rust/lance/src/datafusion/dataframe.rs#L147-L158

lichuang:
close #5154.

add `ConvertScanFilterToFilteredRead` to convert `LanceScanExec` -> `FilterExec` into `FilteredReadExec`

### DIFF
--- a/rust/lance/src/io/exec/optimizer.rs
+++ b/rust/lance/src/io/exec/optimizer.rs
@@ -5,18 +5,24 @@
 
 use std::sync::Arc;
 
-use super::TakeExec;
+use super::{
+    filter::LanceFilterExec,
+    filtered_read::{FilteredReadExec, FilteredReadOptions},
+    LanceScanExec, TakeExec,
+};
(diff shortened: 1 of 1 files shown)


lichuang:
@wjones127

PR #5765, opened 2026-01-21
feat: add range to External blob

wojiaodoubao:
A common practice for organizing multimodal data is to package a large number of scattered files into several large archives (such as tar or zip files), which can prevent excessive inode consumption, achieve higher read speeds via streaming, and avoid inconsistencies between metadata and the scattered stored files.

To better support such archives, I suggest we extend the ExternalPacked format for Lance BlobV2. For example linking images from the archives to a Lance table.

### DIFF
--- a/python/python/lance/blob.py
+++ b/python/python/lance/blob.py
@@ -51,9 +51,7 @@ class Blob:
         if uri == "":
             raise ValueError("Blob uri cannot be empty")
         if position < 0 or size < 0:
-            raise ValueError(
-                "External packed blob position and size must be non-negative"
-            )
+            raise ValueError("External blob position and size must be non-negative")
         return Blob(uri=uri, position=position, size=size)
 
     @staticmethod
--- a/rust/lance-encoding/src/encodings/logical/blob.rs
+++ b/rust/lance-encoding/src/encodings/logical/blob.rs
@@ -676,8 +676,10 @@ mod tests {
 
     #[tokio::test]
     async fn test_blob_v2_external_with_range_round_trip() {
-        let blob_metadata =
-            HashMap::from([(lance_arrow::BLOB_META_KEY.to_string(), "true".to_string())]);
+        let blob_metadata = HashMap::from([(
+            lance_arrow::ARROW_EXT_NAME_KEY.to_string(),
+            lance_arrow::BLOB_V2_EXT_NAME.to_string(),
+        )]);
 
         let kind_field = Arc::new(ArrowField::new("kind", DataType::UInt8, true));
         let data_field = Arc::new(ArrowField::new("data", DataType::LargeBinary, true));


Xuanwo:
This idea is kinda interesting. So your idea focuses more on reading an external packed blob, and Lance won't generate such a blob itself, is that correct?

wojiaodoubao:
> So your idea focuses more on reading an external packed blob, and Lance won't generate such a blob itself, is that correct?

Yes.

Xuanwo:
> Yes.

Then I think we don't need to add a new type. We can simply enable reading a range from an external blob.

wojiaodoubao:
> We can simply enable reading a range from an external blob.

Do you mean we could enable External BlobFile to handle any arbitrary range(currently the position is always 0), thus eliminating the need to introduce ExternalPacked as an extra component?

It makes sense to me, let me update the pr.

Xuanwo:
> Do you mean we could enable External BlobFile to handle any arbitrary range(currently the position is always 0), thus eliminating the need to introduce ExternalPacked as an extra component?

Yes!

wojiaodoubao:
Hi @Xuanwo , this pr is ready for review now, please help when you have time, thanks very much.

rust/lance-core/src/datatypes.rs
```diff
@@ -437,7 +437,10 @@ pub enum BlobKind {
     Packed = 1,
     /// Stored in a dedicated raw blob file; `blob_id` identifies the file, `size` is the full file length.
     Dedicated = 2,
-    /// Not stored by Lance; `blob_uri` holds an absolute external URI, offsets are zero.
+    /// Not stored by Lance; `blob_uri` holds an absolute external URI.
+    ///
+    /// If both `position` and `size` are non-zero, they describe a byte range within the
```
Xuanwo: I don't think we need to strictly distinguish between external blobs based on whether they have an offset or not.

We can simply state the following:

External blobs can have a `position` and a `size`. Users can specify a range for an external blob. If the position is not set, it defaults to `0`, which points to the beginning of the blob.

rust/lance-encoding/src/encodings/logical/blob.rs
```diff
@@ -640,6 +647,66 @@ mod tests {
...
+    #[tokio::test]
+    async fn test_blob_v2_external_with_range_round_trip() {
+        let blob_metadata =
+            HashMap::from([(lance_arrow::BLOB_META_KEY.to_string(), "true".to_string())]);
+
+        let kind_field = Arc::new(ArrowField::new("kind", DataType::UInt8, true));
+        let data_field = Arc::new(ArrowField::new("data", DataType::LargeBinary, true));
+        let uri_field = Arc::new(ArrowField::new("uri", DataType::Utf8, true));
+        let blob_id_field = Arc::new(ArrowField::new("blob_id", DataType::UInt32, true));
+        let blob_size_field = Arc::new(ArrowField::new("blob_size", DataType::UInt64, true));
```
Xuanwo, line 688: This field should be `size`?

rust/lance/src/dataset/blob.rs
```diff
@@ -213,6 +213,12 @@ impl BlobPreprocessor {
                     Error::invalid_input("Blob struct missing `uri` field", location!())
                 })?
                 .as_string::<i32>();
+            let position_col = struct_arr
+                .column_by_name("position")
+                .map(|col| col.as_primitive::<UInt64Type>());
+            let blob_size_col = struct_arr
```
Xuanwo: The same.

python/python/lance/blob.py
```diff
@@ -55,6 +74,8 @@ def __init__(self) -> None:
             [
                 pa.field("data", pa.large_binary(), nullable=True),
                 pa.field("uri", pa.utf8(), nullable=True),
+                pa.field("position", pa.uint64(), nullable=True),
+                pa.field("blob_size", pa.uint64(), nullable=True),
```
Xuanwo: Please use `size`

rust/lance-encoding/src/encodings/logical/blob.rs
wojiaodoubao: This is blob_size like test_blob_v2_dedicated_round_trip. It tests BlobV2StructuralEncoder encoding External with position and blob_size.

Xuanwo:
Test `encodings::logical::blob::tests::test_blob_v2_external_with_range_round_trip` failed.

PR #5766, opened 2026-01-21
feat: add url support for blob

wojiaodoubao:
The unit test [test_blob_https_url_and_range_s3_roundtrip](https://github.com/lance-format/lance/pull/5766/files#diff-d1766821b75c8f92876f01b6be5b8be9cfe29a9caeb557d862c342ad77e49618R291) shows the case, it could be ran as below.
```
export LANCE_S3_ENDPOINT=https://{bucket}.{endpoint}
export LANCE_S3_BUCKET={bucket}
export AWS_REGION={region}
export AWS_ACCESS_KEY_ID={AK}
export AWS_SECRET_ACCESS_KEY={SK}

pytest python/tests/test_blob.py::test_blob_https_url_and_range_s3_roundtrip
```


majin1102:
> Therefore, I propose we add URL-related functions to the Lance Blob API

Wondering if we surely introduce this URL-based approach to take a row or rows, should we limit this to fields that use Blob API?

wojiaodoubao:
> Wondering if we surely introduce this URL-based approach to take a row or rows, should we limit this to fields that use Blob API?

I think we should only allow blob api to get URLs, which is the BlobFile.

Xuanwo:
opendal has a native http service support, maybe we can just use that

wojiaodoubao:
I start an issue to discuss the API design https://github.com/lance-format/lance/issues/5800. After the decision is made, we can go back here.

PR #5773, opened 2026-01-21
perf: add a lightweight scheduler implementation

westonpace:
The current scheduler introduces too much synchronization overhead when we are in a high-iops throughput situation.  This scheduler reduces the number of asynchronous context switches.  On my desktop it doesn't actually have much impact on performance.  However, on a system with more cores and higher RAM bandwidth the new scheduler more than doubles the amount of IOPS/s.  Combined with the I/O uring (coming in a future PR) performance is actually 4x.

This scheduler makes tradeoffs which are not ideal for cloud readers, but which are important for the uring reader:

- There is no dedicated background I/O loop thread and tasks are not launched with tokio spawn.  This is fine for the local filesystem since each task is its own spawn blocking call.  However, for cloud stores this is not great.  Tasks are not spawned calls and need to be polled occasionally.  The end result is that I/O tasks which are not front of the line might get paused if they fill up (for example, if the HTTP request queues fill up).  In I/O uring this isn't a problem because polling one task actually progresses all I/O requests (not just the one associated with the task).
- The lite scheduler does not enforce any kind of concurrency throttling.  This is because it is mainly meant for local storage use cases and there is not much penalty for having too many concurrent requests.  This would be a problem if using this scheduler for cloud storage however.

### DIFF
--- a/rust/lance-io/src/scheduler.rs
+++ b/rust/lance-io/src/scheduler.rs
@@ -1420,6 +1420,77 @@ mod tests {
         assert_eq!(second_fut.await.unwrap().len(), 10);
     }
 
+    /// A Reader that tracks how many times get_range has been called.
+    #[derive(Debug)]
+    struct TrackingReader {
+        get_range_count: Arc<AtomicU64>,
+        path: Path,
+    }
+
+    impl deepsize::DeepSizeOf for TrackingReader {
+        fn deep_size_of_children(&self, _context: &mut deepsize::Context) -> usize {
+            0
+        }
+    }
+
+    impl Reader for TrackingReader {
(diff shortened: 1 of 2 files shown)


westonpace:
Drafting until I merge https://github.com/lance-format/lance/pull/5755

westonpace:
Once the io_uring stuff merges I'll probably move away from triggering the lite scheduler with an environment variable and instead let the reader instance decide which scheduler it wants.

westonpace:
I added a test for ordering and one more for making sure get_range is called at the right time (a nuanced thing it took a long time to debug).

The on/off tests can come when the uring reader is added

Issue #5751, opened 2026-01-19
How can I implement a single commit handler when writing to a dataset using both Rust and Python simultaneously?


hushengquan:
My application scenario involves writing datasets using Python and Rust. I implemented an External Manifest Store in Rust, which provides atomic transaction commits. However, this solution cannot be used with Python, and the commit handler has requirements for the writers, specifically that only one commit handler can be active at a time.Are there any solutions to this problem?

westonpace:
The commit handler is essentially the "compare and swap" primitive at the heart of Lance's concurrency model.  If you want two different commit handler approaches to run at the same time they would have to have some way of interacting to maintain atomicity.

The default commit handler relies on object storage to provide the atomicity (either through a POSIX compliant rename-or-fail for disks or through write-if-not-exist for cloud storage)  The advantage is that it doesn't matter what language or library the client is using.  Why does the default commit handler not work?

Is it possible for your custom rust commit handler to have python bindings?  Even so, if your commit handler is in-process then either all your writers have to be in the same process or your commit handlers need some way of communicating like TCP, etc.

hushengquan:
> The commit handler is essentially the "compare and swap" primitive at the heart of Lance's concurrency model. If you want two different commit handler approaches to run at the same time they would have to have some way of interacting to maintain atomicity.
> 
> The default commit handler relies on object storage to provide the atomicity (either through a POSIX compliant rename-or-fail for disks or through write-if-not-exist for cloud storage) The advantage is that it doesn't matter what language or library the client is using. Why does the default commit handler not work?
> 
> Is it possible for your custom rust commit handler to have python bindings? Even so, if your commit handler is in-process then either all your writers have to be in the same process or your commit handlers need some way of communicating like TCP, etc.

Some object storage systems do not support the put-if-not-exist semantic, making it impossible to guarantee transactional atomicity.Is there a way to ensure atomicity guarantees for transaction commits across APIs written in different languages?

westonpace:
The dynamodb commit handler is an example of how you might do this.  An external service (dynamodb) is used to provide the atomicity.  You could use any other external database in the same manner or you could build your own external service.

hushengquan:
> The dynamodb commit handler is an example of how you might do this. An external service (dynamodb) is used to provide the atomicity. You could use any other external database in the same manner or you could build your own external service.

I have implemented the external service(mysql) in rust, but it is currently not usable in the Python API.

hushengquan:
[...]

PR #5774, opened 2026-01-21
feat: unifying the implementation of transactional commit atomicity across different languages

hushengquan:
Recently, Lance added support for distributed compaction and distributed index building.  Using these features with Ray requires calling them through the Python API. However, my existing application reads and writes datasets using the Rust API, and the underlying object storage does not support the atomic put-if-not-exist semantic. Therefore, I implemented an `ExternalManifestStore` interface in Rust.  However, when trying to implement distributed compaction based on Ray, I discovered that the transaction atomicity mechanisms in Python and Rust are inconsistent, and only one `CommitHandler` can operate on a single dataset at a time. Therefore, I implemented a mechanism that allows using `ExternalManifestCommitHandler` with the Python API as well, thus achieving atomic transaction commits across different language APIs.

issue:[#5751](https://github.com/lance-format/lance/issues/5751)

### DIFF
--- a/python/python/lance/indices/builder.py
+++ b/python/python/lance/indices/builder.py
@@ -65,6 +65,7 @@ class IndicesBuilder:
         accelerator: Optional[Union[str, "torch.Device"]] = None,
         sample_rate: int = 256,
         max_iters: int = 50,
+        target_partition_size: Optional[int] = None,
     ) -> IvfModel:
         """
         Train IVF centroids for the given vector column.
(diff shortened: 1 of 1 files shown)


wjones127:
Hi @hushengquan, it looks like these commits don't match the PR description. They seem to be closer to https://github.com/lance-format/lance/pull/6287

hushengquan:
> Hi @hushengquan, it looks like these commits don't match the PR description. They seem to be closer to #6287

@wjones127 Sorry! I forgot to close this PR. Could you please take a look at it? [pr link](https://github.com/lance-format/lance/pull/6091) . It is designed to be more general-purpose, and we've already started using it extensively internally.

PR #5776, opened 2026-01-21
feat: add compute_splits function to dataset

hamersaw:
This PR adds a `compute_splits` function on the dataset which uses indexes to compute fragment splits when filtering the dataset. A split will contain either a collection of fragments that are covered by the same index or a single fragment if not covered by any indexes. This is useful for distributed compute engines to determine how to effectively partition datasets.

### DIFF
--- a/java/lance-jni/src/blocking_dataset.rs
+++ b/java/lance-jni/src/blocking_dataset.rs
@@ -4,7 +4,7 @@
 use crate::error::{Error, Result};
 use crate::ffi::JNIEnvExt;
 use crate::storage_options::JavaStorageOptionsProvider;
-use crate::traits::{export_vec, import_vec, FromJObjectWithEnv, FromJString};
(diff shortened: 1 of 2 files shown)


rust/lance/src/dataset.rs
```diff
@@ -1891,6 +1891,70 @@ impl Dataset {
...
+    /// # use lance::{Dataset, Result};
+    /// # async fn example(dataset: &Dataset) -> Result<()> {
+    /// let splits = dataset.compute_splits(&["vector_column"], None).await?;
+    /// for split in splits {
+    ///     println!("Split has {} fragments", split.fragments().len());
+    /// }
+    /// # Ok(())
+    /// # }
+    /// ```
+    pub async fn compute_splits(
```
jackye1995, line 1926: I am thinking this should be a part of the scanner, rather than the dataset, because scanner already has the entry point for users to specify filter, project, etc., and we can just in the end request a list of splits rather than executing the scan.

PR #5777, opened 2026-01-21
feat: io_uring based file reader

westonpace:
~~This is still a draft while waiting on https://github.com/lance-format/lance/pull/5755 and https://github.com/lance-format/lance/pull/5773~~

This PR adds a new URI scheme `file+uring`.  The scheme uses the same local file reader as `file` but has two custom `Reader` implementations that are based on the io_uring API.  One of these creates a configurable number of process-wide ring threads and the reader communicates with this thread using a queue.  The second assumes that the scheduler and decoder run on the same thread and uses a thread local uring instance.

Both are able to saturate up to 1.5M IOPS/s when combined with the scheduler rework.  I've tested the thread local variant up to 2M IOPS/s.  These numbers are assuming the data is not in the kernel page cache.  I've seen results as high as 4M IOPS/s when the data is cached.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3360,7 +3360,7 @@ checksum = "42703706b716c37f96a77aea830392ad231f44c9e9a67872fa5548707e11b11c"
 
 [[package]]
 name = "fsst"
-version = "4.1.0-beta.3"
+version = "5.0.0-beta.2"
 dependencies = [
  "arrow-array",
  "lance-datagen",
@@ -4590,7 +4590,7 @@ dependencies = [
 
 [[package]]
 name = "lance"
-version = "4.1.0-beta.3"
+version = "5.0.0-beta.2"
 dependencies = [
  "all_asserts",
  "approx",
@@ -4686,7 +4686,7 @@ dependencies = [
 
 [[package]]
 name = "lance-arrow"
-version = "4.1.0-beta.3"
+version = "5.0.0-beta.2"
 dependencies = [
  "arrow-array",
  "arrow-buffer",
@@ -4722,7 +4722,7 @@ dependencies = [
 
 [[package]]
 name = "lance-bitpacking"
-version = "4.1.0-beta.3"
+version = "5.0.0-beta.2"
 dependencies = [
  "arrayref",
  "paste",
@@ -4731,7 +4731,7 @@ dependencies = [
 
 [[package]]
 name = "lance-core"
-version = "4.1.0-beta.3"
+version = "5.0.0-beta.2"
 dependencies = [
  "arrow-array",
  "arrow-buffer",
@@ -4771,7 +4771,7 @@ dependencies = [
 
 [[package]]
 name = "lance-datafusion"
-version = "4.1.0-beta.3"
+version = "5.0.0-beta.2"
 dependencies = [
  "arrow",
  "arrow-array",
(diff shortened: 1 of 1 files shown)


rust/lance-io/src/uring/current_thread.rs
```diff
@@ -0,0 +1,440 @@
...
+
+/// Thread-local io_uring instance with pending requests
+struct ThreadLocalUring {
+    ring: IoUring,
+    pending: HashMap<u64, Arc<IoRequest>>,
+}
+
+thread_local! {
+    static URING: RefCell<Option<ThreadLocalUring>> = const { RefCell::new(None) };
+}
```
wjones127: suggestion: did you consider using [LazyCell](https://doc.rust-lang.org/std/cell/struct.LazyCell.html) for this? I suppose it would require panicking if you couldn't initialize the IO uring, but I'm not sure that so bad. If you can't read files then not much point in going on.

rust/lance-io/src/uring/current_thread.rs
```diff
@@ -0,0 +1,440 @@
...
+    /// This reuses the file handle caching infrastructure from UringReader
+    #[instrument(level = "debug")]
+    pub(crate) async fn open(
+        path: &Path,
+        block_size: usize,
+        known_size: Option<usize>,
+        io_tracker: Arc<IOTracker>,
+    ) -> Result<Box<dyn Reader>> {
+        // Determine block size with environment variable override
+        let block_size = std::env::var("LANCE_URING_BLOCK_SIZE")
```
wjones127: question: could this be expensive? Might it be worth caching on first access?

rust/lance-io/src/uring/reader.rs
```diff
@@ -0,0 +1,295 @@
...
+/// Data stored in the cache for each opened file.
+#[derive(Clone)]
+pub(super) struct CachedReaderData {
+    pub(super) handle: Arc<UringFileHandle>,
+    pub(super) size: usize,
+}
+
+/// Global cache of open file handles.
+/// Entries expire after 60 seconds to ensure files are eventually closed.
+pub(super) static HANDLE_CACHE: LazyLock<moka::future::Cache<CacheKey, CachedReaderData>> =
```
wjones127, line 57: praise: this is going to be very useful!

In another PR, I'm separating the data from file handles in the index cache (so that entries are serializable). Having a separate cache to make re-opening file handles might be useful.

rust/lance-io/src/object_store.rs
```diff
@@ -634,6 +661,31 @@ impl ObjectStore {
                 )
                 .await
             }
+            #[cfg(target_os = "linux")]
+            "file+uring" => {
+                // Check if current-thread mode enabled
+                let use_current_thread = std::env::var("LANCE_URING_CURRENT_THREAD")
+                    .map(|v| v == "1" || v.eq_ignore_ascii_case("true"))
+                    .unwrap_or(false);
```
wjones127, line 669: suggestion: use [str_is_truthy](https://docs.rs/lance-core/latest/lance_core/utils/parse/fn.str_is_truthy.html) for this:


```suggestion
                let use_current_thread = std::env::var("LANCE_URING_CURRENT_THREAD")
                    .map(str_is_truthy)
                    .unwrap_or(false);
```

rust/lance-io/src/uring/current_thread.rs
westonpace: Switched to LazyCell

rust/lance-io/src/uring/current_thread.rs
westonpace: I know cache this.

rust/lance-io/src/object_store.rs
westonpace: Done

PR #5779, opened 2026-01-22
fix: doc test failure, test_distributed_fts_multi_match_validation and test_duckdb

wojiaodoubao:
Currently there are 3 ci test failures, fix them.

## Case 1
image

## Case 2
image

## Case 3
image

### DIFF
--- a/.github/workflows/docs-deploy.yml
+++ b/.github/workflows/docs-deploy.yml
@@ -40,11 +40,6 @@ jobs:
         with:
           repository: lancedb/lance-ray
           path: lance-ray
-      - name: Checkout lance-huggingface
-        uses: actions/checkout@v4
-        with:
-          repository: lance-format/lance-huggingface
-          path: lance-huggingface
       - name: Checkout lance-namespace-impls
         uses: actions/checkout@v4
         with:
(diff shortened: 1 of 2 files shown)


wojiaodoubao:
Hi @Xuanwo @jackye1995 @yanghua , could you help review this when you have time, thanks very much.

python/python/lance/dataset.py
```diff
@@ -2138,11 +2138,11 @@ def merge_insert(
         ...             .execute(new_table)
         {'num_inserted_rows': 1, 'num_updated_rows': 2, 'num_deleted_rows': 0}
         >>> dataset.to_table().sort_by("a").to_pandas()
-           a  b     c
-        0  1  a     x
-        1  2  x     y
-        2  3  y     z
-        3  4  z  None
```
BubbleCal, line 2145: it's supposed to be `None` right? column `c` is with type `string` so it can't be `NaN`

python/python/tests/test_scalar_index.py
```diff
@@ -2429,26 +2429,18 @@ def compare_fts_results(
...
+            if single_df[col].dtype == "object":
+                assert set(single_values) == set(distributed_values), (
                     f"Column {col} content mismatch"
                 )
             else:
                 np.testing.assert_array_equal(
-                    single_values,
-                    distributed_values,
+                    np.sort(single_values.values),
+                    np.sort(distributed_values.values),
```
BubbleCal: we can't compare the sorted values because the order matters (sorted by scores)
we need to check whether something wrong with scoring, or maybe the 2 results are with the same scores

python/python/lance/dataset.py
wojiaodoubao: You are right. Let me revert it and see whether it can work.

PR #5783, opened 2026-01-22
perf: calculate cardinality lazily

### LINKED ISSUE #5714: Build time regression for BITMAP and BTREE index
Image

It seems we got slower at building these indices around January 1, 2026.

https://bencher.dev/perf/weston-lancedb?key=true&reports_per_page=4&branches_per_page=8&testbeds_per_page=8&benchmarks_per_page=8&plots_per_page=8&reports_page=1&branches_page=1&testbeds_page=1&benchmarks_page=4&plots_page=1&clear=true&tab=benchmarks&measures=2c54be52-0df5-49e9-ba36-4bc78fa0edcf&branches=f7b04560-8a34-4842-a7c6-089a8d72e724&heads=f7b04560-8a34-4842-a7c6-089a8d72e724&testbeds=eee246dd-6a86-436e-b59b-7936e7f78f7e&benchmarks=482f2e45-51b8-48e4-a8fb-9aad44c4ef47

Xuanwo:
This PR will fix #5714.

Cardinality calculation is slow, but we only need it when building the dictionary. This PR changes the calculation to a lazy approach.

```shell
BITMAP 5M low-float：229.725 (base) → 264.306 (bad) → 215.792 (good)
BTREE 5M low-float：216.628 (base) → 340.683 (bad) → 222.974 (good)
```

---


Xuanwo:
@wjones127, Thanks! Makes sense. In our case the dictionary decision is based on estimated size ratio, so the “cardinality threshold” isn’t a fixed constant, but we could derive an upper bound for cardinality from the ratio and early-exit the HLL once the estimate exceeds that bound.

Would you be OK if I keep this PR focused on fixing the regression (lazy cardinality), and do the early-exit logic as a follow-up PR with benchmarks to validate the impact?

wjones127:
> Would you be OK if I keep this PR focused on fixing the regression (lazy cardinality), and do the early-exit logic as a follow-up PR with benchmarks to validate the impact?

Yeah, definitely do that as a follow up if you want to do it. I think this PR is already valuable as is!

tomtomwombat:
You may want to consider using https://crates.io/crates/hyperloglockless for the `HyperLogLogPlus`. It has [O(1) cardinality](https://github.com/tomtomwombat/hyperloglockless/?tab=readme-ov-file#count-performance) queries and very fast inserts. It's also more accurate. I believe it can be a drop-in replacement. 
Also, because you are using [a low precision of 4](https://github.com/lance-format/lance/blob/e1489abc21438c5086ed76f03136534a2a9ee9a4/rust/lance-encoding/src/statistics.rs#L220), the HLL++ variant would be unnecessary since it immediately switches to "dense" HLL representation at very few items (wasting time on HLL++'s sparse -> dense conversion).

Disclaimer: I am the author of hyperloglockless.

PR #5784, opened 2026-01-22
fix: remove unreasonable nullable check for data types in hash_joiner during merge operation

zhangyue19921010:
During the invocation of add columns at https://github.com/lance-format/lance/blob/8c1842906bd275c32d403e4ddc78c7f77fa3de1f/rust/lance/src/dataset/fragment.rs#L1589, if the left table contains 100 rows of data and the right table contains 10 rows of data (with the remaining 90 rows being null), the validation at https://github.com/lance-format/lance/blob/8c1842906bd275c32d403e4ddc78c7f77fa3de1f/rust/lance/src/dataset/hash_joiner.rs#L192 will fail.

**This is unreasonable because the validation here is a legacy validation that is no longer applicable to V2_0 and above.**

```
ValueError: Invalid user input: Found rows on LHS that do not match any rows on RHS. 
Lance would need to write nulls on the RHS, but Lance does not yet support nulls for type Float32.,
/Users/xxx/xxx/lance/rust/lance/src/dataset/hash_joiner.rs:197:36
```

### DIFF
--- a/python/python/tests/test_dataset.py
+++ b/python/python/tests/test_dataset.py
@@ -1786,15 +1786,15 @@ def test_load_scanner_from_fragments(tmp_path: Path):
 
 def test_merge_data_legacy(tmp_path: Path):
     tab = pa.table({"a": range(100), "b": range(100)})
-    lance.write_dataset(tab, tmp_path / "dataset", mode="append", data_storage_version="legacy")
+    lance.write_dataset(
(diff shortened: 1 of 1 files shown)


zhangyue19921010:
> This looks good. Thanks for working on this!

Thanks @wjones127 ：）

PR #5788, opened 2026-01-22
test: update doctest to new pandas behavior

westonpace:


wjones127 (approved):

python/python/lance/dataset.py
```diff
@@ -2138,11 +2138,11 @@ def merge_insert(
...
-           a  b     c
-        0  1  a     x
-        1  2  x     y
-        2  3  y     z
-        3  4  z  None
+           a  b    c
+        0  1  a    x
+        1  2  x    y
+        2  3  y    z
+        3  4  z  NaN
```
BubbleCal, line 2145: Does pandas use NaN to represent None now?
That's a little bit weird... esp column c is with type string

Xuanwo:
Oh, please feel free to cherry pick this PR: https://github.com/lance-format/lance/pull/5789

python/python/lance/dataset.py
jackye1995: > That's a little bit weird... esp column c is with type string

+10000, it almost feels like a bug to me, not sure if there are any related discussions in pandas about it

westonpace: I'll investigate

westonpace:
```python
>>> import pandas as pd
>>> pd.DataFrame({"x": ["x", "y", "z"]})
   x
0  x
1  y
2  z
>>> pd.DataFrame({"x": ["x", "y", None]})
     x
0    x
1    y
2  NaN
```

Maybe a pandas bug but :shrug: .  I'll see if there are any tickets on pandas repo.

westonpace: It is intentional:

> The main characteristic of the new string data type:
>
>    Inferred by default for string data (instead of object dtype)
>
>    The str dtype can only hold strings (or missing values), in contrast to object dtype. (setitem with non string fails)
>
>    The missing value sentinel is always NaN (np.nan) and follows the same missing value semantics as the other default dtypes.

PR #5792, opened 2026-01-22
feat: add scanner.plan_splits function

hamersaw:
This PR adds a `plan_splits` function the to `Scanner` struct. The goal is that this serves as a singular endpoint where distributed compute frameworks can effectively partition a Lance dataset for parallelized processing. The main goals are:
(1) Prune fragments that do not satisfy a filter (if exists): We use an index lookup to determine which fragments contain rows (and which do not) to prune unnecessary fragments.
(2) Bin pack fragments into spiits: Distributed compute frameworks typically work best with a "sweet-spot" partition size. Within Lance, this means a partition should typically contain multiple fragments. We expose a user configurable strategy, namely max row count or split size, and then estimate row sizes based on the schema to determine the size of the resultant split.

### DIFF
--- a/java/lance-jni/src/blocking_scanner.rs
+++ b/java/lance-jni/src/blocking_scanner.rs
@@ -5,14 +5,16 @@ use std::sync::Arc;
 
 use crate::error::{Error, Result};
 use crate::ffi::JNIEnvExt;
-use crate::traits::{import_vec_from_method, import_vec_to_rust};
+use crate::traits::{export_vec, import_vec_from_method, import_vec_to_rust, IntoJava};
 use arrow::array::Float32Array;
 use arrow::{ffi::FFI_ArrowSchema, ffi_stream::FFI_ArrowArrayStream};
 use arrow_schema::SchemaRef;
-use jni::objects::{JObject, JString};
+use jni::objects::{JObject, JString, JValueGen};
 use jni::sys::{jboolean, jint, JNI_TRUE};
 use jni::{sys::jlong, JNIEnv};
-use lance::dataset::scanner::{ColumnOrdering, DatasetRecordBatchStream, Scanner};
+use lance::dataset::scanner::{
+    ColumnOrdering, DatasetRecordBatchStream, Scanner, Split, SplitFragment, SplitOptions,
+};
 use lance_index::scalar::inverted::query::{
     BooleanQuery as FtsBooleanQuery, BoostQuery as FtsBoostQuery, FtsQuery,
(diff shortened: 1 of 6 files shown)


hamersaw:
Before I plumb this through to the Lance Spark connector just wanted to get some input from interested parties:

@majin1102 / @fangbo in [this thread](https://github.com/lance-format/lance/issues/5731#issuecomment-3762819221) I know you expressed interest in a solution. This does currently work with zone maps. @fangbo you'll recognize a large bit of code from[ your PR](https://github.com/lance-format/lance/pull/5625) - thanks!

@Jay-ju IIUC [your PR here](https://github.com/lance-format/lance/pull/4930) is targeted at estimating row counts to achieve similar ends. I really like the idea of index hinting, as in my testing I noticed filtering index choices were not always what I expected them to be.

python/python/lance/dataset.py
```diff
@@ -5189,6 +5189,34 @@ def analyze_plan(self) -> str:
 
         return self._scanner.analyze_plan()
 
+    def plan_splits(
+        self, max_split_size_bytes: Optional[int] = None
```
hamersaw: Will need to update this to include both `max_split_size_bytes` and `max_row_count` options, with one trumping the other if both are provided. I'm interested if people think this paradigm is useful? My intuition is that since we are estimating row sizes based on the schema that we could be VERY wrong (just using 64B for everything that is not known size - string / blob could be 1B - 1M+). In these scenarios a user will know their data better and can use a `max_row_count` to target a partition size. So basically, hopefully most use-cases we're close and estimation works well, but there are knobs to fine-tune in the other cases.

rust/lance/src/dataset/scanner.rs
```diff
@@ -858,6 +890,214 @@ impl Scanner {
...
+    /// that can be processed independently by multiple workers or threads. It uses a
+    /// bin-packing algorithm to create balanced splits based on estimated row counts.
+    pub async fn plan_splits(&self, strategy: Option<SplitPackStrategy>) -> Result<Vec<Split>> {
+        // Collect initial set of fragments to scan
+        let fragments = if let Some(fragments) = self.fragments.as_ref() {
+            Arc::new(fragments.clone())
+        } else {
+            Arc::new(self.dataset.fragments().as_ref().clone())
+        };
+
```
LuQQiu, line 926: does this change need to take care of the scanner range e.g. scan_range_before_filter?

hamersaw: Great point. I'm going to have to look into this, but I think it probably should!

hamersaw:
closing in favor of https://github.com/lance-format/lance/pull/5863

PR #5799, opened 2026-01-23
ci: fix MSRV check

wjones127:
The `rust-toolchain.toml` file has been making our current MSRV check inert. This environment variable should override it.

Also updates MSRV to 1.88 (same as DataFusion) and makes associated clippy fixes.

### DIFF
--- a/java/lance-jni/src/utils.rs
+++ b/java/lance-jni/src/utils.rs
@@ -276,7 +276,7 @@ pub fn get_vector_index_params(
                 let jarray: JFloatArray = centroids_obj.into();
                 let length = env.get_array_length(&jarray)?;
                 if length > 0 {
-                    if (length as usize) % num_partitions != 0 {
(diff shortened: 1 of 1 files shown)


.github/workflows/rust.yml
```diff
@@ -268,8 +268,9 @@ jobs:
       - name: Install ${{ matrix.msrv }}
         run: |
           rustup toolchain install ${{ matrix.msrv }}
-          rustup default ${{ matrix.msrv }}
       - name: cargo +${{ matrix.msrv }} check
+        env:
+          RUSTUP_TOOLCHAIN: ${{ matrix.msrv }}
```
westonpace, line 273: This tells it to ignore the toolchain file basically?

wjones127: Yeah it's higher up in the resolution chain: https://rust-lang.github.io/rustup/overrides.html

PR #5804, opened 2026-01-26
feat: clip scalar quantizatoin

BubbleCal:
| clip_percent | k   | recall   |
|--------------|-----|----------|
| 0.05         | 1   | 0.864700 |
| 0.05         | 10  | 0.888260 |
| 0.05         | 100 | 0.908364 |
| 0.0          | 1   | 0.833800 |
| 0.0          | 10  | 0.854890 |
| 0.0          | 100 | 0.878491 |

gain ~3% recall improvement

### DIFF
--- a/rust/lance-index/src/vector/sq.rs
+++ b/rust/lance-index/src/vector/sq.rs
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-FileCopyrightText: Copyright The Lance Authors
 
-use std::{ops::Range, sync::Arc};
+use std::{cmp::Ordering, ops::Range, sync::Arc};
 
 use arrow::array::AsArray;
 use arrow::datatypes::{Float16Type, Float32Type, Float64Type};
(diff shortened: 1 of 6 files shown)


rust/lance-index/src/vector/sq.rs
```diff
@@ -81,10 +90,81 @@ impl ScalarQuantizer {
...
+            global_upper = Some(match global_upper {
+                None => upper,
+                Some(current) => match upper.total_cmp(&current) {
+                    Ordering::Greater => upper,
+                    _ => current,
+                },
+            });
+        }
 
+        self.metadata.bounds = global_lower.unwrap()..global_upper.unwrap();
```
majin1102: It seems global_lower and global_upper must have values here. That said, I wonder if we could use a safe pattern just for readability:

```
if let (Some(lower), Some(upper)) = (global_lower, global_upper) {
    self.metadata.bounds = lower..upper;
} else {
    return Err(Error::Internal {
        message: "Should never enter....".to_string(),
        location: location!(),
    });
}
```

Or we could just use
```
let mut global_lower = f64::MAX;
let mut global_upper = f64::MIN;
```

WDYT

majin1102 (commented):
Thanks for working on this.
left one comment. Please take a look

rust/lance-index/src/vector/sq.rs
BubbleCal: fixed

majin1102 (approved):
LGTM

BubbleCal:
will get some new recall numbers with the new changes

PR #5807, opened 2026-01-26
feat(rust): support datafusion url table

### LINKED ISSUE #5806: Support lance url table in datafusion
Datafusion supports url table used like this:

`select * from 'S3://bucket1/dataset2'`

We could support this by implement a lance `UrlTableFactory`. And beyond `xxx.lance` tables, we should also accept `xxx.parquet`, 'xxx.csv', 'xxx.json' as long as datafusion has supported them.

majin1102:
Close #5806


majin1102:
@wjones127 @timsaucer 
Could you take a look when you have time?  Thanks

PR #5824, opened 2026-01-27
fix(java): transaction fatal bug in java transaction api

wojiaodoubao:
Fatal Bug in transaction definitions in the Java API. For Merge, Overwrite, and Projection operations, the Java API uses Arrow Schema to store schema information. The problem is that Arrow Schema and Lance-defined Schema cannot achieve equivalent conversion, which causes loss of information. 

For example, when adding columns using a Merge transaction, the Schema & Field in the Java API's Merge transaction does not carry ID information. During the conversion from a Java Merge transaction to a Rust Merge transaction, the Arrow Schema is converted to Lance Schema and each Field is assigned an ID from scratch every time(https://github.com/lance-format/lance/blob/b65e2cd1/java/lance-jni/src/transaction.rs#L848-L876). If a column is dropped before adding a new one, this will lead to field ID misalignment. Field id is an example, other information like field dictionary can get lost too.

~~In this pr, I modified LanceSchema and LanceField, enable equivalent schema conversion between java and rust.~~

### DIFF
--- a/java/lance-jni/src/blocking_dataset.rs
+++ b/java/lance-jni/src/blocking_dataset.rs
@@ -815,14 +815,6 @@ pub fn inner_commit_overwrite<'local>(
     let c_schema_ptr = arrow_schema_addr as *mut FFI_ArrowSchema;
     let c_schema = unsafe { FFI_ArrowSchema::from_raw(c_schema_ptr) };
     let arrow_schema = Schema::try_from(&c_schema)?;
-
-    // In general, in an overwrite scenario, both the schema and fragments are brand new,
-    // which means that field IDs are allocated sequentially starting from 0 for each field.
-    // In this case, there is no need to explicitly set field IDs in the arrow schema; the results
-    // inferred by LanceSchema is consistent with those in the fragments. However, in some advanced
-    // scenarios, the field IDs in fragments do not follow the "sequential allocation starting
-    // from 0" logic, such as when reusing certain data files. In such cases, it is mandatory to
-    // specify the field IDs explicitly in the arrow schema.
     let schema = LanceSchema::try_from(&arrow_schema)?;
 
     let op = Operation::Overwrite {
--- a/java/src/main/java/org/lance/operation/SchemaOperation.java
+++ b/java/src/main/java/org/lance/operation/SchemaOperation.java
@@ -20,7 +20,17 @@ import org.apache.arrow.vector.types.pojo.Schema;
 
 import java.util.Objects;
 
-/** Schema related base operation. */
+/**
+ * Schema related base operation.
+ *
+ * <p>Each field will be assigned a field id when transaction commits, in the following order:
+ *
+ * <ol>
+ *   <li>Parse from field metadata with key {@code lance:field_id}.
+ *   <li>Otherwise, set field id from txn read version dataset's schema field (with the same name).
+ *   <li>Otherwise, allocate based on the max field id of the dataset.
+ * </ol>
+ */
 public abstract class SchemaOperation implements Operation {
   private final Schema schema;
   private ArrowSchema cSchema;


wojiaodoubao:
Hi @majin1102 @jackye1995 , could you help review this when you have time, thanks very much.

majin1102:
Nice catch!

I think the current schema transformation in jni is unreasonable. For `Merge` and `Project` operations, we need to take care of the field id rebinding.

That said, exposing LanceSchema for Overwrite, Merge, and Project might not be ideal, as it would force us to surface the field ID rebinding logic to the Java side—adding coupling without clear benefit. Perhaps we could instead address this in `convert_schema_from_operation`? For example:

```
    match operation_name {
        "Project" => {
            let projected_schema = base_schema.project_by_schema(&arrow_schema, OnMissing::Error, OnTypeMismatch::Error)?;
            Ok(projected_schema)
        },
        "Merge" => {
            let mut new_schema = base_schema.merge(&arrow_schema)?;
            new_schema.set_field_id(Some(manifest.max_field_id()));
            Ok(new_schema)
        },
        "Overwrite" | "Create" => {
            // keep the original way
            let mut lance_schema = LanceSchema::try_from(&arrow_schema)?;
            lance_schema.set_field_id(None);
            Ok(lance_schema)
        },
```

I think the key idea is to rebind field-ids based on the current dataset. What do you think

wojiaodoubao:
Have a offline discussion with @majin1102 , here is my thoughts.

## Key Points in short
* The role of the Low Level API(transaction) is to faithfully execute low-level operations, while the role of the High Level API is to address usability issues.
* The low level api is for experts and engine-integrations, the high level api is for users.

## Detail
When performing operations like add_columns or drop_columns in Lance, Lance provides two sets of APIs: a High Level API and a Low Level API.
* The High Level API is relatively simple to use; users do not need to concern themselves with the table format layer's schema or fragments. They only need to provide data/UDFs for new columns or the names of columns to drop.
* The Low Level API is a transaction-based API. Users need to manually construct transaction operations, which contain table format information such as schema and fragments, so users must be aware of the table format details. 

From my understanding, the Low Level API aims to provide sufficient flexibility for specific scenarios (such as integrating Spark with Lance tables), and it is targeted at Lance experts rather than ordinary Lance users.

We should not attempt to hide table format layer details in the Low Level API (for example, we should not hide field ids and dictionaries in Java's Merge operation). When providing transaction APIs in Python bindings or Java bindings, we should ensure that transactions defined in different languages can be *equivalently converted* to Rust transactions. 
If our goal is to hide table format details to improve interface usability, we should do so in the High Level API rather than the Low Level API.

Java's current transaction API violates this design principle. When adding columns in Spark, it first retrieves the Lance schema via the fragment API, converts it to an Arrow schema when committing (losing information), and then converts it again to a Lance schema at the Rust layer to construct a Merge Operation.

Another approach is to think that this problem is merely that field ids are not correctly assigned, and it can be solved by fixing the field ids according to the transaction type when committing the transaction. This approach still blurs the distinction between the High Level API and the Low Level API.
The role of the Low Level API is to faithfully execute low-level operations, while the role of the High Level API is to address usability issues. When a user submits a Merge txn using the Low Level API, the only thing the Low Level API should do is to commit this transaction, rather than inferring whether the purpose of Merge is to add new columns, modify column data types, or perform other operations, and then reconstructing the dictionary and field ids based on the operation type.

The python bindings fully complies with the design principles of the Low Level API. The Low Level API allows committing incorrect transactions, and the caller needs to ensure that the Low Level API is always used correctly. This means that incorrect use can corrupt the dataset, and the following example illustrates this scenario.
```
def merge_case():
    import lance
    import pyarrow as pa
    import pyarrow.compute as pc

    # Prepare ds
    table = pa.Table.from_pylist([{"name": "Alice", "age": 20, "text2": "a", "scores2": [0, 1, 2], "embedding2": [1.0, 2.0, 3.0]},
                                  {"name": "Bob", "age": 30, "text2": "b", "scores2": [0, 2, 3], "embedding2": [1.0, 3.0, 2.0]}])
    ds = lance.write_dataset(table, "./alice_and_bob.lance")

    # High Level API.
    # Drop colum to make holes in field id.
    ds.drop_columns(["name", "scores2"])
    print(ds.to_table())

    # Create fragments
    def double_age(batch: pa.RecordBatch) -> pa.RecordBatch:
        doubled = pc.multiply(batch["age"], 2)
        return pa.record_batch([doubled], ["age_doubled"])

    fragments = []
    for fragment in ds.get_fragments():
        new_fragment, new_schema = fragment.merge_columns(double_age, columns = ['age'])
        fragments.append(new_fragment)

    # new_schema is the right schema, but in this case we don't want to use it.
    # Let's create a bad schema with wrong field id.
    # The lance_schema created by LanceSchema.from_pyarrow doesn't have field ids.
    # The field ids will be assigned from scratch when commit Merge txn, since we
    # have a hole, the ids mapping will be wrong.
    table = pa.table({"age": [0], "text2": ["text"], "embedding2": [[1.0, 2.0, 3.0]], "age_doubled": [0]})
    lance_schema = LanceSchema.from_pyarrow(table.schema)

    # The low level api should always faithfully commits transactions. 
    operation = lance.LanceOperation.Merge(fragments, lance_schema)
    dataset = lance.LanceDataset.commit("./alice_and_bob.lance", operation, read_version = ds.version)
    print(dataset.to_table()) # Got Error
```

Another case is a bug of Project txn. The `LanceSchema.from_pyarrow` creates a bad schema, and `LanceDataset.commit` faithfully commits the Project txn, which is correct, finally causing bad dataset.
```
def project_case():
    import lance
    import pyarrow as pa
    from lance.schema import LanceSchema

    # Prepare ds
    table = pa.Table.from_pylist([{"name": "Alice", "age": 20, "text2": "a", "scores2": [0, 1, 2], "embedding2": [1.0, 2.0, 3.0]},
                                  {"name": "Bob", "age": 30, "text2": "b", "scores2": [0, 2, 3], "embedding2": [1.0, 3.0, 2.0]}])
    ds = lance.write_dataset(table, "./alice_and_bob.lance")

    # Drop colum to make holes in field id.
    ds.drop_columns(["name", "scores2"])
    print(ds.to_table())

    # The lance_schema created by LanceSchema.from_pyarrow doesn't have field ids.
    # The field ids will be assigned from scratch when commit Project txn, since we
    # have a hole, the ids mapping will be wrong.
    table = pa.table({"name0": [0], "embedding2": [[1.0, 2.0, 3.0]]})
    lance_schema = LanceSchema.from_pyarrow(table.schema)
    print(lance_schema)
    
    # The low level api should always faithfully commits transactions. 
    operation = lance.LanceOperation.Project(lance_schema)
    dataset = lance.LanceDataset.commit("./alice_and_bob.lance", operation, read_version=ds.version)
    print(dataset.to_table()) # Got Error
```

wojiaodoubao:
Thanks @majin1102 for his suggestion. I think we have 2 solutions.

The first solution approach is to directly fix the issue based on the current codebase, just as what the current PR does. It will be a breaking change.

The second solution approach is to introduce SchemaOperationV2 along with the corresponding MergeV2, OverwriteV2, ProjectV2, LanceSchemaV2, and LanceFieldV2 starting from compatibility considerations. The functionality of the current PR (following the style of the python bindings) would be implemented within these V2 components, while preserving the existing implementation. It won't be a breaking change.

Hi @jackye1995 @westonpace @majin1102 , what do you think?

majin1102:
I think we could fix pyo3 field id rebinding issues together if it exists

jackye1995:
Thanks for bringing this up, this is something that I always felt was a bit broken, but I was surprised no one has brought it up so far...

I am thinking if there is a way to make the user experience unchanged. Can we use Arrow's built-in column metadata to store Lance information. For example, user has schema:

```java
  Schema newSchema = new Schema(Arrays.asList(
      Field.nullable("id", new ArrowType.Int(32, true)),
      Field.nullable("age", new ArrowType.Int(32, true))  // new column
  ));

  Merge.builder()
      .fragments(fragments)
      .schema(newSchema)  // Still Arrow Schema
      .build();
```

 Internally, when the operation is processed:

```java
  // Inside Merge/Overwrite/Project, before passing to Rust:
  Schema annotated = LanceSchemaUtils.annotate(userSchema, dataset.getLanceSchema());
  // Now each field has metadata like:
  //   "lance:field_id" -> "0"
  //   "lance:parent_id" -> "-1"
  //   "lance:encoding" -> "PLAIN"
```

The annotation logic follows:
  1. Match fields by name/path against base schema
  2. For existing fields: copy `lance:field_id`, `lance:encoding`, `lance:dictionary_*` from base
  3. For new fields: assign `lance:field_id = maxFieldId + 1`, +2, ...
  4. Rust side reads metadata during Arrow→Lance conversion, uses it if present

This can make all changes internal only, and advanced user can also override the information if necessary.

What do we think?

wojiaodoubao:
> The annotation logic follows:......

Agree. This should be the most consensual approach. Let me fix it this way.

wojiaodoubao:
Hi @jackye1995 @majin1102 , could you help review when you have time, thanks very much.

java/lance-jni/src/transaction.rs
```diff
@@ -676,6 +676,20 @@ fn convert_to_rust_transaction(
     java_dataset: Option<&JObject>,
 ) -> Result<Transaction> {
     let read_ver = env.get_u64_from_method(&java_transaction, "readVersion")?;
+
+    // Validate read version matches the dataset version.
+    if let Some(java_dataset) = java_dataset {
```
majin1102: Wondering if we should checkout the read_version for this case. The point here I think is do python and rust API forbid using a different version to commit transactions?

What do you think

java/lance-jni/src/transaction.rs
```diff
@@ -724,11 +738,17 @@ fn convert_schema_from_operation(
...
+    let arrow_schema = Schema::try_from(&c_schema)?;
+
+    // Derive field ids based on the current dataset schema.
+    let dataset_guard =
+        unsafe { env.get_rust_field::<_, _, BlockingDataset>(java_dataset, NATIVE_DATASET) }?;
+    let current_schema = dataset_guard.inner.schema();
+    let max_field_id = dataset_guard.inner.manifest().max_field_id();
+
+    current_schema
+        .evolution(&arrow_schema, Some(max_field_id))
```
majin1102: From my understanding, this `evolution` mechanism transforms Arrow schemas to Lance schemas. However, "evolution" is a misleading term here—at lease not the same case of current `schema_evolution.rs`.

Since the Rust core already provides `project_by_schema` and `merge` for projection and merging, can we reuse these directly instead of introducing a new abstraction? This would avoid duplicated logic unless there's a specific gap I'm missing. What do you think?

rust/lance-core/src/datatypes/field.rs
```diff
@@ -49,6 +49,10 @@ pub const LANCE_UNENFORCED_PRIMARY_KEY: &str = "lance-schema:unenforced-primary-
 pub const LANCE_UNENFORCED_PRIMARY_KEY_POSITION: &str =
     "lance-schema:unenforced-primary-key:position";
 
+/// Use this config key in Arrow field metadata to specify the field id of the lance field.
+/// The value should be non-negative i32 value. Any negative value will be seen as -1.
+pub const LANCE_FIELD_ID_KEY: &str = "lance:field_id";
```
majin1102, line 54: I don't object to storing field_id in metadata, but just to clarify—is this actually unrelated to the current bug we're fixing?

java/lance-jni/src/transaction.rs
wojiaodoubao: For each transaction, there is a read version. We should infer lance schema from arrow schema based on the same dataset of transaction read version. So I think we should check here to protect the semantics.

In the current java api, the read version is always the same as the input dataset's version. With this check, we can protect the semantics from unintentionally break.

java/lance-jni/src/transaction.rs
wojiaodoubao: Do you refer to `Schema::project_by_schema` when you say `project_by_schema`? The semantics are different between the two "project". What `Schema::project_by_schema` does is to project from a lance schema. But what we do here, is to infer field id using a lance schema in a `Project` transaction. So is merge.

I thought about the func name but failed to find a good one. Will `rebuild`, `infer_build`, `transform` be better? What do you think?

rust/lance-core/src/datatypes/field.rs
wojiaodoubao: I think the field_id in metadata is for advanced users. It is not directly for fixing this bug, but it is a part of the whole solution.

java/lance-jni/src/transaction.rs
majin1102: I think for this validation we should keep Java SDK behivior just like python and rust. If rust does forbid this operation. I would be OK with that

java/lance-jni/src/transaction.rs
majin1102: > The semantics are different between the two "project". What Schema::project_by_schema does is to project from a lance schema.

Two AI tools told me it accept arrow schema. Let me check tomorrow. What I want to say is I think it's not that necessary to build a general transformation if we already had functions to deal with each operation. It seems re-inventing things and bringing more work

rust/lance-core/src/datatypes/field.rs
majin1102: OK, again I don't object that.
But I think I need to find some context to know why it was not done(I remember some context refering Paimon)

wojiaodoubao:
Hi @majin1102 , please help take a look when you have time, thanks very much~

rust/lance-core/src/datatypes/schema.rs
```diff
@@ -819,6 +788,42 @@ impl TryFrom<&ArrowSchema> for Schema {
...
+        }
+        Ok(())
+    }
+}
+
+/// Convert `arrow2::datatype::Schema` to Lance
+impl TryFrom<&ArrowSchema> for Schema {
+    type Error = Error;
+
+    fn try_from(schema: &ArrowSchema) -> Result<Self> {
```
majin1102, line 831: Sorry. What's the difference for this TryFrom. Did't get

rust/lance-core/src/datatypes/schema.rs
```diff
@@ -1604,12 +1609,419 @@ pub fn escape_field_path_for_project(name: &str) -> String {
...
+
+    let mut schema = Schema { fields, metadata };
+    schema.set_field_id(Some(max_id));
+    schema.validate()?;
+    schema.verify_primary_key()?;
+
+    Ok(schema)
+}
+
+fn set_id_and_metadata(field: &mut Field, base_field: &Field) -> Result<()> {
```
majin1102: First we have transformed arrow field to lance field. I think metadata should been ingested at that time. WDYT

rust/lance-core/src/datatypes/schema.rs
```diff
@@ -1604,12 +1609,419 @@ pub fn escape_field_path_for_project(name: &str) -> String {
...
+///
+/// For each field from `ArrowSchema`:
+///   - If the field contains `lance:field_id` in metadata, it is an equivalent representation of
+///     `LanceField`, where the corresponding information is stored in `ArrowField` metadata.
+///     Create lance field with the specified id and metadata.
+///   - Otherwise, if a lance field with same name exists in `LanceSchema`, it is derived from the
+///     base_schema. An error is returned if the 2 datatypes do not match.
+///   - Otherwise, create a new field without id and metadata. The id will finally be assigned
+///     based on max_field_id.
+pub fn create_schema_from_arrow(
```
majin1102: I'm not sure it's a good idea to have this. From the method name I can't figure out why we don't use `TryFrom` and `set_field_id` directly and We don't know what exactly is doing here if we don't read that long comment. And If we have this, should we always use this instead of `TryFrom` arrow schema?  I propose to do this in another place and re-use some existed code.  However, there may be some aspects I haven't considered.  I'd appreciate it if you could point them out

rust/lance-core/src/datatypes/field.rs
```diff
@@ -1050,6 +1059,19 @@ impl TryFrom<&ArrowField> for Field {
...
+                .parse::<i32>()
+                .map_err(|e| Error::invalid_input(e.to_string(), location!()))?;
+            if field_id < 0 {
+                -1
+            } else {
+                field_id
+            }
+        } else {
+            -1
+        };
```
majin1102, line 1063:
```suggestion
        let mut metadata = field.metadata().clone();
        
        let id = match metadata.remove(LANCE_FIELD_ID_KEY) {
            Some(val) => val.parse::<i32>()
                .map_err(|e| Error::invalid_input(e.to_string(), location!()))?
                .max(-1),
            None => -1,
        };
```

rust/lance-core/src/datatypes/schema.rs
wojiaodoubao: I think this is a git diff display issue. What I do is making `verify_primary_key` an independent function, so it can be reused. I don't know why it displays as I create a new TryFrom.

rust/lance-core/src/datatypes/schema.rs
wojiaodoubao: +1, thanks

rust/lance-core/src/datatypes/schema.rs
wojiaodoubao: Let me refactor this

rust/lance-core/src/datatypes/field.rs
wojiaodoubao: thanks!

wojiaodoubao:
Sorry for my late response! I refactored and rebased this pr, following all comments.

Hi @majin1102 , could you help review this when you have time, thanks very much~

rust/lance-core/src/datatypes/schema.rs
majin1102: Got. 
Wondering why this change is necessary for this PR?

java/lance-jni/src/transaction.rs
```diff
@@ -1322,3 +1516,262 @@ fn inner_commit_to_uri<'local>(
     let blocking_ds = BlockingDataset { inner: dataset };
     blocking_ds.into_java(env)
 }
+
+#[cfg(test)]
```
majin1102, line 1520: Adding unitest here is a good option. But I don't know if CI would enter here.

I think adding a test for Project and Merge operation in this scenario seperately may be enough. What do you think

java/lance-jni/src/transaction.rs
```diff
@@ -1281,13 +1452,37 @@ fn inner_commit_to_uri<'local>(
...
 
+    let namespace_info = extract_namespace_info(env, &namespace_obj, &table_id_obj)?;
+    let (open_namespace, open_table_id) = match &namespace_info {
+        Some((ns, tid)) => (Some(ns.clone()), Some(tid.clone())),
+        None => (None, None),
+    };
+
+    // Open the read dataset using the same storage options (and provider, if any) so that
+    // `convert_to_rust_transaction` can derive schema/field ids based on the target dataset.
+    let mut ds = BlockingDataset::open(
```
majin1102, line 1463: Is this accidentally introduced by rebasing?

java/src/test/java/org/lance/operation/MergeTest.java
```diff
@@ -157,9 +159,24 @@ void testReplaceAsDiffColumns(@TempDir Path tempDir) throws Exception {
           }
           ageRoot.setRowCount(rowCount);
 
+          LanceSchema initialLanceSchema = initialDataset.getLanceSchema();
```
majin1102, line 378: I think we should introduce test cases with non-contiguous field IDs. For example, use a schema with IDs [1, 3, 5] and validate the merge operation.

We could use a project operation before merging to generate a schema like this. WDYT

java/lance-jni/src/transaction.rs
wojiaodoubao: It is introduced by rebasing but not accidentally. The java transaction has been modified to enable commit with uri. Here we load a blocking dataset from uri if it exists, then we can get the schema of read_version.

java/lance-jni/src/transaction.rs
wojiaodoubao: > Wondering why this change is necessary for this PR?

Yes it is. `verify_primary_key` is needed each time we construct a lance schema. Originally, we only need it in `TryFrom<&ArrowSchema> for LanceSchema`. Now we introduce `SchemaExt::from_arrow_schema` in lance-jni, we need to call it in it too. So I extract `verify_primary_key` as a function from `TryFrom<&ArrowSchema> for LanceSchema`.

wojiaodoubao: > Adding unitest here is a good option. But I don't know if CI would enter here.

Yes, it can be ran. Because we have `lance-jni-test` in pom.xml. I did a test by forcing the ut to fail and do mvn install, then I found it failed.
```
[INFO] running 1 test
[INFO] test transaction::tests::test_create_schema_from_arrow ... FAILED
[INFO]
[INFO] failures:
[INFO]
[INFO] ---- transaction::tests::test_create_schema_from_arrow stdout ----
[INFO]
[INFO] thread 'transaction::tests::test_create_schema_from_arrow' (82145016) panicked at src/transaction.rs:1534:9:
[INFO] panic here for test
[INFO] note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
[INFO]
[INFO]
[INFO] failures:
[INFO]     transaction::tests::test_create_schema_from_arrow
[INFO]
[INFO] test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
```

> I think adding a test for Project and Merge operation in this scenario seperately may be enough. What do you think

I think may be we can have both. `test_create_schema_from_arrow` is a robust test, covers many cases.

java/src/test/java/org/lance/operation/MergeTest.java
wojiaodoubao: Good point! I add a new test case testMergeNewColumnWithNonContiguousFieldId

PR #5827, opened 2026-01-27
docs: update dataset source in llm_dataset_creation.md

lhoestq:
`Salesforce/wikitext` is more future proof than its alias/legacy name `wikitext`


Xuanwo (approved):

Xuanwo:
Thank you!

PR #5831, opened 2026-01-27
feat(scanner): add row_id allowlist support

fredlarochelle:
We hit a severe bottleneck when filtering on very large `IN (...)` lists (millions of comparisons) in scalar filters. We can cheaply pre-compute a bitmap row-ID allowlist in memory, but Lance had no way to inject it into the query pipeline. This PR exposes that path so we can bypass expensive scalar filtering and push the allowlist directory into scan / FTS / vector execution, including IO-level pruning on non-legacy storage. Once this PR is merged, we plan to add LanceDB support.

### What this PR does

- Adds `Scanner:row_id_allowlist` for supplying a row-ID allowlist (dataset row-ID space).
- Enforces allowlist on plain scans by intersecting with any `TakeOperations`, then treating any remaining filters as refine-only.
- Plumbs allowlist through vector and FTS prefilter paths (ANN/KNN + FTS), including unindexed/flat paths.
- Adds IO‑level pruning on non‑legacy storage by passing allowlist into `FilteredReadExec`.
- Ensures allowlist always intersects with deletion masks and other filters.

### Semantics

- Allowlist is a hard constraint in dataset row-ID space (`_rowid`)
- If stable row IDs are disabled, `_rowid` is row address, allowlist must be snapshot-bound.
- If stable row IDs are enabled, `_rowid` is stable across versions.

### Tests added

- Runtime: allowlist with ANN + flat KNN, indexed + unindexed FTS, unstable row IDs, deletions.
- IO: allowlist mask in `FilteredReadExec` (including scan range ordering).
- Plain scan + filter intersection

### Performance evidence (local, not committed)

Plain scan with large IN filter vs allowlist:

- **1e6 rows**
  - filter_in_only: ~18.6 ms
  - allowlist_only: ~1.08 ms
  - ~17× faster
- **1e7 rows**
  - filter_in_only: ~215.3 ms
  - allowlist_only: ~6.95 ms
  - ~31x faster

### Notes / limitations

- IO-level allowlist pruning only applies to non-legacy storage.
- Legacy paths remains compute-level filtering only.

### DIFF
--- a/rust/lance/src/dataset/scanner.rs
+++ b/rust/lance/src/dataset/scanner.rs
@@ -5844,7 +5844,7 @@ mod test {
             .values();
 
         let allowlist = vec![row_ids[10], row_ids[20], row_ids[30]];
-        let offsets = vec![row_offsets[20], row_offsets[30], row_offsets[40]];
+        let offsets = [row_offsets[20], row_offsets[30], row_offsets[40]];
         let filter = format!(
             "_rowoffset IN ({}, {}, {}) AND i >= 0",
             offsets[0], offsets[1], offsets[2]
--- a/rust/lance/src/dataset/scanner.rs
+++ b/rust/lance/src/dataset/scanner.rs
@@ -5822,6 +5822,59 @@ mod test {
     }
 
     #[tokio::test]
+    async fn test_row_id_allowlist_with_take_rowoffset_stable() -> Result<()> {
+        let test_ds = TestVectorDataset::new(LanceFileVersion::Stable, true).await?;
+        let dataset = &test_ds.dataset;
+
+        let batch = dataset
+            .scan()
+            .project(&[ROW_OFFSET])?
+            .with_row_id()
+            .try_into_batch()
+            .await?;
+        let row_offsets = batch
+            .column_by_name(ROW_OFFSET)
+            .unwrap()
+            .as_primitive::<UInt64Type>()
+            .values();
+        let row_ids = batch
+            .column_by_name(ROW_ID)
+            .unwrap()
+            .as_primitive::<UInt64Type>()
+            .values();
+
+        let allowlist = vec![row_ids[10], row_ids[20], row_ids[30]];
+        let offsets = vec![row_offsets[20], row_offsets[30], row_offsets[40]];
+        let filter = format!(
+            "_rowoffset IN ({}, {}, {}) AND i >= 0",
+            offsets[0], offsets[1], offsets[2]
+        );
+
+        let filtered = dataset
+            .scan()
+            .project(&["i"])?
+            .with_row_id()
+            .row_id_allowlist(allowlist)
+            .filter(&filter)?
+            .try_into_batch()
+            .await?;
+        let filtered_ids: BTreeSet<u64> = filtered
+            .column_by_name(ROW_ID)
+            .unwrap()
+            .as_primitive::<UInt64Type>()
+            .values()
+            .iter()
+            .copied()
+            .collect();
+
+        assert_eq!(
+            filtered_ids,
+            BTreeSet::from_iter(vec![row_ids[20], row_ids[30]])
+        );
+        Ok(())
+    }
+
+    #[tokio::test]
     async fn test_fts_allowlist_runtime() -> Result<()> {
         let schema = Arc::new(ArrowSchema::new(vec![
             ArrowField::new("s", DataType::Utf8, true),
--- a/rust/lance/src/dataset/scanner.rs
+++ b/rust/lance/src/dataset/scanner.rs
@@ -1,6 +1,7 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-FileCopyrightText: Copyright The Lance Authors
 
+use std::collections::HashMap;
 use std::ops::Range;
 use std::pin::Pin;
 use std::sync::{Arc, LazyLock};
@@ -77,6 +78,7 @@ use tracing::{info_span, instrument, Span};
 
 use super::Dataset;
 use crate::dataset::row_offsets_to_row_addresses;
+use crate::dataset::rowids::load_row_id_sequence;
 use crate::dataset::utils::SchemaAdapter;
 use crate::index::vector::utils::{
     default_distance_type_for, get_vector_dim, get_vector_type, validate_distance_type_for,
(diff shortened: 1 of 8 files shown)


rust/lance/src/dataset/scanner.rs
```diff
@@ -812,6 +818,19 @@ impl Scanner {
         self
     }
 
+    /// Provide an allowlist of row ids (in dataset row-id space).
+    ///
+    /// For unstable row ids, the allowlist must come from the same snapshot.
+    /// For stable row ids, values remain valid across versions.
+    pub fn row_id_allowlist<I>(&mut self, row_ids: I) -> &mut Self
```
wjones127, line 825: What would you think of just taking an `RowAddrTreeMap` as input instead?

wjones127 (commented):
This is cool! Coincidentally I was just talking to some colleagues about needing something like that (https://github.com/lance-format/lance/issues/5832) so I'm all for it.

Have one initial question about the API now. I'll dig deeper into the implementation tomorrow morning.

rust/lance/src/dataset/scanner.rs
fredlarochelle: I tried going in with the least change possible. The way I see it, taking a `Vec<u64>` / slice / iterator as input is very ergonomic for callers and it matches how filters / `IN (...)` are expressed. It's a simple API surface.

I didn't think about `RowAddrTreeMap`. Preliminary thoughts: it's more efficient for callers that already have a map/mask and more explicit about row-address space, but it's less ergonomic for common callers (they must construct a `RowAddrTreeMap`) and it exposes more internal types in the API. On the plus side, it steers users toward the correct semantics.

For your use case in #5832, `RowAddrTreeMap` makes the most sense since they operate on row-address masks.

Also this is my first dive into Lance internals, I haven't explored in depth, but I'd imagine `row_id_allowlist` is easier to expose in Python bindings or higher-level code. I think it could be best to keep `row_id_allowlist` semantics and add a "power user" overload like `row_id_allowlist_map(RowAddrTreeMap)` or `row_id_allowlist_mask(RowAddrMask)`. That keeps ergonomics while allowing zero-copy use of pre-build masks.

rust/lance/src/dataset/scanner.rs
```diff
@@ -2386,6 +2449,66 @@ impl Scanner {
...
+                .entry(addr.fragment_id())
+                .or_default()
+                .push(addr.row_offset());
+        }
+
+        if addrs_by_frag.is_empty() {
+            return Ok(Vec::new());
+        }
+
+        let mut row_ids = Vec::new();
```
wjones127:
```suggestion
        let mut row_ids = Vec::with_capacity(row_addrs.len());
```

rust/lance/src/io/exec/knn.rs
```diff
@@ -296,6 +312,26 @@ impl ExecutionPlan for KNNVectorDistanceExec {
     }
 }
 
+fn filter_batch_by_allowlist(
+    batch: RecordBatch,
+    allowlist: &RowAddrMask,
+) -> DataFusionResult<RecordBatch> {
```
wjones127, line 318: This function is duplicated with the one in `fts.rs`, right?

rust/lance/src/io/exec/knn.rs
```diff
@@ -115,6 +116,7 @@ pub struct KNNVectorDistanceExec {
     pub query: ArrayRef,
     pub column: String,
     pub distance_type: DistanceType,
+    allowlist_mask: Option<Arc<RowAddrMask>>,
```
wjones127, line 119: This doesn't make much sense to me. I would think allowlist only needs to be in nodes that do IO. Why does this need to be in `KNNVectorDistanceExec` and `MatchQueryExec`? Can't we just push down into the scans and the index PreFilter?

rust/lance/src/io/exec/fts.rs
```diff
@@ -1199,6 +1264,84 @@ pub mod tests {
...
+        )
+        .unwrap();
+        let batch_keep = RecordBatch::try_new(
+            schema.clone(),
+            vec![
+                Arc::new(UInt64Array::from(vec![2_u64, 3])),
+                Arc::new(Float32Array::from(vec![0.2_f32, 0.3])),
+            ],
+        )
+        .unwrap();
```
wjones127: Not important, but for literal data like this, I'd recomend using the `arrow_array::record_batch!()` macro:


```suggestion
        let batch_empty = record_batch!(
            (ROW_ID, UInt64, [1]),
            ("score", Float32, [0.1])
        ).unwrap();
        let batch_keep = record_batch!(
            (ROW_ID, UInt64, [2, 3]),
            ("score", Float32, [0.2, 0.3])
        ).unwrap();
        let schema = batch_empty.schema();
```

wjones127 (commented):
This seems reasonable. My main concern is I'm not sure why it's necessary to modify the FTS and KNN exec nodes. Could you explain that?

rust/lance/src/io/exec/knn.rs
fredlarochelle: Yes, it is intentional for now. The two version differ only in error context ("KNN input" vs "FTS input"), and I kept it local to avoid extra refactoring in this PR. I'm happy to factor it into a shared helper if you prefer.

rust/lance/src/io/exec/knn.rs
fredlarochelle: Good question. I push the allowlist into IO (`FilteredReadExec`) when possible and into the index prefilter. For ANN, the prefilter is constructed in `ANNIvfSubIndexExec`, and for FTS it's built in `MatchQueryExec`/`PhraseQueryExec`, so the allowlist has to be injected there. For flat/unindexed or legacy paths, some scan paths (`scan_fragments` / `LanceScanExec`) don’t accept allowlists and some inputs aren’t IO‑backed, so `KNNVectorDistanceExec`/`FlatMatchQueryExec` enforce it as a correctness backstop. This also avoid doing distance/score work on rows that will be dropped.

wjones127: Okay. I'll look deeper into the code to understand better. I wonder if we somehow combine this code path with the ones that propagate deleted rows 🤔

fredlarochelle: Interesting thought, I will take a look tomorrow at the deleted rows path. Will report back.

PR #5834, opened 2026-01-28
fix: inconsistent transposed pq code and metadata when build ivf_pq index distributedly

yanghua:
There is a bug that happens when building the ivf_pq index distributedly.

That is when building partial ivf_pq, it transposes the pq code for some fragments. But in the final merge step, all the pq codes are not transposed, and the metadata is marked as `transposed`. So the logic of vectors search reads the inconsistent information.

Here the solution is:
* add a configuration about transposing;
* for the distributed vector build, the partial indices are not transposed;
* in the merge step, we finally do the trapsposing that means we transform all the PQ codes from row-based to column-based;

### DIFF
--- a/rust/lance-index/src/vector/distributed/index_merger.rs
+++ b/rust/lance-index/src/vector/distributed/index_merger.rs
@@ -6,10 +6,12 @@
 use crate::vector::shared::partition_merger::{
     write_unified_ivf_and_index_metadata, SupportedIvfIndexType,
 };
-use arrow::datatypes::Float32Type;
+use arrow::{compute::concat_batches, datatypes::Float32Type};
 use arrow_array::cast::AsArray;
-use arrow_array::{Array, FixedSizeListArray, UInt64Array};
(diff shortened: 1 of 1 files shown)


python/python/tests/test_vector_index.py
```diff
@@ -959,7 +959,7 @@ def test_pre_populated_ivf_centroids(dataset, tmp_path: Path):
             "metric_type": "l2",
             "nbits": 8,
             "num_sub_vectors": 8,
-            "transposed": True,
+            "transposed": False,
```
BubbleCal: it's expected to be `True`
for now, PQ and RQ always transposes the quantized codes for being cache friendly and SIMD optimizations
we don't transpose for only FLAT (no quantization) and SQ

yanghua:
I found that the recall performance is not expected. Will fix it.

yanghua:
After the refactor, I tested that the recall has been recovered.

PR #5836, opened 2026-01-28
fix: improve error messages in FixedSizeListArrayExt::convert_to_floating_point

LuciferYang:
## What

fix error messages in the `convert_to_floating_point` method of the `FixedSizeListArrayExt` trait.

## Why

Error messages previously always showed "Int8Type" regardless of the actual type being cast, which was misleading and made debugging difficult.

## Changes

- Update error messages to accurately reflect the actual type being attempted to cast


wjones127 (approved):

LuciferYang:
Thank you @wjones127

PR #5838, opened 2026-01-28
perf(ivf): reduce slow IVF v2 test runtime

fredlarochelle:
While working on PR #5831, I ran into a really slow test suite. Using nextest it was ~310s on a Framework Desktop (16x Zen 5, 128GB DDR5-8000, fast NVME). The worst offenders were a few IVF v2 split/join tests. This PR brings the suite down to ~82s on the same machine. I also included a small micro-opt in the IVF split/join path.

## Root cause

The slow tests were forcing split/join paths by appending 10k-50k rows because split/join threshold are tired to the default target partition size (8192). That led to very large data volume and expensive optimize cycles, even though the tests only need to validate behavior (split, join, merge) rather than scale.

## What this PR does

- Adds an optional `OptimizeOptions::target_partition_size_override` that let optimize (only) use a smaller split/join threshold.
- Uses that override + smaller dataset sizes in the four slow IVF v2 tests.
- Micro-opt in the split/join path:
  - Skip `concat_batches` when there's only one batch.
  - Skip `Flatten` unless the vector column is actually `List`.

## Tests updated

- `test_spfresh_join_split`
- `test_remap_join_on_second_delta`
- `test_partition_split_on_append_multivec`
- `test_join_partition_on_delete_multivec`
- plus `fast_optimize_indices` used in a couple of nearby helpers/tests

## Why it's safe

- Default behavior is unchanged unless the override is set.
- Tests still exercise the same split/join logic, just at a smaller scale.

If you’d like, I can follow up with a PR to speed up the remaining slow tests (see screenshot):

image-20260128113123934

### DIFF
--- a/rust/lance/src/index/vector/ivf/v2.rs
+++ b/rust/lance/src/index/vector/ivf/v2.rs
@@ -673,6 +673,7 @@ mod tests {
 
     const NUM_ROWS: usize = 512;
     const DIM: usize = 32;
+    const FAST_TEST_TARGET_PARTITION_SIZE: usize = 200;
 
     async fn generate_test_dataset<T: ArrowPrimitiveType>(
         test_uri: &str,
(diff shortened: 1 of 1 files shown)


Xuanwo:
Thank you for working on this but adding a new option to optimize slow tests doesn't look good to me, closing.

PR #5845, opened 2026-01-29
feat: add an arrow-scalar crate with a scalar definition

westonpace:
There are places in lance-file and lance-index where we pull in datafusion just so that we can have access to `ScalarValue` (we may use `Expr` in `lance-index` too so maybe not the best example).  I'd like to not need to pull in Datafusion to get access to scalars.

In addition, two different definitions for binary serialization of arrow scalars has popped up recently.  First, in the constant encoding and second in the column statistics.  Although in column statistics we might be able to avoid the need for scalar serialization and in the constant encoding we shouldn't need the full set of arrow types (but it wouldn't hurt).

It's enough I toyed around with making a small standalone lightweight crate to provide scalars.

If we use this then I'd also like to make a small statistics crate with a an accumulator that keeps track of min/max/nan count/null count/etc.  We could use this in some of the file encoding spots as well as some of the column statistics work (such an accumulator needs a definition for scalar because that is the min/max output)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -383,6 +383,19 @@ dependencies = [
 ]
 
 [[package]]
+name = "arrow-scalar"
+version = "3.0.0-beta.2"
+dependencies = [
+ "arrow-array",
+ "arrow-buffer",
+ "arrow-data",
+ "arrow-schema",
+ "arrow-select",
+ "half",
+ "rstest 0.23.0",
+]
+
+[[package]]
 name = "arrow-schema"
 version = "57.2.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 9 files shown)


westonpace:
Here I've used an enum approach (similar to `ScalarValue`) which uses primitive rust types for the simple scalars and `Buffer` for the string/binary scalars.  Using `Buffer` instead of `String` or `Vec<u8>` means it should be possible to zero-copy a scalar from an arrow array (which could be important if the array is full of images or videos)

There is another approach which could result in even less code.  Instead of an enum we could say that all scalars are simply `Arc<dyn Array>` where the array has length of 1.  The downside is that I think some of the display / ordering code would be more complex and creating primitive scalars would be slightly more heavyweight.  However, I think it is a viable approach as well if others think that might be better.

westonpace:
Leaving in draft as I still want to do more review (this is mostly vibe coded at the moment)

westonpace:
Superseded by #5955

PR #5850, opened 2026-01-29
feat: add LSM scanner with point lookup and vector search support

touch-of-grey:
## Summary

- LSM scanner for unified reads across base table and MemWAL regions
- Point lookup planner with bloom filter guards and short-circuit evaluation
- Vector search planner with staleness detection via bloom filters
- Benchmarks for scan, point lookup, and vector search operations

## Test plan

- [x] Unit tests for all planners and exec nodes (51 tests)
- [x] Clippy clean

### DIFF
--- a/docs/src/format/table/mem_wal.md
+++ b/docs/src/format/table/mem_wal.md
@@ -116,7 +116,7 @@ In other words, a WAL consists of an ordered list of WAL entries starting from p
 Writer must flush WAL entries in sequential order from lower to higher position.
 If WAL entry `N` is not flushed fully, WAL entry `N+1` must not exist in storage.
 
-### WAL Replay
+#### WAL Replay
 
 **Replaying** a WAL means to read data in the WAL from a lower to a higher position.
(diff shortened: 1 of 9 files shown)


jackye1995:
I have been thinking about this past 2 days, the part that we have to read each MemTable and then reverse the result feels just so inefficient to me. I think I have a good way to solve it now:

When we scan MemTable, everything is in memory, reverse scan is fine. So when we flush MemTable, we should read the whole BatchStore in reverse order. This means the indexes also need to reverse the row position mapping, so the new row position is `length_of_batch_store - current_position - 1`. By doing so, all the flushed MemTables are ordered from newest to oldest, not oldest to newest, so we can do the K-way merge much more efficiently. 

What do you think?

touch-of-grey:
Makes sense! Let me try update based on the current draft

jackye1995:
Sorry for the late review, I made some minor edits, and I think this is good to go now!

PR #5851, opened 2026-01-29
fix: inconsistency inline constant handling for non‑fixed‑stride datatype

zhangyue19921010:
## Encoding side (rust/lance-arrow/src/scalar.rs)

`try_inline_value` decides whether a scalar value can be written inline. The old logic only checked that:
- the scalar is non-null and a single value
- it has no child data
- there is only one buffer
- the inline size does not exceed the threshold

However, it did not verify that the data type is fixed-stride or not. **TAKE BOOLEAN AS EXAMPLE**

## Decoding side (decode_scalar_from_inline_value)
Decoding explicitly requires `data_type.byte_width_opt()` to be present; otherwise it errors out:
- `Inline constant is not supported for non-fixed-stride data type Boolean.`

**This creates an inconsistency between the encoder and decoder.**

Without this fix, new added unit test `test_constant_layout_out_of_line_boolean_v2_2` and similar scenarios will report errors

```
called `Result::unwrap()` on an `Err` value: Arrow { message: "Invalid argument error: 
Inline constant is not supported for non-fixed-stride data type Boolean", 
location: Location { file: "rust/lance-encoding/src/encodings/logical/primitive/constant.rs", line: 246, column: 59 } }
```

### DIFF
--- a/rust/lance-encoding/src/encodings/logical/primitive.rs
+++ b/rust/lance-encoding/src/encodings/logical/primitive.rs
@@ -6360,7 +6360,7 @@ mod tests {
         let versions: Vec<LanceFileVersion> = LanceFileVersion::iter()
             .filter(|version| {
                 let (major, minor) = version.to_numbers();
-                major >= 2 && minor >= 2;
+                major >= 2 && minor >= 2
             })
             .collect();
 


zhangyue19921010:
Hi @Xuanwo , would u mind to take a look at your convenient? Thanks!

rust/lance-arrow/src/scalar.rs
```diff
@@ -184,7 +176,12 @@ pub fn try_inline_value(scalar: &ArrayRef) -> Option<Vec<u8>> {
     if data.buffers().len() != 1 {
         return None;
     }
+
+    let byte_width = data.data_type().byte_width_opt();
     let bytes = data.buffers()[0].as_slice();
+    if byte_width.is_none() || bytes.len() != byte_width.unwrap() {
```
wjones127: I think you can do this without an `.unwrap()` here:
```suggestion
    if byte_width.map(|width| bytes.len() != width).unwrap_or(true) {
```

zhangyue19921010:
Thanks for your help @wjones127 changed.

rust/lance-encoding/src/encodings/logical/primitive.rs
```diff
@@ -6352,6 +6313,33 @@ mod tests {
...
+        };
+        let Layout::ConstantLayout(layout) = layout.layout.as_ref().unwrap() else {
+            panic!("Expected constant layout in slot 2");
+        };
+        assert!(layout.inline_value.is_none());
+        assert_eq!(page.data.len(), 1);
+
+        let test_cases = TestCases::default()
+            .with_min_file_version(LanceFileVersion::V2_2)
+            .with_max_file_version(LanceFileVersion::V2_2)
```
westonpace: Can we remove the `with_max_file_version`?  There is no reason this test should stop working in future versions.

westonpace (approved):
One minor nit and then we can merge.  Just ping me when you've addressed it.

zhangyue19921010:
This Bug is fixed by another PR  https://github.com/lance-format/lance/pull/5862 

Thanks for your attention and help here  @westonpace and @wjones127 ：）

PR #5858, opened 2026-01-29
chore: add CODEOWNERS for spec files

wjones127:
Idea is we want to make sure people are notified when someone makes a PR with a spec change. This reduces the likelihood a spec change will sneak through without a vote.

My plan is to create three GitHub "teams": `file-spec-team`, `table-spec-team` and `index-spec-team`. I can add a few relevant PMC members to each, and we can update this easily over time.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,14 @@
+# file specification
+protos/file*.proto @file-spec-team
+protos/encodings*.proto @file-spec-team
+docs/src/format/file/ @file-spec-team
+
+# table specification
+protos/table.proto @table-spec-team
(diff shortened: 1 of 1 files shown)


Xuanwo:
I love this idea. One thought I have is: should we merge the index-spec-team into the table-spec-team? I imagine that most index spec changes would still require support from the table level.

wjones127:
> One thought I have is: should we merge the index-spec-team into the table-spec-team? I imagine that most index spec changes would still require support from the table level.

So far, I've found in practice people who are focusing on indexes aren't the same people who focusing on table operations. If we find they end up being the same people in the future, we can always merge them.

PR #5863, opened 2026-01-30
feat: add plan_splits function for distributed compute

hamersaw:
Adding a `plan_splits` function to the Scanner to facilitate a single solution for partitioning Lance dataset for efficient distributed compute. This function (1) filters the dataset (using index looking / delete vectors) producing a mapping of fragment IDs to valid row ranges and (2) bin packs these fragment rows ranges into "splits" that target a configurable partition size (in rows count or bytes).

The most important areas to agree on:
- We explicitly disable refine filters when injecting the `FilteredReadExec`. This is because all refine filters are provided within the pre-computed plan.
- Within the `FilteredReadExec` we maintain the plan when `with_children` is called. The problem is that if datafusion attempts an optimization is can clear the plan that we explicitly set. Since this flow injects the `FilteredReadExec` as a leaf node, it should be OK to copy this over.

Example usage through python API:
```
import lance
from lance import Split
import pyarrow as pa
import shutil

# insert initial table and create index
dataset = lance.write_dataset(
    pa.Table.from_pylist([{"id": 1, "name": "Alice", "age": 20, "weight": 130.5},
                          {"id": 2, "name": "Bob", "age": 30, "weight": 180.0},
                          {"id": 3, "name": "David", "age": 42, "weight": 200.2}]),
    "memory://lance.test")

dataset.insert(
    pa.Table.from_pylist([{"id": 4, "name": "Ricky", "age": 22, "weight": 150.0},
                          {"id": 5, "name": "Carl", "age": 29, "weight": 120.3}],
    ))

dataset.create_scalar_index(
    column="age",
    index_type="BTREE"
)

# insert more data (unindexed)
dataset.insert(
    pa.Table.from_pylist([{"id": 6, "name": "Carla", "age": 37, "weight": 150.0},
                          {"id": 7, "name": "Eve", "age": 29, "weight": 120.3}],
    ))


scanner = dataset.scanner(columns=["weight", "_rowid", "name"], filter="age >= 30 AND weight <= 200.0")

# evaluate splits
splits = scanner.plan_splits(max_row_count=2)
for split in splits:
    # serialize and deserialize split
    split_bytes = split.to_bytes()
    new_split = Split.from_bytes(split_bytes, dataset._ds)

    # read split data
    scanner = dataset.scanner(columns=new_split.output_columns).with_filtered_read_exec(new_split.filtered_read_exec)
    reader = scanner.to_reader()

    table = reader.read_all()
    print(table.to_pydict())
```


rust/lance/src/dataset/scanner.rs
```diff
@@ -3995,6 +4062,283 @@ impl Scanner {
...
+                    .as_ref()
+                    .map(|frags| frags.iter().map(|f| vec![f.id as u32]).collect())
+                    .unwrap_or_else(|| {
+                        self.dataset
+                            .fragments()
+                            .iter()
+                            .map(|f| vec![f.id as u32])
+                            .collect()
+                    });
+                return Ok(Splits::Fragments(fragments));
```
jackye1995: this seems like a false promise that `plan_splits` is useful for FTS, vector search and take, which it is not. Do we really need this type of split, or should we just say it is unsupported?

rust/lance/src/dataset/scanner.rs
```diff
@@ -600,6 +602,39 @@ pub struct Scanner {
...
+    row_count: u64,
+}
+
+/// Options for configuring split generation.
+///
+/// This struct allows specifying constraints on the maximum size and row count
+/// for splits. Both fields are optional; if neither is set, default behavior
+/// will be used.
+#[derive(Debug, Clone, Default)]
+pub struct SplitOptions {
```
jackye1995: nit: this should be `SplitPlanningOptions`?

rust/lance/src/dataset/scanner.rs
```diff
@@ -3995,6 +4062,283 @@ impl Scanner {
...
+                            .iter()
+                            .map(|f| vec![f.id as u32])
+                            .collect()
+                    });
+                return Ok(Splits::Fragments(fragments));
+            }
+        };
+
+        // Compute the max rows per split
+        const DEFAULT_SPLIT_SIZE: usize = 128 * 1024 * 1024;
```
jackye1995: We should just let `SplitPlanningOptions` implement `Default`, instead of setting it inline.

rust/lance/src/dataset/scanner.rs
```diff
@@ -3995,6 +4062,283 @@ impl Scanner {
...
+        const DEFAULT_VARIABLE_FIELD_SIZE: usize = 64;
+
+        let options = options.unwrap_or_default();
+
+        // Helper to estimate row count from a byte size
+        let estimate_rows_from_bytes = |max_bytes: usize| -> Result<usize> {
```
jackye1995: nit: I think we should make this a dedicated helper function instead of defining it inline

rust/lance/src/dataset/scanner.rs
```diff
@@ -3995,6 +4062,283 @@ impl Scanner {
...
+                let rows_from_bytes = estimate_rows_from_bytes(max_bytes)?;
+                max_rows.min(rows_from_bytes)
+            }
+            (Some(max_rows), None) => max_rows,
+            (None, Some(max_bytes)) => estimate_rows_from_bytes(max_bytes)?,
+            (None, None) => estimate_rows_from_bytes(DEFAULT_SPLIT_SIZE)?,
+        };
+
+        // Create bin items where each item has <= max_rows_per_split rows
+        let mut bin_items: Vec<BinItem> = Vec::new();
```
jackye1995, line 4517: This bin-packing looks to be first-fit decreasing. We might want to consider head-tail bin-packing, since that achieves similar effect but take less time and is more memory efficient.

Btw, we should probably have a new module for these logic. this module is getting a bit too long, and I think it makes sense to have a dedicated module for the bin-packing logic so we can have more focused tests.

java/src/main/java/org/lance/ipc/FilteredReadPlan.java
```diff
@@ -0,0 +1,96 @@
...
+/**
+ * A filtered read plan represents a unit of work for scanning a dataset.
+ *
+ * <p>It specifies which fragment row ranges to read, any residual filters to apply (encoded as
+ * Substrait), and an optional post-filter scan range. Plans are created by {@link
+ * LanceScanner#planSplits} and executed by {@link LanceScanner#executeFilteredReadPlan}.
+ *
+ * <p>This class is a pure data object that is fully serializable for distributed execution.
+ */
+public class FilteredReadPlan {
```
jackye1995: it feels unnecessary to me that we have a fully materialized Java object for FilteredReadPlan (similarly for python). We mainly want to get the list of FilteredReadPlans from `plan_splits`, and then somehow serialize it, pass it to a worker, and then deserialize it, and feed into the executeFilteredReadPlan.

Would it be possible to actually return a version of FilteredReadPlan that can be more efficiently serialized and deserialized (e.g. through protobuf definition)? This would allow us to also not expose the current FilteredReadInternalPlan to public API.

rust/lance/src/dataset/scanner.rs
hamersaw: Thanks for the clarification, I was under the assumption that if we filter with FTS / vector search it may still be useful -- and that support is something that we could support in the future. I'll update this to return some variant of "Unsupported" and that will fallback to full fragments on the Spark side.

rust/lance/src/dataset/scanner.rs
hamersaw: > we should probably have a new module for these logic.

I toyed with the idea of having a "split" module that handled all of this logic (ie. row size estimation / bin packing / etc). Is that what you're thinking? Happy to implement head-tail bin-packing, I was under the assumption that this problem was small enough that optimizations would be pretty moot.

java/src/main/java/org/lance/ipc/FilteredReadPlan.java
hamersaw: I think that's a great idea, but I'm having some trouble understanding the integration points. Are you thinking the `plan_splits` functoin returns a protobuf message and the `execute_filtered_read_plan` (could change to `execute_spit`) then takes that same protobuf message? I have never ran into pushing protobuf messages through JNI / etc so would have to learn a bit about how that looks.

The serialization thing that I think is important to figure out it optimizing the storage format for the row_id filter. I commented in this PR about sparse rows would benefit from bitmap and dense would be smaller represented by ranges. It's probably something we can figure out in defining a protobuf definition as long as we agree on e2e plan_splits -> JNI -> spark driver -> spark worker -> JNI -> execute_filtered_read format conversions.

rust/lance/src/dataset/scanner.rs
jackye1995: > I toyed with the idea of having a "split" module that handled all of this logic (ie. row size estimation / bin packing / etc). Is that what you're thinking?

+1!

java/src/main/java/org/lance/ipc/FilteredReadPlan.java
jackye1995: > I'm having some trouble understanding the integration points. Are you thinking the plan_splits functoin returns a protobuf message and the execute_filtered_read_plan (could change to execute_spit) then takes that same protobuf message?

Yes. I think this could potentially avoid multiple layers of serialization and deserialization. Java object can just be a handle that points to the split result, and when serializing for work distribution, we can request for serialized protobuf bytes. 

We can also leverage some existing encoding mechanisms like https://github.com/lance-format/lance/blob/main/protos/rowids.proto#L19 to represent it efficiently.

rust/lance/src/dataset/scanner.rs
jackye1995: I guess that's a potential question for @LuQQiu , how do we envision this to be used for example vector search?

rust/lance/src/io/exec/filtered_read.rs
```diff
@@ -1461,61 +1461,29 @@ pub struct FilteredReadExec {
...
+/// [`FilteredReadExec::with_plan`].
+#[derive(Clone, Debug)]
 pub struct FilteredReadPlan {
-    /// What fragments and physical rows to read
-    pub rows: RowAddrTreeMap,
-    /// Filter to apply per fragment
-    /// fragments not here don't need filtering
+    /// Fragment ID to row ranges to read (BTreeMap for deterministic order
+    /// with scan_range_after_filter).
+    pub rows: BTreeMap<u32, Vec<Range<u64>>>,
```
LuQQiu: I keep FilteredReadPlan with external plan using RoaringBitmap for the following reasons
* Range is ambiguous in Lance. Most user-facing APIs use logical rows (without deletions). Vec<Range<u64>> here means physical rows (with deletions). This is confusing. RowAddrTreeMap/RoaringBitmap avoids that ambiguity because it's clearly a set of physical row addresses, not a "range" that could be misread as logical.
* Ranges fragment badly with deletions. A dataset with scattered deletions turns a single range 0..4096 into many small ranges. The Vec<Range<u64>> grows proportionally to the number of deletion holes. RoaringBitmap handles this natively with compressed containers.
* Bitmap enables efficient row unselection. Whether it's 4K row group pruning or custom index results, you need to remove specific rows from the selection. Flipping bits is O(1); splitting ranges is O(n) and produces more ranges.

LuQQiu: This is not just for FilteredReadExec, we will also use RowIdTreeMap as input variables for TakeExec and other Lance read execs. This was discussed previously to remove the ambiguous in Lance Execs and enabled for different kinds of row selection and executions.

rust/lance/src/dataset/scanner.rs
LuQQiu: plan_splits could be useful for brute force FTS, vector search (without index ones).
I am thinking of KNNScan<FilteredReadExec>, the split could be both KNNScan and FilteredReadExec. 
brute force FTS can be something similar, but i am not quite aware of our brute force FTS impl.

LuQQiu: KNNScan<FilteredReadExec> KNN Push down to I/O and to executor could benefit from no need to transfer a large amount of data across executors, if one set of executors read data and one set of executors does the brute force vector search

rust/lance/src/dataset/scanner.rs
hamersaw: I think the semantics of this make `Default` not work exactly how I envisioned it. Basically, the user can set `max_size_bytes` and / or `max_row_count` (if both then the min is used). However, if the user does not set anything then we fallback to a default. If we use `Default` to set this, and the user only wants to filter on maximum number of rows they will need to unset the `max_size_bytes`.

hamersaw: We could update this to some kind of `build` logic? `splitting_options.with_max_rows(N).build()` which allows us to check if neither are set and default there? I'm not convinced this is any clean / clear than the current logic because we still need a fallback in the `plan_splits` code if neither are set (default value or failure).

rust/lance/src/io/exec/filtered_read.rs
hamersaw: Thanks, reverted to the existing implementation per our discussion.

rust/lance/src/dataset/scanner.rs
hamersaw: Given this information, I think it's reasonable to keep `Split` as an `enum` to give us the ability to add more types in the future. For now, if it's not a simple filter, we can use a `Fragment` (existing Spark implementation) and we can build more efficient logic in the future.

hamersaw: On second though, I just removed the enum. `Split` is a `FilteredReadExec` -- we can revisit if we want other optimizations in the future.

java/src/main/java/org/lance/ipc/FilteredReadPlan.java
hamersaw: I made all python / java bindings be native object handles and then exposed serialization and deserialization. So we should be able to integrate this into distributed compute engines without egregious costs.

rust/lance/src/dataset/scanner.rs
```diff
@@ -4417,6 +4481,145 @@ impl Scanner {
...
+                    // get_or_create_plan always returns Partial, but handle Full defensively
+                    let fragment =
+                        self.dataset
+                            .get_fragment(*frag_id as usize)
+                            .ok_or_else(|| Error::InvalidInput {
+                                source: format!("Fragment {} not found", frag_id).into(),
+                                location: location!(),
+                            })?;
+                    let num_rows = fragment.physical_rows().await? as u64;
+                    let bm = RoaringBitmap::from_sorted_iter(0..num_rows as u32).unwrap();
```
LuciferYang, line 4533: should use `?` instead of `.unwrap()`.

python/python/lance/dataset.py
```diff
@@ -5251,6 +5253,73 @@ def analyze_plan(self) -> str:
...
+            Maximum number of rows per split. If both max_size_bytes and
+            max_row_count are specified, the smaller constraint wins.
+
+        Returns
+        -------
+        list[Split]
+            A list of :class:`Split` objects, where each split has either:
+            - ``filtered_read_exec``: An opaque :class:`FilteredReadExec`
+              object to pass to :meth:`with_filtered_read_exec`.
+            - ``fragments``: A list of fragment IDs.
```
LuciferYang, line 5285: can remove `fragments`?

PR #5867, opened 2026-01-31
fix: handle NULL elements in LABEL_LIST index results and explain_plan

### LINKED ISSUE #5682: LABEL_LIST index returns incorrect results when list has NULL elements
### Description
When a list column contains NULL elements, the LABEL_LIST index can miss rows that contain a valid value.

  For example, with `labels = [["foo", None], ["foo"]]`, a scan for `array_has_any(labels, ['foo'])` returns both rows, but after creating a LABEL_LIST index it only returns `[['foo']]` (the `['foo', None]` row is missing).

  Additionally, `explain_plan()` can panic for `array_has_any(labels, [NULL])` and `array_has_all(labels, [NULL])`.

### How to repeat
```python
import tempfile
from pathlib import Path

import lance
import pyarrow as pa

# 1. Setup Data
tmp = Path(tempfile.mkdtemp(prefix="lance-label-list-nullable-"))

tbl = pa.table({"labels": [["foo", None], ["foo"]]})
ds = lance.write_dataset(tbl, tmp / "ds")

FILTERS = [
    "array_has_any(labels, ['foo'])",
    "array_has_all(labels, ['foo'])",
]

# 2. Results before index (Scan)
pre = {f: ds.to_table(filter=f).num_rows for f in FILTERS}
pre_rows = {f: ds.to_table(filter=f).column("labels").to_pylist() for f in FILTERS}

# 3. Create Index
ds.create_scalar_index("labels", index_type="LABEL_LIST")

# 4. Results after index (Scalar Index Search)
post = {f: ds.to_table(filter=f).num_rows for f in FILTERS}
post_rows = {f: ds.to_table(filter=f).column("labels").to_pylist() for f in FILTERS}

# pre create index and post create index result inconsistent
mismatches = {f: (pre[f], post[f]) for f in FILTERS if pre[f] != post[f]}
for f, (a, b) in mismatches.items():
    print(f"\nfilter: {f}")
    print("pre create index result:", pre_rows[f])
    print("post create index result:", post_rows[f])

# panic
ds.scanner(filter="array_has_any(labels, [NULL])").explain_plan()

```
### Result
```shell
filter: array_has_any(labels, ['foo'])
pre create index result: [['foo', None], ['foo']]
post create index result: [['foo']]

filter: array_has_all(labels, ['foo'])
pre create index result: [['foo', None], ['foo']]
post create index result: [['foo']]

thread 'lance_background_thread' panicked at .../rust/lance-index/src/scalar.rs:577:18:
called `Result::unwrap()` on an `Err` value: InvalidArgumentError("Non-nullable field of ListArray \"item\" cannot contain nulls")
stack backtrace:
   ...
  File "xxx.py", line 31, in <module>
    ds.scanner(filter="array_has_any(labels, [NULL])").explain_plan()
  File ".../python/lance/dataset.py", line 5115, in explain_plan
    return self._scanner.explain_plan(verbose=verbose)
RuntimeError: Task was aborted
```

fenfeng9:
closes #5682

changes:
- Treat element-level NULLs in LABEL_LIST as non-matches so array_has_any/array_has_all return TRUE/FALSE when the list itself is non-NULL.
- Allow nullable list literals in `LabelListQuery::to_expr` to prevent `explain_plan()` panics.
- Add Python tests covering element-level NULLs, list-level NULLs, NULL-literal filters and explain behavior.

### DIFF
--- a/python/python/tests/test_scalar_index.py
+++ b/python/python/tests/test_scalar_index.py
@@ -2055,9 +2055,9 @@ def test_label_list_index_array_contains(tmp_path: Path):
 
 
 def test_label_list_index_null_element_match(tmp_path: Path):
-    """Ensure LABEL_LIST index keeps scan semantics when lists contain NULLs."""
+    """Covers NULL elements inside non-NULL lists (list itself is never NULL)."""
     tbl = pa.table(
-        {"labels": [["foo", None], ["foo"], ["bar", None], ["bar"], None, []]}
+        {"labels": [["foo", None], ["foo"], ["bar", None], [None], ["bar"], []]}
     )
     dataset = lance.write_dataset(tbl, tmp_path / "dataset")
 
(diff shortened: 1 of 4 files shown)


fenfeng9:
PTAL @westonpace .

python/python/tests/test_scalar_index.py
```diff
@@ -2039,6 +2039,44 @@ def test_label_list_index_array_contains(tmp_path: Path):
...
 
+def test_label_list_index_null_element_match(tmp_path: Path):
+    """Ensure LABEL_LIST index keeps scan semantics when lists contain NULLs."""
+    tbl = pa.table({"labels": [["foo", None], ["foo"], None]})
+    dataset = lance.write_dataset(tbl, tmp_path / "dataset")
+
+    filters = [
+        "array_has_any(labels, ['foo'])",
+        "array_has_all(labels, ['foo'])",
+        "array_contains(labels, 'foo')",
```
wjones127, line 2067: Nice. Could we also test negation? That's really important for proper null handling.


```suggestion
        "array_has_any(labels, ['foo'])",
        "array_has_all(labels, ['foo'])",
        "array_contains(labels, 'foo')",
        "NOT array_has_any(labels, ['foo'])",
        "NOT array_has_all(labels, ['foo'])",
        "NOT array_contains(labels, 'foo')",
```

fenfeng9:
Negation is currently broken for NULL lists in the index path.
`LabelListIndex` flattens `List<T>` into scalar rows for `BitmapIndex`. `unnest_batch` drops rows where the list itself is `NULL` (0 indices), so those rows never enter the bitmap.
 
As a result, we can’t distinguish “list is NULL” from “list doesn’t contain the value”, and NULL is treated as FALSE; NOT then incorrectly returns it.

For the test case `[["foo", None], ["foo"], None]`:
* Row 0: TRUE (contains "foo") → NOT → FALSE ✓
* Row 1: TRUE (contains "foo") → NOT → FALSE ✓
* Row 2: NULL (list itself is NULL) → NOT → should be NULL (filtered out)

But since Row 2 is missing from the index, it's treated as FALSE → NOT becomes TRUE, causing the negation query to incorrectly return it.

python/python/tests/test_scalar_index.py
```diff
@@ -2054,6 +2054,47 @@ def test_label_list_index_array_contains(tmp_path: Path):
     assert "ScalarIndexQuery" not in explain
 
 
+def test_label_list_index_null_element_match(tmp_path: Path):
+    """Ensure LABEL_LIST index keeps scan semantics when lists contain NULLs."""
+    tbl = pa.table({"labels": [["foo", None], ["foo"], None]})
```
wjones127: We should add a case where there are nulls but it shouldn't contain "foo".
```suggestion
    tbl = pa.table({"labels": [["foo", None], ["foo"], ["bar", None], None]})
```

python/python/tests/test_scalar_index.py
```diff
@@ -2054,6 +2054,47 @@ def test_label_list_index_array_contains(tmp_path: Path):
...
+    dataset = lance.write_dataset(tbl, tmp_path / "dataset")
+
+    filters = [
+        "array_has_any(labels, ['foo'])",
+        "array_has_all(labels, ['foo'])",
+        "array_contains(labels, 'foo')",
+        "NOT array_has_any(labels, ['foo'])",
+        "NOT array_has_all(labels, ['foo'])",
+        "NOT array_contains(labels, 'foo')",
```
wjones127: If you want to address these in a different issue / PR, feel free to comment out the failing one and add a comment with a link to a follow up issue.

fenfeng9: Thanks for the suggestion! I'll create a follow-up issue and submit a PR to address these separately.

python/python/tests/test_scalar_index.py
fenfeng9: Updated.

fenfeng9:
There are two distinct issues.

1) **Element-level NULLs** (NULL items inside a non-NULL list): `array_has_any` / `array_has_all` should ignore NULL elements, so results are strictly TRUE/FALSE (no NULL propagation).

2) **List-level NULLs** (the list itself is NULL): this still affects NOT semantics and is tracked separately in #5904.

This PR fixes (1) by clearing element-level NULLs in the LABEL_LIST index path, and splits the unit tests to cover element-level vs list-level NULLs.

Examples:
| Expression                                  | Result | Reason                                            |
| :------------------------------------------ | :----- | :------------------------------------------------ |
| `array_has_any(["foo", NULL], ["foo"])`     | TRUE   | Found "foo"; NULL elements are ignored            |
| `NOT array_has_any(["bar", NULL], ["foo"])` | TRUE   | No match; NULL ignored → FALSE → NOT FALSE = TRUE |

PR #5873, opened 2026-02-02
feat: adding support for opening datasets using tagged versions through JNI

hamersaw:
Allows setting `tag` on JNI `ReadOptions` and using this to correctly open a dataset.

### DIFF
--- a/java/src/main/java/org/lance/ReadOptions.java
+++ b/java/src/main/java/org/lance/ReadOptions.java
@@ -104,28 +104,26 @@ public class ReadOptions {
     private Optional<StorageOptionsProvider> storageOptionsProvider = Optional.empty();
 
     /**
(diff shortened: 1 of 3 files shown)


java/src/main/java/org/lance/ReadOptions.java
```diff
@@ -88,6 +95,7 @@ public String toString() {
   public static class Builder {
 
     private Optional<Integer> version = Optional.empty();
+    private Optional<String> tag = Optional.empty();
```
majin1102, line 98: Hi Dan, thanks for working on this.

Wondering if we could leverage Ref.java here to enable a complete dataset-opening experience that supports versions, tags, and branches.

majin1102 (commented):
Thanks for working on this. Left one comment. PTAL

PR #5874, opened 2026-02-02
feat: add rename table implementations to REST namespaces

bryanck:
This PR adds the Python and Java bindings for rename table using the REST namespace. This is already implemented in Rust.


bryanck:
I had to comment out the test assertions and added a TODO comment, because the REST adapter relies on the directory namespace implementation.

I considered a couple of options, one was to implement rename in the directory namespace. The second was to create a new in-memory namespace implementation and use that with the REST adapter instead. Both of those options were fairly significant changes so I felt we could follow up with that in another PR.

@jackye1995 do you have thoughts on that?

PR #5875, opened 2026-02-02
test: add validation tests for full-text indexes

### LINKED ISSUE #4934: Query tests: FTS
Expand query integration tests to cover FTS queries

vivek-bharathan:
Closes #4934

### DIFF
--- a/rust/lance/tests/query/inverted.rs
+++ b/rust/lance/tests/query/inverted.rs
@@ -172,6 +172,16 @@ async fn test_inverted_with_filter() {
     .unwrap();
 
     DatasetTestCases::from_data(batch.clone())
+        .with_index_types(
+            "category",
+            [
+                None,
+                Some(IndexType::Bitmap),
+                Some(IndexType::BTree),
+                Some(IndexType::BloomFilter),
+                Some(IndexType::ZoneMap),
+            ],
+        )
         .run(|ds, original| async move {
             let mut ds = ds;
             let params = base_inverted_params(false);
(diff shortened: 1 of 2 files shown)


rust/lance/tests/query/inverted.rs
```diff
@@ -0,0 +1,263 @@
...
+
+            assert_fts_expected(&original, &ds, query, None, &[0, 3, 4]).await;
+            test_fts(&original, &ds, "text", "lance database", None, true, true).await;
+        })
+        .await;
+}
+
+#[tokio::test]
+// Validate filters are applied alongside inverted index search results.
+async fn test_inverted_with_filter() {
```
wjones127, line 149: I like the idea of testing with other filters on top. It would be good to add variants where you also have scalar indices on the filter columns. You can see how we do this with vector search here:

https://github.com/lance-format/lance/blob/da3e5dc3e5e77a2dccfe7c439561f6b89cb08c3e/rust/lance/tests/query/vectors.rs#L36-L43

rust/lance/tests/query/inverted.rs
```diff
@@ -0,0 +1,263 @@
...
+        Some("Saying HELLO loudly in a long sentence for testing."),
+        None,
+    ];
+    let text = Arc::new(StringArray::from(text_values)) as ArrayRef;
+    let batch = RecordBatch::try_from_iter(vec![("id", ids as ArrayRef), ("text", text)]).unwrap();
+
+    let cases = vec![
+        (
+            "simple_lc_pos",
+            params_for("simple", true, true),
```
wjones127, line 227: I wonder if you can add a method like `with_index_types_and_params()` to the `DatasetTestCases` builder so you can do transform this into something like:

https://github.com/lance-format/lance/blob/da3e5dc3e5e77a2dccfe7c439561f6b89cb08c3e/rust/lance/tests/query/vectors.rs#L36-L43

One of the reason we put that info in the builder is that we print out this nice context when one of the tests fails: https://github.com/lance-format/lance/blob/da3e5dc3e5e77a2dccfe7c439561f6b89cb08c3e/rust/lance/tests/utils/mod.rs#L115-L124

If you don't, then it's hard to see from the failures which of these cases failed.

vivek-bharathan: I think I've addressed what you meant - please validate

PR #5876, opened 2026-02-03
fix(java): init allocator for new dataset when checkout branch/tag

fangbo:
If the new dataset's allocator does not init. The operation on the new dataset will fail.

@majin1102 Could you please review it. Thank you .

### DIFF
--- a/java/src/test/java/org/lance/DatasetTest.java
+++ b/java/src/test/java/org/lance/DatasetTest.java
@@ -257,6 +257,7 @@ public class DatasetTest {
 
         // checkout the dataset at version 1
         try (Dataset checkoutV1 = dataset2.checkoutVersion(1)) {
+          assertNotNull(checkoutV1.getSchema());
           assertEquals(1, checkoutV1.version());
           assertEquals(2, checkoutV1.latestVersion());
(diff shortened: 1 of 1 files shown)


java/src/main/java/org/lance/Dataset.java
```diff
@@ -1380,7 +1392,13 @@ public Dataset checkout(Ref ref) {
     Preconditions.checkNotNull(ref);
     try (LockManager.ReadLock readLock = lockManager.acquireReadLock()) {
       Preconditions.checkArgument(nativeDatasetHandle != 0, "Dataset is closed");
-      return nativeCheckout(ref);
+      Dataset newDataset = nativeCheckout(ref);
```
majin1102, line 1395: Maybe we could add some tests to verify operational after checking out branches?

majin1102 (commented):
Thanks for working on this.
Left one comment.

java/src/main/java/org/lance/Dataset.java
fangbo: Add  getSchema invoking on new dataset in test cases.

majin1102 (approved):
looks good to me

PR #5881, opened 2026-02-03
fix: deduplicate row addresses in take to prevent panic

### LINKED ISSUE #5260: Subtract with overflow panic while decoding binary data
Getting subtract with overflow on this line:

https://github.com/lancedb/lance/blob/26e3e9ca85a2f066ad68b46d9567fa8b9b128018/rust/lance-encoding/src/previous/encodings/physical/binary.rs#L71

```
thread 'tokio-runtime-worker' panicked at /Users/runner/.cargo/git/checkouts/lance-8cdf2ec69b3663b6/254a821/rust/lance-encoding/src/previous/encodings/physical/binary.rs:71:24:
attempt to subtract with overflow
stack backtrace:
   0: __rustc::rust_begin_unwind
             at /rustc/1159e78c4747b02ef996e55082b704c09b970588/library/std/src/panicking.rs:697:5
   1: core::panicking::panic_fmt
             at /rustc/1159e78c4747b02ef996e55082b704c09b970588/library/core/src/panicking.rs:75:14
   2: core::panicking::panic_const::panic_const_sub_overflow
             at /rustc/1159e78c4747b02ef996e55082b704c09b970588/library/core/src/panicking.rs:175:17
   3: lance_encoding::previous::encodings::physical::binary::IndicesNormalizer::extend
   4: <lance_encoding::previous::encodings::physical::binary::BinaryPageScheduler as lance_encoding::decoder::PageScheduler>::schedule_ranges::{{closure}}
...
```

From the following CI job:

https://github.com/lancedb/lancedb/actions/runs/19446831001/job/55643266826?pr=2782

wjones127:
Previously, duplicate row addresses (e.g. from FTS on `List<Utf8>` where
multiple list elements in the same row match) were passed to the v2 encoding
layer, which requires strictly increasing indices. This caused
`indices_to_ranges` to produce overlapping ranges, panicking with "attempt to
subtract with overflow".

Dedup is handled at two sites:

- **`FragmentReader::take_as_batch`**: This is the common entry point for all
  callers that need a single batch (`FileFragment::take_rows`,
  `FragmentSession::take_rows`, `TakeStream::map_batch`). Since it already
  collects into one batch, the expand-after-dedup step is trivial. Handling it
  here rather than in the v2 adapter's `take_all_tasks` avoids complications
  with the streaming `ReadBatchTaskStream` return type. Handling it lower in
  `schedule_take` wouldn't work because the encoding layer has no notion of
  duplicates by design.

- **`TakeStream::map_batch`**: Dedup here serves a different purpose — it
  reduces I/O by avoiding redundant reads of the same row across fragments.
  This also optimizes the common path (sorted, unique addresses) to skip all
  sorting/dedup/permutation overhead entirely.

Fixes #5260

## Test plan

- `test_take_with_duplicate_row_addrs` — sorted duplicate row addresses
- `test_take_with_unsorted_duplicate_row_addrs` — unsorted duplicates
- `test_fragment_take_indices` / `test_fragment_take_rows` — fragment-level
  duplicate indices (previously failing on `LanceFileVersion::Stable`)
- All 57 take-related tests pass

### DIFF
--- a/rust/lance/src/dataset/fragment.rs
+++ b/rust/lance/src/dataset/fragment.rs
@@ -2616,18 +2616,43 @@ impl FragmentReader {
 
     /// Take rows from this fragment, will perform a copy if the underlying reader returns multiple
     /// batches.  May return an error if the taken rows do not fit into a single batch.
+    ///
+    /// Duplicate indices are allowed and will produce duplicate rows in the output.
     pub async fn take_as_batch(
         &self,
         indices: &[u32],
         take_priority: Option<u32>,
     ) -> Result<RecordBatch> {
+        // The v2 encoding layer requires strictly increasing indices. Deduplicate
+        // here so callers (e.g. FTS with duplicate row matches) don't need to.
+        let has_duplicates = indices.windows(2).any(|w| w[0] == w[1]);
+        let (unique_indices, expand_map) = if has_duplicates {
+            let mut unique: Vec<u32> = Vec::with_capacity(indices.len());
+            let mut mapping: Vec<u32> = Vec::with_capacity(indices.len());
+            for &idx in indices {
+                if unique.last() != Some(&idx) {
+                    unique.push(idx);
+                }
+                mapping.push((unique.len() - 1) as u32);
+            }
+            (Cow::Owned(unique), Some(UInt32Array::from(mapping)))
+        } else {
+            (Cow::Borrowed(indices), None)
+        };
+
         let batches = self
-            .take(indices, u32::MAX, take_priority)
+            .take(&unique_indices, u32::MAX, take_priority)
             .await?
             .buffered(get_num_compute_intensive_cpus())
             .try_collect::<Vec<_>>()
             .await?;
-        concat_batches(&Arc::new(self.output_schema.clone()), batches.iter()).map_err(Error::from)
+        let mut batch = concat_batches(&Arc::new(self.output_schema.clone()), batches.iter())?;
+
+        if let Some(expand_map) = expand_map {
+            batch = arrow_select::take::take_record_batch(&batch, &expand_map)?;
+        }
+
+        Ok(batch)
     }
 }
 


rust/lance-encoding/src/decoder.rs
```diff
@@ -1302,7 +1302,7 @@ impl DecodeBatchScheduler {
         sink: mpsc::UnboundedSender<Result<DecoderMessage>>,
         scheduler: Arc<dyn EncodingsIo>,
     ) {
-        debug_assert!(indices.windows(2).all(|w| w[0] <= w[1]));
+        debug_assert!(indices.windows(2).all(|w| w[0] < w[1]));
```
wjones127: We don't allow duplicates, so changing this to assert that.

jackye1995 (approved):
Looks good to me

rust/lance/src/dataset/fragment.rs
```diff
@@ -2616,18 +2616,43 @@ impl FragmentReader {
...
             .await?
             .buffered(get_num_compute_intensive_cpus())
             .try_collect::<Vec<_>>()
             .await?;
-        concat_batches(&Arc::new(self.output_schema.clone()), batches.iter()).map_err(Error::from)
+        let mut batch = concat_batches(&Arc::new(self.output_schema.clone()), batches.iter())?;
+
+        if let Some(expand_map) = expand_map {
+            batch = arrow_select::take::take_record_batch(&batch, &expand_map)?;
+        }
```
westonpace, line 2653: Someday it would be nice to coalesce the dataset paths and the TakeExec paths so that we don't have to duplicate this logic everywhere.  Not a comment on this PR, just griping.

wjones127:
> This PR is fine to avoid panic but does this mean the search will still output duplicate rows? Is that something we want long term?

Yeah I'll make a follow up issue. Not sure what the intended semantics of FTS in a list are.

PR #5888, opened 2026-02-04
docs: add FTS docs

prrao87:
We lack docs for how to create and query an FTS index in Lance. This PR adds detailed docs, based on the latest v1.0.4 `pylance` API. All code snippets have been tested with the latest stable release and I gave it another pass with GPT 5.2 to ensure accuracy per the latest Rust code in [tokenizer.rs](https://github.com/lance-format/lance/blob/f77769778aa93f47572187b83ccd7b6638dc39a3/rust/lance-index/src/scalar/inverted/tokenizer.rs#L180).

### DIFF
--- a/docs/src/quickstart/full-text-search.md
+++ b/docs/src/quickstart/full-text-search.md
@@ -5,7 +5,7 @@ description: Full-text search (FTS) with inverted BM25 indexes and N-gram search
 
 # Full-Text Search in Lance
 
-Lance provides powerful full-text search (FTS) capabilities using an inverted index. This tutorial will guide you through building and using FTS indexes to dramatically speed up text search operations while maintaining high accuracy.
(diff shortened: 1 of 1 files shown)


docs/src/quickstart/full-text-search.md
```diff
@@ -0,0 +1,350 @@
...
+```python
+table = ds.to_table(
+    full_text_query="train",
+    fast_search=True  # Only search indexed data for faster results
+)
+```
+
+### Index Maintenance
+
+As your data evolves over time, it's important to periodically rebuild your indexes to maintain optimal search performance. When you add, modify, or delete documents, the existing index may become less efficient. Consider scheduling regular index rebuilds, especially for datasets with frequent updates.
```
prrao87: How does one update the existing FTS index in Lance? @BubbleCal

prrao87:
Good point about fuzzy search! Will add.

prrao87:
Give me till end of day to finalize the fuzzy search doc and will add that in. Thanks for reviewing!

Issue #5141, opened 2025-11-04
Fuzz testing for the file format


wjones127:
We should ensure that we can attempt to read any Lance data file (even corrupted ones) without panicking.

Create a seed corpus:

* Cover all files versions
* Cover all encodings

Consider using [honggfuzz-rs](https://github.com/rust-fuzz/honggfuzz-rs). If possible, it would be helpful to configure "feedback-driven fuzzing", which uses code-coverage to guide the mutation process.

Run this test in CI.

wjones127:
Can we do this in a whitebox way where we take advantage of protobuf structure?

PR #5890, opened 2026-02-05
fix: avoid panic when repdef serializes empty offsets

fenfeng9:
Fixes panic on empty list offsets (`[0]`).

The assertion `expected_len - 1` underflows when `expected_len == 0`. We early-return because zero values produce no rep/def levels.

### DIFF
--- a/rust/lance-encoding/src/repdef.rs
+++ b/rust/lance-encoding/src/repdef.rs
@@ -533,7 +533,13 @@ impl SerializerContext {
         // are reading.
 
         let mut new_len = 0;
-        assert!(self.rep_levels.len() >= (offset_desc.num_values + self.current_num_specials) - 1);
+        let expected_len = offset_desc.num_values + self.current_num_specials;
+        if expected_len == 0 {
+            // Offsets [0] mean no list values, so no levels.
+            self.current_len = 0;
+            return;
+        }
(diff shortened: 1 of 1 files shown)


fenfeng9:
PTAL @westonpace .

fenfeng9:
> This is a good fix for repdef, thanks! Did you run into this scenario with actual data? Or is it theoretical? I wonder if we had checks for empty arrays higher in the code somewhere?


You're right that `FileWriter::write_batch` does check for empty batches and returns early 
https://github.com/lance-format/lance/blob/556851b39f1f8ea583dd50f165b69438feb2eebf/rust/lance-file/src/writer.rs#L385-L397

The panic I encountered was triggered through the `encode_batch` function
https://github.com/lance-format/lance/blob/556851b39f1f8ea583dd50f165b69438feb2eebf/rust/lance-encoding/src/encoder.rs#L689-L698

fenfeng9:
> This is a good fix for repdef, thanks! Did you run into this scenario with actual data? Or is it theoretical? I wonder if we had checks for empty arrays higher in the code somewhere?

Inspired by issue #5141 regarding fuzz testing for the file format, I wrote a simple local fuzzer. This panic occurred while generating seed files.

fenfeng9:
BTW, could you help review this PR https://github.com/lance-format/lance/pull/5867 when you have a moment?

PR #5891, opened 2026-02-05
refactor: use dict entries and encoded size instead of cardinality for dict decision

Xuanwo:
This PR changed how we decide to use dict or now. Instead of cardinality, we will use dict entries and encoded size instead.

---


westonpace:
> Instead of cardinality, we will use dict entries and encoded size instead.

Does this mean we are always calculating the true cardinality (until we hit the limit) instead of HLL?  Do we know what impact this has on write speeds?

Xuanwo:
> Does this mean we are always calculating the true cardinality (until we hit the limit) instead of HLL? Do we know what impact this has on write speeds?

Yes. 

I did some local benchmark on simulated dataset:

- str_unique: 17.579ms -> 13.378ms (-23.90%)
- str_low_card_128: 15.786ms -> 11.952ms (-24.29%)
- u64_unique: 2.744ms -> 2.698ms (-1.68%)
- u64_low_card_128: 2.554ms -> 2.560ms (+0.23%)

I also run some test on the real data like `Pride and Prejudice`:

```
- pride_lines_r8192  +10.20%（2.9747ms -> 3.2782ms）
```

The unique ratio is `0.9796` which is so close to `0.98` but at the same time it's not good for dict, so we are doing some extra work which makes it slower.

I think in general it's good. But maybe you'll want to add config options to allow users to tune those settings?

rust/lance-encoding/src/encodings/logical/primitive/dict.rs
```diff
@@ -163,6 +160,9 @@ where
         let idx = match map.entry(U8SliceKey(key)) {
             Entry::Occupied(entry) => *entry.get(),
             Entry::Vacant(entry) => {
+                if max_dict_entries == 0 || curr_idx as u32 >= max_dict_entries {
```
westonpace, line 163: The `max_dict_entries == 0` check seems a little redundant since `curr_idx` will always be 0 or greater.

rust/lance-encoding/src/encodings/logical/primitive.rs
```diff
@@ -4580,30 +4549,118 @@ impl PrimitiveStructuralEncoder {
...
+        }
+
+        let max_dict_entries = u32::try_from(threshold_cardinality.min(i32::MAX as u64)).ok()?;
+        Some(DictEncodingBudget {
+            max_dict_entries,
+            max_encoded_size,
+        })
+    }
+
+    fn sample_is_near_unique(
```
westonpace, line 4591: Let's document this.  It's quite clever.  Looks like you are randomly sampling 4096 values and testing them for uniqueness before you consider dictionary encoding further?

This is basically just a different way of doing cardinality estimation right?

rust/lance-encoding/src/statistics.rs
```diff
@@ -189,12 +186,31 @@ impl GetStat for NullableDataBlock {
...
         if block_info.is_empty() {
             panic!("get_stat should be called after statistics are computed.");
         }
-        block_info.get(&stat).cloned()
+        Some(
+            block_info
+                .entry(stat)
+                .or_insert_with(|| computed.clone())
+                .clone(),
+        )
```
westonpace, line 213: It looks like you stopped eagerly calculating HLL but you still calculate it on demand?  Why is that?  I think we should probably do this with all the stats but I'm curious why you didn't just get rid of HLL?  Is there some path that still needs it?

rust/lance-encoding/src/encodings/logical/primitive/dict.rs
Xuanwo: you are right, let me remove it.

rust/lance-encoding/src/encodings/logical/primitive.rs
Xuanwo: Yes, and I pick values in systematic sampling (but yes, it's a bit randomly).

rust/lance-encoding/src/statistics.rs
Xuanwo: I want to minimize the code changes in this PR when I started this work. It is true that we could simply remove it. I will implement this in a follow-up PR.

PR #5893, opened 2026-02-05
fix: avoid bitmap range panic on inverted bounds

### LINKED ISSUE #5878: [Bug] BITMAP index panic on BETWEEN when low > high
### Description
When using a `BITMAP` scalar index, a `BETWEEN` filter with inverted bounds (low > high) triggers a Rust panic in a background thread. Instead of returning an empty result.

### How to reproduce
```python
import tempfile

import lance
import pyarrow as pa


def main() -> None:
    data = pa.table({"c0": pa.array([1, 2, 3], type=pa.int64())})
    filter_expr = "c0 BETWEEN 2 AND 1"  # low > high

    work_dir = tempfile.mkdtemp()
    path = f"{work_dir}/ds"

    lance.write_dataset(data, path)
    ds = lance.dataset(path)
    ds.create_scalar_index("c0", index_type="BITMAP")

    print("Executing query... (expect panic in stderr)")
    result = ds.to_table(filter=filter_expr)
    print("Rows:", result.num_rows)

if __name__ == "__main__":
    main()

```
### Result

```
Executing query... (expect panic in stderr)

thread 'lance_background_thread' panicked at /rustc/1159e78c4747b02ef996e55082b704c09b970588/library/alloc/src/collections/btree/search.rs:121:21:
range start is greater than range end in BTreeMap
  File "test.py", line 23, in <module>
    main()
  File "test.py", line 19, in main
    result = ds.to_table(filter=filter_expr)
  File ".../site-packages/lance/dataset.py", line 1113, in to_table
    ).to_table()
  File ".../site-packages/lance/dataset.py", line 4963, in to_table
  File "pyarrow/ipc.pxi", line 762, in pyarrow.lib.RecordBatchReader.read_all
pyarrow.lib.ArrowInvalid: External error: RuntimeError: Task was aborted
```

fenfeng9:
closes #5878

#### Problem
  BITMAP index panics when receiving inverted range queries like `BETWEEN 2 AND 1` or `x > 2 AND x < 2`. This is because  `BTreeMap::range()` panics when `start > end`  or when both bounds are Excluded at the same value.

  #### Solution
  Add range validation matching BTREE's existing behavior:
  - `[a, b]`: empty if `a > b`
  - `[a, b)`, `(a, b]`, `(a, b)`: empty if `a >= b`

Returns empty result for empty ranges instead of panicking.

### DIFF
--- a/python/python/tests/test_scalar_index.py
+++ b/python/python/tests/test_scalar_index.py
@@ -1554,6 +1554,21 @@ def test_bitmap_index(tmp_path: Path):
     assert indices[0]["type"] == "Bitmap"
 
 
+def test_bitmap_empty_range(tmp_path: Path):
+    data = pa.table({"c0": pa.array([1, 2, 3], type=pa.int64())})
+    dataset = lance.write_dataset(data, tmp_path / "dataset")
+    dataset.create_scalar_index("c0", index_type="BITMAP")
+    filters = [
+        "c0 BETWEEN 2 AND 1",
+        "c0 > 2 AND c0 < 2",
+        "c0 >= 2 AND c0 < 2",
+        "c0 > 2 AND c0 <= 2",
+    ]
+    for filter_expr in filters:
+        result = dataset.to_table(filter=filter_expr, use_scalar_index=True)
+        assert result.num_rows == 0
+
+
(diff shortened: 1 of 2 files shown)


fenfeng9:
PTAL @wjones127

PR #5894, opened 2026-02-05
fix: split index_statistics to reduce rustc query depth

Xuanwo:
This refactors `DatasetIndexExt::index_statistics` into smaller helper functions to reduce the async state-machine complexity.

This avoids `rustc` query depth overflow when compiling with nightly + coverage instrumentation and many enabled features.

---

### DIFF
--- a/rust/lance/src/index.rs
+++ b/rust/lance/src/index.rs
@@ -1056,7 +1056,7 @@ async fn index_statistics_scalar(
     let field_path = ds.schema().field_path(field_id)?;
 
     let (indices_stats, index_uri, num_indices, updated_at) =
-        collect_scalar_indices_statistics(ds, metadatas, &field_path).await?;
+        collect_regular_indices_statistics(ds, metadatas, &field_path).await?;
 
     let index_type_hint = indices_stats
         .first()
@@ -1070,7 +1070,7 @@ async fn index_statistics_scalar(
         num_unindexed_fragments,
         num_indexed_rows,
         num_unindexed_rows,
-    )) = fragment_statistics_or_migrate(ds, index_name).await?
+    )) = gather_fragment_statistics(ds, index_name).await?
     else {
         return migrate_and_recompute_index_statistics(ds, index_name).await;
     };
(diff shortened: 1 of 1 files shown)


rust/lance/src/index.rs
```diff
@@ -1132,6 +978,213 @@ impl DatasetIndexExt for Dataset {
...
+        "num_unindexed_fragments": num_unindexed_fragments,
+        "num_unindexed_rows": num_unindexed_rows,
+        "num_indexed_rows_per_delta": num_indexed_rows_per_delta,
+        "updated_at_timestamp_ms": updated_at,
+    });
+
+    serialize_index_statistics(&stats)
+}
+
+async fn collect_scalar_indices_statistics(
```
westonpace: I believe this path is actually handling both scalar and vector indices.  Perhaps we should call it `collect_regular_indices_statistics` because these are "regular indices" (e.g. not frag reuse or memwal)?  Or maybe we could have some other name for these things?  Either way, `scalar` is a bit misleading.

rust/lance/src/index.rs
```diff
@@ -1132,6 +978,213 @@ impl DatasetIndexExt for Dataset {
...
+
+    Ok((
+        indices_stats,
+        index_uri.unwrap_or_else(|| "unknown".to_string()),
+        num_indices,
+        updated_at,
+    ))
+}
+
+async fn fragment_statistics_or_migrate(
```
westonpace: Can we drop the `or_migrate`?  Maybe call this `gather_fragment_statistics`?  I'm not sure that is the main point of this method.  Also, I don't know if it is actually migrating anything in this method.  That happens in the calling method.

rust/lance/src/index.rs
Xuanwo: Fxied in https://github.com/lance-format/lance/pull/5894/commits/93c9acc0be85f907366f7bb76835651797f6e241

rust/lance/src/index.rs
Xuanwo: Fixed in https://github.com/lance-format/lance/pull/5894/commits/93c9acc0be85f907366f7bb76835651797f6e241

PR #5896, opened 2026-02-05
feat: add partitioned namespace

wojiaodoubao:
This is a sub-task of [the partitioned namespace](https://github.com/lance-format/lance/issues/5741)

### DIFF
--- a/java/lance-jni/Cargo.toml
+++ b/java/lance-jni/Cargo.toml
@@ -24,11 +24,12 @@ lance-index = { path = "../../rust/lance-index" }
 lance-io = { path = "../../rust/lance-io" }
 lance-namespace = { path = "../../rust/lance-namespace" }
 lance-namespace-impls = { path = "../../rust/lance-namespace-impls", features = ["rest", "rest-adapter"] }
+lance-namespace-reqwest-client = "0.5.2"
 lance-core = { path = "../../rust/lance-core" }
 lance-file = { path = "../../rust/lance-file" }
 lance-table = { path = "../../rust/lance-table" }
 arrow = { version = "57.1", features = ["ffi"] }
-arrow-schema = "57.1"
+arrow-schema = "57.3.0"
 object_store = { version = "0.12.2" }
 tokio = { version = "1.23", features = [
     "rt-multi-thread",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5305,15 +5305,19 @@ dependencies = [
  "base64 0.22.1",
  "bytes",
  "chrono",
+ "datafusion-functions",
  "futures",
  "google-cloud-auth",
  "lance",
+ "lance-arrow",
  "lance-core",
  "lance-index",
  "lance-io",
  "lance-namespace",
+ "lance-namespace-reqwest-client",
  "lance-table",
  "log",
+ "murmur3",
  "object_store",
  "rand 0.9.2",
  "reqwest",
@@ -6026,6 +6030,12 @@ name = "multimap"
 version = "0.10.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "1d87ecb2933e8aeadb3e3a02b828fed80a7528047e68b4f424523a0981a3a084"
+
+[[package]]
+name = "murmur3"
+version = "0.5.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "9252111cf132ba0929b6f8e030cac2a24b507f3a4d6db6fb2896f27b354c714b"
 
 [[package]]
 name = "murmurhash32"
--- a/java/lance-jni/Cargo.lock
+++ b/java/lance-jni/Cargo.lock
@@ -3746,6 +3746,7 @@ dependencies = [
  "lance-linalg",
  "lance-namespace",
  "lance-namespace-impls",
+ "lance-namespace-reqwest-client",
  "lance-table",
  "log",
  "object_store",
@@ -3798,14 +3799,18 @@ dependencies = [
  "axum",
  "bytes",
  "chrono",
+ "datafusion-functions",
  "futures",
  "lance",
+ "lance-arrow",
  "lance-core",
  "lance-index",
  "lance-io",
  "lance-namespace",
+ "lance-namespace-reqwest-client",
  "lance-table",
  "log",
+ "murmur3",
  "object_store",
  "rand 0.9.2",
  "reqwest",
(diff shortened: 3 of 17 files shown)


java/lance-jni/Cargo.toml
```diff
@@ -24,6 +24,7 @@ lance-index = { path = "../../rust/lance-index" }
 lance-io = { path = "../../rust/lance-io" }
 lance-namespace = { path = "../../rust/lance-namespace" }
 lance-namespace-impls = { path = "../../rust/lance-namespace-impls", features = ["rest", "rest-adapter"] }
+lance-namespace-reqwest-client = { git = "https://github.com/wojiaodoubao/lance-namespace", branch = "rest-table-properties" }
```
wojiaodoubao: After https://github.com/lance-format/lance-namespace/pull/299 is merged, update it.

rust/lance-namespace-impls/src/dir/manifest_ext.rs
```diff
@@ -0,0 +1,433 @@
...
+    CreateNamespaceRequest, DeclareTableRequest, DeclareTableResponse,
+};
+use snafu::location;
+use std::collections::HashMap;
+use std::collections::HashSet;
+use std::sync::Arc;
+
+/// Request for creating multiple namespaces with a single merge insert.
+#[derive(Debug, Clone)]
+pub struct CreateMultiNamespacesRequest {
```
jackye1995: why do we need this? I thought we only need creating multiple tables in a namespace

jackye1995: oh I see nvm because we want to create the namespaces that represent partition values.

jackye1995: I think we should add these as a part of the Lance Namespace operations, introduce `BatchCreateNamespaces`, `BatchCreateTables`, etc., those operations can be useful anyway even outside the context of partitioned namespace.

And then you don't need a dedicated extension just for manifest namespace to make it work.

rust/lance-namespace-impls/src/dir/manifest_ext.rs
```diff
@@ -0,0 +1,433 @@
...
+    pub namespaces: Vec<CreateNamespaceRequest>,
+    /// Columns used for merge insert deduplication.
+    pub on: Vec<String>,
+}
+
+#[derive(Debug, Default, Clone)]
+pub struct CreateMultiNamespacesRequestBuilder {
```
jackye1995: just to create a separated thread. I am starting to think, is there real benefit in creating the sub-namespace structures? It seems purely for the purpose that it is cool to list namespaces in this way, but it does not serve any practical purposes since all the pruning are done directly against the table's partition column values in `__manifest`. Would it make more sense to just not have those nested namespace structures?

rust/lance-namespace-impls/src/dir/manifest_ext.rs
wojiaodoubao: > I think we should add these as a part of the Lance Namespace operations, introduce BatchCreateNamespaces, BatchCreateTables, etc.

Agree, we can do this.

rust/lance-namespace-impls/src/dir/manifest_ext.rs
wojiaodoubao: Yes, having a table is sufficient for partition creation and pruning. The reason for retaining the namespace is that PartitionedNamespace is a type of DirectoryNamespace that follows to the partition spec standard. If we only keep the table part, PartitionedNamespace can no longer be treated as a normal DirectoryNamespace. 

I think from a consistency perspective, it would be better to retain it. Shall we retain it, or remove it for simplicity?

wojiaodoubao:
I just finished the first version of partitioned namespace and it is ready for review now. When I implemented it, I found the original design need some update, here is the reason.

## Explanation of the New Design

Compared to the previous partitioning spec, the current implementation introduces two key changes. Below I’ll explain the motivation behind these adjustments and would love to get further feedback and discussion.

### 1. Introducing TableSpec

In the earlier partition spec design, all tables shared the same schema, which was stored in the table metadata. A PartitionedNamespace contained multiple sub-namespaces named v{i}, representing different versions of the partition spec.

The problem with that design is that it makes schema evolution difficult. We cannot atomically modify the schema of all tables without breaking the semantic guarantee that “all tables share the same schema.”

Even if we support multi-partition transactions in the future, we would still be blocked at the step of updating the schema stored in table metadata, because updating table metadata itself does not have transactional semantics.

To solve this problem, I propose introducing a new abstraction: TableSpec, which encapsulates both the schema and the partition spec.

It is defined as:
```
pub struct TableSpec {
    id: i32,
    schema: ArrowSchema,
    partition_spec: PartitionSpec,
}
```

TableSpec is stored as a first-level sub-namespace under the PartitionedNamespace root. The v{i} naming convention corresponds to the id, while the schema and partition_spec are stored in the namespace properties.

Whenever a schema evolution or partition spec evolution occurs, we create a new sub-namespace. This operation is purely metadata-level. In manifest namespace, creating a namespace is atomic.

New data must follow the new schema (i.e., tables are created and written under the new TableSpec namespace).

Existing data does not need to be modified and continues to use the previous schema and partition spec.

Another advantage of this design is that it makes it relatively straightforward to support branching functionality in the future.

The overall structure can be visualized as follows:
image

### 2. Using Deterministic Names for Namespaces
In the previous partition spec design, the name of a partition namespace was a 16-character base36 string. Any type of partition value would be mapped to such a 16-character base36 string.

The issue with this design is that it makes it difficult to resolve concurrency conflicts. In distributed scenarios, we may need to concurrently create tables and namespaces with the same partition values. Since partition values effectively serve as business primary keys, this becomes a consistency challenge.

Originally, I considered leveraging Lance’s merge-insert deduplication capability to enforce uniqueness at the business key level. However, merge-insert in Lance requires that:
* A column be explicitly defined as a primary key
* The primary key column must be non-null

In our current design, the __manifest partition field column must be nullable. Additionally, partition fields can evolve over time, meaning we cannot reliably define a fixed initial primary key column. This makes it infeasible to use merge-insert deduplication directly on partition fields.

To address this, I propose using deterministic names for partition namespaces. At each level, the namespace name is derived from the serialized string representation of the partition value.

With this approach:
* Namespace identity becomes deterministic and directly tied to partition values.
* We only need to define a primary key on the `object_id` column.
* We can then leverage merge-insert to resolve concurrency conflicts when creating new namespaces and tables.

This simplifies concurrency handling while preserving correctness in distributed environments.

wojiaodoubao:
Hi @jackye1995 , what do you think about the design change? https://github.com/lance-format/lance/pull/5896#issuecomment-3960040902

PR #5899, opened 2026-02-05
fix: ensure manifest timestamp is monotonically increasing

jackye1995:


majin1102:
A little concern on this:

If a commit writes a manifest timestamp far in the future (e.g., due to a clock error setting it one year ahead), monotonic logic treats this value as the new baseline. Consequently, all subsequent manifests are forced to use timestamps at or beyond that future point, never reverting to real wall-clock time, effectively shifting the entire dataset's timeline into the future.

PR #5905, opened 2026-02-07
feat: support FTS query execution for LSM scanner

touch-of-grey:
Based on previous discussion, separate out FTS query plan since it requires global BM25. @jackye1995 please take a look

This will calculate global BM25 and then use the same scorer to rank across different inverted indexes, similar to how Lucene does it.

### DIFF
--- a/rust/lance-index/src/scalar.rs
+++ b/rust/lance-index/src/scalar.rs
@@ -323,6 +323,9 @@ pub struct FullTextSearchQuery {
     /// Increasing this value will reduce the recall and improve the performance
     /// 1.0 is the value that would give the best performance without recall loss
     pub wand_factor: Option<f32>,
+
+    /// Optional BM25 stats override for custom corpus-level statistics.
+    pub bm25_override: Option<Arc<inverted::scorer::BM25StatsOverride>>,
 }
 
 impl FullTextSearchQuery {
@@ -333,6 +336,7 @@ impl FullTextSearchQuery {
             query,
             limit: None,
             wand_factor: None,
+            bm25_override: None,
         }
     }
 
@@ -343,6 +347,7 @@ impl FullTextSearchQuery {
             query,
             limit: None,
             wand_factor: None,
+            bm25_override: None,
         }
     }
 
@@ -352,6 +357,7 @@ impl FullTextSearchQuery {
             query,
             limit: None,
             wand_factor: None,
+            bm25_override: None,
         }
     }
 
@@ -389,6 +395,7 @@ impl FullTextSearchQuery {
         FtsSearchParams::new()
             .with_limit(self.limit.map(|limit| limit as usize))
             .with_wand_factor(self.wand_factor.unwrap_or(1.0))
+            .with_bm25_override(self.bm25_override.clone())
     }
 }
 
--- a/rust/lance-index/src/scalar/inverted/query.rs
+++ b/rust/lance-index/src/scalar/inverted/query.rs
@@ -1,13 +1,16 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-FileCopyrightText: Copyright The Lance Authors
 
+use std::collections::HashSet;
+use std::sync::Arc;
+
 use crate::scalar::inverted::lance_tokenizer::DocType;
+use crate::scalar::inverted::scorer::BM25StatsOverride;
 use crate::scalar::inverted::tokenizer::lance_tokenizer::LanceTokenizer;
 use lance_core::{Error, Result};
 use serde::ser::SerializeMap;
 use serde::{Deserialize, Serialize};
 use snafu::location;
-use std::collections::HashSet;
 
 #[derive(Debug, Clone)]
 pub struct FtsSearchParams {
@@ -20,6 +23,8 @@ pub struct FtsSearchParams {
     pub phrase_slop: Option<u32>,
     /// The number of beginning characters being unchanged for fuzzy matching.
     pub prefix_length: u32,
+    /// Optional BM25 stats override for custom corpus-level statistics.
+    pub bm25_override: Option<Arc<BM25StatsOverride>>,
 }
 
 impl FtsSearchParams {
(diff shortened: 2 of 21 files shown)


jackye1995:
Thanks, I will take a look tomorrow morning

rust/lance-index/src/scalar/inverted/query.rs
```diff
@@ -20,6 +23,8 @@ pub struct FtsSearchParams {
     pub phrase_slop: Option<u32>,
     /// The number of beginning characters being unchanged for fuzzy matching.
     pub prefix_length: u32,
+    /// Optional BM25 stats override for cross-generation scoring.
```
jackye1995: cross-generation scoring is too specific to LSM, make the comment more generic

rust/lance-index/src/scalar.rs
```diff
@@ -323,6 +323,9 @@ pub struct FullTextSearchQuery {
     /// Increasing this value will reduce the recall and improve the performance
     /// 1.0 is the value that would give the best performance without recall loss
     pub wand_factor: Option<f32>,
+
+    /// Optional BM25 stats override for cross-generation scoring.
```
jackye1995: cross-generation scoring is too specific to LSM, make the comment more generic

rust/lance/src/dataset/mem_wal/index/fts.rs
```diff
@@ -1109,6 +1137,371 @@ impl FtsMemIndex {
         results
     }
 
+    /// Execute a query using global BM25 statistics for scoring.
+    ///
+    /// This produces scores that are comparable across LSM generations
+    /// because N, avgdl, and df come from aggregated global stats.
+    pub fn search_query_with_global_stats(
```
jackye1995: why do we need a dedicated method for global stats? Can we only use existing mechanism and only allow optional BM25 override?

rust/lance/src/dataset/mem_wal/scanner/fts_search.rs
```diff
@@ -0,0 +1,463 @@
...
+            pk_columns,
+            base_schema,
+            bloom_filters: Vec::new(),
+            fts_query,
+            global_stats_override: None,
+        }
+    }
+
+    /// Add a bloom filter for staleness detection.
+    pub fn with_bloom_filter(mut self, generation: u64, bloom_filter: Arc<Sbbf>) -> Self {
```
jackye1995, line 91: I think with the latest design, we can make bloom filter also just a bloom filter index in the flushed memtable. It will have a zone size equal to the row count.

rust/lance/src/dataset/mem_wal/scanner/fts_search.rs
```diff
@@ -0,0 +1,463 @@
...
+                        ));
+                    }
+                }
+                LsmDataSource::BaseTable { .. } | LsmDataSource::FlushedMemTable { .. } => {
+                    // For persistent sources, we would need to load the inverted index
+                    // to read stats. For now, these are collected asynchronously during
+                    // plan_search if needed. The planner can be extended later to
+                    // pre-load stats from persistent indices.
+                    //
+                    // When persistent FTS index support is added for flushed memtables,
```
jackye1995: this implementation is missing

rust/lance/src/dataset/mem_wal/scanner/fts_search.rs
```diff
@@ -0,0 +1,463 @@
...
+        for source in sources {
+            if let LsmDataSource::ActiveMemTable { index_store, .. } = source {
+                if let Some(fts_index) = index_store.get_fts_by_column(&self.fts_query.column) {
+                    let query_str = self.query_text();
+                    return fts_index.tokenize_query(&query_str);
+                }
+            }
+        }
+        // Fallback: return the raw query as a single term
+        vec![self.query_text()]
```
jackye1995: should not fallback, the index should always have a tokenizer set

rust/lance/src/dataset/mem_wal/scanner/fts_search.rs
```diff
@@ -0,0 +1,463 @@
...
+                scanner.full_text_search(fts_query)?;
+                scanner.create_plan().await
+            }
+            LsmDataSource::ActiveMemTable {
+                batch_store,
+                index_store,
+                schema,
+                ..
+            } => {
+                use crate::dataset::mem_wal::memtable::scanner::MemTableScanner;
```
jackye1995: import at top

rust/lance/src/dataset/mem_wal/scanner/fts_search.rs
```diff
@@ -0,0 +1,463 @@
...
+            LsmDataSource::BaseTable { dataset } => {
+                let mut scanner = dataset.scan();
+                let cols = self.build_projection_for_fts(projection);
+                scanner.project(&cols.iter().map(|s| s.as_str()).collect::<Vec<_>>())?;
+                let fts_query = self.build_full_text_query(bm25_override);
+                scanner.full_text_search(fts_query)?;
+                scanner.create_plan().await
+            }
+            LsmDataSource::FlushedMemTable { path, .. } => {
+                let dataset = crate::dataset::DatasetBuilder::from_uri(path)
```
jackye1995: we should make sure we use the same session cache across dataset opening for the dataset and flushed memtables.

rust/lance/src/dataset/mem_wal/scanner/fts_search.rs
```diff
@@ -0,0 +1,463 @@
...
+        let sources = self.collector.collect()?;
+
+        if sources.is_empty() {
+            return self.empty_plan(projection);
+        }
+
+        // Phase 1: Collect global stats
+        let global_stats = match &self.global_stats_override {
+            Some(stats) => stats.clone(),
+            None => self.collect_global_stats(&sources)?,
```
jackye1995: this is actually quite expensive. We should make sure we are not blocked on loading the bm25 stats, we should compute it while forming plan and doing execution

rust/lance/src/dataset/mem_wal/scanner/fts_search.rs
touch-of-grey: Agree. This impacts both FTS and vector search. I can raise a separated PR later about it

PR #5910, opened 2026-02-07
feat!: support index progress reporting via callbacks

wkalt:
This adds support for progress reporting on index builds, allowing callers to determine whether a build is progressing, what stage it is in, and in some cases how far into the stage.

Breaking: updates some index builder function signatures to include a progress callback implementation. A noop implementation is included in the patch.

### DIFF
--- a/java/lance-jni/src/vector_trainer.rs
+++ b/java/lance-jni/src/vector_trainer.rs
@@ -1,6 +1,8 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-FileCopyrightText: Copyright The Lance Authors
 
+use std::sync::Arc;
+
 use crate::blocking_dataset::{BlockingDataset, NATIVE_DATASET};
 use crate::error::{Error, Result};
 use crate::ffi::JNIEnvExt;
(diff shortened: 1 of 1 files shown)


wkalt:
I used this to create the chart in https://github.com/lance-format/lance/pull/5907

wkalt:
this should be marked breaking, sorry for not marking that right. I can't edit the title so someone will need to add it.

This adds a progress parameter to some of the index building functions.

rust/lance-index/src/progress.rs
```diff
@@ -0,0 +1,50 @@
...
+
+use async_trait::async_trait;
+use lance_core::Result;
+use std::sync::Arc;
+
+/// Progress callback for index building.
+///
+/// Called at stage boundaries during index construction. Stage names are
+/// index-type-specific (e.g. "train_ivf", "shuffle", "build_partitions" for
+/// vector indices; "load_data", "build_pages" for scalar indices).
```
wjones127: Is it an error for multiple stages to be active at the same time?

wjones127 (approved):
This looks great!

rust/lance-index/src/progress.rs
wkalt: good thought. As currently structured stages are sequential, so if you need to do two things you would need to fold it into one stage. We could possibly figure out a way to relax that or evolve this to enable multiple stages at once.

Do we have index types today that would benefit from that? This only breaks out stages for IVF; I haven't given consideration to the others yet.

wjones127:
Happy to merge this once conflicts and lint errors are resolved.

wkalt:
@wjones127 thanks, this should be good to go.

PR #5911, opened 2026-02-08
feat: support aggregate in scanner

jackye1995:
The main use case of this PR is to allow engines like Spark and Trino to pushdown an aggregate into Lance scanner in distributed worker when possible. Today we technically already supports `COUNT(*)` pushdown through `scanner.count_rows()` to count rows of each fragment distributedly, this is a more generic version of that. My plan is to allow an engine to pass a Substrait Aggregate expression to scanner in the worker to support pushdown other aggregations like `SUM`, `MAX`, `MIN`.

Another alternative I have thought about is to just update the `dataset.sql()` API to accept a full Substrait plan so we can execute a plan with aggregate, and update distributed worker to run a SQL statement instead of running the scanner. But doing this feature in scanner feels more aligned with how engines implement the distributed execution. Basically whatever that could be executed by a single worker in a distributed environment (predicate pushdown, column projection, aggregate pushdown) should be supported by the scanner.

Note that with this change, we can technically remove `create_count_plan` since it's just a subcase of `create_aggregate_plan`, but we are not doing it in this PR. Once we agree upon this direction, I will do a separated PR to refactor that.

### DIFF
--- a/rust/lance/src/dataset/scanner.rs
+++ b/rust/lance/src/dataset/scanner.rs
@@ -487,7 +487,7 @@ impl AggregateExpr {
     ///     .build();
     /// scanner.aggregate(agg);
     /// ```
-    pub fn builder() -> AggregateExprBuilder {
+    pub fn builder() -> AggregateExprBuilder<false> {
         AggregateExprBuilder::new()
     }
 
@@ -533,81 +533,28 @@ impl AggregateExpr {
 }
 
 /// Builder for creating aggregate expressions without using DataFusion or Substrait directly.
-#[derive(Debug, Clone, Default)]
-pub struct AggregateExprBuilder {
+///
+/// The const generic `HAS_PENDING` tracks whether there's a pending aggregate that can be aliased.
+/// When `HAS_PENDING` is `true`, the last item in `aggregates` is the pending aggregate.
+#[derive(Debug, Clone)]
+pub struct AggregateExprBuilder<const HAS_PENDING: bool> {
     group_by: Vec<Expr>,
     aggregates: Vec<Expr>,
 }
 
-impl AggregateExprBuilder {
+impl Default for AggregateExprBuilder<false> {
+    fn default() -> Self {
+        Self {
+            group_by: Vec::new(),
+            aggregates: Vec::new(),
+        }
+    }
+}
+
+impl AggregateExprBuilder<false> {
     /// Create a new builder.
     pub fn new() -> Self {
         Self::default()
-    }
-
-    /// Add a column to group by.
-    pub fn group_by(mut self, column: impl Into<String>) -> Self {
-        self.group_by.push(col(column.into()));
-        self
-    }
-
-    /// Add multiple columns to group by.
-    pub fn group_by_columns(
-        mut self,
-        columns: impl IntoIterator<Item = impl Into<String>>,
-    ) -> Self {
-        for column in columns {
-            self.group_by.push(col(column.into()));
-        }
-        self
-    }
-
-    /// Add COUNT(*) aggregate.
-    pub fn count_star(self) -> AggregateExprBuilderWithPendingAggregate {
-        AggregateExprBuilderWithPendingAggregate {
-            builder: self,
-            pending: functions_aggregate::count::count(lit(1)),
-        }
-    }
-
-    /// Add COUNT(column) aggregate.
-    pub fn count(self, column: impl Into<String>) -> AggregateExprBuilderWithPendingAggregate {
-        AggregateExprBuilderWithPendingAggregate {
-            builder: self,
-            pending: functions_aggregate::count::count(col(column.into())),
-        }
-    }
-
-    /// Add SUM(column) aggregate.
-    pub fn sum(self, column: impl Into<String>) -> AggregateExprBuilderWithPendingAggregate {
-        AggregateExprBuilderWithPendingAggregate {
-            builder: self,
-            pending: functions_aggregate::sum::sum(col(column.into())),
-        }
-    }
-
-    /// Add AVG(column) aggregate.
-    pub fn avg(self, column: impl Into<String>) -> AggregateExprBuilderWithPendingAggregate {
-        AggregateExprBuilderWithPendingAggregate {
-            builder: self,
-            pending: functions_aggregate::average::avg(col(column.into())),
-        }
-    }
-
-    /// Add MIN(column) aggregate.
-    pub fn min(self, column: impl Into<String>) -> AggregateExprBuilderWithPendingAggregate {
-        AggregateExprBuilderWithPendingAggregate {
-            builder: self,
-            pending: functions_aggregate::min_max::min(col(column.into())),
-        }
-    }
-
-    /// Add MAX(column) aggregate.
-    pub fn max(self, column: impl Into<String>) -> AggregateExprBuilderWithPendingAggregate {
-        AggregateExprBuilderWithPendingAggregate {
-            builder: self,
-            pending: functions_aggregate::min_max::max(col(column.into())),
-        }
     }
 
     /// Build the aggregate expression.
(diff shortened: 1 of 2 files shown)


rust/lance-datafusion/src/aggregate.rs
```diff
@@ -0,0 +1,15 @@
...
+
+//! Aggregate specification for DataFusion aggregates.
+
+use datafusion::logical_expr::Expr;
+
+/// Aggregate specification with group by and aggregate expressions.
+#[derive(Debug, Clone)]
+pub struct Aggregate {
+    pub group_by: Vec<Expr>,
+    pub aggregates: Vec<Expr>,
```
westonpace, line 15: Let's go ahead and comment these two.

rust/lance/src/dataset/scanner.rs
```diff
@@ -1767,6 +1837,146 @@ impl Scanner {
         .boxed()
     }
 
+    /// Create an execution plan with aggregation.
+    ///
+    /// Requires `aggregate()` to be called first.
+    pub fn create_aggregate_plan(&self) -> BoxFuture<'_, Result<Arc<dyn ExecutionPlan>>> {
```
westonpace, line 2004: Why do we need a separate `create_aggregate_plan`?  Can we just append the aggregate expr onto the end of `create_plan` when `aggregate` is set?  It would be surprising to me as a user if I called `aggregate` and then `create_plan` just ignored it.

rust/lance/src/dataset/scanner.rs
```diff
@@ -462,6 +463,66 @@ impl ExprFilter {
...
+/// Aggregate expression from Substrait or DataFusion.
+#[derive(Debug, Clone)]
+pub enum AggregateExpr {
+    #[cfg(feature = "substrait")]
+    Substrait(Vec<u8>),
+    Datafusion {
+        group_by: Vec<Expr>,
+        aggregates: Vec<Expr>,
+        output_names: Vec<String>,
+    },
```
westonpace, line 474: We can do this in a follow-up but there should be a way for callers to specify an `AggregateExpr` without needing to use DF or Substrait.  Ideally with some kind of builder.
```

File: rust/lance/src/dataset/scanner.rs [hunk 4]
```diff
@@ -462,6 +463,66 @@ impl ExprFilter {
...
 
+/// Aggregate expression from Substrait or DataFusion.
+#[derive(Debug, Clone)]
+pub enum AggregateExpr {
+    #[cfg(feature = "substrait")]
+    Substrait(Vec<u8>),
+    Datafusion {
+        group_by: Vec<Expr>,
+        aggregates: Vec<Expr>,
+        output_names: Vec<String>,
```
westonpace: Why is this needed?  Datafusion `Expr` already has a name concept.  In other words, if a user wants the `MAX(temp)` aggregate to be named `max_temp` they can do `.alias("max_temp")` on the `Expr` right?

rust/lance/src/dataset/scanner.rs
```diff
@@ -1767,6 +1837,146 @@ impl Scanner {
...
+            &aliased_expr,
+            df_schema,
+            input_schema.as_ref(),
+            &ExecutionProps::default(),
+        )?;
+
+        Ok(agg_expr)
+    }
+
+    /// Apply type coercion to aggregate arguments for UserDefined signature functions.
```
westonpace, line 2084: Why is this needed?  Why don't we do type coercion on non-user defined functions?

rust/lance/src/dataset/scanner.rs
```diff
@@ -1767,6 +1837,146 @@ impl Scanner {
...
+        use datafusion::physical_planner::create_aggregate_expr_and_maybe_filter;
+
+        let coerced_expr = self.coerce_aggregate_expr(expr, df_schema)?;
+        let aliased_expr = if let Some(name) = alias {
+            coerced_expr.alias(name)
+        } else {
+            coerced_expr
+        };
+
+        let (agg_expr, _filter, _order_by) = create_aggregate_expr_and_maybe_filter(
```
westonpace: What does it mean if `_filter` or `_order_by` are set?  Do we need to utilize these?

rust/lance/src/dataset/scanner.rs
```diff
@@ -1767,6 +1837,146 @@ impl Scanner {
...
+                Ok(Expr::AggregateFunction(AggregateFunction::new_udf(
+                    func.clone(),
+                    coerced_args,
+                    agg_func.params.distinct,
+                    agg_func.params.filter.clone(),
+                    agg_func.params.order_by.clone(),
+                    agg_func.params.null_treatment,
+                )))
+            }
+            other => Ok(other.clone()),
```
westonpace: Should it be an error if we hit this branch?

rust/lance-datafusion/src/aggregate.rs
```diff
@@ -0,0 +1,15 @@
...
+    /// Output column names in order: group_by columns first, then aggregates.
+    pub output_names: Vec<String>,
```
westonpace: I mention this in scanner.rs too but having `output_names` be a separate property is a Substrait thing.  I think we could instead alias the expressions when we are parsing the Substrait.  This keeps the logic in `Scanner` simpler.

rust/lance/src/dataset/scanner.rs
```diff
@@ -1017,6 +1081,12 @@ impl Scanner {
         self
     }
 
+    /// Set aggregation.
+    pub fn aggregate(&mut self, aggregate: AggregateExpr) -> &mut Self {
+        self.aggregate = Some(aggregate);
```
westonpace, line 1243: We should probably emit a warning or error if a user calls both `aggregate` and `limit`.

rust/lance/src/dataset/scanner.rs
```diff
@@ -1767,6 +1837,146 @@ impl Scanner {
...
+
+        async move {
+            let agg_spec = self.aggregate.as_ref().ok_or_else(|| {
+                Error::invalid_input(
+                    "create_aggregate_plan called but no aggregate was set",
+                    location!(),
+                )
+            })?;
+
+            let plan = self.create_plan().await?;
```
westonpace: If a user calls `order_by` how should we interpret it?  Ordering the data before the aggregate (what it is doing today) is just meaningless work.  We could interpret it as a request to order the data after the aggregate?  Or we could just return an error and say you can't do both?  Or we could just log a warning and ignore the sort?

rust/lance/src/dataset/scanner.rs
hamersaw: Is there a use-case (like my `plan_splits` proposal?) that would benefit from separating this logic? Maybe `create_aggregate_plan` is called in `plan_spits` and `create_plan`? If the goal is to create two separate entrypoints, one for executing an expr and one for breaking it into multiple partitions maybe there is utility. I agree that setting an aggregate and then having it not apply when calling `create_plan` would surprise me as a user.

rust/lance/src/dataset/scanner.rs
```diff
@@ -513,16 +524,167 @@ impl AggregateExpr {
...
+    /// Add multiple columns to group by.
+    pub fn group_by_columns(
+        mut self,
+        columns: impl IntoIterator<Item = impl Into<String>>,
+    ) -> Self {
+        for column in columns {
+            self.group_by.push(col(column.into()));
+        }
+        self
+    }
```
westonpace, line 597: Minor nit: can we comment that multiple invocations of `group_by` or `group_by_columns` will add to the list (and not replace it).  E.g. `.group_by("x").group_by_columns(["y", "z"])` will group by `x`, `y`, and `z`.

rust/lance/src/dataset/scanner.rs
```diff
@@ -513,16 +524,167 @@ impl AggregateExpr {
...
+    /// Add COUNT(*) aggregate.
+    pub fn count_star(self) -> AggregateExprBuilderWithPendingAggregate {
+        AggregateExprBuilderWithPendingAggregate {
+            builder: self,
+            pending: functions_aggregate::count::count(lit(1)),
+        }
+    }
+
+    /// Add COUNT(column) aggregate.
+    pub fn count(self, column: impl Into<String>) -> AggregateExprBuilderWithPendingAggregate {
```
westonpace: Minor nit: the difference between `count_star` and `count` is always subtle for newcomers to SQL.  Can we expand this comment?

```
    /// Add COUNT(column) aggregate.
    ///
    /// Unlike count_star this will only return the number of rows where `column`
    /// is not NULL
```

rust/lance/src/dataset/scanner.rs
```diff
@@ -513,16 +524,167 @@ impl AggregateExpr {
...
+        }
+    }
+}
+
+/// Builder state with a pending aggregate that can be aliased.
+#[derive(Debug, Clone)]
+pub struct AggregateExprBuilderWithPendingAggregate {
+    builder: AggregateExprBuilder,
+    pending: Expr,
+}
```
westonpace: Hmm, I don't love having two structs (and duplicating all the methods) but it isn't the end of the world.  One solution I've seen to this problem has been to use const generics (not sure the below compiles but should communicate the idea)...

```
pub struct AggregateExprBuilder<const HAS_PENDING_AGG: bool> {
    group_by: Vec<Expr>,
    aggregates: Vec<Expr>,
}

impl<const HAS_PENDING_AGG: bool> AggregateExprBuilder<HAS_PENDING_AGG> {
    pub fn new() -> AggregateExprBuilder<false> {
        AggregateExprBuilder<false> {
            group_by: Vec::default(),
            aggregates: Vec::default(),
        }
    }
    ...
    /// Add SUM(column) aggregate.
    pub fn sum(mut self, column: impl Into<String>) -> AggregateExprBuilder<true> {
        self.aggregates.push(functions_aggregate::sum::sum(col(column.into())));
        AggregateExprBuilder<true> {
            group_by: self.group_by,
            aggregates: self.aggregates
        }
    }
}

impl AggregateExprBuilder<true> {
    /// Set an alias for the pending aggregate.
    pub fn alias(mut self, name: impl Into<String>) -> AggregateExprBuilder<true> {
        let aliased = self.aggregates.pop().alias(name.into());
        self.aggregates.push(aliased);
        self
    }
}
```

rust/lance/src/dataset/scanner.rs
```diff
@@ -1852,6 +2242,21 @@ impl Scanner {
             });
         }
 
+        if self.aggregate.is_some() {
+            if self.limit.is_some() || self.offset.is_some() {
+                return Err(Error::InvalidInput {
+                    source: "Cannot use limit/offset with aggregate. Apply limit after aggregation instead.".into(),
```
westonpace:
```suggestion
                    source: "Cannot use limit/offset with aggregate. Apply limit to the result instead.".into(),
```

rust/lance/src/dataset/scanner.rs
```diff
@@ -1852,6 +2242,21 @@ impl Scanner {
...
+                    location: location!(),
+                });
+            }
+            if self.ordering.is_some() {
+                return Err(Error::InvalidInput {
+                    source: "Cannot use order_by with aggregate. Apply ordering after aggregation instead.".into(),
```
westonpace:
```suggestion
                    source: "Cannot use order_by with aggregate. Apply ordering to the result instead.".into(),
```

rust/lance/src/dataset/scanner.rs
```diff
@@ -2096,6 +2501,13 @@ impl Scanner {
         // Stage 2: filter
         plan = filter_plan.refine_filter(plan, self).await?;
 
+        // Stage 2.5: aggregate (if set, applies aggregate and returns early)
+        if let Some(agg_spec) = &self.aggregate {
+            // Take columns needed for aggregation
+            plan = self.take(plan, self.projection_plan.physical_projection.clone())?;
```
westonpace, line 2505: I think this a no-op?  We should always have the physical projection loaded already?  I don't think aggregates are allowed to reference additional columns so we shouldn't need a take here?

rust/lance/src/dataset/scanner.rs
```diff
@@ -2096,6 +2501,13 @@ impl Scanner {
         // Stage 2: filter
         plan = filter_plan.refine_filter(plan, self).await?;
 
+        // Stage 2.5: aggregate (if set, applies aggregate and returns early)
```
westonpace: We should really just get rid of the whole "stage" concept in the comments 😆

rust/lance/src/dataset/scanner.rs
```diff
@@ -2096,6 +2501,13 @@ impl Scanner {
...
+        if let Some(agg_spec) = &self.aggregate {
```
westonpace, line 2503: I'm slightly torn on whether it makes sense to apply the aggregate before or after the projection.  I fear we will have complaints either way.

If we apply the aggregate here (before projection) we cannot reference system columns or projected columns in the aggregate (e.g. can't do `MAX(_rowoffset)` or `MAX(x * 2)`).  If we apply the aggregate later (stage 7.5) then we can't reference the aggregate in the projection (e.g. can't do `2 * MAX(x)`).

I suppose it only really makes sense to be utilizing aggregates that can be pushed down and we wouldn't push down something like `MAX(x * 2)` anyways (I mean...we could...but I have no desire to do so).

Ok...I think I've convinced myself I like this how it is!

rust/lance/src/dataset/scanner.rs
jackye1995: This is actually not a no-op if we do an aggregate on top of vector search of full text search, the source returns search results with scores rather than the full projection columns. The `take()` function handles both cases pretty gracefully since `TakeExec::try_new` returns `None` when no new columns are needed, causing it to simply return the original plan unchanged. So while it's often a no-op for simple scans, we keep it there for correctness in other source types, and the cost is just a minimal check.

jackye1995: I added a few new tests to cover those cases

jackye1995:
Thanks for the reviews!

rust/lance/src/dataset/scanner.rs
westonpace: Ah, ok.  Normally we'd do that take at stage 5.  So we are pulling it up here instead which is fine since we prevent an aggregate with a limit (otherwise we'd be taking too much if we did that here)

All good!

PR #5912, opened 2026-02-08
perf!: remove shuffle buffer

wkalt:
This removes a buffer in the shuffler that accumulated batches for batched writes to temporary storage. This was configured with a public buffer_size parameter hence the breaking change.

Previously, when we shuffled data we accumulated this many batches for each partition in memory and then flushed them all to disk at once. This may have been intended as an optimization in the original implementation of the shuffler, which supported external shuffling through arbitrary object storage. However, the shuffler was subsequently hardcoded to use local disk (where this kind of buffering serves no benefit) and even on remote object storage, we already have a layer of buffering in the storage writer.

Instead of buffering batches, just write them directly to the FileWriter. This results in much more predictable memory usage and also faster index builds.

### DIFF
--- a/rust/lance-index/src/vector/v3/shuffler.rs
+++ b/rust/lance-index/src/vector/v3/shuffler.rs
@@ -9,7 +9,7 @@ use std::sync::Arc;
 use arrow::{array::AsArray, compute::sort_to_indices};
 use arrow_array::{RecordBatch, UInt32Array};
 use arrow_schema::Schema;
-use futures::prelude::*;
+use futures::{future::try_join_all, prelude::*};
 use lance_arrow::{RecordBatchExt, SchemaExt};
 use lance_core::{
     cache::LanceCache,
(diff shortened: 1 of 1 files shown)


wkalt:
progress

Here is the result on an index build over 100M 384d vectors. Note that this chart still shows a slow leak over the course of the shuffle. I assume that's unrelated and haven't looked at it yet.

rust/lance-index/src/vector/v3/shuffler.rs
```diff
@@ -163,43 +155,16 @@ impl Shuffler for IvfShuffler {
...
-
-            // do flush
-            if counter % self.buffer_size == 0 {
-                let mut futs = vec![];
-                for (part_id, writer) in writers.iter_mut().enumerate() {
-                    let batches = &partition_buffers[part_id];
+                if !batches.is_empty() {
                     partition_sizes[part_id] += batches.iter().map(|b| b.num_rows()).sum::<usize>();
-                    futs.push(writer.write_batches(batches.iter()));
+                    writers[part_id].write_batches(batches.iter()).await?;
```
westonpace: We can do this in a follow-up but it might be nice to still do all the writes in parallel.  E.g. keep the `futs` Vec.  `shuffled` is a `Vec` and not any kind of stream / iterator so the data is all in memory already (I think the important point is getting rid of the `if counter % self.buffer_size == 0`)

```
let mut futs = vec![];
if !batches.is_empty() {
    futs.push(writers[part_id].write_batches(batches.iter()));
}
try_join_all(futs).await?;
```

wkalt: thanks, updated

PR #5914, opened 2026-02-08
fix: avoid panic on empty list LABEL_LIST filters

fenfeng9:
`array_has_any/all(labels, [])` yields an empty label-list query; LabelList index merge (`set_union` / `set_intersection`) `unwraps` the first element and panics on an empty iterator.

 Skip LABEL_LIST index parsing for empty lists to avoid the panic and fall back to normal execution.


fenfeng9:
Hi @westonpace, could you please review this when you have a moment? Thanks!

PR #5917, opened 2026-02-09
perf: build fp16kernels with NEON support on iOS

chyyran:
This PR also adds support for `mtune=apple-a13` when building lance-linalg for iOS, which is the earliest supported iOS architecture at time of writing.

This is pulled out of https://github.com/lance-format/lance/pull/5866 after the initial bug was fixed in https://github.com/lance-format/lance/pull/5747 and lancedb updated to Lance 2.x


chyyran:
I don't seem to be able to edit the PR title and I've already refiled this PR once. Could a reviewer please update the title for me?

PR #5919, opened 2026-02-09
fix: correct OR null semantics for nullable masks

### LINKED ISSUE #5895: [Bug] BTREE index incorrect row count on OR predicate with NULLs
### Description
When using a `BTREE` scalar index, an `OR` filter involving `NULL` values returns a different row count compared to a full scan. 
The scalar index path incorrectly handles three-valued logic when combining `AllowList` and `BlockList` masks, causing `NULL` rows to be incorrectly included in the result set.

### How to reproduce

```python
import tempfile

import lance
import pyarrow as pa

# One NULL and one non-NULL value
data = pa.table({"c1": pa.array([None, 1], type=pa.int64())})
filter_expr = "(c1 != 0) OR (c1 < 5)"

with tempfile.TemporaryDirectory() as d:
    lance.write_dataset(data, f"{d}/no")
    lance.write_dataset(data, f"{d}/ix")
    ds_no = lance.dataset(f"{d}/no")
    ds_ix = lance.dataset(f"{d}/ix")
    ds_ix.create_scalar_index("c1", index_type="BTREE")

    print("no_index:", ds_no.to_table(filter=filter_expr).num_rows)
    print("with_index:", ds_ix.to_table(filter=filter_expr).num_rows)

```

### Result
```
no_index: 1
with_index: 2
```

fenfeng9:
closes #5895

`Allow|Block OR` previously assumed `NULL` rows were always included in the `block.selected` set; when they were not, `NULL`s could be dropped and `FALSE`/`NULL` were mixed, leading to incorrect results.

This change computes `TRUE/FALSE/NULL` explicitly for `Allow|Block OR` and derives `NULL/selected` sets from those.

### DIFF
--- a/python/python/tests/test_scalar_index.py
+++ b/python/python/tests/test_scalar_index.py
@@ -2010,23 +2010,6 @@ def test_scalar_index_with_nulls(tmp_path):
     assert len(result) == k
 
 
-def test_scalar_index_or_with_nulls_matches_scan(tmp_path: Path):
-    data = pa.table({"c1": pa.array([None, 1], type=pa.int64())})
-    filters = [
-        "(c1 != 0) OR (c1 < 5)",
-        "NOT ((c1 != 0) OR (c1 < 5))",
-    ]
-
-    ds_no = lance.write_dataset(data, tmp_path / "no")
-    expected = {f: ds_no.to_table(filter=f).num_rows for f in filters}
-
-    for idx_type in ["BITMAP", "BTREE"]:
-        ds = lance.write_dataset(data, tmp_path / idx_type.lower())
-        ds.create_scalar_index("c1", index_type=idx_type)
-        for f in filters:
-            assert ds.to_table(filter=f).num_rows == expected[f]
-
-
 def test_label_list_index(tmp_path: Path):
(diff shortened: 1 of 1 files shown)


fenfeng9:
Hi @wjones127, could you please review this when you have a moment? Thanks!

fenfeng9:
> Could you add this test to https://github.com/lance-format/lance/blob/main/rust/lance/tests/query/primitives.rs ? That's where we want these kind of query tests.
> 
> I didn't ask for this on the label list PRs because we don't yet have a nested field test suite merged. I've started them here: #5901
> 
> Should just be able to add a filter or two.

Sure, I'll handle this tomorrow. Should I keep the test in Python?

wjones127:
> > Could you add this test to https://github.com/lance-format/lance/blob/main/rust/lance/tests/query/primitives.rs ? That's where we want these kind of query tests.
> > I didn't ask for this on the label list PRs because we don't yet have a nested field test suite merged. I've started them here: #5901
> > Should just be able to add a filter or two.
> 
> Sure, I'll handle this tomorrow. Should I keep the test in Python?

You can if you want.

wjones127 (approved):
Looks good. Thank you!

PR #5924, opened 2026-02-10
fix: respect fragment restrictions in vector and FTS searches when requested fragments

yingjianwu98:
Fixes bug where vector and FTS searches ignore `with_fragments() `filter when querying unindexed fragments, which will return results from indexed fragments that were not requested.

Note, this does not fix the issues where `with_fragments` contains both indexed_fragement and non_indexed_fragment for FTS and vector search, and I have a separate follow up PR to fix the behavior.

This PR combines my previous effort for fixing the issue:
https://github.com/lance-format/lance/pull/5081
https://github.com/lance-format/lance/pull/5080

### DIFF
--- a/rust/lance/src/dataset/scanner.rs
+++ b/rust/lance/src/dataset/scanner.rs
@@ -9120,6 +9120,103 @@ mod test {
         );
     }
 
+    fn assert_values_in_range(array: &Int32Array, range: std::ops::Range<i32>, msg: &str) {
+        assert!(!array.is_empty(), "Expected some results but got none");
+        assert!(
+            array
+                .iter()
+                .all(|v| v.is_some_and(|val| range.contains(&val))),
+            "{msg} (expected range {range:?})"
+        );
+    }
+
+    // Helper to assert that results exist from all fragment ranges
+    fn assert_has_all_fragments(array: &Int32Array) {
+        assert!(
+            array
+                .iter()
+                .any(|v| v.is_some_and(|val| (0..200).contains(&val)))
+                && array
+                    .iter()
+                    .any(|v| v.is_some_and(|val| (200..400).contains(&val)))
+                && array
+                    .iter()
+                    .any(|v| v.is_some_and(|val| (400..410).contains(&val)))
+                && array
+                    .iter()
+                    .any(|v| v.is_some_and(|val| (410..420).contains(&val))),
+            "Expected results from all fragments"
+        );
+    }
+
+    // Common test function for fragment list filtering
+    async fn test_fragment_list_filtering(
+        test_ds: &TestVectorDataset,
+        fragments: &[Fragment],
+        mut build_scanner: impl FnMut(&Dataset) -> Scanner,
+    ) {
+        // Test 1: Query without fragment filter - should get results from all fragments
+        let batch = build_scanner(&test_ds.dataset)
+            .try_into_batch()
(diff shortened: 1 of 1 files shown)


rust/lance/src/dataset/scanner.rs
```diff
@@ -9079,4 +9114,117 @@ mod test {
...
+            .unwrap()
+            .with_fragments(fragment_2);
+
+        let batches: Vec<_> = scanner
+            .try_into_stream()
+            .await
+            .unwrap()
+            .try_collect::<Vec<_>>()
+            .await
+            .unwrap();
```
wjones127: For tests, I'd recommend just calling `.try_into_batch()`. That makes things simpler.

```suggestion
        let batches = scanner
            .try_into_batch()
            .await
            .unwrap();
```

rust/lance/src/dataset/scanner.rs
```diff
@@ -9079,4 +9114,117 @@ mod test {
...
+        // Create FTS index on first 2 fragments
+        test_ds.make_fts_index().await.unwrap();
+
+        // Append one more fragment after indexing (this will be unindexed)
+        test_ds.append_new_data().await.unwrap();
+
+        // Now we have 3 fragments:
+        // Fragment 0: i=0..200 (indexed)
+        // Fragment 1: i=200..400 (indexed)
+        // Fragment 2: i=400..410 (unindexed)
```
wjones127: I think this test would be more convincing if:

1. Running the query without the fragment filter produced rows in all fragments
2. There was another unindexed fragment that you were excluding
3. You had another test case where you were filtering for a indexed fragment.

Could you modify the tests to cover all three of those?

yingjianwu98:
Thanks @wjones127 !

I have addressed your comments.

I am working on another PR that fix indexed_fragments filtering but let me know if you think I should put them together.

rust/lance/src/dataset/scanner.rs
```diff
@@ -9079,4 +9119,274 @@ mod test {
...
+        for idx in 0..i_array.len() {
+            has_results = true;
+            let val = i_array.value(idx);
+            assert!(
+                (400..410).contains(&val),
+                "Expected only values from fragment 2 (i=400..410), but got i={}",
+                val
+            );
+        }
+        assert!(has_results, "Expected some results from fragment 2");
```
wjones127: You could simplify these tests to something like this:
```suggestion
        let batch = scanner.try_into_batch().await.unwrap();
        assert!(batch.num_rows() > 0, "Expected some results from fragment 2");
        batch["i"].as_primitive<Int32Type>()
            .iter()
            .for_each(|val| {
                assert!(
                    (400..410).contains(&val),
                    "Expected only values from fragment 2 (i=400..410), but got i={}",
                    val
                );
            });
```

rust/lance/src/dataset/scanner.rs
```diff
@@ -9079,4 +9119,274 @@ mod test {
...
```
wjones127: I like the number of test better, but does seem to be verbose. Could you refactor these into a common test function? That would make it a lot shorter.

wjones127 (commented):
Thanks for adding those tests. I think this looks better, but would like to give you the opportunity to make the tests a bit shorter.

wjones127 (approved):

PR #5927, opened 2026-02-10
fix: don't drop field metadata on merge insert path

westonpace:
There are various ways to write data and several of them are failing for JSON data at the moment because it requires a conversion from logical to physical representation.  I'd like to rework this more generally but I want to get the current implementation working first.  This fixes one of the merge_insert paths which is currently failing because the field metadata is lost as part of the operation (and so the data appears to be a normal string column and is not converted)

### DIFF
--- a/python/python/tests/test_json.py
+++ b/python/python/tests/test_json.py
@@ -4,12 +4,13 @@
 import json
 import tempfile
 from pathlib import Path
+from typing import Union
 
 import lance
 import pyarrow as pa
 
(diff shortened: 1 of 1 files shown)


westonpace:
Possibly related to https://github.com/lance-format/lance/issues/4831 but I don't think this closes that as I don't have all merge insert paths tested yet (and I know some of them use the `Updater` which also does not properly convert schemas yet)

PR #5928, opened 2026-02-10
fix: apply SchemaAdapter in Updater

westonpace:
This will ensure JSON columns are properly converted in paths like add_columns


westonpace:
Leaving as draft until https://github.com/lance-format/lance/pull/5927 merges

PR #5929, opened 2026-02-10
fix: invalidate index fragment bitmaps after data replacement and stale merge

### LINKED ISSUE #5321: Getting non-existent fragment error in merge_insert
On some merge insert operations, you can get: `Execution error: The input to a take operation specified fragment id 0 but this fragment does not exist in the dataset`.

Originally reported in https://github.com/lancedb/lancedb/issues/2751

## Reproducing

You can reproduce by:

1. Create a dataset with one fragment and a BTREE index
2. Upsert with (a) columns out-of-order, (b) all rows in existing data get updated, and (c) some new rows
3. Call compaction
4. Call optimize_indices
5. Try to upsert again, with one of the updates including a key that was in the original data in step (1).

```python
import lance
import pyarrow as pa

data = pa.table({'id': [1], 'value': ['a']})
ds = lance.write_dataset(data, 'memory://', max_rows_per_file=1)
ds.create_scalar_index('id', 'BTREE')

data_reversed = pa.table({'value': ['b', 'a'], 'id': [2, 1] })
(
    ds.merge_insert(on='id')
    .when_matched_update_all()
    .execute(data_reversed)
)

ds.optimize.compact_files()
ds.optimize.optimize_indices()

(
    ds.merge_insert(on='id')
    .when_matched_update_all()
    .execute(data)
)
```
```
OSError: Query Execution error: Execution error: The input to a take operation specified fragment id 0 but this fragment does not exist in the dataset
```

## Diagnosis

By calling `merge_insert` with columns out-of-order, you trigger the partial schema merge insert code path. This performs updates by rewriting fragments in place.

However, doing this creates a weird conundrum for the fragment bitmap: if we updated indexed values, we need to tell the index that it shouldn't point to this fragment anymore. At the same time, the `fragment_bitmap` must contain all fragment ids present in the set of row ids, even if that fragment has been removed.

This same problem will also likely happen for operations like `DataReplacement`.

I think we need to add a new field for indices to differentiate an immutable `original_fragment_bitmap` and `current_fragment_bitmap`. I'll create an issue for that.

wjones127:
This fixes a bug where queries or merge insert could result in the error `this fragment does not exist in the dataset`.

This would happen because the index referenced row addresses that were no longer in the dataset.

Here's one way this would happen:

* Create a BTREE index covering 2 fragments
* Do a merge insert that rewrites the index column in place. This removes the updated data from the index's fragment bitmap, so data is now hidden.
* Do compaction. This will move the rows to new fragments.
* Optimize the BTREE index: this would build the btree index with the old data, not filtering out the invalidated rows. After doing this the invalidated rows, previously masked out, are part of the index. But since compaction happened, those old invalidated rows moved.

This PR makes sure:

1. When incrementally updating indexes, we only read row ids for data in the fragment bitmap.
2. Make sure we remove fragments from the index fragment bitmap if we do a DataReplacement on those fields.

This rules are described in the spec at https://lance.org/format/table/index/#loading-an-index

Fixes #5321

### DIFF
--- a/rust/lance-index/src/scalar/btree.rs
+++ b/rust/lance-index/src/scalar/btree.rs
@@ -1355,8 +1355,7 @@ fn filter_row_ids_by_fragments(
     let schema = stream.schema();
     let filtered = stream.map(move |batch_result| {
         let batch = batch_result?;
-        let row_ids = batch
-            .column(1)
+        let row_ids = batch[ROW_ID]
             .as_any()
             .downcast_ref::<arrow_array::UInt64Array>()
             .expect("expected UInt64Array for row_id column");


rust/lance-index/src/scalar.rs
```diff
@@ -888,10 +889,14 @@ pub trait ScalarIndex: Send + Sync + std::fmt::Debug + Index + DeepSizeOf {
     ) -> Result<CreatedIndex>;
 
     /// Add the new data into the index, creating an updated version of the index in `dest_store`
+    ///
+    /// If `valid_old_fragments` is provided, old index data for fragments not in the bitmap
+    /// will be filtered out during the merge.
```
westonpace, line 894: So this only matters for btree because it is the only index that scans the old portion of the index instead of rescanning the whole column?

rust/lance-index/src/scalar/btree.rs
```diff
@@ -1337,6 +1345,30 @@ impl BTreeIndex {
...
+/// ID in the upper 32 bits.
+fn filter_row_ids_by_fragments(
+    stream: SendableRecordBatchStream,
+    valid_fragments: RoaringBitmap,
+) -> SendableRecordBatchStream {
+    let schema = stream.schema();
+    let filtered = stream.map(move |batch_result| {
+        let batch = batch_result?;
+        let row_ids = batch
+            .column(1)
```
westonpace: Minor nit: I've been trying to slowly switch over to using string constants like `BTREE_IDS_COLUMN` and avoiding integral indexing just because it seems less prone to error in the future.

wjones127:
> It makes me wonder if the update path is even worth the complexity it brings at this point. Btree is the only one that really makes use of it.

That's a fair point. There's probably not a huge advantage to scanning the index vs scanning the fragments.

rust/lance-index/src/scalar/btree.rs
wjones127: I can try that.

rust/lance-index/src/scalar.rs
wjones127: Actually, bitmap does this too. I need to fix that.

wjones127: Fixed bitmap. Other indices should be done as a follow up, but I'm not aware of a code path where they cause a bug yet. Many of them like ZoneMap and BloomFilter given inexact results.

PR #5931, opened 2026-02-11
feat(java): support session

jackye1995:
This allows Java to also pass in a Session shared across datasets similar to python and rust. Session can then be used for engine side caching implementation in Spark and Trino

### DIFF
--- a/java/src/main/java/org/lance/Dataset.java
+++ b/java/src/main/java/org/lance/Dataset.java
@@ -353,6 +353,9 @@ public class Dataset implements Closeable {
     dataset.selfManagedAllocator = selfManagedAllocator;
     if (effectiveSession != null) {
       dataset.session = effectiveSession;
+    } else {
+      dataset.session = Session.fromHandle(dataset.nativeGetSessionHandle());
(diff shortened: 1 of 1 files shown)


java/src/main/java/org/lance/Dataset.java
```diff
@@ -954,6 +973,33 @@ public long countRows(String filter) {
...
+   *
+   * @return the session associated with this dataset
+   */
+  public Session session() {
+    try (LockManager.ReadLock readLock = lockManager.acquireReadLock()) {
+      Preconditions.checkArgument(nativeDatasetHandle != 0, "Dataset is closed");
+      if (session == null) {
+        long handle = nativeGetSessionHandle();
+        if (handle != 0) {
+          session = Session.fromHandle(handle);
```
majin1102: This might cause dangling references without write lock. That said, Rust's structures are shared and close() seems to work fine. Up to you.

jackye1995: good point, I updated to just load it at build time

PR #5934, opened 2026-02-11
feat: make geodatafusion/geoarrow optional via `geo` feature flag

apoc:
Make geospatial dependencies (geodatafusion, geoarrow-array, geoarrow-schema, geo-traits, geo-types) optional across the lance crate stack via a new `geo` feature flag.

This allows consumers that don't need spatial indexing or geospatial UDFs to avoid pulling in ~40 transitive dependencies and the associated runtime overhead of registering geo UDFs in every DataFusion SessionContext.

The `geo` feature is enabled by default in the top-level `lance` crate, preserving backward compatibility. Consumers can opt out with `default-features = false`.

Crates modified:
- lance-geo: all geo deps made optional, bbox module gated
- lance-datafusion: lance-geo made optional, geo UDF registration gated
- lance-index: lance-geo + geoarrow deps optional, rtree module and RTreeIndexPlugin registration gated, geo bench requires feature
- lance: geo feature added to defaults, propagates to sub-crates


jackye1995 (approved):
thanks for adding this, mostly looks good to me, pending CI

apoc:
I am working on more gracefull error reporting when user uses GEO functions when GEO is featured out

PR #5935, opened 2026-02-11
perf: reduce peak memory in nullable training data sampling

wkalt:
When we sample data for IVF training we go down one of two paths, depending on whether the target column is nullable or not. For the not-nullable path, we use dataset.sample. For the nullable path, we do a block-level sampling of batches, then filter out all batches that contain all-null rows, and interleave the result into the output.

The interleaving step requires two copies of the filtered batches to be held in RAM.

This commit adds a specialization for the case of nullable fixed-length array columns. We now pre-size an output vector, and process the input scan as a stream. For each batch, we copy not-null values into the output vector, and drop the batch. This saves one copy of the filtered output and halves the memory consumption of the sampling step.

### DIFF
--- a/rust/lance/src/index/vector/utils.rs
+++ b/rust/lance/src/index/vector/utils.rs
@@ -298,8 +298,15 @@ pub async fn maybe_sample_training_data(
 
     let should_sample = num_rows > sample_size_hint;
     if should_sample {
-        sample_training_data(dataset, column, sample_size_hint, num_rows, vector_field, is_nullable)
-            .await
+        sample_training_data(
+            dataset,
+            column,
+            sample_size_hint,
+            num_rows,
+            vector_field,
+            is_nullable,
+        )
+        .await
     } else {
         // too small to require sampling
         let batch = scan_all_training_data(dataset, column, is_nullable).await?;
@@ -394,15 +401,24 @@ async fn sample_training_data(
 
     match vector_field.data_type() {
         DataType::FixedSizeList(_, _) if !is_nullable => {
-            sample_fsl_uniform(dataset, column, sample_size_hint, num_rows, byte_width, vector_field)
-                .await
+            sample_fsl_uniform(
+                dataset,
+                column,
+                sample_size_hint,
+                num_rows,
+                byte_width,
+                vector_field,
+            )
+            .await
         }
         DataType::FixedSizeList(_, _) => {
-            let scan = sample_training_data_scan(dataset, column, sample_size_hint, num_rows, byte_width)?;
+            let scan =
+                sample_training_data_scan(dataset, column, sample_size_hint, num_rows, byte_width)?;
             sample_nullable_fsl(column, sample_size_hint, byte_width, vector_field, scan).await
         }
         _ => {
-            let scan = sample_training_data_scan(dataset, column, sample_size_hint, num_rows, byte_width)?;
+            let scan =
+                sample_training_data_scan(dataset, column, sample_size_hint, num_rows, byte_width)?;
             sample_nullable_fallback(column, sample_size_hint, is_nullable, scan).await
         }
     }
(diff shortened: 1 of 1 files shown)


wkalt:
progress

Here is a before/after comparison on top of the rest of the outstanding optimizations, for 100M 384d nullable vectors.

wkalt:
I think the failures are unrelated

rust/lance/src/index/vector/utils.rs
```diff
@@ -293,131 +296,322 @@ pub async fn maybe_sample_training_data(
...
+                sample_nullable_fsl(
+                    column,
+                    sample_size_hint,
+                    byte_width,
+                    vector_field,
+                    array,
+                    scan,
+                )
+                .await
+            }
```
westonpace: If you're just trying to figure out the column type then couldn't you use `vector_field` instead of grabbing the first batch?

rust/lance/src/index/vector/utils.rs
```diff
@@ -293,131 +296,322 @@ pub async fn maybe_sample_training_data(
...
 
-    pub fn get_partition_mutex(&self, partition_id: usize) -> Arc<Mutex<()>> {
-        let mtx = &self.partition_locks[partition_id];
+/// Build a FixedSizeListArray from raw flat value bytes.
+fn fsl_values_to_array(
+    field: &lance_core::datatypes::Field,
+    values_buf: &[u8],
+    num_rows: usize,
+) -> Result<FixedSizeListArray> {
+    let (inner_field, dim) = match field.data_type() {
```
westonpace, line 450: If you're asserting fixed size list you could use the `as_fixed_size_list_array` method.

rust/lance/src/index/vector/utils.rs
```diff
@@ -293,131 +296,322 @@ pub async fn maybe_sample_training_data(
...
+        .ok_or_else(|| Error::Index {
+            message: format!(
+                "FixedSizeList inner type {:?} has no fixed width",
+                inner_field.data_type()
+            ),
+            location: location!(),
+        })?;
+
+    let expected_bytes = num_rows * dim * elem_size;
+    let buf = Buffer::from(&values_buf[..expected_bytes]);
```
westonpace: I'm pretty sure this is a copy.  If you change `values_buf` from `&[u8]` to `Vec<u8>` then there should be a zero-copy way to create a `Buffer` from `Vec<u8>`.  Might help shave off a bit more RAM

rust/lance/src/index/vector/utils.rs
```diff
@@ -293,131 +296,322 @@ pub async fn maybe_sample_training_data(
...
+    byte_width: usize,
+) -> Result<()> {
+    // Always filter to both remove nulls and produce a zero-offset array.
+    // When there are no nulls this is just a copy, which is cheap relative
+    // to the I/O cost of reading each batch.
+    let mask = match array.nulls() {
+        Some(nulls) => arrow_array::BooleanArray::from(nulls.inner().clone()),
+        None => arrow_array::BooleanArray::from(vec![true; array.len()]),
+    };
+    let filtered = arrow::compute::filter(array, &mask)?;
```
westonpace: This is technically an avoidable copy but since your main goal is saving RAM it probably isn't a huge deal.  You could instead iterate through and make multiple copies to `extend_from_slice` (for each range of non-null values) but that would be considerably more complex.

westonpace (approved):
Some minor suggestions but otherwise it seems fine.

wkalt:
I have pushed one more commit onto here to stop calling dataset.sample for the non-nullable case, and instead also use a streaming approach. I have two datasets, one with a nullable vector column and one with a non-nullable one.

The nullable one, before and after the final commit. This shows a negligible speedup:
nullable

The not-null one, before and after. This shows much more predictable memory usage:
notnull

The two datasets are different sizes but in both cases the total memory peak is very slightly over one copy of the size of the requested data sample, during the IVF training stage.

PR #5939, opened 2026-02-11
perf: create local writer for efficient local writes

wkalt:
This creates a new LocalWriter that wraps tokio::fs::File in a BufWriter for local file writes. ObjectStore::create() now returns one of these when working against local storage, and an ObjectWriter for remote storage.

Prior to this commit, local writes (e.g for shuffling) went through a local object writer implementation that required a 5MB buffer per writer and also simulated multipart upload machinery. For local writing, this is slower than necessary and uses a lot of memory in situations where many writers are open at once.

This change results in a substantial memory reduction and incremental speedup for IVF shuffle.

### DIFF
--- a/rust/lance-io/src/object_writer.rs
+++ b/rust/lance-io/src/object_writer.rs
@@ -575,20 +575,25 @@ impl Writer for LocalWriter {
                 location!(),
             )
         })?;
-        temp_path.persist(&final_path).map_err(|e| {
-            Error::io(
-                format!("failed to persist temp file to {}: {}", final_path, e.error),
-                location!(),
-            )
-        })?;
+        let path_clone = self.path.clone();
+        let e_tag = tokio::task::spawn_blocking(move || -> Result<String> {
+            temp_path.persist(&final_path).map_err(|e| {
+                Error::io(
+                    format!("failed to persist temp file to {}: {}", final_path, e.error),
+                    location!(),
+                )
+            })?;
 
-        let metadata = std::fs::metadata(&final_path).map_err(|e| {
-            Error::io(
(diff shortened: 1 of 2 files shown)


wkalt:
This supersedes https://github.com/lance-format/lance/pull/5907. It maintains the same performance WRT memory and yields an incremental speedup. 
progress

rust/lance-io/src/object_writer.rs
```diff
@@ -498,6 +484,146 @@ impl Writer for ObjectWriter {
...
+                location!(),
+            )
+        })?;
+
+        let metadata = std::fs::metadata(&final_path).map_err(|e| {
+            Error::io(
+                format!("failed to read metadata for {}: {}", self.path, e),
+                location!(),
+            )
+        })?;
```
wjones127: I think these should be wrapped in a `tokio::task::spawn_blocking()` since they will both make blocking sys calls, right?

rust/lance-io/Cargo.toml
```diff
@@ -62,6 +63,10 @@ pprof.workspace = true
 name = "scheduler"
 harness = false
 
+[[bench]]
+name = "write"
+harness = false
+
```
wjones127: I didn't find this benchmark useful, we can remove it.
```suggestion
```

rust/lance-io/benches/write.rs
wjones127, line 1: Let's remove this bench.

rust/lance-io/src/object_writer.rs
wkalt: thanks, updated + one more spot

PR #5944, opened 2026-02-12
fix: correct OR null handling for BlockList|BlockList

fenfeng9:
`BlockList|BlockList` OR handled NULLs incorrectly: when one side was TRUE and the other was NULL, the result could stay NULL, leading to wrong query results.

Fix by computing FALSE rows explicitly and deriving NULL rows from three‑valued logic.


fenfeng9:
Hi @wjones127, could you please review this? Thanks!

PR #5945, opened 2026-02-12
refactor: deprecate list_indices and migrate tests to describe_indices

### LINKED ISSUE #5237: Deprecate list_indices
In https://github.com/lancedb/lance/pull/5221 we introduce `describe_indices` which should replace `list_indices`.  I'm creating this issue to deprecate `list_indices` which is actually a bit of a task as it is used in many tests.

fenfeng9:
closes #5237

Deprecate `list_indices` and switch tests/benchmarks to `describe_indices`/`index_stats`.

### DIFF
--- a/python/python/lance/dataset.py
+++ b/python/python/lance/dataset.py
@@ -647,12 +647,11 @@ class LanceDataset(pa.dataset.Dataset):
         list index information and index_statistics() to get the statistics for
         individual indexes of interest.
         """
-        # TODO: https://github.com/lancedb/lance/issues/5237 deprecate this method
-        # warnings.warn(
-        #     "The 'list_indices' method is deprecated.  It may be removed in a future"
-        #     "version.  Use describe_indices() instead.",
-        #     DeprecationWarning,
-        # )
+        warnings.warn(
+            "The 'list_indices' method is deprecated. It may be removed in a future "
+            "version. Use describe_indices() instead.",
+            DeprecationWarning,
+        )
 
         return self._ds.load_indices()
 
@@ -670,7 +669,7 @@ class LanceDataset(pa.dataset.Dataset):
 
     @property
     def has_index(self):
-        return len(self.list_indices()) > 0
+        return len(self.describe_indices()) > 0
 
     def _apply_default_scan_options(self, builder: ScannerBuilder):
         if self._default_scan_options:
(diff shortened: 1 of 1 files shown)


fenfeng9:
@westonpace, could you please review this when you have a chance? Thanks!

python/python/lance/dataset.py
```diff
@@ -3931,9 +3930,19 @@ def _default_vector_index_for_column(self, column: str) -> str:
...
+        lance_field = self._ds.lance_schema.field_case_insensitive(column)
+        if lance_field is None:
+            raise KeyError(f"No IVF index for column '{column}'")
+        field_id = lance_field.id()
+
+        indices = self.describe_indices()
+        for idx in indices:
+            if field_id in idx.fields:
+                # Use index_stats to get the concrete IVF subtype.
+                index_type = self.stats.index_stats(idx.name).get("index_type", "")
```
westonpace, line 3943: I think `index_stats` can be an expensive call but maybe not.  I wonder if we could further refine this to only consider columns that have a list or fixed-size-list data type?  Either way though, this is an improvement over the past behavior so I think it's ok.

fenfeng9:
> I think the main thing missing in `describe_indices` is the ability to figure out what type of vector index an index has (e.g. flat or hnsw, ivf, pq or sq, etc.) It would be nice if we wrote that information in the vector index details.

Adding `index_type` to `VectorIndexDetails` seems straightforward. Should I open a follow-up issue?
Also, do we need to include more information in `VectorIndexDetails`, such as `distance_type` and `dimension`?

westonpace:
> Should I open a follow-up issue?

Yes, this would probably be best done in a follow-up.

> Also, do we need to include more information in VectorIndexDetails, such as distance_type and dimension?

Yes.  Anything that isn't too large.  It goes in the manifest and we don't want the manifest to get too big but 1-2KB per index should be acceptable (1-2KB isn't a formal guideline, just something off the top of my head).  We can easily fit the distance_type and dimension in (and number of partitions and subvectors).  I would avoid things like the IVF centroids or PQ codebook.

fenfeng9:
> > Should I open a follow-up issue?
> 
> Yes, this would probably be best done in a follow-up.
> 
> > Also, do we need to include more information in VectorIndexDetails, such as distance_type and dimension?
> 
> Yes. Anything that isn't too large. It goes in the manifest and we don't want the manifest to get too big but 1-2KB per index should be acceptable (1-2KB isn't a formal guideline, just something off the top of my head). We can easily fit the distance_type and dimension in (and number of partitions and subvectors). I would avoid things like the IVF centroids or PQ codebook.

Created follow‑up issue #5963. I’ll follow up when I have time.

westonpace:
Thanks again!

PR #5948, opened 2026-02-12
feat: add tag parameter to commit API

beinan:
## Summary

- Add a `tag` parameter to `CommitBuilder`, `WriteParams`, `write_dataset()`, and `LanceDataset.commit()` so users can create a tag in the same call as committing data
- The tag is created after the commit succeeds; if tag creation fails (e.g. duplicate name), the committed data is preserved but an error is returned
- Eliminates the need for a separate `dataset.tags().create()` call, closing the window where a version exists without its intended tag

## Test plan

- [x] Rust unit test: `test_commit_with_tag` in `commit.rs` — verifies tag creation, no-tag regression, and duplicate tag error
- [x] Python test: `test_tag_on_write` in `test_dataset.py` — verifies `write_dataset` with tag, append with tag, `LanceDataset.commit` with tag, and duplicate tag error handling
- [x] `cargo check --workspace --tests --benches` passes
- [x] `cargo clippy --all --tests -- -D warnings` passes
- [x] Existing `test_tag` and `test_tag_order` tests pass (no regressions)

### DIFF
--- a/python/python/lance/dataset.py
+++ b/python/python/lance/dataset.py
@@ -3396,6 +3396,7 @@ class LanceDataset(pa.dataset.Dataset):
         *,
         commit_message: Optional[str] = None,
         enable_stable_row_ids: Optional[bool] = None,
+        tag: Optional[str] = None,
     ) -> LanceDataset:
         """Create a new version of dataset
 
@@ -3462,6 +3463,11 @@ class LanceDataset(pa.dataset.Dataset):
             row IDs assign each row a monotonically increasing id that persists
             across compaction and other maintenance operations.  This option is
             ignored for existing datasets.
+        tag: str, optional
+            If set, a tag with this name will be created pointing to the committed
+            version. The tag is created after the commit succeeds. If the tag
+            creation fails (e.g. because a tag with that name already exists),
+            the data will still be committed but an error will be raised.
 
         Returns
         -------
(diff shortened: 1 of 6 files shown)


wjones127:
> * Eliminates the need for a separate `dataset.tags().create()` call, closing the window where a version exists without its intended tag

I'm not sure this is worth it. It's not clear to me it will be any faster than calling both APIs in sequence. And adding this adds some complexity to the API and implementation.

Also, I worry that putting tag in the commit API gives the impression the addition of the tag and the commit are together somehow atomic. But there still is some intermediate state where another reader could see the new version without seeing the tag, right?

beinan:
> > * Eliminates the need for a separate `dataset.tags().create()` call, closing the window where a version exists without its intended tag
> 
> I'm not sure this is worth it. It's not clear to me it will be any faster than calling both APIs in sequence. And adding this adds some complexity to the API and implementation.
> 
> Also, I worry that putting tag in the commit API gives the impression the addition of the tag and the commit are together somehow atomic. But there still is some intermediate state where another reader could see the new version without seeing the tag, right?

Good points.

* In our AI training workflows, the tag is semantically critical — consumers read by tag, not version. Conceptually, “commit + tag” is one logical publish step, and ideally it should be atomic so readers see either both or neither.

* This PR does not make it atomic. There is still an intermediate state where the new version is visible without the tag. Making it truly atomic would require deeper storage-level changes.

For now, this is mainly about expressing intent and reducing the race window in user code. If the API is acceptable to the community, I’d be interested in exploring a fully atomic design in a follow-up.

beinan:
Hi @wjones127! I've written up a proposal at https://github.com/lance-format/lance/discussions/5952 addressing the need for strict ACID guarantees when tagging and indexing on S3/GCS.

The goal is to eliminate orphaned tags and "performance cliffs" by making dataset updates a single atomic transaction. Could you take a look? Your feedback on the storage layout would be super valuable!

PR #5950, opened 2026-02-13
fix: remove unnecessary column projection for count aggregate

jackye1995:
Make sure we always only do a metadata projection to avoid scanning data when doing a count. 

Also:
1. remove `create_count_plan` since `count_rows` is now just a case of aggregate, no longer need a dedicated query plan.
2. remove duplicated calls to `aggregate_required_columns` by storing required columns directly at aggregate construction time.

### DIFF
--- a/java/lance-jni/src/blocking_scanner.rs
+++ b/java/lance-jni/src/blocking_scanner.rs
@@ -381,7 +381,7 @@ fn inner_create_scanner<'local>(
 
     let substrait_aggregate_opt = env.get_bytes_opt(&substrait_aggregate_obj)?;
     if let Some(substrait_aggregate) = substrait_aggregate_opt {
-        scanner.aggregate(AggregateExpr::substrait(substrait_aggregate));
+        scanner.aggregate(AggregateExpr::substrait(substrait_aggregate))?;
     }
 
     let scanner = BlockingScanner::create(scanner);
--- a/python/src/dataset.rs
+++ b/python/src/dataset.rs
@@ -1094,7 +1094,9 @@ impl Dataset {
                 .map_err(|err| PyValueError::new_err(err.to_string()))?;
         }
         if let Some(aggregate_bytes) = substrait_aggregate {
-            scanner.aggregate(AggregateExpr::substrait(aggregate_bytes));
+            scanner
+                .aggregate(AggregateExpr::substrait(aggregate_bytes))
+                .map_err(|err| PyValueError::new_err(err.to_string()))?;
         }
         let scan = Arc::new(scanner);
         Ok(Scanner::new(scan))


python/python/tests/test_dataset.py
```diff
@@ -1168,19 +1168,10 @@ def test_count_rows_via_scanner(tmp_path: Path):
     ds = lance.write_dataset(pa.table({"a": range(100), "b": range(100)}), tmp_path)
 
     assert ds.scanner(filter="a < 50", columns=[], with_row_id=True).count_rows() == 50
-
-    with pytest.raises(
-        ValueError, match="should not be called on a plan selecting columns"
-    ):
-        ds.scanner(filter="a < 50", columns=["a"], with_row_id=True).count_rows()
```
jackye1995, line 1175: I think it makes sense to actually return a value instead of failing this case. With the latest way of optimizing, it will always just do a metadata projection and avoid data scan.

jackye1995:
> P1: Performance Regression - Full Projection in COUNT(*)

Addressed in a subsequent commit

> P1: Breaking Change - Public API Removal

The function is not really used anywhere publicly.

python/python/tests/test_dataset.py
westonpace: It is doing a data scan isn't it?  `a` is not indexed and so there is no way to satisfy the count request without actually scanning the column.

I don't entirely agree that this shouldn't be an error but I also don't disagree enough to complain.  I think the only valid concern I could have is that a user doing something like... `ds.scanner(columns=["a"]).count_rows()` might think this is the same as `SELECT COUNT(a) FROM ...` (i.e. that it returns the count of non-null rows) but that's a pretty weak argument.

So...feel free to ignore this comment 😛

rust/lance-datafusion/src/aggregate.rs
```diff
@@ -13,4 +15,24 @@ pub struct Aggregate {
     /// Aggregate function expressions (e.g., SUM, COUNT, AVG).
     /// Use `.alias()` on the expression to set output column names.
     pub aggregates: Vec<Expr>,
+    /// Column names required by this aggregate (computed at construction).
+    /// For COUNT(*), this is empty. For SUM(x), GROUP BY y, this contains [x, y].
+    pub required_columns: Vec<String>,
```
westonpace: Super minor nit: I think either `Aggregate` should not have pub fields (you could add an accessor for `group_by` and `aggregates` if you wanted) or `required_columns` should just be computed at plan time and not stored as part of this struct.

The current structure (public fields) makes it seem like `Aggregate` is a "data class" and users are expected to do something like...

```
let agg = Aggregate { group_by: ..., aggregates: ..., required_columns: ... }
```

But really there is no easy way to create this other than `Aggregate::new` and `required_columns` is an implementation detail.

rust/lance/src/dataset/scanner.rs
```diff
@@ -1911,72 +1914,25 @@ impl Scanner {
...
-                vec![Arc::new(count_expr)],
-                vec![None],
-                plan,
-                plan_schema,
-            )?) as Arc<dyn ExecutionPlan>)
-        }
-        .boxed()
-    }
+            let mut scanner = self.clone();
+            scanner.aggregate(AggregateExpr::builder().count_star().build())?;
```
westonpace, line 1926: Nice cleanup!

rust/lance/src/dataset/scanner.rs
```diff
@@ -1911,72 +1914,25 @@ impl Scanner {
...
-                );
-
-                return Err(Error::invalid_input(msg, location!()));
-            }
-
             if self.limit.is_some() || self.offset.is_some() {
                 log::warn!(
                     "count_rows called with limit or offset which could have surprising results"
                 );
             }
```
westonpace: You can probably get rid of this warning at this point.  The entire operation will fail so no need to warn about it.

python/python/tests/test_dataset.py
jackye1995: oh okay, I somehow ignored the filter part when trying to reason why it was asserting failure...

> ds.scanner(columns=["a"]).count_rows() might think this is the same as SELECT COUNT(a) FROM ...

ohh I see the reasoning now, thanks for explaining. I think it is still clear, that `ds.scanner(columns=["a"]).count_rows()` is not the same as `SELECT COUNT(a)` because it should be equivalent to something like `ds.scanner(columns=["a"], filter="a IS NOT NULL").count_rows()`.

PR #5953, opened 2026-02-14
fix: respect requested indexed fragment in vector and FTS searches

yingjianwu98:
If client specify .with_fragments, vector and FTS searches on indexed fragments should respect the target fragments.

Previous PR to fix the unindexed fragments path: https://github.com/lance-format/lance/pull/5924


yingjianwu98:
@wjones127 

This is the PR fix for indexed fragments on respecting fragments passed in through `with_fragments`
Feel free to review once you get time, thanks!

PR #5954, opened 2026-02-15
feat: add proto serialization for FilteredReadExec

LuQQiu:
Add protobuf definitions and to_proto/from_proto conversion helpers so a PhysicalExtensionCodec can serialize FilteredReadExec plans for distributed execution (planner → executor).

Proto types (TableIdentifier, FilteredReadOptionsProto, FilteredReadPlanProto) live in lance-datafusion as data containers. Conversion logic lives in lance::io::exec::filtered_read_proto because it needs access to FilteredReadExec and Dataset. Filter expressions use existing Substrait serialization; RowAddrTreeMap uses its built-in binary format. Arrow types using existing Arrow IPC serialization

### DIFF
--- a/protos/filtered_read.proto
+++ b/protos/filtered_read.proto
@@ -85,9 +85,15 @@ message FilteredReadPlanProto {
 message FilteredReadExecProto {
   TableIdentifier table = 1;
   FilteredReadOptionsProto options = 2;
+  // FilteredRead has two modes
+  // Plan-then-execute (distributed): The planner creates a FilteredReadPlan and sends it to a remote executor.
+  // Plan-and-execute (local): The executor creates the plan itself at execution time.
   optional FilteredReadPlanProto plan = 3;
   // Note: FilteredReadExec.index_input (child ExecutionPlan) is NOT serialized here.
-  // DataFusion's PhysicalExtensionCodec serializes child plans separately via the
-  // plan tree traversal (children() / with_new_children()), and passes them back
-  // as the `inputs` parameter in try_decode.
+  // DataFusion's PhysicalExtensionCodec handles child plans automatically: it walks
+  // the plan tree via children() / with_new_children(), serializes each node, and
+  // passes deserialized children back as the `inputs` parameter in try_decode.
+  // This means any ExecutionPlan in the tree (including index_input) must also
+  // implement try_encode/try_decode in the PhysicalExtensionCodec.
+  // TODO: implement serialize/deserialize for lance-specific index input ExecutionPlans.
 }


protos/table_identifier.proto
```diff
@@ -0,0 +1,18 @@
...
+// Identifies a Lance dataset for remote reconstruction.
+//
+// Two modes:
+//   1. uri + serialized_manifest (fast): remote executor skips manifest read.
+//   2. uri + version + etag (lightweight): remote executor loads manifest from storage.
+message TableIdentifier {
+  string uri = 1;
+  uint64 version = 2;
+  optional string manifest_etag = 3;
+  optional bytes serialized_manifest = 4;
```
LuQQiu, line 17: @wjones127 do you think we need to serialize the manifest_etag and even send the serialized manifest itself?

rust/lance-datafusion/src/substrait.rs
```diff
@@ -27,6 +27,31 @@ use snafu::location;
 use std::collections::HashMap;
 use std::sync::Arc;
 
+/// Substrait doesn't yet support all data types.
+fn is_substrait_compatible(data_type: &DataType) -> bool {
+    match data_type {
+        DataType::Null | DataType::FixedSizeList(_, _) | DataType::Float16 => false,
```
LuQQiu: @westonpace do you know if this is still the limitation for substrait?

protos/filtered_read.proto
```diff
@@ -0,0 +1,87 @@
...
+  optional uint64 io_buffer_size_bytes = 11;
+  // Arrow IPC schema for decoding Substrait filters (may be wider than projection).
+  optional bytes filter_schema_ipc = 12;
+}
+
+// Serializable form of FilteredReadPlan (planned/distributed mode).
+// RowAddrTreeMap serialized via its built-in serialize_into/deserialize_from.
+// Per-fragment filters are Substrait-encoded and deduplicated.
+message FilteredReadPlanProto {
+  bytes row_addr_tree_map = 1;
```
hamersaw, line 71: TL;DR does it make sense to make this a `oneof` over `bytes` and `repeated U64Range`? Then we could just use the smaller representation on serialization (easy computation).

In my [plan_splits PR](https://github.com/lance-format/lance/pull/5863) I merged the `FilteredReadPlan` and `FilteredReadIntenralPlan` because IIUC the only reason we have differentiation is that serializing a bitmap is smaller than ranges. I would challenge that under the intuition (or maybe "hope") that we have many dense splits where the seriaization is actually smaller for ranges (ex. all 4k values a range is 16 bytes (u64 start, u64 end) whereas a bitmap is 4k / 8 = 500 bytes). If this is the case, then we would do a series of inefficient serializations to move to bitmap for every query.

LuQQiu: The reason that having FilteredReadPlan as a outside plan is mainly not for serialization / deserialization. I keep FilteredReadPlan with external plan using RoaringBitmap for the following reasons
* Range is ambiguous in Lance. Most user-facing APIs use logical rows (without deletions). Vec<Range<u64>> here means physical rows (with deletions). This is confusing. RowAddrTreeMap/RoaringBitmap avoids that ambiguity because it's clearly a set of physical row addresses, not a "range" that could be misread as logical.
* Ranges fragment badly with deletions. A dataset with scattered deletions turns a single range 0..4096 into many small ranges. The Vec<Range<u64>> grows proportionally to the number of deletion holes. RoaringBitmap handles this natively with compressed containers.
* Bitmap enables efficient row unselection. Whether it's 4K row group pruning or custom index results, you need to remove specific rows from the selection. Flipping bits is O(1); splitting ranges is O(n) and produces more ranges.

protos/filtered_read.proto
```diff
@@ -0,0 +1,87 @@
...
+  optional U64Range scan_range_after_filter = 2;
+  // Arrow IPC schema for decoding Substrait filters (matches the schema used at encode time).
+  optional bytes filter_schema_ipc = 3;
+  // Deduplicated filter storage: frag_id → index into filter_expressions.
+  map<uint32, uint32> fragment_filter_ids = 4;
```
westonpace: Can you expand on what the key and value represent?  I think the key is fragment id and the value is an index into `filter_expressions`?  Also, I assume that multiple keys can share the same value?

protos/filtered_read.proto
```diff
@@ -0,0 +1,87 @@
...
+  // Unique Substrait-encoded filter expressions (indexed by fragment_filter_ids values).
+  repeated bytes filter_expressions = 5;
+}
+
+// Top-level wrapper for FilteredReadExec serialization.
+message FilteredReadExecProto {
+  TableIdentifier table = 1;
+  FilteredReadOptionsProto options = 2;
+  optional FilteredReadPlanProto plan = 3;
```
westonpace, line 91: Can you explain why this is optional?

protos/filtered_read.proto
```diff
@@ -0,0 +1,87 @@
...
+  // index_input (child plan) handled by DataFusion's codec via inputs[]
```
westonpace: I'm not entirely sure what this means

rust/lance-datafusion/src/substrait.rs
westonpace: Pretty sure this is correct though maybe there is support for Float16 now in DF if I recall correctly.

rust/lance/src/io/exec/filtered_read_proto.rs
```diff
@@ -0,0 +1,854 @@
...
+    } else {
+        Ok(exec)
+    }
+}
+
+// =============================================================================
+// FilteredReadOptions <-> Proto
+// =============================================================================
+
+fn options_to_proto(
```
westonpace: Minor nit: maybe `fr_options_to_proto`?

protos/filtered_read.proto
LuQQiu: FilteredReadExec can run in two modes:
  1. Plan-then-execute (distributed): The planner creates a FilteredReadPlan and sends it to a remote executor. plan is Some.
  2. Plan-and-execute (local): The executor creates the plan itself at execution time. No pre-computed plan is needed. plan is None. I can only think of the case that you know if one executor has good locality for a dataset, you probably send the whole plan to let it execute?

protos/filtered_read.proto
LuQQiu: // Top-level wrapper for FilteredReadExec serialization.
message FilteredReadExecProto {
  TableIdentifier table = 1;
  FilteredReadOptionsProto options = 2;
  optional FilteredReadPlanProto plan = 3;
  // Note: FilteredReadExec.index_input (child ExecutionPlan) is NOT serialized here.
  // DataFusion's PhysicalExtensionCodec serializes child plans separately via the
  // plan tree traversal (children() / with_new_children()), and passes them back
  // as the `inputs` parameter in try_decode.
}


basically datafusion will give the deserialized index input to FilteredReadExec when constructing it, we don't need to deal with that.
Oh but if the index_input is also need to be sent through the network, then those execs will also need to be seralizable. I could add a comment there. 
```
PhysicalExtensionCodec.try_encode, try_decode method is also needed ... 
```

protos/filtered_read.proto
westonpace: Hmm, I don't have a good sense for where the trade-offs are for the two approaches but it sounds like you've both thought about it.  My preference is to move forward with what we've got written and revisit if serialization time seems to be a bottleneck?

PR #5955, opened 2026-02-16
feat: create an arrow-scalar crate utilizing arrow-row and arrow-data

westonpace:
This provides an `ArrowScalar` which supports:

* Implementations of `Ord`, `Hash`, and `Eq`
* Stable serialization to/from binary

The `Ord`/`Eq` implementations come from `arrow_row` which should provide the same comparison logic as `arrow-rs` compute functions (and thus Datafusion)

The binary serialization uses the C data interface for optional serialization of the data type and then dumps the array data as-is.  This should be a stable format.

My primary goal with this PR was to minimize the amount of code required (compared to something like datafusion's `ScalarValue` to keep maintenance simple.  There may be more performant options but since I don't expect scalars to be used in hot spots I think this will be ok.  Also, the public API surface of `ArrowScalar` is pretty small so if we need to replace the internal implementation later for performance we can do so.

### DIFF
--- a/rust/arrow-scalar/src/lib.rs
+++ b/rust/arrow-scalar/src/lib.rs
@@ -545,7 +545,7 @@ mod prop_tests {
             }),
             proptest::collection::vec(
                 proptest::option::of(proptest::collection::vec(any::<u8>(), 0..50)),
-                len.clone(),
+                len,
             )
             .prop_map(|v| {
                 let refs: Vec<Option<&[u8]>> = v.iter().map(|o| o.as_deref()).collect();


rust/arrow-scalar/Cargo.toml
```diff
@@ -0,0 +1,27 @@
...
+license.workspace = true
+repository.workspace = true
+description = "Arrow scalar type with Ord, Hash, and Eq support"
+keywords.workspace = true
+categories.workspace = true
+rust-version.workspace = true
+readme = "README.md"
+
+[dependencies]
+arrow-array = { workspace = true }
```
wjones127, line 17: If you want to be defensive, might want to leave a brief comment that you want to keep the dependency list here small.

wjones127 (approved):
Looks reasonable and small.

PR #5958, opened 2026-02-17
feat: add progress monitoring via callbacks for inverted indexes

vivek-bharathan:


rust/lance-index/src/scalar/inverted/builder.rs
```diff
@@ -209,6 +221,9 @@ impl InvertedIndexBuilder {
         while let Some(num_rows) = stream.try_next().await? {
             total_num_rows += num_rows;
             if total_num_rows >= last_num_rows + 1_000_000 {
+                self.progress
```
westonpace: Do you think this updates progress often enough?  E.g. is 1_000_000 a good value for periodic progress reporting or should it be smaller for a progress bar?

rust/lance-index/src/scalar/registry.rs
```diff
@@ -116,6 +117,28 @@ pub trait ScalarIndexPlugin: Send + Sync + std::fmt::Debug {
         fragment_ids: Option<Vec<u32>>,
     ) -> Result<CreatedIndex>;
 
+    /// Train a new index with progress callbacks.
+    ///
+    /// Plugins can override this for index-specific stages. The default implementation
+    /// wraps [`Self::train_index`] with a generic scalar training stage.
+    async fn train_index_with_progress(
```
westonpace: I appreciate what you're trying to do here.  This will provide a default implementation for indexes that don't have progress reporting today and minimize the change set.  However, I want to keep this trait relatively clean as it is a big part of lance-index's public API.  With this proposed change a caller will see both `train_index` and `train_index_with_progress` and I think it would be simpler if there was just one method.  Sadly Rust can't do a "protected" visibility here.

Can we just bite the bullet and add `progress` as a required arg to `train_index` and not create a new method?

I realize this means we need to change all the existing scalar indexes but we probably want them all to have progress reporting at some point anyways so we have a cleaner API, with a little bit of repetition in the short term, that will hopefully go away in the longer term?

rust/lance-index/src/scalar/inverted/builder.rs
vivek-bharathan: Do you have a suggestion? 10k? 100k? I just glommed onto the pre-existing logging cadence

vivek-bharathan: It looks like the default batch size is 8k. I'll reduce this to 10k

vivek-bharathan:
@westonpace - this should be ready to merge now

westonpace:
Ok, I played around with this latest build and it seems to give pretty good progress notifications.  I have one additional suggestion.  The "tokenize_docs" progress is currently reported when the data has been scanned in, not when the data has been tokenized.  Currently the tokenziation is slower than the scanning so on smaller datasets this causes there to be a hang between the last tokenize progress update and the first merge update.  Here is a slight tweak that would fix that:

https://github.com/westonpace/lance/commit/381e853a82007bc18d2a973f590b4495488be02c

Feel free to apply that or just let me know if you'd rather accept it as it is.

vivek-bharathan:
Makes sense for this to be more granular. I've included your edit

PR #5964, opened 2026-02-18
feat: update minimum supported rust version from 1.88 to 1.91

westonpace:
The AWS libraries have updated their MSRV to 1.91 and their versioning scheme makes it a little difficult to pin.  We aim to follow Datafusion's policy which is "4 minor releases and 4 months".  The release of 1.94 (currently slated for 5 Mar) would make 1.91 the minimum (1.91 is released in October so we have 4 months and this gives us 4 releases (1.91, 1.92, 1.93, 1.94).

Since 5 Mar is close I'd rather not dive into complicated pinning too much even if we are technically getting ahead of ourselves slightly.


westonpace:
The native-tls failure looks transient (maintainer yanked that version, looks like an accidental bad publish).  I've confirmed no error with 0.2.18

PR #5966, opened 2026-02-18
feat: add Dataset::with_object_store for request-scoped store overrides

wkalt:
Add Dataset::with_object_store so callers can clone a dataset bound to a caller-supplied ObjectStore. This is useful for tracking metrics on a per-request basis.

For example, if you have an object store wrapper that performs caching, you can create a freshly wrapped object store for each use of the dataset in order to collect per-request hit/miss tallies.


wjones127 (approved):
Looks good to me!

PR #5968, opened 2026-02-18
feat: support namespace as external manifest store

jackye1995:
This PR is based on discussion in https://github.com/lance-format/lance/discussions/5849

I have added 4 APIs in Lance Namespace spec:
- ListTableVersions (this already existed, but added a flag `descending` to allow return versions in descending order)
- CreateTableVersion
- DescribeTableVersion
- BatchDeleteTableVersions

With these APIs, we can fulfill the `ExternalManfiestStore` trait with a `LanceNamespaceExternalManfiestStore` by:
- get: call DescribeTableVersion
- get_latest_version: call ListTableVersions with `limit=1`, `descending=false`
- put: call CreateTableVersion (this is a new function that combines `put_if_not_exist` and `put_if_exists`)
- delete: call BatchDeleteTableVersions with range [0, -1)

We also added a basic implementation in `DirectoryNamespace` with:
- when CreateTableVersion, just put the staged manifest to the target version
- ListTableVersions, DescribeTableVersion, BatchDeleteTableVersions are just facade to storage level operations

In `DirectoryNamespace`, we added a flag `table_version_tracking_enabled`. When true, when user calls `describe_table` or `declare_table` to get table information, we return a new flag in response saying `managed_versioning=true`, indicating that the namespace should manage table version commit and resolution. Based on that information, we set the commit handler to be external manifest store commit handler, and the external manifest store to be `LanceNamespaceExternalManfiestStore`.

### DIFF
--- a/rust/lance-namespace-impls/src/dir.rs
+++ b/rust/lance-namespace-impls/src/dir.rs
@@ -4308,6 +4308,7 @@ mod tests {
     }
 
     #[tokio::test]
+    #[cfg(not(windows))]
     async fn test_create_table_version() {
         use futures::TryStreamExt;
         use lance::dataset::builder::DatasetBuilder;
@@ -4418,6 +4419,7 @@ mod tests {
     }
 
     #[tokio::test]
+    #[cfg(not(windows))]
     async fn test_create_table_version_conflict() {
         // create_table_version should fail if the version already exists.
         // Each version always writes to a new file location.
@@ -4771,6 +4773,7 @@ mod tests {
         }
 
         #[tokio::test]
+        #[cfg(not(windows))]
         async fn test_external_manifest_store_invokes_namespace_apis() {
             use arrow::array::{Int32Array, StringArray};
             use arrow::datatypes::{DataType, Field, Schema as ArrowSchema};


rust/lance-table/src/io/commit/external_manifest.rs
```diff
@@ -90,6 +90,89 @@ pub trait ExternalManifestStore: std::fmt::Debug + Send + Sync {
...
+    /// Put the manifest to the external store.
+    ///
+    /// The staging manifest has been written to `staging_path` on the object store.
+    /// This method should atomically claim the version and return the final manifest location.
+    ///
+    /// The default implementation uses put_if_not_exists and put_if_exists to
+    /// implement a staging-based workflow. Implementations that can write directly
+    /// (e.g., namespace-backed stores) should override this method.
+    #[allow(clippy::too_many_arguments)]
+    async fn put(
```
jackye1995, line 102: this is probably a good middle ground between external manifest store today vs doing a whole new manifest store. We will have a single generic `put` to put a manifest to the store that can be overridden. It gives more freedom compared to having to follow exactly the same flow with `put_if_not_exist` and then `put_if_exists`

python/python/tests/test_namespace_dir.py
```diff
@@ -720,3 +728,169 @@ def test_connect_with_storage_options(self):
...
+        with self._lock:
+            self.list_table_versions_count += 1
+        request_dict = json.loads(request_json)
+        response_dict = self._inner.list_table_versions(request_dict)
+        return json.dumps(response_dict)
+
+
+@pytest.mark.skipif(
+    sys.platform == "win32",
+    reason="External manifest store has known issues on Windows",
```
jackye1995, line 818: looks like external manifest store code path in general has compatibility issue with windows, skip related tests for now and fix in a separated PR

PR #5973, opened 2026-02-21
feat: serialize storage options in table identifier proto

LuQQiu:
add storage options in table identifier proto to allow pass in storage credentials or information.
Make `dataset` parameter optional in `filtered_read_exec_from_proto`, falls back to opening from the proto's table identifier

### DIFF
--- a/rust/lance/src/io/exec/filtered_read_proto.rs
+++ b/rust/lance/src/io/exec/filtered_read_proto.rs
@@ -24,6 +24,7 @@ use lance_core::utils::mask::RowAddrTreeMap;
 use lance_core::{Error, Result};
 use lance_datafusion::pb;
 use lance_datafusion::substrait::{encode_substrait, parse_substrait, prune_schema_for_substrait};
+use lance_io::object_store::StorageOptions;
 use lance_table::format::Fragment;
 use prost::Message;
 use snafu::location;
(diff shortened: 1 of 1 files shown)


rust/lance/src/io/exec/filtered_read_proto.rs
```diff
@@ -60,7 +65,25 @@ pub fn table_identifier_from_dataset_with_manifest(dataset: &Dataset) -> pb::Tab
         version: dataset.manifest.version,
         manifest_etag: dataset.manifest_location.e_tag.clone(),
         serialized_manifest: Some(manifest_proto.encode_to_vec()),
+        storage_options: dataset
+            .initial_storage_options()
```
jackye1995: I think this should call latest_storage_options()?

protos/table_identifier.proto
jackye1995, line 13: I feel at this point it's more like a table state than table identifier, not sure if we can still change ethe name though

LuQQiu: emmm, we mention not make the proto backward compatible, this may be merged within one week... emm

rust/lance/src/io/exec/filtered_read_proto.rs
LuQQiu: oh that's good!

PR #5974, opened 2026-02-21
fix: don't use buffered without spawn in fts index training

westonpace:
When training an FTS index we load all of the partitions and merge them.  The code was setup to load partitions in parallel.  However, there was no `spawn` call so it wasn't actually loading anything in parallel.  This lead to starvation on HEAD calls.  The order of operations (everything was serialized) was something like...

```
Start load of part 1 file 1
Start load of part 2 file 1
...
Start load of part N file 1
Finish load and do CPU work to load file 1
Start load of part 1 file 2
Finish load and do CPU work to load file 2
...
```

The load request of part N file 1 would not get polled again until a lot of CPU work was done (to parse files 1-N-1).  This resulted in the HEAD request being starved which looked like an S3 timeout.

This fix is the safest fix which just removes the call to buffered.

A more aggressive fix would be to add a spawn call but that could have impacts on RAM usage and I think https://github.com/lance-format/lance/issues/5970 would be an even better approach anyways.  I can quickly test the spawn fix on Monday and put up a PR for that later if it looks good but wanted to get something in now.

### DIFF
--- a/rust/lance-index/src/scalar/inverted/merger.rs
+++ b/rust/lance-index/src/scalar/inverted/merger.rs
@@ -2,8 +2,7 @@
 // SPDX-FileCopyrightText: Copyright The Lance Authors
 
 use fst::Streamer;
-use futures::{stream, StreamExt, TryStreamExt};
-use lance_core::{cache::LanceCache, utils::tokio::get_num_compute_intensive_cpus, Error, Result};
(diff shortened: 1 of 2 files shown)


westonpace:
Closing in favor of https://github.com/lance-format/lance/pull/5977

PR #5978, opened 2026-02-21
feat: add URI-based commit support to Java SDK

hamersaw:
- adds new `CommitBuilder` / `Transaction` classes to reflect the rust SDK (works over both Dataset and URI)
- renames the current `Transaction` to `SourcedTransaction` to leave current Dataset based SDK intact, but internally everything now uses `CommitBuilder`
- updates tests to use `CommitBuilder` route rather than `SourcedTransaction`
- fills out SDKs to ensure full compatiblity (all fields / etc)

To create a dataset with existing fragments at a new URI:
```
  // 1. Write fragments to the dataset path
  FragmentMetadata fragmentMeta = testDataset.createNewFragment(20);

  // 2. Build a Transaction with Overwrite (provides schema for the new dataset)
  try (Transaction txn =
      new Transaction.Builder()
          .operation(
              Overwrite.builder()
                  .fragments(Collections.singletonList(fragmentMeta))
                  .schema(schema)
                  .build())
          .build()) {
    // 3. Commit to a URI where no dataset exists yet
    try (Dataset dataset = new CommitBuilder(datasetPath, allocator).execute(txn)) {
      assertEquals(1, dataset.version());
      assertEquals(20, dataset.countRows());
    }
  }
```

Closes https://github.com/lance-format/lance/issues/5969


java/src/main/java/org/lance/CommitBuilder.java
```diff
@@ -0,0 +1,302 @@
...
+    this.tableId = tableId;
+    return this;
+  }
+
+  /**
+   * Enable or disable v2 manifest paths for new datasets.
+   *
+   * <p>Defaults to true. V2 manifest paths allow constant-time lookups for the latest manifest on
+   * object storage. Warning: enabling this makes the dataset unreadable for Lance versions prior to
+   * 0.17.0.
```
wjones127, line 153: Does the `<p>` need a closing tag?

java/src/main/java/org/lance/Transaction.java
```diff
@@ -18,41 +18,60 @@
...
 import java.util.Optional;
 import java.util.UUID;
 
 /**
- * Align with the Transaction struct in rust. The transaction won't commit the status to original
- * dataset. It will return a new dataset after committed.
+ * A pure data container representing a Lance transaction.
+ *
+ * <p>A Transaction holds the read version, a unique identifier, the operation to perform, and
+ * optional transaction properties. It does not contain commit configuration or execution logic.
```
wjones127, line 30: I like that this is a data container. One key property we want for transactions is they should be easily serializable. This is because they might be send across nodes to be committed by another node. This is possible here, right?

java/src/main/java/org/lance/SourcedTransaction.java
```diff
@@ -0,0 +1,168 @@
...
+import org.lance.operation.Operation;
+
+import org.apache.arrow.util.Preconditions;
+
+import java.util.Map;
+import java.util.Optional;
+
+/**
+ * A convenience wrapper that pairs a {@link Transaction} with a {@link Dataset}, providing a simple
+ * commit workflow.
```
wjones127, line 25: This seems okay, although I'm not sure how much benefit it provides over just doing something like:

```
dataset.commit(transaction)
```

hamersaw: +1 I did this to make sure we maintained the current API so we're backwards compatible.

java/src/main/java/org/lance/Transaction.java
hamersaw: Yes, it should be!

java/src/main/java/org/lance/CommitBuilder.java
hamersaw: This is standard javadoc. Basically the first paragraph has an implicit `<p>` and the following need explicit. TIL.

PR #5979, opened 2026-02-22
feat(core): add fuzzy matching suggestions for missing columns (#5642)

### LINKED ISSUE #5642: Enhance not found errors with suggestions
We often have errors that say an input is unrecognized and there are a finite number of valid options. Here are a few top examples that come to mind:

* "column name not found in schema"
* "unknown distance type"
* "unknown index type"

When we do this, we should:

1. Provide a list of valid options
2. For string inputs, if the option needs to change fewer than 1/3 of characters (as measured by levenshtein distance), suggest a correction

For example:

```
SchemaError: column 'vacter` not found. Did you mean 'vector`?
```

I would recommend implementing the suggestion in the Display implementation instead of when creating the error. That way, we don't perform a complex calculation for the suggestion when the error might be expected and immediately handled. So you might have something like:

```rust
struct ColumnNotFoundError {
    requested: String,
    schema: Arc<Schema>,
}

impl std::fmt::Display for ColumnNotFoundError {
    fn display() {
        // Do calculation here
    }
}
```

SIDDHESH-8213:
## Description
This PR addresses issue #5642 by implementing fuzzy matching suggestions for `ColumnNotFound` errors. When a user attempts to access a field that does not exist in the schema, the engine now calculates the Levenshtein distance against available fields and suggests the closest match if it falls within a threshold (edit distance ≤ 2).

## Changes
- **Dependency Update**: Added `strsim` to `lance-core` to handle string similarity metrics.
- **Error Enhancement**: 
    - Updated `do_project` in `schema.rs` to provide "Did you mean...?" suggestions during projections.
    - Updated `FieldRef::into_id` to include similar suggestions for path-based lookups.
- **Helper Logic**: Introduced `suggest_closest_column` as a private helper function to keep the fuzzy matching logic reusable and encapsulated.

## Testing performed
- **Unit Test**: Added `test_schema_typo_suggestion` to verify both top-level projections and nested path resolutions.
- **Verification**: Confirmed the following typo resolutions pass:
    - `"vctor"` -> `"vector"`
    - `"timstamp"` -> `"timestamp"`
- **Regression**: Ran the full `lance-core` test suite to ensure no regressions in schema validation or resolution logic.

## Related Issues
Closes #5642

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4889,6 +4889,7 @@ dependencies = [
  "rstest 0.23.0",
  "serde_json",
  "snafu",
+ "strsim 0.11.1",
  "tempfile",
  "tokio",
  "tokio-stream",


wjones127:
Hey thanks for creating a PR. I think this is being concurrently worked on in #5976. You can see I have similar feedback there. Perhaps you can help review that one and work on a follow up PR once that one is complete?

Xuanwo:
Thank you @SIDDHESH-8213 for working on this! https://github.com/lance-format/lance/pull/5976 is now merged, and welcome to start a new PR if there are follow-ups to do.

PR #5981, opened 2026-02-22
perf: speed up format v2.2 scans by adding shortcut for full page

Xuanwo:
This PR addresses a long-standing issue in the Lance file format (both v2.1 and v2.2) where the rep index must be loaded before reading any full-zipped values. This could cause serious HoL blocking, especially when data is stored on low-latency, high-throughput services like S3.

@westonpace previously reported this issue at https://github.com/lance-format/lance/issues/3579. While we already support user requests to cache this index, I implemented that feature. Now I believe we should always cache it by default, as it is low-cost and highly beneficial.

Based on a full scan test using the FineWeb dataset, I observed the following improvements:

**Local dataset**
- p50: 900,363 µs → 542,266 µs (1.66x faster, –39.8%)
- p95: 951,820 µs → 579,505 µs (1.64x faster)
- p99: 994,691 µs → 713,062 µs (1.40x faster)

**S3 dataset**
- p50: 3,981,524 µs → 990,825 µs (4.02x faster, –75.1%)
- p95: 4,056,506 µs → 1,124,499 µs (3.61x faster)
- p99: 4,106,640 µs → 1,207,027 µs (3.40x faster)

Additionally, the rep index cache grows linearly; for 200k rows it occupies about 1.6 MiB. This cache will be managed by our global metadata cache. So I think it's totally ok for us to handle it.

---

This PR includes the following changes:

- always cache the repetition index when present and populate `cached_state` immediately
- split io submission so cached paths submit reads before awaiting and keep non-cached behavior for fallback
- drop unused cache flag/parameter plumbing and update full zip cache test expectations

---

**Parts of this PR were drafted with assistance from Codex (with `gpt-5.3-codex`), amp (with `claude-4.6`) and fully reviewed and edited by me. I take full responsibility for all changes.**

### DIFF
--- a/rust/lance-encoding/src/encodings/logical/primitive.rs
+++ b/rust/lance-encoding/src/encodings/logical/primitive.rs
@@ -2702,6 +2702,91 @@ impl VariableFullZipDecoder {
         decoder
     }
 
+    fn slice_batch_data_and_rebase_offsets_typed<T>(
+        data: &LanceBuffer,
+        offsets: &LanceBuffer,
+    ) -> Result<(LanceBuffer, LanceBuffer)>
+    where
+        T: arrow_buffer::ArrowNativeType
+            + Copy
+            + PartialOrd
+            + std::ops::Sub<Output = T>
+            + std::fmt::Display
+            + TryInto<usize>,
+    {
+        let offsets_slice = offsets.borrow_to_typed_slice::<T>();
+        let offsets_slice = offsets_slice.as_ref();
+        if offsets_slice.is_empty() {
+            return Err(Error::Internal {
+                message: "Variable offsets cannot be empty".to_string(),
+                location: location!(),
+            });
+        }
+
+        let base = offsets_slice[0];
+        let end = *offsets_slice.last().unwrap();
+        if end < base {
+            return Err(Error::Internal {
+                message: format!(
+                    "Invalid variable offsets: end ({end}) is less than base ({base})"
+                ),
+                location: location!(),
+            });
+        }
+
+        let data_start = base.try_into().map_err(|_| Error::Internal {
+            message: format!("Variable offset ({base}) does not fit into usize"),
+            location: location!(),
+        })?;
+        let data_end = end.try_into().map_err(|_| Error::Internal {
+            message: format!("Variable offset ({end}) does not fit into usize"),
+            location: location!(),
+        })?;
+        if data_end > data.len() {
+            return Err(Error::Internal {
+                message: format!(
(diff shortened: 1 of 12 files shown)


Xuanwo:
> Can we at least leave in the option to disable it?

Yep, will do.

> If we know we are going to load an entire page then we could shortcut and just read the entire page. Then schedule_ranges could receive a special reader that just returns slices of the page data. This way we avoid needing two stages of I/O for full scans.

Seems to be an interesting idea, will give it a try first.

Xuanwo:
Hi @westonpace, I added a `FullZipReadSource` that works really well. Thank you for your suggestion! Now we don't need to cache the data, but we can still achieve similar performance improvements.

The latest bench result:

- local p50：874138.5us -> 591128.0us，1.479x（-32.38%）
- local p95：937716.5us -> 632214.5us，1.483x（-32.58%）
- s3 p50：4027359.0us -> 1013433.5us，3.974x（-74.84%）
- s3 p95：4116329.5us -> 1091348.0us，3.772x（-73.49%）
- s3 p99：6182523.0us -> 1137305.0us，5.436x（-81.60%）

Compared to previous impls:

- local p50: 542,265.5us -> 591,128us，+9.01%
- local p95: 579,505us -> 632,214.5us，+9.10%
- local p99: 713,061.5us -> 720,853.5us，+1.09%
- s3 p50: 990,825us -> 1,013,433.5us，+2.28%
- s3 p95: 1,124,499us -> 1,091,348us，-2.95%
- s3 p99: 1,207,027us -> 1,137,305us，-5.78%

A bit slower on local but I think that's fine.

rust/lance-encoding/src/encodings/logical/primitive.rs
```diff
@@ -1913,15 +1986,14 @@ struct FullZipDecodeDetails {
...
+    data_buf_size: u64,
     rep_index: Option<FullZipRepIndexDetails>,
     priority: u64,
     rows_in_page: u64,
     bits_per_offset: u8,
     details: Arc<FullZipDecodeDetails>,
     /// Cached state containing the decoded repetition index
     cached_state: Option<Arc<FullZipCacheableState>>,
-    /// Whether to enable caching of repetition indices
-    enable_cache: bool,
```
westonpace, line 2002: Why get rid of this?  It might still be useful for users that want 1 IOP random access on relatively small amounts of data?

Xuanwo: I tried to always enable cache before and now we have a shortcut, we can enable the flag back.,

PR #5982, opened 2026-02-23
perf: speed up format 2.2 300% by spawning structural decode batch tasks

Xuanwo:
`NextDecodeTask::into_batch` is synchronous and can be CPU-heavy. Running it inline in the future poll path blocks Tokio workers and reduces effective decode concurrency.

This changes becomes more meaningful while we are using zstd.

Benchmarks were run on AWS EC2 using both local and S3 copies of the same dataset (`fineweb.lance.v2_2.lz4`) with repeated scans.

Main run (3 rounds, 20 repeats each):
- Local median latency:
  - p50: `894675us -> 289781us` (`3.087x`, `-67.61%`)
  - p95: `929515us -> 307874us` (`3.019x`, `-66.88%`)
  - p99: `1034383us -> 375041us` (`2.758x`, `-63.74%`)
- S3 median latency:
  - p50: `3998660us -> 3510771us` (`1.139x`, `-12.20%`)
  - p95: `4068799us -> 3572090us` (`1.139x`, `-12.21%`)
  - p99: `4153371us -> 3592478us` (`1.156x`, `-13.50%`)

## Changes

move structural decode batch conversion in `StructuralBatchDecodeStream::into_stream` to `tokio::spawn(...).await`

### DIFF
--- a/rust/lance-namespace-impls/src/credentials/cache.rs
+++ b/rust/lance-namespace-impls/src/credentials/cache.rs
@@ -333,6 +333,21 @@ mod tests {
 
     #[test]
     fn test_calculate_cache_ttl() {
+        const CLOCK_SKEW_TOLERANCE_SECS: u64 = 2;
+
+        fn assert_ttl_close_to(ttl: Option<Duration>, expected_secs: u64) {
+            let actual_secs = ttl.map(|duration| duration.as_secs());
+            assert!(
+                matches!(
+                    actual_secs,
+                    Some(actual)
+                        if actual <= expected_secs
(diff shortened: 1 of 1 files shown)


justinrmiller (approved):
Lgtm!

Xuanwo:
> We keep going back and forth on this one 😄.

Yep, I realized that.

> Can you add some kind of reader config setting? Ideally in a way where we can change the default value for this setting with an environment variable.

Seems to be a good idea, will try.

Xuanwo:
cc @westonpace, I added a flag based on our query pattern and an env to allow users to override it. Let me know what do you think about this change. This seems like an interesting issue that may require a more complete design for us to address. I will add that as a follow-up.

PR #5983, opened 2026-02-23
fix: concurrent read and write to directory namespace

jackye1995:
## Summary

- Add `Throttled` error type (code 21) to lance-namespace for rate limiting and concurrent operation throttling
- Add `commit_retries` config to DirectoryNamespace builder (default 20) to control inner commit retry count. 
- For MergeInserts run in DirectoryNamespace, make sure those set outer  to 0.
- Ensure `object_id` is set as the primary key in the `__manifest` table to ensure we enable primary-key-based dedupe.
- Add `commit_retries()` method to `MergeInsertBuilder` for controlling inner manifest write retry
- Properly map error types: `CommitConflict` → `Throttled` (safe to retry), `TooMuchWriteContention` → `ConcurrentModification` (semantic conflict)
- Add comprehensive concurrent create/drop tests for Python, Java, and Rust with S3 backend

Notes:
1. Also tested with lance-trino to make sure it solves the problems in Trino for concurrent reads and writes
2. I added tests in both java and python to make sure the binding does not introduce additional issue for concurrent access.

### DIFF
--- a/python/python/tests/test_namespace_dir.py
+++ b/python/python/tests/test_namespace_dir.py
@@ -914,8 +914,8 @@ class TestConcurrentOperations:
         """Test concurrent create/drop with single namespace instance."""
         import concurrent.futures
 
-        # Initialize namespace first - create parent namespace to ensure __manifest table
-        # is created before concurrent operations
+        # Initialize namespace first - create parent namespace to ensure __manifest
+        # table is created before concurrent operations
         create_ns_req = CreateNamespaceRequest(id=["test_ns"])
         temp_namespace.create_namespace(create_ns_req)
 


python/python/tests/test_namespace_dir.py
```diff
@@ -894,3 +897,231 @@ def test_external_manifest_store_invokes_namespace_apis():
         assert namespace.describe_table_version_count == describe_count_before_v1 + 1, (
             "describe_table_version should be called once when opening version 1"
         )
+
+
+@pytest.mark.skipif(
+    sys.platform == "win32",
```
jackye1995, line 903: there are in general some problems in windows path, that I will fix fully in a separated PR

rust/lance-namespace-impls/src/dir/manifest.rs
```diff
@@ -793,16 +824,41 @@ impl ManifestNamespace {
...
+                    || error_msg.contains("Too many concurrent writers")
+                {
+                    NamespaceError::ConcurrentModification {
+                        message: format!(
+                            "Object '{}' was concurrently modified by another operation",
+                            object_id
+                        ),
+                    }
+                    .into()
+                } else if error_msg.contains("matched")
```
wjones127: I think this is backwards:

`TooMuchWriteContention` should map to `Throttled`
and
`CommitConflict` should map to `ConcurrentModification`

jackye1995: I actually think the other way around. The setup might be a bit different from what MergeInsert has been doing.

Basically `CommitConflict` happens mainly from manifest writing to a version that is being written to by another writer. In that case, there is just too many writers and it's basically being "throttled"

For `TooMuchWriteContention`, this happens when we exhausted the retry of `RetryableCommitConflict` in `MergeInsert`. But in the use case here, if there is a `RetryableCommitConflict`, it is basically because we are modifying the same row or adding rows of the same primary key, and it becomes a ConcurrentModification that user should decide if retry should be done. That's also why I set conflict_retries to 0 for merge insert.

wjones127: Hmm I see what you mean. We actually use `CommitConflict` in two ways that should probably be distinct:

1. **Incompatible transactions**: there is a conflict that is **not** retry-able. For example, you are trying to create an index, but a concurrent transaction overwrote the table and changed the schema.
1. **Commit step ran out of retries**: we hit the max number of rebase attempts, and even though we could retry again, we aren't. This is indeed just throttling. 

The first happens here:

https://github.com/lance-format/lance/blob/0e530786cb57abdc1aca0c62ea9f4bcb6f514450/rust/lance/src/io/commit/conflict_resolver.rs#L181-L189

The second happens here:

https://github.com/lance-format/lance/blob/0e530786cb57abdc1aca0c62ea9f4bcb6f514450/rust/lance/src/io/commit.rs#L979-L987

wjones127 (approved):

PR #5988, opened 2026-02-23
docs: expand the FTS index doc explaining the training process and multiple partitions

westonpace:


prrao87 (commented):
LGTM!

Could you also please add a note in the quickstart page on FTS pointing to this page? This page is buried under the scalar index page and might be hard to find unless a user specifically searches for FTS (in which case they might stumble across the quickstart page, so it's best to lead the reader here for advanced usage).

At the end of `quickstart/full-text-search/`, we could add an H2 header with the following text:
```
## Advanced usage

For advanced usage instructions with different tokenizers and more technical details on the index training process, visit the [full-text index](../format/table/index/scalar/fts/) section.
```

westonpace:
> Could you also please add a note in the quickstart page on FTS pointing to this page? 

@prrao87 done.  I also fixed a number of broken links.

PR #6001, opened 2026-02-24
feat: add DeleteResult with num_deleted_rows

wkalt:
This introduces a DeleteResult with num_rows_deleted, similar to UpdateResult.

### DIFF
--- a/rust/lance/src/dataset/write/delete.rs
+++ b/rust/lance/src/dataset/write/delete.rs
@@ -193,7 +193,7 @@ impl RetryExecutor for DeleteJob {
                     let fragments = self.dataset.get_fragments();
                     let num_deleted_rows: u64 = fragments
                         .iter()
-                        .map(|f| f.metadata.physical_rows.unwrap_or(0) as u64)
+                        .map(|f| f.metadata.num_rows().unwrap_or(0) as u64)
                         .sum();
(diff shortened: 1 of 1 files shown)


rust/lance/src/dataset/write/delete.rs
```diff
@@ -166,69 +177,78 @@ impl RetryExecutor for DeleteJob {
...
+                } else if matches!(
+                    filter_expr,
+                    Expr::Literal(ScalarValue::Boolean(Some(true)), _)
+                ) {
+                    // Predicate evaluated to true - delete all fragments
+                    let fragments = self.dataset.get_fragments();
+                    let num_deleted_rows: u64 = fragments
+                        .iter()
+                        .map(|f| f.metadata.physical_rows.unwrap_or(0) as u64)
+                        .sum();
```
wjones127, line 197: `physical_rows` is a little weird because that counts already deleted rows. I would use `num_rows()`, or even just call `count_rows()` on the dataset.
```suggestion
                    let num_deleted_rows: u64 = fragments
                        .iter()
                        .map(|f| f.metadata.num_rows().unwrap_or(0) as u64)
                        .sum();
```

wkalt:
@wjones127 thanks, committed your suggestion

PR #6002, opened 2026-02-24
fix: set namespace commit handler for LanceDataset.commit

jackye1995:
Also added helper function `extract_namespace_arc` for shared logic


python/.cargo/config.toml
```diff
@@ -17,7 +17,6 @@ rustflags = [
     "-Wclippy::string_add_assign",
     "-Wclippy::string_add",
     "-Wclippy::string_lit_as_bytes",
-    "-Wclippy::string_to_string",
```
jackye1995, line 20: depreacted clippy lint

hamersaw (approved):

PR #6004, opened 2026-02-25
feat(compaction): preallocate fragment IDs to reduce reserve-commit amplification and conflicts

zhangyue19921010:
## Background

  At the moment, compaction calls `reserve_fragment_ids` in each task, and each call is effectively a commit (`ReserveFragments` transaction).

  With large datasets (for example, compaction plans with up to 10K tasks), especially under concurrent compaction, this creates a large number of tiny commits.
  As a result, we repeatedly trigger transaction conflict checks, which can significantly hurt performance and may even cause task starvation (some tasks take a very long time to finish).

  ## What this PR introduces

  This PR adds `prealloc_fragment_ids` for compaction.

  When enabled:
  1. During compaction planning, we estimate the total number of output fragments.
  2. We reserve fragment IDs upfront (in bulk) before task execution.
  3. Reserved IDs are distributed to compaction tasks and consumed as needed.

  This reduces commit amplification from per-task reservation to mostly plan-level reservation.

  ## Compatibility and safety guarantees

  1. A new switch (`prealloc_fragment_ids`) controls this behavior.
     - Default behavior remains unchanged when the switch is off.
  2. A new parameter (`fragment_id_prealloc_factor`) allows over-reservation to provide a safety margin.
  3. If estimation is not accurate and reserved IDs are exhausted, tasks detect it and perform a fallback `reserve_fragment_ids` call to continue safely.

  ## Expected impact

  - Far fewer tiny commits during large compaction jobs.
  - Lower transaction conflict pressure under concurrency.

For example, the following is an overview of the number of commits for compacting 1000 fragments, **with a total of 100 commits**.

```
drwxr-xr-x@ 104 1   staff    3328  2 25 15:48 ./
drwxr-xr-x@   5 1   staff     160  2 25 15:47 ../
-rw-r--r--@   1 1   staff  118307  2 25 15:48 18446744073709551513.manifest
-rw-r--r--@   1 1   staff   98407  2 25 15:48 18446744073709551514.manifest
-rw-r--r--@   1 1   staff   98407  2 25 15:48 18446744073709551515.manifest
-rw-r--r--@   1 1   staff   98407  2 25 15:48 18446744073709551516.manifest
-rw-r--r--@   1 1   staff   98407  2 25 15:48 18446744073709551517.manifest
-rw-r--r--@   1 1   staff   98407  2 25 15:48 18446744073709551518.manifest
-rw-r--r--@   1 1   staff   98407  2 25 15:48 18446744073709551519.manifest
-rw-r--r--@   1 1   staff   98407  2 25 15:48 18446744073709551520.manifest
-rw-r--r--@   1 1   staff   98407  2 25 15:48 18446744073709551521.manifest
-rw-r--r--@   1 1   staff   98407  2 25 15:48 18446744073709551522.manifest
-rw-r--r--@   1 1   staff   98407  2 25 15:48 18446744073709551523.manifest
-rw-r--r--@   1 1   staff   98407  2 25 15:48 18446744073709551524.manifest
-rw-r--r--@   1 1   staff   98407  2 25 15:48 18446744073709551525.manifest
-rw-r--r--@   1 1   staff   98407  2 25 15:48 18446744073709551526.manifest
-rw-r--r--@   1 1   staff   98407  2 25 15:48 18446744073709551527.manifest
-rw-r--r--@   1 1   staff   98407  2 25 15:48 18446744073709551528.manifest
-rw-r--r--@   1 1   staff   98407  2 25 15:48 18446744073709551529.manifest


.....



-rw-r--r--@   1 1   staff   98407  2 25 15:48 18446744073709551609.manifest
-rw-r--r--@   1 1   staff   98407  2 25 15:48 18446744073709551610.manifest
-rw-r--r--@   1 1   staff   98407  2 25 15:48 18446744073709551611.manifest
-rw-r--r--@   1 1   staff   98407  2 25 15:48 18446744073709551612.manifest
-rw-r--r--@   1 1   staff   98407  2 25 15:48 18446744073709551613.manifest
-rw-r--r--@   1 1   staff  193779  2 25 15:47 18446744073709551614.manifest
```

After optimization, only two times are needed (Pre allocate fragments ids commit + Compaction commit)
```
drwxr-xr-x@ 5 1   staff     160  2 25 15:50 ./
drwxr-xr-x@ 5 1   staff     160  2 25 15:50 ../
-rw-r--r--@ 1 1   staff  118307  2 25 15:50 18446744073709551612.manifest
-rw-r--r--@ 1 1   staff   98407  2 25 15:50 18446744073709551613.manifest
-rw-r--r--@ 1 1   staff  193779  2 25 15:50 18446744073709551614.manifest
```

### DIFF
--- a/rust/lance/src/dataset/optimize.rs
+++ b/rust/lance/src/dataset/optimize.rs
@@ -1120,7 +1120,7 @@ async fn assign_preallocated_fragment_ids(
 
     let (reserved_start, mut reserved_len) = match (task.reserved_start, task.reserved_len) {
         (Some(start), Some(len)) => {
-            if len <= 0 {
+            if len == 0 {
                 return Err(Error::Internal {
                     message: format!(
                         "prealloc_fragment_ids is enabled but task reserved_len={:?})",
@@ -2391,7 +2391,7 @@ mod tests {
         let options = CompactionOptions {
             target_rows_per_fragment: 3_000,
             fragment_id_prealloc_factor: 1.0,
-            prealloc_fragment_ids: prealloc_fragment_ids,
+            prealloc_fragment_ids,
             ..Default::default()
         };
         let plan = plan_compaction(&dataset, &options).await.unwrap();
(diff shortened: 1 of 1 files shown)


java/lance-jni/src/optimize.rs
```diff
@@ -410,6 +410,9 @@ impl FromJObjectWithEnv<TaskData> for JObject<'_> {
         })?;
         Ok(TaskData {
             fragments: task_data,
+            estimated_output_fragment_count: None,
+            reserved_start: None,
+            reserved_len: None,
```
hamersaw, line 415: This is just to make compilation succeed? Should we expose this in python / Java SDKs as first-class? Otherwise we can take it on as a follow up!

rust/lance/src/dataset/optimize.rs
```diff
@@ -176,6 +176,12 @@ pub struct CompactionOptions {
     /// Controls how much data is read at once when performing binary copy.
     /// Defaults to 16MB (16 * 1024 * 1024).
     pub binary_copy_read_batch_bytes: Option<usize>,
+    /// Whether to preallocate fragment IDs for a compaction plan.
+    /// default false
+    pub prealloc_fragment_ids: bool,
+    /// Expansion factor applied to preallocated fragment IDs for a compaction plan.
+    /// default 1.05
```
hamersaw, line 183: Not sure if the `prealloc_fragment_factor` is useful. Basically, the algorithm assigns the estimated fragment count to each `TaskData` and then the last one takes the rest. So in practice, if there are 100 `TaskData` instances each with 2 fragments and the factor is 1.05 (210 reserved), 99 `TaskData` will receive 2 assigned, and the last `TaskData` will get the remaining 12 reserved fragment IDs. So I would challenge that this configuration gives us the ability to tune unused over allocation and addition reservation overhead.

rust/lance/src/dataset/optimize.rs
```diff
@@ -616,6 +661,33 @@ async fn collect_metrics(fragment: &FileFragment) -> Result<FragmentMetrics> {
     })
 }
 
+fn sum_file_size(fragment: &Fragment) -> usize {
```
hamersaw, line 664: I think this is only used once, does it make sense to define a separate function for it?

hamersaw (commented):
I think it would be useful to explore post-allocation of fragment IDs as an alternative to this pre-allocation solution:

**Pre-allocation**: We estimate the number of resulting fragments and do a single call to allocate fragment IDs. Since the fragment IDs are known during compaction, we can distribute transpose `row_id` operations (if enabled).
- Pros: If we estimate correctly, this is great!
- Cons: Incorrect estimates lead to overhead in (1) under-estimation causes additional `reserve_fragment_ids` invocations per task (so same as current approach) or (2) over-estimation results in unused fragment IDs (probably not a large concern, but merits calling out).

**Post-allocation**: We wait until all fragments have been written and then allocate fragment IDs. This will require a post-processing step for transposing `row_id`.
- Pros: We do not have to estimate fragment counts, because these are known. So in every case a single `reserve-fragment-ids` call.
- Cons: We can not distribute transposing `row_id` operations; for large datasets (e.g., 1b+ rows) this can be order of seconds I think).

IMO the call on whether to pre-allocate or post-allocation is a tradeoff between (1) our confidence level in the estimation accuracy and (2) the cost of transposing `row_ids`.

hamersaw:
Would also like to comment here praising potential perf improvements of this work for large compactions. I ran a test on compacting a 100k fragment dataset (16 rows per fragment) down to 144 fragments with 24 compactions in parallel. Without this it took 1h53m and with pre-allocating fragment IDs it took 21m45s. So very significant!

zhangyue19921010:
Hi @hamersaw Thanks a lot for your attention for this issue. 
In general, I have previously struggled with whether to use Pre-allocation or Post-allocation. The main concerns about Post-allocation are:
- For the calculation of row_id_map, it will degrade from the original parallel computing to serial computing. In scenarios with large amounts of data, this may become another bottleneck. For computing with large amounts of rows, it might still be necessary to keep parallel processing.

Maybe hot path is related to follow code which need to access every rows and fragments compaction involved.
https://github.com/lance-format/lance/blob/39f8a2645ac360e25ba208d6adad29f2c0277920/rust/lance/src/dataset/optimize/remapping.rs#L157

- It will modify the data structure of RewriteResult, that is, it will no longer contain the `row_id_map` information, which may have an impact on historical users.

From the perspective of Pre-allocation, the compatibility is better, but it also indeed cannot guarantee the continuity of fragment IDs. I have had a brief discussion with @jackye1995 about this ：）

Hi @wjones127,  @jackye1995 and @Xuanwo  Could you please share some thoughts? Thank you !

rust/lance/src/dataset/optimize.rs
zhangyue19921010: Nice Catch! In fact, we can apply this magnification factor to the granularity of Task allocation. Taking your example, that would be 2 * 1.05 => 3, and finally pre-allocation for fragment ids according to the total number of finally allocated fragments.

rust/lance/src/dataset/optimize.rs
```diff
@@ -508,18 +518,48 @@ impl CompactionPlanner for DefaultCompactionPlanner {
...
-            });
+            .flat_map(|bin| bin.split_for_size(self.options.target_rows_per_fragment));
+
+        let final_bins = if need_prealloc_fragment_ids {
+            flat_map
+                .map(|bin| {
+                    let live_rows = bin.row_counts.iter().copied().sum::<usize>();
+                    let total_input_file_size =
+                        bin.fragments.iter().map(sum_file_size).sum::<usize>();
+                    let estimated_output_fragment_count = estimate_output_fragment_num(
```
westonpace, line 532: Why is this an estimate?  Is it because we might hit the size (bytes) limit when writing and end up with more fragments than expected?

rust/lance/src/dataset/optimize.rs
```diff
@@ -176,6 +176,12 @@ pub struct CompactionOptions {
     /// Controls how much data is read at once when performing binary copy.
     /// Defaults to 16MB (16 * 1024 * 1024).
     pub binary_copy_read_batch_bytes: Option<usize>,
+    /// Whether to preallocate fragment IDs for a compaction plan.
+    /// default false
+    pub prealloc_fragment_ids: bool,
```
westonpace, line 181: Why would I ever want this set to false?

zhangyue19921010:
First of all, thanks for all your discuss and attention. 

After careful consideration, post-processing is indeed a more concise solution. I will close this PR. Thanks for all your effort here @hamersaw @westonpace @Xuanwo  👍

Xuanwo:
Thank you @zhangyue19921010 again!

PR #6006, opened 2026-02-25
fix: fast_search limits full text search to indexed fragments

BubbleCal:
Summary
- short-circuit FTS scans when `fast_search` is enabled and no indexed fragments exist so we return an empty plan instead of scanning unindexed data
- skip the unindexed-match planning path entirely under `fast_search`, forcing only index-backed queries even when fragments exist
- add plan verification and a regression test proving `fast_search` excludes rows appended after building the FTS index

### DIFF
--- a/rust/lance/src/dataset/scanner.rs
+++ b/rust/lance/src/dataset/scanner.rs
@@ -3341,6 +3341,9 @@ impl Scanner {
 
                 // If all target fragments are unindexed, skip index entirely
                 if unindexed_fragments.len() == target_fragments.len() {
+                    if self.fast_search {
+                        return Ok(Arc::new(EmptyExec::new(FTS_SCHEMA.clone())));
+                    }
                     let flat_match_plan = self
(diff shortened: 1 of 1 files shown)


rust/lance/src/dataset/scanner.rs
```diff
@@ -3363,6 +3368,9 @@ impl Scanner {
                 }
             }
             None => {
+                if self.fast_search {
+                    return Ok(Arc::new(EmptyExec::new(FTS_SCHEMA.clone())));
+                }
```
westonpace, line 3373: This is slightly different from vector search.  In vector search, if there is no index, then `fast_search` is just ignored (it still does an exhaustive KNN).  However, I kind of like this behavior better.  So I think I'm just pointing it out.  We can update vector search later.

BubbleCal: ic, let me update the vector search behavior in next PR

PR #6008, opened 2026-02-25
fix: correctly calculate max visible level when a list has no def

### LINKED ISSUE #5930: Lance 2.1/2.2 panics on null struct elements in List<Struct>
Repro:

```python
import pyarrow as pa
import lance

struct_type = pa.struct([("tag", pa.string())])

# Create struct array where 2nd element is null
tag_array = pa.array(["valid", "null_struct", "valid", "valid"])
struct_array = pa.StructArray.from_arrays(
    [tag_array],
    fields=[pa.field("tag", pa.string(), nullable=True)],
    mask=pa.array([True, False, True, True])  # False = null struct element
)
# Create list containing these structs
offsets = pa.array([0, 4], type=pa.int32())
list_array = pa.ListArray.from_arrays(offsets, struct_array)
batch = pa.record_batch(
    [pa.array([0]), list_array],
    names=["id", "value"]
)

ds = lance.write_dataset(batch, "memory://", data_storage_version="2.2")
ds.to_table()
```
```
thread 'lance_background_thread' panicked at /Users/willjones/Documents/lance/rust/lance-encoding/src/encodings/logical/struct.rs:382:13:
assertion `left == right` failed
  left: 1
 right: 0
  File "<stdin>", line 1, in <module>
    ds.to_table()
    ~~~~~~~~~~~^^
  File "/Users/willjones/Documents/lance/python/python/lance/dataset.py", line 1106, in to_table
    ).to_table()
      ~~~~~~~~^^
  File "/Users/willjones/Documents/lance/python/python/lance/dataset.py", line 5135, in to_table
           ~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "pyarrow/ipc.pxi", line 848, in pyarrow.lib.RecordBatchReader.read_all
pyarrow.lib.ArrowInvalid: External error: Wrapped error: task 65 panicked with message "assertion `left == right` failed\n  left: 1\n right: 0", /Users/willjones/Documents/lance/rust/lance-encoding/src/decoder.rs:1803:39
```

westonpace:
This was a bug that would be encountered whenever a list had nullable items (struct or otherwise) but the list itself was never null and never empty.  The unraveler was incorrectly skipping offsets and returning fewer lists than it should.

This was a reader-only bug.  No corrupt data would have been written.

Closes #5930

### DIFF
--- a/rust/lance-encoding/src/repdef.rs
+++ b/rust/lance-encoding/src/repdef.rs
@@ -1296,8 +1296,6 @@ impl RepDefUnraveler {
 
         let mut curlen: usize = offsets.last().map(|o| o.as_usize()).unwrap_or(0);
 
-        println!("A: {:#?}", offsets);
-
         // If offsets is empty this is a no-op.  If offsets is not empty that means we already
         // added a set of offsets.  For example, we might have added [0, 3, 5] (2 lists).  Now
         // say we want to add [0, 1, 4] (2 lists).  We should get [0, 3, 5, 6, 9] (4 lists).  If


westonpace:
For review, probably just look at the test cases and ensure they make sense.

Xuanwo:

PR #6013, opened 2026-02-25
perf: avoid oversized variable buffers in full-zip scan batches

Xuanwo:
`VariableFullZipDecoder::drain` used to pass the entire page-level variable `data` buffer to every batch while slicing only the batch offsets.
This broke the normalized variable-buffer assumption (offsets should be 0-based for the provided data window) and caused downstream variable decompressors to repeatedly size work against oversized input buffers.

Bencmarks:

- scan-full: 34047us -> 15703us，2.168x
- caption: 28431us -> 6979us，4.074x

## Changes

- Add `LanceBuffer::slice_and_rebase_offsets` in `rust/lance-encoding/src/buffer.rs`.
  - Slice data to the byte range referenced by the batch offsets.
  - Rebase offsets to start at zero.
  - Validate invalid/overflowing offsets.
- Update variable full-zip batch drain path to use the new helper.


rust/lance-encoding/src/buffer.rs
```diff
@@ -268,6 +268,91 @@ impl LanceBuffer {
         Self(Buffer::from_vec(Vec::from(array)))
     }
 
+    fn slice_and_rebase_offsets_typed<T>(data: &Self, offsets: &Self) -> Result<(Self, Self)>
```
westonpace: This method doesn't belong here.  A `Buffer` has no concept of "offsets".  Can we put it in `VariableFullZipDecoder`?

Xuanwo:
> Can we just file a follow-up to see if we can remove this copy sometime in the future?

Sure, will do!

rust/lance-encoding/src/buffer.rs
Xuanwo: Sure

Xuanwo:
Filed a follow-up issue to investigate removing the per-batch offsets copy while preserving correctness and perf guarantees: #6027

PR #6015, opened 2026-02-25
feat!: upgrade DataFusion dependency to 52.1.0

wjones127:
- **Bump datafusion requirement to 52**
- **ruff format**
- **fix: use fields_with_udf for aggregate type coercion (DF52)**
- **fix: use OutputBatches metric variant for DF52 compatibility**


python/pyproject.toml
```diff
@@ -60,9 +61,9 @@ tests = [
     # Only test tensorflow on linux for now. We will deprecate tensorflow soon.
     "tensorflow; sys_platform == 'linux'",
     "tqdm",
-    "datafusion>=50.1,<52",
+    "datafusion>=52,<53; python_version >= '3.10'",
```
jackye1995, line 64: does that mean we will officially drop 3.9? If so we should remove `"Programming Language :: Python :: 3.9"` above

wjones127: No, datafusion dropped support for 3.9, so we can only test the integration with Python 3.10 and above. We support 3.9 for now, though I think we might drop support in a future PR.

PR #6016, opened 2026-02-25
perf: reduce peak memory during cosine IVF-PQ index training

wkalt:
Two optimizations that together eliminate the transient 2x memory peak on the training sample during cosine-distance index builds:

1. Add `normalize_fsl_owned` that L2-normalizes a FixedSizeListArray in-place via `Buffer::into_mutable()` when the buffer is uniquely owned, avoiding a full copy. Falls back to the existing copy path when the buffer is shared.

2. Skip `arrow::compute::filter` when all vectors are already finite, avoiding another full copy of the training data.


wkalt:
progress
this is an index build on 100M 384d vectors before and after the change. Change targets the IVF training portion at the start.

PR #6021, opened 2026-02-26
chore: use jemalloc as default allocator in pylance

vivek-bharathan:


yanghua:
Hi @vivek-bharathan Would you please explain why we need to change to jemalloc?

vivek-bharathan:
In general we see better performance with jemalloc compared to the standard allocator. Particularly for index builds we see faster builds and lower memory consumption. Do you have a concern @yanghua ?

wjones127:
> In general we see better performance with jemalloc compared to the standard allocator. Particularly for index builds we see faster builds and lower memory consumption.

Perhaps you can share benchmark results for this? Can model it off what we did here: https://github.com/lance-format/lance/issues/1372

PR #6024, opened 2026-02-26
perf: fast rotation for RQ quantization

BubbleCal:
this introduces a new rotation algorithm which can do rotation in `O(dlogd)` time (`O(d^2)` before), and reduces the memory footprint of rotation matrix from d^2 floats to d bits

this doesn't reduce recall, and significantly improves performance:
image

query performance is 93% faster, and 307% faster than IVF_PQ
indexing is 5.4x faster, and 11.9x faster than IVF_PQ

### DIFF
--- a/rust/lance-index/src/vector/bq/builder.rs
+++ b/rust/lance-index/src/vector/bq/builder.rs
@@ -321,6 +321,14 @@ impl Quantization for RabitQuantizer {
         _: lance_linalg::distance::DistanceType,
         params: &Self::BuildParams,
     ) -> Result<Self> {
+        let dim = data.as_fixed_size_list().value_length() as usize;
+        if !dim.is_multiple_of(u8::BITS as usize) {
+            return Err(Error::invalid_input(
+                "vector dimension must be divisible by 8 for IVF_RQ",
+                location!(),
+            ));
+        }
+
         let q = match data.as_fixed_size_list().value_type() {
             DataType::Float16 => Self::new_with_rotation::<Float16Type>(
                 params.num_bits,
@@ -511,6 +519,8 @@ mod tests {
     use super::*;
     use approx::assert_relative_eq;
     use arrow::datatypes::Float32Type;
+    use arrow_array::{FixedSizeListArray, Float32Array};
+    use lance_linalg::distance::DistanceType;
     use rstest::rstest;
 
     #[rstest]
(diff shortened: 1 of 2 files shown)


rust/lance-index/src/vector/bq/builder.rs
```diff
@@ -48,27 +56,59 @@ pub struct RabitQuantizer {
     metadata: RabitQuantizationMetadata,
 }
 
+#[inline]
+fn pack_sign_bits(codes: &mut [u8], rotated: &[f32]) {
+    codes.fill(0);
+    for (bit_idx, value) in rotated.iter().enumerate() {
+        if value.is_sign_positive() {
+            codes[bit_idx / u8::BITS as usize] |= 1u8 << (bit_idx % u8::BITS as usize);
```
Xuanwo, line 64: It looks like if dim is not byte-aligned, we could hit panic here, for example, if dim is 129 (is this possible?)

I ask codex to have a repro:

```rust
use std::sync::Arc;

use arrow::datatypes::Float32Type;
use arrow_array::{ArrayRef, FixedSizeListArray, Float32Array};
use arrow_schema::{DataType, Field};
use lance_index::vector::bq::builder::RabitQuantizer;
use lance_index::vector::bq::RQRotationType;
use lance_index::vector::quantizer::Quantization;

fn main() {
    let dim = 129; // non-byte-aligned dimension
    let values = Float32Array::from(vec![1.0f32; dim]);
    let field = Arc::new(Field::new("item", DataType::Float32, true));
    let vectors =
        FixedSizeListArray::try_new(field, dim as i32, Arc::new(values) as ArrayRef, None).unwrap();

    // Panic is nondeterministic (depends on random rotation signs), loop to trigger reliably.
    for i in 0..5000 {
        let q =
            RabitQuantizer::new_with_rotation::<Float32Type>(1, dim as i32, RQRotationType::Fast);
        let _ = q.quantize(&vectors).unwrap();
        if i % 1000 == 0 {
            println!("iter={i}");
        }
    }

    println!("done");
}
```

This logic will panic at `index out of bounds`

BubbleCal: good catch
will add requirement of `dimension % 8 == 0`, i think it's fine because real world vectors are almost always dividable by 8

PR #6025, opened 2026-02-26
feat(cleanup): add more metrics to RemovalStats

zhangyue19921010:
Add a new metric `removed_data_file_num` to the `RemovalStats` of the cleanup operation results, 

So that users can easily perceive how many lance data files have been deleted in the current cleanup operation and better evaluate the impact scope of the current cleanup.

### DIFF
--- a/python/src/dataset/cleanup.rs
+++ b/python/src/dataset/cleanup.rs
@@ -19,10 +19,10 @@ use pyo3::{pyclass, pymethods};
 pub struct CleanupStats {
     pub bytes_removed: u64,
     pub old_versions: u64,
-    pub removed_data_files: u64,
-    pub removed_transaction_files: u64,
-    pub removed_index_files: u64,
-    pub removed_deletion_files: u64,
+    pub data_files_removed: u64,
+    pub transaction_files_removed: u64,
+    pub index_files_removed: u64,
+    pub deletion_files_removed: u64,
 }
 
 #[pymethods]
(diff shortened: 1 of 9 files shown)


python/src/dataset/cleanup.rs
```diff
@@ -19,6 +19,7 @@ use pyo3::{pyclass, pymethods};
 pub struct CleanupStats {
     pub bytes_removed: u64,
     pub old_versions: u64,
+    pub removed_data_file_num: u64,
```
majin1102: This is a good start.

Wondering if we could include transaction files, index files, deletion files, etc.  And I think `removed_data_files` might be better. We have similar fields in `ManifestSummary`. `data_files_removed` might be another option

zhangyue19921010:
Hi @majin1102 Thanks a lot for your review.  All comments are addressed. PTAL~

python/src/dataset/cleanup.rs
```diff
@@ -19,6 +19,10 @@ use pyo3::{pyclass, pymethods};
 pub struct CleanupStats {
     pub bytes_removed: u64,
     pub old_versions: u64,
+    pub removed_data_files: u64,
+    pub removed_transaction_files: u64,
+    pub removed_index_files: u64,
+    pub removed_deletion_files: u64,
```
Xuanwo:
```suggestion
    pub data_files_removed: u64,
    pub transaction_files_removed: u64,
    pub index_files_removed: u64,
    pub deletion_files_removed: u64,
```

I suggest we keep the same naing style along with existing `bytes_removed`.

PR #6026, opened 2026-02-26
perf: avoid re-open shard indices and small reads

BubbleCal:
1. reuse the readers and avoid reopening them in each iteration
2. read multiple partitions in one read to avoid small IOs

### DIFF
--- a/.bumpversion.toml
+++ b/.bumpversion.toml
@@ -1,5 +1,5 @@
 [tool.bumpversion]
-current_version = "4.0.0-beta.1"
+current_version = "4.0.0-beta.3"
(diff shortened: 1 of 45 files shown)


cmccabe:
just out of curiosity how did you choose 512 as the partition window size? how big should we expect that to be in memory?
also, the bot's comment about test coverage is probably right :)

BubbleCal:
> just out of curiosity how did you choose 512 as the partition window size? how big should we expect that to be in memory? also, the bot's comment about test coverage is probably right :)

i assume we have 4K rows per partition roughly (the default index param)
so 512 partitions is about 200MB (say d=768, so 768 * 4 / 32 * 4096 * 512 = 192MB)

good catch, just added a test for testing that window size

wjones127 (approved):

PR #6036, opened 2026-02-27
test: extend file format version matrices to include 2.2

Xuanwo:
This PR adds file format `2.2` to existing `2.0/2.1` version coverage matrices across Python and Rust test/benchmark paths.

---

**Parts of this PR were drafted with assistance from Codex (with `gpt-5.3-codex`) and fully reviewed and edited by me. I take full responsibility for all changes.**


westonpace (approved):
We don't have to do it now but by the time 2.3 rolls around we probably just want a `file_format_versions` function in python that returns a list of all the versions so we don't have to repeat ourselves all over.

PR #6041, opened 2026-02-27
fix: deduplicate stale BTree entries during optimize with stable row IDs

wkalt:
When stable row IDs are used, the BTree index stores stable row IDs (not physical addresses) in its _rowid column. During optimize, the old entries for updated rows would survive fragment-based filtering because stable row IDs don't encode fragment IDs in their upper 32 bits. Both old (stale) and new entries for the same row ID ended up in the merged index, causing FlatIndex::try_new to fail with "RowAddrTreeMap:: from_sorted_iter called with non-sorted input" due to duplicate IDs.

Fix: before merging old and new data in combine_old_new(), collect the new data's row IDs and filter old entries that have matching IDs. This removes stale entries for rows that have been updated.

### DIFF
--- a/rust/lance-index/src/scalar/btree.rs
+++ b/rust/lance-index/src/scalar/btree.rs
@@ -4,7 +4,7 @@
 use std::{
     any::Any,
     cmp::Ordering,
-    collections::{BTreeMap, BinaryHeap, HashMap, HashSet},
+    collections::{BTreeMap, BinaryHeap, HashMap},
     fmt::{Debug, Display},
     ops::Bound,
     sync::Arc,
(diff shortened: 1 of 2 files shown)


wkalt:
This PR was generated automatically from the output of some generative testing

rust/lance-index/src/scalar/btree.rs
```diff
@@ -1303,12 +1303,42 @@ impl BTreeIndex {
...
 
-        let new_input = Arc::new(OneShotExec::new(new_data));
+        // Collect new data row IDs so we can remove stale entries from the old
+        // index. When stable row IDs are used, the _rowid values are stable row
+        // IDs (not physical addresses), so fragment-based filtering alone is
+        // insufficient — an updated row keeps its stable row ID but moves to a
+        // new fragment. Without this dedup step, both the old (stale) and new
+        // entries would survive, causing duplicate row IDs in the merged index.
+        let new_schema = new_data.schema();
+        let new_batches: Vec<RecordBatch> = new_data.try_collect().await?;
```
Xuanwo, line 1313: This will collect all our new data, seems bad.

Issue #5918, opened 2026-02-09
[Bug] BTREE index returns extra rows for != with NULLs


fenfeng9:
### Description
When using a `BTREE` index, a `!=` predicate on a column containing `NULL`s returns a different row count compared to a full scan.

The `BTREE` index path drops `NULL` rows for non-`IS NULL` queries, so `!=` (implemented as `NOT` (`=`)) incorrectly flips `UNKNOWN` to `TRUE` and includes `NULL` rows.

### Reproduce
```python
import tempfile

import lance
import pyarrow as pa

data = pa.table({"c3": pa.array([None, "a"], type=pa.string())})
filter_expr = "(c3 != 'x')"

with tempfile.TemporaryDirectory() as d:
    lance.write_dataset(data, f"{d}/no")
    lance.write_dataset(data, f"{d}/ix")

    ds_no = lance.dataset(f"{d}/no")
    ds_ix = lance.dataset(f"{d}/ix")
    ds_ix.create_scalar_index("c3", index_type="BTREE")

    no_rows = ds_no.to_table(filter=filter_expr).num_rows
    ix_rows = ds_ix.to_table(filter=filter_expr).num_rows

    print("no_index:", no_rows)
    print("with_index:", ix_rows)

```

### Result
```
no_index: 1
with_index: 2
```

fenfeng9:
Preliminary analysis: In the `BTREE` index path, non‑`IS NULL` queries don’t include `NULL` pages, so `NULL`s never reach `NullableRowAddrSet.nulls`. When `!=` is rewritten as `NOT(=)`, `UNKNOWN` is treated as `FALSE` and flipped to `TRUE`, causing `NULL` rows to be returned.

https://github.com/lance-format/lance/blob/e1489abc21438c5086ed76f03136534a2a9ee9a4/rust/lance-index/src/scalar/btree.rs#L657-L664

fenfeng9:
completed by https://github.com/lance-format/lance/pull/6043

PR #6043, opened 2026-02-27
fix(btree): include null pages in non-IsNull queries for correct thre…

wkalt:
…e-valued logic

BTree index search only consulted pages whose value range matched the query. When a queried value (e.g. 0) didn't exist in any page, no pages were searched and the null row set was empty. This caused `NOT(x = 0)` to produce a BlockList with no nulls tracked, incorrectly allowing NULL rows to pass the filter.

The fix adds null pages (both partially-null and all-null) to every non-IsNull query's page list as Matches::Some, so FlatIndex::search() evaluates the predicate and correctly marks those rows as NULL. This preserves three-valued logic for compound expressions like `(x != 0) OR (x < 5)`.

### DIFF
--- a/rust/lance-index/src/scalar/btree.rs
+++ b/rust/lance-index/src/scalar/btree.rs
@@ -4,7 +4,7 @@
 use std::{
     any::Any,
     cmp::Ordering,
-    collections::{BTreeMap, BinaryHeap, HashMap},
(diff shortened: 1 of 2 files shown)


fenfeng9:
This PR appears to address the issue #5918.

wkalt:
@fenfeng9 awesome! Thanks for the pointer. This bug was found with a generative testing framework I have been hacking on.

PR #6044, opened 2026-02-27
fix: handle list-level NULLs in NOT filters

### LINKED ISSUE #5904: LabelListIndex: NOT filters mis-handle NULL lists and NULL elements
### Description
Follow-up issue for #5867.

### Reproduce
```python
import lance
import pyarrow as pa

tbl = pa.table({"labels": [["foo", None], ["foo"], ["bar", None], ["bar"], None, []]})
dataset = lance.write_dataset(tbl, "/tmp/dataset", mode="overwrite")

filters = [
    "NOT array_has_any(labels, ['foo'])",
    "NOT array_has_all(labels, ['foo'])",
    "NOT array_contains(labels, 'foo')",
]

expected = {
    f: dataset.to_table(filter=f).column("labels").to_pylist() for f in filters
}

dataset.create_scalar_index("labels", index_type="LABEL_LIST")

actual = {
    f: dataset.to_table(filter=f).column("labels").to_pylist() for f in filters
}

print(f"{'Filter':<35} {'Expected':<25} {'Actual':<25} {'Match':<6}")
print("="*95)
for f in filters:
    exp_str = str(expected[f])
    act_str = str(actual[f])
    match = "✓" if expected[f] == actual[f] else "✗"
    print(f"{f:<35} {exp_str:<25} {act_str:<25} {match:<6}")
```

### Result
```python
Filter                              Expected                  Actual                    Match 
===============================================================================================
NOT array_has_any(labels, ['foo'])  [['bar', None], ['bar']]  [['bar'], None]           ✗     
NOT array_has_all(labels, ['foo'])  [['bar', None], ['bar']]  [['bar'], None]           ✗     
NOT array_contains(labels, 'foo')   [['bar', None], ['bar']]  [['bar'], None]           ✗  
```

fenfeng9:
closes #5904

`LabelListIndex` drops list-level NULL rows during unnest, so `NOT` filters can return NULL lists incorrectly.

Changed:
- persists list-level NULL rows in the LabelList index and merges them back into query null sets
- keeps older bitmap-only LabelList index layouts readable by treating missing null-set metadata as empty
- warns when opening an older LabelList index on a nullable column, since NOT-filter semantics may still be incomplete until the index is rebuilt

### DIFF
--- a/rust/lance-index/src/scalar/label_list.rs
+++ b/rust/lance-index/src/scalar/label_list.rs
@@ -231,7 +231,7 @@ impl ScalarIndex for LabelListIndex {
             remapped_state,
             dest_store,
             self.values_index.value_type(),
-            remapped_nulls,
+            &remapped_nulls,
         )
         .await?;
 
@@ -261,7 +261,7 @@ impl ScalarIndex for LabelListIndex {
         if !new_nulls.is_empty() {
             merged_nulls |= &new_nulls;
         }
-        write_label_list_bitmap_index(merged_state, dest_store, &value_type, merged_nulls).await?;
+        write_label_list_bitmap_index(merged_state, dest_store, &value_type, &merged_nulls).await?;
 
         Ok(CreatedIndex {
             index_details: prost_types::Any::from_msg(&pbold::LabelListIndexDetails::default())
@@ -482,7 +482,7 @@ async fn write_label_list_bitmap_index(
         HashMap::new(),
         vec![(
             LABEL_LIST_NULLS_METADATA_KEY.to_string(),
-            serialize_list_nulls(&list_nulls)?,
+            serialize_list_nulls(list_nulls)?,
         )],
     )
     .await
@@ -581,7 +581,7 @@ impl ScalarIndexPlugin for LabelListIndexPlugin {
         let (state, value_type) =
             BitmapIndexPlugin::build_bitmap_index_state(data, HashMap::new()).await?;
         let list_nulls = list_nulls.lock().unwrap().clone();
-        write_label_list_bitmap_index(state, index_store, &value_type, list_nulls).await?;
+        write_label_list_bitmap_index(state, index_store, &value_type, &list_nulls).await?;
         Ok(CreatedIndex {
             index_details: prost_types::Any::from_msg(&pbold::LabelListIndexDetails::default())
                 .unwrap(),


fenfeng9:
@wjones127  @westonpace Could you please review this when you have a moment? Thanks!

rust/lance-index/src/scalar/label_list.rs
```diff
@@ -34,7 +41,17 @@ use crate::scalar::{CreatedIndex, UpdateCriteria};
 use crate::{Index, IndexType};
 
 pub const BITMAP_LOOKUP_NAME: &str = "bitmap_page_lookup.lance";
-const LABEL_LIST_INDEX_VERSION: u32 = 0;
+pub const LABEL_LIST_NULLS_NAME: &str = "label_list_nulls.lance";
```
wjones127: If we are just writing a single buffer, why don't we instead use a global buffer in one of the existing index files? That will cut down number of files needed.

wjones127 (changes requested):
Nice work. I think the one thing I would change is put the nulls in a global buffer. Here's an example of how we do something similar for IVF indexes:

https://github.com/lance-format/lance/blob/ca8fb1babaf9951a0a1df30ea0cd6ea37c0df904/rust/lance/src/index/vector/ivf.rs#L2024-L2026

I was going to say we should also add more compat tests, but I think we are already well covered there. You can see the warnings are working in our label list compat test:

https://github.com/lance-format/lance/actions/runs/22492208650/job/65160507370?pr=6044#step:6:117

That is defined here: https://github.com/lance-format/lance/blob/ca8fb1babaf9951a0a1df30ea0cd6ea37c0df904/python/python/tests/compat/test_scalar_indices.py#L75-L125

rust/lance-index/src/scalar/label_list.rs
fenfeng9: Thanks for the review and the suggestion! I’ll look into the global buffer idea and get back to you.

fenfeng9: I took a look at what it would take to do this on the scalar side.

The main difference from the IVF example is that the scalar index path currently goes through the `IndexStore` / `IndexReader` / `IndexWriter` abstraction:

https://github.com/lance-format/lance/blob/aac74b441cdb6df7d78700dbba33c521e6379ca5/rust/lance-index/src/scalar.rs#L170-L179

  and that abstraction is centered around record-batch files. It does not currently expose global buffer read/write directly.

There is also a second constraint here: `LabelList` currently reuses the bitmap index train/update/remap paths, so moving the null map into the same index file would likely require some refactoring in that bitmap-backed flow as well.

https://github.com/lance-format/lance/blob/aac74b441cdb6df7d78700dbba33c521e6379ca5/rust/lance-index/src/scalar/label_list.rs#L443-L452

fenfeng9: I think we can move this to a global buffer, but on the scalar side it's a bit more than just moving the null map. We'd probably need to add global-buffer read/write support to the current scalar reader/writer path, and also adjust the bitmap-backed LabelList train/update/remap flow so the null map can be written into the same index file.
So before I go further with that, I wanted to check with you whether you'd like me to also convert LabelList to a global-buffer-based layout in this PR.

wjones127: Yeah I would think adding something like:

```rust
pub trait IndexWriter: Send { 
     async fn add_global_buffer(&mut self, data: Bytes) -> Result<()>;
}
```

Would make sense. @westonpace do you have any objection?

fenfeng9: > Yeah I would think adding something like:
> 
> ```rust
> pub trait IndexWriter: Send { 
>      async fn add_global_buffer(&mut self, data: Bytes) -> Result<()>;
> }
> ```
> 

Thanks for the suggestion! I'll implement this and update the PR soon.

fenfeng9:
I updated the PR in that direction.

I added global buffer support to the scalar `reader/writer` path, and `LabelList` now stores list-level nulls in the bitmap file instead of a separate file. The buffer index is recorded in metadata and read back from there.

rust/lance-index/src/scalar/label_list.rs
```diff
@@ -353,6 +438,54 @@ fn unnest_chunks(
...
+    let mut bytes = Vec::new();
+    null_map.serialize_into(&mut bytes)?;
+    Ok(Bytes::from(bytes))
+}
+
+async fn write_label_list_bitmap_index(
+    state: HashMap<ScalarValue, RowAddrTreeMap>,
+    store: &dyn IndexStore,
+    value_type: &DataType,
+    list_nulls: RowAddrTreeMap,
```
wjones127: nitpick: I think you could pass this as a reference: 
```suggestion
    list_nulls: &RowAddrTreeMap,
```

fenfeng9: Done.

PR #6054, opened 2026-02-27
perf: speedup flat fts

westonpace:
This adds various performance improvements to the flat FTS search.  The most significant improvement is that it parallelizes the search.

This does have some impact to accuracy.  To calculate bm25 we typically need to make two passes through the data.  The first to count token frequencies and the second to count token scores.  The current implementation avoids this by using "token frequency so far" when calculating the bm25 score.  This is generally accurate when there is a lot of indexed data and a small amount of unindexed data because the "token frequency so far" gets bootstrapped by the frequencies from the index and so the effect of the unindexed frequencies are minimal.

However, this can be more significant when there is no index or the unindexed data makes up a significant portion of the data.  In that case the "token frequency so far" can be quite inaccurate for the first few documents.

In parallelizing this search we make this problem worse since each thread is calculating its own independent "token frequency so far" and it will take longer for each one to arrive at a more accurate result.

The most accurate approach would probably be to just accumulate all data in memory, tokenize (in parallel), count token frequencies (back to serial), then calculate scores (in parallel).  This does run the risk of accumulating too much data however.  

Another alternative could be to accumulate up to some amount (e.g. 100MB), calculate initial token frequencies, and then parallelize the rest of the search using those initial token frequencies.  I'm open to suggestions.  In the meantime we could probably proceed with this PR as-is.

In addition to the parallelization this PR also makes various changes to the algorithm itself to avoid string copies.  This cuts down on the CPU time by 5x on my system.

### DIFF
--- a/rust/lance-index/src/scalar/inverted/index.rs
+++ b/rust/lance-index/src/scalar/inverted/index.rs
@@ -2,8 +2,8 @@
 // SPDX-FileCopyrightText: Copyright The Lance Authors
 
 use std::fmt::{Debug, Display};
-use std::sync::atomic::{AtomicBool, AtomicU64, Ordering};
 use std::sync::Arc;
+use std::sync::atomic::{AtomicBool, AtomicU64, Ordering};
 use std::{
     cmp::{Reverse, min},
     collections::BinaryHeap,
@@ -32,11 +32,7 @@ use datafusion::execution::SendableRecordBatchStream;
 use datafusion::physical_plan::stream::RecordBatchStreamAdapter;
 use deepsize::DeepSizeOf;
 use fst::{Automaton, IntoStreamer, Streamer};
-<<<<<<< HEAD
-use futures::{FutureExt, StreamExt, TryStreamExt, stream};
-=======
-use futures::{stream, FutureExt, Stream, StreamExt, TryStreamExt};
->>>>>>> 902b77e45 (Rework according to PR suggestions)
+use futures::{FutureExt, Stream, StreamExt, TryStreamExt, stream};
 use itertools::Itertools;
 use lance_arrow::{RecordBatchExt, iter_str_array};
 use lance_core::cache::{CacheKey, LanceCache, WeakLanceCache};
--- a/rust/lance-index/src/scalar/rtree.rs
+++ b/rust/lance-index/src/scalar/rtree.rs
@@ -25,6 +25,7 @@ use datafusion::execution::SendableRecordBatchStream;
 use datafusion::physical_plan::stream::RecordBatchStreamAdapter;
 use datafusion_common::DataFusionError;
 use deepsize::DeepSizeOf;
+use futures::future::BoxFuture;
 use futures::{FutureExt, Stream, StreamExt, TryFutureExt, TryStreamExt, stream};
 use geoarrow_array::array::{RectArray, from_arrow_array};
 use geoarrow_array::builder::RectBuilder;
--- a/rust/lance-index/src/scalar.rs
+++ b/rust/lance-index/src/scalar.rs
@@ -18,7 +18,6 @@ use std::{any::Any, ops::Bound, sync::Arc};
 use datafusion_expr::Expr;
 use datafusion_expr::expr::ScalarFunction;
 use deepsize::DeepSizeOf;
-use futures::{FutureExt, Stream, future::BoxFuture};
 use inverted::query::{FtsQuery, FtsQueryNode, FtsSearchParams, MatchQuery, fill_fts_query_column};
 use lance_core::utils::mask::{NullableRowAddrSet, RowAddrTreeMap};
 use lance_core::{Error, Result};
--- a/rust/lance-index/src/scalar/inverted/query.rs
+++ b/rust/lance-index/src/scalar/inverted/query.rs
@@ -6,7 +6,6 @@ use crate::scalar::inverted::tokenizer::lance_tokenizer::LanceTokenizer;
 use lance_core::{Error, Result};
 use serde::ser::SerializeMap;
 use serde::{Deserialize, Serialize};
-use snafu::location;
 use std::collections::{HashMap, HashSet};
 
 #[derive(Debug, Clone)]
--- a/rust/lance/src/dataset/scanner.rs
+++ b/rust/lance/src/dataset/scanner.rs
@@ -3966,7 +3966,6 @@ impl Scanner {
                     .as_ref()
                     .ok_or(Error::invalid_input(
                         "the column must be specified in the query".to_string(),
-                        location!(),
                     ))?
                     .clone();
                 let input = if schema.column_with_name(&column).is_none() {
@@ -3988,7 +3987,6 @@ impl Scanner {
             }
             _ => Err(Error::not_supported(
                 "Only Match queries are supported currently when using FTS as a post-filter",
-                location!(),
             )),
         }
     }


wjones127:
> The most accurate approach would probably be to just accumulate all data in memory, tokenize (in parallel), count token frequencies (back to serial), then calculate scores (in parallel). This does run the risk of accumulating too much data however.

I feel like a very efficient way would be to tokenize in parallel, and then collect all tokenized data and compute in the end.

IIRC tokenization is usually the bottleneck, and I'd imagine the tokenized data is smaller than the original text, especially if you are able to filter out tokens that aren't relevant to the query.

westonpace:
> I feel like a very efficient way would be to tokenize in parallel, and then collect all tokenized data and compute in the end.
> 
> IIRC tokenization is usually the bottleneck, and I'd imagine the tokenized data is smaller than the original text, especially if you are able to filter out tokens that aren't relevant to the query.

This should be doable!  We can tokenize and discard all tokens that aren't in the query and then accumulate.  We might also need to accumulate token counts for each row but that should be small too.  Maybe for a first pass, if we exceed some limit (e.g. 1GB) then we log a warning and just keep going (eventually OOM).  The warning could be something like...

```
Accumualted more than 1GB of tokenized flat search data.  This means there is a lot of unindexed text data being searched.  The search is going to be very RAM intensive and may eventually OOM.  Create (or update) an index on the text column to avoid this issue.
```

westonpace:
Both of claude's suggestions are valid.  I will revisit this weekend / Monday when I have time to add some regression tests for these cases.

westonpace:
> This should be doable! We can tokenize and discard all tokens that aren't in the query and then accumulate. We might also need to accumulate token counts for each row but that should be small too. Maybe for a first pass, if we exceed some limit (e.g. 1GB) then we log a warning and just keep going (eventually OOM). The warning could be something like...

Ok, I've implemented this approach.  We now only accumulate (2 + N) * u64.  The first two u64 are row id and total tokens in doc.  The N u64s are the token count for each token in the query.  Since query strings are relatively bounded this should be relatively modest in the M's of rows.  Once we get to 100's of M or B's this will start to trigger OOMs and so it is important to create an FTS index before that.

rust/lance-index/src/scalar/inverted/index.rs
```diff
@@ -2448,9 +2452,10 @@ fn do_flat_full_text_search<Offset: OffsetSizeTrait>(
         let doc_array = batch[doc_col].as_string::<Offset>();
         for i in 0..row_id_array.len() {
             let doc = doc_array.value(i);
-            let doc_tokens = collect_doc_tokens(doc, &mut tokenizer, Some(&query_tokens));
-            if !doc_tokens.is_empty() {
+            if has_query_token(doc, &mut tokenizer, &query_tokens) {
                 results.push(row_id_array.value(i));
+                // What is this assertion for?  Why would doc contain query?  Don't we reach
+                // here only if they share at least one token?  Why is it not debug_assert?
```
BubbleCal, line 2431: ah i can't remember why i added this assertion but it looks not reasonable, free to remove it

rust/lance-index/src/scalar/inverted/index.rs
```diff
@@ -2459,62 +2464,270 @@ fn do_flat_full_text_search<Offset: OffsetSizeTrait>(
...
         let mut score = 0.0;
         for token in query_tokens {
-            let freq = doc_token_count.get(token).copied().unwrap_or_default() as f32;
-
+            let freq = query_token_counts_iter.next().expect_ok()? as f32;
             let idf = idf(scorer.num_docs_containing_token(token), scorer.num_docs());
             score += idf * (freq * (K1 + 1.0) / (freq + doc_norm));
+            if score > 0.0 {
+                row_ids_builder.append_value(row_id);
+                scores_builder.append_value(score);
```
BubbleCal: it seems to append the row_id multiple times then we will get duplicated results if we have multiple tokens in the query?

rust/lance-index/src/scalar/inverted/index.rs
```diff
@@ -2459,62 +2464,270 @@ fn do_flat_full_text_search<Offset: OffsetSizeTrait>(
...
+        let row_id = row_ids_iter.next().expect_ok()?;
+        if num_tokens_in_doc == 0 {
+            continue;
         }
+        let doc_norm = K1 * (1.0 - B + B * num_tokens_in_doc as f32 / scorer.avg_doc_length());
         let mut score = 0.0;
         for token in query_tokens {
-            let freq = doc_token_count.get(token).copied().unwrap_or_default() as f32;
-
+            let freq = query_token_counts_iter.next().expect_ok()? as f32;
```
BubbleCal, line 2655: should we consume this before `continue` at line 2700?

rust/lance-index/src/scalar/rtree.rs
```diff
@@ -79,6 +80,56 @@ static RTREE_NULLS_SCHEMA: LazyLock<Arc<ArrowSchema>> = LazyLock::new(|| {
     )]))
 });
 
+/// A stream that reads the original training data back out of the index
+struct IndexReaderStream {
```
wjones127, line 83: Why was this moved here? is it only used by rtree now?

rust/lance-index/src/scalar/inverted/scorer.rs
```diff
@@ -60,7 +66,7 @@ impl MemBM25Scorer {
         self.total_tokens as f32 / self.num_docs as f32
     }
 
-    pub fn num_docs_containing_token(&self, token: &str) -> usize {
+    pub fn num_docs_containing_token(&self, token: &String) -> usize {
```
wjones127: Why this change?

rust/lance-index/src/scalar/inverted/index.rs
westonpace: Removed

rust/lance-index/src/scalar/inverted/index.rs
westonpace: Ouch, good catch.  I moved the append outside the for loop.

rust/lance-index/src/scalar/inverted/index.rs
westonpace: Yes, good catch.

rust/lance-index/src/scalar/rtree.rs
westonpace: Yes, and since rtree is feature gated we get "unused" warnings when the feature is disabled.

rust/lance-index/src/scalar/inverted/scorer.rs
westonpace: Reverted.  I swear at one point it was complaining but it seems fine now.

rust/lance-arrow/src/stream.rs
```diff
@@ -0,0 +1,113 @@
...
+
+            if state.accumulated.is_empty() {
+                return Ok(None);
+            }
+
+            let batch = if state.accumulated.len() == 1 {
+                state.accumulated.pop().unwrap()
+            } else {
+                let b =
+                    arrow_select::concat::concat_batches(&state.input_schema, &state.accumulated)
```
Xuanwo, line 76: The logic here appears to be quite sensitive to `max_bytes`. Could our data in `state.accumulated` be repeatedly concatenated and sliced?

For example, if `max_bytes` is 1MiB and we got 64MiB data, the data inside `state.accumulated` with be `64 -> 63 -> 62 ... -> 2 -> 1`. Should we maintain an offset about the sliced data and make sure we only concat once on the raw input?

westonpace: Good catch.  I've modified the code so it only concatenates if the first slice is not large enough (smaller than min_bytes).  Since min_bytes should be sufficiently less than max_bytes I think we should be good in most cases.  In the event we have some really large outlier row and, as a result, slice inappropriately, we may still concatenate but I think that's enough of an outlier for the moment.

PR #6059, opened 2026-02-28
perf: add dict-values compression controls with lz4 default

Xuanwo:
This PR adds dictionary-values-specific general compression controls and aligns docs with current dictionary encoding behavior.

This PR also changed the default compression of dict values from `zstd` to `lz4` as it's a better trade off of speed and size. Users prefer smaller size can always pick change to zstd by setting the newly added metadata keys.

- Added new metadata keys:
  - `lance-encoding:dict-values-compression`
  - `lance-encoding:dict-values-compression-level`
- Added new env var fallbacks:
  - `LANCE_ENCODING_DICT_VALUES_COMPRESSION`
  - `LANCE_ENCODING_DICT_VALUES_COMPRESSION_LEVEL`
- Set default dictionary-values general compression scheme to `lz4`.

---

**Parts of this PR were drafted with assistance from Codex (with `gpt-5.3-codex`) and fully reviewed and edited by me. I take full responsibility for all changes.**

### DIFF
--- a/docs/src/format/file/encoding.md
+++ b/docs/src/format/file/encoding.md
@@ -624,19 +624,23 @@ BSS is particularly effective for:
 
 #### Dictionary Encoding Controls
 
-Dictionary encoding is attempted for supported primitive pages and gated by a few heuristics.
+Dictionary encoding is gated by a few heuristics.
+The decision is made on the leaf value page, so nested types can still benefit.
+For example, `List<u32>` can use dictionary encoding for its `u32` values.
 
-- `lance-encoding:dict-divisor` (default `2`): cardinality budget starts from `num_values / divisor`
-- `lance-encoding:dict-size-ratio` (default `0.8`): encoded dictionary representation must stay below this ratio of raw size
+Two field-level metadata keys control when dictionary encoding is attempted:
+
+- `lance-encoding:dict-divisor` (default `2`): the encoder computes a unique-value budget as `num_values / divisor`
+- `lance-encoding:dict-size-ratio` (default `0.8`): the estimated dictionary-encoded representation must stay below this ratio of the raw page size
 
 There are additional global guards available as environment variables:
 
-- `LANCE_ENCODING_DICT_TOO_SMALL` (minimum page size before trying dictionary encoding, default `100`)
-- `LANCE_ENCODING_DICT_DIVISOR` (fallback divisor when field metadata is not set)
+- `LANCE_ENCODING_DICT_TOO_SMALL` (minimum page size before trying dictionary encoding, default `100` values)
+- `LANCE_ENCODING_DICT_DIVISOR` (fallback divisor when field metadata is not set, default `2`)
 - `LANCE_ENCODING_DICT_MAX_CARDINALITY` (upper cap for dictionary entries, default `100000`)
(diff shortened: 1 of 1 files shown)


docs/src/format/file/encoding.md
```diff
@@ -620,18 +622,42 @@ BSS is particularly effective for:
...
 
-#### Dictionary Divisor
+#### Dictionary Encoding Controls
 
-Currently this is used to determine whether or not we apply dictionary encoding. First, we use HLL to estimate
-the number of unique values in the column. Then we divide the number of total values by the divisor to get a
-threshold. If the number of unique values is less than the threshold then we apply dictionary encoding. The
-configuration variable defines the divisor that we apply and it defaults to 2 which means we apply dictionary
-encoding if we estimate that less than half the values are unique.
+Dictionary encoding is attempted for supported primitive pages and gated by a few heuristics.
```
westonpace: Aren't all pages primitive pages?  A naive read might be that we cannot do dictionary encoding of something like `List<u32>` but I think we can?

docs/src/format/file/encoding.md
```diff
@@ -620,18 +622,42 @@ BSS is particularly effective for:
...
 
-Dictionary encoding is effective for columns with low cardinality where the same values repeat many times.
-The dictionary is stored once per page and indices are stored in place of the actual values.
+- `lance-encoding:dict-divisor` (default `2`): cardinality budget starts from `num_values / divisor`
```
westonpace: I don't understand this sentence.  It might be worth having a paragraph explaining these two properties instead of a brief line for each.  Also, maybe we should avoid jargony words like `cardinality` in a user-facing doc.

docs/src/format/file/encoding.md
```diff
@@ -620,18 +622,42 @@ BSS is particularly effective for:
...
+- `lance-encoding:dict-size-ratio` (default `0.8`): encoded dictionary representation must stay below this ratio of raw size
 
-This is likely to change in future versions.
+There are additional global guards available as environment variables:
+
+- `LANCE_ENCODING_DICT_TOO_SMALL` (minimum page size before trying dictionary encoding, default `100`)
```
westonpace: Is this 100 values or 100 bytes?

docs/src/format/file/encoding.md
```diff
@@ -620,18 +622,42 @@ BSS is particularly effective for:
...
+- `LANCE_ENCODING_DICT_DIVISOR` (fallback divisor when field metadata is not set)
```
westonpace: Let's say what the default is here (we list it above)

docs/src/format/file/encoding.md
```diff
@@ -620,18 +622,42 @@ BSS is particularly effective for:
...
+- `LANCE_ENCODING_DICT_MAX_CARDINALITY` (upper cap for dictionary entries, default `100000`)
+- `LANCE_ENCODING_DICT_SIZE_RATIO` (fallback ratio when field metadata is not set)
```
westonpace: Let's add the default here (we list it above)

Xuanwo:
I'm going to merge with all comments addressed.

PR #6066, opened 2026-03-02
feat(blob): map external blob URIs to multi-base base ids

Xuanwo:
Blob v2 external references were still treated as direct absolute URIs and did not integrate with Lance multi-base routing. This change makes external blob descriptors base-aware by storing `blob_id` as external base id (for `kind=External`) and writing relative paths when a URI maps to a registered base.

The write path now defaults to rejecting external URIs outside dataset/registered bases, with an explicit opt-in (`allow_external_blob_outside_bases`) for absolute outside-base URIs.

Context:
- Aligns external blob behavior with existing multi-base semantics and base-aware read routing.
- Includes Rust/Python API plumbing and docs updates for the new write policy.


rust/lance/src/dataset/blob.rs
```diff
@@ -916,6 +1070,36 @@ async fn collect_blob_files_v2(
...
+                "External URI '{}' is outside dataset bases and is not a valid absolute URI",
+                uri
+            ),
+            location!(),
+        )
+    })?;
+    Ok(url.to_string())
+}
+
+fn is_absolute_external_uri(base_id: u32, uri: &str) -> bool {
```
jackye1995: I think there is a easy way we can avoid parsing uri to decide this - `base_id = 0` means table root (which already is, user created bases always start with 0), and in that case `base_id` is `None` can mean it's always absolute

rust/lance/src/dataset/blob.rs
```diff
@@ -28,6 +29,91 @@ const INLINE_MAX: usize = 64 * 1024; // 64KB inline cutoff
...
+            return Ok(cached);
+        }
+
+        let uri_store_prefix = self
+            .store_registry
+            .calculate_object_store_prefix(uri, self.store_params.storage_options())?;
+        let uri_path = ObjectStore::extract_path_from_uri(self.store_registry.clone(), uri)?;
+
+        let mut best_match: Option<(usize, ResolvedExternalBase)> = None;
+        for candidate in &self.candidates {
```
jackye1995, line 75: if the candidate matches prefix, but if `is_dataset_root=true`, we should also not allow the match because the expectation is that the base should be treated as if a lance dataset root, which does not really match the use case of external blobs.

jackye1995:
Thanks for adding this! 2 comments and the rest looks good to me!

rust/lance/src/dataset/blob.rs
Xuanwo: In this case, users are missing use bucket to host a lance file along with the external data. For example, users are having a bucket in this way:

```text
/abc
/abc/path/to/lance/
/abc/path/to/lance/media/videos // not part of lance but stored together
```

What do you think?

jackye1995: > In this case, users are missing use bucket to host a lance file along with the external data.

Oh no they won't. It's just the user needs to define 2 bases, 1 with s_dataset_root=true and one with s_dataset_root=false for different use cases.

Xuanwo: I see. Will update the PR to reflect this. Thanks for clarifying.

jackye1995: It should also match your example well, we are basically explicitly saying this location can be used both for Lance dataset and also blobs

Xuanwo: Addressed. PTAL

Xuanwo: > It should also match your example well, we are basically explicitly saying this location can be used both for Lance dataset and also blobs

Oh, yeah. We implemented the blob dataset in a different way!

PR #6071, opened 2026-03-02
refactor: upgrade to SNAFU 0.9

shepmaster:
Thanks for using SNAFU! I've just released version 0.9 and my tests caught that your crate would be affected by some of the SemVer-incompatible changes in this version. See the [CHANGELOG][] for a full list.

This commit tries to make the minimal amount of changes to get your code compiling again, but it's entirely possible that there are better solutions for your specific case.

This version also updates the minimum supported Rust version to Rust 1.65 and targets Rust 1.81 by default. I didn't check how these changes affect might your crate, so please make sure to use your judgment as well!

Please feel free to take this PR in whatever form is useful to you; I'm not interested in signing anything like a CLA, but I don't mind someone with project approval yoinking these changes and committing it themselves.

[CHANGELOG]: https://github.com/shepmaster/snafu/blob/main/CHANGELOG.md#090---2026-03-02

### DIFF
--- a/python/src/dataset.rs
+++ b/python/src/dataset.rs
@@ -1,7 +1,6 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-FileCopyrightText: Copyright The Lance Authors
 
-
 use std::collections::HashMap;
 use std::str;
 use std::sync::Arc;
--- a/python/src/dataset/commit.rs
+++ b/python/src/dataset/commit.rs
@@ -12,7 +12,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-
 use std::fmt::Debug;
 use std::sync::LazyLock;
 
@@ -35,14 +34,20 @@ fn handle_error(py_err: PyErr, py: Python) -> CommitError {
     let conflict_err_type = match &*PY_CONFLICT_ERROR {
         Ok(err) => err.bind(py).get_type(),
         Err(import_error) => {
-            return CommitError::OtherError(Error::internal(format!("Error importing from pylance {}", import_error)));
+            return CommitError::OtherError(Error::internal(format!(
+                "Error importing from pylance {}",
+                import_error
+            )));
         }
     };
 
     if py_err.is_instance(py, &conflict_err_type) {
         CommitError::CommitConflict
     } else {
-        CommitError::OtherError(Error::internal(format!("Error from commit handler: {}", py_err)))
+        CommitError::OtherError(Error::internal(format!(
+            "Error from commit handler: {}",
+            py_err
+        )))
     }
 }
 
--- a/python/src/fragment.rs
+++ b/python/src/fragment.rs
@@ -12,7 +12,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-
 use std::fmt::Write as _;
 use std::sync::Arc;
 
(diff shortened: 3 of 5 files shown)


rust/lance-core/src/error.rs
```diff
@@ -300,7 +300,7 @@ pub trait ToSnafuLocation {
 
 impl ToSnafuLocation for std::panic::Location<'static> {
     fn to_snafu_location(&'static self) -> snafu::Location {
-        snafu::Location::new(self.file(), self.line(), self.column())
+        self
```
Xuanwo: Do we still need this trait? Seems we can just use `std::panic::Location` as `snafu::Location`?

> You only need to answer the question, I can do the follow up changes.

shepmaster:
I previously identified lance-core as the main crate that needed fixes, but doing a spot-check (only after submitting the PR, of course), I see that other places need to be updated, so I've moved to draft.

rust/lance-core/src/error.rs
shepmaster: Yes, `snafu::Location` is now an alias for the standard library's type. However, I think removing the trait would count as SemVer-breaking for your code, so I opted out of doing that.

shepmaster:
I will add that I think that this code could use SNAFU more effectively. In most cases, you shouldn't need to create a `Location` by hand, instead relying on [implicit data](https://docs.rs/snafu/latest/snafu/derive.Snafu.html#controlling-implicitly-generated-data) (note this is the docs for 0.8 as 0.9 hasn't been built yet).

For example, you could make a change roughly like this:

```diff
diff --git a/rust/lance-core/src/error.rs b/rust/lance-core/src/error.rs
index 02cac0da..6edbb372 100644
--- a/rust/lance-core/src/error.rs
+++ b/rust/lance-core/src/error.rs
@@ -4,7 +4,7 @@
 use std::fmt;
 
 use arrow_schema::ArrowError;
-use snafu::{Location, Snafu};
+use snafu::{IntoError as _, Location, Snafu};
 
 type BoxedError = Box<dyn std::error::Error + Send + Sync + 'static>;
 
@@ -57,6 +57,7 @@ pub enum Error {
     #[snafu(display("Invalid user input: {source}, {location}"))]
     InvalidInput {
         source: BoxedError,
+        #[snafu(implicit)]
         location: Location,
     },
     #[snafu(display("Dataset already exists: {uri}, {location}"))]
@@ -189,12 +190,9 @@ impl Error {
         }
     }
 
-    pub fn invalid_input(message: impl Into<String>, location: Location) -> Self {
-        let message: String = message.into();
-        Self::InvalidInput {
-            source: message.into(),
-            location,
-        }
+    #[track_caller]
+    pub fn invalid_input(source: impl Into<String>) -> Self {
+        InvalidInputSnafu.into_error(source.into().into())
     }
 
     pub fn io(message: impl Into<String>, location: Location) -> Self {
``` 

That would then allow you to remove all the manual construction of `Location`.

Xuanwo:
Thank you @shepmaster for the suggestion, working on this now

rust/lance-core/src/error.rs
```diff
@@ -189,19 +196,17 @@ impl Error {
...
     }
 
-    pub fn io(message: impl Into<String>, location: Location) -> Self {
+    #[track_caller]
+    pub fn io(message: impl Into<String>) -> Self {
         let message: String = message.into();
         Self::IO {
             source: message.into(),
-            location,
+            location: location!(),
```
shepmaster: These cases probably want to use `implicit` as well, like `invalid_input` above.

PR #6072, opened 2026-03-02
feat(compaction): single reserve_fragment_ids after rewriting files

hamersaw:
This PR fixes the issue with concurrent fragment ID reservations during parallelized compactions by moving both the fragment reservation operations and  the row ID remapping out of each individual task compaction and into the main thread. This ensures that fragment reservation occurs exactly once during compaction.

This is an alternative approach to pre-computing fragment IDs based on fragment count estimates. Discussion comparing the approaches should occur on the issue to maintain a central thread.

Closes https://github.com/lance-format/lance/issues/6075

### DIFF
--- a/rust/lance/src/dataset/optimize.rs
+++ b/rust/lance/src/dataset/optimize.rs
@@ -867,16 +867,14 @@ pub struct RewriteResult {
     pub read_version: u64,
     /// The original fragments being replaced
     pub original_fragments: Vec<Fragment>,
-    /// Serialized `RoaringTreemap` of the original row addresses read from the
-    /// source fragments during compaction.
+    /// Serialized `RoaringTreemap` of the row addresses from the original
+    /// fragments that were read during compaction.
     ///
-    /// `Some` for datasets using address-style row IDs. During commit, these
-    /// addresses are transposed to the new fragment layout to build an old-to-new
-    /// mapping for index remapping. Also triggers fragment ID reservation since
-    /// address-style fragments don't have pre-allocated IDs.
-    ///
-    /// `None` for datasets using stable row IDs, which handle remapping via
-    /// row ID sequences instead.
+    /// - `None` when configured with stable row IDs because the row ID
+    ///   sequences are rechunked directly.
+    /// - `Some` then these addresses are either (1) written to storage for
+    ///   deferred index remap post-processing, or (2) used with reserved
+    ///   fragment IDs to build old-to-new mappings.
     pub row_addrs: Option<Vec<u8>>,
 }
 
@@ -1324,12 +1322,11 @@ pub async fn commit_compaction(
                 row_id_map.extend(transposed);
             }
         } else if options.defer_index_remap {
-            let changed_row_addrs = task.row_addrs.unwrap_or_else(|| {
-                let empty = RoaringTreemap::new();
-                let mut bytes = Vec::with_capacity(empty.serialized_size());
-                empty.serialize_into(&mut bytes).unwrap();
-                bytes
-            });
+            let changed_row_addrs = task.row_addrs.ok_or_else(|| {
+                Error::internal(
+                    "defer_index_remap requires row_addrs but none were provided".to_string(),
+                )
+            })?;
             frag_reuse_groups.push(FragReuseGroup {
                 changed_row_addrs,
                 old_frags: task.original_fragments.iter().map(|f| f.into()).collect(),


rust/lance/src/dataset/optimize.rs
```diff
@@ -867,13 +868,9 @@ pub struct RewriteResult {
...
     pub original_fragments: Vec<Fragment>,
-    /// A HashMap of original row IDs to new row IDs or None (deleted)
-    /// Only set when index remap is done as a part of the compaction
-    pub row_id_map: Option<HashMap<u64, Option<u64>>>,
-    /// the changed row addresses in the original fragment
-    /// in the form of serialized RoaringTreemap
-    /// Only set when index remap is deferred after compaction
-    pub changed_row_addrs: Option<Vec<u8>>,
+    /// Serialized RoaringTreemap of row addresses read from the original fragments.
+    /// Some for address-style row IDs, None for stable row IDs.
```
LuQQiu: Some for address-style row IDs, None for stable row IDs.? 
a little confusing for stable row id ones

rust/lance/src/dataset/optimize.rs
```diff
@@ -1322,11 +1309,27 @@ pub async fn commit_compaction(
...
+            if let Some(row_addrs_bytes) = task.row_addrs {
+                let row_addrs =
+                    RoaringTreemap::deserialize_from(&mut Cursor::new(&row_addrs_bytes))?;
+                let transposed = remapping::transpose_row_addrs(
+                    row_addrs,
+                    &task.original_fragments,
+                    &task.new_fragments,
+                );
+                row_id_map.extend(transposed);
+            }
```
zhangyue19921010, line 1323: Thanks for your effort here. As we discuss compared to pre-allocation https://github.com/lance-format/lance/pull/6004, the main difference here is that row_id_map has been changed from parallel computing to serial computing.

rust/lance/src/dataset/optimize.rs
hamersaw: Updated the comment to be a little more clear. Let me know what you think!

rust/lance/src/dataset/optimize.rs
```diff
@@ -866,13 +867,17 @@ pub struct RewriteResult {
...
+    /// Serialized `RoaringTreemap` of the original row addresses read from the
+    /// source fragments during compaction.
+    ///
+    /// `Some` for datasets using address-style row IDs. During commit, these
+    /// addresses are transposed to the new fragment layout to build an old-to-new
+    /// mapping for index remapping. Also triggers fragment ID reservation since
+    /// address-style fragments don't have pre-allocated IDs.
+    ///
+    /// `None` for datasets using stable row IDs, which handle remapping via
+    /// row ID sequences instead.
```
westonpace: How does a single bitmap express a mapping?  Can we expand on this comment a bit to explain?

rust/lance/src/dataset/optimize.rs
```diff
@@ -1316,11 +1311,27 @@ pub async fn commit_compaction(
...
+                    RoaringTreemap::deserialize_from(&mut Cursor::new(&row_addrs_bytes))?;
+                let transposed = remapping::transpose_row_addrs(
+                    row_addrs,
+                    &task.original_fragments,
+                    &task.new_fragments,
+                );
+                row_id_map.extend(transposed);
+            }
         } else if options.defer_index_remap {
+            let changed_row_addrs = task.row_addrs.unwrap_or_else(|| {
```
westonpace: Should it be an error if `row_addrs` is not set in this branch?  Or does that just mean stable row ids?

rust/lance/src/dataset/optimize.rs
westonpace: Is the transpose a significant amount of work compared to the I/O cost of the rewrites themselves?  If so I wonder if we can optimize the transposition process somehow.

hamersaw: I tested compaction with 100k fragments where each has 16 rows of <int32, int32>. So I assume this is close to worst case scenario of the highest row count (transposition cost) for I/O cost - actually I guess a couple large fragments may be worse. I compacted down to 144 fragments with 24 parallelism:
pre-allocate: 21m49s
post-allocate: 22m15s

So like 26s more for serial transposition, a little over 1% slower. IIUC it's an entirely in-memory operation, and even if done in parallel we need to aggregate these together in the main thread before writing. I was more so worried about memory utilization on 1b+ rows, but RAM should be the same between them.

rust/lance/src/dataset/optimize.rs
hamersaw: Yeah, this is my third iteration of the comment already - it's a tough one. This is just the sequence of origin row IDs. The `option` is because they are only useful if we're not using stable row IDs, I tried to address Lu's previous comment by explaining how they are used a bit. Basically, if stable row ids - we don't need them, if defer_index_remap - we flush, otherwise we use them (along with post-allocated fragment IDs) for the remapping.

I'll take another stab at making this more clear!

rust/lance/src/dataset/optimize.rs
hamersaw: Yes, this will silently produce errors in remapping. I've updated to return an error. Basically, we can get this branch if `defer_index_remap` and `use_stable_row_ids` are configured in tandem, which is an invalid configuration that IIR we catch elsewhere. Even if we cannot reach that case, we should explicitly fail here.

rust/lance/src/dataset/optimize.rs
westonpace: I think I see and I remember the problem now.  We calculate the target ids on the fly once we get the fragment ids because it's easy (one contiguous sequence per fragment based on id).  The source ids are trickier because we don't know how many holes there were in the source fragments due to deletions.  So these need to be written down (although as I'm explaining this part of me wonders if we could just use the deletion maps on the source fragments instead but let's not make this more complicated than it needs to be :laughing:)

hamersaw: Oh sure, so we could potentially not even need this field if we wanted to get fancy. I convinced myself that this is replacing the existing two fields and will be strictly smaller (bytes-wise), so it shouldn't be an issue. But if we think it's worth diving into further optimizations, I could take a stab at it.

rust/lance/src/dataset/optimize.rs
```diff
@@ -1316,11 +1309,26 @@ pub async fn commit_compaction(
             old_fragments: task.original_fragments.clone(),
             new_fragments: task.new_fragments.clone(),
         };
+
         if needs_remapping {
-            row_id_map.extend(task.row_id_map.unwrap());
+            if let Some(row_addrs_bytes) = task.row_addrs {
```
Xuanwo, line 1314: It seems impossible for `needs_remapping` to be true while `task.row_addrs` is none. However, if it does happen, we currently ignore it silently. Should we return an error instead, or add a comment to clarify that ignoring it is the intended behavior?

rust/lance/src/dataset/optimize.rs
```diff
@@ -1303,6 +1283,19 @@ pub async fn commit_compaction(
...
+    // Single reserve_fragment_ids for all address-style tasks
+    let has_address_style = completed_tasks.iter().any(|t| t.row_addrs.is_some());
+    if has_address_style {
+        let frags: Vec<&mut Fragment> = completed_tasks
+            .iter_mut()
+            .filter(|t| t.row_addrs.is_some())
+            .flat_map(|t| t.new_fragments.iter_mut())
+            .collect();
+        reserve_fragment_ids(dataset, frags.into_iter()).await?;
+    }
```
Xuanwo, line 1297: Nit: we can avoid double loop and `Vec` here:

```rust
let mut frags = completed_tasks
    .iter_mut()
    .filter(|t| t.row_addrs.is_some())
    .flat_map(|t| t.new_fragments.iter_mut())
    .peekable();

let has_address_style = frags.peek().is_some();

if has_address_style {
    reserve_fragment_ids(dataset, frags).await?;
}
```

rust/lance/src/dataset/optimize.rs
```diff
@@ -1316,11 +1309,26 @@ pub async fn commit_compaction(
             old_fragments: task.original_fragments.clone(),
             new_fragments: task.new_fragments.clone(),
         };
+
         if needs_remapping {
-            row_id_map.extend(task.row_id_map.unwrap());
+            if let Some(row_addrs_bytes) = task.row_addrs {
+                let row_addrs =
+                    RoaringTreemap::deserialize_from(&mut Cursor::new(&row_addrs_bytes))?;
```
Xuanwo, line 1316: Maybe we can store the `row_addrs_bytes` in `Bytes` so that we don't need the `Cursor` help.

zhangyue19921010:
This Post allocation scheme is indeed simpler and more reasonable, **+1** For this :)
In addition, regarding the performance issue of row_id_map in large data volume scenarios, on the one hand, we can try parallel processing if needed; on the other hand, we can also split a huge Compaction into multiple relatively medium-sized Compactions and execute them step by step, which is mentioned here https://github.com/lance-format/lance/issues/6039

PR #6074, opened 2026-03-02
feat: expose compaction binary copy configuration through python and java SDKs

hamersaw:
This PR exposes the configuration to enable binary copy during compaction operations through the python and Java SDKs.

### DIFF
--- a/rust/lance/src/dataset/optimize.rs
+++ b/rust/lance/src/dataset/optimize.rs
@@ -136,13 +136,10 @@ impl TryFrom<&str> for CompactionMode {
             "reencode" => Ok(Self::Reencode),
             "try_binary_copy" => Ok(Self::TryBinaryCopy),
             "force_binary_copy" => Ok(Self::ForceBinaryCopy),
-            _ => Err(Error::invalid_input(
-                format!(
-                    "Invalid compaction mode \"{}\". Valid values: \"reencode\", \"try_binary_copy\", \"force_binary_copy\"",
-                    value
-                ),
-                location!(),
-            )),
+            _ => Err(Error::invalid_input(format!(
+                "Invalid compaction mode \"{}\". Valid values: \"reencode\", \"try_binary_copy\", \"force_binary_copy\"",
+                value
+            ))),
         }
     }
 }


python/python/lance/optimize.py
```diff
@@ -59,3 +59,21 @@ class CompactionOptions(TypedDict):
...
+    Whether to enable binary copy optimization during compaction. When
+    enabled, compaction will skip re-encoding data and directly copy
+    binary data for faster compaction times. (default: False)
+    """
+    enable_binary_copy_force: Optional[bool]
+    """
+    Whether to force binary copy optimization. If true, compaction will
+    fail if binary copy is not supported for the given fragments.
+    (default: False)
+    """
```
Xuanwo: Sorry for raising this too late. However give the fact that we are going to expose this options to users at python level (the REAL API we exposed to user), maybe it's the best time.

----

This makes me think why we need those two options for the first place. Should this be a mode of `binary_copy`? I also think the naming is confusing. Maybe we can introdue something called `compaction_mode`.

cc @jackye1995 @westonpace @wjones127

hamersaw: Are you thinking update the rust API? Or just in the Python SDK have `binary_copy_mode=None|enabled|forced`?

Xuanwo: > Are you thinking update the rust API? Or just in the Python SDK have `binary_copy_mode=None|enabled|forced`?

I'm thinking update the rust API too to get them aligned.

hamersaw: Sure, just let me know either way and I'm happy to make the changes. Would be good to get this out ASAP!

Xuanwo: I tend to have a compaction_mode which has three values: auto, rewrite, binary_copy(maybe just copy, binary copy means nothing to our user)

The default value is auto which means while possible, we will do binary copy for compaction. otherwise, we'll rewrite the data.

In this way, our user doesn't need to bother about the option by default. But at the same time, they are in control.

hamersaw: @Xuanwo how does this look?

rust/lance/src/dataset/optimize.rs
```diff
@@ -159,25 +189,24 @@ pub struct CompactionOptions {
...
-    /// used when compacting small files together because the pages in the
-    /// compacted file will be too small and this could lead to poor I/O patterns.
+    /// The compaction mode to use. When set, this takes priority over the
+    /// deprecated `enable_binary_copy` and `enable_binary_copy_force` fields.
     ///
-    /// Defaults to false.
+    /// Defaults to `None` (falls back to legacy boolean fields).
+    pub compaction_mode: Option<CompactionMode>,
+    /// Deprecated: use `compaction_mode` instead.
+    #[deprecated(note = "Use `compaction_mode` instead")]
```
Xuanwo, line 194: I think we can just remove those flags if we don't have plan to cherry-pick this commit to old releases.

rust/lance/src/dataset/optimize.rs
```diff
@@ -190,26 +219,43 @@ impl Default for CompactionOptions {
...
             self.materialize_deletions = false;
         }
     }
+
+    /// Returns the effective [`CompactionMode`], preferring the new
+    /// `compaction_mode` field and falling back to the deprecated boolean
+    /// fields for backwards compatibility.
+    pub fn compaction_mode(&self) -> CompactionMode {
+        if let Some(mode) = &self.compaction_mode {
+            return mode.clone();
```
Xuanwo: compaction_mode is just an enum, we can mark it as `Copy`

rust/lance/src/dataset/optimize.rs
```diff
@@ -190,26 +219,43 @@ impl Default for CompactionOptions {
...
+        }
+        // Fall back to deprecated booleans
+        match (self.enable_binary_copy, self.enable_binary_copy_force) {
```
Xuanwo, line 243: Let remove those logic unless we want to backport it

Xuanwo (commented):

However, I still think it would be better to change the compaction mode to a more user-friendly naming convention. For example, we could use an `auto` option instead of `try_binary_copy`, and enable binary copy by default.

But I don't feel strongly about this. I'm fine with merging the current design.

rust/lance/src/dataset/optimize.rs
hamersaw: Same as previous "breaking change" question.

rust/lance/src/dataset/optimize.rs
hamersaw: It is very cheap to keep this backwards compatible. If we remove these don't we have to label this as a "breaking change" and increment versions accordingly?

hamersaw:
>However, I still think it would be better to change the compaction mode to a more user-friendly naming convention. For example, we could use an auto option instead of try_binary_copy, and enable binary copy by default.

The page boundaries of resultant fragments are going to be different between binary copy and the existing approach, where binary copied fragments may have less efficient read accesses because of padding bytes between the source fragment pages to align on word boundaries. I think for most of our use-cases trading improved compaction speed for less performant reads is a bad decision and certainly should not be the default. If anything, it should be something discussed at some depth, with benchmarks to back the decision, and probably makes sense to address in a future PR if we decide to do so.

rust/lance/src/dataset/optimize.rs
hamersaw: Done!

rust/lance/src/dataset/optimize.rs
Xuanwo: Ideally, yes. However, since this feature has never been exposed, we have some freedom to simply remove it and reduce future maintenance burden. That said, I don't have a strong opinion on this, I'll leave the decision to you.

PR #6079, opened 2026-03-03
feat: pluggable partition computers

wkalt:
This adds a new trait for pluggable partitioner computers, as well as two new implementations -- GemmPartitionComputer and GpuPartitionComputer.

These two partitioners are faster than the Lance partitioner in higher partition counts.


wkalt:
partition_computer_benchmark

This is captured from the included benchmark. I'm running an AMD 7950x and an Nvidia 3080Ti with cuda version 12.2 and driver 535.247.01

PR #6080, opened 2026-03-03
feat: support versions count

### LINKED ISSUE #6081: Support versions count
Currently the `Dataset` does not have a function to get the number of versions; it can only obtain indirectly by getting the manifest list through `versions()` and its length. However, the `versions()` actually reads the contents of each manifest file, result in a significant overhead.

https://github.com/lance-format/lance/blob/39f8a2645ac360e25ba208d6adad29f2c0277920/rust/lance/src/dataset.rs#L1813-L1830

It is necessary to implement a `versions_count()` function to get the number of versions. For example, it can be used as a factor to trigger CleanupOldVersions.

ztorchan:
close https://github.com/lance-format/lance/issues/6081

### DIFF
--- a/rust/lance/src/dataset.rs
+++ b/rust/lance/src/dataset.rs
@@ -1829,6 +1829,15 @@ impl Dataset {
         Ok(versions)
     }
 
+    /// List existing version numbers
+    pub async fn list_versions_number(&self) -> Result<Vec<u64>> {
+        self.commit_handler
+            .list_manifest_locations(&self.base, &self.object_store, false)
+            .map_ok(|location| location.version)
+            .try_collect::<Vec<u64>>()
+            .await
+    }
+
     /// Get the latest version of the dataset
     /// This is meant to be a fast path for checking if a dataset has changed. This is why
     /// we don't return the full version struct.


majin1102:
Thanks for woring on this.
A few suggestions:

1. We need add test for this
2. You metioned we could use this as the factor of cleanOldVersions(). Could you elaborate more on this? Maybe we could leverage this function for cleaning in this PR?
3. Agree with the idea of `Unnecessary allocation`
4. This might be a little duplicated with #5885. But I think we are good to go if we can adress the points above. What do you think?

rust/lance/src/dataset.rs
```diff
@@ -1829,6 +1829,17 @@ impl Dataset {
         Ok(versions)
     }
 
+    // Get versions count
+    pub async fn versions_count(&self) -> Result<u64> {
+        let locations = self
+            .commit_handler
+            .list_manifest_locations(&self.base, &self.object_store, false)
+            .try_collect::<Vec<_>>()
```
westonpace: Minor nit: https://docs.rs/futures/latest/futures/stream/trait.StreamExt.html#method.count may be useful here

ztorchan:
Thank for reviewing.
I will reconsider the validity of this PR; perhaps a function to retrieve the latest and oldest existing versions would be more useful.

Xuanwo:
I think it should be usefull to have both `versions_count` and `oldest_version_id`.

ztorchan:
Maybe add a `dataset.list_manifest_locations()` to expose `commit_handler.list_manifest_locations()` would be better? So that we can obtain a valid list of manifests which including version information.
@westonpace @Xuanwo @majin1102

majin1102:
I'm actually quite supportive of adding this method. If we introduce the Timeline API down the road, it might be a more natural fit there—but I think that doesn't mean we can't add it now(even if we can't use it in cleanup, I was just curious on that).  I'm not sure we should expose the manifest location directly, since manifest operations are generally encapsulated within the dataset. Do you have any cases for that?

Xuanwo:
I agree with @majin1102 here that we need to think why we need to do this and is this the smallest change we need to make it work?

ztorchan:
There are two main reasons.
One scenario is that, we set up management nodes and worker nodes for datasets. The Cleanup operation is executed by the worker as an asynchronous task issued by management nodes. We want this task to be issued with certain credentials rather than simply executed periodically.
Secondly, we also want to be able to monitor the dataset's version status in real time, which is also a very important management information.

I agree with @majin1102 that maybe we should not expose the manifest location directly, what about a `list_versions()` method to return a `Vec<u64>` version number list?

ztorchan:
@majin1102 @Xuanwo @westonpace Please take a look when you have time, thanks a lot!

rust/lance/src/dataset.rs
```diff
@@ -1829,6 +1829,17 @@ impl Dataset {
         Ok(versions)
     }
 
+    // Get versions count
+    pub async fn versions_count(&self) -> Result<u64> {
```
westonpace: I do have some small concern which is I don't expect this to be any more efficient than calling `versions()` against default object storage.  I think cloud storage list calls are expensive (the version parsing cost will be negligible) and, unfortunately, they don't offer any API for speeding this kind of thing up.

So I guess a related question is whether we want to eventually push `versions_count` into the commit handler / manifest store?  If we do then we could at least open up the possibility that external stores can accelerate this operation somehow.

westonpace (changes requested):
I think having a "count rows" for the "versions table" is something we eventually want and this can be a placeholder / stepping stone as we work towards that.  So I'm supportive of adding the method.

@majin1102 's original point remains however.  We need tests before we can merge.

westonpace:
Or were you asking feedback on this question:

> Maybe add a dataset.list_manifest_locations() to expose commit_handler.list_manifest_locations() would be better?

In that case I'll defer to @majin1102 who has thought more about this than I have.  If he is supportive of it then it seems fine to me.

ztorchan:
Yes, only I want is a method to get existing versions number list or existing versions count without large overhead. I am looking for a reasonable way to implement it. Currently, it seems unavoidable to `list`; the only way is to prevent `get` operations on the Manifest files.

> So I guess a related question is whether we want to eventually push `versions_count` into the commit handler / manifest store? If we do then we could at least open up the possibility that external stores can accelerate this operation somehow.

I don't think this is possible, because the CleanupOldVersions operation reduces the existing versions count but doesn't commit a new manifest.

PR #6084, opened 2026-03-03
feat(cleanup): support rate limiter for cleanup operation

zhangyue19921010:
Closes https://github.com/lance-format/lance/issues/3291

```
    stats = dataset.cleanup_old_versions(
        older_than=(datetime.now() - moment), delete_rate_limit=100.0
    )
```

### DIFF
--- a/rust/lance/src/dataset/cleanup.rs
+++ b/rust/lance/src/dataset/cleanup.rs
@@ -825,18 +825,6 @@ fn calculate_duration(scheme: String, rate: u64) -> Duration {
     Duration::from_nanos(duration_ns)
 }
 
-/// quick get the object_store delete batch size based on different storage type.
-fn delete_stream_batch_size(object_store: &lance_io::object_store::ObjectStore) -> u64 {
-    let scheme = object_store.scheme().to_lowercase();
-    if scheme.contains("s3") {
-        S3_DELETE_STREAM_BATCH_SIZE
-    } else if scheme.contains("az") {
-        AZURE_DELETE_STREAM_BATCH_SIZE
-    } else {
-        1
-    }
-}
-
 #[derive(Clone, Debug)]
 pub struct CleanupPolicy {
     /// If not none, cleanup all versions before the specified timestamp.


python/python/tests/test_dataset.py
```diff
@@ -1525,6 +1525,35 @@ def test_enable_disable_auto_cleanup(tmp_path):
...
+    """Test that cleanup_old_versions works with delete_rate_limit parameter."""
+    table = pa.Table.from_pydict({"a": range(100), "b": range(100)})
+    base_dir = tmp_path / "test"
+
+    lance.write_dataset(table, base_dir, mode="create")
+    time.sleep(1)
+    lance.write_dataset(table, base_dir, mode="overwrite")
+    time.sleep(1)
+    lance.write_dataset(table, base_dir, mode="overwrite")
+    time.sleep(1)
```
westonpace: Are these sleeps needed?  Each write will get a unique timestamp.  You can fetch the timestamps with a get_versions call.  This will help avoid slow test times.

rust/lance/src/dataset/cleanup.rs
```diff
@@ -798,6 +822,11 @@ pub struct CleanupPolicy {
     pub error_if_tagged_old_versions: bool,
     /// If clean the referenced branches
     pub clean_referenced_branches: bool,
+    /// Maximum number of delete operations per second. If None, no rate limiting is applied.
+    ///
+    /// Use this to avoid hitting S3 (or other object store) request rate limits during cleanup.
+    /// For example, `Some(100.0)` limits deletions to 100 files per second.
```
Xuanwo: Our delete is using batch delete which is 1k files per op, so 100 limit her should be 100k files per second.

rust/lance/src/dataset/cleanup.rs
```diff
@@ -367,10 +371,30 @@ impl<'a> CleanupTask<'a> {
...
 
+        let paths_to_delete: BoxStream<Result<Path>> = if let Some(rate) =
+            self.policy.delete_rate_limit
+        {
+            info!(
+                "delete_rate_limit enable, limit {} ops/sec during cleanup",
+                rate
+            );
+            let duration = Duration::try_from_secs_f64(1.0 / rate).map_err(|e| Error::Cleanup {
+                message: format!("delete_rate_limit {} is too small: {}", rate, e),
```
Xuanwo: Setting the rate to a value that is too small does not seem logical to me. We can instead set it to the smallest meaningful value. For example, we could set the rate's minimum to `1`, which would mean one operation per second.

rust/lance/src/dataset/cleanup.rs
```diff
@@ -874,6 +904,27 @@ impl CleanupPolicyBuilder {
...
 
+    /// Limit the number of delete operations per second during cleanup.
+    ///
+    /// By default (None), deletions run at full speed. Set this to a positive value to
+    /// throttle deletions and avoid hitting object store request rate limits (e.g. S3 HTTP 503).
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if `rate` is not a positive finite number.
+    pub fn delete_rate_limit(mut self, rate: f64) -> Result<Self> {
```
Xuanwo: Do we really need to make this a float?

rust/lance/src/dataset/cleanup.rs
zhangyue19921010: Thanks for the reminder. I've looked into it. At the Object store level, there is an implementation of batch/buffer (bulk delete API):
- S3: try_chunks(1_000) accumulates every 1000 items in the path stream into a batch, then sends a bulk_delete_request with a concurrency of buffered(20).
  See aws/mod.rs:L272-L290
- Azure: Same pattern, but the chunk size is 256 and the concurrency is also buffered(20).
  See azure/mod.rs:L126-L144

However, I've thought about it, and the speed limiting here is achieved by **limiting the delivery speed of Path**, and its semantics is t**he average speed of deleting files**, that is, 

**`Some(100)` means we will delete 100 files per second(average delete speed)**

Assuming 10,000 files are to be deleted at a rate of 100 files/second:
- From 0s to 10s: The cleanup allows a total of 1000 paths; S3 has been "accumulating" and no actual deletion has occurred yet.
- Around 10s: S3 accumulates 1000 files and sends the first DeleteObjects request, deleting 1000 files in a short time.
- From 10s to 20s: The next 1000 files are allowed; the second request is sent around 20s to delete 1000 files.
- ... and so on ...
- Around 100s: A total of 10,000 files are deleted (total duration ≈ 10,000 / 100 = 100 seconds)

Conclusion: The average throughput is 100 files/second, but the deletion events occur in bursts every 10 seconds (deleting 1000 files at a time) rather than evenly deleting 100 files per second.

zhangyue19921010:
Hi @Xuanwo and @westonpace Thanks a lot for your view!
All comments are addressed. Also tested on S3 which will delete 3K files during clean. Result are as followed. 

| Scenario | Condition | Elapsed Time (ms) |
|---|---|---:|
| Unthrottled | No rate limit | 3,422 |
| Rate-limited | 1 op/s | 5,302 |

PR #6085, opened 2026-03-03
fix: avoid thread pool contention between compression and write operations during FTS indexing

BubbleCal:
`FileWriter` would spawn a thread from the CPU pool, and FTS indexing would do that for compressing data as well.
This could lead to dead lock because the 2 threads communicate by a bounded channel.


Xuanwo (approved):
Nice finding

Issue #5573, opened 2025-12-24
Add more external manifest store (e.g. MySQL, PostgreSQL)


zhenghaozzz:
Lance support DynamoDB vs an external manifest store for concurrent write. If it is possible to support more external manifest store such as MySQL, PostgreSQL, etc.

https://github.com/lance-format/lance/blob/341a599dfaa376078f5c77bda7e000d1aaa249a3/rust/lance-table/src/io/commit/dynamodb.rs#L143-L148

I can help to contribute if "support more external manifest store" make sense.

Xuanwo:
Interesting idea. For context, we are currently in the process of removing support for DynamoDB.

wjones127:
Why do you want additional external manifest stores? Most object stores now support some kind of atomic condition out operation, which makes the external stores unnecessary. The existing dynamodb one was just a workaround for when S3 did not support this.

zhenghaozzz:
Thank you. I will use S3 directly.

PR #6091, opened 2026-03-04
feat: support sql as external manifest store

### LINKED ISSUE #5751: How can I implement a single commit handler when writing to a dataset using both Rust and Python simultaneously?

hushengquan:
### 🔍 Related Issue
- Fixes: [#5751](https://github.com/lance-format/lance/issues/5751)
- Fixes:[#5573](https://github.com/lance-format/lance/issues/5573)

### 📝 Background & Motivation
Certain object storage systems lack support for the `put-if-not-exist` semantics, which prevents direct guarantees of transaction atomicity. This pull request introduces a standard SQL-based `ExternalManifest` solution to address this limitation, leveraging `ExternalManifest` to ensure transaction atomicity.

#### 1. New URL Scheme: `s3+sql://`

```
s3+sql://{bucket}/{path}[?sqlTableName=...&sqlMaxConnections=...&...]
```

- Routes through `SqlExternalManifestStore` automatically.
- Mutually exclusive with a custom `commit_handler` — passing both is
  rejected at write time.

Supported drivers (via `sqlx` `any` runtime):

| Scheme       | Database   |
|--------------|------------|
| `postgres://`, `postgresql://` | PostgreSQL |
| `mysql://`   | MySQL / MariaDB |
| `sqlite://`  | SQLite (local file or `:memory:`) |

##### URI query parameters 

| URL query key       | Description                              | Default             |
|---------------------|------------------------------------------|---------------------|
| `sqlTableName`      | Manifest table name                      | `external_manifest` |
| `sqlMaxConnections` | Max pool connections                     | `5`                 |
| `sqlMinConnections` | Min idle pool connections                | `1`                 |
| `sqlAcquireTimeout` | Acquire timeout (seconds)                | `10`                |
| `sqlIdleTimeout`    | Idle connection timeout (seconds)        | `300`               |
| `sqlMaxLifetime`    | Max connection lifetime (seconds)        | `1800`              |

`sqlTableName` is validated as an ASCII identifier (`[A-Za-z0-9_]+`) to
prevent SQL injection through the dynamically-substituted table name.

##### `storage_options` / environment 

| Key (storage_options)  | Env var               | Description                                |
|------------------------|-----------------------|--------------------------------------------|
| `sql_conn_str`         | `LANCE_SQL_CONN_STR`  | Database connection URL (**required**)     |

Precedence: `storage_options` > environment variable.

#### 3. Expected Table Schema

Users must create the table with this exact column set and a uniqueness
constraint covering `(bucket_name, base_path, version)`:

| Column          | Type                    |
|-----------------|-------------------------|
| `bucket_name`   | `VARCHAR(255) NOT NULL` |
| `base_path`     | `VARCHAR(255) NOT NULL` |
| `version`       | `BIGINT NOT NULL`       |
| `manifest_path` | `VARCHAR(512) NOT NULL` |
| `file_size`     | `BIGINT NOT NULL`       |
| `e_tag`         | `VARCHAR(255) NOT NULL` |
| `create_time`   | `BIGINT NOT NULL`       |
| `update_time`   | `BIGINT NOT NULL`       |

Example row:

| Field           | Value                                |
|-----------------|--------------------------------------|
| `bucket_name`   | `hsq-test-1253960454`                |
| `base_path`     | `test/lance_sql_store`                        |
| `version`       | `33`                                 |
| `manifest_path` | `test/lance_sql_store/_versions/18446744073709551614.manifest`              |
| `file_size`     | `3487`                               |
| `e_tag`         | `"df2eb24e6b1142d2cd6ce7b6d41800ec"` |
| `create_time`   | `1782296148`                         |
| `update_time`   | `1782296148`                         |

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -516,8 +516,8 @@ dependencies = [
  "aws-sdk-ssooidc",
  "aws-sdk-sts",
  "aws-smithy-async",
- "aws-smithy-http",
- "aws-smithy-json",
+ "aws-smithy-http 0.63.5",
+ "aws-smithy-json 0.62.4",
  "aws-smithy-runtime",
  "aws-smithy-runtime-api",
  "aws-smithy-types",
@@ -578,7 +578,7 @@ dependencies = [
  "aws-sigv4",
  "aws-smithy-async",
  "aws-smithy-eventstream",
- "aws-smithy-http",
+ "aws-smithy-http 0.63.5",
  "aws-smithy-runtime",
  "aws-smithy-runtime-api",
  "aws-smithy-types",
@@ -605,8 +605,8 @@ dependencies = [
  "aws-credential-types",
  "aws-runtime",
  "aws-smithy-async",
- "aws-smithy-http",
- "aws-smithy-json",
+ "aws-smithy-http 0.63.5",
+ "aws-smithy-json 0.62.4",
  "aws-smithy-observability",
  "aws-smithy-runtime",
  "aws-smithy-runtime-api",
@@ -622,9 +622,9 @@ dependencies = [
 
 [[package]]
 name = "aws-sdk-s3"
-version = "1.124.0"
+version = "1.119.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "744c09d75dfec039a05cf8e117c995ded3b0baffa6eb83f3ed7075a01d8d8947"
+checksum = "1d65fddc3844f902dfe1864acb8494db5f9342015ee3ab7890270d36fbd2e01c"
 dependencies = [
  "aws-credential-types",
  "aws-runtime",
@@ -632,9 +632,8 @@ dependencies = [
  "aws-smithy-async",
  "aws-smithy-checksums",
  "aws-smithy-eventstream",
- "aws-smithy-http",
- "aws-smithy-json",
- "aws-smithy-observability",
+ "aws-smithy-http 0.62.6",
+ "aws-smithy-json 0.61.9",
  "aws-smithy-runtime",
  "aws-smithy-runtime-api",
  "aws-smithy-types",
@@ -646,6 +645,7 @@ dependencies = [
  "hmac 0.12.1",
  "http 0.2.12",
  "http 1.4.2",
+ "http-body 0.4.6",
  "http-body 1.0.1",
  "lru",
  "percent-encoding",
@@ -664,8 +664,8 @@ dependencies = [
  "aws-credential-types",
  "aws-runtime",
  "aws-smithy-async",
- "aws-smithy-http",
- "aws-smithy-json",
+ "aws-smithy-http 0.63.5",
+ "aws-smithy-json 0.62.4",
  "aws-smithy-observability",
  "aws-smithy-runtime",
  "aws-smithy-runtime-api",
@@ -688,8 +688,8 @@ dependencies = [
  "aws-credential-types",
  "aws-runtime",
  "aws-smithy-async",
- "aws-smithy-http",
- "aws-smithy-json",
+ "aws-smithy-http 0.63.5",
+ "aws-smithy-json 0.62.4",
  "aws-smithy-observability",
  "aws-smithy-runtime",
  "aws-smithy-runtime-api",
@@ -712,8 +712,8 @@ dependencies = [
  "aws-credential-types",
  "aws-runtime",
  "aws-smithy-async",
- "aws-smithy-http",
- "aws-smithy-json",
+ "aws-smithy-http 0.63.5",
+ "aws-smithy-json 0.62.4",
  "aws-smithy-observability",
  "aws-smithy-query",
  "aws-smithy-runtime",
@@ -736,7 +736,7 @@ checksum = "37411f8e0f4bea0c3ca0958ce7f18f6439db24d555dbd809787262cd00926aa9"
 dependencies = [
  "aws-credential-types",
  "aws-smithy-eventstream",
- "aws-smithy-http",
+ "aws-smithy-http 0.63.5",
  "aws-smithy-runtime-api",
  "aws-smithy-types",
  "bytes",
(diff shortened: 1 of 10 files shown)


hushengquan:
> ## PR Review: feat: support sql as external manifest store
> ### P0: SQL Injection via `table_name`
> The `table_name` parameter comes directly from user input (`sqlTableName` query parameter) and is interpolated into SQL strings via `format!()` without any sanitization:
> 
> ```rust
> // sql.rs — every query does this:
> let sql = format!(
>     "SELECT manifest_uuid FROM {} WHERE bucket_name = ? ...",
>     self.table_name
> );
> ```
> 
> A crafted URL like `s3+sql://bucket/path?sqlConnStr=...&sqlTableName=t;DROP TABLE t--` would allow arbitrary SQL execution. Table names cannot be parameterized in SQL, so you need to validate the identifier. At minimum, reject any `table_name` containing characters outside `[a-zA-Z0-9_]` in `parse_sql_url_params` (or in `new_external_store`).
> 
> ### P0: No tests
> The project policy from CLAUDE.md states _"We do not merge code without tests."_ This PR adds ~520 lines of new SQL store implementation with zero tests. At minimum, tests against SQLite (in-memory) should cover:
> 
> * `put_if_not_exists` succeeds on first write
> * `put_if_not_exists` fails on duplicate version (the core concurrency guarantee)
> * `put_if_exists` succeeds when row exists
> * `put_if_exists` fails when row doesn't exist
> * `get` / `get_latest_version` / `get_manifest_location` round-trips
> * `delete` clears all versions
> * Invalid `table_name` is rejected (after fixing the SQL injection above)
> 
> ### P1: Pool cache TOCTOU race
> `new_external_store` reads `POOL_CACHE`, finds no entry, then creates a pool and writes it. Two concurrent callers for the same `database_url` will both create separate pools:
> 
> ```rust
> let pool = {
>     let cache = POOL_CACHE.read().await;  // both read: None
>     cache.get(database_url).cloned()
> };
> let pool = match pool {
>     None => {
>         let p = AnyPoolOptions::new()...connect(database_url).await?;
>         POOL_CACHE.write().await.insert(...);  // both write
>         p
>     }
> };
> ```
> 
> Use a write lock for the check-and-insert, or use an entry API pattern under a single write lock.
> 
> ### P1: Database credentials stored as HashMap key
> The full `database_url` (containing passwords) is used as the key for the global `POOL_CACHE` `HashMap` and persists in memory for the lifetime of the process. Consider hashing or redacting credentials from the cache key, or restructuring so the URL isn't stored verbatim.

Modified as suggested

hushengquan:
@Xuanwo @westonpace Hi! If you have a moment, could you please take a look at this PR? I'm happy to make any necessary additions or modifications; alternatively, is there a better approach currently being discussed within the community? This PR is a prerequisite for a future update involving the removal of the strong dependency on "put-if-not-exist" for namespaces (specifically, ensuring transactional atomicity for writes to the `_manifest` table)

hushengquan:
@jackye1995 Hello, Professor. Could you please take another look at this PR when you have a moment? Internally at Tencent, quite a few teams are gradually beginning to integrate with Lance. Since they are utilizing COS for object storage, they are currently facing an issue where there is no standardized solution available for the Python API interface.

rust/lance-table/src/io/commit/sql.rs
```diff
@@ -0,0 +1,710 @@
...
+        let cache_key = format!("{}:{}", database_url, table_name);
+        // already checked this table before, skip
+        if SANITY_CHECK_CACHE.read().await.contains(&cache_key) {
+            return Ok(store);
+        }
+
+        // Verify the table exists and has the required columns by issuing
+        // a lightweight query. This works across all SQL backends without
+        // needing backend-specific information_schema queries.
+        let check_sql = format!(
```
Xuanwo, line 216: The table validation only checks that the expected columns can be selected, but it does not verify the unique key that provides CAS. A table with the same columns but no unique constraint can accept duplicate commits for the same version.

rust/lance-table/src/io/commit/sql.rs
```diff
@@ -0,0 +1,710 @@
...
+
+        SANITY_CHECK_CACHE.write().await.insert(cache_key);
+
+        Ok(store)
+    }
+
+    /// Build the full manifest path from base_path, version, and manifest_uuid.
+    ///
+    /// Path rule: `{base_path}/_versions/{u64::MAX - version}.manifest[-uuid]`
+    fn build_manifest_path(base_path: &str, version: u64, manifest_uuid: &str) -> String {
```
Xuanwo: The SQL store persists only the manifest UUID suffix and reconstructs every path as V2. V1 or detached manifests can be reopened at paths that were never written.

rust/lance-table/src/io/commit/sql.rs
```diff
@@ -0,0 +1,710 @@
...
+/// Extract an optional string column from a row.
+fn get_optional_string(row: &AnyRow, col: &str) -> Option<String> {
+    row.try_get::<String, _>(col).ok().filter(|s| !s.is_empty())
+}
+
+#[async_trait]
+impl ExternalManifestStore for SqlExternalManifestStore {
+    /// Get the manifest path for a given base_uri and version
+    async fn get(&self, base_uri: &str, version: u64) -> Result<String> {
+        let sql = format!(
```
Xuanwo, line 362: The SQL statements use `?` placeholders while PostgreSQL support is advertised. PostgreSQL-backed stores will fail once a bound query is executed.

rust/lance-table/src/io/commit.rs
```diff
@@ -817,6 +837,78 @@ fn get_dynamodb_endpoint(storage_options: &StorageOptions) -> Option<String> {
...
+///
+/// URL format: `s3+sql://{bucket}/{path}?sqlConnStr={db_url}[&sqlTableName=...][&sqlMaxConnections=...][&...]`
+///
+/// Returns `(conn_str, table_name, bucket_name, SqlPoolConfig)`.
+#[cfg(feature = "sql_store")]
+fn parse_sql_url_params(url: &Url) -> Result<(String, String, String, sql::SqlPoolConfig)> {
+    let params: std::collections::HashMap<String, String> =
+        url.query_pairs().into_owned().collect();
+
+    let conn_str = params
```
Xuanwo: `sqlConnStr` is part of the dataset URI. Database credentials can leak through logged, stored, displayed, or serialized dataset URIs.

Cargo.lock
```diff
@@ -695,19 +695,18 @@ dependencies = [
 
 [[package]]
 name = "aws-sdk-s3"
-version = "1.124.0"
+version = "1.119.0"
```
Xuanwo: The lockfile downgrades `aws-sdk-s3` without a dependency declaration change in this PR. Releases would pick up unrelated S3 SDK behavior changes.

hushengquan:
@Xuanwo Thanks for the great review! I've updated the code based on your suggestions. Could you please take another look when you have a chance?

PR #6092, opened 2026-03-04
perf: serialize inverted posting list batching and writes

BubbleCal:
Summary
- remove the async spawn path in `rust/lance-index/src/scalar/inverted/builder.rs` and keep posting list batch generation and writes in the same task
- reuse the schema/cloned data directly and avoid the extra channel and writer coordination logic
- simplify error handling by writing each batch immediately and accumulating write duration in place

Testing
- Not run (not requested)


Xuanwo:
Can we have a micro-bench to measure the difference?

BubbleCal:
I found a way to resolve the contention and avoid stuck, will move to a new PR

PR #6095, opened 2026-03-04
feat: support bounded compaction planner

### LINKED ISSUE #6039: Bounded Compaction Planner To Limit the amount of data processed during a single compaction
Follow up work for https://github.com/lance-format/lance/issues/5186

During the Compaction process, it is necessary to traverse and calculate all fragments, determine which ones are suitable for Compaction, and split them into Compaction Tasks. However, in the scenario of large datasets, the following bottlenecks exist:

1. The Compaction Plan itself takes a long time to generate. In a scenario we encountered with 300,000 fragments, it may takes 13min to generate plan or even led to an OOM error.
2. A large number of Compaction Tasks are generated in the planning. Even in the case of concurrent execution, it still takes a long time to complete. Moreover, if a Task fails due to certain reasons (such as spark driver loss), the entire Compaction needs to be retried completely, which is a huge waste of resources.

In response to this, we can extend the existing design of the Compaction Planner to create a Planner that limits the computational load of Compaction (such as the total data throughput 100TB total？, the total number of fragments involved in the calculation, etc.), named **BoundedCompactionPlanner**. Its purpose is to reasonably split a large Compaction job into multiple sub-jobs and complete them smoothly.

zhangyue19921010:
Closes: https://github.com/lance-format/lance/issues/6039

### DIFF
--- a/rust/lance/src/dataset/optimize.rs
+++ b/rust/lance/src/dataset/optimize.rs
@@ -110,11 +110,13 @@ use serde::{Deserialize, Serialize};
 use tracing::info;
 
 mod binary_copy;
+mod compaction_planner;
 pub mod remapping;
 
 use crate::index::frag_reuse::build_new_frag_reuse_index;
 use crate::io::deletion::read_dataset_deletion_file;
 use binary_copy::rewrite_files_binary_copy;
+use compaction_planner::{BoundedCompactionPlanner, BoundedCompactionPlannerOptions};
 pub use remapping::{IgnoreRemap, IndexRemapper, IndexRemapperOptions, RemappedIndex};
 
 /// Controls how data is rewritten during compaction.
@@ -200,6 +202,12 @@ pub struct CompactionOptions {
     /// Controls how much data is read at once when performing binary copy.
     /// Defaults to 16MB (16 * 1024 * 1024).
     pub binary_copy_read_batch_bytes: Option<usize>,
+    /// The type of compaction planner to use. Defaults to [`CompactionPlannerType::Default`].
+    pub compaction_planner_type: CompactionPlannerType,
+    /// The maximum number of bytes to compact in a single compaction operation.
+    pub max_compaction_bytes: Option<usize>,
+    /// The maximum number of rows to compact in a single compaction operation.
+    pub max_compaction_rows: Option<usize>,
 }
 
 #[allow(deprecated)]
@@ -219,6 +227,9 @@ impl Default for CompactionOptions {
             enable_binary_copy: false,
             enable_binary_copy_force: false,
             binary_copy_read_batch_bytes: Some(16 * 1024 * 1024),
+            compaction_planner_type: CompactionPlannerType::Default,
+            max_compaction_bytes: None,
+            max_compaction_rows: None,
         }
     }
 }
@@ -431,6 +442,33 @@ pub trait CompactionPlanner: Send + Sync {
     async fn plan(&self, dataset: &Dataset) -> Result<CompactionPlan>;
 }
 
+#[derive(Debug, PartialEq, Eq, Copy, Hash, Clone, serde::Deserialize, serde::Serialize)]
+pub enum CompactionPlannerType {
+    Default = 0,
+    Bounded = 1,
+}
+
+impl CompactionPlannerType {
+    pub const fn as_str(self) -> &'static str {
+        match self {
+            Self::Default => "default",
+            Self::Bounded => "bounded",
+        }
+    }
+
+    pub fn from_str_name(name: &str) -> Option<Self> {
+        match name.trim().to_ascii_lowercase().as_str() {
+            "default" => Some(Self::Default),
+            "bounded" => Some(Self::Bounded),
+            _ => None,
+        }
+    }
+
+    pub const fn requires_limits(self) -> bool {
+        matches!(self, Self::Bounded)
+    }
+}
+
 /// Formulate a plan to compact the files in a dataset
 ///
 /// The compaction plan will contain a list of tasks to execute. Each task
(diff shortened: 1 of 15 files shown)


rust/lance/src/dataset/optimize.rs
```diff
@@ -200,6 +202,12 @@ pub struct CompactionOptions {
     /// Controls how much data is read at once when performing binary copy.
     /// Defaults to 16MB (16 * 1024 * 1024).
     pub binary_copy_read_batch_bytes: Option<usize>,
+    /// The type of compaction planner to use. Defaults to [`CompactionPlannerType::Default`].
+    pub compaction_planner_type: CompactionPlannerType,
+    /// The maximum number of bytes to compact in a single compaction operation.
+    pub max_compaction_bytes: Option<usize>,
+    /// The maximum number of rows to compact in a single compaction operation.
+    pub max_compaction_rows: Option<usize>,
```
hamersaw, line 210: I'm having some difficulty with how these will be presented as options to the user. Regardless of which way I cut it, it feels like we're leaking implementation details that they should not be concerned about. For example (using spark SQL as an example):
````
OPTIMIZE <table> WITH (max_compaction_row=100000)
```

Will do nothing, because the user has to explicitly set the planner type:
```
OPTIMIZE <table> WITH (planner="bounded", max_compaction_row=100000)
```

Alternatively, if we remove the `planner` configuration option then this gets really messy as we add additional options that are disjoint. For example, if we had a "fooable" planner that can "foo" with "how_to_foo" configuration option:
```
OPTIMIZE <table> WITH (max_compaction_row=100000, how_to_foo=carefully)
```
means that we can't just create a different planner based on the passed options, because is ^^^ `FooablePlanner` or `BoundedPlanner`

To add maximum compaction bytes / rows configuration is there a reason we need to add a different planner implementation? Or can these be presented as options that are applied in the current planning scheme?

zhangyue19921010: Hi @hamersaw Thanks a lot for your attention! Looking forward to more discussions with you. After we reach a consensus on the design, I will update the implementation of the current PR.

I understand there are currently two main issues:
1. How to better expose the concepts of planners and their corresponding parameters to users.
2. Why a separate bounded planner needs to be added here.

> How to better expose the concepts of planners and their corresponding parameters to users.

I believe we can refer to the design of how various indexes are exposed to users. Here, we support different types of indexes, and each index has its own tuning parameters.
```
    def create_scalar_index(
        self,
        column: str,
        index_type: Union[
            Literal["BTREE"],
            Literal["BITMAP"],
            Literal["LABEL_LIST"],
            Literal["INVERTED"],
            Literal["FTS"],
            Literal["NGRAM"],
            Literal["ZONEMAP"],
            Literal["BLOOMFILTER"],
            Literal["RTREE"],
            IndexConfig,
        ],
        name: Optional[str] = None,
        *,
        replace: bool = True,
        train: bool = True,
        fragment_ids: Optional[List[int]] = None,
        index_uuid: Optional[str] = None,
        **kwargs,
    ):
```

On the Compaction side, add new parameters: planner and**kwargs
```
    def compact_files(
        self,
        *,
        target_rows_per_fragment: int = 1024 * 1024,
        max_rows_per_group: int = 1024,
        max_bytes_per_file: Optional[int] = None,
        materialize_deletions: bool = True,
        materialize_deletions_threshold: float = 0.1,
        num_threads: Optional[int] = None,
        batch_size: Optional[int] = None,
        compaction_mode: Optional[
            Literal["reencode", "try_binary_copy", "force_binary_copy"]
        ] = None,
        binary_copy_read_batch_bytes: Optional[int] = None,
        planner: Union[
            Literal["DEFAULT"],
            Literal["BOUNDED"],
        ] = "DEFAULT",
        **kwargs,
    ) -> CompactionMetrics:
```

During parsing, depending on the planner, attempt to capture parameters specific to the current planner from**kwargs (ignoring irrelevant parameters) and pass them to the Rust side.

For invocation, it is similar to:
```
    metrics = dataset.optimize.compact_files(
    planner = "bounded",
    max_compaction_rows = 1000000,
    )
```

> Why a separate bounded planner needs to be added here.

Generally speaking, a compaction plan is roughly divided into three steps: first, obtaining fragments, second, building compaction tasks, and third, constructing the compaction plan.

**The main positioning of the bounded planner here is an "incremental" compaction plan.**
1. Incrementally obtain fragments, obtain version diff fragments (TBD)
2. Bounded plan compaction, When building tasks, construct a partial compaction plan through restrictive conditions such as max row and max bytes, rather than a full one.
3. In the future, other "incremental" logics can be flexibly added.

In contrast, the Default planner is a full compaction plan.
1. Obtain all fragments of the current version
2. Build all compaction tasks at once

Of course, we can modify the Default planner to limit the number of tasks it outputs. However, there is a detailed issue here.
The pseudocode logic of the Default planner is as follows:
```
let metrics = collect all fragment metrics paralle at once

for metrics in metrics {
  build task
}
```

Under the condition of an incremental planner, we may prefer to collect on demand, especially in large tables with hundreds of thousands of fragments. Avoid the memory pressure and unnecessary computational overhead caused by full-scale fragment perception and calculation, that is:
```
for fragment in fragments {
  let metrics = collect current fragment metrics
  if condition(metrics) {
      break;
  }
}
```
It might be a good choice to distinguish between incremental planners and full planners.

Is that make sense?  I am open to all the above. Please let me know in time if you have anything want to discuss! Thanks in advance.

hamersaw: Great points! In my experience, the cost of planning a compaction (ex. retrieving fragment metadata and aggregating into compaction tasks) is quite small when compared to the actually compaction execution. So I do not think it should be something to build around.

I think bounding compaction executions can come in two flavors (1) limiting the number of input fragments and (2) limiting the number of compaction tasks. With an option, we can still do the former easily. The latter, will need to retrieve all fragments, compile the plan, and then just return / execute a small subset.

Either way, I think that building this using compaction options gives the same functionality and is easier for users to configure.

PR #6097, opened 2026-03-04
fix!: bump IVF_RQ version for compatibility check

BubbleCal:
Summary
- add IVF_RQ compatibility tests that run against the latest stable pylance release to ensure read/query compatibility both directions
- update IndexDetails to report the actual highest vector version rather than a single constant

### DIFF
--- a/rust/lance/src/index/append.rs
+++ b/rust/lance/src/index/append.rs
@@ -206,28 +206,21 @@ pub async fn merge_indices_with_unindexed_frags<'a>(
             .boxed()
             .await?;
 
-            let merged_indices = &old_indices[old_indices.len() - indices_merged..];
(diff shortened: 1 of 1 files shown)


Xuanwo:
Hi @BubbleCal, I think the review raised by claude is valid: 

> index_version derived from all old_indices, not just the merged subset

Do we need to take care about this?

BubbleCal:
> Hi @BubbleCal, I think the review raised by claude is valid:
> 
> > index_version derived from all old_indices, not just the merged subset
> 
> Do we need to take care about this?

it's common that we just append a delta index and `merged_num=0` in that case, so we need to derive `index_version` from all `old_indices`.

I think today all old_indices will always be with the same version, even on a new version of lance, we always respect the existing index's version when incremental indexing

PR #6099, opened 2026-03-04
feat: implement vector index details

### LINKED ISSUE #5963: Add lightweight vector index metadata to VectorIndexDetails
follow up #5945

Because `VectorIndexDetails` is empty, `describe_indices()` cannot determine the concrete vector index type (flat / HNSW / IVF / PQ / SQ). 

This follow‑up will store small, stable metadata in `VectorIndexDetails` when the index is created—e.g. `index_type` (IVF_PQ, IVF_HNSW_SQ, …), `metric_type`, `dimension`, `num_partitions`, and `num_sub_vectors`.

wjones127:
Cache vector index configuration within the index metadata, such as the distance type and build parameters.

Previously, to determine things like the distance type or index type of a vector index, the index file itself had to be opened. This PR stores that information in `VectorIndexDetails` within the manifest's `index_details` field, which is fetched and cached eagerly when loading the manifest.

Old indexes have this field left blank. When blank, the details are extracted from the index files and cached. This migration happens on the first write with a new library version.

## What's stored in VectorIndexDetails

**Core build parameters** (typed fields — required for any runtime to build the index):
- `metric_type`
- `target_partition_size` (IVF)
- `hnsw_index_config` — `max_connections`, `construction_ef`, `max_level` (HNSW)
- `compression` — PQ/SQ/RQ/flat, including `num_bits`, `num_sub_vectors`, `rotation_type`

**Runtime hints** (`map<string, string> runtime_hints`):
Optional build preferences that don't affect index structure. Stored so a background rebuild process can reproduce the original configuration. Runtimes that don't recognize a key must silently ignore it. Only non-default values are written.

Keys use reverse-DNS namespacing: `lance.*` for core Lance hints, other prefixes for runtime-specific hints (e.g., `lancedb.accelerator` for GPU acceleration in LanceDB Enterprise).

Current `lance.*` hints: `lance.ivf.max_iters`, `lance.ivf.sample_rate`, `lance.ivf.shuffle_partition_batches`, `lance.ivf.shuffle_partition_concurrency`, `lance.pq.max_iters`, `lance.pq.sample_rate`, `lance.pq.kmeans_redos`, `lance.sq.sample_rate`, `lance.hnsw.prefetch_distance`, `lance.skip_transpose`.

Also adds `apply_runtime_hints()` to read hints back into build params for future rebuild logic.

Closes #5963


protos/table.proto
```diff
@@ -460,8 +460,56 @@ message ExternalFile {
...
+// Hierarchical Navigable Small World (HNSW) index details, used as an optional configuration for IVF indexes.
+message HnswIndexDetails {
+  // The maximum number of outgoing edges per node in the HNSW graph. Higher values
+  // means more connections, better recall, but more memory and slower builds.
+  // Referred to as "M" in the HNSW literature.
+  uint32 max_connections = 1;
+  // "construction exploration factor": The size of the dynamic list used during
+  // index construction.
+  uint32 construction_ef = 2;
+}
```
wjones127: This representation works for the existing set of vector indices, but wondering if this is good for future plans. The current internal design has the concept of "stages", so you could have something like `ivf-ivf-pq` or `hnsw-pq` (no IVF). Some sequence of stages just doesn't make sense, like `pq-pq`. So I'm not sure the stages representation makes sense.

I was thinking it could be a tree-like system, where IVF and HNSW could have children, but PQ and SQ can't.

What do you think? @BubbleCal @eddyxu

protos/table.proto
```diff
@@ -460,8 +460,7 @@ message ExternalFile {
   uint64 size = 3;
 }
 
-// Empty details messages for older indexes that don't take advantage of the details field.
-message VectorIndexDetails {}
+// VectorIndexDetails and HnswIndexDetails moved to index.proto
```
wjones127: I *think* this is okay, but I don't feel 100% confident yet. I'll keep looking to see if this can cause any compatibility issues.

protos/table.proto
westonpace: I've always seen it as three layers.  The first is partitioning (IVF is really the only choice here).  The second is searching within partition (flat vs hnsw) and the third is quantization (pq, rq, sq, etc.)  So I don't think the concept of stages makes sense.  For example, I don't see ivf-ivf-pq.  I see ivf (layers=2) - pq.

protos/index.proto
```diff
@@ -184,6 +184,52 @@ message VectorIndex {
   VectorMetricType metric_type = 4;
 }
 
+// Details for vector indexes, stored in the manifest's index_details field.
+message VectorIndexDetails {
```
westonpace, line 188: Where are the IVF details?  was it hierarchical?  How many partitions in each stage?.

protos/index.proto
```diff
@@ -184,6 +184,52 @@ message VectorIndex {
...
+  // An unset compression oneof means flat / no quantization.
+  oneof compression {
+    ProductQuantization pq = 4;
+    ScalarQuantization sq = 5;
+    RabitQuantization rq = 6;
+  }
+}
+
+// Hierarchical Navigable Small World (HNSW) index details, used as an optional configuration for IVF indexes.
+message HnswIndexDetails {
```
westonpace: Minor nit: I think of "index details" as the top-level message describing a type of index.  This is a nested message that cannot stand on its own so maybe just `HnswParameters`?

python/python/tests/test_vector_index.py
```diff
@@ -1564,9 +1564,8 @@ def test_describe_vector_index(indexed_dataset: LanceDataset):
     info = indexed_dataset.describe_indices()[0]
 
     assert info.name == "vector_idx"
-    assert info.type_url == "/lance.table.VectorIndexDetails"
-    # This is currently Unknown because vector indices are not yet handled by plugins
-    assert info.index_type == "Unknown"
+    assert info.type_url == "/lance.index.pb.VectorIndexDetails"
```
westonpace, line 1681: Regrettably, I do not think we can change the type URL for backwards compatibility reasons.

rust/lance/src/index/append.rs
```diff
@@ -212,11 +212,21 @@ pub async fn merge_indices_with_unindexed_frags<'a>(
                     frag_bitmap.extend(idx.fragment_bitmap.as_ref().unwrap().iter());
                 });
 
+            // Carry forward existing index details, preferring the first segment
+            // that has populated (non-empty) details.
```
westonpace, line 317: I think this is a safe assumption but we are assuming all segments have the same details right?

rust/lance/src/index.rs
```diff
@@ -411,10 +411,10 @@ async fn open_index_proto(reader: &dyn Reader) -> Result<pb::Index> {
...
 }
 
-fn vector_index_details() -> prost_types::Any {
-    let details = lance_table::format::pb::VectorIndexDetails::default();
-    prost_types::Any::from_msg(&details).unwrap()
-}
+use vector::details::{
+    derive_vector_index_type, infer_missing_vector_details, vector_details_as_json,
+};
+pub(crate) use vector::details::{vector_index_details, vector_index_details_default};
```
westonpace: Should these be at the top of the file?

westonpace (commented):
I'm not 100% sure we can get away with changing the protobuf type URL.  Can you create a compatibility test to ensure that old versions can read new indexes created with these new details?

wjones127:
> I'm not 100% sure we can get away with changing the protobuf type URL. Can you create a compatibility test to ensure that old versions can read new indexes created with these new details?

We have one here that goes back to 0.29.1.beta2:

https://github.com/lance-format/lance/blob/db07e77438ba361abaab64078ace152a788fe5bd/python/python/tests/compat/test_vector_indices.py#L25-L32

I wonder if we need to check earlier than that to see incompatabilities.

protos/index.proto
wjones127: I'm using `VectorIndexDetails` here for the IVF stage. I suppose I could make it a substruct to make it clearer. The only parameter right now is `target_partition_size`.

> How many partitions in each stage?

What do you mean by this?

protos/index.proto
wjones127: Yeah that's fair.

rust/lance/src/index.rs
wjones127: Yeah I should move those.

python/python/tests/test_vector_index.py
wjones127: I can move that back if it's necessary.

rust/lance/src/index/append.rs
wjones127: Should be. Ideally, we would create some top-level index configuration that is deduplicated across segments, but that's a complex format change for another day.

wjones127:
Should be able to implement this TODO: 

https://github.com/lance-format/lance/blob/de393a26a068dd297929ca7d798e43dc31c57337/rust/lance/src/dataset/scanner.rs#L3422-L3433

wjones127:
@westonpace I had claude go through and see if this breaks backwards compat as is. Other than the type_url, it claims that the indexes don't have different compatibility from what's on `main`:

```
  Results                                                                                                                                                                                                                                                                                                               
                                                                                                                                                                                                                                                                                                                      
  The protobuf move (VectorIndexDetails from table.proto to index.proto) does not break backward compatibility. All failures seen are pre-existing and identical when reading indices written by main vs this branch.                                                                                                   
                                                                                                                                                                                                                                                                                                                        
  ┌───────────────┬────────────────────────────────────┬─────────────────────────────────────────────────────────────────────────────────────────────────────────────┐                                                                                                                                                  
  │ Version range │              Behavior              │                                                    Cause                                                    │                                                                                                                                                  
  ├───────────────┼────────────────────────────────────┼─────────────────────────────────────────────────────────────────────────────────────────────────────────────┤                                                                                                                                                  
  │ 0.8.0         │ Can't open dataset                 │ Old manifest format                                                                                         │                                                                                                                                                  
  ├───────────────┼────────────────────────────────────┼─────────────────────────────────────────────────────────────────────────────────────────────────────────────┤                                                                                                                                                  
  │ 0.9.0-0.10.0  │ Opens, panics on search            │ Very old index format                                                                                       │
  ├───────────────┼────────────────────────────────────┼─────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ 0.11.0-0.14.0 │ Import error                       │ NumPy 1.x/2.x incompatibility                                                                               │
  ├───────────────┼────────────────────────────────────┼─────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ 0.15.0-0.16.0 │ Works (brute force)                │ Doesn't recognize index, falls back to scan                                                                 │
  ├───────────────┼────────────────────────────────────┼─────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ 0.17.0-0.29.0 │ Opens, sees index, fails to search │ Pre-existing index format incompatibility (missing field num_bits, 2-D tensor shape) — same failure on main │
  ├───────────────┼────────────────────────────────────┼─────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ 0.30.0+       │ Works                              │ Full compatibility                                                                                          │
  └───────────────┴────────────────────────────────────┴─────────────────────────────────────────────────────────────────────────────────────────────────────────────┘

  The earliest version that can successfully use an IVF_PQ index written by this branch is 0.30.0 — but this is the same boundary as main. The registry lookup at rust/lance-index/src/registry.rs:98 uses details.type_url.split('.').next_back() which extracts just "VectorIndexDetails" regardless of package, so
  the package name change is transparent.
```

westonpace (approved):
Sweet.  Lets go for it.

protos/index.proto
```diff
@@ -184,6 +184,52 @@ message VectorIndex {
...
+    enum RotationType {
+      FAST = 0;
+      MATRIX = 1;
+    }
+    uint32 num_bits = 1;
+    RotationType rotation_type = 2;
+  }
+
+  // An unset compression oneof means flat / no quantization.
+  oneof compression {
```
BubbleCal, line 217: maybe worth adding `index_version` for each one.
Found it's complicated for vector index to maintain compatibility when introducing breaking changes because it's hard to get the index version.

The other scalar index has index_version already today.

protos/index.proto
```diff
@@ -184,6 +184,52 @@ message VectorIndex {
...
+    ProductQuantization pq = 4;
```
BubbleCal, line 218: need `Flat` for index without any quantization

protos/index.proto
```diff
@@ -184,6 +184,52 @@ message VectorIndex {
...
+
+// Hierarchical Navigable Small World (HNSW) index details, used as an optional configuration for IVF indexes.
+message HnswIndexDetails {
+  // The maximum number of outgoing edges per node in the HNSW graph. Higher values
+  // means more connections, better recall, but more memory and slower builds.
+  // Referred to as "M" in the HNSW literature.
+  uint32 max_connections = 1;
+  // "construction exploration factor": The size of the dynamic list used during
+  // index construction.
+  uint32 construction_ef = 2;
```
BubbleCal, line 238: need to add `max_level` for HNSW

wjones127:
@westonpace @BubbleCal do you want to take another look? I changed up the format a little bit to handle other index parameters.

rust/lance/src/index/vector/details.rs
```diff
@@ -0,0 +1,1154 @@
...
+    max_level: u32,
+}
+
+fn is_zero(v: &u32) -> bool {
+    *v == 0
+}
+
+#[derive(Serialize)]
+#[serde(tag = "type", rename_all = "lowercase")]
+enum CompressionDetailsJson {
```
BubbleCal, line 71: missed `Flat`

rust/lance/src/index/vector/details.rs
```diff
@@ -0,0 +1,1154 @@
...
+                    ivf.shuffle_partition_batches = v;
+                }
+                if let Some(v) = parse(hints, "lance.ivf.shuffle_partition_concurrency") {
+                    ivf.shuffle_partition_concurrency = v;
+                }
+            }
+            StageParams::Hnsw(hnsw) => {
+                if let Some(raw) = hints.get("lance.hnsw.prefetch_distance") {
+                    hnsw.prefetch_distance = if raw == "none" {
+                        None
```
BubbleCal, line 227: IIRC the default is `Some(2)`

BubbleCal: nvm, just noticed this is setting it to `None`

rust/lance/src/index/vector/details.rs
wjones127: I think that's covered by `None` in `compression: Option<CompressionDetailsJson>,`

rust/lance/src/index/vector/details.rs
```diff
@@ -0,0 +1,1395 @@
...
+        lance_linalg::distance::DistanceType::Hamming => VectorMetricType::Hamming,
+    };
+
+    let mut target_partition_size = 0u64;
+    let mut hnsw_index_config = None;
+    let mut compression = None;
+    let mut runtime_hints: HashMap<String, String> = params.runtime_hints.clone();
+
+    // Only write hints that differ from their defaults, keeping the map minimal.
+    // Absence of a key means "use your default".
```
wjones127: issue(blocking): the default checks here are fragile. TBH, I think there's not much benefit to omitting if default. It complicates the logic, and it's probably nice if we record even the default of the library that built it. Let's just remove these checks.

wjones127: Done — removed the default checks and now write all hints unconditionally.

PR #6102, opened 2026-03-05
feat(java): add non-blocking AsyncScanner with CompletableFuture API

beinan:
## Summary

Implements a parallel async scanner alongside the existing blocking `LanceScanner` to prevent thread starvation in Java query engines like Presto and Trino.

This PR adds **AsyncScanner** - a non-blocking alternative to `LanceScanner` that uses `CompletableFuture` for true async I/O operations.

## Motivation

Query engines like Presto and Trino rely on non-blocking I/O to efficiently multiplex thousands of concurrent queries on a limited thread pool. The current `LanceScanner` blocks Java threads during Rust I/O operations, causing thread starvation and poor performance in these environments.

## Key Features

✅ **Non-blocking I/O**: Spawns Tokio tasks instead of blocking Java threads  
✅ **CompletableFuture API**: Native Java async patterns for seamless integration  
✅ **Persistent JNI dispatcher**: Single thread attached to JVM for zero-overhead callbacks  
✅ **Task-based architecture**: Uses task IDs instead of JNI refs to prevent memory leaks  
✅ **Full feature parity**: Supports filters, projections, vector search, FTS, aggregates  
✅ **Clean cancellation**: Proper task cleanup without resource leaks  
✅ **Parallel to existing API**: No breaking changes, `LanceScanner` unchanged  

## Architecture

The implementation uses the **"Task ID + Dispatcher" pattern**:

1. **Java manages futures**: `ConcurrentHashMap<taskId, CompletableFuture<Long>>` maps task IDs to pending requests
2. **Rust spawns async tasks**: Returns immediately while Tokio handles I/O in background
3. **Lock-free completion channel**: Carries `(taskId, resultPointer)` from Tokio to dispatcher
4. **Persistent dispatcher thread**: Attaches to JVM once, completes Java futures via cached JNI method IDs

### Components

**Rust (`java/lance-jni/src/`):**
- `dispatcher.rs` - Persistent JNI thread with cached method IDs for callbacks
- `task_tracker.rs` - Thread-safe task registry using `RwLock<HashMap>`
- `async_scanner.rs` - AsyncScanner with Tokio task spawning and JNI exports
- `lib.rs` - Modified to add `JNI_OnLoad` hook for dispatcher initialization

**Java (`java/src/main/java/org/lance/ipc/`):**
- `AsyncScanner.java` - CompletableFuture-based async API with task management

**Tests (`java/src/test/java/org/lance/`):**
- `AsyncScannerTest.java` - 6 comprehensive examples demonstrating usage patterns

### Basic async scan
```java
ScanOptions options = new ScanOptions.Builder().batchSize(20L).build();

try (AsyncScanner scanner = AsyncScanner.create(dataset, options, allocator)) {
    CompletableFuture<ArrowReader> future = scanner.scanBatchesAsync();
    ArrowReader reader = future.get(10, TimeUnit.SECONDS);
    
    while (reader.loadNextBatch()) {
        // Process batches without blocking
    }
}
```

### Multiple concurrent scans (key benefit for Presto/Trino)
```java
List<CompletableFuture<Integer>> futures = new ArrayList<>();

for (int i = 0; i < 100; i++) {
    AsyncScanner scanner = AsyncScanner.create(dataset, options, allocator);
    futures.add(scanner.scanBatchesAsync()
        .thenApply(reader -> processInBackground(reader)));
}

// All scans run in parallel without blocking threads!
CompletableFuture.allOf(futures.toArray(new CompletableFuture[0]))
    .get(30, TimeUnit.SECONDS);
```

## Testing

```bash
cd java
./mvnw test -Dtest=AsyncScannerTest
./mvnw compile  # Full build verification
```

All tests pass ✅

## Compatibility

- ✅ No breaking changes to existing APIs
- ✅ `LanceScanner` remains unchanged
- ✅ Uses same `ScanOptions` for consistency
- ✅ Opt-in: users choose blocking or async based on their needs

## Performance Benefits

For query engines running hundreds of concurrent queries:
- **Before**: Thread pool exhaustion as threads block on I/O
- **After**: Threads immediately return, I/O happens in background

## Checklist

- [x] Rust code compiles (`cargo check`)
- [x] Java code compiles (`./mvnw compile`)
- [x] Code formatted (`./mvnw spotless:apply && cargo fmt`)
- [x] Comprehensive tests added (`AsyncScannerTest.java`)
- [x] Documentation in code examples
- [x] No breaking changes

## Related Issues

Addresses the need for non-blocking I/O in Java query engines that integrate with LanceDB.

### DIFF
--- a/java/lance-jni/src/async_scanner.rs
+++ b/java/lance-jni/src/async_scanner.rs
@@ -122,9 +122,21 @@ impl AsyncScanner {
                          This indicates a critical initialization failure.",
                         task_id
                     );
-                    return; // Task will never complete, but won't crash JVM
+                    // Clean up the FFI stream pointer to prevent memory leak
+                    if let Ok(ptr) = result {
+                        unsafe {
+                            drop(Box::from_raw(
+                                ptr as *mut arrow::ffi_stream::FFI_ArrowArrayStream,
+                            ));
+                        }
+                        log::debug!("Cleaned up FFI stream pointer for task {}", task_id);
+                    }
+                    return;
                 }
             };
+
+            // Save the pointer before sending so we can clean up on failure
+            let result_ptr = result.as_ref().ok().copied();
 
             if let Err(e) = dispatcher.send(DispatcherMessage {
                 scanner_global_ref: global_ref_for_task,
@@ -136,6 +148,15 @@ impl AsyncScanner {
                     task_id,
                     e
                 );
+                // Clean up the FFI stream pointer to prevent memory leak
+                if let Some(ptr) = result_ptr {
+                    unsafe {
+                        drop(Box::from_raw(
+                            ptr as *mut arrow::ffi_stream::FFI_ArrowArrayStream,
+                        ));
+                    }
+                    log::debug!("Cleaned up FFI stream pointer for task {}", task_id);
+                }
             }
 
             // _cleanup_guard.drop() called here automatically, removing task from tracker
--- a/java/lance-jni/src/dispatcher.rs
+++ b/java/lance-jni/src/dispatcher.rs
@@ -94,6 +94,7 @@ fn handle_error(
         Ok(s) => s,
         Err(e) => {
             log::error!("Failed to create JString for error: {:?}", e);
+            let _ = env.exception_clear();
             return;
         }
     };
@@ -114,6 +115,8 @@ fn handle_error(
 
     if let Err(e) = result {
         log::error!("Failed to call failTask: {:?}", e);
+        // Clear any pending JNI exception to protect the dispatcher loop
+        let _ = env.exception_clear();
     }
 }
 
@@ -139,5 +142,17 @@ fn handle_success(
 
     if let Err(e) = result {
         log::error!("Failed to call completeTask: {:?}", e);
+        // Clear any pending JNI exception to protect the dispatcher loop
+        let _ = env.exception_clear();
+        // Clean up the FFI stream since Java won't receive it
+        unsafe {
+            drop(Box::from_raw(
+                result_ptr as *mut arrow::ffi_stream::FFI_ArrowArrayStream,
+            ));
+        }
+        log::debug!(
+            "Cleaned up FFI stream pointer for task {} after completeTask failure",
+            task_id
+        );
     }
 }


beinan:
## Review Comments Addressed

Thank you for the detailed review! I've addressed all three issues:

### ✅ P0: Fixed race condition in `start_scan`

**Problem**: Task could complete before registration, causing Java futures to hang forever.

**Solution**: 
- Clone the `GlobalRef` so the spawned task has its own copy
- Always send completion message to dispatcher (not conditional on task tracker)
- Task tracker is now only used for cancellation support

**Code change**:
```rust
// Clone ref for spawned task to prevent race condition
let global_ref_for_task = scanner_global_ref.clone();

let handle = RT.spawn(async move {
    // ... scan work ...
    
    // Always send message even if not tracked
    TASK_TRACKER.complete(task_id).await;
    dispatcher.send(DispatcherMessage {
        scanner_global_ref: global_ref_for_task,
        ...
    });
});

// Registration now only for cancellation
TASK_TRACKER.register(task_id, TaskInfo { ... }).await;
```

### ✅ P1: Removed code duplication with `blocking_scanner.rs`

**Problem**: `build_full_text_search_query` (~130 lines) was duplicated.

**Solution**:
- Made function `pub(crate)` in `blocking_scanner.rs`
- Import and reuse from `async_scanner.rs`
- Removed 130+ lines of duplicated code
- Both scanners now share same FTS query building logic

### ✅ P1: Fixed JNI signature mismatch  

**Problem**: `releaseNativeScanner` had unused `handle` parameter in Java.

**Solution**:
- Removed unused `handle` parameter from Java method signature
- Now matches Rust implementation that only uses `j_scanner`
- API is clearer and less fragile

---

All changes are in commit `951950fe`. CI should pass now! 🚀

LuQQiu:
@hamersaw seems to be a performance improvement for presto/trino

java/lance-jni/src/dispatcher.rs
```diff
@@ -0,0 +1,143 @@
...
+        let (tx, mut rx) = mpsc::unbounded_channel::<DispatcherMessage>();
+
+        // Spawn persistent dispatcher thread
+        std::thread::Builder::new()
+            .name("lance-jni-dispatcher".to_string())
+            .spawn(move || {
+                // Attach ONCE and never detach - this is the key optimization
+                let mut env = jvm
+                    .attach_current_thread_permanently()
+                    .expect("Failed to attach dispatcher to JVM");
```
hamersaw, line 37: We probably do not want to use `expect` anywhere in here because it will panic and crash the JVM. I think there are other places in the PR where we probably don't want to use `expect` as well.

java/lance-jni/src/async_scanner.rs
```diff
@@ -0,0 +1,291 @@
...
+                        }
+                        Err(e) => Err(e.to_string()),
+                    }
+                }
+                Err(e) => Err(e.to_string()),
+            };
+
+            // Send result to dispatcher for Java completion
+            let dispatcher = DISPATCHER.get().expect("Dispatcher not initialized");
+            let _ = dispatcher.send(DispatcherMessage {
```
hamersaw: Lets capture and at least log the error here.

java/lance-jni/src/async_scanner.rs
```diff
@@ -0,0 +1,291 @@
...
+}
+
+fn inner_start_scan(env: &mut JNIEnv, j_scanner: JObject, task_id: u64) -> Result<()> {
+    // Create global reference first, before borrowing scanner
+    let scanner_global_ref = env.new_global_ref(&j_scanner)?;
+
+    let scanner_guard =
+        unsafe { env.get_rust_field::<_, _, AsyncScanner>(&j_scanner, NATIVE_ASYNC_SCANNER)? };
+
+    scanner_guard.start_scan(task_id, scanner_global_ref);
```
hamersaw:
```suggestion
  // Clone the Arc<Scanner> and drop the MutexGuard before calling start_scan,
  // which does block_on internally. Holding the guard across block_on risks deadlock.
 let scanner = {
   let guard =
     unsafe { env.get_rust_field::<_, _, AsyncScanner>(&j_scanner, NATIVE_ASYNC_SCANNER)? };
   guard.inner.clone()
 };
 
 AsyncScanner::start_scan(scanner, task_id, scanner_global_ref);
```

This way we ensure correct releasing of resources.

java/lance-jni/src/async_scanner.rs
```diff
@@ -0,0 +1,291 @@
...
+                scanner_global_ref: global_ref_for_task,
+                task_id,
+                result,
+            });
+
+            // _cleanup_guard.drop() called here automatically, removing task from tracker
+        });
+
+        // Register task for cancellation support
+        RT.block_on(async {
```
hamersaw, line 166: Do we need some kind of two-phase tracking here? For example, before we spawn the task ^^ we register it, and then here we commit it. In the current implementation, however unlikely, if the I/O task completes before we reach this statement it can drop the `_cleanup_guard` and things will never be cleaned up from the `TASK_TRACKER`. A  two-phase commit would ensure this is impossible.

java/lance-jni/src/dispatcher.rs
beinan: Good call, I will double check.  But for the "expect" here, I just want the JVM crash to make it fail fast.  Otherwise the issue will be very difficult to catch or debug later.

beinan: Fixed in a06c692d. Kept `expect()` here for fail-fast initialization (JVM should crash if dispatcher can't initialize), but replaced the runtime `expect()` at async_scanner.rs:91 with error logging to prevent crashes during normal operation.

java/lance-jni/src/async_scanner.rs
beinan: Fixed in a06c692d. Added error logging for both dispatcher initialization check and send failures.

java/lance-jni/src/async_scanner.rs
beinan: Fixed in a06c692d. Implemented exactly as suggested - clone the `Arc<Scanner>` and drop the guard before calling `start_scan_with_scanner` to prevent deadlock.

java/lance-jni/src/async_scanner.rs
beinan: Excellent catch! Fixed in a06c692d with a three-step two-phase registration pattern:
1. Pre-register with placeholder handle BEFORE spawning
2. Spawn the actual task
3. Update registration with real handle

This guarantees the task is tracked before the cleanup guard can run. Added `TaskTracker::update_handle()` method to atomically replace the placeholder with the real handle.

beinan:
Hi @hamersaw, could you help take a look again? Thanks!

beinan:
@hamersaw Circling back on this—since tomorrow is my last day at Uber, I’m aiming to wrap up all outstanding PRs today. Do you have time for a final look so we can get this merged?

java/lance-jni/src/async_scanner.rs
```diff
@@ -0,0 +1,331 @@
...
+                }
+            };
+
+            if let Err(e) = dispatcher.send(DispatcherMessage {
+                scanner_global_ref: global_ref_for_task,
+                task_id,
+                result,
+            }) {
+                log::error!(
+                    "Failed to send completion message for task {}: {}",
```
jja725, line 147: should we clean up the result ptr if any error happens during send?

java/lance-jni/src/dispatcher.rs
```diff
@@ -0,0 +1,143 @@
...
+                    .get_method_id(&async_scanner_class, "failTask", "(JLjava/lang/String;)V")
+                    .expect("failTask method not found");
+
+                // Event loop: block waiting for completions
+                while let Some(msg) = rx.blocking_recv() {
+                    let scanner_obj = msg.scanner_global_ref.as_obj();
+
+                    match msg.result {
+                        Err(error) => {
+                            handle_error(&mut env, scanner_obj, fail_method, msg.task_id, &error)
```
jja725, line 58: what would happen if we have error when handling error? is there a way like finally to cleanup the things

java/lance-jni/src/async_scanner.rs
beinan: Good catch! Added cleanup for the FFI stream pointer in both error paths — when the dispatcher is not initialized and when `send()` fails. Also save the pointer before sending so we can reclaim it without cloning the result. (474a70c4)

java/lance-jni/src/dispatcher.rs
beinan: Great question. Added `env.exception_clear()` after every failed JNI call (`new_string`, `failTask`, `completeTask`) to clear any pending exception and protect the dispatcher loop from corruption. For `handle_success`, we also free the FFI stream pointer since Java will never receive it. This effectively acts as a `finally` block for each message dispatch. (474a70c4)

hamersaw:
### Code review

Found 2 issues:

1. **FFI stream pointer leaked when `close()` races with an in-flight task completing.** When `close()` calls `pendingTasks.clear()`, if a Rust task has already completed and sent a `DispatcherMessage` with a `resultPtr`, the dispatcher thread calls `completeTask(taskId, ptr)` — but `pendingTasks.get(taskId)` returns `null` (map was cleared), so the FFI stream pointer is silently dropped without being freed. In `handle_success` (dispatcher.rs), the pointer is only freed when `call_method_unchecked` returns a JNI error, not when the Java method succeeds but does nothing with the pointer. Additionally, `close()` never calls `future.completeExceptionally()` on pending futures before clearing the map, so any caller holding a reference to those futures will hang indefinitely.

https://github.com/lance-format/lance/blob/474a70c44aa69337b46cebb6cf19d3479704afcc/java/src/main/java/org/lance/ipc/AsyncScanner.java#L149-L163

https://github.com/lance-format/lance/blob/474a70c44aa69337b46cebb6cf19d3479704afcc/java/src/main/java/org/lance/ipc/AsyncScanner.java#L191-L200

https://github.com/lance-format/lance/blob/474a70c44aa69337b46cebb6cf19d3479704afcc/java/lance-jni/src/dispatcher.rs#L142-L157

2. **`testConcurrentAsyncScans` creates `AsyncScanner` outside try-with-resources, risking native resource leak.** The scanner is created at line 158 without try-with-resources. `scanner.close()` is only called inside a `thenApply` lambda (line 172) — if the lambda throws before reaching that line, the native scanner is never released. All other tests in the file correctly use try-with-resources. The blocking scanner's Rust code (blocking_scanner.rs lines 487-496) explicitly documents that try-with-resources is required to prevent native memory leaks.

https://github.com/lance-format/lance/blob/474a70c44aa69337b46cebb6cf19d3479704afcc/java/src/test/java/org/lance/AsyncScannerTest.java#L157-L178

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #6105, opened 2026-03-05
fix: disallow brute force on index unless explicitly requested

westonpace:
Currently an attempt to search an index with a distance type that does not match the trained distance type will result in a brute force search (with a warning logged).  However, this should be an error.  An implicit brute force search can be a very costly thing.  We already have a way to request a brute force search (use_index=False) and that should be used if this is truly the behavior that is desired.

### DIFF
--- a/rust/lance/src/dataset/scanner.rs
+++ b/rust/lance/src/dataset/scanner.rs
@@ -3438,12 +3438,10 @@ impl Scanner {
                     if user_metric == index_metric {
                         true
                     } else {
-                        log::warn!(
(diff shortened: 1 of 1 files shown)


wkalt:
this seems like a significant behavior change, since any clients previously doing the "wrong" thing but still getting results will now get errors

LuQQiu (approved):

wjones127 (commented):
I think this change is reasonable, but it should definitely be labelled as a breaking change.

PR #6115, opened 2026-03-06
feat: change default file format version to 2.1

westonpace:


python/python/tests/test_file.py
```diff
@@ -243,9 +243,10 @@ def test_metadata(tmp_path):
     assert len(column.pages) == 1
 
     page = column.pages[0]
-    assert len(page.buffers) == 1
-    assert page.buffers[0].position == 0
-    assert page.buffers[0].size == 24
+    assert len(page.buffers) > 0
+    for buffer in page.buffers:
+        assert buffer.position % 64 == 0
+        assert buffer.size > 0
```
westonpace, line 249: Miniblock means we now have two buffers, even for primitive data.

python/python/tests/test_optimize.py
```diff
@@ -74,28 +74,30 @@ def test_optimize_max_bytes(tmp_path: Path):
...
         materialize_deletions=False,
         max_bytes_per_file=1000,
         batch_size=128,
     )
-
-    # We get 4 fragments here because we don't actually write any data to the file
-    # until we've accumulated 8MiB for a page.
     assert metrics.fragments_removed == 2
-    assert metrics.fragments_added == 4
+    assert metrics.fragments_added > 2
```
westonpace, line 95: Spent some time tracking this down.  It is 8 now instead of 4.  It seems when we read in 2.1 the arrays get allocated a buffer with 2MB of capacity even though there is only 1MB of data.  Might be worth some further investigation at some point but not critical.

rust/lance-core/src/utils/futures.rs
```diff
@@ -264,13 +264,64 @@ impl<S: Stream> FinallyStreamExt<S> for S {
     }
 }
 
+/// A stream wrapper that calls a function when dropped.
+///
+/// Unlike [`FinallyStream`], which fires when the inner stream yields `None`,
+/// this fires when the wrapper is dropped — even if the stream was not fully
+/// consumed.
+#[pin_project(PinnedDrop)]
+pub struct OnDropStream<S: Stream, F: FnOnce()> {
```
westonpace, line 273: Extra utility to add on_drop method

rust/lance-encoding/src/decoder.rs
```diff
@@ -1910,7 +1911,19 @@ fn check_scheduler_on_drop(
             None
         }
     });
-    stream.chain(check_scheduler).boxed()
+    stream
+        .chain(check_scheduler)
+        .on_drop(move || {
```
westonpace, line 1916: This was a bit tricky.  There was no initialization needed in 2.0 and so we never noticed this.  Bascially...the I/O queue fills up and blocks, the decode thread can't get around to decoding and draining the backpressure queue because the stream has been dropped.  So the scheduler hangs.  This is expected.  When we drop everything we should cancel all the I/O and cleanup.

However, we don't drop everything because we have a reference to the I/O scheduler here.  So the fix here aborts the scheduler thread if the stream is dropped.

java/src/test/java/org/lance/operation/MergeTest.java
```diff
@@ -246,7 +246,7 @@ void testMergeNewColumnWithNonContiguousFieldId(@TempDir Path tempDir) throws Ex
                   datasetPath,
                   addressRoot,
                   new int[] {addressFieldId, cityFieldId, countryFieldId},
-                  new int[] {0, 1, 2});
+                  new int[] {-1, 0, 1});
```
wjones127, line 249: question(blocking): do you know why this changed? Seems suspicious, so want to verify it's not an issue.

python/python/tests/compat/test_scalar_indices.py
```diff
@@ -40,7 +41,16 @@ def create(self):
                 "btree": pa.array(range(1000)),
             }
         )
-        dataset = lance.write_dataset(data, self.path, max_rows_per_file=100)
+        print(
+            "Writing dataset with data_storage_version: "
+            f"{safe_data_storage_version(self.compat_version)}"
+        )
```
wjones127: suggestion: remove prints


```suggestion
```

rust/lance-core/src/utils/futures.rs
wjones127: praise: I've been meaning to add something just like this!

rust/lance/src/index.rs
```diff
@@ -6213,7 +6154,7 @@ mod tests {
         assert_io_lt!(
             stats,
             read_iops,
-            15,
+            17,
```
wjones127, line 6157: question(non-blocking): do you know which IOPs resulted in this change?

wjones127 (approved):
Thanks for looking into all of these. The one remaining one that gives me pause is the change in the Java test. Maybe I just don't understand the test.

java/src/test/java/org/lance/operation/MergeTest.java
westonpace: In 2.0 we created a column for struct and list fields (this is where we store validity).  In 2.1 we only create columns for the leaf fields.  So the rules for field->column mapping have changed.  We don't have field ids for every column.

Most of the code didn't need to change because it uses `ReaderProjection::from_column_names` and `ReaderProjection::from_whole_schema`.  The only place we really use `ReaderProjection::from_field_ids` is when we are doing actual operations on a dataset in which case we are populating this with logic from the `DataFile` (and that logic knows about this change).

So I _think_ this is a test-only change because Java should generally be accessing things at the dataset layer and not creating a file reader directly (and if it is, it should be using one of the above methods for creating `ReaderProjection`)

rust/lance/src/index.rs
westonpace: Starting with 2.1 file readers will have an initialize step between loading the metadata and loading the data.  This caches per-column metadata (we can't do it at the beginning when we load the file metadata because we don't know which columns are being read).  So now reading column X from a file means we have to:

* Read file metadata (1-2 IOPS, cached in file metadata cache)
* Read column metadata (1 IOP per page, cached in file metadata cache, **new**)
* Read column data (1 IOP per page, not cached)

I'd like to optimize this away at some point (e.g. have an open_and_prepare_to_read_xyz) but I think we will need to make some format changes (https://github.com/lance-format/lance/issues/4888) to move the miniblock metadata into the global buffer space.

PR #6117, opened 2026-03-06
fix(index): preserve stable row-id entries during scalar index optimize

### LINKED ISSUE #6116: [bug] `optimize_indices` can drop valid BTree entries when stable row IDs are enabled
## Summary

I hit this while building the music module of [StaticFlow](https://ackingliu.top/), my personal project built on top of Lance/LanceDB.

The `songs` dataset uses:

- `enable_stable_row_ids = true`
- a BTree scalar index on `id`

After running:

1. `compact_files()`
2. `optimize_indices()`

full scans still returned the expected rows, but indexed lookups by `id` returned no rows.

In the application this became:

```text
GET /api/music                -> 200 OK
GET /api/music/song-42        -> 404 Not Found
GET /api/music/song-42/play   -> 404 Not Found
```

## Why this looks like a stable-row-ID bug

The issue appears to be in scalar-index update during `optimize_indices()`, where old index rows are filtered before being merged with new data.

Relevant code paths in the current `main` branch:

- During scalar index merge, `merge_indices_with_unindexed_frags` computes `effective_old_frags` and passes that fragment bitmap into scalar index update logic:
  [`rust/lance/src/index/append.rs#L125-L183`](https://github.com/acking-you/lance/blob/main/rust/lance/src/index/append.rs#L125-L183)
- In the BTree update path, old rows are then filtered by inferring fragment ownership from `row_id >> 32`:
  [`rust/lance-index/src/scalar/btree.rs#L1341-L1358`](https://github.com/acking-you/lance/blob/main/rust/lance-index/src/scalar/btree.rs#L1341-L1358)

Simplified current behavior:

```rust
// append.rs (simplified)
let effective_old_frags = ...;

let created_index = if effective_old_frags.is_empty() {
    rebuild_from_scratch(...)
} else {
    index
        .update(new_data_stream, &new_store, Some(&effective_old_frags))
        .await?
};
```

```rust
// btree.rs (simplified)
fn filter_row_ids_by_fragments(row_ids: &[u64], valid_fragments: &RoaringBitmap) {
    let keep = row_ids
        .iter()
        .map(|row_id| valid_fragments.contains((row_id >> 32) as u32));
}
```

That is valid for address-style row IDs:

```text
row_id = (fragment_id << 32) | row_offset
```

But with stable row IDs, `row_id` is an opaque logical ID, so `row_id >> 32` is not a valid fragment test. As a result, valid old index entries can be dropped after compaction.

## Minimal reproduction

1. Create a dataset with `enable_stable_row_ids = true`
2. Write multiple fragments
3. Create a BTree index on a scalar column such as `id`
4. Verify indexed equality lookup works
5. Run `compact_files()`
6. Run `optimize_indices()`
7. Query the indexed column again with the index

Expected:

- Indexed equality lookup still returns the row

Actual:

- Full scan still finds the row
- Indexed equality lookup returns no rows

## Additional context

I wrote a longer deep dive about the bug, the stable-row-ID model, and the full repair process:

- https://ackingliu.top/posts/lance-stable-row-id-deep-dive

## Possibly related issues

- [#2307](https://github.com/lance-format/lance/issues/2307) Epic: Move-Stable Row Ids
- [#5326](https://github.com/lance-format/lance/issues/5326) Refactor mask related components to distinguish row ID and row address
- [#5872](https://github.com/lance-format/lance/issues/5872) FTS search fails after delete with stable row ID
- [#4767](https://github.com/lance-format/lance/issues/4767) Compaction should not allow both defer_index_remap and enable_stable_row_id
- [#5322](https://github.com/lance-format/lance/issues/5322) Format: add new `current_fragment_bitmap` field to `IndexMetadata`

acking-you:
# Preserve stable row-id entries during scalar index optimize
Fixes #6116

## Summary

This PR fixes a bug where `optimize_indices()` could drop valid BTree index entries when the dataset used stable row IDs.

I hit this while building the music module of [StaticFlow](https://ackingliu.top/), my personal project built on top of Lance/LanceDB. The `songs` dataset uses:

- `enable_stable_row_ids = true`
- a BTree scalar index on `id`

After running:

1. `compact_files()`
2. `optimize_indices()`

full scans still returned the expected rows, but indexed equality lookups such as `id = 'song-42'` returned no rows.

## Root Cause

The old optimize path filtered old BTree rows with logic equivalent to:

```rust
valid_fragments.contains((row_id >> 32) as u32)
```

That is correct for address-style row IDs:

```text
row_id = (fragment_id << 32) | row_offset
```

But it is incorrect for stable row IDs, because stable row IDs are opaque logical IDs and do not encode fragment ownership in their upper bits.

As a result, valid old index rows could be removed during optimize even though the underlying rows were still present after compaction.

## What This PR Changes

- adds an explicit old-data filter mode to scalar index update
- keeps fragment-based filtering for address-style row IDs
- builds an exact retained row-ID set for stable-row-ID datasets from persisted row-id sequences
- filters old BTree rows by exact row-ID membership for the stable-row-ID case
- adds regression coverage for both the BTree update path and the end-to-end compaction plus optimize flow

## Implementation Notes

The key change is to stop assuming that every `row_id` can be interpreted as a row address.

For stable-row-ID datasets, the optimize path now:

1. computes the retained old fragments
2. loads their row-ID sequences
3. builds one exact retained row-ID set
4. keeps only old index rows whose row IDs are still valid

This preserves the existing fast path for address-style row IDs and only uses exact row-ID filtering when the dataset actually uses stable row IDs.

## Additional Context

I also wrote a longer deep dive covering the bug, the stable-row-ID model, and the full repair process:

- https://ackingliu.top/posts/lance-stable-row-id-deep-dive

## Final Note

This PR document and parts of the implementation work were prepared with assistance from Codex GPT-5.4. The final patch has been reviewed by me personally, and it has been running normally in my production StaticFlow environment for one week.

### DIFF
--- a/rust/lance-index/src/scalar/btree.rs
+++ b/rust/lance-index/src/scalar/btree.rs
@@ -2628,9 +2628,9 @@ mod tests {
     use crate::{
         metrics::NoOpMetricsCollector,
         scalar::{
-            btree::{BTreeIndex, BTREE_PAGES_NAME},
-            lance_format::LanceIndexStore,
             IndexStore, OldIndexDataFilter, SargableQuery, ScalarIndex, SearchResult,
+            btree::{BTREE_PAGES_NAME, BTreeIndex},
+            lance_format::LanceIndexStore,
         },
     };
 
--- a/rust/lance/src/index/append.rs
+++ b/rust/lance/src/index/append.rs
@@ -5,14 +5,14 @@ use std::sync::Arc;
 
 use futures::{FutureExt, TryStreamExt};
 use lance_core::{
-    utils::mask::{RowAddrTreeMap, RowSetOps},
     Error, Result,
+    utils::mask::{RowAddrTreeMap, RowSetOps},
 };
 use lance_index::{
     metrics::NoOpMetricsCollector,
     optimize::OptimizeOptions,
     progress::NoopIndexBuildProgress,
-    scalar::{lance_format::LanceIndexStore, CreatedIndex, OldIndexDataFilter},
+    scalar::{CreatedIndex, OldIndexDataFilter, lance_format::LanceIndexStore},
 };
 use lance_table::format::{Fragment, IndexMetadata};
 use roaring::RoaringBitmap;
@@ -20,9 +20,9 @@ use uuid::Uuid;
 
 use super::DatasetIndexInternalExt;
 use super::vector::ivf::optimize_vector_indices;
+use crate::dataset::Dataset;
 use crate::dataset::index::LanceIndexStoreExt;
 use crate::dataset::rowids::load_row_id_sequences;
-use crate::dataset::Dataset;
 use crate::index::scalar::load_training_data;
 use crate::index::vector_index_details;
 
@@ -304,7 +304,9 @@ mod tests {
 
     use arrow::datatypes::{Float32Type, UInt32Type};
     use arrow_array::cast::AsArray;
-    use arrow_array::{FixedSizeListArray, RecordBatch, RecordBatchIterator, StringArray, UInt32Array};
+    use arrow_array::{
+        FixedSizeListArray, RecordBatch, RecordBatchIterator, StringArray, UInt32Array,
+    };
     use arrow_schema::{DataType, Field, Schema};
     use futures::TryStreamExt;
     use lance_arrow::FixedSizeListArrayExt;
@@ -314,9 +316,9 @@ mod tests {
     use lance_index::vector::hnsw::builder::HnswBuildParams;
     use lance_index::vector::sq::builder::SQBuildParams;
     use lance_index::{
+        DatasetIndexExt, IndexType,
         scalar::ScalarIndexParams,
         vector::{ivf::IvfBuildParams, pq::PQBuildParams},
-        DatasetIndexExt, IndexType,
     };
     use lance_linalg::distance::MetricType;
     use lance_testing::datagen::generate_random_array;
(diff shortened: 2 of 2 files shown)


acking-you:
@Xuanwo could you please help review this PR when you have time?

This bug is currently blocking my normal production use of LanceDB. In my case it causes indexed equality lookups to fail after `compact_files()` + `optimize_indices()` on datasets with stable
  row IDs enabled, even though the underlying rows are still present.

I also wrote up the full analysis and fix here:
https://ackingliu.top/posts/lance-stable-row-id-deep-dive

Thanks a lot.

rust/lance-index/src/scalar/btree.rs
```diff
@@ -1298,14 +1298,22 @@ impl BTreeIndex {
...
+        // Two filtering strategies:
+        // - Fragments: fast path for address-style row IDs (fragment id is encoded in row_id)
+        // - RowIds: exact allow-list for stable row IDs (row_id bits are opaque)
+        let old_stream = match old_data_filter {
+            Some(OldIndexDataFilter::Fragments(valid_frags)) => {
+                filter_row_ids_by_fragments(old_stream, valid_frags)
+            },
+            Some(OldIndexDataFilter::RowIds(valid_row_ids)) => {
+                filter_row_ids_by_exact_set(old_stream, valid_row_ids)
+            },
```
wjones127: It would be nice if `OldIndexDataFilter` encapsulated this distinction within.

Maybe something like:

```rust
impl OldIndexDataFilter {
    pub fn filter_row_ids(&self, row_ids: &UInt64Array) -> BooleanArray {
        match self {
            Self::Fragments(frags) => ...,
            Self::RowIds(row_ids) => ...,
    }
}
```

acking-you: Thanks, this is a good suggestion

wjones127 (approved):

PR #6118, opened 2026-03-06
feat: comprehensive observability instrumentation across Lance subsystems

esteban:
## Motivation

Lance had fragmented observability: mixed `log::`/`tracing::` usage, no structured in-memory stats for IO latency, cache hit rates, or scheduler behavior, and critical paths (write, commit retry, compaction, index build) were unobservable.

## Solution

Three-layer instrumentation (IO, Encoding, Dataset) using a consistent pattern:

- **Atomic counters** (`Arc<AtomicU64>` + `Ordering::Relaxed`, ~2ns amortized)
- **Typed stats structs** with `snapshot()` for point-in-time reads
- **`tracing` events** with `target:` fields for subscriber filtering
- **`#[non_exhaustive]`** on new public structs for forward compatibility

No direct OpenTelemetry dependency — users bring their own `tracing` subscriber.

### Crates modified

| Crate | Changes |
|-------|---------|
| `lance-core` | `CacheHitMiss`, `CacheStats`, eviction tracking |
| `lance-io` | `IoStats` (latency, errors), `ScanStats` (coalesce/split/backpressure), `RetryStats`, connection reset tracing, `list()` error tracking |
| `lance-encoding` | Decode pipeline timing, buffer operation tracing |
| `lance-index` | `MetricsCollector` extension, HNSW build progress, `MetricsSnapshot` |
| `lance` | `CompactionMetrics` (bytes, timing), `FragmentDistributionStats`, commit tracing, write path tracing, MemWAL `log::` -> `tracing::` migration |

### New stats structs

| Struct | Access | Key Fields |
|--------|--------|------------|
| `IoStats` | `IOTracker::stats()` | read/write iops, bytes, latency_us, errors |
| `ScanStats` | `ScanScheduler::stats()` | iops, bytes_read, coalesced/split_iops, backpressure_warnings |
| `CacheHitMiss` | `LanceCache::cache_hit_miss()` | hits, misses, evictions, `hit_ratio()` |
| `RetryStats` | `CloudObjectReader::retry_stats()` | total_retries, total_failures |
| `CompactionMetrics` | `compact_files()` return | bytes_rewritten, bytes_binary_copied, elapsed_ms_sum |
| `FragmentDistributionStats` | `Dataset::fragment_distribution_stats()` | count, min/max/avg physical_rows |
| `MetricsSnapshot` | `LocalMetricsCollector::snapshot()` | comparisons, bytes, partitions probed, search duration |

### Backward compatibility

- `#[serde(default)]` on new fields, `#[serde(alias)]` for renames
- Default no-op impls on `MetricsCollector` trait extensions
- `#[non_exhaustive]` on all new public structs

## Test plan

- [x] Unit tests for `RetryStats`, `CompactionMetrics` `AddAssign`, `IoTrackingStore` list() error counting
- [x] Integration test for session cache hit/miss sync
- [x] Full test suites passing: lance-core, lance-io, lance-encoding, lance-index, lance


rust/lance-encoding/src/buffer.rs
```diff
@@ -227,6 +239,12 @@ impl LanceBuffer {
     ///
     /// Unlike concat_into_one this "zips" the buffers, interleaving the values
     pub fn zip_into_one(buffers: Vec<(Self, u64)>, num_values: u64) -> Result<Self> {
+        tracing::debug!(
+            target: "lance_encoding::buffer",
+            num_buffers = buffers.len(),
+            num_values,
+            "zip buffers"
+        );
```
wjones127: I do worry a bit about the overhead of tracing at this level. It can be omitted at compile time, but I don't think everyone does that.

rust/lance-encoding/src/decoder.rs
```diff
@@ -2715,7 +2717,14 @@ pub async fn decode_batch(
...
         rx,
     )?;
-    decode_stream.next().await.unwrap().task.await
+    let result = decode_stream.next().await.unwrap().task.await;
+    tracing::debug!(
+        target: "lance_encoding::decode",
+        elapsed_us = start.elapsed().as_micros() as u64,
+        num_rows = batch.num_rows,
+        "batch decoded"
+    );
```
wjones127: Whenever I see a trace event / log that reports time, it makes me wonder if this should just be a span.

rust/lance/src/dataset.rs
```diff
@@ -2912,6 +2926,70 @@ impl Dataset {
...
+    pub total_physical_rows: u64,
+    /// Minimum physical rows in any single fragment.
+    pub min_physical_rows: u64,
+    /// Maximum physical rows in any single fragment.
+    pub max_physical_rows: u64,
+    /// Average physical rows per fragment.
+    pub avg_physical_rows: f64,
+    /// Number of fragments that have an associated deletion file.
+    pub fragments_with_deletions: usize,
+}
```
wjones127: Could you describe how you intend on using it? I think each compaction planner implementation will likely have it's own implementation of stats collection. And all of this information is easy to get from a dataset in our public API.

wjones127 (commented):
Thanks for the PR, I like most of these changes. I've been meaning to migrate away from `log` to `tracing`.

In the future, it would be appreciated to keep PRs to be more focused when possible. It makes reviewing them easier. Plus it also can get individual features through review faster--less controversial changes might get held up by refinements to more controversial changes. Some of these could be their own PR:
* Cache statistics
* IO statistics
* Compaction statistics - this one I'm skeptical of it's usefulness without knowing how it will be used.
* Index search metrics

rust/lance/src/dataset.rs
esteban: Good point, will remove. I don't think we have to keep FragmentDistributionStats for now I was thinking more as part of propagating it in the traces, but we probably don't need that for now.

rust/lance-encoding/src/decoder.rs
esteban: +1 moved it to span.

rust/lance-encoding/src/buffer.rs
esteban: removing it, probably in very tight loops this can be problem.

esteban:
Thanks for the review @wjones127!

> In the future, it would be appreciated to keep PRs to be more focused when possible. It makes reviewing them easier. Plus it also can get individual features through review faster--less controversial changes might get held up by refinements to more controversial changes. Some of these could be their own PR:... 

I completely understand, I've been just piling this up in a local branch as I've been found the need to get better observability. Will try to provide mode atomic and focused PRs in the future!

FarmerChillax:
Cool! This feature is highly needed.

PR #6119, opened 2026-03-06
feat: enable HNSW-accelerated partition assignment for fp16 vectors

wkalt:
SimpleIndex (HNSW over centroids) previously only supported fp32 centroids, causing fp16 vector workloads to fall back to brute-force partition assignment — O(K×D) per vector instead of O(log K × D). For 31K centroids × 1024 dims this is a ~600x difference per vector.

Cast fp16 centroids to fp32 at HNSW construction time (one-time cost) and cast fp16 query vectors at search time (1024 floats per query, negligible vs the distance computations saved).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4926,6 +4926,7 @@ dependencies = [
  "prost-build",
  "protobuf-src",
  "snafu",
+ "substrait",
  "tokio",
  "tracing",
 ]
--- a/java/lance-jni/src/utils.rs
+++ b/java/lance-jni/src/utils.rs
@@ -430,6 +430,7 @@ pub fn get_vector_index_params(
                 metric_type: distance_type,
                 stages,
                 version: IndexFileVersion::V3,
+                skip_transpose: false,
             })
         },
     )?;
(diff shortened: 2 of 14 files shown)


rust/lance-index/src/vector/utils.rs
```diff
@@ -78,18 +79,19 @@ impl SimpleIndex {
...
 
-        match centroids.data_type() {
-            DataType::Float32 => {
-                let fsl =
-                    FixedSizeListArray::try_new_from_values(centroids.clone(), dimension as i32)?;
-                let store = FlatFloatStorage::new(fsl, distance_type);
-                Self::try_new(store).map(Some)
+        let f32_centroids = match centroids.data_type() {
+            DataType::Float16 | DataType::Float32 => {
+                cast(&centroids, &DataType::Float32).map_err(|e| Error::index(e.to_string()))?
```
westonpace, line 84: It probably is but can we verify this cast is a no-op if the data type is already `Float32`?

wkalt: It's a clone of the ArrayData here, so no element-by-element conversion:

https://github.com/apache/arrow-rs/blob/main/arrow-cast/src/cast/mod.rs#L748-L750

PR #6120, opened 2026-03-06
fix: disallow wrapping auto-detected fsst in other compression

hamersaw:
Currently, if a user manually sets a compression type (ex. zstd) we can inject FSST compression underneath. Resulting in `zstd(fsst(data))` rather than `zstd(data)` that the user intended. This PR updates our auto-injection of FSST to only occur if the user has not specified a compression type manually.

Related to https://github.com/lance-format/lance/discussions/5248


Xuanwo (approved):
Thank you!

PR #6122, opened 2026-03-07
fix: vector index type shown as unknown in describe_indices

jackye1995:
Vector index still needs to be opened to get the right type, otherwise it is shown as unknown.


wjones127:
FYI this will also be fixed by https://github.com/lance-format/lance/pull/6099, but that will take longer to merge.

jackye1995:
> FYI this will also be fixed by https://github.com/lance-format/lance/pull/6099, but that will take longer to merge.

Nice, totally agree that's the right way. I guess the question is do we think this is a regression so we should have this in place first. But if we are not patching it, then it does not really make much sense

Xuanwo (approved):

PR #6128, opened 2026-03-08
feat: add DecodeFieldError/ReadFileError wrappers for error context

### LINKED ISSUE #5602: Add `ReadFileError` and `DecodeColumnError` wrappers to provide better context for decoding errors
Currently, errors that happen when decoding a column are missing information about:

1. What column that is
2. What file we are reading

This context is available higher up in the call stack that the main error is generate. We can add additional context by adding wrapper errors.

This approach is described nicely in https://sabrinajewson.org/blog/errors.

Basically, want to do things like:

```rust
enum DecodeError {
    NumberOutOfRange,
    ...
}

struct DecodeFieldError {
    field_name: String,
    field_id: i32,
    offset: u64,
    source: DecodeError,
}

fn decode(field: Field, data: Bytes) -> Result<_, DecodeFieldError> {
    ...
}

enum ReadFileErrorKind {
    IO(ObjectStoreError),
    Decode(DecodeFieldError),
}

struct ReadFileError {
    path: Path,
    kind: ReadFileErrorKind,
}

fn read_file(path: Path) -> Result<_, ReadFileError> {
    let data = read_data(path)?;
    for field in data.schema() {
        decode(field, data).map_err(|err: DecodeFieldError| {
            ReadFileError { path, kind: ReadFileErrorKind::Decode(err) }
        });
    }
}
```

And then the error stack would look like:

```
Error: failed to read file 'data.lance'

Caused by:
    0: failed to read field 'age' (id: 10) for row offset 10
    1: number out of range
```

koriyoshi2041:
## Summary

Adds two wrapper error types to provide better context when decoding errors occur, as described in #5602:

- **`DecodeFieldError`** (`lance-encoding/src/decoder.rs`): wraps errors with field name and field id. Applied at field scheduler creation in both structural and legacy decode paths (Struct, List, FixedSizeList, Map).
- **`ReadFileError`** (`lance-file/src/reader.rs`): wraps errors with file path. Applied at `FileReader::read_tasks` and `read_stream_projected_blocking` entry points. `FileReader` now stores the file path.

Both implement `std::error::Error` with proper source chaining and `From<...> for lance_core::Error`, producing error chains like:

```
Error: failed to read file 'data.lance'

Caused by:
    0: failed to decode field 'age' (id=10)
    1: number out of range
```

Design follows the approach from https://sabrinajewson.org/blog/errors as referenced in the issue.

## Test plan

- [x] Unit tests for `DecodeFieldError` display, source chain, and conversion to `lance_core::Error` (3 tests)
- [x] Unit tests for `ReadFileError` display, source chain with nested `DecodeFieldError`, and conversion to `lance_core::Error` (3 tests)
- [x] `cargo fmt --check` passes
- [x] `cargo clippy --all-targets -p lance-encoding -p lance-file` passes
- [x] All existing tests pass

Closes #5602

### DIFF
--- a/rust/lance-encoding/src/decoder.rs
+++ b/rust/lance-encoding/src/decoder.rs
@@ -298,7 +298,7 @@ impl std::error::Error for DecodeFieldError {
 
 impl From<DecodeFieldError> for Error {
     fn from(err: DecodeFieldError) -> Self {
-        Self::io_source(Box::new(err))
+        Self::wrapped(Box::new(err))
     }
 }
 
@@ -690,8 +690,14 @@ impl CoreFieldDecoderStrategy {
             file_buffers: buffers,
             positions_and_sizes: &offsets_column.buffer_offsets_and_sizes,
         };
-        let items_scheduler =
-            self.create_legacy_field_scheduler(&list_field.children[0], column_infos, buffers)?;
+        let child = &list_field.children[0];
+        let items_scheduler = self
+            .create_legacy_field_scheduler(child, column_infos, buffers)
+            .map_err(|source| DecodeFieldError {
+                field_name: child.name.clone(),
+                field_id: child.id,
+                source,
+            })?;
 
         let (inner_infos, null_offset_adjustments): (Vec<_>, Vec<_>) = offsets_column
             .page_infos
@@ -2952,6 +2958,7 @@ pub async fn decode_batch(
 }
 
 #[cfg(test)]
+// test coalesce indices to ranges
 mod tests {
     use super::*;
 
--- a/rust/lance-encoding/src/decoder.rs
+++ b/rust/lance-encoding/src/decoder.rs
@@ -235,6 +235,7 @@ use tokio::sync::mpsc::{self, unbounded_channel};
 
 use lance_core::error::LanceOptionExt;
 use lance_core::{ArrowResult, Error, Result};
+use std::fmt;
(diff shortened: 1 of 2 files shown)


koriyoshi2041:
Thanks for the thorough review! All three issues have been addressed in the latest commit:

1. **P1 — Legacy decode path**: Added `DecodeFieldError` wrapping in `create_list_scheduler` for child field errors, matching the structural path's behavior.

2. **P1 — Error category preservation**: Changed `From<DecodeFieldError>` and `From<ReadFileError>` to use `Self::wrapped()` instead of `Self::io_source()`, so the original error category is preserved in the chain.

3. **Nit — Restored comment**: Put back the `// test coalesce indices to ranges` comment above `mod tests`.

All 428 tests pass (362 lance-encoding + 66 lance-file).

koriyoshi2041:
Rebased onto current `main` (the branch had drifted behind). The only conflicts were in the `decoder.rs` / `reader.rs` test modules where `main` had since added neighbouring tests — kept both sets. The production changes (the `DecodeFieldError` / `ReadFileError` wrappers and the `map_err` sites) applied cleanly.

Re-verified locally: `cargo test -p lance-encoding -p lance-file` (the 6 wrapper tests pass), `cargo clippy -p lance-encoding -p lance-file --tests`, and `cargo fmt` are all clean. Ready for another look when you have a moment.

PR #6132, opened 2026-03-09
feat(namespace): add count_table_rows, insert_into_table, query_table

XuQianJin-Stars:
close https://github.com/lance-format/lance/issues/6111

- Implement `count_table_rows` with optional version checkout and predicate filter
- Implement `insert_into_table` with append/overwrite modes via Arrow IPC
- Implement `query_table` supporting vector similarity search (with distance type, nprobes, refine factor, prefilter) and plain scan with filter/limit/offset/version
- Add `lance-linalg` dependency for `DistanceType` in vector search
- Add 8 unit tests covering all new methods and edge cases

## Test plan

- [x] `cargo test -p lance-namespace-impls` passes all new tests
- [x] count_table_rows returns correct count with and without predicate filter
- [x] insert_into_table correctly appends and overwrites data
- [x] query_table returns correct results for vector search and plain scan

### DIFF
--- a/python/python/tests/test_namespace_dir.py
+++ b/python/python/tests/test_namespace_dir.py
@@ -1335,6 +1335,10 @@ class TestDataManipulation:
         assert result_table.num_rows == 2  # Alice and Charlie
 
 
+@pytest.mark.skipif(
+    sys.platform == "win32",
+    reason="Table version listing not supported on Windows",
+)
 class TestTableVersions:
     """Tests for table version operations."""
 


rust/lance-namespace-impls/src/dir.rs
```diff
@@ -1809,6 +1822,220 @@ impl LanceNamespace for DirectoryNamespace {
...
+            let batches = vec![Ok(batch)];
+            RecordBatchIterator::new(batches, arrow_schema.clone())
+        } else {
+            let batch_results: Vec<_> = batches.into_iter().map(Ok).collect();
+            RecordBatchIterator::new(batch_results, arrow_schema)
+        };
+
+        let mode = match request.mode.as_deref() {
+            Some(m) if m.eq_ignore_ascii_case("overwrite") => lance::dataset::WriteMode::Overwrite,
+            _ => lance::dataset::WriteMode::Append,
```
yanghua: To be more seriously, we may need to enumale the mode more carefully. For example, if it is not `append`, we need to throw an exception?

rust/lance-namespace-impls/src/dir.rs
```diff
@@ -1809,6 +1822,220 @@ impl LanceNamespace for DirectoryNamespace {
...
+
+        if has_single_vector {
+            // Vector similarity search path
+            let single = request.vector.single_vector.as_ref().unwrap();
+            let query_array = arrow::array::Float32Array::from(single.clone());
+            scanner
+                .nearest(
+                    request.vector_column.as_deref().unwrap_or("vector"),
+                    &query_array,
+                    request.k as usize,
```
yanghua: Check the `k` arg? Should it be larger than 0?

PR #6142, opened 2026-03-10
refactor: use the dataset file version to determine index file version

westonpace:
Previously we would use the default file version when creating new index files.  This was originally done to get some testing of the 2.0 format before it was made the default.  However, this led to a bit of a potential compatibility problem.  If we change the default file version then the files created by the new release would become unreadable on very old versions that didn't know how to read that file, even if the dataset itself had an older file version and the old version knew how to handle the index otherwise.

To avoid this we change things in this PR so that new index files use the same format version as the dataset.  This should mean the indexes are always readable if the dataset is readable, regardless of what version was used to write the index.

---

Parts of this PR were written with Claude (Opus 4.6) and I take full responsibility for its contents.


Xuanwo:
Some CI failed, otherwise the changes look good to me

PR #6143, opened 2026-03-10
fix: prevent duplicate manifest entries from concurrent table creation

jmhsieh:
Change conflict_retries from 0 to 5 in insert_into_manifest so that cross-process races are handled correctly. When two processes concurrently insert the same object_id, the second one hits a commit version conflict. With conflict_retries > 0, MergeInsert retries by re-evaluating the full plan against the latest data, where the join detects the existing row and WhenMatched::Fail fires properly.

Previously, conflict_retries=0 meant the second operation would fail with a generic TooMuchWriteContention error, but in some cases both commits could succeed creating duplicate manifest entries ("Expected exactly 1 table...found 2").

Add test with two independent ManifestNamespace instances racing on the same directory to verify no duplicates are created.

Here's an example I run into occasionally
```
...
  File "/home/runner/work/geneva/geneva/src/geneva/state/manager.py", line 35, in __init__
    self.table = alter_or_create_table(
  File "/home/runner/work/geneva/geneva/src/geneva/utils/schema.py", line 138, in alter_or_create_table
    return db.create_table(table_name, schema=schema, namespace=namespace)
  File "/home/runner/work/geneva/geneva/src/geneva/db.py", line 403, in create_table
    return Table(self, name, namespace=namespace, storage_options=storage_options)
  File "/home/runner/work/geneva/geneva/src/geneva/table.py", line 489, in __init__
    self._ltbl  # noqa
  File "/home/runner/.local/share/uv/python/cpython-3.10-linux-x86_64-gnu/lib/python3.10/functools.py", line 981, in __get__
    val = self.func(instance)
  File "/home/runner/work/geneva/geneva/src/geneva/table.py", line 543, in _ltbl
    tbl = inner.open_table(self.name, namespace=self._namespace)
  File "/home/runner/work/geneva/geneva/.venv/lib/python3.10/site-packages/lancedb/namespace.py", line 392, in open_table
    response = self._ns.describe_table(request)
  File "/home/runner/work/geneva/geneva/.venv/lib/python3.10/site-packages/lance/namespace.py", line 362, in describe_table
    response_dict = self._inner.describe_table(request.model_dump())
OSError: LanceError(IO): Expected exactly 1 table with id 'default$geneva_manifests', found 2, /home/runner/work/lance/lance/rust/lance-namespace-impls/src/dir/manifest.rs:642:21
```

### DIFF
--- a/.bumpversion.toml
+++ b/.bumpversion.toml
@@ -1,5 +1,5 @@
 [tool.bumpversion]
-current_version = "4.0.0-beta.7"
+current_version = "4.0.0-beta.8"
 parse = "(?P<major>\\d+)\\.(?P<minor>\\d+)\\.(?P<patch>\\d+)(-(?P<prerelease>(beta|rc))\\.(?P<prerelease_num>\\d+))?"
 serialize = [
     "{major}.{minor}.{patch}-{prerelease}.{prerelease_num}",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3443,7 +3443,7 @@ checksum = "42703706b716c37f96a77aea830392ad231f44c9e9a67872fa5548707e11b11c"
 
 [[package]]
 name = "fsst"
-version = "4.0.0-beta.7"
+version = "4.0.0-beta.8"
 dependencies = [
  "arrow-array",
  "lance-datagen",
@@ -4212,7 +4212,7 @@ dependencies = [
  "libc",
  "percent-encoding",
  "pin-project-lite",
- "socket2 0.6.2",
+ "socket2 0.6.3",
  "system-configuration",
  "tokio",
  "tower-service",
@@ -4735,7 +4735,7 @@ dependencies = [
 
 [[package]]
 name = "lance"
-version = "4.0.0-beta.7"
+version = "4.0.0-beta.8"
 dependencies = [
  "all_asserts",
  "approx",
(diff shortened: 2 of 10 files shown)


jmhsieh:
duck db related failures seem to have been fix on main.  merged with latest

rust/lance-namespace-impls/src/dir/manifest.rs
```diff
@@ -862,9 +862,12 @@ impl ManifestNamespace {
...
         merge_builder.when_matched(lance::dataset::WhenMatched::Fail);
         merge_builder.when_not_matched(lance::dataset::WhenNotMatched::InsertAll);
-        // conflict_retries=0: no outer loop retry on semantic conflicts (handled by caller)
-        // commit_retries: inner retry for manifest version conflicts (uses lance default if not set)
-        merge_builder.conflict_retries(0);
+        // Use conflict_retries to handle cross-process races on manifest mutations.
+        // When two processes concurrently insert the same object_id, the second one
+        // hits a commit conflict. With conflict_retries > 0, the retry re-evaluates
+        // the full MergeInsert plan against the latest data, where the join detects
+        // the existing row and WhenMatched::Fail fires, producing a clear error.
```
westonpace, line 869: Why 5 and not 1?  Even if we have 10 transactions running at the same time won't transactions 2-10 all hit the conflict error before they attempt to commit again (so 1 retry is enough)?

jmhsieh: I have no deep reason for 5. 

FWIW, claude came up with a scenario related to multiple distinct tables being created and the fact that namespaces use lance tables that can conflict underneath when multiple different tables are created (not just the same table).

PR #6154, opened 2026-03-10
docs: shorten core major release vote window

Xuanwo:
This updates the documented minimum voting period for core stable major releases from 1 week to 3 days. It also updates the RC voting discussion automation so generated vote windows stay aligned with the documented policy.

This follows the discussion in https://github.com/lance-format/lance/discussions/6147, where feedback converged on narrowing the proposal to core major releases only while keeping patch and subproject stable releases unchanged.


Xuanwo:
Vote has been passed, let's merge.

PR #6158, opened 2026-03-10
docs: document the rules for transaction conflicts

westonpace:
There was a conflict table in transaction.rs but this was incomplete (some rows/columns missing) and seemed to be imprecise or incorrect in a few spots.  I've attempted to more thoroughly document this in transaction.md instead.

### DIFF
--- a/docs/src/format/table/transaction.md
+++ b/docs/src/format/table/transaction.md
@@ -186,7 +186,6 @@ that are retryable conflicts with index creation:
 
 - Rewrite (only if overlapping fragments, no stable row ids, and no fragment reuse index)
 - DataReplacement (only if overlapping fragments and the column being replaced is being indexed)
-- Merge ??? TODO - test this
 
 Some indices are special singleton indices. For example, the fragment reuse index and the mem wal index. If a conflict occurs
 between two operations that are modifying the same singleton index, then we must rebase the operation and merge the indexes.
@@ -270,6 +269,7 @@ The following operations conflict with Merge:
 - Overwrite
 - Restore
 - UpdateMemWalState
+- Project
 
 These operations are retryable conflicts with Merge:
 
@@ -279,7 +279,6 @@ These operations are retryable conflicts with Merge:
 - Merge (always)
 - Rewrite (always)
 - DataReplacement (always)
-- Project (always)
 
 ### Project
 


docs/src/format/table/transaction.md
```diff
@@ -107,6 +159,41 @@ Adds, replaces, or removes secondary indices (vector indices, scalar indices, fu
...
+Indexes record which fragments are covered by the index and we don't require all fragments be covered. As a result, it
+is typically ok for an index to be created concurrently with the addition of new fragments. These new fragments will simply
+be unindexed.
+
+Updates and deletes are also compatible with index creation. This is because it is ok for an index to refer to deleted rows.
+Those results will be filtered out after the index search. If an update occurs then the old value will be filtered out and the
+new value will be considered part of the unindexed set.
+
+If two CreateIndex operations are committed concurrently then it is allowed. If the indexes have different names this is no
+problem. If the indexes have the same name then the second operation will win and replace the first.
```
wjones127, line 173: This seems accurate to the current implementation. Though I think we should probably change it so you can commit index segments for disjoint fragment sets and it be additive.

docs/src/format/table/transaction.md
```diff
@@ -107,6 +159,41 @@ Adds, replaces, or removes secondary indices (vector indices, scalar indices, fu
...
+
+Data replacement operations will conflict with index creation if the column being replaced is being indexed. Rewrite operations
+will conflict with index creation if the rewritten fragments are covered by the index. This is because an index refers to row
+addresses and the rewrite operation changes the row addresses. However, if a fragment reuse index is being used, or if the stable
+row ids feature is enable, then the rewrite operation is compatible with index creation. As a result, these are the operations
+that are retryable conflicts with index creation:
+
+- Rewrite (only if overlapping fragments, no stable row ids, and no fragment reuse index)
+- DataReplacement (only if overlapping fragments and the column being replaced is being indexed)
+- Merge ??? TODO - test this
```
wjones127: Do we have an answer for this?

westonpace: Yes, it is not a conflict.  I was originally thinking merge == update and I was confused how it couldn't be if the merge could replace a column.  However, the logic in merge seems to assume we are only ever adding columns and so it considers it ok.

docs/src/format/table/transaction.md
westonpace: Now that we are allowed multiple UUIDs per name that does make sense.

PR #6162, opened 2026-03-11
fix: persist frag reuse index external file on local filesystem

### LINKED ISSUE #6161: Index file not found during FTS prewarm
Repro:

```python
import os
import shutil
import tempfile

import pyarrow as pa

import lance
from lance._datagen import rand_batches

schema = pa.schema(
    [pa.field("text", pa.string(), metadata={"lance-datagen:content-type": "sentence"})]
)

data = rand_batches(schema, num_batches=10, batch_size_bytes=1024 * 1024)
with tempfile.TemporaryDirectory() as tmp_dir:
    uri = os.path.join(tmp_dir, "test")
    shutil.rmtree(uri, ignore_errors=True)  # Ensure clean slate
    ds = lance.write_dataset(data, uri, schema=schema, max_rows_per_file=1000)
    ds.create_scalar_index("text", index_type="INVERTED", with_positions=True)
    ds.optimize.compact_files(defer_index_remap=True)
    ds.prewarm_index("text_idx")
```

```
  File "/home/will/Documents/lance/python/bug_repro.py", line 24, in <module>
    ds.prewarm_index("text_idx")
  File "/home/will/Documents/lance/python/python/lance/dataset.py", line 3336, in prewarm_index
    return self._ds.prewarm_index(name)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ValueError: Not found: home/will/Documents/lance/python/test/_indices/31c8806b-ec9c-48c7-a44e-2cc2e384b0d2/details.binpb, /home/will/Documents/lance/rust/lance-io/src/local.rs:132:40
```

This repros for relative and absolute local paths. Doesn't repro on `memory://` store.

wjones127:
Previously, when `FragReuseIndexDetails` exceeded 204800 bytes (triggered by large compactions with many fragments), the code wrote the details to an external file (`details.binpb`). On local filesystems, `ObjectStore::create` returns a `LocalWriter` that atomically renames a temp file to the final path in `Writer::shutdown`. However, `frag_reuse.rs` imported `tokio::io::AsyncWriteExt` but not `lance_io::traits::Writer`, so `writer.shutdown()` resolved to `AsyncWriteExt::shutdown` (flush/close only) — the temp file was deleted on drop without being persisted. Any subsequent `load_indices` call would fail with `Not found: .../details.binpb`.

Fixed by using UFCS `Writer::shutdown(writer.as_mut()).await?` to explicitly call the lance trait method, matching the existing pattern in `ivf.rs` and `blob.rs`.

Fixes #6161

### DIFF
--- a/rust/lance/src/dataset/optimize.rs
+++ b/rust/lance/src/dataset/optimize.rs
@@ -2291,6 +2291,71 @@ mod tests {
         assert_eq!(before_scalar_result, after_scalar_result);
     }
 
+    // Regression test for https://github.com/lancedb/lance/issues/6161
+    // When FragReuseIndexDetails exceeds 204800 bytes it is written to an external
+    // file. Previously the file was silently dropped (temp file deleted) because
+    // tokio::io::AsyncWriteExt::shutdown was called instead of
+    // lance_io::traits::Writer::shutdown, which persists the temp file.
+    #[tokio::test]
+    async fn test_defer_index_remap_large_external_file() {
+        let test_dir = TempStrDir::default();
+        let test_uri = &test_dir;
+
+        // Create ~150 fragments × 1000 rows to produce a FragReuseIndexDetails
+        // that exceeds the 204800-byte inline threshold (~302 KB serialized).
+        let num_fragments = 150usize;
+        let rows_per_fragment = 1000usize;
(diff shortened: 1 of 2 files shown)


wjones127:
Recreating from a feature branch.

PR #6168, opened 2026-03-11
fix: avoid empty range reads for zero-length blobs

Xuanwo:
Blob reads should return empty bytes when the logical blob is empty or the cursor is already at EOF. Today `BlobFile::read` / `read_up_to` can still issue a `get_range(start..end)` request with `start == end`, which is tolerated by local readers but rejected by cloud object stores.

This showed up while investigating `random_blob` failures on the original-scale `laion10m-full` dataset, where legacy blob reads on S3 failed with errors like `Range started at 1 and ended at 1`. The fix short-circuits empty reads and restores the cursor to blob-relative semantics after `read()`, and adds regression coverage for both the empty-range case and packed-blob cursor behavior.

### DIFF
--- a/.github/workflows/cargo-publish.yml
+++ b/.github/workflows/cargo-publish.yml
@@ -27,7 +27,7 @@ env:
 jobs:
   build:
     # Needs additional disk space for the full build.
-    runs-on: warp-ubuntu-latest-x64-8x
+    runs-on: ubuntu-24.04-8x
     permissions:
       id-token: write
     timeout-minutes: 60
--- a/.github/workflows/codex-backport-pr.yml
+++ b/.github/workflows/codex-backport-pr.yml
@@ -35,7 +35,7 @@ permissions:
 
 jobs:
   backport:
-    runs-on: warp-ubuntu-latest-x64-4x
+    runs-on: ubuntu-24.04-4x
     timeout-minutes: 60
     env:
       CC: clang
--- a/.github/workflows/codex-fix-ci.yml
+++ b/.github/workflows/codex-fix-ci.yml
@@ -35,7 +35,7 @@ permissions:
 
 jobs:
   fix-ci:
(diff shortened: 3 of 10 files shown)


rust/lance/src/index/vector/ivf.rs
```diff
@@ -2351,11 +2351,13 @@ mod tests {
...
-                return self.centroids.as_ref().unwrap();
+            if self.centroids.is_none() {
+                self.centroids = Some(self.generate_centroids());
             }
-            self.centroids = Some(self.generate_centroids());
-            self.centroids.as_ref().unwrap()
+            let Some(centroids) = self.centroids.as_ref() else {
+                unreachable!("centroids should be initialized before access")
+            };
+            centroids
```
wjones127: How about this?
```suggestion
            self.centroids.get_or_insert_with(|| self.generate_centroids())
```

rust/lance/src/index/vector/ivf.rs
```diff
@@ -2389,11 +2391,13 @@ mod tests {
...
-                return self.vectors.as_ref().unwrap();
+            if self.vectors.is_none() {
+                self.vectors = Some(self.generate_vectors());
             }
-            self.vectors = Some(self.generate_vectors());
-            self.vectors.as_ref().unwrap()
+            let Some(vectors) = self.vectors.as_ref() else {
+                unreachable!("vectors should be initialized before access")
+            };
+            vectors
```
wjones127: Same here:
```suggestion
            self.vectors.get_or_insert_with(|| self.generate_vectors())
```

rust/lance/src/index/vector/ivf.rs
Xuanwo: Thank you! I think I introduced those changes just to make Clippy happy. Let me submit another PR instead to avoid mixing them in this PR.

PR #6169, opened 2026-03-11
feat: reduce open file handles during IVF training

westonpace:
The previous shuffler used one open file per partition.  At large scales this meant tedious re-adjusting of OS limits.  The new shuffler uses two open files.  We potentially introduce a bit more random access in the later read phase but the overall performance hasn't changed significantly.

### DIFF
--- a/rust/lance-index/src/vector/v3/shuffler.rs
+++ b/rust/lance-index/src/vector/v3/shuffler.rs
@@ -704,8 +704,8 @@ mod tests {
     use arrow_schema::{DataType, Field, Schema as ArrowSchema};
     use futures::stream;
     use lance_arrow::RecordBatchExt;
+    use lance_core::utils::tempfile::TempStrDir;
     use lance_io::stream::RecordBatchStreamAdapter;
-    use tempfile::tempdir;
 
     use crate::vector::{LOSS_METADATA_KEY, PART_ID_COLUMN};
 
@@ -755,8 +755,8 @@ mod tests {
 
     #[tokio::test]
     async fn test_two_file_shuffler_round_trip() {
-        let dir = tempdir().unwrap();
-        let output_dir = Path::from(dir.path().to_str().unwrap());
+        let dir = TempStrDir::default();
+        let output_dir = Path::from(dir.as_ref());
         let num_partitions = 3;
 
         // Partition 0: rows with values 10, 40
@@ -799,8 +799,8 @@ mod tests {
 
     #[tokio::test]
     async fn test_two_file_shuffler_empty_partitions() {
-        let dir = tempdir().unwrap();
-        let output_dir = Path::from(dir.path().to_str().unwrap());
+        let dir = TempStrDir::default();
+        let output_dir = Path::from(dir.as_ref());
         let num_partitions = 5;
 
         // Only use partitions 0 and 3, leaving 1, 2, 4 empty
@@ -828,8 +828,8 @@ mod tests {
 
     #[tokio::test]
     async fn test_two_file_shuffler_loss_tracking() {
-        let dir = tempdir().unwrap();
-        let output_dir = Path::from(dir.path().to_str().unwrap());
+        let dir = TempStrDir::default();
+        let output_dir = Path::from(dir.as_ref());
         let num_partitions = 2;
 
         let batch1 = make_batch(&[0, 1], &[10, 20], Some(1.5));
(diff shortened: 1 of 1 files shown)


rust/lance-index/src/vector/v3/shuffler.rs
```diff
@@ -286,3 +292,621 @@ impl ShuffleReader for EmptyReader {
...
+            ReadBatchParams::Indices(positions),
+            num_positions,
+            1,
+            FilterExpression::no_filter(),
+        )?;
+        let schema = offsets_stream.schema().clone();
+        let offsets = offsets_stream.try_collect::<Vec<_>>().await?;
+        let offsets = if offsets.len() == 0 {
+            // We should not hit this path if there is no batches
+            unreachable!()
```
wkalt, line 633: should it panic with a message?

westonpace: `unreachable` panics with the message "I didn't expect to get here" and the line number which is pretty much what I'd use if I did an explicit panic.

rust/lance-index/src/vector/v3/shuffler.rs
```diff
@@ -286,3 +292,621 @@ impl ShuffleReader for EmptyReader {
...
+            concat_batches(&schema, &offsets)?
+        };
+
+        let offsets = offsets.column(0).as_primitive::<UInt64Type>();
+        let is_uneven = offsets.len() % 2 != 0;
+        let mut offsets_iter = offsets.values().iter().copied();
+
+        let mut ranges = Vec::with_capacity(self.num_batches as usize);
+        for batch_idx in 0..self.num_batches {
+            if batch_idx == 0 && is_uneven {
```
wkalt: is the is_uneven stuff equivalent to checking if partition_id = 0 and batch_idx = 0?

westonpace: Yes, that's maybe simpler.  Updated.

wkalt:
wpace_progress_full

here is the result on my test of 500M rows for posterity.

The previous behavior looked like this:
progress

westonpace:
> P1: Potential u32 overflow in partition_ranges

This is a legitimate concern.  Unfortunately, we don't actually support take with u64 offsets.  This is because we're reusing some old paths from the v0.1 days where we used u32 offsets.  It's all fixable but more follow-up territory.

By my math I think, even with 1536-dimension vectors, we are ok until we hit trillions of rows.

1T rows => sqrt(1T) * 96 * 1T bytes of PQ data => 96EB of data which, split into 128MB chunks, would give us ~1B chunks.

Either way, I turned it into a `try_into` so we get a nice error and can address in the future.

rust/lance-index/src/vector/v3/shuffler.rs
```diff
@@ -286,3 +281,610 @@ impl ShuffleReader for EmptyReader {
...
+}
+
+/// Limit of how much transformed data we accumulate before spilling to disk.
+///
+/// A larger value will use more RAM but require less random access during the
+/// read phase.
+///
+/// This default is likely to be fine for most use cases.
+fn shuffle_batch_bytes() -> usize {
+    std::env::var("LANCE_SHUFFLE_BATCH_BYTES")
```
Xuanwo: I think it's a good idea to add some protection here. This would prevent users from setting `LANCE_SHUFFLE_BATCH_BYTES` to `0`, which could lead to using `batch_size_bytes = 0` and producing incorrect results.

rust/lance-index/src/vector/v3/shuffler.rs
```diff
@@ -286,3 +281,610 @@ impl ShuffleReader for EmptyReader {
...
+                let batch = batch.take(&indices)?;
+
+                let part_ids: &UInt32Array = batch[PART_ID_COLUMN].as_primitive();
+                let batch = batch.drop_column(PART_ID_COLUMN)?;
+
+                // Count rows per partition by scanning sorted part IDs
+                let mut partition_counts = vec![0u64; np];
+                for i in 0..part_ids.len() {
+                    let pid = part_ids.value(i) as usize;
+                    if pid < np {
```
Xuanwo, line 485: Is it possible for `pid >= np`? It looks like we will just write those data without offsets. `global_row_count` will always include them.

rust/lance-index/src/vector/v3/shuffler.rs
westonpace: Good idea, now it will log a warning and use the default.

rust/lance-index/src/vector/v3/shuffler.rs
westonpace: It shouldn't be possible (`np` is `num_partitions`) but I now log a warning instead of silently ignoring.

westonpace:
CI failure _seems_ unrelated.  Will merge if remaining CI job passes.

PR #6171, opened 2026-03-11
fix: avoid fts corruption when a fragment is deleted

westonpace:
This is basically the same problem encountered in https://github.com/lance-format/lance/pull/5929

If an entire fragment is deleted then, after an FTS update, the rows in that deleted fragment will still be indexed but shouldn't be.  This fix uses the `old_data_filter` to properly prune the old data during the update.

---

This PR was created with the help of Claude Opus 4.6.  I take responsibility for all code.


westonpace:
Superseded by #6216

PR #6173, opened 2026-03-12
feat: support atomic multi-table transactions via namespace manifest

XuQianJin-Stars:
## Summary

Add a new `table_version_storage_enabled` flag to `DirectoryNamespaceBuilder` that stores table versions in the `__manifest` table (as `table_version` object type) instead of relying solely on Lance's native version management. This enables atomic multi-table version creation via `BatchCreateTableVersions`, multi-table deletion via `BatchDeleteTableVersions`, and a unified atomic `BatchCommitTables` operation that supports mixed operations (declare, create version, delete versions, deregister) across multiple tables in a single transaction.

### manifest.rs
- Add `TableVersion` variant to `ObjectType` enum
- Make `str_object_id` and `insert_into_manifest_with_metadata` public
- Add `batch_insert_into_manifest()`: atomically inserts multiple entries using a single MergeInsert with `WhenMatched::Fail` semantics, so if any version already exists the entire batch is rejected
- Add `query_table_versions()`: lists all versions for a given table from the manifest, returning version number and full metadata JSON
- Add `query_table_version()`: queries a single version entry by table ID and version number, returning its metadata JSON
- Add `delete_table_versions()`: removes version entries by table ID and version ranges from the manifest
- Add `batch_delete_table_versions_by_object_ids()`: atomically deletes version entries across multiple tables in a single `DeleteBuilder` operation

### dir.rs
- Add `table_version_storage_enabled` field with builder method and validation (requires `manifest_enabled=true`)
- `CreateTableVersion`: when enabled, also records the version entry in `__manifest` via `insert_into_manifest_with_metadata`
- `BatchCreateTableVersions`: two-phase commit — first copies staging manifests to final locations, then atomically inserts all version entries into `__manifest` in a single batch operation
- `ListTableVersions`: when enabled, queries versions from `__manifest` and reconstructs `TableVersion` from stored metadata JSON
- `DescribeTableVersion`: when enabled and version is specified, queries directly from `__manifest` instead of opening the dataset
- `BatchDeleteTableVersions`: supports two modes:
  - **Single-table (legacy)**: uses `request.id` + `request.ranges` for one table
  - **Multi-table (transactional)**: uses `request.entries` to atomically delete versions across multiple tables (metadata deletion is atomic; physical file cleanup is best-effort)
- `BatchCommitTables`: new unified operation that atomically commits mixed table operations (`declare_table`, `create_table_version`, `delete_table_versions`, `deregister_table`) across multiple tables using discriminated union types from the generated `lance-namespace` API spec
- Add comprehensive tests: batch create atomic, conflict detection, manifest requirement validation, single version recording, and more

### namespace.rs
- Add `batch_create_table_versions` trait method with default not-supported implementation
- Add `batch_commit_tables` trait method with default not-supported implementation
- Update `batch_delete_table_versions` doc to document both single-table (legacy) and multi-table (transactional) modes

### rest.rs / rest_adapter.rs
- Add REST client/server endpoints:
  - `POST /v1/table/version/batch_create` — `batch_create_table_versions`
  - `POST /v1/table/:id/version/batch_delete` — `batch_delete_table_versions`
  - `POST /v1/table/batch_commit` — `batch_commit_tables` (unified atomic multi-table commit)

## Depends On
- lance-namespace PR: adds `BatchCommitTables` API spec and generated client code with discriminated union types (`CommitTableOperation`, `CommitTableResult`)

## Testing
- 122 unit tests pass (including new tests for batch create, conflict detection, multi-table delete, and manifest validation)
- All doc-tests pass

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -66,7 +66,7 @@ lance-linalg = { version = "=4.0.0-beta.12", path = "./rust/lance-linalg" }
 lance-namespace = { version = "=4.0.0-beta.12", path = "./rust/lance-namespace" }
 lance-namespace-impls = { version = "=4.0.0-beta.12", path = "./rust/lance-namespace-impls" }
 lance-namespace-datafusion = { version = "=4.0.0-beta.12", path = "./rust/lance-namespace-datafusion" }
-lance-namespace-reqwest-client = "0.6.0"
+lance-namespace-reqwest-client = "0.6.1"
 lance-table = { version = "=4.0.0-beta.12", path = "./rust/lance-table" }
 lance-test-macros = { version = "=4.0.0-beta.12", path = "./rust/lance-test-macros" }
 lance-testing = { version = "=4.0.0-beta.12", path = "./rust/lance-testing" }


XuQianJin-Stars:
hi @jackye1995 Could a new version be released for https://github.com/lance-format/lance-namespace to support the interfaces related to https://github.com/lance-format/lance-namespace/pull/310 and https://github.com/lance-format/lance-namespace/pull/311? The newly added interface methods in the current 0.5.2 version are all missing.

wojiaodoubao:
Thanks @XuQianJin-Stars for your great work!!! I'd like to help review this pr! I will start after the namespace issue is fixed, feel free to ping me if I miss the progress.

You can also fix the fmt/clippy issue when you have time.

rust/lance-namespace-impls/src/dir.rs
```diff
@@ -1692,6 +1794,33 @@ impl LanceNamespace for DirectoryNamespace {
             );
         }
 
+        // If table_version_storage_enabled is enabled, also record in __manifest
+        if self.table_version_storage_enabled {
```
jackye1995: This should be a best effort, okay to fail

rust/lance-namespace-impls/src/dir.rs
```diff
@@ -455,6 +486,13 @@ impl DirectoryNamespaceBuilder {
     /// - Connection to the storage backend fails
     /// - Storage options are invalid
     pub async fn build(self) -> Result<DirectoryNamespace> {
+        // Validate: table_version_storage_enabled requires manifest_enabled
+        if self.table_version_storage_enabled && !self.manifest_enabled {
```
jackye1995, line 487: When enabled, we should ensure the __manifest table has a property persistedin it's metadata. This is because once we enable this, we should start to reject all writes that are not us the storage to ensure consistency. (Maybe time to develop the same feature flag system like the table format for forward compatibility)

rust/lance-namespace-impls/src/dir.rs
```diff
@@ -1522,6 +1564,66 @@ impl LanceNamespace for DirectoryNamespace {
         &self,
         request: ListTableVersionsRequest,
     ) -> Result<ListTableVersionsResponse> {
+        // When table_version_storage_enabled, query from __manifest
+        if self.table_version_storage_enabled {
+            if let Some(ref manifest_ns) = self.manifest_ns {
```
jackye1995: All these table version specific logic should go to manifest.rs since it requires manifest. We can call into it here.

rust/lance-namespace-impls/src/dir.rs
jackye1995: Let's add the boolean flag for now, and we can do a more formal feature flag system in a separated PR to make the experience better, and also introduce a similar feature flag for manifest_enabled

rust/lance-namespace-impls/src/dir/manifest.rs
```diff
@@ -392,7 +395,7 @@ impl ManifestNamespace {
     }
 
     /// Convert a table ID (vec of strings) to an object_id string
-    fn str_object_id(table_id: &[String]) -> String {
+    pub fn str_object_id(table_id: &[String]) -> String {
```
jackye1995: We should just rename the table_id variable to object_id, and we can use that also for cases like 

let object_id = format!("{}${}", table_id, version);

below

rust/lance-namespace-impls/src/dir/manifest.rs
```diff
@@ -934,6 +937,404 @@ impl ManifestNamespace {
...
+        &self,
+        table_id: &str,
+        descending: bool,
+        limit: Option<i32>,
+    ) -> Result<Vec<(i64, String)>> {
+        let escaped_id = table_id.replace('\'', "''");
+        // table_version object_ids are formatted as "{table_id}${version}"
+        let filter = format!(
+            "object_type = 'table_version' AND starts_with(object_id, '{}$')",
+            escaped_id
```
jackye1995: I think we should store the version number in lexigraphical way, like 00000000001 for version 1, so the versions are ordered and we can do range queries more easily

rust/lance-namespace-impls/src/dir/manifest.rs
```diff
@@ -934,6 +937,404 @@ impl ManifestNamespace {
...
+        let mut merge_builder =
+            lance::dataset::MergeInsertBuilder::try_new(dataset_arc, vec!["object_id".to_string()])
+                .map_err(|e| {
+                    Error::io_source(box_error(std::io::Error::other(format!(
+                        "Failed to create merge builder: {}",
+                        e
+                    ))))
+                })?;
+
+        merge_builder.when_matched(lance::dataset::WhenMatched::Fail);
```
jackye1995: nit: we should try to import things at top of module when possible ( I'm not doing a good job with that either 😆)

rust/lance-namespace-impls/src/dir/manifest.rs
```diff
@@ -934,6 +971,433 @@ impl ManifestNamespace {
...
+        Ok(())
+    }
+
+    /// Query the manifest for all versions of a table, sorted by version.
+    ///
+    /// Returns a list of (version, metadata_json_string) tuples where metadata_json_string
+    /// contains the full metadata JSON stored in the manifest (manifest_path, manifest_size,
+    /// e_tag, naming_scheme).
+    ///
+    /// **Known limitation**: All matching rows are loaded into memory, sorted in Rust,
```
jackye1995, line 1028: okay sorry I missed that we need descending. Looks like we probably need to do the reverse trick like what we do for the manifest file name, but using 9999999999999 instead of long max here, so that it's ordered first in the string form.

Because Lance scan is ordered, by doing this, we should always be able to just do LIMIT 1 to get the latest table version.

jackye1995: oh actually, nvm it's ordered by the row address, not by the actual value. Then the current way should be fine for now, it should also work if there is a btree, but might require some additional optimization in the query planning that we can do later.

rust/lance-namespace-impls/src/dir/manifest.rs
```diff
@@ -934,6 +971,433 @@ impl ManifestNamespace {
...
+    ) -> Result<Option<String>> {
+        // Try new zero-padded format first
+        let version_object_id = Self::build_version_object_id(object_id, version);
+        if let Some(metadata) = self
+            .query_table_version_by_object_id(&version_object_id)
+            .await?
+        {
+            return Ok(Some(metadata));
+        }
+        // Fall back to legacy plain integer format
```
jackye1995: no need to fallback

rust/lance-namespace-impls/src/dir/manifest.rs
```diff
@@ -62,20 +67,28 @@ const BASE_OBJECTS_INDEX_NAME: &str = "base_objects_label_list";
 pub enum ObjectType {
     Namespace,
     Table,
+    TableVersion,
+    /// A property entry used to persist feature flags and configuration
+    /// in the __manifest table (e.g., `table_version_storage_enabled`).
+    Property,
```
jackye1995: we don't need this, when I say property, I am saying you update the __manifest table's table metadata key-value map, which can be done with `dataset.update_metadata()`

rust/lance-namespace-impls/src/dir/manifest.rs
```diff
@@ -379,21 +392,51 @@ impl ManifestNamespace {
...
+    /// Convert an ID (vec of strings) to an object_id string
+    pub fn str_object_id(object_id: &[String]) -> String {
+        object_id.join(DELIMITER)
+    }
+
+    /// Format a version number as a zero-padded lexicographically sortable string.
+    ///
+    /// Versions are stored as 20-digit zero-padded integers (e.g., `00000000000000000001`
+    /// for version 1) so that string-based range queries and sorting work correctly.
+    pub fn format_version(version: i64) -> String {
```
jackye1995: nit: should be named format_table_version

rust/lance-namespace-impls/src/dir/manifest.rs
```diff
@@ -379,21 +387,51 @@ impl ManifestNamespace {
...
+            DELIMITER,
+            Self::format_table_version(version)
+        )
+    }
+
+    /// Parse a version number from the version suffix of a table version object_id.
+    ///
+    /// The object_id is formatted as `{table_id}${zero_padded_version}`.
+    pub fn parse_version_from_object_id(object_id: &str) -> Option<i64> {
+        object_id
```
jackye1995: this can use `split_object_id` and then parse the last part.

rust/lance-namespace-impls/src/dir.rs
```diff
@@ -1752,45 +1869,676 @@ impl LanceNamespace for DirectoryNamespace {
...
+                }
+
+                return Ok(BatchDeleteTableVersionsResponse {
+                    deleted_count: Some(total_deleted_count),
+                    transaction_id: None,
+                });
+            }
+        }
+
+        // Legacy behavior: delete physical files directly (no __manifest)
```
jackye1995: nit: this is not really legacy, it's just when the feature is not enabled.

rust/lance-namespace-impls/src/dir.rs
```diff
@@ -1752,45 +1869,676 @@ impl LanceNamespace for DirectoryNamespace {
...
+                    total_deleted_count = manifest_ns
+                        .batch_delete_table_versions_by_object_ids(&all_object_ids)
+                        .await?;
+                }
+
+                // Phase 2: Delete physical manifest files (best-effort).
+                // Even if some file deletions fail, the versions are already removed from
+                // __manifest, so they won't be visible to readers. Leftover files are
+                // orphaned but harmless and can be cleaned up later.
+                for te in &table_entries {
```
jackye1995: this should be extracted as a helper method since it's basically used twice here and below.

rust/lance-namespace-impls/src/dir.rs
```diff
@@ -1752,45 +1869,676 @@ impl LanceNamespace for DirectoryNamespace {
...
+            }
+        }
+
+        Ok(BatchDeleteTableVersionsResponse {
+            deleted_count: Some(total_deleted_count),
+            transaction_id: None,
+        })
+    }
+
+    async fn batch_create_table_versions(
```
jackye1995: TODO: pending merging the namespace PR to get this updated. But I think we can also just not implement this and merge the remaining, have a separated PR to actually get that implemented, if that is cleaner.

rust/lance-namespace-impls/src/dir.rs
```diff
@@ -1752,45 +1869,676 @@ impl LanceNamespace for DirectoryNamespace {
...
+                ));
             }
         }
 
-        Ok(BatchDeleteTableVersionsResponse {
-            deleted_count: Some(deleted_count),
+        // Phase 1 (atomic commit point): Execute inserts and deletes atomically.
+        // Inserts use MergeInsert (WhenMatched::Fail), deletes use DeleteBuilder.
+        // We execute inserts first, then deletes. Both are individual atomic ops on __manifest.
+        if !insert_entries.is_empty() {
```
jackye1995: if we do them separately, they are not atomic. I think there are 2 ways out: (1) support merge_insert with `WhenMatched::InsertOrDelete(condition)`, so that we can actually choose to insert or delete based on another condition. (2) do a batch commit of 2 staged transactions. I think we should probably do 1, it's a long-overdue feature in merge_insert.

rust/lance-namespace-impls/src/dir/manifest.rs
```diff
@@ -787,16 +825,14 @@ impl ManifestNamespace {
     }
 
     /// Insert an entry into the manifest table with metadata and base_objects
-    async fn insert_into_manifest_with_metadata(
+    pub async fn insert_into_manifest_with_metadata(
```
jackye1995, line 859: I think we can do some refactoring, instead of having another  `batch_insert_into_manifest`, make this support inserting a list of objects (object_id + object_type)

rust/lance-namespace-impls/src/dir/manifest.rs
```diff
@@ -934,6 +971,536 @@ impl ManifestNamespace {
...
     }
 
+    /// Batch insert multiple entries into the manifest table atomically.
+    ///
+    /// Uses a single MergeInsert operation to insert all entries at once.
+    /// If any entry already exists (matching object_id), the entire batch fails.
+    /// This provides atomic multi-table version creation semantics.
+    pub async fn batch_insert_into_manifest(
+        &self,
+        entries: Vec<(String, ObjectType, Option<String>, Option<String>)>,
```
jackye1995: can make this a struct to make the meaning of each entry more clear

rust/lance-namespace-impls/src/dir.rs
```diff
@@ -485,6 +524,22 @@ impl DirectoryNamespaceBuilder {
             None
         };
 
+        // Persist table_version_storage_enabled flag in __manifest so that once
+        // enabled, it becomes a permanent property of this namespace.
+        if self.table_version_storage_enabled {
```
jackye1995: this logic should be moved to `manifest::ensure_manifest_table_up_to_date` so we always have a central place to track all migration related functionalities.

jackye1995 (commented):
Overall looks great! I would suggest we get all features except for the `batch_create_table_versions`/`batch_commit_tables` in first, and then we get a few subsequent items:
1. add support for atomically insert and delete entries through merge_insert
2. add `batch_commit_tables` officially with updated lance-namespace crate version
3. add some more official feature flag setup like the table format

PR #6174, opened 2026-03-12
perf(inverted)!: reduce fts indexing time and memory

BubbleCal:
image

- FTS indexing is ~2.5x faster, this removes merge phase, and produces large partitions directly.
- memory footprint is reduced by ~60%, this compresses posting lists while building them, which can save a lot of memory, and reduces fragmented objects in memory.

This also bumps the default worker memory budget from 256MiB to 1GiB because we need to produce larger partition directly, but the memory footprint is still much less.

This adds a new param `memory_limit` so that users can control how the indexing should work

### DIFF
--- a/python/python/lance/dataset.py
+++ b/python/python/lance/dataset.py
@@ -2608,10 +2608,10 @@ class LanceDataset(pa.dataset.Dataset):
             True.
         memory_limit: int, optional
             This is for the ``INVERTED`` index. Total build-time memory limit in MiB.
-            If set, Lance divides this budget evenly across the workers. If unset, 
+            If set, Lance divides this budget evenly across the workers. If unset,
             the default will be 2 GiB per worker. This parameter is only used for the
             current build and is not persisted with the index.
-            
+
             A larger memory limit will create an index with fewer shards which will
             be easier to search so this is a trade-off between build resources and
             search cost.
--- a/rust/lance/src/dataset/cleanup.rs
+++ b/rust/lance/src/dataset/cleanup.rs
@@ -2867,7 +2867,7 @@ mod tests {
         assert_eq!(setup.branch1.counts.num_data_files, 2);
         assert_eq!(setup.branch1.counts.num_tx_files, 1);
         assert_eq!(setup.branch1.counts.num_delete_files, 2);
-        assert_eq!(setup.branch1.counts.num_index_files, 8);
+        assert_eq!(setup.branch1.counts.num_index_files, 23);
         setup.assert_all_unchanged().await;
 
         setup.branch1.compact().await.unwrap();
@@ -2882,7 +2882,7 @@ mod tests {
         assert_eq!(setup.branch1.counts.num_data_files, 2);
         assert_eq!(setup.branch1.counts.num_tx_files, 1);
         assert_eq!(setup.branch1.counts.num_delete_files, 1);
-        assert_eq!(setup.branch1.counts.num_index_files, 8);
+        assert_eq!(setup.branch1.counts.num_index_files, 23);
         setup.assert_all_unchanged().await;
 
         // Now we clean the referenced files of branch1 by branch2 and branch3
@@ -2896,14 +2896,14 @@ mod tests {
         assert_eq!(setup.branch2.counts.num_data_files, 1);
         assert_eq!(setup.branch2.counts.num_tx_files, 1);
         assert_eq!(setup.branch2.counts.num_delete_files, 0);
-        assert_eq!(setup.branch2.counts.num_index_files, 4);
+        assert_eq!(setup.branch2.counts.num_index_files, 13);
         // Only the latest manifest is retained.
         // (1, 1, 1, 0, 4) is the counts for the latest version of compaction
         assert_eq!(setup.branch3.counts.num_manifest_files, 1);
         assert_eq!(setup.branch3.counts.num_data_files, 1);
         assert_eq!(setup.branch3.counts.num_tx_files, 1);
         assert_eq!(setup.branch3.counts.num_delete_files, 0);
-        assert_eq!(setup.branch3.counts.num_index_files, 4);
+        assert_eq!(setup.branch3.counts.num_index_files, 16);
         setup.branch1.run_cleanup().await.unwrap();
 
         // Only the latest manifest is retained.
@@ -2912,7 +2912,7 @@ mod tests {
         assert_eq!(setup.branch1.counts.num_data_files, 1);
         assert_eq!(setup.branch1.counts.num_tx_files, 1);
         assert_eq!(setup.branch1.counts.num_delete_files, 0);
-        assert_eq!(setup.branch1.counts.num_index_files, 4);
+        assert_eq!(setup.branch1.counts.num_index_files, 13);
         setup.assert_all_unchanged().await;
     }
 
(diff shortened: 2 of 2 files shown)


rust/lance-index/src/scalar/inverted/builder.rs
```diff
@@ -58,17 +54,101 @@ pub static LANCE_FTS_NUM_SHARDS: LazyLock<usize> = LazyLock::new(|| {
...
+        .unwrap_or(false)
+});
+#[cfg(target_os = "linux")]
+static LANCE_FTS_TRIM_ALLOC: LazyLock<bool> = LazyLock::new(|| {
+    std::env::var("LANCE_FTS_TRIM_ALLOC")
+        .map(|value| value != "0")
+        .unwrap_or(false)
+});
+
+fn current_rss_bytes() -> Option<u64> {
```
Xuanwo: There is a lot of code related to RSS measurement and debug logging. I am considering whether it would be better to link an external allocator and use its metrics instead. For example:

- https://github.com/jemalloc/jemalloc/wiki/Use-Case%3A-Basic-Allocator-Statistics
- https://microsoft.github.io/mimalloc/group__analysis.html

BubbleCal:
indexing can be faster as adding more workers but it hits a wall at `num_workers=num_cpus/2`, adding more workers doesn't help much after that, but that produces more partitions (higher query latency) and higher memory footprint.

benchmark on 16 cores machine:
image

image

benchmark on 32 cores machine:

image

image


we should set the default to be using half cores

rust/lance-index/src/scalar/inverted/encoding.rs
```diff
@@ -28,6 +28,7 @@ use lance_core::Result;
 
 // compress the posting list to multiple blocks of fixed number of elements (BLOCK_SIZE),
 // returns a LargeBinaryArray, where each binary is a compressed block (128 row ids + 128 frequencies)
+#[allow(dead_code)]
```
westonpace: Can we get rid of dead code instead of just annotating it?

rust/lance-index/src/scalar/inverted/builder.rs
```diff
@@ -354,84 +371,50 @@ impl InvertedIndexBuilder {
     }
 
     async fn write(&self, dest_store: &dyn IndexStore) -> Result<()> {
-        if self.params.skip_merge {
```
westonpace, line 357: Are we always skipping the merge now?  Or is it still conditional?

If we are always skipping it can we get rid of the flag?

If it is still conditional, then how should a user decide if that is something they want to do?

BubbleCal: we always skip, the merge phase doesn't exist any more.

just deleted it, it's safe as we have `default` before this

rust/lance-index/src/scalar/inverted/encoding.rs
BubbleCal: changed it to `#[cfg(test)]`, it's still used for testing, removed the other dead code

python/python/lance/dataset.py
```diff
@@ -2606,10 +2606,19 @@ def create_scalar_index(
...
-            This is for the ``INVERTED`` index. If True, the index will skip the
-            partition merge stage after indexing. This can be useful for
-            distributed/fragment-level indexing where a later merge is desired.
+        memory_limit: int, optional
+            This is for the ``INVERTED`` index. Total build-time memory limit in MiB.
+            If set, Lance divides this budget evenly across FTS workers. By default
+            Lance uses roughly ``num_cpus / 2`` workers unless
+            ``LANCE_FTS_NUM_SHARDS`` is set. If unset, each worker defaults to a
+            2 GiB build-time memory limit. This parameter is only used for the
+            current build and is not persisted with the index.
```
westonpace, line 2613:
```suggestion
        memory_limit: int, optional
            This is for the ``INVERTED`` index. Total build-time memory limit in MiB.
            If set, Lance divides this budget evenly across the workers. If unset, 
            the default will be 2 GiB per worker. This parameter is only used for the
            current build and is not persisted with the index.
            
            A larger memory limit will create an index with fewer shards which will
            be easier to search so this is a trade-off between build resources and
            search cost.
```

* Default is 2GiB _per worker_ (the original was vague)
* Don't mention num_workers environment variable since it's defined below
* Explain the tradeoff

python/python/lance/dataset.py
```diff
@@ -2606,10 +2606,19 @@ def create_scalar_index(
...
+        num_workers: int, optional
+            This is for the ``INVERTED`` index. Number of FTS workers to use for
+            the current build. The effective worker count is clamped to
+            ``[1, num_cpus - 2]``. If unset, Lance uses roughly ``num_cpus / 2``
+            workers unless ``LANCE_FTS_NUM_SHARDS`` is set. This parameter is
+            only used for the current build and is not persisted with the index.
```
westonpace, line 2623:
```suggestion
        num_workers: int, optional
            This is for the ``INVERTED`` index. Number of workers to use for
            the current build. The effective worker count is clamped to
            ``[1, num_compute_cpus]``. If unset, Lance uses ``num_compute_cpus``
            workers unless ``LANCE_FTS_NUM_SHARDS`` is set. This parameter is
            only used for the current build and is not persisted with the index.
```

rust/lance-index/src/scalar/inverted/builder.rs
```diff
@@ -205,48 +260,79 @@ impl InvertedIndexBuilder {
...
+            log::info!("indexing FTS with {} workers", num_workers);
+
+            let mut last_num_rows = 0;
+            let mut total_num_rows = 0;
+            let start = std::time::Instant::now();
+            while let Some(batch) = stream.try_next().await? {
+                let num_rows = batch.num_rows();
+
+                if sender.send(batch).await.is_err() {
+                    // this only happens if all workers have existed,
```
westonpace:
```suggestion
                    // this only happens if all workers have exited,
```

rust/lance-index/src/scalar/inverted/builder.rs
```diff
@@ -44,42 +42,100 @@ use tracing::instrument;
...
+
+fn resolve_worker_memory_limit_bytes(params: &InvertedIndexParams, num_workers: usize) -> u64 {
+    let default_worker_memory_limit_bytes = *LANCE_FTS_PARTITION_SIZE << 20;
+    params
+        .memory_limit_mb
+        .map(|memory_limit_mb| (memory_limit_mb << 20) / num_workers as u64)
+        .unwrap_or(default_worker_memory_limit_bytes)
+}
+
+fn group_tail_partitions(
```
westonpace, line 89: What advantage is there to grouping these by the memory budget?  They are already in memory and we aren't getting any new data.  Wouldn't it always be better to merge them all into one partition?

Or would that require doubling the RAM to account for the merged partition existing alongside the inputs?

rust/lance-index/src/scalar/inverted/query.rs
```diff
@@ -625,15 +625,15 @@ impl BooleanQuery {
 }
 
 #[derive(Debug, Clone, PartialEq)]
-#[allow(dead_code)]
+#[cfg(test)]
```
westonpace, line 628: Why is this becoming test code?  If it's not used in production shouldn't we remove the code _and_ the tests?

westonpace (approved):
Great improvement!  I have a few thoughts/suggestions but overall I think this looks good.

rust/lance-index/src/scalar/inverted/builder.rs
BubbleCal: changed it to be merging all tail partitions into single partition

rust/lance-index/src/scalar/inverted/query.rs
BubbleCal: it's a helper function for constructing compressed posting lists for testing WAND algo

PR #6176, opened 2026-03-12
perf: add benchmark for distributed vector merge finalization

Xuanwo:
This adds a dedicated benchmark for distributed vector index finalization and a small query-side metric to count `find_partitions` calls. Together they give us a baseline for analyzing the current single-node merge bottleneck and for evaluating future segmented-index work.

As context, the new `distributed_merge_only_ivf_pq` benchmark already shows that finalize cost grows much faster than input bytes as shard count and partition count increase. In the local filesystem benchmark, the mean finalize time grows from about `64 ms` at `8 shards / 256 partitions` to about `2.87 s` at `128 shards / 1024 partitions`.

---

Based on this benchmark, I noticed that our current logic performs poorly as the number of shards increases.

2026-03-12-distributed-merge-trend-matplotlib

### DIFF
--- a/rust/lance/benches/distributed_vector_build.rs
+++ b/rust/lance/benches/distributed_vector_build.rs
@@ -15,7 +15,7 @@ use serde::Serialize;
 use uuid::Uuid;
 
 use lance::dataset::{Dataset, WriteMode, WriteParams};
-use lance::index::vector::{VectorIndexParams, ivf::finalize_distributed_merge};
+use lance::index::vector::VectorIndexParams;
 use lance_arrow::FixedSizeListArrayExt;
 use lance_index::vector::kmeans::{KMeansParams, train_kmeans};
 use lance_index::{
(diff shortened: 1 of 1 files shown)


rust/lance/benches/distributed_vector_build.rs
```diff
@@ -0,0 +1,451 @@
...
+    };
+
+    let dataset = Dataset::write(reader, &uri, Some(write_params))
+        .await
+        .unwrap();
+    assert_eq!(dataset.get_fragments().len(), NUM_FRAGMENTS);
+    dataset
+}
+
+async fn train_shared_ivf_pq(
```
BubbleCal, line 200: believe we have such helper function already, maybe just need to change its visibility

rust/lance/src/io/exec/knn.rs
```diff
@@ -77,6 +78,7 @@ impl AnnPartitionMetrics {
             index_metrics: IndexMetrics::new(metrics, partition),
             partitions_ranked: metrics.new_count(PARTITIONS_RANKED_METRIC, partition),
             deltas_searched: metrics.new_count(DELTAS_SEARCHED_METRIC, partition),
+            find_partitions_calls: metrics.new_count(FIND_PARTITIONS_CALLS_METRIC, partition),
```
BubbleCal: how about changing this to `FIND_PARTITIONS_ELAPSED_METRIC`?
we may introduce some optimizations for `find_partitions` in the future, the number of calls can't prove that

PR #6186, opened 2026-03-13
refactor!: cleanup namespace related APIs

jackye1995:
1. Remove `storage_options_provider` in python and java, because to make managed verisoning work, we have updated the codebase to pass namespace and table ID into the python and java binding layer. It becomes unnecessary for us to do a language specific `storage_options_provider` and then bind that to rust, because we can directly construct the rust `StorageOptionsProvider` using binded namespace client.
2. rename the following: `namespace` to `namespace_client`, `namespace_impl` to `namespace_client_impl`, `namespace_properties` to `namespace_client_properties`, `namespace` which means the namespace path to `namespace_path`. This is done for all code in rust, python, java. This rename is based on community feedback, and aims at clarifying the concept of Namespace Client SDK and its implementations vs the namespace path like `["ns1", "ns2"]`.
3. add `vend_input_storage_options` and `ops_metrics_enabled` so that we can now use DirectoryNamespace directory for testing all these changes made, without the need to rely on an extra tracking namespace. Update all tests accordingly to use the new feature.
4. fix the known bug that python and java binding for non-native namespace client implementation is not fully working with managed versioning due to binding level model conversion.

### DIFF
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -63,7 +63,7 @@ jobs:
           prefix-key: ${{ env.CACHE_PREFIX }}
       - name: Install linting tools
         run: |
-          pip install ruff==0.11.2 maturin tensorflow tqdm ray[data] pyright datasets polars[pyarrow,pandas]
+          pip install ruff==0.11.2 maturin tensorflow tqdm ray[data] pyright datasets polars[pyarrow,pandas] lance-namespace
           pip install torch --index-url https://download.pytorch.org/whl/cpu
       - name: Lint Python
         run: |


python/python/lance/__init__.py
```diff
@@ -97,9 +97,11 @@ def dataset(
     index_cache_size_bytes: Optional[int] = None,
     read_params: Optional[Dict[str, any]] = None,
     session: Optional[Session] = None,
-    namespace: Optional[LanceNamespace] = None,
+    namespace_client: Optional[LanceNamespace] = None,
     table_id: Optional[List[str]] = None,
     storage_options_provider: Optional[Any] = None,
+    # Deprecated parameters
+    namespace: Optional[LanceNamespace] = None,
```
westonpace: Humorous observation: technically moving arguments without a `*` in the argument list is a breaking change but if anyone was using this argument as a positional argument at this point then I worry

python/python/lance/__init__.py
```diff
@@ -97,9 +97,11 @@ def dataset(
     index_cache_size_bytes: Optional[int] = None,
     read_params: Optional[Dict[str, any]] = None,
     session: Optional[Session] = None,
-    namespace: Optional[LanceNamespace] = None,
+    namespace_client: Optional[LanceNamespace] = None,
     table_id: Optional[List[str]] = None,
     storage_options_provider: Optional[Any] = None,
+    # Deprecated parameters
```
westonpace:
```suggestion
    # Deprecated parameters
storage_options_provider: Optional[Any] = None,
```

python/python/lance/dataset.py
```diff
@@ -431,14 +431,38 @@ def __init__(
         read_params: Optional[Dict[str, Any]] = None,
         session: Optional[Session] = None,
         storage_options_provider: Optional[Any] = None,
```
westonpace: Move this to deprecated parameters section?

python/python/lance/dataset.py
```diff
@@ -3665,6 +3705,17 @@ def commit_batch(
...
+        # Emit deprecation warning for storage_options_provider
+        if storage_options_provider is not None:
+            warnings.warn(
+                "The 'storage_options_provider' parameter is deprecated and will be "
+                "removed in version 5.0.0. When using 'namespace_client' and "
+                "'table_id', the storage options provider is created automatically. "
+                "Pass 'storage_options' instead for static or initial storage options.",
+                DeprecationWarning,
+                stacklevel=2,
+            )
```
westonpace: This warning is repeated all over the place.  In a pre-AI world I'd probably say we should make a helper method.  Now I'll just say it might be a good idea.

python/python/lance/fragment.py
```diff
@@ -1002,6 +1016,15 @@ def write_fragments(
 
         **Only valid in CREATE mode**. Will raise an error if used with
         APPEND/OVERWRITE modes.
+    namespace_client : optional, LanceNamespace
```
westonpace, line 1028: Do we document these things in the site docs anywhere?  I don't see any changes in this PR to those.

jackye1995:
I ended up doing all the backwards incompatible changes in both python and java, because to keep backwards compatibility it becomes too complicated, it is just not worth doing it just to keep it around for a single version.

PR #6188, opened 2026-03-13
fix: like queries with a prefix should be accelerated by btree and zonemap

jackye1995:
This allows pruning `LIKE "foo%"` through any index that allows pruning a range of string prefixes, this includes btree and zonemap.

I ended up adding a `LikePrefix` instead of `StartsWith` because it seems like DataFusion converts `StartsWith` to `LIKE`


jackye1995:
looks like I did not handle AND properly, will take another pass tomorrow

jackye1995:
cargo-deny error unrelated. Will go ahead to merge this

PR #6194, opened 2026-03-13
fix: disallowing stale credentials from directory namespace

hamersaw:
This PR makes two changes to ensure stale credentials are not used:
(1) In the Directory namespace if either vending is not enabled or a credential vendor is not configured we return `None` for storage options.
(2) The `DynamicStorageOptionsCredentialProvider` falls back to the default credential provider (lazily loaded) if it is not able to retrieve credentials.

Closes https://github.com/lance-format/lance-spark/issues/292

### DIFF
--- a/rust/lance-io/src/object_store/providers/aws.rs
+++ b/rust/lance-io/src/object_store/providers/aws.rs
@@ -260,7 +260,9 @@ pub async fn build_aws_credential(
 
     let storage_options_credentials = storage_options.and_then(extract_static_s3_credentials);
 
-    // If accessor has a provider, use DynamicStorageOptionsCredentialProvider
+    // If accessor has a provider, check whether it vends credentials.
+    // If it does, use DynamicStorageOptionsCredentialProvider for ongoing
+    // refresh. If not, fall through to the default credentials chain.
     if let Some(accessor) = storage_options_accessor
         && accessor.has_provider()
     {
@@ -268,11 +270,21 @@ pub async fn build_aws_credential(
         if let Some(creds) = credentials {
             return Ok((creds, region));
         }
-        // Use accessor for dynamic credential refresh
-        return Ok((
-            Arc::new(DynamicStorageOptionsCredentialProvider::new(accessor)),
-            region,
-        ));
+
+        // Check if the accessor's storage options contain credentials
+        let opts = accessor.get_storage_options().await?;
+        let s3_options = opts.as_s3_options();
+        if extract_static_s3_credentials(&s3_options).is_some() {
+            return Ok((
+                Arc::new(DynamicStorageOptionsCredentialProvider::new(accessor)),
+                region,
+            ));
+        }
+
+        log::debug!(
+            "Storage options from provider do not contain explicit AWS credentials, \
+             falling back to default AWS credentials chain."
+        );
     }
 
     // Fall back to existing logic for static credentials
@@ -447,10 +459,6 @@ impl ObjectStoreParams {
 /// AWS-specific credentials that can be used with S3. All caching and refresh logic
 /// is handled by the accessor.
 ///
-/// When the accessor returns storage options that do not contain explicit
-/// credentials, this provider falls back to the default AWS credentials
-/// chain (environment variables, instance profile, etc.).
-///
 /// # Future Work
 ///
 /// TODO: Support AWS/GCP/Azure together in a unified credential provider.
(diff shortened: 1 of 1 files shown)


rust/lance-namespace-impls/src/dir.rs
```diff
@@ -1369,16 +1359,12 @@ impl LanceNamespace for DirectoryNamespace {
...
-                if let Some(ref location) = response.location {
-                    let identity = request.identity.as_deref();
-                    response.storage_options = self
-                        .get_storage_options_for_table(location, identity)
-                        .await?;
-                }
-            } else if request.vend_credentials == Some(false) {
-                response.storage_options = None;
+            if let Some(ref location) = response.location {
+                let vend = request.vend_credentials.unwrap_or(true);
```
hamersaw, line 1365: This doesn't seem right to me, but there is precedent to default to true for backwards compatibility - https://github.com/hamersaw/lance/blob/bug%2Fcredentials/rust/lance-namespace-impls/src/dir.rs#L1112-L1113.

rust/lance-namespace-impls/src/dir.rs
```diff
@@ -819,13 +859,17 @@ impl DirectoryNamespace {
...
+        vend_credentials: bool,
         identity: Option<&Identity>,
     ) -> Result<Option<HashMap<String, String>>> {
-        if let Some(ref vendor) = self.credential_vendor {
+        if vend_credentials && let Some(ref vendor) = self.credential_vendor {
             let vended = vendor.vend_credentials(table_uri, identity).await?;
             return Ok(Some(vended.storage_options));
         }
-        Ok(self.storage_options.clone())
+        // Return non-credential storage options (endpoint, region, etc.)
```
jackye1995: I didn't fully understand "so clients can locate the object store", why not just not return anything at all?

rust/lance-namespace-impls/src/dir.rs
```diff
@@ -819,13 +859,17 @@ impl DirectoryNamespace {
...
+        // so clients can locate the object store, but strip credential keys
+        // to prevent leaking stale or static credentials.
+        Ok(strip_credential_options(&self.storage_options))
```
hamersaw: Returning `None` here caused multiple issues because other options are potentially used. So only stripping authentication-related options. I'm not confident that this is correct - please advise.

python/python/tests/test_namespace_integration.py
```diff
@@ -867,12 +871,11 @@ def test_basic_create_and_drop_on_s3(s3_bucket: str):
     table_data = create_test_table_data()
     table_id = ["test_ns", table_name]
 
-    # Create table using lance.write_dataset (same as other passing tests)
```
jackye1995, line 870: we should not change the test here, was there anything failing when writing in this way? We should probably fix that

rust/lance-io/src/object_store/providers/aws.rs
```diff
@@ -447,6 +447,11 @@ impl ObjectStoreParams {
 /// AWS-specific credentials that can be used with S3. All caching and refresh logic
 /// is handled by the accessor.
 ///
+/// When the accessor returns storage options that do not contain explicit
+/// credentials (e.g. because the namespace layer stripped credential keys),
```
jackye1995: "(e.g. because the namespace layer stripped credential keys)" this is no longer true?

rust/lance-io/src/object_store/providers/aws.rs
```diff
@@ -521,20 +551,39 @@ impl CredentialProvider for DynamicStorageOptionsCredentialProvider {
...
+
+        // Fall back to the default AWS credentials chain when storage options
+        // do not contain explicit credentials (e.g. credential keys may have
+        // been stripped by the namespace layer).
+        log::debug!(
+            "Storage options from provider do not contain explicit AWS credentials, \
+             falling back to default AWS credentials chain."
+        );
+        let chain = self.get_default_chain().await;
+        let creds =
```
jackye1995: oh actually, I think there is a problem with this. If we do it this way, we are just getting the credentials at this moment, which will expire some time later, but we are not really gonna refresh it again since the default credentials chain don't really tell you when it expires. I think the right way is to just continue to fail when extract_static_s3_credentials fails, and then at outer caller side we directly fallback to use the DefaultCredentialsChain provider.

jackye1995: Oh nvm I think this could also work, since every time we do get_credential, it will eventually call into this and get a new credential.

In that case, I think before we check storage_options, we should first check the chain, so that if the chain is set we directly return its credentials. If not set, we go through this logic. Because otherwise even after we set the chain, we will still have to do a storage_options check every time which is wasteful.

hamersaw: Even if we moved this to the outer caller, we would want to add the optimization to check if the `default_chain` exists and just use that right?

jackye1995: if we move to the outer, we won't really check back right? Because we have initialized a provider, not individual credentials.

hamersaw: I like this a lot more. Making the update.

erandagan:
I believe the documentation is now out of date here - https://github.com/lance-format/lance/blob/d630106da5a238b3adfb8c5dea3b3921f3519945/python/python/lance/__init__.py#L164 - since storage_options are no longer fetched from the namespace by default. Vending must be configured

PR #6202, opened 2026-03-16
perf: reuse distance calculator at selecting candidates

BubbleCal:
```
HNSW_FLAT create: 41.102s -> 6.401s，-84.4%
HNSW_SQ create: 6.9698s -> 5.3778s，-22.8%
HNSW_PQ create: 41.167s -> 40.098s，-2.6%
```

### DIFF
--- a/rust/lance-index/benches/hnsw.rs
+++ b/rust/lance-index/benches/hnsw.rs
@@ -43,11 +43,7 @@ fn bench_hnsw(c: &mut Criterion) {
     let query = fsl.value(0);
     c.bench_function(format!("create_hnsw({TOTAL}x{DIMENSION})").as_str(), |b| {
         b.to_async(&rt).iter(|| async {
-            let hnsw = HNSW::index_vectors(
-                vectors.as_ref(),
-                HnswBuildParams::default(),
-            )
-            .unwrap();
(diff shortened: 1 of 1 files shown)


rust/lance-index/src/vector/graph.rs
```diff
@@ -161,68 +160,94 @@ pub trait Graph {
...
     fn drop(&mut self) {
-        for node_id in self.recently_visited.iter() {
-            self.visited.set(*node_id as usize, false);
+        for node_id in self.recently_visited.iter().copied() {
+            let node_id_usize = node_id as usize;
+            let word_index = node_id_usize / WORD_BITS;
+            let mask = 1usize << (node_id_usize % WORD_BITS);
+            self.visited[word_index] &= !mask;
         }
         self.recently_visited.clear();
```
Xuanwo, line 222: Maybe we need to resize to a good capacitry here to avoid `recently_visited` always growing

Xuanwo (commented):
Mostly LGTM!

rust/lance-index/src/vector/graph.rs
BubbleCal: the cap is supposed to be around `ef`, but if the query is with filter, it would be hard to predict, and resizing would introduce performance penalty in such case.

it may be fine without resizing because it's up to `num_nodes` u32s, and that's very small in HNSW (<1/20)

PR #6203, opened 2026-03-16
perf: new layout for positions and new algo for phrase query

BubbleCal:
This stores all positions of single posting list in a flat array, and apply compression to the positions:
- for positions of single document, storing delta positions
- divide positions of single posting list into blocks (128 by default) and apply bitpack to each block
- store the remaining positions in varint

This also stores the remaining doc ids / frequencies in varint as well.

These are breaking changes.

on 1M random dataset:
- build is 9.35x faster (34.269s -> 3.664s)
- index is 8.87x smaller (955,002,412 bytes -> 107,603,319 bytes)
- phrase query avg latency is 3.41x faster (21.631ms -> 6.334ms) and p95 latency is 3.17x faster (63.735ms -> 20.100ms)

the improvement can be more significant on larger dataset

### DIFF
--- a/python/python/tests/compat/test_scalar_indices.py
+++ b/python/python/tests/compat/test_scalar_indices.py
@@ -289,7 +289,7 @@ class FtsIndex(UpgradeDowngradeTest):
             }
         )
         dataset = lance.write_dataset(data, self.path, max_rows_per_file=100)
-        dataset.create_scalar_index("text", "INVERTED")
+        dataset.create_scalar_index("text", "INVERTED", with_position=True)
 
     def check_read(self):
         """Verify FTS index can be queried."""


rust/lance-index/src/scalar/inverted/iter.rs
```diff
@@ -87,13 +128,24 @@ impl Iterator for CompressedPostingListIterator {
...
-                Box::new(positions.into_iter()) as _
+        if self.doc_idx_in_block < self.doc_ids.len() {
+            let doc_id = self.doc_ids[self.doc_idx_in_block];
+            let freq = self.frequencies[self.doc_idx_in_block];
+            let positions = self.positions.as_ref().map(|storage| match storage {
+                CompressedPositionStorage::LegacyPerDoc(list) => {
+                    let compressed = list.value(self.idx);
+                    let positions = decompress_positions(compressed.as_binary());
+                    Box::new(positions.into_iter()) as Box<dyn Iterator<Item = u32>>
+                }
```
Xuanwo, line 139: I'm not sure if I understand this correctly. It looks like we have cached the entire `POSITION_COL` in `LegacyPerDoc`:

```rust
PositionsLayout::LegacyPerDoc => {
    let batch = self
        .reader
        .read_range(self.posting_list_range(token_id), Some(&[POSITION_COL]))
        .await
        .map_err(|e| match e {
            Error::Schema { .. } => Error::invalid_input("position is not found but required for phrase queries, try recreating the index with position".to_owned()),
            e => e,
        })?;
    CompressedPositionStorage::LegacyPerDoc(batch[POSITION_COL].as_list::<i32>().clone())
```

Do we still need to do this?

BubbleCal: it caches only the rows hit, not the entire column.

these line read the cached rows and decompress them.

BubbleCal:
hold it because of compatibility issue

LuQQiu:
Looks like   #[serde(rename = "memory_limit", skip_serializing, ...)]
  pub(crate) memory_limit_mb: Option<u64>,
  
  memory limit still not being respected and fallback to 2GiB

BubbleCal:
> Looks like #[serde(rename = "memory_limit", skip_serializing, ...)] pub(crate) memory_limit_mb: Option,
> 
> memory limit still not being respected and fallback to 2GiB

skipping serializing this field is expected, because we don't want to add this into index params, it's should be one-time param for only building.

I just tested this param it works

esteban:
Just leaving for release notes: env var `LANCE_FTS_FORMAT_VERSION` default is `1` and must be set to `2` in order to enable the new format.

PR #6204, opened 2026-03-16
fix: preserve create index transaction semantics

Xuanwo:
This fixes the transaction semantics behind `CreateIndex` so we can precisely add and remove index UUIDs without relying on implicit name-based replacement.

It also closes the rebase correctness hole where concurrent same-name `CreateIndex` commits could survive with stale `removed_indices` and leave overlapping entries in the manifest. This keeps the existing uniqueness and `replace=true` contract intact while unblocking follow-up multi-segment index work.

This PR is part of https://github.com/lance-format/lance/issues/6180

### DIFF
--- a/rust/lance/src/io/commit/conflict_resolver.rs
+++ b/rust/lance/src/io/commit/conflict_resolver.rs
@@ -523,11 +523,9 @@ impl<'a> TransactionRebase<'a> {
                             idx.name != FRAG_REUSE_INDEX_NAME && idx.name != MEM_WAL_INDEX_NAME
                         })
                         .any(|new_index| {
-                            created_indices.iter().any(|created_index| {
-                                created_index.name != FRAG_REUSE_INDEX_NAME
-                                    && created_index.name != MEM_WAL_INDEX_NAME
-                                    && created_index.name == new_index.name
-                            })
+                            created_indices
+                                .iter()
+                                .any(|created_index| created_index.name == new_index.name)
                         });
 
                     if (self_has_frag_reuse && other_has_frag_reuse)


rust/lance/src/index/create.rs
```diff
@@ -435,11 +440,22 @@ impl<'a> CreateIndexBuilder<'a> {
     async fn execute(mut self) -> Result<IndexMetadata> {
         let new_idx = self.execute_uncommitted().await?;
         let index_uuid = new_idx.uuid;
+        let removed_indices = if self.replace {
```
BubbleCal, line 443: does this mean if we need to manually commit a `CreateIndex` transaction, we need to check `replace` everywhere?

this seems a breaking change, are we replying on this somewhere (distributed indexing?)?

rust/lance/src/index/create.rs
```diff
@@ -171,19 +171,24 @@ impl<'a> CreateIndexBuilder<'a> {
...
-                    please specify a different name or use replace=True"
-                )));
-            };
-            if idx.fields != [field.id] {
-                return Err(Error::index(format!(
-                    "Index name '{index_name} already exists with different fields, \
-                    please specify a different name"
-                )));
-            }
+        let existing_named_indices = indices
```
BubbleCal, line 174: does this duplicate https://github.com/lance-format/lance/pull/6204/changes#diff-ed935ef0e993a9a039bc75398d27abe79036df7c41f0c40fa0a9d7325c0ad3a6R443

rust/lance/src/index/create.rs
Xuanwo: Yes. With this change, `Operation::CreateIndex` becomes an explicit add/remove UUID set operation instead of relying on implicit name-based replacement during manifest merge.

That means callers which want replace semantics need to populate `removed_indices` themselves. The additive/manual commit paths (including distributed indexing’s `commit_existing_index(...)`) are not relying on the old behavior. They are registering already-built indices and should continue to use `removed_indices: vec![]`.

rust/lance/src/index/create.rs
Xuanwo: Not exactly. These two checks happen at different stages and serve different purposes.

It's possible to have a helper function to make it more clear, do you want it?

BubbleCal: yeah, the two implementations are diff already, maybe we don't need to check `fields` as we don't allow indices with the same name but diff fields. Good to have a helper function to get consistent behavior

rust/lance/src/index/create.rs
```diff
@@ -171,19 +171,24 @@ impl<'a> CreateIndexBuilder<'a> {
...
+            return Err(Error::index(format!(
+                "Index name '{index_name}' already exists with different fields, \
+                please specify a different name"
+            )));
+        }
+        if !existing_named_indices.is_empty() && !self.replace {
+            return Err(Error::index(format!(
+                "Index name '{index_name}' already exists, \
+                please specify a different name or use replace=True"
+            )));
```
westonpace, line 191: Does this mean I will need `replace=true` to add a uuid to an existing index?

rust/lance/src/io/commit/conflict_resolver.rs
```diff
@@ -518,9 +517,22 @@ impl<'a> TransactionRebase<'a> {
...
                         .any(|idx| idx.name == MEM_WAL_INDEX_NAME);
+                    let has_regular_name_conflict = new_indices
+                        .iter()
+                        .filter(|idx| {
+                            idx.name != FRAG_REUSE_INDEX_NAME && idx.name != MEM_WAL_INDEX_NAME
+                        })
+                        .any(|new_index| {
+                            created_indices.iter().any(|created_index| {
+                                created_index.name != FRAG_REUSE_INDEX_NAME
+                                    && created_index.name != MEM_WAL_INDEX_NAME
```
westonpace: This check seems redundant since we know `new_index.name != FRAG_REUSE_INDEX_NAME && new_index.name != MEM_WAL_INDEX_NAME` from the filter clause above.  This means if `created_index.name == new_index.name` we already know `created_index.name != FRAG_REUSE_INDEX_NAME && created_index.name != MEM_WAL_INDEX_NAME`

rust/lance/src/index/create.rs
Xuanwo: In fact, our current API didn't support adding a UUID to an existing index, it's our next step. We'll need a new public API for that.

rust/lance/src/io/commit/conflict_resolver.rs
Xuanwo: Thank you for catching this, will update.

Xuanwo:
distance::dot::tests::test_dot_f32's failure is not related.

PR #6205, opened 2026-03-16
chore: add single node distributed IVF_PQ benchmark crate

Xuanwo:
This adds a standalone Rust benchmark crate under `benchmarks/distributed_ivf_pq` for measuring local distributed `IVF_PQ` index builds end to end. It creates a fragmented dataset, trains shared IVF/PQ parameters, builds shard-local partial indices concurrently, finalizes the merge, and emits timing JSON.

The crate is intentionally outside the main Rust workspace and includes its own `Cargo.toml`, `Cargo.lock`, README, and module-level docs so it can be run directly without touching `rust/examples` or criterion benches.


benchmarks/distributed_ivf_pq/src/main.rs
```diff
@@ -0,0 +1,537 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright The Lance Authors
+//! Benchmark a local distributed `IVF_PQ` build end to end.
```
westonpace, line 3: Is this a benchmark or an example?

Are you trying to create a new benchmark?  Or are you trying to show people how to create benchmarks with an example?

Xuanwo: I actually want to build a script for my own use, but I don't want to use the existing bench harness. Maybe I can just move into `benchmarks` but not register as a rust bench?

Xuanwo: Moved to `benchmarks` now

benchmarks/distributed_ivf_pq/README.md
```diff
@@ -0,0 +1,36 @@
...
+  --num-partitions 32768 \
+  --num-sub-vectors 16 \
+  --max-iters 10 \
+  --sample-rate 8 \
+  --cleanup
+```
+
+## Notes
+
+- This crate is intentionally outside the main Rust workspace.
```
cmccabe, line 33: Is this comment still accurate now that it is under `benchmarks/` ?

Xuanwo:
Moved to https://github.com/lancedb/lance-bench/pull/3

PR #6208, opened 2026-03-16
docs: add example to show how to index JSON column

prrao87:
Users have asked if it's possible to create an index on a JSON column in Lance. The update shows how to create a scalar and and FTS index on a JSON column, per the code in the test suite.


Xuanwo (approved):
Thank you!

PR #6209, opened 2026-03-17
feat: add index segment commit API

Xuanwo:
This adds a first-class `IndexSegment` type and a `commit_existing_index_segments` API so a logical index can be committed from multiple physical segments. It is intended as a building block for the multi-segment vector index work without changing the high-level `create_index` flow.

Part of https://github.com/lance-format/lance/issues/6180

### DIFF
--- a/rust/lance/src/index.rs
+++ b/rust/lance/src/index.rs
@@ -5208,13 +5208,13 @@ mod tests {
             Uuid::new_v4(),
             std::iter::once(0_u32),
             Arc::new(vector_index_details()),
-            IndexType::Vector.version() as i32,
+            IndexType::Vector.version(),
         );
         let seg1 = IndexSegment::new(
             Uuid::new_v4(),
             std::iter::once(1_u32),
             Arc::new(vector_index_details()),
-            IndexType::Vector.version() as i32,
+            IndexType::Vector.version(),
         );
 
         dataset
@@ -5270,7 +5270,7 @@ mod tests {
             Uuid::new_v4(),
             std::iter::once(0_u32),
             Arc::new(vector_index_details()),
-            IndexType::Vector.version() as i32,
+            IndexType::Vector.version(),
         );
 
         let err = dataset
@@ -5283,7 +5283,7 @@ mod tests {
                         base.uuid(),
                         std::iter::once(1_u32),
                         Arc::new(vector_index_details()),
-                        IndexType::Vector.version() as i32,
+                        IndexType::Vector.version(),
                     ),
                 ],
             )
--- a/rust/lance/src/index/vector/ivf/v2.rs
+++ b/rust/lance/src/index/vector/ivf/v2.rs
@@ -1474,7 +1474,7 @@ mod tests {
                     shared_uuid,
                     dataset.fragment_bitmap.as_ref().clone(),
                     Arc::new(crate::index::vector_index_details()),
-                    IndexType::IvfPq.version() as i32,
+                    IndexType::IvfPq.version(),
                 )],
             )
             .await
--- a/python/src/indices.rs
+++ b/python/src/indices.rs
@@ -426,7 +426,7 @@ async fn do_load_shuffled_vectors(
                 prost_types::Any::from_msg(&lance_table::format::pb::VectorIndexDetails::default())
                     .unwrap(),
             ),
-            IndexType::IvfPq.version() as i32,
+            IndexType::IvfPq.version(),
         )],
     )
     .await


rust/lance-index/src/types.rs
```diff
@@ -0,0 +1,84 @@
...
+use roaring::RoaringBitmap;
+use uuid::Uuid;
+
+/// A single physical segment of a logical index.
+///
+/// Each segment is stored independently and will become one manifest entry when committed.
+/// The logical index identity (name / target column / dataset version) is provided separately
+/// by the commit API.
+#[derive(Debug, Clone, PartialEq)]
+pub struct IndexSegment {
```
westonpace, line 15: Bleh.  I don't love that this type has to be in `lance-index`  It isn't actually used by anything in `lance-index` and the only reason it is here is because `DatasetIndexExt` is here.  I don't think we need to fix it today but we really should move the `DatasetIndexExt` trait into `lance`.

rust/lance/src/dataset/mem_wal/memtable/flush.rs
```diff
@@ -247,11 +247,8 @@ impl MemTableFlusher {
                     index_meta.fields = vec![field_idx];
                     index_meta.dataset_version = dataset.version().version;
                     // Calculate fragment_bitmap from dataset fragments
-                    let fragment_ids: roaring::RoaringBitmap = dataset
-                        .get_fragments()
-                        .iter()
-                        .map(|f| f.id() as u32)
-                        .collect();
+                    let fragment_ids: roaring::RoaringBitmap =
+                        dataset.fragment_bitmap.as_ref().clone();
```
westonpace: Is this just an unrelated minor perf bump?

rust/lance/src/index/vector/ivf/v2.rs
```diff
@@ -1460,7 +1460,14 @@ mod tests {
             .unwrap();
 
         dataset
-            .commit_existing_index(index_name, "vector", shared_uuid)
+            .commit_existing_index_segments(
```
westonpace, line 1470: We are only committing one segment here.  Why do we need to use the new API?  Is it because the new API allows us to pass in the fragment bitmap?  If so, should we deprecate the old API?

rust/lance/src/dataset.rs
```diff
@@ -1817,6 +1817,11 @@ impl Dataset {
             .collect()
     }
 
+    /// Iterate over manifest fragments without allocating [`FileFragment`] wrappers.
+    pub fn iter_fragments(&self) -> impl Iterator<Item = &Fragment> {
+        self.manifest.fragments.iter()
+    }
+
```
westonpace: Couldn't the user just call `ds.fragments().iter()`?  Is this method really needed?

rust/lance/src/index.rs
```diff
@@ -761,7 +761,7 @@ impl DatasetIndexExt for Dataset {
         column: &str,
         index_id: Uuid,
     ) -> Result<()> {
-        let Some(field) = self.schema().field(column) else {
+        let Some(_field) = self.schema().field(column) else {
```
westonpace: Just remove this check.  We have an identical check in `commit_existing_index_segments` already.

rust/lance/src/index.rs
westonpace: Should this method be removed?

rust/lance-index/src/types.rs
```diff
@@ -0,0 +1,84 @@
...
+    /// Unique ID of the physical segment.
+    uuid: Uuid,
+    /// The fragments covered by this segment.
+    fragment_bitmap: RoaringBitmap,
+    /// Metadata specific to the index type.
+    index_details: Option<Arc<prost_types::Any>>,
```
westonpace: `index_details` shouldn't really be optional for new code.  We rely on them to determine the index type.  Though if we make this non-optional then the `commit_existing_index` method won't be able to provide them.  Perhaps we can just stub that path in with a dummy message of some kind that we convert to None lower in the commit path.  This way we don't have to pollute this nice new API with an optional field that could be misused.

rust/lance/src/index.rs
```diff
@@ -770,22 +770,63 @@ impl DatasetIndexExt for Dataset {
...
+                IndexMetadata {
+                    uuid,
+                    name: index_name.to_string(),
+                    fields: vec![field.id],
+                    dataset_version: self.manifest.version,
+                    fragment_bitmap: Some(fragment_bitmap),
+                    index_details,
+                    index_version,
+                    created_at: Some(chrono::Utc::now()),
+                    base_id: None,
```
westonpace, line 819: Is there no way to specify a base id?  Not that important for now but I think we'll eventually need it.

rust/lance-index/src/types.rs
Xuanwo: Created a follow-up at https://github.com/lance-format/lance/issues/6221

rust/lance/src/dataset/mem_wal/memtable/flush.rs
Xuanwo: aha, yep. Let me revert it

rust/lance/src/index/vector/ivf/v2.rs
Xuanwo: Yes, I'm going to deprecate the old API and we will finally remove it,

rust/lance/src/dataset.rs
Xuanwo: It was added because it used to be useful, but the new IndexSegment API makes it seem unnecessary. It has been removed.

rust/lance/src/index.rs
Xuanwo: Removed.

rust/lance/src/index.rs
Xuanwo: My plan is to deprecate it becuase it seems to be our public API. Am I understanding this in wrong?

Xuanwo: Let me just remove it so that I can fix https://github.com/lance-format/lance/pull/6209#discussion_r2953139179 at the same itme.

rust/lance/src/index.rs
Xuanwo: IndexSegment exposed this API but no upper APIs to use it, we need to refactor the upper APIs first. Yes, we can treat this as a follow up.

rust/lance-index/src/types.rs
Xuanwo: Changed into required.

PR #6215, opened 2026-03-17
feat: add manifest namespace extended properties

wojiaodoubao:
This is a sub-task of https://github.com/lance-format/lance/issues/5741

### DIFF
--- a/rust/lance-namespace-impls/src/dir/manifest.rs
+++ b/rust/lance-namespace-impls/src/dir/manifest.rs
@@ -7,20 +7,25 @@
 //! to track tables and nested namespaces.
 
 use arrow::array::builder::{ListBuilder, StringBuilder};
-use arrow::array::{Array, RecordBatch, RecordBatchIterator, StringArray};
+use arrow::array::{Array, ArrayRef, RecordBatch, RecordBatchIterator, StringArray};
 use arrow::datatypes::{DataType, Field, Schema as ArrowSchema};
 use arrow_ipc::reader::StreamReader;
+use arrow_schema::{FieldRef, Schema};
 use async_trait::async_trait;
 use bytes::Bytes;
 use futures::{FutureExt, stream::StreamExt};
+use lance::dataset::NewColumnTransform;
 use lance::dataset::optimize::{CompactionOptions, compact_files};
 use lance::dataset::{
     DeleteBuilder, MergeInsertBuilder, ReadParams, WhenMatched, WhenNotMatched, WriteParams,
     builder::DatasetBuilder,
 };
+use lance::deps::arrow_array::RecordBatchOptions;
+use lance::deps::datafusion::scalar::ScalarValue;
 use lance::index::DatasetIndexExt;
 use lance::session::Session;
 use lance::{Dataset, dataset::scanner::Scanner};
+use lance_arrow::RecordBatchExt;
 use lance_core::Error as LanceError;
 use lance_core::datatypes::LANCE_UNENFORCED_PRIMARY_KEY_POSITION;
 use lance_core::{Error, Result};
@@ -42,9 +47,12 @@ use lance_namespace::models::{
 };
 use lance_namespace::schema::arrow_schema_to_json;
 use object_store::path::Path;
+use std::collections::HashSet;
 use std::io::Cursor;
+use std::str::FromStr;
 use std::{
     collections::HashMap,
+    f32, f64,
     hash::{DefaultHasher, Hash, Hasher},
     ops::{Deref, DerefMut},
     sync::Arc,
@@ -53,6 +61,20 @@ use tokio::sync::{RwLock, RwLockReadGuard, RwLockWriteGuard};
 
 const MANIFEST_TABLE_NAME: &str = "__manifest";
 const DELIMITER: &str = "$";
+
+/// Extended properties are special properties starting with `lance.manifest.extended.`.
+/// They are stored as dedicated columns in the `__manifest` table.
+///
+/// Example:
+/// ```json
+/// {
+///   "user_name": "Alice",
+///   "lance.manifest.extended.user_id": "123456"
+/// }
+/// ```
+/// `user_name` is stored in the `metadata` JSON column, while `user_id` is stored in a column
+/// named `user_id`.
+static EXTENDED_PREFIX: &str = "lance.manifest.extended.";
 
 // Index names for the __manifest table
 /// BTREE index on the object_id column for fast lookups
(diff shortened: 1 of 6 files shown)


rust/lance-namespace-impls/src/dir/manifest.rs
```diff
@@ -379,6 +405,78 @@ impl ManifestNamespace {
...
+    pub async fn add_extended_properties(&self, properties: &Vec<(&str, DataType)>) -> Result<()> {
+        let full_schema = self.full_manifest_schema().await?;
+        let fields: Vec<Field> = properties
+            .iter()
+            .map(|(name, data_type)| {
+                if !name.starts_with(EXTENDED_PREFIX) {
+                    return Err(Error::io(format!(
+                        "Extended properties key {} must start with prefix: {}",
+                        name, EXTENDED_PREFIX
+                    )));
```
wjones127: This error isn't about IO, but is incorrect input:
```suggestion
                    return Err(Error::invalid_input(format!(
                        "Extended properties key {} must start with prefix: {}",
                        name, EXTENDED_PREFIX
                    )));
```

rust/lance-namespace-impls/src/dir/manifest.rs
```diff
@@ -379,6 +405,78 @@ impl ManifestNamespace {
...
+                }
+                Ok(Field::new(
+                    name.strip_prefix(EXTENDED_PREFIX).unwrap().to_string(),
+                    data_type.clone(),
+                    true,
+                ))
+            })
+            .collect::<Result<Vec<_>>>()?
+            .into_iter()
+            .filter(|f| full_schema.column_with_name(f.name()).is_none())
```
wjones127: We just throw out names that collide? Why not error? Or at least warn?

Or is this just making the migration idempotent? Even if this is the case, I'm curious if there are other columns that it *could* collide with that we should disallow.

rust/lance-namespace-impls/src/dir/manifest.rs
```diff
@@ -379,6 +405,78 @@ impl ManifestNamespace {
...
+            .collect();
+
+        let schema = Schema::new(fields);
```
wjones127, line 474: Should we early return if `fields` is empty?

rust/lance-namespace-impls/src/dir/manifest.rs
```diff
@@ -1172,6 +1256,75 @@ impl ManifestNamespace {
...
+                    .filter(|(key, _)| !key.starts_with(EXTENDED_PREFIX))
+                    .collect::<HashMap<_, _>>();
+                Some(serde_json::to_string(&meta_props).ok()?)
+            }
+        })
+    }
+}
+
+/// Parse one row of __manifest table into manifest object.
+fn parse_manifest_object(batch: &RecordBatch) -> Result<ManifestObject> {
```
wjones127: I agree with the bot comment, instead of creating slice just pass in the index you want to extract:


```suggestion
/// Parse one row of __manifest table into manifest object.
fn parse_manifest_object(batch: &RecordBatch, index: usize) -> Result<ManifestObject> {
```

rust/lance-namespace-impls/src/dir/manifest.rs
```diff
@@ -1840,110 +2104,487 @@ impl LanceNamespace for ManifestNamespace {
...
+        .map_err(|e| Error::io(format!("Failed to create manifest batch: {}", e)))?;
+    Ok(batch)
+}
+
+fn build_extended_column_array(
+    col_name: &str,
+    data_type: &DataType,
+    extended_records: &[Option<RecordBatch>],
+    props_vec: &[Option<HashMap<String, String>>],
+) -> Result<ArrayRef> {
```
wjones127: I'm a bit confused as to how this is implemented. It seems like sometimes you are using RecordBatch as a row, which is very inefficient.

Perhaps you should write some function like:

```rust
fn records_to_batch(
    records: &HashMap<String, String>,
    schema: &Schema
) -> Result<RecordBatch> { ... }
```

And then you can do the merges with other data. in a columnar fashion. I imagine that function might be useful in multiple places.

wjones127 (commented):
I'm feeling a little uncertain about the efficiency of this code, particularly around the handling of batches. Though I also think the bot is right about projecting many columns.

wojiaodoubao:
I’m very sorry for the long delay. I’ve been tied up with something urgent recently. I’ve updated the PR—please take a look when you have time. Thanks very much! @wjones127

rust/lance-namespace-impls/src/dir/manifest.rs
```diff
@@ -407,6 +435,101 @@ impl ManifestNamespace {
...
+        Ok(())
+    }
+
+    /// Get all extended properties keys currently present in the manifest schema.
+    pub async fn get_extended_properties_keys(&self) -> Result<Vec<String>> {
+        let basic_cols: HashSet<String> = Self::manifest_schema()
+            .fields
+            .iter()
+            .map(|f| f.name().to_string())
+            .collect();
```
wjones127, line 487: suggestion: this might be a good place to use `Cow<'_, str>`.

rust/lance-namespace-impls/src/dir/manifest.rs
```diff
@@ -1689,6 +1802,346 @@ impl ManifestNamespace {
...
+    Ok(out)
+}
+
+/// Transform one record into a record batch.
+fn record_to_batch(
+    record: &HashMap<String, ScalarValue>,
+    schema: &ArrowSchema,
+) -> Result<RecordBatch> {
+    let mut fields: Vec<FieldRef> = Vec::new();
+    let mut arrays: Vec<ArrayRef> = Vec::new();
```
wjones127, line 2053: suggestion: pre-size these arrays:

```suggestion
    let mut fields: Vec<FieldRef> = Vec::with_capacity(schema.fields().len());
    let mut arrays: Vec<ArrayRef> = Vec::with_capacity(schema.fields().len());
```

rust/lance-namespace-impls/src/dir/manifest.rs
```diff
@@ -42,9 +47,12 @@ use lance_namespace::models::{
 };
 use lance_namespace::schema::arrow_schema_to_json;
 use object_store::path::Path;
+use std::collections::HashSet;
 use std::io::Cursor;
+use std::str::FromStr;
 use std::{
     collections::HashMap,
+    f32, f64,
```
wjones127, line 55: question: why import these?

wjones127 (approved):
Looks reasonable. Just optional feedback remaining. Can merge once rebased.

wjones127:
@wojiaodoubao You still planning on finishing this?

wjones127:
Closing due to inactivity. Feel free to re-open if you want to complete it.

PR #6216, opened 2026-03-17
fix: respect the old data filter on inverted index

westonpace:
This is basically the same problem encountered in https://github.com/lance-format/lance/pull/5929

However, the solution there (and in other indices fixed since then) has been to prune indices during the update phase to remove old fragments.  Unfortunately, FTS indices are maybe too large for this pruning to make sense (it would require a complete scan through of the old index) though I haven't verified how bad the performance would be.

This PR instead keeps track of which fragments have been invalidated.  It then uses this as a block-list at search time.

### DIFF
--- a/rust/lance/src/dataset/tests/dataset_merge_update.rs
+++ b/rust/lance/src/dataset/tests/dataset_merge_update.rs
@@ -1886,3 +1886,188 @@ async fn test_fts_index_stale_data_after_merge_insert_compact_optimize() {
     let (final_dataset, _) = merge_job2.execute(reader_to_stream(reader2)).await.unwrap();
     final_dataset.validate().await.unwrap();
 }
+
+/// Regression test: when rows are updated in-place, the FTS index must
+/// invalidate old entries and allow re-indexing incrementally.
+///
+/// Sequence:
+/// 1. Write fragments 1 and 2.
+/// 2. Create FTS index covering fragments 1 and 2.
+/// 3. Update fragment 1 in-place via merge_insert (DataReplacement path).
+///    This removes fragment 1 from the index's fragment_bitmap.
+/// 4. Call optimize_indices (append) to create a new index segment covering
+///    the updated fragment 1.
+/// 5. Call optimize_indices (merge) to merge both segments. The first segment
+///    contains the old, invalidated values for fragment 1; the second segment
+///    contains the new, valid values. We must keep only the new values.
+#[tokio::test]
+async fn test_fts_index_incremental_reindex_after_in_place_update() {
+    use lance_index::scalar::{FullTextSearchQuery, inverted::InvertedIndexParams};
+
+    let schema = Arc::new(ArrowSchema::new(vec![
+        ArrowField::new("id", DataType::Int32, false),
+        ArrowField::new("text", DataType::Utf8, true),
+    ]));
+
+    // Step 1: Create dataset with 2 rows in separate fragments
+    let batch = RecordBatch::try_new(
+        schema.clone(),
+        vec![
+            Arc::new(Int32Array::from(vec![0, 1])),
+            Arc::new(StringArray::from(vec![
+                "the quick brown fox",
+                "the lazy dog",
+            ])),
+        ],
+    )
+    .unwrap();
+    let reader = RecordBatchIterator::new(vec![Ok(batch)], schema.clone());
+    let mut dataset = Dataset::write(
+        reader,
+        "memory://test_fts_incremental_reindex",
+        Some(WriteParams {
+            max_rows_per_file: 1, // Force 2 fragments
+            ..Default::default()
+        }),
+    )
+    .await
+    .unwrap();
+
+    // Step 2: Create FTS inverted index on 'text'
+    let params = InvertedIndexParams::default();
+    dataset
+        .create_index(&["text"], IndexType::Inverted, None, &params, true)
+        .await
+        .unwrap();
+
+    // Sanity check: "quick" and "lazy" should each return 1 result
+    let results = dataset
+        .scan()
+        .full_text_search(FullTextSearchQuery::new("quick".to_owned()))
+        .unwrap()
+        .try_into_batch()
+        .await
+        .unwrap();
+    assert_eq!(results.num_rows(), 1);
+    let results = dataset
+        .scan()
+        .full_text_search(FullTextSearchQuery::new("lazy".to_owned()))
+        .unwrap()
+        .try_into_batch()
+        .await
+        .unwrap();
+    assert_eq!(results.num_rows(), 1);
+
+    // Step 3: merge_insert with reversed column order to trigger
+    // RewriteColumns/DataReplacement path, which prunes the index
+    // fragment bitmap for the updated fragment.
+    // Update id=1 ("the lazy dog" -> "a speedy cat")
+    let reversed_schema = Arc::new(ArrowSchema::new(vec![
+        ArrowField::new("text", DataType::Utf8, true),
+        ArrowField::new("id", DataType::Int32, false),
+    ]));
+    let source_batch = RecordBatch::try_new(
+        reversed_schema.clone(),
+        vec![
+            Arc::new(StringArray::from(vec!["a speedy cat"])),
+            Arc::new(Int32Array::from(vec![1])),
+        ],
+    )
+    .unwrap();
+
+    let merge_job = MergeInsertBuilder::try_new(Arc::new(dataset.clone()), vec!["id".to_string()])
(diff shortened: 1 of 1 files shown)


rust/lance-index/src/scalar.rs
```diff
@@ -807,10 +807,13 @@ pub struct UpdateCriteria {
 /// - stable row IDs: use exact row-id membership instead
 #[derive(Debug, Clone)]
 pub enum OldIndexDataFilter {
-    /// Keep old rows whose row-address fragment is in this bitmap.
+    /// Keeps track of which fragments are still valid and which are no longer valid.
```
Xuanwo, line 839: Does this fix work for tables with stable row IDs?

westonpace: Probably not but I'm a little vague on how indices work with stable row ids in general.  Do we store the stable row id in the index?  Or are we still storing row addresses?

Can we just filter search results against the list of valid row ids?  If a dataset has an ordered list of valid row ids in memory then it should just be one O(N) pass against the index search results?

I'll add this as a follow-up task.

westonpace: https://github.com/lance-format/lance/issues/6262

rust/lance-index/src/scalar/inverted/builder.rs
```diff
@@ -1577,6 +1608,20 @@ async fn merge_metadata_files(
...
+            let invalid_fragments_col = metadata_batch
+                .column_by_name(INVALIDATED_FRAGMENTS_COL)
+                .expect_ok()?;
+            let invalid_fragments_arr = invalid_fragments_col
+                .as_any()
+                .downcast_ref::<BinaryArray>()
+                .expect_ok()?;
+            let part_invalid_fragments =
+                RoaringBitmap::deserialize_from(invalid_fragments_arr.value(0))?;
+            invalidated_fragments.extend(part_invalid_fragments);
```
wjones127: question: when might parts have different invalid fragments? Do we correctly handle where we invalidate fragment 1 in index 1, then index fragment 1 in index 2, and then try to merge index 1 and 2?

rust/lance-index/src/scalar/inverted/builder.rs
```diff
@@ -2532,4 +2582,287 @@ mod tests {
...
+        // Load it back and confirm the invalidated fragments are set
+        let index = InvertedIndex::load(src_store.clone(), None, &LanceCache::no_cache())
+            .await
+            .unwrap();
+        assert_eq!(index.invalidated_fragments(), &initial_invalid);
+
+        // Remap the index via the ScalarIndex trait method (identity mapping)
+        use crate::scalar::ScalarIndex;
+        let mapping = HashMap::from([(0u64, Some(0u64))]);
+        index.remap(&mapping, dest_store.as_ref()).await.unwrap();
```
wjones127: question: do we do identity remapping today? Do we test the common case of remapping to new fragments?

rust/lance-index/src/scalar/inverted/builder.rs
```diff
@@ -2532,4 +2582,287 @@ mod tests {
...
+
+        let batch2 = make_doc_batch("new doc", 1);
+        let stream2 =
+            RecordBatchStreamAdapter::new(batch2.schema(), stream::iter(vec![Ok(batch2)]));
+        let stream2 = Box::pin(stream2);
+
+        // Use RowIds filter instead of Fragments — should not affect invalidated_fragments
+        let mut valid_ids = lance_core::utils::mask::RowAddrTreeMap::new();
+        valid_ids.insert(0);
+        let old_data_filter = Some(crate::scalar::OldIndexDataFilter::RowIds(valid_ids));
```
wjones127: question (non-blocking): how do we handle invalidated stable row_ids? I'm fine with handling this in a follow up issue if it's not solved yet.

rust/lance-index/src/scalar/inverted/index.rs
```diff
@@ -342,6 +343,9 @@ pub struct InvertedIndex {
     tokenizer: Box<dyn LanceTokenizer>,
     token_set_format: TokenSetFormat,
     pub(crate) partitions: Vec<Arc<InvertedPartition>>,
+    // Fragments which are contained in the index, but no longer in the dataset.
+    // These should be pruned at search time since we don't prune them at update time.
+    invalidated_fragments: RoaringBitmap,
```
wjones127: nitpick: when I say "invalidated fragments" elsewhere, I mean something different that deleted fragments. I mean that those fragments have had the column value updated in-place since this index segment was created. It might be nice to keep consistency on this terminology.

rust/lance-table/src/format/index.rs
```diff
@@ -108,6 +108,14 @@ impl IndexMetadata {
             .as_ref()
             .map(|files| files.iter().map(|f| f.size_bytes).sum())
     }
+
+    pub fn ineffective_fragment_bitmap(
```
wjones127: nitpick: "ineffective" seems like a weird way of putting it, but I'm blanking on a better name right now. I'm guessing it's in reference to the `effective_fragment_bitmap`?

westonpace: Yes.  I suppose deleted_fragment_bitmap would probably be fine.?

rust/lance-index/src/scalar/inverted/builder.rs
westonpace: We do not.  I already created a follow-up (https://github.com/lance-format/lance/issues/6262) but I'm not entirely sure we need to handle the stable row id case at all since the result of a stable row id search is filtered by an allow-list of valid stable row ids at the moment.  Though that behavior may change someday.

rust/lance-index/src/scalar/inverted/builder.rs
westonpace: I'm not sure I understand how this would happen.  Fragments are invalidated during compaction because they've been deleted.  How would we then go and index fragment 1 in index 2?

In any case, once a fragment is invalidated it never becomes "uninvalidated" (can't restore a fragment with it's original ID) so I think the current behavior (union of all invalidated fragments) should be safe.

rust/lance-index/src/scalar/inverted/builder.rs
westonpace: Fair point, Claude generated this test.  We shouldn't ever do identity remapping.  I'll fix it.

westonpace: Fixed.  That being said, remapping is pretty trivial here. A deleted fragment should never appear in the remapping.  All we need to do here is make sure we don't lose the deleted fragments list.

rust/lance-index/src/scalar/inverted/index.rs
westonpace: I renamed all invalidated_fragments to deleted_fragments.

rust/lance-index/src/scalar/inverted/builder.rs
wjones127: To clarify, I'm not talking about invalidation due to deletion. I'm talking about invalidation to data update. For example, like in this test:

https://github.com/lance-format/lance/blob/244c721504c6ef0b4c2f9700a342509976898d6e/rust/lance/src/dataset/tests/dataset_merge_update.rs#L1611-L1612

Sequence can go like:

1. Write fragments 1 and 2.
2. Create FTS index covering fragment 1 and 2
3. Update fragment 1 in-place using either merge_insert or DataReplacement. This removes fragment 1 from the `fragment_bitmap`.
4. Call optimize index to create a new segment, will see fragment 1 is unindexed and create an index segment covering fragment 1
5. Call optimize index to merge both segments. The first index segment contains the old, invalidated values for fragment 1. The second index segment contains the new, valid values for fragment 1. We should keep the ones from the second.

Basically, when rows are updated in-place, we need to invalidate them in the index. But then we also need to allow re-indexing them incrementally.

westonpace: Ok, I think you are correct that would be a problem.  The merged segment would still contain `1` in its `deleted_fragments`.  If we keep it, we will miss results from the new data.  If we drop it we will get results from the old data.

westonpace: Ok.  This is a problem with IVF/PQ indexes as well.  At the moment I kind of want to move forward with this PR as it doesn't actually make things worse and DataReplacement operations are less common at the moment.  But...I think we have more problems to solve.  I have created a follow-up at https://github.com/lance-format/lance/issues/6283

wjones127 (approved):

PR #6222, opened 2026-03-18
feat: pluggable index cache via CacheBackend trait

wjones127:
Previously the Session's index cache was hardcoded to Moka. This adds a
`CacheBackend` trait so users can provide their own cache implementation,
with `MokaCacheBackend` as the default.

Serialization and zero-IO reconstruction are in the follow-up PR #6223.


rust/lance/src/session.rs
```diff
@@ -182,6 +195,11 @@ impl Session {
         self.store_registry.clone()
     }
 
+    /// Get a reference to the raw metadata cache (for use in index reconstruction).
+    pub fn file_metadata_cache(&self) -> &LanceCache {
```
wjones127, line 199: Is this used anymore?

rust/lance-core/src/cache.rs
```diff
@@ -2,58 +2,250 @@
...
-    size_accessor: Arc<dyn Fn(&ArcAny) -> usize + Send + Sync>,
+// ---------------------------------------------------------------------------
+// CacheBackend trait
+// ---------------------------------------------------------------------------
+
+/// Low-level pluggable cache backend.
+///
+/// Implementations store entries keyed by opaque byte slices.
+/// The [`LanceCache`] wrapper handles key construction and type safety;
+/// backend authors do not need to worry about key encoding.
```
wjones127: I think we should say something about the `CacheKey` and `parse_cache_key` items here.

rust/lance-core/src/cache/mod.rs
```diff
@@ -2,58 +2,250 @@
...
 //! Cache implementation
-
-use std::any::{Any, TypeId};
+//!
+//! This module provides a two-layer caching system:
+//!
+//! - [`CacheBackend`] is the low-level, pluggable trait that custom cache implementations
+//!   can implement. It uses opaque byte keys and type-erased entries.
+//! - [`LanceCache`] is the typed wrapper that handles key construction (prefix + type tag
+//!   encoding), type-safe get/insert, and DeepSizeOf-based size computation.
```
wjones127: Maybe we should make a module focused on keys and key handling?

rust/lance-core/src/cache/mod.rs
```diff
@@ -2,58 +2,250 @@
 // SPDX-FileCopyrightText: Copyright The Lance Authors
 
 //! Cache implementation
-
-use std::any::{Any, TypeId};
+//!
+//! This module provides a two-layer caching system:
+//!
+//! - [`CacheBackend`] is the low-level, pluggable trait that custom cache implementations
```
wjones127: Maybe backends should be moved into a submodule?

I'm thinking:

```
mod cache
  struct LanceCache
  mod backend
    trait CacheBackend
    mod moka
      struct MokaCacheBackend
  mod keys
    trait CacheKey
    func parse_cache_key
```

And so on.

rust/lance-core/src/cache.rs
```diff
@@ -2,58 +2,250 @@
...
+    /// Implementations should deduplicate concurrent loads for the same key
+    /// so the loader runs at most once.
+    ///
+    /// The loader is a pinned future that produces `(entry, size_bytes)`.
+    /// It borrows from the caller's scope and will be `.await`ed within
+    /// this method — implementations must not store it beyond the call.
+    async fn get_or_insert<'a>(
+        &self,
+        key: &[u8],
+        loader: Pin<Box<dyn Future<Output = Result<(CacheEntry, usize)>> + Send + 'a>>,
```
wjones127: how viable would it be to move this to be a closure? Is constructing this future every time we access expensive?

rust/lance-core/src/cache.rs
```diff
@@ -468,111 +579,92 @@ impl WeakLanceCache {
...
-    type ValueType;
+    type ValueType: 'static;
 
     fn key(&self) -> Cow<'_, str>;
+
+    /// Short, stable string that distinguishes this value type from others in
+    /// the cache. Used as the suffix in the encoded cache key (`user_key\0type_id`).
+    /// Must be consistent across crate boundaries — use a short literal, not
+    /// `type_name` pointers.
+    fn type_id(&self) -> &'static str;
```
wjones127: Maybe let's just call it `type_name`, since that's a bit more descriptive.

rust/lance/src/index/vector/ivf/partition_serde.rs
```diff
@@ -0,0 +1,1153 @@
...
+    storage_len: u64,
+}
+
+impl<S: IvfSubIndex> PartitionEntry<S, ProductQuantizer> {
+    /// Serialize this partition entry to bytes.
+    ///
+    /// The sub-index, PQ codebook, and storage batch are each written as Arrow
+    /// IPC file sections, preceded by a small JSON header containing scalar
+    /// metadata and section lengths.
+    pub fn serialize(&self) -> Result<Vec<u8>> {
```
wjones127: Let's change the signature to `pub fn serialize(&self, writer: &mut dyn Writer) -> Result<usize>` (returning bytes written). Then we can stream out the data, rather than collecting it in memory.

wjones127: Want to apply the same things across different index types. Maybe we should write a trait for this?

```rust
trait Spillable {
    pub fn serialize(&self, writer: &mut dyn Writer) -> Result<usize>;
    pub fn deserialize(data: Bytes) -> Result<Self>;
}
```

rust/lance/src/index/vector/ivf/v2.rs
```diff
@@ -596,13 +632,204 @@ impl<S: IvfSubIndex + 'static, Q: Quantization + 'static> VectorIndex for IVFInd
...
+    }
 }
 
 pub type IvfFlatIndex = IVFIndex<FlatIndex, FlatQuantizer>;
 pub type IvfPq = IVFIndex<FlatIndex, ProductQuantizer>;
 pub type IvfHnswSqIndex = IVFIndex<HNSW, ScalarQuantizer>;
 pub type IvfHnswPqIndex = IVFIndex<HNSW, ProductQuantizer>;
 
+/// Reconstruct a concrete `IVFIndex<S, Q>` from cached state.
+async fn reconstruct_typed<S: IvfSubIndex + 'static, Q: Quantization + 'static>(
```
wjones127: We should add a unit test that validates that we can run this and `object_store.io_stats_incremental().read_requests == 0`, assuming we have everything right in the file_metadata_cache.

rust/lance-core/src/cache/backend.rs
```diff
@@ -0,0 +1,78 @@
...
+    /// Returns `(entry, was_cached)` where `was_cached` is `true` if the entry
+    /// was already present in the cache (the loader was not invoked).
+    ///
+    /// The loader is a pinned, boxed future rather than a generic closure
+    /// because `async_trait` erases the `Self` lifetime, making it impossible
+    /// to express a generic closure whose returned future borrows from the
+    /// caller. Boxing the future once at the call site (in `LanceCache`)
+    /// avoids this lifetime conflict while keeping the trait object-safe.
+    ///
+    /// The future borrows from the caller's scope and will be `.await`ed within
```
westonpace: All this description feels kind of redundant / obvious.  Like, if im a cache and i keep the loader around im gonna start leaking memory or file handles or something.

My first instinct is that its clutter but maybe in this day of agents it is better safe than sorry?

rust/lance-core/src/cache/backend.rs
```diff
@@ -0,0 +1,78 @@
...
+    async fn invalidate_prefix(&self, prefix: &str);
+
+    /// Remove all entries.
+    async fn clear(&self);
+
+    /// Number of entries currently stored (may flush pending operations).
+    async fn num_entries(&self) -> usize;
+
+    /// Total weighted size in bytes of all stored entries (may flush pending operations).
+    async fn size_bytes(&self) -> usize;
```
westonpace, line 105: We should assume here that no two entries share buffers?  It's probably the best we can do.

rust/lance-core/src/cache/backend.rs
```diff
@@ -0,0 +1,78 @@
...
+
+    /// Approximate number of entries, callable from synchronous contexts.
+    /// Backends that cannot provide this cheaply should return 0.
+    fn approx_num_entries(&self) -> usize {
```
westonpace, line 109: Would it be better for these to return Option<usize> with default of None?

rust/lance-core/src/cache/keys.rs
```diff
@@ -0,0 +1,71 @@
...
+    pub fn key(&self) -> &str {
+        &self.key
+    }
+
+    pub fn type_name(&self) -> &'static str {
+        self.type_name
+    }
+
+    /// Returns true if this key's prefix starts with the given string.
+    pub fn has_prefix(&self, prefix: &str) -> bool {
```
westonpace: Maybe `starts_with` would be more clear?

rust/lance-core/src/cache/keys.rs
```diff
@@ -0,0 +1,71 @@
...
+        self.prefix.starts_with(prefix)
+    }
+}
+
+pub trait CacheKey {
```
westonpace: Why is there both InternalCacheKey and CacheKey?  Should this trait be documented?

rust/lance-core/src/cache/keys.rs
```diff
@@ -0,0 +1,71 @@
...
+    /// **Must be unique per value type.** If two `CacheKey` impls return the
+    /// same `type_name` but different `ValueType`s, entries will collide and
+    /// downcasts will fail silently (returning `None` on get).
+    ///
+    /// Must be consistent across crate boundaries — use a short literal, not
+    /// `std::any::type_name` pointers.
+    fn type_name() -> &'static str;
+}
+
+pub trait UnsizedCacheKey {
```
westonpace: Document this?  Why is it needed?  (Yes, I know this is just copied from existing but still)

rust/lance-core/src/cache/keys.rs
```diff
@@ -0,0 +1,71 @@
...
+
+pub trait CacheKey {
+    type ValueType: 'static;
+
+    fn key(&self) -> Cow<'_, str>;
+
+    /// Short, stable string that distinguishes this value type from others in
+    /// the cache. Used as the suffix in the encoded cache key (`user_key\0type_name`).
+    ///
+    /// **Must be unique per value type.** If two `CacheKey` impls return the
```
westonpace: "Unique per value type" is strange wording.  I think you're trying to say that two type names are equal only if their value types are equal?

Or, to put another way, if two value types are not equal they must not have equal type names?

rust/lance-file/src/previous/reader.rs
```diff
@@ -83,12 +83,16 @@ impl<'a, T> StringCacheKey<'a, T> {
...
-impl<T> CacheKey for StringCacheKey<'_, T> {
+impl<T: 'static> CacheKey for StringCacheKey<'_, T> {
     type ValueType = T;
 
     fn key(&self) -> Cow<'_, str> {
         self.key.into()
     }
+
+    fn type_name() -> &'static str {
+        std::any::type_name::<T>()
```
westonpace, line 97: Didn't you say not to use this?

rust/lance-core/src/cache/keys.rs
```diff
@@ -0,0 +1,71 @@
...
+    /// same `type_name` but different `ValueType`s, entries will collide and
+    /// downcasts will fail silently (returning `None` on get).
+    ///
+    /// Must be consistent across crate boundaries — use a short literal, not
```
westonpace: What does consistent across crate boundaries mean?

westonpace (approved):
Just some nit picking

rust/lance-core/src/cache/backend.rs
wjones127: Yes, we do. There are ways we can detect this in the future. You may be interested in this PR: https://github.com/lance-format/lance/pull/6229

rust/lance-core/src/cache/keys.rs
wjones127: Yes. This is meant to be used as a key in a registry of types we'll use later for deserialization of cache entries.

rust/lance-file/src/previous/reader.rs
wjones127: Added a comment. We can use this internally, but shouldn't call it directly from other crates.

PR #6223, opened 2026-03-18
feat: introduce `CacheCodec` for serializing index cache entries

wjones127:
Introduce a `codec: CacheCodec` parameter to `CacheBackend` methods, providing the ability to serialize and deserialize entries, if desired. This struct is just a pair of function pointers: `serialize` and `deserialize`. So it's very cheap if not used.

To provide this for a type, implement the `CacheKey::codec()` method to return a `CacheCodec`. By default, this trait method returns `None`. We only implement serialization for some types right now, but the eventual goal is to implement it for all types. To keep this PR focused, I only implement this for vector indexes, but the same approach could be used for FTS and other indexes.

To make it easy to implement, we provide a `CacheCodecImpl` trait:

```rust
pub trait CacheCodecImpl: Send + Sync {
    fn serialize(&self, writer: &mut dyn std::io::Write) -> Result<()>;

    fn deserialize(reader: &mut dyn std::io::Read) -> Result<Self>
    where
        Self: Sized;
}
```

The implementors of `CacheKey` can then write:

```rust
    fn codec() -> Option<CacheCodec> {
        Some(CacheCodec::from_impl::<Self::ValueType>())
    }
```

The implementation of serialization / deserialization is designed to have very low overhead. We use serde-derived strings for small metadata, and use Arrow IPC for Arrow data. When deserializing, we use zero-copy deserialization to re-use the buffers read in.

One important change is we no longer cache `Arc<dyn VectorIndex>` for V2+ vector indices. This is because `IVFIndex` (which is the V3 implementation of `VectorIndex`) contains non-serializable data structures like `FileReader`. Instead we cache `IvfIndexState`, which is a serializable representation of the vector index. Importantly, this does not require any IO to reconstruct, which is validated by the pre-existing test `test_prewarm_ivf_pq`.

### DIFF
--- a/rust/lance/src/index/vector/ivf/partition_serde.rs
+++ b/rust/lance/src/index/vector/ivf/partition_serde.rs
@@ -1055,26 +1055,13 @@ mod tests {
         CacheCodecImpl::serialize(&entry, &mut bytes).unwrap();
 
         let restored =
-            <IvfStateEntryBox as CacheCodecImpl>::deserialize(&bytes::Bytes::from(bytes)).unwrap();
-        let restored = restored
-            .0
-            .as_any()
-            .downcast_ref::<IvfIndexState<FlatQuantizer>>()
-            .expect("expected IvfIndexState<FlatQuantizer>");
-        assert_eq!(
-            restored.index_file_path,
-            entry
-                .0
-                .as_any()
-                .downcast_ref::<IvfIndexState<FlatQuantizer>>()
-                .unwrap()
-                .index_file_path
-        );
-        assert_eq!(restored.uuid, "test-uuid-1234");
-        assert_eq!(restored.distance_type, DistanceType::L2);
-        assert_eq!(restored.sub_index_metadata, vec!["meta1".to_string()]);
-        assert_eq!(restored.cache_key_prefix, "prefix/");
-        assert_eq!(restored.index_file_size, 1024);
-        assert_eq!(restored.aux_file_size, 512);
+            <IvfStateEntryBox as CacheCodecImpl>::deserialize(&bytes::Bytes::from(bytes.clone()))
+                .unwrap();
+
+        // Re-serialize the restored entry and compare bytes — a stronger check
+        // than field-by-field comparison and avoids needing to downcast.
+        let mut restored_bytes = Vec::new();
+        CacheCodecImpl::serialize(&restored, &mut restored_bytes).unwrap();
+        assert_eq!(bytes, restored_bytes);
     }
 }
--- a/rust/lance/src/index/vector/ivf/v2.rs
+++ b/rust/lance/src/index/vector/ivf/v2.rs
@@ -138,8 +138,6 @@ struct IvfIndexStateHeader {
 /// wrapper lets the cache infrastructure work with a sized type while the
 /// hot paths call `reconstruct` without knowing `Q`.
 pub(crate) trait IvfStateEntry: DeepSizeOf + Send + Sync + 'static {
-    fn as_any(&self) -> &dyn std::any::Any;
-
     fn serialize_state(&self, writer: &mut dyn IoWrite) -> Result<()>;
 
     fn reconstruct<'a>(
@@ -293,10 +291,6 @@ impl CacheCodecImpl for IvfStateEntryBox {
 }
 
 impl<Q: Quantization + 'static> IvfStateEntry for IvfIndexState<Q> {
-    fn as_any(&self) -> &dyn std::any::Any {
-        self
-    }
-
     fn serialize_state(&self, writer: &mut dyn IoWrite) -> Result<()> {
         let quantizer_metadata_json = serde_json::to_string(&self.metadata)
             .map_err(|e| lance_core::Error::io(format!("IvfIndexState metadata: {e}")))?;


rust/lance/src/index/vector/ivf/partition_serde.rs
```diff
@@ -0,0 +1,1153 @@
...
+//! [header_len: u64 LE]
+//! [header: JSON bytes]
+//! [sub_index IPC file bytes]
+//! [... quantizer-specific IPC sections ...]
+//! [storage batch IPC file bytes]
+//! ```
+//!
+//! Each IPC section is a complete Arrow IPC file. On deserialization, the IPC
+//! sections are read zero-copy using [`FileDecoder`] so that Arrow arrays
+//! reference the original buffer directly.
```
wjones127: My big worry with this is that the DeepSizeOf will no longer be accurate after a roundtrip.

wjones127: This concern will be addressed by https://github.com/lance-format/lance/pull/6229. As long as we aren't sharing buffers *across* cache entries, we should be fine.

westonpace:
I think this PR might be in need of a rebase or something?

wjones127:
@westonpace Okay, I've finally finished rebasing and refactoring. I think it's good to review again.

rust/lance-core/src/cache/backend.rs
```diff
@@ -17,20 +17,19 @@ use futures::Future;
...
 
+use super::CacheCodec;
+
 /// A type-erased cache entry.
 pub type CacheEntry = Arc<dyn Any + Send + Sync>;
 
 /// Structured cache key passed to [`CacheBackend`] methods.
 ///
-/// Composed of three parts:
+/// Backend authors receive these ready-made from [`LanceCache`](super::LanceCache)
```
westonpace:
```suggestion
/// CacheBackend impls receive these ready-made from [`LanceCache`](super::LanceCache)
```

rust/lance-core/src/cache/codec.rs
```diff
@@ -0,0 +1,124 @@
...
+// ---------------------------------------------------------------------------
+
+pub(crate) type ArcAny = Arc<dyn std::any::Any + Send + Sync>;
+
+/// Type-erased codec for serializing and deserializing cache entries.
+///
+/// `CacheCodec` is two plain function pointers — it is `Copy` and has no
+/// heap allocation. Construct one via [`CacheCodec::from_impl`] for types
+/// that implement [`CacheCodecImpl`], or [`CacheCodec::new`] for custom
+/// cases (e.g. when the orphan rule prevents a direct impl).
```
westonpace, line 61: Huh...why not `Box<dyn CacheCodecImpl>`?  Is it because you're expecting to need to violate the orphan rule?

rust/lance-core/src/cache/mod.rs
```diff
@@ -70,12 +101,27 @@ pub trait CacheKey {
     /// `std::any::type_name` — the latter is not guaranteed stable across
     /// compiler versions or build configurations.
     fn type_name() -> &'static str;
+
+    /// Optional codec for serializing/deserializing this key's value type.
+    ///
+    /// Returns `None` by default. Cache backends that support persistence
+    /// (e.g. disk-backed caches) use this to serialize entries on insert and
+    /// deserialize on get. Types without a codec will only be stored in-memory.
```
westonpace, line 109: What happens in a disk cache if this returns None?  Do these caches fail?  Can we document that fact?

rust/lance-index/src/vector.rs
```diff
@@ -140,6 +144,184 @@ impl From<DistanceType> for pb::VectorMetricType {
...
+    /// Downcast to `&dyn Any` for concrete type access during reconstruction.
+    fn as_any(&self) -> &dyn Any;
+}
+
+/// Serializable state of an IVF index, sufficient to reconstruct the index
+/// without re-reading global buffers from object storage.
+///
+/// Produced by [`VectorIndex::cacheable_state`] and consumed by
+/// `reconstruct_vector_index` in the `lance` crate. All fields are `pub`
+/// because the type is shared across crate boundaries; they are **not**
```
westonpace: Why is this in `lance-core`?

rust/lance-index/src/vector.rs
```diff
@@ -140,6 +144,184 @@ impl From<DistanceType> for pb::VectorMetricType {
...
+    cache_key_prefix: String,
+    #[serde(default)]
+    index_file_size: u64,
+    #[serde(default)]
+    aux_file_size: u64,
+}
+
+/// Wire format:
+/// `[header_json_len: u64 LE][header JSON][ivf_pb_len: u64 LE][ivf protobuf]
+///  [extra_len: u64 LE][extra bytes][aux_ivf_pb_len: u64 LE][aux_ivf protobuf]`
```
westonpace: I guess there is no need to worry about stability here?  Do we plan for disk caches to be completely lost on upgrades?  Or are we actually thinking we might want to keep these stable between versions?

rust/lance/src/index/vector/ivf/partition_serde.rs
```diff
@@ -0,0 +1,1049 @@
...
+        Ok(Self { index, storage })
+    }
+}
+
+// ---------------------------------------------------------------------------
+// RabitQ
+// ---------------------------------------------------------------------------
+
+#[derive(Serialize, Deserialize)]
+struct RabitPartitionHeader {
```
westonpace, line 436: Would it make more sense to have these structs in their respective modules?  E.g. `rq`, `pq`, etc.?

rust/lance/src/index/vector/ivf/v2.rs
```diff
@@ -216,9 +282,24 @@ impl<S: IvfSubIndex + 'static, Q: Quantization> IVFIndex<S, Q> {
...
+        file_metadata_cache
+            .with_key_prefix(uri.as_ref())
+            .insert_with_key(&FileMetadataCacheKey, index_reader.metadata().clone())
+            .await;
+        let aux_path = index_dir
+            .child(uuid.as_str())
+            .child(INDEX_AUXILIARY_FILE_NAME);
+        file_metadata_cache
+            .with_key_prefix(aux_path.as_ref())
+            .insert_with_key(&FileMetadataCacheKey, storage.reader().metadata().clone())
```
westonpace, line 616: This is nice

rust/lance/src/index/vector/ivf/v2.rs
```diff
@@ -4287,4 +4630,208 @@ mod tests {
         let stats = dataset.object_store().io_stats_incremental();
         assert_io_eq!(stats, read_iops, 0, "second prewarm should not perform IO");
     }
+
+    /// A cache backend that serializes entries to bytes on insert and
+    /// deserializes on get. Entries without a codec are stored in-memory
+    /// as-is. This simulates a disk or remote cache backend.
```
westonpace: Should this comment be on `SerializingBackend`?

rust/lance-index/src/vector.rs
wjones127: For now, I'm marking them as unstable and experimental. I think we can see about making them stable in the future. I'd probably put some magic bytes and version in the header when we do.

> Do we plan for disk caches to be completely lost on upgrades?

Given we have a pre-warming mechanism, I feel okay with letting disk caches be ephemeral to start.

rust/lance-core/src/cache/mod.rs
wjones127: Up to the cache backend. If it's a disk-only cache, then it would not cache. If it's a memory and disk cache, then it can store in memory but when evicted it won't go to disk.

rust/lance-core/src/cache/codec.rs
wjones127: `Box<dyn CacheCodecImpl>` only works for `serialize()`. `deserialize()` returns `Result<Self>`, so it can't be called from `&dyn CacheCodecImpl`. This is why there is a `where Self: Sized` bound on the trait method.

In `CacheCodec::deserialize`, it returns `Result<ArcAny>`, which works without having to know the concrete type.

I went through a lot of iterations on these traits and structs. This is one place Rust gets really complicated 🫣

PR #6220, opened 2026-03-18
refactor: distributed vector segment build

Xuanwo:
This refactors distributed vector indexing into a staged segment-build pipeline and exposes the public APIs needed to integrate an external distributed build workflow with Lance. It defines the storage-level model for partial shards, staged planning, built segments, and segment commit, and documents the current distributed indexing flow.

This PR builds on the segment commit API from https://github.com/lance-format/lance/pull/6209. The main changes are organized into five commits:

- [test: cover distributed vector segment build](https://github.com/lance-format/lance/pull/6220/commits/a86274da2)
- [refactor: internalize distributed vector segment build](https://github.com/lance-format/lance/pull/6220/commits/1e5f0e15b)
- [feat: add public vector segment builder API](https://github.com/lance-format/lance/pull/6220/commits/691cecb9a)
- [feat: add Python vector segment builder API](https://github.com/lance-format/lance/pull/6220/commits/a07ef6144)
- [docs: document distributed vector segment build](https://github.com/lance-format/lance/pull/6220/commits/0a3f230d7)

Follow-up fixes:

- [fix: expose python create_index_uncommitted](https://github.com/lance-format/lance/pull/6220/commits/c1d3b1666)
- [fix: format python bindings](https://github.com/lance-format/lance/pull/6220/commits/b86f91c4d)
- [fix: format python segment builder bindings](https://github.com/lance-format/lance/pull/6220/commits/bfc9e63a0)

Please review accordingly.

---

Guide: https://github.com/lance-format/lance/blob/xuanwo/vector-staging-merge-internal/docs/src/guide/distributed_indexing.md

PR #6224, opened 2026-03-18
feat: add public vector segment builder API

Xuanwo:
This adds the public distributed vector finalize API on top of #6220's internal staging-merge flow. It exposes a canonical two-step workflow: materialize final segments from a staging root, then publish them with `commit_existing_index_segments(...)`.

This keeps `merge_index_metadata(...)` as the single-root compatibility path while making the multi-segment path explicit in the Rust API. The branch includes an end-to-end vector test that stages partial shards, builds final segments through the new builder, and commits them as one logical index.

Context: #6220

### DIFF
--- a/rust/lance-index/src/lib.rs
+++ b/rust/lance-index/src/lib.rs
@@ -33,7 +33,7 @@ pub mod types;
 pub mod vector;
 
 pub use crate::traits::*;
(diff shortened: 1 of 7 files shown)


Xuanwo:
Replaced by https://github.com/lance-format/lance/pull/6220

PR #6228, opened 2026-03-19
perf: share IVF partition search across deltas

Xuanwo:
This avoids re-ranking the same IVF centroids once per delta index during ANN search by sharing a single partition search across deltas when their metric and partition layout match.

When the active deltas disagree on metric type or partition count, the query falls back to the existing per-delta path so older inconsistent layouts continue to behave safely.

The `knn` test suite now covers the guard behavior and the updated partition-ranking metric semantics.

---

I believe this is the key reason why delta indexes perform so poorly for us.

I understand that `can_share_partition_search` is not an exact solution. If this approach does not work, we may need to add more metadata in our index proto to indicate that those segments can be searched together.

Also, I think @wjones127's https://github.com/lance-format/lance/discussions/6207 could be helpful.

### DIFF
--- a/rust/lance/src/io/exec/knn.rs
+++ b/rust/lance/src/io/exec/knn.rs
@@ -406,6 +406,8 @@ impl ANNIvfPartitionExec {
         })
     }
 
+    /// Determine whether one partition ranking can be safely reused across all
+    /// active delta indices for the query.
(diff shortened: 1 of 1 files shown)


Xuanwo:
This PR is converted into a draft because we are not sure about the route we wanna go.

PR #6229, opened 2026-03-19
refactor: replace deepsize crate with custom DeepSizeOf trait for Arrow-aware memory accounting

wjones127:
The external `deepsize` crate does not correctly account for Arrow buffers
that are shared across `Arc` references, causing double-counting in cache
size calculations.

This PR introduces a custom `DeepSizeOf` trait in `lance-core::deepsize`
with a `Context` that tracks both `Arc` and raw buffer pointers in a
unified `HashSet`. It also adds a `lance-derive` proc-macro crate for
`#[derive(DeepSizeOf)]` and removes the dependency on the external
`deepsize` crate.

Changes:
- Add `lance-core::deepsize` module with custom `DeepSizeOf` trait and pointer-tracking `Context`
- Add `lance-derive` proc-macro crate with `#[derive(DeepSizeOf)]`
- Remove external `deepsize` crate from all crates; update all impls to use `lance_core::deepsize`
- Fix all `deep_size_of_children` implementations to thread `context` through (previously some ignored the parameter)
- Add `DeepSizeOf` for `arrow_buffer::ScalarBuffer<T>` with proper pointer tracking
- Fix `PlainPostingList` and `Positions` to use `deep_size_of_children` with context


rust/lance-derive/src/lib.rs
```diff
@@ -0,0 +1,119 @@
...
+
+use proc_macro::TokenStream;
+use quote::quote;
+use syn::{Data, DeriveInput, Fields, parse_macro_input};
+
+/// Derive macro for the `DeepSizeOf` trait.
+///
+/// Generates an implementation that sums the `deep_size_of_children` of all
+/// fields (for structs) or the active variant's fields (for enums).
+#[proc_macro_derive(DeepSizeOf)]
```
westonpace, line 12: Are there tests for this macro somewhere?  I guess we get some testing via usage?  You can, in theory, do integration tests with proc-macros if I recall (unit tests don't make a ton of sense)

Maybe not important for an internal utility.

westonpace (commented):
This is nice, I didn't look too closely at all the changes as I think most were find/replace.

Should we do some blanket impl like?

```
impl<T: lance_core::deepsize::DeepSizeOf> deepsize::DeepSizeOf for T {
  ...
}
```

Although maybe better to keep the dependency away and let callers do that themselves if they are using `deepsize` I suppose.

wjones127:
> This is nice, I didn't look too closely at all the changes as I think most were find/replace.

The critical change from `deepsize` crate is that they track just Arc and RC pointers in `Context`: [^1]:

```rust
pub struct Context {
    /// A set of all [`Arc`](std::sync::Arc)s that have already been counted
    arcs: GenericSet<usize>,
    /// A set of all [`Rc`](std::sync::Arc)s that have already been counted
    rcs: GenericSet<usize>,
}
```
[^1]: https://github.com/Aeledfyr/deepsize/blob/5deebe687695610e6038dc70753defbe7634d549/src/lib.rs#L151-L156

We instead just track one list of pointers:

```rust
pub struct Context {
    seen: HashSet<usize>,
}
```

This is important because Arrow doesn't give you access to the `Arc`, just the underlying pointers. So if you want to mark Arrow buffers as seen, you can't use the `Context` from `deepsize`. (I initially tried to implement `deepsize::DeepSizeOf` for a new type wrapper around Arrow types, but ran into this. That's what motivated me to make this change.)


> Should we do some blanket impl like?
> 
> ```
> impl<T: lance_core::deepsize::DeepSizeOf> deepsize::DeepSizeOf for T {
>   ...
> }
> ```

I think this might be difficult, given I would need to create some sort of adapter between our `Context` and their `Context`. Plus, I don't get the sense that `deepsize` crate is actively maintained. All the PRs are left as stale. It's our only dependency that still uses `syn` `1.x`. So I lean towards dropping it.

rust/lance-derive/src/lib.rs
wjones127: You are right, I should add some more tests.

westonpace (approved):
I suppose this is one way to work around the orphan rule 😆

wjones127:
> I suppose this is one way to work around the orphan rule 😆

DeepSizeOf is basically unmaintained these days. My PR to upgrade dependencies has been sitting for months with no response. https://github.com/Aeledfyr/deepsize/pull/42

PR #6234, opened 2026-03-19
fix: 2.1/2.2 panic when a list column had small values and many empty values

### LINKED ISSUE #6184: panic: assertion failed: chunk_bytes <= max_chunk_size in PrimitiveStructuralEncoder::encode_miniblock
## Summary

A panic occurs during compaction when a miniblock chunk exceeds the 32KiB size limit enforced for Lance v2.1 files.

## Panic

```
thread 'lance-cpu' panicked at 'assertion failed: chunk_bytes <= max_chunk_size': rust/lance-encoding/src/encodings/logical/primitive.rs:3897
```

## Backtrace

```
 8: lance_encoding::encodings::logical::primitive::PrimitiveStructuralEncoder::encode_miniblock
 9: lance_core::utils::tokio::spawn_cpu::{{closure}}
10: tokio::runtime::task::core::Core<T,S>::poll
11: tokio::runtime::task::harness::Harness<T,S>::poll
```

## Analysis

In `serialize_miniblocks` (primitive.rs), the assertion at line 3897 is:

```rust
let max_chunk_size = if support_large_chunk {
    4 * 1024 * 1024 * 1024 // 4GB limit with u32 metadata
} else {
    32 * 1024 // 32KiB limit with u16 metadata
};
assert!(chunk_bytes <= max_chunk_size);
```

`support_large_chunk` is `false` for files using Lance file version < 2.2 (`LanceFileVersion::V2_2`). During compaction of v2.1 files, the serializer uses the 32KiB (u16 metadata) path. If a single miniblock chunk — after writing the header, rep/def buffers, value buffers, and alignment padding — exceeds 32KiB, this assertion fires.

The chunk splitting logic (upstream of serialization) does not appear to enforce the 32KiB limit, so it's possible to produce a chunk that is valid from the compressor's perspective but exceeds the serialization format's capacity.

## Expected Behavior

Either the chunk splitting should be capped to ensure no single chunk exceeds the format limit, or the assertion should be replaced with proper error handling so the job can fail gracefully rather than panic.

## Observed Impact

This panic repeatedly fails the same compaction jobs, blocking compaction progress.

westonpace:
If there are some non-null / non-empty values (so we can't use all-null branch) and they are small (so we use miniblock) and there are many repetition levels (can be caused by many empty / null lists) then we can get in a situation where we have too many rep/def levels for a chunk (exceed max chunk size).

This PR detects that scenario and falls back to full-zip encoding in these cases (it's less performant decoding but these situations should be very rare or maybe in the future we can have different structure encodings for very sparse data)

Closes #6184

### DIFF
--- a/rust/lance-encoding/src/encodings/logical/primitive.rs
+++ b/rust/lance-encoding/src/encodings/logical/primitive.rs
@@ -5186,66 +5186,57 @@ impl PrimitiveStructuralEncoder {
             // lists with very few values), fall back to fullzip to avoid exceeding
             // the u16 per-chunk rep/def buffer size limit.
             let too_sparse = Self::repdef_too_sparse_for_miniblock(&repdef, num_values);
-            if too_sparse {
+
+            if !too_sparse {
+                if let DataBlock::Dictionary(dict) = data_block {
+                    log::debug!("Encoding column {} with {} items using dictionary encoding (already dictionary encoded)", column_idx, num_values);
+                    let (mut indices_data_block, dictionary_data_block) = dict.into_parts();
+                    // TODO: https://github.com/lancedb/lance/issues/4809
+                    // If we compute stats on dictionary_data_block => panic.
+                    // If we don't compute stats on indices_data_block => panic.
+                    // This is messy.  Don't make me call compute_stat ever.
+                    indices_data_block.compute_stat();
+                    return Self::encode_miniblock(
+                        column_idx,
+                        &field,
+                        compression_strategy.as_ref(),
+                        indices_data_block,
+                        repdef,
+                        row_number,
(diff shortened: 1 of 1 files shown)


Xuanwo (approved):
Thank you for this fix!

wjones127:
Does this close #6184?

westonpace:
> Does this close https://github.com/lance-format/lance/issues/6184?

Impossible to say for certain without a concrete reproducer but given that the other case is also many empty lists I think we can say yes.

I actually got a slightly different error but I think both would be fixed by this root cause.  In my case the rep levels were so verbose they exceeded u16.  However, I think the #6184 error would also be possible if the rep levels were verbose enough to exceed 16KiB but not u16.

PR #6242, opened 2026-03-20
ci: add dev artifact builds for python and java

Xuanwo:
This adds a separate GitHub Actions workflow to build temporary Python and Java dev artifacts from any ref without touching the existing release state machine. The workflow builds in a temporary source copy, applies dev-only version overrides there, and uploads the resulting artifacts for testing.

This keeps the current beta / RC / stable flows unchanged while making it easier to validate branch-specific changes before release.

### DIFF
--- a/.github/workflows/dev-artifacts.yml
+++ b/.github/workflows/dev-artifacts.yml
@@ -1,0 +1,434 @@
+name: Build Dev Artifacts
+
+on:
+  workflow_dispatch:
+    inputs:
+      ref:
+        description: "Git ref to checkout (branch, tag, or SHA)"
+        required: false
+        default: "main"
+        type: string
+      build_python:
+        description: "Build Python dev artifacts"
+        required: true
+        default: true
+        type: boolean
+      build_java:
+        description: "Build Java dev artifacts"
+        required: true
+        default: true
+        type: boolean
+      debug:
+        description: "Keep debug symbols in Python wheels"
+        required: true
+        default: true
+        type: boolean
+
+permissions:
+  contents: read
+
+jobs:
+  python-linux:
+    if: ${{ inputs.build_python }}
+    timeout-minutes: 60
+    name: Python Linux 3.${{ matrix.python-minor-version }} ${{ matrix.config.platform }} manylinux${{ matrix.config.manylinux }}
+    strategy:
+      matrix:
+        python-minor-version: ["9"]
+        config:
+          - platform: x86_64
+            manylinux: "2_17"
+            extra_args: ""
+            runner: ubuntu-22.04
+          - platform: x86_64
+            manylinux: "2_28"
+            extra_args: "--features fp16kernels"
+            runner: ubuntu-22.04
+          - platform: aarch64
+            manylinux: "2_17"
+            extra_args: ""
+            runner: ubuntu-24.04-arm64-4x
+          - platform: aarch64
+            manylinux: "2_28"
+            extra_args: "--features fp16kernels"
+            runner: ubuntu-24.04-arm64-4x
+    runs-on: ${{ matrix.config.runner }}
+    steps:
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
+        with:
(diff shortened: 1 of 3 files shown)


westonpace:
Uploads them where?  Looks like uploads them as GH artifacts?  Are there any storage limitations or costs here?

ci/set_dev_build_version.py
```diff
@@ -0,0 +1,96 @@
...
+
+PYTHON_VERSION_RE = re.compile(r'^version = "([^"]+)"$', re.MULTILINE)
+JAVA_VERSION_RE = re.compile(r"(<version>)([^<]+)(</version>)", re.MULTILINE)
+
+
+def normalize_base_version(version: str) -> str:
+    return version.split("-", 1)[0]
+
+
+def build_dev_version(base_version: str, run_number: str, sha: str, language: str) -> str:
```
jackye1995, line 20: why use `run_number` here? I think it would be more clear to just use the current beta version, like `4.0.0-beta10+g1234abcd`, so we know which source beta version is this coming from

jackye1995:
Nice, I have been thinking about doing this for a long time, thank you for taking this up!

One problem in the current approach is that the rust version is still the beta version, not changed together to also include the github hash. It would be nice to also make that consistent, because the writer version would be stored in the Lance file written. Is that possible to do? If that's too complicated, this is also good enough.

.github/workflows/dev-artifacts.yml
```diff
@@ -0,0 +1,434 @@
...
+          manylinux: ${{ matrix.config.manylinux }}
+          args: "--release ${{ (!inputs.debug) && '--strip' || '' }} ${{ matrix.config.extra_args }}"
+          maturin-version: "1.10.2"
+          before-script-linux: |
+            set -e
+            yum install -y openssl-devel clang \
+              && curl -L https://github.com/protocolbuffers/protobuf/releases/download/v24.4/protoc-24.4-linux-aarch_64.zip > /tmp/protoc.zip \
+              && unzip /tmp/protoc.zip -d /usr/local \
+              && rm /tmp/protoc.zip
+      - name: Upload wheels
```
jackye1995, line 137: I think we can publish this still to fury? maybe time to actually setup the maven one as well

ci/set_dev_build_version.py
```diff
@@ -0,0 +1,96 @@
...
+
+
+def update_java_version(build_root: Path, version: str) -> None:
+    pom_xml = build_root / "java" / "pom.xml"
+    content = pom_xml.read_text()
+    match = JAVA_VERSION_RE.search(content)
+    if match is None:
+        raise ValueError("Could not find java package version")
+    pom_xml.write_text(
+        JAVA_VERSION_RE.sub(lambda match: f"{match.group(1)}{version}{match.group(3)}", content, count=1)
```
jackye1995, line 63: why not just use bump-my-version to do this?

ci/set_dev_build_version.py
Xuanwo: I use an abstracted version of by design. This is because those artifacts could be fetched from an open PR, which is unrelated to any beta version.

.github/workflows/dev-artifacts.yml
Xuanwo: Yes, we need to publish to fury. But I'm not sure, do we have a limitation from fury? Can we setup a lifecycle for fury that all dev artifacts that older than 30 days should be removed?

Xuanwo:
> One problem in the current approach is that the rust version is still the beta version, not changed together to also include the github hash. It would be nice to also make that consistent, because the writer version would be stored in the Lance file written. Is that possible to do? If that's too complicated, this is also good enough.

No, I don't think rust version need a change. Users want to try an ongoing PR can just specify the commit

ci/set_dev_build_version.py
Xuanwo: I don't want a version bump.

To clarify, I'm not trying to do a release. I'm just to build artifacts from a given branch / PR so that we can try and test it.

Xuanwo:
We don't really need this, let's close.

PR #6245, opened 2026-03-20
fix: filter out detached versions when scanning manifests

jackye1995:
I am using detached manifests to do parallel data loading, and hit this bug that the finally committed table cannot be read because it cannot parse the detached manifest names. This is because we parse detached version as manifest scheme v2, and then try to parse version from it and fails. 

I think it's pending more discussion if we should treat detached manifests as a specific scheme, I tried to do that but it looks like the current code has some assumption that the file is either v1 or v2 scheme and detached is just a case within, so refactor gets pretty big and might not be worth it. This PR just do a quick fix so that we can at least open a dataset with detached manifests.

I also added the ability to list detached manifests, the functionality seems missing so far.

### DIFF
--- a/rust/lance-table/src/io/commit.rs
+++ b/rust/lance-table/src/io/commit.rs
@@ -115,6 +115,19 @@ impl ManifestNamingScheme {
         }
     }
 
+    /// Parse a detached version from a filename like `d123456.manifest`.
+    ///
+    /// Returns the full version number with the detached mask bit set.
+    pub fn parse_detached_version(filename: &str) -> Option<u64> {
+        if !filename.starts_with(DETACHED_VERSION_PREFIX) {
+            return None;
+        }
+        let without_prefix = &filename[DETACHED_VERSION_PREFIX.len()..];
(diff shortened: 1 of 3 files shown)


rust/lance-table/src/io/commit.rs
```diff
@@ -428,6 +446,38 @@ fn list_manifests<'a>(
...
+fn detached_manifest_location_from_meta(
+    meta: object_store::ObjectMeta,
+) -> Option<ManifestLocation> {
+    let filename = meta.location.filename()?;
+    let version = ManifestNamingScheme::parse_detached_version(filename)?;
+    Some(ManifestLocation {
+        version,
+        path: meta.location,
+        size: Some(meta.size),
+        naming_scheme: ManifestNamingScheme::V2,
```
jackye1995, line 459: maybe we should add a flag `is_detached` in the `ManifestLocation` struct, but that goes to the topic in the description of if we want to treat it as a naming scheme directly, so keep this unchanged for now.

Issue #6240, opened 2026-03-20
`namespace_exists()` / `table_exists()` throw `RuntimeError` instead of `NamespaceNotFoundError` / `TableNotFoundError`


jorritsandbrink:
As per their docstring, `DirectoryNamespace.namespace_exists()` and `DirectoryNamespace.table_exists()` are supposed to raise `NamespaceNotFoundError` / `TableNotFoundError` when the namespace/table does not exist, however, they return a `RuntimeError` with message _"Failed to downcast NamespaceError source, falling back to runtime error. This may indicate a version mismatch."_.

Repro:

```py
import tempfile

from lance.namespace import DirectoryNamespace, NamespaceExistsRequest, TableExistsRequest


ns = DirectoryNamespace(root=tempfile.mkdtemp())
ns.namespace_exists(NamespaceExistsRequest(id=["nonexistent_namespace"]))
# ns.table_exists(TableExistsRequest(id=["nonexistent_table"]))
```

Result:

```sh
[2026-03-20T15:20:33Z WARN  lance::error] Failed to downcast NamespaceError source, falling back to runtime error. This may indicate a version mismatch. Source type: "Namespace 'nonexistent_namespace' not found"
Traceback (most recent call last):
  File "/home/j/repos/dlt-3546/tmp.py", line 7, in <module>
    ns.namespace_exists(NamespaceExistsRequest(id=["nonexistent_namespace"]))
    ~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/j/.cache/uv/archive-v0/HiBtUgMOpG-E8F-FVIxL_/lib/python3.13/site-packages/lance/namespace.py", line 351, in namespace_exists
    self._inner.namespace_exists(request.model_dump())
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^
RuntimeError: Namespace error: Namespace 'nonexistent_namespace' not found, /home/runner/work/lance/lance/rust/lance-namespace-impls/src/dir/manifest.rs:1722:17
```

Using `pylance==3.0.1`.

PR #6252, opened 2026-03-22
fix: namespace_exists/table_exists raise NamespaceNotFoundError/TableNotFoundError

lilei1128:
namespace_exists() and table_exists() were raising RuntimeError instead of
  the documented NamespaceNotFoundError / TableNotFoundError because the
  implementations in dir.rs and manifest.rs used Error::namespace_source(String)
  — a plain boxed string — which cannot be downcast to NamespaceError in
  infer_error(), triggering the fallback RuntimeError path.

  Fix by constructing proper NamespaceError variants:
  - manifest.rs namespace_exists: NamespaceError::NamespaceNotFound
  - manifest.rs table_exists: NamespaceError::TableNotFound
  - dir.rs namespace_exists: NamespaceError::Unsupported (child NS without manifest)
  - dir.rs table_exists: NamespaceError::TableNotFound

  Also fix rest_adapter.rs error_to_response to dispatch on NamespaceError
  error codes instead of fragile string matching, and rest.rs HTTP helpers to
  parse the numeric error code from the response body and reconstruct the
  correct NamespaceError variant, so the full REST round-trip also raises the
  right Python exception.

  Update tests to assert TableNotFoundError / NamespaceNotFoundError specifically.

### DIFF
--- a/rust/lance-namespace-impls/src/dir.rs
+++ b/rust/lance-namespace-impls/src/dir.rs
@@ -1348,9 +1348,16 @@ impl LanceNamespace for DirectoryNamespace {
         // Atomically check table existence and deregistration status
         let status = self.check_table_status(&table_name).await;
 
-        if !status.exists || status.is_deregistered {
+        if !status.exists {
             return Err(NamespaceError::TableNotFound {
                 message: format!("'{}'", table_name),
+            }
+            .into());
+        }
+
+        if status.is_deregistered {
+            return Err(NamespaceError::TableNotFound {
+                message: format!("'{}' is deregistered", table_name),
             }
             .into());
         }
--- a/python/python/tests/test_namespace_dir.py
+++ b/python/python/tests/test_namespace_dir.py
@@ -262,14 +262,14 @@ class TestTableOperations:
         memory_namespace.table_exists(exists_req)
 
     def test_table_not_exists(self, memory_namespace):
-        """Test checking if a non-existent table exists raises TableNotFoundError."""
+        """Test checking if a non-existent table exists."""
         # Create parent namespace
         create_ns_req = CreateNamespaceRequest(id=["workspace"])
         memory_namespace.create_namespace(create_ns_req)
 
         exists_req = TableExistsRequest(id=["workspace", "nonexistent"])
 
-        with pytest.raises(TableNotFoundError):
+        with pytest.raises(Exception):
             memory_namespace.table_exists(exists_req)
 
     def test_drop_table(self, memory_namespace):
@@ -291,7 +291,7 @@ class TestTableOperations:
 
         # Verify table no longer exists
         exists_req = TableExistsRequest(id=["workspace", "test_table"])
-        with pytest.raises(TableNotFoundError):
+        with pytest.raises(Exception):
             memory_namespace.table_exists(exists_req)
 
     def test_deregister_table(self, temp_namespace):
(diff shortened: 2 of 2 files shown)


lilei1128:
close #6240

rust/lance-namespace-impls/src/dir/manifest.rs
```diff
@@ -2239,9 +2240,10 @@ impl LanceNamespace for ManifestNamespace {
         if self.manifest_contains_object(&object_id).await? {
             Ok(())
         } else {
-            Err(Error::namespace_source(
-                format!("Namespace '{}' not found", object_id).into(),
-            ))
+            Err(NamespaceError::NamespaceNotFound {
+                message: format!("Namespace '{}' not found", object_id),
+            }
+            .into())
```
wjones127, line 2246: suggestion: This message duplicates what is in the `Display` implementation for this error:

https://github.com/lance-format/lance/blob/baa13abc15feae141e22fe4a6f9c7a02106cc88d/rust/lance-namespace/src/error.rs#L180-L182

The error will print as:

```
Namespace not found: Namespace '{name}' not found
```

Consider just putting the quoted object id as the message:

```suggestion
            Err(NamespaceError::NamespaceNotFound {
                message: format!("'{}'", object_id)
            }
            .into())
```

Then the message would be like:

```
Namespace not found: '{name}'
```

rust/lance-namespace-impls/src/dir.rs
```diff
@@ -1346,15 +1349,17 @@ impl LanceNamespace for DirectoryNamespace {
         let status = self.check_table_status(&table_name).await;
 
         if !status.exists {
-            return Err(Error::namespace_source(
-                format!("Table does not exist: {}", table_name).into(),
-            ));
+            return Err(NamespaceError::TableNotFound {
+                message: format!("Table does not exist: {}", table_name),
+            }
+            .into());
```
wjones127: suggestion: similar to namespace error comment, this message duplicates the same text. Consider shortening it to:
```suggestion
            return Err(NamespaceError::TableNotFound {
                message: format!("'{}'", table_name),
            }
            .into());
```

python/python/tests/test_namespace_rest.py
```diff
@@ -236,14 +237,14 @@ def test_table_exists(self, rest_namespace):
...
-        """Test checking if a non-existent table exists."""
+        """Test checking if a non-existent table exists raises TableNotFoundError."""
         # Create parent namespace
         create_ns_req = CreateNamespaceRequest(id=["workspace"])
         rest_namespace.create_namespace(create_ns_req)
 
         exists_req = TableExistsRequest(id=["workspace", "nonexistent"])
 
-        with pytest.raises(Exception):
+        with pytest.raises(TableNotFoundError):
```
wjones127: praise: I'm very happy to see us asserting more specific errors here 😄

lilei1128: Thanks! Happy to help make our error handling clearer 😊

rust/lance-namespace-impls/src/dir/manifest.rs
lilei1128: Thank you for your correction.

wjones127:
I've received a request from @hamersaw to merge #6275 in favor of this PR, as it contains these changes plus some more. I'll close this out, but thank you for working on this. It was reviewing this I was able to see the duplicate info in the errors more easily 🙌

Issue #5654, opened 2026-01-07
File writer panics on unsupported types


westonpace:
There are still a few data types we do not support (FSL<List<...>>, FSL<FSL<...>>, etc.)  The file writer currently panics if it encounters a type it does not recognize.  This should be an error as it is possible for users to trigger panics with bad input otherwise.

PR #6253, opened 2026-03-22
fix: return errors for unsupported fixed-size-list child types

myandpr:
## Summary
- reject unsupported `FixedSizeList` child combinations during schema/type conversion
- return schema errors instead of letting these cases reach later panic paths
- replace the old panic-based tests with a small matrix of unsupported `FixedSizeList` child cases

## Implementation
This keeps the fix at the schema/type conversion boundary instead of deeper in the writer stack.

Unsupported `FixedSizeList` child types are now rejected while converting Arrow `DataType` into Lance `LogicalType`. Supported paths remain unchanged, including primitive nested FSL and the existing `FixedSizeList<Struct>` version gating.

## Testing
- cargo test -p lance-encoding unsupported_fsl_child_types -- --nocapture
- cargo test -p lance-encoding test_fixed_size_list_struct_requires_v2_2 -- --nocapture

### DIFF
--- a/rust/lance-core/src/datatypes.rs
+++ b/rust/lance-core/src/datatypes.rs
@@ -123,6 +123,17 @@ fn timeunit_to_str(unit: &TimeUnit) -> &'static str {
     }
 }
 
+fn is_supported_fixed_size_list_child(data_type: &DataType, nested: bool) -> bool {
+    match data_type {
+        DataType::Struct(_) => !nested,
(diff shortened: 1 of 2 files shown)


myandpr:
Hey @westonpace , this should address #5654 by returning errors for unsupported FixedSizeList child combinations instead of panicking. 

Would appreciate your feedback when you have a chance. Thanks!

myandpr:
@wjones127 Thanks for the review — appreciate it!🙌

PR #6256, opened 2026-03-23
feat: add OpenDAL Azdls backend for abfss:// with use_opendal flag

burlacio:
When use_opendal=true is set in storage options, abfss:// URLs now use OpenDAL's Azdls service (Azure Data Lake Storage Gen2 DFS API) instead of MicrosoftAzureBuilder. This provides native HNS capabilities like rename and create_dir that are not available via the Blob API.

The az:// scheme continues to use OpenDAL's Azblob when use_opendal=true.

Changes:
- Add opendal/services-azdls feature to azure feature flag (Cargo.toml)
- Refactor build_opendal_azure_store to route abfss->Azdls, az->Azblob
- Add tests: Azdls provider selection, Azblob assertion, capability comparison

### DIFF
--- a/rust/lance-io/src/object_store/providers/azure.rs
+++ b/rust/lance-io/src/object_store/providers/azure.rs
@@ -521,12 +521,11 @@ mod tests {
 
         // Build az:// operator (uses Azblob backend)
         let az_url = Url::parse("az://test-container/path").unwrap();
-        let az_operator = AzureBlobStoreProvider::build_opendal_operator(&az_url, &common_opts)
-            .unwrap();
+        let az_operator =
+            AzureBlobStoreProvider::build_opendal_operator(&az_url, &common_opts).unwrap();
 
         // Build abfss:// operator (uses Azdls backend)
-        let abfss_url =
-            Url::parse("abfss://testfs@testaccount.dfs.core.windows.net/data").unwrap();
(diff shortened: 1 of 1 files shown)


burlacio:
> ## PR Review
> Clean, well-structured change. A few observations:
> 
> **Minor concern — unknown keys passed to OpenDAL operators:** The `config_map` is cloned from `storage_options` which includes `use_opendal` and potentially other non-OpenDAL keys. These are passed directly into `Operator::from_iter::<Azdls>(config_map)` / `Operator::from_iter::<Azblob>(config_map)`. If OpenDAL silently ignores unknown keys this is fine (and it's existing behavior for `az://`), but worth verifying this doesn't cause unexpected warnings or failures with the Azdls service specifically.
> 
> **Tests look good** — the capability comparison test (`test_azdls_capabilities_differ_from_azblob`) is a nice way to validate that the correct backend is selected without needing live infra.
> 
> No P0/P1 issues found. LGTM.

I could try a builder approach instead of the config map. wdyt @Xuanwo ?
- https://docs.rs/opendal/latest/opendal/services/struct.Azblob.html#via-builder
- https://docs.rs/opendal/latest/opendal/services/struct.Azdls.html#via-builder

Xuanwo:
> I could try a builder approach instead of the config map. wdyt @Xuanwo ?

Following the same pattern as other code is enough to me. Let's go

PR #6261, opened 2026-03-23
feat: public APIs for trivial column access (raw byte ranges)

wkalt:
Add APIs that let downstream consumers read raw value bytes directly from the file without going through the decoder pipeline, when the column encoding is simple enough (non-nullable, fixed-width, flat, uncompressed).

New APIs:
- ColumnInfo::is_trivially_decodable() — true when on-disk bytes match the in-memory representation. Handles MiniBlock, FullZip, and legacy v2.0 flat encodings.
- ColumnInfo::trivial_page_descriptors() — returns per-page metadata (buffer offsets, bits_per_value, has_large_chunk) for locating raw data in the file.
- miniblock_raw_data_ranges() — parses miniblock chunk metadata words to produce (offset, length, num_rows) tuples pointing past headers to raw value data.
- PageInfo::miniblock_layout() / full_zip_layout() — extract the specific layout variant from the encoding protobuf.
- CachedFileMetadata::column_index(name) — maps a column name to its index in column_infos, handling FSL child resolution.

These enable a speedup for brute-force KNN when vector data is cached in memory, by allowing the caller to reinterpret cached pages as &[f32] without Arrow allocation or DataFusion plan execution.

### DIFF
--- a/rust/lance-encoding/src/decoder.rs
+++ b/rust/lance-encoding/src/decoder.rs
@@ -355,6 +355,246 @@ impl ColumnInfo {
             .map(|page| page.encoding.is_structural())
             .unwrap_or(false)
     }
+
+    /// True when on-disk bytes are the same as the in-memory representation:
+    /// non-nullable, fixed-width, uncompressed flat encoding with no dictionary.
+    /// Callers can safely reinterpret the raw bytes as typed values.
+    ///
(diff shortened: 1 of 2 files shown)


wkalt:
this is proof of concept for a fast path to KNN scanning in-memory lance data pages. Not expected to merge or be the approach we ultimately take.

PR #6263, opened 2026-03-23
fix: preserve row last-updated metadata for rewrite-columns updates

### LINKED ISSUE #6000: [Bug] FileFragment.update_columns does not update last_updated_at_version_meta

myandpr:
Closes #6000

## Summary

Fix `_row_last_updated_at_version` for `RewriteColumns` updates on stable-row-id datasets.

Previously, `update_columns` could advance the dataset version while leaving `_row_last_updated_at_version` stale for rewritten fragments.

The original bug was that `FileFragment.update_columns` rewrote column data but did not propagate enough row-level information for commit to refresh `last_updated_at_version_meta` correctly.

## Approach

This change makes rewrite-columns version refresh source-aware:

- `update_columns` now records matched physical row offsets and carries them through commit as transient fragment state
- commit refreshes row-version metadata only for actually updated fragments
- if partial row offsets are present, it does a partial per-row refresh
- if upstream already computed precise `last_updated_at_version_meta` (for example merge-insert rewrite-columns), commit preserves it
- otherwise, `RewriteColumns` falls back to full-fragment refresh

This keeps the existing API unchanged and limits the change to commit-time metadata handling plus the transient fragment metadata needed to carry row offsets across Rust, Python, and Java round-trips.

## Tests

Added coverage for:

- `fragment.update_columns` partial row refresh
- non-zero fragment partial refresh
- rewrite-columns fallback without transient row offsets
- untouched fragments in multi-fragment datasets staying unchanged
- merge-insert rewrite-columns preserving precise per-row last-updated metadata
- Python JSON round-trip preserving pending rewrite offsets
- Java fragment round-trip preserving pending rewrite offsets

## Validation

```bash
cargo test -p lance test_fragment_update -- --nocapture
python -m pytest python/python/tests/test_fragment.py -k "last_updated_version or rewrite_columns_without_pending_offsets or untouched_fragments or non_zero_fragment or preserves_last_updated_version_across_json"
python -m pytest python/python/tests/test_dataset.py -k merge_insert_partial_rewrite_columns_preserves_last_updated_versions
```

### DIFF
--- a/java/lance-jni/src/fragment.rs
+++ b/java/lance-jni/src/fragment.rs
@@ -5,7 +5,7 @@ use arrow::array::{RecordBatch, RecordBatchIterator, StructArray};
 use arrow::ffi::{FFI_ArrowArray, FFI_ArrowSchema, from_ffi_and_data_type};
 use arrow::ffi_stream::{ArrowArrayStreamReader, FFI_ArrowArrayStream};
 use arrow_schema::DataType;
-use jni::objects::{JIntArray, JValue, JValueGen};
+use jni::objects::{JIntArray, JLongArray, JValue, JValueGen};
 use jni::{
     JNIEnv,
     objects::{JObject, JString},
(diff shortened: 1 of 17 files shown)


myandpr:
note: Only linux-build is failing now, in the distributed IVF/PQ test under rust/lance/src/index/vector/ivf/v2.rs(`test_distributed_vector_build_commits_multiple_segments_and_preserves_query_results::case_2_ivf_pq`). This looks unrelated to the row-version metadata changes in this PR.

myandpr:
This CI failure is caused by an existing flaky test on , not by the change in this PR.

I opened a fix here: #6268

The flaky test was introduced by #6220 and has already failed on  as well:
- https://github.com/lance-format/lance/actions/runs/23450834811
- https://github.com/lance-format/lance/actions/runs/23460892697

The fix makes the test probe all IVF partitions before comparing Top-K row ids, which removes the probing-related nondeterminism while keeping the same strict equality assertion.

Will merge #6268 first, then this PR should be able to rerun cleanly.

myandpr:
This CI failure is caused by an existing flaky test on main, not by the change in this PR.

I opened a fix here: #6268

The flaky test was introduced by #6220 and has already failed on main as well:
- https://github.com/lance-format/lance/actions/runs/23450834811
- https://github.com/lance-format/lance/actions/runs/23460892697

The fix makes the test probe all IVF partitions before comparing Top-K row ids, which removes the probing-related nondeterminism while keeping the same strict equality assertion.

Will merge #6268 first, then this PR should be able to rerun cleanly.

myandpr:
Note: The current CI failure looks unrelated to the changes in this PR. It is failing in a distributed IVF test on linux-arm due to the same flaky exact Top-K row-id assertion pattern we saw before, not in the row-version metadata paths touched here. 

I’ve opened a follow-up PR to fix it: #6281.

myandpr:
Hey @wjones127 @jackye1995 could you take a look this PR when you have a chance? 

It fixes _row_last_updated_at_version handling for rewrite-columns updates on stable-row-id datasets, including merge-insert-related cases. 

Any feedback would be much appreciated!

PR #6265, opened 2026-03-24
feat: support hamming clustering

jackye1995:
Add support for SIMD accelerated pairwise hamming distance computation, and the ability to compute a cluster of binary vectors that are within a given hamming distance threshold, these are considered similar or potentially duplicated vectors of the original representation. 

Also expose the feature in python for easy consumption.


rust/lance-linalg/src/distance/hamming.rs
```diff
@@ -86,6 +96,640 @@ pub fn hamming_distance_arrow_batch(
...
+                result.push(id_a, id_b, dist);
+            }
+        }
+    }
+
+    result
+}
+
+/// Extract u64 hashes from a FixedSizeList<UInt8, 8> Arrow array.
+pub fn extract_hashes_from_fixed_list(array: &FixedSizeListArray) -> Result<Vec<u64>> {
```
wjones127, line 443: issue(blocking): this function ignores `array.offset()` and nulls. I think you should at least handle offset, and error if there are nulls.

rust/lance-linalg/src/distance/hamming.rs
```diff
@@ -86,6 +96,640 @@ pub fn hamming_distance_arrow_batch(
...
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+/// Compute hamming distances for a query against multiple targets.
+/// Uses SIMD acceleration when available.
+#[inline]
+pub fn hamming_batch_u64(query: u64, targets: &[u64], results: &mut [u32]) {
+    debug_assert_eq!(targets.len(), results.len());
```
wjones127, line 177: issue(blocking): The safety of the below functions relies on this assertion. Make this a real assertion instead to avoid out-of-bound writes.


```suggestion
    assert_eq!(targets.len(), results.len());
```

rust/lance/src/index/vector/hamming.rs
```diff
@@ -0,0 +1,939 @@
...
+            )));
+        }
+    }
+
+    // Get total row count
+    let total_rows: usize = dataset
+        .get_fragments()
+        .iter()
+        .filter_map(|f| f.metadata().physical_rows)
+        .sum();
```
wjones127, line 297: issue(blocking): if you use physical rows here, you will be sampling deleted rows too. That's likely a bad idea. This also means the offsets passed later to `take()` will be out-of-bounds.

Instead, use `count_rows`:


```suggestion
    let total_rows: usize = dataset.count_rows(None).await?;
```

wjones127 (changes requested):

Xuanwo:
Closing this because #7379 was merged and covers Hamming clustering across the Rust implementation and Python bindings. Please reopen if this PR contains remaining scope that is not covered by #7379.

PR #6266, opened 2026-03-24
feat: add aimd throttled object store

### LINKED ISSUE #6237: Implement a generic AIMD rate limit throttle
## Parent Issue

Part of [OSS-636](https://linear.app/lancedb/issue/OSS-636/feature-request-implement-aimd-additive-increasemultiplicative) (Implement AIMD Adaptive Retry)

## Summary

Implement a generic AIMD (Additive Increase / Multiplicative Decrease) rate limit throttle as a standalone, reusable component. This should be a leaky bucket (or potentially GCRA) throttle whose rate is controlled by the AIMD algorithm.

## Details

The throttle should support:

* **Configurable parameters**: initial rate, min rate, max rate, reduction factor, increase increment, adjust window, max attempts
* **Rate adjustment logic**:
  * **Multiplicative Decrease**: When throttled, reduce rate by the reduction factor (e.g., halve it), enforcing a minimum floor
  * **Additive Increase**: When successful, slowly increase rate by a fixed increment, enforcing an optional ceiling
* **Failure categorization**: Classify errors as throttling/congestion (triggers rate decrease), recoverable (retry only), or unrecoverable (fail immediately)
* **Window-based decisions**: Track request/throttle counts over a configurable window to decide when to adjust rates

The simplest knob to use may be the number of concurrent requests (concurrency), but a rate limit with an AIMD-controlled leaky bucket throttle is proposed. An AIMD-controlled concurrency limit would also be acceptable.

## Open Questions

* Should rate limits be global across all requests, or should there be independent limits for read (including head), write, and list operations?

### LINKED ISSUE #6238: Create a proxy ObjectStore that applies the AIMD throttle
## Parent Issue

Part of [OSS-636](https://linear.app/lancedb/issue/OSS-636/feature-request-implement-aimd-additive-increasemultiplicative) (Implement AIMD Adaptive Retry)

Depends on [OSS-638](https://linear.app/lancedb/issue/OSS-638/implement-a-generic-aimd-rate-limit-throttle) (Implement a generic AIMD rate limit throttle)

## Summary

Create a proxy/wrapper `ObjectStore` implementation that applies the AIMD rate limit throttle (from [OSS-638](https://linear.app/lancedb/issue/OSS-638/implement-a-generic-aimd-rate-limit-throttle)) to all storage operations. This proxy sits between the application and the actual storage backend, transparently enforcing rate limits.

## Details

The proxy `ObjectStore` should:

* Wrap any existing `ObjectStore` implementation
* Apply the AIMD throttle before forwarding requests to the underlying store
* Detect throttling responses (e.g., HTTP 429) from the backend and feed them back to the AIMD rate adjuster
* Distinguish between error types: throttling errors (should reduce rate), transient errors (should retry without rate change), and permanent errors (should propagate immediately)
* Be configurable per storage account/bucket rather than being a single global throttle

## Context

Currently `object_store` only has `ThrottledStore` (designed for testing, not production) and its retry config is hard-coded to exponential backoff with jitter. The errors from `object_store` are wrapped in `object_store::Error` which is too generic to distinguish 404 from 429. This proxy needs to handle error classification at the right layer.

westonpace:
This does not hook the throttle up anywhere yet, that will come in a future PR.

Closes #6237
Closes #6238

### DIFF
--- a/rust/lance-io/src/object_store/throttle.rs
+++ b/rust/lance-io/src/object_store/throttle.rs
@@ -45,16 +45,19 @@ use tracing::debug;
 ///
 /// However, the builtin object stores all use RetryError when retries are configured and
 /// throttle errors are returned.  Sadly, RetryError is not a public type, so we have to
-/// infer it from the error message.  These error messages currently look like:
+/// infer it from the error message.  This is potentially dangerous because these errors
+/// often include the URI itself and that URI could have any characters in it (e.g. if we
+/// look for 429 then we might match a 429 in a UUID).These error messages currently look like:
 ///
 /// ", after ... retries, max_retries: ..., retry_timeout: ..."
 ///
-/// So, as a crude heuristic, which should work for the builtin object stores, and might
-/// match custom object stores, we simply look for the string "retries" in the error message.
+/// So, as a crude heuristic, which should work for the builtin object stores, but won't
+/// work for custom object stores, we simply look for the string "retries, max_retries"
+/// in the error message.
 pub fn is_throttle_error(err: &object_store::Error) -> bool {
     // Only Generic errors can carry throttle responses
     if let object_store::Error::Generic { source, .. } = err {
-        source.to_string().contains("retries")
+        source.to_string().contains("retries, max_retries")
     } else {
         false
     }


rust/lance-io/src/object_store/throttle.rs
```diff
@@ -0,0 +1,940 @@
...
+use tracing::debug;
+
+use super::WrappingObjectStore;
+
+/// Check whether an `object_store::Error` represents a throttle response
+/// (HTTP 429 / 503) from a cloud object store.
+///
+/// The `object_store` crate surfaces these as `Error::Generic` with the HTTP
+/// status or cloud-specific message embedded in the source chain. We match
+/// against known patterns from S3, GCS, and Azure.
```
wkalt: I wonder if we could upstream an improvement for this

rust/lance-io/src/object_store/throttle.rs
```diff
@@ -0,0 +1,940 @@
...
+pub fn is_throttle_error(err: &object_store::Error) -> bool {
+    // Only Generic errors can carry throttle responses
+    if let object_store::Error::Generic { source, .. } = err {
+        let msg = source.to_string();
+        // Check for common throttle patterns from cloud stores
+        msg.contains("429")
```
wkalt: the bot had a good comment about these. 429 or 503 could occur in the paths of index files or transaction files, so if those paths are also present in the errors then we could make an incorrect inference here.

I wonder if we could find the specific errors for each cloud provider documented. AWS ones seem to be here: https://docs.aws.amazon.com/AmazonS3/latest/API/ErrorResponses.html.

rust/lance-io/src/object_store/throttle.rs
westonpace: Yeah, I've simplified this logic a bit, which should work as long as there are retries configured on the object store.  However, I like the opendal approach, which is roughly the equivalent of adding `is_temporary` to `object_store::Generic`.  I'll work on an upstream PR.

rust/lance-io/src/object_store/throttle.rs
westonpace: Getting to the HTTP response from object store is rather tricky and actually probably impossible due to the fact that most errors are private types hiding behind the `Error` trait so all you can do is grab the error as a string.

I've simplified the logic for now.  We look for the word `retries, max_retries` which is still potentially in a URI but given the whitespace in there I think we're probably ok (I suspect any whitespace in the URI will be %20 or +)  We can worry about a better way after we upstream.

westonpace:
@claude review this PR

rust/lance-io/src/object_store/throttle.rs
rpgreen: Is this something we could contribute up to object_store?

PR #6269, opened 2026-03-24
refactor!: remove staging from distributed vector indexing

Xuanwo:
This refactors distributed vector indexing to remove the staging-root workflow and treat worker outputs as segments written directly under `indices/<segment_uuid>/`. Segment planning and build now operate on segments directly, and vector indexing no longer uses `merge_index_metadata`.

The API and docs are updated around `with_segments(...)` / `plan_segments(...)`, and the focused distributed vector tests were updated to cover the new workflow.

Java's API will be affected

### DIFF
--- a/rust/lance-index/src/vector/bq/storage.rs
+++ b/rust/lance-index/src/vector/bq/storage.rs
@@ -23,7 +23,7 @@ use lance_linalg::simd::{self};
 use lance_table::utils::LanceIteratorExtension;
 use num_traits::AsPrimitive;
 use prost::Message;
-use serde::{Deserialize, Deserializer, Serialize};
+use serde::{Deserialize, Serialize};
 
 use crate::frag_reuse::FragReuseIndex;
 use crate::pb;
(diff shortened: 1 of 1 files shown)


python/python/lance/indices/__init__.py
```diff
@@ -27,13 +27,7 @@ class IndexFileVersion(str, Enum):
     V3 = "V3"
 
 
-class SupportedDistributedIndices(str, Enum):
+class SupportedScalarDistributedIndices(str, Enum):
```
BubbleCal: we don't need this for vectors because it just directly calls the single-node build path right?

rust/lance/src/index/vector/ivf/v2.rs
BubbleCal, line 1: I believe we have a bunch of code which assuming all delta indices (AKA segments in this proposal) share the same centroids, need to carefully check those lines on the `optimize_indices` path

Xuanwo: Yes, we do. All our code path have this assumption now. We need a new design for non-share centroids.

python/python/lance/indices/__init__.py
Xuanwo: Yes, we don't need this anymore.

PR #6270, opened 2026-03-24
feat: clarify logical indices and physical index segments

Xuanwo:
This change makes the logical-index and physical-segment split explicit in the user-facing index APIs without breaking existing behavior. `describe_indices` remains the logical view, `describe_index_segments` becomes the explicit physical-segment view, and index statistics now expose `num_segments` / `segments` alongside the legacy fields for compatibility.

The Rust, Python, and Java bindings now use the same model so segment-aware callers do not need to infer semantics from raw manifest metadata. I validated the Rust path with `cargo test -p lance test_optimize_delta_indices -- --nocapture` and the Java path with `./mvnw -q -Dtest=DatasetTest#testDescribeIndicesByName test`.


rust/lance-index/src/traits.rs
```diff
@@ -77,6 +77,13 @@ pub trait IndexDescription: Send + Sync {
     /// IndexMetadata for each segment of the index.
     fn metadata(&self) -> &[IndexMetadata];
 
+    /// Returns the physical index segments that make up this logical index.
+    ///
+    /// This is an alias for [`Self::metadata`] with a less ambiguous name.
```
wjones127, line 77: praise: +1 I like explicitly calling it segments.

rust/lance-index/src/traits.rs
```diff
@@ -243,6 +252,21 @@ pub trait DatasetIndexExt {
         })
     }
 
+    /// Describe physical index segments.
+    ///
+    /// When `name` is provided, only segments belonging to the named logical
+    /// index are returned. Otherwise, all index segments in the current dataset
+    /// version are returned.
+    async fn describe_index_segments(&self, name: Option<&str>) -> Result<Vec<IndexMetadata>> {
```
wjones127: question(blocking): Users can already call `ds.describe_indices(IndexCriteria::default().for_name(name)).await?.first().map(|idx| idx.segments)` to get this. I worry adding a new method clutters our API. Do you think it's worth adding this method? Is this called commonly enough where it's worth making a simplified API?

I say this in part because I think `segments` are a low-level concept that most users won't care about.

Xuanwo: I think it's a good point. Let's remove it from the public API.

python/python/lance/dataset.py
```diff
@@ -657,9 +659,22 @@ def list_indices(self) -> List[Index]:
         return self._ds.load_indices()
 
     def describe_indices(self) -> List[IndexDescription]:
-        """Returns index information for all indices in the dataset."""
+        """Returns logical index information aggregated across all segments."""
         return self._ds.describe_indices()
 
+    def describe_index_segments(
+        self, index_name: Optional[str] = None
+    ) -> List[IndexSegmentDescription]:
```
wjones127: if we removed in Rust, should we also remove it in Python and Java? Or is there a reason to keep it in these bindings?

Xuanwo: ooops, I overlook them

PR #6275, opened 2026-03-24
feat: actually using namespace error variants

hamersaw:
In the rust SDK we have a `NamespaceError` that contains a message and an error code. Additionally, in both the Java and python SDKs we have logic to parse this type (and error codes) to produce native Java exceptions. The problem, is that in the `namespace-impls` we are never using them. Therefore the rust `NamespaceError` type is never used and the Java / python SDK conversion code the parse and throw native errors is essentially dead code. In this PR, we update the `namespace-impls` to actually throw the correct error types.

Closes https://github.com/lance-format/lance/issues/6240


hamersaw:
Mentioning parallel work that updates to throw on `TableNodeFound` - https://github.com/lance-format/lance/pull/6267

hamersaw:
Another effort to add `TableNotFound` - https://github.com/lance-format/lance/pull/6252

rust/lance-namespace-impls/src/dir/manifest.rs
```diff
@@ -2101,26 +2124,29 @@ impl LanceNamespace for ManifestNamespace {
                 properties: info.metadata,
                 ..Default::default()
             }),
-            None => Err(Error::namespace_source(
-                format!("Namespace '{}' not found", object_id).into(),
-            )),
+            None => Err(NamespaceError::NamespaceNotFound {
+                message: format!("Namespace '{}' not found", object_id),
+            }
+            .into()),
```
wjones127, line 2127: suggestion: This message duplicates what is in the `Display` implementation for this error:

https://github.com/lance-format/lance/blob/baa13abc15feae141e22fe4a6f9c7a02106cc88d/rust/lance-namespace/src/error.rs#L180-L182

The error will print as:

```
Namespace not found: Namespace '{name}' not found
```

Consider just putting the quoted object id as the message:

```suggestion
            Err(NamespaceError::NamespaceNotFound {
                message: format!("'{}'", object_id)
            }
            .into())
```

Then the message would be like:

```
Namespace not found: '{name}'
```

rust/lance-namespace-impls/src/dir/manifest.rs
```diff
@@ -2131,9 +2157,10 @@ impl LanceNamespace for ManifestNamespace {
 
         let object_id = namespace_id.join(DELIMITER);
         if self.manifest_contains_object(&object_id).await? {
-            return Err(Error::namespace_source(
-                format!("Namespace '{}' already exists", object_id).into(),
-            ));
+            return Err(NamespaceError::NamespaceAlreadyExists {
+                message: format!("Namespace '{}' already exists", object_id),
+            }
+            .into());
```
wjones127, line 2160: issue(non-blocking): similar problem here: `message` duplicates the `Display` implementations of `NamespaceError::NamespaceAlreadyExists`:

https://github.com/lance-format/lance/blob/baa13abc15feae141e22fe4a6f9c7a02106cc88d/rust/lance-namespace/src/error.rs#L184-L186

rust/lance-namespace-impls/src/dir/manifest.rs
wjones127: You could also change the error itself to table a table name instead of a custom message.

hamersaw: Great call out. I updated all of these that I could find. But just like replacing all the errors, I'm sure I missed one and will have to come back and fix it :laughing:

PR #6282, opened 2026-03-24
feat: respect RUST_LOG config if available

jackye1995:
This allows user to use RUST_LOG to look at logs from other crates if necessary to make debugging easier.


jackye1995:
We can just alternatively use LANCE_LOG for the purpose. I had the misunderstanding that LANCE_LOG only configures lance crate logging behavior.

Issue #6288, opened 2026-03-25
Optimize merge_partial_vector_auxiliary_files performance for 'non-PQ' index with ShardMergeReader


hushengquan:
When processing the `merge_partial_vector_auxiliary_files` stage for the `ivf-sq` index on a dataset of 5 million records with 8 workers and 2236 partitions, the operation takes an excessively long time (**3780 seconds**), which significantly impacts the overall efficiency of vector index merging.

PR #6289, opened 2026-03-25
perf: use ShardMergeReader for non-PQ index types

hushengquan:
## Problem Description
When processing the `merge_partial_vector_auxiliary_files` stage for the `ivf-sq` index on a dataset of 5 million records with 8 workers and 2236 partitions, the operation takes an excessively long time (**3780 seconds**), which significantly impacts the overall efficiency of vector index merging.(#6288 )

## Root Cause
The current implementation of merging partial vector auxiliary files lacks shard-based parallelism optimization, leading to serial processing bottlenecks when handling a large number of partitions and datasets.

## Proposed Solution
Use `ShardMergeReader` to leverage shard-level parallelism for merging partial vector auxiliary files. This optimization distributes the merge workload across multiple shards, reducing the overall processing time.

## Performance Metrics (Before/After)
| Scenario (8 workers / 2236 partitions / 5M dataset) | Before Optimization | After Optimization | Improvement |
|------------------------------------------------------|---------------------|--------------------|-------------|
| `merge_partial_vector_auxiliary_files` (ivf_sq index)| 3780 seconds        | 208 seconds        | ~18.1x faster |

## Additional Context
- Index type: `ivf_sq`
- Dataset size: 5 million records
- Worker count: 8
- Partition count: 2236
- Optimization core: `ShardMergeReader` (shard-based parallel merge logic)

### DIFF
--- a/rust/lance-index/src/vector/distributed/index_merger.rs
+++ b/rust/lance-index/src/vector/distributed/index_merger.rs
@@ -13,7 +13,6 @@ use arrow_array::{Array, FixedSizeListArray, RecordBatch};
 use futures::StreamExt as _;
 use lance_arrow::{FixedSizeListArrayExt, RecordBatchExt};
 use lance_core::{Error, ROW_ID_FIELD, Result};
-use std::ops::Range;
 use std::sync::Arc;
 
 use crate::IndexMetadata as IndexMetaSchema;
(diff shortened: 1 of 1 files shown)


Xuanwo:
Thank you very much for this PR! I think this change is great, but I want to raise one point: we are also limited by memory usage. A non-PQ index uses more memory than a PQ index, that’s why we didn’t have this feature in the first place. 

IVF_SQ/IVF_FLAT can have large payloads and large partitions. Therefore, I believe we should ensure our new logic works well in multi‑batch and large‑partition scenarios, and make sure we do not store an entire partition in memory before writing it out.

hushengquan:
> Thank you very much for this PR! I think this change is great, but I want to raise one point: we are also limited by memory usage. A non-PQ index uses more memory than a PQ index, that’s why we didn’t have this feature in the first place.
> 
> IVF_SQ/IVF_FLAT can have large payloads and large partitions. Therefore, I believe we should ensure our new logic works well in multi‑batch and large‑partition scenarios, and make sure we do not store an entire partition in memory before writing it out.

Got it, thank you!

PR #6290, opened 2026-03-25
fix: clean up table_version entries on drop_table to fix cascade drop_namespace

XuQianJin-Stars:
## What

Clean up `table_version` entries from the `__manifest` table during `drop_table` to fix cascade `drop_namespace` failures.

## Why

When `table_version_storage_enabled` is `true`, version records are stored in the `__manifest` table with `object_type='table_version'` and `object_id='{table_object_id}${version}'`. Previously, `drop_table` only deleted the table entry itself (matching `object_id` exactly) but left all associated version entries behind.

This caused `drop_namespace` with `Cascade` behavior to fail because it counted these orphaned `table_version` entries as child objects of the namespace — the namespace appeared non-empty even after all tables had been dropped.

### 1. Refactor: Extract `delete_by_filter` helper

The existing `delete_from_manifest` method contained inline logic for executing a delete against the manifest dataset and running subsequent inline optimization. This logic was extracted into a shared private method `delete_by_filter(&self, filter: &str, error_context: &str)` so it can be reused.

### 2. New method: `delete_all_table_versions_for_table`

Added `delete_all_table_versions_for_table(&self, table_object_id: &str)` which constructs a filter:

This matches all version entries for the given table using the `starts_with` predicate on the `object_id` column (since version object IDs are formatted as `{table_object_id}${zero_padded_version}`). The method delegates to `delete_by_filter` and is a no-op when no matching rows exist.

### 3. Call cleanup in both `drop_table` and `deregister_table`

- In `drop_table`: after deleting the table entry from the manifest, `delete_all_table_versions_for_table` is called to remove all version records before deleting the physical data directory.
- In `deregister_table`: same cleanup is performed to prevent orphaned version entries from blocking future `drop_namespace` calls.

### 4. Tests

Added two new parameterized tests (`#[case::with_optimization(true)]` / `#[case::without_optimization(false)]`):

- **`test_drop_table_with_versions_then_drop_namespace`**: End-to-end scenario — creates a namespace, creates a table, inserts a `table_version` entry, drops the table, then verifies `drop_namespace` succeeds (previously it would fail).
- **`test_delete_all_table_versions_for_table`**: Unit-level verification that `delete_all_table_versions_for_table` removes all version entries without affecting the table entry itself.

## Testing

- [x] New unit tests cover the fix scenario
- [x] Both inline-optimization-enabled and disabled paths are tested via `#[rstest]` parameterization

### DIFF
--- a/rust/lance-namespace-impls/src/dir/manifest.rs
+++ b/rust/lance-namespace-impls/src/dir/manifest.rs
@@ -312,6 +312,10 @@ pub struct ManifestNamespace {
     /// Number of retries for commit operations on the manifest table.
     /// If None, defaults to [`lance_table::io::commit::CommitConfig`] default (20).
     commit_retries: Option<u32>,
+    /// Whether table version records are stored as rows in the `__manifest` table.
+    /// When false, drop_table / deregister_table can skip scanning for version rows
+    /// since none are ever written.
+    table_version_storage_enabled: bool,
     /// Serialize manifest mutations within a single namespace instance so concurrent
     /// create/drop calls do not compete with each other on the same in-memory snapshot.
     manifest_mutation_lock: Arc<Mutex<()>>,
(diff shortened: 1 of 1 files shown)


yanghua:
@claude review

yanghua:
@claude review

PR #6293, opened 2026-03-25
feat: move rate limiting to the object store

### LINKED ISSUE #6239: Move rate limiting from ScanScheduler into the object store layer
## Parent Issue

Part of [OSS-636](https://linear.app/lancedb/issue/OSS-636/feature-request-implement-aimd-additive-increasemultiplicative) (Implement AIMD Adaptive Retry)

Depends on [OSS-639](https://linear.app/lancedb/issue/OSS-639/create-a-proxy-objectstore-that-applies-the-aimd-throttle) (Create a proxy ObjectStore that applies the AIMD throttle)

## Summary

Move the existing concurrency-based rate limiting out of `ScanScheduler` and into the object store layer (via the proxy ObjectStore from [OSS-639](https://linear.app/lancedb/issue/OSS-639/create-a-proxy-objectstore-that-applies-the-aimd-throttle)). The I/O buffer rate limiting will remain in `ScanScheduler`.

## Details

The current concurrency throttling in Lance has several weaknesses:

1. **Read-path only**: It is applied through `ScanScheduler`, which only affects the read path and even then does not cover all reads
2. **Misplaced scope**: It is both per-scan and global. The per-scan throttle is too local (why should we care about how many concurrent requests a single scan makes) and the global throttle is too global (what if a user has multiple storage accounts or buckets)
3. **Not universally applicable**: Throttling may not make sense for all object stores (e.g., when using NVMe with extremely high IOPS rates, just trying to obtain a semaphore can be a bottleneck)

By moving rate limiting into the object store layer:

* It will apply to both read and write paths
* It can be scoped per storage account/bucket rather than globally
* It can be disabled for backends where it doesn't make sense (e.g., local NVMe)
* `ScanScheduler` retains only I/O buffer management responsibility, simplifying its role

westonpace:
Closes #6239


westonpace:
Keeping in draft until #6266 merges

rust/lance-io/src/object_store/providers/aws.rs
```diff
@@ -44,12 +45,12 @@ impl AwsStoreProvider {
         storage_options: &StorageOptions,
         is_s3_express: bool,
     ) -> Result<Arc<dyn OSObjectStore>> {
-        let max_retries = storage_options.client_max_retries();
```
Xuanwo, line 47: client_max_retries is not useful anymore, do we need to remove it?

rust/lance-io/src/object_store/providers/aws.rs
Xuanwo, line 1: Seems multipart upload process is not covered. Is it expected?

rust/lance-io/src/object_store/throttle.rs
```diff
@@ -354,7 +461,7 @@ impl AimdThrottledStore {
 impl ObjectStore for AimdThrottledStore {
```
Xuanwo, line 634: I think we should consider overriding our own `rename_if_not_exists` handling.

Currently, the logic retries the entire `copy and delete` process. If the copy succeeds but the delete fails due to throttling, the entire operation is retried. After that, the copy will fail because the object already exists.

rust/lance-io/src/object_store/providers/aws.rs
westonpace: It should still be used.  We actually rely on object_store's retries to make the AIMD throttle work.  The object_store errors have no "is_temporary" so we have no other way of knowing if an error is a temporary error except to see if object_store applied retries to it.

So by default we should do 3 object store retries now.  Each time those fail we apply an AIMD retry (longer backoff, cut throttle).  So in total we get 9 retries instead of 10.

rust/lance-io/src/object_store/throttle.rs
westonpace: I'll create a follow-up, I don't think we use `rename_if_not_exists` in most cases anymore.

westonpace:
Test results:

## Test Description

On Azure, 40 VMs (8 cores each) hammer the same blob with random takes, each take grabbing 1024 rows.  Each VM has 8 processes so there are a total of 320 processes taking from the same blob.  This blob is in a standard class storage account.  This should aggressively trigger rate limiting errors.  Everything runs for 5 minutes.

## With AIMD disabled, previous settings (note, red line is success in this chart for some reason)

8,330 failed takes, 50,176 rows taken

image

## With AIMD enabled, default settings

0 Errors, 9,951,232 rows taken

image

## Aggressive AIMD (1 object_store retry, 5 AIMD retries)

0 Errors, 7,606,272 rows taken

image

(will be updated with additional results)

rust/lance-io/src/object_store/providers/aws.rs
westonpace: Good catch.  I've added multipart to the retry handling.  I still can't apply the outer retry loop to the delete stream or list stream methods.  These return a stream of items and there is no way of mapping results to underlying object store requests so there is nothing I can retry there.  Since these operations are hopefully rare I think it will be ok.

PR #6294, opened 2026-03-25
feat!: support sampling selected fragments

Xuanwo:
This extends `Dataset::sample` to accept optional fragment IDs so grouped training jobs can sample from selected fragments without reimplementing deletion-aware row offset mapping. It rejects empty fragment selections and unknown fragment IDs while preserving the existing behavior for whole-dataset sampling.

### DIFF
--- a/rust/lance/src/dataset/tests/dataset_io.rs
+++ b/rust/lance/src/dataset/tests/dataset_io.rs
@@ -45,8 +45,6 @@ use lance_table::io::manifest::read_manifest;
 use object_store::path::Path;
 use rstest::rstest;
 
-use crate::utils::test::{DatagenExt, FragmentCount, FragmentRowCount};
-
 #[tokio::test]
 async fn test_truncate_table() {
     let tmpdir = tempfile::tempdir().unwrap();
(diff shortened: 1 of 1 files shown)


rust/lance/src/dataset/tests/dataset_io.rs
```diff
@@ -1437,6 +1437,118 @@ async fn test_fast_count_rows(
...
+        &test_uri,
+        Some(WriteParams {
+            max_rows_per_file: 4,
+            max_rows_per_group: 2,
+            data_storage_version: Some(data_storage_version),
+            ..Default::default()
+        }),
+    )
+    .await
+    .unwrap();
```
wjones127, line 1503: suggestion(non-blocking): Could just use `.into_ram_dataset()` to simplify this test:
```suggestion
    let dataset = gen_batch()
        .col("i", array::step::<Int32Type>())
        .into_ram_dataset(FragmentCount::from(2), FragmentRowCount::from(4))
        .await
        .unwrap();
```

rust/lance/src/dataset/tests/dataset_io.rs
```diff
@@ -1437,6 +1437,118 @@ async fn test_fast_count_rows(
...
```
wjones127, line 1536: suggestion(non-blocking): similar here:
```suggestion
    let dataset = gen_batch()
        .col("i", array::step::<Int32Type>())
        .into_ram_dataset(FragmentCount::from(2), FragmentRowCount::from(4))
        .await
        .unwrap();
```

rust/lance/src/dataset/tests/dataset_io.rs
Xuanwo: Thank you, updated.

rust/lance/src/dataset/tests/dataset_io.rs
Xuanwo: Updated.

rust/lance/src/dataset/tests/dataset_io.rs
```diff
@@ -1437,6 +1439,96 @@ async fn test_fast_count_rows(
...
+        err.to_string()
+            .contains("does not accept an empty fragment_ids list")
+    );
+}
+
+#[rstest]
+#[tokio::test]
+async fn test_sample_with_unknown_fragment_ids_rejected(
+    #[values(LanceFileVersion::Legacy, LanceFileVersion::Stable)]
+    _data_storage_version: LanceFileVersion,
```
wjones127: issue: my suggestion messed this up. Sorry! I think it's more important that we test with both, don't you?

Xuanwo: Yes, I have the same feeling. Will change back.

PR #6297, opened 2026-03-25
feat: add storage_options to BasePath

cmccabe:
For Azure, different storage accounts can have buckets (containers) of the same name. Also storage account is the unit of rate limit cap.

When expressed as base, today we try to match base based on the prefix, and that would mean it can only match 1 bucket but not the others if user defines multiple bases of the same buckets in different storage accounts.

This PR adds the ability to define storage_options as a part of the base definition.

### DIFF
--- a/protos/table.proto
+++ b/protos/table.proto
@@ -219,6 +219,10 @@ message BasePath {
   bool is_dataset_root = 3;
   // Note: This absolute path will be directly used by Path:parse(),
   string path = 4;
+  // Storage options relevant to identifying the base path. For example, azure_storage_account_name
+  // goes here. Credentials and options which don't affect the identity of the data being stored
+  // should not be placed here.
+  map<string, string> storage_options = 5;
 }
 
 // Auxiliary Data attached to a version.
--- a/rust/lance/src/dataset/blob.rs
+++ b/rust/lance/src/dataset/blob.rs
@@ -1352,6 +1352,7 @@ mod tests {
                         name: Some("blob_base".to_string()),
                         path: base_uri,
                         is_dataset_root,
+                        storage_options: Default::default(),
                     }]),
                     target_bases: Some(vec![1]),
                     ..Default::default()
@@ -1894,6 +1895,7 @@ mod tests {
                         name: Some("external".to_string()),
                         path: base_uri,
                         is_dataset_root: false,
+                        storage_options: Default::default(),
                     }]),
                     ..Default::default()
                 }),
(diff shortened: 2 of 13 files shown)


rust/lance/src/dataset/utils.rs
```diff
@@ -264,3 +266,112 @@ impl SchemaAdapter {
         ))
     }
 }
+
+pub fn object_store_params_for_base_path<'a>(
+    base_path: &BasePath,
+    input_params: Option<&'a ObjectStoreParams>,
+) -> Cow<'a, ObjectStoreParams> {
+    if let Some(params) = input_params
+        && base_path.storage_options.is_empty()
```
Xuanwo, line 275: if `base_path.storage_options.is_empty()`, we will return the `input_params` directly so that the base_path_prefix doesn't work anymore.

rust/lance/src/dataset/utils.rs
```diff
@@ -264,3 +266,112 @@ impl SchemaAdapter {
...
+    {
+        return Cow::Borrowed(params);
+    }
+    let mut merged_storage_options = base_path.storage_options.clone();
```
Xuanwo, line 279: Hi, can you add a comments for object_store_params_for_base_path to declare the expected merge order? It seems we can to use input_params to override the base path's params.

Xuanwo (commented):
Thank you for working on this, only some questions.

protos/table.proto
```diff
@@ -219,6 +219,10 @@ message BasePath {
   bool is_dataset_root = 3;
   // Note: This absolute path will be directly used by Path:parse(),
   string path = 4;
+  // Storage options relevant to identifying the base path. For example, azure_storage_account_name
+  // goes here. Credentials and options which don't affect the identity of the data being stored
+  // should not be placed here.
+  map<string, string> storage_options = 5;
```
jackye1995, line 225: inconveniently, this is a spec change which means it needs a vote

rust/lance/src/dataset/blob.rs
jackye1995, line 1: In this file, we have a place that needs to be changed around `ExternalBaseResolver::resolve_external_uri`

This becomes interesting with this change, because I can now have multiple bases of the same prefix, and the resolution logic cannot determine which one to use.

I think there is a way to resolve this, by checking the external link's query parameters, for example I can write `az://container/file.jpg?base=1` to indicate the exact base to use. @Xuanwo curious about your opinion on this.

Another alternative commonly used is `az://container@account/...` but that feels too Azure/OCI specific.

cmccabe:
https://github.com/lance-format/lance/discussions/6307

PR #6300, opened 2026-03-26
fix(namespace): support nested types in convert_json_arrow_type

jiaoew1991:
`convert_json_arrow_type` only handled scalar types (int/float/utf8/binary), causing deserialization failures for any schema containing list, struct, large_binary, large_utf8, fixed_size_list, map, decimal, or date/time types.

This made `arrow_type_to_json` and `convert_json_arrow_type` asymmetric: serialization worked for all types but deserialization rejected most of them with "Unsupported Arrow type".

In practice this broke the DuckDB lance extension's fast schema-from-REST path — tables with list/struct columns fell back to opening the S3 dataset for every DESCRIBE, making SHOW ALL TABLES ~20x slower than necessary.

Add support for: float16, large_utf8, large_binary, fixed_size_binary, decimal32/64/128/256, date32, date64, timestamp, duration, list, large_list, fixed_size_list, struct, and map.

Add a roundtrip test covering all supported types.

### DIFF
--- a/rust/lance-namespace/src/schema.rs
+++ b/rust/lance-namespace/src/schema.rs
@@ -268,34 +268,31 @@ pub fn convert_json_arrow_type(json_type: &JsonArrowDataType) -> Result<DataType
         "float32" => Ok(DataType::Float32),
         "float64" => Ok(DataType::Float64),
 
-        // Decimal types
+        // Decimal types - encoding: precision * 1000 + scale
+        // Decoding must handle negative scale: precision = ((encoded + 128) / 1000)
         "decimal32" => {
             let encoded = json_type.length.unwrap_or(0);
-            Ok(DataType::Decimal32(
-                (encoded / 1000) as u8,
-                (encoded % 1000) as i8,
-            ))
+            let precision = ((encoded + 128) / 1000) as u8;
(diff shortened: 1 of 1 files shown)


rust/lance-namespace/src/schema.rs
```diff
@@ -261,10 +264,111 @@ pub fn convert_json_arrow_type(json_type: &JsonArrowDataType) -> Result<DataType
...
+        "float16" => Ok(DataType::Float16),
         "float32" => Ok(DataType::Float32),
         "float64" => Ok(DataType::Float64),
+
+        // Decimal types
+        "decimal32" => {
+            let encoded = json_type.length.unwrap_or(0);
+            Ok(DataType::Decimal32(
+                (encoded / 1000) as u8,
+                (encoded % 1000) as i8,
```
BubbleCal: for decimal the scale can be negative, so we need to restore precision and scale in such way:
```
let precision = ((encoded + 128) / 1000) as u8;
let scale = (encoded - precision as i64 * 1000) as i8;
```

need tests for:
- Decimal32(10, -2)
- Decimal128(9, -2)
- Decimal256(38, 10)

PR #6305, opened 2026-03-26
feat: support segmented inverted index build and search

Xuanwo:
This PR teaches inverted/FTS indices to participate in the segment-based build workflow and to search across multiple committed segments with a shared BM25 scorer. It keeps the current on-disk inverted format intact while aligning FTS with the newer `execute_uncommitted() -> create_index_segment_builder() -> commit_existing_index_segments()` path.

image

image


rust/lance-index/src/scalar/inverted/index.rs
```diff
@@ -455,11 +457,41 @@ impl InvertedIndex {
...
+            for idx in 0..expanded.len() {
+                let token = expanded.get_token(idx);
+                if seen.insert(token.to_string()) {
+                    expanded_tokens.push(token.to_string());
+                    expanded_positions.push(expanded.position(idx));
+                }
+            }
+        }
+        Ok(Tokens::with_positions(
+            expanded_tokens,
```
BubbleCal: this won't work
it would duplicate every token `num_partitions` times right?

rust/lance/src/io/exec/fts.rs
```diff
@@ -289,18 +386,38 @@ impl ExecutionPlan for MatchQueryExec {
...
-                )
-                .boxed()
-                .await?;
+            let tokens = Arc::new(tokens);
+            let params = Arc::new(params);
+            let limit = params.limit.unwrap_or(usize::MAX);
+            let mut candidates = std::collections::BinaryHeap::new();
+            for index in &indices {
+                let (doc_ids, scores) = index
+                    .bm25_search(
```
BubbleCal: should we do concurrent search here?

rust/lance/src/io/exec/fts.rs
```diff
@@ -675,28 +782,33 @@ impl ExecutionPlan for FlatMatchQueryExec {
...
-                .await?;
-            let inverted_idx = match index_meta {
-                Some(index_meta) => {
-                    let uuid = index_meta.uuid.to_string();
-                    let index = ds
-                        .open_generic_index(&column, &uuid, &metrics.index_metrics)
-                        .await?;
-                    index.as_any().downcast_ref::<InvertedIndex>().cloned()
+            let segments = load_segments(&ds, &column).await?;
+            let (tokenizer, base_scorer) = match segments {
```
BubbleCal, line 827: worth parallelizing this!

rust/lance/src/io/exec/fts.rs
```diff
@@ -878,63 +990,75 @@ impl ExecutionPlan for PhraseQueryExec {
...
-                )
-                .boxed()
-                .await?;
+            let tokens = Arc::new(tokens);
+            let params = Arc::new(params);
+            let limit = params.limit.unwrap_or(usize::MAX);
+            let mut candidates = std::collections::BinaryHeap::new();
+            for index in &indices {
+                let (doc_ids, scores) = index
+                    .bm25_search(
```
BubbleCal: and here

wjones127 (commented):
A key benchmark I would recommend doing for this is FTS search over 13 fragments, varying segment size: 1, 2, 4, 6 segments (plus one unindexed fragment). That would help us see how robust the performance is over varying numbers of segments. Ideally we can get a somewhat flat line.

BubbleCal (approved):

PR #6306, opened 2026-03-26
fix: multiple improvements for gh workflows

esteban:


kevinjqliu:
perhaps also revert https://github.com/lance-format/lance/commit/a018e0dfd3561fc6fdc0c3f7f04e4f21d2ff06a9

esteban:
@kevinjqliu Right, I'm considering seriously. Thank you!

.github/workflows/claude-code-review.yml
```diff
@@ -37,11 +30,12 @@ jobs:
 
       - name: Run Claude Code Review
         id: claude-review
-        uses: anthropics/claude-code-action@26ec041249acb0a944c0a47b6c0c13f05dbc5b44 # v1
+        uses: anthropics/claude-code-action@673eb13aa77026be5c507eda12322c1a58b80f0b # v1
+        env:
+          CLAUDE_CODE_SUBPROCESS_ENV_SCRUB: '1'
```
westonpace, line 35: Can you add a comment explaining what this does?

kevinjqliu:
yes! and please rotate the github secrets (in case its already compromised)

.github/workflows/claude-code-review.yml
```diff
@@ -68,5 +62,5 @@ jobs:
           # See https://github.com/anthropics/claude-code-action/blob/main/docs/usage.md
           # or https://code.claude.com/docs/en/cli-reference for available options
           claude_args: |
-            --allowed-tools "Bash(gh issue view:*),Bash(gh search:*),Bash(gh issue list:*),Bash(gh pr comment:*),Bash(gh pr diff:*),Bash(gh pr view:*),Bash(gh pr list:*)"
+            --allowed-tools "Bash(gh pr comment ${{ github.event.pull_request.number }}:*),Bash(gh pr diff ${{ github.event.pull_request.number }}:*),Bash(gh pr view ${{ github.event.pull_request.number }}:*)"
```
westonpace, line 65: Hmm, restricting `comment` seems fine.  The rest are all read-only and it's a public repo right?  Would these be a problem?

That being said I don't really see why it would need to see other PRs or issues.  I guess the only case I can think of is when a PR is built for an issue and references back to it (e.g. "closes 5472 see issue for rationale")

.github/workflows/claude-code-review.yml
esteban: Updated the right version, hash should be 3ac52d0da9f8ec9ca7b4dc23bb477e36ef9c77a9 for 1.0.79: https://github.com/anthropics/claude-code-action/commit/3ac52d0da9f8ec9ca7b4dc23bb477e36ef9c77a9

cmccabe (approved):

.github/workflows/claude-code-review.yml
esteban: The risk is that other PRs/comments could be used for prompt injection. For sure this limits the value of Claude Code to a small context of the PR. 🤷🏻‍♂️

.github/workflows/claude-code-review.yml
westonpace: Oh, I meant, what is `CLAUDE_CODE_SUBPROCESS_ENV_SCRUB`?  I don't want someone turning it off in the future.

kevinjqliu:
please run `uvx --from zizmor zizmor --offline .github/workflows/` see if there are other issues (for this workflow and others) 
And other repos!!

PR #6310, opened 2026-03-27
perf: optimize stable row_id index build from O(rows) to O(fragments)

jiaoew1991:
## Summary

When `enable_stable_row_id` is enabled, the first `take_rows()` call triggers a full `RowIdIndex` build. On large datasets this cold start was extremely slow (18.9s for 968M rows, 220s for 4.26B rows).

### Root Causes

**1. O(total_rows) segment expansion in `decompose_sequence`**

The original code expanded every `U64Segment` element-by-element, even for `Range` segments with no deletions. For a `Range(0..273711)` with no deletions, this meant 273K iterations, deletion vector checks, temporary allocations, and re-compression — only to produce the same Range back. Across 18,243 fragments averaging 233K rows, this totaled **4.26 billion iterations** with ~32 GB of temporary allocations.

**2. O(N²) fragment lookup in `load_row_id_index`**

The original code called `fragments.iter().find()` (O(N) linear search) for each of N fragments, resulting in O(N²) comparisons. `try_join_all` spawned all N futures at once, overwhelming the async runtime. `get_deletion_vector()` was called unconditionally even for fragments without deletion files.

### Solution

**Fix 1: O(1) fast path for Range segments without deletions.** When a fragment has no deletions and its row_id sequence is a `Range`, construct the index chunk directly without iterating.

**Fix 2: HashMap lookup + conditional deletion vector loading.** Use a `HashMap<u32, &FileFragment>` for O(1) lookup, `buffer_unordered` for controlled concurrency, and skip `get_deletion_vector()` when there's no deletion file.

### Results

| Dataset | Before | After | Speedup |
|---------|--------|-------|---------|
| 968M rows, 3,540 fragments | 18.9s | 150ms | **126x** |
| 4.26B rows, 18,243 fragments | 220s | 89ms | **2,471x** |

## Test plan

- [x] All existing rowids tests pass (45 tests)
- [x] Clippy clean
- [x] New test `test_large_range_segments_no_deletions` validates fast path correctness at boundaries and performance (100 fragments × 250K rows completes < 1s)
- [ ] Verify on real datasets with deletions to ensure slow path correctness

### DIFF
--- a/rust/lance-table/src/rowids/index.rs
+++ b/rust/lance-table/src/rowids/index.rs
@@ -650,6 +650,72 @@ mod tests {
         );
     }
 
+    #[test]
+    fn test_large_range_segments_with_deletions() {
+        let rows_per_fragment = 1_000u64;
+        let num_fragments = 10u32;
+        let mut offset = 0u64;
+
+        let fragment_indices: Vec<FragmentRowIdIndex> = (0..num_fragments)
+            .map(|frag_id| {
+                let start = offset;
+                offset += rows_per_fragment;
+
+                // Delete every 3rd row (offsets 0, 3, 6, ...) within each fragment.
+                let mut deleted = roaring::RoaringBitmap::new();
+                for i in (0..rows_per_fragment as u32).step_by(3) {
+                    deleted.insert(i);
+                }
+
+                FragmentRowIdIndex {
+                    fragment_id: frag_id,
+                    row_id_sequence: Arc::new(RowIdSequence(vec![U64Segment::Range(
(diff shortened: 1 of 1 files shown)


rust/lance-table/src/rowids/index.rs
```diff
@@ -547,6 +590,66 @@ mod tests {
         })
     }
 
+    #[test]
+    fn test_large_range_segments_no_deletions() {
```
BubbleCal, line 594: please add a test case that contains deletions

PR #6313, opened 2026-03-27
refactor!: align distributed index build around segments

Xuanwo:
This refactor removes the old `IndexSegment` / segment-builder workflow from distributed vector indexing and aligns the public flow around `IndexMetadata` returned from fragment-level builds. Worker builds now produce uncommitted index metadata, callers optionally merge caller-defined metadata groups, and the final metadata set is committed as one logical index.

```python
# Build uncommitted index metadata
segment_0 = dataset.create_index_uncommitted(...)
segment_1 = dataset.create_index_uncommitted(...)

# Merge segments if needed
merged_segment = dataset.merge_existing_index_segments([
    segment_0,
    segment_1,
])

# Commit segments
dataset.commit_existing_index_segments("vector_idx", "vector", [merged_segment])
```


wjones127:
> ```python
> # Build index segments
> segment_uuid_0 = dataset.create_index_uncommitted(...)
> segment_uuid_1 = dataset.create_index_uncommitted(...)
> 
> # Merge index segments if needed
> merged_uuid = dataset.merge_existing_index_segments([
>     segment_uuid_0,
>     segment_uuid_1,
> ])
> 
> # Commit index segments
> dataset.commit_existing_index_segments("vector_idx", "vector", [merged_uuid])
> ```

issue: this creates a new commit workflow (though `commit_existing_index_segments`) which I worry clutters the API. Have you considered just reusing the existing `dataset.commit()` API?

For example, here's an alternative API for writing two index segments that uses the existing commit workflow:

```python
segment_0: Transaction = dataset.create_index("vector", "IVF_PQ", fragments=[0, 1], committed=False)
segment_1: Transaction = dataset.create_index("vector", "IVF_PQ", fragments=[2, 3], committed=False)

operation = lance.LanceOperation.CreateIndex(...)
transaction = Transaction(segment_0.read_version, operation)

dataset.commit(transaction)
```

Xuanwo:
> ```
> operation = lance.LanceOperation.CreateIndex(...)
> transaction = Transaction(segment_0.read_version, operation)
>
> dataset.commit(transaction)
> ```

Thank you @wjones127! I think this API makes sense. Do you also suggest that we remove `create_index_uncommitted` at the same time?

python/src/indices.rs
```diff
@@ -505,21 +431,42 @@ async fn do_load_shuffled_vectors(
...
+    ds.object_store
+        .put(
+            &sidecar_path,
+            &table_pb::IndexMetadata::from(&sidecar).encode_to_vec(),
+        )
+        .await
+        .infer_error()?;
+    ds.commit_existing_index_segments(index_name, column, vec![index_id])
+        .await
+        .infer_error()?;
```
wjones127: question: is this written out as a method of inter-process communication? If so, this is inconsistent with how we handle that elsewhere. In distributed fragment writes, we return the metadata directly and leave it to the caller to choose how to send the data. They caller can write to object store like this, or they can send via another IPC mechanism. Is there a reason that couldn't be done here too?

wjones127:
> Do you also suggest that we remove `create_index_uncommitted` at the same time?

I'm not as sure about that. We definitely aren't consistent with how we handle uncommitted. For example, we use `LanceFragment.create()` for append:

https://github.com/lance-format/lance/blob/36e8b2d263bbbf606aa599a57c9b6f7cbd7d6664/python/python/tests/test_dataset.py#L1602

But we use `.execute_uncommitted` for merge insert:

https://github.com/lance-format/lance/blob/36e8b2d263bbbf606aa599a57c9b6f7cbd7d6664/python/python/tests/test_dataset.py#L1697-L1701

I think most important would be to re-use `Dataset.commit()` as the main entry point to commit transactions.

python/src/indices.rs
Xuanwo: Yep, we used to return `IndexMetadata` directly but our internal users are not happy with that. They don't want to handle the details and perfer to have APIs more like our old ones where they just call `create_index_uncommitted() -> commit_exising_index()`.

I'm open to provider both ways: we can return `IndexMetdata` and write the sidecar files at the sametime and users can choose how to handle them.

Do you feel that looks better to you?

Xuanwo:
> I think most important would be to re-use `Dataset.commit()` as the main entry point to commit transactions.

I agree with that. 

We are luck of some utils that building `Transaction` from index segments. Do you think it's a good idea for this refactor to move on with `commit_existing_index_segments` (this API replaced the old `commit_existing_index` and not an entire new API).

I'm happy to work as a follow up https://github.com/lance-format/lance/issues/6317 to refactor this part and unify them all in the `Dataset.commit()`.

python/src/indices.rs
wjones127: > I'm open to provider both ways: we can return `IndexMetdata` and write the sidecar files at the sametime

I'd rather not make them pay the IO cost to create that file and then have to clean it up later if the caller isn't going to use it. So I wouldn't want both.

My preference is to keep the API low-level and agnostic to the distributed framework.

We should try to make it easy to use these APIs. If there's something missing for making it easy to write `IndexMetadata` out, then perhaps we can add that. Maybe it would be good if we could make it just two lines:

```python
metadata = create_index_uncommitted()
metadata.write_to("s3://bucket/path")
```

wjones127:
> We are luck of some utils that building `Transaction` from index segments. Do you think it's a good idea for this refactor to move on with `commit_existing_index_segments` (this API replaced the old `commit_existing_index` and not an entire new API).

Yeah, I agree we can do that refactor in a follow up.

rust/lance/src/dataset/index.rs
```diff
@@ -242,18 +242,16 @@ mod tests {
...
-            ),
-            IndexSegment::new(
-                second_segment_uuid,
-                [target_fragments[1].id() as u32],
-                built_index.index_details.clone().unwrap(),
-                built_index.index_version,
-            ),
+            IndexMetadata {
+                uuid: first_segment_uuid,
+                fragment_bitmap: Some(std::iter::once(target_fragments[0].id() as u32).collect()),
```
BubbleCal, line 247: can this work?
```suggestion
                fragment_bitmap: Some(vec![target_fragments[0].id() as u32]),
```

Xuanwo: We are collecting `RoaringBitmap` here, so it doesn't work

PR #6314, opened 2026-03-27
chore: cleanup dead functions and dead_code annotations

Xuanwo:
This PR removes dead functions and narrows a few test-only helpers so the workspace no longer depends on stale `dead_code` suppressions in live paths. It intentionally avoids field-level cleanup and keeps the scope limited to function-oriented removals and warning fixes after re-review.


HaochengLIU:
+1 from my end
feel free to merge it when you feel comfortable

PR #6320, opened 2026-03-27
feat: clean up data files on failed writes

### LINKED ISSUE #6124: Automatic cleanup of data files on failed writes
If we fail on a write, such as an append, we should delete the data files we wrote.

This will already be cleaned up eventually during GC, but it would be nice to have best-effort cleanup immediately on failure.

wjones127:
Previously, if a write operation failed mid-stream, the data files already
written to storage were left as orphans until GC reclaimed them (7+ days by
default).

This adds best-effort cleanup at two levels:

1. `do_write_fragments`: on error in the write loop, deletes all data files
   it opened (both completed and in-progress). External-base files are
   skipped with a warning since their object store isn't available at
   cleanup time.

2. `update` / `merge_insert` `execute_impl`: if `apply_deletions` fails
   after data files were successfully written, cleans up those data files.

Every path we clean up was generated by a writer we opened in the same call,
stored in local-only data structures, so there is no risk of deleting
pre-existing files.

Fixes #6124

## Test plan

- [x] New test: single-file write failure cleans up partial data file
- [x] New test: multi-file write failure (file boundary crossed before error) cleans up all files
- [x] All 183 existing `dataset::write` tests pass

### DIFF
--- a/rust/lance/src/dataset/write.rs
+++ b/rust/lance/src/dataset/write.rs
@@ -3409,4 +3409,51 @@ mod tests {
             "All data files (including completed ones) should be cleaned up on failure"
         );
     }
+
+    /// Verifies the external-base branch in `cleanup_data_fragments`: files with
+    /// `base_id == Some(_)` are skipped (logged but not deleted via the dataset's
+    /// object store), while same-fragment files with `base_id == None` are deleted.
+    #[tokio::test]
+    async fn test_cleanup_data_fragments_skips_external_base() {
+        use lance_core::utils::tempfile::TempStrDir;
+
+        let test_dir = TempStrDir::default();
+        let test_uri = test_dir.as_str();
+
+        let (object_store, base_dir) =
+            ObjectStore::from_uri_and_params(Default::default(), test_uri, &Default::default())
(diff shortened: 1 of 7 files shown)


rust/lance/src/dataset/write.rs
```diff
@@ -535,73 +535,168 @@ pub async fn do_write_fragments(
...
+    if let Err(e) = loop_result {
+        let in_progress_path = writer
+            .as_ref()
+            .filter(|w| w.base_id().is_none())
+            .map(|w| w.path().to_owned());
+        // Drop the writer before deleting: closes any open file handle (some
+        // platforms can't delete open files) and aborts in-progress multipart
+        // uploads via `ObjectWriter::drop`.
+        drop(writer.take());
+        cleanup_partial_write(&object_store, base_dir, &fragments, in_progress_path).await;
```
wjones127: nit: it seems we have to leak a lot of info out of the writer to cleanup a resource the writer creates. How about we make sure the writer always cleans up a resource it created instead?

BubbleCal (approved):

PR #6325, opened 2026-03-28
feat: expose stable row ids

ivscheianu:
Expose the stable row ID manifest flag to Java via `Dataset.usesStableRowIds()`.

Downstream consumers (e.g. lance-spark) need to check whether a dataset was created with stable row IDs enabled. The Rust manifest tracks this flag internally, but it was not accessible from the Java API.

The only workaround was to mirror the flag into `Dataset.getConfig()` via an explicit `UpdateConfig` commit after table creation. This had two consequences:

1. **Unnecessary version bump.** Every table creation with stable row IDs produced two commits instead of one: the initial `CREATE`, followed by a separate `UpdateConfig` commit just to persist the flag in the config map. This means the table starts at version 2 instead of version 1.

2. The above mentioned action will also cause **CDF (Change Data Feed) breakage.** CDF tracks row-level changes across dataset versions, and tests assert exact version numbers for created/updated rows (e.g. "row inserted at version 1, updated at version 3"). The extra version bump shifts every subsequent version by +1, breaking all CDF assertions across the test suite. 

With `Dataset.hasStableRowIds()`, consumers can read the manifest flag directly: no config map mirror needed, no extra commit, no version inflation.

A companion PR on lance-spark (https://github.com/lance-format/lance-spark/pull/351) adds the rest of the needed changes.

**NOTE:** he changes are related, but can be merged independently. No synchronization is needed, they can exist without each other without problems


wjones127 (approved):
This looks reasonable, assuming CI passes.

PR #6326, opened 2026-03-28
fix: sync structured primary key fields after metadata update

### LINKED ISSUE #6324: UpdateConfig.fieldMetadataUpdates does not set protobuf 'unenforced_primary_key' field
**Problem**

When using `UpdateConfig.fieldMetadataUpdates()` (or the deprecated `fieldMetadata()`) to set `lance-schema:unenforced-primary-key=true` and `lance-schema:unenforced-primary-key:position=N` on a field, the metadata key-value pairs are stored correctly in the field metadata map, but the protobuf `unenforced_primary_key` boolean and `unenforced_primary_key_position` fields on the Field message are not populated. As a result, `LanceField.isUnenforcedPrimaryKey()` returns false even though the field metadata contains the correct keys.

**Steps to reproduce**

```
// Create a table with a non-nullable int field (id=0)
Dataset ds = Dataset.create(alloc, uri, schema, writeParams);

// Set PK metadata via UpdateConfig
Map<Integer, UpdateMap> updates = new HashMap<>();
Map<String, String> meta = new HashMap<>();
meta.put("lance-schema:unenforced-primary-key", "true");
meta.put("lance-schema:unenforced-primary-key:position", "1");
updates.put(0, UpdateMap.builder().updates(meta).replace(false).build());

UpdateConfig op = UpdateConfig.builder().fieldMetadataUpdates(updates).build();
Transaction txn = new Transaction.Builder().readVersion(ds.version()).operation(op).build();
Dataset newDs = new CommitBuilder(ds).execute(txn);

// Reopen and check
LanceField field = newDs.getLanceSchema().fields().get(0);
field.getMetadata();                  // → {lance-schema:unenforced-primary-key=true, ...} ✓
field.isUnenforcedPrimaryKey();       // → false ✗
field.getUnenforcedPrimaryKeyPosition(); // → empty ✗
```

**Expected behavior**

After committing an `UpdateConfig` with `lance-schema:unenforced-primary-key` field metadata, `LanceField.isUnenforcedPrimaryKey()`` should return true and `getUnenforcedPrimaryKeyPosition()` should return the correct position. The lance-core library should recognize these well-known metadata keys and translate them to the corresponding protobuf fields, consistent with how it handles them when creating a table from an Arrow schema with the same metadata keys.

**Environment**

- lance-core version: 4.1.0-beta.3
- Java: 17

wombatu-kun:
**Summary**  
Fixes https://github.com/lance-format/lance/issues/6324
 
`UpdateConfig.fieldMetadataUpdates()` correctly persists `lance-schema:unenforced-primary-key` and `lance-schema:unenforced-primary-key:position` in the field's metadata HashMap, but the corresponding structured field `unenforced_primary_key_position` (used for protobuf serialization and `is_unenforced_primary_key()`) was never updated. This caused primary key information set via metadata updates to be lost on the next dataset reload.

**Root cause**  
 `apply_update_map()` in `transaction.rs` only modifies `field.metadata` (a HashMap<String, String>) without syncing the structured `Field.unenforced_primary_key_position` field that drives protobuf serialization and the `is_unenforced_primary_key()`  API.                                                                                                                             

**Fix**
- Added `Field::sync_embedded_metadata()` that re-parses well-known metadata keys (`lance-schema:unenforced-primary-key`, `lance-schema:unenforced-primary-key:position`) and updates the corresponding structured fields.
- Call `sync_embedded_metadata()` after every `apply_update_map()` on field metadata in `Transaction::build_manifest()`.
- Refactored `TryFrom<&ArrowField>` for Field to reuse `sync_embedded_metadata()` instead of duplicating the parsing logic.


rust/lance-core/src/datatypes/field.rs
```diff
@@ -998,6 +998,23 @@ impl Field {
...
+
+    /// Re-parse well-known metadata keys and update the corresponding structured fields.
+    ///
+    /// Call this after modifying `field.metadata` directly (e.g., via UpdateConfig)
+    /// to keep structured fields like `unenforced_primary_key_position` in sync.
+    pub fn sync_embedded_metadata(&mut self) {
+        self.unenforced_primary_key_position = self
+            .metadata
+            .get(LANCE_UNENFORCED_PRIMARY_KEY_POSITION)
+            .and_then(|s| s.parse::<u32>().ok())
```
hamersaw: I think this will silently swallow invalid values. Since this is publicly exposed, it should probably return a `Result` to propagate the parse errors. We could add a few tests the validate this.

wombatu-kun: Fixed and added tests.  
And CI failed for some other reason, i suppose.  
(`FAILED python/tests/test_scalar_index.py::test_create_inverted_index_progress_callback - assert 75 == 150`)

wombatu-kun: @hamersaw all tests pass!

rust/lance-core/src/datatypes/field.rs
```diff
@@ -1124,8 +1136,10 @@ impl TryFrom<&ArrowField> for Field {
             nullable: field.is_nullable(),
             children,
             dictionary: None,
-            unenforced_primary_key_position,
-        })
+            unenforced_primary_key_position: None,
+        };
+        result.sync_embedded_metadata()?;
```
hamersaw: I would prefer a refactoring where we don't create a type with `None` and then immediately update it, but not a blocker.

hamersaw (commented):
Thanks for the PR (and patience on review)! Overall it looks great! I think the testing is a bit excessive and not validating the fix proposed in this PR. I'm wondering if we can have one test for set + sync and then one for merge insert. IMO that would adequately cover this.

wombatu-kun:
Thanks for the review! Trimmed the test suite in dataset::metadata to the two you suggested — test_update_field_metadata_sets_unenforced_primary_key (set + sync: covers both the legacy boolean flag path and explicit :position in one test) and test_update_field_metadata_primary_key_used_by_merge_insert (end-to-end via MergeInsert). Dropped the three parse-error tests.

rust/lance-core/src/datatypes/field.rs
wombatu-kun: Refactored in c4c3d9d2 — extracted a parse_unenforced_primary_key_position helper so TryFrom<&ArrowField> now computes the value before the struct literal, no more None + sync_embedded_metadata() update.

hamersaw (approved):
Looks great! Thanks.

wombatu-kun:
> Looks great! Thanks.

Oh, it seems you don't have write access to this repo. Do you know somebody who has it? Could you tag them please to help with merging?

hamersaw:
Thanks for keeping this from going stale. I think we merged a fix for this issue in https://github.com/lance-format/lance/pull/6706. But IMO refactoring this out to a separate function is a little cleaner and still has merit. Thoughts?

Issue #6322, opened 2026-03-28
Java: Add blob v2 write support with external blob references


jackye1995:
## Summary

Java SDK currently only supports reading blob v2 datasets and writing legacy blob v1 (via `lance-encoding:blob` metadata). We need full blob v2 write support, including external blob references with byte ranges.

### Reading (✅ Supported)
- `Dataset.takeBlobs()` and `Dataset.takeBlobsByIndices()` work with blob v2
- `BlobFile` provides file-like access with `read()`, `seek()`, `tell()`, `size()`

### Writing (⚠️ Legacy Only)
Java tests use legacy blob format:
```java
// TestUtils.java:642-645
private static final String BLOB_META_KEY = "lance-encoding:blob";
private static final String BLOB_META_TRUE = "true";
Field blobField = new Field("blobs",
    new FieldType(true, ArrowType.LargeBinary.INSTANCE, null, 
        Map.of("lance-encoding:blob", "true")),
    Collections.emptyList());
```

### 1. Schema Helpers

```java
import org.lance.schema.BlobField;

// Create a blob v2 field
Field blobField = BlobField.create("video");
Field nullableBlobField = BlobField.nullable("thumbnail");

Schema schema = new Schema(Arrays.asList(
    Field.notNullable("id", new ArrowType.Int(64, true)),
    BlobField.nullable("blob")
));
```

### 2. Blob Value Builder

```java
import org.lance.blob.Blob;
import org.lance.blob.BlobArray;

// Build blob values - supports inline bytes, external URIs, and byte ranges
BlobArray blobs = BlobArray.builder()
    .add(new byte[] {0x01, 0x02, 0x03})                    // inline bytes
    .add("s3://bucket/path/video.mp4")                      // external URI (full file)
    .add(Blob.fromUri("s3://bucket/archive.tar")            // external with byte range
              .position(4096)
              .size(8192))
    .addNull()                                              // null value
    .build(allocator);

// Or from lists
BlobArray blobs = BlobArray.fromBytes(allocator, listOfByteArrays);
BlobArray blobs = BlobArray.fromUris(allocator, listOfUriStrings);
```

### 3. Writing to Dataset (Simple Path)

```java
try (VectorSchemaRoot root = VectorSchemaRoot.create(schema, allocator)) {
    // ... populate id column ...
    
    // Set blob column using BlobArray
    BlobArray blobArray = BlobArray.builder()
        .add(Blob.fromUri("s3://bucket/video1.mp4"))
[...]

Issue #6321, opened 2026-03-28
Support materializing external blob references on write


jackye1995:
## Use Case

When writing blob v2 data with external URI references (e.g., `Blob.from_uri("s3://bucket/file.mp4", position=4096, size=8192)`), users may want Lance to:

1. Fetch the bytes from the external location
2. Store them as inline/packed/dedicated (based on size thresholds)
3. Manage the data fully within the Lance dataset

This is useful when:
- The external data source may be deleted or become unavailable after initial ingestion
- Users want Lance to own and manage all blob data going forward
- Data is being migrated from external storage into Lance-managed storage

## Current Behavior

External blob references are always kept as external (`BlobKind::External`). The URI is stored as metadata, and Lance reads from the external location on access. The data is never copied into Lance storage.

See `BlobPreprocessor::preprocess` in `rust/lance/src/dataset/blob.rs:400-420`:
```rust
if has_uri {
    let (external_base_id, external_uri_or_path) =
        self.resolve_external_reference(uri_val).await?;
    kind_builder.append_value(BlobKind::External as u8);
    // URI stored as reference, data NOT fetched
}
```

## Proposed Solution

Add a `materialize_external_blobs` option to the write path that, when enabled:

1. Fetches bytes from external URIs (respecting `position` and `size` if specified)
2. Routes the fetched data through the normal inline/packed/dedicated logic based on size:
   - ≤ 64KB → Inline
   - 64KB - 4MB → Packed
   - > 4MB → Dedicated
3. Stores the data in Lance-managed storage

### Python API

```python
lance.write_dataset(
    table,
    "./dataset.lance",
    data_storage_version="2.2",
    materialize_external_blobs=True,  # new option
)
```

### Rust API

```rust
WriteParams {
    materialize_external_blobs: true,
    ..Default::default()
}
```

## Considerations

- **Async I/O**: Fetching external blobs should use async object store reads
- **Error handling**: Clear errors when external URIs are unreachable
- **Progress/logging**: For large datasets, provide visibility into fetch progress
- **Parallelism**: Batch fetch multiple external blobs concurrently for performance
- **Partial failures**: Define behavior when some external fetches fail (fail-fast vs continue)

## Workaround

Currently users must fetch external data manually before writing:

```python
def fetch_blob(uri, position=0, size=None):
    # fetch bytes from uri
    ...

data = fetch_blob("s3://bucket/archive.tar", position=4096, size=8192)
[...]

PR #6330, opened 2026-03-30
feat(java): add allowExternalBlobOutsideBases to WriteParams

beinan:
## Summary

Add support for writing blob v2 columns with external URI references that are outside registered base paths. This enables use cases like INSERT INTO SELECT across Lance tables where the target table stores external blob references pointing to the source table's blob files instead of copying the actual blob bytes.

## Changes

- **WriteParams.java**: Add `allowExternalBlobOutsideBases` Optional<Boolean> field, getter, and builder method
- **Fragment.java**: Pass the new field through `createWithFfiArray` and `createWithFfiStream` native methods
- **fragment.rs (JNI)**: Thread the new `Optional<Boolean>` parameter through all fragment creation functions to `extract_write_params`
- **utils.rs (JNI)**: Parse the new parameter and set `allow_external_blob_outside_bases` on Rust `WriteParams`
- **blocking_dataset.rs (JNI)**: Pass `JObject::null()` for the new param in `Dataset.write()` path (not needed there)

## Context

This is a prerequisite for lance-spark blob JOIN support (lance-format/lance-spark#355). When blob data flows through Spark's shuffle during JOIN + INSERT INTO, the target table needs to write external blob references pointing to the source table's physical blob files. The Rust `BlobPreprocessor` already supports this via `allow_external_blob_outside_bases`, but the Java SDK had no way to set it.

Ref: #6321, #6322

## Test plan

- [x] Rust JNI code compiles cleanly (no errors in changed files)
- [ ] Java unit tests (CI)

### DIFF
--- a/java/lance-jni/src/blocking_dataset.rs
+++ b/java/lance-jni/src/blocking_dataset.rs
@@ -631,6 +631,7 @@ fn create_dataset<'local>(
         &storage_options_obj,
         &initial_bases,
         &target_bases,
+        &JObject::null(), // allow_external_blob_outside_bases not used for Dataset.write()
     )?;
 
     // Set up namespace commit handler and storage options provider if namespace is provided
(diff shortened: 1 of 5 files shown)


beinan:
> looks good, pending CI fixes

ci passed

PR #6331, opened 2026-03-30
feat: add timestamps to tag metadata

### LINKED ISSUE #6333: Add tag creation and update timestamps to the Lance format
Currently, tag metadata in the Lance format is enough to resolve a tag to a target version or branch, but it does not expose lifecycle timestamps. For higher-level management services, we want the Lance format itself to provide both tag creation
  time and last update time, so the dataset can serve as the source of truth for tag metadata.

  This is useful for management workflows such as:

  - showing tag creation and last modification time in APIs and UIs
  - sorting and filtering tags by creation time or recency
  - supporting audit and troubleshooting workflows
  - avoiding the need for an external metadata store just to track tag timestamps

  We should extend tag metadata in the Lance format to include:

  - createdAt: the time when the tag was first created
  - updatedAt: the time when the tag was last updated

  Expected behavior:

  - when a tag is created, both createdAt and updatedAt should be set
  - when a tag is updated to point to a different reference, createdAt should remain unchanged and updatedAt should be refreshed
  - older datasets without these fields should remain readable for backward compatibility
  - for legacy tags that do not have createdAt, the field may remain absent rather than being backfilled with an inferred value

  This metadata should be surfaced consistently to upper-layer integrations and management services so they can consume tag timestamps directly from Lance without reconstructing them from external state.

majin1102:
Add created_at and updated_at into `TagContents`.

Close #6333

### DIFF
--- a/java/src/main/java/org/lance/Tag.java
+++ b/java/src/main/java/org/lance/Tag.java
@@ -36,15 +36,6 @@ public class Tag {
    *
    * <p>Timestamps are system-generated metadata and are not part of the public Java input surface.
    */
-  private Tag(String name, String branch, long version, int manifestSize, Instant updatedAt) {
-    this(name, branch, version, manifestSize, null, updatedAt);
-  }
-
-  /**
-   * Constructor used by JNI when reading tag metadata from native code.
-   *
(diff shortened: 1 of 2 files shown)


majin1102:
squash this PR into #6364

PR #6332, opened 2026-03-30
refactor: remove row-id ordering from vector index merge

Xuanwo:
This removes the `ROW_ID` sort from distributed vector segment finalization. The merge path now preserves incoming batch order instead of canonicalizing partition contents by `ROW_ID`, which avoids the `FixedSizeList` `take` overflow that was hitting very large IVF_PQ partitions.

A focused Rust regression test now covers large `FixedSizeList` partition batches and verifies we preserve input order without triggering the Arrow overflow.


rust/lance/src/index/vector/builder.rs
```diff
@@ -93,19 +94,140 @@ use super::{
 /// If the batch has no ROW_ID column or has fewer than 2 rows, it is
 /// returned unchanged. When sorting, the relative order of rows with the
 /// same ROW_ID is preserved.
-fn stable_sort_batch_by_row_id(batch: &RecordBatch) -> Result<RecordBatch> {
+fn stable_sort_batch_by_row_id(batch: &RecordBatch) -> Result<Vec<RecordBatch>> {
```
BubbleCal: I think we don't need to sort the batches any more today, maybe just removing it?

Xuanwo: Sure, working on it

PR #6341, opened 2026-03-30
fix: avoid full scan for nullable vector fragment sampling

Xuanwo:
This fixes an OOM in vector index training when `fragment_ids` are provided for a nullable FixedSizeList vector column. Instead of materializing all selected training rows and then sampling, the nullable fragment-filtered path now samples in chunks and filters nulls batch by batch, keeping memory bounded by the requested sample size.

### DIFF
--- a/rust/lance/src/index/vector/utils.rs
+++ b/rust/lance/src/index/vector/utils.rs
@@ -2,14 +2,15 @@
 // SPDX-FileCopyrightText: Copyright The Lance Authors
 
 use std::collections::HashSet;
+use std::pin::Pin;
 use std::sync::Arc;
 
 use arrow::array::ArrayData;
 use arrow::datatypes::DataType;
 use arrow_array::new_empty_array;
-use arrow_array::{Array, ArrayRef, FixedSizeListArray, RecordBatch, UInt32Array, cast::AsArray};
+use arrow_array::{Array, ArrayRef, FixedSizeListArray, RecordBatch, cast::AsArray};
 use arrow_buffer::{Buffer, MutableBuffer};
-use futures::StreamExt;
+use futures::{Stream, StreamExt, stream};
 use lance_arrow::DataTypeExt;
 use lance_core::datatypes::Schema;
 use lance_linalg::distance::DistanceType;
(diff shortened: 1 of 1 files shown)


rust/lance/src/index/vector/utils.rs
```diff
@@ -603,6 +618,120 @@ async fn sample_nullable_fsl(
...
+    let mut seen_offsets = HashSet::with_capacity(sample_size_hint.min(num_rows));
+    let mut rng = SmallRng::from_os_rng();
+
+    const TAKE_CHUNK_SIZE: usize = 8192;
+
+    while num_non_null < sample_size_hint && seen_offsets.len() < num_rows {
+        let remaining_needed = sample_size_hint - num_non_null;
+        let request_size = remaining_needed
+            .saturating_mul(2)
+            .max(remaining_needed)
```
wjones127: question: what is this `.max()` doing? I would think `2 * remaining_needed` is always returned here.

rust/lance/src/index/vector/utils.rs
```diff
@@ -427,16 +428,35 @@ async fn sample_training_data(
...
+                num_rows,
+                byte_width,
+                vector_field,
+                fragment_ids,
+            )
+            .await;
+        }
+
+        let batch =
+            scan_all_training_data(dataset, column, is_nullable, Some(fragment_ids)).await?;
```
westonpace: Why do we need to scan all training data in this case?  Should we add a follow-up?

rust/lance/src/index/vector/utils.rs
```diff
@@ -603,6 +618,120 @@ async fn sample_nullable_fsl(
...
 }
 
+/// Sample nullable FixedSizeList vectors from a fragment-filtered dataset
+/// without materializing all selected rows in memory.
+///
+/// `Dataset::sample` can undershoot after null filtering, while the old
+/// `scan_all_training_data` fallback eagerly loaded every selected row. This
+/// helper keeps sampling bounded chunks from the selected fragments, filtering
+/// nulls batch-by-batch until enough training vectors have been collected.
+async fn sample_nullable_fsl_from_fragments(
```
westonpace: Couldn't we just call `sample_nullable_fsl` but with a scan that was based on the given fragments instead of a full scan?  This feels like a DRY violation?

rust/lance/src/index/vector/utils.rs
```diff
@@ -603,6 +618,120 @@ async fn sample_nullable_fsl(
...
+            .min(num_rows - seen_offsets.len());
+        let mut sampled_offsets = if num_rows - seen_offsets.len() <= request_size.saturating_mul(4)
+        {
+            let mut unseen_offsets = (0..num_rows as u64)
+                .filter(|offset| !seen_offsets.contains(offset))
+                .collect::<Vec<_>>();
+            unseen_offsets.shuffle(&mut rng);
+            unseen_offsets.truncate(request_size);
+            seen_offsets.extend(unseen_offsets.iter().copied());
+            unseen_offsets
```
wjones127: nitpick: I worry this branch adds complexity without bringing more performance benefit.

rust/lance/src/index/vector/utils.rs
Xuanwo: covered and fixed.

rust/lance/src/index/vector/utils.rs
Xuanwo: Makes sense, now changed to reuse logic here.

rust/lance/src/index/vector/utils.rs
Xuanwo: Removed.

rust/lance/src/index/vector/utils.rs
Xuanwo: Makes sense, removed.

PR #6342, opened 2026-03-30
feat: support bf16 from pytorch dataset

eddyxu:
# Summary

Support round-trip to use bf16 from PyTorch

Co-authored-by: Claude Opus 4.6 <noreply@anthropic.com>

### DIFF
--- a/rust/lance-index/src/scalar/bitmap.rs
+++ b/rust/lance-index/src/scalar/bitmap.rs
@@ -606,8 +606,7 @@ impl ScalarIndex for BitmapIndex {
         dest_store: &dyn IndexStore,
         _old_data_filter: Option<super::OldIndexDataFilter>,
     ) -> Result<CreatedIndex> {
-        let state = self.load_bitmap_index_state().await?;
-        BitmapIndexPlugin::do_train_bitmap_index(new_data, state, dest_store).await?;
+        BitmapIndexPlugin::streaming_build_and_write(new_data, Some(self), dest_store).await?;
 
         Ok(CreatedIndex {
             index_details: prost_types::Any::from_msg(&pbold::BitmapIndexDetails::default())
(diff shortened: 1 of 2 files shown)


python/python/lance/torch/data.py
```diff
@@ -52,6 +52,29 @@ def __call__(
...
+
+def _bf16_to_tensor(arr: pa.Array) -> torch.Tensor:
+    """Convert a bfloat16 extension array to a torch.bfloat16 tensor.
+
+    Reinterprets the raw bytes as uint16 and views as bfloat16,
+    since they share the same 2-byte memory layout.
+    Null values are replaced with NaN.
+    """
+    storage = arr.storage if isinstance(arr.type, pa.ExtensionType) else arr
+    buf = storage.buffers()[1]
```
westonpace, line 68: Should we do a sanity check that the data type of storage is a 16-bit type at this point?

python/python/lance/torch/data.py
```diff
@@ -52,6 +52,29 @@ def __call__(
...
+    offset = storage.offset * 2  # 2 bytes per bf16 value
+    np_uint16 = np.frombuffer(buf, dtype=np.uint16, count=len(storage), offset=offset)
+    tensor = torch.from_numpy(np_uint16.copy()).view(torch.bfloat16)
```
westonpace: Is the copy here so that the resulting buffer can be mutable?

python/python/lance/torch/data.py
```diff
@@ -52,6 +52,29 @@ def __call__(
...
+    if arr.null_count > 0:
+        null_mask = torch.from_numpy(arr.is_null().to_numpy(zero_copy_only=False))
```
westonpace, line 96: Seems like there should be a way to do this without a copy but maybe not.

python/python/lance/torch/data.py
```diff
@@ -104,22 +127,34 @@ def _to_tensor(
...
             pa.types.is_integer(arr.type)
             or pa.types.is_floating(arr.type)
             or pa.types.is_boolean(arr.type)
         ):
             tensor = torch.from_numpy(arr.to_numpy(zero_copy_only=False))
 
-            if uint64_as_int64 and tensor.dtype == torch.uint64:
+            if (
+                uint64_as_int64 and tensor.dtype == torch.uint64
+            ):  # ← inside numeric branch ✓
```
westonpace: Kind of a strange comment.  I'm not really sure what it means.

westonpace (approved):
Just some nits

python/python/lance/torch/data.py
eddyxu: fixed.

python/python/lance/torch/data.py
eddyxu: asked claude / codex to do double checks, to make this opportunist

PR #6343, opened 2026-03-30
feat: support DataFusion Expr in DeleteBuilder

### LINKED ISSUE #6339: feature: support datafusion expression for row deletions
currently, only SQL string predicate is supported:
https://github.com/lance-format/lance/blob/main/rust/lance/src/dataset/write/delete.rs#L82
the scanner creates a SQL expression filter:
https://github.com/lance-format/lance/blob/main/rust/lance/src/dataset/scanner.rs#L1162
which calls into `parse_filer()` and convert the SQL to datafusion expression:
https://github.com/lance-format/lance/blob/main/rust/lance-datafusion/src/planner.rs#L845

instead, i would suggest passing a datafusion expression directly into the table deletion API, so no conversion is needed (similarly, to how the query filter work, where they support both SQL and Expr)

wjones127:
Previously, `DeleteBuilder` only accepted SQL string predicates. Callers who already had a `datafusion::logical_expr::Expr` (e.g. from query planning) had to serialize it to a string only for it to be reparsed inside.

This adds `DeleteBuilder::from_expr(dataset, expr)` as a companion to the existing `DeleteBuilder::new(dataset, sql)`, following the dual-input pattern already used in `Scanner` (`filter` vs `filter_expr`). Internally both paths store an `ExprFilter` and dispatch to the appropriate scanner method. The `Operation::Delete { predicate }` field serializes the expr via `expr.to_string()` when needed for transaction logging.

Closes #6339

## Test plan
- [x] New `test_delete_with_expr_filter` test verifies expr-based deletion matches the equivalent SQL path
- [x] All existing delete tests pass unchanged

### DIFF
--- a/rust/lance/src/dataset/write/delete.rs
+++ b/rust/lance/src/dataset/write/delete.rs
@@ -2,6 +2,7 @@
 // SPDX-FileCopyrightText: Copyright The Lance Authors
 
 use crate::dataset::rowids::get_row_id_index;
+use crate::dataset::scanner::ExprFilter;
 use crate::{
     Dataset,
     dataset::transaction::{Operation, Transaction},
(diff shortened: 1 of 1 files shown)


yuvalif:
@wjones127 thanks for the fix!
do you plan on adding this API to lancedb as well (https://github.com/lancedb/lancedb/issues/3204) ?

PR #6349, opened 2026-03-30
feat: add proto serialization for ANNIvfPartitionExec and ANNIvfSubIndexExec

LuQQiu:
add proto serialization for ANNIvfPartitionExec and ANNIvfSubIndexExec
so the plan can be serialize and executed by remote distributed worker


protos/ann_ivf.proto
```diff
@@ -0,0 +1,40 @@
...
+  VectorQueryProto query = 1;
+  TableIdentifier table = 2;
+  repeated string index_uuids = 3;
+}
+
+message ANNIvfSubIndexExecProto {
+  VectorQueryProto query = 1;
+  TableIdentifier table = 2;
+  string index_name = 3;
+  repeated string segment_uuids = 4;
```
wkalt: are these called "segment_uuids" and "index_uuids" on the lance side? It would be better to call them "segment_ids"/"index_ids" if possible so we don't need to put the encoding in the name. But if this matches the lance side, ok.

LuQQiu: Good catch, make sure align with the original exec /struct names

PR #6352, opened 2026-03-31
feat: add planned blob reads with source-level coalescing

Xuanwo:
This PR improves blob I/O in two complementary ways: `BlobFile` instances that resolve to the same physical object now share a lazy `BlobSource` and can opportunistically coalesce concurrent reads before handing them to Lance's existing scheduler, and datasets now expose a planned `read_blobs` API for materializing blob payloads directly. It also adds explicit cursor-preserving range reads for `BlobFile` across Rust, Python, and Java, with end-to-end Python coverage for the new API and the edge cases it uncovered.

This keeps the optimization aligned with Lance's existing scheduler model while giving callers a higher-level path for sequential and batched blob access.

## Python example

```python
import lance

dataset = lance.dataset("/path/to/dataset")
blobs = dataset.read_blobs(
    "images",
    indices=[0, 4, 8],
    target_request_bytes=8 * 1024 * 1024,
    max_gap_bytes=64 * 1024,
    max_concurrency=4,
    preserve_order=True,
)

for row_address, payload in blobs:
    print(row_address, len(payload))
```

### DIFF
--- a/python/python/tests/test_memory_leaks.py
+++ b/python/python/tests/test_memory_leaks.py
@@ -5,6 +5,8 @@ from __future__ import annotations
 
 import gc
 import os
+import subprocess
+import sys
 from typing import Callable
 
 import lance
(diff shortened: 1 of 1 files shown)


rust/lance/src/dataset/blob.rs
```diff
@@ -509,29 +517,222 @@ pub async fn preprocess_blob_batches(
...
+                source.drain_pending_reads(scheduler).await;
+            });
+        }
+
+        response_rx.await.map_err(|_| {
+            Error::internal("Blob source read task dropped the response".to_string())
+        })?
+    }
+
+    async fn drain_pending_reads(self: Arc<Self>, scheduler: FileScheduler) {
```
jackye1995, line 826: Correct me if I am wrong, but I think this only drains the reads that are going on concurrently. I think in addition to this, we would like to something that can be more optimized for sequential read use cases, for example we can expose a reader that just gives blobs from a list of row positions.

In that use case, we can further minimize the amount of calls we make to object store. Consider the example that we read blobs at position [0-1024), and the caller is iterating through all blobs,  and each image is like 100KB. Then we will read around 100MB, and technically we could just do minimum 1 request to object store. It becomes completely just limited by how much data we would like to buffer in a single call to object store to determine how much object store calls we make to read all images.

Adding more complexity, suppose the user is taking some range with gaps, e.g. [0, 1020], [1023]. Then there is a decision that the stream reader has to make about if it should still fetch the continuous range in a single call, or split it to a separated call. And there could be a threshold set like max_gap_size to determine this.

If we do that, I think this could be turned to a more controlled way to do even concurrent read. Considering again if we read rows [0, 4096), then instead of trying to read all images concurrently and depend on the file scheduler to collapse the requests, we can just split it to a shard of 4 streams reading continuous 1024 images to match whatever hardware the workflow is run on.

Let me know what you think about that!

Xuanwo: Thank you for the suggestion! I have impelmented a new `read_blobs` API which gives users more control over the blob reading.

PR #6356, opened 2026-03-31
feat: support ingest mode for external blob writes

### LINKED ISSUE #6321: Support materializing external blob references on write

Xuanwo:
This adds `external_blob_mode="ingest"` for blob v2 writes so datasets can import external blob bytes into Lance-managed storage instead of always persisting URI references. The write path now streams external sources through `lance-io` reader streams for packed and dedicated blobs while preserving inline materialization for small payloads. 

Closes #6321.

### DIFF
--- a/python/src/dataset.rs
+++ b/python/src/dataset.rs
@@ -46,9 +46,10 @@ use lance::dataset::{
 };
 use lance::dataset::{ColumnAlteration, ProjectionRequest};
 use lance::dataset::{
-    Dataset as LanceDataset, DeleteBuilder, MergeInsertBuilder as LanceMergeInsertBuilder,
-    ExternalBlobMode, ReadParams, UncommittedMergeInsert, UpdateBuilder, Version, WhenMatched,
-    WhenNotMatched, WhenNotMatchedBySource, WriteMode, WriteParams,
+    Dataset as LanceDataset, DeleteBuilder, ExternalBlobMode,
+    MergeInsertBuilder as LanceMergeInsertBuilder, ReadParams, UncommittedMergeInsert,
+    UpdateBuilder, Version, WhenMatched, WhenNotMatched, WhenNotMatchedBySource, WriteMode,
+    WriteParams,
     fragment::FileFragment as LanceFileFragment,
     progress::WriteFragmentProgress,
     scanner::Scanner as LanceScanner,


rust/lance/src/dataset/blob.rs
```diff
@@ -200,6 +201,71 @@ pub struct BlobPreprocessor {
...
+    source_store_registry: Arc<ObjectStoreRegistry>,
+    source_store_params: ObjectStoreParams,
+}
+
+/// A logical slice of an external blob that can be materialized or streamed into Lance-managed
+/// storage.
+struct ExternalBlobSource {
+    reader: Box<dyn Reader>,
+    start: usize,
+    size: usize,
```
wjones127: issue(non-blocking): This is a pre-existing issue, but using usizes for file ranges means 32-bit systems won't handle large files correctly. I'm not sure how much we care, but worth noting.

wjones127:
It looks like there's a few Windows and linting issues to fix, but otherwise looks good

rust/lance/src/dataset/blob.rs
Xuanwo: got it, let's change it

PR #6359, opened 2026-03-31
feat: support distributed IVF_RQ segment builds

Xuanwo:
This PR wires IVF_RQ into the distributed vector index segment build path so workers can build and commit shard-local RQ segments through the existing distributed indexing workflow. It also adds the missing RQ handling in the distributed auxiliary metadata plumbing and regression coverage for distributed RQ segment build plus auxiliary merge.

Current limitation: IVF_RQ still does not have a pre-trained/shared RQ metadata contract. Different RQ segments can therefore carry different rotations and centroids, so merging multiple IVF_RQ segments into a single physical segment is not a supported workflow yet. I plan to follow up by adding shared/pre-trained RQ metadata so segment merge can be supported safely.

### DIFF
--- a/rust/lance/src/index/vector.rs
+++ b/rust/lance/src/index/vector.rs
@@ -702,11 +702,34 @@ pub(crate) async fn build_distributed_vector_index(
         }
 
         IndexType::IvfRq => {
-            return Err(Error::index(format!(
-                "Build Distributed Vector Index: invalid index type: {:?} \
-            is not supported in distributed mode; skipping this shard",
-                index_type
-            )));
+            let StageParams::RQ(rq_params) = &stages[1] else {
+                return Err(Error::index(format!(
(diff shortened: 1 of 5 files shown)


rust/lance/src/index/vector.rs
```diff
@@ -702,11 +702,34 @@ pub(crate) async fn build_distributed_vector_index(
...
+                shuffler,
+                Some(ivf_params),
+                Some(rq_params.clone()),
+                (),
+                frag_reuse_index,
+            )?
+            .with_ivf(ivf_model)
+            // For distributed shards, keep RQ codes in row-major layout.
+            // A single packing pass is performed in the distributed merge stage.
+            .with_transpose(false)
```
BubbleCal, line 728: this is nice!

PR #6363, opened 2026-03-31
feat(java): support fragment-scoped IVF and PQ training in VectorTrainer

summaryzb:
This PR based on https://github.com/lance-format/lance/pull/6296 which extends the Java `VectorTrainer` API to support fragment-scoped IVF centroid training and residual-based PQ codebook training, enabling distributed index build workflows where training is restricted to specific data fragments rather than the full dataset.


summaryzb:
@Xuanwo PTAL

PR #6364, opened 2026-03-31
feat: add branch/tag metadata maps and tag timestamps

majin1102:
## What Changed

  - Changed branch and tag `metadata` to structured key/value maps.
  - Added tag timestamp fields:
    - `createdAt` / `created_at`
    - `updatedAt` / `updated_at`
  - Exposed the new tag metadata shape consistently in Rust, Python, and Java.
  - Updated docs and tests to match the new ref metadata model.

  ## Rust

  - `TagContents` now includes:
    - `metadata: HashMap<String, String>`
    - `created_at: Option<DateTime<Utc>>`
    - `updated_at: Option<DateTime<Utc>>`
  - Tag creation initializes timestamps.
  - Tag updates preserve `created_at` and refresh `updated_at`.
  - Ref serialization tests now cover:
    - map-based metadata
    - tag timestamps
    - backward compatibility for missing optional fields

  ## Python

  - Python tag/branch metadata now uses dict-based metadata.
  - Tag listing APIs now return:
    - `branch`
    - `version`
    - `created_at`
    - `updated_at`
    - `manifest_size`
    - `metadata`
  - Updated the public `Tag` `TypedDict`.
  - Updated tests for tag metadata and timestamp behavior.
  - Fixed the `Tags.list()` docstring to describe the full returned metadata.

  ## Java

  - Java `Tag` now includes:
    - `Map<String, String> metadata`
    - `Optional<Instant> createdAt`
    - `Optional<Instant> updatedAt`
  - JNI tag listing now passes timestamps and metadata maps into Java.
  - Updated Java tests to validate timestamp and metadata behavior.

  ## Docs

  - Updated the branch/tag format spec to document tag timestamps and map-based metadata.
  - Updated the tags and branches guide examples to show the current tag metadata shape.


docs/src/format/table/branch_tag.md
```diff
@@ -118,4 +119,7 @@ Each tag file is a JSON file with the following fields:
 |-----------------|--------|----------|--------------------------------------------------------------------------|
 | `branch`        | string | Yes      | Branch name being tagged. `null` or absent indicates main branch.        |
 | `version`       | number |          | Version number being tagged within that branch.                          |
+| `createdAt`     | string | Yes      | RFC 3339 timestamp for when the tag was first created.                  |
+| `updatedAt`     | string | Yes      | RFC 3339 timestamp for the latest tag reference update.                 |
```
westonpace, line 123: Since we have `manifest_size` should this be `created_at` and `updated_at` for consistency?

Or actually, looking at the test, I wonder if this is supposed to be `manifestSize` below?

python/src/dataset.rs
```diff
@@ -1609,7 +1609,10 @@ impl Dataset {
         for (tag_name, tag_content) in tags {
             let dict = PyDict::new(py);
             dict.set_item("version", tag_content.version)?;
+            dict.set_item("created_at", tag_content.created_at)?;
```
westonpace, line 1613: Do you also need to set `branch`?

python/python/lance/dataset.py
```diff
@@ -5817,6 +5821,12 @@ def update(
         """
         self._ds.update_tag(tag, reference)
 
+    def replace_metadata(self, tag: str, metadata: Dict[str, str]) -> None:
+        """
+        Replace metadata for an existing tag.
+        """
```
westonpace, line 5832: Can you expand on this a little?  Will this merge in the metadata or completely replace any existing metadata?  Will this cause updated_at to change?

westonpace (approved):
The change seems clean and straightforward.  What is your motivation for branch / tag metadata?  Arguably this might be a spec change (requiring 3 votes) though it isn't actually modifying any .proto files.

majin1102:
> The change seems clean and straightforward. What is your motivation for branch / tag metadata? Arguably this might be a spec change (requiring 3 votes) though it isn't actually modifying any .proto files.

Thanks for driving this forward.

For reference, the motivation thread is here:
https://github.com/lance-format/lance/discussions/6361

And the vote thread is here:
https://github.com/lance-format/lance/discussions/6374
Would you mind voting there as well?

PR #6365, opened 2026-03-31
feat: compile-time feature-gated Rust backtrace for JNI error diagnostics

summaryzb:
No regression for release builds, Python layer also can reuse this mechanism

The overall design employs a two-tiered strategy of "compile-time switching + runtime gating":

1. **Compile-time switching:**  Two versions of the `MaybeBacktrace` type are defined through conditional compilation using `#[cfg(feature = "backtrace")]`, when the feature is disabled error size, layout, and runtime behavior are unchanged as before

2. **Runtime Gating:** With the `backtrace` feature enabled, actual backtrace capture is still controlled by the runtime environment variable `RUST_BACKTRACE=1`

3. **Java integration:** The transformation logic calls `err.backtrace()` to obtain an optional backtrace. If exists, it is  passed to the corresponding Java exception class. Users can enable this feature during Maven builds using `-Drust.features=backtrace`.

### DIFF
--- a/java/lance-jni/src/error.rs
+++ b/java/lance-jni/src/error.rs
@@ -181,29 +181,36 @@ impl std::fmt::Display for Error {
 
 impl From<LanceError> for Error {
     fn from(err: LanceError) -> Self {
+        let backtrace_suffix = err
+            .backtrace()
+            .map(|bt| format!("\n\nRust backtrace:\n{}", bt))
+            .unwrap_or_default();
+        let message = format!("{}{}", err, backtrace_suffix);
+
         match &err {
             LanceError::DatasetNotFound { .. }
             | LanceError::DatasetAlreadyExists { .. }
             | LanceError::CommitConflict { .. }
-            | LanceError::InvalidInput { .. } => Self::input_error(err.to_string()),
-            LanceError::IO { .. } => Self::io_error(err.to_string()),
-            LanceError::Timeout { .. } => Self::timeout_error(err.to_string()),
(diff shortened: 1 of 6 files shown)


summaryzb:
@hamersaw PTAL

java/lance-jni/src/error.rs
```diff
@@ -175,28 +175,35 @@ impl std::fmt::Display for Error {
 
 impl From<LanceError> for Error {
     fn from(err: LanceError) -> Self {
+        let backtrace_suffix = err
+            .backtrace()
+            .map(|bt| format!("\n\nRust backtrace:\n{}", bt))
+            .unwrap_or_default();
+        let message = format!("{}{}", err, backtrace_suffix);
+
```
wjones127, line 189: nitpick: it's a little sad to see all the error fields collapsed into a string, makes it harder to introspect errors and format them nicely. Pre-existing issue, so not worth fixing in this PR, but want to point it out.

java/lance-jni/Cargo.lock
```diff
@@ -3187,6 +3187,17 @@ version = "3.0.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "8bb03732005da905c88227371639bf1ad885cc712789c011c31c5fb3ab3ccf02"
 
+[[package]]
+name = "io-uring"
+version = "0.7.11"
```
wjones127: question(non-blocking): why is this being pulled in here now?

wjones127 (approved):
This looks good. If you can rebase and fix CI, I'd be happy to re-review and merge.

PR #6366, opened 2026-03-31
feat: add cuVS accelerator support

Xuanwo:
This adds a first Python-side `accelerator="cuvs"` path for `IVF_PQ` without changing the existing `accelerator="cuda"` torch behavior. The new path wires cuVS training outputs back into Lance's existing index build flow and also enables the same backend selection through `IndicesBuilder.prepare_global_ivf_pq`.

The scope is intentionally narrow: build-time training only, `IVF_PQ` only, and no Rust ABI or query-path changes yet. I also added focused Python tests for cuVS dispatch and the new validation rules.

### DIFF
--- a/python/python/lance/dataset.py
+++ b/python/python/lance/dataset.py
@@ -2960,7 +2960,10 @@ class LanceDataset(pa.dataset.Dataset):
                 )
 
             if use_cuvs:
-                from .cuvs import one_pass_train_ivf_pq_on_cuvs
+                from .cuvs import (
+                    one_pass_assign_ivf_pq_on_cuvs,
+                    one_pass_train_ivf_pq_on_cuvs,
+                )
 
                 LOGGER.info("Doing one-pass ivfpq cuVS training")
                 timers["ivf+pq_train:start"] = time.time()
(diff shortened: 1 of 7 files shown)


python/python/lance/dataset.py
```diff
@@ -3229,7 +3263,12 @@ def create_index(
             The number of sub-vectors for PQ (Product Quantization).
         accelerator : str or ``torch.Device``, optional
             If set, use an accelerator to speed up the training process.
-            Accepted accelerator: "cuda" (Nvidia GPU) and "mps" (Apple Silicon GPU).
+            Accepted accelerator:
+
+            - "cuda" or ``torch.device(...)`` for the existing torch-based path
+            - "mps" for Apple Silicon GPU
+            - "cuvs" for the explicit cuVS-based IVF_PQ training path
```
BubbleCal: does this require nvidia GPU? may comment here

BubbleCal (approved):

python/python/lance/dataset.py
Xuanwo: Oh, yes, will update the comments.

PR #6367, opened 2026-03-31
refactor: rename "region" to "shard" in mem_wal implementation

hamersaw:
Avoid confusion with object store regions (e.g., AWS regions) which are unrelated to the MemWAL concept of a unique writer/reader instance.

Closes https://github.com/lance-format/lance/issues/6355


Xuanwo (approved):
Thank you!

PR #6368, opened 2026-04-01
fix: ensure durable writes actually wait for WAL flush

hamersaw:
track_batch_for_wal was returning a pre-resolved watcher instead of the actual BatchDurableWatcher, so durable_write=true never blocked waiting for the WAL flush to complete. Also include close_duration in benchmark timing for accurate end-to-end throughput reporting.

### DIFF
--- a/rust/lance/src/dataset/mem_wal/wal.rs
+++ b/rust/lance/src/dataset/mem_wal/wal.rs
@@ -587,6 +587,39 @@ mod tests {
         assert!(!watcher.is_durable());
     }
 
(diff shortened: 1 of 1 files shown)


hamersaw:
> The change seems fine but it raises a question. Why wasn't this caught by a failing unit test? Do we need unit tests to ensure that the durable write feature is actually working as intended?

I just don't think there was a test. I just noticed it because I turned on durable writes and they were completing in like single ms. I just added a unit test that _should_ catch this, but timing things are tough sometimes.

PR #6371, opened 2026-04-01
perf: reduce bitmap index build to 1 bitmap in RAM

wkalt:
Bitmap index build and update previously materialized every unique value's bitmap in a HashMap before writing to disk. For high-cardinality columns (e.g. URLs with millions of unique values across thousands of fragments) this caused OOM during both initial build and append.

Replace the materialize-then-write pipeline with a streaming merge that processes one value at a time:

- Request sorted input via TrainingOrdering::Values so the scanner delivers (value, row_id) pairs in value order
- Stream through the sorted data, accumulating row_ids for the current value and flushing each bitmap to disk as soon as the value changes
- During append, merge-join the sorted new data with the old index's BTreeMap keys, loading each old bitmap on demand and dropping it after write

### DIFF
--- a/rust/lance-index/src/scalar/bitmap.rs
+++ b/rust/lance-index/src/scalar/bitmap.rs
@@ -606,8 +606,7 @@ impl ScalarIndex for BitmapIndex {
         dest_store: &dyn IndexStore,
         _old_data_filter: Option<super::OldIndexDataFilter>,
     ) -> Result<CreatedIndex> {
-        let state = self.load_bitmap_index_state().await?;
-        BitmapIndexPlugin::do_train_bitmap_index(new_data, state, dest_store).await?;
+        BitmapIndexPlugin::streaming_build_and_write(new_data, Some(self), dest_store).await?;
 
         Ok(CreatedIndex {
             index_details: prost_types::Any::from_msg(&pbold::BitmapIndexDetails::default())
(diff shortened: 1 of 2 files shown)


wkalt:
before and after for 10M unique md5 hashes

bitmap_comparison

rust/lance-index/src/scalar/bitmap.rs
```diff
@@ -618,14 +617,86 @@ impl ScalarIndex for BitmapIndex {
...
+
+    /// Write the current batch to disk.
+    async fn flush(&mut self) -> Result<()> {
+        if self.keys.is_empty() {
+            return Ok(());
+        }
+        let keys_array =
+            ScalarValue::iter_to_array(self.keys.drain(..).collect::<Vec<_>>().into_iter())
+                .unwrap();
+        let mut binary_builder = BinaryBuilder::new();
```
wjones127: suggestion: you could presize this builder:

```suggestion
        let total_size = self.serialized.iter().map(|b| b.len()).sum();
        let mut binary_builder = BinaryBuilder::with_capacity(self.serialized.len(), total_size);
```

Although it might be just as efficient to do:

```rust
let bitmaps_array: BinaryArray = self.serialized.iter().collect();
```

rust/lance-namespace-impls/src/dir/manifest.rs
```diff
@@ -913,7 +913,7 @@ impl ManifestNamespace {
             schema.clone(),
             vec![
                 Arc::new(StringArray::from(object_ids)),
-                Arc::new(StringArray::from(object_types.to_vec())),
+                Arc::new(StringArray::from(object_types.clone())),
```
wjones127: issue: these changes in namespace code seem unrelated and unnecessary

rust/lance-index/benches/inverted.rs
```diff
@@ -201,12 +201,7 @@ fn bench_inverted(c: &mut Criterion) {
         .with_phrase_slop(Some(0));
     let phrase_pairs = sample_words
         .windows(2)
-        .map(|pair| {
-            Arc::new(Tokens::new(
-                pair.iter().map(|s| s.to_string()).collect(),
-                DocType::Text,
-            ))
-        })
+        .map(|pair| Arc::new(Tokens::new(pair.to_vec(), DocType::Text)))
```
wjones127: issue: this seems unrelated?

wkalt: hmm, I wonder if my lint fix triggered on this. Will clean this up thanks.

wjones127 (approved):

PR #6375, opened 2026-04-01
perf: fix O(N²) version column scan with deletion vectors

pengw0048:
## Problem

Scanning `_row_created_at_version` or `_row_last_updated_at_version` is extremely slow on fragments with deletion vectors — **53 seconds for 1M rows** (vs 0.03s for a regular data column on the same table).

This makes the `delta()` API (`get_inserted_rows()` / `get_updated_rows()`) unusable on any table that has had deletions.

## Root Cause

In `apply_row_id_and_deletes()` (`lance-table/src/utils/stream.rs`), the version column is built by:

```rust
sequence.versions()
    .skip(r.start as usize)
    .take((r.end - r.start) as usize)
```

`versions()` creates a fresh iterator from the start of the RLE sequence. For each range in the selection, `skip(r.start)` walks through all prior elements — O(rows) per range. With deletion vectors creating many small ranges, this becomes O(rows × ranges).

## Fix

Replace with `version_values_for_selection()` that:
- **Fast-paths single-run fragments** (common case: all rows same version) with `vec![version; count]` — O(1)
- **Binary search over precomputed run offsets** for O(log(runs)) per position in the multi-run case

## Benchmark

1M rows, 33% deleted, `_row_created_at_version` scan:

| | Before | After |
|---|---|---|
| Time | 53s | 0.02s |
| Complexity | O(rows × ranges) | O(rows × log(runs)) |

## Minimal Reproduction

```python
import lance, pyarrow as pa, numpy as np, time

uri = "/tmp/test_version_col_perf.lance"
lance.write_dataset(
    pa.table({"val": np.random.randint(0, 1000, 1_000_000)}),
    uri, mode="overwrite", enable_stable_row_ids=True,
)
ds = lance.dataset(uri)
ds.delete("val < 333")
ds = lance.dataset(uri)

t0 = time.time()
ds.scanner(columns=["_row_created_at_version"]).to_table()
print(f"_row_created_at_version with deletions: {time.time()-t0:.2f}s")

t0 = time.time()
ds.scanner(columns=["val"]).to_table()
print(f"normal column with deletions: {time.time()-t0:.2f}s")
```

### DIFF
--- a/rust/lance-table/src/utils/stream.rs
+++ b/rust/lance-table/src/utils/stream.rs
@@ -346,9 +346,12 @@ pub fn apply_row_id_and_deletes(
 
         if config.with_row_last_updated_at_version {
             let version_arr = if let Some(sequence) = &config.last_updated_at_sequence {
-                Arc::new(UInt64Array::from(
-                    version_values_for_selection(sequence, &config.params, batch_offset, num_rows)?,
-                ))
+                Arc::new(UInt64Array::from(version_values_for_selection(
+                    sequence,
+                    &config.params,
+                    batch_offset,
+                    num_rows,
+                )?))
             } else {
                 // Default to version 1 if sequence not provided
                 Arc::new(UInt64Array::from(vec![1u64; num_rows as usize]))
@@ -359,9 +362,12 @@ pub fn apply_row_id_and_deletes(
 
         if config.with_row_created_at_version {
             let version_arr = if let Some(sequence) = &config.created_at_sequence {
-                Arc::new(UInt64Array::from(
-                    version_values_for_selection(sequence, &config.params, batch_offset, num_rows)?,
-                ))
+                Arc::new(UInt64Array::from(version_values_for_selection(
+                    sequence,
+                    &config.params,
+                    batch_offset,
+                    num_rows,
+                )?))
             } else {
                 // Default to version 1 if sequence not provided
                 Arc::new(UInt64Array::from(vec![1u64; num_rows as usize]))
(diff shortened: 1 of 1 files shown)


rust/lance-table/src/utils/stream.rs
```diff
@@ -161,6 +161,59 @@ fn apply_deletions_as_nulls(batch: RecordBatch, mask: &BooleanArray) -> Result<R
...
+            Some(start)
+        })
+        .collect();
+    let total_len: usize = sequence.runs.iter().map(|r| r.len()).sum();
+
+    for r in &selection {
+        for pos in r.start..r.end {
+            let pos = pos as usize;
+            if pos >= total_len {
+                versions.push(1); // fallback
```
BubbleCal: i think we will get an error at https://github.com/lance-format/lance/pull/6375/changes#diff-e3ed3f2ff36ff45b28ec19c75b4b9fadcfd857e2ced681ceb22b07fdc4e705b4R360 if we don't have enough rows, but this would hide the error

rust/lance-table/src/utils/stream.rs
```diff
@@ -161,6 +161,59 @@ fn apply_deletions_as_nulls(batch: RecordBatch, mask: &BooleanArray) -> Result<R
...
+                continue;
+            }
+            // Binary search for the run containing this position
+            let run_idx = match run_offsets.binary_search(&pos) {
```
BubbleCal, line 204: need one more test to cover this path

BubbleCal (approved):
Good to go once we have all greens

pengw0048:
@BubbleCal could you approve the workflows again? Fixed the formatting and the windows test failure seemed unrelated.

PR #6376, opened 2026-04-01
feat: support vector query pruning by index segments

Xuanwo:
This change adds `with_index_segments()` for vector queries and makes ANN planning prune to the selected index segments instead of always searching the full logical index. It also makes `with_fragments()` participate in segment selection and flat fallback computation so fragment-filtered and segment-filtered searches stay correct when only part of the logical index is queried. 

This feature will make distributed search much faster to avoid loading not related index segments.

---

FTS should also support this, will add after https://github.com/lance-format/lance/pull/6305 been merged.


rust/lance/src/dataset/scanner.rs
```diff
@@ -1040,6 +1046,14 @@ impl Scanner {
         self
     }
 
+    /// Restrict vector index search to the specified index segments.
+    ///
+    /// This setting is only supported for vector search.
```
westonpace, line 1051: Do we plan to eventually extend this to FTS search as well?

rust/lance/src/dataset/scanner.rs
```diff
@@ -2167,6 +2181,19 @@ impl Scanner {
             }
         }
 
+        if let Some(index_segments) = &self.index_segments {
+            if index_segments.is_empty() {
+                return Err(Error::invalid_input(
+                    "with_index_segments does not accept an empty segment list".to_string(),
+                ));
+            }
```
westonpace: We could probably do this part of the validation on the `with_index_segments` method?  I'm not sure how much we should really worry about eager validation but we do some, for example, when users set a filter.

rust/lance/src/dataset/scanner.rs
```diff
@@ -3421,63 +3448,137 @@ impl Scanner {
...
+                    let use_this_index = match q.metric_type {
+                        Some(user_metric) => {
+                            if user_metric == index_metric {
+                                true
+                            } else {
+                                log::warn!(
+                                    "Requested metric {:?} is incompatible with index metric {:?}, falling back to brute-force search",
+                                    user_metric,
+                                    index_metric
+                                );
```
westonpace: Since this branch is arguably a new path (we should only hit it if the user called `with_index_segments` right?) maybe we can return an error here instead of logging a warning?  We want to eventually return an error on the existing path but don't yet out of fear of disruption / backwards compatibility.

rust/lance/src/dataset/scanner.rs
```diff
@@ -3557,27 +3660,33 @@ impl Scanner {
...
-                .await?;
+        let fallback_fragments = if let Some(target_fragments) = &self.fragments {
+            let indexed_fragments = self.get_indexed_frags(indexed_segments);
+            target_fragments
+                .iter()
+                .filter(|fragment| !indexed_fragments.contains(fragment.id as u32))
+                .cloned()
+                .collect::<Vec<_>>()
+        } else if self.index_segments.is_some() {
+            Vec::new()
```
westonpace, line 3676: So specifying index segments has the same effect as fast_search?  I guess this is probably obvious but can we add a note to the `with_index_segments` method that specifying this will mean that no unindexed data will be searched?

rust/lance/src/dataset/scanner.rs
```diff
@@ -3557,27 +3660,33 @@ impl Scanner {
...
+        } else {
+            self.dataset.unindexed_fragments(index_name).await?
+        };
+
+        if !fallback_fragments.is_empty() {
             let mut q = q.clone();
-            q.metric_type = Some(idx.metric_type());
+            let metric_type = q.metric_type.ok_or_else(|| {
+                Error::internal(
```
westonpace: Why do we have this new error path?  What sequence of actions will encounter this error?

rust/lance/src/dataset/scanner.rs
```diff
@@ -10280,6 +10472,154 @@ full_filter=name LIKE Utf8(\"test%2\"), refine_filter=name LIKE Utf8(\"test%2\")
...
+        test_ds.append_data_with_range(400, 410).await.unwrap();
+        test_ds.append_data_with_range(410, 420).await.unwrap();
+        let fragments = test_ds.dataset.fragments();
+
+        let mut scanner = test_ds.dataset.scan();
+        scanner
+            .nearest("vec", &query, 420)
+            .unwrap()
+            .with_fragments(vec![fragments[0].clone(), fragments[2].clone()])
+            .with_index_segments(vec![segment_ids[0]]);
```
westonpace: Cool, good to see a test that has both specified.

rust/lance/src/dataset/scanner.rs
Xuanwo: Yes, all our index will eventually support FTS.

rust/lance/src/dataset/scanner.rs
Xuanwo: Makes sense. Let's move it.

rust/lance/src/dataset/scanner.rs
Xuanwo: Makes sense, will update.

rust/lance/src/dataset/scanner.rs
Xuanwo: Will update.

rust/lance/src/dataset/scanner.rs
Xuanwo: Over protect from codex and I didn't review it out locally. Will remove it.

Issue #6387, opened 2026-04-02
Support byte-sized batch limits in file reader


westonpace:
## Summary

Allow the batch size to be specified in bytes instead of rows when reading lance files. This gives callers control over memory usage per batch, which is especially important for variable-width data where row count is a poor proxy for memory consumption.

## Approach

Add an `estimate_decoded_bytes` method to the page and field decoder traits. Before each drain, the batch stream queries this to compute a row count that fits the byte budget. A post-decode feedback loop measures actual batch sizes and corrects the estimate for subsequent batches. No file format changes required — all estimates use data already available at decode time.

See [investigation notes](https://github.com/lancedb/lance/blob/main/rust/lance-encoding/AGENTS.md) and worktree at `feat-byte-sized-batches-file-reader` for full design details.

## Tasks

- [ ] Add `batch_size_bytes` config option — wire `batch_size_bytes: Option<u64>` through `SchedulerDecoderConfig` into `BatchDecodeStream`. When set, replaces fixed `rows_per_batch`. When unset, existing row-based behavior unchanged.
- [ ] Modify `BatchDecodeStream::next_batch_task()` for byte-based row selection — query `estimate_decoded_bytes` across children, compute row count that fits the budget, use cross-column coordination loop (estimate → check → adjust → converge).
- [ ] Add post-decode feedback loop — after `into_batch()` measures actual batch size (`decoder.rs:2551`), feed measured bytes-per-row back to refine row count for subsequent batches.
- [ ] Add `estimate_decoded_bytes` to `StructuralPageDecoder` trait — default implementation returns conservative fallback so the system works before all encodings are covered.
- [ ] Add `estimate_decoded_bytes` to `StructuralFieldDecoder` trait — field-level decoder spans pages, knows data type, delegates to page decoders. Struct decoder aggregates across children.
- [ ] Implement estimates for exact fixed-width encodings — Flat, Constant, InlineBitpacking, OutOfLineBitpacking, RLE, ByteStreamSplit, PackedStruct. All `num_rows * known_width`.
- [ ] Implement estimate for General compression (LZ4/Zstd) — read length prefix from compressed frame header (8-byte u64 for Zstd, 4-byte u32 for LZ4) to get exact decompressed size without decompressing.
- [ ] Implement estimate for Dictionary encoding — inspect loaded dictionary DataBlock. Fixed-width: use value width. Variable-width: scan offsets for max value size, bound = `num_rows * max_value_size`.
[...]

PR #6388, opened 2026-04-02
feat: add batch_size_bytes to encoding decode stream

westonpace:
## Summary

This adds some initial framework for #6387 .  It adds the `batch_size_bytes` option which will be wired up in a future PR.  It provides a basic implementation that uses a guess-and-check strategy to try and figure out how many rows to use.  This is inexact and some batches it emits will be too large.  Future PRs will add accurate sizing from the decode layers to hopefully avoid the need for guessing entirely.  Still, this is enough to get the feature wired up and then we can improve it later.

- Add `batch_size_bytes: Option<u64>` to `SchedulerDecoderConfig` and thread it through the structural v2.1 decode path
- When set, compute rows-per-batch from byte estimates instead of the fixed `batch_size` row count
- After each batch decodes, measure actual bytes-per-row and feed it back so subsequent batches converge toward the target byte size
- Feedback degrades gradually (midpoint) when actual size is smaller than the estimate; adopts immediately when larger to avoid OOM
- Only the v2.1+ `StructuralBatchDecodeStream` is modified; v2.0 `BatchDecodeStream` is unchanged (logs a warning if the option is set)
- Wiring this through to the file reader and Python/Java bindings will be done in a follow-up PR

## Test plan

- [x] `test_estimate_bytes_per_row` — unit test for the schema-based byte estimator
- [x] `test_byte_sized_batches_fixed_width` — 1000 rows × 4 Int32 columns, `batch_size_bytes=1600` → 10 batches of exactly 100 rows, roundtrip verified
- [x] `test_byte_sized_batches_none_unchanged` — `batch_size_bytes=None` still uses `rows_per_batch` (no behavioral change)
- [x] `test_byte_sized_batches_feedback_convergence` — 100-byte strings with 64-byte schema estimate; verifies second/third batches converge to ~50 rows after feedback
- [x] `cargo clippy -p lance-encoding --tests -p lance-file -- -D warnings` clean
- [x] `cargo fmt --all -- --check` clean

### DIFF
--- a/rust/lance-encoding/src/decoder.rs
+++ b/rust/lance-encoding/src/decoder.rs
@@ -1693,12 +1693,16 @@ impl<T: RootDecoderType> RecordBatchReader for BatchDecodeIterator<T> {
 /// binary, lists) a rough default is used. The estimate is used as a
 /// starting point when `batch_size_bytes` is set; a post-decode feedback
 /// loop corrects it after the first batch.
+///
+/// This estimate ignores validity bitmaps at the moment.  We can't infer
+/// their presence simply from the data_type and their impact is probably
+/// fairly negligible.
 fn estimate_bytes_per_row(data_type: &DataType) -> f64 {
     if let Some(w) = data_type.byte_width_opt() {
         return w as f64;
     }
     match data_type {
-        DataType::Boolean => 1.0,
+        DataType::Boolean => 1.0 / 8.0,
         DataType::Utf8 | DataType::Binary | DataType::LargeUtf8 | DataType::LargeBinary => 64.0,
         DataType::Struct(fields) => fields
             .iter()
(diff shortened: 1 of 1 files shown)


rust/lance-encoding/src/decoder.rs
```diff
@@ -1682,6 +1683,102 @@ impl<T: RootDecoderType> RecordBatchReader for BatchDecodeIterator<T> {
...
+/// accounting only for the portion of buffers that belongs to the
+/// batch's row range. Unlike `get_array_memory_size()`, this does
+/// not over-count when arrays share a larger underlying page buffer.
+fn batch_data_size(batch: &RecordBatch) -> u64 {
+    batch
+        .columns()
+        .iter()
+        .map(|c| array_data_size(c.as_ref()))
+        .sum()
+}
```
westonpace: I don't like this.  I'm going to make a prequel PR to address getting the size of decoded batches

rust/lance-encoding/src/decoder.rs
```diff
@@ -1686,6 +1687,35 @@ impl<T: RootDecoderType> RecordBatchReader for BatchDecodeIterator<T> {
...
+/// For fixed-width types this is exact. For variable-width types (strings,
+/// binary, lists) a rough default is used. The estimate is used as a
+/// starting point when `batch_size_bytes` is set; a post-decode feedback
+/// loop corrects it after the first batch.
+fn estimate_bytes_per_row(data_type: &DataType) -> f64 {
+    if let Some(w) = data_type.byte_width_opt() {
+        return w as f64;
+    }
+    match data_type {
+        DataType::Boolean => 1.0,
```
wjones127: nitpick: technically the data is bitpacked, so I would think this would be more like `1.0 / 8.0`. But probably fine to leave as 1.0. Might be worth a comment.

westonpace: Good point.  Since we're using floats I went ahead and switched to 1.0 / 8.0.  This also led me to realize we forgot to account for validity bitmaps but I don't see a simple way to do so and this is just an estimate so I just made a comment for now.

PR #6390, opened 2026-04-02
fix: branch_identfier unstable for legacy branches

majin1102:
## Problem

Legacy branches, i.e. branches whose `BranchContents` were written without a persisted `branch_identifier`, currently deserialize through `BranchIdentifier::none()`. That fallback generates a fresh random UUID on each read, so the same unchanged branch can surface a different `branch_identifier` across repeated loads.

This makes branch identity unstable in both Python and Java for legacy datasets. On the Python side, `branches.list()` / `branches_ordered()` expose `branch_identifier` directly, so callers that diff, cache, or snapshot branch metadata can observe false changes even when the branch itself has not changed. On the Java side, the same legacy branch can also appear with a different identifier across refreshes, which makes equality-style comparisons unstable as well.

## Summary
- stabilize fallback branch identifiers for legacy branch metadata by replacing the missing-identifier sentinel with a deterministic synthetic UUID during branch metadata reads
- keep the fallback logic localized to Rust branch metadata loading so Python and Java continues returning stable `branch_identifier` values without API shape changes
- add a lightweight Rust regression test that exercises `BranchContents::from_path` on in-memory branch metadata and verifies stable repeated reads plus distinct identifiers for different branch names

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7968,6 +7968,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "sha1_smol"
+version = "1.0.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "bbfa15b3dddfee50a0fff136974b3e1bde555604ba463834a7eb7deb6417705d"
+
+[[package]]
 name = "sha2"
 version = "0.10.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -9200,6 +9206,7 @@ dependencies = [
  "getrandom 0.4.2",
  "js-sys",
  "serde_core",
+ "sha1_smol",
  "wasm-bindgen",
 ]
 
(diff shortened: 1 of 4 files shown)


majin1102:
@codex review

majin1102:
@codex review

Deniwn22 (commented):
Logic looks correct. One suggestion: the early return on line prevents the cleanup function from running — worth adding a finally block.

majin1102:
> Logic looks correct. One suggestion: the early return on line prevents the cleanup function from running — worth adding a finally block.

Hi， @Deniwn22, thanks for your attention.

This PR has been open for quite a while. There might be some context I don't get. Would you mind explain more on your thought?

majin1102:
@codex review

jackye1995 (approved):
looks good to me!

PR #6391, opened 2026-04-02
feat: thread data_size through decode pipeline

westonpace:
## Summary
- Threads accurate `data_size` (in bytes) from `DataBlock::data_size()` at the encoding layer through the full decode pipeline to the final `RecordBatch`
- Implements `DataBlock::data_size()` for `Struct` and `Dictionary` variants (were `todo!()`)
- Uses the accurate data size for the "batch is too large" warning instead of Arrow's `get_array_memory_size()`, which over-reports due to shared page buffers
- Changes `DecodeArrayTask::decode()` to return `(ArrayRef, u64)` so data size flows through naturally

## Test plan
- [x] All 364 existing `lance-encoding` tests pass
- [x] `cargo clippy -p lance-encoding --tests -- -D warnings` clean
- [x] `cargo clippy -p lance-file --tests -- -D warnings` clean
- [x] `cargo fmt --all -- --check` clean

### DIFF
--- a/rust/lance-encoding/src/decoder.rs
+++ b/rust/lance-encoding/src/decoder.rs
@@ -1472,9 +1472,13 @@ impl BatchDecodeStream {
                     // Real decode work happens inside into_batch, which can block the current
                     // thread for a long time. By spawning it as a new task, we allow Tokio's
                     // worker threads to keep making progress.
-                    tokio::spawn(async move { next_task.into_batch(emitted_batch_size_warning) })
+                    let (batch, _data_size) =
+                        tokio::spawn(
+                            async move { next_task.into_batch(emitted_batch_size_warning) },
+                        )
                         .await
-                        .map_err(|err| Error::wrapped(err.into()))?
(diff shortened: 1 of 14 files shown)


rust/lance-encoding/src/decoder.rs
```diff
@@ -1472,9 +1472,13 @@ impl BatchDecodeStream {
                     // Real decode work happens inside into_batch, which can block the current
                     // thread for a long time. By spawning it as a new task, we allow Tokio's
                     // worker threads to keep making progress.
-                    tokio::spawn(async move { next_task.into_batch(emitted_batch_size_warning) })
+                    let (batch, _data_size) =
```
westonpace, line 1475: I plan on using this in a future PR soon

westonpace:
Ostensibly this PR stands on its own because it makes the warning log message that we print more accurate.  The true reason for the change though is to enable https://github.com/lance-format/lance/pull/6388 to be more accurate.

wjones127 (approved):
I'm excited for byte-size batches!

PR #6393, opened 2026-04-02
fix: preserve multipart part ordering in throttle wrapper

westonpace:
## Summary
- Call `target.put_part` synchronously in `ThrottledMultipartUpload::put_part` to lock in part ordering at creation time, rather than deferring the call into the async future body where await order would determine part order.
- Remove the unnecessary `Arc<Mutex<...>>` wrapper around the inner upload target since `&mut self` already prevents concurrent `put_part` calls.
- Add test `test_throttled_multipart_reorders_parts` that verifies parts are ordered by creation, not by await order.

## Test plan
- [x] `cargo test -p lance-io throttle` — all 21 tests pass

### DIFF
--- a/rust/lance-io/src/object_store/throttle.rs
+++ b/rust/lance-io/src/object_store/throttle.rs
@@ -1608,9 +1608,8 @@ mod tests {
         let fut_a = upload.put_part(PutPayload::from_static(b"AAAA"));
         let fut_b = upload.put_part(PutPayload::from_static(b"BBBB"));
 
-        // Await in REVERSE order. Because ThrottledMultipartUpload defers the
-        // inner put_part call to the future body, fut_b's data reaches the
-        // inner upload first and is assigned the lower index.
+        // Await in REVERSE order. Part ordering should be determined by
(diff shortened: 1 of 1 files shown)


rust/lance-io/src/object_store/throttle.rs
```diff
@@ -1611,4 +1594,40 @@ mod tests {
...
+        let path = Path::from("test/multipart_ordering.bin");
+        let mut upload = throttled.put_multipart(&path).await.unwrap();
+
+        // Create futures for two parts in order: A then B.
+        let fut_a = upload.put_part(PutPayload::from_static(b"AAAA"));
+        let fut_b = upload.put_part(PutPayload::from_static(b"BBBB"));
+
+        // Await in REVERSE order. Because ThrottledMultipartUpload defers the
+        // inner put_part call to the future body, fut_b's data reaches the
+        // inner upload first and is assigned the lower index.
```
wkalt: the comments in the test are no longer accurate with the code changes

westonpace: Fixed

PR #6397, opened 2026-04-03
fix: quote column names in merge_insert join to preserve case sensiti…

YSBF:
…vity

DataFusion's DataFrame::join() lowercases unquoted identifiers, which causes merge_insert to fail when column names contain uppercase letters (e.g. "Stock", "Date"). The join key "target_Stock" is lowercased to "target_stock" internally, but the schema retains "target_Stock", causing a schema mismatch error:

  No field named target_stock. Did you mean '?table?.Stock'?

Fix: wrap join column names in double quotes so DataFusion treats them as quoted identifiers and preserves their original casing. This is consistent with how prefix_columns() already quotes column names when aliasing.

Affects both the FullCompatible and Subschema join paths in create_full_table_joined_stream(). The indexed path (create_indexed_joined_stream) is not affected because it uses Column::new_with_schema() which does direct schema lookup.

### DIFF
--- a/rust/lance/src/dataset/write/merge_insert.rs
+++ b/rust/lance/src/dataset/write/merge_insert.rs
@@ -790,17 +790,21 @@ impl MergeInsertJob {
         let session_ctx = SessionContext::new_with_config(session_config);
         let schema = source.schema();
         let new_data = session_ctx.read_one_shot(source)?;
+        // Quote column names to preserve case sensitivity — DataFusion's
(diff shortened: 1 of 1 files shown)


wjones127 (changes requested):
Hi @YSBF. I thought I had already fixed this in https://github.com/lance-format/lance/pull/5465.

Could you add a unit test that fails without this change?

wjones127:
Hi @YSBF, I'll close this for now. But feel free to re-open if you can reproduce a failure.

PR #6402, opened 2026-04-03
feat: optimize one segmented vector segment per run

Xuanwo:
This changes segmented vector index optimize so the default rebalance path keeps segment boundaries and rewrites only the single worst segment in each run. It builds on #6400's logical vector index / IVF view work and avoids the current behavior where segmented optimize treats the logical index as one physical index.

I also added a regression test that creates a skewed two-segment IVF index and verifies that optimize replaces only the oversized segment while leaving the other segment untouched.

### DIFF
--- a/rust/lance/src/index/vector/ivf.rs
+++ b/rust/lance/src/index/vector/ivf.rs
@@ -336,7 +336,7 @@ pub(crate) fn select_segment_for_single_rebalance(
             .map(|dt| dt.timestamp_millis())
             .unwrap_or(i64::MIN);
 
-        let split_candidate = (split_partition_count > 0).then(|| SegmentRebalanceCandidate {
+        let split_candidate = (split_partition_count > 0).then_some(SegmentRebalanceCandidate {
             segment_id: metadata.uuid,
             score: split_partition_count,
             created_at_ms,
@@ -347,7 +347,7 @@ pub(crate) fn select_segment_for_single_rebalance(
             best_split = Some(candidate);
         }
 
-        let join_candidate = (join_partition_count > 0).then(|| SegmentRebalanceCandidate {
+        let join_candidate = (join_partition_count > 0).then_some(SegmentRebalanceCandidate {
             segment_id: metadata.uuid,
             score: join_partition_count,
             created_at_ms,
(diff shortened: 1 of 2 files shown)


rust/lance/src/index/vector/ivf.rs
```diff
@@ -270,6 +271,106 @@ impl std::fmt::Debug for IVFIndex {
...
+
+        let created_at_ms = metadata
+            .created_at
+            .map(|dt| dt.timestamp_millis())
+            .unwrap_or(i64::MIN);
+
+        let split_candidate =
+            (max_partition_size > split_threshold).then(|| SegmentRebalanceCandidate {
+                segment_id: metadata.uuid,
+                score: max_partition_size - split_threshold,
```
BubbleCal: maybe use `the number of partitions which are over split_threshold` as the score?

rust/lance/src/index/vector/ivf.rs
```diff
@@ -270,6 +271,106 @@ impl std::fmt::Debug for IVFIndex {
...
+            && candidate_is_better(candidate, best_split)
+        {
+            best_split = Some(candidate);
+        }
+
+        let join_candidate =
+            (num_partitions > 1 && min_partition_size < join_threshold).then(|| {
+                SegmentRebalanceCandidate {
+                    segment_id: metadata.uuid,
+                    score: join_threshold - min_partition_size,
```
BubbleCal: same as L346

rust/lance/src/index/vector/ivf.rs
```diff
@@ -270,6 +271,106 @@ impl std::fmt::Debug for IVFIndex {
...
+                    && (candidate.created_at_ms, candidate.segment_id.as_bytes())
+                        < (
+                            current_best.created_at_ms,
+                            current_best.segment_id.as_bytes(),
+                        ))
+        }
+    }
+}
+
+fn index_type_for_segmented_optimize(index: &dyn VectorIndex) -> Result<IndexType> {
```
BubbleCal, line 302: NIT: we have a helper function doing the same thing IIRC

Xuanwo: got it, removed!

PR #6405, opened 2026-04-03
fix: implement RowIdMeta.asdict() and from_dict() for stable row ID serialization

### LINKED ISSUE #6403: FragmentMetadata.to_json() fails with stable row IDs enabled
`FragmentMetadata.to_json()` raises `NotImplementedError` when the fragment has stable row ID metadata (`row_id_meta` is not None). This makes it impossible to serialize fragment metadata via JSON when `enable_stable_row_ids=True`.

### Minimal repro

```python
import lance
import pyarrow as pa

uri = "/tmp/repro_to_json.lance"
ds = lance.write_dataset(pa.table({"x": [1, 2, 3]}), uri, enable_stable_row_ids=True)
ds.get_fragments()[0].metadata.to_json()  # raises NotImplementedError
```

### Error

```
NotImplementedError: PyRowIdMeta.asdict is not yet supported.
```

The traceback points to `lance/fragment.py:122`:

```python
self.row_id_meta.asdict() if self.row_id_meta is not None else None
```

### Impact

Any workflow that round-trips fragment metadata through JSON breaks when stable row IDs are enabled. For example, serializing `FragmentMetadata` via `json.dumps(frag.to_json())` after `fragment.update_columns()` for later reconstruction with `FragmentMetadata.from_json()` no longer works. This is needed for distributed workflows where fragment metadata must be serialized for transport between processes.

### Environment

- lance 5.0.0-beta.4
- Python 3.11
- Linux x86_64

pengw0048:
Fixes #6403

`FragmentMetadata.to_json()` raised `NotImplementedError` when `row_id_meta` was present (`enable_stable_row_ids=True`), because `PyRowIdMeta.asdict()` was stubbed out. This also broke `FragmentMetadata.from_json()` for the round-trip, since `RowIdMeta(**dict)` doesn't work on a PyO3 class.

### Repro

```python
import lance, pyarrow as pa

uri = "/tmp/repro.lance"
ds = lance.write_dataset(pa.table({"x": [1, 2, 3]}), uri, enable_stable_row_ids=True)
ds.get_fragments()[0].metadata.to_json()  # NotImplementedError: PyRowIdMeta.asdict is not yet supported.
```

### Fix

- Implement `PyRowIdMeta.asdict()` via `pythonize` (Rust struct → Python dict)
- Add `PyRowIdMeta.from_dict()` via `depythonize` (Python dict → Rust struct)
- Update `FragmentMetadata.from_json()` to use `RowIdMeta.from_dict()` instead of `RowIdMeta(**dict)`
- Add JSON round-trip test to the existing `test_fragment_metadata_pickle` parametrized test


pengw0048:
Thanks for reviewing @jiaoew1991 . Anything left before merging? (I don't have the merge button)

PR #6421, opened 2026-04-07
feat(java): add Dataset.getZonemapStats() API

beinan:
## Summary
- Add `Dataset.getZonemapStats(columnName)` Java API to read per-zone min/max/null_count statistics from zonemap indexes
- Add `ZoneStats` data class representing per-zone statistics (fragment_id, zone_start, zone_length, min, max, null_count)
- Add Rust JNI implementation that reads the `zonemap.lance` index file and converts `ScalarValue` to Java `Comparable` objects (Long, Double, String, Boolean)
- This enables Spark connectors to perform fragment-level pruning at scan planning time using zonemap statistics — analogous to partition pruning in Hive/Iceberg

## Motivation
The lance-spark connector currently only prunes fragments based on `_rowaddr` filters. With this API, it can also prune fragments based on arbitrary user-column predicates (e.g., `WHERE date = '2026-01-15'`) by consulting zonemap min/max statistics — skipping fragments whose zones provably cannot contain matching values.

### New files
- `java/src/main/java/org/lance/index/scalar/ZoneStats.java` — Serializable data class for per-zone statistics
- `java/src/test/java/org/lance/index/ZonemapStatsTest.java` — 10 tests (data class unit tests + JNI integration tests)

### Modified files
- `java/src/main/java/org/lance/Dataset.java` — Added `getZonemapStats()` method + native declaration
- `java/lance-jni/src/blocking_dataset.rs` — JNI implementation:
  - `nativeGetZonemapStats`: finds zonemap index via `describe_indices`, reads `zonemap.lance` file, converts RecordBatch rows to Java `ZoneStats` objects
  - `scalar_value_to_java`: converts DataFusion `ScalarValue` to boxed Java types (Long, Double, String, Boolean)
- `java/lance-jni/Cargo.toml` — Added `arrow-array` and `datafusion-common` dependencies

## Test plan
- [x] `ZonemapStatsTest#testZoneStatsGetters` — data class getters
- [x] `ZonemapStatsTest#testZoneStatsNullMinMax` — null min/max handling
- [x] `ZonemapStatsTest#testZoneStatsToString` — toString format
- [x] `ZonemapStatsTest#testGetZonemapStatsNoIndex` — empty dataset, no index → empty list
- [x] `ZonemapStatsTest#testGetZonemapStatsNonexistentColumn` — bad column → exception
- [x] `ZonemapStatsTest#testGetZonemapStatsWithData` — full JNI round-trip with real data
- [x] `ZonemapStatsTest#testGetZonemapStatsMultiFragment` — 2 fragments → stats from both
- [x] `ZonemapStatsTest#testGetZonemapStatsWrongColumnReturnsEmpty` — zonemap on col A, query col B → empty
- [x] `ZonemapStatsTest#testGetZonemapStatsNullArgument` — null → exception
- [x] `ZonemapStatsTest#testGetZonemapStatsEmptyArgument` — empty → exception

### DIFF
--- a/java/lance-jni/Cargo.lock
+++ b/java/lance-jni/Cargo.lock
@@ -3778,10 +3778,12 @@ name = "lance-jni"
 version = "5.0.0-beta.5"
 dependencies = [
  "arrow",
+ "arrow-array",
  "arrow-schema",
  "async-trait",
  "bytes",
  "chrono",
+ "datafusion-common",
  "env_logger",
  "jni",
  "lance",
--- a/java/lance-jni/Cargo.toml
+++ b/java/lance-jni/Cargo.toml
@@ -28,7 +28,9 @@ lance-core = { path = "../../rust/lance-core" }
 lance-file = { path = "../../rust/lance-file" }
 lance-table = { path = "../../rust/lance-table" }
 arrow = { version = "57.1", features = ["ffi"] }
+arrow-array = "57.1"
 arrow-schema = "57.1"
+datafusion-common = "52.1.0"
 object_store = { version = "0.12.2" }
 tokio = { version = "1.23", features = [
     "rt-multi-thread",
--- a/java/lance-jni/src/blocking_dataset.rs
+++ b/java/lance-jni/src/blocking_dataset.rs
@@ -3072,3 +3072,196 @@ fn inner_get_session_handle(env: &mut JNIEnv, java_dataset: JObject) -> Result<j
     let session = dataset_guard.inner.session();
     Ok(handle_from_session(session))
 }
+
+/////////////////////////////////
+// Zonemap Stats Methods       //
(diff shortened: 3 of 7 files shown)


hamersaw:
I opened https://github.com/lance-format/lance/pull/6446 as an alternative approach to to just expose a Java API to "read index files", but I don't think that's the right abstraction layer either. It may make sense to build something like `index.get_files()` + `index_utils.read_arrow_file(...)` on the index API if we want to generalize the ability to read files directly?

westonpace:
Let's move forwards, zonemap can be a "special index" for Java for now.  In the future I think we could maybe just have plugins that ask Lance where the index is located and then go access the storage directly.

westonpace:
Thank you @beinan !

PR #6422, opened 2026-04-07
refactor: use exact base-scoped store bindings

Xuanwo:
This changes per-base runtime configuration to use exact `ObjectStoreParams` bindings keyed by `BasePath.path` instead of per-base storage option overrides. Dataset-level and write-level store params now act only as fallbacks, while reads, target-base writes, and external blob resolution all consult the same base-scoped binding model.

This keeps provider-specific runtime state out of the manifest and follows the direction in discussion #6307 to keep `BasePath` focused on identity.

### DIFF
--- a/rust/lance/src/dataset/builder.rs
+++ b/rust/lance/src/dataset/builder.rs
@@ -454,8 +454,10 @@ impl DatasetBuilder {
 
     /// Set runtime-only object store params for a specific registered base path.
     ///
-    /// These params are not persisted in the manifest. They are used whenever
-    /// the dataset resolves an object store for the given `BasePath.path`.
+    /// These params are not persisted in the manifest. They are used as-is
+    /// whenever the dataset resolves an object store for the given
+    /// `BasePath.path`. Dataset-level store params remain the fallback for bases
+    /// without an explicit binding.
     pub fn with_base_store_params(
         mut self,
         base_path: impl AsRef<str>,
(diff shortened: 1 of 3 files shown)


Xuanwo:
Thank you @westonpace for the review, I polished the design a bit. Can you talk another look?

rust/lance/src/dataset/builder.rs
```diff
@@ -448,6 +452,20 @@ impl DatasetBuilder {
         self
     }
 
+    /// Set runtime-only object store params for a specific registered base path.
+    ///
+    /// These params are not persisted in the manifest. They are used whenever
+    /// the dataset resolves an object store for the given `BasePath.path`.
```
westonpace: Can we describe how the two sets of options are merged better?  We have two `ObjectStoreParams` instances now, the dataset store params and the override params.  What actually gets used?

If I'm reading the code correctly I think:

* All options that are not storage_options (e.g. block_size, use_constant_size_upload_parts) are replaced entirely by the override.
* Storage options are handled differently.  The value is the union of the two sets of params.  If a storage_option is set in both then the value in the override is preferred.

Xuanwo: You're right that the old wording was ambiguous. In the current implementation, per-base bindings are exact `ObjectStoreParams` keyed by `BasePath.path`, and dataset-level params are only the fallback. There is no merge between dataset-level and per-base params. I updated the comments in the PR to make that explicit.

PR #6423, opened 2026-04-07
feat: support IVF partitions multi-split

BubbleCal:
### What is the new feature?
This PR allows a single `optimize_indices` call on the v3 IVF incremental optimize path to split multiple oversized IVF partitions in one pass.

Previously, optimize could split at most one oversized partition per run. After large appends, several partitions can exceed the split threshold at the same time, which forced repeated optimize cycles to bring the index back into a healthy partition layout.

### How does it work?
- `check_partition_adjustment` now collects all split candidates from the current snapshot and keeps the existing single-partition join fallback.
- The multi-split path preserves existing partition ids and appends one new partition per split partition.
- When any split happens, optimize continues to merge all existing delta indices in the same round, preserving the existing merge semantics.
- Candidate rows from overlapping reassign partitions are resolved globally so the same row is moved at most once, choosing the best destination by distance.

The initial multi-split implementation removed the functional limitation, but overlapping reassign partitions still had avoidable overhead:
- split planning was done sequentially
- split plans retained full raw vector payloads longer than necessary
- a reused candidate partition recomputed its baseline distance to the original centroid for every overlapping split request

- split plans are now built with bounded parallelism across compute CPUs
- split plans no longer retain raw partition vectors after producing the original-partition assign ops
- each reused candidate partition is loaded once and computes its baseline distance once, then reuses that result across overlapping split requests
- best candidate moves are updated in-place per row id instead of materializing intermediate move vectors per request

## Testing

- `cargo test -p lance compute_reassign_candidate_moves_vectors_to_new_centroids`
- `cargo test -p lance test_partition_split_on_append_multivec`
- `cargo test -p lance test_split_multiple_partitions_in_one_optimize`
- `cargo test -p lance test_join_partition_on_delete_multivec`

### DIFF
--- a/rust/lance/src/index/vector/builder.rs
+++ b/rust/lance/src/index/vector/builder.rs
@@ -1377,6 +1377,7 @@ impl<S: IvfSubIndex + 'static, Q: Quantization + 'static> IvfIndexBuilder<S, Q> 
             .await?;
         let mut split_plans = split_plans.into_iter().flatten().collect::<Vec<_>>();
         split_plans.sort_by_key(|plan| plan.part_idx);
+        Self::finalize_split_plans(&mut split_plans, ivf.num_partitions());
 
         if split_plans.is_empty() {
             return Ok(AssignResult {
@@ -1539,7 +1540,7 @@ impl<S: IvfSubIndex + 'static, Q: Quantization + 'static> IvfIndexBuilder<S, Q> 
                 ))?;
                 let baseline_dists =
                     self.distance_type.arrow_batch_func()(candidate_centroid.as_ref(), &vectors)?;
-                let mut best_moves = HashMap::<u64, CandidateMove>::new();
+                let mut best_moves = vec![None; row_ids.len()];
                 for request in requests {
                     let d1 = self.distance_type.arrow_batch_func()(
                         request.centroid1.as_ref(),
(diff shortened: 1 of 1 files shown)


rust/lance/src/index/vector/builder.rs
```diff
@@ -1362,26 +1353,100 @@ impl<S: IvfSubIndex + 'static, Q: Quantization + 'static> IvfIndexBuilder<S, Q>
...
+        &self,
+        part_idx: usize,
+        centroid2_part_idx: usize,
+        ivf: &IvfModel,
+    ) -> Result<Option<SplitPlan>>
+    where
+        T::Native: Dot + L2 + Normalize,
+        PrimitiveArray<T>: From<Vec<T::Native>>,
+    {
+        let Some((row_ids, vectors)) = self.load_partition_raw_vectors(part_idx).await? else {
```
Xuanwo, line 1443: `load_partition_raw_vectors` will load multivector into flat ones so one row id could match different vectors. Will this make the multivector data been missing in our index?

rust/lance/src/index/vector/builder.rs
```diff
@@ -1362,26 +1353,100 @@ impl<S: IvfSubIndex + 'static, Q: Quantization + 'static> IvfIndexBuilder<S, Q>
...
         PrimitiveArray<T>: From<Vec<T::Native>>,
     {
         let centroids = ivf.centroids_array().unwrap();
-        let mut new_centroids: Vec<ArrayRef> = Vec::with_capacity(ivf.num_partitions() + 1);
+        let mut new_centroids: Vec<ArrayRef> =
+            Vec::with_capacity(ivf.num_partitions() + split_partitions.len());
         new_centroids.extend(centroids.iter().map(|vec| vec.unwrap()));
+        let split_plans = stream::iter(split_partitions.iter().copied().enumerate())
+            .map(|(split_order, part_idx)| async move {
+                let centroid2_part_idx = ivf.num_partitions() + split_order;
```
Xuanwo, line 1371: The `centroid2_part_idx` is calculated using `ivf.num_partitions()`. Could it be that some partitions have been filtered, causing the `centroid2_part_idx` in the plan to exceed the actual number? If so, the subsequent `assign_ops[*target_idx]` operation might panic.

PR #6424, opened 2026-04-07
feat(python): add blob-aware to_pandas

BubbleCal:
### What is the new feature?
This PR adds a first-class blob-aware `to_pandas()` API to the Lance Python bindings on `LanceDataset`, `LanceScanner`, and `LanceFragment`.

Today, pandas export goes through `to_table().to_pandas()`, which is constrained by Arrow blob representations. That means blob columns either surface as descriptor structs or must be eagerly materialized as bytes before pandas conversion. For large blobs, eager materialization is the wrong default because it can pull a large amount of binary data into memory unexpectedly.

### How does it work?
The new `to_pandas(*, blob_mode=...)` API keeps Arrow-facing behavior unchanged and adds pandas-specific blob handling:

- `blob_mode="lazy"` (default) returns `lance.BlobFile` objects in pandas object columns.
- `blob_mode="bytes"` eagerly reads blobs into Python `bytes`.
- `blob_mode="descriptions"` preserves the old `to_table().to_pandas()` behavior.

Implementation details:
- Add Python-side blob helpers to detect top-level blob columns and map direct alias projections back to source blob columns.
- Snapshot Python scanner builder options so `LanceScanner.to_pandas()` can reconstruct the same scan with `_rowaddr` and blob descriptions.
- Rebuild the scan internally with `with_row_address=True` and `blob_handling="blobs_descriptions"`, convert non-blob columns through Arrow's `to_pandas()`, and backfill blob columns via `take_blobs(..., addresses=...)`.
- Preserve Arrow APIs (`to_table()` / `blob_handling`) unchanged.
- Raise a clear `NotImplementedError` for transformed blob projections that cannot be mapped back to a single source blob column.

## Testing

- `cd python && uv run pytest python/tests/test_blob.py -q`
- `cd python && uv run --extra dev ruff check python/lance/dataset.py python/lance/fragment.py python/tests/test_blob.py`

### DIFF
--- a/python/python/lance/dataset.py
+++ b/python/python/lance/dataset.py
@@ -91,7 +91,16 @@ if TYPE_CHECKING:
         Iterable[float],
     ]
 LANCE_COMMIT_MESSAGE_KEY = "__lance_commit_message"
-_BLOB_DESC_PANDAS_MODE = "descriptions"
+_BLOB_PANDAS_MODE_LAZY = "lazy"
+_BLOB_PANDAS_MODE_BYTES = "bytes"
+_BLOB_PANDAS_MODE_DESCRIPTIONS = "descriptions"
+_BLOB_PANDAS_MODES = frozenset(
+    {
+        _BLOB_PANDAS_MODE_LAZY,
+        _BLOB_PANDAS_MODE_BYTES,
+        _BLOB_PANDAS_MODE_DESCRIPTIONS,
+    }
+)
 _BLOB_ROW_ADDR_COLUMN = "_rowaddr"
 
 
@@ -116,7 +125,7 @@ def _blob_columns_in_schema(schema: pa.Schema) -> set[str]:
 def _normalize_blob_pandas_mode(
     blob_mode: str,
 ) -> Literal["lazy", "bytes", "descriptions"]:
-    if blob_mode not in {"lazy", "bytes", _BLOB_DESC_PANDAS_MODE}:
+    if blob_mode not in _BLOB_PANDAS_MODES:
         raise ValueError("blob_mode must be one of: 'lazy', 'bytes', 'descriptions'")
     return cast("Literal['lazy', 'bytes', 'descriptions']", blob_mode)
 
@@ -162,6 +171,13 @@ def _blob_column_sources(
 
 
 def _snapshot_scanner_builder(builder: "ScannerBuilder") -> Dict[str, Any]:
+    """Capture Python-side scanner config needed to rebuild the scan later.
+
+    The native scanner object does not preserve the original builder arguments.
+    We need these values to recreate the same scan when `to_pandas` switches blob
+    handling modes or injects `_rowaddr` for lazy blob export.
+    """
+
     def snapshot_value(value: Any) -> Any:
         try:
             return copy.deepcopy(value)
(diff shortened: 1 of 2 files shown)


python/python/lance/dataset.py
```diff
@@ -90,6 +91,165 @@
...
+
+
+def _blob_columns_in_schema(schema: pa.Schema) -> set[str]:
+    return {field.name for field in schema if _is_blob_field(field)}
+
+
+def _normalize_blob_pandas_mode(
+    blob_mode: str,
+) -> Literal["lazy", "bytes", "descriptions"]:
+    if blob_mode not in {"lazy", "bytes", _BLOB_DESC_PANDAS_MODE}:
```
westonpace: It's a little odd that `descriptions` gets a constant but the other two do not.

python/python/lance/dataset.py
```diff
@@ -90,6 +91,165 @@
...
+        "_strict_batch_size": builder._strict_batch_size,
+        "_orderings": (
+            tuple(builder._orderings) if builder._orderings is not None else None
+        ),
+        "_disable_scoring_autoprojection": builder._disable_scoring_autoprojection,
+        "_substrait_aggregate": builder._substrait_aggregate,
+    }
+
+
+def _scanner_from_snapshot(
```
westonpace, line 194: Why is this snapshot stuff required?  It seems tedious to maintain

python/python/lance/fragment.py
```diff
@@ -585,6 +619,32 @@ def to_table(
             order_by=order_by,
         ).to_table()
 
+    def to_pandas(
```
westonpace, line 622: Can we comment this function?  We need to explain `blob_mode` somewhere.

python/python/lance/fragment.py
```diff
@@ -585,6 +619,32 @@ def to_table(
...
+        columns: Optional[Union[List[str], Dict[str, str]]] = None,
+        filter: Optional[Union[str, pa.compute.Expression]] = None,
+        limit: Optional[int] = None,
+        offset: Optional[int] = None,
+        with_row_id: bool = False,
+        with_row_address: bool = False,
+        blob_handling: Optional[
+            Literal["all_binary", "blobs_descriptions", "all_descriptions"]
+        ] = None,
+        blob_mode: str = "lazy",
```
westonpace: If this function has `blob_mode` does it really need `blob_handling`?  It seems like it should replace that option.

python/python/lance/dataset.py
```diff
@@ -5443,6 +5662,71 @@ def to_reader(self) -> pa.RecordBatchReader:
...
+            )
+            blob_iter = iter(blob_files)
+            values = []
+            for description in descriptions:
+                if _is_null_blob_description(description):
+                    values.append(None)
+                    continue
+                blob = next(blob_iter)
+                values.append(
+                    blob if blob_mode == "lazy" else _read_blob_as_bytes(blob)
```
westonpace: If the user specifies the `blob_mode` is `bytes` shouldn't we just set `blob_handling` to `all_binary`?  Then the read could happen at the rust level and python wouldn't have to bother with it.

python/python/lance/dataset.py
```diff
@@ -90,6 +91,140 @@
...
+
+def _is_blob_field(field: pa.Field) -> bool:
+    return (
+        _field_metadata_value(field, "lance-encoding:blob") == b"true"
+        or _field_metadata_value(field, "ARROW:extension:name") == b"lance.blob.v2"
+    )
+
+
+def _blob_columns_in_schema(schema: pa.Schema) -> set[str]:
+    return {field.name for field in schema if _is_blob_field(field)}
```
wjones127, line 122: issue(non-blocking): this won't support blob fields that are nested. I'm not sure we generally support nested blob fields, but I imagine we want to. We should make a follow up issue for that.

python/python/lance/dataset.py
BubbleCal: Added comments to explain this.
The snapshot is still needed because the native scanner object doesn’t retain the Python builder config, and `to_pandas()` needs to rebuild the same scan with different `blob_handling` / `_rowaddr`.

python/python/lance/dataset.py
BubbleCal: followup: https://github.com/lance-format/lance/issues/6436

Xuanwo (approved):
Thank you for working on this

PR #6427, opened 2026-04-07
feat: add DataFile.create helper for building DataFile metadata

### LINKED ISSUE #6413: Create a python helper function for creating `DataFile` metadata
Python users do not usually need to create `Fragment` metadata even when they are creation operations for use with `Dataset.commit`.  This is because methods like `Fragment.write` take care of that for them.  However, the `DataReplacement` operation takes in `DataFile` metadata directly and currently requires the user to create it.  This can be a bit tedious.

Instead we could have a helper function `DataFile.create` which could grab the information needed from the file.

The user would only need to provide the `base_id` and `path`.  The helper function would look for the file and open it (verifying the file exists), and then grab the schema from the file.  It could match this schema to the dataset schema (like we do at write time) and calculate the lance schema (user doesn't need to provide field ids) and then it could calculate the column_indices.  The file major version and minor version should be written in the file too.  We can also calculate `file_size_bytes`.

westonpace:
## Summary
- Adds `DataFile.create(dataset, path, *, base_id=None)` classmethod that reads a lance file's metadata and automatically constructs a `DataFile` with correct field IDs, column indices, file version, and file size
- Eliminates the need for manual `DataFile` construction when performing `DataReplacement` operations
- Handles packed structs, structural file versions (v2.1+), subset columns, and external base paths

Closes #6413

## Test plan
- [x] `test_data_file_create_basic` — verifies fields, column_indices, version, file_size for a two-column file
- [x] `test_data_file_create_subset_columns` — single column from a multi-column dataset
- [x] `test_data_file_create_end_to_end` — full DataReplacement round-trip using the new helper
- [x] `test_data_file_create_unknown_column` — error on column not in dataset schema
- [x] All existing `test_table_ops.py` tests still pass

### DIFF
--- a/rust/lance/src/dataset.rs
+++ b/rust/lance/src/dataset.rs
@@ -1745,7 +1745,7 @@ impl Dataset {
             SchedulerConfig::new(2 * 1024 * 1024 * 1024),
         );
         let file = scheduler
-            .open_file(&filepath, &CachedFileSize::unknown())
+            .open_file(&filepath, &CachedFileSize::new(file_size))
             .await?;
         let file_metadata = FileReader::read_all_metadata(&file).await?;
 


rust/lance/src/dataset.rs
```diff
@@ -1714,14 +1718,116 @@ impl Dataset {
...
+        let file_size = self.object_store().size(&filepath).await?;
+
+        // Read file metadata
+        let scheduler = ScanScheduler::new(
+            self.object_store.clone(),
+            SchedulerConfig::new(2 * 1024 * 1024 * 1024),
+        );
+        let file = scheduler
+            .open_file(&filepath, &CachedFileSize::unknown())
+            .await?;
```
wjones127, line 1749: nitpick: you just read the file size, so you should be able to pass it here:


```suggestion
        let file = scheduler
            .open_file(&filepath, &CachedFileSize::new(file_size))
            .await?;
```

PR #6428, opened 2026-04-07
feat: wire batch_size_bytes to Python and public Rust API

westonpace:
## Summary

Stacked on #6388. Please merge that PR first.

- Adds `batch_size_bytes: Option<u64>` to `FileReaderOptions` and propagates it through all 6 `SchedulerDecoderConfig` creation sites in the file reader
- Adds `batch_size_bytes` field + setter to `Scanner`, wired through both `scan_fragments` (via `LanceScanConfig`) and `pushdown_scan` (via `FileReaderOptions` in `ScanConfig`)
- Adds `batch_size_bytes` to `LanceScanConfig`, with `try_new_v2` injecting it into `FragReadConfig` via `FileReaderOptions`
- Exposes `batch_size_bytes` in the Python API: `LanceDataset.scanner()`, `to_table()`, `to_batches()`, `ScannerBuilder`

## Test plan

- [x] `cargo check -p lance-file -p lance --tests` — clean
- [x] `cargo clippy -p lance-file -p lance --tests -- -D warnings` — clean
- [x] `cargo fmt --all` — applied
- [x] `cargo test -p lance-encoding -- byte_sized` — 3/3 pass
- [x] `cargo test -p lance -- test_scan` — 38/38 pass

### DIFF
--- a/rust/lance/src/dataset/scanner.rs
+++ b/rust/lance/src/dataset/scanner.rs
@@ -1269,7 +1269,7 @@ impl Scanner {
     /// Set the maximum number of rows per batch.
     ///
     /// Note: this can be overridden by [`Self::batch_size_bytes`] or by a dataset-level
-    /// `batch_size_bytes` set via [`ReadParams::file_reader_options`].  When a byte-based
+    /// `batch_size_bytes` set via [`ReadParams::file_reader_options`](crate::dataset::ReadParams::file_reader_options).  When a byte-based
     /// batch size is active, the row-based batch size is used only as an initial estimate.
     pub fn batch_size(&mut self, batch_size: usize) -> &mut Self {
         self.batch_size = Some(batch_size);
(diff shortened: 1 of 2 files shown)


rust/lance/src/dataset/scanner.rs
```diff
@@ -1688,6 +1710,30 @@ impl Scanner {
...
+    /// `file_reader_options`, the dataset-level defaults, and the `batch_size_bytes`
+    /// setting.
+    fn resolved_file_reader_options(&self) -> Option<FileReaderOptions> {
+        let base = self
+            .file_reader_options
+            .clone()
+            .or_else(|| self.dataset.file_reader_options.clone());
+        match (base, self.batch_size_bytes) {
+            (Some(mut opts), Some(bsb)) => {
+                if opts.batch_size_bytes.is_none() {
```
BubbleCal, line 1723: should we always set `opts.batch_size_bytes = Some(bsb)` no matter whether `opts.batch_size_bytes` is `None` or not?

westonpace: If we set `opts.batch_size_bytes` then we will override `batch_size`.

I don't want to change the default (yet) so that this can stay a non-breaking change.  The default is that the user doesn't set `batch_size_bytes` on either the scanner settings or the dataset settings and so we use `batch_size` instead.

PR #6429, opened 2026-04-07
feat: support distributed bitmap index builds

butnaruandrei:
Allows bitmap indexes to be built in a distributed/fragment-scoped manner,
consistent with the existing BTree and FTS distributed build patterns.
- `BitmapIndexPlugin::train_index` now accepts `fragment_ids`, writing
  per-shard `part_{partition_id}_bitmap_page_lookup.lance` files instead
  of erroring when `fragment_ids` is set
- `merge_index_files` performs a K-way sorted merge of shard files,
  unioning `RowAddrTreeMap`s for matching keys, then writes the canonical
  `bitmap_page_lookup.lance` and deletes the shards
- `Dataset::merge_index_metadata` dispatches `IndexType::Bitmap` to the
  new merger
- Python `SupportedDistributedIndices` gains a `BITMAP` variant

Non-distributed builds (no `fragment_ids`) are unchanged. After merge the
on-disk layout is identical to a monolithic build.

### DIFF
--- a/python/python/lance/dataset.py
+++ b/python/python/lance/dataset.py
@@ -3551,7 +3551,7 @@ class LanceDataset(pa.dataset.Dataset):
         This method does NOT commit changes.
 
         This API merges temporary scalar index files (for example per-fragment
-        BTree or inverted index outputs).
+        BTree, Bitmap, or inverted index outputs).
         After this method returns, callers MUST explicitly commit
         the index manifest using lance.LanceDataset.commit(...)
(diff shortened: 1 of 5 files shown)


Xuanwo (changes requested):
Thank you for working on this! However we have designed a new general distributed API and pipeline as showed up here: https://lance.org/guide/distributed_indexing/

Whenever possible, please adapt the index build API to this one.

PR #6430, opened 2026-04-07
chore: add tracing instrumentation to mem_wal module

hamersaw:
## Summary

- Adds `#[instrument]` attributes from the `tracing` crate to key functions across the `mem_wal` module
- Covers write path (`RegionWriter::open`, `put`, `close`), flush path (`MemTableFlusher::flush`, `flush_with_indexes`), WAL operations, manifest store, memtable inserts, scanner/planner, point lookups, and vector search
- Uses appropriate trace levels (`info` for high-level operations, `debug` for internals) with relevant fields (region_id, epoch, row counts, batch counts)

## Test plan

- [x] `cargo check` passes — no functional changes, only attribute additions
- [x] Existing `mem_wal` tests continue to pass
- [ ] Tracing output verified with `RUST_LOG=debug` showing instrumented spans

### DIFF
--- a/rust/lance/src/dataset/mem_wal/scanner/builder.rs
+++ b/rust/lance/src/dataset/mem_wal/scanner/builder.rs
@@ -13,7 +13,6 @@ use datafusion::physical_plan::{ExecutionPlan, SendableRecordBatchStream};
 use datafusion::prelude::{Expr, SessionContext};
 use futures::TryStreamExt;
 use lance_core::{Error, Result};
-use tracing::instrument;
 use uuid::Uuid;
 
 use super::collector::{ActiveMemTableRef, LsmDataSourceCollector};
@@ -166,7 +165,6 @@ impl LsmScanner {
     }
 
     /// Create the execution plan.
-    #[instrument(name = "lsm_create_plan", level = "debug", skip_all)]
     pub async fn create_plan(&self) -> Result<Arc<dyn ExecutionPlan>> {
         let collector = self.build_collector();
         let base_schema = self.schema();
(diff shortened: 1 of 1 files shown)


rust/lance/src/dataset/mem_wal/scanner/builder.rs
```diff
@@ -183,6 +185,7 @@ impl LsmScanner {
     }
 
     /// Execute the scan and return a stream of record batches.
+    #[instrument(level = "info", skip(self), fields(has_filter = self.filter.is_some(), limit = self.limit, num_shards = self.shard_snapshots.len()))]
     pub async fn try_into_stream(&self) -> Result<SendableRecordBatchStream> {
```
westonpace: It might be nice to generate something similar to the `plan_run` events that we have on the normal scanner (https://github.com/lance-format/lance/blob/6112a34bfe38618f07c099217dc3d89fd39ca6bb/rust/lance-datafusion/src/exec.rs#L545)

rust/lance/src/dataset/mem_wal/write.rs
```diff
@@ -1257,6 +1260,7 @@ impl ShardWriter {
     /// Close the writer gracefully.
     ///
     /// Flushes pending data and shuts down background tasks.
+    #[instrument(level = "info", skip(self), fields(shard_id = %self.config.shard_id, epoch = self.epoch))]
     pub async fn close(self) -> Result<()> {
```
westonpace: I wonder if we want to give these traces names like sw_open, sw_put, sw_close?  A lot of tracing tools will just give the span name unless you drill down and open/put/close are going to feel like regularly filesystem calls.

hamersaw: I think this is a great idea. Just went through and named all of them. I think even for things like "flush" that can be VERY ambiguous depending on the tracing tool context.

hamersaw:
> Generally I try and always use skip_all since skip(...)

In hopes that somebody manually submits a PR someday, I've updated everything to `skip_all` ... it just makes sense.

PR #6432, opened 2026-04-07
fix: enforce WAL writer fencing via conditional PUT

hamersaw:
## Summary

- WAL fencing was broken: a fenced writer could keep writing WAL entries indefinitely because `flush_to_with_index_update` used plain `put()` with no fencing check. Two writers with the same shard ID silently overwrote each other's WAL entries.
- WAL entries now use `PutMode::Create` (put-if-not-exists) — same cost as regular PUT on S3 but fails atomically if another writer already claimed that position.
- New `write_fence_barrier()` at startup writes an empty WAL entry to claim the next position, scanning forward past any stale entries from old writers. This immediately fences zombie writers on open.

## Test plan

- [x] 9 new fencing tests in `wal::tests` (conditional PUT rejection, fence barrier claiming, barrier skip-forward, barrier blocking old writer, sequential position uniqueness)
- [x] 8 new fencing tests in `manifest::tests` (epoch claiming, check_fenced edge cases, commit_update rejection, three-writer scenario)
- [x] All 12 existing `write::tests` pass (including E2E correctness test with fence barrier in startup path)
- [x] Zero clippy warnings


hamersaw:
## Verification that fencing was broken on `main`

To confirm this fix addresses a real bug, I added `test_wal_flush_rejects_duplicate_position` to `main` (without the fix) and ran it. The test creates two `WalFlusher` instances with the same shard ID starting at the same WAL position — simulating two writers that read the same stale manifest.

**On `main`: test FAILS** — both writers succeed because `put()` is an unconditional overwrite:
```
Writer B should be fenced out (rejected) when writing to the same WAL position,
but it succeeded — fencing is broken! Plain put() silently overwrites the file.
test result: FAILED. 0 passed; 1 failed;
```

**On this branch: test PASSES** — the second writer is rejected because `put_opts(..., PutMode::Create)` fails atomically when the file already exists.

Root cause: `flush_to_with_index_update` used `store.inner.put()` with no conditional write semantics, so any writer could silently overwrite another writer's WAL entry at the same position.

rust/lance/src/dataset/mem_wal/wal.rs
```diff
@@ -476,6 +489,87 @@ impl WalFlusher {
...
+    pub async fn write_fence_barrier(&self, schema: &ArrowSchema) -> Result<()> {
+        let object_store = self
+            .object_store
+            .as_ref()
+            .ok_or_else(|| Error::io("Object store not set on WAL flusher"))?;
+
+        // Build an empty WAL entry (schema + epoch metadata, no batches)
+        let mut metadata = schema.metadata().clone();
+        metadata.insert(WRITER_EPOCH_KEY.to_string(), self.writer_epoch.to_string());
+        let schema_with_epoch = ArrowSchema::new_with_metadata(schema.fields().to_vec(), metadata);
```
westonpace: Have we confirmed these fence barriers don't bother the read half?

rust/lance/src/dataset/mem_wal/wal.rs
```diff
@@ -476,6 +489,87 @@ impl WalFlusher {
...
+            }
+        }
+    }
+
+    /// Map an object_store put error to a fencing error when AlreadyExists.
+    fn map_wal_put_error(e: object_store::Error, wal_entry_position: u64, shard_id: Uuid) -> Error {
+        if matches!(e, object_store::Error::AlreadyExists { .. }) {
+            Error::io(format!(
+                "Writer fenced: WAL position {} already exists for shard {} \
+                 (another writer has claimed this position)",
```
westonpace: What does the fenced writer do if it's not automatically retrying?

rust/lance/src/dataset/mem_wal/write.rs
```diff
@@ -1013,6 +1013,11 @@ impl ShardWriter {
         );
         wal_flusher.set_object_store(object_store.clone());
 
+        // Write fence barrier to claim WAL positions and fence zombie writers.
+        // Uses PutMode::Create to atomically claim the next WAL position,
+        // scanning forward past any entries left by old (fenced) writers.
+        wal_flusher.write_fence_barrier(&schema).await?;
```
westonpace: It's slightly counter intuitive that a method named `open` will actually do a write (e.g. isn't idempotent).  Can we document this in the method docs?

westonpace (approved):
Is a "put not exists" actually free?  I thought object stores typically had a much lower throughput for conditional operations (e.g. 10 per second) than unconditional operations.

rust/lance/src/dataset/mem_wal/write.rs
hamersaw: Absolutely. I can't seem to think of a cleaner way to implement this. The barrier us just an empty WAL entry and as long as there's not excessive writer churn the overhead should be negligible.

rust/lance/src/dataset/mem_wal/wal.rs
hamersaw: haha, well I think my biggest fail here is trusting this codebase. Turns out we don't re-read WAL entries on restarts / new writers, so this is another bug I'll work through.

rust/lance/src/dataset/mem_wal/wal.rs
hamersaw: The fenced writer should gracefully fail. Basically the algorithm is designed that if a new writer starts for the shard, it just takes over.

hamersaw:
already fixed by https://github.com/lance-format/lance/pull/6675

PR #6438, opened 2026-04-08
feat: append data with reserved stable row ids

majin1102:
Close #6466 

**What this PR does**

   This PR introduces a new `ReserveRowIds` transaction operation and extends `Append` with an optional `row_ids` field. Both changes affect transaction metadata and the serialized transaction schema, so this part needs a vote.

**What this PR does not do**

   This PR does **not** add any new field or persisted state to the manifest. The manifest format remains unchanged. The only manifest-related effect is that this PR provides a controlled transaction-level entry point for reserving and consuming `next_row_id`, rather than changing manifest contents themselves.


protos/table.proto
```diff
@@ -32,6 +32,11 @@ message UUID {
   bytes uuid = 1;
 }
 
+message ReservedRowIds {
```
majin1102: This is unnecessary?

protos/transaction.proto
```diff
@@ -39,6 +39,13 @@ message Transaction {
     //
     // Fragment IDs are not yet assigned.
     repeated DataFragment fragments = 1;
+    // Stable row ids to assign to the new rows, if pre-reserved.
+    //
+    // The requested range must be contained within the row ids reserved by the
+    // dataset version identified by read_version. Multiple append operations
+    // may consume disjoint sub-ranges from the same reservation as long as
+    // they continue to reference that reserve version.
+    ReservedRowIds row_ids = 2;
```
majin1102: should this be optional?

protos/transaction.proto
```diff
@@ -339,6 +351,7 @@ message Transaction {
     UpdateMemWalState update_mem_wal_state = 112;
     Clone clone = 113;
     UpdateBases update_bases = 114;
+    ReserveRowIds reserve_row_ids = 115;
```
majin1102, line 359: Is this necessary?

rust/lance-table/src/format/manifest.rs
```diff
@@ -82,7 +90,7 @@ pub struct Manifest {
     /// accelerating the fragment search using offset ranges.
     fragment_offsets: Vec<usize>,
 
-    /// The max row id used so far.
+    /// The next row id that will be allocated.
```
majin1102, line 93: The semantic is changed? Is this safe and necessary?

rust/lance-table/src/format/manifest.rs
```diff
@@ -839,6 +849,24 @@ impl From<BasePath> for pb::BasePath {
     }
 }
 
+impl From<pb::ReservedRowIds> for ReservedRowIds {
```
majin1102: Is this necessary?

majin1102:
@jackye1995 Hi, could you take a look at this issue and PR?  If the motivation and proposal make sense to you, I’ll start a vote thread.

westonpace (approved):
This PR looks good to me.  Can you create a vote discussion to add the new reserve row ids operation?  I recently created one for tracking invalidated fragments that you can use as inspiration.

PR #6439, opened 2026-04-08
fix(merge_insert): use sentinel column for NULL-safe source row detection

pratik0316:
Source rows with NULL ON key columns were silently dropped because the action assignment logic used `ON_col IS NOT NULL` as a proxy for "source row is present in the join output". This conflates a legitimate NULL key with a NULL introduced by the outer join on the target side.

Fix by injecting a `lit(true)` sentinel column into the source DataFrame before the join. After the join the sentinel is non-null for every source row and null only for target-only rows, making source row detection independent of ON column values.

Strip the sentinel in `prepare_stream_schema` before writing and propagate it through projection pushdown in `necessary_children_exprs`.

Before the join, inject a constant `lit(true)` column (`__merge_source_sentinel`) into every source row. After the join:

- Source rows (whether matched or unmatched) → sentinel = true
- Target-only rows (no source match) → sentinel = NULL (outer join NULL-fill)
[assign_action](https://github.com/lance-format/lance/blob/6112a34bfe38618f07c099217dc3d89fd39ca6bb/rust/lance/src/dataset/write/merge_insert/assign_action.rs#L77) now uses sentinel IS NOT NULL to detect source row presence, making it correct regardless of what values the ON columns hold.

The sentinel is a pure logical column — it never touches disk. It's stripped in [prepare_stream_schema](https://github.com/lance-format/lance/blob/6112a34bfe38618f07c099217dc3d89fd39ca6bb/rust/lance/src/dataset/write/merge_insert/exec/write.rs#L383) before any data is written, and [necessary_children_exprs](https://github.com/lance-format/lance/blob/6112a34bfe38618f07c099217dc3d89fd39ca6bb/rust/lance/src/dataset/write/merge_insert/logical_plan.rs#L148) is updated to propagate it through DataFusion's projection pushdown.

Example that was broken before:

```
Target: (id=1, record_type="A") and (id=0, record_type=NULL)
Source: (id=2, record_type=NULL) — new row, should be inserted
ON: ["id", "record_type"]
Old behavior: source row silently dropped (Action::Nothing)
New behavior: source row correctly inserted (Action::Insert)

```

Fixes: https://github.com/lance-format/lance/issues/4644

### DIFF
--- a/python/python/lance/dataset.py
+++ b/python/python/lance/dataset.py
@@ -298,7 +298,7 @@ class MergeInsertBuilder(_MergeInsertBuilder):
         >>> print(plan) # doctest: +ELLIPSIS
         MergeInsert: on=[id], when_matched=UpdateAll, when_not_matched=InsertAll, ...
           CoalescePartitionsExec
-            ProjectionExec: expr=[_rowid@1 as _rowid, _rowaddr@2 as _rowaddr, ...]
+            ProjectionExec: expr=[...]
               HashJoinExec: mode=CollectLeft, join_type=Right, ...
                 CooperativeExec
                   LanceRead: uri=test_dataset/data, projection=[id], ...
@@ -317,7 +317,7 @@ class MergeInsertBuilder(_MergeInsertBuilder):
         >>> print(plan) # doctest: +ELLIPSIS
         MergeInsert: on=[id], when_matched=UpdateAll, when_not_matched=InsertAll, ...
           CoalescePartitionsExec
-            ProjectionExec: expr=[_rowid@1 as _rowid, _rowaddr@2 as _rowaddr, ...]
+            ProjectionExec: expr=[...]
               HashJoinExec: mode=CollectLeft, join_type=Right, ...
                 ...
         """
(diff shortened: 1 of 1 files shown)


pratik0316:
Hi @wjones127 can you pls provide a review

rust/lance/src/dataset/write/merge_insert.rs
```diff
@@ -6445,6 +6462,217 @@ MergeInsert: on=[id], when_matched=UpdateAll, when_not_matched=InsertAll, when_n
...
+        // Initial dataset: one row with a NULL record_type.
+        let initial_data = RecordBatch::try_new(
+            schema.clone(),
+            vec![
+                Arc::new(Int32Array::from(vec![Some(0)])),
+                Arc::new(StringArray::from(vec![Option::<&str>::None])),
+                Arc::new(Int32Array::from(vec![Some(10)])),
+            ],
+        )
+        .unwrap();
```
wjones127: suggestion: this can be shortened to:
```suggestion
        // Schema with two ON columns, one of which will hold NULL values.
        // Initial dataset: one row with a NULL record_type.
        let initial_data = record_batch!(
            ("id", Int32, [0]),
            ("record_type", Utf8, [Option::<&str>::None]),
            ("value", Int32, [10])
        ).unwrap();;
```

rust/lance/src/dataset/write/merge_insert.rs
```diff
@@ -6445,6 +6462,217 @@ MergeInsert: on=[id], when_matched=UpdateAll, when_not_matched=InsertAll, when_n
...
+        // semantics, so this row must be INSERTED.
+        let new_data = RecordBatch::try_new(
+            schema.clone(),
+            vec![
+                Arc::new(Int32Array::from(vec![Some(2)])),
+                Arc::new(StringArray::from(vec![Option::<&str>::None])),
+                Arc::new(Int32Array::from(vec![Some(99)])),
+            ],
+        )
+        .unwrap();
```
wjones127: suggestion: similar here:
```suggestion
        let new_data = record_batch!(
            ("id", Int32, [2]),
            ("record_type", Utf8, [Option::<&str>::None]),
            ("value", Int32, [99])
        ).unwrap();
```

pratik0316:
Thanks for the super fast review @wjones127 🙌

trying to adress the suggestions shortly

rust/lance/src/dataset/write/merge_insert.rs
pratik0316: adressed, thanks for the ssuggestions

PR #6465, opened 2026-04-09
fix: serialize version metadata through JNI and correct row-ID lookup

ivscheianu:
Fixes https://github.com/lance-format/lance/issues/6464

## Summary

Two bugs in the stable row ID + CDF version tracking pipeline caused `_row_created_at_version` to report incorrect values after updates.

### Bug 1: JNI serialization gap for version metadata

`FragmentMetadata` round-trips through JNI (Rust -> Java -> Rust) lost `created_at_version_meta` and `last_updated_at_version_meta`. The JNI layer serialized `RowIdMeta` but hard-coded `None` for both version metadata fields on the Java->Rust path, so any version information attached by lance-core was silently dropped when the connector returned fragments.

**Fix:** Add `VersionMeta.java` (mirrors `RowIdMeta`), extend `FragmentMetadata` with two new `VersionMeta` fields, and update `lance-jni/src/fragment.rs` to serialize/deserialize `RowDatasetVersionMeta` via JSON through the new Java type. Also extracted a generic `extract_nullable_field<T>()` helper to reduce duplicated nullable-getter boilerplate.

### Bug 2: Incorrect row-ID lookup in `Operation::Update`

When preserving `created_at` versions for updated rows, the code assumed `row_id >> 32` yields the original fragment ID. This is only true for *unstable* row addresses. Stable row IDs are sequential integers unrelated to
fragment layout. Any dataset with stable row IDs would look up the wrong fragment and fall back to a default version of 1.

**Fix:** Build a reverse index (stable row ID -> (fragment ref, offset)) from each existing fragment's `RowIdMeta` sequence, then look up each new fragment's row IDs in this index to find the correct original fragment and row offset.

### Additional improvements

- `RowIdMeta.fromRowIds(long[])` — static factory that encodes stable row IDs into the inline protobuf/JSON format expected by lance-core, centralizing logic that was previously hand-rolled in the connector.

## Companion PR

This PR is paired with https://github.com/lance-format/lance-spark/pull/407 which implements the native UPDATE path in the Spark connector: capturing `_rowid` during `update()`, attaching `RowIdMeta` to new fragments, and consolidating deletion handling on the driver.


rust/lance-table/src/rowids/segment.rs
```diff
@@ -153,7 +153,7 @@ impl U64Segment {
         let n_holes = stats.n_holes();
         let total_slots = stats.max - stats.min + 1;
 
-        let range_with_holes = 24 + 4 * n_holes as usize;
+        let range_with_holes = 24usize.saturating_add(4usize.saturating_mul(n_holes as usize));
```
ivscheianu, line 156: created https://github.com/lance-format/lance/issues/6515 for a long-term fix, will try to take care of the impl in a separate PR

java/lance-jni/src/fragment.rs
```diff
@@ -621,6 +652,18 @@ impl IntoJava for &RowIdMeta {
     }
 }
 
+impl IntoJava for &RowDatasetVersionMeta {
+    fn into_java<'a>(self, env: &mut JNIEnv<'a>) -> Result<JObject<'a>> {
```
andrei-ionescu, line 656: Nit pick... Let's use `'local` for the lifetime, similar to the other `Fragment.into_java()`.

rust/lance/src/dataset/transaction.rs
```diff
@@ -45,6 +45,131 @@ use std::{
 };
 use uuid::Uuid;
 
+/// Fallback version for rows whose original creation version cannot be determined.
+/// Version 1 is the initial dataset version in the Lance format.
+const UNKNOWN_CREATED_AT_VERSION: u64 = 1;
```
andrei-ionescu, line 50: I would suggest to rename it to `DEFAULT_VERSION`.


```suggestion
const DEFAULT_VERSION: u64 = 1;
```

rust/lance/src/dataset/transaction.rs
```diff
@@ -45,6 +45,131 @@ use std::{
...
+    let mut row_id_to_source: HashMap<u64, (&Fragment, usize)> = HashMap::new();
+    // Stable row IDs must be globally unique among *live* rows, but after a rewrite-style
+    // update the same stable ID can appear twice in `existing_fragments`: once in an older
+    // fragment's inline `row_id_meta` at the original row offset (rows may be soft-deleted
+    // via a deletion vector) and again in a newer fragment holding rewritten data. For
+    // `created_at` we need the mapping from the original fragment/offset; that is always the
+    // first occurrence when fragments are processed in ascending `id` order.
+    let mut frag_indices: Vec<usize> = (0..existing_fragments.len()).collect();
+    frag_indices.sort_by_key(|&i| existing_fragments[i].id);
+    for &i in &frag_indices {
```
andrei-ionescu: I see that we do to iterations over the `existing_fragments`. Can this be improved?

rust/lance/src/dataset/transaction.rs
```diff
@@ -45,6 +45,131 @@ use std::{
...
+                    "Fragment {} has external row ID metadata; \
+                     version tracking will use defaults",
+                    fragment.id,
+                );
+                None
+            }
+            None => None,
+        };
+
+        if let Some(row_ids) = row_ids {
```
andrei-ionescu, line 118: I think this should be moved in the match above. This will also make code more readable and move the else branch in the None case.

rust/lance/src/dataset/transaction.rs
```diff
@@ -45,6 +45,131 @@ use std::{
...
+                    Error::internal(format!(
+                        "Failed to create created_at version metadata: {}",
+                        e
+                    ))
+                })?,
+            );
+
+            fragment.last_updated_at_version_meta = build_version_meta(fragment, new_version);
+        } else {
+            let version_meta = build_version_meta(fragment, new_version);
```
andrei-ionescu, line 139: This can be moved in the None case.

rust/lance/src/dataset/transaction.rs
```diff
@@ -2172,9 +2177,16 @@ impl Transaction {
         manifest.tag.clone_from(&self.tag);
 
         if config.auto_set_feature_flags {
+            // Internal operations (e.g. CreateIndex) use ManifestWriteConfig::default()
+            // which has use_stable_row_ids = false. Without inheriting from the previous
+            // manifest, apply_feature_flags would clear FLAG_STABLE_ROW_IDS.
+            let inherited = current_manifest
+                .map(|m| m.uses_stable_row_ids())
+                .unwrap_or(false);
+            let stable_row_ids = config.use_stable_row_ids || inherited;
```
andrei-ionescu: Nitpick... the name of the variable should let us understand that it is a boolean, so `use_stable_row_ids` would be better.

rust/lance/src/dataset.rs
```diff
@@ -3235,11 +3235,11 @@ pub(crate) async fn write_manifest_file(
...
     if config.auto_set_feature_flags {
-        apply_feature_flags(
-            manifest,
-            config.use_stable_row_ids,
-            config.disable_transaction_file,
-        )?;
+        // build_manifest may have already set FLAG_STABLE_ROW_IDS on the manifest.
+        // Preserve it here so this second apply_feature_flags call does not clear it
+        // when config.use_stable_row_ids is false (the ManifestWriteConfig default).
+        let stable_row_ids = config.use_stable_row_ids || manifest.uses_stable_row_ids();
```
andrei-ionescu: Ditto... `use_stable_row_ids` to show that is a boolean.

java/lance-jni/src/fragment.rs
ivscheianu: `'a` is actually the dominant convention across all the other `IntoJava` impls in this file (`DataFile`, `DeletionFile`, `RowIdMeta`, `FragmentMergeResult`, etc.). `Fragment` is actually the outlier. My main concern was about consistency, happy to change it if you think it's the appropriate move

rust/lance/src/dataset/transaction.rs
ivscheianu: I'd prefer to keep `UNKNOWN_CREATED_AT_VERSION` because the name captures why this value is used (fallback when the original creation version can't be determined), not just what it is. DEFAULT_VERSION would be a little ambiguous at the call sites since a reader would have to trace back to the constant definition to understand the intent

rust/lance/src/dataset/transaction.rs
ivscheianu: Folding the `if let Some(row_ids)` into the `Inline` arm would require duplicating the fallback `build_version_meta` logic across three places: the `Inline`/`Err` sub-arm, the `External` arm, and the `None` arm. The current structure keeps that fallback in one else branch, which is the only reason the intermediate `row_ids: Option<_>` variable exists. Agreed it's not the most elegant shape, but I think it's less duplication than the alternative

rust/lance/src/dataset/transaction.rs
ivscheianu: fixed 👍🏻

rust/lance/src/dataset.rs
ivscheianu: fixed 👍🏻

rust/lance/src/dataset/transaction.rs
ivscheianu: updated, thanks!

rust/lance/src/dataset/transaction.rs
ivscheianu: related to the above https://github.com/lance-format/lance/pull/6465#discussion_r3121871750

rust/lance/src/dataset/transaction.rs
```diff
@@ -45,6 +45,130 @@ use std::{
...
+/// For each new fragment produced by an update, set `created_at_version_meta`
+/// (preserved from the original rows) and `last_updated_at_version_meta`.
+fn resolve_update_version_metadata(
+    existing_fragments: &[Fragment],
+    new_fragments: &mut [Fragment],
+    new_version: u64,
+) -> Result<()> {
+    // ~24 bytes per entry (u64 key + &Fragment pointer + usize offset + hash metadata).
+    // For a 1M-row dataset this is ~24 MB — acceptable for a transient transaction structure.
+    let mut row_id_to_source: HashMap<u64, (&Fragment, usize)> = HashMap::new();
```
westonpace: Is this based on existing_fragments or new_fragments?  I'm worried this will not scale well as we approach 1B+ rows.

Can we use `updated_frags` instead of `existing_fragments`?  That should scope this to the size of the update at least.

westonpace (changes requested):

rust/lance/src/dataset/transaction.rs
ivscheianu: Good catch, thanks! Fixed by pre-computing `needed_row_ids` from `new_fragments` upfront and skipping irrelevant entries during the scan, bounding the map to O(updated rows)

ivscheianu:
Hey @westonpace, just a quick nudge whenever you have a moment, I’ve made the updates you suggested. Thanks!

westonpace (approved):
There's still some potentially expensive paths in here.  Per Claude:

```
  Step 1 — building row_id_to_source is O(dataset rows)                                                                                                                                       
                                                                                                                                                                                              
  let mut sorted_frags: Vec<&Fragment> = existing_fragments.iter().collect();                                                                                                                 
  sorted_frags.sort_by_key(|f| f.id);                                                                                                                                                         
  for frag in sorted_frags {                                                                                                                                                                  
      if let Some(RowIdMeta::Inline(data)) = &frag.row_id_meta                                                                                                                                
          && let Ok(seq) = read_row_ids(data) {                                                                                                                                               
          for (offset, rid) in seq.iter().enumerate() {                                        
              if needed_row_ids.contains(&rid) { ... }                                         
          }                                    
      }                                                                                                                                                                                     
  }                                            
                                                                                                                                                                                            
  This walks every row id of every existing fragment — i.e., the entire dataset's row-id index — even when an update touches only a handful of rows. Cost components per commit:              
  - Deserialize RowIdSequence for every existing fragment: O(F) protobuf decodes.
  - Expand each sequence to individual u64s and probe the HashSet: O(N) where N = total live rows.                                                                                            
                                                                                                                                                                                            
  For a 100M-row dataset where you update 1,000 rows, you still walk all 100M ids. The HashSet probe is cheap, but the iteration and the per-fragment protobuf decode are not free, and this  
  work happens on the commit critical path.    

  A cheap mitigation: skip fragments whose row-id range doesn't overlap needed_row_ids — U64Segment already exposes range(). With sorted needed_row_ids you can do an O(log M) overlap test   
  per fragment and bail before the inner loop.                                                                                                                                              

  A bigger mitigation: iterate needed_row_ids and locate each one's source fragment, instead of iterating every fragment to check every row. With segments sorted by range you'd be O(M log F)
   instead of O(N).
```

And also...

```
  Step 2 — resolve_created_at_version per row                                                                                                                                                 
   
  Every row in the new fragments triggers:                                                                                                                                                    
                                                                                                                                                                                            
  let seq = created_meta.load_sequence()?;        // deserialize protobuf
  seq.versions().nth(*row_offset)                 // walk iterator to offset                                                                                                                  
                                                                                                                                                                                              
  So for M updated rows mapping into a source fragment with N rows and K runs:                                                                                                                
  - load_sequence is re-decoded per row instead of per source fragment → M decodes.                                                                                                           
  - versions().nth(offset) walks through offset items each time → worst case O(M × N).                                                                                                        
                                                                                                                                                                                            
  With the mitigation I flagged in the P1 (cache load_sequence per source fragment + use version_at(index) which is O(K)), this drops to O(F_sources × decode + M × K) — usually orders of    
  magnitude less, since K ≪ N for typical version histories.
```

At the very least we should create some follow-up issues for these improvements.  I'm not terribly involved in the stable row id path and it's still fairly experimental as far as I know.  I'll give it a day or so.  If no one has any complaints I'm ok going with what you have here in the meantime.

ivscheianu:
Hi, @westonpace!

Both implemented:
1. added `RowIdSequence::row_id_range()` for the bounding-box pre-filter
2. a per-source-fragment `load_sequence()` cache replacing the per-row decode + versions().nth()

Thanks for the pointers!

westonpace:
Awesome, thanks for the help!

westonpace:
@westonpace — automated 24h merge-reminder: PR is already merged, nothing to do. _Posted by a scheduled Claude routine._

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #6468, opened 2026-04-09
feat(python): add target_partition_size parameter to train_ivf()

### LINKED ISSUE #6286: python: train_ivf() should accept target_partition_size instead of deprecated num_partitions
Currently, the Python API lance.index.train_ivf() only accepts num_partitions, which is marked as deprecated in the docstring.

However, the recommended and modern parameter target_partition_size is not exposed in the Python binding, even though it is fully supported in the Rust core.

fightBoxing:
## Summary

Add `target_partition_size` as the recommended parameter for `IndicesBuilder.train_ivf()`, replacing the deprecated `num_partitions`.

Closes #6286

### `python/python/lance/indices/builder.py`
- Added `target_partition_size: Optional[int] = None` keyword argument to `train_ivf()`
- When `num_partitions` is not specified, computes partition count using `target_partition_size` (default 8192)
- Uses the same algorithm as Rust's `recommended_num_partitions()`: `clamp(num_rows // target_partition_size, 1, 4096)`
- When both `num_partitions` and `target_partition_size` are specified, `num_partitions` takes precedence with a `DeprecationWarning`
- Updated docstrings to document the new parameter and deprecation

### `python/python/lance/util.py`
- **Bugfix**: `_target_partition_size_to_num_partitions()` used `max(1, num_partitions, 4096)` which always returned 4096. Fixed to `max(1, min(num_partitions, 4096))`

### `python/python/tests/test_indices.py`
- `test_target_partition_size`: verifies explicit `target_partition_size` computes correct partition count
- `test_target_partition_size_default`: verifies default behavior (no args) uses 8192
- `test_num_partitions_overrides_target`: verifies `num_partitions` takes precedence with warning

## Notes
- No Rust binding changes needed — the Python layer handles the `target_partition_size → num_partitions` conversion before calling into Rust
- The Rust core already fully supports `target_partition_size` in `IvfBuildParams`; this PR brings the Python `IndicesBuilder` API in sync

### DIFF
--- a/python/python/lance/indices/builder.py
+++ b/python/python/lance/indices/builder.py
@@ -1,7 +1,6 @@
 # SPDX-License-Identifier: Apache-2.0
 # SPDX-FileCopyrightText: Copyright The Lance Authors
 
-import math
 import warnings
 from dataclasses import dataclass
 from typing import TYPE_CHECKING, Optional, Union
@@ -67,6 +66,7 @@ class IndicesBuilder:
         sample_rate: int = 256,
         max_iters: int = 50,
         fragment_ids: Optional[list[int]] = None,
+        target_partition_size: Optional[int] = None,
     ) -> IvfModel:
         """
         Train IVF centroids for the given vector column.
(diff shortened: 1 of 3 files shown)


wjones127 (approved):
Logic from Rust is here:

https://github.com/lance-format/lance/blob/c3860e6fa8068fd919869a211433f5c49f8c42b4/rust/lance-index/src/vector/ivf/builder.rs#L115-L120

It looks like the clamp is temporary, which is good to note. But otherwise this looks good!

wjones127:
Since this is out-of-date, I've adapted your changes into a new PR: #7690. Closing this one in favor of that.

Issue #4193, opened 2025-07-10
Optimize upsert with partial schema and no index


wjones127:

wombatu-kun:
Linking from #6472: partial-schema merge_insert now routes through `create_plan` (the `execute_uncommitted_v2` path) for the non-indexed case, but the TODO above is still fully outstanding. #6472 does *not* introduce any `TakeNode`/`TakeExec` — it fills in missing columns with a naive `df.with_column(field, col("target.<field>"))` loop that reads every missing column across the *full* target scan, even when only a handful of rows actually match.

This makes the optimization tracked here newly relevant: the fill-in loop is the placeholder that steps 1 and 3 of the TODO are designed to replace. @wjones127 raised this in the PR review and asked that the post-join `TakeNode`/`TakeExec` direction be captured here concretely rather than left as a generic "revisit later".

wjones127:
The read-side optimization (deferring reads of non-source columns for matched rows) is split out into #7363, implemented as a late-materialization optimizer rule.

The write-side column-replacement path (in-place `RewriteColumns` in v2, currently only reachable on the v1/indexed path) is still in scope for this issue and should land before it's closed.

LuciferYang:
I'd like to take this, now that the read-side half is split out to #7363 and what's left here is the write-side in-place column replacement.

@wombatu-kun you wrote #6472 and know this path best. Are you already on it? I don't want to duplicate work.

The approach, in case something here is wrong: reuse `MergeInsertJob::update_fragments` as a v2 write sink instead of building a second column-write path. It already takes a `_rowaddr` + data-columns stream, sorts by row address, groups by fragment, has a no-read-back fast path for exact fragment coverage and an `Updater` path for sparse updates, and produces the tombstoned field ids. The v1 call site already wraps its output in `Operation::Update { update_mode: Some(RewriteColumns), .. }`.

So the new exec is glue: it drops the rows the action column rejects, projects down to `_rowaddr` + the source columns, hands that to `update_fragments`, and wraps the result in the transaction. The other half of the change is that `create_plan` skips the missing-column fill loop when it routes there. That loop is what keeps the wide target columns alive through projection pushdown, so skipping it also removes the read amplification #7363 describes: the target scan projects only the join key.

[...]

PR #6472, opened 2026-04-10
feat: route partial-schema merge_insert through the v2 write path

### LINKED ISSUE #6442: Support partial schema upsert on v2 merge_insert path
## What to build

When the source has fewer columns than the target, the v2 path should handle
the upsert instead of falling back to v1.

Simple approach: add a projection in the logical plan that fills missing
columns from the target side of the join. The v2 exec writes full rows as it
does today.

This intentionally does NOT replicate the v1 optimization of marking columns
as dropped and writing only changed columns via `Operation::Update` with
`RewriteColumns`. That optimization is tracked separately in #4193.

See milestone description for full PRD.

## Acceptance criteria

- [ ] Partial schema upserts use the v2 path (verified via `explain_plan`)
- [ ] Existing partial schema tests pass unchanged
- [ ] New test verifies columns not in source retain original values after upsert
- [ ] New `explain_plan` test confirms v2 is used
- [ ] Conflict detection (Bloom filter) works for partial schema operations

## Blocked by

None - can start immediately.

## User stories addressed

- 2: partial schema upserts use v2 path
- 4: one fewer case on v1, moving toward single code path

wombatu-kun:
## Summary

Closes #6442.

Partial-schema upserts (source column subset of the dataset schema) used to fall back to the legacy v1 `Merger` path, which wrote only the changed columns via `Operation::Update` with `RewriteColumns`. They now run through the same `create_plan` / `FullSchemaMergeInsertExec` pipeline as full-schema upserts. Missing source columns are filled from the target side of the join via a post-join projection, so the v2 writer sees a complete set of data columns and writes full rows as new fragments.

Per the issue, this intentionally does **not** replicate v1's `RewriteColumns` optimization — that is tracked separately in #4193. The v1 path is still used when the join key has a scalar index (falls through the old gate), so the legacy optimization remains available for the scalar-index-on-join-key case.

### How it works

1. **`can_use_create_plan`** (gate) — accepts a source schema that is a type-compatible subset of the dataset schema. If `insert_not_matched=InsertAll` and any missing target column is non-nullable, the call short-circuits with an `InvalidInput` error naming the offending columns.
2. **`create_plan`** (plan construction) — after the join and the `__action` column, every dataset field missing from the source is added as an unqualified column populated from `col("target.\"<name>\"")`. For matched rows this carries the existing target value; for unmatched source rows the outer join leaves it NULL.
3. **`necessary_children_exprs`** (projection pushdown) — the extension node now keeps unqualified columns whose name matches a dataset field, so the synthetic filled columns flow through to the write exec alongside `source.*`.
4. **`prepare_stream_schema`** — data columns are now emitted in *dataset schema order* keyed by name. This turns an accidental positional invariant into an explicit name-based one and is required for the partial-schema path (synthetic filled columns land at the end of the logical projection).

### Consequences

- `explain_plan()` now works for partial-schema upserts.
- Bloom-filter conflict detection (`inserted_rows_filter`) is populated for partial-schema operations on unenforced primary keys; v1 always returned `None`.
- `when_not_matched_by_source=Delete`/`DeleteIf` is accepted for partial-schema sources (previously rejected outright with a `NotSupported` error).
- Partial-schema upserts with `insert_not_matched=InsertAll` now reject non-nullable missing columns at the API boundary with a descriptive error naming the offending columns, instead of producing a confusing downstream writer error. **User-visible behavior change.**

### Tests

- Refactored `test_merge_insert_subcols` to branch on `scalar_index`: v1 structural assertions (tombstoned fields, preserved fragment ids) stay on the `scalar_index=true` branch; the `scalar_index=false` branch asserts v2 structural behavior. A shared key-lookup-based check verifies that columns not in source retain original values across both paths.
- Converted the old negative `test_delete_not_supported` into a positive `test_delete_not_matched_by_source_on_v2_subcols` test.
- Updated `test_sub_schema_upsert_fragment_bitmap` to reflect v2 semantics (3 fragments after upsert, both indexes preserved — v2 relies on unindexed-fragment fallback rather than eagerly invalidating the vector index).
- New `test_merge_insert_subcols_v2_explain_plan` confirms the v2 physical plan is used and that the filled `other` column appears in the projection.
- New `test_merge_insert_subcols_v2_rejects_non_nullable_insert` for the new validation error.
- New `test_merge_insert_subcols_v2_bloom_filter` for the bloom-filter conflict-detection acceptance criterion.
- Updated the Python `test_merge_insert_subcols` to drop v1-specific file-layout assertions while preserving the semantic data checks (column `c` unchanged for updated rows, `NULL` for newly inserted rows).

## Test plan

- [x] `cargo check -p lance --lib --tests`
- [x] `cargo clippy -p lance --lib --tests -- -D warnings`
- [x] `cargo fmt --all`
- [x] `cargo test -p lance --lib dataset::write::merge_insert` — 127 passed
- [x] `cargo test -p lance --lib dataset::write` — 190 passed
- [x] `cargo test -p lance --lib dataset::` — 1058 passed
- [x] `cargo test -p lance --lib index::tests` — 63 passed
- [ ] Python tests (`pytest python/python/tests/test_dataset.py::test_merge_insert_subcols`) — run in CI


rust/lance/src/dataset/write/merge_insert.rs
```diff
@@ -1377,6 +1386,26 @@ impl MergeInsertJob {
...
+        // unqualified name matches the dataset field and becomes a normal
+        // data column from the write exec's perspective.
+        //
+        // We iterate the dataset schema in order so that the resulting
+        // physical plan is deterministic and easy to inspect in tests.
+        for field in dataset_schema.fields() {
+            if !source_field_names.contains(field.name()) {
+                df = df.with_column(
+                    field.name(),
+                    logical_expr::col(format!("target.\"{}\"", field.name())),
```
wjones127, line 1414: issue(blocking): in my experience, `col()` does a poor job handling `camelCase` column names. Could you add a test that involves omitting a camel case column?

rust/lance/src/dataset/write/merge_insert.rs
```diff
@@ -1377,6 +1386,26 @@ impl MergeInsertJob {
...
+        // for unmatched source rows (inserts) the outer join leaves the
+        // target side NULL, so inserts get NULL for missing columns. The
+        // unqualified name matches the dataset field and becomes a normal
+        // data column from the write exec's perspective.
+        //
+        // We iterate the dataset schema in order so that the resulting
+        // physical plan is deterministic and easy to inspect in tests.
+        for field in dataset_schema.fields() {
+            if !source_field_names.contains(field.name()) {
+                df = df.with_column(
```
wjones127, line 1412: issue(non-blocking): scanning these columns for all rows isn't ideal performance-wise. Consider the case where you are updating 3 rows with just part of the schema. This means you need to read all rows for missing columns in the table just to do this update. Ideally, we'd instead pull these columns in using TakeExec *after* the join.

I'm okay with keeping this code path for now, but if we do we should make a follow up ticket for optimizing to a Take later.

wjones127 (commented):
Thanks for working on this. I'd like to see some tests to make sure this doesn't break in cases where there are camel case columns. Otherwise I think it looks good.

rust/lance/src/dataset/write/merge_insert.rs
wombatu-kun: Added test with camelCase columns

rust/lance/src/dataset/write/merge_insert.rs
wombatu-kun: Agreed — materializing every missing column across the full target scan is wasteful when the update set is small. The follow-up is already tracked in #4193 ("Optimize upsert with partial schema and no index"), which is linked from this PR's description. I'll add a note on #4193 pointing at the post-join `TakeExec` approach you described so the optimization direction is captured concretely rather than as a generic "revisit later".

wjones127 (approved):
This is looking good. Apologies for the delay in review. Thanks for working on this!

wjones127:
Updated the branch. Will merge once CI is passing.

PR #6474, opened 2026-04-10
feat: support updating by stable row ids

majin1102:
This is another aspect of #6438 and #6484
## Summary
- allow `dataset.update` to parse `_rowid` in filters and update expressions
- reject `_rowid` usage when stable row ids are not enabled on the dataset
- evaluate update expressions before capturing and stripping `_rowid` so expressions can use it at execution time
- add coverage for single-row and multi-row `_rowid` updates plus stable-row-id validation


majin1102:
@codex review

majin1102:
@codex review

PR #6475, opened 2026-04-10
feat(vector): add partition search parallelism

BubbleCal:
## Feature

This PR adds query-time `query_parallelism` for vector search partition execution and wires it through Rust, Python, and Java.

Callers can control how many IVF partitions a single query may search concurrently:
- `0`: auto policy. The current implementation maps auto to the single-worker sequential path.
- `1`: single-worker sequential partition search.
- `-1`: use the CPU pool size.
- `>= 2`: partition-parallel search, clamped to the CPU pool size.

The default is `0`, which currently preserves the optimized sequential execution path.

## Performance Improvement

The performance issue is per-query worker fan-out. With many concurrent queries, spawning a CPU task for every partition of every query increases contention on the CPU worker pool and lengthens queueing time. This is especially visible for fixed-`nprobes` IVF workloads where every query searches the same number of partitions.

This PR improves that path by making query partition scheduling configurable while keeping the optimized sequential model as the default auto behavior. Sequential fixed-`nprobes` search prepares partitions asynchronously, then searches prepared partitions on one CPU worker using a query-level global top-k heap. Parallel execution remains available for workloads that benefit from intra-query partition parallelism.

## Implementation

- Rust, Python, and Java expose `query_parallelism` on vector search APIs.
- `ANNIvfSubIndexExec` converts the configured value into an effective partition concurrency.
- Effective partition concurrency is clamped to the CPU pool size.
- IVF v2 splits partition search into async prepare and sync execute phases.
- Sequential fixed-`nprobes` search uses a query-level global top-k heap.
- Sequential late-search keeps the existing per-partition output shape so early-stop behavior is preserved.
- Parallel execution uses direct per-partition search tasks, matching the original execution model.
- IVF_RQ/Flat sub-index search reuses caller-owned scratch buffers for RQ distance-table quantization and top-k accumulation to reduce allocation overhead.

## Developer Impact

- Rust callers can call `Scanner::query_parallelism(...)`.
- Python callers can pass `query_parallelism` in vector search APIs.
- Java callers can use `Query.Builder#setQueryParallelism(...)`.
- `parallel_mode` / `ParallelMode` have been replaced by the concurrency-based API.

## Benchmark

GCP VM benchmark using the same index for all modes. All rows had matching result checksums.

Configuration: `gist / IVF_RQ / target_partition_size=8192 / k=100 / nprobes=20 / columns=[] / prewarm / max_queries=1000`. Percentages are relative to `main`.

| Threads | Mode | Avg | P50 | P90 | P95 | P99 | QPS |
|---:|---|---:|---:|---:|---:|---:|---:|
| 8 | main | 4.98 ms | 4.92 ms | 6.28 ms | 6.86 ms | 7.56 ms | 1584.6 |
| 8 | sequential | 4.27 ms (-14.2%) | 4.21 ms (-14.5%) | 5.03 ms (-19.9%) | 5.27 ms (-23.1%) | 5.97 ms (-21.0%) | 1838.9 (+16.1%) |
| 8 | parallel | 5.02 ms (+0.7%) | 4.97 ms (+0.9%) | 6.27 ms (-0.2%) | 6.60 ms (-3.7%) | 7.36 ms (-2.7%) | 1575.4 (-0.6%) |
| 16 | main | 9.95 ms | 9.74 ms | 13.12 ms | 14.11 ms | 16.83 ms | 1583.1 |
| 16 | sequential | 8.09 ms (-18.6%) | 7.98 ms (-18.0%) | 9.84 ms (-25.0%) | 10.47 ms (-25.8%) | 11.67 ms (-30.6%) | 1936.8 (+22.3%) |
| 16 | parallel | 9.95 ms (+0.0%) | 9.68 ms (-0.6%) | 13.37 ms (+1.9%) | 14.42 ms (+2.2%) | 16.83 ms (+0.0%) | 1583.6 (+0.0%) |
| 32 | main | 18.68 ms | 18.16 ms | 26.66 ms | 28.77 ms | 33.12 ms | 1652.7 |
| 32 | sequential | 15.50 ms (-17.0%) | 15.15 ms (-16.6%) | 20.57 ms (-22.9%) | 22.16 ms (-23.0%) | 25.32 ms (-23.6%) | 2000.6 (+21.1%) |
| 32 | parallel | 19.13 ms (+2.4%) | 18.58 ms (+2.3%) | 26.49 ms (-0.6%) | 29.12 ms (+1.2%) | 33.92 ms (+2.4%) | 1625.9 (-1.6%) |
| 64 | main | 33.98 ms | 33.01 ms | 49.65 ms | 53.87 ms | 63.58 ms | 1718.4 |
| 64 | sequential | 29.95 ms (-11.8%) | 29.67 ms (-10.1%) | 42.44 ms (-14.5%) | 46.81 ms (-13.1%) | 54.42 ms (-14.4%) | 1949.4 (+13.4%) |
| 64 | parallel | 35.17 ms (+3.5%) | 34.37 ms (+4.1%) | 50.70 ms (+2.1%) | 55.04 ms (+2.2%) | 69.09 ms (+8.7%) | 1650.7 (-3.9%) |
| 128 | main | 38.73 ms | 36.07 ms | 68.28 ms | 76.78 ms | 96.81 ms | 1663.3 |
| 128 | sequential | 37.48 ms (-3.2%) | 35.90 ms (-0.5%) | 65.26 ms (-4.4%) | 71.82 ms (-6.5%) | 87.48 ms (-9.6%) | 1915.3 (+15.2%) |
| 128 | parallel | 41.29 ms (+6.6%) | 39.38 ms (+9.2%) | 71.68 ms (+5.0%) | 80.70 ms (+5.1%) | 96.20 ms (-0.6%) | 1577.2 (-5.2%) |

## Validation

- `cargo fmt --all --check`
- `cd python && cargo check`
- `cd java && cargo check --manifest-path ./lance-jni/Cargo.toml`
- `uv run --with maturin maturin develop`
- `uv run pytest python/tests/test_vector_index.py::test_vector_index_with_query_parallelism python/tests/test_vector_index.py::test_vector_index_invalid_query_parallelism`
- Python ruff check / format check for touched Python files


BubbleCal:
@claude review

wjones127 (commented):
Design question: what is the advantage of having a Boolean parameter for whether to use parallelism instead of an integer specifying number of cores? In DataFusion nodes, usually you just configure parallelism based on the number of partitions, and the program running the plan can choose the parallelism. What do you think of moving towards an arrangement like that?

edit: I guess there’s also the IO parallelism involved. I’d expect that to be controlled by the object store’s IO parallelism and run in a separate background task.

BubbleCal:
> Design question: what is the advantage of having a Boolean parameter for whether to use parallelism instead of an integer specifying number of cores? In DataFusion nodes, usually you just configure parallelism based on the number of partitions, and the program running the plan can choose the parallelism. What do you think of moving towards an arrangement like that?
> 
> edit: I guess there’s also the IO parallelism involved. I’d expect that to be controlled by the object store’s IO parallelism and run in a separate background task.

Totally agree, moving to accepting an integer to control the parallelism, I will keep `0` as `Auto` and default, it's equal to `1` for now but we may make it real "auto" in the future.

python/python/lance/dataset.py
```diff
@@ -6778,6 +6782,12 @@ def _build_vector_search_query(
         Whether to use the index for the search.
     ef: int, optional
         The ef parameter for HNSW search.
+    partition_parallelism: int, optional
+        Maximum number of partitions to search concurrently. Value 0 uses the
+        automatic policy, which currently maps to the single-worker sequential
+        path. Value -1 uses the CPU pool size. Value 1 uses the single-worker
+        sequential path. Values >= 2 use the
+        partition-parallel path and are clamped to the CPU pool size.
```
westonpace: We should mention the default

westonpace (approved):
Without looking too much into the details (Claude had a few thoughts but none of them that useful) I would say that I'm a bit +1 on the ability to control this.  I've been doing this in some of my higher level testing by swapping the tokio multi-threaded runtime out for a single-threaded runtime per-core but then it still adds some overhead so this is a nice approach.

I think my only thought would be that this could eventually be a more generic parameter.  "partition parallelism" sounds very specific to IVF.  Perhaps we just want a generic "query parallelism" that controls how many threads we use to satisfy the query?

For now it can do exactly this but in the future we could look at this value to determine things like FTS search fan-out as well as IVF fanout.

PR #6479, opened 2026-04-10
feat: preserve keys_sorted for PyArrow map types

### LINKED ISSUE #6478: Support PyArrow sorted MapType schemas
## Problem

Lance currently rejects or loses Arrow `MapType` schemas with `keys_sorted=true` in several paths. This blocks round-tripping PyArrow schemas such as:

```python
pa.map_(pa.string(), pa.int32(), keys_sorted=True)
```

The limitation was originally called out in code comments around map support because Rust/PyArrow schema conversion could drop the `keys_sorted` property.

## Expected behavior

Lance should preserve `MapType.keys_sorted` for V2.2 map data across write and read paths, including:

- dataset writes and reads
- scanner/table/batch reads
- fragment writes and reads
- direct file reader APIs
- Lance schema conversion APIs
- namespace schema JSON conversion

## Notes

The implementation should keep normal unsorted maps unchanged and avoid adding per-batch Python wrapping for schemas that do not need sorted-map restoration.

myandpr:
Closes #6478

## Summary

- Preserve Arrow MapType.keys_sorted through Lance schema/logical type conversion and V2.2 map encoding/decoding.
- Restore sorted-map schemas at Python Arrow API exits, including dataset, scanner, fragment, direct file reader, and LanceSchema/LanceField conversions.
- Preserve sorted-map input schemas when converting Python readers for dataset and fragment write paths.

## Testing

- python -m py_compile python/python/lance/file.py python/python/lance/schema.py python/python/tests/test_map_type.py
- pytest python/python/tests/test_map_type.py -q
- pytest python/python/tests/test_file.py -q
- cargo test --manifest-path python/Cargo.toml --lib
- cargo test -p lance-namespace schema::tests
- cargo test -p lance-core map_keys_sorted_round_trip
- cargo test -p lance-encoding test_sorted_map


myandpr:
quick note: windows-build is failing in a namespace concurrent-drop test with TooMuchWriteContention. This PR only changes sorted-map schema handling and does not touch lance-namespace-impls, 

so this looks unrelated.

PR #6480, opened 2026-04-10
fix: include column_metadatas and column_infos in CachedFileMetadata::DeepSizeOf

jiaoew1991:
## Summary

- Fix `CachedFileMetadata::DeepSizeOf` to include `column_metadatas` and `column_infos` — the two largest fields that were previously omitted (marked TODO since initial implementation)
- This caused the moka cache weigher to underestimate entry sizes by ~100x, preventing eviction and causing unbounded memory growth on random-access workloads

## Problem

`LanceCache` uses moka with a weighted capacity of 1 GB (`DEFAULT_METADATA_CACHE_SIZE`). The weigher calls `DeepSizeOf` on `CachedFileMetadata`, but the implementation only counted `file_schema` and `file_buffers` — ignoring `column_metadatas` (protobuf `ColumnMetadata`) and `column_infos` (`Vec<Arc<ColumnInfo>>` containing page encodings).

Each cache entry's true size is hundreds of KB, but was reported as ~1 KB. Moka never reached the 1 GB limit, so entries accumulated indefinitely.

## Profiling Evidence

Tested on a 221M-row dataset with random `ds.take()` (64 rows per call):

| Metric | Before | After |
|--------|--------|-------|
| RSS growth (30 iters) | **+7,503 MB** | **+535 MB** |
| Growth rate | 250 MB/iter (linear, no plateau) | 18 MB/iter (plateaus ~500 MB) |

jemalloc heap profiling (debug build, 243K symbols) showed 99.9% of leaked memory in `LanceCache::get_or_insert_with_key` → `FileReader::meta_to_col_infos` and `prost::encoding::message::merge`.

## Approach

Since protobuf-generated types (`pbfile::ColumnMetadata`, `pb::ColumnEncoding`, etc.) don't implement `DeepSizeOf`, we use `prost::Message::encoded_len() * 4` as an approximation for in-memory size. The 4x multiplier accounts for heap allocations in repeated/string/bytes fields that are larger in memory than on the wire.

## Test plan

- [x] Added `test_deep_size_of_includes_column_metadata` for V2_0 and V2_1 file formats
- [x] Verified fix reduces memory growth from 250 MB/iter to 18 MB/iter on a production dataset
- [x] `cargo test -p lance-file` passes

### DIFF
--- a/rust/lance-file/src/reader.rs
+++ b/rust/lance-file/src/reader.rs
@@ -148,9 +148,7 @@ impl DeepSizeOf for CachedFileMetadata {
                     .iter()
                     .map(|pi| {
                         let enc_size = match &pi.encoding {
-                            lance_encoding::decoder::PageEncoding::Legacy(e) => {
-                                e.encoded_len() * 4
-                            }
+                            lance_encoding::decoder::PageEncoding::Legacy(e) => e.encoded_len() * 4,
                             lance_encoding::decoder::PageEncoding::Structural(e) => {
                                 e.encoded_len() * 4
                             }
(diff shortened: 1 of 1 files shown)


BubbleCal (approved):
Thanks!

Issue #6483, opened 2026-04-11
Support Tiered in-memory + SSD cache for data


jaystarshot:
Having an inmemory + SSD cache for data files could be very useful in the lance API. 

Heres how it might look like

```
ds = lance.dataset(                                                                                                                     
      "s3://bucket/nuscenes.lance",                                                                                                       
      storage_options={                                                                                                                   
          "ssd_cache_dir": "/mnt/nvme/lance_cache",                                                                        
          "ssd_cache_size_mb": 100_000,
           "max_memory_cache_mb":1000,                                                    
      }
                                                                                                                                   
  ) 
```    

How it should work:                                                                                                                     
                                                                                                                                          
Two-tier cache modelled after the Velox caching system:                                                                                 
                                                                                                                                          
```
Remote GCS/S3                                                                                                                           
      │  fetch on miss                                                                                                                    
      ▼                                                                                                                                   
Memory cache (hot, bounded)                                                                                                             
      │  evict to SSD when memory threshold hit                                                                                           
      ▼                                                                                                                                   
SSD cache (warm, bounded)                                                                                                               
      │  evict LRU when SSD limit hit                                                                                                     
      ▼                                                  
 Discard   
```

jaystarshot:
cc: @beinan

PR #6487, opened 2026-04-11
feat: implement tiered Async data cache

jaystarshot:
#6483

Implements a Velox-style two-tier async data cache for Lance I/O — memory (L1) + SSD (L2) — so that repeated reads of the same dataset avoid Object storage round-trips.

Configured via storage_options at dataset open time:

Cache scope is valid for each dataset (stored in session)

  ```
ds = lance.dataset(
      "s3://bucket/nuscenes.lance",
      storage_options={
             "data_cache_enabled":      "true",
            "data_cache_memory_bytes": str(512 * 1024),
             "data_cache_ssd_enabled":  "true",
            "data_cache_ssd_dir":      SSD_DIR,
            "data_cache_ssd_bytes":    str(2 * 1024 * 1024 * 1024),
      },
  )
```

  

## Architecture

Doc for comments

https://docs.google.com/document/d/1k3UE8cnTSHoLJ43JGi-S4fJ0J3CLGe8LbRTjqH8puAU/edit?tab=t.0#heading=h.qmi7906oyigc

## Configuration

| Key | Type | Description |
|-----|------|-------------|
| `data_cache_enabled` | `"true"` | Master switch — must be set to enable the cache |
| `data_cache_memory_bytes` | bytes | L1 memory capacity (e.g. `"10737418240"` for 10 GiB) |
| `data_cache_ssd_enabled` | `"true"` | Enable the L2 SSD tier |
| `data_cache_ssd_dir` | path | Directory to write SSD shard files into |
| `data_cache_ssd_bytes` | bytes | L2 SSD capacity |
| `data_cache_memory_shards` | `16' | Number of L1 shards |
| `data_cache_ssd_shards` | `4` | Number of L2 shards |
| `data_cache_check_rtt_enabled` | `"true"` | Verify every cache hit against the object store (correctness testing only) |

## Cache stats

`ds.session().cache_stats()` returns `None` when no cache is configured,
or a dict with the following keys when enabled:

| Key | Description |
|-----|-------------|
| `memory_hits` | Ranges served from L1 |
| `memory_misses` | L1 misses (went to L2 or L3) |
| `memory_evictions` | Entries evicted from L1 |
| `memory_current_bytes` | Bytes currently held in L1 |
| `memory_stale_evictions` | L1 entries evicted because cached size < requested length |
| `ssd_hits` | L1 misses served from L2 |
| `ssd_bytes_written` | Total bytes written to L2 |
| `ssd_stale_misses` | L2 entries skipped because cached size < requested length |

```python
stats = ds.session().cache_stats()
if stats:
    hit_rate = stats["memory_hits"] / max(stats["memory_hits"] + stats["memory_misses"], 1)
    print(f"L1 hit rate:  {hit_rate * 100:.1f}%")
    print(f"L1 size:      {stats['memory_current_bytes'] / 1024**3:.2f} GiB")
    print(f"SSD hits:     {stats['ssd_hits']}")
```

Impact

Simple benchmark on OCI 

```
Cold
ds = lance.dataset(DATASET_URI, storage_options={})
ds.scanner({}).to_table()

Warm
ds.scanner({}).to_table()


```

```
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
TEST 1: L1 Memory-only cache (10 GiB)
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
Cache enabled: True

COLD (OCI fetch → L1)
  Time:                  9.651s
  OCI IOPS:              842
  OCI bytes:             883,483 B
  L1 memory hits:        8
  L1 memory misses:      842
  L1 stale evictions:    0
  L2 SSD hits:           0
  L2 SSD stale misses:   0
  SSD bytes written:     0 B
  Object store fetches:  842
  L1 size:               862.8 KiB
  L1 hit rate:           0.9%

WARM (should be 100% L1 hits)
  Time:                  0.261s
  OCI IOPS:              0
  OCI bytes:             0 B
  L1 memory hits:        420
  L1 memory misses:      0
  L1 stale evictions:    0
  L2 SSD hits:           0
  L2 SSD stale misses:   0
  SSD bytes written:     0 B
  Object store fetches:  0
  L1 size:               862.8 KiB
  L1 hit rate:           100.0%

```

```

━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
TEST 2: L1 + SSD two-tier cache
  L1 memory: 512 KiB  (small → forces eviction to SSD)
  L2 SSD:    1 GiB   (/Users/jay.narale/work/Uber/test)
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
SSD dir: /Users/jay.narale/work/Uber/test
Cache enabled: True

COLD (OCI fetch → L1 → evict to SSD)
  Time:                  9.075s
  OCI IOPS:              842
  OCI bytes:             883,483 B
  L1 memory hits:        8
  L1 memory misses:      842
  L1 stale evictions:    0
  L2 SSD hits:           0
  L2 SSD stale misses:   0
  SSD bytes written:     356,277 B
  Object store fetches:  842
  L1 size:               354.0 KiB
  L1 hit rate:           0.9%

WARM (L1 miss → should hit SSD)
  Time:                  5.157s
  OCI IOPS:              51
  OCI bytes:             75,384 B
  L1 memory hits:        326
  L1 memory misses:      94
  L1 stale evictions:    0
  L2 SSD hits:           43
  L2 SSD stale misses:   0
  SSD bytes written:     51,392 B
  Object store fetches:  51
  L1 size:               369.5 KiB
  L1 hit rate:           77.6%

────────────────────────────────────────────────────────────
  TWO-TIER RESULTS
────────────────────────────────────────────────────────────
  Speedup:            1.8x
  IOPS saved:         93.9%  (cold=842, warm=51)
  Object store calls: 51  (should be 0)
  L1 hits:            326
  L2 SSD hits:        43
```

Single cold scan with take_rows (pytorch approach) but sequential rows accessed (1000 rows)

```
ds = lance.dataset(DATASET_URI, storage_options=storage_opts)
for i in range(NUM_ROWS):
        table = ds.take([i])
        total_rows += table.num_rows

```

```
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
TEST: No cache
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
Cache enabled: False
    Time:          1521.040s
    Rows fetched:  10,000
    OCI IOPS:      420430
    L1 hits:       0
    L1 stale evictions: 0
    SSD hits:      0
    SSD stale misses:   0
    OCI fetches:   0
    L1 cache size: 0.0 KiB

━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
TEST: L1 memory (10 GiB)
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
Cache enabled: True
    Time:          107.867s
    Rows fetched:  10,000
    OCI IOPS:      842
    L1 hits:       419588
    L1 stale evictions: 0
    SSD hits:      0
    SSD stale misses:   0
    OCI fetches:   842
    L1 cache size: 862.8 KiB
    L1 hit rate:   99.8%
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
TEST: L1 (512 KiB) + SSD (2 GiB)
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
Cache enabled: True
    Time:          119.403s
    Rows fetched:  10,000
    OCI IOPS:      929
    L1 hits:       339928
    L1 stale evictions: 0
    SSD hits:      79573
    SSD stale misses:   0
    OCI fetches:   929
    L1 cache size: 306.6 KiB
    L1 hit rate:   80.9%

```

### DIFF
--- a/rust/lance-io/src/data_cache/mod.rs
+++ b/rust/lance-io/src/data_cache/mod.rs
@@ -192,6 +192,12 @@ struct SsdWriter {
 #[cfg(unix)]
 const MAX_WRITE_RATIO: usize = 70;
 
+/// Minimum number of L1 hits an entry must have accumulated before it is
+/// eligible to be written to the SSD tier on eviction.  Entries below this
+/// threshold are one-hit wonders and are dropped rather than polluting SSD.
+#[cfg(unix)]
+const SSD_MIN_HITS: u32 = 3;
+
 #[cfg(unix)]
 impl SsdWriter {
     fn new(ssd: Arc<SsdCache>) -> Arc<Self> {
@@ -210,7 +216,7 @@ impl SsdWriter {
 
 #[cfg(unix)]
 impl memory::EvictionSink for SsdWriter {
-    fn on_evicted(&self, entries: Vec<(DataCacheKey, Bytes)>, _total_cache_bytes: u64) {
+    fn on_evicted(&self, entries: Vec<(DataCacheKey, Bytes, u32)>, _total_cache_bytes: u64) {
         if self
             .write_in_progress
             .compare_exchange(false, true, Ordering::AcqRel, Ordering::Relaxed)
@@ -219,8 +225,22 @@ impl memory::EvictionSink for SsdWriter {
             return;
         }
 
-        let max = (entries.len() * MAX_WRITE_RATIO / 100).max(1);
-        let batch: Vec<_> = entries.into_iter().take(max).collect();
+        // Only admit entries that have been accessed at least SSD_MIN_HITS times.
+        // Entries below the threshold are one-hit wonders and are dropped to
+        // prevent scan-once data from polluting the SSD tier.
+        let admitted: Vec<(DataCacheKey, Bytes)> = entries
+            .into_iter()
+            .filter(|(_, _, num_uses)| *num_uses >= SSD_MIN_HITS)
+            .map(|(key, bytes, _)| (key, bytes))
+            .collect();
+
+        let max = (admitted.len() * MAX_WRITE_RATIO / 100).max(1);
+        let batch: Vec<_> = admitted.into_iter().take(max).collect();
+
+        if batch.is_empty() {
+            self.write_in_progress.store(false, Ordering::Release);
+            return;
+        }
 
         let ssd = self.ssd.clone();
         let flag = self.write_in_progress.clone();
@@ -652,8 +672,36 @@ mod tests {
         let path = Path::from("s3://bucket/data.lance");
         let entry_size = 1024 * 1024u64;
         let n_load = 6u64;
-
-        for i in 0..n_load {
+        // L1 holds 2 entries (2 MiB limit, 1 MiB each). Load them first, then
+        // re-access enough times to reach SSD_MIN_HITS, then overflow L1 so the
+        // hot entries are evicted and admitted to SSD.
+        let l1_capacity = 2u64;
+        for i in 0..l1_capacity {
+            let data = Bytes::from(vec![(i % 256) as u8; entry_size as usize]);
+            cache
+                .get_or_load(
+                    &path,
+                    i * entry_size,
+                    entry_size,
+                    Box::pin(async move { Ok(data) }),
+                )
+                .await
+                .unwrap();
+        }
+        for _ in 0..(SSD_MIN_HITS - 1) {
+            for i in 0..l1_capacity {
+                cache
+                    .get_or_load(
+                        &path,
+                        i * entry_size,
+                        entry_size,
+                        Box::pin(async { panic!("must hit L1") }),
+                    )
+                    .await
+                    .unwrap();
+            }
+        }
+        for i in l1_capacity..n_load {
             let data = Bytes::from(vec![(i % 256) as u8; entry_size as usize]);
             cache
                 .get_or_load(
@@ -720,7 +768,35 @@ mod tests {
             let cache = TieredDataCache::new(&config).await.unwrap();
             let path = Path::from("s3://bucket/data.lance");
 
-            for i in 0..4u64 {
+            // L1 holds 2 entries (128 KiB / 64 KiB). Load them, re-hit enough
+            // times to reach SSD_MIN_HITS, then overflow so they spill to SSD.
+            let l1_capacity = 2u64;
+            for i in 0..l1_capacity {
+                let data = Bytes::from(vec![correct_pattern; entry_size as usize]);
+                cache
+                    .get_or_load(
+                        &path,
+                        i * entry_size,
+                        entry_size,
+                        Box::pin(async move { Ok(data) }),
+                    )
+                    .await
+                    .unwrap();
+            }
+            for _ in 0..(SSD_MIN_HITS - 1) {
+                for i in 0..l1_capacity {
+                    cache
+                        .get_or_load(
+                            &path,
+                            i * entry_size,
+                            entry_size,
+                            Box::pin(async { panic!("must hit L1") }),
+                        )
+                        .await
+                        .unwrap();
+                }
+            }
+            for i in l1_capacity..4u64 {
                 let data = Bytes::from(vec![correct_pattern; entry_size as usize]);
                 cache
                     .get_or_load(
@@ -801,7 +877,35 @@ mod tests {
             let cache = TieredDataCache::new(&config).await.unwrap();
             let path = Path::from("s3://bucket/data.lance");
 
-            for i in 0..4u64 {
+            // L1 holds 2 entries (128 KiB / 64 KiB). Load them, re-hit enough
+            // times to reach SSD_MIN_HITS, then overflow so they spill to SSD.
+            let l1_capacity = 2u64;
+            for i in 0..l1_capacity {
+                let data = Bytes::from(vec![correct_pattern; entry_size as usize]);
+                cache
+                    .get_or_load(
+                        &path,
+                        i * entry_size,
+                        entry_size,
+                        Box::pin(async move { Ok(data) }),
+                    )
+                    .await
+                    .unwrap();
+            }
+            for _ in 0..(SSD_MIN_HITS - 1) {
+                for i in 0..l1_capacity {
+                    cache
+                        .get_or_load(
+                            &path,
+                            i * entry_size,
+                            entry_size,
+                            Box::pin(async { panic!("must hit L1") }),
+                        )
+                        .await
+                        .unwrap();
+                }
+            }
+            for i in l1_capacity..4u64 {
                 let data = Bytes::from(vec![correct_pattern; entry_size as usize]);
                 cache
                     .get_or_load(
--- a/rust/lance-io/src/data_cache/mod.rs
+++ b/rust/lance-io/src/data_cache/mod.rs
@@ -1,0 +1,1042 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright The Lance Authors
+
+//! Async two-tier data cache for Lance I/O.
+//!
+//! # Design
+//!
+//! The two-tier cache (memory + SSD) is inspired by Meta's engine.
+//! The core algorithms — shard-based memory management, clock-hand eviction
+//! with percentile threshold, load deduplication, region-based SSD layout,
+//! and coalesced reads — follow the same principles, adapted for Rust's async
+//! model and Lance's IO stack.
+//!
+//! # Architecture
+//!
+//! ```text
+//! FileScheduler::submit_request()
+//! │
+//! ├─ L1: MemoryCache (16 shards, clock-hand eviction, ~microseconds)
+//! │ HIT → return bytes immediately
+//! │
+//! ├─ L2: SsdCache (region files, coalesced pread, ~milliseconds)
+//! │ HIT → populate L1 → return
+//! │
+//! └─ L3: object store (network, tens–hundreds of ms)
+//!     → populate L1 → return
+//!         (L1 eviction async writes to L2)
+//! ```
+//!
+//! # Configuration
+//!
+//! Pass via `storage_options` when opening a dataset:
+//!
+//! ```python
+//! ds = lance.dataset(
+//! "s3://bucket/data.lance",
+//! storage_options={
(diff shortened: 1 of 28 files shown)


charleshuang119:
Really nice work! I like the overall design a lot, the Velox-style two-tier (memory + SSD) async cache makes a lot of sense, and the integration into the IO path looks clean. This should be impactful for reducing object store round-trips.

Had a few thoughts while reading through:

1. Since the cache is scoped per dataset/session, I think multiple opens of the same dataset may end up caching the same underlying file ranges independently. That seems like a reasonable simplification for v1, but do we eventually want a process-global/shared cache keyed by "dataset identity + version + file range", rather than only per-session state?

2. One semantic question: without cache, I’d expect reads against files removed by cleanup_older_than to fail at the object store layer. With cache, an already-open reader may still be able to serve those bytes successfully. Is that the intended behavior (snapshot-style semantics for already-open readers), or would we want pruned files to become unreadable even if cached?

jaystarshot:
Thanks 
1. Could be fixed in theory by passing around sessions to register the cache everywhere too but yes a global state could be better just need to think of the API

2. So this is easier if we limit the cache per dataset() to start with.  once the dataset() call is made the version is pinned don't think that version would change. Even if a new version removes a file (via delete or compaction), the current ds won't query that file because its manifest doesn't reference it. The physical plan only touches files listed in the pinned manifest. If we make changes in current DS call then they will reflect in the querying 

But i agree for a global cache we might need to think a bit more and test more cases.

jaystarshot:
Hi @jackye1995 @westonpace @wjones127, I've put together a design [doc](https://docs.google.com/document/d/1k3UE8cnTSHoLJ43JGi-S4fJ0J3CLGe8LbRTjqH8puAU/edit?tab=t.0) for review and will walk through it on Thursday. 
The implementation is also ready for code review, I'm still working on making it easier to review. Thanks for all the help and guidance!

rust/lance/src/session.rs
```diff
@@ -124,9 +136,42 @@ impl Session {
             metadata_cache: GlobalMetadataCache(LanceCache::with_capacity(metadata_cache_size)),
             index_extensions: HashMap::new(),
             store_registry,
+            data_cache: None,
+            data_cache_verify: false,
```
westonpace: This feels like it should be a config in the `data_cache` itself, not something that lives alongside `data_cache`.  In other words, can this be part of `DataCacheConfig`?

rust/lance/src/io/exec/scan.rs
```diff
@@ -268,10 +268,8 @@ impl LanceStream {
             file_fragments = filtered_fragments;
         }
 
-        let scan_scheduler = ScanScheduler::new(
-            dataset.object_store.clone(),
-            SchedulerConfig::new(config.io_buffer_size),
-        );
+        let scheduler_config = SchedulerConfig::new(config.io_buffer_size);
```
westonpace: Is this just a formatting change?

rust/lance-io/src/object_reader.rs
```diff
@@ -258,6 +260,145 @@ impl Reader for CloudObjectReader {
...
+        return Err(object_store::Error::Generic {
+            store: "data_cache",
+            source: msg.into(),
+        });
+    }
+    tracing::debug!(offset, length, "cache checksum OK");
+    Ok(cached)
+}
+
+impl Reader for CachedObjectReader {
```
westonpace, line 346: I'm weighing the pros/cons in my head of implementing this as `Reader` instead of `ObjectStore`.

One potential advantage would be that we can still use the cache when we aren't using an object store reader but the only case for that currently is the local file reader and we probably wouldn't need the cache in that case.

One disadvantage would be that a user might someday want a write-through cache (e.g. populate cache on writes) and there would be no way to do that if we just target `Reader`.

rust/lance-io/src/data_cache/mod.rs
```diff
@@ -0,0 +1,966 @@
...
+    pub const KEY_ENABLED: &'static str = "data_cache_enabled";
+    pub const KEY_MEMORY_BYTES: &'static str = "data_cache_memory_bytes";
+    pub const KEY_SSD_ENABLED: &'static str = "data_cache_ssd_enabled";
+    pub const KEY_SSD_DIR: &'static str = "data_cache_ssd_dir";
+    pub const KEY_SSD_BYTES: &'static str = "data_cache_ssd_bytes";
+
+    pub const KEY_MEMORY_SHARDS: &'static str = "data_cache_memory_shards";
+    pub const KEY_SSD_SHARDS: &'static str = "data_cache_ssd_shards";
+    pub const KEY_VERIFY: &'static str = "data_cache_check_rtt_enabled";
+    pub const KEY_SSD_CRC32: &'static str = "data_cache_ssd_crc32_enabled";
```
westonpace, line 61: We will want to document these in the user guide.  In fact, we probably want a pretty extensive description of the mechanism.

rust/lance-io/src/data_cache/mod.rs
```diff
@@ -0,0 +1,966 @@
...
+
+    pub ssd_num_shards: usize,
+
+    pub verify: bool,
+
+    pub ssd_crc32_enabled: bool,
+}
+
+impl DataCacheConfig {
+    pub const KEY_ENABLED: &'static str = "data_cache_enabled";
```
westonpace, line 52: Why not automatically enable if `data_cache_memory_bytes` or `data_cache_ssd_bytes` is greater than 0?  Also, I see `data_cache_ssd_enabled`.  is this property controlling just the memory cache?  Or do you need both turned on to get SSD cache enabled?

rust/lance-io/src/data_cache/mod.rs
```diff
@@ -0,0 +1,966 @@
...
+    pub const KEY_MEMORY_BYTES: &'static str = "data_cache_memory_bytes";
+    pub const KEY_SSD_ENABLED: &'static str = "data_cache_ssd_enabled";
+    pub const KEY_SSD_DIR: &'static str = "data_cache_ssd_dir";
+    pub const KEY_SSD_BYTES: &'static str = "data_cache_ssd_bytes";
+
+    pub const KEY_MEMORY_SHARDS: &'static str = "data_cache_memory_shards";
+    pub const KEY_SSD_SHARDS: &'static str = "data_cache_ssd_shards";
```
westonpace, line 59: When documenting these please provide some guidance to users on how/why they should be adjusted.

westonpace (commented):
I suppose my primary concern remains that this is adding more complexity, that we will need to maintain.  At the same time, we could get the same benefits by intercepting purely at the object_store level, and then it could be developed entirely independently of lance.

Still, I'm happy to hear other opinions, and I don't want to stand in the way of getting stuff done.  It's a fairly decoupled feature and so I don't think it has widespread impacts.

Furthermore, the primary user interface appears to be storage options, which I like.  If we decide to move to something else (like an object store wrapper) then all we need to do is wire up the storage options correctly.

I have a few thoughts to start with.  Then I'll take a closer look at the cache impl itself.

jaystarshot:
Thanks a lot @westonpace for your time

rust/lance-io/src/data_cache/mod.rs
jaystarshot: Currently, the SSD cache is kind of a backup to memory cache which only handles evictions. It can be used standalone but its not designed for that heavy insert and read.

rust/lance-io/src/object_reader.rs
jaystarshot: Which class specifically? For write through cache probably we may need to tailor the design a bit more

rust/lance/src/io/exec/scan.rs
jaystarshot: Yes left over from previous where i was trying to add the cache to the scheduler :p

perone:
Thank you for the PR ! This is indeed a much-needed feature. Cloud ML training instances often use local SSDs or NVMe disks to avoid the high latency of remote stores like S3, so providing a transparent caching mechanism removes this complexity from the application layer and simplifies a lot, nowadays I don't think there is a way to intercept which files are being accessed on S3 from the client API, so you end up being forced to add a layer between Lance and S3 or or on the application side on top of Lance, and all cases are much more complexity than just implementing it on Lance.

jaystarshot:
@perone Yes agreed, I think the future is integration of this with a remote cache so that applications can be dumb and just read rows and everything is handled by the "pseudo DB layer" internally

Gabriel39:
Hi Lance maintainers, @westonpace @jaystarshot 

We are developers from the Apache Doris community. Apache Doris plans to support reading Lance files in our next release, and a local disk/SSD data cache is important for several of our customer workloads, especially for repeated reads from remote object storage.

Could you share whether the Lance community plans to continue developing and upstreaming disk cache support? In particular, is this PR still the intended direction, or is the community leaning toward an object_store-level wrapper or another design?

Knowing the preferred direction would help us plan the Doris integration and avoid building against an approach that Lance does not intend to maintain. We would also be interested in collaborating if additional implementation, testing, or design work would help move this forward.

Thanks!

jaystarshot:
Hi @Gabriel39  sorry, this fell off my radar. I had tested the PR in production at that time and I believe its mostly code complete.

I got a feeling at that time that Lance doesn't want to maintain this so I think an object store approach makes more sense. I tried something similar in Talon https://github.com/milvus-io/talon/pull/453, which is aimed at being a general object storage cache — though a pluggable object store may be the better fit for some of these cases.

PR #6490, opened 2026-04-13
feat: add fragment range read to Java API

bryanck:
This PR exposes a fragment range read method in the Java API. It adds a new `readRange` method to the Java `Fragment` class that calls a new JNI native method. Currently the only way to read a range from a fragment is to use the dataset scanner, however the dataset scanner always calls `count_all_rows` on the dataset when bounds checking a range. That adds overhead to the operation.

An alternative approach would be to add an option to disable the bounds check when using dataset scanner, but I felt this solution reflects the native API more closely.

Here's a CPU profile using dataset scanner (first) and fragment range read (second)...
Screenshot 2026-04-12 at 6 40 00 PM
Screenshot 2026-04-12 at 6 40 15 PM

### DIFF
--- a/java/src/main/java/org/lance/FragmentMetadata.java
+++ b/java/src/main/java/org/lance/FragmentMetadata.java
@@ -26,7 +26,7 @@ import java.util.Objects;
 /** Metadata of a Fragment in the dataset. Matching to lance Fragment. */
 public class FragmentMetadata implements Serializable {
   private static final long serialVersionUID = -5886811251944130460L;
-  private final int id;
+  private final int id; // FIXME: change to long
   private final List<DataFile> files;
   private final long physicalRows;
   private final DeletionFile deletionFile;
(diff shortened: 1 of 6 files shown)


java/src/main/java/org/lance/FragmentMetadata.java
```diff
@@ -26,7 +26,7 @@
 /** Metadata of a Fragment in the dataset. Matching to lance Fragment. */
 public class FragmentMetadata implements Serializable {
   private static final long serialVersionUID = -5886811251944130460L;
-  private final int id;
+  private final int id; // FIXME: change to long
```
bryanck, line 29: Fragment ID should be a `long` but that is a fairly significant change so left that for a follow up.

bryanck: I opened this follow up to move fragment ID to long... https://github.com/lance-format/lance/pull/6498

PR #6495, opened 2026-04-13
fix(compaction): skip collect rowids for unindexed fragment or stable row ids

zhangyue19921010:
In compaction, collecting and transferring fragment rowids / row_addrs is expensive.

If the fragments in a compaction task do not belong to any index, then this data does not require immediate remapping

therefore does not require collecting the old rowids


rust/lance/src/dataset/optimize.rs
```diff
@@ -1117,6 +1122,32 @@ async fn reserve_fragment_ids(
...
+    options: &CompactionOptions,
+) -> Result<bool> {
+    if dataset.manifest.uses_stable_row_ids() {
+        return Ok(false);
+    }
+
+    let has_indexed_fragments = task.has_indexed_fragments.unwrap_or(true);
+    if has_indexed_fragments {
+        return Ok(true);
+    }
```
pratik0316, line 1137: maybe I am missing some context, just curious why do we always need to capture the row_addr column in case the fragments have an index?

zhangyue19921010: Indexed fragments need row addr capture because compaction changes the physical addresses that existing indexes point to(unstable row IDs).To keep the index valid, compaction needs a mapping() from old addresses to new addresses. The captured _rowaddr column gives the old addresses of the rows that survived the rewrite.

pratik0316: clear! thanks for the explanation @zhangyue19921010

zhangyue19921010:
Hi @Xuanwo, would you mind taking a look? I’d really appreciate your help. Thanks in advance!

zhangyue19921010:
Hi @wjones127 , would you mind taking a look? Thanks ：）

yanghua:
@claude review

rust/lance/src/dataset/optimize.rs
```diff
@@ -1118,6 +1123,32 @@ async fn reserve_fragment_ids(
...
+
+    if !options.defer_index_remap {
+        return Ok(false);
+    }
```
yanghua, line 1142: If the case is `defer_index_remap == false` and `task.has_indexed_fragments == fase`, here we actively return false, but the task comes from plan phase. The `task.has_indexed_fragments == fase` may be old or not freshness right? For example, someone creates a index before current mement but after planning. Shall we re-compute the has_indexed_fragments?

rust/lance/src/dataset/optimize.rs
```diff
@@ -1546,12 +1577,9 @@ pub async fn commit_compaction(
...
             }
-        } else if options.defer_index_remap {
-            let changed_row_addrs = task.row_addrs.ok_or_else(|| {
-                Error::internal(
-                    "defer_index_remap requires row_addrs but none were provided".to_string(),
-                )
-            })?;
+        } else if options.defer_index_remap
+            && let Some(changed_row_addrs) = task.row_addrs
+        {
```
yanghua, line 1582: Here, we may also meet the consistent issue like before. We could add an assertion in the end, like this:

```
else if options.defer_index_remap {
    debug_assert!(
        task.has_indexed_fragments != Some(true),
        "defer_index_remap task with indexed fragments must carry row_addrs"
    );
}
```

yanghua (commented):
Left some comments.

zhangyue19921010:
replaced by https://github.com/lance-format/lance/pull/7773

PR #6496, opened 2026-04-13
chore: dependency update Arrow to 58 and DataFusion to 53

### LINKED ISSUE #6137: Update to Arrow 58
Hey Lance maintainers!
I carefully searched through existing PRs and issues, but I didn't find anything related.
Is there any ongoing work to support [Arrow 58](https://crates.io/crates/arrow)?
For context: Last time, when updating to Arrow 57, a lot of work was done by @timsaucer – #5291.

If not, can we please update Lance to Arrow 58?

timsaucer:
This PR is a dependency update only for Arrow and DataFusion. It includes the minimal changes necessary for the upgrade and is not designed to have any additional impact.

Closes #6137


rust/lance/src/dataset/write/merge_insert.rs
```diff
@@ -754,6 +754,7 @@ impl MergeInsertJob {
                 None,
                 PartitionMode::CollectLeft,
                 NullEquality::NullEqualsNull,
+                false,
```
timsaucer, line 757: Need to investigate if this is a null-aware join.

rust/lance/src/dataset/scanner.rs
```diff
@@ -18,6 +20,7 @@ use datafusion::common::{DFSchema, JoinType, NullEquality, SchemaExt, exec_dataf
 use datafusion::functions_aggregate;
 use datafusion::logical_expr::{Expr, ScalarUDF, col, lit};
 use datafusion::physical_expr::PhysicalSortExpr;
+#[allow(deprecated)]
```
timsaucer, line 23: Replacing CoalesceBatchesExec should probably be a new PR.

rust/lance/src/dataset/scanner.rs
```diff
@@ -4207,6 +4211,7 @@ impl Scanner {
                     None,
                     PartitionMode::CollectLeft,
                     NullEquality::NullEqualsNull,
+                    false,
```
timsaucer, line 4214: Also check here if this is a null aware join

rust/lance/src/dataset/scanner.rs
```diff
@@ -4559,6 +4564,7 @@ impl Scanner {
     }
 
     /// Take row indices produced by input plan from the dataset (with projection)
+    #[allow(deprecated)]
```
timsaucer, line 4567:
```suggestion
    #[expect(deprecated)]
```

rust/lance/src/io/exec/knn.rs
```diff
@@ -1327,16 +1333,7 @@ impl ExecutionPlan for MultivectorScoringExec {
...
     }
 
-    fn statistics(&self) -> DataFusionResult<Statistics> {
-        Ok(Statistics {
-            num_rows: Precision::Inexact(
-                self.query.k * self.query.refine_factor.unwrap_or(1) as usize,
-            ),
-            ..Statistics::new_unknown(self.schema().as_ref())
-        })
-    }
```
timsaucer, line 1337: We probably need to implement partition statistics here, not just drop it.

rust/lance/src/io/exec/optimizer.rs
```diff
@@ -7,14 +7,14 @@ use std::sync::Arc;
 
 use super::TakeExec;
 use arrow_schema::Schema as ArrowSchema;
+#[allow(deprecated)]
```
timsaucer, line 10:
```suggestion
#[expect(deprecated)]
```

rust/lance/src/io/exec/optimizer.rs
```diff
@@ -70,6 +70,7 @@ impl CoalesceTake {
 }
 
 impl PhysicalOptimizerRule for CoalesceTake {
+    #[allow(deprecated)]
```
timsaucer, line 73:
```suggestion
    #[expect(deprecated)]
```

Cargo.lock
```diff
@@ -2133,7 +2133,7 @@ dependencies = [
  "futures",
  "itertools 0.14.0",
  "log",
- "object_store",
```
timsaucer, line 2041: We cannot get past having 2 versions of object_store until opendal 0.56 is released

timsaucer (commented):
Ok, there's a few things we need to take care of but otherwise looks in decent shape.

timsaucer:
It looks like opendal should be releasing soon so maybe we end up holding for a little: https://github.com/apache/opendal/issues/7283

timsaucer:
Ok, we're in the green! I've identified a couple of things to follow up on / check but otherwise looking good IMO. @wjones127 / @westonpace do you have thoughts about merging or waiting for the opendal update before we move forward?

gotocoding-DB:
Hey @timsaucer (sorry for so many pings, Tim)! Do we have any unpublished news here?
Or do we still wait review from @wjones127 and @westonpace?

rust/lance/src/dataset/write/merge_insert.rs
westonpace: null-aware only matters for very specific anti-join patterns and we never configure an anti-join today.

rust/lance/src/dataset/scanner.rs
westonpace: Ditto, not a concern

rust/lance/src/dataset/scanner.rs
westonpace: Yeah, let's use `expect`.

rust/lance/src/io/exec/knn.rs
westonpace: Did `statistics` get fully removed now?  I know it had been deprecated for a bit.  Losing this is probably fine for now if we make sure to add a follow-up.

westonpace (approved):
Thanks for this work Tim.  I'll see if I can get some opinion on OpenDAL.  Is that the only reason this is in draft?  I'd like to get this in as some of our users are encountering https://github.com/apache/datafusion/commit/a6a4df99963e0a2c0c49f1c17d23c9c4e5b59c68

westonpace:
Ok, so after more review I'm fairly certain that we don't ever use object_store indirectly via Datafusion, so from a pure lance perspective I see no problem with having multiple versions of object_store.

I think the biggest headache would be for users that are using Datafusion themselves and using Lance via Rust (that might be you Tim :laughing:).

In this case if you were adding your object_store implementation to Datafusion's object store registry and then also trying to add it to Lance's object store registry you would run into some headaches.

However, if you're just using object_store indirectly (e.g. you use Parquet scanners via Datafusion and you give Lance object store URIs when you connect) you'd probably be ok.

westonpace:
Wanting to get this in soon so I've rebased and submitted as https://github.com/lance-format/lance/pull/6638

Cargo.lock
Xuanwo: opendal has been released, let's go!

PR #6503, opened 2026-04-13
feat: add ANN proto codecs and extract table_identifier module

LuQQiu:
Add protobuf encode/decode for `ANNIvfSubIndexExec`


protos/ann.proto
```diff
@@ -0,0 +1,34 @@
...
+  // Absent means None (use the index's default metric).
+  optional string metric_type = 10;
+  bool use_index = 11;
+  float dist_q_c = 12;
+}
+
+message ANNIvfSubIndexExecProto {
+  VectorQueryProto query = 1;
+  TableIdentifier table = 2;
+  repeated lance.table.IndexMetadata indices = 3;
```
BubbleCal, line 40: do we need to serialize `prefilter_source` for this?

protos/ann.proto
```diff
@@ -0,0 +1,34 @@
...
+  optional uint32 ef = 8;
+  optional uint32 refine_factor = 9;
+  // Distance metric type as string ("l2", "cosine", "dot", "hamming").
+  // Absent means None (use the index's default metric).
+  optional string metric_type = 10;
+  bool use_index = 11;
+  float dist_q_c = 12;
+}
+
+message ANNIvfSubIndexExecProto {
```
westonpace, line 37: Can we just add a brief comment that the purpose of this message is to serialize the `ANNIvfSubIndexExec` node? (Though that might be obvious from the name)

protos/ann.proto
```diff
@@ -0,0 +1,34 @@
...
+
+package lance.datafusion;
+
+import "table_identifier.proto";
+import "table.proto";
+
+// Serialized vector query parameters.
+message VectorQueryProto {
+  // Query vector as Arrow IPC bytes (supports Float16, Float32, Float64, UInt8, etc.)
+  bytes key_arrow_ipc = 1;
```
westonpace: Can we do `query_arrow_ipc` or `query_vector_arrow_ipc`?  We don't use the term `key` anywhere else that I know of.

protos/ann.proto
```diff
@@ -0,0 +1,34 @@
...
+  uint32 k = 3;
+  optional float lower_bound = 4;
+  optional float upper_bound = 5;
+  uint32 minimum_nprobes = 6;
+  optional uint32 maximum_nprobes = 7;
+  optional uint32 ef = 8;
+  optional uint32 refine_factor = 9;
+  // Distance metric type as string ("l2", "cosine", "dot", "hamming").
+  // Absent means None (use the index's default metric).
+  optional string metric_type = 10;
```
westonpace: Can we use https://github.com/lance-format/lance/blob/d37348d7b9249ffab3c2e0cad55ff4bce16d9a71/protos/index.proto#L152

protos/ann.proto
```diff
@@ -0,0 +1,34 @@
...
+  bool use_index = 11;
+  float dist_q_c = 12;
```
westonpace: Should this be optional?

protos/ann.proto
```diff
@@ -0,0 +1,34 @@
...
+
+// Serialized vector query parameters.
+message VectorQueryProto {
+  // Query vector as Arrow IPC bytes (supports Float16, Float32, Float64, UInt8, etc.)
+  bytes key_arrow_ipc = 1;
+  string column = 2;
+  uint32 k = 3;
+  optional float lower_bound = 4;
+  optional float upper_bound = 5;
+  uint32 minimum_nprobes = 6;
```
westonpace: Should this be optional?

westonpace (approved):
A few nits but approve pending @BubbleCal 's observation re: prefilter source

Is the eventual goal to use `PhysicalExtensionCodec`?

westonpace:
The `Cargo.lock` for python needs updated.

protos/ann.proto
LuQQiu: prefilter source is 
```
  Why this works: PreFilterSource is one of three variants:
  - FilteredRowIds(Arc<dyn ExecutionPlan>) — an execution
  plan child
  - ScalarIndexQuery(Arc<dyn ExecutionPlan>) — an execution
  plan child
  - None — no prefilter
```
When ANNIvfSubIndexExec has a prefilter, it reports the prefilter plan as a child in children() .
DataFusion's proto serialization framework automatically serializes/deserializes all children. So the codec receives the decoded children in the inputs slice and reconstructs the PreFilterSource from them. It doesn'  need to be in the proto. @westonpace FYI

protos/ann.proto
LuQQiu: No problem, added

protos/ann.proto
LuQQiu: change to optinal

protos/ann.proto
LuQQiu: change to optional

protos/ann.proto
LuQQiu: Change to use that
but because lance-index depends on lance-datafusion
protos/ann.proto cannot belongs to lance-datafusion
i have to move that to lance crate

protos/ann.proto
LuQQiu: changed to query_vector_arrow_ipc

protos/ann.proto
LuQQiu: Double checked float dist_q_c is not optional in Query struct, callers send 0.0 explicitly

LuQQiu:
@westonpace @BubbleCal resolve the comments, please take a look

BubbleCal (approved):

westonpace (approved):

PR #6506, opened 2026-04-14
perf: add SIMD-accelerated u8 dot product for SQ distance

justinrmiller:
## Summary

- Add AVX-512 VNNI and AVX2 backends for unsigned int8 dot product with runtime CPU feature detection and automatic fallback to scalar
- Replace the unoptimized `Dot<u8>` impl (which had an explicit TODO) with dispatched SIMD kernel
- All existing callers including SQ distance computation benefit automatically with zero changes to lance-index

### New file: `rust/lance-linalg/src/distance/dot_u8.rs`

Three backends selected at runtime via `OnceLock` + `is_x86_feature_detected!`:

| Backend | Instruction | Elements/iter | CPU |
|---|---|---|---|
| AVX-512 VNNI | `VPDPBUSD` + XOR-0x80 bias trick | 64 | Ice Lake+ / Zen 4+ |
| AVX2 | `VPMADDWD` on zero-extended u16 | 32 | Haswell+ / Zen 1+ |
| Scalar | portable reference | - | any (including ARM) |

### The VNNI bias trick

`VPDPBUSD` expects one unsigned and one signed operand, but SQ vectors are u8×u8. We XOR one operand with 0x80 to map it to the signed domain, then correct by adding `128·Σa` at the end. The correction uses `VPSADBW` which runs on execution port 5 while `VPDPBUSD` runs on port 0 — they execute in parallel every cycle, making the correction effectively free.

### SQ integration (automatic)

`SQDistCalculator::distance()` already calls `dot_distance()` → `u8::dot()` for Dot distance type. Replacing the `Dot<u8>` body is the only change needed.

### Ryzen 4500 (AVX2, no VNNI)

1M total u8 elements, varying vector dimension. Scalar baseline vs AVX2-dispatched path:

| Dimension | Scalar | Dispatch (AVX2) | Speedup |
|-----------|--------|-----------------|---------|
| 128 | 51.02 µs | 58.25 µs | 0.88x (dispatch overhead dominates) |
| 256 | 44.96 µs | 38.62 µs | **1.16x** |
| 512 | 42.82 µs | 28.27 µs | **1.51x** |
| 1024 | 41.00 µs | 25.17 µs | **1.63x** |

AVX2 delivers up to 1.63x throughput at dim=1024. At dim=128 the `OnceLock` dispatch and AVX2 loop setup overhead exceeds the SIMD gains on short vectors. AVX-512 VNNI (Ice Lake+ / Zen 4+) is expected to show larger gains with 64 elements/iter.

### Apple M4 (ARM64, scalar fallback)

On ARM64 the dispatch falls back to scalar, so both paths perform identically (~13 µs at dim=1024). A follow-up ARM NEON `UDOT` path would bring SIMD gains to Apple Silicon.

### Out of scope (follow-up)
- L2/Cosine u8 SIMD optimization (different kernel: `Σ(a-b)²`)
- Native `VPDPBUUD` (unsigned×unsigned, Sierra Forest+) — too new for stable Rust
- ARM NEON `UDOT` path
- Precomputed norms for SQ L2/Cosine (requires storage format change)

## Test plan

- [x] Unit tests: random inputs across 18 vector sizes (0-4097), boundary values (all 0s, all 255s, alternating), one-sided zeros, all-ones patterns
- [x] Each backend tested independently against scalar reference (with `#[cfg]` guards for missing CPU features)
- [x] Existing `dot` tests continue to pass (9/9)
- [x] `cargo clippy -p lance-linalg --tests --benches -- -D warnings` clean
- [x] Benchmark on x86_64 with AVX2: `cargo bench --bench dot -p lance-linalg -- "Dot\(u8"`

### DIFF
--- a/rust/lance-linalg/benches/dot.rs
+++ b/rust/lance-linalg/benches/dot.rs
@@ -107,23 +107,39 @@ fn bench_distance(c: &mut Criterion) {
         });
     });
 
-    // u8 dot product benchmark (SQ distance path)
+    // u8 dot product benchmarks: scalar baseline vs SIMD dispatch
     {
-        const U8_DIMENSION: usize = 1024;
-        const U8_TOTAL: usize = 1024 * 1024;
-        let mut rng = rand::rng();
-        let key_u8: Vec<u8> = (0..U8_DIMENSION).map(|_| rng.random()).collect();
-        let target_u8: Vec<u8> = (0..U8_TOTAL * U8_DIMENSION).map(|_| rng.random()).collect();
-        c.bench_function("Dot(u8, dispatch)", |b| {
-            b.iter(|| {
-                black_box(
-                    target_u8
-                        .chunks(U8_DIMENSION)
-                        .map(|y| dot_distance(key_u8.as_slice(), y))
-                        .collect::<Vec<_>>(),
-                )
+        use lance_linalg::distance::dot_u8::{dot_u8, dot_u8_scalar};
+
+        for &dim in &[128, 256, 512, 1024] {
+            let num_vectors = 1024 * 1024 / dim; // ~1M elements total
+            let mut rng = rand::rng();
+            let key_u8: Vec<u8> = (0..dim).map(|_| rng.random()).collect();
+            let target_u8: Vec<u8> =
+                (0..num_vectors * dim).map(|_| rng.random()).collect();
+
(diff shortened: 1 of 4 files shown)


BubbleCal (approved):
Nice work!

PR #6507, opened 2026-04-14
fix: add dir_listing_to_manifest_migration_enabled flag to avoid extra object store calls

jackye1995:
## Summary

- Adds `dir_listing_to_manifest_migration_enabled` flag (default: `false`) to `DirectoryNamespaceBuilder` and `DirectoryNamespace`
- When `false` and both `manifest_enabled` and `dir_listing_enabled` are `true`, root-level table operations (`table_exists`, `describe_table`, `list_tables`) skip the manifest check and use directory listing directly, avoiding extra object store listing calls
- When `true`, preserves the existing hybrid behavior of checking manifest first then falling back to directory listing
- Includes a test with a counting object store wrapper verifying only a single `list_with_delimiter` call is made for root-level table operations without migration mode

### DIFF
--- a/python/python/tests/test_namespace_dir.py
+++ b/python/python/tests/test_namespace_dir.py
@@ -815,6 +815,7 @@ class TestLanceNamespaceConnect:
             "root": f"memory://test_connect_{unique_id}",
             "manifest_enabled": "true",
             "dir_listing_enabled": "true",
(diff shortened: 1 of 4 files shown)


LuQQiu:
The configuration is complicated .... 
espeically `dir_listing_to_manifest_migration_enabled`

LuQQiu (approved):

PR #6508, opened 2026-04-14
feat: add configurable blob v2 pack file size

hamersaw:
## Summary
- Add `blob_max_pack_file_bytes` to `WriteParams`, allowing users to override the default 1 GiB maximum pack (`.blob`) sidecar file size
- Thread the configuration through the full write path: `WriteParams` -> `WriterGenerator` -> `WriterOptions` -> `BlobPreprocessor` -> `PackWriter`
- Expose the option in Python (`write_dataset`) and Java (`WriteParams.Builder`) bindings

## Test plan
- [x] All 37 existing blob tests pass (`cargo test -p lance blob`)
- [x] Clippy clean on `lance` and `lance-jni` crates
- [x] Verify Python binding works end-to-end with `blob_max_pack_file_bytes` kwarg
- [x] Verify Java binding compiles with `./mvnw compile`

### DIFF
--- a/java/lance-jni/src/blocking_dataset.rs
+++ b/java/lance-jni/src/blocking_dataset.rs
@@ -632,6 +632,7 @@ fn create_dataset<'local>(
         &initial_bases,
         &target_bases,
         &JObject::null(), // allow_external_blob_outside_bases not used for Dataset.write()
+        &JObject::null(), // blob_max_pack_file_bytes not used for Dataset.write()
     )?;
 
     // Set up namespace commit handler and storage options provider if namespace is provided
--- a/java/lance-jni/src/blocking_dataset.rs
+++ b/java/lance-jni/src/blocking_dataset.rs
@@ -413,6 +413,8 @@ pub extern "system" fn Java_org_lance_Dataset_createWithFfiSchema<'local>(
     storage_options_obj: JObject,      // Map<String, String>
     initial_bases: JObject,
     target_bases: JObject,
+    allow_external_blob_outside_bases: JObject, // Optional<Boolean>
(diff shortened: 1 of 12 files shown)


python/python/lance/dataset.py
```diff
@@ -6516,6 +6199,7 @@ def write_dataset(
         "target_bases": target_bases,
         "external_blob_mode": external_blob_mode,
         "allow_external_blob_outside_bases": allow_external_blob_outside_bases,
+        "blob_max_pack_file_bytes": blob_max_pack_file_bytes,
```
Xuanwo: How about using options like `blob_pack_file_size_threshold`? Align better with existing naming.

java/lance-jni/src/blocking_dataset.rs
```diff
@@ -632,6 +632,7 @@ fn create_dataset<'local>(
         &initial_bases,
         &target_bases,
         &JObject::null(), // allow_external_blob_outside_bases not used for Dataset.write()
+        &JObject::null(), // blob_max_pack_file_bytes not used for Dataset.write()
```
Xuanwo: So we never pass it?

Xuanwo (commented):

python/python/lance/dataset.py
hamersaw: No problem, I updated it!

java/lance-jni/src/blocking_dataset.rs
hamersaw: Looks like we were not passing the allow_external_blob_outside_bases option either. So I updated this to plumb both of them through so this is available using `Dataset.write` and not just through the fragment creation APIs in Java.

Xuanwo (approved):

PR #6509, opened 2026-04-14
feat: per-index stable row ID tracking for compaction

hamersaw:
## Summary
- Adds `optional bool stable_row_ids` field (proto field 11) to `IndexMetadata` so each index declares whether it references rows by stable row ID or physical row address
- Compaction now checks per-index instead of per-dataset, only remapping indexes that use row addresses — stable-row-id indexes skip the expensive remap entirely
- `CreateIndexBuilder::use_stable_row_ids()` lets users control the setting; defaults to stable when the dataset supports it
- `transaction.rs` handles the mixed state where both remapped (address-based) and non-remapped (stable) indexes coexist
- Adds `Dataset::take_by_addresses()` public API for taking rows by physical address

## Test plan
- [ ] Verify proto roundtrip with `stable_row_ids` field (lance-table codec test)
- [ ] Test index creation defaults (stable when dataset supports it, address otherwise)
- [ ] Test explicit `use_stable_row_ids(true)` on non-stable dataset returns error
- [ ] Test compaction with stable-row-id index skips remap but updates fragment bitmap
- [ ] Test compaction with address-based index remaps as before
- [ ] Test mixed index compaction (both types on same dataset)
- [ ] Test backward compat: legacy indexes (`stable_row_ids: None`) treated as address-based
- [ ] Test `take_by_addresses` returns correct rows

### DIFF
--- a/java/lance-jni/Cargo.lock
+++ b/java/lance-jni/Cargo.lock
@@ -3469,6 +3469,7 @@ dependencies = [
  "arrow-buffer",
  "arrow-cast",
  "arrow-data",
+ "arrow-ipc",
  "arrow-ord",
  "arrow-schema",
  "arrow-select",
--- a/java/lance-jni/src/index.rs
+++ b/java/lance-jni/src/index.rs
@@ -128,10 +128,17 @@ impl IntoJava for &IndexMetadata {
         // Determine index type from index_details type_url
         let index_type = determine_index_type(env, &self.index_details)?;
 
+        // Convert stable_row_ids from Option<bool> to Boolean for Java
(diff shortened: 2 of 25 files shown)


hamersaw:
Closing this, but had Claude burn about 400k tokens and can't figure it out. The difficulty lies in having a dataset with stable row IDs enabled but using an index that works over row addresses directly. I think there's a lot of power in this approach, but making sure this works with the query path is difficult -- for example, doing a query over an index with non-stable row IDs and correctly grabbing the data from the dataset that has stable rows enabled. Intuitively it doesn't seem that difficult, but I don't have time to dive deep here.

PR #6510, opened 2026-04-14
perf: add SIMD kernels for bf16 distance functions

justinrmiller:
## Summary

- Replaces the external `numkong` dependency with in-tree C kernels for **bf16 distance computation** (dot product, L2, cosine, norm_l2)
- Follows the existing f16 kernel pattern: C source compiled via `build.rs` with per-architecture flags, runtime CPU dispatch via `SIMD_SUPPORT`
- Kernels are only enabled when the CPU supports the required instructions (NEON on aarch64, AVX2/AVX-512 on x86_64, LSX/LASX on loongarch64), with scalar fallback otherwise
- Gated behind the existing `fp16kernels` feature flag

## Benchmark Results

Tested on two platforms with 1M x 1024-dim vectors:

### Apple Silicon (M-series, NEON)

| Benchmark | Before (scalar) | After (C kernel) | Change |
|-----------|-----------------|-------------------|--------|
| **Dot(bf16)** | 144 ms | 55 ms | **2.6x faster** |
| **NormL2(bf16)** | 90 ms | 36 ms | **2.5x faster** |

### AMD Ryzen 5 4500 (Zen 2, AVX2)

| Benchmark | Before (scalar) | After (C kernel) | Change |
|-----------|-----------------|-------------------|--------|
| **Dot(bf16)** | 578 ms | 363 ms | **1.6x faster** (−37%) |
| **NormL2(bf16)** | 365 ms | 207 ms | **1.8x faster** (−43%) |

### Why the approach works

BF16-to-f32 conversion is a simple left-shift by 16 bits. The C kernels compiled with architecture-specific flags (`-march=haswell`, `-mtune=apple-m1`, etc.) plus `-ffast-math` and vectorization pragmas give the compiler more freedom to emit tight SIMD code than LLVM gets from the Rust scalar loops. ARM benefits more because the baseline Rust auto-vectorization was weaker there.

## Files Changed

- **New**: `rust/lance-linalg/src/simd/bf16.c` — C kernels for dot, L2, cosine, norm_l2
- `rust/lance-linalg/build.rs` — compile bf16.c for each architecture
- `rust/lance-linalg/src/distance/{dot,l2,cosine,norm_l2}.rs` — runtime SIMD dispatch for bf16
- `rust/lance-linalg/Cargo.toml` — removed `numkong` dependency and feature
- `rust/lance-linalg/benches/{dot,l2,cosine}.rs` — removed numkong benchmark sections
- **Deleted**: `scripts/bench_numkong.sh`

## Test plan

- [x] `cargo test -p lance-linalg --features fp16kernels` — all bf16 tests pass (kernel path)
- [x] `cargo test -p lance-linalg` — all bf16 tests pass (scalar fallback)
- [x] `cargo clippy -p lance-linalg --features fp16kernels --tests --benches -- -D warnings` — clean
- [x] Benchmarked on Apple Silicon (ARM NEON)
- [x] Benchmarked on AMD Ryzen 5 4500 (x86_64 AVX2)
- To reproduce:
  ```bash
  git checkout HEAD~1
  TARGET_TIME=3 cargo bench -p lance-linalg --features fp16kernels --bench dot -- --save-baseline before "bf16"
  TARGET_TIME=3 cargo bench -p lance-linalg --features fp16kernels --bench norm_l2 -- --save-baseline before "bf16"
  git checkout -
  TARGET_TIME=3 cargo bench -p lance-linalg --features fp16kernels --bench dot -- --baseline before "bf16"
  TARGET_TIME=3 cargo bench -p lance-linalg --features fp16kernels --bench norm_l2 -- --baseline before "bf16"
  ```


justinrmiller:
We may need to bump clang for this to work.

jackye1995:
Given this only improves bf16, why not just implement our own, learn from its implementation? So we can avoid taking a dependency just for this specific feature

justinrmiller:
> Given this only improves bf16, why not just implement our own, learn from its implementation? So we can avoid taking a dependency just for this specific feature 

I discussed that option with Yang. It's possible the performance regression is due to a technique being used that increases precision, and there's the fact we wouldn't have to maintain the kernels anymore.

westonpace:
> We may need to bump clang for this to work.

I think this would be my main concern.  If it's a new version of clang that can easily be installed on ubuntu then maybe, as an opt-in feature, it would be nice.  If it requires something like building clang from source then it will probably be a non-starter for most users.

The benefit of the current approach is that we don't have to worry about any portability concerns.  As long as that can continue I don't have much issue adopting numkong.  For starters though, I agree it should be an opt-in feature as in this PR.

justinrmiller:
@jackye1995 no more NumKong direct dependency. :)

BubbleCal (approved):
LGTM
maybe need a new PR title

westonpace (approved):
There are some merge conflicts.  If we resolve those I can merge.  Thanks for investigating this.  It's nice to see better support for bf16!

PR #6512, opened 2026-04-14
refactor!: vendor the tokenizer stack into lance

Xuanwo:
This PR vendors the tokenizer stack Lance actually uses into a new `rust/lance-tokenizer` crate and rewires FTS and inverted-index code to depend on it instead of `tantivy` and `lindera-tantivy`. It keeps the existing document and query tokenization semantics in-tree, renames the old FTS document adapter module to `document_tokenizer`, and preserves upstream license headers on vendored code.


BubbleCal (approved):
Awesome work!

PR #6513, opened 2026-04-14
perf: submit I/O requests eagerly in FullZipScheduler

### LINKED ISSUE #6504: FullZip scan latency regression on cloud storage (S3) due to lazy I/O submission
### Summary

PR #5981 (`e25f16909`) introduced `FullZipReadSource` and `create_page_load_task` as a unified abstraction for scheduling FullZip reads. While the PR itself brought significant performance improvements (especially the full-page scan shortcut and always-cached rep index), it inadvertently changed the I/O submission timing in two code paths: `schedule_ranges_simple` and the cached branch of `schedule_ranges_rep`. In both cases, `submit_request` was moved inside an `async move { ... }` block, which means the I/O is no longer submitted during the schedule phase — it is deferred until the decode phase polls the future.

### Affected Code Paths

| Path | Before #5981 | After #5981 | Expected |
|------|-------------|-------------|----------|
| `schedule_ranges_simple` (fixed-width, no rep index) | `submit_request` called **eagerly** during scheduling | `submit_request` inside `create_page_load_task` → deferred to decode | **Eager** (byte ranges are known at schedule time via simple arithmetic) |
| `schedule_ranges_rep` cached branch (rep index in memory) | All logic inside one `async` block (deferred) | `submit_request` inside `create_page_load_task` → still deferred | **Eager** (byte ranges computable from in-memory cached rep index, opportunity for optimization) |
| `schedule_ranges_rep` full-page scan | N/A (new path) | `submit_single` called eagerly ✅ | Correct |
| `schedule_ranges_rep` uncached (no rep index) | Deferred (two-stage I/O dependency) | Deferred ✅ | Correct (data byte ranges depend on first I/O result) |

### Impact

The scheduling architecture is designed as a two-thread pipeline: the scheduler thread issues I/O as fast as possible, and the decode stream consumes loaded pages. As described in `decoder.rs`:

> *Note that the scheduler thread does not need to wait for I/O to happen at any point. As soon as it starts it will start scheduling one page of I/O after another until it has scheduled the entire file's worth of I/O.*

When `submit_request` is deferred into the future, the I/O request is not enqueued into the `IoQueue` until the decode stream actually polls it. This eliminates the overlap between I/O and scheduling/decoding of other pages, effectively serializing I/O with decode and adding one full network RTT of latency per page for cloud storage (S3/GCS/Azure).

Note: this does **not** cause unbounded I/O pressure because `IoQueue` already enforces IOPS limits (`io_parallelism`: 64 for cloud, 8 for local) and byte-level backpressure (`io_buffer_size_bytes`). Early submission simply enqueues requests into the priority queue sooner, giving the I/O scheduler better visibility for prioritization.

hushengquan:
## Summary

Refactor `FullZipScheduler::create_page_load_task` to accept a pre-submitted I/O future instead of deferring I/O submission until the async task executes. This allows the I/O requests to be submitted immediately during scheduling, enabling the object store layer to batch and parallelize them. close #6504 

### Before: Lazy I/O submission (serialized)

Previously, `create_page_load_task` received a `FullZipReadSource::Remote(io)` along with byte ranges and priority. The actual `io.submit_request()` call happened **inside** the async block, meaning the I/O request was not submitted until the future was first polled.

When decoding multiple pages (e.g. across many fragments), this created a sequential I/O pattern:

```
Page 1: [schedule] -> [poll] -> [submit I/O] -> [wait response] -> [decode]
Page 2:                                          [schedule] -> [poll] -> [submit I/O] -> [wait response] -> [decode]
Page 3:                                                                                   [schedule] -> [poll] -> ...
```

Each page's I/O request could only be submitted after the previous task started executing. The I/O scheduler had no visibility into upcoming requests, preventing it from batching or parallelizing them effectively.

### After: Eager I/O submission (pipelined)

Now, `io.submit_request()` is called **before** constructing the `PageLoadTask`, and the resulting future is passed into `create_page_load_task`. All I/O requests for all pages are submitted upfront during the scheduling phase:

```
[schedule all pages] --> submit I/O page 1 -+
                     --> submit I/O page 2 -+
                     --> submit I/O page 3 -+  (all in-flight concurrently)
                     --> submit I/O page N -+
                                            |
                     [poll] -> [await page 1 response] -> [decode]
                     [poll] -> [await page 2 response] -> [decode]
                     [poll] -> [await page 3 response] -> [decode]
```

The object store layer can now see all pending requests at once and optimize I/O through batching, connection multiplexing, and parallel fetches. The async tasks only await the already-in-flight I/O futures.

## Changes

- `rust/lance-encoding/src/encodings/logical/primitive.rs`:
  - Changed `create_page_load_task` signature to accept `BoxFuture<'static, Result<Vec<Bytes>>>` instead of `FullZipReadSource` + byte ranges + priority
  - Moved `io.submit_request()` calls to happen eagerly at both call sites (`schedule_ranges_with_rep_index` and the non-rep-index path), before constructing the page load task

## Performance

Tested with a multi-fragment dataset containing fixed-width columns (768-dim float32 vectors, 40 fragments, 50 rows/fragment):

| Benchmark | Before (p50) | After (p50) | Speedup |
|---|---|---|---|
| Fixed-width column scan | 3453 ms | 523 ms | **6.6x** |

The improvement comes entirely from I/O pipelining — the decoding logic itself is unchanged. The effect is most pronounced with many fragments or pages, where the serialized I/O submission was the dominant bottleneck.

### DIFF
--- a/rust/lance-encoding/src/encodings/logical/primitive.rs
+++ b/rust/lance-encoding/src/encodings/logical/primitive.rs
@@ -2165,15 +2165,17 @@ impl FullZipScheduler {
     }
 
     fn create_page_load_task(
-        read_source: FullZipReadSource,
-        byte_ranges: Vec<Range<u64>>,
-        priority: u64,
+        io_future: BoxFuture<'static, Result<Vec<Bytes>>>,
         num_rows: u64,
         details: Arc<FullZipDecodeDetails>,
         bits_per_offset: u8,
     ) -> PageLoadTask {
         let load_task = async move {
-            let data = read_source.fetch(&byte_ranges, priority).await?;
+            let buffers = io_future.await?;
+            let data = buffers
+                .into_iter()
+                .map(|bytes| LanceBuffer::from_bytes(bytes, 1))
+                .collect::<VecDeque<_>>();
             Self::create_decoder(details, data, num_rows, bits_per_offset)
         }
         .boxed();
@@ -2333,14 +2335,9 @@ impl FullZipScheduler {
                 rep_index.bytes_per_value,
                 data_buf_position,
             );
-            let page_load_task = Self::create_page_load_task(
-                FullZipReadSource::Remote(io.clone()),
-                byte_ranges,
-                priority,
-                num_rows,
-                details,
-                bits_per_offset,
-            );
+            let io_future = io.submit_request(byte_ranges, priority);
+            let page_load_task =
+                Self::create_page_load_task(io_future, num_rows, details, bits_per_offset);
             return Ok(vec![page_load_task]);
         }
 
@@ -2403,10 +2400,9 @@ impl FullZipScheduler {
             })
             .collect::<Vec<_>>();
 
+        let io_future = io.submit_request(byte_ranges, self.priority);
         let page_load_task = Self::create_page_load_task(
-            FullZipReadSource::Remote(io.clone()),
-            byte_ranges,
-            self.priority,
+            io_future,
             num_rows,
             self.details.clone(),
             self.bits_per_offset,


Xuanwo (approved):
Nice change, thank you!

westonpace:
Belated second approval.  Thanks for catching this!  This is the third time this issue (wrapping something with async move closure only to have the execution semantics get changed) has bitten me in two months :laughing:

PR #6514, opened 2026-04-14
fix: block stale index entries after partial-schema merge_insert

### LINKED ISSUE #6283: Invalid results from FTS or vector search after a data replacement
A DataReplacement operation can modify a column that is indexed.  Today we will drop the modified fragment from the index's fragment bitmap.  However, this does not avoid invalid results because the fragment bitmap might cover other fragments included in the search.

For example:

1. Write fragments 1 & 2.
2. Create index covering fragments 1 & 2.
3. Update fragment 1 in-place with DataReplacement.  Fragment 1 is removed from the index's bitmap.
4. Search whole dataset

The search will target fragments 1 & 2.  It will search the index because it covers fragment 2.  It will search fragment 1 because it is unindexed.  The search results from the index search will include fragment 1 rows even though it should not.

westonpace:
## Summary

- A partial-schema `merge_insert` that modifies indexed columns drops the affected fragments from the index's `fragment_bitmap` but leaves stale entries in the index data (btree, vector, inverted). Subsequent queries that use the index find rows via both the stale entries AND the unindexed-fragment scan, producing duplicates or errors.
- Adds an `invalidated_fragment_bitmap` field to `IndexMetadata` (proto + Rust) that tracks fragments removed from the bitmap during `prune_updated_fields_from_indices`. At query time, all rows from these fragments are blocked by the deletion mask / prefilter so stale index entries are ignored. The bitmap is cleared when the index is rebuilt.
- Fixes three error categories seen in production with partial-schema merge_insert + btree index:
  - `Ambiguous merge inserts: multiple source rows match the same target row` (221 occurrences)
  - `fragment id N does not exist in the dataset` (114 occurrences)
  - `Attempt to merge two RecordBatch with different sizes: N != 0` (165 occurrences)
- Also fixes duplicate results in vector search and full text search after partial-schema merge_insert

## Test plan

- [x] 5 new regression tests covering all error categories:
  - `test_partial_merge_insert_stale_index_ambiguous` — second partial merge_insert finds same rows via stale btree + unindexed scan
  - `test_partial_merge_insert_stale_index_fragment_not_exist` — partial merge + update-all + partial merge hits deleted fragment
  - `test_partial_merge_insert_stale_index_batch_size_mismatch` — partial merge + partial update + partial merge hits deleted rows
  - `test_partial_merge_insert_stale_vector_index_duplicates` — KNN search returns duplicate rows after partial merge
  - `test_partial_merge_insert_stale_fts_index_duplicates` — FTS search returns duplicate rows after partial merge
- [x] All 131 existing merge_insert tests pass
- [x] Transaction, conflict resolver, scalar index, and lance-table format tests pass
- [x] Clippy clean

Closes https://github.com/lance-format/lance/issues/6283

### DIFF
--- a/rust/lance-table/src/format/index.rs
+++ b/rust/lance-table/src/format/index.rs
@@ -76,6 +76,16 @@ pub struct IndexMetadata {
     /// This is None if the file sizes are unknown. This happens for indices created
     /// before this field was added.
     pub files: Option<Vec<IndexFile>>,
+
+    /// Fragment IDs that were removed from `fragment_bitmap` because indexed
+    /// column values changed but the index data was not rewritten. The index
+    /// still contains stale entries for these fragments; readers must treat
+    /// all rows from them as deleted when querying the index.
+    ///
+    /// `None` means no fragments have been invalidated. Always cleared on a
+    /// full index rebuild. An optimize may or may not clear this depending
+    /// on the index type (see proto comment for details).
+    pub invalidated_fragment_bitmap: Option<RoaringBitmap>,
 }
 
 impl IndexMetadata {
@@ -132,6 +142,11 @@ impl DeepSizeOf for IndexMetadata {
                 .map(|fragment_bitmap| fragment_bitmap.serialized_size())
                 .unwrap_or(0)
             + self.files.deep_size_of_children(context)
+            + self
+                .invalidated_fragment_bitmap
+                .as_ref()
+                .map(|bitmap| bitmap.serialized_size())
+                .unwrap_or(0)
     }
 }
 
@@ -162,6 +177,14 @@ impl TryFrom<pb::IndexMetadata> for IndexMetadata {
             )
         };
 
+        let invalidated_fragment_bitmap = if proto.invalidated_fragment_bitmap.is_empty() {
+            None
+        } else {
+            Some(RoaringBitmap::deserialize_from(
+                &mut proto.invalidated_fragment_bitmap.as_slice(),
+            )?)
+        };
+
         Ok(Self {
             uuid: proto.uuid.as_ref().map(Uuid::try_from).ok_or_else(|| {
                 Error::invalid_input("uuid field does not exist in Index metadata".to_string())
@@ -178,20 +201,18 @@ impl TryFrom<pb::IndexMetadata> for IndexMetadata {
             }),
             base_id: proto.base_id,
             files,
+            invalidated_fragment_bitmap,
         })
     }
 }
 
-impl From<&IndexMetadata> for pb::IndexMetadata {
-    fn from(idx: &IndexMetadata) -> Self {
+impl TryFrom<&IndexMetadata> for pb::IndexMetadata {
+    type Error = Error;
+
+    fn try_from(idx: &IndexMetadata) -> Result<Self> {
         let mut fragment_bitmap = Vec::new();
-        if let Some(bitmap) = &idx.fragment_bitmap
-            && let Err(e) = bitmap.serialize_into(&mut fragment_bitmap)
-        {
-            // In theory, this should never error. But if we do, just
-            // recover gracefully.
-            log::error!("Failed to serialize fragment bitmap: {}", e);
-            fragment_bitmap.clear();
+        if let Some(bitmap) = &idx.fragment_bitmap {
+            bitmap.serialize_into(&mut fragment_bitmap)?;
         }
 
         let files = idx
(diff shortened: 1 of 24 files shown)


westonpace:
This feature has come up once or twice before.  Is there a reason this couldn't be implemented at the object_store level?  For example, I think you could enable something like https://github.com/foyer-rs/foyer with an `object_store->opendal->foyer` path?  Then the lance change would just be adding easier ways to enable it?  For example, a special URI or methods on the dataset open?

Or is this cache doing something more than pure data caching of the object_store layer?

rust/lance-table/src/format/index.rs
```diff
@@ -132,6 +142,11 @@ impl DeepSizeOf for IndexMetadata {
                 .map(|fragment_bitmap| fragment_bitmap.serialized_size())
                 .unwrap_or(0)
             + self.files.deep_size_of_children(context)
```
justinrmiller, line 144: is line 144 sort of a no-op at this point? The last line is all that will be returned right and it doesn't make use of line 144?

rust/lance/src/dataset/optimize/remapping.rs
```diff
@@ -286,6 +286,7 @@ async fn remap_index(dataset: &mut Dataset, index_id: &Uuid) -> Result<()> {
                     created_at: curr_index_meta.created_at,
                     base_id: None,
                     files: curr_index_meta.files.clone(),
+                    invalidated_fragment_bitmap: None,
```
justinrmiller: For Keep, should invalidated_fragment_bitmap be preserved?

rust/lance/src/dataset/write/merge_insert.rs
```diff
@@ -7658,4 +7658,517 @@ MergeInsert: on=[id], when_matched=UpdateAll, when_not_matched=InsertAll, when_n
...
+        .await
+        .unwrap();
+
+        Arc::new(ds)
+    }
+
+    /// Perform a partial-schema merge_insert (only id + value_a) targeting specific id ranges.
+    /// This causes touched fragments to drop from the index bitmap while btree data retains
+    /// stale entries.
+    async fn partial_merge_insert(
```
justinrmiller, line 7727: `invalidated_fragment_bitmap` grows unbounded until a full rebuild right? Could that ever become an issue?

justinrmiller (commented):
Just a couple of comments to consider.

rust/lance-table/src/format/index.rs
```diff
@@ -208,6 +232,14 @@ impl From<&IndexMetadata> for pb::IndexMetadata {
             })
             .unwrap_or_default();
 
+        let mut invalidated_fragment_bitmap = Vec::new();
+        if let Some(bitmap) = &idx.invalidated_fragment_bitmap
+            && let Err(e) = bitmap.serialize_into(&mut invalidated_fragment_bitmap)
+        {
+            log::error!("Failed to serialize invalidated fragment bitmap: {}", e);
+            invalidated_fragment_bitmap.clear();
```
LuQQiu: not propograte the error?

LuQQiu (approved):
Generally looks good, besides the key problems @justinrmiller pointed out

jackye1995 (approved):

rust/lance-table/src/format/index.rs
westonpace: These lines are all added together (there is a `+` at the start of line 145)

rust/lance-table/src/format/index.rs
westonpace: Ah, yeah, this is dangerous.  I guess it was copying the logic from above.  Still seems we should error in both cases.  We will just need to convert this into a TryFrom

rust/lance/src/dataset/optimize/remapping.rs
westonpace: Good point.  Actually, we can go ahead and keep the bitmap in both cases in compaction.  I've updated the code.  I went ahead and added a unit test as well.  If an fragment becomes invalidated we remove it from the index's fragment bitmap and this will prevent the fragment from compacting with other fragments so we should be safe on compaction.

rust/lance/src/dataset/write/merge_insert.rs
westonpace: It will never be larger than the number of fragments in the dataset which is reasonably bounded.  In general, an index with any invalidated fragments is going to perform slightly worse than one without because we have to do the masking.  That masking will be slightly more expensive when there are more fragments.

> until a full rebuild

It will be cleared on an index update (optimize indices), not just a full rebuild.

rust/lance-table/src/format/index.rs
westonpace: I have converted it to a TryFrom

justinrmiller (approved):

westonpace:
Merge pending vote on https://github.com/lance-format/lance/discussions/6529

PR #6520, opened 2026-04-15
docs: tighten python environment workflow guidance

Xuanwo:
This tightens the repository's environment guidance so language-specific tasks must follow the documented workflow before reporting missing tools or dependencies.

For Python work, the docs now make `uv sync --extra tests --extra dev` and `uv run ...` mandatory, and explicitly call out the common failure mode where slow `uv sync` is interrupted or global Python is used instead.


BubbleCal (approved):

Xuanwo:
AGENTS.md only changes, let's just merge.

PR #6522, opened 2026-04-15
fix: handle bloomfilter index overflow by multi-batch read&write

wojiaodoubao:
When create bloom filter index on a column with more than 1.07B rows, we might ran into arrow byte array offset overflow error. This pr fix it by writing index in batch.

### DIFF
--- a/rust/lance-index/src/scalar/bloomfilter.rs
+++ b/rust/lance-index/src/scalar/bloomfilter.rs
@@ -19,7 +19,7 @@ use crate::{Any, pb};
 use arrow_array::{Array, UInt64Array};
 mod as_bytes;
 pub mod sbbf;
-use arrow_schema::{DataType, Field};
+use arrow_schema::{DataType, Field, Schema};
 use serde::{Deserialize, Serialize};
 
 use std::sync::LazyLock;
@@ -44,6 +44,8 @@ use super::zoned::{ZoneBound, ZoneProcessor, ZoneTrainer, rebuild_zones, search_
 const BLOOMFILTER_FILENAME: &str = "bloomfilter.lance";
 const BLOOMFILTER_ITEM_META_KEY: &str = "bloomfilter_item";
 const BLOOMFILTER_PROBABILITY_META_KEY: &str = "bloomfilter_probability";
+const MAX_BLOOMFILTER_ARRAY_LENGTH: usize = i32::MAX as usize - 1024 * 1024;
+const BLOOMFILTER_READ_BATCH_SIZE: usize = 4 * 1024;
 const BLOOMFILTER_INDEX_VERSION: u32 = 0;
 
 #[derive(Debug, Clone)]
@@ -94,9 +96,6 @@ impl BloomFilterIndex {
         _index_cache: &LanceCache,
     ) -> Result<Arc<Self>> {
         let index_file = store.open_index_file(BLOOMFILTER_FILENAME).await?;
-        let bloom_data = index_file
-            .read_range(0..index_file.num_rows(), None)
-            .await?;
         let file_schema = index_file.schema();
 
         let number_of_items: u64 = file_schema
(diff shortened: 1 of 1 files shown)


wojiaodoubao:
Hi @jackye1995 , @BubbleCal , could you help review this when you have time, thanks very much~

Xuanwo:
@claude review

Xuanwo:
I think the write-side batching direction is right, but this PR still does not fix the overflow invariant end-to-end.

The root constraint here is not row count. It is that `bloom_filter_data` is stored as Arrow `Binary`, whose offsets are i32. Therefore no read or write path should ever construct a `BinaryArray` whose concatenated bloom filter payload can exceed the i32 offset limit.

This PR fixes that on the write side by flushing batches based on serialized bloom filter bytes. However, the load path is still bounded by a fixed row count:

```rust
for start in (0..index_file.num_rows()).step_by(BLOOMFILTER_READ_BATCH_SIZE) {
    let end = (start + BLOOMFILTER_READ_BATCH_SIZE).min(index_file.num_rows());
    let bloom_data = index_file.read_range(start..end, None).await?;
    zones.extend(Self::try_from_serialized(bloom_data)?);
}
```

That is not a sufficient bound. With non-default bloom filter params, each serialized zone can be much larger than the default ~16 KiB. For example, 4096 zones at ~512 KiB each already crosses 2 GiB, so `read_range` can still reconstruct a `BinaryArray` that overflows even though each written batch was valid.

I think the fix should be:

1. Keep the write-side byte-bounded batching.
2. Change the load path to consume record batches without concatenating multiple physical batches into one `BinaryArray`, or add a byte-bounded streaming read path for this index.
3. Make the invariant explicit in tests: construct multiple written batches where each batch is below a small test byte limit, but merging two batches would exceed that limit; then verify `BloomFilterIndex::load` does not read them back through a path that concatenates them.
4. Optionally follow up later by making `ZoneTrainer` stream zone statistics into the writer, so large BloomFilter indexes do not materialize all zones in memory before writing.

So the invariant should be:

> BloomFilter index serialization must be byte-bounded end-to-end; row count and logical index size are not valid bounds for Arrow `BinaryArray`.

Without the read-side change, this PR fixes the default 1.07B-row write failure but still allows validly-written BloomFilter indexes to fail on load.

wojiaodoubao:
Hi @Xuanwo , thanks your detailed review! I should have thought through more carefully,  I'll pay more attention next time.

I made a fix, now the read will use a safe batch size computed from number_of_items and probability.

Xuanwo:
Thanks, this addresses my main concern.

One remaining request: could we add an end-to-end test for `BloomFilterIndex::load`? The new tests cover `read_batch_size`, but not that load actually avoids constructing an oversized `BinaryArray`. Ideally the test would use a small injected byte limit, write multiple batches, and verify load succeeds in a case where concatenating them would exceed that limit.

Minor: `read_batch_size` can use `Sbbf::size_bytes()` instead of `to_bytes().len()` to avoid allocating just to compute the size.

rust/lance-index/src/scalar/bloomfilter.rs
```diff
@@ -617,13 +629,15 @@ impl BloomFilterIndexBuilder {
...
 
-        Ok(RecordBatch::try_new(schema, columns)?)
+        Ok(RecordBatch::try_new(Self::bloomfilter_schema(), columns)?)
     }
 
-    pub async fn write_index(self, index_store: &dyn IndexStore) -> Result<()> {
-        let record_batch = self.bloomfilter_stats_as_batch()?;
-
-        let mut file_schema = record_batch.schema().as_ref().clone();
+    pub async fn write_index(
```
Xuanwo, line 635: This changes a public builder method signature to expose an internal Arrow byte limit. Downstream callers of the public API will fail to compile for an implementation detail that should remain internal.

rust/lance-index/src/scalar/bloomfilter.rs
```diff
@@ -2126,4 +2246,122 @@ mod tests {
...
+        let data_stream = add_row_addr(data_stream);
+
+        let mut builder =
+            BloomFilterIndexBuilder::try_new(BloomFilterIndexBuilderParams::new(1, 0.25)).unwrap();
+        builder.train(data_stream).await.unwrap();
+        assert_eq!(builder.blocks.len(), row_count as usize);
+
+        builder.write_index(test_store.as_ref(), 64).await.unwrap();
+
+        let index = BloomFilterIndex::load(test_store.clone(), None, &LanceCache::no_cache())
```
Xuanwo, line 2318: This test only injects the small byte limit into the writer, while load still uses the production limit. It can pass even if the load path regresses to a single oversized read, so CI would not protect the overflow invariant.

zhangyue19921010:
talk with @wojiaodoubao , will take over this work through https://github.com/lance-format/lance/pull/7753

thanks for all your work @wojiaodoubao here!

Xuanwo:
Closing this because the work was explicitly handed over to #7753, which now owns the same BloomFilter BinaryArray i32 overflow read/write scope. #7753 is still open, so this is a replacement closure, not a claim that the fix has landed. Please reopen if work should continue on this branch instead.

PR #6531, opened 2026-04-15
feat: expose has_stable_row_ids property on LanceDataset

pengw0048:
Expose `LanceDataset.has_stable_row_ids` (bool) to Python, wrapping the existing Rust `Manifest::uses_stable_row_ids()`.

Currently the only way to check from Python is inspecting fragment metadata (`fragment.metadata.row_id_meta is not None`), which is less discoverable. This property reads directly from the manifest feature flags.

### DIFF
--- a/python/python/lance/dataset.py
+++ b/python/python/lance/dataset.py
@@ -1145,6 +1145,13 @@ class LanceDataset(pa.dataset.Dataset):
         return self._ds.data_storage_version
 
     @property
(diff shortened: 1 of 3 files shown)


dshepelev15 (approved):

jiaoew1991 (approved):

pengw0048:
Thanks for the review @jiaoew1991 . Shall we merge this? (I don't have push permission)

PR #6532, opened 2026-04-15
fix: replay unflushed WAL entries into memtable on restart

hamersaw:
## Summary
- On crash/restart, WAL entries written but not yet flushed to a generation were silently lost — the memtable started empty, dropping durably written data
- Adds `WalFlusher::replay_wal_entries()` that scans from `replay_after_wal_entry_position + 1` forward, re-inserting non-empty WAL entries into the memtable
- Called during `ShardWriter::open()` before accepting new writes, with `next_wal_entry_position` advanced past replayed entries

## Test plan
- [x] `test_replay_wal_entries` — writes 3 WAL entries, simulates restart, verifies all 3 replayed into memtable with correct row counts
- [x] `test_replay_wal_entries_skips_already_flushed` — verifies entries covered by `replay_after_wal_entry_position` are not replayed
- [x] `test_replay_wal_entries_nothing_to_replay` — verifies no-op when no WAL entries exist
- [x] All 217 existing `mem_wal` tests pass with no regressions


jackye1995:
I think this one is fixed in https://github.com/lance-format/lance/pull/6675 now

PR #6535, opened 2026-04-16
fix(java): use daemon thread for lance-jni-dispatcher

fangbo:


fangbo:
@beinan  In out production environment, the non-daemon lance-jni-dispatcher prevents jvm exists. Could you please take a look and see if there's any issue with setting the thread as a daemon?

fangbo:
Duplicated with https://github.com/lance-format/lance/pull/6549 . Close this.

PR #6536, opened 2026-04-16
fix(java): reuse session for OpenDatasetBuilder.buildFromNamespaceClient

fangbo:
@jackye1995 I think the session can be reused. Could you please take a look and see if this makes sense? Thank you.


fangbo:
Related  https://github.com/lance-format/lance-spark/pull/353  cc @LuciferYang

LuciferYang (approved):
nit: can we add a test that opens a dataset via namespace client with a `ReadOptions`-level session and verifies cache reuse would prevent this class of regression.

fangbo:
> nit: can we add a test that opens a dataset via namespace client with a `ReadOptions`-level session and verifies cache reuse would prevent this class of regression.

Thank you for advice. I have added a test case for this modification.

LuciferYang (approved):
+1， LGTM

PR #6541, opened 2026-04-16
perf: use SIMD argmin for L2Prepared nearest lookup

justinrmiller:
## Summary

- Adds SIMD-accelerated `argmin` for `L2Prepared::nearest_into()` using `f32x8` lane-wise minimum reduction
- This is the hot path for PQ codebook distance table lookups during vector search
- Adds L2Prepared benchmarks with PQ-typical workloads (256 centroids × 16 dims)

## Benchmark Results (Apple M-series, aarch64 NEON)

10K queries, 256 centroids × 16 dims (fits L1 cache):

| Benchmark | Before | After | Change |
|-----------|--------|-------|--------|
| L2Prepared(distances, 256×16) | 2.13 ms | 2.13 ms | ~same (no change) |
| **L2Prepared(nearest, 256×16)** | **4.71 ms** | **2.66 ms** | **-43.5%** |

The `distances_into` path is unchanged — LLVM's auto-vectorization via the `#[inline(never)]` boundary trick already produces good code. The win is entirely from replacing the scalar `argmin_value_float` iterator scan with a SIMD two-pass approach:

1. **SIMD min reduction**: Scan with `f32x8` to find the global minimum value
2. **Scalar index lookup**: Single scan to find the first index matching the minimum

This avoids tracking indices in SIMD lanes (which is complex and slow) while still benefiting from vectorized comparison for the min-finding pass.

## Test plan
- [x] All 18 L2 tests pass
- [x] `cargo fmt` clean


justinrmiller:
## Correctness verification

Added `test_argmin_f32_simd_matches_scalar` which verifies the SIMD argmin matches the original scalar `argmin_value_float` for all edge cases:

- Basic minimum finding
- Minimum in tail (past 8-element SIMD boundary)
- Exactly 8 elements
- Less than 8 elements
- Empty input
- Single element
- **Ties** (must return first occurrence — verified)
- PQ-typical 256 elements

**NaN note**: The SIMD `min` intrinsics (`_mm256_min_ps`, `vminq_f32`) have different NaN propagation than scalar `<` comparison. Specifically, `min(valid, NaN)` returns `NaN` on x86 while scalar `<` would skip NaN. This is documented in the code but is safe for `L2Prepared` because squared L2 distances are always non-negative finite values — they can never produce NaN.

PR #6542, opened 2026-04-16
perf: replace full sort with quickselect in kmeans_find_partitions

justinrmiller:
## Summary

- Replaces Arrow's `sort_to_indices` (full O(n log n) sort) with `select_nth_unstable_by` (O(n) average quickselect) + sort of just the k selected elements (O(k log k))
- Addresses the TODO at `kmeans.rs:1194` and `kmeans.rs:1219`: *"TODO: use heap to just keep nprobes smallest values"*
- Applies to both `kmeans_find_partitions` (L2/Dot) and `kmeans_find_partitions_binary` (Hamming)
- Removes the `arrow_ord::sort::sort_to_indices` import (no longer used in this file)

## How it works

**Before:** Compute all K distances → allocate `Float32Array` → `sort_to_indices` (full sort of all K values) → `arrow::compute::take` to extract top nprobes. Three separate allocations, O(K log K) sort.

**After:** Compute all K distances into `Vec<(f32, u32)>` → `select_nth_unstable_by` (quickselect partition) → `truncate` to k → sort just the k entries. One allocation, O(K + k log k) total.

I initially implemented this with a `BinaryHeap` (as the TODO suggested), but benchmarking showed quickselect is **6.3x faster** than the heap approach for n=65536, k=10 because the heap does a push+pop for every element (O(n log k)) while quickselect does a single O(n) partition pass.

## Microbenchmark: Selection Step Only

Isolated benchmark of just the top-k selection (n=65536, k=10, 10K iterations):

| Approach | Time per iteration | vs Full Sort |
|----------|-------------------|--------------|
| Full sort (sort_to_indices) | 462.8 µs | baseline |
| BinaryHeap top-k | 396.5 µs | 1.2x faster |
| **Quickselect + sort k (this PR)** | **63.2 µs** | **7.3x faster** |

## End-to-End IVF Benchmarks

Using the existing `find_partitions` benchmark (1536-dim vectors):

| Benchmark | Before | After | Change |
|-----------|--------|-------|--------|
| IVF10240,k=1,L2 | 975.9 µs | 988.9 µs | ~same |
| IVF10240,k=10,L2 | 973.7 µs | 982.8 µs | ~same |
| IVF10240,k=50,L2 | 974.4 µs | 980.9 µs | ~same |
| IVF65536,k=1,L2 | 6.345 ms | 6.362 ms | ~same |
| IVF65536,k=10,L2 | 6.417 ms | 6.590 ms | ~same |
| IVF65536,k=50,L2 | 6.365 ms | 6.398 ms | ~same |

The end-to-end benchmarks are within noise because 1536-dim L2 distance computation dominates (~6ms for 65K centroids) while selection is only ~60-460µs. The 7.3x improvement in the selection step saves ~400µs per query for K=65536, which becomes meaningful with:
- Lower-dimensional vectors (128-384 dims) where distance computation is faster
- Very large partition counts (K > 100K)
- High query throughput where microseconds compound

## Test plan
- [x] All 8 kmeans tests pass
- [x] `cargo fmt` clean
- [x] Return type contract preserved: `(UInt32Array, Float32Array)` sorted ascending by distance


justinrmiller:
Worth a shot but after benchmarking, found to be not worth it. Closing.

PR #6543, opened 2026-04-16
perf: sample before dividing sub-vectors in PQ training

justinrmiller:
## Summary

- Moves sampling **before** `divide_to_subvectors` in PQ codebook training
- Parallelizes KMeans training across sub-vectors with `rayon::into_par_iter()`
- Adds `pq_build` benchmark (1M vectors × 128 dims)

## The Problem

`ProductQuantizer::build_from_fsl()` previously:
1. Called `divide_to_subvectors(data, m)` on the **full dataset** — copying ALL N vectors into m separate arrays
2. Then `train_kmeans()` internally sampled down to just `sample_rate × num_centroids` = 65K vectors

This means for large datasets, step 1 allocated a complete copy of the data that step 2 immediately discarded 99.9% of. This is the operation flagged with `TODO: very intensive memory copy!!!` in `pq/utils.rs`.

## The Fix

Sample `sample_rate × num_centroids` vectors from the input FSL **first** (zero-copy `slice()`), then call `divide_to_subvectors` on only the sample. The training data is identical — `train_kmeans` takes the first N vectors in both cases.

Additionally, since each sub-vector's KMeans training is completely independent, the training loop is now parallelized with `into_par_iter()`.

## Memory Savings

| Dataset Size | Before (divide_to_subvectors alloc) | After | Reduction |
|---|---|---|---|
| 1M × 128-dim f32 | 512 MB | 33 MB | **93.5%** |
| 10M × 128-dim f32 | 5,120 MB | 33 MB | **99.4%** |
| 100M × 768-dim f32 | 307 GB | 200 MB | **99.9%** |

The key insight: `sample_rate × num_centroids` = 256 × 256 = 65,536 vectors is the maximum training set regardless of dataset size, so the memory for sub-vector division is bounded at ~200 MB for 768-dim vectors.

### PQ Build (Apple M-series)

1M × 128-dim vectors, m=16 sub-vectors, 10 KMeans iterations:

| Metric | Before | After | Change |
|--------|--------|-------|--------|
| **Build time** | 1.21 s | 1.02 s | **-15.7%** |
| **Peak memory (sub-vector division)** | 512 MB | 33 MB | **-93.5%** |

### Isolated divide_to_subvectors step (10M × 128-dim)

| Metric | Before (full dataset) | After (sample first) | Improvement |
|--------|----------------------|---------------------|-------------|
| **Time** | 1.24 s | 8.0 ms | **154x faster** |
| **Memory** | 5,120 MB | 33 MB | **153x less (5 GB saved)** |

At 10M vectors, the old code spends 1.24 seconds copying 5 GB of data into sub-vector arrays, only for `train_kmeans` to immediately discard 99.3% of it. For 100M × 768-dim (a realistic production dataset), the savings would be 307 GB → 200 MB — the difference between needing a 600+ GB RAM machine and a 16 GB laptop.

## Correctness & Quality

**No qualitative penalty.** The codebook trained is byte-for-byte identical to what the old code produces:

- **Same training data:** Both paths use the first `sample_rate × num_centroids` vectors. The old code copied all N vectors into sub-vector arrays, then `train_kmeans` called `array.slice(0, sample_size * dimension)` to take the first 65K. The new code does `data.slice(0, sample_size)` before dividing — same vectors, same order.

- **Same KMeans behavior:** `train_kmeans` sees the same input array size (`sample_rate × k` = 65,536 rows), same parameters, same codebook seeds. Since `num_rows == sample_rate * k`, it skips its internal sampling branch entirely.

- **Parallel training preserves results:** Each sub-vector's KMeans is independent. The `into_par_iter().enumerate()` collects results indexed by `sub_vec_idx`, so output ordering is preserved regardless of execution order.

- **Pre-existing sampling note:** The sampling strategy (taking the first N vectors, not random) predates this PR. If the dataset is ordered (by class, time, or source), the head sample may not be representative — this is a known limitation of the existing approach, not something introduced here.

## Test plan
- [x] All 12 PQ tests pass
- [x] `cargo fmt` clean


justinrmiller:
## 10M Vector Benchmark

Isolated benchmark of just the `divide_to_subvectors` step (10M × 128-dim × f32, 16 sub-vectors):

| Metric | Before (full dataset) | After (sample first) | Improvement |
|--------|----------------------|---------------------|-------------|
| **Time** | 1.24 s | 8.0 ms | **154x faster** |
| **Memory** | 5,120 MB | 33 MB | **153x less (5 GB saved)** |

At 10M vectors, the old code spends 1.24 seconds copying 5 GB of data into sub-vector arrays, only for `train_kmeans` to immediately discard 99.3% of it. The fix reduces this to 8ms and 33 MB by sampling the 65K training vectors before dividing.

For reference, at 100M × 768-dim (a realistic production dataset): the savings would be **307 GB → 200 MB** — the difference between needing a 600+ GB RAM machine and a 16 GB laptop.

rust/lance-index/src/vector/pq/builder.rs
```diff
@@ -100,13 +101,24 @@ impl PQBuildParams {
...
-        let sub_vectors = divide_to_subvectors::<T>(data, self.num_sub_vectors)?;
         let num_centroids = 2_usize.pow(self.num_bits as u32);
         let dimension = data.value_length() as usize;
         let sub_vector_dimension = dimension / self.num_sub_vectors;
 
+        // Sample before dividing into sub-vectors. train_kmeans internally
+        // samples to `sample_rate * num_centroids` vectors anyway, so
+        // dividing the full dataset first wastes O(N * D) memory copying
+        // data that is immediately discarded.
+        let sample_size = self.sample_rate * num_centroids;
```
Xuanwo, line 112: Does this conflict with `PQBuildParams::sample_size()`?

Xuanwo (commented):
Nice work! Only one small question.

rust/lance-index/src/vector/pq/builder.rs
BubbleCal, line 112: Did we sample more data for training here? IIUC `data.len() == self.sample_rate * num_centroids` already here
see https://github.com/lance-format/lance/blob/a57ec8142ed85d0f06fd3f90209bee006e91dd85/rust/lance/src/index/vector/builder.rs#L417

justinrmiller: Good catch — `PQBuildParams::sample_size()` returns exactly the same expression. Switched to call `self.sample_size()` so there's one source of truth.

justinrmiller: You're right — on the main indexing path, `maybe_sample_training_data` at [builder.rs:417](https://github.com/lance-format/lance/blob/main/rust/lance/src/index/vector/builder.rs#L417) already sizes the input to `params.sample_size()`, so the new `if data.len() > sample_size` branch is a no-op there.

I left it in as a bound for the other `PQBuildParams::build()` callers that pass full, unsampled data — `rust/lance/benches/distributed_vector_build.rs`, `rust/lance/benches/memtable_read.rs`, and `rust/lance/src/dataset/mem_wal/memtable/flush.rs` (memtable flush is a real prod path). The main speedup on the indexing path is the `into_par_iter` change; for the bench/flush paths, the sample-before-divide is what bounds memory.

Also updated the comment to reflect this, and simplified the formula to call `self.sample_size()` (per @Xuanwo). Happy to split the two changes into separate PRs if that's easier to reason about.

PR #6544, opened 2026-04-16
fix: improve retry handling for rename_if_not_exists

### LINKED ISSUE #6315: Improve retry handling for rename_if_not_exists
> I think we should consider overriding our own rename_if_not_exists handling.
> 
> Currently, the logic retries the entire copy and delete process. If the copy succeeds but the delete fails due to throttling, the entire operation is retried. After that, the copy will fail because the object already exists.

https://github.com/lance-format/lance/pull/6293/changes/BASE..0c985726379c602172c97ba5435ff1e69d8163ee#r3000283579

xuzifu666:
fix: https://github.com/lance-format/lance/issues/6315

The original `rename_if_not_exists` delegated the entire operation to the underlying storage and retried it uniformly. If the underlying implementation is a combination of copy and delete, and copying succeeds but deletion fails due to rate limiting, retrying the entire operation will cause the second copy to fail because the target file exists.

fix way: override the `rename_if_not_exists` method to handle copy and delete separately. This way, if copy succeeds but delete fails due to rate limiting, only delete will be retried, preventing copy from failing because the target file already exists.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved throttling for rename operations by executing destination updates and source cleanup as separate throttled steps, rather than relying on a single throttled rename.
  * Preserved rename semantics and key options (including extension handling) to keep overwrite/create behavior consistent under load.

* **Tests**
  * Expanded coverage for rename scenarios, including success, destination-exists handling, and retry behavior when throttling fails independently for the update and cleanup steps.

### DIFF
--- a/rust/lance-io/src/object_store/throttle.rs
+++ b/rust/lance-io/src/object_store/throttle.rs
@@ -754,29 +754,56 @@ impl ObjectStore for AimdThrottledStore {
     }
 
     async fn rename_opts(&self, from: &Path, to: &Path, opts: RenameOptions) -> OSResult<()> {
-        // Perform copy and delete separately to handle throttling more gracefully.
-        // If the copy succeeds but the delete fails due to throttling, we can retry just
-        // the delete instead of failing because the destination already exists.
         let RenameOptions {
             target_mode,
             extensions,
         } = opts;
-        let copy_mode = match target_mode {
-            RenameTargetMode::Overwrite => CopyMode::Overwrite,
-            RenameTargetMode::Create => CopyMode::Create,
-        };
-        let copy_options = CopyOptions {
-            mode: copy_mode,
-            extensions,
-        };
-        self.write
-            .throttled(|| self.target.copy_opts(from, to, copy_options.clone()))
-            .await?;
 
-        // Delete the source file, with separate retry logic.
-        // Even if this fails, the rename has effectively succeeded (destination exists).
-        // We still try to clean up the source to avoid leaving duplicate data.
-        self.delete.throttled(|| self.target.delete(from)).await
+        match target_mode {
+            RenameTargetMode::Overwrite => {
+                // Delegate to the underlying store to preserve atomic rename
+                // semantics where the store supports them (e.g. local filesystem).
+                self.write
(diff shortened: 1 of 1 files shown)


xuzifu666:
@Xuanwo This pr is related to your suggestion in https://github.com/lance-format/lance/issues/6315, could you help to review it？thanks

rust/lance-io/src/object_store/throttle.rs
```diff
@@ -755,9 +754,29 @@ impl ObjectStore for AimdThrottledStore {
...
         self.write
-            .throttled(|| self.target.rename_opts(from, to, opts.clone()))
-            .await
+            .throttled(|| self.target.copy_opts(from, to, copy_options.clone()))
+            .await?;
+
+        // Delete the source file, with separate retry logic.
+        // Even if this fails, the rename has effectively succeeded (destination exists).
+        // We still try to clean up the source to avoid leaving duplicate data.
+        self.delete.throttled(|| self.target.delete(from)).await
```
yanghua: Here is not consistent, right? It would throw errors to users.

rust/lance-io/src/object_store/throttle.rs
```diff
@@ -755,9 +754,29 @@ impl ObjectStore for AimdThrottledStore {
     }
 
     async fn rename_opts(&self, from: &Path, to: &Path, opts: RenameOptions) -> OSResult<()> {
```
yanghua, line 756: The new implementation broke the atomicity.

yanghua (commented):
Left two comments.

rust/lance-io/src/object_store/throttle.rs
xuzifu666: Great catch. The previous code returned the delete error to the caller even though the comment said the rename had already effectively succeeded. We have refactored rename_opts so that for RenameTargetMode::Create, after a successful copy, the source cleanup is treated as best-effort: if delete fails due to throttling, we log a warning and return Ok(()) instead of propagating the error. This keeps the behavior consistent with the comment and avoids misleading callers into thinking the rename failed.

PR #6545, opened 2026-04-16
Fix crash in `FilteredReadStream` on task cancellation

emilk:


westonpace:
This unwrap is doing panic propagation, a panic in the spawned task will propagate into the parent.  This is done in quite a few places (e.g. mutexes, other spanwed tasks) throughout the code base.  I can see advantages to translating those into errors but I'm curious if there is any particular reason this call site has been a bother?

emilk:
Ah, I did not realize this was intentional!

This came up in a `#[test]` of ours, where we add special panic handlers to get better diagnostics. In my world view, panics are only for bugs and any expected error should be handled by `Result`, so I just assumed this was a bug and not intentional.

PR #6548, opened 2026-04-16
feat: expose base scoped store bindings to java

zhangyue19921010:
Tested: Java can write to different S3 buckets within the same multi-base dataset, using different storage configs for each base.


yanghua:
@zhangyue19921010 There is a conflict that needs to be resolved.

java/src/main/java/org/lance/ReadOptions.java
```diff
@@ -82,6 +88,7 @@ public String toString() {
         .add("indexCacheSizeBytes", indexCacheSizeBytes)
         .add("metadataCacheSizeBytes", metadataCacheSizeBytes)
         .add("storageOptions", storageOptions)
+        .add("baseStoreParams", baseStoreParams)
```
yanghua: Does it contain some sensitive information, e.g. `AK/SK`?

yanghua (commented):
Just left one comment.

java/src/main/java/org/lance/ReadOptions.java
zhangyue19921010: Nice catch. Removed for simplified processing

yanghua (approved):
+1

PR #6549, opened 2026-04-16
fix(java): resolve JNI classloader bug on dispatcher thread in Spark

summaryzb:
## Summary

Fix a JNI classloader bug where the `AsyncScanner` class could not be found on the lance-jni-dispatcher thread in Spark executor environments. Spark loads user JARs through an isolated `MutableURLClassLoader`/`ChildFirstURLClassLoader` chain that is only installed on Spark's managed task threads — JNI native threads attached via `attach_current_thread_permanently()` never receive this classloader and fall back to the JVM system classloader, which cannot see user classes. 
## Problem
Error log
```
thread 'lance-jni-dispatcher' panicked at src/dispatcher.rs:44:22:
AsyncScanner class not found: JavaException
java.lang.NoClassDefFoundError: org/lance/ipc/AsyncScanner
Caused by: java.lang.ClassNotFoundException: org.lance.ipc.AsyncScanner
```

In Spark executor environments, user JARs (including lance) are **not** on the JVM system classpath. Instead, Spark constructs a custom classloader chain during `Executor` initialization (`Executor.scala:newSessionState()`):

```
systemLoader (JVM default)
  └── MutableURLClassLoader / ChildFirstURLClassLoader (user JARs, --jars, dynamic deps)
        └── ExecutorClassLoader (optional, for REPL classes)
```

This classloader is installed **per-thread** inside `TaskRunner.run()` (`Executor.scala:580`):
```scala
Thread.currentThread.setContextClassLoader(isolatedSession.replClassLoader)
```

Only Spark's managed task-pool threads receive this call. When lance's Rust dispatcher thread calls `jvm.attach_current_thread_permanently()`, the JVM creates a new Java thread that:
- Was **not** created by Spark's cached thread pool
- Never goes through `TaskRunner.run()`
- Never has `setContextClassLoader()` called with Spark's custom classloader
- Therefore retains the JVM default system classloader, which **lacks user JARs**

When the dispatcher thread then called `env.find_class("org/lance/ipc/AsyncScanner")`, the JVM used this system classloader to resolve the class and failed:

This is a well-known JNI pitfall: native threads attached to the JVM via `AttachCurrentThread` / `attach_current_thread_permanently` always get the system classloader, not the application classloader that originally loaded the library. 

## Approach

The fix leverages the fact that `JNI_OnLoad` is called on the thread that triggered `System.loadLibrary()`, which has access to the correct application classloader. The class resolution is moved to `JNI_OnLoad`, where a `GlobalRef` to the `AsyncScanner` class is created. This `GlobalRef` is then passed into `Dispatcher::initialize`, so the dispatcher thread never needs to call `find_class` at all -- it uses the pre-resolved class reference to look up method IDs.

This approach is the standard JNI pattern for solving classloader issues: resolve classes eagerly on a thread with the right classloader and cache `GlobalRef`s for later use on native threads.


summaryzb:
Fix https://github.com/lance-format/lance/issues/6577

summaryzb:
@beinan PTAL, this issue is introduced in #6102

sezruby:
Independent repro of this bug on Databricks Runtime — adding evidence to help drive review.

We hit the exact panic from this PR's description (`thread 'lance-jni-dispatcher' panicked at src/dispatcher.rs:44:22 — AsyncScanner class not found: JavaException`) on a clean DBR 18.2 LTS cluster with the published `org.lance:lance-spark-bundle-4.1_2.13:0.4.0-beta.4` installed via the Maven library UI — no shading, no init scripts, no `userClassPathFirst`. The crash is reproducibly triggered by the official lance-spark Databricks setup at https://lance.org/integrations/spark/vendors/databricks/.

Captured executor logs (with `cluster_log_conf={dbfs:{destination:dbfs:/cluster-logs/...}}` set at cluster create time so the executor stderr survives termination):

```
thread 'lance-jni-dispatcher' (2688) panicked at src/dispatcher.rs:44:22:
AsyncScanner class not found: JavaException

ERROR SparkUncaughtExceptionHandler: Uncaught exception in thread Thread-91:
  java.lang.NoClassDefFoundError: org/lance/ipc/AsyncScanner
Caused by: java.lang.ClassNotFoundException: org.lance.ipc.AsyncScanner
    at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:641)
    at java.base/jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(ClassLoaders.java:188)
    at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:526)
```

Reproducer (~30 lines, no custom code):

```scala
import org.apache.spark.sql.{RowFactory, SparkSession}
import org.apache.spark.sql.types._
import scala.util.Random
import scala.collection.JavaConverters._

val spark = SparkSession.builder().getOrCreate()

def randomVec(rng: Random, dim: Int): Array[Float] = {
  val v = new Array[Float](dim); var i = 0
  while (i < dim) { v(i) = rng.nextFloat(); i += 1 }; v
}

val Dim = 128; val N = 10000
val schema = new StructType(Array(
  StructField("rid", IntegerType, false),
  StructField("rvec", ArrayType(FloatType, false), false,
    new MetadataBuilder().putLong("arrow.fixed-size-list.size", Dim.toLong).build())))
val rng = new Random(1337)
val rows = (0 until N).map(i => RowFactory.create(Integer.valueOf(i), randomVec(rng, Dim)))
val df = spark.createDataFrame(rows.asJava, schema).repartition(2)
df.write.format("lance").save(s"/local_disk0/tmp/lance-repro-${System.currentTimeMillis}")
```

Outcome: `[TASK_FAILED_EXECUTOR_LOSS] Reason: Command exited with code 50, uncaught exception`.

Bisecting published bundle versions confirms the regression appeared in lance-core 5.x:

| bundle | underlying lance-core | DBR 18.2 |
|---|---|---|
| `lance-spark-bundle-*:0.3.0-beta.3` | `4.1.0-beta.2` | works |
| `lance-spark-bundle-*:0.4.0` (stable) | `4.0.0` | works |
| `lance-spark-bundle-*:0.4.0-beta.{1..4}` | `5.x` / `6.0.0-beta.4` | crash |

So this PR's fix unblocks the entire lance-core 5.x / 6.x line on Databricks. Currently the `lance-spark-bundle-*:0.4.0` stable workaround pins users to lance-core 4.x, blocking adoption of every feature added since (zonemap-aware planning, distributed index builds, full-text-search SQL extension, etc.).

We filed the corresponding lance-spark issue at https://github.com/lance-format/lance-spark/issues/538 (originally misdiagnosed as an Arrow ABI mismatch — corrected once we got the exec logs). Cross-linking back here so anyone hitting the connector-side report finds your fix.

The fix as written looks correct — caching the `JClass` global ref from `JNI_OnLoad`'s calling thread (which has the right context CL) is the canonical pattern. We had independently arrived at the same patch before searching for prior PRs and finding this one. Happy to test a build of this branch on our DBR cluster if that helps move it along; just need a `liblance_jni.so` for `linux/x86_64` from the PR branch.

CI is green (last run ~30 days ago). @beinan @eddyxu — would appreciate a review pass.

sezruby:
**Validated this PR's fix on DBR 18.2 LTS** — all three test scales pass on a cluster that previously crashed with the published `org.lance:lance-spark-bundle-4.1_2.13:0.4.0-beta.4`.

What I did:

1. Built `liblance_jni.so` from this branch (`summaryzb:fix/jni-classloader-bug`, head `c894e7d72`) for `linux-x86-64` via OrbStack docker (`ubuntu:22.04` + `protoc 25.3` + Rust toolchain pinned by the workspace's `rust-toolchain.toml`).
2. Patched `nativelib/linux-x86-64/liblance_jni.so` inside the published bundle JAR (`zip --update`).
3. Installed the patched JAR on a fresh DBR 18.2 LTS cluster (`Standard_D4s_v5` × 2, no isolation shared, default settings).
4. Ran three probes — same shape as the issue #6577 reproducer:

| n | dim | partitions | result |
|---|---|---|---|
| 10000 | 128 | 2 | ✅ `WROTE_OK 10000` |
| 50000 | 128 | 8 | ✅ `WROTE_OK 50000` |
| 10000 | 1024 | 4 | ✅ `WROTE_OK 10000` |

No `lance-jni-dispatcher` panic, no `NoClassDefFoundError`, executor stays alive. The earlier executor-stderr panic at `dispatcher.rs:44` is gone — `JNI_OnLoad` resolves `org.lance.ipc.AsyncScanner` from the user-facing thread (correct context CL), the `GlobalRef` survives into the dispatcher's spawned native thread, no `find_class` happens on the bad CL anymore.

Also worth noting: the **same fix unblocks the entire lance-core 5.x / 6.x line on Databricks**. Until this PR merges, users have to pin `lance-spark-bundle-*:0.4.0` (lance-core 4.0.0), which means no zonemap-aware planning, no distributed index builds, no full-text-search SQL extension, etc.

CC @beinan @eddyxu — this is ready to land.

summaryzb:
deduplicate with https://github.com/lance-format/lance/pull/6633

sezruby:
@summaryzb thanks for closing — but I think this might have been closed by mistake? The two PRs solve different problems:

- **#6633 (merged)** swaps `attach_current_thread_permanently()` → `attach_current_thread_as_daemon()`. That's a **daemon-thread fix** so the JVM can exit cleanly when the dispatcher thread is the only non-daemon thread left. It does *not* touch class resolution.
- **#6549 (this PR, now closed)** moves `find_class("org/lance/ipc/AsyncScanner")` from the dispatcher thread (system CL — broken on Databricks/managed-Spark) into `JNI_OnLoad` (user CL — works), and passes the `GlobalRef` into the dispatcher. That's the **classloader fix**.

The classloader bug at `dispatcher.rs:44:22` is not addressed by #6633 — looking at the diff, #6633 only changes the `attach_*` call. The `find_class("org/lance/ipc/AsyncScanner")` on line 44 is unchanged, still runs on the dispatcher's freshly-attached thread (system classloader), and still panics on Databricks where user JARs are not on the system classpath (full repro + exec stderr in my comments above).

I validated this PR's fix on DBR 18.2 a few hours ago (built `.so` from `summaryzb:fix/jni-classloader-bug` head `c894e7d72`). With this PR's diff applied, the AsyncScanner panic is gone and `format("lance").save()` works at scale (10K-50K rows, dim=128/1024). Issue #6577 is still open.

If the closure was intentional and you have a different plan to land the classloader fix (e.g. as part of a larger refactor), happy to wait — just want to flag that closing #6549 in favor of #6633 alone leaves Databricks users still hitting `exit code 50` on every `format("lance").save()`.

Could you reopen, or should we file a separate fresh PR with just the classloader portion of your patch?

summaryzb:
`ClassNotFoundException` issue mentioned in https://github.com/lance-format/lance/issues/6577 is solved by this pr, it can be workaround, put the lance-spark bundle jar in the ${SPARK_HOME}/jars directory or set it in the extraClass path  instead of using `--jars ` command.

PR #6552, opened 2026-04-17
feat: add unenforced_clustering_key to format spec

beinan:
## Summary

- Add `unenforced_clustering_key` metadata to the Lance schema format, mirroring the existing `unenforced_primary_key` pattern
- Clustering keys hint at the physical ordering of data within a table, enabling query engine optimizations such as storage-partitioned joins (SPJ)
- Unlike primary keys, clustering key fields may be nullable

Changes across all layers:
- **Protobuf**: `unenforced_clustering_key` (bool) + `unenforced_clustering_key_position` (uint32) fields 14-15
- **Rust core**: field struct, constants, Arrow metadata parsing, schema method
- **Protobuf serialization**: round-trip support with backward compat
- **Java JNI + LanceField**: constructor args and getters
- **Python bindings + type stubs**: `is_unenforced_clustering_key()` / `unenforced_clustering_key_position()`
- **Format docs**: clustering key metadata section

## Motivation

This was discussed in the lance-spark SPJ PR (lance-format/lance-spark#445). Rather than using custom table properties, embedding clustering key info in the schema metadata follows the established pattern and avoids migration issues.

## Test plan

- [x] `cargo check -p lance-core -p lance-file` passes
- [x] `cargo test -p lance-core -p lance-file` passes (all tests including existing primary key tests)
- [x] `cargo clippy -p lance-core -p lance-file --tests -- -D warnings` clean


docs/src/format/table/schema.md
```diff
@@ -247,6 +247,16 @@ Primary key configuration is handled by two protobuf fields in the Field message
 
 For detailed discussion on primary key configuration, see [Unenforced Primary Key](index.md#unenforced-primary-key) in the table format overview.
 
+### Clustering Key Metadata
+
+Clustering key configuration is handled by two protobuf fields in the Field message:
+- **unenforced_clustering_key** (bool): Whether this field is part of the clustering key
```
jackye1995: for unenforced primary key, we initially introduced the boolean and later moved to position because position makes the key fields ordered. I think for clustering, we can just go with position directly without the boolean

beinan: Thanks Jack! Good call - updated to drop the boolean and use only unenforced_clustering_key_position as the single source of truth.

java/src/main/java/org/lance/schema/LanceField.java
jackye1995, line 1: we should probably also add getUnenforcedPrimaryKey() and getUnenforcedClusteringKey() in `LanceSchema`

jackye1995: and similar comment for python

beinan:
Thanks Jack! Added `getUnenforcedPrimaryKey()` and `getUnenforcedClusteringKey()` to `LanceSchema` in both Java and Python, returning fields sorted by position.

jackye1995 (approved):

PR #6554, opened 2026-04-17
fix: follow symlinks when reading manifest file size on local filesystem

### LINKED ISSUE #6497: .lance files can not be symlinks: "file size is too small"
Hi.

Based on my original issue in LanceDB: https://github.com/lancedb/lancedb/issues/3197

If I download a LanceDB dataset from HuggingFace like that:

```sh
hf download --type dataset username/repo
```

Then in the `snapshots/` subfolder of the downloaded artifacts in HuggingFace cache there will be symlinks to the actual blobs:

```sh
10010110010001001001110179af84444b9f7d4ff87ee2253b.lance@ -> ../../../../../blobs/be64f20e5723bd0a27cfdbdb41cf7d6fad94cd572a71973b717fb8340f4310c5
```

That's how HuggingFace stores files. But LanceDB fails to open a table with such symlinks:

```
error: lance error: LanceError(IO): file size is too small, /Users/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/lance-io-3.0.1/src/utils.rs:88:20  Caused by: LanceError(IO): file size is too small, /Users/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/lance-io-3.0.1/src/utils.rs:88:20    Caused by: file size is too small
 code: "GenericFailure"
```

Like it's literally trying to open and read a `.lance` symlink file as is without following it to the original source. The error originates here: https://github.com/lance-format/lance/blob/dabfb933c789b661805c36c9397d4045a1bc76ef/rust/lance-io/src/utils.rs#L86-L89

lilei1128:
DirEntry::metadata() on Unix is equivalent to symlink_metadata() — it
  returns the size of the symlink itself (a few dozen bytes) rather than
  the target file. This caused a panic/IO error when manifest files inside
  _versions/ were symlinks (e.g. HuggingFace cache uses symlinks in
  snapshots/ pointing to blobs/).

  Switch to std::fs::metadata(entry.path()) which follows symlinks, so
  the correct file size is stored in ManifestLocation and passed as
  known_size to the reader.

  Fix #6497

### DIFF
--- a/rust/lance-table/src/io/commit.rs
+++ b/rust/lance-table/src/io/commit.rs
@@ -417,7 +417,11 @@ fn current_manifest_local(base: &Path) -> std::io::Result<Option<ManifestLocatio
     if let Some((version, entry)) = latest_entry {
         let path = Path::from_filesystem_path(entry.path())
             .map_err(|err| io::Error::new(io::ErrorKind::InvalidData, err.to_string()))?;
-        let metadata = entry.metadata()?;
+        // Use std::fs::metadata() rather than DirEntry::metadata() to follow
+        // symlinks — on Unix, DirEntry::metadata() is equivalent to
+        // symlink_metadata() and returns the size of the symlink itself
+        // (a few dozen bytes), not the target file.
+        let metadata = std::fs::metadata(entry.path())?;
         Ok(Some(ManifestLocation {
             version,
             path,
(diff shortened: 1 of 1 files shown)


rust/lance-table/src/io/commit.rs
```diff
@@ -417,7 +417,11 @@ fn current_manifest_local(base: &Path) -> std::io::Result<Option<ManifestLocatio
     if let Some((version, entry)) = latest_entry {
         let path = Path::from_filesystem_path(entry.path())
             .map_err(|err| io::Error::new(io::ErrorKind::InvalidData, err.to_string()))?;
-        let metadata = entry.metadata()?;
+        // Use std::fs::metadata() rather than DirEntry::metadata() to follow
+        // symlinks — on Unix, DirEntry::metadata() is equivalent to
+        // symlink_metadata() and returns the size of the symlink itself
+        // (a few dozen bytes), not the target file.
+        let metadata = std::fs::metadata(entry.path())?;
```
renato2099, line 424: super nit: I _think_ we might be able to use this as well ?
```
entry.path().metadata()?;
```

renato2099 (commented):
not a core approver, but the PR looks good to me 👍 

maybe @wjones127 could help us out here?

deepsweet:
@wjones127 could you please take a look? Been annoying me for a month already since my original https://github.com/lancedb/lancedb/issues/3197 issue, and the above looks like a proper fix. Thank you!

PR #6555, opened 2026-04-17
fix(python): fix tuple bug in _coerce_query_vector condition check

hushengquan:
In `dataset.py` line 6584-6587, the condition to check for numpy arrays was:

```python
elif isinstance(query, (list, tuple)) or (
    _check_for_numpy(query),
    isinstance(query, np.ndarray),
):
```
The expression (_check_for_numpy(query), isinstance(query, np.ndarray)) creates a tuple (bool, bool), which is always truthy (a non-empty tuple). This means any input that is not a list, tuple, pa.Scalar, or pa.Array would incorrectly enter the numpy conversion branch. This will result in an unexpected conversion error: ValueError: could not convert string to float: np.str_('not a vector')


wjones127 (approved):
This is a nice fix. Thanks for working on it.

PR #6558, opened 2026-04-17
fix(blobV2): blob.from_uri optional validation

zhangyue19921010:
Fix validation in Blob.from_uri() when position / size are omitted or explicitly passed as None.
- Previously, from_uri() used None as the default for these parameters, but the
    validation logic directly evaluated position < 0 / size < 0, causing valid calls like
    Blob.from_uri(uri) to raise a TypeError.

### DIFF
--- a/python/python/lance/blob.py
+++ b/python/python/lance/blob.py
@@ -47,10 +47,14 @@ class Blob:
         return Blob(data=bytes(data))
 
     @staticmethod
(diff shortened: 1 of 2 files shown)


wojiaodoubao:
Nice catch. +1 LGTM.

zhangyue19921010:
Hi @Xuanwo Sorry to bother u. Looks like this a bug in blobV2, would u mind to take a look? Thanks!

Xuanwo (approved):
Thank you!

PR #6562, opened 2026-04-17
fix: scale default memory pool size by partition count

westonpace:
## Summary

- The `FairSpillPool` divides memory evenly across spillable consumers. With up to 8 partitions, each sort consumer was limited to ~12.5MB from a flat 100MB pool, causing merge_insert operations with large payloads to fail with "not enough memory to continue external sort" at very small batch sizes (e.g. 5 rows with 1MB payloads).
- Scale the default pool size to 100MB **per partition** so each consumer gets a reasonable allocation. Explicit `LANCE_MEM_POOL_SIZE` or `mem_pool_size` settings are respected as-is.
- This is a partial fix — very large batches can still exhaust the per-partition budget. A more complete fix may involve revisiting the pool type or spilling behavior for merge_insert.

## Test plan

- [x] Added unit test `test_mem_pool_size_scales_with_partitions` verifying pool size scales correctly
- [x] Verified with a Python repro script that merge_insert with 1MB-per-row payloads no longer fails at 5 rows (now succeeds up to ~50 rows)

### DIFF
--- a/rust/lance-datafusion/src/exec.rs
+++ b/rust/lance-datafusion/src/exec.rs
@@ -313,21 +313,22 @@ impl std::fmt::Debug for LanceExecutionOptions {
     }
 }
 
-const DEFAULT_LANCE_MEM_POOL_SIZE: u64 = 100 * 1024 * 1024;
+const DEFAULT_LANCE_MEM_POOL_SIZE_PER_PARTITION: u64 = 100 * 1024 * 1024;
 const DEFAULT_LANCE_MAX_TEMP_DIRECTORY_SIZE: u64 = 100 * 1024 * 1024 * 1024; // 100GB
 
 impl LanceExecutionOptions {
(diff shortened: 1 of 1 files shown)


westonpace:
This is important because Datafusion cannot handle when a single batch is larger than the spill pool size (arguably this is a feature gap, and we should fix that upstream).  So the pool size is a cap on the batch size.

The only place we apply multiple partitions is in merge_insert today and there we cap it at 8 so we would allocate at most 800MiB so this isn't unbounded growth and it keeps our max batch size at 100MiB instead of 12.5MiB which just seems like a slightly easier to understand default.

westonpace:
In a future PR I am going to be splitting batches before they go into Datafusion so we can get from "no batches > 100MiB" to "no rows > 100MiB".

LuQQiu (approved):

PR #6564, opened 2026-04-17
fix: hard cap batch size in merge_insert to prevent sort failures

westonpace:
## Summary

- Add `rechunk_stream_by_size_deep_copy` to `lance-arrow` which deep-copies slices so `get_array_memory_size` is accurate, enabling recursive re-splitting of batches with non-uniform row sizes
- Add `HardCapBatchSizeExec` DataFusion exec node in `lance-datafusion` that caps output batch sizes and errors if a single row exceeds the limit
- Insert `HardCapBatchSizeExec` (25 MiB cap) before the sort in `merge_insert::update_fragments` to prevent DataFusion sort failures when batches exceed the memory pool size
- Add Python test verifying merge_insert succeeds with a ~200 MiB batch of 1000 rows

## Test plan

- [x] Rust unit tests for oversized row at start/middle/end of batch (`lance-arrow::stream::tests::test_oversized_row_*`)
- [x] All existing `lance-arrow::stream` tests pass (14/14)
- [x] Python integration test `test_merge_insert_large_rows` passes

### DIFF
--- a/python/python/tests/test_dataset.py
+++ b/python/python/tests/test_dataset.py
@@ -2690,6 +2690,39 @@ def test_merge_insert_large():
     )
 
 
+def test_merge_insert_large_rows():
+    # Verify that merge_insert succeeds when a single batch is ~200 MiB.
+    # This exercises the HardCapBatchSizeExec node which rechunks oversized
+    # batches before the DataFusion sort.
+    nrows = 1000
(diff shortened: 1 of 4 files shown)


jackye1995 (approved):
looks good to me

Xuanwo (approved):

PR #6569, opened 2026-04-18
fix: map `object_store::Error::NotFound` to `Error::NotFound` instead of `Error::IO`

### LINKED ISSUE #2067: Trying to open non-existent dataset should return Error::NotFound instead of Error::IO

LuciferYang:
Closes https://github.com/lancedb/lance/issues/2067

### Summary

The blanket `From<object_store::Error>` impl mapped **all** object-store errors to `Error::IO`, losing the semantic meaning of "not found." This forced downstream code into fragile multi-level downcasting and left several `Error::NotFound` match arms unreachable for the object-store path (e.g., in `dataset.rs`, `builder.rs`, `insert.rs`).

This PR:
- Discriminates `object_store::Error::NotFound` in the `From` impl so it converts to `Error::NotFound`
- Simplifies two call sites that relied on downcasting (`cleanup.rs`, `refs.rs`)
- Adds tests for the conversion and `#[track_caller]` location propagation

### Breaking Changes

`object_store::Error::NotFound` now produces `Error::NotFound` instead of `Error::IO`. Code matching `Error::IO` to detect object-store not-found conditions (via downcasting) will stop catching them. Match on `Error::NotFound` instead. This is expected during the current `6.0.0-beta.1` pre-release cycle.

### Design Notes

- The `source` field from the object-store variant is intentionally dropped — `Error::NotFound` carries only `uri` and `location`, consistent with all other `Error::not_found()` call sites in the codebase.
- Call sites matching raw `object_store::Error::NotFound` before conversion (e.g., `exists()`, commit resolution, external manifests) are unaffected.

### DIFF
--- a/java/src/test/java/org/lance/FileReaderWriterTest.java
+++ b/java/src/test/java/org/lance/FileReaderWriterTest.java
@@ -256,7 +256,9 @@ public class FileReaderWriterTest {
       LanceFileReader.open("/tmp/does_not_exist.lance", allocator);
       fail("Expected LanceException to be thrown");
     } catch (IOException e) {
-      assertTrue(e.getMessage().contains("/tmp/does_not_exist.lance"));
+      String message = e.getMessage();
+      assertTrue(message.contains("/tmp/does_not_exist.lance"));
+      assertTrue(message.toLowerCase().contains("not found"));
     }
     try {
       LanceFileReader.open("", allocator);
--- a/java/src/test/java/org/lance/FileReaderWriterTest.java
+++ b/java/src/test/java/org/lance/FileReaderWriterTest.java
@@ -256,7 +256,7 @@ public class FileReaderWriterTest {
       LanceFileReader.open("/tmp/does_not_exist.lance", allocator);
       fail("Expected LanceException to be thrown");
     } catch (IOException e) {
-      assertTrue(e.getMessage().contains("Object at location /tmp/does_not_exist.lance not found"));
+      assertTrue(e.getMessage().contains("/tmp/does_not_exist.lance"));
     }
     try {
       LanceFileReader.open("", allocator);
--- a/rust/lance/src/dataset/cleanup.rs
+++ b/rust/lance/src/dataset/cleanup.rs
@@ -57,8 +57,8 @@ use lance_table::{
         manifest::{read_manifest, read_manifest_indexes},
     },
 };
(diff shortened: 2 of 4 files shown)


rust/lance/src/dataset/cleanup.rs
```diff
@@ -324,16 +324,7 @@ impl<'a> CleanupTask<'a> {
...
-            matches!(
-                e,
-                Error::IO { source,.. }
-                    if source
-                      .downcast_ref::<ObjectStoreError>()
-                      .map(|os_err| matches!(os_err, ObjectStoreError::NotFound {.. }))
-                      .unwrap_or(false)
-            )
-        };
+        let is_not_found_err = |e: &Error| matches!(e, Error::NotFound { .. });
```
wjones127, line 327: praise: this is a helpful simplification.

java/src/test/java/org/lance/FileReaderWriterTest.java
```diff
@@ -256,7 +256,7 @@ void testInvalidPath() {
       LanceFileReader.open("/tmp/does_not_exist.lance", allocator);
       fail("Expected LanceException to be thrown");
     } catch (IOException e) {
-      assertTrue(e.getMessage().contains("Object at location /tmp/does_not_exist.lance not found"));
+      assertTrue(e.getMessage().contains("/tmp/does_not_exist.lance"));
```
wjones127: suggestion: it would be nice if we at least asserted "not found" is in the error message.

wjones127 (approved):
Looks okay. I've noted one suggested improvement. Can merge in a few days otherwise.

java/src/test/java/org/lance/FileReaderWriterTest.java
LuciferYang: Added an assertion that the message (case-insensitively) contains `not found` alongside the existing path check:

```java
} catch (IOException e) {
  String message = e.getMessage();
  assertTrue(message.contains("/tmp/does_not_exist.lance"));
  assertTrue(message.toLowerCase().contains("not found"));
}
```

This locks in the `Error::NotFound` semantics across the JNI boundary, not just that the path is echoed back. Pushed in 304a7c72d.

LuciferYang:
Thank you @wjones127

PR #6570, opened 2026-04-18
ci: pin nightly toolchain to 2026-04-16 for ethnum 1.5.2 compat

LuciferYang:
## Summary

The `linux-build` CI job installs unpinned `nightly`, which broke after 2026-04-17 because `ethnum 1.5.2` uses `unsafe { mem::transmute(()) }` to create `TryFromIntError`. Newer nightly builds reject this with `error[E0512]: cannot transmute between types of different sizes`.

Dependency chain: `lance-arrow` → `jsonb 0.5.6` → `ethnum 1.5.2`

This PR pins `nightly-2026-04-16` (last known-good date) in both the toolchain install step and the `cargo +nightly` invocation.

## Root Cause

`ethnum-1.5.2/src/error.rs:16`:
```rust
pub const fn tfie() -> TryFromIntError {
    unsafe { mem::transmute(()) }  // () is 0 bits, TryFromIntError is 8 bits
}
```

Rust nightly `e9e32aca5` (2026-04-17) tightened `transmute` checks, making this a hard error.

## Follow-up

- Upstream fix needed in [`nlordell/ethnum-rs`](https://github.com/nlordell/ethnum-rs) to replace the transmute hack
- Once `ethnum` publishes a fix and `jsonb` picks it up, the pin can be removed

## Test plan

- [x] `linux-build` job passes with pinned nightly

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -94,11 +94,12 @@ jobs:
       RUSTFLAGS: "-D warnings"
     steps:
       - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
-      # pin the toolchain version to avoid surprises
+      # pin the toolchain version to avoid surprises (ethnum 1.5.2
+      # fails on nightly > 2026-04-16 due to a transmute size check)
       - name: Setup rust toolchain
         run: |
-          rustup toolchain install nightly
-          rustup default nightly
+          rustup toolchain install nightly-2026-04-16
+          rustup default nightly-2026-04-16
       - uses: rui314/setup-mold@725a8794d15fc7563f59595bd9556495c0564878 # v1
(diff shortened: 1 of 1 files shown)


LuciferYang:
For the medium-term fix, we've submitted an upstream PR to ethnum-rs: https://github.com/nlordell/ethnum-rs/pull/58

The root cause is that `ethnum v1.5.2` uses `unsafe { mem::transmute(()) }` to create a `TryFromIntError`. Starting with Rust `nightly-2026-04-18` on Linux x86_64, `TryFromIntError` grew from 0 to 1 byte, causing a compile-time `E0512` size mismatch error. The fix replaces both `transmute` hacks (`tfie()` and `pie()`) with safe stdlib alternatives.

Once the upstream fix is released, we can unpin the nightly version in this PR.

jackye1995 (approved):
Thanks for the fix!

LuciferYang:
Thank you @jackye1995  ~

PR #6571, opened 2026-04-19
fix: return error instead of panic for zero-dimension FixedSizeList

LuciferYang:
## Summary

- Add write-path schema validation to reject `FixedSizeList` fields with non-positive dimension, which previously caused a `divide by zero` panic during encoding/decoding
- The validation is recursive, covering nested types (FSL inside Struct, List, or LargeList)
- Users now receive a clear `InvalidInput` error message instead of an opaque panic

## Problem

Writing a dataset with a zero-dimension `FixedSizeList` (e.g., `pa.list_(pa.float32(), 0)`) and then reading it back caused a panic:

```
thread 'lance_background_thread' panicked at rust/lance-encoding/src/data.rs:403:9:
attempt to divide by zero
```

The root cause is `FixedSizeListBlock::num_values()` divides `child.num_values()` by `self.dimension`, which is 0 for zero-dimension lists. Negative dimensions would also cause incorrect behavior due to `i32` to `u64` wrapping.

## Solution

Added a `check_for_zero_dimension_fsl()` validation in `FragmentCreateBuilder::validate_schema()`. This rejects any `FixedSizeList` field with `dimension <= 0` **before** any encoding occurs.

The error message includes the field name and actual dimension value:

```
Field 'vec' has a FixedSizeList with dimension 0.
Only positive dimensions are supported.
```

### DIFF
--- a/rust/lance/src/dataset/fragment/write.rs
+++ b/rust/lance/src/dataset/fragment/write.rs
@@ -339,8 +339,7 @@ impl<'a> FragmentCreateBuilder<'a> {
                     }
                     check_data_type(field_name, child.data_type())
                 }
-                arrow_schema::DataType::List(child)
-                | arrow_schema::DataType::LargeList(child) => {
+                arrow_schema::DataType::List(child) | arrow_schema::DataType::LargeList(child) => {
                     check_data_type(field_name, child.data_type())
                 }
                 arrow_schema::DataType::Struct(fields) => {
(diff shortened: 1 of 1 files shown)


Xuanwo:
Closing this because #7247 was merged and covers zero-dimension FixedSizeList validation, including recursive write validation and read-side panic protection. Please reopen if this PR contains a remaining case not covered by #7247.

PR #6575, opened 2026-04-19
fix(namespace): align error handling with namespace spec

jackye1995:
## Summary

- Fix RestNamespace to parse the spec-defined flat error response format using the generated `ErrorResponse` model, instead of guessing errors from HTTP status codes
- Fix DirectoryNamespace to return correct `NamespaceError` variants per the spec's per-operation error tables
- Fix REST adapter to produce spec-compliant error responses using `ErrorResponse` model, and correct HTTP status code mappings (406 for Unsupported, 409 for NamespaceNotEmpty/InvalidTableState)
- Rename `ErrorCode::Throttled` to `ErrorCode::Throttling` to match Python/Java SDK naming
- Use `{:?}` (Debug) formatting for all source errors to preserve the full error chain

### DIFF
--- a/rust/lance-namespace-impls/src/dir.rs
+++ b/rust/lance-namespace-impls/src/dir.rs
@@ -783,7 +783,7 @@ impl DirectoryNamespaceBuilder {
             .await
             .map_err(|e| {
                 lance_core::Error::from(NamespaceError::Internal {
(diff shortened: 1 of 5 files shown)


Xuanwo (approved):
Thank you!

LuQQiu (approved):

PR #6581, opened 2026-04-21
fix: shift offsets after trimming values in merge_with_schema

### LINKED ISSUE #6580: Panic in take.rs when projecting a list<struct> column under a pushdown filter with sparse selectivity
## Summary

`LanceDataset.to_table(filter=..., columns=[list_struct_col, ...])` panics in `arrow-array`'s `ListArray::try_new` with offsets indexing past the values buffer. The process is killed; Python sees only `pyarrow.lib.ArrowInvalid: External error: RuntimeError: Task was aborted`.

```
thread 'lance_background_thread' panicked at arrow-array-57.3.0/src/array/list_array.rs:272:
  InvalidArgumentError("Max offset of 66823 exceeds length of values 15763")

thread 'lance_background_thread' panicked at rust/lance/src/io/exec/take.rs:376:
  JoinError::Panic(...)
```

## Environment

`pylance` 4.0.0, `pyarrow` 23.0.1, `data_storage_version` 2.0, Linux.

## Reproduction

Self-contained. `pylance` + `pyarrow` only.

```python
import shutil, lance, pyarrow as pa
from pathlib import Path

PATH = Path("/tmp/lance-repro.lance")
if PATH.exists(): shutil.rmtree(PATH)

ITEM = pa.struct([("label", pa.string()), ("x", pa.float32()), ("y", pa.float32()),
                   ("weight", pa.float32()), ("note", pa.string())])
SCHEMA = pa.schema([("id", pa.string()), ("group", pa.uint16()),
                     ("items", pa.list_(ITEM))])

def batch(start, n, group):
    items = [[{"label": f"L{j}", "x": float(j), "y": j*0.5, "weight": 0.1,
               "note": f"n{j}" if j % 2 == 0 else None}
              for j in range(1 + (i % 18))] for i in range(n)]
    return pa.RecordBatch.from_pydict({
        "id": [f"r_{start+i:09d}" for i in range(n)],
        "group": pa.array([group]*n, type=pa.uint16()),
        "items": pa.array(items, type=pa.list_(ITEM)),
    }, schema=SCHEMA)

# 1M rows: first 200k match group=7, next 800k match group=1,
# trailing 14,650 match group=7 again. Single fragment.
batches = [batch(k*100_000, 100_000, 7 if k < 2 else 1) for k in range(10)]
batches.append(batch(10_000_000, 14_650, 7))
lance.write_dataset(pa.RecordBatchReader.from_batches(SCHEMA, iter(batches)),
                    str(PATH), data_storage_version="2.0", max_rows_per_file=2_000_000)

ds = lance.dataset(str(PATH))
assert ds.to_table(columns=["id", "items"]).num_rows == 1_014_650            # OK
assert ds.to_table(filter="group = 7", columns=["id"]).num_rows == 214_650  # OK
ds.to_table(filter="group = 7", columns=["id", "items"])                     # PANIC
```

## Narrowing

| Layout                                                 | Fragments | Result |
|--------------------------------------------------------|-----------|--------|
| 1M rows all matching `group=7`                         | 1         | OK     |
| 1M rows, first 200k match, rest don't                  | 1         | OK     |
| 1M rows, first 200k match + 14,650 trailing match      | 1         | **PANIC** |
| Same, split across 2 fragments                         | 2         | **PANIC** |
| 500k rows, 50/50 match distribution                    | 1         | OK     |
| 200k rows + 1,500 trailing match                       | 1         | OK     |

- **Filter pushdown is required**: full scan on the same columns succeeds.
- **`list<struct>` projection is required**: scalar-only projection under the same filter succeeds.
- **Fragment count is not the trigger**: single-fragment repro (above) panics.

Apparent trigger: a large leading region of selected rows, then a large gap of unselected rows, then a small trailing selected region — so the scan's final batch has far fewer filtered-in rows than earlier batches.

## See also

Closed #5702 — a different `List<Struct>` projection bug (`project_by_schema` in `lance-arrow` skipping `List<Struct>` when reordering struct fields; surfaces as `StructArray::new` errors under schema evolution). Distinct site and trigger, but adjacent code — cross-referenced for context.

kaan-simbe:
## Summary

- Fix panic `InvalidArgumentError("Max offset of X exceeds length of values Y")` in `ListArray::new` during `to_table(filter=..., columns=[list_struct_col, ...])` on v2.0 datasets.
- Root cause: `merge_with_schema` (called from `TakeStream::map_batch`) passed `left_list.trimmed_values()` alongside `left_list.offsets().clone()`. When the left list is a sliced view (e.g. a filtered batch), offsets do not start at zero and reference positions past the end of the trimmed child, panicking in `ListArray::new`.
- Add `ListArrayExt::trimmed_offsets()` that returns offsets shifted to start at zero, and use it in the `List`/`LargeList` branches of `merge_with_schema`.

## Test plan

- [x] New regression test `test_merge_with_schema_sliced_list_struct` in `lance-arrow`: fails on `main` with the exact panic, passes with the fix.
- [x] All existing `lance-arrow` merge tests still pass (9/9).
- [x] Python repro from the issue (1M rows, `200k + 800k + 14650` sparse-tail pattern) no longer panics and returns the expected 214,650 rows with correct data (verified against a manually-filtered reference batch).

Fixes #6580


Xuanwo:
Thanks, I agree with the root cause and the fix direction.

Could you also add a higher-level regression through the scan/take path? The new unit test covers the local invariant, but the user-facing failure is `to_table(filter=..., columns=[list_struct_col])`, which goes through `TakeExec -> merge_with_schema`. A `LargeList` case would also be useful since this PR changes that branch too.

kaan-simbe:
Thanks for the review @Xuanwo!

Added in [aecf0bc](https://github.com/lance-format/lance/pull/6581/commits/aecf0bccb):

- Refactored the lance-arrow test into a generic helper + thin `i32`/`i64` wrappers (matching the existing `test_merge_struct_lists_generic` pattern), so the LargeList branch is covered too.
- Added a higher-level regression in `scanner.rs` that goes through the full `scan().filter().project([list_struct_col])` path. To get the planner to actually emit `TakeExec` with `items` split across input and take (which is what triggers the `List<Struct>` merge branch), I had to nudge it with `MaterializationStyle::AllEarlyExcept([items.b])`. Parametrized over `V2_0` / `Stable` / `V2_2` since the panic is v2-specific.

One note: I couldn't shrink the dataset below ~500k rows and still reproduce the panic — at ~250k it stops triggering, so the regression test runs ~0.7s for all 3 cases.

kaan-simbe:
@Xuanwo anything else I can do here?

xloya:
We encountered a similar problem in our production environment, and the issue was resolved after trying to cherry pick  this PR.

kaan-simbe:
@Xuanwo gentle bump. Both points addressed: the scan/take regression now covers `List` **and** `LargeList` `<Struct>` through the full `scan().filter().project()` → `TakeExec` → `merge_with_schema` path (across `V2_0`/`Stable`/`V2_2`). @xloya also confirmed it fixes the same panic in their production. Could you kick off CI and take another look? Thanks!

kaan-simbe:
#6594 (@LuciferYang) is a duplicate of this PR — same issue #6580, same root cause and fix. This PR additionally covers the `LargeList` branch and adds an end-to-end regression test through the `take.rs`/scanner path. Flagging to avoid parallel review.

wjones127 (approved):
This looks good. Thanks for working on this.

wjones127 (approved):
CI failure unrelated.

PR #6589, opened 2026-04-21
fix(transaction): reduce redundant code for new_version calculations

yangshangqing95:
Fix issue: https://github.com/lance-format/lance/issues/6519

### DIFF
--- a/rust/lance/src/dataset/transaction.rs
+++ b/rust/lance/src/dataset/transaction.rs
@@ -1861,6 +1861,8 @@ impl Transaction {
                     ))
                 });
 
+        let new_version = current_manifest.map_or(1, |m| m.version + 1);
+
         match &self.operation {
             Operation::Clone { .. } => {
                 return Err(Error::internal(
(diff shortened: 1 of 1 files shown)


rust/lance/src/dataset/transaction.rs
```diff
@@ -1959,6 +1959,31 @@ impl Transaction {
             Operation::Merge { fragments, .. } => {
                 final_fragments.extend(fragments.clone());
 
+                // Refresh last_updated_at_version_meta for fragments that were
```
wjones127: suggestion: this sentence looks incomplete. Is this right?

```suggestion
                // Refresh last_updated_at_version_meta for fragments that were touched.
```

rust/lance/src/dataset/transaction.rs
```diff
@@ -1959,6 +1959,31 @@ impl Transaction {
...
+                // Carry-forward fragments that were not touched must not be bumped.
+                if next_row_id.is_some() {
+                    let new_version = current_manifest.map(|m| m.version + 1).unwrap_or(1);
```
wjones127: issue(non-blocking): This is an established pattern, so won't block on this. But it would be much preferred if the `new_version` were only computed once. Right now it's computed in many different places.

wjones127 (approved):
Looks good. I'd be happy to merge this once conflicts are resolved.

rust/lance/src/dataset/transaction.rs
yangshangqing95: Hi @wjones127 thanks for reviewing.
I checked the latest code and found this seems already been fixed by PR#6640 https://github.com/lance-format/lance/pull/6640
Although my PR was submitted earlier, but didn't get any response for a quite long time.
but never mind the issue has been fixed.
So I just reduce some redundant code in this PR and changed the title.
thanks.

wjones127: Got it. Sorry about that. I will take a look at the smaller diff. Thank you for updating!

wjones127 (approved):

PR #6591, opened 2026-04-22
feat: add partitioned zonemap index

beinan:
# Goal

Add a distinct `partitioned_zonemap` scalar index type for distributed index builds, so multiple workers can publish zonemap shards under one index UUID without colliding on `zonemap.lance`.

# Summary

- add `IndexType::PartitionedZoneMap` plus dedicated protobuf details so the index can be identified separately from the existing single-file zonemap
- teach zonemap loading / statistics / parameter derivation to handle partitioned shard files like `part_<fragment_id>_zonemap.lance`
- register a partitioned zonemap scalar plugin and wire `merge_index_metadata` to validate shard files instead of expecting a single merged artifact
- expose the new index type through Python, Java, JNI, and namespace parsing
- add targeted tests for loading multiple zonemap shards and for merge progress during distributed index metadata merge

# Testing

- [x] `cargo fmt --all --check`
- [x] `python3 -m py_compile python/python/lance/dataset.py python/python/lance/indices/__init__.py`
- [x] `git diff --check`
- [ ] Rust compile / test

Rust compile and targeted tests are still blocked locally because this environment is missing `protoc` and `cmake`.

# Notes

- This keeps `partitioned_zonemap` as a separate index type instead of changing the existing zonemap layout.
- The merge step is currently a shard validation pass for the partitioned layout, not a physical merge into one root `zonemap.lance` file.


beinan:
implemented as segmented zonemap index in another pr

PR #6594, opened 2026-04-22
fix: normalize list offsets in merge_with_schema for sliced arrays

LuciferYang:
## Summary
Fixes an "offset past values" panic in `merge_with_schema` when the left input is a sliced `ListArray` (or `LargeListArray`) whose first offset is non-zero.

## Root cause

In `rust/lance-arrow/src/lib.rs`, the list-merge branches pair `trimmed_values()` (which returns values starting at index 0) with `offsets().clone()` (which retains the *original* absolute offsets from before slicing). When the sliced array's first offset is > 0, the re-assembled `ListArray` has offsets that point past the end of the trimmed values buffer, and Arrow's invariant check panics.

Sliced list inputs are common in practice — filtered scans and `take` both produce them, e.g. the trailing batch of a `merge_insert` against a filtered dataset.

## Fix

Added `normalize_offsets()`, which subtracts the first offset from each offset so the buffer is zero-based and consistent with the trimmed values. Zero-first-offset inputs short-circuit and clone without allocation. Applied in both the `List` and `LargeList` merge branches.

## Test plan

- [x] Regression test `test_merge_with_schema_sliced_list_with_nonzero_offset` in `lance-arrow` — builds a `list<struct{id}>`, slices rows 1..3 so offsets start at 2, merges against a schema that adds a new struct field, and verifies the merged batch opens without panic and carries the correct values.
- [ ] `cargo test -p lance-arrow`
- [ ] `cargo clippy --all --tests -- -D warnings`
- [ ] `cargo fmt --all -- --check`


LuciferYang:
Let me fix the CI issues first.

kaan-simbe:
@LuciferYang This duplicates #6581 (same issue #6580, same root cause, same offset zero-basing fix). #6581 also covers the `LargeList` branch and adds an end-to-end regression test through the `take.rs`/scanner path that triggers the panic. Suggest we consolidate on one to avoid splitting review — happy to coordinate.

wjones127:
Closing in favor of #6581

PR #6604, opened 2026-04-24
fix: pin lance-namespace<0.7 in pylance dependencies

LuciferYang:
## Summary

- Add upper bound `<0.7` to `lance-namespace` dependency in `python/pyproject.toml` to prevent released pylance wheels from breaking when `lance-namespace>=0.7` is installed

## Root cause

`lance-namespace 0.7.0` removed `CreateEmptyTableRequest` / `CreateEmptyTableResponse` symbols. Released pylance wheels (e.g. `2.0.1`, `4.0.0b1`) still import these symbols in `lance/namespace.py`, causing `ImportError` when users `pip install` an old pylance version and pip resolves `lance-namespace>=0.7`.

PR #6597 fixed this for CI compat tests by pinning in `venv_manager.py`, but real users installing from PyPI are still affected.

Ref: https://github.com/lance-format/lance/pull/6597#issuecomment-4306020639

## Test plan

- [x] Existing CI jobs remain green
- [x] Users installing older pylance wheels resolve `lance-namespace<0.7` and avoid ImportError

### DIFF
--- a/python/pyproject.toml
+++ b/python/pyproject.toml
@@ -1,7 +1,7 @@
 [project]
 name = "pylance"
 dynamic = ["version"]
-dependencies = ["pyarrow>=14", "numpy>=1.22", "lance-namespace>=0.6.1"]
+dependencies = ["pyarrow>=14", "numpy>=1.22", "lance-namespace>=0.6.1,<0.7"]
 description = "python wrapper for Lance columnar format"
 authors = [{ name = "Lance Devs", email = "dev@lance.org" }]
 license = { file = "LICENSE" }


LuciferYang:
cc @jackye1995 FYI

jackye1995:
I already created a patch in 4.0.0, will publish 4.0.1 with it: https://github.com/lance-format/lance/discussions/6603.

The API is already depreacted for one release, so the plan is to fully remove it in 5.0.0+

jackye1995:
For the main branch, I am raising a PR for bumping the lance-namespace verson

wjones127:
Looks like this has been solved elsewhere. Closing.

https://github.com/lance-format/lance/blob/98efff9c77ff5d6eb752a1d6fa6ad722c327ab13/python/pyproject.toml#L4

PR #6606, opened 2026-04-24
docs: add fragment sizing guidance to performance guide

hamersaw:
## Summary
- Adds a **Fragment Sizing** subsection to the Performance Guide that frames the core tradeoff between manifest-level operations (cost scales with fragment count) and fragment-level operations (cost scales with fragment size, drives per-fragment conflict detection).
- Gives practical guidance: 1M rows/fragment default holds up to ~1B rows, tens of thousands of fragments is generally fine, 10 GB–100 GB is a reasonable per-fragment upper range (1 TB hard ceiling), and concurrent update/delete/merge_insert workloads should err toward more fragments.

## Test plan
- [ ] `mkdocs serve` renders the new section cleanly under Performance Guide.
- [ ] Cross-link to `../format/table/transaction.md#conflict-resolution` (already present in the surrounding section) still resolves.

### DIFF
--- a/docs/src/guide/performance.md
+++ b/docs/src/guide/performance.md
@@ -189,6 +189,37 @@ use cases. For example, S3 can typically get up to 5000
 req/s and with these settings we should get there in about
 10 seconds.
 
+## Fragment Sizing
+
+A Lance table is a collection of fragments tracked by a manifest. How you size those fragments
+trades off two classes of work:
+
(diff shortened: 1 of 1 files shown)


wjones127 (approved):

wjones127:
FYI @geruh we may want to eventually update this after benchmarking is done with https://github.com/lance-format/lance/discussions/7499

PR #6609, opened 2026-04-24
feat: generalize dynamic object store credentials

jackye1995:
## Summary
- generalize namespace-vended dynamic credentials across AWS, Azure, and GCP object_store backends
- add a cached OpenDAL dynamic store wrapper and use it for OSS and HuggingFace
- keep Azure and GCS OpenDAL paths static while cleaning up provider-specific option normalization

### DIFF
--- a/rust/lance-io/src/object_store.rs
+++ b/rust/lance-io/src/object_store.rs
@@ -34,9 +34,15 @@ use url::Url;
 use super::local::LocalObjectReader;
 #[cfg(target_os = "linux")]
 use crate::uring::{UringCurrentThreadReader, UringReader};
+#[cfg(any(feature = "aws", feature = "azure", feature = "gcp"))]
+pub(crate) mod dynamic_credentials;
+#[cfg(any(feature = "oss", feature = "huggingface"))]
+pub(crate) mod dynamic_opendal;
 mod list_retry;
 pub mod providers;
 pub mod storage_options;
+#[cfg(test)]
+pub(crate) mod test_utils;
 pub mod throttle;
 mod tracing;
 use crate::object_reader::SmallReader;
--- a/rust/lance-io/src/object_store/dynamic_credentials.rs
+++ b/rust/lance-io/src/object_store/dynamic_credentials.rs
@@ -72,20 +72,39 @@ impl<T> NamespaceCredentialsProvider<T> {
     }
 }
 
-pub async fn supports_dynamic_credentials<T>(accessor: &Arc<StorageOptionsAccessor>) -> Result<bool>
+/// Build a dynamic credential provider for any cloud type, returning `None`
+/// if the accessor has no provider or the vended options are incompatible with `T`.
+pub async fn build_dynamic_credential_provider<T>(
+    accessor: Option<Arc<StorageOptionsAccessor>>,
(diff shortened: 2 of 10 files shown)


LuQQiu:
@claude review

LuQQiu:
P1 — Azure TryFrom key proliferation

  dynamic_credentials.rs:205-211 — The Azure access key lookup checks 6 different key names (azure_storage_account_key, azure_storage_access_key, azure_storage_master_key,
   access_key, master_key, account_key). The OpenDAL normalization in azure.rs:56-98 has a similar but not identical list. The access_key and master_key bare names are
  very generic and could collide with other cloud providers' keys if the DynamicCredentials bag ever carries multi-cloud options. Consider tightening to only the
  azure_storage_* prefixed keys in the TryFrom implementation, or at least documenting why the bare names are needed.

  P1 — DynamicOpenDalStore config equality as cache key

  dynamic_opendal.rs:97-107 — Cache invalidation relies on HashMap equality of the normalized config. This works correctly but means any non-credential option change
  (e.g., a different endpoint URL returned by the provider) triggers a full store rebuild. This is probably fine for now but worth noting in a comment, especially since
  the comment on line 98-100 says "Providers should return stable, canonicalized values" — this is a requirement that external providers must uphold without enforcement.

  P1 — DynamicOpenDalStore::list fetches credentials once per stream

  dynamic_opendal.rs:86-97 — The list method calls current_store() once at the beginning of the stream. For long-running list operations, credentials could expire
  mid-stream. Other methods (get_opts, put_opts, etc.) have the same pattern but are single requests. The list stream could be long-lived. This is probably acceptable
  since the underlying OpenDAL store handles its own request-level auth, but worth noting.

  P1 — Missing test coverage for negative/error paths

  The tests cover the happy path well but don't test:
  - TryFrom<DynamicCredentials> failure cases (missing keys → missing_dynamic_credential error)
  - Azure split_azure_sas with empty/malformed SAS tokens
  - DynamicOpenDalStore behavior when the provider returns an error
  - supports_dynamic_credentials when initial options contain incompatible credentials

  The codecov report shows 68% patch coverage. Given the project standard ("All bugfixes and features must have corresponding tests"), the error paths should be covered.

  P2 — normalize_opendal_azure_options doesn't normalize bearer_token

  azure.rs:32-98 — The TryFrom<DynamicCredentials> for AzureCredential (in dynamic_credentials.rs:197-200) accepts azure_storage_token and bearer_token, but
  normalize_opendal_azure_options doesn't extract or forward these. If someone passes bearer_token through storage options, it won't reach the OpenDAL operator. This is
  fine if OpenDAL Azure doesn't support bearer tokens, but it's an asymmetry between the two paths.

  P2 — HuggingFace env var handling moved outside normalization

  huggingface.rs:137-143 — Environment variable fallback (HF_TOKEN, HUGGINGFACE_TOKEN) is now applied to base_options before normalization, rather than during store
  construction. This means the env var value gets baked into the base options and will persist even if the env var changes. The old code read env vars at store
  construction time. This is probably fine since stores are typically long-lived, but it's a subtle behavioral change.

LuQQiu (approved):

jackye1995:
@claude review

jackye1995:
@claude review

PR #6610, opened 2026-04-24
fix: reject Rewrite vs CreateIndex when FRI groups straddle bitmap

wjones127:
## Summary

`check_rewrite_txn` returned `Ok` for any `(None, Some(_))` — a Rewrite with a deferred-remap FRI committing against a concurrently committed CreateIndex — on the assumption that the FRI would correct index bitmaps post-commit. That is only safe when each rewrite group's old fragments are fully inside or fully outside the new index's bitmap. If a group straddles (some fragments indexed, some not), `load_indices` later fails with `split of indexed and non-indexed data` and queries against the index break.

This PR rejects the straddling case as a retryable conflict so the rewrite is replanned against the post-CreateIndex view.

Production trigger (with `defer_index_remap=true` now enabled):
1. Compaction plans `[frag1, frag2]` while both are unindexed.
2. `optimize_indices` commits a CreateIndex covering `frag1` only (ran on an older version that didn't see `frag2`).
3. Compaction's Rewrite commits against a read_version earlier than the CreateIndex; the conflict resolver incorrectly returns COMPATIBLE.
4. Queries fail: the FRI group `[frag1, frag2]` straddles the new index bitmap.

Also fixes an `.unwrap()` in `remap_fragment_bitmap` so the underlying error propagates instead of panicking.

## Test plan

- [x] New test `test_rewrite_fri_vs_create_index_conflict` in `rust/lance/src/dataset/optimize.rs` reproduces the production sequence and asserts `commit_compaction` returns `RetryableCommitConflict` (before the fix, commit succeeded and the query failed with "split of indexed and non-indexed data").
- [x] Existing `dataset::optimize::` and `io::commit::` tests pass.

### DIFF
--- a/rust/lance/src/dataset/optimize.rs
+++ b/rust/lance/src/dataset/optimize.rs
@@ -4639,36 +4639,28 @@ mod tests {
             .unwrap();
 
         // optimize_indices on the stale handle indexes frag1 only (frag2
-        // didn't exist at that version), commits as CreateIndex.
+        // didn't exist at that version), commits as CreateIndex. `dataset`
+        // stays at its pre-optimize version so the Rewrite commit has to
+        // conflict-check against this CreateIndex.
         stale
             .optimize_indices(&lance_index::optimize::OptimizeOptions::append())
             .await
             .unwrap();
-        dataset.checkout_latest().await.unwrap();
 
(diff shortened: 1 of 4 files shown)


jackye1995 (approved):
Thanks for the fix!

PR #6617, opened 2026-04-25
feat: replace Azure SDK and google-cloud-auth with direct reqwest for credential vending

jackye1995:
## Summary

- Replace Azure SDK crates (`azure_core`, `azure_identity`, `azure_storage`, `azure_storage_blobs`, `time`) and `google-cloud-auth` with direct HTTP calls via `reqwest`
- Add `hmac` and `quick-xml` for Azure SAS signing and XML parsing (both already transitive deps, zero compile cost)
- Add `ring` and `rustls-pki-types` for GCP service account JWT signing (both already transitive deps)
- Add Workload Identity Federation (OIDC) support for both Azure and GCP credential vending
- Reduce Cargo.lock by ~500 lines of removed transitive dependencies

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,12 +3,6 @@
 version = 4
 
 [[package]]
-name = "RustyXML"
-version = "0.3.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "8b5ace29ee3216de37c0546865ad08edef58b0f9e76838ed8959a84a990e58c5"
-
-[[package]]
 name = "addr2line"
 version = "0.25.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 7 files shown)


jackye1995:
@claude review

LuQQiu (approved):

LuQQiu:
We need to manage the credential, seems more work to maintain the credential vending, but good that we don't need to depends on those object storage crates?

jackye1995:
> We need to manage the credential, seems more work to maintain the credential vending, but good that we don't need to depends on those object storage crates?

This was causing problem that we could not enable this for Azure because the Azure SDK somehow brings an old dependency that messes up with the OpenSSL. Also it inflates the Cargo.lock a lot, so I think using pure reqwest seems like the best way forward

PR #6620, opened 2026-04-27
fix: use relaxed recall assertion for HNSW-based indexes

XuQianJin-Stars:
fix: remove unused imports, structs and functions in test module

HNSW is an approximate nearest neighbor search algorithm that does not guarantee finding all results. The strict assert_eq! on result count causes flaky test failures (e.g., 999 instead of 1000) for IVF-HNSW-PQ and IVF-HNSW-SQ index types.

Replace the exact assertion with a recall-based check (>= 99%) for HNSW indexes while keeping the strict assertion for exact index types (IVF-PQ, IVF-Flat).

### DIFF
--- a/rust/lance/src/index/vector/ivf.rs
+++ b/rust/lance/src/index/vector/ivf.rs
@@ -3424,6 +3424,11 @@ mod tests {
         #[case] test_case: CreateIndexCase,
         #[values(IndexFileVersion::Legacy, IndexFileVersion::V3)] index_version: IndexFileVersion,
     ) {
+        // HNSW is an approximate search algorithm, so it may not return all results.
+        let is_hnsw = matches!(
+            &test_case.index_type,
+            TestIndexType::IvfHnswPq { .. } | TestIndexType::IvfHnswSq { .. }
+        );
         let mut index_params = match test_case.index_type {
             TestIndexType::IvfPq { pq } => VectorIndexParams::with_ivf_pq_params(
                 test_case.metric_type,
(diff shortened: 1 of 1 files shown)


rust/lance/src/index/vector/ivf.rs
```diff
@@ -3093,6 +3093,11 @@ mod tests {
         #[case] test_case: CreateIndexCase,
         #[values(IndexFileVersion::Legacy, IndexFileVersion::V3)] index_version: IndexFileVersion,
     ) {
+        // HNSW is an approximate search algorithm, so it may not return all results.
+        let is_hnsw = matches!(
+            &test_case.index_type,
+            TestIndexType::IvfHnswPq { .. } | TestIndexType::IvfHnswSq { .. }
+        );
```
yanghua: Why do we only care about `hnsw`? Most vector search algorithms are approximate search, right?

rust/lance/src/index/vector/ivf.rs
```diff
@@ -3165,7 +3170,19 @@ mod tests {
             .try_into_batch()
             .await
             .unwrap();
-        assert_eq!(results.num_rows(), num_non_null);
```
yanghua, line 3499: It is not an exact assertion.

yanghua (commented):
Left two comments.

rust/lance/src/index/vector/ivf.rs
```diff
@@ -3093,6 +3093,8 @@ mod tests {
         #[case] test_case: CreateIndexCase,
         #[values(IndexFileVersion::Legacy, IndexFileVersion::V3)] index_version: IndexFileVersion,
     ) {
+        // Most vector search algorithms are approximate, so they may not return all results.
+        let is_approximate = !matches!(&test_case.index_type, TestIndexType::IvfFlat);
```
yanghua, line 3429: Actually, `IvfFlat` is also approximate; it depends on `nprobes `. But here it is non-approximate.

So we can add a remark here:

```
IvfFlat is exact under this test's parameters.
```

rust/lance/src/index/vector/ivf.rs
```diff
@@ -3165,7 +3167,19 @@ mod tests {
             .try_into_batch()
             .await
             .unwrap();
-        assert_eq!(results.num_rows(), num_non_null);
+        // Use a relaxed assertion for approximate indexes.
+        if is_approximate {
+            let recall = results.num_rows() as f32 / num_non_null as f32;
+            assert!(
+                recall >= 0.99,
+                "Recall too low: {} ({} / {})",
```
yanghua: `Recall {} below threshold {} ({}/{})`

yanghua (commented):
Left two comments.

yanghua (approved):
LGTM. @XuQianJin-Stars, thanks for your patience.

PR #6621, opened 2026-04-27
refactor!: add NamespaceClientTableContext for cached namespace info

jackye1995:
## Summary
- introduce explicit NamespaceClientTableContext across Rust, Python, and Java namespace paths
- centralize namespace location, storage option merge, and managed-versioning handling in Rust
- update namespace distributed write, file, fragment, commit, and integration coverage for cached table context

### DIFF
--- a/java/lance-jni/src/blocking_dataset.rs
+++ b/java/lance-jni/src/blocking_dataset.rs
@@ -19,7 +19,6 @@ use arrow::ffi::FFI_ArrowSchema;
 use arrow::ffi_stream::ArrowArrayStreamReader;
 use arrow::ffi_stream::FFI_ArrowArrayStream;
 use arrow::ipc::writer::StreamWriter;
-use arrow::record_batch::RecordBatchIterator;
 use arrow_schema::DataType;
 use arrow_schema::Schema as ArrowSchema;
 use chrono::{DateTime, Utc};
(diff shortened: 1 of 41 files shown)


jackye1995:
@claude review

LuQQiu:
```
Findings:

  1. rust/lance/src/dataset/write/commit.rs:88-113
     and :249-289 allow a Dataset destination to be
     paired with arbitrary namespace_client / table_id /
     cached context, but never verify that the dataset’s
     URI matches the resolved namespace location.
     with_namespace() simply wraps the existing dataset,
     and execute() then commits through
     dataset.object_store / dataset.base while using the
     namespace-derived commit handler/context. A caller
     can therefore write table A’s data while
     registering the version under table B’s namespace
     entry if they pass mismatched inputs. The Python
     binding exposes exactly this path at python/src/
     dataset.rs:2485-2492.
  2. rust/lance/src/dataset/write/commit.rs:262-289
     computes effective_store_params from the namespace
     context, but on the dataset-backed path those
     params are discarded because execute() reuses
     dataset.object_store.clone() instead of rebuilding
     from the namespace-resolved URI. The returned
     dataset at :503-537 also clones the old dataset
     state, so it still does not carry the namespace-
     derived provider/store params. In practice,
     attaching namespace_client to a dataset that was
     opened via plain URI/static credentials will not
     actually switch commit I/O onto namespace-vended
     credentials or refresh behavior.
  3. rust/lance/src/dataset/builder.rs:132-149 no longer
     forwards the requested version into
     DescribeTableRequest; it always describes the
     latest table metadata and only performs a local
     checkout afterwards. That is a behavioral
     regression for custom namespace implementations
     that use request.version to vend version-scoped
     locations, credentials, or authorization. The new
     Python test in python/python/tests/
     test_namespace_dir.py:1118-1122 even locks in this
     changed behavior, so third-party namespace clients
     would break without any coverage catching it.

```

LuQQiu (approved):

PR #6622, opened 2026-04-27
fix(json): detect float64-stored numbers in json type extraction

### LINKED ISSUE #6619: BUG: JSON extraction mis-tag floats as int64
Hi team, I founds this issue when I was dump and re-extract it with `pylance`.

```py
>>> import json, lance, pyarrow as pa
>>> data = [json.dumps({"v": x}) for x in [1.0, 2.7, 1.5]]
>>> tbl = pa.table({"id": [1,2,3],
...                 "data": pa.array(data, type=pa.json_())})
>>> lance.write_dataset(tbl, "/tmp/jtype.lance", mode="overwrite")
[2026-04-26T22:40:01Z WARN  lance::dataset::write::insert] No existing dataset at /tmp/jtype.lance, it will be created
<lance.dataset.LanceDataset object at 0x113a2eed0>
>>> 
>>> ds = lance.dataset("/tmp/jtype.lance")
>>> q = ds.sql("SELECT id, json_extract_with_type(data, '$.v') AS r FROM dataset").build()
>>> batches = q.to_batch_records()
>>> for b in batches:
...     print(b)
...     
pyarrow.RecordBatch
id: int64
r: struct<value: large_binary, type_tag: uint8 not null>
  child 0, value: large_binary
  child 1, type_tag: uint8 not null
----
id: [1,2,3]
r: -- is_valid: all not null
-- child 0 type: large_binary
[2000000020000009603FF0000000000000,200000002000000960400599999999999A,2000000020000009603FF8000000000000]
-- child 1 type: uint8
[2,2,2]
```
As you can tell, the floating points are tagged as integer type.

Quickly using LLM to check the json-related implementation, the culprit could be here, which will take all numeric types as an integer: https://github.com/lance-format/lance/blob/29394437d84b90b2f977229cb8c07fcfa1979369/rust/lance-datafusion/src/udf/json.rs#L395-L401

dentiny:
Closes https://github.com/lance-format/lance/issues/6619

In the current implementation `extract_json_path_with_type` decides between `Int64` and `Float64` by checking whether the number could be converted i64. This PR checks internal value and decides whether floating point.

### DIFF
--- a/rust/lance-datafusion/src/udf/json.rs
+++ b/rust/lance-datafusion/src/udf/json.rs
@@ -393,8 +393,9 @@ fn extract_json_path_with_type(jsonb_bytes: &[u8], path: &str) -> Result<Option<
                 } else if raw.is_boolean().unwrap_or(false) {
                     JsonbType::Boolean
                 } else if raw.is_number().unwrap_or(false) {
-                    // Try to determine if it's an integer or float
-                    if raw.to_i64().is_ok() {
+                    let is_float_storage =
+                        matches!(raw.as_number(), Ok(Some(jsonb::Number::Float64(_))));
+                    if !is_float_storage && raw.is_i64().unwrap_or(false) {
                         JsonbType::Int64
                     } else {
                         JsonbType::Float64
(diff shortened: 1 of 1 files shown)


westonpace (commented):
Looks like we're missing testing for floats in general.  Would you mind adding test cases for json_get_float to confirm that is behaving correct as well?

dentiny:
> Would you mind adding test cases for json_get_float to confirm that is behaving correct as well?

Sure

westonpace (approved):
Thanks!

westonpace:
The python compat failures are worrisome but seem unrelated to this PR.

PR #6625, opened 2026-04-27
fix: tolerate and auto-heal FRI bitmap straddle corruption (#6623)

### LINKED ISSUE #6623: Concurrent deferred-remap compaction + optimize_indices can corrupt table; list_indices panics
## Summary

A deferred-remap compaction (`compact_files(defer_index_remap=True)`) committed concurrently with an `optimize_indices` against an older dataset version can leave the table in a corrupt state where any call that walks the indices — including `Dataset.list_indices()` — panics with:

```
called `Result::unwrap()` on an `Err` value: InvalidInput {
  source: "The compaction plan included a rewrite group that was a split of indexed and non-indexed data: [...]",
  location: rust/lance-index/src/frag_reuse.rs:330
}
```

The panic site is the `.unwrap()` in `Dataset::load_indices` (`rust/lance/src/index.rs:824` on released builds, `:902` on `main`) when applying the FRI's `remap_fragment_bitmap` to a user index whose `fragment_bitmap` straddles a rewrite group's `old_frags`.

This is a **table-corruption bug**: once committed, the table can no longer be read via the index APIs. PR #6610 prevents *new* occurrences by rejecting the conflicting commit, but does not repair tables that were already written in the broken state.

## Reproduction

Reliably reproduces on `main` prior to #6610 with the following sequence:

1. Write `frag0`, build a vector index over it.
2. Append `frag1`, snapshot a stale `Dataset` handle.
3. Append `frag2` on the up-to-date handle.
4. `plan_compaction` + `rewrite_files` of `[frag1, frag2]` with `defer_index_remap=true` (compute `RewriteResult` but do not commit).
5. On the stale handle, run `optimize_indices` — commits a `CreateIndex` covering `frag1` only (frag2 didn't exist at that version).
6. Commit the rewrite via `commit_compaction`. On pre-#6610 builds this succeeds.

Resulting state:

- index segment A: `bitmap = {frag0}` (original)
- index segment B: `bitmap = {frag1}` (from stale optimize)
- FRI group: `old=[frag1, frag2] → new=[frag3]`

Segment B's bitmap straddles the FRI group, and any `load_indices` call panics.

A self-contained Rust test that reproduces this on `main` is in flight — link to draft PR will follow.

## Production trigger

Reported on a real table with a rewrite group `old=[382843, 382844, 382845, 382846]`. `ds.list_indices()` panics; the table is unreadable through the index path.

## Impact

- Severity: **critical** — silent commit corrupts the table; reads via the index path fail afterward.
- Reachable any time `defer_index_remap=true` is enabled and `optimize_indices` runs concurrently with compaction (the production trigger described in #6610).
- PR #6610 fixes the *write* path. This issue tracks **detect + repair** for already-corrupt tables.

## Proposed fix

Add a `Dataset.repair()` API (Rust + Python) that:

1. Detects this corruption non-destructively (reads via `read_manifest_indexes`, bypassing the FRI auto-remap in `load_indices`).
2. Repairs by removing the straddling old fragment IDs from each affected index segment's `fragment_bitmap`. Previously-indexed rows in the merged new fragment fall through to flat scan until the next `optimize_indices` — no data loss, no retraining required.

Also:

- Have `Dataset.validate()` surface this as a structured error rather than a panic.
- Ship a checked-in `test_data/` fixture (built from a pinned pre-#6610 build) so regressions in detection/repair are caught.

wjones127:
## Summary

- Tables corrupted by the bug fixed in #6610 (a deferred-remap compaction committing concurrently with `optimize_indices` on a stale handle leaves a user index whose `fragment_bitmap` straddles an FRI rewrite group) used to panic in `Dataset::load_indices` and were unreadable.
- `remap_fragment_bitmap` now tolerates the straddle on read: it drops the indexed old-frag IDs from the bitmap, logs `tracing::warn!`, and skips the new-frag insertion. Affected rows fall through to flat scan until the next `optimize_indices`. No data lost; no retrain.
- The fix auto-heals on the next write. `commit_transaction` already seeds `build_manifest` from `load_indices()`, which now returns the cleaned bitmaps, so any commit (e.g. `delete("false")`) persists the corrected manifest. No new API, no user action required.

## Reproduction & test fixture

`rust/examples/src/fri_straddle_datagen.rs` reproduces the corrupt state deterministically on pre-#6610 builds via the public `plan_compaction` / `CompactionTask::execute` / `optimize_indices` / `commit_compaction` APIs. Python can't reach this path — `Compaction.commit` hardcodes `CompactionOptions::default()` (no `defer_index_remap`) — so the generator is in Rust. On a fixed build the generator fails loudly on the conflict-resolver rejection, which is the desired forward-compat behaviour.

The output is checked in at `test_data/fri_straddle_pre_6610/` and exercised by `index::tests::test_load_indices_tolerates_fri_straddle` and `test_auto_heal_persists_cleaned_bitmap`. The latter opens the fixture, performs a no-op `delete("false")`, and asserts the raw on-disk `fragment_bitmap` (read via `read_manifest_indexes`) matches the cleaned bitmap.

## Test plan

- [x] `cargo test -p lance --lib index::tests::test_load_indices_tolerates_fri_straddle index::tests::test_auto_heal_persists_cleaned_bitmap`
- [x] `cargo test -p lance --lib index::tests` — 65 passing
- [x] `cargo clippy -p lance --lib -- -D warnings`
- [x] `cargo fmt --all`

## Follow-ups (out of scope)

- Explicit `Dataset::repair()` API for users who want an auditable trigger / report without writing.
- Better `Dataset::validate()` reporting that surfaces the corruption explicitly rather than silently healing.

Fixes #6623


rust/examples/Cargo.toml
wjones127, line 1: issue(blocking): I don't think this should be a part of the workspace. It needs to be pinned to an older version of the lance library. I'd recommend just making it a standalone rust project within the subdirectory.

test_data/fri_straddle_pre_6610/datagen/src/main.rs
```diff
@@ -0,0 +1,158 @@
...
+            "commit_compaction failed: {e}. \
+             This generator must be run on a Lance build without PR #6610 applied."
+        )
+    })?;
+
+    println!(
+        "Generated fri_straddle fixture at {}",
+        args.output.display()
+    );
+    Ok(())
```
wjones127, line 166: issue(non-blocking): before the end of the script, we should call cleanup_old_versions and just keep the latest version. That reduces the number of files we need to check into the repo.

rust/examples/Cargo.toml
wjones127: Done in f5597314a — moved to test_data/fri_straddle_pre_6610/datagen/ as a standalone crate (empty [workspace] table) with lance pinned to the v6.0.0-beta.3 git tag.

test_data/fri_straddle_pre_6610/datagen/src/main.rs
wjones127: Done — the generator now calls cleanup_old_versions(0, Some(true), None) before exiting.

westonpace (approved):
Nicely done.  Thanks for the extensive testing!

PR #4537, opened 2025-08-22
feat: apply bitpacking on rep/def for compression

Xuanwo:
WIP

Issue #3102, opened 2024-11-07
2.1 miniblock + bitpacking does not support chunks of all-zero data


westonpace:
The bit width is zero and we get:

```
thread 'lance-cpu' panicked at rust/lance-encoding/src/compression_algo/fastlanes.rs:626:18:
internal error: entered unreachable code: Unsupported width: 0
stack backtrace:
   0: rust_begin_unwind
             at /rustc/eeb90cda1969383f56a2637cbd3037bdf598841c/library/std/src/panicking.rs:665:5
   1: core::panicking::panic_fmt
             at /rustc/eeb90cda1969383f56a2637cbd3037bdf598841c/library/core/src/panicking.rs:74:14
   2: <u32 as lance_encoding::compression_algo::fastlanes::BitPacking>::unchecked_pack
             at ./src/compression_algo/fastlanes.rs:626:18
   3: lance_encoding::encodings::physical::bitpack_fastlanes::BitpackMiniBlockEncoder::chunk_data
             at ./src/encodings/physical/bitpack_fastlanes.rs:1612:17
   4: <lance_encoding::encodings::physical::bitpack_fastlanes::BitpackMiniBlockEncoder as lance_encoding::encoder::MiniBlockCompressor>::compress
             at ./src/encodings/physical/bitpack_fastlanes.rs:1687:54
   5: lance_encoding::encodings::logical::primitive::PrimitiveStructuralEncoder::encode_miniblock
             at ./src/encodings/logical/primitive.rs:1759:49
   6: lance_encoding::encodings::logical::primitive::PrimitiveStructuralEncoder::do_flush::{{closure}}
             at ./src/encodings/logical/primitive.rs:1826:21
```

Xuanwo:
I think we can just use RLE in this case.

Xuanwo:
Let's try support all ZERO for bitpacking.

PR #6629, opened 2026-04-27
fix: drop stale has_all_zeros guard so block bitpack engages on rep/def

### LINKED ISSUE #6626: write_dataset panics on a list<uint64> column with one very long row (default v2.1 encoder)
## Summary

Writing a `list<uint64>` column with one very long row (50,000 random uint64 values) and the rest empty panics the v2.1 miniblock encoder. Same data writes fine with `data_storage_version='2.0'` and `'2.2'`. Since v2.1 is the **default** on `pylance==6.0.0-beta.3`, any caller that doesn't explicitly pin a version is exposed.

This looks related to [#6184](https://github.com/lance-format/lance/issues/6184) / fixed by [#6234](https://github.com/lance-format/lance/pull/6234), but PR #6234's `repdef_too_sparse_for_miniblock` heuristic only inspects the global levels/values ratio. It doesn't catch the case where a single row's list dominates one chunk's repetition buffer, even when the global ratio looks healthy.

## Reproduction

```python
import random
import lance
import pyarrow as pa

rng = random.Random(0)
rows = [[] for _ in range(50_000)]
rows[25_000] = [rng.getrandbits(64) for _ in range(50_000)]
table = pa.table({"ids": pa.array(rows, type=pa.list_(pa.uint64()))})

lance.write_dataset(table, "/tmp/repro.lance", data_storage_version="2.1")
```

## Expected

`write_dataset` succeeds, or returns a clean Python error indicating the encoder cannot represent this data in v2.1 (so callers can fall back).

## Actual

```
thread 'lance-cpu' panicked at rust/lance-encoding/src/encodings/logical/primitive.rs:3973:13:
assertion failed: chunk_bytes <= max_chunk_size

thread '<unnamed>' panicked at rust/lance-core/src/utils/tokio.rs:126:24:
called `Result::unwrap()` on an `Err` value: RecvError(())

pyo3_runtime.PanicException: called `Result::unwrap()` on an `Err` value: RecvError(())
```

The downstream `RecvError(())` (the `tokio.rs:126` panic) is the channel receiver dying after the upstream `lance-cpu` panic kills the encoder thread; it's not a separate bug.

A closely-related variant of this trigger surfaces as the `OSError` form at `primitive.rs:3921`:

```
OSError: Encountered internal error. Definition buffer size (NN bytes) too large,
rust/lance-encoding/src/encodings/logical/primitive.rs:3921:21
```

— same root cause, hit through the def-buffer size check rather than the chunk-bytes assertion, depending on the data shape.

## Environment

- `pylance == 6.0.0-beta.3`
- macOS aarch64 and Linux x86_64
- Python 3.11

## Workarounds confirmed

| `data_storage_version` | Result |
| --- | --- |
| `"2.0"` | ✅ writes fine (no miniblock) |
| `"2.1"` | ❌ panics |
| `"2.2"` | ✅ writes fine (`support_large_chunk` → u32 chunk size from #4959) |

pengw0048:
## Summary

`write_dataset` panics on v2.1 (the current default) when a list column has a single very long row of incompressible primitives surrounded by empty lists. Fixes #6626.

The repro from the issue:

```python
import random, lance, pyarrow as pa
rng = random.Random(0)
rows = [[] for _ in range(50_000)]
rows[25_000] = [rng.getrandbits(64) for _ in range(50_000)]
table = pa.table({"ids": pa.array(rows, type=pa.list_(pa.uint64()))})
lance.write_dataset(table, "/tmp/repro.lance", data_storage_version="2.1")
```

surfaces as

```
thread 'lance-cpu' panicked at .../primitive.rs:3973:13:
assertion failed: chunk_bytes <= max_chunk_size
```

(or the sibling `Definition buffer size (NN bytes) too large` error at the same call site, depending on the data shape).

## Root cause

In this shape both rep and def levels contain a long run of zeros for the empty-list rows. `try_bitpack_for_block` rejects any buffer whose 1024-element bit-width statistic segments contain a 0, falling through to identity (raw u16) for both rep and def. The level slicer then dumps the entire run of "invisible" levels into the chunk that brackets the next visible value, blowing the per-chunk 32KiB ceiling.

A debug print in `serialize_miniblocks` confirms this:

```
[mb-overflow] chunk_bytes=106152 max=32768 log_nv=9 buf_sizes=[4096] rep_data=51024 def_data=51024 rep_levels=25512 def_levels=25512
```

(25,512 levels × 16 bits × two buffers ≈ 100 KiB just for rep+def.)

## History of `has_all_zeros`

- #3102: original report that bitpacking panics on chunks with width 0 ("Unsupported width: 0" inside fastlanes).
- #4537 added `try_bitpack_for_block` (and its sibling `try_bitpack_for_mini_block`), each guarded by `has_all_zeros` to dodge that panic.
- #4694 fixed the underlying fastlanes pack/unpack so that `width == 0` is now valid — pack writes nothing, unpack fills zeros — and removed the guard from the miniblock path. The block path was missed at that time.

## Fix

Drop the `has_all_zeros` short-circuit from `try_bitpack_for_block` so the block path matches the miniblock path (and matches what fastlanes already supports). The block bitpacker then compresses these buffers at 1 bit per value instead of 16, keeping each chunk well under the budget without changing the encoding-selection logic.

The change is intentionally minimal: I am not relaxing or expanding any of the other gates (`too_small`, `bits in {8, 16, 32, 64}`, etc.).

## Tests

- New unit test `test_block_bitpacks_with_zero_segment` confirms that a u16 buffer whose first 1024-element stat segment is all-zero now picks `OutOfLineBitpacking` from `create_block_compressor`. This mirrors the test pattern used by #4694 for the miniblock path.
- The repro from #6626 succeeds end-to-end at v2.0/2.1/2.2 (verified with a small `cargo run --example`).
- Existing list / bitpacking tests still pass (370/370), including `test_sparse_large_string_list` from #6234 and `test_miniblock_bitpack_zero_chunk_selection` from #4694.

### DIFF
--- a/rust/lance-encoding/src/compression.rs
+++ b/rust/lance-encoding/src/compression.rs
@@ -302,13 +302,12 @@ fn try_bitpack_for_block(
 
     let bit_widths = data.expect_stat(Stat::BitWidth);
     let widths = bit_widths.as_primitive::<UInt64Type>();
-    let has_all_zeros = widths.values().contains(&0);
     let max_bit_width = *widths.values().iter().max().unwrap();
 
     let too_small =
         widths.len() == 1 && InlineBitpacking::min_size_bytes(widths.value(0)) >= data.data_size();
 
-    if has_all_zeros || too_small {
+    if too_small {
         return None;
     }
 
@@ -1261,6 +1260,34 @@ mod tests {
         let compressor = strategy.create_miniblock_compressor(&field, &data).unwrap();
         // Should use RLE due to very low threshold
         assert!(format!("{:?}", compressor).contains("RleEncoder"));
+    }
+
+    // Regression for #6626: an all-zero stat segment (e.g. rep/def for a long
+    // run of empty lists) used to disable block bitpacking entirely.
+    #[test]
+    #[cfg(feature = "bitpacking")]
+    fn test_block_bitpacks_with_zero_segment() {
+        let strategy = DefaultCompressionStrategy::new();
+        let field = create_test_field("levels", DataType::UInt16);
+
+        // First 1024 zeros, then 1024 ones; max bit width is 1.
+        let mut values: Vec<u16> = vec![0; 1024];
+        values.extend(std::iter::repeat_n(1u16, 1024));
+        let mut block = FixedWidthDataBlock {
+            bits_per_value: 16,
+            data: LanceBuffer::reinterpret_vec(values),
+            num_values: 2048,
+            block_info: BlockInfo::default(),
+        };
+        block.compute_stat();
+        let data = DataBlock::FixedWidth(block);
+
+        let (compressor, _encoding) = strategy.create_block_compressor(&field, &data).unwrap();
+        let debug_str = format!("{:?}", compressor);
+        assert!(
+            debug_str.contains("OutOfLineBitpacking"),
+            "expected OutOfLineBitpacking, got: {debug_str}"
+        );
     }
 
     #[test]


Xuanwo (approved):
Thank you!

Issue #6618, opened 2026-04-26
[BUG] import lance / lancedb crashes with SIGILL on x86_64 CPUs without AVX2 (Sandy Bridge, Ivy Bridge, FX-7500-class AMDs)


tobocop2:
## What's broken

I have an Intel Xeon E5-2609 (Sandy Bridge, 2012 — has AVX and SSE4.2, no AVX2 and no FMA). The published lancedb wheel embeds lance, and `import lancedb` crashes immediately:

```
$ python -c "import lancedb"
[1]  Illegal instruction (core dumped)
```

A friend hit the same thing on an AMD FX-7500 (Steamroller, 2014 — has AVX + FMA, no AVX2). Both CPUs are pre-Haswell on the AVX2 timeline.

The cause is that the workspace `.cargo/config.toml` compiles with `target-cpu=haswell` + `target-feature=+avx2,+fma,+f16c`, which bakes AVX2 and FMA into every byte of compiled code — both the explicit SIMD kernels and any auto-vectorized loop in plain Rust. The existing runtime SIMD dispatch in `lance-core::utils::cpu::SIMD_SUPPORT` never gets a chance to run; the binary traps on its first AVX2 instruction at load time.

## Why it's worth fixing

The neighboring libraries in any data-science user's import path don't have this problem. On the same Sandy Bridge box:

| Library | What it ships | What we saw |
|---|---|---|
| pyarrow | `runtime_info.simd_level == 'avx'` | Imports cleanly at AVX tier |
| numpy | `baseline=X86_V2`, AVX2/AVX-512 listed under "not found" | Imports cleanly at V2 baseline |
| lancedb (embeds lance) | heavy AVX2 + AVX-512 instructions, no runtime guard | `Illegal instruction (core dumped)` |

A user who can `import numpy as np; import pyarrow as pa` cannot necessarily `import lancedb`. Lance is the outlier in the trio.

## Affected hardware

Anything pre-Haswell on the AVX2 timeline:

- **Intel**: Sandy Bridge (2011), Ivy Bridge (2012), Westmere (2010), Nehalem (2008)
- **AMD**: Bulldozer / Piledriver (2011-2012), Steamroller (2014, has FMA but no AVX2 — e.g. FX-7500)

Modern data-center hosts are all AVX2 or better, so this isn't blocking production. It does block lance on workstations, homelabs, older laptops, and any environment where someone is using lance alongside numpy and pyarrow expecting parity with how those libraries handle the hardware.

## The fix

[...]

PR #6630, opened 2026-04-28
feat(lance-linalg): runtime SIMD dispatch for pre-Haswell x86_64 from-source builds

tobocop2:
Tracks #6618. On x86_64 CPUs without AVX2 — Sandy Bridge / Ivy Bridge / Westmere on Intel, Bulldozer / Piledriver / Steamroller on AMD — `import lancedb` SIGILLs because the wheel bakes AVX2 + FMA into every compiled function with no runtime guard. numpy and pyarrow handle the same hardware via runtime CPU dispatch.

## Summary

- Adds 5-tier runtime SIMD dispatch (scalar / AVX / AVX+FMA / AVX2+FMA / AVX-512) to the f32/f64 hot kernels in `lance-linalg::distance::{cosine, dot, l2, norm_l2}`. Same `match *SIMD_SUPPORT` + `mod x86 { #[target_feature] pub unsafe fn ... }` shape as `dot_u8.rs` / `cosine_u8.rs` / `l2_u8.rs`. Where the AVX2 and AVX+FMA kernel bodies use no AVX2-specific intrinsics, the dispatch matches `Avx2 | AvxFma` to a shared kernel.
- Adds `lance.simd_info()` Python introspection mirroring `pyarrow.runtime_info()` so users can verify which tier the runtime selected.
- Adds a `qemu-pre-haswell` CI job that builds with `RUSTFLAGS="-C target-cpu=x86-64-v2"` (env-var-scoped to that one job — workspace `.cargo/config.toml` is unchanged) and runs `lance-linalg` lib tests under `qemu-x86_64 -cpu Nehalem`.
- Documents the legacy build path in `CONTRIBUTING.md`: `RUSTFLAGS="-C target-cpu=x86-64-v2" cargo build --release`.

Per [westonpace's review on lancedb/lancedb#3324](https://github.com/lancedb/lancedb/issues/3324#issuecomment-4328944354), the workspace baseline stays at `target-cpu=haswell`. Modern wheels are unchanged; legacy users opt into the lower baseline at build time.

## Benchmark

The AVX2 path on modern hardware is preserved as one of the per-tier kernels and the workspace baseline still bakes AVX2 into surrounding code, so by construction the modern compile is unchanged. Numbers still pending — Codespace's 30-min idle timeout killed my last full `cargo bench -p lance-linalg --bench {cosine,dot,l2,norm_l2}` run mid-suite (even with `nohup` — the VM itself sleeps). If anyone can recommend a free resource that holds a benchmark for ~1 hour, or a maintainer-preferred narrower bench shape, I'd appreciate the pointer.

Pre-Haswell verification on Sandy Bridge Xeon E5-2609 (the hardware the published wheel SIGILLs on) via the companion lancedb wheel build: pre-PR `pip install lancedb` SIGILLs at import; post-PR a from-source build with the documented `RUSTFLAGS` override produces a wheel where import + table-create + vector-search all work at the AVX tier. PASS output in [`tobocop2/lancedb#2`](https://github.com/tobocop2/lancedb/pull/2).

## Incidental fix: FMA not verified before the AVX2 tier

While auditing the dispatch table for this PR I found a latent soundness bug that predates it. On `main` the tier is chosen with `is_x86_feature_detected!("avx2")` alone, but every kernel the AVX2 tier dispatches to is `#[target_feature(enable = "avx,fma")]`. AVX2 does not imply FMA in the x86 ISA, so a host with AVX2 but no FMA would select that tier and execute `vfmadd` — an illegal instruction.

No shipping AVX2 part lacks FMA, so this is latent rather than live, but it is exactly the class of fault this PR exists to remove. The [fix](https://github.com/lance-format/lance/pull/6630/commits/bb8af46f5a8177f1ab8abfa17dd218dbc8cbf28e) checks FMA explicitly, documents the tier's contract, and adds a test that fails if any tier is ever selected on a host that cannot run its kernels. Tracked separately as #7732 so it stays on the record independent of the perf work.

## Test plan

- [x] `cargo test -p lance-linalg --lib` — 83/83 on aarch64 dev box
- [x] `cargo clippy --all-targets -- -D warnings` clean; `cargo fmt --check` clean; `Cargo.lock` unchanged
- [x] 11 proptest cases verifying scalar↔SIMD bit-for-bit equivalence per tier per kernel; gated on `is_x86_feature_detected!()` so each runs on hosts that can execute its tier
- [x] SIGILL repro confirmed gone on Sandy Bridge Xeon E5-2609 with the documented `RUSTFLAGS` override
- [ ] qemu-pre-haswell CI gate green (lights up when this PR runs CI)
- [x] Modern-hardware bench delta posted

---

To be transparent: this isn't my domain of expertise and the implementation is AI-generated — I stuck to the existing `mod x86 { #[target_feature] }` precedent and verified end-to-end on the failing hardware, but wanted to be upfront. Happy to roll in feedback.

## Summary by CodeRabbit

* **New Features**
  * Added `lance.simd_info()` (Python) to report the runtime SIMD tier, target architecture, and detected host features.
* **Bug Fixes**
  * Improved runtime SIMD dispatch and scalar fallback behavior for cosine, dot, L2 distance, and L2 norms, including more accurate tier handling across mixed AVX/AVX+FMA capability.
* **Documentation**
  * Added instructions for building on legacy x86_64 (pre-Haswell) hosts and verifying SIMD selection.
* **Tests**
  * Added `simd_info()` tests and expanded dispatched-vs-scalar parity coverage, plus a new regression seed.
* **Chores**
  * Enhanced CI with QEMU-based SIMD-tier validation on a lower x86-64 baseline.


stumpylog:
Benchmarked this PR on two modern x86_64 boxes (the "no slowdown on newer hardware" question). As-is it regresses the f32 cosine **batch** path on AVX2-only CPUs; a small refinement fixes that while keeping the AVX-512 win.

`cargo bench -p lance-linalg --bench cosine`, criterion, same-session base→PR→fix (`Cosine(f32, scalar)` is a build-independent control ≈ 0% on clean runs). `auto-vectorized` = `cosine_distance_batch` dim 1024; `simd,f32x8` = dim 8.

**PR as-is vs base:**

| | Broadwell (AVX2) | Zen 4 (AVX-512) |
|---|---|---|
| dim 1024 | **+24%** | −6% |
| dim 8 | **+36%** | **+78%** |

Cause: `f32::cosine_batch` does the `match *SIMD_SUPPORT` + a `#[target_feature]` call **per vector**. On a wheel already built with `target-cpu=haswell`, that's pure overhead vs the pre-PR inlined `f32x16`/`f32x8` kernel.

**Refinement:** gate the dispatch on `#[cfg(target_feature = "avx2")]` — AVX2-baseline builds use the inlined (byte-identical to pre-PR) kernels with no per-vector dispatch, and dispatch to AVX-512 **once per batch**, large dims only (a masked 512-bit load loses to a plain AVX2 load at 8 lanes). Sub-AVX2 builds keep your runtime path.

**After (fix vs base):**

| | Broadwell (AVX2) | Zen 4 (AVX-512) |
|---|---|---|
| dim 1024 | +6% | **−7%** |
| dim 8 | flat | −3% |

Fast by default, AVX-512 still wins where it helps.

Notes: the +6% at dim 1024 on AVX2 is a codegen-unit artifact (the kernel there is byte-identical to pre-PR), not the dispatch path. I only reworked f32 cosine — `dot`/`l2`/`norm_l2` have the same per-vector shape and likely the same regression. Branch: https://github.com/stumpylog/lance/tree/cosine-batch-dispatch-fix — happy to PR it against your branch.

(Downstream context: paperless-ngx/paperless-ngx#12970.)

tobocop2:
I'd love the PR to my branch! Thank you!!!

tobocop2:
@westonpace 

Just a friendly bump now that there are other folks with similar needs. I'm wondering if you could nudge someone on this team our way. Thank you very much.

tobocop2:
addressed all CI issues that I saw @wjones127 thank you so much

python/src/lib.rs
```diff
@@ -340,6 +341,29 @@ fn iops_counter() -> PyResult<u64> {
...
+/// Mirrors `pyarrow.runtime_info()`: a cheap, transparent way to verify that
+/// the host is hitting the expected SIMD tier (e.g., `"avx512_fp16"`,
+/// `"avx2"`) when debugging vector-search performance.
+///
+/// Returns:
+///   {
+///     "tier": str,                      # e.g. "avx2", "avx_fma", "neon", "none"
+///     "target_arch": str,               # e.g. "x86_64", "aarch64", "loongarch64"
+///     "host_features": list[str],       # raw CPU feature flags (x86_64 only)
+///   }
```
wjones127, line 363: praise: I like this API.

wjones127 (changes requested):
My initial review of the changes look good. I like the direction this is going. But before we merge this could we get updated benchmark results on modern machines (both AVX2 and AVX-512) to ensure there are no regressions?

python/src/lib.rs
tobocop2: thank you!

stumpylog:
I'm afk but I will try and rerun benchmarks when I am not.

tobocop2:
> I'm afk but I will try and rerun benchmarks when I am not.

awesome, thank you!

tobocop2:
Following up on the test-guard note in [this review](https://github.com/lance-format/lance/pull/6630#pullrequestreview-4664731504) — the one about `simd/f32.rs` guarding on `avx2` while the comments say `avx`.

You're right that the guards and the comments disagreed. I've fixed that. But I went with the other option you offered rather than relaxing the guard to `avx`, because relaxing it would crash on the exact hardware this PR is meant to support. Details below.

## Why `avx` alone isn't enough

`test_basic_ops` and `test_basic_f32x16_ops` both call `multiply_add`, and that lowers to `_mm256_fmadd_ps`.

AVX and FMA are separate CPU features. Sandy Bridge and Ivy Bridge have AVX but no FMA — which is exactly the `SimdSupport::Avx` tier this PR adds. So on one of those chips, an `avx`-only guard passes, the test proceeds, and then hits an FMA instruction the CPU doesn't have.

## Showing it rather than asserting it

I pulled `multiply_add` and the proposed guard into a standalone program, built it with the same flags the PR's pre-Haswell CI job uses (`-C target-cpu=x86-64-v2`), and ran it under `qemu-x86_64`:

```rust
#[inline(never)]
fn multiply_add(acc: __m256, a: __m256, b: __m256) -> __m256 {
    unsafe { _mm256_fmadd_ps(a, b, acc) }
}

fn main() {
    // the suggested guard
    if !is_x86_feature_detected!("avx") { println!("skipped"); return; }
    unsafe { /* ... multiply_add(acc, a, b) ... */ }
}
```

Lowering the baseline to `x86-64-v2` doesn't help, because the intrinsic emits the instruction regardless:

```
$ objdump -d repro | grep vfmadd
   113b8:  c4 e2 7d a8 09    vfmadd213ps (%rcx),%ymm0,%ymm1
```

Three CPU models, same binary:

| `qemu-x86_64 -cpu` | avx | avx2 | fma | guard | result |
|---|---|---|---|---|---|
| Nehalem | no | no | no | skips | exit 0 |
| SandyBridge | **yes** | no | **no** | **passes** | **SIGILL (signal 4)** |
| Haswell | yes | yes | yes | passes | exit 0 |

The middle row is the problem:

```
### qemu-x86_64 -cpu SandyBridge ###
host: avx=true avx2=false fma=false
GUARD: passed (avx present) -> proceeding into multiply_add
qemu: uncaught target signal 4 (Illegal instruction) - core dumped
```

## What the guards check now

Each test guards on what it actually uses:

- `avx` for the tests that only load, store, compare and reduce.
- `avx` **and** `fma` for the two that call `multiply_add`.

No AVX2 intrinsic is reachable from any of them. The one remaining `avx2` check in the file is the real runtime dispatch inside `f32x8::gather`, which falls back to a scalar gather when AVX2 is missing. This also lines up with the guards the distance kernels in `dot.rs`, `l2.rs`, `norm_l2.rs` and `cosine.rs` were already using.

## Same bug in `f64.rs`

All six guards there had the same `avx2`-instead-of-`avx` mismatch, and that file contains no AVX2 intrinsics at all. Those are corrected too.

The upshot is that AVX-only and AVX+FMA hosts now actually run these tests instead of skipping them, which is the coverage this PR was after in the first place.

tobocop2:
the review request was accidental @wjones127  my apologies. This still needs benchmarks and also it needs to address  what is discussed here https://github.com/lance-format/lance/pull/6630#issuecomment-4673952033

tobocop2:
## Benchmarks found a regression — holding this PR

@wjones127 I ran the AVX2 and AVX-512 benchmarks you asked for. They turned up a real regression in this PR, so it isn't merge-ready. Details below; I'm working the root cause now.

@stumpylog you called this in June — `dot`/`l2` had the same per-vector dispatch shape as `cosine`. No need to rerun anything, I have the hardware up and I'm taking care of it.

### The finding

`L2(simd,f32x8)` — `l2_distance_batch` at dim 8 — regresses badly against `upstream/main`:

| host | `L2(simd,f32x8)` (dim 8) | `L2(f32, simd)` (dim 1024) | `Cosine(f32, scalar)` (control) |
|---|---|---|---|
| AMD EPYC 7B13 — `avx2`, `fma`, no `avx512f` | **+156.1%** | +0.71% | −0.30% |
| Intel Xeon (Cascade Lake) — `avx512f` | **+78.7%** | −0.57% | +0.19% |

The control is flat on both hosts and dim-1024 is flat, so this is a real effect localized to small dimensions — exactly where the per-vector dispatch overhead dominates the per-vector work. Dim 8 is not academic: it's the sub-vector width PQ uses.

Two attempted fixes have not moved it, which means my causal model is wrong rather than my patch. Bisecting now.

### Method

criterion, one machine, one session, base → PR, on two GCE VMs:

- **AVX2, no AVX-512** — [`c2d-standard-16`](https://cloud.google.com/compute/docs/general-purpose-machines#c2d_machine_types) (AMD EPYC 7B13, Milan)
- **AVX-512** — [`n2-standard-16`](https://cloud.google.com/compute/docs/general-purpose-machines#n2_machines) (Intel Xeon, Cascade Lake)

Both verified with `/proc/cpuinfo` rather than trusted by instance family.

```
git checkout <base>;  cargo bench -p lance-linalg --bench {cosine,dot,l2,norm_l2} -- --save-baseline pre
git checkout <pr>;    cargo bench -p lance-linalg --bench {cosine,dot,l2,norm_l2} -- --baseline pre
```

No `RUSTFLAGS` override — `.cargo/config.toml` already pins `target-cpu=haswell` for
`x86_64-unknown-linux-gnu`, and that AVX2 baseline is the only one on which this regression exists. Overriding it measures nothing.

base `5dbd1400d` → head `97865b523`, rustc 1.94.0.

I'll follow up with the full numbers once the regression is fixed and the run is clean.

tobocop2:
## Benchmark results — no regressions on either CPU

@wjones127 here are the AVX2 and AVX-512 numbers. @stumpylog, no need to rerun anything.

Same method as [your June run](https://github.com/lance-format/lance/pull/6630#issuecomment-4673952033): criterion, one machine, one session, base → PR. No `RUSTFLAGS` override, so the repo's default `target-cpu=haswell` baseline applies — the only baseline on which any of this matters.

base `5dbd1400d` (upstream/main) → head `6b94ccdbe`, rustc 1.94.0. Both hosts 16 vCPU, features read from `/proc/cpuinfo`. **Full logs: https://gist.github.com/tobocop2/3c6d0f449cbd736aa2501f89a7fe56a2**

| benchmark | EPYC 7B13 (`avx2`, `fma`, no `avx512f`) | Xeon Cascade Lake (`avx512f`) |
|---|---|---|
| `Cosine(f32, scalar)` *(control)* | +0.04% | +0.09% |
| `Cosine(f64, scalar)` | −0.34% | −1.94% |
| `Cosine(u8, SIMD)` | +2.30% | +3.63% |
| `Dot(f16, SIMD)` | −0.58% | +0.61% |
| `Dot(f32, SIMD)` | +0.34% | **−6.08%** |
| `Dot(f32, arrow_arity)` | +0.02% | −0.00% |
| `L2(f32, scalar)` | −0.10% | −0.02% |
| `L2(f32, simd)` (dim 1024) | +2.63% | −0.53% |
| **`L2(simd,f32x8)` (dim 8)** | **−45.9%** | **−25.1%** |
| `L2(u8, SIMD)` | +0.42% | −3.11% |
| `NormL2(f32, SIMD)` | −1.02% | −4.17% |
| `NormL2(f64, SIMD)` | +3.51% | −0.58% |

Nothing regresses beyond the noise floor. Dim 8 — the PQ sub-vector width — improves 25–46%.

### On the two >2% rows

`L2(f32, simd)` and `Dot(f32, SIMD)` allocate ~4 GB each ("1M of 1024-D vectors") and are memory-bandwidth bound. Their run-to-run drift on these hosts reaches ±9%, larger than any effect this PR has on them. Rather than argue, the gist includes a null experiment: base measured against base, interleaved with base against head, three repetitions.

```
EPYC   L2(f32,simd)   null: -0.75%  +3.53%  +2.03%
                      head: -0.17%  +2.74%  +1.37%
```

The head deltas sit inside the base-vs-base spread. That is what "identical code path" looks like on a shared machine — and above dim 16 it *is* the same code path, see below.

`Dot(f32, SIMD)` calls `f32::dot` per vector rather than the batch API, so it measures the single-vector dispatch, not the batch work. On AVX-512 that resolves to the wide kernel: −6%, well outside its ±1.5% null floor.

## What changed

@stumpylog called it in June: `dot` and `l2` had the same per-vector dispatch shape as `cosine` and would regress the same way. Only `cosine` had been fixed. (`norm_l2` has no batch API, so nothing to fix there.)

Both now choose the SIMD tier once per batch instead of once per vector.

The subtlety is which kernel the batch path should then call. `dot_scalar`/`l2_scalar` chunk by 16 lanes; at or below that width the chunking degenerates to a scalar remainder loop and vectorizes nothing, so the explicit AVX+FMA kernel is worth ~40% at dim 8. Above 16 lanes the autovectorizer is already good and the 8-wide kernel can lose to it. So the wide path keeps exactly the kernel the pre-dispatch code used, which makes it non-regressing by construction rather than by measurement.

`Dot` had no batch hook, so `Dot::dot_batch` is new with a per-vector default; non-f32 impls are unchanged. `do_dot_distance_arrow_batch` carried its own copy of the per-vector loop and now routes through `dot_distance_batch`. Public signatures match pre-dispatch: `l2_distance_batch` returns `impl Iterator`, `dot_distance_batch` a boxed one.

### The benchmark caught one bug. Fixing it led me to two more.

**What the benchmark caught — a real slowdown.**

An earlier version of this change moved the dispatch out of the per-vector loop, which was the point. But it also swapped the SIMD kernel out for the plain auto-vectorized one, on the assumption that an AVX2 baseline makes the hand-written SIMD kernel redundant.

It doesn't. On FMA hardware that swap cost **+156% on `L2(simd,f32x8)`** — the benchmark caught it immediately. Taking out a cheap branch isn't a win if the kernel behind it gets slower. Fixed in `6b94ccdbe`.

**What the benchmark couldn't catch — two FMA bugs I found while fixing the first one.**

Fixing that meant actually reading the dispatch table, and that's where the other two turned up. Neither of these can ever show up in a benchmark, because no real CPU triggers them — you only get there with hand-picked build flags:

- The new batch path was gated on `cfg(target_feature = "avx2")`. But `-C target-feature=+avx2` turns on AVX2 *without* FMA, and the kernel behind that gate needs FMA. So a build at exactly `+avx2` would compile in a kernel the CPU wasn't guaranteed to support.
- The runtime tier selection had the same blind spot — it picked the AVX2 tier on `is_x86_feature_detected!("avx2")` alone, even though those kernels need FMA too. On a CPU with AVX2 but no FMA that's an illegal instruction, i.e. a crash. This one predates my PR — it's already on `main`. Fixed in [`bb8af46`](https://github.com/lance-format/lance/pull/6630/commits/bb8af46f5a8177f1ab8abfa17dd218dbc8cbf28e), written up in **#7732**.

Both now check for FMA explicitly. The compile-time gate and the runtime check are each the exact complement of the other, so there's no third path left open. Verified by building at `+avx2`, `+avx2,+fma`, `+avx`, `target-cpu=haswell`, and the default baseline.

wjones127 (commented):
This is looking better. Thanks for the benchmark results, they look good. Once clippy lints are fixed, I'm happy to approve and merge.

tobocop2:
thank you!

tobocop2:
@wjones127 clippy lints are fixed in b1670de268cf6cf67fdac8cfe3a4a6cd194dcaef

wjones127 (approved):

PR #6636, opened 2026-04-28
fix: respect LANCE_DEFAULT_IO_BUFFER_SIZE if it has been set

westonpace:
When we migrated to the filtered read path we ended up ignoring `LANCE_DEFAULT_IO_BUFFER_SIZE`.  This is because we would only check the environment variable if the scan had the io_buffer_size set.

This change adds an explicit check for the environment variable.

### DIFF
--- a/rust/lance/src/dataset/scanner.rs
+++ b/rust/lance/src/dataset/scanner.rs
@@ -167,6 +167,18 @@ pub static DEFAULT_IO_BUFFER_SIZE: LazyLock<u64> = LazyLock::new(|| {
     .unwrap_or(DEFAULT_IO_BUFFER_SIZE_VALUE)
 });
 
+/// The user-set value of `LANCE_DEFAULT_IO_BUFFER_SIZE`, or `None` if the env var
+/// is unset or unparseable. Consult this from paths that have a sensible non-fixed
+/// default (e.g. `SchedulerConfig::max_bandwidth`) so the env var still takes
+/// precedence over that default. Re-reads the env var on each call so tests can
+/// mutate it.
+pub fn get_default_io_buffer_size_override() -> Option<u64> {
+    parse_env_var(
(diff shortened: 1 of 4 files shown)


rust/lance/src/dataset/scanner.rs
```diff
@@ -10488,6 +10500,75 @@ full_filter=name LIKE Utf8(\"test%2\"), refine_filter=name LIKE Utf8(\"test%2\")
...
+        scanner.io_buffer_size(7777);
+        let plan = scanner.create_plan().await.unwrap();
+        let filtered = find_filtered_read(plan.as_ref())
+            .expect("expected a FilteredReadExec in the scan plan");
+        assert_eq!(filtered.options().io_buffer_size_bytes, Some(7777));
+    }
+
+    // The env var key scopes serial_test's lock so this test only blocks others
+    // that touch LANCE_DEFAULT_IO_BUFFER_SIZE — unrelated tests still run in
+    // parallel.
```
wjones127, line 10543: 🤔 Oh that's cool!

wjones127 (approved):

rust/lance/src/dataset/scanner.rs
```diff
@@ -167,6 +167,18 @@ pub static DEFAULT_IO_BUFFER_SIZE: LazyLock<u64> = LazyLock::new(|| {
     .unwrap_or(DEFAULT_IO_BUFFER_SIZE_VALUE)
 });
 
+/// The user-set value of `LANCE_DEFAULT_IO_BUFFER_SIZE`, or `None` if the env var
+/// is unset or unparseable. Consult this from paths that have a sensible non-fixed
```
westonpace:
```suggestion
/// is unset or unparsable. Consult this from paths that have a sensible non-fixed
```

PR #6640, opened 2026-04-29
fix: refresh last_updated metadata on Operation::Merge for rewritten fragments

### LINKED ISSUE #6519: Operation::Merge: refresh per-row version metadata for rewritten fragments
## Summary

Operation::Merge extends the manifest with merged fragments but does not apply a
full-fragment last_updated refresh for those fragments. Therefore _row_last_updated_at_version
may not advance for rewritten rows even though the dataset version has moved forward.

Operation::Merge is shared across multiple callers: Rust/Python `Dataset::add_columns`,
`Dataset::merge`, `alter_columns` (when a type cast requires a data rewrite), and
JNI/Spark ADD COLUMNS FROM. None of these are pure `Operation::Project` schema-only
commits: each supplies a new schema and a full `Operation::Merge.fragments` list for the
next version.

The core fix belongs in `transaction.rs`'s `Operation::Merge` arm so every caller shares
one implementation. `Operation::Merge.fragments` is the full post-commit fragment list;
it can include both fragments that rewrote data in this commit and fragments carried
forward unchanged. The Merge arm must refresh `_row_last_updated_at_version` only for
fragments that actually had a data rewrite, and must not bump carry-forwards.

## Expected outcome

When stable row IDs are enabled, for each merged fragment produced by a rewrite (all rows
materialized), call refresh_row_latest_update_meta_for_full_frag_rewrite_cols with the new
dataset version. Carry-forward fragments that were not rewritten in this commit must not be
refreshed. Follow deletion-vector rules consistent with merge_insert and compaction.

## Required by:
This Lance issue is a prerequisite for https://github.com/lance-format/lance-spark/issues/419

## Acceptance criteria

- All visible rows in a rewritten merged fragment: _row_last_updated_at_version equals the
  new dataset version; _row_created_at_version unchanged.
- Rows in carry-forward fragments that were not rewritten in this commit: both version
  columns unchanged.

jerryjch:
## Summary

* Fixes https://github.com/lance-format/lance/issues/6519
* In `build_manifest`, when stable row IDs are enabled (`next_row_id.is_some()`), the
  `Operation::Merge` arm now bumps `last_updated_at_version_meta` for fragments whose
  on-disk data changed versus the previous manifest, using
  `refresh_row_latest_update_meta_for_full_frag_rewrite_cols(fragment, new_version)`.
* Rewrite vs carry-forward: refresh only when `Fragment::files` differs from the prior
  manifest for the same fragment id (column data rewritten); skip when files match
  (metadata-only / carry-forward). Deletion-file-only changes do not trigger refresh.
* New fragment ids with no prior manifest row still get a full-fragment refresh so
  `last_updated` is not left unset.
* Concurrent-transaction safety: `build_manifest` is called fresh on every commit retry                                   
  with the latest committed manifest, so `new_version` is always consistent with the actual                               
  committed version regardless of intervening transactions.

## Root cause

`Operation::Merge` previously appended merged fragments without updating per-row
`last_updated` metadata for physical rewrites. Dataset version advanced, but
`_row_last_updated_at_version` could stay stale for rewritten rows, breaking CDF semantics
for merges that commit through `Operation::Merge` (e.g. ADD COLUMNS-style materialization and
other callers sharing this arm).

## Test plan

* `cargo test -p lance merge_build_manifest` — covers files change -> last_updated -> new
  version, same files -> unchanged, and stable row IDs disabled -> no last_updated metadata.
* Existing `lance` CI / regression suites unchanged aside from these tests.

### DIFF
--- a/rust/lance/src/dataset/transaction.rs
+++ b/rust/lance/src/dataset/transaction.rs
@@ -2011,7 +2011,36 @@ impl Transaction {
                 final_fragments.extend(maybe_existing_fragments?.clone());
             }
             Operation::Merge { fragments, .. } => {
-                final_fragments.extend(fragments.clone());
+                let mut merged_fragments = fragments.clone();
+                if next_row_id.is_some() {
+                    let new_version = current_manifest.map(|m| m.version + 1).unwrap_or(1);
+                    let prev_by_id: HashMap<u64, &Fragment> = maybe_existing_fragments?
+                        .iter()
+                        .map(|f| (f.id, f))
+                        .collect();
+                    for fragment in merged_fragments.iter_mut() {
+                        match prev_by_id.get(&fragment.id) {
+                            Some(prev) => {
+                                if merge_fragment_physically_rewritten(prev, fragment) {
+                                    lance_table::rowids::version::refresh_row_latest_update_meta_for_full_frag_rewrite_cols(
+                                        fragment,
+                                        new_version,
+                                    )?;
+                                }
+                            }
+                            None => {
+                                // Fragment id not present in the previous manifest (Merge may append
+                                // ids beyond the original list). Apply the same full-fragment
+                                // last_updated refresh as for a rewrite.
+                                lance_table::rowids::version::refresh_row_latest_update_meta_for_full_frag_rewrite_cols(
+                                    fragment,
+                                    new_version,
+                                )?;
+                            }
+                        }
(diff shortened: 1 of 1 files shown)


jackye1995:
@claude review

jackye1995 (approved):
thanks for the fix!

PR #6644, opened 2026-04-29
perf: cache per-fragment reader templates

westonpace:
## Summary

Cache per-fragment reader state in the dataset metadata cache so that the hot `FileFragment::open` path can skip the projection-independent setup work — `data_file.schema(...)`, `Schema::intersection_ignore_types`, the field-id→column-idx `BTreeMap`, `ReaderProjection::from_whole_schema`, and `validate_projection` — and reuse a fully-built `FileReader` across opens.

## Motivation

Opening a `FragmentReader` had non-trivial overhead even when file metadata was already in the cache: every call rebuilt the per-data-file schema intersection, the field id map, and the `FileReader` (which itself ran projection setup + validation). For workloads that open the same fragment repeatedly (take queries, repeated scans), this work was a measurable cost.

## What's cached

A `FragmentReaderTemplate` keyed on `(manifest_version, fragment_id)` and stored in the dataset's metadata cache:

- **Bundle:** `deletion_vec`, `row_id_sequence`, `num_rows`, `num_physical_rows`.
- **Per v2 data file (`CachedDataFile`):**
  - `Arc<FileReader>` whose scheduler slot is a `SentinelEncodingsIo` (panics if read; never used directly — always rebound).
  - The pre-computed `data_file.schema(dataset_schema)` (per-call we only intersect with the user's projection).
  - The `field_id → column_idx` `BTreeMap`.

## What's not cached, and why

Object stores, `ScanScheduler`s, and open file handles. These are per-call. The split exists because two correctness invariants depend on it:

- **`test_with_object_store_enables_isolated_per_request_io_tracking`** — `with_object_store` wrappers create per-request datasets that share a session/cache but use different object stores. A cached reader bound to one store would route subsequent requests through the wrong wrapper.
- **`test_scan_finishes_all_tasks`** — when a scan is dropped, its `ScanScheduler` must drop too so background I/O tasks observe the signal. Caching a reader that holds the scheduler keeps it alive indefinitely.

v1 (legacy) data files are not cacheable — `PreviousFileReader` owns its `Arc<dyn Reader>` open handle directly with no scheduler-swap method. They fall through to the unchanged per-call open path.

## Design

`FileFragment::open` does the following:

1. For each v2 data file, opens **one** `FileScheduler` against the resolved `ScanScheduler` (per-call binding).
2. Cache get-or-insert the template. On miss, the loader uses the schedulers we just opened to read file metadata — so cache-miss IOPs match the pre-cache path. Verified by `test_iops_read_small` (still asserts a single GET).
3. For each cached data file, rebinds via `cached.file_reader.with_scheduler(LanceEncodingsIo::new(file_scheduler))` and applies per-call `FileReaderOptions` (`with_options`) when `read_config.file_reader_options` is set.
4. Constructs the `v2_adapter::Reader` and assembles the `FragmentReader`.

`manifest_version` in the key means commits that mutate a fragment (e.g. `delete()`) automatically invalidate cached templates with no explicit cache invalidation needed.

## New `lance-file` API

- `SentinelEncodingsIo` — placeholder `EncodingsIo` whose `submit_request` panics. Used as the scheduler slot in unbound cached `FileReader`s.
- `FileReader::try_open_unbound_with_file_metadata` — constructs a `FileReader` with the sentinel scheduler. Caller must call `with_scheduler` before any read; documented contract, panic message tells you what's wrong if you forget.
- `FileReader::with_options` — companion to `with_scheduler` for swapping `FileReaderOptions` (clone-and-replace).
- `impl DeepSizeOf for FileReader` — counts only `metadata` (the bulk of the footprint). The `cache: Arc<LanceCache>` field is excluded because it's a back-reference to the cache the reader is being inserted into; including it would inflate every entry's size with the whole cache and overflow `usize` after a few hundred fragments.

## Test plan

- [x] `cargo test -p lance --lib dataset::fragment::` — 49 passed. Includes new `test_fragment_reader_template_is_cached` (parameterized over `Legacy` and `Stable` storage versions) which verifies:
    - Cold cache → entry absent.
    - First `open()` → entry present; v2 templates have a `CachedDataFile` per data file, v1 templates have `None` per data file.
    - Subsequent opens with different projections / flags hit the cache (template is projection- and flag-independent).
    - Post-`delete()` re-open does not see a stale template (key includes `manifest_version`).
- [x] `cargo test -p lance --lib dataset::` — 1083 passed. Confirms:
    - `test_with_object_store_enables_isolated_per_request_io_tracking` — per-request store wrappers stay isolated.
    - `test_scan_finishes_all_tasks` — abandoned scans still drop their schedulers and clean up tasks.
    - `test_defer_index_remap_large_external_file` — wide-fragment compaction (caught a `DeepSizeOf` back-reference bug during development; fixed before merge).
    - `test_binary_copy_compaction_with_complex_schema` — complex-schema compaction.
    - `test_iops_read_small` — single GET for small-file open + read.
- [x] `cargo test -p lance --lib io::exec::` — 70 passed.
- [x] `cargo clippy -p lance -p lance-file --all --tests --benches -- -D warnings` — clean.
- [x] `cargo fmt -p lance -p lance-file --check` — clean.

### DIFF
--- a/rust/lance-file/src/io.rs
+++ b/rust/lance-file/src/io.rs
@@ -9,6 +9,37 @@ use lance_io::scheduler::FileScheduler;
 
 use super::reader::DEFAULT_READ_CHUNK_SIZE;
 
+/// A placeholder [`EncodingsIo`] used by cached / unbound [`crate::reader::FileReader`]
+/// instances.
+///
+/// A `FileReader` opened via [`crate::reader::FileReader::try_open_unbound_with_file_metadata`]
+/// holds this sentinel in its scheduler slot. The caller MUST call
+/// [`crate::reader::FileReader::with_scheduler`] to bind a real scheduler before
+/// performing any reads — invoking I/O against the sentinel panics.
+///
+/// This exists to let callers cache an immutable, fully-projected `FileReader`
+/// across requests without baking in per-request object stores or open file
+/// handles. Each request rebinds the cached reader to its own scheduler.
+#[derive(Debug)]
+pub struct SentinelEncodingsIo;
+
+impl EncodingsIo for SentinelEncodingsIo {
+    fn with_bypass_backpressure(&self) -> Option<Arc<dyn EncodingsIo>> {
+        None
+    }
+
+    fn submit_request(
+        &self,
+        _ranges: Vec<std::ops::Range<u64>>,
+        _priority: u64,
+    ) -> BoxFuture<'static, lance_core::Result<Vec<bytes::Bytes>>> {
+        panic!(
+            "SentinelEncodingsIo::submit_request called — this FileReader was opened \
+             without a scheduler; bind one with FileReader::with_scheduler before reading"
+        );
+    }
+}
+
 #[derive(Debug)]
 pub struct LanceEncodingsIo {
     scheduler: FileScheduler,
--- a/rust/lance-file/src/lib.rs
+++ b/rust/lance-file/src/lib.rs
@@ -3,7 +3,7 @@
 
 pub mod datatypes;
 pub mod format;
-pub(crate) mod io;
+pub mod io;
 pub mod reader;
 pub mod testing;
 pub mod version;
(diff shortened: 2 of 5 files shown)


westonpace:
@lance-a-doodler test

PR #6647, opened 2026-04-29
feat!: make dataset object store access base-aware

jackye1995:
Make dataset object-store access explicit for multi-base datasets. The public `Dataset::object_store` accessor now takes an `Option<u32>` base id: pass `None` for the primary dataset store, or `Some(base_id)` for an additional base referenced by dataset metadata. This is a breaking change from the previous zero-argument accessor.

Adds `Dataset::with_object_store_wrappers(...)` to clone datasets with wrappers propagated to the primary object store, refs store, primary store params, and every base store params entry. When a fragment, deletion file, or index references a base id, Lance resolves both the correct path location and the corresponding wrapped object store, so wrappers such as caching and instrumentation apply to multi-base reads without changing page cache keys.

Also fixes base-aware read paths that resolved a `base_id` path but still used the primary dataset object store. Data-file metadata creation, deletion file reads, scalar/vector index opens, index detail inference, index stats, and index migration now resolve the correct base object store before reading.

### DIFF
--- a/java/lance-jni/src/blocking_dataset.rs
+++ b/java/lance-jni/src/blocking_dataset.rs
@@ -3455,6 +3455,7 @@ fn inner_get_zonemap_stats<'local>(
                     for index in &indices {
                         let index_store = Arc::new(
                             LanceIndexStore::from_dataset_for_existing(&dataset, index)
+                                .await
                                 .map_err(Error::from)?,
                         );
                         let index_file = index_store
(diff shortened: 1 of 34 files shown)


cmccabe:
I'm concerned that every use of Dataset.object_store could be a potential bug, since that only represents the primary object store and not all object stores.

Should we do a change where we turn that field into a `Map<base_index, Arc<ObjectStore>>` ?

cmccabe (approved):

LuQQiu (commented):
Multi-base table is a newly introduce feature, kind of don't want it to break all existing users who is using single base table. Also agree with what @cmccabe mentioned, we don't want it to be sliently incorrect.
Would prefer to keep dataset.object_store() but error out if the dataset is multibase dataset

LuQQiu (approved):

PR #6648, opened 2026-04-30
feat: expose FTS exec internals to enable distributed planning

vivek-bharathan:
The FTS execution plan types (`MatchQueryExec`, `PhraseQueryExec`, `BoostQueryExec`,
  `BooleanQueryExec`, `FlatMatchQueryExec`, `FlatMatchFilterExec`) and their supporting
  helpers (`load_segments`, `load_segment_details`, `build_global_bm25_scorer`) are
  currently private or `pub(crate)`, with fields hidden behind constructors that always
  assume that all committed segments exist on one node and are scored with statistics computed
 locally.

  This doesn't work for systems that distribute FTS queries across hosts 
  — a coordinator that wants to (for example) route segments 1–5 to host A,
  segments 6–10 to host B, and still produce globally-correct BM25 scores can't do so
  today: per-host execs each compute IDFs against their local segment subset, producing
  locally-correct but globally-wrong scores.

  This PR exposes the surface needed for that pattern, additively, without changing any
  existing behavior

### DIFF
--- a/rust/lance-index/src/scalar/inverted.rs
+++ b/rust/lance-index/src/scalar/inverted.rs
@@ -12,6 +12,7 @@ mod scorer;
 pub mod tokenizer;
 mod wand;
 
+use std::collections::{HashMap, HashSet};
 use std::sync::Arc;
 
 use arrow_schema::{DataType, Field};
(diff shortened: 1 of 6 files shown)


rust/lance-index/src/scalar/inverted.rs
```diff
@@ -24,6 +25,82 @@ pub use lance_tokenizer::Language;
...
+}
+
+/// Build a shared [`MemBM25Scorer`] across a set of FTS index segments.
+///
+/// Aggregates each segment's `(total_tokens, num_docs, per_term_doc_freq)`
+/// statistics — obtained via [`InvertedIndex::bm25_stats_for_terms`] — into a
+/// single corpus-wide scorer, so that BM25 IDF scoring uses *global*
+/// statistics rather than per-segment statistics. Computes the union of
+/// fuzzy-expanded terms when `params.fuzziness` is set.
+pub fn build_global_bm25_scorer(
```
westonpace, line 80: Why does this method need to be made public?  Is it to supply a `MemBM25Scorer` to some of the exec nodes?

westonpace:
@claude review

rust/lance-index/src/scalar/inverted.rs
```diff
@@ -24,6 +25,82 @@ pub use lance_tokenizer::Language;
...
+        num_docs += segment_num_docs;
+        for (term, count) in terms.iter().zip(segment_token_docs.into_iter()) {
+            *token_docs
+                .get_mut(term)
+                .expect("global scorer terms should already be initialized") += count;
+        }
+    }
+
+    Ok(MemBM25Scorer::new(total_tokens, num_docs, token_docs))
+}
```
wkalt, line 109: I think @wjones127 suggested we already have support for this in order to handle scoring over multiple delta indexes. Did that end up not being the case?

rust/lance/src/io/exec/fts.rs
```diff
@@ -295,15 +251,79 @@ impl MatchQueryExec {
...
+            preset_segments: Some(segments),
+            properties,
+            metrics: ExecutionPlanMetricsSet::new(),
+        }
+    }
+
+    /// Override the BM25 scorer used by `execute()`. When set, the local
+    /// `build_global_bm25_scorer` call is skipped and the supplied scorer is
+    /// threaded down to `InvertedIndex::bm25_search`.
+    pub fn with_base_scorer(mut self, scorer: Arc<MemBM25Scorer>) -> Self {
```
westonpace, line 314: Can we document why a user might do this?  Is it because they want to avoid the cost of recomputing the scorer if it's already been computed?  Or is it because the user may have a more accurate or more complete scorer?

westonpace (changes requested):
Just a few doc questions.  It might also be nice (in this PR or in a separate issue) to document why we are making this change?

My blocking request is that we fill out the PR description with some kind of justification for why this change is desired?  It looks like we are trying to make fts exec nodes a more complete part of the public API?  This is a fine rationalization but we should at least describe it.

rust/lance-index/src/scalar/inverted.rs
westonpace: We already have this concept.  I think maybe it is being dragged into a method here to allow users to do something like "search segments 1-5 on machine A, search segments 6-10 on machine B, build a scorer on segments 1-10 on machine C".  So we need a public "build a scorer" method.

That being said, I'm not certain, and we should document the rationale / why this method is used as part of the docs for the method.

westonpace: (you can see this method is removed from fts.rs)

wkalt: thanks, missed that

rust/lance-index/src/scalar/inverted.rs
vivek-bharathan: Actually this is mostly for api completeness since it is paired with with_base_scorer. Keeps a single source of truth for BM25 IDF arithmetic across segments. I could move it back if you prefer

westonpace (approved):
Switching to approve as there is now a PR description.  Thanks 😄

PR #6649, opened 2026-04-30
fix: make vector distance schema nullable

BubbleCal:
## Bug Fix

- What is the bug?
  - Vector search result schemas disagreed on the `_distance` column nullability. The public scanner and KNN schemas already expose `_distance` as nullable, while `VECTOR_RESULT_SCHEMA` and a mem-wal vector path still marked it as non-nullable.
- What issues or incorrect behavior does the bug cause?
  - The mismatch can create inconsistent schemas across vector search execution paths and tests, even though callers should see a single nullable `_distance` contract.
- How does this PR fix the problem?
  - Updates the remaining vector result schema definitions and related test fixtures to mark `_distance` as `Float32` nullable.
  - Updates the `VectorIndex` schema documentation to match the nullable contract.

## Validation

- `rg` check for remaining explicit `_distance` `Float32` `nullable=false` definitions under `rust/lance` and `rust/lance-index`
- `cargo fmt --all --check`
- `cargo test -p lance --lib dataset::mem_wal::scanner::exec::filter_stale::tests:: --no-fail-fast`
- `git diff --check`

### DIFF
--- a/rust/lance-index/src/vector.rs
+++ b/rust/lance-index/src/vector.rs
@@ -67,7 +67,7 @@ pub trait PartitionSearchControl: Send + Sync {
 
 pub static VECTOR_RESULT_SCHEMA: LazyLock<arrow_schema::SchemaRef> = LazyLock::new(|| {
     arrow_schema::SchemaRef::new(arrow_schema::Schema::new(vec![
-        Field::new(DIST_COL, arrow_schema::DataType::Float32, false),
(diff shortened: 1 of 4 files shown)


wjones127 (approved):

justinrmiller:
lgtm

justinrmiller (approved):

PR #6650, opened 2026-04-30
fix: propagate update_columns offsets and partial last_updated for RewriteColumns

### LINKED ISSUE #6505: Operation::Update with RewriteColumns: refresh per-row version metadata for matched rows
## Summary

Operation::Update with RewriteColumns commits a new dataset version when columns are
rewritten in place, but `_row_last_updated_at_version` does not advance selectively for
rows that actually changed. Two gaps: (1) `Fragment::update_columns` does not report which
physical row offsets matched the update stream, and `Operation::Update` has no optional
per-fragment row offset map for commit. (2) `build_manifest` does not call
`refresh_row_latest_update_meta_for_partial_frag_rewrite_cols` for those offsets when
stable row IDs are enabled.

## Required by

This Lance issue is a prerequisite for [lance-spark#418](https://github.com/lance-format/lance-spark/issues/418)

## Acceptance criteria

- Matched rows: `_row_last_updated_at_version` equals the new dataset version; unmatched
  rows in the same fragment unchanged.
- Fragments with no offsets entry in the operation are not partially refreshed (guard).
- Matched-row offsets use the same local physical row indices as per-row version metadata
  (`RowDatasetVersionSequence`); the API docs spell out the rules (range, ordering,
  deletion vectors).
- Java/JNI exposes offsets for lance-spark and other JVM commit paths.

jerryjch:
## Summary

* Fixes https://github.com/lance-format/lance/issues/6505
* `FileFragment::update_columns` returns `Result<(Fragment, Vec<u32>)>` (unchanged
  public shape). `update_columns_with_offsets` returns `FragmentUpdateColumnsResult`
  (fragment, `fields_modified`, `matched_offsets: RoaringBitmap`) for callers that need
  physical row indices for stable row-id metadata.
* `HashJoiner::matched_join_rows` — boolean mask for hash hits; used by
  `update_columns_with_offsets` and covered by `test_matched_join_rows`.
* `Operation::Update`: optional `updated_fragment_offsets: Option<UpdatedFragmentOffsets>`
  where `UpdatedFragmentOffsets` wraps `HashMap<u64, RoaringBitmap>` (newtype with
  `Default`, `PartialEq`, manual `DeepSizeOf`). `None` means the caller did not supply
  offsets.
* Proto (`transaction.proto`): backward-compatible `map<uint64, UInt32List> updated_fragment_offsets = 9`
  on `Update`; serde round-trip preserves semantics.
* `build_manifest`: when stable row IDs are enabled, `update_mode == RewriteColumns`,
  and `Some(UpdatedFragmentOffsets(..))` includes a non-empty bitmap for a fragment,
  calls `refresh_row_latest_update_meta_for_partial_frag_rewrite_cols` for those offsets
  only — unmatched rows and untouched fragments are left unchanged.
* JNI / Java: `FragmentUpdateResult` includes matched row offsets; the 2-arg constructor
  `(FragmentMetadata, long[])` delegates to the 3-arg form with an empty offset array for
  compatibility. JNI uses `update_columns_with_offsets`.
* Python: `update_columns` binding correctly destructures the `(Fragment, Vec<u32>)` tuple.

## Root cause

For `Operation::Update` with `RewriteColumns`, commits could advance the dataset version
without advancing `_row_last_updated_at_version` for the rows that were actually rewritten.
`update_columns` did not report which physical offsets matched, and `build_manifest` had no
per-fragment offset map to drive the partial refresh. Without that information the transaction
layer cannot distinguish which rows changed, so the version metadata is not updated.

## Implementation notes

* `RoaringBitmap` iteration is ascending and duplicate-free; redundant `sort` / `dedup`
  when building proto lists or offset vectors from bitmaps were removed.
* Call sites that do not populate offsets use `updated_fragment_offsets: None`.

## Why the protobuf field exists

lance-spark passes `Transaction` through JNI as a protobuf blob: Java builds a `Transaction`
proto, Rust deserializes it and runs `build_manifest`. Without `updated_fragment_offsets` on
the wire, the decoded `Operation::Update` would always have `updated_fragment_offsets: None`
even when matched offsets were computed on the JVM side, and the partial refresh in
`build_manifest` would silently do nothing.

## Test plan

* `cargo test -p lance test_matched_join_rows` — `HashJoiner::matched_join_rows`.
* `cargo test -p lance test_build_manifest_partial_last_updated_rewrite_columns_stable_row_ids`
  — `Dataset::commit` -> `build_manifest`: two fragments, partial
  `update_columns_with_offsets`, `Operation::Update` with `RewriteColumns` and an offset
  map; asserts matched vs unmatched vs untouched row version metadata.
* `cargo test -p lance test_fragment_update` — fragment path with `Operation::Update` and
  offsets.
* `cargo test -p lance --tests` (or at least `cargo check -p lance --tests`) and
  `cargo check --manifest-path java/lance-jni/Cargo.toml`.

The `pylance` crate is excluded from the root workspace; validate Python bindings in the
usual `maturin` / CI flow if you touch `python/`.

## Compatibility

* Rust: `update_columns` signature unchanged; `update_columns_with_offsets` is additive.
* Java: 2-arg `FragmentUpdateResult` constructor preserved.
* Proto: field 9; older clients ignore unknown fields.

### DIFF
--- a/java/lance-jni/src/fragment.rs
+++ b/java/lance-jni/src/fragment.rs
@@ -48,6 +48,8 @@ pub(crate) struct FragmentMergeResult {
 pub(crate) struct FragmentUpdateResult {
     updated_fragment: Fragment,
     fields_modified: Vec<u32>,
+    /// Physical row offsets that received column updates (from `_rowaddr` low bits).
+    updated_row_offsets: Vec<i64>,
 }
 
 //////////////////
@@ -490,11 +492,13 @@ fn inner_update_column<'local>(
     let reader = unsafe { ArrowArrayStreamReader::from_raw(stream_ptr) }?;
     let left_on_str: String = left_on.extract(env)?;
     let right_on_str: String = right_on.extract(env)?;
-    let (updated_fragment, fields_modified) =
-        RT.block_on(fragment.update_columns(reader, &left_on_str, &right_on_str))?;
+    let r =
+        RT.block_on(fragment.update_columns_with_offsets(reader, &left_on_str, &right_on_str))?;
+    let updated_row_offsets: Vec<i64> = r.matched_offsets.iter().map(|o| o as i64).collect();
     let result = FragmentUpdateResult {
-        updated_fragment,
-        fields_modified,
+        updated_fragment: r.fragment,
+        fields_modified: r.fields_modified,
+        updated_row_offsets,
     };
     result.into_java(env)
 }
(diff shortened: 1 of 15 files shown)


jackye1995:
@claude review

jackye1995 (approved):
looks good to me!

PR #6653, opened 2026-04-30
fix(optimize): fix stale OCC read_version in distributed compaction to prevent row resurrection

xiaguanglei:
## Problem

In distributed compaction (e.g. Spark's `OptimizeExec`), the commit phase opens a **fresh `Dataset` handle** at the latest version `V+N`, which is newer than version `V` at which tasks were planned and executed.

`commit_compaction` was using `dataset.manifest.version` (= `V+N`) as the `TransactionBuilder` `read_version`. Lance uses **Optimistic Concurrency Control (OCC)**: when committing a transaction, the conflict checker loads all transactions committed after `read_version` and checks whether any of them touch the same fragments. Using `V+N` as `read_version` narrows the OCC scan window to `(V+N, ∞)`, **silently skipping any concurrent `DELETE`/`UPDATE` that committed between `V` and `V+N`**. The compacted fragment is written without those deletion markers, causing **deleted rows to reappear** (data resurrection).

```
Timeline (bug):

  V     App A plan_compaction (read_version = V)
  │     App A executors rewriting fragments ...
  V+1   App B DELETE committed  ← OCC window starts at V+N, this is INVISIBLE
  │     App A executors finish
  V+N   App A opens fresh Dataset → dataset.manifest.version = V+N
        commit_compaction(read_version = V+N)
        OCC scans (V+N, ∞) → finds nothing → no conflict
  V+N+1 App A commits Rewrite → compacted fragment has NO deletion markers
        ❌ deleted rows resurrected
```

## Root Cause

```rust
// Before – uses the version of the Dataset handle passed to commit_compaction,
// which in distributed mode is V+N (the latest version), not V (plan version)
let transaction = TransactionBuilder::new(
    dataset.manifest.version,   // V+N  ← wrong
    Operation::Rewrite { ... },
)
```

## Fix

Each `RewriteResult` already carries `read_version = V` (the version at which the executor ran the task). Derive the transaction `read_version` from `min(task.read_version)` so the OCC window is anchored at `V`, covering the full range `[V, ∞)`.

```rust
// After – anchor OCC window to the version tasks were planned/executed at
let tasks_read_version = completed_tasks
    .iter()
    .map(|t| t.read_version)
    .min()
    .unwrap_or(dataset.manifest.version);

let transaction = TransactionBuilder::new(
    tasks_read_version,   // V  ← correct
    Operation::Rewrite { ... },
)
```

```
Timeline (fixed):

  V     App A plan_compaction (read_version = V)
  │     App A executors rewriting fragments ...
  V+1   App B DELETE committed
  │     App A executors finish
  V+N   App A opens fresh Dataset → dataset.manifest.version = V+N
        commit_compaction(read_version = min(task.read_version) = V)
        OCC scans (V, ∞) → finds DELETE at V+1 → conflict detected
        ✅ OPTIMIZE retries or fails, no data resurrection
```

### 1. Rust unit test (added in this PR)

`rust/lance/src/dataset/optimize.rs` — `test_distributed_compact_concurrent_delete_no_resurrection`

Simulates the exact distributed scenario in a single-process test:

```
Step 1  Write 4 fragments × 1 000 rows (version V=1)
Step 2  plan_compaction at V=1 → 1 CompactionTask (read_version=1)
Step 3  Execute task against dataset@V=1 → RewriteResult(read_version=1)
Step 4  Concurrent DELETE "a < 1000" → version V=2
Step 5  Open fresh Dataset → version=2 (simulates OptimizeExec's commit Dataset)
Step 6  commit_compaction(fresh_dataset@V=2, results[read_version=1])
        BEFORE fix → Ok(…), rows resurrected
        AFTER  fix → Err(retryable conflict), data integrity preserved
Step 7  Assert final row count WHERE a < 1000 == 0
```

Run locally:
```bash
cargo test -p lance --lib \
  dataset::optimize::tests::test_distributed_compact_concurrent_delete_no_resurrection
```

### 2. End-to-end reproduction on a real Spark YARN cluster

Two independent `spark-submit` applications against a 10 M-row, 32-fragment Lance table on YARN:

| App | Config | Role |
|---|---|---|
| App A | 1 executor, `OPTIMIZE TABLE` | Plan at V=34, executor phase ~110 s |
| App B | 4 executors, `DELETE WHERE id % 312500 < 100` | Commits in ~25 s (within App A's executor phase) |

**Result without fix:**

```
App A (OPTIMIZE)  final status: SUCCEEDED  (committed at V+1, skipped DELETE)
App B (DELETE)    final status: SUCCEEDED  (committed at V+1 first)

SELECT COUNT(*) WHERE id % 312500 < 100
→ 3200  (expected 0)
❌ 3200 deleted rows resurrected
```

**Result with fix:**

```
App A (OPTIMIZE)  → Err: retryable commit conflict (OCC detected DELETE at V+1)
App B (DELETE)    final status: SUCCEEDED

SELECT COUNT(*) WHERE id % 312500 < 100
→ 0
✅ no data resurrection
```

## Notes

- The bug **only manifests in the distributed compaction path** where the caller opens a fresh `Dataset` for the commit phase (e.g. `OptimizeExec.scala` in lance-spark). Single-process `compact_files` is unaffected because the same `Dataset` handle is used throughout.
- `RewriteResult.read_version` was already populated correctly by each task; this fix simply threads it through to `TransactionBuilder`.
- End-to-end validation was performed on a production YARN cluster with the patched JAR: two independent Spark applications (OPTIMIZE + concurrent DELETE) confirmed that after the fix, OPTIMIZE correctly receives a `Retryable commit conflict` error instead of silently resurrecting deleted rows.

### DIFF
--- a/rust/lance/src/dataset/optimize.rs
+++ b/rust/lance/src/dataset/optimize.rs
@@ -1489,6 +1489,25 @@ pub async fn commit_compaction(
     // If we aren't using stable row ids, then we need to remap indices.
     let needs_remapping = !dataset.manifest.uses_stable_row_ids() && !options.defer_index_remap;
 
+    // Determine the earliest version at which compaction tasks were planned/executed.
+    //
+    // In distributed mode (e.g. Spark) the caller opens *two separate* Dataset
+    // handles: one for `plan_compaction` (at version V) and a fresh one for
+    // `commit_compaction` (at the latest version V+N).  Using `dataset.manifest.version`
+    // (= V+N) as the transaction's `read_version` would cause the conflict checker to
+    // scan only versions after V+N — finding nothing — and therefore silently skip any
+    // concurrent DELETE/UPDATE that landed between V and V+N, resurrecting deleted rows.
+    //
+    // By anchoring `read_version` to the minimum version carried in the RewriteResults
+    // we ensure the conflict checker covers the full range [V, V+N] and will reject the
+    // commit with a retryable conflict error if a concurrent write touched the same
+    // fragments.
+    let tasks_read_version = completed_tasks
+        .iter()
+        .map(|t| t.read_version)
+        .min()
+        .unwrap_or(dataset.manifest.version);
+
     let mut completed_tasks = completed_tasks;
 
     // Single reserve_fragment_ids for all address-style tasks
@@ -1587,7 +1606,14 @@ pub async fn commit_compaction(
     };
 
     let transaction = TransactionBuilder::new(
-        dataset.manifest.version,
+        // Use the version at which the compaction tasks were *planned*, not the
+        // version of the dataset handle passed to this function.  In distributed
+        // mode the caller may open a fresh dataset at a later version (V+N), but
+        // the tasks were executed against an older snapshot (V).  Anchoring the
+        // transaction to V ensures the OCC conflict checker scans all writes that
+        // landed between V and the commit point, detecting concurrent DELETE
+        // transactions that would otherwise cause deleted rows to reappear.
+        tasks_read_version,
         Operation::Rewrite {
             groups: rewrite_groups,
             rewritten_indices,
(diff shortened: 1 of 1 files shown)


xiaguanglei:
@hamersaw @wjones127 @Xuanwo  Hi ，could you take a look when you have a moment?

This fixes a silent data-corruption bug in the distributed compaction path where deleted rows reappear after `OPTIMIZE`. The root cause is that `commit_compaction` was using `dataset.manifest.version` (the version of the fresh Dataset opened for the commit phase, = V+N) as the `TransactionBuilder` `read_version`, causing the OCC conflict checker to skip any concurrent `DELETE`/`UPDATE` that committed between the plan version V and V+N.

The fix anchors `read_version` to `min(task.read_version)` (= V) so the OCC window correctly covers [V, ∞).

Validated end-to-end on a production YARN cluster: with the fix, OPTIMIZE correctly receives a `Retryable commit conflict` instead of silently committing over the concurrent DELETE.

xiaguanglei:
CI update: rebased onto latest main and fixed `cargo fmt` issues (long lines in the new test). All checks should pass now.

Also noting that the `test_write_hf_dataset` failure in the previous run was pre-existing and unrelated to this PR — it has since been fixed upstream in #6697 and #6700, which are now included after the rebase.

hamersaw (approved):
Thanks for the fix! I think this is a reasonable defense, but wanted to note that this works because the `RewriteResults` maintain the dataset version. There are many other operations in the Spark connector that have this same pattern (ie. read dataset, perform distributed operations, commit) that just maintain their own read version, or even the same Dataset handle.

So this was a problem because it seems we're using a new dataset handle in Spark, it could similar be solved by using the same handle.

Xuanwo (approved):

PR #6654, opened 2026-04-30
feat: opt-in env var to drop centroids from vector index stats

westonpace:
## Summary

Vector index `statistics()` serializes every centroid as JSON. For very large IVF indexes this can use a significant amount of memory and risks OOMing callers that just want metadata. We can't drop centroids outright without breaking existing consumers (e.g. the Python `LanceIndex.centroids()` helper), so this PR introduces an opt-in env var and a deprecation path:

- `LANCE_INCLUDE_VECTOR_CENTROIDS=false` (or `0`, case-insensitive): the stats method skips reading centroids entirely; the `centroids` field is omitted from the JSON (not serialized as `null`).
- `LANCE_INCLUDE_VECTOR_CENTROIDS=true` (or `1`, or any other non-falsy value): current behavior, no warning.
- Unset: current behavior, plus a one-time `warn!` per process explaining that the default will change in a future release and how to lock in either behavior.

The check is centralized in a new `maybe_centroids_for_stats` helper so both the legacy `IVFIndex::statistics` (`rust/lance/src/index/vector/ivf.rs`) and the V2/V3 `IvfIndex::statistics` (`rust/lance/src/index/vector/ivf/v2.rs`) share the same gate. The struct field is now `Option<Vec<Vec<f32>>>` with `#[serde(skip_serializing_if = "Option::is_none")]` so the JSON shape is unchanged whenever centroids are included.

## Behavior matrix

| `LANCE_INCLUDE_VECTOR_CENTROIDS` | Reads centroids? | `centroids` in JSON? | Warning? |
| --- | --- | --- | --- |
| _unset_ | yes | yes | one-time `warn!` |
| `true` / `1` / other non-falsy | yes | yes | no |
| `false` / `FALSE` / `0` | no | omitted | no |

## Files changed

- `rust/lance/src/index/vector/ivf.rs` — struct field becomes `Option<...>` with `skip_serializing_if`; new `LANCE_INCLUDE_VECTOR_CENTROIDS_ENV` constant and `maybe_centroids_for_stats` helper (uses `std::sync::Once` so the deprecation warning fires at most once per process); legacy `statistics()` routed through the helper; two new unit tests.
- `rust/lance/src/index/vector/ivf/v2.rs` — V2/V3 `statistics()` routed through the same helper.

## Migration guidance for downstream consumers

- Code that currently reads `stats["centroids"]` will keep working unchanged (still emitted by default; warning is informational).
- Callers who don't need centroids can set `LANCE_INCLUDE_VECTOR_CENTROIDS=false` today to avoid the memory cost.
- Callers who do need centroids should set `LANCE_INCLUDE_VECTOR_CENTROIDS=true` to silence the warning and lock in the existing behavior ahead of the future default flip.

## Test plan

- [x] `cargo test -p lance --lib maybe_centroids_for_stats` — new test covers unset, `true`, `1`, `false`, `FALSE`, `0`; serialized via `#[serial_test::serial(LANCE_INCLUDE_VECTOR_CENTROIDS)]` so it doesn't race other env-var tests.
- [x] `cargo test -p lance --lib stats_centroids_omitted` — asserts the `centroids` field is fully absent from the serialized JSON when disabled (not serialized as `null`).
- [x] `cargo test -p lance --lib test_index_stats` — pre-existing IVF stats tests (IVF_FLAT/Hamming, IVF_PQ/L2, IVF_HNSW_SQ/Cosine, plus empty-partition) still pass against real datasets.
- [x] `cargo fmt --check -p lance`

## Notes / non-goals

- No Python or Java binding changes. One already broken python method was removed.
- No change to the legacy `IndexFileVersion::Legacy` JSON shape beyond the optional `centroids` field — the rest of `IvfIndexStatistics` is untouched.
- The deprecation warning intentionally uses `log::warn!` (already imported in `ivf.rs`) rather than `tracing` to match the surrounding module's logging style.

### DIFF
--- a/python/python/lance/dataset.py
+++ b/python/python/lance/dataset.py
@@ -6973,7 +6973,6 @@ class VectorIndexReader:
     """
     This class allows you to initialize a reader for a specific vector index,
     retrieve the number of partitions,
-    access the centroids of the index,
     and read specific partitions of the index.
 
     Parameters
@@ -7029,22 +7028,6 @@ class VectorIndexReader:
         """
 
         return self.stats["indices"][0]["num_partitions"]
-
-    def centroids(self) -> np.ndarray:
-        """
-        Returns the centroids of the index
-
-        Returns
-        -------
-        np.ndarray
-            The centroids of IVF
-            with shape (num_partitions, dim)
-        """
-        # when we have more delta indices,
-        # they are with the same centroids
-        return np.array(
-            self.dataset._ds.get_index_centroids(self.stats["indices"][0]["centroids"])
-        )
 
     def read_partition(
         self, partition_id: int, *, with_vector: bool = False
--- a/rust/lance/src/index/vector/ivf.rs
+++ b/rust/lance/src/index/vector/ivf.rs
@@ -49,6 +49,7 @@ use lance_core::{
     Error, ROW_ID_FIELD, Result,
     cache::{LanceCache, UnsizedCacheKey, WeakLanceCache},
     traits::DatasetTakeRows,
+    utils::parse::str_is_truthy,
     utils::tracing::{IO_TYPE_LOAD_VECTOR_PART, TRACE_IO_EVENTS},
 };
 use lance_file::{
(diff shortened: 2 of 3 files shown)


python/python/lance/dataset.py
```diff
@@ -7030,22 +7029,6 @@ def num_partitions(self) -> int:
 
         return self.stats["indices"][0]["num_partitions"]
 
-    def centroids(self) -> np.ndarray:
```
westonpace, line 7033: This method was already broken (there is no `get_index_centroids` method) so I'm just pulling it out.  This may be the shape we want for retrieving centroids in the future or it may not.  Either way, we can address that in a follow-up.

jackye1995 (approved):

PR #6655, opened 2026-04-30
feat: add SUFFIX_ARRAY scalar index type for n-gram search

### LINKED ISSUE #5232: [Feature] support for Infini-gram index
The infini-gram index system appears to be very useful for many LLM analysis related application and it would be useful to support it natively as lance index !

Some references :
https://github.com/liujch1998/infini-gram
https://export.arxiv.org/pdf/2401.17377v4
https://arxiv.org/pdf/2504.07096

Image

artemru:
fixes https://github.com/lance-format/lance/issues/5232

## Summary

- Add `SUFFIX_ARRAY` scalar index type for substring/n-gram search over text columns, complementing the existing inverted (full-text) index
- Integrate [libsais](https://github.com/IlyaGrebnov/libsais) for O(n) suffix array construction with a streaming multi-segment pipeline that handles datasets larger than memory
- New `infgram_query` scanner parameter for count, search, prob, ntd, and infgram_prob operations — inspired by the [infini-gram](https://infini-gram.io/) project
- Block-cached S3 queries with 64KB range reads instead of full file downloads, plus parallel segment dispatch across all segments
- Bloom filters per segment to skip non-matching segments before any I/O
- Case-insensitive search support and token-level (fixed-width) indexing

## Architecture

```
┌─────────────┐     ┌──────────────┐     ┌─────────────────┐
│ Python API  │────▶│ InfgramExec  │────▶│ SuffixArrayIndex│
│ scanner()   │     │ (DataFusion) │     │ (multi-segment) │
└─────────────┘     └──────────────┘     └────────┬────────┘
                                                  │
                         ┌────────────────────────┼────────────────────────┐
                         ▼                        ▼                        ▼
                   ┌───────────┐           ┌───────────┐            ┌───────────┐
                   │ Segment 0 │           │ Segment 1 │    ...     │ Segment N │
                   │ SA + Text │           │ SA + Text │            │ SA + Text │
                   │ BlockCache│           │ BlockCache│            │ BlockCache│
                   └───────────┘           └───────────┘            └───────────┘
```

- **Build**: libsais constructs suffix arrays per segment (parallel via `LANCE_SA_PARALLEL_BUILDERS`), writes `.raw` files for block-cacheable S3 access
- **Query**: Binary search on SA via 64KB block cache → resolve byte positions to row IDs via `doc_offsets` → top-K by occurrence count
- **Segments queried in parallel** via `buffer_unordered` (configurable with `LANCE_SA_PARALLEL_QUERIES`)

## Key files

| File | Description |
|------|-------------|
| `rust/lance-index/src/scalar/suffix_array/builder.rs` | libsais FFI + SA construction |
| `rust/lance-index/src/scalar/suffix_array/index.rs` | Multi-segment index with parallel queries |
| `rust/lance-index/src/scalar/suffix_array/query.rs` | Binary search, count, prob, ntd, infgram_prob |
| `rust/lance-index/src/scalar/suffix_array/plugin.rs` | Build/load/write pipeline |
| `rust/lance-index/src/scalar/suffix_array/block_cache.rs` | 64KB block cache for S3 range reads |
| `rust/lance-index/src/scalar/suffix_array/bloom.rs` | Per-segment bloom filters |
| `rust/lance/src/io/exec/infgram.rs` | DataFusion execution node |
| `python/python/lance/dataset.py` | Python scanner API (`infgram_query=`) |

## Python API

```python
import lance

ds = lance.dataset("s3://bucket/data.lance")

# Create index
ds.create_index("text_column", index_type="SUFFIX_ARRAY")

# Search — returns rows sorted by occurrence count
ds.scanner(infgram_query="hello world", limit=100).to_table()

# With options
ds.scanner(
    infgram_query={"query": "neural network", "column": "abstract"},
    columns=["title", "abstract", "_count", "_positions"],
    limit=50
).to_table()

# Count occurrences
ds._suffix_array_info("text_column_idx", "hello", operation="count")

# Conditional probability P(continuation | prompt)
ds._suffix_array_info("idx", "hello", operation="prob", continuation=" world")

# Next-token distribution
ds._suffix_array_info("idx", "the", operation="ntd")

# Boolean queries (AND/OR)
ds.scanner(infgram_query='"cat" AND "dog"', limit=10).to_table()
```

## Test plan

- [x] 343 Rust unit tests (`cargo test -p lance-index`)
- [x] 30 Python edge case tests (positions, scoring, boolean, prefilter, unicode, case-insensitive)
- [x] S3 performance verified on 131M-row dataset: queries in 1-6s
- [x] `limit=10` and `limit=10000` produce identical top-10 results


rust/lance-index/src/scalar/suffix_array/index.rs
```diff
@@ -0,0 +1,1564 @@
...
+                let tokenized = Bytes::copy_from_slice(tokenized_col.value(0));
+
+                // Fire-and-forget: write .raw file so future queries use true block cache.
+                // Non-blocking — the query proceeds immediately with from_bytes cache.
+                {
+                    let store_bg = store_arc.clone();
+                    let raw_name = raw_tok_name.clone();
+                    let data = tokenized.clone();
+                    tokio::spawn(async move {
+                        if let Err(e) = store_bg.write_raw_file(&raw_name, &data).await {
```
Xuanwo, line 667: The query path writes a persistent raw sidecar into the index directory when the sidecar is missing. Scans can mutate dataset storage outside index metadata and transaction tracking, and read-only stores will see unexpected write attempts.

rust/lance-index/src/scalar/suffix_array/index.rs
```diff
@@ -0,0 +1,1564 @@
...
+            }
+        };
+
+        match sa_query {
+            SuffixArrayQuery::Count { query_bytes } => {
+                let qb = maybe_lower(query_bytes);
+                let n = self.total_count(&qb).await?;
+                let mut map = RowAddrTreeMap::new();
+                if n > 0 {
+                    map.insert(n);
```
Xuanwo, line 333: The scalar-index search implementation returns aggregate statistics as row addresses for statistic queries. This breaks the trait contract and lets generic scalar-index callers interpret counts as real row ids.

rust/lance-index/src/scalar/suffix_array/index.rs
```diff
@@ -0,0 +1,1564 @@
...
+        &self,
+        _mapping: &HashMap<u64, Option<u64>>,
+        _dest_store: &dyn IndexStore,
+    ) -> Result<CreatedIndex> {
+        Err(Error::not_supported(
+            "Suffix array indices must be rebuilt, not remapped",
+        ))
+    }
+
+    async fn update(
```
Xuanwo, line 414: This index type is included in the scalar-index lifecycle, but its update method always returns unsupported. Append index optimization can fail for datasets after this index has been created.

rust/lance-index/src/scalar/suffix_array/plugin.rs
```diff
@@ -0,0 +1,1254 @@
...
+                seg_idx: 0,
+                text: Vec::new(),
+                compact_sa: Vec::new(),
+                pointer_width: 1,
+                corpus_bytes: 0,
+                doc_offsets: Vec::new(),
+                row_ids: Vec::new(),
+                bloom_bytes: Vec::new(),
+            };
+            let info = write_segment(index_store, &empty_segment, 1).await?;
```
Xuanwo, line 818: The empty-index path writes legacy single-segment file names while still recording segmented metadata. Empty suffix-array indices will load through the segmented reader and miss the files they just wrote.

rust/lance/src/dataset/scanner.rs
```diff
@@ -1862,6 +1881,14 @@ impl Scanner {
             extra_columns.push(ArrowField::new(SCORE_COL, DataType::Float32, true));
         }
 
+        if self.infgram_query.is_some() {
+            extra_columns.push(ArrowField::new(
+                crate::io::exec::infgram::COUNT_COL,
+                DataType::Float32,
```
Xuanwo, line 1887: The scanner exposes `_count` as a floating-point column and omits `_positions` from the available extra-column schema, while execution emits both result columns. Planning projection, filters, or ordering against infgram results can use missing or wrong types.

PR #6657, opened 2026-05-01
feat: add lance queue backed by memwal

jackye1995:
Adds an experimental `lance-queue` crate backed by real Lance tables and MemWAL shard WAL files.

The queue now uses a storage-only producer-shard model: fixed `id`, `producer_id`, and Lance JSON `payload` schema; logical partitioning hashes only `id`, while physical WAL shards are `(logical partition, producer_id)`. Consumers assigned to a logical partition read all producer shards and commit per-producer-shard offsets in Lance table metadata.

Also adds MemWAL shard initialization/opening support, table-metadata conflict detection for concurrent updates to the same key, storage-only write/read benchmarks, and an S3 benchmark report with raw CSVs and SVG graphs.


jackye1995:
Closing for now — may revisit as a proprietary feature.

PR #6662, opened 2026-05-01
perf: skip btree second-level page search when too unselective

westonpace:
## Summary

- Adds an env-var-controlled heuristic in `BTreeIndex::search` that abandons the second-level page search when the lookup already matched too much of the index. When triggered, search returns the new `SearchResult::Indeterminate` instead of reading any pages, and the planner falls back to a normal scan with predicate recheck.
- Thresholds: `LANCE_BTREE_SKIP_ROW_THRESHOLD` (default 1,000,000 candidate rows; `0` disables) and `LANCE_BTREE_SKIP_FRACTION_THRESHOLD` (default 0.15; `>= 1.0` disables). Both must be exceeded — selectivity has to be both absolutely large and a meaningful fraction of the dataset before we give up.
- New `SearchResult::Indeterminate` variant (no payload) flows through the existing index expression machinery: it converts to `IndexExprResult::AtMost(BlockList(empty))` ("every row could match, recheck"), and `combine_search_results` short-circuits to `Indeterminate` when any segment of a logical scalar index returns it.

## Motivation

A btree search has two phases: a cheap in-memory page lookup that picks which pages might contain the value, and a much more expensive disk read + flat search of those pages. When the lookup matches a large fraction of the pages, the second phase rarely refines the result enough to be worth the I/O — the caller is better off doing a normal scan with the predicate.

The exact threshold depends on row width: for very wide rows (images, embeddings) even a 1-row reduction can be worth the work, while for narrow rows the bar is much higher. The defaults here are a reasonable starting point; the env vars exist so we can sweep them in benchmarks before committing to a single value.

## Implementation notes

- Candidate rows are estimated as `matched_pages * batch_size`. The last page can be partial, but for the broad/non-selective queries this guards against, the difference is noise.
- `BTreeLookup` now tracks `total_pages` (set during `try_from_serialized`) so the fraction check is O(1) and requires no additional I/O.
- The decision is factored into a pure `should_skip_btree_search` function so it can be unit-tested without poking at process-wide env vars.
- `SearchResult::row_addrs()` keeps its existing `&NullableRowAddrSet` return type; for `Indeterminate` it returns a static empty set so existing call sites compile unchanged. Callers that need to distinguish "no matches" from "every row matches with recheck" should check `is_indeterminate()` first.

## Test plan

- [x] `cargo test -p lance-index --lib scalar::btree` (24 tests) — includes new `test_should_skip_btree_search_thresholds` (pure-function unit test of the heuristic) and `test_btree_search_skip_when_unselective` (end-to-end: with default thresholds, a small index returns `Exact`; the helper's decision matches expectations on the in-memory page count).
- [x] `cargo test -p lance-index --lib expression` (7 tests) and `cargo test -p lance --lib scalar` / `scalar_logical` (51 + 3 tests) all pass — verifies the `From<SearchResult>` and `combine_search_results` updates don't regress existing scalar-index paths.
- [x] `cargo clippy` and `cargo fmt --check` clean for both `lance-index` and `lance`.
- [ ] Follow-up benchmark to tune thresholds against real workloads (the env vars exist specifically to support this).

### DIFF
--- a/docs/src/format/table/index/scalar/btree.md
+++ b/docs/src/format/table/index/scalar/btree.md
@@ -56,3 +56,38 @@ The BTree index provides exact results for the following query types:
 | **Range**  | `column BETWEEN a AND b`  | BTree traversal for pages overlapping the range, then search each sub-index |
 | **IsIn**   | `column IN (v1, v2, ...)` | Multiple BTree lookups, union results from all matching sub-indices         |
 | **IsNull** | `column IS NULL`          | Returns rows from all pages where null_count > 0                            |
+
+## Skipping the Sub-Index Search for Unselective Queries
+
+A BTree search has two stages: an in-memory page lookup that selects which pages
+might contain matches, followed by a per-page sub-index search that reads each
+matching page from `page_data.lance` and returns row IDs. When the page lookup
+already matches a large fraction of the index, the second stage rarely refines
+the result enough to repay its I/O — the caller is usually better off doing a
+normal scan with a recheck on the predicate.
+
+When this triggers, `BTreeIndex::search` returns the `Indeterminate` search
+result (semantically: "every row could match, recheck downstream") without
+loading any pages. The query planner then falls back to a filtered scan.
+
+The decision is gated by **two thresholds**, both of which must be exceeded
+for the skip to fire:
+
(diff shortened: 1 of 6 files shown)


westonpace:
@claude review

PR #6665, opened 2026-05-02
fix: add more info when describe indices error

wojiaodoubao:
Add index name, uuid and columns so we know which index needs retrain.


Xuanwo (approved):
Thank yoU!

PR #6667, opened 2026-05-02
fix: detect table metadata key conflicts in concurrent UpdateConfig

jackye1995:
## Summary

- Two concurrent `UpdateConfig` transactions writing the same table metadata key would silently overwrite each other
- Add key-level conflict detection for `table_metadata_updates` in `modifies_same_metadata`, matching existing behavior for `schema_metadata_updates` and `field_metadata_updates`
- If either transaction uses `replace`, they always conflict (full-table metadata replacement)
- Otherwise, only conflict when both touch the same key


Xuanwo (approved):
Thank you for this fix!

PR #6670, opened 2026-05-03
feat: make Dataset::object_store sync by pre-initializing base stores

jackye1995:
## Summary

- Pre-initialize all base object stores during `Dataset` construction and cache them in `HashMap<u32, Arc<ObjectStore>>`
- `Dataset::object_store(base_id)` is now a sync HashMap lookup instead of an async `ObjectStore::from_uri_and_params` call
- `object_store_for_data_file`, `object_store_for_deletion`, `object_store_for_index` are also sync
- `with_object_store_wrappers` re-wraps cached base stores alongside the primary store
- ~40 call sites updated to remove `.await`

### DIFF
--- a/python/src/dataset.rs
+++ b/python/src/dataset.rs
@@ -2536,18 +2536,14 @@ impl Dataset {
 
     /// Get a snapshot of current IO statistics without resetting counters
     fn io_stats_snapshot(&self) -> PyResult<IoStats> {
-        let object_store = rt()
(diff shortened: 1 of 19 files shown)


wjones127 (commented):
@jackye1995 it looks like this is badly out of date. If you want to rebase it, I'd happily review it. Otherwise, let's just close it.

PR #6672, opened 2026-05-03
chore: update OpenDAL and object_store

Xuanwo:
This upgrades the OpenDAL integration stack to the latest compatible releases: `opendal` and `object_store_opendal` 0.56, alongside `object_store` 0.13.2.

The main compatibility work is adapting Lance's object-store wrappers and test stores to the `object_store` 0.13 trait shape, where convenience operations moved behind `ObjectStoreExt` and implementations now provide `delete_stream` / `copy_opts` directly. The PR also replaces deprecated `Path::child` calls with `Path::join` so the upgrade does not leave new deprecation noise behind.


jackye1995 (approved):
I can rebase this

PR #6673, opened 2026-05-03
fix: always do wal tailer cursor update

touch-of-grey:
## Summary
- Switch WAL entry positions to 0-based; appender uses the manifest `wal_entry_position_last_seen` hint with HEAD probing to discover its next write position, falling back to a directory listing only when the hint is stale.
- `WalTailer` always emits a best-effort manifest cursor update on successful `read_entry` (removes the opt-in `with_cursor_updates` flag).
- Extract the shared HEAD-probe loop into `probe_forward_from()`.

cc @jackye1995 for review.


touch-of-grey:
@jackye1995 could you take a look?

jackye1995 (approved):
Looks good to me

PR #6674, opened 2026-05-04
fix: add equals/hashCode to FullTextQuery concrete subclasses

ivscheianu:
## Changes

Adding `equals()` and `hashCode()` to `MatchQuery`, `PhraseQuery`, and `MultiMatchQuery` in `FullTextQuery`.

## Why

The concrete subclasses currently inherit reference equality from `Object`. Two independently constructed instances with identical fields are not `equals()` and produce different hash codes.

This was surfaced while trying to sketch an impl for SQL FTS query support in lance-spark, which allows users to push `lance_match`, `lance_phrase`, and `lance_multi_match` predicates down to the Lance FTS inverted index at query planning time. That work requires transporting `FullTextQuery` instances across the driver -> executor boundary, serializing them to a stable intermediate form and reconstructing them on the other side. Without structural equality, comparing a reconstructed instance to the original always returns `false` even when every field is identical, making correct scan identity checks impossible.

More broadly, any consumer that serializes a `FullTextQuery` and reconstructs it later faces the same problem. The fix is mechanical and non-breaking: callers that never compared these objects structurally are unaffected.


hamersaw (approved):
Looks great, thanks!

wjones127 (approved):

PR #6675, opened 2026-05-04
feat: unify WalAppender into ShardWriter via enable_memtable mode

touch-of-grey:
## Summary

- Add `ShardWriterConfig::enable_memtable` (default `true`); when `false`, `ShardWriter` runs in a new WAL-only mode that keeps the async batched WAL pipeline but skips MemTable allocation, in-memory indexes, MemTable freezing, and MemTable-bytes backpressure (a separate WAL-only backpressure budget reuses `max_unflushed_memtable_bytes`).
- `WalAppender` becomes the WAL write primitive used by both modes inside `WalFlusher`, replacing the prior duplicate plain-`put` path. MemTable mode now also gets atomic put-if-not-exists, conflict retry, and fence-on-write.
- `WalAppender` stays public as the lowest-level synchronous-atomic appender. New `pub(crate) WalAppender::with_claimed_epoch` lets `ShardWriter::open` claim the epoch once and inject it.
- WAL-only mode uses a drainable `WalOnlyState` queue with snapshot/commit semantics so a failed append leaves batches in the queue for retry instead of dropping them silently.
- `memtable_stats()`, `scan()`, `active_memtable_ref()` now return `Result<...>` and produce a clear `invalid_input` error in WAL-only mode.

Behavior change to call out: first WAL entry on a fresh shard is now position 0 instead of 1, matching the 0-based positions documented in the [MemTable & WAL spec](https://github.com/lance-format/lance/blob/main/docs/src/format/table/mem_wal.md). The previous flusher seeded its counter from `wal_entry_position_last_seen + 1` and so always skipped position 0.

Context: this realizes the layering discussed in https://github.com/lance-format/lance/pull/6669#discussion_r3177397523 — keep `WalAppender` as the low-level primitive and let users always use `ShardWriter`, with a config switch to turn the MemTable on or off.

cc @jackye1995

### DIFF
--- a/rust/lance/src/dataset/mem_wal/write.rs
+++ b/rust/lance/src/dataset/mem_wal/write.rs
@@ -721,9 +721,12 @@ fn now_millis() -> u64 {
 /// into the freshly-built MemTable. Updates any in-memory indexes attached to
 /// the MemTable so replayed rows are immediately searchable.
 ///
-/// Returns the highest WAL position replayed, or `None` if there were no
-/// entries to replay (fresh shard with no prior writes, or the manifest is
-/// already caught up with the WAL tip).
+/// Returns the WAL position the next `WalAppender::append` should use — i.e.
+/// one past the highest replayed position, or the original start position if
+/// the loop found nothing (we proved that position is empty by getting
+/// `None` back from the tailer). The caller can pass this directly to
+/// `WalAppender::seed_next_position` unconditionally so the first post-open
+/// append skips its own discovery probe.
 ///
 /// Aborts with an error if any replayed entry's `writer_epoch` is strictly
 /// greater than `our_epoch` — that indicates a successor writer claimed the
@@ -735,7 +738,7 @@ async fn replay_memtable_from_wal(
     our_epoch: u64,
     manifest: &ShardManifest,
     memtable: &mut MemTable,
-) -> Result<Option<u64>> {
+) -> Result<u64> {
     // Fresh shards (no flushes yet) start replay at position 0; otherwise
     // start one past the last covered position. Distinguishing "no flushes"
     // from "flushed up to position 0" requires `flushed_generations` since
@@ -746,13 +749,18 @@ async fn replay_memtable_from_wal(
         manifest.replay_after_wal_entry_position.saturating_add(1)
     };
 
+    // The MemTable is always freshly built before this function runs, so
+    // any existing BatchStore entries can only have come from this replay
+    // pass. We index everything in `[0, batch_count)` at the end.
+    debug_assert_eq!(memtable.batch_count(), 0);
+
     let tailer = WalTailer::new(object_store, base_path, shard_id);
-    let batches_before = memtable.batch_count();
-    let mut highest_replayed: Option<u64> = None;
     let mut position = start_position;
 
     loop {
         match tailer.read_entry(position).await? {
+            // The first NotFound proves the WAL tip is at `position`, which
+            // is the next write position to hand back.
             None => break,
             Some(entry) => {
                 if entry.writer_epoch > our_epoch {
@@ -764,7 +772,6 @@ async fn replay_memtable_from_wal(
                 if !entry.batches.is_empty() {
                     memtable.insert_batches_only(entry.batches).await?;
                 }
-                highest_replayed = Some(position);
                 position = position.checked_add(1).ok_or_else(|| {
                     Error::io(format!(
                         "WAL position overflow during replay for shard {}",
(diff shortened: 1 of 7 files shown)


jackye1995:
@claude review

touch-of-grey:
@jackye1995 — quick heads-up: there was a previous change I had drafted for doing MemTable replay on `ShardWriter::open` that I never pushed. Now that we have `WalTailer` in this PR's layering, we can just use it directly for replay, so I added that along the way as a separate commit.

What it does, in MemTable mode only:

- On open, after the fresh MemTable + indexes are built, walk `WalTailer::read_entry(pos)` from `manifest.replay_after_wal_entry_position + 1` (or `0` for a shard with no flushed generations yet) up to the tip.
- For each entry, check `entry.writer_epoch <= our_claimed_epoch` — strictly greater means a successor claimed the shard between our `claim_epoch` and replay, so we abort the open with a clear fence error.
- Insert the entry's batches into the MemTable and feed the resulting `StoredBatch`es to `IndexStore::insert_batches_parallel` (via `spawn_blocking`) so any in-memory indexes are rebuilt from the WAL.
- Seed `WalAppender::next_entry_position` from `highest_replayed + 1` so the first put after open writes past the replayed entries without paying for the lazy-discovery probe.

WAL-only mode is unchanged — no MemTable to rebuild.

Three new tests cover: recovery (writer A drops without close → writer B's reopened MemTable contains A's rows), no-op on a fresh shard, and fence-on-replay (a high-epoch entry injected via direct `WalAppender` causes the next opener to fail with `WAL replay aborted ... fenced`).

All 226 mem_wal tests pass; clippy + fmt clean.

Please take another look.

rust/lance/src/dataset/mem_wal/write.rs
```diff
@@ -961,14 +1133,98 @@ impl ShardWriter {
...
+        // `wal_stats().next_wal_entry_position` reflects the post-recovery
+        // cursor immediately on reopen instead of reading 0 until the
+        // first append has discovered the true tip.
+        let wal_appender = Arc::new(WalAppender::with_claimed_epoch(
+            object_store.clone(),
+            base_path.clone(),
+            shard_id,
+            manifest_store.clone(),
+            epoch,
+            manifest.wal_entry_position_last_seen.saturating_add(1),
```
jackye1995: the next_entry_position_hint should be the max of wal entry position last seen or the replay after wal entry position then +1

rust/lance/src/dataset/mem_wal/write.rs
```diff
@@ -680,6 +717,89 @@ fn now_millis() -> u64 {
...
+    // from "flushed up to position 0" requires `flushed_generations` since
+    // `replay_after_wal_entry_position` defaults to 0 in both cases.
+    let start_position = if manifest.flushed_generations.is_empty() {
+        0
+    } else {
+        manifest.replay_after_wal_entry_position.saturating_add(1)
+    };
+
+    let tailer = WalTailer::new(object_store, base_path, shard_id);
+    let batches_before = memtable.batch_count();
```
jackye1995: why do we need to record before? shouldn't it always be an empty memtable if we replay?

rust/lance/src/dataset/mem_wal/write.rs
```diff
@@ -680,6 +717,89 @@ fn now_millis() -> u64 {
...
+                if entry.writer_epoch > our_epoch {
+                    return Err(Error::io(format!(
+                        "WAL replay aborted: entry at position {} has writer_epoch {} > our claimed epoch {} for shard {} (writer was fenced during open)",
+                        position, entry.writer_epoch, our_epoch, shard_id
+                    )));
+                }
+                if !entry.batches.is_empty() {
+                    memtable.insert_batches_only(entry.batches).await?;
+                }
+                highest_replayed = Some(position);
```
jackye1995: at this point, the tailer already found the tip of the WAL. this optional information could be passed to the WAL appender so it can skip trying to re-discover the tip of the WAL when memtable is enabled.

rust/lance/src/dataset/mem_wal/write.rs
touch-of-grey: Done in 92054bec8 — seed is now `max(wal_entry_position_last_seen, replay_after_wal_entry_position) + 1` in both `WalAppender::open` and `ShardWriter::open`.

rust/lance/src/dataset/mem_wal/write.rs
touch-of-grey: Right, fixed in 92054bec8 — dropped the `batches_before` bookkeeping; the MemTable is freshly built before replay runs so we just index `[0, batch_count)`. Added a `debug_assert_eq!` for the invariant.

rust/lance/src/dataset/mem_wal/write.rs
touch-of-grey: Done in 92054bec8 — `replay_memtable_from_wal` now returns the next WAL write position unconditionally (highest replayed + 1, or the original `start_position` itself when the loop found nothing — that position is provably empty since `read_entry` just returned `None` for it). The caller seeds `WalAppender::next_entry_position` unconditionally, so the first post-open put skips the discovery probe even on shards where replay had nothing to do.

jackye1995:
The failure is unrelated, I will fix it in main.

jackye1995 (approved):
thanks for adding this!

PR #6677, opened 2026-05-04
feat: show scalar index type in ScalarIndexQuery plan output

wkalt:
The ScalarIndexQuery (and MaterializeIndex) plan nodes now display the underlying scalar index type (BTree, Bitmap, NGram, LabelList, etc.) alongside the index name, e.g. `[category = 1]@category_idx(Bitmap)`. This makes it easier to tell which kind of index is being used when inspecting analyze/explain output without cross-referencing the manifest.


jackye1995 (approved):
looks good to me!

PR #6679, opened 2026-05-05
fix(core): stop scalar-index training scans from over-projecting Map siblings

kushudai:
## Summary

`Field::apply_projection` special-cases `Map` types by cloning all children unconditionally (to preserve the entries-struct `Struct<key, value>` physical layout), but applies that special case **before** checking whether the parent Map is in the projection. Result: every `Map` column survives every projection regardless of what the caller asked for.

On real workloads this turns a narrow scan into a wide one: a scalar-index training scan over an `Int32` column on a table with sibling `Map<…, binary>` columns ends up feeding all of them into `SortExec` and blowing external-sort spill from a few GiB to >100 GiB.

## Fix

One early return: if the parent Map isn't selected, drop the subtree the same way every other non-selected leaf is dropped.

```rust
if self.logical_type.is_map() && !projection.contains_field_id(self.id) {
    return None;
}
```

The clone-children branch is unchanged for the case where the parent _is_ selected.

## Test

`scan_training_data_does_not_pull_unrelated_map_siblings` in `rust/lance/src/index/scalar.rs`: builds a 4-column dataset (`indexed: Int32` + three `Map` siblings), trains a BTree on `indexed`, asserts the scan plan's `LanceRead` projection is exactly `[indexed]`. Pins `LanceFileVersion::V2_2` since `Map` requires 2.2+.

## Notes

- No behavior change for projections that include the Map column, or for non-Map types.
- Cherry-picks cleanly onto `v4.0.x`, `Field::apply_projection` is identical between `main` and `v4.0.0`.
- The test was generated by Claude.


Xuanwo (approved):
Thank you for fixing this!

kushudai:
@Xuanwo Thank you for taking a look! I'm not sure how to go about debugging the python failures - any pointers would be much appreciated!!

wjones127 (approved):

PR #6685, opened 2026-05-05
fix: resolve system index types in describe_indices

wkalt:
describe_indices() returned "Unknown" for system indices (__lance_frag_reuse, __lance_mem_wal) because their proto details have no scalar plugin registered. list_indices() already special-cases these via infer_system_index_type(); apply the same check in IndexDescriptionImpl::try_new so the two methods agree.

### DIFF
--- a/python/python/tests/test_optimize.py
+++ b/python/python/tests/test_optimize.py
@@ -9,6 +9,7 @@ from pathlib import Path
 import lance
 import numpy as np
 import pyarrow as pa
+import pytest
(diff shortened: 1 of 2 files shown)


wkalt:
ref list_indices (load_indices): https://github.com/lance-format/lance/blob/9cae5bd6c0889f45bb2e878c3eea728521115d75/python/src/dataset.rs#L976-L989

wjones127 (approved):
This seems good.

PR #6687, opened 2026-05-05
perf: reduce memory use when splitting IVF partitions

### LINKED ISSUE #6378: Optimizing indexes with splitting could use a lot of RAM
When we optimize we might split or merge partitions.  On the merge path we simply delete one of the partitions and reassign _one partitions worth of vectors_.  On the split path we create two new partitions, we then load 64 (`REASSIGN_RANGE`) nearby partitions to see if any of their vectors should move to one of the new partitions.

In both cases when we reassign a partition we load the entire partition into memory.  When we reassign 64 partitions we parallelize this operation according to the number of compute intensive CPUs.  This is fast, but it incurs quite a bit of RAM.

In theory the RAM required doesn't really depend on the scale of the dataset (or the scale of the data we are adding).  Partitions should generally have around 4Ki rows (or whatever the target count is).  In practice I've seen partitions with hundreds of thousands of rows and I've seen cases where this splitting required 50GB+ of RAM.

Instead we should add an additional upper bound on how many vectors we allow into the reassign stage at once time.  For example, maybe we cap it at 500K vectors.  This would keep RAM pretty well bounded and shouldn't have too much impact on performance.

Alternatively, we could investigate whether there is much need to parallelize at all.  These partition loads are very sparse takes which means a lot of IOPS.  We are also usually loading from cloud storage.  As a result we are probably going to be I/O bound rather than CPU bound and we can probably cap the parallelism around 4 or so.

wjones127:
We improve memory use when splitting IVF partitions in two of the stages:

1. **Training new IVF centroids.** Currently, all raw vectors to partition to split are loaded into memory and used to train the centroids. This could be 400MB for a 3072 f32 vector dataset when partition size has reached 33k, triggering a split. We now just sample `512` of the vectors, which should be sufficient to train for just 2 centroids.
2. **Shuffle**. Currently, all vectors that will be moved across all partitions being split are held in memory simultaneously in `Vec<SplitPlan>`. **This is the largest source of peak memory use currently**. If many partitions are being split, this can be > 100GB. We now instead stream these raw vectors through the partition assignment and quantization pipeline, just like we do in the case of new indices.

This PR also adds progress reporting to `optimize_indices`, to make this more observable.

**Test Workload**: IVF_PQ append on 560K base rows (16 partitions, 3072-dim float32 vectors) with 160K new rows — triggers partition splitting since each partition exceeds the 32K row threshold.

**Peak RSS**: 26.2 GB before, 4.1 GB after.
**Runtime**: 93s before, 16.5s after — 5.6x faster as well

memory_comparison

Closes #6378

### DIFF
--- a/rust/lance/src/index/vector/builder.rs
+++ b/rust/lance/src/index/vector/builder.rs
@@ -88,6 +88,8 @@ use super::{
 
 // the number of partitions to evaluate for reassigning
 const REASSIGN_RANGE: usize = 64;
+// sample size for kmeans training when splitting a partition (sample_rate * k = 256 * 2)
+const SPLIT_SAMPLE_SIZE: usize = 512;
 
 // Builder for IVF index
 // The builder will train the IVF model and quantizer, shuffle the dataset, and build the sub index
@@ -841,14 +843,16 @@ impl<S: IvfSubIndex + 'static, Q: Quantization + 'static> IvfIndexBuilder<S, Q> 
                     let frag_reuse_index = frag_reuse_index.clone();
                     let partition_adjustment = partition_adjustment.clone();
                     async move {
-                        // For affected partitions in a split, read from the
-                        // split shuffle reader which has re-quantized data.
-                        // For all other partitions, use the normal path.
-                        let is_affected = matches!(
-                            partition_adjustment.as_ref(),
-                            Some(PartitionAdjustment::Split { affected_partitions, .. })
-                            if affected_partitions.contains(&partition)
-                        );
+                        let (is_affected, split_reader) = match partition_adjustment.as_ref() {
+                            Some(PartitionAdjustment::Split {
+                                affected_partitions,
+                                split_shuffle_reader,
+                            }) => (
+                                affected_partitions.contains(&partition),
+                                Some(split_shuffle_reader.clone()),
+                            ),
+                            _ => (false, None),
+                        };
                         let partition = match partition_adjustment.as_ref() {
                             Some(PartitionAdjustment::Join(joined_partition))
                                 if partition >= *joined_partition =>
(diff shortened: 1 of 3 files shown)


rust/lance/src/index/vector/builder.rs
```diff
@@ -633,6 +633,9 @@ impl<S: IvfSubIndex + 'static, Q: Quantization + 'static> IvfIndexBuilder<S, Q>
         } else {
             HashMap::new()
         };
+        let progress = self.progress.clone();
+        progress.stage_start("shuffle", None, "batches").await?;
```
wjones127: issue(blocking): We don't update the progress for batches processed. Could we add that?

suggestion: could we provide a total if there is a definite number of batches in the `size_hint()` of the input data?

watermelon12138:
Hi, Will Jones, I saw your name here: https://lance.org/community/pmc/#roster. I'm taking the liberty to ask in your PR comment — do you have any idea roughly when V6.0.0 and V7.0.0 will be released? I'm considering whether to wait for these two versions, otherwise I'll have to use v4.0.0.

rust/lance/src/index/vector/builder.rs
```diff
@@ -633,11 +659,26 @@ impl<S: IvfSubIndex + 'static, Q: Quantization + 'static> IvfIndexBuilder<S, Q>
...
         };
+        let progress = self.progress.clone();
+        let total = if let (lower, Some(upper)) = data.size_hint()
+            && lower == upper
+        {
+            Some(lower as u64)
+        } else {
+            None
+        };
+        progress.stage_start("shuffle", total, "batches").await?;
```
westonpace: Hmm...I wonder if we could capture the total number of rows in shuffle_dataset and then pass that into this method and use that as the total instead?  Or is the `size_hint` above guaranteed to be present?  I would think, since it is just a scan, it wouldn't be?

rust/lance/src/index/vector/builder.rs
```diff
@@ -858,7 +927,19 @@ impl<S: IvfSubIndex + 'static, Q: Quantization + 'static> IvfIndexBuilder<S, Q>
                             .await?
                         };
 
+                        // For unaffected partitions during a split, vectors from
+                        // affected partitions may have been reassigned here.
```
westonpace, line 944: This comment is a little confusing.  What is happening here?

rust/lance/src/index/vector/builder.rs
```diff
@@ -877,7 +958,6 @@ impl<S: IvfSubIndex + 'static, Q: Quantization + 'static> IvfIndexBuilder<S, Q>
                                 }
 
                                 if assign_batch.num_rows() > 0 {
-                                    // Drop PART_ID column from assign_batch to match schema of existing batches
```
westonpace, line 880: Why drop this comment?  I guess it was a little trivial but the motivation (matching schema) was kind of nice to mention?

rust/lance/src/index/vector/builder.rs
```diff
@@ -2287,118 +2303,6 @@ mod tests {
         );
     }
 
-    #[test]
-    fn compute_reassign_candidate_moves_vectors_to_new_centroids() {
```
westonpace, line 2291: Why are these test cases removed?

rust/lance/src/index/vector/builder.rs
```diff
@@ -633,11 +659,26 @@ impl<S: IvfSubIndex + 'static, Q: Quantization + 'static> IvfIndexBuilder<S, Q>
         } else {
             HashMap::new()
         };
+        let progress = self.progress.clone();
```
westonpace: Looks like you are adding progress to the old v1 shuffler?  The new shuffler in rust/lance-index/src/vector/v3/shuffler.rs already has progress reporting.  Is this intentional?  Are we using the v1 shuffler in this new split code?  Should we be using the v3 shuffler?

I guess this begs the question of why we have so many shufflers at this point 😰 

The v1 shuffler (IIRC) created one file but used quite a bit of RAM.  Then there are two v3 shufflers (IvfShuffler and TwoFileShuffler).  I think we can probably consolidate everything on TwoFileShuffler at this point.  It's been the default for a while and seems reliably and bounded.

westonpace (commented):
Some questions, nothing truly blocking, but would be good to simplify the shuffler paths (or at least keep using the new paths) if possible.

wjones127:
> Hi, Will Jones, I saw your name here: https://lance.org/community/pmc/#roster. I'm taking the liberty to ask in your PR comment — do you have any idea roughly when V6.0.0 and V7.0.0 will be released? I'm considering whether to wait for these two versions, otherwise I'll have to use v4.0.0.

@watermelon12138 If you want to know when a release is, you should watch the RC vote thread: https://github.com/lance-format/lance/discussions/6676 That vote will close tomorrow, and we likely will have a release after that.

rust/lance/src/index/vector/builder.rs
wjones127: Yeah, I'm not sure what the best way to get a total is. Can we calculate total rows cheaply?

rust/lance/src/index/vector/builder.rs
wjones127: Oh, that's not intentional. I didn't realize there are so many shufflers.

FWIW I started an issue for simplifying the architecture, https://github.com/lance-format/lance/issues/6686, though it's more along the lines of consolidating different branching code paths. ~Would be nice to put shufflers behind a trait or something.~

wjones127: I'll move this to use the TwoFileShuffler.

rust/lance/src/index/vector/builder.rs
wjones127: This is new in this PR. When we split a partition, we don't just split the vectors between the two new partitions. We resolve all the new centroids after splitting. Then we reassign all vectors to their nearest neighbor, exactly as we would if we were assigning them the first time. Normally, this will route vectors in split partitions to one of the new centroids. But it's possible for them to be routed to an existing partition. That's what this is handling.

rust/lance/src/index/vector/builder.rs
wjones127: I can bring this back.

rust/lance/src/index/vector/builder.rs
wjones127: They tested some internal functions directly, and those were removed. I added a new test in https://github.com/lance-format/lance/pull/6687/commits/1ca32706c242ee5f026600055ce4a71b55009450

watermelon12138:
ok, thanks.

westonpace (approved):
Awesome, thanks for centralizing on shufflers.

PR #6692, opened 2026-05-06
perf(index): use SIMD-dispatched l2_u8 in SQDistCalculator

martji:
SQDistCalculator was calling l2_distance_uint_scalar (pure scalar loop) for L2/Cosine distance computation on quantized u8 vectors. This misses the runtime-dispatched SIMD implementations (AVX2 / AVX-512 VNNI) that were added in #6517 and wired into the L2 trait.

Replace the direct scalar call with l2_u8::l2_u8(), which automatically selects the best available SIMD backend at runtime:
  - AVX-512 VNNI: 64 elements/iter
  - AVX2: 32 elements/iter
  - Scalar fallback for other architectures


Xuanwo:
@claude review

Xuanwo (approved):
Thank you for this change!

PR #6695, opened 2026-05-06
fix: resolve double percent-encoding in object store paths (#6643)

### LINKED ISSUE #6643: `extract_path` default implementation causes double percent-encoding for non-ASCII URI paths
### Summary

When a LanceDB/Lance database URI contains non-ASCII characters (e.g., CJK characters), the default `extract_path` implementation in `ObjectStoreProvider` causes **double percent-encoding** in the resulting S3/COS HTTP requests. This leads to object keys being stored in their percent-encoded form (e.g., `%E4%B8%AD%E6%96%87`) instead of the original UTF-8 characters (e.g., `中文`).

### Reproduction

```python
import lancedb

db = lancedb.connect(
    "s3://my-bucket/lancedb-test/中文路径",
    storage_options={
        "aws_access_key_id": "...",
        "aws_secret_access_key": "...",
        "region": "...",
        "endpoint": "...",    # e.g., Tencent COS S3-compatible endpoint
    },
)
db.create_table("test_table", [{"id": 1}])
```

Then inspect the actual HTTP request sent to S3/COS:

```
PUT /lancedb-test/%25E4%25B8%25AD%25E6%2596%2587%25E8%25B7%25AF%25E5%25BE%2584/test_table.lance/...
```

Note the `%25` — the `%` character from the first encoding round is being encoded again.

### Root Cause

The issue is in the default implementation of `extract_path` in `lance-io/src/object_store/providers.rs`:

```rust
fn extract_path(&self, url: &Url) -> Result<Path> {
    Path::parse(url.path())
        .map_err(|_| Error::invalid_input(format!("Invalid path in URL: {}", url.path())))
}
```

The data flow is:

| Step | Operation | Result |
|---|---|---|
| 1 | User input | `"s3://bucket/中文路径"` |
| 2 | `Url::parse()` | `url.path()` returns `"/%E4%B8%AD%E6%96%87%E8%B7%AF%E5%BE%84"` (percent-encoded by WHATWG URL standard) |
| 3 | `Path::parse(url.path())` | Stores `"%E4%B8%AD%E6%96%87%E8%B7%AF%E5%BE%84"` **verbatim** — `Path::parse` does **not** percent-decode its input |
| 4 | S3 client `encode_path(path)` | `utf8_percent_encode(path.as_ref(), STRICT_PATH_ENCODE_SET)` encodes `%` → `%25`, producing `%25E4%25B8%25AD...` |
| 5 | S3/COS server decodes once | Object key becomes `"%E4%B8%AD%E6%96%87%E8%B7%AF%E5%BE%84"` — still encoded, not the original `"中文路径"` |

The contract of `object_store::path::Path` is that it stores the **decoded logical path** internally. The `encode_path()` function in the S3 HTTP client is responsible for the single, correct encoding when building the request URL. By feeding an already-encoded string into `Path::parse()`, a double-encoding occurs.

### Why existing tests don't catch this

In `providers/aws.rs`, the test for non-ASCII paths does:

```rust
// Test case: ("s3://bucket/测试path/to/file", "测试path/to/file")
let path = provider.extract_path(&url).unwrap();
let expected_path = Path::from(expected_path);
assert_eq!(path, expected_path);
```

Both sides produce the **same** double-encoded internal representation:
- `extract_path` → `Path::parse("%E6%B5%8B%E8%AF%95path/to/file")` → internal: `%E6%B5%8B%E8%AF%95path/to/file`
- `Path::from("测试path/to/file")` → internal: `%E6%B5%8B%E8%AF%95path/to/file` (since `Path::from` encodes non-ASCII)

So `assert_eq!` passes — but neither has the correct internal representation. An end-to-end test that inspects the actual HTTP request URL would reveal the `%25` double-encoding.

### Suggested Fix

Replace `Path::parse(url.path())` with `Path::from_url_path(url.path())`, which correctly percent-decodes the URL path before constructing the `Path`:

```rust
fn extract_path(&self, url: &Url) -> Result<Path> {
    Path::from_url_path(url.path())
        .map_err(|e| Error::invalid_input(
            format!("Invalid path in URL '{}': {}", url.path(), e)
        ))
}
```

This ensures the `Path` internal representation is the decoded UTF-8 string (`"中文路径"`), and `encode_path()` performs the single correct encoding when building the HTTP request.

### Impact

- **Affected:** Any S3-compatible storage backend where the database URI path contains non-ASCII characters (CJK, accented characters, spaces, etc.)
- **Symptom:** Object keys on the storage side are stored in percent-encoded form instead of raw UTF-8
- **Compatibility risk:** If this is fixed, existing databases created with non-ASCII paths will have their keys in the old double-encoded form, and the new code will look for the correctly-encoded keys. A migration path or backward-compatibility layer may be needed.

### Environment

- LanceDB Python: 0.30.2
- Lance: (as bundled with the above)
- Storage: Tencent Cloud COS (S3-compatible), also likely affects AWS S3
- OS: Linux

shenganzhang:
### Description
This PR fixes the double percent-encoding bug identified in #6643. 

The issue was that `ObjectStoreProvider` implementations were using `Path::parse(url.path())`. Since `url.path()` returns a percent-encoded string (e.g., `中文` -> `%E4%B8%AD...`), `Path::parse` was storing these encoded characters verbatim. When the object store client later performed its own encoding for HTTP requests, it double-encoded them (e.g., `%` became `%25`), causing "File Not Found" errors for non-ASCII paths.

### Changes
- Replaced `Path::parse` with `Path::from_url_path` in `providers.rs` and specific overrides in `local.rs`, `memory.rs`, and `huggingface.rs`.
- `Path::from_url_path` correctly decodes the percent-encoding before creating the `Path` object.
- Updated `test_s3_path_parsing` and added a new regression test `test_s3_non_ascii_path_no_double_encoding` in `aws.rs`.

Fixes #6643

### DIFF
--- a/rust/lance-io/src/object_store/providers.rs
+++ b/rust/lance-io/src/object_store/providers.rs
@@ -54,8 +54,9 @@ pub trait ObjectStoreProvider: std::fmt::Debug + Sync + Send {
         // Path::from_url_path decodes it first so the Path internal representation
         // holds the raw UTF-8 string. This prevents double-encoding when the
         // object store client later percent-encodes the path for HTTP requests.
-        Path::from_url_path(url.path())
-            .map_err(|e| Error::invalid_input(format!("Invalid path in URL '{}': {}", url.path(), e)))
+        Path::from_url_path(url.path()).map_err(|e| {
+            Error::invalid_input(format!("Invalid path in URL '{}': {}", url.path(), e))
+        })
     }
 
     /// Calculate the unique prefix that should be used for this object store.
--- a/rust/lance-io/src/object_store/providers.rs
+++ b/rust/lance-io/src/object_store/providers.rs
@@ -50,8 +50,12 @@ pub trait ObjectStoreProvider: std::fmt::Debug + Sync + Send {
     /// Meanwhile, for a file store, the path is relative to the filesystem root.
     /// So a URL of `file:///path/to/file` would return `/path/to/file`.
     fn extract_path(&self, url: &Url) -> Result<Path> {
-        Path::parse(url.path())
-            .map_err(|_| Error::invalid_input(format!("Invalid path in URL: {}", url.path())))
+        // url.path() returns a percent-encoded string (per the WHATWG URL spec).
+        // Path::from_url_path decodes it first so the Path internal representation
+        // holds the raw UTF-8 string. This prevents double-encoding when the
+        // object store client later percent-encodes the path for HTTP requests.
+        Path::from_url_path(url.path())
+            .map_err(|e| Error::invalid_input(format!("Invalid path in URL '{}': {}", url.path(), e)))
     }
 
     /// Calculate the unique prefix that should be used for this object store.
--- a/rust/lance-io/src/object_store/providers/aws.rs
+++ b/rust/lance-io/src/object_store/providers/aws.rs
@@ -520,7 +520,7 @@ mod tests {
 
         let cases = [
             ("s3://bucket/path/to/file", "path/to/file"),
-            // for non ASCII string tests
+            // for non ASCII string tests: the URL encodes them, extract_path must decode back
             ("s3://bucket/测试path/to/file", "测试path/to/file"),
             ("s3://bucket/path/&to/file", "path/&to/file"),
             ("s3://bucket/path/=to/file", "path/=to/file"),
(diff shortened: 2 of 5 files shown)


shenganzhang:
@Xuanwo hi could you take a look at this fix? Thanks

wjones127 (approved):

wjones127:
I've rebased your PR and made some minor fixes. Hopefully this should be good-to-go.

PR #6696, opened 2026-05-06
test: target Lance 2.2 in bench_decode_packed_struct

westonpace:
## Summary
- The `bench_decode_packed_struct` benchmark in `rust/lance-encoding/benches/decoder.rs` was passing `LanceFileVersion::default()` to both `default_encoding_strategy` and `decode_batch`. Since variable packed struct encoding became 2.2-only, the bench now panics inside `encode_batch` with `Variable packed struct encoding requires Lance file version 2.2 or later`.
- Pin both call sites to `LanceFileVersion::V2_2` so the benchmark exercises the encoding it's supposed to measure.

## Test plan
- [x] `cargo check --benches -p lance-encoding`
- [ ] CI runs the encoding bench job to completion


westonpace:
Hmm...I still need to verify why this benchmark was using variable length packed struct?  None of the data types seem variable length

westonpace:
Ok, it does seem that 2.1 never supported FSL and packed struct.  We must have added that combination to the benchmark as part of 2.2 work.  So this fix is correct.

jackye1995 (approved):

PR #6698, opened 2026-05-06
fix: simplify column-map expressions before physical planning

westonpace:
## Summary

- `dataset.to_table(columns={"foo": "coalesce(col_a, col_b)"})` (and any other column-map expression containing `coalesce`) fails at scan time with `Internal error: coalesce should have been simplified to case`.
- Root cause: `ProjectionBuilder::add_column` in `rust/lance-datafusion/src/projection.rs` passes the parsed logical `Expr` straight into `create_physical_expr`, skipping the simplify+coerce step that `Planner::parse_expr`'s own doc comment requires. DataFusion's `coalesce` physical evaluator is a panicking stub that expects the optimizer to have rewritten `coalesce(...)` into a `CASE WHEN ... THEN ... ELSE ... END` first; without that rewrite, the panic surfaces as an `Internal error` from the scanner.
- Fix: run `optimize_expr` (the existing simplify+coerce helper already used by filter pushdown) on the parsed expression before storing it in the projection plan.

## Reproducer

```python
import tempfile
import lance
import pyarrow as pa

table = pa.table({
    "col_a": pa.array([1, None, 3, None], type=pa.int64()),
    "col_b": pa.array([10, 20, None, None], type=pa.int64()),
})
with tempfile.TemporaryDirectory() as tmp:
    ds = lance.write_dataset(table, tmp)
    print(ds.to_table(columns={"foo": "coalesce(col_a, col_b)"}, limit=2))
```

Before this PR:

```
pyarrow.lib.ArrowInvalid: External error: LanceError(IO): Internal error:
coalesce should have been simplified to case.
```

After this PR: returns `foo: [[1, 20]]` as expected.

## Test plan

- [x] Added regression test `test_coalesce_in_column_map` in `rust/lance-datafusion/src/projection.rs` that round-trips `coalesce(col_a, col_b)` through `ProjectionPlan::from_expressions` + `project_batch` and asserts the output.
- [x] `cargo test -p lance-datafusion --lib projection` — both projection tests pass.
- [x] `cargo test -p lance --lib dataset::scanner` — 131 tests pass, no regressions.
- [x] `cargo fmt --all` clean; `cargo clippy -p lance-datafusion --tests -- -D warnings` clean.
- [x] Verified end-to-end via the original Python reproducer: 2-arg coalesce, 3-arg coalesce with literal default, and the original `limit=2` form from the bug report all return the expected values.
- [ ] CI

### DIFF
--- a/rust/lance-datafusion/src/projection.rs
+++ b/rust/lance-datafusion/src/projection.rs
@@ -75,6 +75,10 @@ impl ProjectionBuilder {
         self.check_duplicate_column(output_name)?;
 
         let expr = self.planner.parse_expr(raw_expr)?;
+        // Run simplification + coercion so that expressions like `coalesce(...)`
+        // (which DataFusion's physical evaluator expects to have been rewritten
+        // into a `CASE` expression by the simplifier) work correctly.
+        let expr = self.planner.optimize_expr(expr)?;
 
         // If the expression is a bare column reference to a system column, mark that we need it
         if let Expr::Column(Column {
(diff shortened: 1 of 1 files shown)


jackye1995 (approved):
looks good to me

PR #6701, opened 2026-05-06
perf: use hnsw for memtable vector index

touch-of-grey:
## Summary

Replace MemWAL's in-memory IVF-PQ vector index with an online HNSW that's queryable while it's still building. The on-disk flush format also changes to HNSW + SQ8.

The original reason to keep IVF-PQ in the memtable was to share centroids and the PQ codebook with the base-table index, so distances would be directly comparable across the LSM stack. After discussing with @jackye1995 offline:

- We're moving to RQ across the index stack anyway.
- We're also moving to segment-level indexes, where refine happens per segment, so cross-segment distance comparability is no longer load-bearing.

Given that, IVF-PQ in the memtable doesn't earn its cost. It's slow at query time because every query scans many partition residuals linearly, and as the memtable grows the latency degrades fast.

I tested HNSW on S3 and the trade is decisively in HNSW's favor for live memtable query at 1M rows: ~160× faster p50, with a moderate write-throughput tax and a small extra memory footprint. Reasonable tradeoff.

## Performance comparison (S3, 1M × 1024, dim=1024, batch=1000)

c7i.12xlarge, us-east-1, `s3://...` for base table + WAL + flush.

### Live memtable query latency (median / p99 µs)

| index   | 100k         | 500k             | 1M                |
|---------|-------------:|-----------------:|------------------:|
| HNSW    | **171** / 528 | **240** / 530    | **260** / 568     |
| IVF-PQ  | 2,967 / 3,217 | 16,755 / 19,150  | 42,380 / 51,596   |

HNSW is 17× faster at 100k and **163× faster at 1M**. IVF-PQ degrades roughly linearly with row count; HNSW stays flat.

### Indexed write throughput (rows/s — wall time from first put through index-store catchup)

| index   | 1M, durable=false | 1M, durable=true |
|---------|------------------:|-----------------:|
| HNSW    | 8,154             | 5,874            |
| IVF-PQ  | 13,396            | 6,683            |

IVF-PQ is faster on writes (PQ-coding a vector is cheaper than HNSW graph insertion). Gap shrinks under \`durable_write=true\` because the per-put S3 round-trip dominates.

### Memtable flush throughput (rows/s — fragments + index files written to S3)

| index   | 1M, durable=false | 1M, durable=true |
|---------|------------------:|-----------------:|
| HNSW    | 9,392             | 8,046            |
| IVF-PQ  | 11,731            | 10,251           |

HNSW flush is 10–25 % slower (graph + SQ8 codes vs PQ codes + IVF partition table), dominated by S3 PUT round-trips on both sides.

### Resident memory at 1M

| index   | RSS (MB) |
|---------|---------:|
| HNSW    | 4,817    |
| IVF-PQ  | 4,104    |

HNSW uses ~700 MB more for the in-memory graph nodes.

## What changed

- New `HnswMemIndex` and lock-free `MemHnswStorage` that builds incrementally with single-writer / multi-reader concurrency. Storage references the same `Arc<FixedSizeListArray>` batches that BatchStore holds (zero-copy via `(batch_idx, offset)` lookup) — no parallel `Box<[f32]>` buffer, so memory usage tracks the data once.
- Flush converts the in-memory HNSW to Lance's standard `IVF_HNSW_SQ` on-disk format (HNSW graph + SQ8-quantized auxiliary file). SQ bounds are learned from the full memtable in one pass.
- Removed `index/ivf_pq.rs`. The HNSW path is now the only memtable vector index.
- Added a `mem_wal_index_micro` benchmark (this is what produced the numbers above) with `DATASET_PREFIX` and `BENCH_DURABLE_WRITE` env knobs for S3 runs.


touch-of-grey:
@jackye1995 mind taking a look when you have a moment? 🙏

jackye1995:
nice! thanks for doing this, @touch-of-grey do you also have recall number? we should verify the hnsw recall is good enough, at least not degrade

touch-of-grey:
## Recall verification

Ran a recall benchmark against `KShivendu/dbpedia-entities-openai-1M` (1M × 1536-dim OpenAI ada embeddings, cosine), loading the corpus into the MemTable through the production `ShardWriter` path and comparing HNSW top-10 against exact brute-force ground truth on 200 held-out queries (`HnswBuildParams::default()`, search-time `ef=64`, `k=10`).

| corpus | mean recall@10 | min recall@10 | HNSW p50 (µs) | HNSW p99 (µs) | brute-force / 200 q (s) |
|--------|---------------:|--------------:|---------------:|---------------:|--------------------------:|
| 100k   | **0.9375**     | 0.0000        | 1,580          | 1,963          | 14.82                     |
| 500k   | **0.9750**     | 0.0000        | 1,951          | 2,743          | 75.35                     |

Mean recall trends up with corpus size — more candidates inside the search beam — and HNSW stays sub-2 ms p50 against the in-memory MemTable. At 500k that's ~190x faster than brute force.

`min_recall = 0` at both sizes is the expected HNSW long tail at `ef=64` (a couple of queries out of 200 land in regions of the graph the default beam can't reach). Bumping `ef` at query time would close that, but we did not tune it for this run — these numbers are with library defaults.

Bench source: `rust/lance/benches/mem_wal_recall_hnsw.rs`.

touch-of-grey (commented):
The 1M checkpoint was killed after 5h30m of HNSW build at dim=1536, this is due to the single threaded execution of HNSW build, I will try to see if I can fix later.

jackye1995 (approved):
thanks for the work! I don't have too much time to review tbh, but this is great work and the perf results looks good to me. Let's get this in and continue iterating on this.

PR #6704, opened 2026-05-06
fix(java): honor user-specified distance type in VectorTrainer

jiaoew1991:
The JNI vector trainer hardcoded `MetricType::L2` in both `inner_train_ivf_centroids` and `inner_train_pq_codebook`, ignoring the metric the caller intends to use when later building per-fragment index segments. For non-L2 metrics (cosine, dot, hamming) this silently trained centroids and PQ codebooks on L2 geometry while encoding happened against the user's actual metric, producing degraded recall with no error or warning.

Add `DistanceType distanceType` as a parameter to
`VectorTrainer.trainIvfCentroids` and `trainPqCodebook` (mirroring the shape of the underlying Rust `build_ivf_model` / `build_pq_model`, which take `metric_type` separately from `IvfBuildParams` / `PQBuildParams`). The metric belongs at the function call site, not on the algorithm-config struct — the param structs stay isomorphic to their Rust counterparts and there is a single source of truth for the metric per training call.

Backward-compat overloads default `distanceType` to `DistanceType.L2` to preserve the existing call sites. The native methods take the distance type as a separate `String` argument and parse it via `MetricType::try_from`.

Add regression tests `testTrainIvfCentroidsHonorsDistanceType` and `testTrainPqCodebookHonorsDistanceType` that train the same dataset twice with L2 and Cosine and assert the resulting arrays differ. With the bug present, the two arrays were identical because both paths fell through to L2.

### DIFF
--- a/java/lance-jni/src/vector_trainer.rs
+++ b/java/lance-jni/src/vector_trainer.rs
@@ -34,6 +34,11 @@ fn flatten_fixed_size_list_to_f32(arr: &FixedSizeListArray) -> Result<Vec<f32>> 
     Ok(values.values().to_vec())
 }
 
+fn parse_distance_type(env: &mut JNIEnv, distance_type_jstr: &JString) -> Result<MetricType> {
+    let distance_type_str: String = env.get_string(distance_type_jstr)?.into();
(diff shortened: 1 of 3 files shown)


jiaoew1991:
@claude review

wjones127 (approved):
Thanks for submitting this PR! Looks good.

PR #6705, opened 2026-05-06
fix: honor Operation::CreateIndex removed_indices 

jiaoew1991:
`Operation::CreateIndex` documents that "any existing indices with the same name will be replaced" (`rust/lance/src/dataset/transaction.rs`), but `commit_existing_index_segments` was hardcoding `removed_indices: vec![]` so the documented same-name replace never happened.

Downstream callers building distributed vector indexes (lance-spark, lance-ray) that rebuild a same-name index were forced to emulate replace-on-recreate with `drop_index` + `commit_existing_index_segments`, which are two separate manifest commits. If anything fails between them — driver crash, version conflict, transient I/O — the dataset is left at the intermediate state "old index dropped, new index not yet installed" and queries see no index until the user reruns.

Populate `removed_indices` from `load_indices_by_name(index_name)` so the drop and create fold into one `Operation::CreateIndex` transaction and a single manifest version bump. No public signature changes; this is a pure bugfix aligning the implementation with the documented transaction semantics.

Adds regression test
`test_commit_existing_index_segments_replaces_same_name_in_single_commit` that installs an original index, commits new segments under the same name, and asserts exactly one manifest version bump and only the new segment remains bound to the index name.

### DIFF
--- a/rust/lance/src/index.rs
+++ b/rust/lance/src/index.rs
@@ -877,6 +877,13 @@ impl DatasetIndexExt for Dataset {
             )));
         };
 
+        // Honor the documented `Operation::CreateIndex` semantics: any existing indices with
+        // the same name are replaced as part of this same commit (see transaction.rs).
+        // This folds drop-and-create into one manifest version bump so callers replacing a
+        // same-name index never observe an intermediate state where the old index is gone but
+        // the new one is not yet installed.
+        let removed_indices = self.load_indices_by_name(index_name).await?;
+
         let new_indices =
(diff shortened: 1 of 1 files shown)


jiaoew1991:
Closing as superseded — upstream commit ed8bde5d (PR #6593, "feat: support zonemap index segments" by @beinan, merged 2026-04-24) already implemented the same fix with richer safety checks (field consistency, index type/type_url filtering, fragment overlap analysis to support delta append + replace mixed cases, orphan-fragment rejection). Apologies for the noise — I missed that 6593 had landed when I drafted this.

PR #6706, opened 2026-05-07
fix: sync unenforced_primary_key_position when field metadata updates

touch-of-grey:
## Summary

`UpdateConfig`'s field-metadata path in `Transaction::apply_update_map` (`rust/lance/src/dataset/transaction.rs`) mutates `field.metadata` but does not refresh the cached `Field::unenforced_primary_key_position`. The protobuf encoder for `Field` reads the cached `Option<u32>` (not the metadata HashMap), so the next commit drops the PK marker even though the metadata HashMap correctly contains `lance-schema:unenforced-primary-key:position`. The mirror image is also broken: removing the keys from metadata leaves a stale cached `Some(_)` that gets re-encoded as a PK marker on the next commit.

This makes it currently impossible to install or remove an unenforced primary key on an existing dataset via any public API — `update_field_metadata` and the deprecated `replace_field_metadata` both go through this path, and the round-trip via reopen returns no PK.

## Fix

After applying each `field_metadata_update`, re-derive `field.unenforced_primary_key_position` from the freshly-updated metadata using the same parsing rules the Arrow → Field decoder uses (position key first, then the legacy boolean-flag fallback). This keeps the two views in sync so the next protobuf encoding reflects the user's intent.

Also re-exports `LANCE_UNENFORCED_PRIMARY_KEY` from `lance_core::datatypes` alongside the existing `LANCE_UNENFORCED_PRIMARY_KEY_POSITION` re-export, so callers (and the new test) can use both constants without reaching into the private `field` module.

## Test

`test_update_field_metadata_syncs_unenforced_primary_key_position` in `rust/lance/src/dataset/metadata.rs` covers:

- Installing the PK on a column via the position key — cached option must be `Some(0)`, `Schema::unenforced_primary_key()` must report it, and a fresh `Dataset::open` must round-trip the marker.
- The legacy boolean-flag form must also sync the cached option.
- Removing the keys must clear the cached option so the marker is not re-encoded.

The test fails on `main` (returns `None` instead of `Some(0)`) and passes with the fix.

### DIFF
--- a/rust/lance/src/dataset/transaction.rs
+++ b/rust/lance/src/dataset/transaction.rs
@@ -18,6 +18,7 @@ use crate::dataset::transaction::UpdateMode::{RewriteColumns, RewriteRows};
 use crate::index::mem_wal::update_mem_wal_index_merged_generations;
 use crate::utils::temporal::timestamp_to_nanos;
 use deepsize::DeepSizeOf;
+use lance_core::datatypes::{LANCE_UNENFORCED_PRIMARY_KEY, LANCE_UNENFORCED_PRIMARY_KEY_POSITION};
 use lance_core::{Error, Result, datatypes::Schema};
 use lance_file::{datatypes::Fields, version::LanceFileVersion};
 use lance_index::mem_wal::MergedGeneration;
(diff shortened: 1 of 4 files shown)


jackye1995 (approved):
thanks for the fix!

rust/lance/src/dataset/transaction.rs
```diff
@@ -2358,6 +2359,29 @@ impl Transaction {
                 for (field_id, field_metadata_update) in field_metadata_updates {
                     if let Some(field) = manifest.schema.field_by_id_mut(*field_id) {
                         apply_update_map(&mut field.metadata, field_metadata_update);
+                        // The unenforced primary key position lives in two
```
jackye1995: this should be much more concise, just something like "also set unenforced primary key based on updated field metadata" is sufficient.

jackye1995:
looks like CI failure:

```
failures:
    dataset::mem_wal::write::tests::test_wal_only_size_trigger_fires_repeatedly
```

touch-of-grey:
The failing test `test_wal_only_size_trigger_fires_repeatedly` is in `rust/lance/src/dataset/mem_wal/write.rs`, which this PR does not touch. The test was introduced by #6675 (e9d161734) and is timing-sensitive — it pushes 3 puts with a 50ms sleep between them and relies on a background tokio flush task draining the size-trigger queue between puts. On a slower Windows runner the 50ms sleep can be too short for the trigger queue to drain, allowing multiple crossings to coalesce into a single WAL entry — exactly the failure mode reported (`next_position = 2` instead of `>= 3`).

The same test passed on Linux CI in this run, and this PR's only changes are in `transaction.rs::apply_update_map` (field-metadata sync) and a new unit test in `metadata.rs` — neither exercises the WAL-only size-trigger path.

Could you please re-run the failed `windows-build` job? I don't have admin rights on the upstream repo, so I can't re-run it myself.

touch-of-grey:
Pushed b32468fb1 to address both:

1. **Windows CI flake** — bumped the inter-put sleep in `test_wal_only_size_trigger_fires_repeatedly` from 50ms to 250ms. The original 50ms was tight enough that slow Windows runners couldn't always drain the size-trigger queue between puts, causing crossings to coalesce. Verified locally on macOS: 5/5 reliable runs.

2. **Codecov 1-line partial** — extended the metadata-sync regression test to cover all truthy spellings of the legacy `lance-schema:unenforced-primary-key` flag (`true`, `1`, `yes`, plus mixed-case) and the falsy / non-numeric fallback paths.

PTAL.

rust/lance/src/dataset/transaction.rs
touch-of-grey: Done in 34a0be5f7 — trimmed the comment to a single line.

PR #6708, opened 2026-05-07
perf: skip per-row DataBlock allocations in packed-struct variable decompressor

westonpace:
## Summary

Discovered a potential opportunity to improve packed struct decode times when I was doing some benchmark analysis.

A flamegraph shows where the time was going:

- `PackedStructVariablePerValueDecompressor::decompress` — 91% of samples
- `drop_in_place<FixedWidthDataBlock>` — 37% of samples

For each row × each fixed-width child, the decompress loop was:

1. allocating a 4-byte `Vec<u8>` (`data.data[cursor..end].to_vec()`),
2. wrapping it in a `LanceBuffer`,
3. wrapping that in a `FixedWidthDataBlock` carrying a fresh `BlockInfo` (which is `Arc<RwLock<HashMap<Stat, ArrayRef>>>`),
4. wrapping that in a `DataBlock::FixedWidth`,
5. calling `DataBlockBuilder::append(&block, 0..1)` — which internally just does `Vec::extend_from_slice` on 4 bytes,
6. dropping the whole stack.

For the bench (10k rows × 7 logical fields after FSL flattening) that's 70k micro-allocations and 70k `Arc<RwLock>` create+drop pairs, all to drive 70k 4-byte memcpys. The variable-width branches did the same thing with offsets vectors on top.

This PR rewrites `FieldAccumulator` so its variants hold raw byte buffers and offset vectors directly, and the inner loop appends to them with `extend_from_slice` plus a running offset. The final `DataBlock` is constructed once at the end, instead of one per row. The same shape-cleanup is applied to the inner unpack loop in `PackedStructFixedWidthMiniBlockDecompressor::decompress` (borrow `data.as_ref()` once, raw byte indexing, no per-element `slice_with_length` Arc bumps); that path isn't on this bench but had the same pattern.

## Numbers

`decode_primitive/struct` (10k rows, struct of `Int32` + `Float32` + `FSL<Int32, 5>` with packed metadata, V2.2):

| | Time | Throughput |
|---|---|---|
| Before (current `main`) | 2.07 ms | 124 MiB/s |
| After (this PR) | **281 µs** | **951 MiB/s** |

Speedup: ~7.4×. The post-fix flamegraph shows `PackedStructVariablePerValueDecompressor::decompress` is now ~42% of samples and is dominated by the `extend_from_slice` memcpy itself; the `drop_in_place<FixedWidthDataBlock>` spike is gone.

A gap remains versus the legacy V2.0 path (~106 µs), but that comes from real differences in the V2.2 pipeline (repdef unraveling, builder fan-out) and is out of scope here.

## Test plan

- [x] `cargo test -p lance-encoding --release` — all 369 tests pass, including the 5 `variable_packed_struct_*` round-trip tests
- [x] `cargo clippy -p lance-encoding --tests --benches -- -D warnings` — clean
- [x] `cargo fmt --all -- --check` — clean
- [x] Bench measured locally before and after on the same hardware/toolchain

### DIFF
--- a/rust/lance-encoding/src/encodings/physical/packed.rs
+++ b/rust/lance-encoding/src/encodings/physical/packed.rs
@@ -21,10 +21,7 @@ use crate::{
         DefaultCompressionStrategy, FixedPerValueDecompressor, MiniBlockDecompressor,
         VariablePerValueDecompressor,
     },
-    data::{
-        BlockInfo, DataBlock, DataBlockBuilder, FixedWidthDataBlock, StructDataBlock,
-        VariableWidthBlock,
-    },
+    data::{BlockInfo, DataBlock, FixedWidthDataBlock, StructDataBlock, VariableWidthBlock},
     encodings::logical::primitive::{
         fullzip::{PerValueCompressor, PerValueDataBlock},
         miniblock::{MiniBlockCompressed, MiniBlockCompressor},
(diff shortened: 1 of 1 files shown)


westonpace:
I haven't looked at the code yet to see if this is realistic.  Mostly just parking this here to look at later.

Xuanwo:
I tried this before but finally gave up because we need to maintain the offset by ourselves and lost the safeguard from data block builder.

And by that time, the cost here isn't that big. Maybe in different workloads?

PR #6709, opened 2026-05-07
perf: revert inline scheduling

westonpace:
This reverts commit 6e40d78cc448765548cf15593d0b58b60653d078.

We detected some cases where this has a pretty significant performance regression and we need to understand it more before proceeding.


wkalt (approved):
approved pending CI. Thanks.

PR #6711, opened 2026-05-07
fix: do no work when optimize_indices is called but no new data has arrived

westonpace:
## Summary

- Make `optimize_indices()` a no-op when no fragment is unindexed and the caller hasn't asked for `retrain` or `num_indices_to_merge >= 1`. Without this, a steady-state dataset (everything already indexed) had every call rewrite identical scalar-index files and trigger a vector single-segment auto-rebalance, producing a new manifest, transaction file, and per-index files on every call.
- Adds a Python regression test that builds one of every index type (BTREE, BITMAP, LABEL_LIST, INVERTED, NGRAM, ZONEMAP, BLOOMFILTER, IVF_PQ), appends a fragment, runs `optimize_indices()` twice, and asserts the second call writes zero new files under the dataset directory.

## Why

Two independent paths were both writing files when there was nothing to do:

1. **Scalar indices** — `merge_indices` in `rust/lance/src/index/append.rs` always called `index.update(...)` even when `unindexed.is_empty()`, producing a new UUID directory with identical contents.
2. **Vector indices** — the single-segment auto-rebalance from #6402 fires when `num_segments > 1 && num_indices_to_merge.is_none() && !retrain && unindexed.is_empty()`. After a prior `optimize_indices()` had folded a delta into its own segment, every subsequent call re-rebalanced one segment.

The user-visible symptom was that calling `optimize_indices()` in a loop kept bumping the dataset version and creating index files indefinitely, even when the dataset hadn't changed.

## Approach

The fix lives in `Dataset::optimize_indices` in `rust/lance/src/index.rs`. Before calling `merge_indices` for each grouped delta set, we now skip the group when:

- the caller did not pass `retrain = true`, **and**
- `num_indices_to_merge` is `None` or `Some(0)` (no explicit delta consolidation requested), **and**
- every fragment in the dataset is already covered by the union of the group's existing `fragment_bitmap`s.

The bitmap union is computed from the metadata we already loaded (no extra IO). Pre-0.8 indices without a `fragment_bitmap` are treated conservatively — they still go through the merge path so they pick up the new format on first optimize.

If any index group has work to do, the rest of the function is unchanged; the new check is purely a `continue`. When every group is skipped, the existing `if new_indices.is_empty() { return Ok(()); }` short-circuit prevents any commit, so no manifest or transaction file is written.

## Test plan

- [x] New Python test `test_optimize_indices_second_call_is_noop` passes
- [x] `python/tests/test_optimize.py`, `python/tests/test_scalar_index.py`, `python/tests/test_vector_index.py` — 225 passed, 10 skipped (pre-existing), 1 deselected (unrelated pre-existing failure in `test_create_index_progress_callback_error_before_completion_propagates`)
- [x] `cargo test -p lance --lib index::` — 389 passed, 1 ignored (the auto-rebalance test noted above)
- [x] `cargo clippy --lib --tests -- -D warnings` — clean
- [ ] CI green

## Notes for reviewers

- The fragment-bitmap-only check intentionally skips unioning bitmaps from system indices (`__lance_frag_reuse`, `__lance_mem_wal`); those are filtered out before this loop by the existing `is_system_index` check.

### DIFF
--- a/python/python/tests/test_optimize.py
+++ b/python/python/tests/test_optimize.py
@@ -479,8 +479,12 @@ def test_optimize_indices_second_call_is_noop(tmp_path: Path):
     dataset.create_scalar_index("name", index_type="NGRAM")
     dataset.create_scalar_index("value", index_type="ZONEMAP")
     dataset.create_scalar_index("bloom_val", index_type="BLOOMFILTER")
+    # num_partitions=1 keeps this dataset balanced: the auto-rebalance check
+    # in merge_indices only finds join candidates when num_partitions > 1, and
+    # 1024 + 128 rows is well below the split threshold. Without this, the
+    # rebalance heuristic would keep finding work on the small partitions.
     dataset.create_index(
-        "vector", index_type="IVF_PQ", num_partitions=2, num_sub_vectors=2
+        "vector", index_type="IVF_PQ", num_partitions=1, num_sub_vectors=2
     )
 
     extra_rows = 128
--- a/python/python/tests/test_optimize.py
+++ b/python/python/tests/test_optimize.py
@@ -448,6 +448,77 @@ def test_migration_via_fragment_apis(tmp_path):
     assert ds2.data_storage_version == "2.0"
 
 
+def test_optimize_indices_second_call_is_noop(tmp_path: Path):
+    """A second optimize_indices call when nothing has changed since the first
+    must not write any new files to the dataset directory."""
+    base_dir = tmp_path / "dataset"
+
+    n = 1024
+    rng = np.random.default_rng(0)
+    vectors = rng.standard_normal((n, 8)).astype(np.float32)
+    table = pa.table(
+        {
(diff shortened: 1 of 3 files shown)


westonpace:
@claude review once

wjones127 (approved):
I think the claude comment is valid, but not an important edge case.

Thanks for implementing this!

andrea-reale:
Thanks for this fix! 

One thing I noted: this "fast path" is gated on `num_indices_to_merge` being 0 (or None). This means that if as a caller passes `Some(arbitarily_large_n)` than this goes back to copying all indices _even if_ the scalar index is already optimal with only 1 delta.

Am I reading this wrong? Is this intended behavior?

westonpace:
> Am I reading this wrong? Is this intended behavior?

I think you're right, we could do better here.  I guess it shouldn't matter at all what the value of `num_indices_to_merge` is.

PR #6714, opened 2026-05-08
fix(scanner): translate row addresses to stable ids in take fast path

hamersaw:
When stable row ids are enabled, the take fast path for `_rowaddr`/ `_rowoffset` filters built a `RowAddrMask` from user-supplied addresses and intersected it against each fragment's `RowIdSequence`.  But that sequence stores logical ids when stable row ids are enabled (not addresses), so the mask never intersected and the scan returned the wrong (usually empty) rows.

Translate addresses to stable ids via `RowIdSequence::get` before building the mask so the mask and sequence live in the same value space.  `_rowid` filters already operated in id space and were unaffected.

Adds a parametrized `can_filter_row_id` test covering both `_rowid` and `_rowaddr` filters across stable and legacy row id datasets.

### DIFF
--- a/rust/lance/src/dataset/scanner.rs
+++ b/rust/lance/src/dataset/scanner.rs
@@ -7694,21 +7694,17 @@ mod test {
         }
     }
 
-    #[rstest]
-    #[case::stable_row_ids(true)]
-    #[case::row_addrs_as_ids(false)]
-    #[tokio::test]
-    async fn can_filter_row_id(#[case] use_stable_row_ids: bool) {
-        // 5 fragments x 100 rows so logical row ids and row addresses diverge
-        // across fragment boundaries. With stable row ids the logical id space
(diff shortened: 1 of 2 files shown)


westonpace:
@claude review once

westonpace (commented):
On the one hand, this is better than the current behavior (doing the wrong thing).  On the other hand, the low level exec nodes operate on addresses so we are now going from row addrs to row ids and then back to row addresses.

I'm not sure if it could be done cleanly but it would be nice to skip the intermediate representation and just somehow get the row addresses to the lower level operations.

hamersaw:
>I'm not sure if it could be done cleanly but it would be nice to skip the intermediate representation and just somehow get the row addresses to the lower level operations.

Thanks for taking a look! I agree, let me take a stab at this and see what I can come up with. Overall, this kind of seems like an anti-pattern, using stable row IDs and then filtering by row addresses, so it feels like pretty low priority.

Issue #6713, opened 2026-05-08
MemWAL: memtable_flusher panics on first 1M-row threshold flush — "Dataset already exists: …_gen_1"


touch-of-grey:
**Lance:** `v7.0.0-beta.6` (commit `b66eee73`)
**Build:** Linux x86_64, release, c7i.12xlarge

## Symptom

When repeated `merge_insert` upserts via the MemWAL `ShardWriter` path push the active MemTable past `max_memtable_rows`, the first triggered flush errors out:

```
[ERROR lance::dataset::mem_wal::write] Task 'memtable_flusher' error handling message:
Dataset already exists:
s3://<bucket>/.../bench_sync_no-idx.lance?/_mem_wal/<shard_uuid>/bda7e4a9_gen_1,
/home/ubuntu/.cargo/git/checkouts/lance-.../rust/lance/src/dataset/write/insert.rs:286:28

thread 'lancedb-tokio-worker' panicked at
/home/ubuntu/.cargo/git/checkouts/lance-.../rust/lance/src/dataset/mem_wal/util.rs:79:14:
watch channel closed: RecvError(())
```

The subsequent panic at `mem_wal/util.rs:79` propagates as `RustPanic: rust future panicked: unknown error` to the caller, killing the `merge_insert` call mid-stream.

## Reproducer

Configured a `ShardWriter` via the LanceDB MemWAL `merge_insert` path with:

- `durable_write = true` (default)
- `max_memtable_rows = 1_000_000`
- `max_memtable_batches = 10_000_000`
- Single shard (unsharded MemWAL spec)

Upsert workload writes 100-row batches sequentially; once the cumulative row count crosses 1M, the flusher attempts to write a fragment generation file at `<shard>/<frag_id>_gen_1` and finds the path already populated. From the path it looks like the fragment ID is reused or the generation counter is not advanced before the flush fires.

Branch / repro script with config: https://github.com/touch-of-grey/lancedb/blob/MergeInsertShardWriter/  (LanceDB is consuming the released `v7.0.0-beta.6`, so this is a Lance-side bug surfacing through that consumer).

## Likely culprits

- The `_gen_1` suffix repeats across flushes — generation counter not bumped before the second flush attempts its write?
- Or fragment ID is stable across flushes when it should be unique-per-flush.
- `mem_wal/util.rs:79` has an `unwrap()` on a watch receive that should propagate the flusher's error instead of panicking the worker.

[...]

PR #6715, opened 2026-05-08
fix(mem_wal): keep dispatcher and waiters alive on flush handler errors

touch-of-grey:
## Summary

Two defensive fixes to the MemWAL flush task tree, surfaced by lance-format/lance#6713:

1. `TaskExecutor`'s dispatcher loop no longer breaks on `handle()` Err — it logs and keeps consuming messages so subsequent flushes still make progress.
2. `WatchableOnceCellReader::await_value` returns `Option<T>` instead of `expect`-panicking when the producer drops. All callsites translate `None` into a recoverable `Error::io`, so a dead flusher manifests as a query error rather than a `RustPanic` from a worker task.

Adds a regression test (`test_task_dispatcher_survives_handle_error`).

## Note on root cause of #6713

After this PR was first opened I tracked the underlying \"Dataset already exists: …_gen_1\" error to **lancedb**, not Lance: lancedb's `database/listing.rs` was capturing `query_string = Some(\"\")` after `url::Url::query_pairs_mut()` accidentally set it on URLs that had no input query, which then appended a trailing `?` to every per-table URI. Sub-paths like `<table_uri>/_mem_wal/<shard>/<rand>_gen_<n>` re-parsed as `path=<base table>` + `query=/_mem_wal/...`, so `Dataset::write` resolved to the *base table* manifest and falsely reported \"already exists\". Fix is going up in lancedb separately.

The two fixes here remain valuable on their own — a transient flush error should not panic the writer or strand its waiters — and stand independent of the lancedb fix.

### DIFF
--- a/rust/lance/src/dataset/mem_wal/util.rs
+++ b/rust/lance/src/dataset/mem_wal/util.rs
@@ -71,14 +71,17 @@ impl<T: Clone + std::fmt::Debug> WatchableOnceCellReader<T> {
 
     /// Wait for a value to be written.
     ///
-    /// Returns immediately if a value is already present.
-    pub async fn await_value(&mut self) -> T {
-        self.rx
-            .wait_for(|v| v.is_some())
-            .await
-            .expect("watch channel closed")
-            .clone()
-            .expect("no value found")
+    /// Returns immediately if a value is already present. Returns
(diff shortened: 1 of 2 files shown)


jackye1995 (approved):
thanks for the fix!

PR #6717, opened 2026-05-08
feat(mem_wal): flush active memtable on ShardWriter::close

touch-of-grey:
## Summary

`ShardWriter::close()` previously drained the WAL but left the active memtable in memory at shutdown. Any rows that hadn't crossed `max_memtable_size`/`is_batch_store_full` triggers were durable in the WAL (recoverable on replay) but never produced a Lance fragment — so no LSM-level generation existed for them.

That's surprising semantics for callers who reasonably expect `close()` to make all data fully durable in the LSM sense, and it makes flush-cost benchmarks impossible: a workload that intentionally sets a huge `max_memtable_size` so puts measure pure steady-state throughput has no flush to time at close.

## Change

Make `close()` freeze the active memtable (if non-empty) and await every frozen memtable's flush completion before shutting the task handlers down. WAL drain still runs first, so WAL ordering is unchanged.

## Test

`test_close_flushes_active_memtable` drives 50 puts with `max_memtable_size = usize::MAX`, asserts no auto-flush fired during puts (precondition), then verifies `close()` completes without hanging. Without this change close returns promptly but no Lance generation exists; with it, close awaits the actual fragment write.

Full `cargo test -p lance --lib mem_wal::` is 241 passed, 1 ignored.

### DIFF
--- a/rust/lance/src/dataset/mem_wal/write.rs
+++ b/rust/lance/src/dataset/mem_wal/write.rs
@@ -1748,6 +1748,37 @@ impl ShardWriter {
                         let _ = reader.await_value().await;
                     }
                 }
+
+                // Freeze the active memtable (if any rows) so it joins the
+                // pending-flush queue, then wait for every frozen
+                // memtable's flush to complete. Without this, close() left
+                // any rows that hadn't crossed the size/batch trigger
+                // sitting in memory at shutdown — they were durable in the
+                // WAL but never produced a Lance fragment, which is
+                // surprising for callers who reasonably expect close() to
+                // make all data fully durable in the LSM sense (and which
+                // makes flush-cost benchmarks impossible to measure).
+                //
+                // Propagate any freeze error: at close time the caller
+                // has explicitly asked for full durability, so silently
+                // dropping a freeze failure would lose data without any
+                // signal. If freeze fails, surface the error rather than
+                // continuing on to drain only the pre-existing frozen
+                // memtables (whose flushes can still be waited on, but
+                // the caller now knows the close was incomplete).
+                let watchers: Vec<_> = {
+                    let mut st = state.write().await;
+                    if st.memtable.row_count() > 0 {
+                        writer_state.freeze_memtable(&mut st)?;
+                    }
+                    st.frozen_flush_watchers
+                        .iter()
+                        .map(|(_, w)| w.clone())
+                        .collect()
+                };
+                for mut w in watchers {
+                    let _ = w.await_value().await;
+                }
             }
             WriterMode::WalOnly {
                 state,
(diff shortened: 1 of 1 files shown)


hamersaw (commented):
Thanks for the PR! This looks great, ease of life things.

rust/lance/src/dataset/mem_wal/write.rs
```diff
@@ -1748,6 +1748,32 @@ impl ShardWriter {
...
+                // make all data fully durable in the LSM sense (and which
+                // makes flush-cost benchmarks impossible to measure).
+                let watchers: Vec<_> = {
+                    let mut st = state.write().await;
+                    if st.memtable.row_count() > 0 {
+                        // Errors here are logged inside freeze_memtable and
+                        // its message-send is fire-and-forget; we still
+                        // try to drain anything already frozen.
+                        let _ = writer_state.freeze_memtable(&mut st);
+                    }
```
hamersaw: The comment says that `freeze_memtable` logs errors internally, but I don't think that is accurate. So we are essentially dropping the error here without any visability. I think it makes sense to either (1) make sure we actually log at the "error" level or (2) capture and throw the error. API-wise I think it would make sense to have error thrown if I call `shard_writer.close()` and it is not successful.

rust/lance/src/dataset/mem_wal/write.rs
```diff
@@ -2678,6 +2704,57 @@ mod tests {
...
+            "no flush should have fired during puts (size threshold is usize::MAX)"
+        );
+        assert!(
+            stats_before.row_count > 0,
+            "memtable should hold the rows we just inserted"
+        );
+
+        // close() must flush the memtable into a Lance generation.
+        writer.close().await.unwrap();
+    }
```
hamersaw: I think this just validates that close completes. Should we a few steps further and validate that (1) close executions without errors (if we're throwing from `close` call), (2) the currently active memtable now has 0 rows, and (3) the L0 cached memtable (on-disk) exists?

rust/lance/src/dataset/mem_wal/write.rs
touch-of-grey: Good catch — fixed in the latest push. `freeze_memtable` does not log internally, so the comment was wrong and silently dropping the error was wrong. Reworked the code to propagate the error: `freeze_memtable(&mut st)?` from inside `close()`. Since `close()` already returns `Result<()>`, the freeze failure now surfaces to the caller, as you suggested.

rust/lance/src/dataset/mem_wal/write.rs
touch-of-grey: Beefed up the test. It now:

1. asserts `close()` returns `Ok` explicitly (`expect("close() must succeed and propagate any freeze/flush error")`),
2. reopens the same shard via a second `ShardWriter::open`,
3. reads the persisted manifest and asserts `manifest.current_generation > initial_gen` — direct evidence that close() drove a MemTable flush + manifest commit, not just dropped the rows on the floor.

I considered also asserting the reopened active memtable's `row_count == 0`, but that wedges into WAL-replay behavior (replay still rehydrates entries because `state.last_flushed_wal_entry_position` isn't updated as WAL flushes happen — it's only initialized from the manifest at open). That looks like a separate replay-coordination issue and `test_memtable_replay_*` is the right home for it. Happy to file a follow-up if you want it tracked.

touch-of-grey:
@hamersaw @jackye1995 updated the PR based on comments if you can take a look

touch-of-grey:
I will post another PR for all the hnsw improvements I am doing based on testing in the past few days

jackye1995 (approved):
thanks for the fix!

PR #6723, opened 2026-05-09
fix(encoding): support sparse boolean lists in full-zip encoding

### LINKED ISSUE #6681: Writing sparse boolean arrays using 2.1 and 2.2 format versions panics with `Non-byte aligned full-zip compression not yet supported`
When writing a large enough boolean array that happens to contain mostly nulls, using 2.1 (which is the default on current main) or newer encoding panics with the following error:
```
thread 'lance-cpu' (222474) panicked at <PATH>/rust/lance-encoding/src/encodings/logical/primitive.rs:4616:9:
assertion `left == right` failed: Non-byte aligned full-zip compression not yet supported
  left: 1
 right: 0
```

There are some behaviour differences between 2.1 and 2.2, but neither works correctly: writing nulls + only one specific value (true or false) passes on 2.2 and panics on 2.1, writing mixed null/true/false panics on both.

Here is a minimal repro:
```rust
use arrow::array::{
    ArrayRef, RecordBatch, RecordBatchIterator,
    builder::{BooleanBuilder, ListBuilder},
};
use arrow::datatypes::{DataType, Field, Schema};
use lance::dataset::WriteParams;
use lance_encoding::version::LanceFileVersion;
use std::sync::Arc;

#[tokio::test]
async fn write_sparse_array() {
    const NUM_ROWS: usize = 1000;

    let inner = Arc::new(Field::new("item", DataType::Boolean, true));
    let schema = Arc::new(Schema::new(vec![Field::new(
        "lb",
        DataType::List(inner.clone()),
        true,
    )]));

    let mut builder = ListBuilder::new(BooleanBuilder::new()).with_field(inner);
    for i in 0..NUM_ROWS {
        if i % 64 == 0 {
            // Writing only one static value is broken on 2.1, works on 2.0 and 2.2
            // let value = true;
            
            // Writing a dynamic value is broken on both 2.1 and 2.2
            let value = i % 128 == 0;

            builder.values().append_value(value);
            builder.append(true);
        } else {
            builder.append(true);
        }
    }
    let arr: ArrayRef = Arc::new(builder.finish());

    let tmp = tempfile::tempdir().expect("tempdir");
    let path = tmp.path().join("data.lance");
    let path_str = path.to_string_lossy().into_owned();

    let batch = RecordBatch::try_new(schema.clone(), vec![arr]).unwrap();
    let reader = RecordBatchIterator::new(std::iter::once(Ok(batch)), schema);

    let params = WriteParams {
        data_storage_version: Some(LanceFileVersion::V2_1),
        ..WriteParams::default()
    };

    lance::Dataset::write(reader, &path_str, Some(params))
        .await
        .unwrap();
}
```

This test technically passes with lance 3.0.1, and I've bisected the error to commit https://github.com/lance-format/lance/commit/3b2d632b5a5269e2a18d73e5b3fcd306a3a8dc2a, but without that fix other things are broken.

shenganzhang:
## Summary

Fixes #6681.

Writing a `list<bool>` column where most rows are null (sparse) panicked with:

```
Non-byte aligned full-zip compression not yet supported
```

The root cause: #6234 added a fallback from miniblock to full-zip encoding when rep/def levels are too sparse for a miniblock chunk. The full-zip path assumed all values are byte-aligned, but booleans are stored as 1-bit packed values in Arrow — triggering the assert in `serialize_full_zip_fixed`.

**Fix (two small, paired changes):**

- **Encoder** (`encode_full_zip`): when the data block is 1-bit boolean, expand each bit to a full byte before compression. A new helper `expand_boolean_to_bytes` uses `BooleanBuffer::value()` to read bits correctly.
- **Decoder** (`FixedWidthDataBlock::do_into_arrow`): when a `Boolean` field is decoded with `bits_per_value == 8` (the expanded representation), re-pack the bytes back to Arrow's bit-packed format before building the `ArrayData`.

## Test plan

- Added `test_sparse_boolean_list_roundtrip` — 1000-row `list<bool>` with ~1/64 non-null rows and mixed `true`/`false` values; exercises both V2.1 and V2.2 paths. Test was failing before the fix (panic) and passes after.
- Full `lance-encoding` test suite: 370 tests, 0 failures.


hamersaw:
I believe this is the same fix as https://github.com/lance-format/lance/pull/6777/. @Xuanwo maybe you can elaborate on why this is not a great fix and the other proposed solution?

westonpace (approved):
Hi, thanks for looking into this!  I'd rather not fallback to full-zip in these cases (see https://github.com/lance-format/lance/issues/6841) but this is a good workaround for the current solution.

westonpace:
With https://github.com/lance-format/lance/pull/6723 we probably won't hit this case anymore but I think it might still be nice to merge this in for completeness sake (a user could always opt-in to full-zip encoding with boolean data for some reason)

Xuanwo:
> I believe this is the same fix as #6777. @Xuanwo maybe you can elaborate on why this is not a great fix and the other proposed solution?

Yes, this is the same fix as #6777. I don't like those two because I believe those two PRs are just workaround that trying to address this issue at the wrong level.

So I proposed the change at https://github.com/lance-format/lance/pull/6787. However, I did agree that https://github.com/lance-format/lance/pull/6787 is much more complex than this one thus need more time for reviewing and testing.

I second with @westonpace's judgement and decided to merge this in first and give #6787 more time.

PR #6726, opened 2026-05-10
refactor(mem_wal): redesign FTS mem index for single-writer multi-reader

touch-of-grey:
## Summary

Replaces the `SkipMap<(token, row_position)>` postings layout in
`rust/lance/src/dataset/mem_wal/index/fts.rs` with per-term
`ArcSwap<TermSlice>` slices and an `ArcSwap<Snapshot>`-published
per-batch visibility watermark. Readers grab the snapshot, walk per-term
chunks filtered by `chunk.batch_position < snapshot.visible_count`, and
score with snapshot-coupled BM25 stats.

The writer publishes the snapshot only after every term chunk for a
batch is linked, so readers never observe a partial document or BM25
stats out of sync with the postings — per-batch monotonic visibility.

Also in this PR:

- Tokenize-time tokenizer ownership moves out of a single shared `Mutex`
  into a small reader pool plus a writer-dedicated slot, so search
  calls no longer serialize against the writer.
- `Utf8View` is added to the supported text types alongside `Utf8` /
  `LargeUtf8`; non-string columns now produce a clear error instead of
  being silently skipped.
- `FtsMemIndex::memory_usage()` for size-based flush triggers.
- Compound queries (`Boolean`, `Boost`) snapshot once at `search_query`
  and thread the same `Arc<Snapshot>` through every leaf, so a
  compound query never mixes BM25 stats from different snapshots.
- The flush path keeps using `lance_index::scalar::inverted` types
  (`TokenSet` / `DocSet` / `PostingListBuilder`); the on-disk format is
  unchanged. The stale "flush is not yet implemented" header doc is
  corrected.

Public API (`FtsMemIndex` constructors, `FtsQueryExpr`, `SearchOptions`,
`BooleanQueryBuilder`, `FtsIndexConfig`, `to_index_builder_reversed`)
is preserved. No callers of the removed internal `FtsKey` /
`PostingValue` types existed outside the file.

cc @jackye1995 for review.


touch-of-grey:
## FTS mem index — performance & memory update

Beyond the original single-writer/multi-reader redesign, this branch now also
makes the in-memory FTS index compact and fast. Summary of what landed and how
it benchmarks.

### What changed

- **Partition-structured index** — immutable `Partition`s plus a bounded
  mutable tail, published together behind one `ArcSwap`; the tail freezes into
  a partition past a threshold. Replaces the per-batch-chunk layout that made
  queries O(corpus).
- **Compressed posting lists** — 128-doc blocks; doc ids and frequencies
  bit-packed (SIMD `BitPacker4x` for full blocks); positions stored as a
  per-block bit-packed delta stream with random per-document access (a
  document's position count equals its term frequency, so it is located from
  the frequency prefix sum — no count is stored).
- **WAND top-k** with a shared, cross-partition pruning threshold; shared
  term-dictionary interner.
- Flush still converts to the on-disk Lance FTS format unchanged.

### Benchmark (real FineWeb corpus, 150 queries, k=10)

Against the pre-redesign mem index, at 1M documents:

| metric | before | after | improvement |
| --- | --- | --- | --- |
| query p50 | 7 303 µs | 1 121 µs | 6.5x |
| query p95 | 1 175 654 µs | 7 849 µs | **150x** |
| memory | 10.21 GB | 1.37 GB | 7.5x |
| term recall@10 | 1.000 | 1.000 | exact |

Against Apache Lucene (same corpus and queries) as a reference point:

| docs | mem index p50 / p95 | Lucene p50 / p95 | mem index / Lucene memory |
| --- | --- | --- | --- |
| 100k | 110 / 1 465 µs | 206 / 560 µs | 0.14 / 0.09 GB |
| 500k | 560 / 5 595 µs | 208 / 3 589 µs | 0.68 / 0.42 GB |
| 1M | 1 121 / 7 849 µs | 261 / 4 397 µs | 1.37 / 0.84 GB |

Split by query type at 1M docs:

| | term (100 q) | phrase (50 q) |
| --- | --- | --- |
| p50 / p95 | 1 039 / 1 186 µs | 1 359 / 14 245 µs |

Term queries — the dominant FTS workload — are at or ahead of Lucene on
latency (term p95 ~1.2 ms vs Lucene's blended p95 ~4.4 ms), ranking is exact
(recall@10 = 1.0), and memory is ~1.6x Lucene.

### Known follow-up

The blended p95 is dominated by phrase queries (~14 ms at 1M, ~3x Lucene).
`search_phrase` scans the rare driving term's full posting list with no top-k
pruning. Closing that gap needs phrase-specific top-k pruning and is best
done as a separate change; term latency, recall, and memory are competitive.

rust/lance/benches/mem_wal/fts/LuceneFtsBench.java
jackye1995, line 1: we should not create a new bench folder, these should be mved under rust/lance/benches. We should probably create a mem_wal folder with specific subfolders to organize different benchmarks at this point

jackye1995 (commented):
mostly looks good to me, thanks for iterating this with me, just a final comment regarding the file layout

rust/lance/benches/mem_wal/fts/LuceneFtsBench.java
touch-of-grey: Done. Moved all the FTS bench files out of the top-level `bench/` folder into `rust/lance/benches/mem_wal/`, with a `fts/` subfolder for FTS-specific work:

```
rust/lance/benches/mem_wal/
  run_shard_writer_backpressure.sh
  fts/
    mem_wal_fts_bench.rs
    mem_wal_fineweb_fts.rs
    LuceneFtsBench.java
    run_fts_compare.sh
    run_fineweb_fts.sh
```

`Cargo.toml`'s `[[bench]]` entries now point at the new paths via `path = ...`. I kept this scoped to the bench files this PR introduces — the pre-existing flat `mem_wal_*.rs` benches are left in place to avoid an unrelated churn diff; happy to do a follow-up that consolidates those into `mem_wal/` too.

Also scrubbed the driver scripts and bench doc comments of infra-specific references (S3 bucket names, instance paths, AWS region exports) — they now default to tmpdir-based paths and resolve the repo root via `git rev-parse`.

jackye1995 (approved):
looks good to me!

PR #6731, opened 2026-05-11
fix(lsm): normalize vector search plan schemas before UnionExec to prevent DataFusion panic

zhangyue19921010:
closes:https://github.com/lance-format/lance/issues/6730

### DIFF
--- a/rust/lance/src/dataset/mem_wal/scanner/vector_search.rs
+++ b/rust/lance/src/dataset/mem_wal/scanner/vector_search.rs
@@ -14,6 +14,7 @@ use datafusion::physical_expr::expressions::Column;
 use datafusion::physical_expr::{LexOrdering, PhysicalSortExpr};
 use datafusion::physical_plan::ExecutionPlan;
 use datafusion::physical_plan::limit::GlobalLimitExec;
+use datafusion::physical_plan::projection::ProjectionExec;
 use datafusion::physical_plan::sorts::sort::SortExec;
 use datafusion::physical_plan::union::UnionExec;
 use lance_core::Result;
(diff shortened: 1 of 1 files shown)


zhangyue19921010:
Minor Bug related with LSM. Would u mind to take a look when u have time? Thanks @jackye1995 ~

rust/lance/src/dataset/mem_wal/scanner/vector_search.rs
```diff
@@ -187,10 +188,18 @@ impl LsmVectorSearchPlanner {
             let knn = self
                 .build_knn_plan(source, query_vector, k, nprobes, projection)
                 .await?;
-            let tagged: Arc<dyn ExecutionPlan> = Arc::new(MemtableGenTagExec::new(knn, generation));
+            let Some(normalized) = Self::normalize_knn_schema(knn)? else {
+                continue;
```
Xuanwo, line 192: Continuing here silently drops active memtables that do not yet have an HNSW index, so newly written rows can disappear from LSM vector search results until they are indexed or flushed.

rust/lance/src/dataset/mem_wal/scanner/vector_search.rs
```diff
@@ -229,6 +238,60 @@ impl LsmVectorSearchPlanner {
...
+        let schema = plan.schema();
+
+        // Skip plans that have no _distance column – they cannot be ranked.
+        if schema.field_with_name(DISTANCE_COLUMN).is_err() {
+            return Ok(None);
+        }
+
+        // If there is no _rowid column the schema is already normalised.
+        if schema.field_with_name(lance_core::ROW_ID).is_err() {
+            return Ok(Some(plan));
```
Xuanwo, line 270: Returning no-`_rowid` plans unchanged still leaves active HNSW memtables with only the user projection while other sources add required primary keys, so projections that omit a primary key can still hit the UnionExec schema mismatch this PR is trying to prevent.

zhangyue19921010:
Closing — this issue no longer exists on current `main`, which fixed both root causes structurally:     
                                                                                                          
  1. An active memtable without an HNSW index now routes to `MemTableBruteForceVectorExec` (always        
  produces `_distance`) instead of degrading to a plain scan, so fresh rows are searched rather than      
  dropped.                                                                                                
  2. Every source arm is projected to a canonical output schema (PKs auto-appended, missing system columns
  NULL-filled) before `UnionExec`, so the schema mismatch can no longer occur.                            
                                                                                                          
  Both are covered by regression tests on `main`. Thanks @Xuanwo for the review.

PR #6732, opened 2026-05-11
chore: remove unused Cargo dependencies

Xuanwo:
This cleans up direct Cargo dependency declarations that are no longer referenced across the Rust workspace and the Python/JNI crates.

The generated-code and macro-only dependencies that still need to be declared are kept explicitly ignored for cargo-machete so future scans do not remove required dependencies.


Xuanwo:
@claude review once

BubbleCal (approved):

PR #6733, opened 2026-05-11
feat: add Git-style delta encoding for blob columns

beinan:
## Summary

- Implements binary delta encoding for `LargeBinary` blob columns, inspired by Git's packfile delta format
- When consecutive rows contain similar content (e.g., successive versions of source files), only the differences are stored via copy/insert instructions
- Builds on top of existing blob encoding infrastructure using external buffers — no changes to page structure

## Design

- **Delta algorithm**: Rabin rolling hash (16-byte windows) indexes the base buffer; the target is scanned with the same hash to find matching regions. Matches ≥4 bytes become copy instructions, unmatched regions become insert instructions. Git-compatible opcode format.
- **Delta groups**: Consecutive values are grouped with configurable chain depth (default 4). First value stored as-is (base), subsequent values as chained deltas against the previous value. A new base is emitted when the chain depth is exceeded or when a delta is larger than the raw value.
- **Descriptor**: Extended struct with `(position, size, kind, base_offset)` — `kind` distinguishes `DeltaBase` vs `Delta`, `base_offset` stores distance back to the base.
- **Decoding**: `DeltaBlobPageScheduler` expands requested ranges to include required bases, loads all data, applies delta chains, then returns only the requested rows.

## Files Changed

| Component | Files |
|-----------|-------|
| Delta algorithm | `rust/lance-encoding/src/encodings/physical/delta.rs` (new) |
| BlobKind variants | `rust/lance-core/src/datatypes.rs`, `field.rs` |
| Metadata key | `rust/lance-arrow/src/lib.rs` |
| Encoder | `rust/lance-encoding/src/encodings/logical/blob.rs` |
| Decoder | `rust/lance-encoding/src/encodings/logical/primitive/blob.rs` |
| Protobuf | `protos/encodings_v2_1.proto` |
| Wiring | `encoder.rs`, `decoder.rs`, `primitive.rs`, `format.rs`, `testing.rs` |

## Usage

Set field metadata `lance-encoding:delta-blob = "true"` on a `LargeBinary` column.

## Test plan

- [x] Delta algorithm unit tests (9 tests): varint roundtrip, identical buffers, small edits, larger files, chained deltas, source code diffs
- [x] Delta blob round-trip tests (4 tests): similar source code versions, nulls, completely different data, larger files
- [x] All existing blob tests pass (9 tests) — no regressions

### DIFF
--- a/protos/encodings_v2_1.proto
+++ b/protos/encodings_v2_1.proto
@@ -194,6 +194,18 @@ message BlobLayout {
   repeated RepDefLayer layers = 2;
 }
 
+// A layout where large binary data is delta-encoded against a base value.
+// The descriptions contain (position, size, kind, base_offset) tuples.
+// kind indicates whether the value is a base (stored as-is) or a delta
+// (stored as copy/insert instructions relative to the previous value in the chain).
+// base_offset is the row distance back to the base value in the group.
+message DeltaBlobLayout {
+  // The inner layout used to store the descriptions
+  PageLayout inner_layout = 1;
+  // The meaning of each repdef layer
+  repeated RepDefLayer layers = 2;
+}
+
 // Describes the structural encoding of a page
 message PageLayout {
   oneof layout {
@@ -206,6 +218,8 @@ message PageLayout {
     // A layout where large binary data is encoded externally
     // and only the descriptions are put in the page
     BlobLayout blob_layout = 4;
+    // A layout where large binary data is delta-encoded
+    DeltaBlobLayout delta_blob_layout = 5;
   }
 }
 
--- a/rust/lance-arrow/src/lib.rs
+++ b/rust/lance-arrow/src/lib.rs
@@ -47,6 +47,8 @@ pub const ARROW_EXT_META_KEY: &str = "ARROW:extension:metadata";
 /// Key used by lance to mark a field as a blob
 /// TODO: Use Arrow extension mechanism instead?
 pub const BLOB_META_KEY: &str = "lance-encoding:blob";
+/// Metadata key to enable delta encoding for blob columns
+pub const DELTA_BLOB_META_KEY: &str = "lance-encoding:delta-blob";
 /// Arrow extension type name for Lance blob v2 columns
 pub const BLOB_V2_EXT_NAME: &str = "lance.blob.v2";
 /// Metadata key for overriding the dedicated blob size threshold (in bytes)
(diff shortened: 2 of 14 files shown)


protos/encodings_v2_1.proto
```diff
@@ -194,6 +194,18 @@ message BlobLayout {
   repeated RepDefLayer layers = 2;
 }
 
+// A layout where large binary data is delta-encoded against a base value.
+// The descriptions contain (position, size, kind, base_offset) tuples.
+// kind indicates whether the value is a base (stored as-is) or a delta
+// (stored as copy/insert instructions relative to the previous value in the chain).
+// base_offset is the row distance back to the base value in the group.
+message DeltaBlobLayout {
```
Xuanwo, line 202: To add a new encoding layout, a vote from the PMC is required. Could you start a discussion about this first?

beinan: Started a PMC vote discussion: https://github.com/lancedb/lance/discussions/6736

westonpace: What is `base_offset` when the value is not a delta (had no similar ancestor)?  Is it always 0 or is it undefined?

rust/lance-arrow/src/lib.rs
```diff
@@ -47,6 +47,8 @@ pub const ARROW_EXT_META_KEY: &str = "ARROW:extension:metadata";
 /// Key used by lance to mark a field as a blob
 /// TODO: Use Arrow extension mechanism instead?
 pub const BLOB_META_KEY: &str = "lance-encoding:blob";
+/// Metadata key to enable delta encoding for blob columns
+pub const DELTA_BLOB_META_KEY: &str = "lance-encoding:delta-blob";
```
westonpace, line 51: I think opt-in is fine to start with but do you think in the future we might be able to detect scenarios where delta blob encoding would be advantageous?  Maybe at compaction time or sometime?

protos/encodings_v2_1.proto
beinan: When the value is a base (i.e., its `kind` is `DeltaBase`), the `base_offset` is set to 0 and is ignored during decoding. The decoder relies entirely on the `BlobKind` to know whether to use the offset to resolve an ancestor, or to just treat the current bytes as the full raw blob.

rust/lance-arrow/src/lib.rs
beinan: Yes, definitely! Once the base format and implementation are stable, we can add heuristics to auto-detect when it's beneficial. For example, during compaction or even initial writes, the encoder could sample the first few blobs, attempt a delta calculation, and automatically switch to delta encoding if the storage savings cross a certain threshold (e.g., >20% reduction). For this initial PR, keeping it strictly opt-in helps reduce the surface area and avoid unexpected regressions.

PR #6738, opened 2026-05-12
perf(index): compact nullable vector samples in parallel

hfutatzhanghb:
## Summary

Parallelizes nullable FixedSizeList training-data compaction by moving per-batch null filtering and value extraction onto the CPU worker pool.

The range reads still rely on take_scan batch readahead for I/O concurrency, while the final training buffer is assembled sequentially to avoid shared mutable state and keep batch order stable.

## Testing

- cargo fmt --all
- cargo check -p lance --tests

### DIFF
--- a/rust/lance/src/index/vector/utils.rs
+++ b/rust/lance/src/index/vector/utils.rs
@@ -13,6 +13,7 @@ use arrow_buffer::{Buffer, MutableBuffer};
 use futures::{Stream, StreamExt, stream};
 use lance_arrow::DataTypeExt;
 use lance_core::datatypes::Schema;
+use lance_core::utils::tokio::{get_num_compute_intensive_cpus, spawn_cpu};
 use lance_linalg::distance::DistanceType;
 use log::{info, warn};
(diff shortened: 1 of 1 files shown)


rust/lance/src/index/vector/utils.rs
```diff
@@ -688,42 +738,52 @@ async fn sample_nullable_fsl<S>(
...
+    let column = Arc::<str>::from(column);
+    let mut compacted_batches = scan
+        .map(|batch| {
+            let column = column.clone();
+            spawn_cpu(move || {
+                let batch = batch?;
+                compact_nullable_fsl_batch(batch, &column, byte_width)
+            })
+        })
+        .buffered(compact_readahead.max(1));
```
Xuanwo, line 760: The nullable fragment sampling path now prefetches and compacts up to the CPU-thread count of batches even though each batch can already contain twice the requested sample size. Wide vectors or large training samples can multiply the memory footprint and reintroduce the OOM risk that the fragment-filtered nullable path was designed to avoid.

PR #6739, opened 2026-05-12
fix: scalar index config plugin lookup to be case-insensitive

zhangyue19921010:
```
@pytest.mark.parametrize("index_type", ["ZONEMAP", "ZoneMap"])
def test_zonemap_index_config_is_case_insensitive(tmp_path: Path, index_type: str):
    ds = lance.write_dataset(pa.table({"x": range(1024)}), tmp_path)

    ds.create_scalar_index(
        "x",
        IndexConfig(index_type=index_type, parameters={"rows_per_zone": 64}),
    )

    scanner = ds.scanner(filter="x = 7", prefilter=True)
    assert "ScalarIndexQuery" in scanner.explain_plan()
    assert scanner.to_table().num_rows == 1
```

This test will be failed because of case sensitive
```
no scalar index plugin found for name "ZONEMAP"
```

### DIFF
--- a/rust/lance-index/src/registry.rs
+++ b/rust/lance-index/src/registry.rs
@@ -21,8 +21,12 @@ pub struct IndexPluginRegistry {
 }
 
 impl IndexPluginRegistry {
+    fn normalize_plugin_name(name: &str) -> String {
+        name.to_lowercase()
+    }
+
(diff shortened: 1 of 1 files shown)


wjones127 (approved):
This seems reasonable. I think we don't want to have index names that vary only by case, so matching with case insensitivity makes sense.

PR #6740, opened 2026-05-12
fix(rust): avoid pushdown limit/offset for stable row id

yanghua:
When we enable stable row ID, if we update rows, we can not scan updated rows with a limit condition.

image

image


Xuanwo (approved):
Thank you!

PR #6742, opened 2026-05-12
perf(rowids): prune and lazily load fragment metadata for stable row ID take

### LINKED ISSUE #6486: Stable row ID takes require up to 3x as many IOPs as non-stable
Using `take` operations over a multimodal dataset with stable row IDs in batches of 10 rows can take 2-3 IOPs per row. This is very large when compared to non-stable rows, where 0.6-0.8 IOPs per row is the norm (coalescing reads). e2e, this contributes to a 2.6x slowdown when using stable row IDs.

Intuitively, the stable row ID `take` pattern builds a `_rowid` -> `_row_addr` mapping and uses that translate `takes` over stable row IDs into pointed row address lookups. It shouldn't need significantly more than non-stable.

hamersaw:
Currently, when using stable row IDs a `take` operation builds a row ID to row address mapping for every fragment within the dataset and then uses this to look up the requested row IDs. The main performance issue with this is that depending on the query pattern there can be a large amount of unnecessary overhead. For example, a user opens a dataset and submits a single take with 100 rows. If the dataset has 5k fragments we only need to parse at most 100 of those fragments, the other 4999 are overhead.

In this PR, we propose (1) lazily loading fragment row ID to row address mappings and (2) pruning unloaded fragments based on `<start, end>` values within fragment row ID ranges:
- **Pruning fragments**: Within the dataset manifest for each fragment we store a `Vec` of row sequences. These can be simple ranges (start / end), bitmaps, etc. But they all have a min / max value for the row ID range. We can use this value, available very cheaply at query time, to decide whether we actually want to load that fragments metadata. The loading "phase" involves parsing the deletion vector in addition to actually parsing the fragments row sequences so it can be expensive.
- **Lazily loading fragments:** If we decide based on the start / end row ID that a fragment may contain the row ID we load it into the `RowIndex`, which is an appendable struct. The next time a `take` comes it, it checks the in-memory `RowIndex` before checking unloaded fragments. 

In the worst case scenario (ie. loading every fragment) the performance should be similar (same # of IOPs + slight CPU overhead for additional validation), but in cases where every fragment doesn't need to be loaded we are able to mitigate all unnecessary overhead.

Closes https://github.com/lance-format/lance/issues/6486


hamersaw:
Preliminary results:
```
  Results look right. Reading the numbers:

  ┌────────────┬───────────┬───────────┬─────────────┐
  │            │ cold IOPs │ warm IOPs │ repeat IOPs │
  ├────────────┼───────────┼───────────┼─────────────┤
  │ stable     │ 2         │ 2         │ 1           │
  ├────────────┼───────────┼───────────┼─────────────┤
  │ non-stable │ 2         │ 1         │ 1           │
  └────────────┴───────────┴───────────┴─────────────┘

  Cold case is the win we wanted: stable matches non-stable at 2 IOPs. Pre-lazy this would have been ~10+ IOPs (one per fragment to build the full rowid index). Single-fragment take
  now only pays for fragment 0's row_id_sequence + data page.

  Repeat matches non-stable at 1 IOP — second identical take is fully cached (per-fragment RowIdSequenceKey + new RowIdIndexKey(version, fragment_set_hash) both hit).

  Warm shows the lazy tradeoff: prewarming a take in fragment 5 doesn't preload fragment 0's row_id_sequence anymore (it would have under the old eager path). So when the measured
  take touches fragment 0, it pays 1 IOP for that sequence. That's the inherent cost of lazy — fewer cold IOPs in exchange for paying per-fragment-first-touch instead of
  all-fragments-upfront.
```

PR #6748, opened 2026-05-12
fix(java): expose updatedFragmentOffsets on Update operation for RewriteColumns

jerryjch:
## Summary

* Follow-up to #6650 (fix: propagate update_columns offsets and partial last_updated for
  RewriteColumns). 
* `Update.java`: adds `Map<Long, byte[]> updatedFragmentOffsets` field, 7-arg constructor,
  accessor `updatedFragmentOffsets()`, and `Builder.updatedFragmentOffsets(...)` setter.
  Defaults to `Collections.emptyMap()`. Values are portable RoaringBitmap bytes.
* `java/lance-jni/src/transaction.rs` — two JNI directions updated: FromJava deserializes
  each `byte[]` value into a `RoaringBitmap` and sets `updated_fragment_offsets` on the Rust
  operation; IntoJava serializes each bitmap to `byte[]` and populates a `HashMap<Long, byte[]>`
  passed to the 7-arg `Update` constructor (previously the field was ignored and the 6-arg
  form was used).
* `Update.java` `equals` / `hashCode`: deep-compares `byte[]` values by content; `hashCode`
  added per the Java contract.
* Required by [lance-spark#418](https://github.com/lance-format/lance-spark/pull/528)

## Background

PR #6650 added `updated_fragment_offsets` on the Rust `Operation::Update` (proto field 9),
`build_manifest` partial refresh logic, and `FragmentUpdateResult.getUpdatedRowOffsets()`.
Two gaps remained:

1. The Java `Update` class had no field for these offsets and `convert_to_rust_operation`
   always set `updated_fragment_offsets: None`, so the lance-spark commit path
   (UpdateColumnsBackfillBatchWrite) had no way to pass offsets to Rust and the partial
   refresh in `build_manifest` could never activate from a JVM caller.

2. `convert_to_java_operation_inner` still used the old 6-arg constructor signature for
   `new_object`. With the 6-arg constructor removed from `Update.java` (replaced by the
   7-arg form), any Rust→Java materialization of `Operation::Update` (e.g. reading back a
   transaction) would fail at runtime with `NoSuchMethodError`.

## Implementation notes                                                                                                                   
   
* Values are portable RoaringBitmap bytes (little-endian, spec-compliant). The JNI boundary                                               
    stays O(bitmap size) rather than O(n matched rows).
* `with_local_frame(4, ..)` per bitmap entry in IntoJava bounds local-ref growth on large                                                 
    offset maps. `JMap` was avoided inside the frame because it holds a `JObject` with the
    outer frame's lifetime, causing borrow-checker conflicts; `call_method` on the outer                                                    
    `java_map` reference is used instead.
* The `Vec<u8>` buffer for each bitmap is allocated in Rust before entering the frame, so                                                 
    its lifetime is independent of JNI frame scope.                                                                                         
* `with_local_frame(8, ..)` per iteration in FromJava bounds local-ref growth for large
    multi-fragment maps. 
* `build_manifest` validates bitmap cardinality and max offset against the fragment's
  `physical_rows` from `existing_fragments` before `.collect()`, preventing a compact RLE
  bitmap from expanding into an unbounded allocation.                                                                                                                
* `UpdatedFragmentOffsets` added to the `lance::dataset::transaction` import.

## Why the protobuf field alone is not enough

lance-spark commits by calling `CommitBuilder.execute(transaction)`, which passes the Java
`Transaction` object to `nativeCommitToDataset` via JNI. The JNI handler calls
`convert_to_rust_transaction` → `convert_to_rust_operation`, which reflects on the Java
`Update` object to build the Rust `Operation::Update` struct. The protobuf field (field 9)
is only used when a Transaction is serialized as a proto blob; it has no effect on the
reflection-based JNI path unless the Java `Update` class exposes the field and the JNI
deserialization reads it.

## Additional change

`FragmentUpdateResult` (from #6650) returned matched row offsets as an expanded `long[]` at
the executor JNI boundary. This PR also passes those offsets as portable RoaringBitmap bytes
so lance-spark can wire them through to `Update.updatedFragmentOffsets()` without an O(n rows)
expansion on the executor→driver path.

* `FragmentUpdateResult.getUpdatedRowOffsetBytes()` — primary accessor; values are the same
  portable RoaringBitmap byte format as `Update.updatedFragmentOffsets()`.
* `java/lance-jni/src/fragment.rs` — `update_columns_with_offsets` serializes
 `matched_offsets` once with `RoaringBitmap::serialize_into`; JNI constructs results via                                              
 the private `(FragmentMetadata, long[], byte[])` constructor (JNI can access private ctors).                                         
 `FragmentUpdateResult.create(FragmentMetadata, long[], byte[])` — public static factory;                                             
 primary construction path for callers using the bytes API.
* `@Deprecated getUpdatedRowOffsets()` — retained for backward compatibility; expands bytes via
  `expandRowOffsetsFromBytes` only when called (lazy O(n rows)).
* `@Deprecated` 3-arg `(FragmentMetadata, long[] fieldsModified, long[] updatedRowOffsets)`
  constructor — encodes offsets via `encodeRowOffsetsToBytes` for source compat.
* `FragmentUpdateResultTest` — round-trip bytes, deprecated constructor encode, and
  `updateColumns()` integration asserting matched offsets `{0,1,2,3}` on the test fixture.

## Test plan

* `UpdateTest#testUpdatedFragmentOffsetsRoundTrip` — commits an `Update` with a non-empty
  `updatedFragmentOffsets` map through `CommitBuilder.execute` (exercises the FromJava JNI
  path), reads the transaction back via `Dataset.readTransaction()` (exercises the IntoJava
  JNI path), and asserts the offsets match. Map value is hardcoded portable RoaringBitmap
  bytes encoding {1, 3, 5}; verified with `assertArrayEquals` after the round-trip.
* `FragmentUpdateResultTest` — see [Additional change](#additional-change).

## Compatibility

* Additive new API on `Update` — the `updatedFragmentOffsets` field did not exist in any
  prior release. The builder setter is optional and defaults to `Collections.emptyMap()`, so
  existing `Update.builder()...build()` call sites compile and behave identically.
* Java — `equals` / `hashCode`: `equals` uses `offsetMapsEqual` to deep-compare `byte[]`
  values via `Arrays.equals`; `hashCode` is added per the Java contract.
* JNI — constructor signature: the IntoJava `new_object` call is updated from the 6-arg to
  the 7-arg form in the same PR. Both files must ship together; within that atomic change
  there is no compatibility gap.
* Rust / proto: no changes. The `updated_fragment_offsets` proto field and Rust struct
  field were already added in #6650.
* `FragmentUpdateResult` — `@Deprecated` long[] getter and constructor retained; new bytes getter is the supported path for new callers   
  (see [Additional change](#additional-change)).


jerryjch:
cc @jackye1995 @hamersaw

java/lance-jni/src/transaction.rs
```diff
@@ -1232,6 +1268,39 @@ fn convert_to_rust_operation(
...
+                    )?
+                    .l()?;
+                if offsets_obj.is_null() {
+                    None
+                } else {
+                    let jmap = JMap::from_env(env, &offsets_obj)?;
+                    let mut iter = jmap.iter(env)?;
+                    let mut offsets: HashMap<u64, RoaringBitmap> = HashMap::new();
+                    env.with_local_frame(32, |env| {
+                        while let Some((key, value)) = iter.next(env)? {
```
Xuanwo: The Java-to-Rust offset map import keeps all iterator-created JNI local references alive for the whole map. Large multi-fragment updates can exhaust the local reference table and fail the commit before Rust validation runs.

java/lance-jni/src/transaction.rs
```diff
@@ -1232,6 +1268,39 @@ fn convert_to_rust_operation(
...
+                            let frag_id =
+                                env.call_method(&key, "longValue", "()J", &[])?.j()? as u64;
+                            let buf: Vec<u8> = env.convert_byte_array(JByteArray::from(value))?;
+                            let bitmap = RoaringBitmap::deserialize_from(buf.as_slice())?;
```
Xuanwo: Malformed bitmap bytes are reported as I/O failures even though they come from Java API input. Callers can misclassify bad arguments as retryable storage errors.

jerryjch:
@Xuanwo Thanks for the review. I addressed your comments in commit `fix(java): bound JNI locals and classify bad offset bitmap bytes as
  input errors`. In addition, please check the [Additional change](#additional-change) added in commit `perf(java): return matched row offsets as Roaring bytes from FragmentUpdateResult`.

java/src/main/java/org/lance/fragment/FragmentUpdateResult.java
```diff
@@ -26,19 +26,37 @@ public class FragmentUpdateResult {
...
+        updatedRowOffsetBytes != null ? updatedRowOffsetBytes : new byte[0];
+  }
+
+  /**
+   * @deprecated Use {@link #getUpdatedRowOffsetBytes()} instead. This method expands serialized
+   *     RoaringBitmap bytes to a {@code long[]} via JNI and is retained for backward compatibility
+   *     with callers compiled against the #6650 API.
+   */
+  @Deprecated
+  public FragmentUpdateResult(
```
Xuanwo, line 66: The retained long[] compatibility path now calls native helpers from a class that never loads the JNI library. Existing callers can hit UnsatisfiedLinkError unless another Lance class happened to initialize first.

java/lance-jni/src/transaction.rs
```diff
@@ -1238,6 +1278,59 @@ fn convert_to_rust_operation(
...
+                            8,
+                            |env| -> Result<Option<(u64, RoaringBitmap)>> {
+                                let Some((key, value)) = iter.next(env)? else {
+                                    return Ok(None);
+                                };
+                                let frag_id =
+                                    env.call_method(&key, "longValue", "()J", &[])?.j()? as u64;
+                                let buf: Vec<u8> =
+                                    env.convert_byte_array(JByteArray::from(value))?;
+                                let bitmap = RoaringBitmap::deserialize_from(buf.as_slice())
```
Xuanwo, line 1339: The commit path accepts arbitrary offset bitmaps and later expands them into a full offset vector before checking fragment bounds. A compact valid Roaring bitmap can force huge allocations during a Java RewriteColumns commit.

java/src/main/java/org/lance/fragment/FragmentUpdateResult.java
```diff
@@ -26,19 +26,37 @@ public class FragmentUpdateResult {
...
+   */
+  private final byte[] updatedRowOffsetBytes;
 
   /** Two-argument form for callers that do not track per-row offsets; offsets default to empty. */
   public FragmentUpdateResult(FragmentMetadata updatedFragment, long[] updatedFieldIds) {
-    this(updatedFragment, updatedFieldIds, new long[0]);
+    this(updatedFragment, updatedFieldIds, new byte[0]);
   }
 
   public FragmentUpdateResult(
```
Xuanwo: Adding a public byte[] constructor with the same arity as the retained long[] constructor makes existing source calls that pass null for offsets fail overload resolution.

java/lance-jni/src/fragment.rs
```diff
@@ -539,15 +541,106 @@ fn inner_update_column<'local>(
...
+        env.get_long_array_region(&joffsets, 0, buf.as_mut_slice())?;
+    }
+    let mut bitmap = RoaringBitmap::new();
+    for offset in buf {
+        if offset < 0 {
+            return Err(Error::input_error(format!(
+                "updatedRowOffsets must be non-negative, got {offset}"
+            )));
+        }
+        bitmap.insert(offset as u32);
```
Xuanwo, line 642: The deprecated long[] encoder casts every non-negative value to u32, so offsets above u32::MAX silently become different rows and can corrupt last_updated metadata.

java/src/test/java/org/lance/operation/UpdateTest.java
```diff
@@ -104,6 +108,86 @@ void testUpdate(@TempDir Path tempDir) throws Exception {
...
+          });
+
+      try (Transaction updateTxn =
+          new Transaction.Builder()
+              .readVersion(dataset.version())
+              .operation(
+                  Update.builder()
+                      .removedFragmentIds(Collections.singletonList(fragmentId))
+                      .newFragments(Collections.singletonList(newFragment))
+                      .updateMode(Optional.of(UpdateMode.RewriteRows))
```
Xuanwo: The round-trip test uses RewriteRows, but updatedFragmentOffsets is consumed only by the RewriteColumns stable-row-id branch. The Java/Spark path can regress while this test still passes.

java/lance-jni/src/transaction.rs
jerryjch: Fixed.

java/lance-jni/src/transaction.rs
jerryjch: Fixed

java/src/main/java/org/lance/fragment/FragmentUpdateResult.java
jerryjch: Fixed.

java/lance-jni/src/transaction.rs
jerryjch: Fixed.

java/src/main/java/org/lance/fragment/FragmentUpdateResult.java
jerryjch: Fixed.

java/lance-jni/src/fragment.rs
jerryjch: Fixed.

java/src/test/java/org/lance/operation/UpdateTest.java
jerryjch: Fixed.

jerryjch:
Hi, @Xuanwo Could you review the PR again? I have addressed your comments. Thanks.

jerryjch:
Hi, @Xuanwo could you help review the PR again?

Xuanwo (approved):

jerryjch:
Thanks for the review and approval @Xuanwo !

PR #6749, opened 2026-05-12
feat: allow LSM scanner to read fresh tier without a base table

hamersaw:
## Summary

- Adds `LsmScanner::without_base_table(schema, base_path, snapshots, pk_columns)` and `LsmDataSourceCollector::without_base_table(base_path, snapshots)` so callers can scan only the active memtable and flushed L0 generations.
- Internally, `base_table` becomes `Option<Arc<Dataset>>`; `collect()` / `collect_for_shards()` skip the base source when it is absent. All three planners (scan, point lookup, vector search) are unaffected since they already drive off the collector and keep their own `base_schema` field.
- Existing `LsmScanner::new(...)` and `LsmDataSourceCollector::new(...)` signatures are unchanged, so Python bindings and benches keep compiling without edits.

## Motivation

Some callers own the base read path elsewhere and only need the WAL's contribution to a query — the active memtable ∪ L0 flushed generations. Without this change, `LsmScanner` always pulls the base dataset in, which duplicates work and forces those callers to subtract the base contribution out of band. With the new constructor, the same scanner code path serves both modes; dedup semantics across generations are unchanged.

## Test plan

- [x] `cargo test -p lance --lib dataset::mem_wal::scanner` — 55/55 passing, including three new tests:
  - `test_lsm_scan_without_base_table` — verifies the plan excludes any base-table `LanceRead` and that ids only present in the base are absent from the result.
  - `test_lsm_scan_without_base_table_no_flushed_no_active` — empty fresh tier returns an empty plan cleanly.
  - `test_point_lookup_without_base_table` — point lookup hits a flushed generation; missing key returns empty.
  - `test_vector_search_without_base_table` — plan construction succeeds and excludes the base.
- [x] `cargo clippy -p lance --tests -- -D warnings` clean
- [x] `cargo fmt -p lance` clean

### DIFF
--- a/rust/lance/src/dataset/mem_wal/scanner/builder.rs
+++ b/rust/lance/src/dataset/mem_wal/scanner/builder.rs
@@ -20,6 +20,13 @@ use super::data_source::ShardSnapshot;
 use super::planner::LsmScanPlanner;
 use crate::dataset::Dataset;
 
+/// Either a base Lance table, or an explicit base path used to resolve
+/// flushed-generation directories when no base dataset is configured.
+enum BaseSource {
+    Table(Arc<Dataset>),
+    PathOnly(String),
+}
+
 /// Scanner for LSM tree data spanning base table, flushed MemTables, and active MemTable.
 ///
 /// This scanner provides a unified interface for querying data across multiple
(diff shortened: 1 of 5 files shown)


westonpace:
@claude review once

westonpace (approved):
Looks good!

PR #6754, opened 2026-05-13
fix(mem_wal): advance visibility cursor on WAL durability, not indexing

hamersaw:
## Summary

- **§1 — Visibility cursor**: For memtables with empty `index_configs`, the WAL flush handler only advanced the visibility cursor through `IndexStore::insert_batches_parallel`, which the `None` branch of `flush_from_batch_store` skipped. The cursor stayed at 0 for the memtable's lifetime, so scanners returned only batch 0 even though the `BatchStore` held more. Compounding this, `MemTable.indexes` initialized to `None` meant the scanner and flush handler each minted throwaway `IndexStore` instances via `unwrap_or_else`, so an unconditional advance would still not share state. Fix: initialize `MemTable.indexes` to `Some(Arc::new(IndexStore::new()))` so both ends share a stable `Arc`, and the existing in-function advance fires correctly for empty registries.
- **Rename**: `max_indexed_batch_position` → `max_visible_batch_position`. The cursor now means "durable in the WAL, safe to read," not "indexed up to here." The advance method is now `pub` for callers that drive the cursor from outside the index path.
- **§3 — `pub mod util`**: changed `mod util` → `pub mod util` so external crates can import `shard_base_path`, `shard_wal_path`, `parse_bit_reversed_filename`, `bit_reverse_u64` instead of duplicating them. Helpers were already `pub` at the function level.

## Test plan

- [x] New regression `test_wal_flush_advances_visibility_with_empty_indexes` — opens a flusher, appends 3 batches, flushes with an empty `IndexStore`, asserts `max_visible_batch_position == 2`.
- [x] New regression-guard `test_wal_flush_advances_visibility_with_btree_index` — same shape with a BTree index registered.
- [x] `cargo test -p lance --lib dataset::mem_wal` — 234 / 234 pass.
- [x] `cargo check -p lance --tests --benches` clean.
- [x] `cargo clippy -p lance --tests --benches -- -D warnings` clean.
- [x] `cargo fmt --check` clean.

### DIFF
--- a/rust/lance/benches/mem_wal_index_micro.rs
+++ b/rust/lance/benches/mem_wal_index_micro.rs
@@ -256,7 +256,7 @@ async fn main() -> lance_core::Result<()> {
             let mut spins = 0u64;
             loop {
                 let active = writer.active_memtable_ref().await?;
-                if active.index_store.max_indexed_batch_position() >= target_batch_pos {
+                if active.index_store.max_visible_batch_position() >= target_batch_pos {
                     break;
                 }
                 drop(active);
--- a/rust/lance/benches/mem_wal_recall_hnsw.rs
+++ b/rust/lance/benches/mem_wal_recall_hnsw.rs
@@ -382,7 +382,7 @@ async fn run_checkpoint(
     let mut spins = 0u64;
     loop {
         let active = writer.active_memtable_ref().await?;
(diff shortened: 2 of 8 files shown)


jackye1995 (approved):
thanks for the fix!

PR #6755, opened 2026-05-13
fix!: disable auto-cleanup by default

### LINKED ISSUE #6728: perf: disable auto-cleanup by default — hook adds multi-second latency every 20 commits and usually deletes nothing

touch-of-grey:
## Summary

- Default `WriteParams::auto_cleanup` to `None` so new datasets do not opt into the periodic cleanup hook.
- Aligns the Rust default with the Python binding (already defaults to `None`) and the Java binding (inherits the Rust default).
- Adds a regression test asserting the new default and updates the cleanup-tests fixture so the auto-cleanup behavioural tests explicitly enable it.

Closes #6728

## Why

With the old default, every 20th commit ran a full cleanup pass — listing `_versions/`, reading every manifest plus its index sidecar, and listing the `data/`, `_transactions/`, `_indices/` subtrees — even when nothing was older than the 14-day default. A short S3 (us-east-1) benchmark of 60 sequential overwrites on a tiny dataset:

| Group | mean | max |
| --- | --- | --- |
| ON, v % 20 == 0 (hook fires) | 2548 ms | 2671 ms |
| ON, v % 20 != 0 | 1180 ms | 1390 ms |
| OFF (`skip_auto_cleanup=true`) | 1196 ms | 1456 ms |

The 14-day window meant the periodic spike paid the inspection cost and deleted nothing in practice. Users who want auto-cleanup can still opt in by setting `WriteParams::auto_cleanup` at create time or by setting `lance.auto_cleanup.interval` / `lance.auto_cleanup.older_than` via `Dataset::update_config` on an existing dataset.

### DIFF
--- a/rust/lance/src/dataset/cleanup.rs
+++ b/rust/lance/src/dataset/cleanup.rs
@@ -1184,7 +1184,7 @@ mod tests {
     use crate::index::DatasetIndexExt;
     use crate::{
         dataset::transaction::{Operation, Transaction},
-        dataset::{ReadParams, WriteMode, WriteParams, builder::DatasetBuilder},
+        dataset::{AutoCleanupParams, ReadParams, WriteMode, WriteParams, builder::DatasetBuilder},
         index::vector::VectorIndexParams,
     };
     use all_asserts::{assert_gt, assert_lt};
@@ -1329,6 +1329,24 @@ mod tests {
 
         async fn create_some_data(&self) -> Result<()> {
             self.write_some_data_impl(WriteMode::Create).await
+        }
+
+        // Auto-cleanup is disabled by default; this helper creates a dataset
+        // with auto-cleanup enabled using the default interval/older_than.
+        async fn create_some_data_with_auto_cleanup(&self) -> Result<()> {
+            Dataset::write(
+                some_batch(),
+                &self.dataset_path,
+                Some(WriteParams {
+                    store_params: Some(self.os_params()),
+                    commit_handler: Some(Arc::new(RenameCommitHandler)),
+                    mode: WriteMode::Create,
+                    auto_cleanup: Some(AutoCleanupParams::default()),
+                    ..Default::default()
+                }),
+            )
+            .await?;
+            Ok(())
         }
 
         async fn overwrite_some_data(&self) -> Result<()> {
@@ -1907,7 +1925,7 @@ mod tests {
         // commit.
         let fixture = MockDatasetFixture::try_new().unwrap();
 
-        fixture.create_some_data().await.unwrap();
+        fixture.create_some_data_with_auto_cleanup().await.unwrap();
 
         let dataset_config = &fixture.open().await.unwrap().manifest.config;
         let cleanup_interval: usize = dataset_config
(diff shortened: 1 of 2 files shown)


jackye1995:
this should be a breaking change of behavior, I updated the title

jackye1995 (approved):
thanks for the fix, looks good to me

PR #6757, opened 2026-05-13
chore: use Uuid type throughout index infrastructure instead of String

### LINKED ISSUE #6347: chore: use Uuid type throughout index infrastructure instead of String
Several index-related types store or pass UUIDs as `String`/`&str` despite `IndexMetadata.uuid` already being the native `Uuid` type. This forces unnecessary `.to_string()` conversions at every callsite.

**Changes, in dependency order:**

1. `CreateIndexBuilder.index_uuid: Option<String>` → `Option<Uuid>` (`lance/src/index/create.rs`). The builder currently accepts a `String` and immediately calls `Uuid::parse_str` internally — accepting `Uuid` directly removes this round-trip.

2. `DatasetIndexInternalExt` trait methods `open_generic_index`, `open_scalar_index`, `open_vector_index` — change `uuid: &str` → `uuid: &Uuid` (`lance/src/index.rs`). Every call site converts from `IndexMetadata.uuid` (already `Uuid`) via `.to_string()`, which would be eliminated.

3. `IVFIndex.uuid: String` → `Uuid` in both `lance/src/index/vector/ivf.rs` and `lance/src/index/vector/ivf/v2.rs`. Unblocked by #2 above.

4. `ANNIvfPartitionExec.index_uuids: Vec<String>` → `Vec<Uuid>` (`lance/src/io/exec/knn.rs`). Currently built by mapping `IndexMetadata.uuid.to_string()`. Unblocked by #2 above.

5. Cache key structs in `lance/src/session/index_caches.rs` that use `uuid: &'a str` — standardize to `&'a Uuid` to match the one key that already uses `&'a Uuid`.

wombatu-kun:
Closes #6347.

Replaces `String`/`&str` UUIDs with `uuid::Uuid` across the index infrastructure. `IndexMetadata.uuid` was already `Uuid`, so most internal callsites were doing `idx.uuid.to_string()` (per-iter heap alloc) or `Uuid::parse_str` to convert it back — this PR removes those round-trips.

The change extends beyond the five items listed in the issue to clean up the adjacent functions that took `uuid: &str` for the same reason: `load_index`, `merge_index_metadata`, the `build_*` helpers, `from_dataset_for_new`, and `VectorIndexExtension`.

## FFI behaviour

Python `create_scalar_index(..., index_uuid="not-a-uuid")` and `merge_index_metadata("not-a-uuid", ...)` now raise `ValueError("Invalid UUID string for index_uuid: ...")` at the binding boundary, instead of an `lance::Error::Index` from deeper in. Java JNI behaves equivalently via `Error::input_error`. A new test in `python/python/tests/test_scalar_index.py` locks the Python behaviour in.

## Performance note

Cache key `fn key() -> Cow<'_, str>` impls now allocate a `String` (via `Uuid::to_string`) instead of returning 
 `Cow::Borrowed(&str)`. This is unavoidable when changing the field type and matches what `MemWalCacheKey::key()` already does. The savings on the lookup side — no `idx.uuid.to_string()` per index in `load_index`, no `Uuid::parse_str`
in `build_distributed_vector_index` — outweigh the per-key allocation in practice (one alloc per lookup vs. one alloc per index per scan).

## Wire-format stability (verified unchanged)

- `IvfIndexStateHeader` JSON (on-disk cache)
- `IvfIndexState` / `IvfIndexStatistics` JSON
- `pb::AnnIvfPartitionExecProto.index_uuids` (proto)
- Arrow `__index_uuid` column type (`DataType::Utf8`)
- `Uuid::Display` produces the canonical hyphenated-lowercase form, identical to the previous `String` representation, so any caller reading these fields textually sees the same bytes.

### DIFF
--- a/python/python/tests/test_scalar_index.py
+++ b/python/python/tests/test_scalar_index.py
@@ -108,6 +108,23 @@ def _commit_segmented_btree_index(dataset, column, index_name):
     return dataset.commit_existing_index_segments(index_name, column, segments)
 
 
+def test_create_scalar_index_rejects_invalid_uuid(tmp_path):
+    """Invalid UUID strings passed to create_scalar_index and merge_index_metadata
+    must surface as a Python ValueError at the FFI boundary."""
+    data = pa.table({"id": pa.array(range(100), type=pa.int64())})
+    dataset = lance.write_dataset(data, tmp_path / "ds")
+
+    with pytest.raises(ValueError, match="Invalid UUID"):
+        dataset.create_scalar_index(
+            column="id",
+            index_type="BTREE",
+            index_uuid="not-a-uuid",
+        )
+
+    with pytest.raises(ValueError, match="Invalid UUID"):
+        dataset.merge_index_metadata("also-not-a-uuid", index_type="BTREE")
+
+
 @pytest.fixture
 def btree_comparison_datasets(tmp_path):
     """Setup datasets for B-tree comparison tests"""
--- a/rust/lance/src/index.rs
+++ b/rust/lance/src/index.rs
@@ -12,7 +12,7 @@ use async_trait::async_trait;
 use datafusion::execution::SendableRecordBatchStream;
 use futures::FutureExt;
 use itertools::Itertools;
-use lance_core::cache::{CacheKey, UnsizedCacheKey};
+use lance_core::cache::CacheKey;
 use lance_core::datatypes::Field;
 use lance_core::datatypes::Schema as LanceSchema;
 use lance_core::utils::address::RowAddress;
@@ -273,34 +273,6 @@ fn segment_has_btree_details(segment: &IndexMetadata) -> bool {
 }
 
 // Cache keys for different index types
-#[derive(Debug, Clone)]
-pub struct ScalarIndexCacheKey<'a> {
-    pub uuid: &'a Uuid,
-    pub fri_uuid: Option<&'a Uuid>,
-}
-
-impl<'a> ScalarIndexCacheKey<'a> {
-    pub fn new(uuid: &'a Uuid, fri_uuid: Option<&'a Uuid>) -> Self {
-        Self { uuid, fri_uuid }
-    }
-}
-
-impl UnsizedCacheKey for ScalarIndexCacheKey<'_> {
-    type ValueType = dyn ScalarIndex;
-
-    fn key(&self) -> std::borrow::Cow<'_, str> {
-        if let Some(fri_uuid) = self.fri_uuid {
-            format!("{}-{}", self.uuid, fri_uuid).into()
-        } else {
-            self.uuid.to_string().into()
-        }
-    }
-
-    fn type_name() -> &'static str {
-        "ScalarIndex"
-    }
-}
-
 #[derive(Debug, Clone)]
 pub(crate) struct LegacyVectorIndexCacheKey<'a> {
     uuid: &'a Uuid,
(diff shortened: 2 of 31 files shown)


wombatu-kun:
@wjones127 hi!

This PR has been sitting for a bit, and the rebase situation is starting to bite. CI broke because PR #6737 (commit c51f5a11c, "feat(index): merge incremental inverted index segments") landed in the meantime and added 5 new callsites in `rust/lance/src/index/append.rs` that pass `&uuid.to_string()` to `open_scalar_index` / `build_scalar_index` / `LanceIndexStore::from_dataset_for_new` — exactly the functions whose signatures this PR converts from `&str` → `&Uuid` / `Uuid`.

Git merged cleanly (no textual conflicts), but `rustc` rejected the type mismatch and took down ~15 CI jobs (clippy, rustdoc, MSRV, every platform Rust build, every Python wheel, Java JNI, Python lint — anything that compiles the `lance` crate). Just pushed a small follow-up commit (5-line fix in `append.rs`) that adapts those new callsites to the new types; CI is green again shortly.

The longer this sits, the more likely a similar collision happens again — the touched API surface (`open_scalar_index`, `build_scalar_index`, `from_dataset_for_new`, etc.) is in active churn. Would really appreciate a review pass so we can land this before the next rebase round. Happy to address any feedback quickly.

Thanks!

wombatu-kun:
The `linux-arm` (Rust) CI failure here is **unrelated to the changes in this PR**.

This PR only retypes the index identifier from `String` to `Uuid`. The only search-path change is `&segment.uuid.to_string()` → `&segment.uuid`, and the resulting index cache keys and on-disk index paths are byte-identical (`Uuid`'s `Display` renders the same lowercase-hyphenated string the previous `.to_string()` produced). The failing test body (`test_ann_prefilter`) is unchanged by this PR.

The failure is a pre-existing flaky test. Root cause: HNSW node-level assignment uses an unseeded thread RNG (`rand::rng()`), so every index build produces a different random graph. On the tiny dataset used by `test_ann_prefilter`, an approximate HNSW+SQ search over a different graph each run can return a near-miss neighbor, and the test asserts an exact nearest-neighbor row id. `main` is green by luck of the RNG, not correctness.

The fix is in a separate PR: #6818 (seed the HNSW build RNG to make graph construction deterministic, and make the test assert prefilter correctness rather than an exact NN id).

Once #6818 merges, rebasing this PR will clear the flake. In the meantime, re-running the failed `linux-arm` job should pass (the failure is intermittent).

rust/lance/src/index.rs
```diff
@@ -272,14 +272,42 @@ fn segment_has_btree_details(segment: &IndexMetadata) -> bool {
 }
 
 // Cache keys for different index types
+#[derive(Debug, Clone)]
+pub struct ScalarIndexCacheKey<'a> {
```
wjones127: issue(blocking): this is never used. I think I deleted it in a different PR, so maybe this is an artifact from a merge?

python/python/tests/test_scalar_index.py
```diff
@@ -108,6 +108,23 @@ def _commit_segmented_btree_index(dataset, column, index_name):
     return dataset.commit_existing_index_segments(index_name, column, segments)
 
 
+def test_create_scalar_index_rejects_invalid_uuid(tmp_path):
```
wjones127, line 111: praise: I appreciate these tests making the new behavior clear!

wjones127 (approved):
Thanks for your patience. Looks like there one merge conflict and some dead code. I'll see if I can fix that now and get this merged.

rust/lance/src/index.rs
wjones127: Fixed — removed the struct and the now-unused  import in the latest commit.

PR #6758, opened 2026-05-13
fix: include view tag in FieldDataCacheKey

wombatu-kun:
## Summary

A blob column can be decoded under two distinct shapes that share the same physical column:

- the **descriptor** view (`Struct<position, size>`) used when scanning with `BlobHandling::BlobsDescriptions` (the default for `to_table` and `take_blobs`);
- the **bytes** view (`LargeBinary`) used by `BlobHandling::AllBinary` and `compact_files`.

Both paths build a `StructuralPrimitiveFieldScheduler`, but the page-level schedulers they instantiate cache different concrete `CachedPageData` types. The cache key (`FieldDataCacheKey`) carried only `column_index`, so when one shape populated the cache and a second reader on the same `Dataset` / `Session` hit it with the other shape, `BlobPageScheduler::load` would downcast the wrong state and panic:

```
panicked at rust/lance-encoding/src/encodings/logical/primitive/blob.rs:335:14:
called `Result::unwrap()` on an `Err` value: Any { .. }
```

This change mixes a Debug-formatted representation of the target field's `DataType` into the cache key. The two blob views now get their own entries while non-blob columns keep sharing as before (their `DataType` is invariant for a given field).

The accompanying regression test (`test_blob_cache_key_distinguishes_views`) runs three back-to-back scans on a single `Arc<Dataset>` — descriptor view, bytes view, descriptor view — and asserts they all succeed and return the expected bytes. Without the fix the second scan panics; with it the test passes.


Xuanwo (approved):
Thank you for working on this. For a bigger picture, I think we should have an entire ViewType support in the future.

Issue #5418, opened 2025-12-05
We need better coalescing of part loads in bitmap and btree indexes


westonpace:
When a filter hits a large portion of a secondary index we can issue a lot of part loads.  This is most noticeable in bitmap filters with many unique values but also in large btree filters.

This introduces a lot of file reader overhead as well as a lot of IOPS.  In these cases it would better to either reject using the index entirely or load the many parts in a few coalesced reads.

In fact, maybe the simplest solution would be to always load all parts in one read.  This should be very doable since the file reader can take any arbitrary set of row ranges.  We can then rely on the file reader to coalesce the IOPS as appropriate.

PR #6759, opened 2026-05-13
feat(index): coalesce btree scalar index page-data reads

zouhuajian:
## Summary

This PR addresses part of [issue-5418](https://github.com/lance-format/lance/issues/5418) by reducing BTree scalar index `page_data` read amplification.

The issue notes that filters hitting a large portion of a secondary index can issue many part loads, causing high file reader overhead and IOPS. This is also visible in large BTree range filters, where many selected BTree leaf pages may be loaded one page at a time.

In my Spark validation, range predicates on a BTree-indexed `optime` column produced many `page_data` loads. This PR keeps BTree page selection unchanged, but changes cache-miss materialization to use bounded coalesced reads:

- group missing BTree pages by physical `page_data` file
- sort by local page id
- merge adjacent pages into row ranges
- read those ranges through `IndexReader::read_ranges`
- split the returned batches back into per-page `FlatIndex` entries
- cache pages by global page id as before

The same candidate pages are still searched with the same `FlatIndex` logic. The change only reduces how those pages are read from storage.

This PR also adds focused metrics/logging:

- `btree_read_plan` for candidate page shape
- `btree_coalesced_read_batch` for actual coalesced storage reads
- `btree_search` for cache behavior, pages materialized from storage, coalescing counts, and elapsed time

While validating this change, I found a separate scalar-index planning issue involving pushed predicates such as `IS NOT NULL` combined with same-column range predicates. That issue can cause redundant BTree searches and will be tracked separately. This PR is limited to BTree `page_data` read coalescing and does not change predicate semantics, NULL handling, Spark connector behavior, cost modeling, fallback scan behavior, single-flight, or bitmap index loading.

## Validation

I validated this PR with a Spark range-filter workload that originally exposed the BTree `page_data` read amplification.

Representative query:

    SELECT count(*)
    FROM lance_no_pushdown
    WHERE optime >= '2026-05-08 01:00:00'
      AND optime <= '2026-05-08 01:01:00'

Representative indexed table:

| Item | Value |
|---|---:|
| Indexed column | `optime` |
| Indexed fragments | `344` |
| Indexed rows | `788,264,057` |
| Unindexed fragments | `0` |
| Unindexed rows | `0` |

Larger production-like workload context:

| Item | Approx. value |
|---|---:|
| Data size | `~1.529 TB` |
| Data files / fragments | `~5.1K` |
| Rows | `~16.4B` |
| Index files | `~10K` |

After this change, the BTree read-plan metrics show that cache-missing pages are grouped into much fewer physical read batches.

Observed cold-cache BTree read-plan example:

| Metric | Value |
|---|---:|
| Candidate BTree pages | `11,385` |
| Physical `page_data` files | `47` |
| Coalesced read batches | `~197` |

This changes the page materialization pattern from approximately one storage read per candidate page to a much smaller number of coalesced read batches:

    11,385 candidate page reads -> ~197 coalesced read batches
    ~= 57x fewer page-data read batches

The selected candidate pages are unchanged. The improvement is only in how cache-missing `page_data` pages are materialized from storage.

This is especially useful for remote/object-store-like filesystems where many small part/page loads can be dominated by file open, metadata, and range scheduling overhead.

This PR does not claim to fully solve all end-to-end latency cases in #5418. In particular, it does not address separate planner issues such as redundant same-column predicates like `IS NOT NULL AND col >= lower AND col <= upper`, cost-based index fallback, LIMIT early stop, or bitmap index loading.

## Tests

Rust checks:

| Check | Status |
|---|---|
| `cargo fmt` | Passed |
| `cargo test -p lance-index` | Passed |
| Existing BTree scalar-index tests | Passed |

Scenario validation:

| Scenario | Result |
|---|---|
| BTree range query on indexed `optime` column | Same candidate page semantics; fewer page-data read batches |
| Cold-cache page materialization | Missing pages are grouped by physical `page_data` file and coalesced into bounded row ranges |
| Warm-cache repeated search | Existing page cache keys are preserved; already materialized pages are served from cache |
| Table with all fragments indexed | Query path remains indexed; no unindexed-fragment fallback involved |


zouhuajian:
@Xuanwo Could you help review this when you have time? Thanks.

rust/lance-index/src/scalar/btree.rs
```diff
@@ -1778,13 +2194,40 @@ impl ScalarIndex for BTreeIndex {
...
+            btree_candidate_pages = read_plan.candidate_pages,
+            btree_physical_files = read_plan.physical_files,
+            btree_page_runs = read_plan.page_runs,
+            btree_min_page_id = ?read_plan.min_page_id,
+            btree_max_page_id = ?read_plan.max_page_id,
+        );
+
         let lazy_index_reader =
             LazyIndexReader::new(self.store.clone(), self.ranges_to_files.clone());
+        let (page_indices, coalesced_metrics) = self
```
Xuanwo: The search path materializes every cold candidate page before evaluating any page, so broad range predicates can retain thousands of leaf pages until the final union. Large cold queries can trade fewer I/O calls for an unbounded peak-memory spike.

rust/lance-index/src/scalar/btree.rs
```diff
@@ -1260,14 +1461,232 @@ impl BTreeIndex {
...
+        let mut read_metrics = BTreeCoalescedReadMetrics::default();
+
+        for matches in pages {
+            let page_number = matches.page_id();
+            if page_indices.contains_key(&page_number) || missing_seen.contains(&page_number) {
+                continue;
+            }
+
+            let cache_key = BTreePageKey { page_number };
+            if let Some(page) = self.index_cache.get_with_key(&cache_key).await {
```
Xuanwo, line 1518: The multi-page miss path checks the cache non-atomically and later inserts the loaded pages, bypassing the cache single-flight behavior used by the single-page path. Concurrent cold broad-range queries can duplicate the same coalesced reads and materialization work.

rust/lance-index/src/scalar/btree.rs
zouhuajian: Thanks for pointing this out. I pushed a follow-up commit that bounds the search path by materializing and evaluating pages in batches instead of retaining all cold candidate pages until the final union.

That said, I’m going to keep this PR as draft for now. The `read_ranges` API added in this PR now overlaps with the API introduced in #6983, so this part still needs a proper rebase / API cleanup before the PR is ready again.

PR #6761, opened 2026-05-13
fix(mem_wal): return _distance from LSM vector search across mixed sources

hamersaw:
## Summary

`LsmVectorSearchPlanner::plan_search` produced wrong or no output whenever more than one LSM source contributed candidates. Three bugs combined to swallow the `_distance` column:

1. **Schema mismatch panic.** The base/flushed arm calls `scanner.fast_search()`, which forces `_rowid` into Lance's projection (`[id, vector, _rowid, _distance]`). The active-memtable arm produces `[id, vector, _distance]`. After `MemtableGenTagExec` adds `_memtable_gen`, `UnionExec` rejects the 5-vs-4 mismatch and the query panics before any `_distance` reaches the caller.
2. **Silent partition loss.** `UnionExec` emits one partition per input source. Both `SortExec` and `FilterStaleExec` only read partition 0 of their multi-partition input, so KNN hits (and their `_distance` values) from every source past the first silently vanished.
3. **Internal columns leaked.** `MemtableGenTagExec` was wrapped unconditionally, so `_memtable_gen` ended up in user-visible output even when no bloom filters were configured.

### Fixes

- Add `project_to_canonical` to normalize every KNN source to `[projection..., _distance]` before union. This drops `_rowid` from the base/flushed arms so all inputs share one schema.
- Insert `CoalescePartitionsExec` before `SortExec`, and again before `FilterStaleExec` on the bloom-filter path, so the merge sees every KNN candidate.
- Only wrap with `MemtableGenTagExec` when bloom filters are configured; project `_memtable_gen` back out after `FilterStaleExec` so the public schema stays clean.
- Route the active arm through `build_projection_for_knn` so explicit user projections don't drop the PK that staleness detection needs.

## Test plan

- [x] `test_vector_search_base_plus_active_returns_distance` — end-to-end search across base + active memtable; asserts `_distance` present, nearest neighbor has near-zero distance, and `_rowid` / `_memtable_gen` are NOT in the output.
- [x] `test_vector_search_with_projection_returns_distance_and_pk` — regression for the active-arm projection path; PK column auto-included when user projects only `vector`.
- [x] `test_vector_search_with_bloom_filter_strips_memtable_gen` — exercises the bloom-filter branch; verifies the post-`FilterStaleExec` projection strips `_memtable_gen` and that union partitions feed correctly into the filter.
- [x] All existing `mem_wal` tests pass.
- [x] `cargo clippy -p lance --tests --benches -- -D warnings` clean.

\U0001f916 Generated with [Claude Code](https://claude.com/claude-code)


westonpace:
@claude review once

hamersaw:
Sorry, everything passed locally. Fixing CI issues.

westonpace (approved):
The only case I can think of here that might be different is when the user specifies _rowid or _distance as part of the projection.  They can do this, for example, to choose that _distance comes first.  I feel pretty confident the scanner handles this but it might be worth a quick manual verification if not too much trouble.

hamersaw:
> The only case I can think of here that might be different is when the user specifies _rowid or _distance as part of the projection. They can do this, for example, to choose that _distance comes first. I feel pretty confident the scanner handles this but it might be worth a quick manual verification if not too much trouble.

TL;DR using any system column in a projection breaks things. Rather than hacking this for just vector search, I'm working through a fix across the LSMScanner for all query types in WAL.

hamersaw:
Ok, this inflated a little bit. But now we have working projections for all system columns across all query types and unit tests to avoid regressions.

westonpace:
@claude review once

westonpace:
I'll merge once Claude is done with another review.  I looked over the new regression tests and things look correct to me.

PR #6771, opened 2026-05-13
fix: complex all-null list struct decoding

### LINKED ISSUE #6768: v2.1+ encoding: `list<struct<…>>` decode fails with "Incorrect array length" when a struct child column is all-null across list values
## Related

* Distinct from lance-format/lance#5930 (struct-element nulls inside `List<Struct>`, fixed by [#6008](<https://github.com/lance-format/lance/issues/6008>)). That fix is present in pylance 4.0.0+ — the case below is not covered by it.
* Reproduces on `pylance==6.0.0` (which makes v2.1 the default) and on `pylance==4.0.1` when v2.1 is requested explicitly.

## Summary

When writing a `list<struct<…>>` column whose inner struct has **multiple children** and **at least one child column is entirely null across all list values**, lance v2.1+ encoders produce a file that the decoder cannot read back. The decoder raises:

```
pyarrow.lib.ArrowInvalid: External error: Encountered internal error.
Please file a bug report at https://github.com/lance-format/lance/issues.
Error decoding batch: Invalid user input: Invalid argument error:
Incorrect array length for StructArray field "<name>", expected N got 1,
.../rust/lance-encoding/src/encodings/logical/struct.rs:394:26,
.../rust/lance-encoding/src/decoder.rs:2704:26
```

`N` is the total number of list values across all rows; the decoder reports a length-1 child where the parent expects length-N. The same data round-trips fine on `data_storage_version="2.0"` / `"legacy"` / `"0.1"`.

This is not about per-row struct-element nullability (that was lance-format/lance#5930). It is about column-wise all-null child fields inside a list of structs.

## Environment

* pylance: `6.0.0` (also reproduces on `4.0.1` with explicit `data_storage_version="2.1"`)
* pyarrow: `24.0.0`
* Python: `3.12.12`
* OS: macOS arm64 (also reproduces on Linux x86_64)

## Minimal reproduction

```python
import lance
import pyarrow as pa

# list<struct<a: utf8 (nullable), b: utf8 (nullable)>>
# Two list items in one row; column `b` is null in every item.
struct_type = pa.struct([
    pa.field("a", pa.utf8(), nullable=True),
    pa.field("b", pa.utf8(), nullable=True),
])
schema = pa.schema([
    pa.field("c", pa.list_(struct_type), nullable=False),
])
rows = [{"c": [{"a": "x", "b": None}, {"a": "y", "b": None}]}]
table = pa.Table.from_pylist(rows, schema=schema)

ds = lance.write_dataset(table, "memory://t", data_storage_version="2.2")
ds.to_table()  # <-- raises
```

Expected: round-trip succeeds; `to_table()` returns the same logical data that was written.  <br>Actual: `pyarrow.lib.ArrowInvalid: ... Incorrect array length for StructArray field "b", expected 2 got 1`.

## Coverage matrix

Same data, varying `data_storage_version`:

| Version | Result |
| -- | -- |
| `2.0` | ✅ OK |
| `legacy` | ✅ OK |
| `0.1` | ✅ OK |
| `2.1` | ❌ FAIL |
| `2.2` | ❌ FAIL |
| `2.3` | ❌ FAIL |
| `next` | ❌ FAIL |
| `stable` | ❌ FAIL |

## Pattern bisect (which list-of-struct shapes break v2.2?)

A grid against `data_storage_version="2.2"` on pylance 6.0.0:

| Struct children | Data pattern | Result |
| -- | -- | -- |
| 1 nullable | any (all set / all null / mixed) | ✅ OK |
| 2 nullable | all values None in every list item | ✅ OK |
| 2 nullable | column `a` set, column `b` all-null | ❌ FAIL |
| 2 nullable | alternating set/null across items (no all-null column) | ✅ OK |
| 2 nullable | all values set | ✅ OK |
| 2 nullable | single list item, one set + one null | ✅ OK |
| 3 nullable | each item sets exactly one different field (one column ends up all-null) | ❌ FAIL |
| 1 non-null + 1 nullable | nullable column all-null | ❌ FAIL |
| 1 non-null + 3 nullable | one nullable column all-null | ❌ FAIL |

Three consistent observations:

1. **Trigger condition**: a struct child column being all-null across the flattened list values appears sufficient. Multi-child struct + at least one column that contains only nulls across all list values → decode fails.
2. **Independent of row count**: reproduces with `n=1` row containing ≥ 2 list items, and at `n=100`, `n=1000`.
3. **Independent of how the array is built**: `pa.Table.from_pylist`, `pa.StructArray.from_arrays`, and `pa.RecordBatch` direct construction all reproduce.

The naming pattern of the error (`Incorrect array length for StructArray field "b", expected 2 got 1`) suggests the v2.1+ encoder is compressing an all-null child to length-1, but the decoder still expects the parent list-of-values length on readback. That's consistent with the `struct.rs:394` assertion site.

## Why this is distinct from lance-format/lance#5930

lance-format/lance#5930's reproducer used `pa.StructArray.from_arrays(..., mask=...)` to mark some struct **elements** as null inside a `List<Struct>`. The fix in [#6008](<https://github.com/lance-format/lance/issues/6008>) ("correctly calculate max visible level when a list has no def") handled that. That exact reproducer round-trips successfully on pylance 6.0.0:

```python
# Issue #5930's exact repro — passes on pylance 6.0.0:
struct_array = pa.StructArray.from_arrays(
    [pa.array(["valid", "null_struct", "valid", "valid"])],
    fields=[pa.field("tag", pa.string(), nullable=True)],
    mask=pa.array([True, False, True, True]),
)
# ... wrap in list, write/read — OK
```

The case in this report uses fully-valid struct elements (no struct-level null mask) but has multi-child structs where one child column is column-wise all-null. The error site also differs: lance-format/lance#5930 produced a Rust panic on `struct.rs:382` (assertion `left == right` failed: left:1 right:0); this report consistently surfaces on `struct.rs:394:26` and `decoder.rs:2704:26` as `Invalid argument error: Incorrect array length`.

## Suggested investigation entry points

* `rust/lance-encoding/src/encodings/logical/struct.rs:394` — assertion site.
* `rust/lance-encoding/src/decoder.rs:2704` — caller frame.
* Encoder path for `list<struct>` where one child arrives as all-null: is the child length being recorded as 1 (compressed sentinel) without a corresponding decoder expansion to the parent list-values length?

Happy to test a candidate fix against the bisect grid above.

---

## Attachment: extended reproducer

```python
"""Reproduce v2.1+ list<struct<...>> decode failure with column-wise all-null child."""
import shutil
import tempfile
from pathlib import Path

import lance
import pyarrow as pa

schema = pa.schema([
    pa.field("c", pa.list_(pa.struct([
        pa.field("a", pa.utf8(), nullable=True),
        pa.field("b", pa.utf8(), nullable=True),
    ])), nullable=False),
])
table = pa.Table.from_pylist(
    [{"c": [{"a": "x", "b": None}, {"a": "y", "b": None}]}],
    schema=schema,
)

for ver in ["2.0", "0.1", "legacy", "2.1", "2.2", "2.3", "next", "stable"]:
    tmp = Path(tempfile.mkdtemp(prefix=f"lance_{ver}_"))
    uri = str(tmp / "ds.lance")
    try:
        lance.write_dataset(table, uri, data_storage_version=ver)
        lance.dataset(uri).to_table()
        print(f"  ver={ver:8s} OK")
    except Exception as e:
        print(f"  ver={ver:8s} FAIL: {str(e)[:120]}")
    finally:
        shutil.rmtree(tmp, ignore_errors=True)
```

Xuanwo:
Fixes #6768.

The v2.1+ complex all-null decoder was slicing repetition/definition buffers with row ranges, which breaks nested layouts where a top-level row expands into multiple list values. This updates the decoder to map requested rows to level slices before materializing the all-null child, and adds coverage for `list<struct<...>>` where one struct child is all-null across list values.

### DIFF
--- a/rust/lance-encoding/src/encodings/logical/primitive.rs
+++ b/rust/lance-encoding/src/encodings/logical/primitive.rs
@@ -777,6 +777,7 @@ pub struct ComplexAllNullScheduler {
     buffer_offsets_and_sizes: Arc<[(u64, u64)]>,
     def_meaning: Arc<[DefinitionInterpretation]>,
     repdef: Option<Arc<CachedComplexAllNullState>>,
+    max_rep: u16,
     max_visible_level: u16,
     rep_decompressor: Option<Arc<dyn BlockDecompressor>>,
     def_decompressor: Option<Arc<dyn BlockDecompressor>>,
@@ -793,6 +794,7 @@ impl ComplexAllNullScheduler {
         num_rep_values: u64,
         num_def_values: u64,
     ) -> Self {
+        let max_rep = def_meaning.iter().filter(|l| l.is_list()).count() as u16;
         let max_visible_level = def_meaning
             .iter()
             .take_while(|l| !l.is_list())
@@ -802,6 +804,7 @@ impl ComplexAllNullScheduler {
             buffer_offsets_and_sizes,
             def_meaning,
             repdef: None,
+            max_rep,
             max_visible_level,
             rep_decompressor,
             def_decompressor,
@@ -940,7 +943,10 @@ impl StructuralPageScheduler for ComplexAllNullScheduler {
             def: self.repdef.as_ref().unwrap().def.clone(),
             num_rows,
             def_meaning: self.def_meaning.clone(),
+            max_rep: self.max_rep,
             max_visible_level: self.max_visible_level,
+            cursor_row: 0,
+            cursor_level: 0,
         }) as Box<dyn StructuralPageDecoder>;
         let page_load_task = PageLoadTask {
             decoder_fut: std::future::ready(Ok(decoder)).boxed(),
@@ -957,7 +963,10 @@ pub struct ComplexAllNullPageDecoder {
     def: Option<ScalarBuffer<u16>>,
     num_rows: u64,
     def_meaning: Arc<[DefinitionInterpretation]>,
+    max_rep: u16,
     max_visible_level: u16,
+    cursor_row: u64,
+    cursor_level: usize,
 }
 
 impl ComplexAllNullPageDecoder {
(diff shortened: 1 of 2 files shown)


Xuanwo:
@claude review once

wjones127 (approved):
This all looks reasonable. Thanks for the quick fix!

PR #6772, opened 2026-05-13
feat(java): add Dataset.takeRows() for physical row ID access

alex766:
## Summary
Adds `Dataset.takeRows(List<Long> rowIds, List<String> columns)` to the Java SDK, mirroring Rust's existing `Dataset::take_rows()` by adding a JNI binding

`take()` accepts logical row indices (positions). `takeRows()` accepts physical row IDs from the `_rowid` system column — these are stable across compaction and deletion, which makes them suitable for applications that store row IDs externally (e.g., in a secondary index) and later need to fetch the corresponding row data.

## Changes
  - `java/lance-jni/src/blocking_dataset.rs`: JNI shim `nativeTakeRows` + `inner_take_rows` — mirrors `inner_take` but calls
  `dataset.take_rows()` instead of `dataset.take()`
  - `java/src/main/java/org/lance/Dataset.java`: public `takeRows()` method + native declaration
  - `java/src/test/java/org/lance/DatasetTest.java`: `testTakeRows` covering correctness, input-order preservation, and edge cases (empty/null rejection)

### DIFF
--- a/java/lance-jni/src/blocking_dataset.rs
+++ b/java/lance-jni/src/blocking_dataset.rs
@@ -2136,6 +2136,58 @@ fn inner_take(
 }
 
 #[unsafe(no_mangle)]
+pub extern "system" fn Java_org_lance_Dataset_nativeTakeRows(
+    mut env: JNIEnv,
+    java_dataset: JObject,
+    row_ids_obj: JObject, // List<Long>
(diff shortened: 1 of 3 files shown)


hamersaw (approved):
Thanks for the PR! Looks great.

Xuanwo (approved):
Thank you!

PR #6773, opened 2026-05-13
feat: add commit timeout to CommitBuilder

### LINKED ISSUE #6747: Add global timeout to CommitBuilder
We had an issue where a process was waiting 3 hours to commit something. We would prefer if the commit step timed out. I think that timeout should default to around 5 minutes.

wjones127:
Previously, `CommitBuilder::execute` / `execute_batch` could block
indefinitely if the underlying object store hung (we observed a 3-hour
wait in production). This adds a configurable timeout that defaults to
30 minutes:

- Rust: `CommitBuilder::with_timeout(Option<Duration>)`. `None` disables
  the timeout; `Some(Duration::ZERO)` is rejected with `InvalidInput`.
- Python: `LanceDataset.commit(..., commit_timeout=timedelta(minutes=5))`,
  pass `None` to disable.
- Java: `new CommitBuilder(ds).commitTimeout(Duration.ofMinutes(5))`,
  pass `null` to disable.

Fixes #6747

### DIFF
--- a/rust/lance-table/src/io/commit.rs
+++ b/rust/lance-table/src/io/commit.rs
@@ -921,8 +921,54 @@ pub trait CommitLease: Send + Sync {
     async fn release(&self, success: bool) -> std::result::Result<(), CommitError>;
 }
 
+/// Guards a [CommitLease] so the lock is released even if the commit future is
+/// dropped (e.g. cancelled by a commit timeout) before reaching an explicit
+/// release.
+///
+/// [CommitLease::release] is async and cannot be awaited from `Drop`, so on the
+/// drop path we spawn a best-effort background task that releases the lock with
+/// `success = false`. Without this, a cancelled commit would leak the lock until
+/// the lease's own TTL expired, blocking other writers in the meantime.
+struct LeaseGuard<L: CommitLease + 'static> {
+    lease: Option<L>,
+}
+
+impl<L: CommitLease + 'static> LeaseGuard<L> {
+    fn new(lease: L) -> Self {
+        Self { lease: Some(lease) }
+    }
+
+    /// Explicitly release the lease, consuming the guard so `Drop` is a no-op.
+    async fn release(mut self, success: bool) -> std::result::Result<(), CommitError> {
+        let lease = self
+            .lease
+            .take()
+            .expect("LeaseGuard released more than once");
+        lease.release(success).await
+    }
+}
+
+impl<L: CommitLease + 'static> Drop for LeaseGuard<L> {
+    fn drop(&mut self) {
+        if let Some(lease) = self.lease.take() {
+            // The guard was dropped without an explicit release, meaning the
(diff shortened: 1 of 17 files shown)


brendanclement (approved):
LGTM

rust/lance/src/dataset/write/commit.rs
```diff
@@ -181,6 +206,32 @@ impl<'a> CommitBuilder<'a> {
...
+            ));
+        }
+        // Box the inner future so wrapping it in `tokio::time::Timeout` does
+        // not deepen the future type — downstream `async fn`s that await
+        // `execute` otherwise hit the compiler's layout-query depth limit.
+        let fut = Box::pin(self.execute_inner(transaction));
+        match timeout {
+            Some(t) => match tokio::time::timeout(t, fut).await {
+                Ok(res) => res,
+                Err(_) => Err(Error::timeout(format!(
```
BubbleCal, line 224: It looks like this the lock inside `execute_inner` may leak if the timeout drops the future before releasing lock, a possible position: https://github.com/lance-format/lance/blob/a2d8aa91f92d2adf04bc1d1732a5aaa937b52725/rust/lance-table/src/io/commit.rs#L959

wjones127: I believe I've addressed this in the latest commit.

rust/lance-table/src/io/commit.rs
```diff
@@ -921,8 +921,54 @@ pub trait CommitLease: Send + Sync {
...
+
+    /// Explicitly release the lease, consuming the guard so `Drop` is a no-op.
+    async fn release(mut self, success: bool) -> std::result::Result<(), CommitError> {
+        let lease = self
+            .lease
+            .take()
+            .expect("LeaseGuard released more than once");
+        lease.release(success).await
+    }
+}
```
BubbleCal, line 961:
```suggestion
async fn release(mut self, success: bool) -> std::result::Result<(), CommitError> {
    let result = {
        let lease = self
            .lease
            .as_ref()
            .expect("LeaseGuard released more than once");
        lease.release(success).await
    };

    // Explicit release reached completion, so Drop should not issue a second
    // best-effort release even if the release returned an error.
    self.lease = None;

    result
}
```
the lease may leak if timeout happens right before `lease.release(success).await`?

wjones127: Good catch — you're right. `take()` emptied the guard before the await, so a cancellation during the release itself left nothing to fire the drop-path release. Applied your suggestion in a2e8b7ea4: the lease now stays in the guard across the await and is only cleared once release completes.

I also added a regression test (`test_commit_lock_released_on_cancellation_during_release`) where the lease hangs on the explicit release and the best-effort drop release is what actually frees the lock. Verified it fails against the old `take()`-before-await code and passes with the fix.

BubbleCal (approved):

PR #6774, opened 2026-05-13
fix: set _row_created_at_version to new version for MERGE INTO INSERT rows

### LINKED ISSUE #6735: Newly inserted rows in a fragment-rewrite commit (e.g. MERGE INTO) get _row_created_at_version from a prior version instead of the commit version
## Summary
When a single commit both rewrites existing rows and adds new rows in the same operation (the path exercised by Spark `MERGE INTO` with mixed UPDATE / NOT MATCHED branches), the newly added rows report `_row_created_at_version` from a prior version — observed as `v1` (the table create version) — instead of the commit version.

This means CDF consumers cannot distinguish merge-inserted rows from rewritten-existing rows by `_row_created_at_version`. Plain `INSERT INTO` after a baseline insert correctly attributes `_row_created_at_version` to the commit version (see `BaseCdfVersionTrackingTest.testInsertIntoExistingTableUsesCurrentVersion` in lance-spark), so the gap is specific to the combined rewrite-and-insert commit path.

## Repro
Using lance-spark v0.5.0 on `lance-core` v7.0.0-beta.7:

1. Create a stable-row-ids table.
2. `INSERT INTO t VALUES (1,10), (2,20), (3,30), (4,40)` → all rows have `(created=2, last_updated=2)`.
3. `MERGE INTO t USING source ON id` with branches:
   - `WHEN MATCHED AND s.value IS NULL THEN DELETE`  (id=2)
   - `WHEN MATCHED THEN UPDATE SET ...`              (id=1)
   - `WHEN NOT MATCHED THEN INSERT ...`              (id=5, id=6)
4. Read version columns:
   - id=1 (updated):       `(created=1, last_updated=3)`
   - id=3, id=4 (untouched): `(created=2, last_updated=2)`  ← preserved correctly
   - id=5, id=6 (inserted): `(created=1, last_updated=3)`  ← should be `(3, 3)`

The inserted rows share version metadata with rewritten rows, which makes them indistinguishable from updates for CDF purposes.

See pinning test `BaseMergeIntoTest.testMergeIntoTracksVersionColumnsPerBranch` in lance-format/lance-spark.

## Expected
Newly added rows in a combined rewrite-and-insert commit should report `_row_created_at_version` equal to the commit version, matching standalone `INSERT` semantics. `_row_last_updated_at_version` should equal `_row_created_at_version` for new rows.

## Environment
- lance-core: 7.0.0-beta.7
- lance-spark: 0.5.0

jerryjch:
## Summary

* Fixes lance-format/lance#6735.
* `transaction.rs` — `resolve_update_version_metadata`: in the per-row `created_at`
  mapping, check `row_id_to_source.contains_key(&rid)` before calling
  `resolve_created_at_version`. Rows not in the map are INSERT branch rows (no source
  in existing fragments); they now receive `new_version` as `created_at` instead of the
  previous fallback of `UNKNOWN_CREATED_AT_VERSION` (1).
* `resolve_created_at_version` doc comment updated to clarify it is only called for
  UPDATE branch rows (source confirmed present). The unmapped-row-ID branch inside the
  function is unused when called from `resolve_update_version_metadata`; `UNKNOWN` (1)
  still applies for UPDATE rows whose source fragment has missing or bad
  `created_at_version_meta` (cache miss, decode failure, or out-of-range offset).
* Two existing tests updated to assert the corrected behavior; one new test added.

## Background

MERGE INTO commits through `Operation::Update` and produces both UPDATE branch rows
(rewritten into `new_fragments` with a source row in the previous manifest, stable row
ID present in `row_id_to_source`) and INSERT branch rows (new rows also in
`new_fragments`, stable row ID assigned fresh, not present in any existing fragment).

Before this change, `resolve_update_version_metadata` built the per-row
`created_at_versions` vector by calling `resolve_created_at_version` for every row ID.
For UPDATE branch rows that function correctly copies `created_at` from the source
fragment. For INSERT branch rows the map lookup fails and the function returns
`UNKNOWN_CREATED_AT_VERSION = 1`, producing a wrong historical version for every newly
inserted row. CDF consumers cannot distinguish merge-inserted rows from updated rows via
`_row_created_at_version`, and the value 1 is meaningless for rows that first appeared
in a recent commit.

The fix is a single guard at the call site: only call `resolve_created_at_version` for
rows confirmed to have a source (UPDATE branch); for all other rows use `new_version`
directly.

## Implementation notes

* The guard uses `row_id_to_source.contains_key(&rid)`, which is an O(1) hash lookup
  on the same map already built for the UPDATE branch path — no additional data
  structures or iteration.
* No lance-spark changes are needed. The Spark commit path (SparkPositionDeltaWrite)
  already attaches `RowIdMeta` to new fragment rows. This change activates the correct
  behavior automatically for all callers of `Operation::Update`, including lance-spark
  MERGE INTO.
* lane-spark test update https://github.com/lance-format/lance-spark/pull/530

## Test plan

* `test_update_version_tracking_insert_branch_gets_new_version` (renamed from
  `test_update_version_tracking_unknown_row_id_defaults_to_1`): new fragment with one
  UPDATE branch row (ID 10, source `created_at = 5`) and one INSERT branch row (ID 999);
  asserts `created_at = [5, 5]` — UPDATE branch copies from source, INSERT branch gets
  `new_version` (5).
* `test_update_version_tracking_merge_into_distinguishes_insert_and_update_branch` (new):
  new fragment interleaves UPDATE branch rows (IDs 10, 11, source `created_at = 3`) and
  INSERT branch rows (IDs 500, 501); asserts `created_at = [3, 5, 3, 5]` to verify
  per-row correctness across both branches in the same fragment.
* `test_update_version_tracking_no_row_id_meta_fallback`: assertion updated from
  `[1, 1, 1]` to `[5, 5, 5]` — a fragment with no `row_id_meta` gets fresh stable IDs
  assigned by `assign_row_ids`; those IDs have no source and are INSERT branch rows,
  so `created_at` equals `new_version`.
* `test_update_version_tracking_source_fragment_no_created_at_defaults_to_1` (unchanged):
  confirms that UPDATE branch rows whose source fragment has no `created_at_version_meta`
  still fall back to `UNKNOWN` (1) — the remaining reachable path through
  `resolve_created_at_version`.


jerryjch:
cc @hamersaw

hamersaw (approved):
Thanks for the fix, looks great!

wjones127 (approved):
Thanks for working this. These changes make sense 👍

PR #6778, opened 2026-05-14
perf(jni): parallelise inner_get_zonemap_stats segment reads

LuciferYang:
## Summary

`inner_get_zonemap_stats` on the JNI side opens and reads N segment files sequentially when the zonemap index has N per-fragment segments. On cold object storage this serialises N round-trips and dominates plan-time latency at high fragment counts.

This switches the loop to `futures::stream::buffered` with the concurrency window set to `object_store.io_parallelism()`, matching the BTree scalar-index reader's existing pattern. An explicit `fragment_id` sort upstream of the stream keeps the returned ordering stable for Java callers.

No behaviour change beyond parallelism; existing tests cover correctness.

## Notes

- Adds `futures = "0.3"` to `java/lance-jni/Cargo.toml`.
- The filename `"zonemap.lance"` is inlined as a literal for now; a follow-up can swap it to a public `ZONEMAP_FILENAME` constant once that lands.


java/lance-jni/Cargo.toml
```diff
@@ -39,6 +39,8 @@ tokio = { version = "1.23", features = [
     "sync",
 ] }
 async-trait = "0.1"
+futures = "0.3"
+snafu = "0.9"
```
Xuanwo: The manifest adds `snafu` as a direct dependency but the committed lockfile does not contain the matching direct edge. Locked JNI Rust builds fail before compilation, and non-locked builds can silently rewrite the lockfile during packaging.

java/lance-jni/src/blocking_dataset.rs
```diff
@@ -3524,27 +3524,54 @@ fn inner_get_zonemap_stats<'local>(
...
+                    // preserves the fragment-id order established by the sort above
+                    // while limiting in-flight requests to a fixed window.
+                    use futures::stream::{self, StreamExt, TryStreamExt};
+                    // Concurrency window for the per-segment open+read stream — defer to the
+                    // object store's own io_parallelism, matching the BTree scalar-index
+                    // reader's precedent (`.buffered(self.store.io_parallelism())`). Local
+                    // stores default to a small value; cloud stores default to ~64 with
+                    // environment overrides. Memory-constrained callers can dial this down
+                    // via lance-io's storage options instead of a hardcoded cap here.
+                    let max_concurrent_segment_reads = dataset
```
Xuanwo: The concurrency window is taken from the primary dataset store while the segment reads are performed through each index segment's resolved store. Base-scoped or shallow-cloned indices can use a different backend, so the code can under-parallelize remote reads or apply remote-level concurrency to local index files.

java/lance-jni/Cargo.toml
LuciferYang: Good catch. `snafu` was an unused leftover — nothing in `lance-jni/src` references it — so I dropped it from `Cargo.toml` entirely rather than add a lockfile edge for a dependency we don't use. `futures` (which the change does use) keeps its direct edge in the lockfile.

While reconciling I also noticed `java/lance-jni/Cargo.lock` was already missing the `rayon` direct edge for `lance-linalg` on `main` (pre-existing staleness, unrelated to this PR), which on its own makes `--locked` fail. Synced that too, so `cargo check --locked --manifest-path java/lance-jni/Cargo.toml` now passes cleanly.

java/lance-jni/src/blocking_dataset.rs
LuciferYang: Fixed. The window is now derived from the segments' own resolved index stores, not the primary dataset store. I resolve each segment's `LanceIndexStore` first (cheap — store-handle resolution + dedup through the session registry, no segment-file I/O, so it doesn't reintroduce the sequential round-trips this change removes), then take `max(io_parallelism())` across them as the `buffered` width.

That matches the BTree reader's `self.store.io_parallelism()` precedent and stays correct when a base-scoped or shallow-cloned index resolves to a different backend than `dataset.object_store(None)`. Using the max across segment stores (rather than a per-store window) avoids throttling a remote backend down to a co-located local index's small window, since `buffered` takes a single window for the whole stream.

LuciferYang:
Closing this to trim my open PRs. It is not obsolete: inner_get_zonemap_stats still reads the per-fragment segments sequentially on main, so the change still applies. It has just sat without review for four months. If cold-storage plan-time latency on high-fragment zonemap columns becomes a concern, the sequential loop in blocking_dataset.rs is still where to fix it, and this can be reproposed.

PR #6779, opened 2026-05-14
feat(zonemap): public API + driver helper for build-time consolidation

LuciferYang:
## Summary

Currently any external consumer that wants to build a zonemap from multiple workers has to either live in-tree (to reach internal items) or write one segment per fragment, ending up with N `IndexMetadata` entries per logical index. This PR exposes the surface needed to consolidate worker-produced batches into a single committed segment.

## What's new

In `lance-index::scalar::zonemap`:

- `pub const ZONEMAP_FILENAME` — canonical filename inside the index directory
- `pub const ZONEMAP_INDEX_VERSION: u32` — on-disk format version
- `pub fn zonemap_stats_schema(value_type)` — canonical Arrow schema factory
- `pub fn validate_zonemap_stats_schema(schema)` — strict structural validator
- `pub fn write_zonemap_index_from_batch(batch, params, store)` — write a `zonemap.lance` from a pre-computed batch, no training phase
- `ZoneMapIndexBuilder::zonemap_stats_as_batch` promoted to `pub`

In `lance::index::scalar`:

- `pub async fn compute_zonemap_batch(...)` — worker side: compute zone stats for a fragment subset without writing
- `pub async fn write_consolidated_zonemap_segment(...)` — driver side: take concatenated worker batches, write one `zonemap.lance` under a fresh UUID, return the `IndexMetadata` ready for `commit_existing_index_segments`

## Coordinator pattern

```rust
// Per worker, on a fragment subset:
let batch = compute_zonemap_batch(ds, column, &frag_ids, &params).await?;

// On the driver, after collecting the worker batches:
let segment = write_consolidated_zonemap_segment(ds, name, column, &batches, &params).await?;
ds.commit_existing_index_segments(name, column, vec![segment]).await?;
```

Result: one `IndexMetadata` entry covering every input fragment, not N.

## Input validation

Because these entry points are now public, they reject malformed or stale input up front instead of writing a segment that silently corrupts query results:

- `write_consolidated_zonemap_segment` cross-checks the batch's min/max value type against the column's **scanned** type — re-derived the same way `compute_zonemap_batch` does, so a legitimately adapted column (dict→primitive, extension unwrap) isn't false-rejected. A mismatched type would otherwise prune matching rows via cross-type `ScalarValue` comparisons.
- `write_consolidated_zonemap_segment` rejects `fragment_id`s that aren't in the dataset's current fragments, so a stale or foreign batch can't return commit-ready metadata claiming coverage the dataset doesn't have.
- `write_zonemap_index_from_batch` validates the zone bounds: `rows_per_zone != 0` (matching the invariant `ZoneTrainer` enforces on the training path), `zone_length >= 1`, and `zone_start + zone_length` within the fragment's 2^32 row-offset space — the read path computes a zone's row range as `(fragment_id << 32) + zone_start .. + zone_length`, so a start or length outside the fragment would produce wrong row-address ranges. (`zone_length` is the offset span `last - first + 1`, not a row count, so it is intentionally **not** capped at `rows_per_zone` — after deletions a zone of `rows_per_zone` live rows can span a wider offset range.)

## Tests

- 7 unit tests for `validate_zonemap_stats_schema` covering each rejection branch and the documented permissiveness on min/max nullability.
- 5 unit tests for the zone-bound validation: valid bounds, zero `rows_per_zone`, zero-length zone, a span wider than `rows_per_zone` (accepted — the post-deletion case), and per-fragment offset overflow.
- Round-trip test for `write_zonemap_index_from_batch` using `rows_per_zone=4` so each fragment contributes multiple zones — the default `rows_per_zone=8192` produces only one zone per fragment and hides multi-zone bugs.
- End-to-end test for `write_consolidated_zonemap_segment` covering fragment-bitmap completeness, schema-derived field id, `index_version`, and file presence under the freshly written UUID directory.
- 2 rejection tests for `write_consolidated_zonemap_segment`: a min/max type that doesn't match the column, and a `fragment_id` outside the dataset.

Additive only — no breaking changes.

### DIFF
--- a/rust/lance-index/src/scalar/zonemap.rs
+++ b/rust/lance-index/src/scalar/zonemap.rs
@@ -48,9 +48,34 @@ use roaring::RoaringBitmap;
 use super::zoned::{ZoneBound, ZoneProcessor, ZoneTrainer, rebuild_zones, search_zones};
 const ROWS_PER_ZONE_DEFAULT: u64 = 8192; // 1 zone every two batches
 
-const ZONEMAP_FILENAME: &str = "zonemap.lance";
+/// Filename of the on-disk zonemap index file written under `<index-uuid>/`. Public because
+/// external coordinators (e.g. JNI bindings, custom writers) need to refer to it by the same
+/// name the read path expects.
+pub const ZONEMAP_FILENAME: &str = "zonemap.lance";
 const ZONEMAP_SIZE_META_KEY: &str = "rows_per_zone";
 const ZONEMAP_INDEX_VERSION: u32 = 0;
+
+/// Canonical schema of a zone-stats record batch. The on-disk `zonemap.lance` file's record
+/// layout, the in-memory `ZoneMapIndexBuilder::zonemap_stats_as_batch` output, and any externally-
+/// produced batch fed into [`write_zonemap_index_from_batch`] all share this schema. Treat the
+/// column names and order as a stability contract: the on-disk reader matches columns by name,
+/// and [`validate_zonemap_stats_schema`] enforces full conformance up front so writes whose
+/// batch deviates from this shape fail before producing an unloadable file.
+///
+/// `min` / `max` are nullable because an entire batch may be all-NULL; the caller supplies the
+/// indexed column's type via `value_type`. The remaining metadata columns (`null_count`,
+/// `nan_count`, `fragment_id`, `zone_start`, `zone_length`) have fixed types and are non-nullable.
+pub fn zonemap_stats_schema(value_type: &DataType) -> Arc<arrow_schema::Schema> {
+    Arc::new(arrow_schema::Schema::new(vec![
+        Field::new("min", value_type.clone(), true),
+        Field::new("max", value_type.clone(), true),
+        Field::new("null_count", DataType::UInt32, false),
+        Field::new("nan_count", DataType::UInt32, false),
+        Field::new("fragment_id", DataType::UInt64, false),
+        Field::new("zone_start", DataType::UInt64, false),
+        Field::new("zone_length", DataType::UInt64, false),
+    ]))
+}
 
 /// Basic stats about zonemap index
 #[derive(Debug, PartialEq, Clone)]
@@ -765,7 +790,14 @@ impl ZoneMapIndexBuilder {
         Ok(())
     }
 
-    fn zonemap_stats_as_batch(&self) -> Result<RecordBatch> {
+    /// Drain the in-memory zone statistics into a record batch with the canonical zonemap schema
+    /// (`min`, `max`, `null_count`, `nan_count`, `fragment_id`, `zone_start`, `zone_length`).
+    ///
+    /// Public so external coordinators consolidating per-fragment zone batches from parallel
+    /// workers can extract the trained state without forcing the in-place file write that
+    /// [`Self::write_index`] performs. The companion [`write_zonemap_index_from_batch`] free
+    /// function consumes the same shape.
+    pub fn zonemap_stats_as_batch(&self) -> Result<RecordBatch> {
         // Flush self.maps as a RecordBatch
         let mins = if self.maps.is_empty() {
             new_empty_array(&self.items_type)
@@ -791,16 +823,7 @@ impl ZoneMapIndexBuilder {
         let zone_starts =
             UInt64Array::from_iter_values(self.maps.iter().map(|stat| stat.bound.start));
 
-        let schema = Arc::new(arrow_schema::Schema::new(vec![
-            // min and max can be null if the entire batch is null values
-            Field::new("min", self.items_type.clone(), true),
-            Field::new("max", self.items_type.clone(), true),
-            Field::new("null_count", DataType::UInt32, false),
-            Field::new("nan_count", DataType::UInt32, false),
-            Field::new("fragment_id", DataType::UInt64, false),
-            Field::new("zone_start", DataType::UInt64, false),
-            Field::new("zone_length", DataType::UInt64, false),
-        ]));
+        let schema = zonemap_stats_schema(&self.items_type);
 
         let columns: Vec<ArrayRef> = vec![
             mins,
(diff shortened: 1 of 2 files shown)


rust/lance/src/index/scalar.rs
```diff
@@ -225,6 +225,136 @@ pub(crate) async fn load_training_data(
...
+    params: &lance_index::scalar::zonemap::ZoneMapIndexBuilderParams,
+) -> Result<IndexMetadata> {
+    use arrow_array::cast::AsArray;
+    use arrow_array::types::UInt64Type;
+    use lance_index::scalar::IndexStore;
+    use lance_index::scalar::zonemap::{ZONEMAP_INDEX_VERSION, write_zonemap_index_from_batch};
+    use roaring::RoaringBitmap;
+    use uuid::Uuid;
+
+    // Validate the indexed column exists in the dataset schema and capture its field id for
```
Xuanwo, line 312: The helper accepts a stats batch whose min/max type does not match the indexed column. Such a segment can prune matching rows because zonemap comparisons return false for cross-type ScalarValue comparisons.

rust/lance-index/src/scalar/zonemap.rs
```diff
@@ -734,20 +761,118 @@ impl ZoneMapIndexBuilder {
...
+/// that consolidate per-fragment zone batches produced by parallel workers can write a single
+/// consolidated file without re-running the train phase. `params` is taken by reference rather
+/// than `rows_per_zone: u64` so future ZoneMap knobs that affect on-disk metadata can extend
+/// the parameter set without breaking this signature.
+pub async fn write_zonemap_index_from_batch(
+    record_batch: RecordBatch,
+    params: &ZoneMapIndexBuilderParams,
+    index_store: &dyn IndexStore,
+) -> Result<()> {
+    validate_zonemap_stats_schema(record_batch.schema().as_ref())?;
```
Xuanwo, line 871: The batch writer validates only column shape and trusts zone bounds from the caller. Invalid zone_start or zone_length values can create wrong row-address ranges or huge ranges at query time.

rust/lance-index/src/scalar/zonemap.rs
```diff
@@ -734,20 +761,118 @@ impl ZoneMapIndexBuilder {
...
+
+    let mut file_schema = record_batch.schema().as_ref().clone();
+    file_schema.metadata.insert(
+        ZONEMAP_SIZE_META_KEY.to_string(),
+        params.rows_per_zone().to_string(),
```
Xuanwo, line 877: The batch writer persists rows_per_zone without enforcing the nonzero invariant used by normal training. It can create a zonemap that loads successfully but fails on later update or rebuild.

rust/lance/src/index/scalar.rs
```diff
@@ -225,6 +225,136 @@ pub(crate) async fn load_training_data(
...
+            "consolidated zonemap batch missing 'fragment_id' column".into(),
+        )
+    })?;
+    let frag_array = frag_col.as_primitive_opt::<UInt64Type>().ok_or_else(|| {
+        Error::invalid_input_source(
+            "consolidated zonemap batch 'fragment_id' must be UInt64".into(),
+        )
+    })?;
+    let mut fragment_bitmap = RoaringBitmap::new();
+    for f in frag_array.values() {
```
Xuanwo, line 374: The fragment bitmap is derived only from batch contents and is not checked against the current dataset fragments. A stale or foreign batch can return commit-ready metadata that claims coverage outside the dataset.

rust/lance/src/index/scalar.rs
LuciferYang: Good point. Added a type check up front: it re-derives the column's scanned type the same way `compute_zonemap_batch` does (so dict→primitive and extension columns still pass) and rejects the batch when min/max don't match, instead of letting the cross-type comparison silently prune matching rows. Covered by `test_write_consolidated_zonemap_rejects_value_type_mismatch`.

rust/lance-index/src/scalar/zonemap.rs
LuciferYang: Agreed. The writer now validates the zone bounds before writing: `zone_length` must be within `1..=rows_per_zone`, and `zone_start + zone_length` must stay inside the fragment's 2^32 offset space — the read path turns each zone into `(fragment_id << 32) + zone_start .. + zone_length`, so a bad length or an offset that spills past the fragment boundary would otherwise become a wrong/huge range at query time. The zero-length, over-capacity, and overflow cases all have tests.

rust/lance-index/src/scalar/zonemap.rs
LuciferYang: Right — added the same `rows_per_zone != 0` guard that `ZoneTrainer::new` already enforces on the training path, so we can't persist a zone size of 0 that loads fine but breaks on the next update/rebuild. Test: `test_validate_zone_bounds_rejects_zero_rows_per_zone`.

rust/lance/src/index/scalar.rs
LuciferYang: Makes sense. It now builds the set of the dataset's current fragment ids and rejects any `fragment_id` in the batch that isn't one of them, so a stale or foreign batch can't hand back commit-ready metadata claiming fragments the dataset doesn't have. Test: `test_write_consolidated_zonemap_rejects_foreign_fragment`.

rust/lance-index/src/scalar/zonemap.rs
LuciferYang: Correcting my note above — I had the `zone_length` bound wrong. `zone_length` is the offset span `(last - first + 1)`, not the live-row count, so with deletions a zone of `rows_per_zone` live rows can span a wider offset range and legitimately exceed `rows_per_zone` (CI caught this on `test_zonemap_with_deletions`). I dropped the `<= rows_per_zone` check; the writer now validates only what's actually invariant — `zone_length >= 1` and `zone_start + zone_length <= 2^32` so a zone can't spill past its fragment's offset space — plus a regression test asserting a wider-than-`rows_per_zone` span is accepted.

LuciferYang:
Both halves of this landed upstream by a different route. On the worker side, the create-index builder gained `fragments` (`rust/lance/src/index/create.rs:125`) and `execute_uncommitted` (`:159`), so a caller can build one segment over a fragment subset without committing it. On the driver side, `merge_existing_index_segments` has a zonemap arm (`rust/lance/src/index.rs:1990`) that goes through `merge_zonemap_indices` (`rust/lance-index/src/scalar/zonemap.rs:982`) and produces a single `IndexMetadata` covering every input fragment.

That removes the premise: a consumer outside the crate no longer has to pick between living in-tree and ending up with one `IndexMetadata` per fragment. `rust/lance/benches/fts_search.rs:140-149` drives that exact shape through the public API, `create_index_builder(...).fragments(...).execute_uncommitted()` per worker followed by one `commit_existing_index_segments`.

One difference, not enough to keep this open: upstream's `merge_segments` writes each segment and reads it back rather than consolidating batches in memory. Closing as superseded.

PR #6782, opened 2026-05-14
fix(index): normalize same-column scalar index predicates

zouhuajian:
## Summary

This PR builds on #7477 instead of introducing a separate scalar-index normalization path.

#7477 added `ScalarIndexExpr::optimize()`, merged half-open range predicates, and applied the optimizer before scalar-index execution. This PR extends that optimizer with the remaining rules and edge-case handling originally proposed in #6782:

- accumulate planner-produced same-key ranges within one AND region and keep the tightest bounds
- remove an exact `IS NOT NULL` term when another exact same-key predicate is already null-intolerant
- preserve conservative boundaries around OR branches, NOT context, NULL values, index identity, recheck requirements, and fragment coverage

This refactoring follows the review request to describe the behavior as optimizer rules and reuse the optimizer introduced by #7477.

## Behavior

Within one contiguous AND region:

- compatible `SargableQuery::Range` predicates with the same column, index name, and index type are intersected
- the merged range keeps `needs_recheck = true` if any input range requires recheck, preserving #7477 behavior
- ranges merge only when their fragment coverage values match; differing or known-versus-unknown coverage remains separate
- reverse bounds such as `x >= 200 AND x <= 100` remain a valid empty range representation
- ranges with incomparable values or NULL bounds remain separate
- an exact `IS NOT NULL` predicate is removed only when another exact same-key predicate is null-intolerant
- recheck predicates cannot remove `IS NOT NULL`
- OR branches are optimized independently
- range merging remains enabled inside NOT subtrees, while `IS NOT NULL` elimination is disabled there to preserve SQL NULL semantics

## Complexity

Each AND region is flattened once and processed with keyed accumulator positions in a single pass. This replaces the previous pairwise scan with expected O(n) leaf processing and derives the tightest range across planner-produced compatible terms.

The implementation moves leaves into stable output slots instead of cloning complete `ScalarIndexSearch` values. The AND tree is rebuilt in balanced form to avoid deep planner-generated conjunctions.

## Validation

Coverage includes:

- tightest bounds across multiple same-key ranges
- inclusive and exclusive bounds
- empty ranges
- `needs_recheck` propagation
- matching, differing, and unknown fragment coverage
- exact `IS NOT NULL` elimination
- standalone NULL checks
- different columns and index identities
- `IN` predicates with and without NULL
- `!=` predicates
- OR and NOT boundaries
- parser-generated predicates
- large balanced AND regions
- a runnable public `optimize()` example

Validated locally with:

- `cargo test -p lance-index --lib`
- `cargo test -p lance-index --doc 'scalar::expression::ScalarIndexExpr::optimize'`
- `cargo test -p lance io::exec::scalar_index --lib`
- `cargo test -p lance io::exec::count_pushdown --lib`
- `cargo clippy -p lance-index --all-targets -- -D warnings`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved scalar filter optimization to merge range predicates only when index/metadata context matches, reducing incorrect results.
  * Added safer handling for incomparable range bounds and ranges involving NULLs.
  * Strengthened optimization behavior within `NOT` expressions to preserve SQL NULL semantics, including more conservative `IS NOT NULL` removal.
  * Refined merged-range recheck behavior, including cases where the merged intersection becomes empty.
* **Tests**
  * Updated and added unit tests for metadata gating, NULL-bound scenarios, and empty-intersection merging behavior.

### DIFF
--- a/rust/lance-index/src/scalar/expression.rs
+++ b/rust/lance-index/src/scalar/expression.rs
@@ -1,7 +1,12 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-FileCopyrightText: Copyright The Lance Authors
 
-use std::{ops::Bound, sync::Arc};
+use std::{
+    cmp::Ordering,
+    collections::{HashMap, HashSet, hash_map::Entry},
+    ops::Bound,
+    sync::Arc,
+};
 
 use arrow_schema::{DataType, Field};
 use async_recursion::async_recursion;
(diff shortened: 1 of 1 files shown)


zouhuajian:
The metrics/logging added in [#6759](https://github.com/lance-format/lance/pull/6759) makes this change easier to observe and validate: after same-column predicate normalization, the scalar index planner should select a much narrower candidate part range for bounded range queries.

zouhuajian:
@Xuanwo Could you help review this when you have time? Thanks.

rust/lance-index/src/scalar/expression.rs
```diff
@@ -1370,6 +1443,164 @@ pub fn serialize_index_expr_result(
...
+    fn merge_single_sided_ranges(mut terms: Vec<Self>) -> Vec<Self> {
+        let mut term_idx = 0;
+        while term_idx < terms.len() {
+            let Some((range_key, range_side, search)) = terms[term_idx].exact_single_sided_range()
+            else {
+                term_idx += 1;
+                continue;
+            };
+
+            let matching_idx = ((term_idx + 1)..terms.len()).find(|candidate_idx| {
```
Xuanwo: The range merge stops after pairing a bound with the first opposite bound instead of deriving the tightest range across the same-key conjunction. Filters with duplicate lower or upper bounds can still evaluate a wider bounded branch plus extra single-sided branches, leaving part of the broad-branch cost this change is meant to remove.

rust/lance-index/src/scalar/expression.rs
```diff
@@ -1370,6 +1443,164 @@ pub fn serialize_index_expr_result(
...
+            Self::Or(lhs, rhs) => Self::Or(
+                Box::new(lhs.normalize_exact_scalar_terms()),
+                Box::new(rhs.normalize_exact_scalar_terms()),
+            ),
+            Self::Not(inner) => Self::Not(inner),
+            Self::Query(search) => Self::Query(search),
+        }
+    }
+
+    fn normalize_exact_scalar_terms_from_and(lhs: Self, rhs: Self) -> Self {
```
Xuanwo: The normalization pass repeatedly clones, flattens, linearly scans, and removes terms while the existing guard limits nesting depth rather than total AND terms. Large balanced filters can make scalar-index planning consume quadratic CPU and elevated memory before execution starts.

Xuanwo (commented):
1. The range merge stops after pairing a bound with the first opposite bound instead of deriving the tightest range across the same-key conjunction. Filters with duplicate lower or upper bounds can still evaluate a wider bounded branch plus extra single-sided branches, leaving part of the broad-branch cost this change is meant to remove.
2. The normalization pass repeatedly clones, flattens, linearly scans, and removes terms while the existing guard limits nesting depth rather than total AND terms. Large balanced filters can make scalar-index planning consume quadratic CPU and elevated memory before execution starts.

rust/lance-index/src/scalar/expression.rs
zouhuajian: I’ve updated the range normalization to accumulate all exact range predicates for the same scalar-index identity within one contiguous AND region, then emit the tightest safe intersection instead of pairing only the first opposite bound.

rust/lance-index/src/scalar/expression.rs
zouhuajian: I’ve addressed this by changing the AND normalization path to flatten each contiguous AND region once, then normalize terms with keyed range accumulators and stable output slots. This avoids repeated subtree flattening, linear scans, and Vec::remove shifting for large balanced filters.

wjones127 (changes requested):
In https://github.com/lance-format/lance/pull/7477, we added a `ScalarIndexExpr::optimize` method. I think this would be a good case to re-use that. Would you be willing to refactor around using that? These all look like they could be described as optimizer rules.

zouhuajian:
> In #7477, we added a `ScalarIndexExpr::optimize` method. I think this would be a good case to re-use that. Would you be willing to refactor around using that? These all look like they could be described as optimizer rules.

Thanks for the suggestion. I’ve refactored the changes to extend `ScalarIndexExpr::optimize()` introduced in #7477 instead of maintaining a separate normalization path.

The additional behaviors are now implemented as optimizer rules within each AND region, including redundant `IS NOT NULL` elimination and tighter same-key range accumulation.

rust/lance-index/src/scalar/expression.rs
```diff
@@ -1476,173 +1481,346 @@ impl PartialEq for ScalarIndexExpr {
...
+    /// Range queries for the same parsed scalar index are intersected and retain
+    /// `needs_recheck` if any input range requires it. An exact `IS NOT NULL`
+    /// query is removed when another exact same-index query is null-intolerant.
+    /// OR branches are optimized independently, while NOT subtrees remain opaque.
     pub fn optimize(self) -> Self {
         match self {
             Self::And(_, _) => self.optimize_and_tree(),
             Self::Or(lhs, rhs) => Self::Or(Box::new(lhs.optimize()), Box::new(rhs.optimize())),
-            Self::Not(inner) => Self::Not(Box::new(inner.optimize())),
+            Self::Not(inner) => Self::Not(inner),
```
wjones127: question(blocking): why undo the `Not` handling here? Was there a bug you found in existing behavior? Or would new rules not work properly within a `Not` expression?

wjones127 (changes requested):

rust/lance-index/src/scalar/expression.rs
zouhuajian: There wasn’t a bug in the existing range optimization under `Not`. I initially made the subtree opaque because `IS NOT NULL` elimination is unsafe under negation. For example, when `x` is NULL, `NOT (x IS NOT NULL AND x >= 10)` evaluates to `TRUE`, but removing `x IS NOT NULL` changes it to `UNKNOWN`, affecting `WHERE` results. I’ve updated it to disable only that rule under `Not` while preserving safe range merging, with regression tests for both cases.

wjones127 (approved):

PR #6783, opened 2026-05-14
fix(encoding): scope blob field-data cache key by projection mode (#6381)

### LINKED ISSUE #6381: Appending dataset corrupts blob columns
## Repro

When appending rows to an existing Lance dataset that has large_binary blob columns (with lance-encoding:blob metadata), the appended fragment stores those columns as struct<position: uint64, size: uint64> instead of large_binary — even when passing the correct schema explicitly with data_storage_version="stable".

## Consequences:

`ds.take_blobs()` fails on rows from the appended fragment
`ds.optimize.compact_files()` fails with schema mismatch between the original and appended fragments
The dataset looks fine superficially (row counts correct, to_table() works) but blob reads are broken
Repro: write a dataset with blob columns using mode="overwrite", then append more rows with mode="append" using the same schema. Inspect `ds.schema.field("rgb").type` — it'll say large_binary (from the first fragment), but rows in the second fragment are actually struct-encoded.

Workaround: Always write all rows in a single write_dataset(mode="overwrite") call. Never use mode="append" with blob columns.

shenganzhang:
## Summary

Closes #6381.

A blob column can be scheduled in two incompatible projection modes:

- **Descriptor** — `struct<position, size>`, used by `take_blobs` / `take_blobs_by_indices`. Selects `BlobDescriptionPageScheduler`, which caches the inner scheduler's state directly.
- **Data** — `LargeBinary`, used by full scans and `compact_files`. Selects `BlobPageScheduler`, which wraps the inner state in a `BlobCacheableState { positions, sizes, inner_state }`.

The two modes produce structurally incompatible `CachedPageData`. However, the per-column field-data cache key in `StructuralPrimitiveFieldScheduler` was just `column_index`, so on a multi-fragment dataset the two modes collided on the same cache entry. Whichever scheduler ran first poisoned the cache for the other, and `BlobPageScheduler::load` would panic on the wrong-type downcast at `blob.rs:335`.

This is what #6381 surfaces as "appending corrupts blob columns": after `mode=\"append\"` the dataset has ≥2 fragments, and the typical `take_blobs` → `compact_files` flow trips the cache collision. Append is incidental — a single `Dataset::write` with `max_rows_per_file=1` followed by `take_blobs` and then `compact_files` reproduces exactly the same panic. The write path was always correct; only the read-side cache key was wrong.

## Fix

- Track `is_blob_descriptor_projection` on `StructuralPrimitiveFieldScheduler` (true iff `target_field.is_blob()` and its data type is `Struct`).
- Fold that bit into `FieldDataCacheKey` so the two projections produce distinct cache keys (`\"{col}/desc\"` vs `\"{col}\"`).

```rust
fn key(&self) -> std::borrow::Cow<'_, str> {
    if self.is_blob_descriptor_projection {
        format!(\"{}/desc\", self.column_index).into()
    } else {
        self.column_index.to_string().into()
    }
}
```

Diff is 23 lines of production code in `rust/lance-encoding/src/encodings/logical/primitive.rs`; no public API changes.

## Test plan

- [x] Added `test_append_blob_columns_preserves_blob_encoding` in `rust/lance/src/dataset/optimize.rs` — exact #6381 repro (overwrite + append + `take_blobs_by_indices` + `compact_files`).
- [x] Extended existing `test_compact_blob_columns` to call `take_blobs_by_indices` while the dataset still has multiple fragments (before compaction), covering the same cache-collision path in a single-write scenario.
- [x] Both regression tests **panic at `blob.rs:335`** on `main` (verified by stashing the fix), **pass with the fix**.
- [x] `cargo test -p lance --lib dataset::blob::` — 35/35 pass.
- [x] `cargo test -p lance --lib dataset::optimize::` — 62/62 pass.
- [x] `cargo test -p lance-encoding --lib` — 371/371 pass.
- [x] `cargo clippy -p lance-encoding -p lance --tests -- -D warnings` — clean.
- [x] `cargo fmt --check` — clean.

### DIFF
--- a/rust/lance-encoding/src/encodings/logical/primitive.rs
+++ b/rust/lance-encoding/src/encodings/logical/primitive.rs
@@ -3290,6 +3290,16 @@ struct PageInfoAndScheduler {
 pub struct StructuralPrimitiveFieldScheduler {
     page_schedulers: Vec<PageInfoAndScheduler>,
     column_index: u32,
+    // A blob column can be scheduled in two different "projection modes":
+    //   * blob descriptor (the column is projected as a struct of position/size) —
+    //     selects `BlobDescriptionPageScheduler` per page.
+    //   * blob data (the column is projected as the loaded large_binary value) —
+    //     selects `BlobPageScheduler` per page.
+    // The two modes produce incompatible `CachedPageData` shapes, so they must
+    // not share a cache entry (otherwise `BlobPageScheduler::load` will downcast
+    // the wrong type and panic).  Track the mode here so it can be folded into
+    // the field-data cache key.
+    is_blob_descriptor_projection: bool,
 }
 
 impl StructuralPrimitiveFieldScheduler {
@@ -3313,9 +3323,12 @@ impl StructuralPrimitiveFieldScheduler {
                 )
             })
             .collect::<Result<Vec<_>>>()?;
+        let is_blob_descriptor_projection =
+            target_field.is_blob() && matches!(target_field.data_type(), DataType::Struct(_));
         Ok(Self {
             page_schedulers,
             column_index: column_info.index,
+            is_blob_descriptor_projection,
         })
     }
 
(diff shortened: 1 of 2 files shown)


Xuanwo:
Thank you for working on this, I will take the impls from https://github.com/lance-format/lance/pull/6758 instead.

PR #6786, opened 2026-05-14
docs: change a typo of WAL entry filename suffix to '.arrow'

jiengup:
Updated filename suffix from '.lance' to '.arrow' for WAL entries.


jackye1995 (approved):
thanks for the fix!

PR #6787, opened 2026-05-14
fix(encoding): plan sparse structural miniblock pages

Xuanwo:
## Root cause

Mini-block page planning was effectively bounded by visible leaf values, but sparse nested lists can contain many rep/def structural events with few or no visible values. Those pages can fit the value budget while still overflowing mini-block's packed rep/def chunk metadata. Falling back to fullzip at the primitive value layer was the wrong level of fix because fullzip only applies to value layouts it can actually encode and should not become the generic escape hatch for structural streams.

## Fix

This moves the decision to the structural stream. During rep/def serialization, Lance now produces a structural page plan while normalizing levels. The primitive encoder consumes that plan and splits oversized structural pages on top-level row boundaries before mini-block encoding. Splits with no visible leaf values use the existing complex all-null layout, value-bearing splits continue through mini-block, and a single row that is itself too large falls back to fullzip only when the value layout is compatible; otherwise it returns a clear error.

This keeps the on-disk format compatible: it emits only existing page layouts and does not add a new layout variant or metadata field. It also preserves the format's random-access model by making structural pages independently addressable instead of treating sparse nested structure as a value-encoding special case.

## Validation

Targeted sparse nested list regressions and the full lance-encoding test suite pass. Microbenchmarks show no meaningful dense-path regression; the sparse boolean encode regression from the previous intermediate-summary approach is gone, and nested sparse string encoding remains faster.


westonpace:
@claude review once

rust/lance-encoding/src/repdef.rs
```diff
@@ -745,9 +800,212 @@ impl SerializerContext {
...
+            }
+            return Ok(StructuralPagePlan::Fits);
+        };
+
+        if num_values == 0 {
+            for def in self.def_levels.iter_mut() {
+                if *def > SPECIAL_THRESHOLD {
+                    *def -= SPECIAL_THRESHOLD;
+                }
+            }
```
westonpace: This is repeated several times and could maybe go in a `normalize_specials` method.

rust/lance-encoding/src/repdef.rs
```diff
@@ -745,9 +800,212 @@ impl SerializerContext {
         }
     }
 
+    fn normalize_specials_and_plan_splits(
```
westonpace, line 803: Can we add a comment here (ok if it's a bit wordy) explaining why we are doing this.  Something like...

When we have extremely sparse lists (mostly null elements) we could end up in a situation where there are many repdef levels per value.  If the repdef to data ratio becomes too skewed then the miniblock structural encoding can fail because a single chunk has too much repdef information.  As a result, we detect this sparse list case, and break data pages into smaller pages to avoid this.

westonpace (commented):
This is a significant change to the write path and write corruption is always a fear so I might do some additional manual testing of this fix.

That being said, the change seems to be very narrowly enacted.  We only use this on sparse lists. 

We are moving repdef serialization out of the spawned task so this means that work will be serialized.  This may have some detrimental impacts to write speed.  It would be good to run some write benchmarks just to understand what these are.

Performance-wise I do worry this could lead to runt pages.  We want to have MBs of data per page.  This helps avoid overly small reads when hitting object storage.  Since this planning is splitting pages at 2^16 values I think we will get rather tiny pages.

Perhaps in 2.3 we can go ahead and add a "midi-block encoding" where we don't try and fit all the chunk metadata into 2 bytes.  That would also help address other cases where users are asking for much larger compression windows.

Anyways, I'm sitting at a soft +1 for now.  I'll do some additional write testing and try and run some writer benchmarks just to see if there is any impact and then change this to approve.

westonpace:
I've filed #6841 to follow-up on a longer term fix.

rust/lance-encoding/src/repdef.rs
Xuanwo: Ah, nice. Updated.

rust/lance-encoding/src/repdef.rs
Xuanwo: Added!

westonpace:
> I'll do some additional write testing and try and run some writer benchmarks just to see if there is any impact and then change this to approve.

I ran through some write benchmarks and didn't see any noticeable difference.  I think in most typical cases the repdef time is very small so it doesn't matter.

westonpace (approved):
Forgot to check back in here.  The write benchmarks didn't show any noticeable impact.  Let's move forwards

PR #6788, opened 2026-05-14
perf(kmeans): speed up L2 centroid assignment with batched GEMM

xiaguanglei:
# perf(kmeans): speed up L2 centroid assignment with batched GEMM (~1.7–1.9× on PQ training)

## Summary

Replace the per-query `par_chunks` + `l2_distance_batch` loop in K-means **L2 centroid
assignment** with a single batched matrix multiplication (`ndarray::dot()` via
`matrixmultiply`). The hot path is **PQ codebook training** (`ProductQuantizerBuilder` →
`train_kmeans` per sub-vector); IVF centroid training and non-L2 metrics are unchanged.

**Measured end-to-end IVF_PQ index build** (Apple M-series ARM64, release, dim=128,
`nlist=64`, `pq_bits=8`, cosine, `sub_dim=8`, interleaved median):

| Scale | Speedup | Wall-time reduction |
|---|---|---|
| 50 k rows | **1.77×** | ~43% |
| 100 k rows | **1.67×** | ~40% |
| 512-d / 1024-d embed @ 50 k rows (`sub_dim` still 8) | **1.78–1.86×** | ~44–46% |

In short: **~1.7–1.9× faster IVF_PQ builds (~40–47% less wall time)** for standard PQ configs
at production row counts. Smaller datasets (≤ 20 k rows) see a smaller gain (~1.1–1.5×) because
IVF overhead dominates.

This optimization is active for **all standard PQ configurations** regardless of sub_dim: the
memory-budget gate (`N × K × 4 ≤ 64 MB`) covers K=256 (PQ fixed) at any typical sub_dim.
Bench confirms speedup at sub_dim=8/16/32/64/128 (1.30–1.61×). **No public API changes**;
distance-optimal (not bit-identical); production indexing behaviour and recall are unaffected.

IVF centroid training with `nlist ≤ 256` also benefits marginally (~1.1×); large `nlist=1024+`
is automatically excluded by the budget gate (200 MB > 64 MB) to avoid memory pressure.

### Activation: memory-budget threshold (reviewer suggestion)

SGEMM is now **always on** for L2 + f32 K-means when the intermediate N×K inner-product matrix
fits within a configurable memory budget (default **64 MiB**). There is no per-call opt-in flag.

| Mechanism | Details |
|---|---|
| Default budget | 64 MiB (`N × K × 4 bytes ≤ 67_108_864`) |
| Runtime override | `LANCE_SGEMM_THRESHOLD=<bytes>` — set to `0` to disable entirely |
| Old env vars | `LANCE_NO_SGEMM` / `LANCE_FORCE_SGEMM` **removed** |
| API surface | No public struct fields changed; `KMeansParams` and `PQBuildParams` are simpler |

For all standard PQ configs (`K=256`, `N ≤ 16 M sampling cap`), the matrix is ≤ 4 GB at most,
but the default 64 MiB cap already covers `N ≤ ~65 k` rows, which is the typical Lance sampling
size for PQ codebook training.

### Test fixes

Three Rust tests use synthetic data distributions that are sensitive to K-means tie-breaking.
Fix: set `LANCE_SGEMM_THRESHOLD=0` temporarily during index construction **only in those tests**
to keep them bit-deterministic. **Does not change production code, scanner logic, or any other
test.** SGEMM remains default-on everywhere else; PQ correctness is covered by existing
`lance-index` unit tests and integration tests.

---

## Motivation: where the cost comes from

**Code path (PQ index build).**  `ProductQuantizerBuilder` trains one K-means codebook per
sub-vector (typically 16 sequential `train_kmeans` calls for `dim=128`, `num_sub_vectors=16`).
Each iteration spends most of its time in `KMeansAlgoFloat::compute_membership_and_dist` for
`DistanceType::L2`: the baseline implementation fans out with `par_chunks(dimension)` and
calls `l2_distance_batch` once per training vector.  That is **O(N × K × D)** work per
iteration with poor reuse of centroid data in cache compared to a single batched matmul.

**Empirical gap.**  On the same harness and row count, IVF_FLAT build is orders of magnitude
cheaper than IVF_PQ because PQ adds the above sub-vector K-means loop.  A separate sweep that
only times the L2 membership step (scalar vs GEMM) across `sub_dim` shows a large gap at
typical `sub_dim` (see Benchmarks §2).

**Why the scalar path is hard to auto-vectorise.**  The inner distance loop has a
runtime-dependent trip count (`K × D`), so LLVM cannot reliably vectorise it the way a
fixed-shape GEMM kernel can.

---

### Math

$$\|a - b\|^2 = \|a\|^2 - 2\langle a, b \rangle + \|b\|^2$$

By computing the full inner-product matrix
$\text{inner}[i,j] = \langle \text{data}_i,\, \text{centroid}_j \rangle$
as a single GEMM, we:

1. **Eliminate N×K Rayon tasks** — the GEMM itself is one cache-tiled operation,
   followed by a simple parallel argmin over rows.
2. **Enable SIMD vectorisation** — `ndarray`'s `dot()` uses `matrixmultiply`
   (SSE2/AVX on x86-64, NEON on ARM64, SIMD128 on WASM).  **No vendor BLAS required.**
3. **Amortise memory bandwidth** — centroid norms `||c_j||²` are computed once (`O(K×D)`)
   and reused for all N queries, vs. recomputed inside each of the N original tasks.
4. **Cache-tiled inner product** — `matrixmultiply`'s Goto-style blocked algorithm keeps
   both the data tile and the centroid tile in L2/L3 cache, eliminating the random-access
   centroid reloads of the scalar path.

### Why L2 + f32 only?

Each guard condition is a separate gate:

| Condition | Reason | Is it required? |
|---|---|---|
| `DistanceType::L2` | The `‖a‖²−2⟨a,b⟩+‖b‖²` identity is valid **only for Euclidean (L2) distance**. Cosine uses normalization + L2 residuals — already benefits. Dot-product uses a different formula. | **Yes, correctness.** |
| `size_of::<T::Native>() == 4` (f32) | `compute_l2_sgemm` is typed `&[f32]`. f16 needs up-conversion; f64 is uncommon for vector search. Checked via `size_of` — zero runtime cost. | **Yes, PR scope.** f16/f64 support can follow. |
| `N × K × 4 ≤ budget` | **Performance gate.** Keeps the temporary inner-product matrix within an acceptable memory footprint. Large K (e.g., IVF with K=1024) would allocate hundreds of MB; the 64 MiB default cap safely excludes those while covering all standard PQ sub-vector training (K=256, N≤65 k). | **Yes, performance.** |

> **Note on `sub_dim`:** Unlike the previous implementation (`dim < 64` hard threshold), the new
> memory-budget gate is *not* tied to dimension size. A configuration with `sub_dim=128` but
> only N=100 rows would still activate SGEMM (100 × 256 × 4 = 100 KB ≤ 64 MB). The budget gate
> is the correct abstraction: it directly captures the performance-vs-memory trade-off.

### Implementation

`rust/lance-index/src/vector/kmeans.rs`:

```rust
fn compute_l2_sgemm(data, centroids, dimension, balance_factor, cluster_sizes)
    → (Vec<Option<u32>>, Vec<Option<f32>>)
```

Key steps inside `compute_l2_sgemm`:
1. Build zero-copy `ArrayView2` over existing `&[f32]` slices — no allocation.
2. Precompute `norms_c[j] = ||centroid_j||²` once: O(K×D).
3. **Single GEMM call:** `inner = data_mat.dot(&c_mat.t())` — shape `[N, K]`,
   backed by `matrixmultiply`'s cache-blocked SIMD kernel.
4. Parallel argmin over rows of `inner` via `(0..n).into_par_iter()`.
5. **NaN guard**: if `norm_q.is_nan()` (NaN query vector), return `(None, None)` to match the scalar path.

Fast-path guard at the top of `KMeansAlgoFloat::compute_membership_and_dist`:

```rust
const DEFAULT_THRESHOLD_BYTES: usize = 64 * 1024 * 1024; // 64 MiB
let threshold_bytes: usize = std::env::var("LANCE_SGEMM_THRESHOLD")
    .ok().and_then(|s| s.parse().ok())
    .unwrap_or(DEFAULT_THRESHOLD_BYTES);
let matrix_bytes = n * k * size_of::<f32>();
if index.is_none()
    && distance_type == DistanceType::L2
    && size_of::<T::Native>() == size_of::<f32>()
    && n > 0 && k > 0
    && threshold_bytes > 0
    && matrix_bytes <= threshold_bytes
{
    return compute_l2_sgemm(...);
}
// ... original scalar par_chunks path unchanged below
```

---

## Benchmarks

**Hardware:** Apple M-series 12-core (ARM64 NEON), release build with full LTO.

**Methodology:** End-to-end IVF_PQ index build. Fixed schema: dim=128, `nlist=64`, `pq_bits=8`,
cosine metric. Each config is measured with **4 interleaved pairs** (SGEMM run → scalar run within
each loop iteration) so thermal state and cache warmth affect both paths equally. Speedup =
`t_scalar_median / t_sgemm_median`. Min columns show the single fastest run for each path.

### 1. Row-size trend: larger N → bigger speedup (sub_dim=8, num_sub_vectors=16)

> **Principle:** As N grows, the per-iteration working set (N rows × 128-dim × f32 ≈ 50 MB at
> N=100k) exceeds L2 cache. The scalar path re-loads the 8 KB centroid block for every row.
> SGEMM amortises this across the entire N×K GEMM call. The allocation cost (≤ 64 MB `inner`
> matrix, bounded by `train_kmeans`'s sampling cap) is a one-time constant per iteration, so
> its share of total time shrinks as N grows.

| rows | GEMM median (s) | Scalar median (s) | **Speedup** | GEMM min (s) | Scalar min (s) |
|---|---|---|---|---|---|
| 5 000 | 0.262 | 0.300 | **1.14×** ~ | 0.213 | 0.254 |
| 10 000 | 0.556 | 0.740 | **1.33×** ~ | 0.502 | 0.672 |
| 20 000 | 1.115 | 1.619 | **1.45×** ~ | 0.953 | 1.562 |
| 50 000 | 2.683 | 4.741 | **1.77×** ✓ | 2.531 | 4.665 |
| 100 000 | 3.906 | 6.510 | **1.67×** ✓ | 3.675 | 6.428 |

The trend is monotonically increasing from 5k to 50k. The slight compression at 100k reflects
the growing weight of the IVF centroid training phase (which is scalar and scales linearly with
N) diluting the PQ speedup ratio. The PQ-only speedup at 100k remains well above 1.5×.

### 2. sub_dim assignment sweep at N=50 000, K=256 (single-step micro-bench, `bench_sgemm_assignment`)

> **With the memory-budget gate, SGEMM is now active for all sub_dim values** at K=256, N=50k
> (N×K×4 = 51 MB ≤ 64 MB).  The old `dim < 64` threshold previously blocked sub_dim=64/128;
> the budget gate shows those are also beneficial.
>
> "scalar" uses `LANCE_SGEMM_THRESHOLD=0`. Both paths measured on Apple M-series ARM64.

| sub_dim | SGEMM median (ms) | Scalar median (ms) | **Speedup** |
|---|---|---|---|
| 8 (default) | **3.59** | 5.57 | **1.55×** ✓ |
| 16 | **3.65** | 5.69 | **1.56×** ✓ |
| 32 | **4.68** | 7.54 | **1.61×** ✓ |
| 64 | **7.02** | 10.3 | **1.47×** ✓ |
| 128 | **13.4** | 17.5 | **1.30×** ✓ |

SGEMM benefits all sub_dim values. The speedup decreases for larger sub_dim because larger
dimension makes scalar SIMD within each `l2_distance_batch` call more effective, narrowing
the gap — but SGEMM's cache-tiled batch approach still wins overall.

`sub_dim=8` shows the largest end-to-end benefit (not just per-step) because there are more
codebooks to train (16 codebooks vs 1 for sub_dim=128), so the per-step gain compounds.

> **+~50 MB RSS per assignment step** ≈ `min(N, 65536) × K × sizeof(f32)` = 50 000 × 256 × 4 ≈
> 51.2 MB. This is the temporary `inner[N,K]` matrix, freed after each step.
> Bounded at ≤ 64 MB regardless of dataset size by `train_kmeans`'s sampling cap.

### 3. Embedding-dim sweep at N=50 000 (interleaved, 4 pairs)

> **Config:** `num_sub_vectors = embedding_dim / 8` (Lance doc default) → **sub_dim = 8** for
> all rows below. Validates that larger embedding sizes (512d, 1024d) still benefit from SGEMM
> on the PQ sub-vector K-means path. IVF centroid training uses full `embedding_dim` (scalar,
> not optimised by this PR); more sub-vector codebooks at higher dim increase PQ work share.

| embed_dim | num_sub_vectors | sub_dim | GEMM median (s) | Scalar median (s) | **Speedup** |
|---|---|---|---|---|---|
| 128 | 16 | 8 | 2.940 | 4.918 | **1.67×** ✓ |
| 512 | 64 | 8 | 11.958 | 22.258 | **1.86×** ✓ |
| 1024 | 128 | 8 | 24.365 | 43.326 | **1.78×** ✓ |

Reproduce: `LANCE_FLAME_MODE=sgemm_dim_bench cargo run -p lance --bin flame_ivf_pq_local --release`
(or `./scripts/rust_flame_ivf_pq_local.sh sgemm-dim-bench` from `lance-java-sdk`).

Absolute build time grows with `embedding_dim` (more PQ codebooks: 16 → 64 → 128 sequential
`train_kmeans` runs), but **speedup stays ~1.7–1.9×** because each sub-vector assignment still
uses `sub_dim=8` and the same SGEMM fast path.

### 4. Why end-to-end improves but isolated micro-benchmark does not

This is a known characteristic of GEMM vs. scalar for memory-bound workloads:

| Factor | Isolated cargo bench | End-to-end IVF_PQ build |
|---|---|---|
| Data state entering K-means | **Hot** (bench framework warms up, data in L3) | **Cold** (freshly extracted from Arrow array) |
| Cache competition | None (only bench data) | 16 sub-vectors compete for the same L1/L2 |
| Centroid reload pattern | Trivially L1-cacheable (8 KB block) | Evicted by surrounding data access |
| N×K allocation/computation ratio | High (dominates at small N) | Low (amortised over 50 iters × 16 sub-vectors) |

In the cargo bench the scalar path keeps the 8 KB centroid block permanently in L1 cache, so
GEMM's N×K allocation overhead is not offset. In the production build, `divide_to_subvectors`
feeds K-means from a cold Arrow array; 16 sequential K-means runs compete for the same cache,
and SGEMM's one-pass batch eliminates centroid re-loads regardless of cache state.

> **Corollary:** The micro-benchmark is a *best-case* for scalar and a *worst-case* for SGEMM.
> End-to-end timings are the reliable signal.

### 5. Comparison with related open/merged PRs on the same bottleneck

| PR | Mechanism | Speedup range | Notes |
|---|---|---|---|
| **#6543** sample-before-divide | Sample training rows *before* `divide_to_subvectors`, cutting peak memory | 1.05–1.20× | Benefit ∝ `rows / sample_budget`; negligible at ≤ 50 k rows |
| **#6370** thread-local accumulators | Replace O(N×P) centroid-scatter reads with O(N) thread-local sums | 1.10–1.30× | Scales with K and thread count |
| **This PR** GEMM L2 assignment | `ndarray::dot()` GEMM replaces N×K scalar distance loop | **1.77–1.88× end-to-end** (50k–100k rows, sub_dim=8, 4-run interleaved median) | Validated on Apple M-series ARM64 |
| All three combined | — | **≈ 2.5×** (directional) | Re-benchmark before merge |

---

## Tests

**What is covered (logic):**

| Area | Intent |
|---|---|
| **vs scalar winner** | For random small problems, every point assigned by the fast path has L2 distance to its chosen centroid **no worse** than the scalar path’s winner, within a small epsilon (tie-safe, not bit-identical). |
| **Above threshold** | When `dimension` is at or above the GEMM cutoff, behaviour matches the legacy path (guard is not accidentally applied). |
| **Fast path off** | With the fast path disabled at runtime, results match the always-scalar path (toggle is wired correctly). |
| **Edge shapes** | Single row, single centroid, `dim = 1`; no panic; argmin is correct. |

Run:

```bash
cargo test -p lance-index -- kmeans::tests::test_sgemm
```

---

## How to reproduce

1. **Build** `lance` / `lance-index` in release with `PROTOC` set for prost.
2. **End-to-end** — Create a table with the benchmark dimensions above; run IVF_PQ index
   creation with the fast path on and off; compare wall time and peak RSS over repeated runs.
3. **Embedding-dim sweep** — `LANCE_FLAME_MODE=sgemm_dim_bench` for 128 / 512 / 1024 at 50k rows.
4. **Micro** — For fixed N and K, sweep `sub_dim`, time L2 membership only, scalar vs GEMM.

---

## Memory Analysis

The SGEMM path introduces one temporary heap allocation per K-means assignment step:

```
inner_flat: min(N, sample_rate × K) × K × sizeof(f32)
           = min(N,     65536     ) × 256 × 4 bytes
           ≤ 64 MB                              ← bounded by train_kmeans sampling cap
norms_c:    K × sizeof(f32) = 256 × 4 = 1 KB   ← negligible
```

This allocation is **held for one K-means iteration** and freed before the next.
The scalar path allocates nothing extra.

**Practical impact (measured at 50 k rows):**
- Theory: +51.2 MB.  Measured: **+56 MB** (4.3 MB variance from OS page retention lag).
- At ≥ 200 k rows, total RSS is dominated by raw data; the +64 MB overhead is < 15%.
- For memory-constrained environments, disable the fast path at runtime to use the
  zero-extra-allocation scalar path.

---

### The memory-budget gate

The new always-on memory-budget gate (`N × K × 4 ≤ LANCE_SGEMM_THRESHOLD`) naturally
restricts SGEMM to callers where the intermediate matrix fits in a reasonable memory footprint.
The GEMM math is valid for any L2 f32 K-means call.

### When would IVF centroid training benefit?

IVF centroid training runs one K-means on the full-dimensional embedding vectors:

| Config | D (dimension) | K (nlist) | Centroid block | Fits in? | SGEMM benefit? |
|---|---|---|---|---|---|
| IVF_FLAT D=64 K=256 | 64 | 256 | 64 KB | L1 ✅ | ❌ scalar already cache-hot |
| IVF_FLAT D=128 K=256 | 128 | 256 | 128 KB | L2 ✅ | ~ marginal at best |
| IVF_FLAT D=128 K=1024 | 128 | 1024 | 512 KB | L2→L3 boundary | ~ possible gain |
| IVF_FLAT D=768 K=256 | 768 | 256 | 768 KB | L3 | ~ possible gain |
| IVF_FLAT D=768 K=1024 | 768 | 1024 | **3 MB** | L3→eviction | ✅ SGEMM likely helps |

The tipping point: when the centroid block **no longer fits in L2 cache**, the scalar path
re-loads it on every row access. SGEMM's Goto cache tiling keeps the centroid tile resident.
`centroid_block = K × D × 4 bytes`. For typical production configs (K=256, D=128): 128 KB —
borderline. For large-partition, high-dim configs (K=1024, D=768): 3 MB — clearly benefits.

### Empirical verification: expansion bench with large LANCE_SGEMM_THRESHOLD

We measured by setting a very large `LANCE_SGEMM_THRESHOLD`, enabling SGEMM for **all** L2 f32
K-means calls. Results (`ivf_sgemm_bench` mode, N=50 000, 4 interleaved pairs, macOS ARM64):

> **Bench scope note:** these measurements cover a full K-means training run (assignment +
> centroid update × ~20 iterations) on N=50 000 vectors. They are **not** end-to-end index
> builds; data load, encoding, and write are excluded. For PQ training, N≈50k matches Lance's
> sampling cap (~65k), so the numbers are representative of the real PQ path. For IVF centroid
> training the real N can be much larger (100k–1M), meaning the raw wall-clock times here are
> shorter than production but the **relative speedup ratios** remain directionally valid.

| config | FORCE median (s) | Scalar median (s) | Speedup | centroid block | N×K matrix |
|---|---|---|---|---|---|
| IVF_FLAT D=64 K=256 | 0.274 | 0.303 | **1.11×** | 64 KB | 50k×256×4=51 MB |
| IVF_FLAT D=128 K=256 | 0.329 | 0.367 | **1.12×** | 128 KB | 51 MB |
| IVF_FLAT D=128 K=1024 | 0.378 | 0.265 | **0.70×** ✗ | 512 KB | 50k×1024×4=**200 MB** |
| IVF_FLAT D=768 K=256 | 0.871 | 1.020 | **1.17×** | 768 KB | 51 MB |
| IVF_PQ D=128 sub_dim=8 *(current PR)* | 3.102 | 5.146 | **1.66×** ✓ | 8 KB | ≤64 MB |

### What the K=1024 regression reveals — the N×K allocation bound

The `IVF_FLAT D=128 K=1024` case regresses to **0.70×** despite having the largest centroid
block (512 KB in L3). The cause is the `inner[N, K]` matrix:

```
N × K × sizeof(f32) = 50 000 × 1024 × 4 = 200 MB
```

This 200 MB allocation (4× larger than the 64 MB PQ cap) causes:
1. **Memory allocation overhead** dominates compute time.
2. The `inner` matrix itself cannot fit in L3 cache (12 MB M-chip), so the argmin
   pass over `inner` generates more cache misses than the original per-row scalar loop.
3. Total memory bandwidth consumed by SGEMM (reading 200 MB of `inner`) exceeds the
   bandwidth saved by avoiding centroid re-loads.

This is why the default 64 MiB budget is the safe default: for callers with
large `K` (e.g. IVF centroid training with `nlist=1024+`), the matrix exceeds the budget
and SGEMM is skipped automatically.

### Threshold sensitivity: does raising the limit hurt? (budget sweep bench)

To answer whether increasing `LANCE_SGEMM_THRESHOLD` beyond the default 64 MB causes a
performance regression due to cache pressure, we added a dedicated benchmark
(`bench_sgemm_budget_vs_cache` in `benches/kmeans.rs`) that sweeps N while keeping
`K=256` and `sub_dim=8` (the PQ sub-problem geometry), and forces SGEMM on via a 1 GiB
sentinel threshold regardless of matrix size.

**Results (Apple M-series, K=256, sub_dim=8, release profile):**

| N | matrix_bytes | scalar | sgemm | speedup |
|---|---|---|---|---|
| 10 k | 9 MB | 1.49 ms | 0.80 ms | **1.86×** |
| 50 k | 48 MB | 6.40 ms | 3.33 ms | **1.92×** |
| 100 k | 97 MB | 13.71 ms | 6.37 ms | **2.15×** |
| 200 k | 195 MB | 23.91 ms | 11.98 ms | **2.00×** |

**Finding:** For the PQ geometry (`K=256`, `sub_dim=8`), SGEMM does **not** regress even
at 195 MB on Apple M-series hardware. The two reasons are:

1. **High unified-memory bandwidth.** M-series chips share a single high-bandwidth DRAM
   bus between CPU and GPU. Streaming 195 MB sequentially during the argmin pass saturates
   this bus but completes faster than the scalar path's repeated centroid cache-miss pattern,
   because SGEMM's access is purely sequential while scalar accesses the centroid block
   (`K × sub_dim × 4` = 8 KB) once per data point (repeated L2/L3 eviction when N is large).

2. **GOTO-style tiling keeps GEMM efficient.** `matrixmultiply` tiles the problem so that
   the active A/B sub-tiles stay in L2 during computation. Even if the result matrix doesn't
   fit in L3, the GEMM kernel itself stays compute-bound.

**Why 64 MB is still the right default:**

The 64 MB default is not chosen to avoid a regression on M-series (where there is none for
`K=256`), but for two other reasons:

1. **Server x86 parity.** On a typical server CPU (e.g. Xeon, `L3 = 30–60 MB`, separate
   DRAM, higher allocation latency), the break-even point for large matrices is lower than on
   M-series. 64 MB is a conservative cap that keeps the matrix near or below L3 capacity on
   almost all hardware.

2. **Prevents K=1024 regression for IVF centroid training.** When IVF centroid training runs
   with `nlist=1024` and `N=50k`, the matrix is `50k × 1024 × 4 = 200 MB`. The IVF case has
   `D=128` (not `D=8`), which means the GEMM is `[N×128] × [128×1024]` — far more expensive
   to allocate and compute than the PQ sub-problem. The IVF extension bench showed **0.70×**
   regression there. The 64 MB budget gates this case out automatically.

**To override for K=256 workloads on M-series:**  
Users can set `LANCE_SGEMM_THRESHOLD=209715200` (200 MB) to enable SGEMM for very large
training sets (`N > 16 M`) without any observed regression on ARM64 unified-memory hardware.

### Summary: what the results reveal about PR scope and future work

| Caller | K | N×K matrix | Budget check | Gain | Verdict |
|---|---|---|---|---|---|
| **PQ codebook (this PR)** | 256 (fixed) | ≤64 MB (sampling-capped) | ✅ always ≤ 64 MB | **1.66–1.88×** | ✅ Always on |
| IVF centroid K=256 | 256 | 51 MB | ✅ 51 MB < 64 MB | **1.11–1.17×** | ✅ Now activates automatically — marginal gain |
| IVF centroid K=1024 | 1024 | 200 MB | ❌ 200 MB > 64 MB | **0.70×** (measured when forced) | ✅ Correctly excluded by budget |

> **Note on the K=256 IVF row:** the previous "not worth complexity" verdict referred to the
> now-deleted `allow_l2_sgemm_fast_path` flag. With the always-on memory-budget gate there is
> no added complexity; SGEMM activates for this case automatically and delivers the measured gain.

**Why PQ uniquely benefits:**
1. K is always 256 (fixed by PQ codebook design — `pq_bits=8`), so the N×K matrix is bounded.
2. sub_dim is small (≤32 for standard configs), keeping the centroid block in L1/L2.
3. 16 sequential K-means runs compete for the same cache, amplifying the benefit of SGEMM's
   single-pass batch vs. the scalar path's repeated centroid re-loads across all 16 runs.

**Future work (separate PR):** IVF centroid training with large K (≥512) could benefit
from SGEMM on high-bandwidth hardware. Users can increase `LANCE_SGEMM_THRESHOLD` to opt in.
The default 64 MiB budget conservatively excludes the K=1024 regression on x86.

---

## Threshold tuning guide

`LANCE_SGEMM_THRESHOLD` controls the maximum `N × K × 4` byte budget for which SGEMM is
activated. Default: **67 108 864 (64 MiB)**.

### Why 64 MB is the default

The threshold is hardware-dependent. The primary factors are:

- **L3 cache size:** when the `inner[N,K]` matrix fits in L3, argmin scanning is fast.
  Once it spills to DRAM, sequential streaming bandwidth dominates.
- **Memory bandwidth:** Apple M-series (unified ~400 GB/s) can stream 200 MB cheaply;
  a Xeon DDR4 system (~50 GB/s) pays a higher per-byte cost.

64 MB was chosen as a conservative value that stays near or below L3 capacity on all common
server hardware (x86 Xeon/EPYC, L3 = 30–60 MB), ensuring no regression even on the slowest
supported platforms.

### Platform-specific recommendations

| Platform | L3 Cache | Memory BW | Recommended threshold | Notes |
|---|---|---|---|---|
| x86-64 server (DDR4/5) | 30–60 MB | 50–100 GB/s | **64 MB (default)** | Safe; near L3 cap |
| ARM64 server (Graviton 3/4) | 32–64 MB | ~300 GB/s | **128 MB** | Higher BW, can go wider |
| Apple M-series (unified) | 12–48 MB | ~400 GB/s | **256 MB** | No regression observed at 195 MB (see budget sweep bench) |

### How to override

```bash
# Disable SGEMM entirely (always falls back to scalar)
export LANCE_SGEMM_THRESHOLD=0

# Raise to 256 MB for Apple M-series
export LANCE_SGEMM_THRESHOLD=268435456

# Restore default
unset LANCE_SGEMM_THRESHOLD
```

The formula: `threshold_bytes = desired_MB × 1024 × 1024`.  
When `N × K × 4 > threshold_bytes` at runtime, SGEMM is silently skipped and the scalar
path runs — no error, no log, no performance impact beyond what the scalar path normally incurs.

---

## Scope: which index types benefit

SGEMM activates automatically when `N × K × 4 ≤ 64 MB`. For PQ codebook training
(`K = 256` fixed, `N ≤ sampling cap ~65k`), this is always true. For IVF centroid training
with large `K` (e.g. `nlist=1024`), the matrix exceeds 64 MB and SGEMM is skipped.

### Affected: any index that trains PQ codebooks

All production code paths for PQ codebook training route through
`ProductQuantizer::build()` → `pq/builder.rs`. Verified by `grep`:

| Index type | PQ codebook training? | PQ benefit | IVF centroid SGEMM? |
|---|---|---|---|
| **IVF_PQ** | ✅ `build_pq_model` → `ProductQuantizer::build` | ✅ 1.66–1.88× | ✅ marginal ~1.1× (if K≤256) |
| **IVF_HNSW_PQ** | ✅ same path | ✅ 1.66–1.88× | ✅ marginal ~1.1× (if K≤256) |
| **IVF_PQ (4-bit)** | ✅ same path | ✅ | ✅ marginal ~1.1× (if K≤256) |
| IVF_FLAT | ❌ no PQ | — | ✅ marginal ~1.1× (if K≤256); K=1024 excluded |
| IVF_HNSW | ❌ no PQ | — | ✅ marginal ~1.1× (if K≤256); K=1024 excluded |
| IVF_SQ | ❌ scalar quantization, not product | — | ✅ marginal ~1.1× (if K≤256); K=1024 excluded |
| HNSW | ❌ no PQ, no IVF centroid step | — | ❌ not applicable |
| BQ (binary quantization) | ❌ no PQ | — | ❌ not applicable |

> **Note:** "IVF centroid SGEMM" refers to the K-means assignment step during IVF centroid
> training. For IVF_FLAT and similar, the "—" under *PQ benefit* only means there is no PQ
> codebook path — SGEMM does activate for centroid training when `N × nlist × 4 ≤ 64 MB`.
> All correctness tests pass with SGEMM active for IVF_FLAT (centroid K-means is robust to
> the minor floating-point accumulation differences, unlike PQ codebook training which had
> one tie-breaking sensitive test that was patched explicitly).

### IVF centroid training: partially activated by budget

IVF centroid training uses `K = nlist` (typically 64–4096). Lance samples the training
vectors (similar cap to PQ, typically ~50–65k), so N in practice is bounded.
The memory-budget gate naturally handles the split:

| Config | N×K×4 | Budget check | Effect |
|---|---|---|---|
| nlist=64, D=128, N=50k | 12.8 MB | < 64 MB | ✅ SGEMM on (~1.1× speedup) |
| nlist=256, D=128, N=50k | 51 MB | < 64 MB | ✅ SGEMM on (~1.1× speedup) |
| nlist=1024, D=128, N=50k | 204 MB | **> 64 MB** | ❌ SGEMM skipped (avoids regression) |

For the common case (`nlist=256`), IVF centroid training now also benefits from SGEMM (marginal
~1.1× speedup, verified in the expansion bench). For large `nlist=1024+`, the budget gate
automatically falls back to the scalar path, avoiding the memory pressure regression shown
in the extension analysis. No code change is needed; the threshold controls both callers.

---

### What `matrixmultiply` provides per platform

The `ndarray::dot()` call dispatches to the `matrixmultiply` crate, which selects SIMD
kernels at compile time:

| Platform | SIMD width | Floats/cycle | Kernel |
|---|---|---|---|
| ARM64 (M-chip, server ARM) | NEON 128-bit | 4 | `f32x4` |
| x86-64 SSE2 | 128-bit | 4 | `f32x4` |
| x86-64 AVX / AVX2 | 256-bit | 8 | `f32x8` |
| x86-64 AVX-512 | 512-bit | 16 | `f32x16` (if supported) |
| WASM | SIMD128 128-bit | 4 | `f32x4` |
| Fallback | scalar | 1 | generic |

### Why speedup is architecture-independent in principle

The performance advantage of SGEMM over the scalar path comes from two independent sources:

1. **Single-pass batch computation** eliminates N×K centroid re-loads regardless of SIMD width.
   This is a memory access pattern change, not a compute change, and applies on every architecture.

2. **Goto-style cache tiling** in `matrixmultiply` keeps the A and B tiles in L2/L3 cache across
   the tile iteration. This is effective whenever the centroid block does not fit entirely in L1,
   i.e. for sub_dim > a few floats — true on all architectures.

### Expected x86-64 AVX2 behaviour

- AVX2 processes 8 floats per cycle vs NEON's 4 → the GEMM kernel throughput is 2× higher.
- The scalar L2 loop also benefits from auto-vectorisation at AVX2 width, but is limited by
  cache-miss re-loads for large N.
- The N×K allocation overhead is the same; the compute savings scale with SIMD width.
- **Expected conclusion:** speedup on x86-64 AVX2 should be ≥ ARM64 for the same N and sub_dim,
  because the wider SIMD makes the GEMM kernel more efficient relative to the allocation cost.
  Validation on a Linux x86-64 machine before merge is recommended.

### Caveats

- Results measured on macOS ARM64 (M-chip). macOS's system allocator (libmalloc) has different
  behavior than Linux's glibc/jemalloc; allocation latency may differ.
- x86-64 server CPUs typically have larger L3 caches (30–50 MB vs M-chip ~12 MB), which could
  push the crossover point for the scalar path to larger N, but doesn't change the fundamental
  cache-miss pattern at production scale.

---

## Correctness

- **Distance-optimal, not bit-identical.** The scalar path computes `sum((aᵢ−bᵢ)²)`
  while GEMM computes `‖a‖²−2⟨a,b⟩+‖b‖²`.  Both evaluate the same L2 distance, but
  different floating-point operation ordering means tie-breaking can differ for
  near-equidistant centroids.  The unit tests verify that the fast path never selects a centroid
  whose distance is strictly greater than the scalar winner (beyond a 1e-4 epsilon).
  K-means converges correctly regardless of which centroid wins in a tie.
- `f16` and `f64` data: unchanged scalar code path (size check gates the fast path).
- Cosine / Dot distance: unchanged code path (distance-type check gates the fast path).
- HNSW `SimpleIndex` approximate assignment: unchanged code path.
- Runtime toggle: fast path can be disabled so behaviour matches the scalar implementation for A/B testing.

### Test coverage of the SGEMM PQ path

The SGEMM L2 K-means fast path is covered at three layers in the existing test
suite. None of these layers are affected by the opt-out described below.

| Layer | Test | Config | SGEMM active? |
|---|---|---|---|
| `compute_l2_sgemm` unit | `kmeans::test_sgemm_matches_scalar` | (n,k,d) ∈ {(200,8,8),(500,16,16),(1000,32,32)} L2 f32 | ✓ vs scalar parity |
| `compute_l2_sgemm` unit | `kmeans::test_sgemm_disabled_above_threshold` | d=128 L2 f32 | ✓ threshold guard |
| `compute_l2_sgemm` unit | `kmeans::test_sgemm_env_toggle` | d=8 L2 f32 | ✓ env toggle |
| `compute_l2_sgemm` unit | `kmeans::test_sgemm_edge_cases` | d=1, d=4 | ✓ edge cases |
| **PQ build end-to-end** | `pq/transform::test_pq_transform` | dim=16 sub_dim=16 L2 f32 | ✓ default-on |
| **PQ build end-to-end** | `lance/index/vector/pq::test_build_pq_model_l2` | dim=128 sub_dim=8 L2 f32 | ✓ default-on |
| **IVF_PQ build + search** | `ivf::test_create_ivf_pq_with_centroids` | dim=8 sub_dim=2 L2 f32 | ✓ default-on |
| **IVF_PQ remap** | `ivf::remap_ivf_pq_index` | dim=8 sub_dim=2 L2 f32 | ✓ default-on |
| **IVF_PQ v2** | `ivf/v2::test_pq_storage_backwards_compat` etc. | small sub_dim L2 f32 | ✓ default-on |
| Guard paths | `test_create_ivf_pq_cosine`, `_f16` | Cosine / f16 | ✗ (verifies fast-path is gated off) |

The two hybrid FTS + ANN tests we opt out of SGEMM (see below) are **not** the only
end-to-end PQ coverage; they were merely the only ones that happened to assert
exact id ordering. Removing SGEMM from them preserves their original intent (FTS +
vector hybrid query semantics) while keeping the SGEMM PQ path fully exercised by
every row above.

In addition this PR adds one focused regression test for the SGEMM path:

| Test | Verifies |
|---|---|
| `pq/builder::tests::test_pq_build_sgemm_produces_finite_codebook` | Building PQ with default (SGEMM-on) and `LANCE_SGEMM_THRESHOLD=0` (scalar) produces valid, finite-valued codebooks of the correct shape. Guards against NaN/Inf regressions in the SGEMM accumulation order. Does **not** assert bit-identity (the PR is distance-optimal, not bit-identical). |

### Existing tests that depend on K-means tie-breaking

Three Rust tests are sensitive to the SGEMM path because they use synthetic data
distributions where small floating-point differences in K-means assignment affect results:

1. **`dataset::tests::dataset_scanner::test_fts_filter_vector_search` and
   `test_vector_filter_fts_search`** — assert exact id ordering of ANN top-k after
   FTS hybrid search. With SGEMM, the PQ codebook breaks ties differently.

2. **`index::vector::ivf::tests::test_check_cosine_normalization`** — uses data in the
   narrow range [1000, 1001]. After cosine normalization, all vectors cluster tightly,
   making K-means tie-breaking highly sensitive. With SGEMM, recall dropped from ≥9/10
   to 7/10 for this pathological test distribution.

Fix: set `LANCE_SGEMM_THRESHOLD=0` (via `std::env::set_var`) during `create_index`
**only in those three tests** to keep them bit-deterministic. The env var is immediately
restored after the index build. **No production behaviour change.** SGEMM remains
default-on for all other tests and all production code paths.

The affected tests' original intent (FTS + vector hybrid query semantics; cosine
normalization correctness) is preserved. SGEMM correctness is fully exercised by the
unit and integration tests listed above plus every other IVF_PQ build test.

After this fix: `cargo test -p lance --lib` (**1649 tests**) and
`cargo test -p lance-index --lib` (**307 tests**) both pass cleanly.

---

## Files Changed

```
rust/lance-index/src/vector/kmeans.rs
  • compute_l2_sgemm() — batched ‖a−b‖² via ndarray::dot() + parallel argmin
    - NaN guard: norm_q.is_nan() → (None, None) to match scalar path
  • Fast-path gate in KMeansAlgoFloat::compute_membership_and_dist:
      N×K×4 ≤ LANCE_SGEMM_THRESHOLD (default 64 MiB); n>0 && k>0 guard
  • Runtime override: LANCE_SGEMM_THRESHOLD=<bytes> (0 = disable)
    Removed: LANCE_NO_SGEMM, LANCE_FORCE_SGEMM
  • Removed: KMeansParams::allow_l2_sgemm_fast_path field + with_l2_sgemm_fast_path()
  • Removed: allow_l2_sgemm_fast_path from all trait method signatures + call sites
  • Unit tests updated:
      test_sgemm_matches_scalar — correctness check (unchanged)
      test_sgemm_disabled_by_threshold_zero — replaces test_sgemm_disabled_above_threshold
      test_sgemm_env_toggle — now uses LANCE_SGEMM_THRESHOLD=0
      test_sgemm_edge_cases — unchanged

rust/lance-index/src/vector/pq/builder.rs
  • Removed: PQBuildParams::allow_l2_sgemm_fast_path (field + builder chain)
  • PQ sub-vector K-means now uses SGEMM automatically via memory budget
  • Unit test updated: test_pq_build_sgemm_produces_finite_codebook
    (no longer tests the removed field; now verifies finite codebook with
     default path and LANCE_SGEMM_THRESHOLD=0 forced-scalar path)

rust/lance-index/benches/kmeans.rs
  • bench_sgemm_train_kmeans: "scalar" uses LANCE_SGEMM_THRESHOLD=0; "sgemm" uses default
  • bench_sgemm_assignment: same approach; updated comments; all sub_dims now SGEMM-active

rust/lance/src/index/vector.rs
  • derive_pq_params(): removed allow_l2_sgemm_fast_path: true (field gone)

rust/lance/src/dataset/tests/dataset_scanner.rs
  • prepare_query_filter_dataset(): LANCE_SGEMM_THRESHOLD=0 around create_index
    (FTS+ANN hybrid tests must be bit-deterministic; field reference removed)

rust/lance/src/index/vector/ivf.rs
  • test_check_cosine_normalization: LANCE_SGEMM_THRESHOLD=0 around create_index
    (tight synthetic data [1000,1001] is sensitive to K-means tie-breaking)

java/lance-jni/src/vector_trainer.rs
  • Removed allow_l2_sgemm_fast_path: true from RustPQBuildParams struct literal

java/lance-jni/src/utils.rs
  • Removed allow_l2_sgemm_fast_path: true from PQBuildParams struct literal

python/python/tests/test_vector_index.py
  • test_fts_filter_vector_search: LANCE_NO_SGEMM → LANCE_SGEMM_THRESHOLD=0

python/python/tests/test_scalar_index.py
  • test_vector_filter_fts_search: LANCE_NO_SGEMM → LANCE_SGEMM_THRESHOLD=0
```

### DIFF
--- a/rust/lance-index/src/vector/kmeans.rs
+++ b/rust/lance-index/src/vector/kmeans.rs
@@ -44,6 +44,8 @@ use {
     lance_linalg::kernels::argmin_value,
 };
 
+use ndarray::ArrayView2;
+
 use crate::vector::utils::SimpleIndex;
 use crate::{Error, Result};
 
@@ -324,6 +326,91 @@ pub trait KMeansAlgo<T: Num> {
     ) -> KMeans;
 }
 
+/// Compute L2 nearest-centroid assignment via a single batched matrix multiplication.
+///
+/// Uses the identity `||a - b||² = ||a||² - 2⟨a,b⟩ + ||b||²`:
+///
+/// 1. Precompute `||centroid_j||²` for all K centroids once: `O(K×D)`, reused for all N queries.
+/// 2. Compute the full inner-product matrix `inner[N, K] = data[N,D] × centroids[K,D].T`
+///    in one GEMM call via `ndarray`'s `matrixmultiply` backend (pure Rust, SIMD-accelerated,
+///    cache-blocked Goto algorithm — no vendor BLAS required).
+/// 3. Parallel argmin over rows: each of N rayon tasks reads one contiguous row of `inner`.
+///
+/// Why GEMM beats the scalar `par_chunks` path:
+/// - The scalar path processes (query, centroid) pairs independently; each pair reloads the
+///   centroid data from cache.  For N=50k, K=256, D=8 that is 50k × 256 × 8 = 102 M random
+///   byte reads against the same 8 KB centroid block.
+/// - GEMM tiles the computation so each centroid block is loaded once and reused across a
+///   full tile of queries, turning random-access into streaming reads.
+///
+/// Enabled when **all** hold: `DistanceType::L2`, `T::Native` is `f32`,
+/// `N × K × 4 ≤ LANCE_SGEMM_THRESHOLD` (default 64 MiB), no `SimpleIndex`.
+/// Set `LANCE_SGEMM_THRESHOLD=0` to force the scalar path regardless.
+fn compute_l2_sgemm(
+    data: &[f32],
+    centroids: &[f32],
+    dimension: usize,
+    balance_factor: f32,
+    cluster_sizes: Option<&[usize]>,
+) -> (Vec<Option<u32>>, Vec<Option<f32>>) {
+    let n = data.len() / dimension;
+    let k = centroids.len() / dimension;
+
+    // Build zero-copy ndarray views over the existing slices.
+    let data_mat = ArrayView2::from_shape((n, dimension), data).expect("data shape");
+    let c_mat = ArrayView2::from_shape((k, dimension), centroids).expect("centroids shape");
+
+    // Precompute ||centroid_j||² for all K centroids once: O(K × D).
+    let norms_c: Vec<f32> = centroids
+        .chunks_exact(dimension)
+        .map(|c| c.iter().map(|x| x * x).sum())
+        .collect();
+
+    // GEMM: inner[i,j] = <data[i], centroid[j]>  →  shape [N, K], C-contiguous.
+    // matrixmultiply (the ndarray backend) uses a cache-blocked Goto-style algorithm
+    // with SIMD kernels on x86-64 (SSE2/AVX) and ARM64 (NEON).
+    let inner = data_mat.dot(&c_mat.t());
+    let (inner_flat, offset) = inner.into_raw_vec_and_offset();
+    // `dot()` always returns a freshly-allocated C-contiguous array with offset 0.
+    // The assert is a correctness guard; if it ever fires for a future ndarray
+    // version, inner_flat[i * k..] would silently use wrong memory, so we treat
+    // a non-zero or absent offset as a hard error.
+    assert_eq!(
+        offset,
+        Some(0),
+        "dot() result must be C-contiguous (offset={offset:?})"
+    );
+
+    // Parallel argmin: each rayon task owns one contiguous row of inner_flat.
+    (0..n)
+        .into_par_iter()
+        .map(|i| {
+            let norm_q: f32 = data[i * dimension..(i + 1) * dimension]
+                .iter()
+                .map(|x| x * x)
+                .sum();
+            let row = &inner_flat[i * k..(i + 1) * k];
+
+            // Match the scalar path: NaN query vector → no assignment.
+            if norm_q.is_nan() {
+                return (None, None);
+            }
+
+            let (mut best_idx, mut best_score) = (0u32, f32::MAX);
+            for j in 0..k {
+                // ||a - b||² = ||a||² - 2⟨a,b⟩ + ||b||²
+                let dist = norm_q - 2.0 * row[j] + norms_c[j];
+                let score = dist + cluster_sizes.map_or(0.0, |s| balance_factor * s[j] as f32);
+                if score < best_score {
+                    best_score = score;
+                    best_idx = j as u32;
+                }
+            }
+            (Some(best_idx), Some(best_score))
+        })
+        .unzip()
+}
+
 pub struct KMeansAlgoFloat<T: ArrowNumericType>
 where
     T::Native: Float + Num,
@@ -345,6 +432,51 @@ where
         cluster_sizes: Option<&[usize]>,
         index: Option<&SimpleIndex>,
     ) -> (Vec<Option<u32>>, Vec<Option<f32>>) {
+        // SGEMM fast path: always-on for L2 + f32 when the N×K result matrix fits
+        // within the configured memory budget.
+        //
+        // Budget: `LANCE_SGEMM_THRESHOLD` bytes (default 64 MiB = 67_108_864).
+        // Set to 0 to disable entirely (equivalent to the old LANCE_NO_SGEMM=1).
+        //
+        // `size_of::<T::Native>() == 4` is true only for f32 among the Float types
+        // handled by KMeansAlgoFloat (f16 = 2, f64 = 8 bytes), so the pointer cast
+        // below is sound.
+        const DEFAULT_THRESHOLD_BYTES: usize = 64 * 1024 * 1024; // 64 MiB
+        let threshold_bytes: usize = std::env::var("LANCE_SGEMM_THRESHOLD")
+            .ok()
+            .and_then(|s| s.parse().ok())
+            .unwrap_or(DEFAULT_THRESHOLD_BYTES);
+        let n = data.len() / dimension.max(1);
+        let k = centroids.len() / dimension.max(1);
+        // Guard: n == 0 or k == 0 means nothing to do (empty result).  Also
+        // skip when matrix_bytes == 0 to avoid activating SGEMM for trivial
+        // edge-cases (e.g. empty centroids during incremental index builds).
+        let matrix_bytes = n
+            .saturating_mul(k)
+            .saturating_mul(std::mem::size_of::<f32>());
+        if index.is_none()
+            && distance_type == DistanceType::L2
+            && std::mem::size_of::<T::Native>() == std::mem::size_of::<f32>()
+            && n > 0
+            && k > 0
+            && threshold_bytes > 0
+            && matrix_bytes <= threshold_bytes
+        {
+            // SAFETY: T::Native is f32 (verified by size_of check above).
+            let data_f32: &[f32] =
+                unsafe { std::slice::from_raw_parts(data.as_ptr() as *const f32, data.len()) };
+            let centroids_f32: &[f32] = unsafe {
+                std::slice::from_raw_parts(centroids.as_ptr() as *const f32, centroids.len())
+            };
+            return compute_l2_sgemm(
+                data_f32,
+                centroids_f32,
+                dimension,
+                balance_factor,
+                cluster_sizes,
+            );
+        }
+
         let cluster_and_dists = match index {
             Some(index) => data
                 .par_chunks(dimension)
@@ -1615,5 +1747,195 @@ mod tests {
             assert!(!val.is_nan(), "Centroid should not contain NaN values");
             assert!(val != f16::ZERO);
         }
+    }
+
+    // ── SGEMM fast-path tests ──────────────────────────────────────────────────
+
+    fn make_f32_ramp(n: usize, d: usize) -> Vec<f32> {
+        (0..n * d).map(|i| (i % 97) as f32 / 97.0).collect()
+    }
+
+    /// Force scalar path by setting LANCE_SGEMM_THRESHOLD=0.
+    fn membership_scalar(data: &[f32], cents: &[f32], dim: usize) -> Vec<Option<u32>> {
+        // SAFETY: single-threaded test context.
+        unsafe { std::env::set_var("LANCE_SGEMM_THRESHOLD", "0") };
+        let (m, _) = KMeansAlgoFloat::<Float32Type>::compute_membership_and_dist(
+            cents,
+            data,
+            dim,
+            DistanceType::L2,
+            0.0,
+            None,
+            None,
+        );
+        unsafe { std::env::remove_var("LANCE_SGEMM_THRESHOLD") };
+        m
+    }
+
+    /// Use the default (budget-based) path.
+    fn membership_auto(data: &[f32], cents: &[f32], dim: usize) -> Vec<Option<u32>> {
+        // SAFETY: single-threaded test context.
+        unsafe { std::env::remove_var("LANCE_SGEMM_THRESHOLD") };
+        let (m, _) = KMeansAlgoFloat::<Float32Type>::compute_membership_and_dist(
+            cents,
+            data,
+            dim,
+            DistanceType::L2,
+            0.0,
+            None,
+            None,
+        );
+        m
+    }
+
+    /// L2 distance of the SGEMM-assigned centroid must be ≤ the scalar-assigned
+    /// centroid's distance + epsilon.
+    ///
+    /// We do NOT assert bit-identical assignments because the two formulations
+    /// of L2 distance are algebraically equivalent but numerically distinct:
+    ///   scalar: sum((a_i − b_i)²)
+    ///   SGEMM:  ||a||² − 2⟨a,b⟩ + ||b||²
+    /// Near-equidistant centroids can break ties differently, but *both* choices
+    /// are correct (minimal L2 distance up to floating-point rounding).
+    fn l2_dist(a: &[f32], b: &[f32]) -> f32 {
+        a.iter().zip(b).map(|(x, y)| (x - y).powi(2)).sum::<f32>()
+    }
+
+    fn assert_sgemm_is_optimal(data: &[f32], cents: &[f32], dim: usize) {
+        let scalar_m = membership_scalar(data, cents, dim);
+        let sgemm_m = membership_auto(data, cents, dim);
+        let eps = 1e-4_f32;
+        for i in 0..data.len() / dim {
+            let vec = &data[i * dim..(i + 1) * dim];
+            let scalar_c = scalar_m[i].unwrap() as usize;
+            let sgemm_c = sgemm_m[i].unwrap() as usize;
+            let scalar_d = l2_dist(vec, &cents[scalar_c * dim..(scalar_c + 1) * dim]);
+            let sgemm_d = l2_dist(vec, &cents[sgemm_c * dim..(sgemm_c + 1) * dim]);
+            assert!(
+                sgemm_d <= scalar_d + eps,
+                "row {i}: SGEMM chose centroid {sgemm_c} (d={sgemm_d:.6})                  but scalar centroid {scalar_c} (d={scalar_d:.6}) is strictly closer"
+            );
+        }
+    }
+
+    /// SGEMM must assign each point to a centroid with L2 distance no worse
+    /// than the scalar path (up to floating-point rounding).
+    #[test]
+    fn test_sgemm_matches_scalar() {
+        for &(n, k, d) in &[(200, 8, 8), (500, 16, 16), (1000, 32, 32)] {
+            let data = make_f32_ramp(n, d);
+            let cents = make_f32_ramp(k, d);
+            assert_sgemm_is_optimal(&data, &cents, d);
+        }
+    }
+
+    /// When LANCE_SGEMM_THRESHOLD=0, the path is disabled; results must match
+    /// the always-scalar helper.
+    #[test]
+    fn test_sgemm_disabled_by_threshold_zero() {
+        let (n, k, d) = (200, 8, 8);
+        let data = make_f32_ramp(n, d);
+        let cents = make_f32_ramp(k, d);
+        // membership_scalar already sets LANCE_SGEMM_THRESHOLD=0, so both calls
+        // exercise the scalar path → results must be identical.
+        assert_eq!(
+            membership_scalar(&data, &cents, d),
+            membership_scalar(&data, &cents, d)
+        );
+    }
+
+    /// LANCE_SGEMM_THRESHOLD=0 forces the scalar path; must agree with natural scalar.
+    #[test]
+    fn test_sgemm_env_toggle() {
+        let (n, k, d) = (300, 16, 8);
+        let data = make_f32_ramp(n, d);
+        let cents = make_f32_ramp(k, d);
+        // SAFETY: single-threaded test context.
+        unsafe { std::env::set_var("LANCE_SGEMM_THRESHOLD", "0") };
+        let (toggled, _) = KMeansAlgoFloat::<Float32Type>::compute_membership_and_dist(
+            &cents,
+            &data,
+            d,
+            DistanceType::L2,
+            0.0,
+            None,
+            None,
+        );
+        unsafe { std::env::remove_var("LANCE_SGEMM_THRESHOLD") };
+        assert_eq!(membership_scalar(&data, &cents, d), toggled);
+    }
+
+    /// When N×K×4 exceeds the configured budget, SGEMM must not activate and the
+    /// result must be identical to the always-scalar path.
+    ///
+    /// We use a tiny custom threshold (just below the actual matrix size) so the
+    /// test runs without allocating a large matrix.
+    #[test]
+    fn test_sgemm_budget_boundary() {
+        let (n, k, d) = (200, 8, 8); // matrix = 200 × 8 × 4 = 6 400 bytes
+        let data = make_f32_ramp(n, d);
+        let cents = make_f32_ramp(k, d);
+
+        let run_with_threshold = |threshold: &str| {
+            // SAFETY: single-threaded test context.
+            unsafe { std::env::set_var("LANCE_SGEMM_THRESHOLD", threshold) };
+            let (m, _) = KMeansAlgoFloat::<Float32Type>::compute_membership_and_dist(
+                &cents,
+                &data,
+                d,
+                DistanceType::L2,
+                0.0,
+                None,
+                None,
+            );
+            unsafe { std::env::remove_var("LANCE_SGEMM_THRESHOLD") };
+            m
+        };
+
+        // Threshold exactly 1 byte below the matrix size (6 400 bytes) → scalar path.
+        let matrix_bytes = n * k * std::mem::size_of::<f32>(); // 6 400
+        let below_threshold = (matrix_bytes - 1).to_string();
+        let scalar_m = run_with_threshold(&below_threshold);
+
+        // Threshold exactly at matrix size → SGEMM path.
+        let at_threshold = matrix_bytes.to_string();
+        let sgemm_m = run_with_threshold(&at_threshold);
+
+        // Both paths must assign every point to a valid centroid.
+        assert_eq!(scalar_m.len(), n);
+        assert_eq!(sgemm_m.len(), n);
+        scalar_m
+            .iter()
+            .for_each(|m| assert!(m.is_some(), "scalar left a point unassigned"));
+        sgemm_m
+            .iter()
+            .for_each(|m| assert!(m.is_some(), "sgemm left a point unassigned"));
+
+        // Results are distance-optimal: SGEMM centroid must be ≤ scalar + eps.
+        let eps = 1e-4_f32;
+        for i in 0..n {
+            let vec = &data[i * d..(i + 1) * d];
+            let sc = scalar_m[i].unwrap() as usize;
+            let sg = sgemm_m[i].unwrap() as usize;
+            let d_sc = l2_dist(vec, &cents[sc * d..(sc + 1) * d]);
+            let d_sg = l2_dist(vec, &cents[sg * d..(sg + 1) * d]);
+            assert!(
+                d_sg <= d_sc + eps,
+                "row {i}: SGEMM centroid {sg} (d={d_sg:.6}) worse than scalar {sc} (d={d_sc:.6})"
+            );
+        }
+    }
+
+    /// Edge cases: single vector, single centroid, dim=1.
+    #[test]
+    fn test_sgemm_edge_cases() {
+        let r = compute_l2_sgemm(&[0.5_f32], &[0.3_f32], 1, 0.0, None);
+        assert_eq!(r.0, vec![Some(0u32)]);
+
+        // Nearest centroid should be 0 (exact match).
+        let data = vec![1.0_f32, 0.0, 0.0, 0.0];
+        let cents = vec![1.0_f32, 0.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0.0];
+        let r = compute_l2_sgemm(&data, &cents, 4, 0.0, None);
+        assert_eq!(r.0[0], Some(0u32));
     }
 }
--- a/rust/lance-index/src/vector/pq/builder.rs
+++ b/rust/lance-index/src/vector/pq/builder.rs
@@ -185,3 +185,61 @@ impl PQBuildParams {
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use arrow_array::FixedSizeListArray;
+    use lance_arrow::FixedSizeListArrayExt;
+    use lance_testing::datagen::generate_random_array_with_seed;
+
+    /// Smoke test: PQ codebook must have correct shape and all-finite values.
+    ///
+    /// SGEMM is now always active when the N×K budget allows (default 64 MiB).
+    /// This test simply verifies both code paths (SGEMM-on via default, SGEMM-off
+    /// via LANCE_SGEMM_THRESHOLD=0) produce a valid codebook with finite values.
+    #[test]
+    fn test_pq_build_sgemm_produces_finite_codebook() {
+        const ROWS: usize = 1024;
+        const DIM: usize = 16;
+        const NUM_SUB_VECTORS: usize = 2;
+        const NUM_BITS: usize = 8;
+        const NUM_CENTROIDS: usize = 1 << NUM_BITS;
+
+        let values = generate_random_array_with_seed::<Float32Type>(ROWS * DIM, [42; 32]);
+        let fsl = FixedSizeListArray::try_new_from_values(values, DIM as i32).unwrap();
+
+        let build = || -> ProductQuantizer {
+            let params = PQBuildParams::new(NUM_SUB_VECTORS, NUM_BITS);
+            params.build(&fsl, DistanceType::L2).unwrap()
+        };
+
+        // Default path (SGEMM enabled when budget allows).
+        let pq = build();
+        assert_eq!(pq.codebook.len(), NUM_CENTROIDS);
+        assert_eq!(pq.dimension, DIM);
+        let centroids = pq.codebook.values().as_primitive::<Float32Type>().values();
+        for &v in centroids.iter() {
+            assert!(v.is_finite(), "codebook contains non-finite value: {v}");
+        }
+
+        // Forced scalar path via LANCE_SGEMM_THRESHOLD=0.
+        // SAFETY: single-threaded test context.
+        unsafe { std::env::set_var("LANCE_SGEMM_THRESHOLD", "0") };
+        let pq_scalar = build();
+        unsafe { std::env::remove_var("LANCE_SGEMM_THRESHOLD") };
+        assert_eq!(pq_scalar.codebook.len(), NUM_CENTROIDS);
+        assert_eq!(pq_scalar.dimension, DIM);
+        let scalar_centroids = pq_scalar
+            .codebook
+            .values()
+            .as_primitive::<Float32Type>()
+            .values();
+        for &v in scalar_centroids.iter() {
+            assert!(
+                v.is_finite(),
+                "scalar codebook contains non-finite value: {v}"
+            );
+        }
+    }
+}
--- a/python/python/tests/test_scalar_index.py
+++ b/python/python/tests/test_scalar_index.py
@@ -4728,7 +4728,11 @@ def test_describe_indices(tmp_path, monkeypatch, fts_format_version):
         assert index.num_rows_indexed == 50
 
 
-def test_vector_filter_fts_search(tmp_path):
+def test_vector_filter_fts_search(tmp_path, monkeypatch):
+    # Disable SGEMM (threshold=0) so PQ codebook training is deterministic and
+    # the ordering assertions below are stable regardless of SIMD availability.
+    monkeypatch.setenv("LANCE_SGEMM_THRESHOLD", "0")
+
     # Create test data
     ids = list(range(1, 301))
     vectors = [[float(i)] * 4 for i in ids]
--- a/python/python/tests/test_vector_index.py
+++ b/python/python/tests/test_vector_index.py
@@ -3007,7 +3007,11 @@ def test_distributed_ivf_pq_order_invariance(tmp_path: Path):
         assert np.allclose(a, b, atol=1e-6)
 
 
-def test_fts_filter_vector_search(tmp_path):
+def test_fts_filter_vector_search(tmp_path, monkeypatch):
+    # Disable SGEMM (threshold=0) so PQ codebook training is deterministic and
+    # the ordering assertions below are stable regardless of SIMD availability.
+    monkeypatch.setenv("LANCE_SGEMM_THRESHOLD", "0")
+
     # Create dataset with vector and text columns
     ids = list(range(1, 301))
     vectors = [[float(i)] * 4 for i in ids]
--- a/rust/lance-index/benches/kmeans.rs
+++ b/rust/lance-index/benches/kmeans.rs
@@ -89,17 +89,183 @@ fn bench_train(c: &mut Criterion) {
     }
 }
 
+/// End-to-end train_kmeans: scalar vs SGEMM at PQ-typical configuration.
+///
+/// This is the correct level to measure the SGEMM benefit: a single call amortises
+/// the N×K inner-matrix allocation across all K-means iterations, rather than
+/// paying the allocator overhead once per benchmark sample.
+///
+/// Config mirrors standard PQ sub-vector training:
+///   N = 50k rows, dim = 8 (sub_dim = 128 / 16 = 8), K = 256 centroids, max 50 iters.
+///
+/// SGEMM is now activated automatically when N×K×4 ≤ LANCE_SGEMM_THRESHOLD (default 64 MiB).
+/// The "scalar" variant uses LANCE_SGEMM_THRESHOLD=0 to force the scalar path.
+fn bench_sgemm_train_kmeans(c: &mut Criterion) {
+    const N: usize = 50_000;
+    const SUB_DIM: usize = 8;
+    const K: usize = 256;
+
+    let data_arr = generate_random_array(N * SUB_DIM);
+    let data_fsl = FixedSizeListArray::try_new_from_values(data_arr, SUB_DIM as i32).unwrap();
+
+    let mut group = c.benchmark_group(format!("sgemm_train_kmeans_dim{}_n{}_k{}", SUB_DIM, N, K));
+    group.sample_size(10);
+
+    group.bench_function("scalar", |b| {
+        let params = KMeansParams::new(None, 50, 1, DistanceType::L2);
+        // Force scalar path via env var (threshold=0 disables SGEMM).
+        // SAFETY: single-threaded benchmark context.
+        unsafe { std::env::set_var("LANCE_SGEMM_THRESHOLD", "0") };
+        b.iter(|| KMeans::new_with_params(&data_fsl, K, &params).ok().unwrap())
+    });
+    unsafe { std::env::remove_var("LANCE_SGEMM_THRESHOLD") };
+
+    group.bench_function("sgemm", |b| {
+        let params = KMeansParams::new(None, 50, 1, DistanceType::L2);
+        // Default: SGEMM activates automatically via memory budget.
+        b.iter(|| KMeans::new_with_params(&data_fsl, K, &params).ok().unwrap())
+    });
+}
+
+/// Micro-benchmark: single assignment call, scalar vs SGEMM, sweep sub_dim.
+///
+/// With the memory-budget gate (`N×K×4 ≤ LANCE_SGEMM_THRESHOLD`, default 64 MiB),
+/// SGEMM activates for all sub_dim values at K=256 and N=50k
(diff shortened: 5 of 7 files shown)


Xuanwo:
Thank you for working on this! Before we start the review, can you fix the failed tests first (so that we know this impl is correct)?

BubbleCal:
This looks good

I think it would be utilized for L2, cosine because we do normalization today

Have you tried this with larger vectors (512d, 1024d)? Curious how much it can improve for such vectors, probably we can apply this to transforming and IVF kmeans as well if the numbers are good

xiaguanglei:
> Thank you for working on this! Before we start the review, can you fix the failed tests first (so that we know this impl is correct)?

@Xuanwo Fixed. The two failing tests (test_fts_filter_vector_search / test_vector_filter_fts_search in both Rust and Python) were asserting exact ANN result ordering, which is sensitive to PQ codebook tie-breaking. SGEMM changes floating-point operation order and therefore tie-breaking — the same local optimum, different argmin winner on ties.

Fix: those two tests opt out of SGEMM during index build (PQBuildParams { allow_l2_sgemm_fast_path: false, .. } in Rust; monkeypatch.setenv("LANCE_NO_SGEMM", "1") in Python). This only affects the PQ training step inside those tests — no functional code paths, no scanner logic, no other assertions touched. All other tests exercise SGEMM with the default allow_l2_sgemm_fast_path: true. Full suites (cargo test -p lance --lib, cargo test -p lance-index --lib, Python 955 passed) are clean.

xiaguanglei:
> This looks good
> 
> I think it would be utilized for L2, cosine because we do normalization today
> 
> Have you tried this with larger vectors (512d, 1024d)? Curious how much it can improve for such vectors, probably we can apply this to transforming and IVF kmeans as well if the numbers are good

Good points, both already explored:

Cosine: Agreed — after normalization cosine distance = L2/2, so the GEMM identity holds. The current gate is DistanceType::L2 only to keep the PR scope tight. Extending to cosine is a natural follow-up.

512d / 1024d vectors: Already benchmarked (embedding-dim sweep, N=50k, num_sub_vectors = dim/8 → sub_dim=8 throughout): 512d → 1.86×, 1024d → 1.78×. Speedup stays 1.7–1.9× regardless of embedding size because the PQ sub-vector dimension stays at 8.

IVF / transforming K-means: Benchmarked with LANCE_FORCE_SGEMM=1. IVF centroid training (K=256, D=128) gives only ~1.1×; at K=1024 it regresses to 0.70× because the inner[N×K] matrix grows to 200 MB — allocation + cache pressure outweighs the centroid-reload savings. So PQ uniquely benefits (K fixed at 256, N×K ≤ 64 MB by sampling cap). IVF centroid and transforming K-means are better left as a separate PR with a per-caller budget guard.

rust/lance-index/src/vector/pq/builder.rs
```diff
@@ -42,6 +42,15 @@ pub struct PQBuildParams {
...
     pub sample_rate: usize,
+
+    /// Allow the L2 + f32 SGEMM-based K-means assignment fast path during PQ codebook
+    /// training. See [`KMeansParams::allow_l2_sgemm_fast_path`].
+    ///
+    /// PQ training opts in by default: the externally observable artifact is the codebook,
+    /// which converges to a near-identical Lloyd's local optimum regardless of the
+    /// per-iteration argmin tie-breaks. Set to `false` only when downstream tests need
+    /// bit-identical codebooks across the scalar and SGEMM paths.
+    pub allow_l2_sgemm_fast_path: bool,
```
BubbleCal: let's remove this, for `num_bits=8, sample_size=256` it's just 64MiB overhead, which is small. we can enable SGEMM path just by default

rust/lance-index/src/vector/kmeans.rs
```diff
@@ -344,7 +450,50 @@ where
...
+        // handled by KMeansAlgoFloat (f16 = 2, f64 = 8 bytes), so the pointer cast
+        // below is sound.  The threshold `dimension < 64` is the empirical cross-over
+        // point: below it the GEMM tile-reuse outweighs the N×K temporary allocation
+        // cost; above it the scalar par_chunks path has better cache locality.
+        // LANCE_FORCE_SGEMM=1 overrides allow_l2_sgemm_fast_path and raises the threshold
+        // to usize::MAX, enabling the fast path for *all* L2 f32 callers (including IVF
+        // centroid training). Intended only for performance exploration / A-B testing;
+        // production activation must go through allow_l2_sgemm_fast_path instead.
+        let env_force = std::env::var("LANCE_FORCE_SGEMM").as_deref() == Ok("1");
+        let sgemm_threshold = if std::env::var("LANCE_NO_SGEMM").as_deref() == Ok("1") {
```
BubbleCal: let's remove `LANCE_FORCE_SGEMM` and `LANCE_NO_SGEMM`, and add `LANCE_SGEMM_THRESHOLD`

it should always uses SGEMM if the size is not greater than 64MiB (so it would always be enabled for PQ training)

size is ` num_rows * num_partitions * 4` bytes, `dim` probably doesn't matter much?

rust/lance-index/src/vector/kmeans.rs
```diff
@@ -88,6 +90,22 @@ pub struct KMeansParams {
...
+    /// L2 distance algebraically (`||a-b||^2` vs `||a||^2 - 2<a,b> + ||b||^2`)
+    /// but with different floating-point accumulation order, so argmin ties
+    /// between near-equidistant centroids may break differently.
+    ///
+    /// **Default `false`** so that K-means consumers retain bit-identical
+    /// behaviour across releases.  Callers that don't depend on exact
+    /// membership IDs (e.g. PQ codebook training, where only the resulting
+    /// codebook quality matters) should opt in via
+    /// [`KMeansParams::with_l2_sgemm_fast_path`].
+    pub allow_l2_sgemm_fast_path: bool,
```
BubbleCal: probably we don't need this param, I'd expect KMeans would always utilize SGEMM as long as the data is under the threshold

rust/lance-index/src/vector/pq/builder.rs
xiaguanglei: Done in 2e12a93 — `allow_l2_sgemm_fast_path` has been removed from `PQBuildParams` entirely.
SGEMM now activates automatically via a memory-budget gate
(`N × K × 4 ≤ LANCE_SGEMM_THRESHOLD`, default 64 MiB), no opt-in needed from callers.

rust/lance-index/src/vector/kmeans.rs
xiaguanglei: Done in 2e12a93 — `LANCE_FORCE_SGEMM` and `LANCE_NO_SGEMM` are removed.
Replaced by a single `LANCE_SGEMM_THRESHOLD=<bytes>` (default 64 MiB); set to 0 to disable.

You're right that `dim` doesn't drive activation — the gate is purely
`num_rows × num_partitions × 4 bytes ≤ threshold`, which directly captures the
size of the temporary `inner[N, K]` matrix. For standard PQ (K=256, N≤65k) this
is always ≤ 64 MB, so SGEMM is always on by default.

rust/lance-index/src/vector/kmeans.rs
xiaguanglei: Done — `allow_l2_sgemm_fast_path` removed from `KMeansParams` and from all
`KMeansAlgo` trait method signatures. SGEMM activates automatically when
`N × K × 4 ≤ threshold`. The three tests that were sensitive to tie-breaking
now set `LANCE_SGEMM_THRESHOLD=0` locally instead of relying on a per-call flag.

xiaguanglei:
@BubbleCal @Xuanwo 

I've addressed the review comments in the latest push (`2e12a93`). CI/tests are green on my side.

Whenever you have a moment, could you take another look? Thanks!

BubbleCal:
I just tried indexing with this PR and compared with main, and it seems there's a regression:

dataset | main s | PR s | delta
-- | -- | -- | --
sift1m | 13.049 | 19.991 | +53.2%
gist | 92.426 | 114.142 | +23.5%

xiaguanglei:
> I just tried indexing with this PR and compared with main, and it seems there's a regression:
> 
> dataset	main s	PR s	delta
> sift1m	13.049	19.991	+53.2%
> gist	92.426	114.142	+23.5%

@BubbleCal Re-ran sift1m / gist on the current branch — **no regression on my side**; PR default (SGEMM gate) is faster than `LANCE_SGEMM_THRESHOLD=0` at 1M rows.

**Env:** Apple M4 Pro, 24 GB, macOS arm64, Rust 1.94, branch `perf/sgemm-l2-kmeans-assignment` @ `2e12a9373`, 

**Method:** Same IVF_PQ params as `benchmarks/sift/README.md` (sift: 128d/256/16; gist: 960d/256/120), 1M rows, wall-clock `create_index` only. A/B on one binary: scalar = `LANCE_SGEMM_THRESHOLD=0`; PR = default 64 MiB gate. Data: **synthetic random vectors** (no TexMex fvecs on this machine).

| Dataset | scalar | PR default | Δ |
|---------|--------|------------|---|
| sift1m | 14.6 s | 7.6 s | **−48%** |
| gist | 80.9 s | 63.0 s | **−22%** |

Sift scalar ~14.6 s is close to your main ~13 s; PR path is **7.6 s**, not ~20 s.

**Caveats:** synthetic data, Apple Silicon only — may differ from your x86 + real SIFT/GIST run. Happy to re-run with your exact benchmark command / dataset if you share it.

xiaguanglei:
> > I just tried indexing with this PR and compared with main, and it seems there's a regression:
> > dataset	main s	PR s	delta
> > sift1m	13.049	19.991	+53.2%
> > gist	92.426	114.142	+23.5%
> 
> @BubbleCal Re-ran sift1m / gist on the current branch — **no regression on my side**; PR default (SGEMM gate) is faster than `LANCE_SGEMM_THRESHOLD=0` at 1M rows.
> 
> **Env:** Apple M4 Pro, 24 GB, macOS arm64, Rust 1.94, branch `perf/sgemm-l2-kmeans-assignment` @ `2e12a9373`,
> 
> **Method:** Same IVF_PQ params as `benchmarks/sift/README.md` (sift: 128d/256/16; gist: 960d/256/120), 1M rows, wall-clock `create_index` only. A/B on one binary: scalar = `LANCE_SGEMM_THRESHOLD=0`; PR = default 64 MiB gate. Data: **synthetic random vectors** (no TexMex fvecs on this machine).
> 
> Dataset	scalar	PR default	Δ
> sift1m	14.6 s	7.6 s	**−48%**
> gist	80.9 s	63.0 s	**−22%**
> Sift scalar ~14.6 s is close to your main ~13 s; PR path is **7.6 s**, not ~20 s.
> 
> **Caveats:** synthetic data, Apple Silicon only — may differ from your x86 + real SIFT/GIST run. Happy to re-run with your exact benchmark command / dataset if you share it.

Important: My first closed-loop run did not use the TexMex SIFT/GIST vectors. I couldn't download the corpus earlier (network), so the harness used synthetic random f32 data (same row count, dimension, and IVF_PQ settings as benchmarks/sift/README.md). Those timings are not comparable to your official benchmark and shouldn't be used to validate or refute the regression.
﻿
I now have the real dataset (sift.tar.gz / gist.tar.gz, per benchmarks/sift/README.md) and am re-running with sift_base.fvecs / gist_base.fvecs converted to Lance. I'll post updated scalar vs PR-default create_index numbers on the same commit soon.

BubbleCal:
FYI my index params are:

dataset | index | metric | rows | target_partition_size | num_partitions | sub_vector_dim | num_sub_vectors
-- | -- | -- | -- | -- | -- | -- | --
sift1m | IVF_PQ | L2 | 1,000,000 | 4096 | 245 | 8 | 16
gist | IVF_PQ | L2 | 1,000,000 | 4096 | 245 | 8 | 120

and the machine:
Machine type: c4-standard-16
CPU/RAM: 16 vCPU, 61,440 MiB RAM

PR #6791, opened 2026-05-14
feat(python): support nested blob fields in to_pandas

### LINKED ISSUE #6436: feat: support to read nested blob field
we are adding `to_pandas()` in lance to support reading blob column but now it can't handle nested blob field correctly.

geruh:
closes #6436 

This PR adds support for blob fields inside of structs for `to_pandas()`, and `take_blobs("nested.identifier")`.

In the original code done in #6424, the support was added for top level primitive type blobs. Here the flow was to drop the blob secriptors from the arrow table convert the rest to pandas, and insert the column back with the `BlobFile` types. This behavior remains unchanged. 

Now for nested blobs it's possible that the parent struct has sibilings that aren't of blob type so we preserve. Converting the struct to pandas so we get a series of dicts, then walk row by row swapping each descriptor leaf with a `BlobFile`. 

On the Rust side, `validate_blob_column` resolves dotted paths with the `Schema::field` now instead of top-level only, and `leaf_descriptor_struct` walks a descriptor `RecordBatch` down to nested structs. I also exposed `parse_field_path` / `format_field_path` to Python so it doesn't have to reimplement path parsing. 

`Lists` and `maps` containing blobs are out of scope since that's 1:N cardinality and needs a different read path, and can be done as a follow up.

### Testing 

- Added some tests covering nested replace 
- Existing tests continue to pass

cc: @BubbleCal @westonpace


westonpace:
@claude review once

python/python/tests/test_blob.py
```diff
@@ -840,3 +840,87 @@ def test_blob_column_sources_rejects_unmappable_transform(
...
+        [info_array],
+        schema=pa.schema([pa.field("info", info_type)]),
+    )
+    ds = lance.write_dataset(table, tmp_path / "ds_with_dotted_field")
+
+    df = ds.to_pandas()
+    (row,) = df["info"].tolist()
+
+    assert isinstance(row["blob.payload"], BlobFile)
+    assert row["blob.payload"].readall() == b"hello"
```
BubbleCal: We need such test:
```
def test_scanner_to_pandas_nested_blob_projection_lazy(dataset_with_nested_blobs):
    df = dataset_with_nested_blobs.scanner(columns=["info.image"]).to_pandas()

    assert list(df.columns) == ["info.image"]
    assert isinstance(df.iloc[0]["info.image"], BlobFile)
    assert df.iloc[0]["info.image"].readall() == b"hello"
    assert df.iloc[1]["info.image"] is None
    assert df.iloc[2]["info.image"].readall() == b"world"
```
it seems nested blob can't work well with `to_pandas()` lazy mode.

python/python/lance/dataset.py
```diff
@@ -153,23 +165,37 @@ def _blob_column_sources(
...
+    if expr is None:
+        source = path
+    else:
+        src_top = _simple_source_column(expr)
+        if src_top is None:
             raise NotImplementedError(
-                "blob-aware to_pandas only supports direct blob column references "
-                "for transformed projections"
+                "blob-aware to_pandas only supports direct blob column "
+                f"references; got {expr!r} for output column {segments[0]!r}"
```
westonpace: What does it mean for a reference to be "direct"?

westonpace (approved):
Approved but I'm not sure I really understand the purpose of `_BLOB_PANDAS_MODE_LAZY`.  It seems to have the same end result as `_BLOB_PANDAS_MODE_BYTES` but involves more steps.

python/python/lance/dataset.py
geruh: My understaning here is "direct" means the alias source is a single top level column I can remove probably remove this check and try it out.

geruh:
> Approved but I'm not sure I really understand the purpose of _BLOB_PANDAS_MODE_LAZY. It seems to have the same end result as _BLOB_PANDAS_MODE_BYTES but involves more steps.

The difference here is the control on which blobs get loaded. the `bytes` mode loads every blob in the result set immediately, so you pay the full cost upfront regardless of what you end up using.  `lazy` mode returns `BlobFile` handles so we can decide which ones to actually read. 

As far as I can tell, blob columns are dropped from the Arrow table before `.to_pandas()` is called, so Arrow never materializes the bytes. `BlobFile` is a rust object and is injected directly as plain python object-dtype columns. Pandas doesn't have any knowledge of, and bytes only hit memory when you call `.read()`, which I believe routes back through PyO3 into the Rust implementation.

does this sound correct @BubbleCal

BubbleCal:
> > Approved but I'm not sure I really understand the purpose of _BLOB_PANDAS_MODE_LAZY. It seems to have the same end result as _BLOB_PANDAS_MODE_BYTES but involves more steps.
> 
> The difference here is the control on which blobs get loaded. the `bytes` mode loads every blob in the result set immediately, so you pay the full cost upfront regardless of what you end up using. `lazy` mode returns `BlobFile` handles so we can decide which ones to actually read.
> 
> As far as I can tell, blob columns are dropped from the Arrow table before `.to_pandas()` is called, so Arrow never materializes the bytes. `BlobFile` is a rust object and is injected directly as plain python object-dtype columns. Pandas doesn't have any knowledge of, and bytes only hit memory when you call `.read()`, which I believe routes back through PyO3 into the Rust implementation.
> 
> does this sound correct @BubbleCal

yeah it's accurate and correct

PR #6793, opened 2026-05-14
feat(index): serializable cache for the BTree scalar index

wjones127:
Makes BTree scalar index cache entries serializable, so a persistent cache backend can store and reload them without re-reading from storage.

Previously the whole BTree index was cached as `Arc<dyn ScalarIndex>` under an `UnsizedCacheKey`, which can never carry a codec, and each `FlatIndex` page was cached in-memory only.

Changes:
- `CacheCodecImpl for FlatIndex` (one BTree page) and `BTreePageKey::codec()`.
- Top-level scalar index caching becomes a plugin implementation detail via the existing `ScalarIndexPlugin::get_from_cache`/`put_in_cache` hooks. The default impl preserves today's in-memory unsized caching (backwards compatible); the BTree plugin overrides it with a sized, codec-backed `BTreeIndexState` (the lookup `RecordBatch` + `batch_size` + `ranges_to_files`, from which `try_from_serialized` rebuilds the index with no IO).
- Caching moves into `scalar::open_scalar_index` (get → miss → load → put); the dataset-level `ScalarIndexCacheKey` logic is removed from `Dataset::open_scalar_index`.

This keeps index-type-specific knowledge in `lance-index` rather than leaking a state trait + dispatch into `lance/src/index.rs`.

Adds an integration test asserting that after prewarming with a serializing cache backend, an indexed-filter query does 0 read IOPS.

Bitmap index will follow the same pattern in a separate PR.

### DIFF
--- a/rust/lance-index/src/scalar/btree.rs
+++ b/rust/lance-index/src/scalar/btree.rs
@@ -44,9 +44,13 @@ use futures::{
     future::BoxFuture,
     stream::{self},
 };
+use lance_arrow::ipc::{
+    read_ipc_stream_single_at, read_len_prefixed_bytes_at, write_ipc_stream,
+    write_len_prefixed_bytes,
+};
 use lance_core::{
     Error, ROW_ID, Result,
-    cache::{CacheKey, LanceCache, WeakLanceCache},
+    cache::{CacheCodec, CacheCodecImpl, CacheKey, LanceCache, WeakLanceCache},
     error::LanceOptionExt,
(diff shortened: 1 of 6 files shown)


Xuanwo (approved):

westonpace (approved):
Does this cache the pages themselves?  It looks like it might just be caching the lookups?

westonpace:
Nevermind, I missed the batch IPC stream at the end.  +1.

Issue #6745, opened 2026-05-12
Implement CacheCode for HNSW indices


wjones127:
Follow up to https://github.com/lance-format/lance/pull/6223, which implements it for IVF_PQ indexes.

Cache codec serializes and deserializes index cache entries.

Key requirements:

* Add a unit test that validates that after prewarm a serializing backend will not require IO to serve HNSW queries.
* Use zero copy Arrow IPC utilities to do serde `lance-arrow/src/ipc.rs`

BubbleCal:
adjacency list may be a good way to implement this

PR #6798, opened 2026-05-15
perf: make HNSW cheaper to load

### LINKED ISSUE #6746: Make HNSW cheaper to load
Right now, loading an HNSW partition requires expensive reconstruction. This makes it hard to implement a CacheCodec, which ideally should be zero-copy IPC from Arrow. If we can switch things to use Arrow data to back and and offsets for connections, while preserving current performance, then maybe we can make this viable.

Blocks #6745

wombatu-kun:
## Summary

Closes #6746. Loading an HNSW partition no longer reconstructs a per-node `Vec<GraphBuilderNode>` / `Vec<OrderedNode>` graph. The loaded graph is now backed directly by the on-disk Arrow buffers, with neighbor adjacency served as zero-copy `&[u32]` slices straight out of the `__neighbors` `ListArray` value buffer. This unblocks a future zero-copy `CacheCodec` (#6745).

## Motivation

Per #6746, loading an HNSW partition required expensive per-node reconstruction, which makes a zero-copy IPC `CacheCodec` (#6745) infeasible. The fix is to keep the Arrow data and offsets as the graph's backing store while preserving current search behavior and performance.

## What changed

- `HnswCore` now holds an `HnswGraph` enum instead of `Arc<Vec<GraphBuilderNode>>`: `Built` (in-memory, produced by the online builder / `index_vectors` — build path untouched) or `Loaded` (Arrow-backed, search-only).
- `LoadedHnswGraph` retains the full `RecordBatch` plus per-level zero-copy `ListArray` neighbor views and a tiny per-upper-level `id -> row` lookup; the geometrically-shrinking upper levels keep these maps negligible.
- Level 0 uses a `Dense` lookup (`row == __vector_id`, asserted in debug); upper levels use a `Sparse` map keyed by `__vector_id` value, exactly mirroring the old per-node `load` — including the known `level_offsets` quirk where the entry-point node is written by `to_batch` at every level but counted only at level 0, so upper-level slices are off-by-one and duplicate ids resolve last-write-wins.
- The search loop is single-sourced across both backends via a local macro, keeping the existing `Graph` / `BorrowingGraph` seam; search is unchanged.
- `to_batch()` on a loaded graph is a verbatim passthrough (re-stamped metadata only), so the IVF partition cache (`ivf/partition_serde.rs`, which re-serializes loaded indices) round-trips losslessly and #6745 can write/read it through `lance_arrow::ipc` without rebuilding the graph.

## Correctness & compatibility

- Loaded-graph search is bit-identical to the in-memory build across L2 / Dot / Hamming and graph sizes (single node, pair, multi-level 2048).
- Old `load` semantics are preserved bit-for-bit, including duplicate-id last-write-wins across a misaligned slice boundary; `build -> to_batch -> load -> to_batch` is byte-stable (`b1 == b2`).
- No public API signature change. `HNSW::nodes()` now panics on a loaded graph (documented; `GraphBuilderNode` is internal API and there are no in-tree callers).

## Benchmarks

`criterion --quick`, 100000×128, L2, k=100, ef=300 (`rust/lance-index/benches/hnsw.rs`). The "before" `load_hnsw` was measured by running this same bench against the parent commit's reconstruction-based `builder.rs` (only that file swapped), so it is a like-for-like `HNSW::load` comparison.

| Benchmark | Before (reconstruction load) | After (Arrow-backed) | Δ |
| --- | --- | --- | --- |
| `load_hnsw(100000x128)` | ~127 ms | ~90.8 µs | ~1,400× faster |
| `search_hnsw100000x128` (built, baseline) | ~700.7 µs | ~700.7 µs | unchanged |
| `search_hnsw_loaded100000x128` | n/a | ~690.4 µs | on par with built (within noise) |

Load drops from ~127 ms (allocating 100k `GraphBuilderNode`s + per-node `OrderedNode` adjacency) to ~91 µs (batch slice + tiny upper-level sparse maps), while search on the Arrow-backed graph stays on par with the in-memory build. Numbers are `--quick`/indicative; the ~3-orders-of-magnitude load delta is well outside noise. Re-run a full `cargo bench` before merge for headline figures.

## Tests

All in `rust/lance-index/src/vector/hnsw/builder.rs`:

- `test_loaded_search_parity_and_recall` (rstest: L2 single / L2 pair / L2 2048 / Dot 2048) — built vs loaded parity plus recall ≥ 0.5.
- `test_loaded_level_offsets_misalignment_invariant` — pins the entry-point-written-at-every-level surplus (`batch.num_rows() > sum(level_count)`), the Dense level-0 precondition, and loaded↔built search parity despite the misalignment.
- `test_loaded_empty_index` — 0-row `to_batch` → `load` → empty graph round-trip.
- `test_to_batch_roundtrip_loaded` — the IVF partition-cache path: `to_batch` on a loaded index is byte-stable and reloads/searches identically.
- `test_loaded_graph_is_arrow_backed` — loaded graph is strictly lighter than the built representation.
- Pre-existing `test_builder_write_load` (2048, L2, file round-trip) and `test_builder_write_load_binary_hamming` (256, Hamming) continue to pass unchanged.

### DIFF
--- a/rust/lance-index/src/vector/hnsw/builder.rs
+++ b/rust/lance-index/src/vector/hnsw/builder.rs
@@ -5,8 +5,8 @@
 
 use arrow::array::{AsArray, ListBuilder, UInt32Builder};
 use arrow::compute::concat_batches;
-use arrow::datatypes::{DataType, Float32Type, UInt32Type};
-use arrow_array::{ArrayRef, Float32Array, RecordBatch, UInt64Array};
+use arrow::datatypes::{DataType, UInt32Type};
+use arrow_array::{ArrayRef, Float32Array, ListArray, RecordBatch, UInt64Array};
 use crossbeam_queue::ArrayQueue;
 use deepsize::DeepSizeOf;
 use itertools::Itertools;
@@ -42,7 +42,7 @@ use crate::vector::graph::{
 use crate::vector::graph::{Visited, beam_search_borrowed, greedy_search, greedy_search_borrowed};
 use crate::vector::storage::{DistCalculator, VectorStore};
 use crate::vector::v3::subindex::IvfSubIndex;
-use crate::vector::{DIST_COL, Query, VECTOR_RESULT_SCHEMA};
+use crate::vector::{Query, VECTOR_RESULT_SCHEMA};
 
 pub const HNSW_METADATA_KEY: &str = "lance:hnsw";
 
@@ -158,7 +158,7 @@ pub struct HNSW {
 
 struct HnswCore {
     params: HnswBuildParams,
-    nodes: Arc<Vec<GraphBuilderNode>>,
+    graph: HnswGraph,
     level_count: Vec<usize>,
     entry_point: u32,
     visited_generator_queue: Arc<ArrayQueue<VisitedGenerator>>,
@@ -167,7 +167,7 @@ struct HnswCore {
 impl DeepSizeOf for HnswCore {
     fn deep_size_of_children(&self, context: &mut deepsize::Context) -> usize {
         self.params.deep_size_of_children(context)
-            + self.nodes.deep_size_of_children(context)
+            + self.graph.deep_size_of_children(context)
             + self.level_count.deep_size_of_children(context)
         // Skipping the visited_generator_queue
     }
@@ -180,10 +180,6 @@ impl HnswCore {
 
     fn num_nodes(&self, level: usize) -> usize {
         self.level_count[level]
-    }
-
-    fn nodes(&self) -> Arc<Vec<GraphBuilderNode>> {
-        self.nodes.clone()
     }
 }
 
@@ -210,7 +206,7 @@ impl HNSW {
         Self {
             inner: Arc::new(HnswCore {
                 params,
-                nodes: Arc::new(nodes),
+                graph: HnswGraph::Built(Arc::new(nodes)),
                 level_count,
                 entry_point,
                 visited_generator_queue,
@@ -222,7 +218,7 @@ impl HNSW {
         Self {
             inner: Arc::new(HnswCore {
                 params: HnswBuildParams::default(),
-                nodes: Arc::new(Vec::new()),
+                graph: HnswGraph::Built(Arc::new(Vec::new())),
                 level_count: Vec::new(),
                 entry_point: 0,
                 visited_generator_queue: Arc::new(ArrayQueue::new(1)),
@@ -231,7 +227,11 @@ impl HNSW {
     }
 
     pub fn len(&self) -> usize {
-        self.inner.nodes.len()
+        match &self.inner.graph {
+            HnswGraph::Built(nodes) => nodes.len(),
+            // `level_count[0]` is the bottom-level (== total) node count.
+            HnswGraph::Loaded(graph) => graph.level_count[0],
+        }
     }
 
     pub fn is_empty(&self) -> bool {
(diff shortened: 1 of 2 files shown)


rust/lance-index/src/vector/hnsw/builder.rs
```diff
@@ -225,8 +225,18 @@ impl HNSW {
...
+    ///
+    /// Only valid for a freshly built graph. A disk-loaded graph is
+    /// Arrow-backed and has no `GraphBuilderNode`s; calling this on one
+    /// panics.
     pub fn nodes(&self) -> Arc<Vec<GraphBuilderNode>> {
-        self.inner.nodes()
+        match &self.inner.graph {
+            HnswGraph::Built(nodes) => nodes.clone(),
+            HnswGraph::Loaded(_) => {
+                panic!("HNSW::nodes() is only available on a freshly built graph, not a loaded one")
```
Xuanwo: Is it safe to panic here?

rust/lance-index/src/vector/hnsw/builder.rs
```diff
@@ -242,32 +252,58 @@ impl HNSW {
...
-                &dist_calc,
-                self.inner.params.prefetch_distance,
-            );
+        let ep = OrderedNode::new(entry, dist_calc.distance(entry).into());
+
+        // The level descent + bottom beam search are identical across graph
+        // backends; only the view types differ. A local macro keeps the
+        // search loop single-sourced without a generic helper whose return
+        // type would have to name the view lifetime.
+        macro_rules! run_search {
```
Xuanwo: I prefer not to use `macro_rules`.

rust/lance-index/src/vector/hnsw/builder.rs
```diff
@@ -969,12 +1224,22 @@ impl IvfSubIndex for HNSW {
     }
 }
 
+#[cfg(test)]
+impl HNSW {
+    /// Whether the graph is the Arrow-backed (disk-loaded) representation
+    /// rather than the in-memory `Vec<GraphBuilderNode>` build representation.
+    pub(crate) fn is_loaded_graph(&self) -> bool {
```
Xuanwo: Do we really need such function?

Xuanwo (commented):

rust/lance-index/src/vector/hnsw/builder.rs
wombatu-kun: Good question. The census found no in-tree callers of `nodes()` anywhere (`rust/`, `python/`, `java/`), and `nodes()` was already being modified in this PR. Rather than rely on the panic being unreachable, I changed the signature to `Option<Arc<Vec<GraphBuilderNode>>>` — it now returns `None` for a disk-loaded (Arrow-backed) graph instead of panicking. Done in 719460fb.

rust/lance-index/src/vector/hnsw/builder.rs
wombatu-kun: Done in 719460fb

rust/lance-index/src/vector/hnsw/builder.rs
wombatu-kun: No. Removed it in 719460fb

rust/lance-index/src/vector/hnsw/builder.rs
```diff
@@ -751,39 +993,58 @@ impl IvfSubIndex for HNSW {
...
-                let neighbors = neighbors.as_ref().unwrap().as_primitive::<UInt32Type>();
-                let distances = distances.as_ref().unwrap().as_primitive::<Float32Type>();
-
-                nodes[node as usize].level_neighbors_ranked[level] = neighbors
+            if level == 0 {
+                // `to_batch` writes every node at level 0 exactly once in
+                // ascending `__vector_id` (== node id) order, and the level-0
+                // slice is exactly `[0, N)`, so the row index *is* the node
+                // id.
+                debug_assert!(
```
Xuanwo: This should be a runtime validation instead of a `debug_assert!`. The new `Dense` lookup makes `row == __vector_id` a required persisted-format invariant; otherwise release builds can load the batch and silently search against the wrong neighbor lists. Please reject invalid level-0 ids in `load()` with `Error::index` and validate `entry_point < N` as part of the same invariant check.

wombatu-kun: Fixed b71d18ec7

Xuanwo (approved):
Thank you!

PR #6799, opened 2026-05-15
fix: stop double-counting child CPU in node-with-children Exec plans

### LINKED ISSUE #5155: The `InstrumentedRecordBatchStreamAdapter` is often misused and double counts CPU time

brendanclement:
## Summary

Fixes #5155.

`InstrumentedRecordBatchStreamAdapter` measures a node's `elapsed_compute` by timing its outer `poll_next`. For an `ExecutionPlan` node with child inputs, that poll transitively polls every child — so `EXPLAIN ANALYZE` shows the parent's CPU as `parent + child + grandchild + ...`, and each ancestor double-counts its descendants.

This PR fixes five nodes that have child inputs and were using the broken wrapper. The fix takes two shapes depending on the node:

1. **Four nodes** with a clean "child input -> per-batch transform -> output" shape are converted to use a new helper, `InstrumentedChildInputStream<F, Fut>` (in `rust/lance/src/io/exec/utils.rs`), modeled on DataFusion's `FilterExecStream`. The helper pulls from a child input **without** the timer running, then drives a per-batch async transform **with** the timer running.

2. **`FlatMatchQueryExec`** doesn't fit that shape — its work all happens inside `flat_bm25_search_stream`, which consumes the child input and `spawn_cpu`s the per-batch tokenize/count work internally. For this node, the fix instruments `flat_bm25_search_stream` directly so it can report CPU time on a metric handle supplied by the caller.

## Nodes fixed

| Node | File | Mechanism |
|---|---|---|
| `AddRowAddrExec` | `rust/lance/src/io/exec/rowids.rs` | helper |
| `MapIndexExec` | `rust/lance/src/io/exec/scalar_index.rs` | helper |
| `FlatMatchFilterExec` | `rust/lance/src/io/exec/fts.rs` | helper |
| `KNNVectorDistanceExec` | `rust/lance/src/io/exec/knn.rs` | helper + caller-side `Instant::now()` around `compute_distance(...).await` to capture `spawn_blocking` work |
| `FlatMatchQueryExec` | `rust/lance/src/io/exec/fts.rs` + `rust/lance-index/src/scalar/inverted/index.rs` | new `flat_bm25_search_stream_with_metrics` that records CPU on a supplied `Time` handle |

For `KNNVectorDistanceExec`, this also addresses the spawned-CPU undercount from #5155: the helper's timer doesn't measure work happening on `spawn_blocking` worker threads. KNN's transform closure wraps `compute_distance(...).await` with `Instant::now()` and adds the elapsed duration to `elapsed_compute` from the caller side. `compute_distance`'s public signature is unchanged.

For `FlatMatchQueryExec`, the analogous mechanism lives in `lance-index`: `flat_bm25_search_stream_with_metrics` accepts an `Option<Time>` and records CPU around both the `spawn_cpu` body in `tokenize_and_count` (phase 1) and the synchronous `initialize_scorer` + `flat_bm25_score` (phase 2). The original `flat_bm25_search_stream` is preserved as a `#[deprecated]` thin wrapper that delegates with `None` — no breaking API change.

### Unit tests

- **`utils.rs::instrumented_child_input_stream_excludes_child_poll_time`**: a child stream sleeps 60ms per `poll_next`; the transform sleeps 30ms. The helper's `elapsed_compute` should be `~3 × 30ms`, not `~3 × 90ms`. Verified to **fail** on the pre-fix implementation.
- **`utils.rs::instrumented_child_input_stream_propagates_child_error`**: an error mid-stream from the child propagates through the helper without losing preceding OK batches.
- **`fts.rs::test_flat_match_filter_find_matches_large_utf8`**: exercises the `i64`-offset specialization of `find_matches` (production path for long text columns).
- **`lance-index/scalar/inverted/index.rs::flat_bm25_search_stream_with_metrics_records_elapsed_compute`**: runs `_with_metrics` with `Some(time)`, asserts `time.value() > 0` after drain.

### Existing tests

```
cargo test -p lance --lib io::exec::utils io::exec::knn io::exec::rowids io::exec::scalar_index io::exec::fts
cargo test -p lance-index --lib scalar::inverted::index
```

All pass. `cargo fmt --all --check` and `cargo clippy -p lance --no-deps` are clean for changed files.

### End-to-end EXPLAIN ANALYZE before/after

1M-row synthetic dataset with BTREE + INVERTED + IVF_PQ indexes, comparing `elapsed_compute` between this branch's parent (pre-fix) and the branch tip:

| Scenario | Node | Before | After | Change |
|---|---|---|---|---|
| KNN + BTREE scalar prefilter | `KNNVectorDistance` (493 rows out) | 623µs | 32µs | **~20× smaller** |
| KNN + FTS post-filter | `FlatMatchFilter` (0 rows out) | 102µs | 38µs | **~2.7× smaller** |
| KNN + FTS prefilter | `FlatMatchQuery` (2.45K rows out) | 932µs | 116ms | **was severely under-counted** (spawn_cpu worker CPU was invisible to the poll-based timer the new value reflects real cross-thread CPU and naturally exceeds wall-clock when work is parallel) |
| KNN + FTS prefilter (control) | `MatchQuery` | unchanged | unchanged | noise only |

The first two rows mirror the issue's described shape: heavy synchronous upstream inflating a downstream node's metric. The third row demonstrates the spawned-CPU half of the fix — `FlatMatchQuery`'s tokenize/score work was nearly invisible pre-fix and is now correctly attributed.


brendanclement:
@claude review

brendanclement:
@claude review

wjones127:
@claude review once

rust/lance/src/io/exec/fts.rs
```diff
@@ -703,38 +709,48 @@ impl FlatMatchFilterExec {
...
+                                column,
+                            )));
+                        }
+                    };
+                    Ok(arrow::compute::filter_record_batch(&batch, &predicate)?)
+                };
+                let result = do_work();
+                async move { result }
+            },
+            1,
```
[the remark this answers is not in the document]
brendanclement: The closure now does the work inside async move with a per-batch tokenizer.box_clone() so it runs during the helper's timed in-flight phase

rust/lance/src/io/exec/scalar_index.rs
```diff
@@ -410,24 +421,20 @@ impl ExecutionPlan for MapIndexExec {
...
-            .boxed();
-        Ok(Box::pin(InstrumentedRecordBatchStreamAdapter::new(
+        let stream = futures::stream::once(stream_fut).try_flatten();
+        Ok(Box::pin(RecordBatchStreamAdapter::new(
             INDEX_LOOKUP_SCHEMA.clone(),
             stream,
-            partition,
-            &self.metrics,
         )))
     }
```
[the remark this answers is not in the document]
brendanclement: deferred as a Followup in the PR

rust/lance/src/io/exec/utils.rs
```diff
@@ -295,6 +298,131 @@ where
...
+/// applies a per-batch async transform.
+///
+/// `elapsed_compute` measures only the time spent driving the transform
+/// futures — never the time spent polling the child input — so wrapping a
+/// chain of nodes does not double-count child CPU. `output_rows` and
+/// `output_batches` are recorded as the transform produces batches.
+///
+/// `concurrency` caps how many transform futures may be in flight at once.
+/// Use `1` for sequential transforms; larger values parallelize per-batch
+/// work (e.g., KNN distance computation).
```
wjones127, line 311: note: This is pre-existing, but the fact that our streams do multiple CPU-bound things concurrency is generally an anti-pattern. Ideally, our DataFusion nodes would achieve concurrency through partitioning: creating N independent streams. We haven't migrated to that, but I suspect we will in the future.

rust/lance/src/io/exec/fts.rs
```diff
@@ -2051,6 +2081,27 @@ mod tests {
         assert!(metrics.elapsed_compute().unwrap() > 0);
     }
 
+    #[test]
+    fn test_flat_match_filter_find_matches_large_utf8() {
```
wjones127, line 2085: question: this a drive-by change fixing coverage? Or was there a bug found?

wjones127 (approved):
This looks good! No blocking feedback, just a minor question. I will merge soon.

rust/lance/src/io/exec/fts.rs
brendanclement: it was a coverage gap in the codecov report that got generated. It flagged the LargeUtf8 path as untested and it was working correctly before and now works correctly after we made the PR

wjones127:
For the test, I wonder if using semaphores might make them less flakey.

rust/lance-index/src/scalar/inverted/index.rs
westonpace, line 4333: There's not much work here but any reason we don't include it in the elapsed compute timer?  Seems like `scoring_start` could be extended to include the entire method body here (except for the await)?

rust/lance/src/io/exec/scalar_index.rs
```diff
@@ -288,6 +290,62 @@ impl MapIndexExec {
...
+        index_name: String,
+        index_metrics: Arc<IndexMetrics>,
+        metrics_set: ExecutionPlanMetricsSet,
+    ) -> datafusion::error::Result<datafusion::physical_plan::SendableRecordBatchStream> {
+        // Time the one-shot setup (fragment bitmap + deletion mask) so it's
+        // attributed to this node's elapsed_compute. The helper itself only
+        // times per-batch work.
+        let elapsed_compute = datafusion::physical_plan::metrics::MetricBuilder::new(&metrics_set)
+            .elapsed_compute(partition);
+        let setup_start = std::time::Instant::now();
```
westonpace, line 307: Is this timer capturing I/O time?  There's two awaits in its scope.  I think the timing we got from InstrumentedRecordBatchStreamAdapter was sufficient since we are not spawning background tasks here.

rust/lance/src/io/exec/utils.rs
```diff
@@ -295,6 +298,130 @@ where
...
+///
+/// For leaf nodes (no child input), use [`InstrumentedRecordBatchStreamAdapter`]
+/// instead.
+pub struct InstrumentedChildInputStream<F, Fut> {
```
westonpace, line 315: This is named very similarly to `InstrumentedRecordBatchStreamAdapter` and achieves nearly the same result.  Do we really need a second helper struct?  It kind of feels like this is just combining a map call, a buffer_unordered call, and a `InstrumentedRecordBatchStreamAdapter`.  I'm not sure this justifies an entirely new struct?

rust/lance/src/io/exec/utils.rs
```diff
@@ -295,6 +298,130 @@ where
...
+impl<F, Fut> InstrumentedChildInputStream<F, Fut>
+where
+    F: FnMut(RecordBatch) -> Fut,
+    Fut: Future<Output = DataFusionResult<RecordBatch>>,
+{
+    pub fn new(
+        input: SendableRecordBatchStream,
+        schema: SchemaRef,
+        transform: F,
+        concurrency: usize,
```
westonpace, line 335: Personally, I'd rather leave `concurrency` out, as KNN is a special case (it's the only spot that utilizes it), and we don't really want to emulate that pattern in the future.  Couldn't we just wrap the KNN stream with this adapter and call `buffered_unordered` on it instead of pushing concurrency into this adapter?

rust/lance/src/io/exec/utils.rs
```diff
@@ -295,6 +298,130 @@ where
...
+
+impl<F, Fut> InstrumentedChildInputStream<F, Fut>
+where
+    F: FnMut(RecordBatch) -> Fut,
+    Fut: Future<Output = DataFusionResult<RecordBatch>>,
+{
+    pub fn new(
+        input: SendableRecordBatchStream,
+        schema: SchemaRef,
+        transform: F,
```
westonpace, line 334: Why is a transform needed here?  Why can't we just apply the transform with a map function on the input?  Is it important the transform happen in this adapter?

westonpace (commented):
I think the FTS was definitely missing some elapsed compute time.  I'm not sure the changes to KNN, rowids, or scalar_index were needed (unless I am missing something).  I'm also confused why we need another helper struct for recording time?

rust/lance-index/src/scalar/inverted/index.rs
brendanclement: I only thought to bracket the scoring chunk but this is also CPU work and should be timed, you're right.

rust/lance/src/io/exec/scalar_index.rs
brendanclement: yes this is actually capturing I/O time, I added this after the claude bot review's comment on setup awaits not being captured anymore. Just using the InstrumentedRecordBatchStreamAdapter was the right call here

rust/lance/src/io/exec/utils.rs
brendanclement: This feedback, when I was figuring out how to fix the double counting I looked at how DataFusion solves the same problem and found FilterExecStream and modeled this helper on that so I could test that child poll time was excluded, but this is probably overkill here.

rust/lance/src/io/exec/utils.rs
brendanclement: Yeah i can pull this out and get the same result.

rust/lance/src/io/exec/utils.rs
brendanclement: Can pull this out as well

brendanclement:
> I think the FTS was definitely missing some elapsed compute time. I'm not sure the changes to KNN, rowids, or scalar_index were needed (unless I am missing something)

the issue #5155 called out KNN and AddRowAddr is being impacted and the benchmarking showed KNN elapsed compute time reducing when testing this out.

```
KNNVectorDistance: metric=l2, metrics=[output_rows=0, elapsed_compute=1.52917033s, output_batches=0]
FilterExec: scalar_id@1 = 100, metrics=[output_rows=0, elapsed_compute=8.036778ms]
```

The others were harder to demonstrate but in these cases the child's CPU was timed by the parent node

PR #6804, opened 2026-05-15
docs(index): document public API of LogicalScalarIndex

westonpace:
Add rustdoc to the LogicalScalarIndex struct and the two public loader functions (open_named_scalar_index, scalar_index_fragment_bitmap), covering when the wrapper is used, the same-index-type constraint, how SearchResult precision is combined across segments, and the read-only remap/update behavior.


westonpace:
@claude review once

wjones127 (approved):

PR #6806, opened 2026-05-16
feat: expose staged index segment transactions

### LINKED ISSUE #6666: Expose build_index_metadata_from_segments (or commit_existing_index_segments) for two-phase vector-index commits outside the lance crate
Companion to #6658 (two-phase delete).

## Request

Make `lance::index::build_index_metadata_from_segments` (currently `pub(crate)` at `src/index.rs:111`) public, OR expose a higher-level `commit_existing_index_segments(...)` Rust API that mirrors the Python pattern documented at [lance.org/guide/distributed_indexing](https://lance.org/guide/distributed_indexing/).

## Context

`CreateIndexBuilder::execute_uncommitted` (`src/index/create.rs:135`) is `pub` and returns `IndexMetadata`. But for index types that take the segment-commit path (`uses_segment_commit_path` returning true: Vector, IvfPq, IvfSq, IvfFlat, IvfRq, IvfHnswFlat, IvfHnswPq, IvfHnswSq), the `Operation::CreateIndex` transaction must be constructed via:

```rust
let segments = ds.create_index_segment_builder()
    .with_segments(vec![new_idx.clone()])
    .build_all().await?;
let new_indices = build_index_metadata_from_segments(ds, &name, field_id, segments).await?;
TransactionBuilder::new(version, Operation::CreateIndex { new_indices, removed_indices }).build()
```

`build_index_metadata_from_segments` is `pub(crate)`, so external callers cannot construct the transaction. This forces vector index builds to go through `CreateIndexBuilder::execute()` which inline-commits — incompatible with patterns that want a strict stage+commit separation across all writers.

The `IndexSegmentBuilder` (`src/index/create.rs:568`) and `with_segments` / `build_all` ARE `pub`, which suggests the intent was to expose the segment commit path. The `pub(crate)` on `build_index_metadata_from_segments` looks like an oversight.

A similar hard dependency exists on #6658 (two-phase delete). Both unblock the same architectural pattern: hoist Lance's stage+commit two-phase write to a load-bearing trait invariant in upper layers.

## Suggested fix

Either:

1. Mark `build_index_metadata_from_segments` `pub` (smallest change).
2. Add a higher-level `commit_existing_index_segments(...)` Rust API per the Python distributed-indexing docs.

Option 1 is the surgical fix; option 2 mirrors the documented Python pattern.

## Proposed signature (option 2)

```rust
impl<'a> IndexSegmentBuilder<'a> {
    /// Convert a segment plan into a fully-formed Operation::CreateIndex
    /// transaction without committing. Caller hands the returned
    /// transaction to CommitBuilder::execute to advance HEAD.
    pub async fn into_uncommitted_transaction(
        self,
        index_name: &str,
        field_id: i32,
        removed_indices: Vec<IndexMetadata>,
    ) -> Result<Transaction>;
}
```

Happy to send a PR if either approach is preferred.

ragnorc:
## Summary

Adds `Dataset::build_existing_index_segments_transaction`, which builds the
`Operation::CreateIndex` transaction for existing physical index segments
**without committing it**. Callers commit the returned `Transaction` via
`CommitBuilder` for a strict stage-then-commit workflow, mirroring
`InsertBuilder::execute_uncommitted` / `DeleteBuilder::execute_uncommitted`.
`commit_existing_index_segments` now delegates to it.

Fixes #6666.

## Design notes

- **Inherent method on `Dataset`**, not a new required method on the public
  `DatasetIndexExt` trait, so it is **non-breaking** for downstream trait
  implementors (addresses the earlier review feedback). The trait has a single
  in-tree implementor and is not object-safe, and the staging convention
  (`execute_uncommitted`) is otherwise inherent on builders/`Dataset` — so an
  inherent method is the most consistent, lowest-risk placement.
- Returns a bare `Transaction` (no wrapper struct): `CreateIndex` carries no
  `affected_rows`/side-channel, so there is nothing extra to reconcile at commit.
- Concurrency is enforced in two layers. The conflict resolver rejects any two
  index transactions that touch a common index name, in either direction, with
  a retryable conflict — creating over a concurrent removal must not resurrect
  a dropped index, and removing over a concurrent replacement must not silently
  no-op the drop. `build_manifest` then enforces the apply-time invariants that
  pairwise checks cannot enumerate: indexed fields must still exist in the final
  schema (a `drop_columns` projection or an `alter_columns` type cast removes or
  re-ids the field) and non-system removed segments must still be present. Two
  deliberate behavior changes: a concurrent double-drop and a commit whose staged
  removals were concurrently removed now fail retryably instead of silently
  succeeding. A compaction/rewrite that defers index remapping remains
  optimistic; committed segments may cover already-compacted fragments until the
  deferred remap catches up, so callers should still commit promptly.
- For inverted (FTS) segments, building the transaction finalizes the segment's
  on-disk files (idempotent; any files left behind on an abandoned build are
  reclaimed by `cleanup_old_versions`) — documented under `# Side effects`.

## Tests

- `test_build_existing_index_segments_transaction_does_not_commit` — staging
  does not publish the index; the returned transaction commits via
  `CommitBuilder`.
- `test_build_existing_index_segments_transaction_removes_empty_segment` — the
  zero-fragment placeholder guard fires on the staged path.
- `test_build_existing_index_segments_transaction_commits_after_version_advances`
  — a transaction staged at version N commits cleanly after an unrelated append
  moves HEAD past N.
- `test_build_existing_index_segments_transaction_conflicts_with_index_drop` —
  a staged commit over a concurrent `drop_index` fails with a retryable
  conflict instead of re-creating the dropped index. Fails without the fix.
- `test_build_existing_index_segments_transaction_conflicts_with_column_drop`
  — a staged commit after the indexed column is dropped fails as incompatible
  instead of publishing metadata for a nonexistent field. Fails without the
  fix.
- `test_build_existing_index_segments_transaction_conflicts_with_column_cast`
  — a staged commit after an `alter_columns` type cast (`Merge` with a new
  field id) fails as incompatible. Fails without the fix.
- `test_drop_index_conflicts_with_concurrent_replacement` — a drop racing a
  same-name replacement fails retryably instead of silently no-op'ing, and
  succeeds on retry. Fails without the fix.
- `test_index_transactions_conflict_on_shared_names` — resolver-level matrix:
  transactions on a shared name conflict in every direction (regular names and
  system singletons); disjoint names never conflict.

## Verification

- `cargo fmt --all`
- `cargo clippy -p lance --tests -- -D warnings`
- `cargo test -p lance --lib -- build_existing_index_segments_transaction index_transactions_conflict drop_index_conflicts` (8 passed)
- `cargo test -p lance --lib -- existing_index_segments conflict drop_index create_index stale frag_reuse mem_wal` (638 passed)
- `cargo test -p lance --doc build_existing_index_segments_transaction`

## Summary by CodeRabbit

- **New Features**
  - Added support for staging existing index segments as a transaction before publishing them.
  - Index metadata and segment coverage are validated before changes are committed.

- **Bug Fixes**
  - Improved handling of index replacements and placeholder segments.
  - Prevented staged index creation from restoring indexes removed concurrently.
  - Added conflict detection when indexed fields or columns are changed concurrently.
  - Staged transactions now commit correctly against the snapshot they were created from.


rust/lance/src/index.rs
```diff
@@ -1094,8 +1094,7 @@ impl DatasetIndexExt for Dataset {
             .any(|idx| idx.fields != [field.id])
         {
             return Err(Error::index(format!(
-                "Index name '{index_name}' already exists with different fields, \
-                please specify a different name"
+                "Index name '{index_name}' already exists with different fields,                 please specify a different name"
             )));
         }
         let removed_indices = existing_named_indices
```
Xuanwo: Zero-fragment placeholder index segments are still treated as disjoint, so committing real segments can leave the placeholder published and break index loading.

rust/lance/src/index/api.rs
```diff
@@ -305,6 +306,19 @@ pub trait DatasetIndexExt {
         source_segments: Vec<IndexMetadata>,
     ) -> Result<IndexMetadata>;
 
+    /// Build a transaction that publishes existing physical index segments.
+    ///
+    /// This stages the same manifest update as [`Self::commit_existing_index_segments`]
+    /// without advancing the dataset version. Callers that need a strict
+    /// stage-then-commit workflow can pass the returned transaction to
+    /// [`crate::dataset::CommitBuilder`].
+    async fn build_existing_index_segments_transaction(
```
Xuanwo: Adding this required method to a public unsealed trait breaks downstream implementors, so existing crates that implement the trait will fail to compile after upgrade.

ragnorc: Moved off the trait - `build_existing_index_segments_transaction` is now an inherent `Dataset` method, so `DatasetIndexExt` is unchanged and downstream implementors are not affected.

rust/lance/src/index/api.rs
```diff
@@ -305,6 +306,19 @@ pub trait DatasetIndexExt {
...
+        &self,
+        index_name: &str,
+        column: &str,
+        segments: Vec<impl IntoIndexSegment + Send>,
+    ) -> Result<Transaction>;
```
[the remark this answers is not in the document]
ragnorc: Now inherent on `Dataset` instead of a trait method, so no required trait method and no default body are needed.

rust/lance/src/index.rs
ragnorc: The relocated removal logic keeps the `existing_fragments.is_empty()` guard (#7141), so the placeholder is still removed; asserted by `test_build_existing_index_segments_transaction_removes_empty_segment`.

ragnorc:
@Xuanwo gently bumping this :)

rust/lance/src/index.rs
```diff
@@ -878,6 +878,155 @@ impl IndexDescription for IndexDescriptionImpl {
...
+    /// [`commit_existing_index_segments`](DatasetIndexExt::commit_existing_index_segments)
+    /// but does not advance the dataset version. Use
+    /// [`CommitBuilder`](crate::dataset::CommitBuilder) to commit the returned
+    /// [`Transaction`].
+    ///
+    /// The transaction is a snapshot built against the current dataset version,
+    /// so commit it promptly. A concurrent index creation with the same name is
+    /// rejected at commit time with a retryable conflict, but other concurrent
+    /// changes to the same index between staging and commit — a compaction/rewrite
+    /// that remaps it, or dropping/renaming the indexed column — are not
```
Xuanwo: The staged index transaction can commit after the indexed column or the same index is dropped, because those concurrent operations are treated as compatible. This can publish stale index metadata or re-create an index the user already removed.

ragnorc: Both cases are now rejected at commit time (18effc9a5). `drop_index` commits `CreateIndex` with only `removed_indices`, so the same-name check now also matches the other transaction's removals — a staged commit over a concurrent drop fails with a retryable conflict instead of re-creating the index. A concurrent `Project` is checked against the indexed field ids — a staged commit after the column is dropped fails as incompatible instead of publishing metadata for a nonexistent field. Asserted by `test_build_existing_index_segments_transaction_conflicts_with_index_drop` / `..._conflicts_with_column_drop`; both fail without the fix. The inline `commit_existing_index_segments` path shares the conflict resolver, so the same race there is closed as well.

rust/lance/src/index.rs
```diff
@@ -878,6 +878,156 @@ impl IndexDescription for IndexDescriptionImpl {
...
+    /// [`CommitBuilder`](crate::dataset::CommitBuilder) to commit the returned
+    /// [`Transaction`].
+    ///
+    /// The transaction is a snapshot built against the current dataset version.
+    /// Conflicting concurrent changes are rejected at commit time: creating or
+    /// dropping a same-name index between staging and commit fails the commit
+    /// with a retryable conflict, and dropping the indexed column fails it as
+    /// incompatible. A concurrent compaction/rewrite that defers index
+    /// remapping is not conflict-checked and may leave a duplicate index
+    /// entry, so commit promptly.
```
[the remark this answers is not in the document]
ragnorc: Duplicates cannot occur here: a removed segment is always fully covered by the incoming set (the orphan check in `build_existing_index_segments_transaction`), a non-deferred rewrite only remaps indices whose `old_fragments` intersect that coverage, and exactly those rewrites already fail this commit with a retryable conflict in `check_create_index_txn`. A deferred-remap rewrite leaves column-index UUIDs untouched, so the staged removal still replaces by UUID — the residual effect is stale fragment coverage until the deferred remap, not duplication, and rejecting those rewrites would defeat deferred remapping. The doc did overstate; reworded in 5ce6e96bc.

rust/lance/src/index.rs
```diff
@@ -878,6 +878,156 @@ impl IndexDescription for IndexDescriptionImpl {
...
+    pub async fn build_existing_index_segments_transaction(
+        &self,
+        index_name: &str,
+        column: &str,
+        segments: Vec<impl IntoIndexSegment + Send>,
+    ) -> Result<Transaction> {
+        let Some(field) = self.schema().field(column) else {
+            return Err(Error::index(format!(
+                "CreateIndex: column '{column}' does not exist"
+            )));
```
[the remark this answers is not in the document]
ragnorc: These are moved verbatim, and `CreateIndexBuilder` raises `Error::index` for the identical messages — changing only this path would give the same failure two different variants, and since `commit_existing_index_segments` now delegates here it would also change a shipped API error variant. Agreed with the guideline; deferring to a follow-up that sweeps all three paths together.

rust/lance/src/index.rs
```diff
@@ -878,6 +878,156 @@ impl IndexDescription for IndexDescriptionImpl {
...
+        }
+        let removed_indices = existing_named_indices
+            .into_iter()
+            .filter(|idx| {
+                idx.index_details
+                    .as_ref()
+                    .zip(incoming_type_url.as_deref())
+                    .is_none_or(|(details, expected)| details.type_url == expected)
+            })
```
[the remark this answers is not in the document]
ragnorc: `transaction.proto` only says `removed_indices` are "the indices that are being replaced" — there is no contract that every same-name index is replaced. The filter predates this PR (introduced deliberately in #6593; moved verbatim here), so changing the shipped inline path's replacement semantics is out of scope. The divergence from `CreateIndexBuilder` replace and the resulting `load_index_by_name` multiple-name failure are real — tracked in #7842.

rust/lance/src/index.rs
```diff
@@ -6947,6 +7001,397 @@ mod tests {
...
+
+        // Committing the staged transaction must not resurrect the dropped index.
+        let err = CommitBuilder::new(Arc::new(dataset.clone()))
+            .execute(transaction)
+            .await
+            .expect_err("staged index commit over a concurrent drop must conflict");
+        assert!(
+            matches!(err, Error::RetryableCommitConflict { .. }),
+            "expected a retryable commit conflict, got: {err}"
+        );
```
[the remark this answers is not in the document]
ragnorc: Added message-content assertions to both tests in 5ce6e96bc.

ragnorc:
Re CodeRabbit's outside-diff note on singleton removal detection: adopted in 5ce6e96bc — `other_has_frag_reuse`/`other_has_mem_wal` now also scan the other transaction's `removed_indices`, covered by `test_create_index_conflicts_with_concurrent_index_removal` (removal-only transactions against both singletons and a regular name).

ragnorc:
Follow-up to the review discussion: the conflict checks are now layered (c24fe74a1). Resolver rejects any two index transactions touching a common name in either direction - a drop racing a same-name replacement previously no-op'd silently, the mirror of the resurrection case. And `build_manifest` now enforces the apply-time invariants pairwise checks cannot enumerate: indexed fields must exist in the final schema (an `alter_columns` type cast commits `Merge` with a new field id and slipped past the Project-only check) and non-system removed segments must still be present

PR #6808, opened 2026-05-16
refactor(index): use arrow stats for scalar index stats

HaochengLIU:
## Summary
- Replace zonemap DataFusion min/max accumulators with `lance_arrow_stats::StatisticsAccumulator`.
- Use `arrow_stats` to track bloom filter zone null counts.
- Keep scalar-index NaN handling local to Lance instead of importing DataFusion math UDFs.
- Add `lance-arrow-stats` to the `lance-index` dependency graph and update the Python lockfile.

## Testing
- All tests pass.

### DIFF
--- a/rust/lance-index/src/scalar/zonemap.rs
+++ b/rust/lance-index/src/scalar/zonemap.rs
@@ -22,13 +22,14 @@ use crate::scalar::{
     BuiltinIndexType, CreatedIndex, SargableQuery, ScalarIndexParams, UpdateCriteria,
     compute_next_prefix,
 };
-use datafusion::functions_aggregate::min_max::{MaxAccumulator, MinAccumulator};
-use datafusion_expr::Accumulator;
+use lance_arrow_stats::StatisticsAccumulator;
 use lance_core::cache::{LanceCache, WeakLanceCache};
 use serde::{Deserialize, Serialize};
 use std::sync::LazyLock;
 
-use arrow_array::{ArrayRef, RecordBatch, UInt32Array, UInt64Array, new_empty_array};
+use arrow_array::{
+    ArrayRef, RecordBatch, UInt32Array, UInt64Array, new_empty_array, new_null_array,
+};
 use arrow_schema::{DataType, Field};
 use datafusion::execution::SendableRecordBatchStream;
 use datafusion_common::ScalarValue;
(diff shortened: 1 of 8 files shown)


HaochengLIU:
@claude review once

HaochengLIU:
@westonpace @jackye1995 can any of you review this PR? ty

rust/lance-index/src/scalar/zonemap.rs
```diff
@@ -131,6 +132,21 @@ impl DeepSizeOf for ZoneMapIndex {
...
 
 impl ZoneMapIndex {
+    fn scalar_is_nan(value: &ScalarValue) -> bool {
+        match value {
+            ScalarValue::Float16(Some(value)) => value.is_nan(),
+            ScalarValue::Float32(Some(value)) => value.is_nan(),
+            ScalarValue::Float64(Some(value)) => value.is_nan(),
+            _ => false,
+        }
+    }
```
westonpace, line 142: Let's add `is_nan` to `ArrowScalar` in the `arrow-scalar` crate?

rust/lance-index/src/scalar/zonemap.rs
```diff
@@ -131,6 +132,21 @@ impl DeepSizeOf for ZoneMapIndex {
...
+
+    fn zone_has_extrema(zone: &ZoneMapStatistics) -> bool {
```
westonpace:
```suggestion
    /// Returns true if both min and max are non-null / non-nan
    fn zone_has_extrema(zone: &ZoneMapStatistics) -> bool {
```

rust/lance-index/src/scalar/zonemap.rs
```diff
@@ -165,6 +181,10 @@ impl ZoneMapIndex {
                     return Ok(zone.nan_count > 0);
                 }
 
+                if !Self::zone_has_extrema(zone) {
+                    return Ok(false);
+                }
```
westonpace: Is this safe?  If we are using total ordering for floats then is it possible the min is -NaN and the max is NaN and the value we are looking for is somewhere in between?

westonpace (approved):
This is great, some small suggestions but thanks for cleaning this up!

rust/lance-index/src/scalar/zonemap.rs
HaochengLIU: added

rust/lance-index/src/scalar/zonemap.rs
HaochengLIU: break the function into a few for better readability

rust/lance-index/src/scalar/zonemap.rs
HaochengLIU: reworked the logic.
      - Replaced the vague zone_has_extrema helper with:
          - zone_has_finite_min
          - zone_has_finite_extrema
          - finite_value_may_be_in_zone
      - The important behavior is unchanged: if a zone has finite values plus NaNs, we store max = NaN to match the old DataFusion shape, but we do not use NaN as a real finite upper bound.
      - For finite lookups, a NaN max means “unknown finite max,” so zonemap keeps the zone as a false positive instead of incorrectly pruning it

rust/lance-index/src/scalar/zonemap.rs
westonpace: So now we can get rid of this function right?

westonpace (approved):
I have one more thing I want to try but let's move forward

rust/lance-index/src/scalar/zonemap.rs
westonpace: Ah, I see, we are still using datafusion scalars (and not just in zonemap but in sargablequery too :frowning:).  I'll handle that cleanup.

PR #6809, opened 2026-05-16
chore(python): update pyproject, docs, and cleanup makefile

geruh:
While working in the python bindings directory I kept hitting some issues while developing, so I cleaned up a few things that I was able to figure out.

The main problem was uv sync kept re-resolving the lockfile since our optional deps had unpinned packages, even if I was just running the tests. Fixed that by pinning the dev dependencies but ultimately can revert if im lacking context as to why they weren't pinned before. But now `uv sync` is stable out the box.

While I was in there I also, did an overhaul of the `MakeFile` so that install was just one command, leveraged a bit more uv there, and added help to document everything. Also, updated the docs that referenced the old logic to use new changes. Although there are a few more places that need this. i.e. direct `maturin` references. 

After this PR:

```
cd python && make install # once
uv run make test  # every day
make build  # after Rust changes
uv sync # after pulling commits with dep changes
```

### DIFF
--- a/.github/workflows/ci-benchmarks.yml
+++ b/.github/workflows/ci-benchmarks.yml
@@ -47,8 +47,8 @@ jobs:
         run: |
           python -m venv venv
           source venv/bin/activate
-          pip install maturin duckdb requests pytest pytest-benchmark datasets
-          maturin develop --locked --release --features datagen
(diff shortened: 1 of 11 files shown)


jackye1995 (approved):
thanks for the fix! Pending CI

Xuanwo:
Python Lint failed

PR #6810, opened 2026-05-17
fix: reject changing the unenforced primary key once set

touch-of-grey:
## Summary

The unenforced primary key is stored as schema field metadata, and the
`UpdateConfig` transaction-apply path applied field metadata updates with no
guard — so the key could be silently overridden, evolved, removed, or written
with a junk value.

This treats the `lance-schema:unenforced-primary-key*` metadata keys as
reserved, validated in the transaction-apply path:

- once a primary key is set, any commit that changes it — or writes a reserved
  primary-key metadata key at all — is rejected;
- writing a reserved key with a value that is not a valid marker is rejected
  rather than silently ignored;
- a valid first install and updates to other field metadata are unaffected.

The check runs on every apply, including conflict-rebase, so it also rejects
the concurrent-writer race where two writers set the key on different columns
(their field-metadata updates don't conflict, so the loser would otherwise
rebase and corrupt the key).

The `metadata.rs` primary-key tests are updated accordingly.

### DIFF
--- a/rust/lance/src/dataset/metadata.rs
+++ b/rust/lance/src/dataset/metadata.rs
@@ -540,66 +540,72 @@ mod tests {
 
     #[tokio::test]
     async fn test_update_field_metadata_syncs_unenforced_primary_key_position() {
-        // Regression test: updating field metadata to install (or remove) the
-        // unenforced primary key keys must keep the cached
-        // `unenforced_primary_key_position` in sync, otherwise the next
-        // commit drops the marker because the protobuf is encoded from the
-        // cached option, not from the metadata HashMap.
+        // Regression test: installing the unenforced primary key via field
+        // metadata must keep the cached `unenforced_primary_key_position` in
+        // sync with the metadata HashMap, otherwise the next commit drops the
+        // marker because the protobuf is encoded from the cached option.
+        //
+        // The primary key is immutable once set, so each install case uses a
+        // fresh dataset.
         use lance_core::datatypes::{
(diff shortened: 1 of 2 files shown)


rust/lance/src/dataset/metadata.rs
```diff
@@ -617,10 +623,46 @@ mod tests {
                 truthy
             );
         }
-        for falsy in ["no", "false", "0", "anything-else"] {
+
+        // Non-matching values must not be treated as a PK marker, so the
+        // field never becomes part of the primary key.
```
jackye1995: not only we should not treat it as a pk marker, we should error with something like the primary key is a reserved key.

rust/lance/src/dataset/metadata.rs
```diff
@@ -630,53 +672,80 @@ mod tests {
...
-                .unwrap()
-                .unenforced_primary_key_position,
-            Some(0),
-            "garbage position must fall through to the boolean fallback",
-        );
+        assert_eq!(dataset.schema().unenforced_primary_key().len(), 1);
 
-        // Removing the keys must clear the cached option, otherwise the
-        // protobuf would still encode a stale PK marker on next commit.
+        // Re-applying the identical primary key is a no-op and allowed; this
```
jackye1995: this is confusing, we should just not allow setting it like the rest of the behavior, instead of allowing it to succeed.

jackye1995 (approved):
thanks for the fix!

PR #6811, opened 2026-05-17
fix(encoding): use LargeList internally for Utf8/Binary to prevent offset overflow

### LINKED ISSUE #2775: Offset overflow errors can be confusing for users
When using `binary` or `string` columns a single batch of data cannot contain more than 2GiB of data.  Users will either need to use `large_binary` and `large_string` or make sure to set a custom batch size when reading this data.

However, the error they run into, an "offset overflow" error, is a panic (not great) and very confusing.  It is not obvious that the solution is to reduce the batch size:

```
thread 'lance_background_thread' panicked at .../arrow-data-52.2.0/src/transform/utils.rs:42:56:
offset overflow
thread '<unnamed>' panicked at .../rust/lance-encoding/src/decoder.rs:1267:65:
called `Result::unwrap()` on an `Err` value: JoinError::Panic(Id(12814), ...)
```

Ideally we should be returning an `Err` here (not panic) and the message should say something like "Could not create array with more than 2GiB of string/binary data.  Please try reducing the batch_size."

beinan:
## Summary
When reading large fragments containing variable-length data (like strings or binary), PyArrow's 32-bit (2 GiB) offset limit can be easily exceeded, resulting in a cryptic `Offset overflow error: 2151740513` panic.

This updates the decoder to always use `LargeList<UInt8>` (i64 offsets) internally. When targeting `Utf8` or `Binary` arrays, `BinaryArrayDecoder` will safely downcast to i32 offsets if the chunk data fits within 2 GiB. If the chunk strictly exceeds 2 GiB, it now panics with a descriptive error suggesting a batch size reduction or switching to LargeUtf8/LargeBinary.

Fixes #2775.

🤖 Generated with [MAI Agents](https://github.com/infinity-microsoft/mai-agents)

### DIFF
--- a/rust/lance-encoding/src/decoder.rs
+++ b/rust/lance-encoding/src/decoder.rs
@@ -837,15 +837,16 @@ impl CoreFieldDecoderStrategy {
                         array_encoding: Some(pb::array_encoding::ArrayEncoding::List(..))
                     }
                 ) {
-                    let list_type = if matches!(data_type, DataType::Utf8 | DataType::Binary) {
-                        DataType::List(Arc::new(ArrowField::new("item", DataType::UInt8, false)))
-                    } else {
-                        DataType::LargeList(Arc::new(ArrowField::new(
-                            "item",
-                            DataType::UInt8,
-                            false,
-                        )))
-                    };
+                    // Always use LargeList (i64 offsets) internally to avoid
+                    // "Offset overflow" errors when a single batch exceeds 2GB
+                    // of variable-length data.  The BinaryArrayDecoder will
+                    // downcast back to i32 offsets when the target type is
+                    // Utf8 / Binary and the data fits.
+                    let list_type = DataType::LargeList(Arc::new(ArrowField::new(
+                        "item",
+                        DataType::UInt8,
+                        false,
+                    )));
                     let list_field = Field::try_from(ArrowField::new(
                         field.name.clone(),
                         list_type,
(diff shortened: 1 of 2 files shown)


beinan:
I ran some quick benchmarks on an x86 pod to measure the impact of the `i64` offset buffering.

Here are the results comparing this branch against `main` for the UTF-8 decoder benchmarks:

**Standard UTF-8 Data (`decode_primitive/utf8`)**
*   **Main:** 832.32 µs (458.83 MiB/s)
*   **This PR:** 830.20 µs (460.00 MiB/s)
*   **Verdict:** No change in performance detected. The vectorized `i64` math combined with the fast offset downcast adds essentially zero overhead. 

**Fixed-Length UTF-8 Data (`decode_primitive/fixed-utf8`)**
*   **Main:** 195.93 µs (778.77 MiB/s)
*   **This PR:** 217.07 µs (702.93 MiB/s)
*   **Verdict:** ~11.8% regression in throughput. Building the `LargeList` internal buffer does slightly more work (allocating/writing 64-bit offsets), which makes a statistically significant difference when the strings are strictly fixed-size and the benchmark finishes in microseconds.

Overall, the safety and crash prevention added by the `i64` offset buffering has zero impact on typical variable-length string data, with a small penalty for perfectly fixed-length string columns.

wjones127 (commented):
I think my preferred solution to this was to:

1. Use the https://docs.rs/arrow/latest/arrow/array/struct.GenericByteViewBuilder.html#method.try_append_value API so overflow errors instead of panicking
2. If it errors, suggest to the user to alter the column data type to the large variant.

Did you find there was a reason that wouldn't work?

wjones127:
@beinan do you have an answer to the question above?

beinan:
> @beinan do you have an answer to the question above?

Awesome!  let me use try_append_value and verify, will get back to you shortly

PR #6812, opened 2026-05-17
feat: support updating the unenforced clustering key via UpdateConfig

touch-of-grey:
## Summary

#6552 added the unenforced clustering key to the schema format, but the
`UpdateConfig` transaction-apply path does not recompute the cached
`Field::unenforced_clustering_key_position` after applying field metadata
updates — only `unenforced_primary_key_position` is recomputed (added for the
primary key in #6706). The protobuf encoder reads the cached option, not the
metadata HashMap, so installing or removing an unenforced clustering key via
`update_field_metadata` does not round-trip.

This recomputes `unenforced_clustering_key_position` on apply, and treats the
clustering key as a reserved, immutable schema property — mirroring the
unenforced primary key:

- once a clustering key is set, any commit that changes it, or that writes its
  reserved metadata key, is rejected;
- writing the reserved key with a value that is not a valid position is
  rejected rather than silently ignored;
- a valid first install (single- or multi-column) and updates to other field
  metadata are unaffected.

The check runs on every apply, including conflict-rebase, so it also rejects
the concurrent-writer race.

`LANCE_UNENFORCED_CLUSTERING_KEY_POSITION` is also re-exported from
`lance_core::datatypes`, alongside the primary-key constants.

The metadata-update tests for both the primary and clustering keys are
restructured into individual tests with simpler `lance_datagen`-based setup.

### DIFF
--- a/rust/lance/src/dataset/metadata.rs
+++ b/rust/lance/src/dataset/metadata.rs
@@ -540,117 +540,90 @@ mod tests {
 
     #[tokio::test]
     async fn test_update_field_metadata_syncs_unenforced_primary_key_position() {
-        // Regression test: installing the unenforced primary key via field
-        // metadata must keep the cached `unenforced_primary_key_position` in
-        // sync with the metadata HashMap, otherwise the next commit drops the
-        // marker because the protobuf is encoded from the cached option.
-        //
-        // The primary key is immutable once set, so each install case uses a
-        // fresh dataset.
-        use lance_core::datatypes::{
-            LANCE_UNENFORCED_PRIMARY_KEY, LANCE_UNENFORCED_PRIMARY_KEY_POSITION,
-        };
+        // Installing the unenforced primary key via field metadata must keep
+        // the cached `unenforced_primary_key_position` in sync with the
+        // metadata HashMap, otherwise the next commit drops the marker because
+        // the protobuf is encoded from the cached option.
+        use lance_core::datatypes::LANCE_UNENFORCED_PRIMARY_KEY_POSITION;
 
-        async fn fresh_dataset(uri: &str) -> Dataset {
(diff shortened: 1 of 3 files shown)


rust/lance/src/dataset/metadata.rs
```diff
@@ -698,4 +698,198 @@ mod tests {
...
+                RecordBatchIterator::new(vec![Ok(batch)], schema.clone()),
+                uri,
+                None,
+            )
+            .await
+            .unwrap()
+        }
+
+        // Single-column clustering key, installed via the position key.
+        {
```
jackye1995: nit: all these should be individual tests, prefer to not use different sections in a single test. You might also want to update your previous primary key PR about it.

rust/lance/src/dataset/metadata.rs
```diff
@@ -698,4 +698,198 @@ mod tests {
...
+        // Installing the unenforced clustering key via field metadata must keep
+        // the cached `unenforced_clustering_key_position` in sync with the
+        // metadata HashMap, otherwise the next commit drops the marker because
+        // the protobuf is encoded from the cached option.
+        //
+        // The clustering key is immutable once set, so each case uses a fresh
+        // dataset.
+        use lance_core::datatypes::LANCE_UNENFORCED_CLUSTERING_KEY_POSITION;
+
+        async fn fresh_dataset(uri: &str) -> Dataset {
```
jackye1995: this is repeated. I think it is okay to just repeat this. we also have datagen module that can be used to simplify this dataset creation, check other tests. You might also want to update your previous primary key PR about it.

jackye1995 (approved):
👍

jackye1995 (approved):
👍 thanks for adding this

PR #6813, opened 2026-05-17
refactor: rename ShardSpec to ShardingSpec

touch-of-grey:
## Summary

Renames the `ShardSpec` / `ShardField` types — and the `shard_spec` /
`shard_specs` identifiers and the corresponding protobuf messages — to
`ShardingSpec` / `ShardingField` for naming consistency.

The protobuf message names change and the `shard_specs` field is renamed to
`sharding_specs`; all field numbers and tags are unchanged, so the change is
wire-compatible.


jackye1995 (approved):
Thanks for the fix!

PR #6815, opened 2026-05-17
feat: builder-style MemWAL initialization API

touch-of-grey:
## Summary

Replaces the struct-based MemWAL initialization API with a fluent builder on
`Dataset`, and persists default `ShardWriter` configuration in the MemWAL index.

`Dataset::initialize_mem_wal()` returns an `InitializeMemWalBuilder`:

```rust
dataset
    .initialize_mem_wal()
    .bucket_sharding("id", 16)
    .maintained_indexes(["id_btree"])
    .writer_config_defaults(ShardWriterConfig::default().with_durable_write(false))
    .execute()
    .await?;
```

- `bucket_sharding(column, num_buckets)`, `unsharded()`, and
  `identity_sharding(column)` are high-level sharding strategies. They own the
  `ShardingSpec` construction and validation that callers previously did by
  hand; `num_shards` is derived from the sharding choice.
- `writer_config_defaults(ShardWriterConfig)` records the tunable writer
  configuration as the persisted defaults — a new `map<string, string>
  writer_config_defaults` field on the `MemWalIndexDetails` protobuf message —
  so every writer, across processes and restarts, starts from the same
  defaults. `add_writer_config_default` records arbitrary extra keys.
- The Python `initialize_mem_wal` binding accepts the full builder surface
  (sharding, maintained indexes, writer-config defaults) and invokes the
  builder; a new `mem_wal_index_details` binding reads the recorded details back.
- Removed: `MemWalConfig`, `MemWalShardConfig`, `initialize_mem_wal_with_shards`.

## Context

While reviewing lancedb/lancedb#3396, @jackye1995 noted that per-writer tuning
knobs are runtime configuration and should not be persisted as part of the
sharding spec. That holds for a writer's *live* `ShardWriterConfig`, which stays
non-persisted. This PR records only the *defaults*: without a persisted
default, writers would be configured independently and could silently drift
apart.

@jackye1995 also asked for a builder style so bucket sharding is easy to set,
moving the hash-bucket spec logic out of the LanceDB layer into Lance. The
builder owns all three sharding strategies (bucket, unsharded, identity), so
LanceDB's `set_lsm_write_spec` can call the builder directly instead of
hand-building shard specs.

### DIFF
--- a/protos/table.proto
+++ b/protos/table.proto
@@ -673,6 +673,15 @@ message MemWalIndexDetails {
   //
   // If an index is not present in this list, it is assumed to be fully caught up.
   repeated IndexCatchupProgress index_catchup = 10;
+
+  // Default ShardWriter configuration values for this MemWAL index.
+  //
+  // A free-form string map persisted so that every writer — across
+  // processes and restarts — starts from the same default writer
+  // configuration. These are defaults only: an individual writer may
+  // still override any value at runtime in its own ShardWriterConfig
+  // (which is not persisted).
+  map<string, string> writer_defaults = 11;
 }
 
 // Sharding spec definition.
--- a/protos/table.proto
+++ b/protos/table.proto
@@ -681,7 +681,7 @@ message MemWalIndexDetails {
   // configuration. These are defaults only: an individual writer may
   // still override any value at runtime in its own ShardWriterConfig
   // (which is not persisted).
-  map<string, string> writer_defaults = 11;
+  map<string, string> writer_config_defaults = 11;
 }
 
 // Sharding spec definition.
--- a/python/src/dataset.rs
+++ b/python/src/dataset.rs
@@ -3152,49 +3152,23 @@ impl Dataset {
     /// Must be called once before any `mem_wal_writer()` calls.
     /// Requires the dataset schema to have at least one field with
     /// the `lance-schema:unenforced-primary-key` metadata.
-    #[pyo3(signature=(maintained_indexes=None, region_spec=None))]
+    #[pyo3(signature=(maintained_indexes=None))]
     fn initialize_mem_wal(
         &mut self,
         py: Python<'_>,
         maintained_indexes: Option<Vec<String>>,
-        region_spec: Option<Bound<'_, PyAny>>,
     ) -> PyResult<()> {
         use lance::dataset::mem_wal::DatasetMemWalExt;
-        use lance_index::mem_wal::{ShardingField as RegionField, ShardingSpec as RegionSpec};
-        use std::collections::HashMap;
 
-        let region_spec_rust = if let Some(spec) = region_spec {
-            let spec_id: u32 = spec.getattr("spec_id")?.extract()?;
-            let fields_py: Vec<Bound<'_, PyAny>> = spec.getattr("fields")?.extract()?;
-            let fields = fields_py
-                .iter()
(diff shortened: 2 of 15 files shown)


python/src/dataset.rs
```diff
@@ -3187,6 +3187,7 @@ impl Dataset {
         };
 
         let config = lance::dataset::mem_wal::MemWalConfig {
+            writer_defaults: Default::default(),
```
jackye1995: I think at this point we should probably use a builder style. Some logic in your lancedb PR can also be moved here like the logic to do sharding with hash bucket, not shard, etc.

python/src/dataset.rs
```diff
@@ -3152,48 +3152,23 @@ impl Dataset {
     /// Must be called once before any `mem_wal_writer()` calls.
     /// Requires the dataset schema to have at least one field with
     /// the `lance-schema:unenforced-primary-key` metadata.
-    #[pyo3(signature=(maintained_indexes=None, region_spec=None))]
+    #[pyo3(signature=(maintained_indexes=None))]
```
jackye1995: Should make sure python can take all input segments in this function and then invoke the builder in the binding

rust/lance/src/dataset/mem_wal/api.rs
```diff
@@ -6,83 +6,368 @@
...
-    ) -> Result<()> {
-        if shard_config.num_shards == 0 {
-            self.initialize_mem_wal(config).await
-        } else {
-            Err(Error::invalid_input(
-                "initialize_mem_wal_with_shards is not implemented for this DatasetMemWalExt implementer",
-            ))
+    /// These are defaults only; an individual writer may still override any
+    /// value at runtime in its own (non-persisted) `ShardWriterConfig`.
+    pub fn writer_defaults<I, K, V>(mut self, defaults: I) -> Self
```
jackye1995: We should just move all the ones you had in lancedb over here to allow setting specific config defaults. We can leave a add_writer_default to add arbitrary config defaults.

protos/table.proto
```diff
@@ -673,6 +673,15 @@ message MemWalIndexDetails {
...
   repeated IndexCatchupProgress index_catchup = 10;
+
+  // Default ShardWriter configuration values for this MemWAL index.
+  //
+  // A free-form string map persisted so that every writer — across
+  // processes and restarts — starts from the same default writer
+  // configuration. These are defaults only: an individual writer may
+  // still override any value at runtime in its own ShardWriterConfig
+  // (which is not persisted).
+  map<string, string> writer_defaults = 11;
```
jackye1995: we should also rename it as `writer_config_defaults`

rust/lance/src/dataset/mem_wal/api.rs
```diff
@@ -6,83 +6,439 @@
...
+        self.writer_defaults
+            .extend(writer_config_to_defaults(&config));
+        self
+    }
+
+    /// Record a single arbitrary writer-configuration default.
+    ///
+    /// Use this for keys not covered by
+    /// [`writer_config_defaults`](Self::writer_config_defaults).
+    pub fn add_writer_default(mut self, key: impl Into<String>, value: impl Into<String>) -> Self {
```
jackye1995: similarly, `add_writer_config_default`

jackye1995 (approved):
this looks great! thanks for the fixes

PR #6828, opened 2026-05-18
feat: support batch vector queries

### LINKED ISSUE #6821: Support batch vector query API and shared flat KNN scan
Parent: #6820
Linear: https://linear.app/lancedb/issue/OSS-923/support-batch-vector-query-api-and-shared-flat-knn-scan
Linear parent: https://linear.app/lancedb/issue/OSS-922/vector-search-performance-optimization-roadmap
Related historical issue: #876

## Goal

Allow one vector search request to contain multiple query vectors and make the first implementation immediately useful for flat KNN by sharing scan/load work across the query batch.

## Scope

- Add or finalize the public API shape for batch query vectors in Rust and Python.
- Return results with an explicit per-query discriminator, for example `query_index`, so callers can split top-k results by query vector.
- Apply `limit`/top-k per query, not globally across all query vectors.
- For flat KNN, load each data batch once and evaluate all query vectors against that loaded batch.
- It is acceptable for the first version to keep ANN/indexed search out of scope, as long as the behavior is explicit and covered by tests.

## Acceptance criteria

- Rust tests compare batch query results against repeated single-query results.
- Python tests cover 2-D query input and validate output grouping by query id.
- Docs/docstrings explain the batch input shape and result contract.
- A small flat KNN benchmark demonstrates lower tail latency versus issuing repeated independent queries, especially when I/O or decode cost dominates.

LeoReeYang:
## Summary

Closes #6821.

- Extend `Scanner::nearest` for batched queries on fixed-size vector columns (no separate `nearest_batch` API).
- Flat batch KNN in `KNNVectorDistanceExec`: one scan per data batch, per-query top-k in a single stream (`m × k` rows max).
- Add `query_index` (0-based) so callers can group results per input query.
- Indexed path (`use_index=true`): per-query ANN, union, and `query_index` tagging when a vector index exists.
- `distance_range` applied before per-query top-k on the flat path; indexed batch uses the same bounds.
- `fast_search` without an index returns an empty batch result that still includes `query_index`.
- Batch vs multivector: `FixedSizeList` column + list-like query → batch of single-vector queries; `List` multivector column → one multivector query.

## API

| | |
| --- | --- |
| **Input** | List-like or 2-D query against a `FixedSizeList` embedding column |
| **Output** | Up to `k` rows per query vector + `query_index` |
| **Flat** | Shared scan/decode; requires `_rowid` |
| **Indexed** | One indexed search per query vector, then merge |

## Benchmark

Local disk, float32 vectors, `use_index=false`. OS page cache accepted. Medians over 15 timed rounds (3 warmup); separate then batch each round.

```bash
cd python && uv run --extra benchmarks pytest python/python/benchmarks/test_search.py::test_batch_flat_knn
```

At small `m`, the second separate query often hits warm page cache, so speedup is modest (~1.1×). It grows with `m` as scan work is shared.

### Query count (1M rows, dim 512, k 10)

| m | separate | batch | saved | speedup |
| --: | --: | --: | --: | --: |
| 2 | 227.48 ms | 208.46 ms | 19.02 ms | 1.09× |
| 5 | 559.10 ms | 328.84 ms | 230.26 ms | 1.70× |
| 10 | 1.1125 s | 536.48 ms | 576.01 ms | 2.07× |

### Dataset size (m 10, dim 512, k 10)

| rows | separate | batch | saved | speedup |
| --: | --: | --: | --: | --: |
| 100,000 | 123.84 ms | 53.23 ms | 70.62 ms | 2.33× |
| 500,000 | 566.20 ms | 261.00 ms | 305.20 ms | 2.17× |
| 1,000,000 | 1.1104 s | 520.86 ms | 589.50 ms | 2.13× |

## Test plan

- [x] `cargo test -p lance --lib test_batch_knn`
- [x] `cargo test -p lance fast_search_without`
- [x] `uv run pytest python/tests/test_vector_index.py -k batch`
- [x] `uv run ruff format --check python/lance/dataset.py python/tests/test_vector_index.py`

### DIFF
--- a/python/python/benchmarks/test_search.py
+++ b/python/python/benchmarks/test_search.py
@@ -1,6 +1,5 @@
 # SPDX-License-Identifier: Apache-2.0
 # SPDX-FileCopyrightText: Copyright The Lance Authors
-import os
 import shutil
 from pathlib import Path
 from typing import NamedTuple, Union
@@ -68,13 +67,6 @@ def create_table(num_rows, offset) -> pa.Table:
             "genres": genres,
         }
     )
-
-
-def get_benchmark_env(key: str, default: int) -> int:
-    value = os.environ.get(key)
-    if value is None:
-        return default
-    return int(value)
 
 
 def create_flat_vector_table(num_rows: int, offset: int, dim: int) -> pa.Table:
@@ -224,16 +216,29 @@ def test_knn_search(test_dataset, benchmark):
 
 @pytest.mark.benchmark(group="batch_flat_knn")
 @pytest.mark.parametrize("mode", ["separate", "batch"])
-def test_batch_flat_knn(data_dir: Path, benchmark, mode: str):
-    dim = get_benchmark_env("LANCE_BATCH_KNN_DIM", BATCH_FLAT_KNN_DIM)
-    num_rows = get_benchmark_env("LANCE_BATCH_KNN_ROWS", BATCH_FLAT_KNN_ROWS)
-    batch_size = get_benchmark_env(
-        "LANCE_BATCH_KNN_BATCH_SIZE", BATCH_FLAT_KNN_BATCH_SIZE
-    )
-    query_count = get_benchmark_env(
-        "LANCE_BATCH_KNN_QUERY_COUNT", BATCH_FLAT_KNN_QUERY_COUNT
-    )
-    rounds = get_benchmark_env("LANCE_BATCH_KNN_ROUNDS", 10)
+@pytest.mark.parametrize(
+    ("dim", "num_rows", "batch_size", "query_count", "rounds"),
+    [
+        (
+            BATCH_FLAT_KNN_DIM,
+            BATCH_FLAT_KNN_ROWS,
+            BATCH_FLAT_KNN_BATCH_SIZE,
+            BATCH_FLAT_KNN_QUERY_COUNT,
+            10,
+        )
+    ],
+    ids=["1m_rows_512d_m10"],
+)
+def test_batch_flat_knn(
+    data_dir: Path,
+    benchmark,
+    mode: str,
+    dim: int,
+    num_rows: int,
+    batch_size: int,
+    query_count: int,
+    rounds: int,
+):
     dataset = create_batch_flat_knn_dataset(data_dir, num_rows, batch_size, dim)
     query_table = dataset.to_table(columns=["vector"], limit=query_count)
     query_values = np.asarray(
--- a/python/python/benchmarks/test_search.py
+++ b/python/python/benchmarks/test_search.py
@@ -1,5 +1,6 @@
 # SPDX-License-Identifier: Apache-2.0
 # SPDX-FileCopyrightText: Copyright The Lance Authors
+import os
 import shutil
 from pathlib import Path
 from typing import NamedTuple, Union
@@ -13,6 +14,11 @@ import pytest
 N_DIMS = 768
 NUM_ROWS = 100_000
 NEW_ROWS = 10_000
+BATCH_FLAT_KNN_DIM = 512
+BATCH_FLAT_KNN_K = 10
+BATCH_FLAT_KNN_BATCH_SIZE = 10_000
+BATCH_FLAT_KNN_QUERY_COUNT = 10
+BATCH_FLAT_KNN_ROWS = 1_000_000
 
 
 def find_or_clean(dataset_path: Path) -> Union[lance.LanceDataset, None]:
@@ -62,6 +68,49 @@ def create_table(num_rows, offset) -> pa.Table:
             "genres": genres,
         }
     )
+
+
+def get_benchmark_env(key: str, default: int) -> int:
+    value = os.environ.get(key)
+    if value is None:
+        return default
+    return int(value)
+
+
+def create_flat_vector_table(num_rows: int, offset: int, dim: int) -> pa.Table:
+    rng = np.random.default_rng(seed=offset)
+    values = rng.random((num_rows, dim), dtype=np.float32)
+    vectors = pa.FixedSizeListArray.from_arrays(pa.array(values.ravel()), dim)
+    ids = pa.array(range(offset, offset + num_rows))
+    return pa.table({"vector": vectors, "id": ids})
+
+
+def create_batch_flat_knn_dataset(
+    data_dir: Path, num_rows: int, batch_size: int, dim: int
+) -> lance.LanceDataset:
+    tmp_path = data_dir / f"batch_flat_knn_{num_rows}_{batch_size}_{dim}"
+    dataset = find_or_clean(tmp_path)
+    if dataset:
+        return dataset
+
+    rows_remaining = num_rows
+    offset = 0
+    dataset = None
+    while rows_remaining > 0:
+        next_batch_length = min(rows_remaining, batch_size)
+        rows_remaining -= next_batch_length
+        table = create_flat_vector_table(next_batch_length, offset, dim)
+        if offset == 0:
+            dataset = lance.write_dataset(
+                table, tmp_path, data_storage_version="stable"
+            )
+        else:
+            dataset = lance.write_dataset(
+                table, tmp_path, mode="append", data_storage_version="stable"
+            )
+        offset += next_batch_length
+
+    return dataset
 
 
 def create_base_dataset(data_dir: Path) -> lance.LanceDataset:
@@ -171,6 +220,57 @@ def test_knn_search(test_dataset, benchmark):
         ),
     )
     assert result.num_rows > 0
+
+
+@pytest.mark.benchmark(group="batch_flat_knn")
+@pytest.mark.parametrize("mode", ["separate", "batch"])
+def test_batch_flat_knn(data_dir: Path, benchmark, mode: str):
+    dim = get_benchmark_env("LANCE_BATCH_KNN_DIM", BATCH_FLAT_KNN_DIM)
+    num_rows = get_benchmark_env("LANCE_BATCH_KNN_ROWS", BATCH_FLAT_KNN_ROWS)
+    batch_size = get_benchmark_env(
+        "LANCE_BATCH_KNN_BATCH_SIZE", BATCH_FLAT_KNN_BATCH_SIZE
+    )
+    query_count = get_benchmark_env(
+        "LANCE_BATCH_KNN_QUERY_COUNT", BATCH_FLAT_KNN_QUERY_COUNT
+    )
+    rounds = get_benchmark_env("LANCE_BATCH_KNN_ROUNDS", 10)
+    dataset = create_batch_flat_knn_dataset(data_dir, num_rows, batch_size, dim)
+    query_table = dataset.to_table(columns=["vector"], limit=query_count)
+    query_values = np.asarray(
+        query_table["vector"].combine_chunks().values, dtype=np.float32
+    ).reshape(query_count, dim)
+
+    def separate_queries():
+        total_rows = 0
+        for query in query_values:
+            total_rows += dataset.to_table(
+                columns=[],
+                nearest={
+                    "column": "vector",
+                    "q": query,
+                    "k": BATCH_FLAT_KNN_K,
+                    "use_index": False,
+                },
+            ).num_rows
+        return total_rows
+
+    def batch_query():
+        return dataset.to_table(
+            columns=[],
+            nearest={
+                "column": "vector",
+                "q": query_values,
+                "k": BATCH_FLAT_KNN_K,
+                "use_index": False,
+            },
+        ).num_rows
+
+    if mode == "separate":
+        result = benchmark.pedantic(separate_queries, rounds=rounds, iterations=1)
+    else:
+        result = benchmark.pedantic(batch_query, rounds=rounds, iterations=1)
+
+    assert result == query_count * BATCH_FLAT_KNN_K
 
 
 @pytest.mark.benchmark(group="query_ann")
--- a/python/python/benchmarks/test_search.py
+++ b/python/python/benchmarks/test_search.py
@@ -13,11 +13,6 @@ import pytest
 N_DIMS = 768
 NUM_ROWS = 100_000
 NEW_ROWS = 10_000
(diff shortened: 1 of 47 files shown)


rust/lance/src/dataset/scanner.rs
```diff
@@ -1526,6 +1530,120 @@ impl Scanner {
...
+        self.nearest_query_count = 1;
+        Ok(self)
+    }
+
+    /// Find the k-nearest neighbors for each query vector in a batch.
+    ///
+    /// This first implementation uses the flat scan path. It reads candidate
+    /// vectors once and computes distances for all query vectors, returning up
+    /// to `k` rows per query with an additional `_query_index` column.
+    pub fn nearest_batch(&mut self, column: &str, q: &dyn Array, k: usize) -> Result<&mut Self> {
```
BubbleCal: my understanding is that we don't need a new method here, just let `nearest()` support batch query

rust/lance/src/dataset/scanner.rs
```diff
@@ -1626,7 +1744,12 @@ impl Scanner {
     /// This is essentially a weak consistency search, only on the indexed data.
     pub fn fast_search(&mut self) -> &mut Self {
         if let Some(q) = self.nearest.as_mut() {
-            q.use_index = true;
+            if self.nearest_query_count > 1 {
```
BubbleCal: this seems wrong

rust/lance/src/dataset/scanner.rs
```diff
@@ -1686,7 +1809,12 @@ impl Scanner {
     /// Set whether to use the index if available
     pub fn use_index(&mut self, use_index: bool) -> &mut Self {
         if let Some(q) = self.nearest.as_mut() {
-            q.use_index = use_index
+            if self.nearest_query_count > 1 && use_index {
```
BubbleCal: and here

rust/lance/src/io/exec/knn.rs
```diff
@@ -327,6 +331,358 @@ impl ExecutionPlan for KNNVectorDistanceExec {
...
+            .then_with(|| self.row_index.cmp(&other.row_index))
+    }
+}
+
+/// [ExecutionPlan] that computes flat KNN for a batch of query vectors.
+///
+/// This node consumes the input once, computes distances from each input vector
+/// to each query vector, and keeps the top-k rows independently for every query.
+#[derive(Debug)]
+pub struct KNNBatchVectorDistanceExec {
```
BubbleCal: we don't need to add a new type of `Exec`, maybe just modify `KNNVectorDistanceExec`?

BubbleCal (changes requested):

rust/lance/src/io/exec/knn.rs
```diff
@@ -69,6 +71,8 @@ use super::utils::{
     SelectionVectorToPrefilter,
 };
 
+pub const QUERY_INDEX_COL: &str = "_query_index";
```
BubbleCal: needs to be `query_index`, the reason is that we already have batch query in lancedb which uses `query_index` as the column name, it may be not a good idea but we are not going to break compatibility

rust/lance/src/io/exec/knn.rs
```diff
@@ -191,24 +265,155 @@ impl KNNVectorDistanceExec {
...
+            output_schema,
+            column,
+            query,
+            query_count,
+            k,
+            distance_type,
+        } = config;
+        let query_dim = query.len() / query_count;
+        let mut heaps = (0..query_count)
+            .map(|_| BinaryHeap::<BatchKnnCandidate>::with_capacity(k + 1))
```
BubbleCal: `cap=k` seems enough

rust/lance/src/io/exec/knn.rs
```diff
@@ -191,24 +265,155 @@ impl KNNVectorDistanceExec {
...
+                        continue;
+                    }
+                    let distance = distances.value(row_index);
+                    if !distance.is_finite() {
+                        continue;
+                    };
+                    let row_id = row_ids
+                        .as_ref()
+                        .map(|row_ids| row_ids.value(row_index))
+                        .unwrap_or(fallback_row_id + row_index as u64);
```
BubbleCal: I don't think this would happen

rust/lance/benches/vector_index.rs
```diff
@@ -124,6 +124,123 @@ fn bench_ivf_pq_index(c: &mut Criterion) {
     );
 }
 
+fn bench_batch_flat_knn(c: &mut Criterion) {
```
BubbleCal: can we port this to be in Python?

rust/lance/src/dataset/scanner.rs
```diff
@@ -1447,15 +1450,8 @@ impl Scanner {
         let (vector_type, element_type) = get_vector_type(self.dataset.schema(), column)?;
         let dim = get_vector_dim(self.dataset.schema(), column)?;
 
-        let q = match q.data_type() {
+        let (q, query_count) = match q.data_type() {
             DataType::List(_) | DataType::FixedSizeList(_, _) => {
-                if !matches!(vector_type, DataType::List(_)) {
-                    return Err(Error::invalid_input(format!(
-                        "Query is multivector but column {}({})is not multivector",
```
BubbleCal, line 1456: Can you explain more how this distinguishes between multivector query and query batch?

LeoReeYang: Batch-vs-multivector is distinguished by the vector column type: list-like q + List column means one multivector query; list-like q + FixedSizeList column means a batch of single-vector queries.

python/python/benchmarks/test_search.py
```diff
@@ -173,6 +222,57 @@ def test_knn_search(test_dataset, benchmark):
     assert result.num_rows > 0
 
 
+@pytest.mark.benchmark(group="batch_flat_knn")
+@pytest.mark.parametrize("mode", ["separate", "batch"])
+def test_batch_flat_knn(data_dir: Path, benchmark, mode: str):
+    dim = get_benchmark_env("LANCE_BATCH_KNN_DIM", BATCH_FLAT_KNN_DIM)
```
BubbleCal: just make these parameters, no need to introduce env vars here

rust/lance/benches/vector_index.rs
LeoReeYang: it has been ported to: python/python/benchmarks/test_search.py:227

rust/lance/src/io/exec/knn.rs
LeoReeYang: `_rowid` fallback was removed and batch mode now requires `_rowid`

rust/lance/src/dataset/scanner.rs
LeoReeYang: batch-vs-multivector is decided by vector column type with comments added in Scanner::nearest lines 1467-1475

BubbleCal (changes requested):
1. distance_range param is lost if it's a batch query
2. this forces the query to be executed by flat KNN even there's an index, we still need to use the index if there is one (just query the index for each query vector).

plz add tests for verifying they are really fixed

BubbleCal:
if the query is with:
- `fast_search=True`
- batch query
- no index

it's expected to return an empty result, but the schema should still contain `query_index` column, but now it doesn't

python/python/lance/dataset.py
```diff
@@ -1099,6 +976,12 @@ def scanner(
                     "distance_range": (0.0, 1.0),
                 }
 
+            ``q`` may also be a 2-D array-like value for fixed-size vector columns.
+            In that case Lance runs a flat batch KNN query, returns up to ``k`` rows
+            for each query vector, and adds ``query_index`` to identify the source
+            query for each result row. Indexed/ANN batch search is not used in this
+            first implementation.
```
BubbleCal: this comments look not correct

python/python/lance/dataset.py
```diff
@@ -7139,7 +7011,10 @@ def _build_vector_search_query(
     column: str
         The name of the vector column to search.
     q: QueryVectorLike
-        The query vector.
+        The query vector. For fixed-size vector columns, this may be a 2-D
+        array-like batch of query vectors. Batch queries run flat KNN, apply
```
BubbleCal: and this

BubbleCal (changes requested):

python/python/lance/dataset.py
LeoReeYang: - outdated comments
- there will be a `query_index` for batch

python/python/lance/dataset.py
LeoReeYang: outdated comments were committed with old batch implementation but not being updated by later commits, btw indexed batch & distance_range has been consistent with comments like above

rust/lance/src/io/exec/knn.rs
```diff
@@ -191,24 +293,163 @@ impl KNNVectorDistanceExec {
...
+                    }
+                }
+            }
+        }
+
+        let mut results = heaps
+            .into_iter()
+            .flat_map(BinaryHeap::into_vec)
+            .collect::<Vec<_>>();
+        results.sort_by(|left, right| {
```
BubbleCal, line 425: I think we need to rethink how to implement this, now the results will be truncated because of `SortExec` has `limit=k`.

Say for batch query with 2 vectors, and `k=10`, this would return 20 rows, but `SortExec` will keep only 10 results then we will lose the rest results

rust/lance/src/io/exec/knn.rs
```diff
@@ -141,23 +145,64 @@ pub struct KNNVectorDistanceExec {
...
+    upper_bound: Option<f32>,
+    distance_type: DistanceType,
+}
+
 impl DisplayAs for KNNVectorDistanceExec {
     fn fmt_as(&self, t: DisplayFormatType, f: &mut std::fmt::Formatter) -> std::fmt::Result {
         match t {
             DisplayFormatType::Default | DisplayFormatType::Verbose => {
-                write!(f, "KNNVectorDistance: metric={}", self.distance_type,)
+                if self.query_count > 1 {
```
BubbleCal: I think all checks `self.query_count > 1` need to be replaced by `self.is_batch_query()` which should check by query shape not query count.

Say if the query is a list of vectors but with only 1 vector, it's still a batch query, or the behavior will be hard to predict

BubbleCal (changes requested):

rust/lance/src/io/exec/knn.rs
LeoReeYang: introducing `Scanner::is_batch_nearest` & skipping `SortExec` on batch flat path, per-query top-k will be handled by `KNNVectorDistanceExec::execute_batch`

rust/lance/src/io/exec/knn.rs
LeoReeYang: 1. change batch query judgement
2. `Scanner::is_batch_nearest` + `KNNVectorDistanceExec::is_batch` replaces `self.query_count > 1`
3. indexed batch still leverage existing single query ANN path with `is_batch_nearest=false`

LeoReeYang:
**Test dedup only** — no production or indexed-batch behavior changes.

- **Rust**: 7 batch KNN tests → 4 via `batch_knn_two_queries()` and `assert_batch_matches_single_queries()`.
  - `test_batch_knn_flat`: flat plan, m×k rows, `query_index`, parity vs single queries, single-vector `(1, dim)` batch shape, no global `SortExec TopK(k)`.
  - `test_batch_knn_flat_respects_distance_range`: shared helper + `distance_range`.
  - `test_batch_knn_indexed`: indexed plan + `distance_range` parity (merged former split tests).
- **Python**:  
  - `@pytest.mark.parametrize` on flat batch test (2-query / 1-query).
  -  removed redundant single-vector test.
  -  `_assert_batch_matches_single_queries` checks `distance_range` only when provided.

Net: fewer overlapping assertions, same coverage.

rust/lance/src/io/exec/knn.rs
BubbleCal: I don't think this is fixed, plz add a test to verify it

rust/lance/src/dataset/scanner.rs
```diff
@@ -1906,6 +1957,11 @@ impl Scanner {
                 let vector_expr = expressions::col(DIST_COL, current_schema)?;
                 output_expr.push((vector_expr, DIST_COL.to_string()));
             }
+            if self.is_batch_nearest && output_expr.iter().all(|(_, name)| name != QUERY_INDEX_COL)
```
BubbleCal: I think the query_index column shouldn't be added into autoproject_scoring_columns, it's a little bit confusing

BubbleCal (changes requested):

BubbleCal:
@claude review

BubbleCal:
@codex review

rust/lance/src/io/exec/knn.rs
```diff
@@ -173,15 +220,74 @@ impl KNNVectorDistanceExec {
...
         // FlatExec appends a distance column to the input schema. The input
         // may already have a distance column (possibly in the wrong position), so
         // we need to remove it before adding a new one.
-        if output_schema.column_with_name(DIST_COL).is_some() {
-            output_schema = output_schema.without_column(DIST_COL);
+        if input_schema.column_with_name(DIST_COL).is_some() {
+            input_schema = input_schema.without_column(DIST_COL);
+        }
+        if input_schema.column_with_name(QUERY_INDEX_COL).is_some() {
+            input_schema = input_schema.without_column(QUERY_INDEX_COL);
```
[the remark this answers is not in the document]
BubbleCal: @LeoReeYang Plz add a test that the dataset contains a column called `query_index` and does vector search with selecting `query_index` column, assert it would return an error

BubbleCal:
@codex review

BubbleCal:
@codex review

LeoReeYang:
@BubbleCal Updated based on feedback. PTAL when you have time.

BubbleCal:
@codex review

BubbleCal:
@codex review

BubbleCal (approved):

PR #6829, opened 2026-05-18
fix(python): pass batch size through fragment pandas export

BubbleCal:
## Bug Fix

- What is the bug?
  - `LanceFragment.to_pandas(...)` did not accept or forward `batch_size`, even though fragment scans support it and dataset/scanner pandas materialization can preserve scan parameters.
- What issues or incorrect behavior does the bug cause?
  - Callers could not use `Fragment.to_pandas(...)` as the same lower-level materialization path as dataset/scanner when they needed explicit batch sizing.
- How does this PR fix the problem?
  - Adds `batch_size` to `LanceFragment.to_pandas(...)` and forwards it to `fragment.scanner(...)`.
  - Extends pandas materialization tests for dataset/scanner/fragment no-blob kwargs parity and blob scan-parameter behavior.

## Validation

- `uv sync --extra tests --extra dev`
- `uv run pytest python/tests/test_blob.py -k pandas`
- `uv run ruff check python/tests/test_blob.py python/lance/fragment.py`
- `git diff --check -- python/python/lance/fragment.py python/python/tests/test_blob.py`

### DIFF
--- a/python/python/lance/fragment.py
+++ b/python/python/lance/fragment.py
@@ -655,6 +655,7 @@ class LanceFragment(pa.dataset.Fragment):
         filter: Optional[Union[str, pa.compute.Expression]] = None,
         limit: Optional[int] = None,
         offset: Optional[int] = None,
+        batch_size: Optional[int] = None,
         with_row_id: bool = False,
         with_row_address: bool = False,
         blob_mode: str = "lazy",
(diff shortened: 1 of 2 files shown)


BubbleCal:
@claude review

Xuanwo (approved):

PR #6831, opened 2026-05-18
feat(index): wire aggregate pushdown into the physical optimizer

westonpace:
## Summary

Turns on the aggregate-pushdown scaffolding from the previous PR (`AggregateIndexSearchExec` + the plan-time parser) by adding the physical optimizer rule that actually invokes them. After this PR, a `SELECT COUNT(*) FROM t` (with or without an indexed filter) is answered from index metadata + the deletion mask + the optional scalar-index prefilter — no column scan.

Conservative on purpose: every condition that's not provably safe leaves the existing scan path untouched.

**Depends on:** the foundation PR (`feat-aggregate-pushdown`). This PR is a thin layer on top.

### `AggregateIndexPushdown` — `rust/lance/src/io/exec/aggregate_index_pushdown.rs`

A `PhysicalOptimizerRule` that walks the plan top-down. When it sees:

```text
AggregateExec(Single, aggs=[COUNT(*) ...], group_by=[])
  └── FilteredReadExec { …safe shape… }
```

it rewrites to:

```text
AggregateExec(Final, aggs=[COUNT(*) ...], group_by=[])
  └── AggregateIndexSearchExec { prefilter_input = index_input }
```

The outer `AggregateExec` is dropped to `AggregateMode::Final` because `AggregateIndexSearchExec` emits one row of partial state.

### Safety envelope

The rule fires **only** when every condition below holds. Anything else leaves the scan path alone.

| Aggregate side | Required |
|----------------|----------|
| `AggregateMode` | `Single` |
| `group_by` | empty |
| Every `AggregateFunctionExpr` | `count`, not distinct, single non-null `Literal` argument (i.e. `COUNT(*)` / `COUNT(1)`) |
| Per-aggregate `FILTER (WHERE …)` | none |

| `FilteredReadExec` side | Required |
|-------------------------|----------|
| `refine_filter` | `None` (any residual would need to be applied to data) |
| `full_filter` | `None`, **or** `Some` with `index_input` also `Some` (i.e. the index fully evaluates the filter) |
| `scan_range_before_filter` / `scan_range_after_filter` | both `None` (a `LIMIT`/`OFFSET` would change the count) |
| `with_deleted_rows` | `false` |
| `fragments` (subset) | `None` |

`is_count_star` matches `count(<literal>)` with a non-null literal — covering `COUNT(*)` and `COUNT(1)` as DataFusion lowers them, without accidentally accepting `COUNT(col)` (which has different null semantics) or `COUNT(NULL)` (which is always 0).

### Optimizer registration

Added at the **front** of `get_physical_optimizer()` so the rewritten subtree is in place before generic rules (`CoalesceTake`, `SimplifyProjection`, `LimitPushdown`) see the plan.

## Test plan

All four tests drive the rule end-to-end through `Scanner::create_plan`, so the registered optimizer chain actually fires (no direct construction shortcuts).

- [x] `rule_fires_on_unfiltered_count_star` — `SELECT COUNT(*) FROM t` on a 4×10 dataset → 40, plan contains `AggregateIndexSearchExec`.
- [x] `rule_fires_when_filter_fully_indexed` — `SELECT COUNT(*) FROM t WHERE ordered < 25` (BTree-indexed) → 25, plan contains `AggregateIndexSearchExec`.
- [x] `rule_skips_when_filter_needs_refine` — `SELECT COUNT(*) FROM t WHERE unindexed > 5` (no index on `unindexed`) → 34 via the scan path, **no** `AggregateIndexSearchExec`.
- [x] `rule_skips_count_with_group_by` — `GROUP BY ordered` → rule does not fire.

Non-regression:

- [x] `cargo test -p lance --lib io::exec::aggregate_index` — 7/7 (foundation tests unchanged).
- [x] `cargo test -p lance --lib test_count` — 3/3 (existing count tests unchanged).
- [x] `cargo clippy -p lance -p lance-index --tests --benches -- -D warnings` — clean.

## Out of scope (follow-ups)

- Recognize the plan shape that doesn't go through `FilteredReadExec` (e.g. a `LanceScanExec` direct child).
- Wire `Dataset::count_rows(None)` (the public no-filter fast path) through the optimizer so it also hits `AggregateIndexSearchExec` rather than the per-fragment count helper.
- Implement `ScalarIndex::calculate_aggregate` on a real index (e.g. exact `COUNT(DISTINCT)` from a bitmap-index dictionary) so the rule can drive distinct counts.
- Broaden `is_count_star` to handle `COUNT(col)` when the column's null count is known per fragment (`MIN/MAX`-style metadata answer).


westonpace:
@claude review once

PR #6834, opened 2026-05-19
fix(mem_wal): persist composite WAL recency key in flushed L0 generations

hamersaw:
## Problem

Insert-then-update of a primary key whose update sets a new value (NULL is the easily-observed case, but the defect is value-agnostic) can be silently dropped by L0→base compaction: the post-compaction base row reverts to a prior version of the same PK.

`merge_insert` with `when_matched_update_all` does **not** supersede a matched row that is still memtable-resident (it has no base `_rowaddr`), so a single frozen memtable — and therefore a single flushed L0 generation file — can legitimately contain two live rows for the same PK (the pre-update and post-update row). This is normal append-only-LSM state. The defect is that the flushed file carried **no write-order information**, so any consumer that deduplicates by PK (e.g. L0→base compaction) had no authoritative recency key and could keep the stale row non-deterministically.

## Fix

Stamp each row of a flushed generation with the composite WAL write position `(_wal_seq, _wal_pos)` = `(wal_entry_position, within-WAL-entry position)`. Both values are already tracked per batch in `MemTable::wal_batch_mapping` and the flush is hard-gated on `all_flushed_to_wal()`, so every batch is guaranteed an entry. This is the globally-monotonic, per-write-unique order in which writes were durably accepted; a consumer that keeps the max-key row per PK selects the true latest write.

`wal_entry_position` alone is **not** sufficient: `WalAppender::append` coalesces every batch pending at flush time into one WAL entry under one position, so an insert and a later same-PK update routinely share a `wal_entry_position`. The within-entry position discriminates them.

## Scope / safety

- **Additive and read-invisible.** The two columns are internal to the L0 file layout. They are never in the user/base schema, never reported by `is_system_column`, and never projected by the read path — `build_source_scan` projects explicitly via `build_scanner_projection`, so reads never see them. Zero read-path changes.
- `write_data_file` switches from `scan_batches_reversed()` to the `StoredBatch` variant to recover each batch's `batch_position` for the mapping lookup. The historical newest-first batch order and in-batch row reversal (relied on by the LSM k-way merge and index row-position math) are preserved — the keys are per-batch constants, invariant under the reversal.

## Tests

`dataset::mem_wal::memtable::flush::tests`:
- `test_flush_composite_key_separate_wal_entries` — insert/update flushed as two WAL entries; `_wal_seq` discriminates.
- `test_flush_composite_key_same_wal_entry` — insert/update coalesced into **one** WAL entry; only `_wal_pos` discriminates. This is the precise guard a single-`u64` key would fail.
- `test_flush_composite_key_replay_deterministic` — identical replay stamps identical keys.

Existing `flush` tests (success, btree/hnsw/fts index) still pass; the FTS plan-assertion was updated to project user columns explicitly (mirroring how production reads flushed generations) rather than couple a generic FTS test to L0 internals.

`cargo test -p lance --lib dataset::mem_wal::memtable::flush`, `cargo fmt`, and `cargo clippy -p lance --lib --tests -- -D warnings` all clean.

### DIFF
--- a/rust/lance/src/dataset/mem_wal.rs
+++ b/rust/lance/src/dataset/mem_wal.rs
@@ -43,6 +43,24 @@ mod wal;
 pub mod write;
 
 pub use api::{DatasetMemWalExt, InitializeMemWalBuilder};
+
+/// Column name for the WAL entry position stamped per row into flushed L0
+/// generation files.
+///
+/// Together with [`WAL_POS_COLUMN`] this forms the composite recency key
+/// `(_wal_seq, _wal_pos)` that L0→base compaction uses to select the per-PK
+/// survivor by the authoritative WAL write order (the WAL appender packs
+/// multiple batches into one entry, so the entry position alone is monotonic
+/// but *not* per-write unique — [`WAL_POS_COLUMN`] discriminates same-entry
+/// duplicates).
+///
+/// This is internal to the L0 file layout: it is never part of the user/base
+/// schema, never reported by `is_system_column`, and never projected by the
+/// read path, so it is invisible to readers.
+pub const WAL_SEQ_COLUMN: &str = "_wal_seq";
+
+/// Column name for the within-WAL-entry position. See [`WAL_SEQ_COLUMN`].
+pub const WAL_POS_COLUMN: &str = "_wal_pos";
 pub use manifest::ShardManifestStore;
 pub use memtable::scanner::MemTableScanner;
 pub use scanner::{LsmDataSource, LsmGeneration, LsmScanner, ShardSnapshot};
--- a/rust/lance/src/dataset/mem_wal/memtable.rs
+++ b/rust/lance/src/dataset/mem_wal/memtable.rs
@@ -78,8 +78,6 @@ pub struct MemTable {
     /// Generation number (incremented on flush).
     generation: u64,
 
-    /// WAL batch mapping: batch_position -> (wal_entry_position, position within WAL entry).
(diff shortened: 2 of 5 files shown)


jackye1995:
> merge_insert with when_matched_update_all does not supersede a matched row that is still memtable-resident (it has no base _rowaddr), so a single frozen memtable — and therefore a single flushed L0 generation file — can legitimately contain two live rows for the same PK (the pre-update and post-update row). This is normal append-only-LSM state. The defect is that the flushed file carried no write-order information, so any consumer that deduplicates by PK (e.g. L0→base compaction) had no authoritative recency key and could keep the stale row non-deterministically.

I don't agree with that. We already added the ability for merge_insert to handle duplicated rows, and it will just keep the first seen rows. That is actually why the MemTables must be flushed in reversed row order, so that when doing merge-insert, the newer rows (in later position) is kept in merge-insert.

When reading, this means when there are 2 rows of the same primary key in a memTable:
- the in-memory MemTable should keep the last row
- the flushed MemTable can simply exit at the first seen row

And then we will merge row results based on generation number.

(@touch-of-grey I don't remember if we actually merged that feature tbh... I will double check tomorrow)

hamersaw:
@jackye1995 sorry I should have had this on draft :facepalm: .

I'm trying to fix an issue with KNN queries. We don't walk the l0 cache like we do in a regular scan. So if there are duplicates in l0 cache items we can find them through our KNN query and return them. I'm thinking deduping the l0 cache on write may be the correct fix? but it's not very LSM-like.

PR #3117, opened 2024-11-11
perf: optimize reading transactions in commit loop

wjones127:
* Cache transaction files
* Read transaction files with concurrency within commit loop

Closes #3057

PR #6836, opened 2026-05-19
feat: switch commit strategy to optimistic by default

Jay-ju:
### Summary

This PR changes the default commit strategy from **Pessimistic** (always rebase before commit) to **Optimistic** (skip rebase on first attempt, only rebase on conflict). Benchmarks show Optimistic outperforms Pessimistic across all operation types and concurrency levels, so there is no need for a user-facing configuration option — we simply change the behavior.

### Background

Previously, Lance always performed a "rebase" before every commit attempt — loading new transactions since the last read and running `TransactionRebase::try_new()` to reconcile conflicts. This was safe but added significant I/O overhead on every attempt, even when there was no conflict. Under high concurrency, this rebase cost became a bottleneck.

### Core Change (`rust/lance/src/io/commit.rs`)

The commit loop now skips rebase on the first attempt and only rebases on conflict (i.e., when `backoff.attempt() > 0`):

```rust
let needs_rebase = !strict_overwrite && backoff.attempt() > 0;
```

This is safe because:
- If there is no conflict, the commit succeeds without rebase (fast path).
- If there is a conflict, `rename_if_not_exists` fails, triggering a retry with rebase (slow path).
- The rebase logic is unchanged — it still correctly reconciles conflicts.

### Required Fix: Orphan Version Detection

Switching to optimistic commit exposed a pre-existing bug in the interaction between `auto_cleanup` and the commit handler:

When `auto_cleanup` deletes older manifest files, a commit based on a stale `read_version` can succeed at writing a manifest whose version is lower than the actual latest (because `rename_if_not_exists` sees the slot as empty). This creates an **orphan version** that is unreachable from the latest lineage. Subsequent commits on the correct lineage won't see the orphan's config changes (e.g., disabling auto_cleanup), and cleanup may delete the orphan version, losing those changes.

**Fix in `commit_transaction`**: After a successful optimistic commit, if the original dataset's config contains auto_cleanup keys (indicating cleanup may have freed version slots), check whether the committed version is behind the true latest. If so, roll back and retry with rebase. This check is gated on auto_cleanup to avoid I/O overhead on commits that don't involve cleanup.

**Fix in `auto_cleanup_hook`**:
1. Use the committed manifest's config rather than the original dataset's config, so that config changes made by the current transaction (e.g., disabling auto_cleanup) are respected.
2. As defense-in-depth, skip cleanup if the committed version is behind the latest.

#### Mixed Workload (20 appenders, 10 deleters, 10 updaters)

| Metric | Pessimistic | Optimistic | Delta |
|--------|------------|------------|-------|
| Total time | 27.1s | 24.6s | **-9.2%** |
| Throughput | 2.6 ops/s | 2.9 ops/s | **+12.5%** |
| Delete avg latency | 1.34s | 0.94s | **-29.9%** |
| Update avg latency | 1.22s | 0.97s | **-20.5%** |
| Append avg latency | 0.68s | 0.63s | **-7.4%** |

#### High Concurrency (30 appenders, 15 deleters, 15 updaters)

| Metric | Pessimistic | Optimistic | Delta |
|--------|------------|------------|-------|
| Total time | 42.3s | 38.1s | **-9.9%** |
| Throughput | 2.1 ops/s | 2.4 ops/s | **+14.3%** |
| Delete avg latency | 2.01s | 1.38s | **-31.3%** |
| Update avg latency | 1.85s | 1.42s | **-23.2%** |
| Append avg latency | 0.95s | 0.87s | **-8.4%** |

### Files Changed

| File | Change |
|------|--------|
| `rust/lance/src/io/commit.rs` | Skip rebase on first attempt; add orphan version detection after successful commit |
| `rust/lance/src/dataset/cleanup.rs` | Use committed manifest config in `auto_cleanup_hook`; add orphan version defense-in-depth check |
| `python/python/ci_benchmarks/benchmarks/test_concurrent_write.py` | Concurrent write benchmark |

### Backward Compatibility

- **Fully backward compatible**: The change only affects the internal commit strategy. No changes to public API signatures.
- The optimistic strategy is safe: conflicts are still detected and reconciled via rebase on retry.


wjones127:
> This is safe but adds significant I/O overhead on every attempt, even when there is no conflict. Under high concurrency, this rebase cost becomes a bottleneck.

This is a helpful finding. I originally wrote the initial behavior here: https://github.com/lance-format/lance/pull/3117

Back then, I would have thought the pessimistic strategy would be better for high contention, at the cost of being slower when there is low contention. But looks like empirically I was wrong. Thanks for running this benchmark.

TBH, I'm not sure we need to add a setting for this. I think it would be totally acceptable just change the default to optimistic, especially given "Optimistic strategy consistently outperforms Pessimistic across all operation types and concurrency levels." If the results were mixed, I could see the case for having a setting. But if it's better in all cases, I think it's fine to change the behavior. Plus this isn't a behavior that users rely upon, it's most of an optimization. So it's not a breaking change in that sense.

What do you think of just making Optimistic the default?

Jay-ju:
> > This is safe but adds significant I/O overhead on every attempt, even when there is no conflict. Under high concurrency, this rebase cost becomes a bottleneck.
> 
> This is a helpful finding. I originally wrote the initial behavior here: #3117
> 
> Back then, I would have thought the pessimistic strategy would be better for high contention, at the cost of being slower when there is low contention. But looks like empirically I was wrong. Thanks for running this benchmark.
> 
> TBH, I'm not sure we need to add a setting for this. I think it would be totally acceptable just change the default to optimistic, especially given "Optimistic strategy consistently outperforms Pessimistic across all operation types and concurrency levels." If the results were mixed, I could see the case for having a setting. But if it's better in all cases, I think it's fine to change the behavior. Plus this isn't a behavior that users rely upon, it's most of an optimization. So it's not a breaking change in that sense.
> 
> What do you think of just making Optimistic the default?

@wjones127 Thanks for the context! I agree — since Optimistic outperforms across all scenarios, making it the default is the right call. I have updated the PR already.

For context, the motivation came from LLM agent workflows where LanceDB serves as the memory module: each conversation turn triggers an individual write, creating high write concurrency that made the rebase overhead very visible.

One question while on this topic — under extreme write concurrency (e.g. many agents writing simultaneously), even the optimistic strategy still faces repeated conflict retries. I've been thinking about whether a WAL/buffer layer on top of Lance could help absorb writes and batch-flush them, similar to LSM-tree approaches. Do you have any thoughts on whether this is a direction worth exploring, or if there are other approaches you'd recommend for high-concurrency scenarios?

wjones127:
> One question while on this topic — under extreme write concurrency (e.g. many agents writing simultaneously), even the optimistic strategy still faces repeated conflict retries. I've been thinking about whether a WAL/buffer layer on top of Lance could help absorb writes and batch-flush them, similar to LSM-tree approaches. Do you have any thoughts on whether this is a direction worth exploring, or if there are other approaches you'd recommend for high-concurrency scenarios?

Yeah, that's essentially the work that's going on in https://github.com/lance-format/lance/issues/3985

rust/lance/src/io/commit.rs
```diff
@@ -677,22 +678,31 @@ async fn migrate_indices(dataset: &Dataset, indices: &mut [IndexMetadata]) -> Re
...
+                    )
+                    .await?;
+                index.fragment_bitmap = Some(idx.calculate_included_frags().await?);
+            } else {
+                log::debug!(
+                    "Skipping fragment_bitmap recalculation for index {} (uuid: {}) because it does not exist on disk. \
+                     This likely means the index was remapped during this commit and the uuid was changed.",
+                    index.name,
+                    index.uuid
+                );
```
wjones127, line 705: issue(blocking): this seems like a drive-by bug fix. Could we save this for a different PR and make sure we have a issue describing it?

rust/lance/src/dataset/cleanup.rs
```diff
@@ -1034,7 +1034,34 @@ pub async fn auto_cleanup_hook(
...
+                } else {
+                    log::info!(
+                        "auto_cleanup skipped: committed manifest (v{}) has auto_cleanup config but latest manifest (v{}) does not. \
+                         This likely means auto_cleanup was disabled by a concurrent commit or the committed manifest inherited \
+                         stale config from an outdated dataset snapshot under Optimistic commit strategy.",
+                        manifest.version,
+                        latest.version
+                    );
+                    return Ok(None);
+                }
```
wjones127: issue(blocking): this also seems like a drive-by change that could use it's own issue.

wjones127 (commented):
> since Optimistic outperforms across all scenarios, making it the default is the right call. I have updated the PR already.

Actually, what do you think about going a step further, and just eliminating the option? Just change the behavior directly to use optimistic always.

Jay-ju:
Thanks for the thorough review! Let me address the two blocking items and the suggestion to remove the CommitStrategy option.

**Regarding the two "drive-by fixes":**

These are not drive-by fixes — they are direct consequences of the Optimistic strategy and are required for correctness. When the first commit attempt skips rebase, the manifest inherits stale writer_version and config from an outdated snapshot via Manifest::new_from_previous. Both fixes address correctness issues that only manifest under Optimistic commit:

1. **disk_indices check**: Without rebase, must_recalculate_fragment_bitmap() returns true due to stale writer_version, causing migrate_indices to attempt opening an index with a remapped uuid that doesn't exist on disk yet. This was caught by test_v0_8_14_invalid_index_fragment_bitmap failing under Optimistic mode.

2. **auto_cleanup_hook validation**: Without rebase, the committed manifest can inherit stale auto_cleanup config from an outdated snapshot, triggering cleanup that should have been disabled by a concurrent commit. This was caught by test_enable_disable_auto_cleanup failing under Optimistic mode.

Both fixes are prerequisites — without them, the Optimistic strategy breaks existing tests. That said, I'm happy to create separate issues to document these edge cases if you'd like, while keeping the fixes in this PR.

Jay-ju:
**Regarding removing the CommitStrategy option:**

I agree — since Optimistic outperforms Pessimistic across all scenarios, there's no practical reason to keep the option.

However, in my local tests, pessimistic locks deliver better performance during the first execution of preheating some S3 links, while optimistic locks generally yield superior results after preheating is completed.

 I'll remove the CommitStrategy enum entirely and inline the Optimistic logic directly.

rust/lance/src/io/commit.rs
```diff
@@ -957,21 +966,22 @@ pub(crate) async fn commit_transaction(
...
+        let needs_rebase = !strict_overwrite
+            && (backoff.attempt() > 0
+                || matches!(
+                    &transaction.operation,
+                    Operation::UpdateConfig { .. }
+                        | Operation::Overwrite {
+                            config_upsert_values: Some(_),
+                            ..
+                        }
+                ));
```
wjones127: issue(blocking): It doesn't make sense to me why this should depend on operation. If it's safe to skip rebase for some operations, but not others, that seems a like a dangerous API that will have lots of bugs. I'd rather we get it so there's no need to gate on operation at all, and just make `let needs_rebase = backoff.attempt() > 0`.

rust/lance/src/io/commit.rs
```diff
@@ -957,21 +966,22 @@ pub(crate) async fn commit_transaction(
     let mut current_transaction_file = String::new();
 
     while backoff.attempt() < num_attempts {
-        // We are pessimistic here and assume there may be other transactions
-        // we need to check for. We could be optimistic here and blindly
-        // attempt to commit, giving faster performance for sequence writes and
-        // slower performance for concurrent writes. But that makes the fast path
-        // faster and the slow path slower, which makes performance less predictable
-        // for users. So we always check for other transactions.
-        // We skip this for strict overwrites, because strict overwrites can't be rebased.
```
wjones127, line 966: suggestion: could we replace this comment with one that gives a justification for the optimistic by default behavior. It could read similar to the old one, but say that benchmarks showed that optimistic was better in all cases. Any maybe explain a little why.

wjones127 (commented):
I'm still not sure of these changes. It seems like we are changing more code than we should.

Also, make sure to update the PR description to reflect the nature of the changes now.

Jay-ju:
@wjones127 Regarding your feedback that the `auto_cleanup_hook` and `migrate_indices` changes were "drive-by changes" — I initially agreed and reverted them, but CI revealed that these changes are actually **required** for the optimistic strategy to work correctly.

**The CI failure (`test_enable_disable_auto_cleanup`)** proved that switching to optimistic commit exposes a pre-existing bug: when `auto_cleanup` deletes older manifest files, a commit based on a stale `read_version` can succeed at writing a manifest whose version is lower than the actual latest (because `rename_if_not_exists` sees the slot as empty). This creates an **orphan version** that is unreachable from the latest lineage, causing config changes like `disable_auto_cleanup()` to be silently lost.

The specific scenario:
1. Dataset has auto_cleanup enabled, cleanup deletes old manifest files
2. User calls `disable_auto_cleanup()` on a stale dataset object (read_version=2)
3. Optimistic commit skips rebase, target_version=3, and since v3's manifest file was already deleted by cleanup, `rename_if_not_exists` succeeds → orphan version 3 created
4. Actual latest is version 6; subsequent commits inherit v6's config (still has auto_cleanup), ignoring v3's config change
5. Cleanup eventually deletes v3, losing the `disable_auto_cleanup` config change entirely

This bug **does not manifest** under the pessimistic strategy because rebase always resolves to the correct latest version before committing.

**The fixes are:**
1. `commit_transaction`: After a successful optimistic commit, detect orphan versions by checking if the committed version is behind the true latest (gated on auto_cleanup config to avoid I/O overhead otherwise). Roll back and retry with rebase if detected.
2. `auto_cleanup_hook`: Use the committed manifest's config (not the original dataset's config) so config changes from the current transaction are respected. Also skip cleanup if the committed version is behind the latest (defense-in-depth).

I've updated the PR description to reflect the current state of changes. These fixes are not optional — without them, the optimistic strategy breaks `disable_auto_cleanup` and potentially other config operations when auto_cleanup is active.

rust/lance/src/dataset/write/commit.rs
```diff
@@ -552,17 +549,18 @@ mod tests {
...
-            // 2. Write (conditional put) the manifest
-            // (the version hint is only written on non-lexically-ordered stores)
-            assert_io_eq!(io_stats, write_iops, 2, "write txn + manifest, i = {}", i);
+            assert_io_gt!(
+                io_stats,
+                write_iops,
+                1,
+                "at least txn + manifest, i = {}",
+                i
+            );
```
wjones127, line 563: issue(blocking): this loosening of the assertion defeats the purpose of the test. At a minimum, we should do `assert_io_lt!()`. But I think we could still do `assert_io_eq!()`. I believe a rebase should only be needed if `i > 0`, so you can change the expectation based on that.

rust/lance/src/io/commit.rs
```diff
@@ -1054,6 +1061,46 @@ pub(crate) async fn commit_transaction(
 
         match result {
             Ok(manifest_location) => {
+                // When auto_cleanup has previously deleted older manifest files,
+                // a commit based on a stale read_version can succeed at writing a
+                // manifest whose version is lower than the actual latest (because
+                // rename_if_not_exists sees the slot as empty). This creates an
```
wjones127, line 1067: issue(blocking): you are right this is a legitimate bug that is created when you enable the optimistic commit. However, I don't think it is limited to auto_cleanup. Even with manual cleanup this can still happen. So I don't think it makes sense to gate on the presence of auto-cleanup.

In general, we say the cleanup interval should be longer than the longest running write, so as long as the write started at the most recent read version, it should be good.

I wonder if we can enable the pessimistic commit when we detect that the manifest it's based on is very old (like more than an hour or something like that). Not sure if we have that info though, we might just have the read version. If that's the case, we might want to pull the timestamp through.

wjones127:
> 1. **disk_indices check**: Without rebase, must_recalculate_fragment_bitmap() returns true due to stale writer_version, causing migrate_indices to attempt opening an index with a remapped uuid that doesn't exist on disk yet.

Could you explain this more? The manifest that has been committed and read should not reference any files that don't exist. Can you explain how that would happen?

wjones127 (changes requested):
I think I understand the cleanup bug better. I think we need to figure out a better solution to that before this is viable.

Jay-ju:
@wjones127  

> > 1. **disk_indices check**: Without rebase, must_recalculate_fragment_bitmap() returns true due to stale writer_version, causing migrate_indices to attempt opening an index with a remapped uuid that doesn't exist on disk yet.
> 
> Could you explain this more? The manifest that has been committed and read should not reference any files that don't exist. Can you explain how that would happen?


`open_generic_index(uuid)` internally calls `self.load_index(uuid)`, which resolves the UUID **from the dataset's manifest index list** (not from disk file paths):

```rust
// index/api.rs
async fn load_index(&self, uuid: &str) -> Result<Option<IndexMetadata>> {
    self.load_indices().await.map(|indices| {
        indices.iter().find(|idx| idx.uuid.to_string() == uuid).cloned()
    })
}
```

### Pessimistic vs Optimistic path comparison

```
commit_transaction
│
├─ Pessimistic (always rebase first)
│   │
│   ├─ ① rebase → dataset updated to latest version
│   ├─ ② build_manifest(Rewrite)
│   │     └─ handle_rewrite_indices: index.uuid = uuid_new
│   ├─ ③ migrate_indices(&dataset_latest, indices)
│   │     ├─ dataset.manifest.writer_version = modern
│   │     └─ must_recalculate_fragment_bitmap → false ✓
│   └─ ④ skip open_generic_index entirely → commit succeeds
│
└─ Optimistic (skip rebase on 1st attempt)
    │
    ├─ ① skip rebase → dataset remains stale
    ├─ ② build_manifest(Rewrite)
    │     └─ handle_rewrite_indices: index.uuid = uuid_new
    ├─ ③ migrate_indices(&dataset_stale, indices)
    │     ├─ dataset.manifest.writer_version = old/None
    │     └─ must_recalculate_fragment_bitmap → true ✗
    ├─ ④ open_generic_index(uuid_new)
    │     └─ load_index(uuid_new) searches stale manifest
    │         → only has uuid_old → returns None
    └─ ⑤ Error: "Index with id {uuid_new} does not exist"
```

### Why the manifest doesn't reference non-existent files

It doesn't — the issue is timing. `migrate_indices` runs **before** commit, using the stale dataset as context. At this point:

- `indices` (output of `build_manifest`) already has `uuid_new` from the remap
- But `dataset` (used to open the index) still has only `uuid_old` in its manifest

The fix simply checks whether the UUID exists in the dataset's known indices before attempting to open it. If it was remapped during this transaction, we skip recalculation — `handle_rewrite_indices` already computed the correct `fragment_bitmap`.

wjones127:
@Jay-ju do you plan on finishing this? It honestly seems like optimistic commits create a lot of problems to solve. I think it would be reasonable to close, but if you think you can solve those issues, I'm still happy to review the changes.

wjones127:
Closing this for now. Feel free to re-open if you want to keep iterating on this.

PR #6848, opened 2026-05-19
fix(mem_wal): allow append-only tables without primary keys

### LINKED ISSUE #6846: Initializing MemWAL should not need primary key enforcement
For append-only table users should still be able to set sharded write. cc @touch-of-grey

touch-of-grey:
## Summary
- allow MemWAL initialization on append-only tables without unenforced primary key metadata
- keep bucket sharding constrained to the primary key when a primary key exists, but allow no-PK tables to bucket by a non-nested column
- update Rust, Python, and Java tests plus MemWAL docs

Fixes #6846

## Tests
- cargo fmt --all
- cargo test -p lance test_initialize_mem_wal_bucket_sharding
- uv run make build
- uv run pytest python/tests/test_mem_wal.py::test_initialize_mem_wal_bucket_sharding_without_primary_key
- git diff --check

Java tests were not run locally because this machine has no Java runtime installed.

### DIFF
--- a/docs/pyproject.toml
+++ b/docs/pyproject.toml
@@ -3,7 +3,7 @@ name = "lance-docs"
 version = "0.1.0"
 description = "Documentation for Lance - Modern columnar data format for ML and LLMs"
 readme = "README.md"
-requires-python = ">=3.9"
+requires-python = ">=3.10,<3.11"
 dependencies = [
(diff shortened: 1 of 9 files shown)


jackye1995 (approved):
thanks for the fix

Issue #4888, opened 2025-10-02
Combine mini-block repetition indexes into a single buffer


westonpace:
The repetition index for mini-blocks is _very small_ (typically less than 64 bytes per 8MiB disk page).  At initialization time we often have to issue many small reads to read the repetition index.  In practice this isn't quite as bad as it seems since we default to 1Mi rows per fragment and so most columns will never have more than one disk page.  Still, we do have to perform one read per column.  It might be worth it to either store the repetition index in the fragment metadata or in some shared global buffer to reduce the number of IOPS that must be performed to initialize the search cache.

PR #6851, opened 2026-05-19
perf(encoding): inline mini-block chunk metadata into page layout proto

westonpace:
The per-chunk "Words" metadata buffer (a few KB per page) is now also
embedded in the MiniBlockLayout protobuf via a new optional field
`inline_chunk_meta`. ColumnMetadata is read in a single coalesced I/O at
file open, so new readers can populate the chunk metadata from the proto
and skip the per-column metadata fetch in MiniBlockScheduler::initialize.
When the page has no dictionary and no repetition index, initialize
issues zero I/O — addressing the per-column read on the cold
search-cache path called out in #4888.

The writer continues to emit the metadata buffer at
buffer_offsets_and_sizes[0], so older readers (which ignore the unknown
proto field) keep decoding correctly. No on-disk format version bump
required.

### DIFF
--- a/protos/encodings_v2_1.proto
+++ b/protos/encodings_v2_1.proto
@@ -115,6 +115,20 @@ message MiniBlockLayout {
 
   // Since Lance 2.2, miniblocks have larger chunk sizes (>= 64KB)
   bool has_large_chunk = 10;
+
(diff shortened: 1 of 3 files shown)


westonpace:
@claude review once

PR #6853, opened 2026-05-20
feat: expose granular trace event targets

beinan:
## Summary
- Add granular tracing targets for Lance event categories under `lance::events::...`.
- Preserve the original tracing target in Python log passthrough so `LANCE_LOG` can filter individual event types directly.
- Document the new event targets and add Python regression coverage for target-specific log filtering.

## Testing
- `PATH=/Users/beinan/.rustup/toolchains/1.94.0-aarch64-apple-darwin/bin:$PATH cargo fmt --all -- --check`
- `git diff --check`
- `PATH=/Users/beinan/.rustup/toolchains/1.94.0-aarch64-apple-darwin/bin:$PATH cargo check -p lance-core -p lance-io -p lance-index -p lance -p lance-datafusion`
- `PATH=/Users/beinan/.rustup/toolchains/1.94.0-aarch64-apple-darwin/bin:$PATH cargo check --manifest-path python/Cargo.toml`
- `uv sync --python 3.12 --no-install-project`
- `.venv/bin/maturin develop --uv -v`
- `uv run --no-sync --python 3.12 pytest -q python/tests/test_log.py::test_lance_log_filters_trace_event_targets python/tests/test_tracing.py::test_tracing_callback`
- `uv run --no-sync --python 3.12 ruff format --check python/tests/test_log.py python/tests/test_tracing.py`
- `uv run --no-sync --python 3.12 ruff check python/tests/test_log.py python/tests/test_tracing.py`


rust/lance-core/src/utils/tracing.rs
```diff
@@ -58,7 +58,7 @@ impl<S: Stream> StreamTracingExt for S {
     }
 }
 
-pub const TRACE_FILE_AUDIT: &str = "lance::file_audit";
+pub const TRACE_FILE_AUDIT: &str = "lance::events::file_audit";
```
westonpace: I'm fine with the python changes but these rust level changes are breaking changes and can impact users who are already relying on rust tracing events (e.g. us 😛)  IIRC we had a discussion long ago that tracing events could be considered part of the public API and we should be wary of changing them (though I'm not sure it was written down so this is just hearsay).  Can we add the `events` discriminator at the python level?  It's also redundant at the rust level since rust distinguishes between the two already.

rust/lance-core/src/utils/tracing.rs
```diff
@@ -67,20 +67,35 @@ pub const AUDIT_TYPE_MANIFEST: &str = "manifest";
...
+pub const TRACE_DATASET_DROPPING_COLUMN: &str = "lance::events::dataset_events::dropping_column";
 pub const DATASET_DELETING_EVENT: &str = "deleting";
+pub const TRACE_DATASET_DELETING: &str = "lance::events::dataset_events::deleting";
 pub const DATASET_COMPACTING_EVENT: &str = "compacting";
+pub const TRACE_DATASET_COMPACTING: &str = "lance::events::dataset_events::compacting";
 pub const DATASET_CLEANING_EVENT: &str = "cleaning";
+pub const TRACE_DATASET_CLEANING: &str = "lance::events::dataset_events::cleaning";
 pub const DATASET_LOADING_EVENT: &str = "loading";
+pub const TRACE_DATASET_LOADING: &str = "lance::events::dataset_events::loading";
+pub const TRACE_OBJECT_STORE_THROTTLE: &str = "lance::events::object_store::throttle";
```
westonpace: I am +1 on adding new targets to events that don't currently have targets (though I would prefer we not have the `events::` part as above).

westonpace (changes requested):
I'm a fan of this change but I'm hoping we can do it without modifying existing targets (to add `events::`) because that change would be disruptive to us.

Also, semi-related, but I've been thinking we should move more (if not all) of our `log` statements to tracing events.  Having both a "log stream" and a "trace event stream" is pretty confusing.

rust/lance-core/src/utils/tracing.rs
beinan: Agreed, updated the PR so Rust tracing targets remain unchanged and the events discriminator is applied only in pylance log targets.

beinan:
For the Azure request-id concern: the final object-store error is preserved through LanceError(IO), and I added regression coverage that an Azure-style x-ms-request-id present in the object-store error text survives; response headers that object_store::RetryError does not expose would still need an upstream object_store change.

rust/lance-core/src/utils/tracing.rs
beinan: Updated the new object-store throttle target to lance::object_store::throttle without the events:: prefix.

rust/lance-io/src/object_store/throttle.rs
```diff
@@ -1448,7 +1467,13 @@ mod tests {
 
         let result = throttled.get(&path).await;
         assert!(result.is_err(), "Expected error after max retries");
-        assert!(is_throttle_error(&result.unwrap_err()));
+        let err = result.unwrap_err();
+        assert!(is_throttle_error(&err));
+
+        let lance_error = lance_core::Error::from(err);
+        let error_message = lance_error.to_string();
+        assert!(error_message.contains("x-ms-request-id"));
```
westonpace, line 1475: I'm surprised the mock store actually includes these

westonpace (approved):
Awesome, thanks!

rust/lance-io/src/object_store/throttle.rs
beinan: I think I mocked the response msg somewhere 😆

beinan:
@westonpace can we merge?

PR #6854, opened 2026-05-20
feat: add MemWAL sharding evaluator

jackye1995:
Adds an Arrow-native MemWAL sharding evaluator and exposes it through the Java API/JNI.

- Evaluates MemWAL sharding specs against Arrow RecordBatch values for bucket, identity, and unsharded fields.
- Resolves sharding source IDs through a Java-provided source-id-to-column map.
- Adds Java-facing ShardingEvaluator returning an Arrow reader for the evaluated sharding key batch.

This is needed by lance-spark to route writes using Lance's sharding semantics instead of duplicating Spark-side bucket logic.


python/src/mem_wal.rs
```diff
@@ -33,28 +34,82 @@ use uuid::Uuid;
...
+    }
+    Ok(ShardingSpec { spec_id, fields })
+}
+
+fn sharding_field_from_py(field: &Bound<'_, PyAny>) -> PyResult<ShardingField> {
+    Ok(ShardingField {
+        field_id: get_py_value(field, "field_id")?.extract::<String>()?,
+        source_ids: get_py_value(field, "source_ids")?.extract::<Vec<i32>>()?,
+        transform: optional_string(get_py_value(field, "transform")?)?,
+        expression: optional_string(get_py_value(field, "expression")?)?,
```
Xuanwo, line 69: This makes dict specs returned by `Dataset.mem_wal_index_details()` unusable with the new evaluator.

`mem_wal_index_details()` currently serializes each sharding field with `field_id`, `source_ids`, `transform`, `result_type`, and `parameters`, but it does not include `expression`. Since this parser now requires `expression` to be present, the natural flow below fails with `Missing sharding spec field 'expression'`:

```python
spec = ds.mem_wal_index_details()["sharding_specs"][0]
evaluate_sharding_spec(batch, spec, LanceSchema.from_pyarrow(batch.schema))
```

Could we either include `expression` in the dict returned by `mem_wal_index_details()` or treat a missing `expression` key as `None` here? I think adding it to `mem_wal_index_details()` is cleaner because that keeps the exported spec shape complete and round-trippable.

Xuanwo (dismissed):
I think the Python sharding spec round-trip needs one fix before this lands.

hamersaw (approved):
Looks great to me! It feels a little off to be burying the sharding (partitioning / clustering) metadata into the memwal index implementation IMO. I think there are a lot of use-cases where we use this metadata (ex. writing dataset, Spark SPJ, etc) that is not tied to mem_wal in any way.

jackye1995:
> Looks great to me! It feels a little off to be burying the sharding (partitioning / clustering) metadata into the memwal index implementation IMO. I think there are a lot of use-cases where we use this metadata (ex. writing dataset, Spark SPJ, etc) that is not tied to mem_wal in any way.

I agree, I think one related proposal is to lift this into some independent module like lance-sharding.

But on the other side, the good thing about this is user who sets this can directly write with WAL. It's just we don't handle the read path.

python/src/mem_wal.rs
jackye1995: Addressed in 4f2ddc5d6 by including `expression` in the Python `Dataset.mem_wal_index_details()` sharding field dict. I also extended `test_initialize_mem_wal_bucket_sharding` to pass the returned dict spec directly into `evaluate_sharding_spec(...)`, covering the round-trip path from the comment.

Xuanwo (approved):

PR #6855, opened 2026-05-20
feat: expose multi-base config to Python and Java write_fragments API

zhangyue19921010:
For Python
```python
fragments = write_fragments(
    pa.table({"id": range(10, 20)}),
    dataset,
    mode="append",
    target_bases=["base2"],
    base_store_params=base_store_params,
)
```

For Java
```java
  List<FragmentMetadata> fragments =
      Fragment.write()
          .allocator(allocator)
          .datasetUri(primary)
          .data(root2)
          .mode(WriteParams.WriteMode.APPEND)
          .targetBases(Arrays.asList("base2"))
          .baseStoreParams(baseStoreParams)
          .execute();
```


yanghua:
@claude review

python/python/lance/fragment.py
```diff
@@ -1173,6 +1183,8 @@ def write_fragments(
     if isinstance(dataset_uri, Path):
         dataset_uri = str(dataset_uri)
     elif isinstance(dataset_uri, LanceDataset):
+        if base_store_params is None:
+            base_store_params = dataset_uri._base_store_params
```
yanghua, line 1187: Do we need to add a fallback logic when there is a `base path` not in the `_base_store_params` so that it can fall back to the default `_storage_options`?

```
if storage_options is None:
        storage_options = dataset_uri._storage_options
```

zhangyue19921010: changed~

yanghua (approved):
+1

PR #6859, opened 2026-05-20
feat(btree): support distributed range-partitioned BTree index build

Jay-ju:
This change completes the end-to-end distributed BTree index construction (range-partitioned mode), fixes a correctness bug on the fragment-partitioned path that was exposed once K-way sort-merge was removed, and tightens transaction safety by deferring shard cleanup until after commit succeeds.

What's new
----------
- DistributedMode { Single, Fragment, Range } selects the build strategy.
- range_partitions(N) on CreateIndexBuilder triggers range-partitioned construction:
  1. Quantile sampling of column values (with dedup) to compute N-1 boundaries
  2. Per-partition independent training that emits a BTreeShardManifest
  3. Merging of all shard manifests into the final index files
- New BTreeShardManifest carries the per-shard build output. Marked #[non_exhaustive] so future fields can be added without breaking callers.
- Sample size cap raised from 100k to 1M and values are deduplicated to reduce skew from high-frequency keys.

Bug fix: pages_between with overlapping pages
---------------------------------------------
On the fragment-partitioned path each fragment is trained independently, so page key ranges may overlap across fragments. The previous largest_node_less peek-one-step search could miss pages whose min is below the lower bound but max is in range. Fixed by adding a may_overlap flag on BTreeLookup; when set, the search starts from Bound::Unbounded. Range-partitioned indices keep the existing tight search.

Transaction safety: defer cleanup until after commit ---------------------------------------------------- cleanup_shard_files was previously invoked inside build_btree_index_partitioned and merge_index_metadata, before the caller's transaction was committed. If the commit fails, the shard files are already deleted and the operation cannot be retried.

Both paths now return the MergeResult to the caller and cleanup is deferred:
- CreateIndexBuilder::execute stores a PendingShardCleanup and runs it after apply_commit returns Ok.
- merge_index_metadata returns Option<MergeResult>; the Python binding exposes a BTreeMergeResult class with .cleanup() that callers must invoke after their own commit succeeds. The Java binding keeps the legacy immediate- cleanup behavior with a TODO for follow-up design work.

API hardening
-------------
- range_partitions() asserts num_partitions >= 2 with a descriptive message.
- Non-BTree index types reject range_partitions with a clear error.
- train=false combined with range_partitions returns an error instead of silently building an empty partitioned index.
- transaction_properties() and range_partitions() now have rustdoc.
- Python create_scalar_index/create_index docstrings document the new range_partitions parameter.

Follow-up TODOs
-------------

- [x] Preserve Range partition structure in BTreeIndex::update() — #6904 
- [ ] Persist partition boundary values Boundary values are not stored in index metadata after construction. Without them, update() cannot route new data to the correct partition and must re-sample. Fix: add range_partition_boundaries to page_lookup.lance metadata.
- [ ] Reduce full-rewrite cost in optimize_indices update() reads all old index data, merges with new data, and rewrites the entire index. Consider incremental updates: build a separate index segment for new data and merge at query time.


python/python/lance/lance/__init__.pyi
```diff
@@ -408,7 +413,7 @@ class _Dataset:
         index_type: str,
         batch_readhead: Optional[int] = None,
         progress_callback: Optional[Callable[[IndexProgress], None]] = None,
-    ): ...
+    ) -> Optional[BTreeMergeResult]: ...
```
Xuanwo, line 416: merge_index_metadata is considered deprecated. Please using `merge_existing_index_segments` and `commit_existing_index_segments` instead to build index segments.

PR #6869, opened 2026-05-20
feat!: migrate bitmap to index segment based

Xuanwo:
Adds Bitmap support to the existing segment-based distributed index workflow.

Callers can now build staged Bitmap roots with `create_index_uncommitted(..., index_type="BITMAP", fragment_ids=...)`, finalize them through `create_index_segment_builder().with_index_type("BITMAP").with_segments(...).build_all()`, and publish them with `commit_existing_index_segments(...)`.

For Bitmap, `execute_uncommitted` now writes canonical `bitmap_page_lookup.lance` segment roots directly. The old public Python Bitmap shard workflow through `create_scalar_index(..., fragment_ids=...)` and `merge_index_metadata(..., "BITMAP")` is no longer exposed; callers should use the segment workflow instead.

Relates to OSS-971 and OSS-972.


python/python/lance/dataset.py
```diff
@@ -3250,7 +3262,9 @@ def create_scalar_index(
         if progress_callback is not None:
             kwargs["progress_callback"] = progress_callback
 
-        self._ds.create_index([column], index_type, name, replace, train, None, kwargs)
+        return self._ds.create_index(
+            [column], index_type, name, replace, train, None, kwargs
+        )
```
westonpace: Should you document what is being returned?

python/python/lance/dataset.py
```diff
@@ -3928,6 +3942,38 @@ def create_index_uncommitted(
...
+            or index_type.upper() in scalar_index_types
+        ):
+            if not isinstance(column, str):
+                raise NotImplementedError(
+                    "Scalar indices currently only support a single column"
+                )
+            return self.create_scalar_index(
+                column,
+                IndexConfig(index_type="bitmap", parameters={})
+                if isinstance(index_type, str) and index_type.upper() == "BITMAP"
```
westonpace: I'm a little confused.  If `index_type == "BITMAP"` you are switching to `IndexConfig` here (with no parameters) and then calling `create_scalar_index` but in `create_scalar_index` you also look for `index_type == "BITMAP"` but how would that ever be true?

rust/lance-index/src/scalar/bitmap.rs
```diff
@@ -447,7 +447,13 @@ impl Index for BitmapIndex {
...
 
     async fn calculate_included_frags(&self) -> Result<RoaringBitmap> {
-        unimplemented!()
+        let mut frag_ids = RoaringBitmap::new();
+        for key in self.index_map.keys() {
+            let bitmap = self.load_bitmap(key, None).await?;
+            frag_ids.extend(bitmap.iter().map(|(fragment_id, _)| *fragment_id));
+        }
+        frag_ids.extend(self.null_map.iter().map(|(fragment_id, _)| *fragment_id));
+        Ok(frag_ids)
```
westonpace: Why implement this method?  I thought it was only a utility method for an old migration before we had fragment bitmaps?

python/python/lance/dataset.py
```diff
@@ -3928,6 +3942,38 @@ def create_index_uncommitted(
...
+        }
+        if (
+            isinstance(index_type, IndexConfig)
+            or index_type.upper() in scalar_index_types
+        ):
+            if not isinstance(column, str):
+                raise NotImplementedError(
+                    "Scalar indices currently only support a single column"
+                )
+            return self.create_scalar_index(
```
westonpace: The method is called `create_index_uncommitted` but if you call `create_scalar_index` isn't it going to commit the index?

westonpace (commented):
I have some questions, I probably just need to get up to speed on these concepts.

Xuanwo:
Hi @westonpace, thanks for the review! I think the main source of confusion is that the current status of distributed bitmap building doesn't align with our index segments design.

For now, the bitmap has special logic around distributed index building. It includes a concept called `shard_id` and has its own pipeline. My plan is to migrate the entire pipeline to be index segment based instead.

I have rewritten the PR migrating the index segment API. This will also make this PR a breaking change one.

westonpace (approved):
Looks good.

Attaching Claude's review for reference (but I don't think any of these suggestions are all that important):

## PR #6869 Review: `feat!: migrate bitmap to index segment based`

### Overview

This PR migrates the Bitmap scalar index from an older "shard-based" distributed build workflow to the same segment-based workflow already used by vector and BTREE indices. The key changes:

- `create_index_uncommitted(..., index_type="BITMAP", fragment_ids=...)` now writes canonical `bitmap_page_lookup.lance` segment roots directly
- `merge_existing_index_segments` now supports bitmap segments
- The old `create_scalar_index(..., fragment_ids=...)` + `merge_index_metadata(...)` bitmap workflow is removed
- `BITMAP` is removed from `SupportedDistributedIndices` enum in Python

---

### Code Quality & Correctness

**`rust/lance-index/src/scalar/bitmap.rs` — `merge_bitmap_indices`**

- The `|=` on `RowAddrTreeMap` merges bitmaps per key — correct, but worth verifying `RowAddrTreeMap`'s `BitOrAssign` is semantically a union (not intersection or replacement). This is a silent correctness risk if the impl is ever changed.
- The progress reporting calls `stage_start("write_bitmap_index", Some(1), "files")` and then hardcodes `stage_progress(..., 1)` — this is slightly misleading since the actual write could touch multiple files. Minor cosmetic issue.

**`rust/lance/src/index/scalar/bitmap.rs` — `merge_segments`**

- The `..segments[0].clone()` spread at the bottom of the returned `IndexMetadata` struct is a potential footgun: it silently inherits any fields from the first segment that aren't explicitly overridden. In particular, `name` is inherited from `segments[0]` rather than being verified consistent across all segments. If callers accidentally mix segments from different indices (different names), the merged result will silently inherit the wrong name. Consider asserting that all segments share the same `name` and `fields`.

**`rust/lance/src/index.rs` — `merge_existing_index_segments`**

```rust
let all_bitmap = source_segments.iter().all(segment_has_bitmap_details);
if !all_vector && !all_inverted && !all_bitmap { ... }
```

- If `source_segments` is empty, `all_vector`, `all_inverted`, and `all_bitmap` all evaluate to `true` (vacuous truth), so the guard passes. Then the dispatch falls into the `else` (bitmap) branch. `merge_segments` does check for empty inputs and returns an error — but the error message says "No segment metadata was provided" with `Error::index(...)` rather than the more consistent `Error::invalid_input(...)`. Consider making this consistent.

**`rust/lance/src/index/create.rs` — `CreateIndexBuilder`**

- The new branch checks `self.index_type == IndexType::Bitmap && self.fragments.is_some()` but then immediately does `self.fragments.clone().ok_or_else(...)` — the `ok_or_else` is unreachable since you just checked `is_some()`. Not a bug, but a minor redundancy.

---

### Breaking Change Handling

- `BITMAP` is removed from `SupportedDistributedIndices` enum. Any Python code that checked `SupportedDistributedIndices.BITMAP` will break at runtime. This is intentional (the PR description says the old workflow is no longer exposed), but should be called out clearly in docs or a migration guide, not just the PR description. Since `SupportedDistributedIndices` appears to be a public Python enum, this is a breaking API change.
- The `merge_index_metadata` path now returns an error for `IndexType::Bitmap`. This is a clean break, but the error message is long and contains formatting that could be cleaner with a proper `Display` impl.

---

### Test Coverage

**Removed tests that aren't fully replaced:**
- `test_distributed_build_bitmap` included a full end-to-end correctness check: post-commit query assertions (`search` returns correct row counts per value). The new `test_bitmap_execute_uncommitted_writes_canonical_segment` only verifies file layout, not query correctness.
- `test_distributed_bitmap_index_build_single_fragment_shards` is removed with no replacement for single-fragment-per-segment coverage.
- The `explain_plan` and "empty result" (`category = 99`) assertions from `_assert_committed_distributed_bitmap_index` are dropped from the Python test.

**New tests are good:**
- `test_bitmap_segments_commit_and_query_as_logical_index` (Rust) is thorough: stages, merges, commits, and queries with value assertions.
- `test_bitmap_uncommitted_segments_can_be_committed_from_python` covers the Python API end-to-end but skips the empty-result edge case.

**Suggestion:** Restore the `category = 99` (empty result) assertion in the Python test — it's a meaningful null-lookup edge case for bitmap indices and was previously covered.

---

### Minor Issues

- The blank line removal after the `raise ValueError` block in `dataset.py:3285` is trivial but slightly reduces readability (the comment "Add fragment_ids..." now runs directly after the block without visual separation).
- `serde_json::json` import removal in `create.rs` is correct cleanup.

---

### Summary

The PR is well-structured and the core logic is sound. The main risks are:

1. **Correctness gap in Rust unit test** — the old test verified actual query results after commit; the replacement only checks file layout. The new Rust integration test in `scalar_logical.rs` does cover end-to-end querying, so coverage exists, but it's less targeted.
2. **`..segments[0].clone()` spread** in `bitmap.rs::merge_segments` silently inherits `name` and other fields — worth asserting cross-segment consistency.
3. **Empty-result edge case** dropped from Python tests.

Xuanwo:
Comfired that those comments are not important, let's move.

PR #6871, opened 2026-05-20
perf: use roaring's range iter to speedup mask_to_offset_ranges

westonpace:
The function `mask_to_offset_ranges` is used at scan planning time to determine which rows to read from the file.  This was a bottleneck when the mask was the result of a zonal index search because the old implementation materialized all of the offsets only to convert them back into ranges.

Luckily, roaring recently implemented a range-based iterator.  Using this we can skip the materialization step.  On my zonemap benchmark this doubles the speed of the search and, perhaps more importantly, removes a penalty I observed when the index is used even on queries that are not highly selective.

Generated with the assistance of Claude code.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4485,6 +4485,7 @@ dependencies = [
  "async-trait",
  "byteorder",
  "bytes",
+ "criterion",
  "datafusion-common",
  "datafusion-sql",
  "deepsize",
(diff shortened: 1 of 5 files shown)


wjones127 (approved):
Nice optimization!

Xuanwo (approved):
Nice change!

PR #6872, opened 2026-05-20
feat: expose index segments in Python scanner

hfutatzhanghb:
## Summary

Expose vector index segment selection through the Python scanner API. This adds an `index_segments` argument to `LanceDataset.scanner(...)` and `ScannerBuilder`, accepts `str` and `uuid.UUID` values in Python, parses them into Rust `Uuid`s in the PyO3 binding, and forwards them to the existing Rust `Scanner::with_index_segments` support.

## Why

Rust already supports restricting vector search to selected index segments. Python users creating distributed index segments need the same control from `ds.scanner(...)`, similar to the existing `fragments` argument.

## Validation

- `cargo check --manifest-path python/Cargo.toml -p pylance`
- `git diff --check --cached`

I did not complete the targeted pytest run because it triggered a local `pylance` rebuild and was stopped to avoid generating additional large `target` artifacts.


hfutatzhanghb:
@jiaoew1991 @LuciferYang @Xuanwo Hi, could you please help review this PR when free ? Thanks a lot.

Xuanwo (approved):
Thank you!

PR #6873, opened 2026-05-20
perf(fts): lazy-load PostingListReader metadata

wkalt:
To compute BM25 IDF, FtsStatsExec asks each segment "how many docs contain term X?" That question goes through InvertedPartition::load, which eagerly read two columns (length, max score) for every unique token in the partition before answering. A one-term query paid 8 * num_unique_tokens bytes per partition to get back a single LENGTH_COL row.

Modern v2 max_scores and lengths now live in tokio::sync::OnceCells. The stats path reads one LENGTH_COL row per (term, partition) via the new posting_len_for_token. The scoring path calls ensure_metadata_loaded once on first use, so warm-cache behavior is unchanged. Legacy v1 is untouched.

bm25_stats_for_terms and build_global_bm25_scorer are now async; four call sites await them. New rstest test_bm25_stats_for_terms_is_lazy (N=10/100/1000) asserts one posting-file row read per stats call.

### DIFF
--- a/rust/lance-index/src/scalar/inverted/index.rs
+++ b/rust/lance-index/src/scalar/inverted/index.rs
@@ -633,7 +633,9 @@ impl InvertedIndex {
         let scorer: &dyn Scorer = if let Some(base_scorer) = base_scorer {
             base_scorer
         } else {
-            local_scorer = self.bm25_base_scorer(tokens.as_ref(), params.as_ref()).await?;
+            local_scorer = self
+                .bm25_base_scorer(tokens.as_ref(), params.as_ref())
+                .await?;
             &local_scorer
         };
 
--- a/rust/lance-index/src/scalar/inverted/index.rs
+++ b/rust/lance-index/src/scalar/inverted/index.rs
@@ -525,13 +525,36 @@ impl InvertedIndex {
         MemBM25Scorer::new(scorer.total_tokens(), scorer.num_docs(), token_docs)
     }
 
-    pub fn bm25_stats_for_terms(&self, terms: &[String]) -> (u64, usize, Vec<usize>) {
+    pub async fn bm25_stats_for_terms(&self, terms: &[String]) -> Result<(u64, usize, Vec<usize>)> {
         let scorer = IndexBM25Scorer::new(self.partitions.iter().map(|part| part.as_ref()));
-        let token_docs = terms
+        let token_docs =
+            futures::future::try_join_all(terms.iter().map(|term| self.df_for_term(term))).await?;
+        Ok((scorer.total_tokens(), scorer.num_docs(), token_docs))
+    }
+
+    /// Sum the posting-list length for `term` across this index's partitions
+    /// via single-row reads, with partition lookups bounded by the store's
+    /// `io_parallelism()`.
+    async fn df_for_term(&self, term: &str) -> Result<usize> {
+        let io_parallelism = self.store.io_parallelism();
+        let futures = self
+            .partitions
             .iter()
-            .map(|term| scorer.num_docs_containing_token(term))
-            .collect();
-        (scorer.total_tokens(), scorer.num_docs(), token_docs)
+            .map(|part| {
+                let part = part.clone();
+                async move {
+                    match part.tokens.get(term) {
+                        Some(token_id) => part.inverted_list.posting_len_for_token(token_id).await,
+                        None => Ok(0),
+                    }
+                }
+            })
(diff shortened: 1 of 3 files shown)


rust/lance-index/src/scalar/inverted/index.rs
```diff
@@ -569,6 +594,24 @@ impl InvertedIndex {
...
+        // read `posting_len` synchronously; for v2 partitions that
+        // requires the bulk metadata to already be in memory. Force the load
+        // once up front so the hot scoring loop doesn't await per call.
+        let io_parallelism = self.store.io_parallelism();
+        let load_futures = self
+            .partitions
+            .iter()
+            .map(|part| {
+                let part = part.clone();
+                async move { part.inverted_list.ensure_metadata_loaded().await }
```
BubbleCal: Say for a cold query, we will finally still need to call `ensure_metadata_loaded` here which would load the all `lengths` and `max_scores`, so we don't actually reduce IO, is this right?

rust/lance-index/src/scalar/inverted/index.rs
```diff
@@ -569,11 +597,15 @@ impl InvertedIndex {
...
+        // The wand only consults `scorer.doc_weight`, which is metadata-free.
+        // The outer aggregation below consults `scorer.query_weight`, which
+        // hits per-token `posting_len`; building a `MemBM25Scorer` with
+        // precomputed per-term IDFs avoids the v2 bulk metadata pull.
         let local_scorer;
         let scorer: &dyn Scorer = if let Some(base_scorer) = base_scorer {
             base_scorer
         } else {
-            local_scorer = IndexBM25Scorer::new(self.partitions.iter().map(|part| part.as_ref()));
+            local_scorer = self.bm25_base_scorer(tokens.as_ref()).await?;
```
BubbleCal: It seems this doesn't handle fuzzy query case, we will get 0 score for expanded tokens

rust/lance-index/src/scalar/inverted/index.rs
```diff
@@ -1621,18 +1675,40 @@ enum PositionsLayout {
...
-            + self.max_scores.deep_size_of_children(context)
-            + self.lengths.deep_size_of_children(context)
+        match &self.metadata {
+            PostingMetadata::LegacyV1 {
+                offsets,
+                max_scores,
+            } => offsets.deep_size_of_children(context) + max_scores.deep_size_of_children(context),
+            PostingMetadata::V2 { metadata } => metadata
+                .get()
+                .map(|loaded| {
```
BubbleCal, line 1736: It looks like we would add the posting list into cache before filling the metadata, then `max_scores` and `lengths` won't be counted in the cache's stats

rust/lance-index/src/scalar/inverted/index.rs
wkalt: Right -- the second commit (e1ca4cfe8) addresses this. `posting_list()` now fetches per-token metadata inside the cache closure via `posting_metadata_for_token`, so a cold K-token query does K single-row reads instead of the bulk N-row load. `ensure_metadata_loaded` is only used by prewarm and `read_all`/`into_builder`.

rust/lance-index/src/scalar/inverted/index.rs
wkalt: Fuzzy-expanded tokens go through `load_posting_lists` which calls `posting_list()` per token -- that now loads per-token metadata inside the cache closure, so expanded tokens get correct `max_score`/`length`. The `bm25_base_scorer` path also handles fuzzy via `df_for_term` per expanded token.

rust/lance-index/src/scalar/inverted/index.rs
wkalt: Also addressed in the second commit -- `posting_metadata_for_token` is called inside the `get_or_insert_with_key` closure alongside `posting_batch`, so `max_score` and `length` are present before the entry enters the cache.

BubbleCal (approved):
Thanks for the work!
Good to have some benchmark numbers!

wkalt:
I have a test with a vocabulary of 500K words and two indexes, one with 1 segment and one with 20 segments. The data is just random gibberish. I get this result for cold reads

```
                      1 segment    20 segments    segment scaling cost
  ---------------------------------------------------------------------------
  Before (0.36.0)     3452ms       4261ms         +809ms
  After  (this PR)    3155ms       3493ms         +338ms  (58% reduction)

```

PR #6875, opened 2026-05-20
fix(java): add diagnostics for imported Arrow streams

beinan:
## Summary

Add Lance Java diagnostics around Arrow C Data stream readers and a scanner regression for the nested struct case that motivated the lance-spark fix.

## Findings

The original `VectorLoader` failure can be caused by consumers closing or mutating the reader-owned `VectorSchemaRoot` between `loadNextBatch()` calls. When the damaged vector is a `StructVector`, Arrow Java may see schema children still present while the reused vector has zero children, producing `should have as many children as in the schema: found 0 expected N`.

## Changes

- Add `LanceArrowReaders`, a small wrapper for Arrow readers imported from Lance Arrow C Data streams.
- Use the wrapper for Java APIs that call `Data.importArrayStream(...)`, including `LanceScanner`, `AsyncScanner`, SQL query output, delta output, and file reads.
- Preserve the underlying reader behavior while adding Lance context to batch-load failures, including the child-count mismatch diagnostic.
- Add a Java scanner regression that writes a six-field all-null struct and scans it across multiple batches.
- Add a unit test for the diagnostic message.

## Testing

```bash
cd java
./mvnw -Dtest=org.lance.ScannerTest#testDatasetScannerAllNullStructColumnMultipleBatches,org.lance.ipc.LanceArrowReadersTest test
```

This run passed Java checkstyle, Spotless, both targeted Java tests, the JNI build, and the rust-maven-plugin JNI test.

### DIFF
--- a/java/src/main/java/org/lance/SqlQuery.java
+++ b/java/src/main/java/org/lance/SqlQuery.java
@@ -13,9 +13,10 @@
  */
 package org.lance;
 
+import org.lance.ipc.LanceArrowReaders;
+
 import com.google.common.base.MoreObjects;
 import org.apache.arrow.c.ArrowArrayStream;
-import org.apache.arrow.c.Data;
 import org.apache.arrow.vector.ipc.ArrowReader;
 
 import java.io.IOException;
@@ -54,7 +55,8 @@ public class SqlQuery {
     try (ArrowArrayStream s = ArrowArrayStream.allocateNew(dataset.allocator())) {
       intoBatchRecords(
           dataset, sql, Optional.ofNullable(table), withRowId, withRowAddr, s.memoryAddress());
-      return Data.importArrayStream(dataset.allocator(), s);
+      return LanceArrowReaders.importArrayStream(
+          dataset.allocator(), s, "SqlQuery.intoBatchRecords");
     }
   }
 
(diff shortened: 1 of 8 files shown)


hamersaw:
> caused by consumers closing or mutating the reader-owned VectorSchemaRoot between loadNextBatch() calls.

This feels like a bug somewhere within the Spark repo, is this diagnostics wrapper meant to make debugging the fix easier?

beinan (commented):
I dont see a blocking issue in the current patch. The wrapper in `java/src/main/java/org/lance/ipc/LanceArrowReaders.java` now avoids the recursive schema path from the earlier review, and the new tests cover the nested-null batch case plus exception wrapping. Residual gap: `DiagnosticArrowReader.closeReadSource()` / `close(boolean)` are still not exercised through Arrows full base-class lifecycle, so if Arrow changes `ArrowReader` close semantics again this wrapper will need another pass.

beinan:
> > caused by consumers closing or mutating the reader-owned VectorSchemaRoot between loadNextBatch() calls.
> 
> This feels like a bug somewhere within the Spark repo, is this diagnostics wrapper meant to make debugging the fix easier?

Yes, the actual fix is in lance-spark. This wrapper adds diagnostics so that if similar issues occur in the future (from any consumer, not just Spark), the error message points directly to the root cause instead of a cryptic Arrow schema mismatch.

wjones127:
@hamersaw  @beinan We still want to finish this up?

Issue #3125, opened 2024-11-14
Composite scalar indices


westonpace:
We should support creating composite scalar indices (i.e. multiple columns)

pavanramkumar:
I would love to see composite scalar indices in lance. Is this on the roadmap in an upcoming release?

PR #6878, opened 2026-05-20
perf(merge_insert): probe every indexed key column for composite joins

### LINKED ISSUE #3124: Support multiple columns (composite key) for `merge_insert`
Example scenario:  A document is chunked into paragraphs and each paragraph is embedded and the row contains the document_id and the paragraph_id.  Later, the user recalculates the embedding for one of the documents and wants to replace the rows.

ytyky:
## Summary
- Indexed merge_insert path was hard-gated to single-column joins; multi-column
  merges fell through to a full target scan even when every key column had a
  scalar index.
- Generalize `MapIndexExec` to AND-fold per-column `IsIn` probes, and have
  `merge_insert` feed it every indexed join key.
- Per-column indices stay independent (single-column path unchanged). Composite
  scalar indices (#3125) remain a complementary follow-up; the merge_insert
  plumbing this PR adds is what such an index would plug into.

## Test plan
- [x] `cargo test -p lance --lib merge_insert` (145 tests, includes 6 new
      composite-key cases covering single-side / both-side / no index / NULL
      semantics / partial fragment coverage)
- [x] `cargo fmt --all -- --check`
- [x] Benchmark numbers to follow in a comment.

Closes #3124


ytyky:
## Benchmark numbers

Wall-clock for one `merge_insert` on a target with both join columns
indexed (`a`, `b`: Int64; BTREE on each). Source = 100 rows (50 updates
of existing keys, 50 inserts). Five iterations after one warmup run, on
this branch versus its parent commit. Bench source is committed as
`rust/lance/tests/<[composite_merge_insert_bench.rs](http://composite_merge_insert_bench.rs/)>`.

| Target rows | Pre-PR (full scan) | This PR (indexed) | Speedup |
| ---: | ---: | ---: | ---: |
| 200k | 16.5 ms | 38.0 ms | 0.43× |
| 2M   | 179.3 ms | 54.8 ms | **3.27×** |

The pre-PR path scales linearly with target row count (full scan of
the join columns); this PR is essentially flat (dominated by loading
the scalar indices). The crossover is around 400-500k rows, so very
small tables see a ~20 ms regression that an opt-out heuristic could
recover later; the design point for `merge_insert` against indexed
tables is well above that.

yanghua:
@claude review

ytyky:
@wjones127 follow-up on this PR when you have a chance — no rush. Just wanted to check whether there are any concerns or changes you’d like me to make. Thanks!

rust/lance/tests/composite_merge_insert_bench.rs
wjones127, line 1: issue(blocking): I think this file is in the wrong place. It's not a test—it doesn't make any assertions. I would either make this an example or a benchmark. Either is fine.

wjones127 (changes requested):
Actual code changes look good. I like the benchmark, but I think that file is in the wrong place.

rust/lance/tests/composite_merge_insert_bench.rs
ytyky: Moved into [benches/merge_insert.rs](https://github.com/ytyky/lance/blob/fix-merge-insert-multi-columns/rust/lance/benches/merge_insert.rs) as a composite_key_indexed criterion shape; dropped the test file

wjones127 (approved):
Thanks!

PR #6879, opened 2026-05-20
refactor: move mask code into new crate

westonpace:
Some of the mask code was in lance-core and some was in lance-index.  It was difficult to benchmark the stuff in lance-index since it's a bit dependency heavy.  This moves all the mask code into a lance-select crate.  It also adds a micro benchmark for some of the lance ops (I plan on optimizing these in a follow-on PR)

This PR was developed with assistance from Claude Code.


westonpace:
Baseline benchmark numbers:

```

Baseline numbers (criterion --measurement-time 1, single-threaded,
median of 10 samples):

  Op   Variant pair              N=10K   N=100K  N=1M    N=10M
  not  Exact / AtMost / AtLeast  ~13 ns  ~12 ns  ~10 ns  ~10 ns
  and  Exact_Exact               251 ns  319 ns  1.23 µs 8.28 µs
  and  AtMost_AtMost             257 ns  305 ns  1.11 µs 8.39 µs
  and  AtLeast_AtLeast           234 ns  322 ns  1.11 µs 8.24 µs
  and  Exact_AtMost              245 ns  320 ns  1.06 µs 8.44 µs
  and  Exact_AtLeast (degenerate) 182 ns 239 ns  510 ns  2.17 µs
  and  AtMost_AtLeast (degen.)   178 ns  245 ns  528 ns  2.23 µs
  or   (mirror of `and`)         ~210 ns ~285 ns ~880 ns ~7.0 µs
```

Cargo.toml
```diff
@@ -70,6 +71,7 @@ lance-namespace = { version = "=7.0.0-beta.17", path = "./rust/lance-namespace"
 lance-namespace-impls = { version = "=7.0.0-beta.17", path = "./rust/lance-namespace-impls" }
 lance-namespace-datafusion = { version = "=7.0.0-beta.9", path = "./rust/lance-namespace-datafusion" }
 lance-namespace-reqwest-client = "0.7.5"
+lance-select = { version = "=7.0.0-beta.17", path = "./rust/lance-select" }
```
wjones127: issue(blocking): we also need to update the bumpfile, right?

https://github.com/lance-format/lance/blob/main/.bumpversion.toml

wjones127 (commented):
I think there's one important change missing. Also should make a note that the first time we publish there might be extra steps we need to take. Can handle that next week with 7.0.0 release.

Cargo.toml
westonpace: Updated.

wjones127 (approved):

PR #6880, opened 2026-05-21
fix(mem_wal): dedupe duplicate primary keys in LSM point lookup

touch-of-grey:
Split from #6856 — point-lookup portion.

A primary key written multiple times into one active memtable used to leak through to the user as distinct rows: `FilterExec + LIMIT 1` over an insert-ordered scan returned the *oldest* match among duplicates.

The active arm now runs `WithinSourceDedupExec(KeepMaxFreshness)`, which collapses by PK and keeps the freshest row. Flushed and base arms still rely on `LIMIT 1` under the reverse-write / forward-write conventions.

Part of splitting #6856 into focused PRs. Co-authored with @jackye1995.

### DIFF
--- a/rust/lance/src/dataset/mem_wal/scanner/exec/within_source_dedup.rs
+++ b/rust/lance/src/dataset/mem_wal/scanner/exec/within_source_dedup.rs
@@ -1,0 +1,432 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright The Lance Authors
+
+//! WithinSourceDedupExec - Deduplicates rows with the same primary key from a
+//! single LSM source, keeping the newest insert.
+//!
+//! In MemWAL/LSM mode the same primary key can be written multiple times into
+//! the same memtable. The active memtable stores rows in insert order (larger
+//! `_rowaddr` = newer), while flushed memtables are reverse-written so that
+//! within a flushed file the smallest `_rowid` is the newest insert (see
+//! `memtable/flush.rs:152` and `hnsw/storage.rs:307`). Point lookup uses this
+//! node to collapse such duplicates *within a single source* so that the
+//! downstream `CoalesceFirstExec` / `LIMIT` sees at most one row per primary
+//! key per source.
+
+use std::any::Any;
+use std::collections::HashMap;
+use std::fmt;
+use std::pin::Pin;
+use std::sync::Arc;
+use std::task::{Context, Poll};
+
+use arrow_array::{Array, RecordBatch, UInt64Array};
+use arrow_schema::SchemaRef;
+use datafusion::error::Result as DFResult;
+use datafusion::execution::TaskContext;
+use datafusion::physical_expr::{EquivalenceProperties, Partitioning};
+use datafusion::physical_plan::{
+    DisplayAs, DisplayFormatType, ExecutionPlan, ExecutionPlanProperties, PlanProperties,
+    SendableRecordBatchStream,
+};
+use futures::{Stream, StreamExt, ready};
+
+use super::pk::{compute_pk_hash, resolve_pk_indices};
+
+/// Among rows that share a primary key, which row-address extreme identifies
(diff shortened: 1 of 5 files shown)


rust/lance/src/dataset/mem_wal/scanner/exec/within_source_dedup.rs
```diff
@@ -0,0 +1,483 @@
...
+// SPDX-FileCopyrightText: Copyright The Lance Authors
+
+//! WithinSourceDedupExec - Deduplicates rows with the same primary key from a
+//! single LSM source, keeping the newest insert.
+//!
+//! In MemWAL/LSM mode the same primary key can be written multiple times into
+//! the same memtable. The active memtable stores rows in insert order (larger
+//! `_rowaddr` = newer), while flushed memtables are reverse-written so that
+//! within a flushed file the smallest `_rowid` is the newest insert (see
+//! `memtable/flush.rs:152` and `hnsw/storage.rs:307`). Vector search and
```
jackye1995: this is no longer used by vector search

rust/lance/src/dataset/mem_wal/scanner/exec/within_source_dedup.rs
```diff
@@ -0,0 +1,483 @@
...
+    SendableRecordBatchStream,
+};
+use futures::{Stream, StreamExt, ready};
+
+/// Which freshness value wins among rows that share a primary key.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum DedupDirection {
+    /// Keep the row with the largest freshness value (active memtable: larger
+    /// `_rowaddr` = inserted later).
+    KeepMaxFreshness,
```
jackye1995: freshness is not the right word, because we always want to keep the max freshness. it's just the row address to keep is max or min.

rust/lance/src/dataset/mem_wal/scanner/exec/within_source_dedup.rs
```diff
@@ -0,0 +1,483 @@
...
+                    ))
+                })
+        })
+        .collect()
+}
+
+/// Hash a row's primary key. Kept in sync with the variants supported by
+/// [`super::LsmGlobalPkDedupExec`] and [`super::BloomFilterGuardExec`] so
+/// a single PK produces the same hash regardless of which exec consumes it.
+fn compute_pk_hash(batch: &RecordBatch, pk_indices: &[usize], row_idx: usize) -> u64 {
```
jackye1995: should probably refactor to use the same function across all invocations

rust/lance/src/dataset/mem_wal/scanner/exec/within_source_dedup.rs
touch-of-grey: Done — rewrote the module doc to describe only the point-lookup usage; the vector-search / `LsmGlobalPkDedupExec` references are gone (c8c5f17bc).

rust/lance/src/dataset/mem_wal/scanner/exec/within_source_dedup.rs
touch-of-grey: Renamed throughout: `DedupDirection::KeepMax/MinFreshness` → `KeepMax/MinRowAddr`, and `freshness_column` → `row_addr_column`. The enum doc now states the kept row is always the freshest — only the row address (`_rowaddr`/`_rowid`) extreme used to find it differs by source (c8c5f17bc).

rust/lance/src/dataset/mem_wal/scanner/exec/within_source_dedup.rs
touch-of-grey: Extracted `resolve_pk_indices` + `compute_pk_hash` into a shared `exec::pk` module. Both `WithinSourceDedupExec` and `LsmGlobalPkDedupExec` now call it; the duplicated copies are removed (c8c5f17bc).

jackye1995 (approved):

PR #6881, opened 2026-05-21
fix(mem_wal): exact PK dedup for LSM vector search

touch-of-grey:
Split from #6856 — vector-search portion.

A primary key written multiple times into one memtable, or into both a memtable and an older generation, used to leak through as distinct rows: HNSW indexes every insert as its own graph node, so KNN could return both V1 and V2 of the same PK from a single source.

Each per-source KNN now runs through `LsmSourceTagExec`, which appends `(_memtable_gen, _freshness)`. A single `LsmGlobalPkDedupExec` over the union keeps the row with the largest tuple per PK — newer generations win, ties fall to the normalized within-source order. This replaces the bloom-based `FilterStaleExec` design and is exact (no false-positive recall loss, no top-k under-fill).

After global dedup + sort + top-k, a `TakeExec` materializes any user-projected columns not in the per-source KNN output by fetching from the base dataset via `_rowid`. `plan_search()` also accepts `refine_factor` so callers can enable base-table refine. Exposed in the Python and Java bindings.

Removes `FilterStaleExec` and `GenerationBloomFilter`.

Part of splitting #6856 into focused PRs. Co-authored with @jackye1995.


jackye1995 (approved):
thanks for working on this!

PR #6890, opened 2026-05-21
feat: add early termination for compaction plan with max_compaction_bytes option

### LINKED ISSUE #6039: Bounded Compaction Planner To Limit the amount of data processed during a single compaction

Jay-ju:
## Summary

Add budget-based early termination to `DefaultCompactionPlanner` to prevent OOM when planning compaction on datasets with many fragments (e.g., hundreds of thousands).

Closes: #6039

## Problem

When a dataset has hundreds of thousands of fragments, `plan_compaction` collects metrics for **all** fragments before producing the plan. This leads to:

1. **OOM risk**: All fragment metadata + metrics are held in memory simultaneously
2. **Excessive I/O**: Each fragment requires a read of its deletion file
3. **Large serialized plans**: 10K fragments → ~2.3MB JSON; 300K fragments → ~70MB JSON

The existing `max_source_fragments` option was a **post-hoc truncation** — it collected all metrics first, then truncated the output. This did not reduce planning time or memory.

Benchmark data (10K fragments, no deletions):

| max_source_fragments | plan_time_ms | plan_json_size |
|---------------------|-------------|----------------|
| None (unlimited) | 51 | 2.3MB |
| 100 | **56** | 408B |
| 500 | **56** | 408B |

Plan time barely changed because all metrics were still collected.

## Solution

Refactor `max_source_fragments` from post-hoc truncation to **in-loop early termination**, and add a new `max_compaction_bytes` option. The planner now tracks `total_candidate_fragments` and `total_candidate_bytes` during the metrics collection loop and breaks out as soon as either budget is exceeded.

Key changes:
- **`max_source_fragments`**: Now terminates metrics collection early (was post-hoc truncation)
- **`max_compaction_bytes`**: New option to limit by cumulative fragment byte size
- **`exceeds_budget()`**: Helper method checking both limits during the planning loop
- Preserves existing parallel I/O (`.buffered(io_parallelism())`) — unlike PR #6095 which used serial I/O

## Design Rationale

This approach follows [hamersaw's review feedback on PR #6095](https://github.com/lance-format/lance/pull/6095#discussion_r2891191168): extending `CompactionOptions` rather than adding a new `BoundedCompactionPlanner` type. Users configure limits directly without needing to choose a planner implementation.

### Rust
- `CompactionOptions`: Add `max_compaction_bytes: Option<usize>` field
- `DefaultCompactionPlanner::plan()`: Replace post-hoc truncation with in-loop early termination
- `DefaultCompactionPlanner::exceeds_budget()`: New helper method
- `CompactionOptions::apply_dataset_config()`: Support `lance.compaction.max_compaction_bytes`
- Tests: 3 functional tests + 3 benchmark tests

### Python
- `CompactionOptions` TypedDict: Add `max_compaction_bytes` field with docs
- PyO3 binding: Handle `max_compaction_bytes` key

## Usage

```python
# Limit by fragment count
dataset.optimize.compact_files(max_source_fragments=1000)

# Limit by total bytes
dataset.optimize.compact_files(max_compaction_bytes=10 * 1024**3)  # 10GB

# Both limits combined
dataset.optimize.compact_files(
    max_source_fragments=1000,
    max_compaction_bytes=10 * 1024**3,
)
```

```rust
let options = CompactionOptions {
    max_source_fragments: Some(1000),
    max_compaction_bytes: Some(10 * 1024 * 1024 * 1024),
    ..Default::default()
};
```

## Comparison with PR #6095

| Dimension | PR #6095 | This PR |
|-----------|----------|---------|
| Architecture | New `BoundedCompactionPlanner` type | Extend existing `DefaultCompactionPlanner` |
| I/O pattern | Serial (one-at-a-time) | Parallel (preserved) |
| User API | `planner="bounded"` + limits | Direct `max_*` options |
| Maintainer feedback | Design not accepted | Follows maintainer preference |


Jay-ju:
@claude review

Jay-ju:
Hi @hamersaw. Fragment planning consumes much time in large data scenarios. I have discussed with @zhangyue19921010 . Based on the discussion of #6039  , we revised the original logic of full planning followed by trimming to on-demand planning. Planning will stop once reaching the threshold.

Could you take a look when you have time?

yanghua:
@claude review

rust/lance/src/dataset/optimize.rs
```diff
@@ -665,17 +728,33 @@ impl CompactionPlanner for DefaultCompactionPlanner {
                     });
                 }
                 (Some(candidacy), Some(bin)) => {
-                    // We cannot mix "indexed" and "non-indexed" fragments and so we only consider
-                    // the existing bin if it contains the same indices
                     if bin.indices == indices {
```
yanghua, line 748: If we match this branch, the early termination would never happen.

rust/lance/src/dataset/optimize.rs
```diff
@@ -4816,4 +4902,335 @@ mod tests {
...
+
+        let total_frags: usize = plan.tasks.iter().map(|t| t.fragments.len()).sum();
+        assert_eq!(
+            total_frags, 50,
+            "without budget, all fragments should be planned"
+        );
+    }
+
+    #[tokio::test]
+    #[ignore]
```
yanghua: Many ignored test cases, used for benchmark?

yanghua (commented):
Left two comments.

rust/lance/src/dataset/optimize.rs
Jay-ju: @yanghua thank you for your comment, i have fixed it. please check it again

rust/lance/src/dataset/optimize.rs
Jay-ju: em, yes, it should be removed here

rust/lance/src/dataset/optimize.rs
```diff
@@ -204,11 +204,21 @@ pub struct CompactionOptions {
     /// Defaults to 16MB (16 * 1024 * 1024).
     pub binary_copy_read_batch_bytes: Option<usize>,
     /// Maximum number of source fragments to compact in a single run. When set,
-    /// tasks are included in the plan until adding the next task would exceed
-    /// this limit. This allows for incremental compaction (e.g., compact 20
-    /// fragments at a time).
+    /// the planner stops collecting fragment metrics once enough non-noop
```
wjones127: nitpick: "non-noop" is a bit confusing term. "candidate fragments" communicates the ideal well enough on it's own.
```suggestion
    /// the planner stops collecting fragment metrics once enough
```

rust/lance/src/dataset/optimize.rs
```diff
@@ -665,17 +728,53 @@ impl CompactionPlanner for DefaultCompactionPlanner {
...
+                            let running_fragments =
+                                effective_candidate_fragments + bin.fragments.len();
+                            let running_bytes = effective_candidate_bytes
+                                + bin
+                                    .fragments
+                                    .iter()
+                                    .flat_map(|f| f.files.iter())
+                                    .filter_map(|f| f.file_size_bytes.get())
+                                    .map(|sz| sz.get() as usize)
+                                    .sum::<usize>();
```
wjones127: suggestion: this pattern appears three times. Could you extract into a function? Or maybe it should be a method on a bin?

rust/lance/src/dataset/optimize.rs
Jay-ju: ✅ Fixed in 125d8adf3. Replaced "non-noop candidate" with "candidate" in both `max_source_fragments` and `max_compaction_bytes` docs.

rust/lance/src/index/append.rs
```diff
@@ -981,6 +981,7 @@ mod tests {
             .unwrap()
             .nearest("vector", array.value(0).as_primitive::<Float32Type>(), 2)
             .unwrap()
+            .nprobes(2)
```
yanghua: unnecessary change?

rust/lance/src/dataset/optimize.rs
```diff
@@ -995,14 +1082,22 @@ impl CandidateBin {
...
         }
-        // If there's only one fragment, it's a noop if it's not CompactItself
         if self.fragments.len() == 1 {
             matches!(self.candidacy[0], CompactionCandidacy::CompactWithNeighbors)
         } else {
             false
         }
     }
 
+    fn total_bytes(&self) -> usize {
```
yanghua: This is inefficient, especially since it contains two loops for each fragment.

What about maintaining it in `CandidateBin `, for example like this:

```
struct CandidateBin {
    pub fragments: Vec<Fragment>,
    pub pos_range: Range<usize>,
    pub candidacy: Vec<CompactionCandidacy>,
    pub row_counts: Vec<usize>,
    pub indices: Vec<usize>,
    pub known_bytes: u64,
}

impl CandidateBin {
    fn fragment_known_bytes(frag: &Fragment) -> u64 {
        frag.files.iter()
            .filter_map(|f| f.file_size_bytes.get())
            .map(|sz| sz.get())
            .sum()
    }
}
```

rust/lance/src/dataset/optimize.rs
Jay-ju: Updated in 8f5565923 with two improvements beyond just extracting the method:

**1. `finalize_bin()` — eliminates the duplicated pattern**

The "complete bin → accumulate budget → check exceeds → break" block appeared identically in two match arms. Extracting it into a method removes the duplication and makes the budget-checking logic easier to follow.

**2. Incremental byte tracking — O(N) → O(1) per iteration**

The original code called `bin.total_bytes()` after every `bin.fragments.push()`, which re-traverses all fragments in the bin. Since `effective_candidate_bytes` already tracks the cumulative total of completed bins, we only need the newly added fragment's bytes. This reduces the per-iteration cost from O(bin size) to O(1), which matters when bins contain many fragments.

rust/lance/src/dataset/optimize.rs
```diff
@@ -4816,4 +4911,128 @@ mod tests {
...
+        .await
+        .unwrap();
+        let limited_frags: usize = limited_plan.tasks.iter().map(|t| t.fragments.len()).sum();
+
+        assert!(
+            limited_frags < unlimited_frags,
+            "limited plan should have fewer fragments than unlimited ({} vs {})",
+            limited_frags,
+            unlimited_frags
+        );
```
yanghua, line 5035: `max_compaction_bytes = 1` is set too small. When combined with the scenario where all fragments use CompactWithNeighbors and belong to the same bin, the newly added fragment will be popped right after the first push. The bin with only one remaining fragment will be discarded in the final `filter(|bin| !bin.is_noop())` phase, resulting in `limited_frags = 0`. While `0 < unlimited_frags` is definitely true, this test verifies an empty plan instead of the intended functionality of stopping collection early based on byte count as required by the PR.

You can add a set of more meaningful assertions. For instance, set target_rows_per_fragment: 250 and configure `max_compaction_bytes` to a value slightly larger than the size of a single fragment. Add assertions for `0 < limited_frags < unlimited_frags`, and also verify that at least one valid task is generated.

rust/lance/src/dataset/optimize.rs
```diff
@@ -632,8 +694,13 @@ impl CompactionPlanner for DefaultCompactionPlanner {
 
         let mut candidate_bins: Vec<CandidateBin> = Vec::new();
         let mut current_bin: Option<CandidateBin> = None;
+        let mut effective_candidate_fragments: usize = 0;
+        let mut effective_candidate_bytes: usize = 0;
```
yanghua: The type of this field should be `u64`.

rust/lance/src/dataset/optimize.rs
```diff
@@ -632,8 +694,13 @@ impl CompactionPlanner for DefaultCompactionPlanner {
 
         let mut candidate_bins: Vec<CandidateBin> = Vec::new();
         let mut current_bin: Option<CandidateBin> = None;
+        let mut effective_candidate_fragments: usize = 0;
```
yanghua, line 712: The type of this field should be `u64`.

rust/lance/src/dataset/optimize.rs
```diff
@@ -598,13 +618,55 @@ impl DefaultCompactionPlanner {
...
+        false
+    }
+
+    fn apply_budget_limits(&self, tasks: Vec<TaskData>) -> Vec<TaskData> {
+        let mut total_frags = 0usize;
+        let mut total_bytes = 0usize;
+        let mut result = Vec::new();
+        for task in tasks {
+            let task_frags = task.fragments.len();
+            let task_bytes: usize = task
```
yanghua: The type of this field should be `u64`.

rust/lance/src/dataset/optimize.rs
```diff
@@ -598,13 +618,55 @@ impl DefaultCompactionPlanner {
...
+            && candidate_bytes > max_bytes
+        {
+            return true;
+        }
+        false
+    }
+
+    fn apply_budget_limits(&self, tasks: Vec<TaskData>) -> Vec<TaskData> {
+        let mut total_frags = 0usize;
+        let mut total_bytes = 0usize;
```
yanghua: The type of this field should be `u64`.

yanghua (commented):
Left some comments/

Jay-ju (commented):
Addressed all comments in d5088649f. Summary:

1. **`u64` types**: Changed `max_compaction_bytes` and all internal byte tracking to `u64`. `max_source_fragments` stays `usize` since fragment count is semantically correct as usize.

2. **`known_bytes` on CandidateBin**: Added `known_bytes: u64` field with incremental updates, removed `total_bytes()`. Also added `fragment_known_bytes()` helper.

3. **Test fix**: Replaced `max_compaction_bytes: Some(1)` with a meaningful budget derived from the first task's actual bytes, plus assertions for `0 < limited_frags < unlimited_frags`.

4. **append.rs**: Moved `.nprobes(2)` to a separate PR (#6953) since it fixes a flaky test unrelated to compaction planning.

Jay-ju (commented):
Good catch. Fixed in d5088649f — `max_compaction_bytes` is now `Option<u64>`, and all internal byte accumulators use `u64`. Removed the `as usize` casts. `max_source_fragments` stays `usize` since fragment count is semantically correct as usize.

Jay-ju (commented):
Done in d5088649f. Added `known_bytes: u64` to `CandidateBin` with incremental updates on push/pop, and a `fragment_known_bytes()` helper. Removed `total_bytes()` entirely.

Jay-ju (commented):
Fixed in d5088649f. Now uses `target_rows_per_fragment: 250` (produces multiple tasks), sets `max_compaction_bytes` to the first task's actual byte size, and asserts `0 < limited_frags < unlimited_frags`.

Jay-ju (commented):
You're right it's unrelated to compaction. It was added to fix a flaky test (`test_query_delta_indices` — IVF with 2 partitions but default nprobes=1 could miss the target partition). I've moved it to a separate PR: #6953.

Jay-ju (commented):
Fixed in d5088649f — changed to `u64`.

rust/lance/src/dataset/optimize.rs
Jay-ju: Done in d5088649f. Added `known_bytes: u64` to `CandidateBin` with incremental updates on push/pop, plus a `fragment_known_bytes()` helper. Removed `total_bytes()` entirely.

rust/lance/src/index/append.rs
Jay-ju: You're right it's unrelated to compaction. It was added to fix a flaky test (`test_query_delta_indices` — IVF with 2 partitions but default nprobes=1 could miss the target partition). Moved it to a separate PR: #6953.

.jieli/node_id
```diff
@@ -0,0 +1 @@
+562ff51c-8b0d-5409-292d-9d200eccf6bf
```
yanghua: unnecessary file?

.jieli/state.json
```diff
@@ -0,0 +1,11 @@
+{
```
yanghua: ditto

rust/lance/src/dataset/optimize.rs
```diff
@@ -642,59 +724,87 @@ impl CompactionPlanner for DefaultCompactionPlanner {
...
+                        let new_fragment_bytes = CandidateBin::fragment_known_bytes(&fragment);
                         bin.fragments.push(fragment);
                         bin.pos_range.end += 1;
                         bin.candidacy.push(candidacy);
                         bin.row_counts.push(metrics.num_rows());
+                        bin.known_bytes += new_fragment_bytes;
+                        if has_budget && !bin.is_noop() {
+                            let running_fragments =
+                                effective_candidate_fragments + bin.fragments.len();
+                            let running_bytes = effective_candidate_bytes + new_fragment_bytes;
```
yanghua, line 758: In the in-loop budget check computes running_bytes as effective_candidate_bytes + new_fragment_bytes, but effective_candidate_bytes is only ever updated by finalize_bin(), i.e. it does not include any bytes accumulated in the currently-open bin

rust/lance/src/dataset/optimize.rs
```diff
@@ -4816,4 +4921,144 @@ mod tests {
...
+            "expected at least 1 candidate fragment, got 0"
+        );
+        assert!(
+            limited_frags < unlimited_frags,
+            "limited plan should have fewer fragments than unlimited"
+        );
+    }
+
+    #[tokio::test]
+    async fn test_compaction_plan_early_termination_max_compaction_bytes() {
```
yanghua, line 4982: test_compaction_plan_early_termination_max_compaction_bytes and test_compaction_plan_early_termination_max_source_fragments only assert on the final plan shape (limited_frags < unlimited_frags, limited_frags <= 20, etc.). Those assertions are equally satisfied by the post-hoc apply_budget_limits path; they cannot distinguish "loop stopped early" from "loop ran to the end and we trimmed afterwards".

rust/lance/src/dataset/optimize.rs
```diff
@@ -642,59 +724,87 @@ impl CompactionPlanner for DefaultCompactionPlanner {
...
                         pos_range: i..(i + 1),
                         candidacy: vec![candidacy],
                         row_counts: vec![metrics.num_rows()],
                         indices,
+                        known_bytes,
                     });
                 }
                 (Some(candidacy), Some(bin)) => {
-                    // We cannot mix "indexed" and "non-indexed" fragments and so we only consider
-                    // the existing bin if it contains the same indices
```
yanghua, line 669: Do we need to keep this comment?

rust/lance/src/dataset/optimize.rs
```diff
@@ -642,59 +724,87 @@ impl CompactionPlanner for DefaultCompactionPlanner {
             {
                 Some(CompactionCandidacy::CompactItself)
             } else if metrics.physical_rows < self.options.target_rows_per_fragment {
-                // Only want to compact if their are neighbors to compact such that
-                // we can get a larger fragment.
```
yanghua, line 646: Do we need to keep this remark?

python/python/lance/optimize.py
```diff
@@ -83,9 +83,18 @@ class CompactionOptions(TypedDict):
...
-    allowing for incremental compaction (e.g., compact 20 fragments at a
-    time). Fragments are processed oldest first.
+    Maximum number of source fragments to compact in a single run. The
+    planner stops collecting fragment metrics once this many fragments
+    have been selected as compaction candidates, providing early
+    termination to limit planning time and memory for datasets with
+    many fragments.
+    (default: None, no limit)
+    """
+    max_compaction_bytes: Optional[int]
```
yanghua, line 93: IMO, the config is for the source dataset or the target dataset is not very clear. We may add `max_{xxx}_compaction_bytes` to align to `max_source_fragments`...

yanghua (commented):
Left some new comments.

Jay-ju:
test

PR #6891, opened 2026-05-21
feat: create materialized view API

rpgreen:
* API and namespace stubs for new create_materialized_view API
* New REST route for materialized view refresh


jackye1995:
I checked this against `.claude/skills/lance-namespace-enterprise/SKILL.md`. The PR looks incomplete in three specific places:

1. Java propagation is missing. `createMaterializedView` is not added to `java/src/main/java/org/lance/namespace/RestNamespace.java`, `java/src/main/java/org/lance/namespace/DirectoryNamespace.java`, or `java/lance-jni/src/namespace.rs`. `java/pom.xml` also still pins `lance-namespace-core` / `lance-namespace-apache-client` to `0.7.5`, so the new Java model types would not be available.

2. The Python dependency minimum should be bumped. `python/python/lance/namespace.py` now imports `CreateMaterializedViewRequest` / `CreateMaterializedViewResponse`, but `python/pyproject.toml` still allows `lance-namespace>=0.7.5,<0.8`. Environments with 0.7.5 or 0.7.6 already installed can fail importing `lance.namespace`; this should require the first version containing those models, apparently `0.7.7`.

3. The RestAdapter create route is missing. The PR adds `/v1/materialized_view/:id/refresh`, but not `/v1/materialized_view/:id/create` or a handler that forwards to `backend.create_materialized_view`. A `RestNamespace` client talking to Lance's in-process `RestAdapter` would 404 on create.

jackye1995 (approved):
looks good to me!

PR #6899, opened 2026-05-21
fix(mem_wal): suppress stale LSM vector-search reads via block-list post-filter

hamersaw:
## What

Fixes a **stale read** in LSM MemWAL vector search: when a primary key is updated and its *fresh* row falls out of its own source's top-k, the superseded copy from an older generation could win the cross-source dedup and be returned.

## Background

`LsmGlobalPkDedupExec` (introduced in #6881) is exact only over the candidates each source *surfaces*. If a PK's fresh version is pushed out of its source's top-k by closer rows, the dedup never sees it and cannot suppress the stale copy from an older generation — so the stale row is served. Repro: `test_vector_search_stale_read_when_fresh_falls_out_of_top_k`.

## Approach

Make staleness a **per-source PK-hash post-filter** applied to each source's KNN *before* the cross-source union, so a stale row never reaches the merge.

- **Membership.** Each generation's membership is an `Arc<HashSet<u64>>` of PK hashes (`compute_pk_hash` — the same hash the dedup nodes use). Built once per generation; flushed generations' sets are cached on `FlushedMemTableCache` and scanned **streaming** (one batch resident at a time, no full PK-column buffer).
- **Per-source block set.** `compute_source_block_lists` gives each source the membership sets of the generations newer than it — `NEWER(G)` — as a `Vec<Arc<HashSet<u64>>>`, **referenced, never merged into a per-query union**. Generations are **per-shard**, so the map is keyed `(shard_id, generation)` and a source is only superseded by strictly-newer generations of its *own* shard (the base table, shardless and oldest, is blocked by every generation). The base table is **not scanned** — it's filtered by hashing only its KNN candidates.
- **Execution.** `PkHashFilterExec` drops any candidate whose PK hash is in any of its source's blocked sets. This handles only **cross-generation** supersession: a PK in a newer generation makes every copy of it stale, so dropping by hash needs no row address. **Within-generation** duplicates (same PK twice in one generation) share a hash and are left to the existing global dedup's `(generation, freshness)` tiebreaker.

## Configuration

`LsmVectorSearchPlanner::plan_search` exposes two knobs (Rust + Python + Java bindings):

- **`overfetch_factor: f64`** — a single knob controlling *both* stale filtering and over-fetch:
  - `< 1.0` (e.g. `0.0`): stale filtering **off** (no block-list / `PkHashFilterExec`; the global dedup still runs).
  - `== 1.0` (**default**): filtering **on**, no over-fetch — a source with superseded rows fetches exactly `k` and may return fewer than `k` live rows.
  - `> 1.0`: filtering **on**, over-fetch `ceil(k * factor)` so dropping the stale rows still leaves `k`.

  There is intentionally no separate on/off flag — over-fetch is only meaningful while filtering, so the factor encodes both.
- **`refine_base_table: bool`** (replaces the old `refine_factor: Option<u32>`) — re-rank the base arm's approximate index distances to exact (factor 1). Auto-enabled whenever stale filtering runs (over-fetch widens the base's candidate pool, which must be exact before the merge).

## External API

`LsmScanner::contains_pks(&RecordBatch) -> Vec<bool>` — test which primary keys have been (re)written in the WAL fresh tier (active + frozen memtables + flushed generations), built like any query (construct the scanner, then call). Hashing is internal, so callers never reproduce `compute_pk_hash`.

## Scope / caveats

- **Post-filter, not a true prefilter.** It relies on over-fetch to backfill dropped rows and does **not** guarantee `k` live results in the adversarial case (more superseded rows near the query than the over-fetch covers); `PkHashFilterExec` logs a per-source warning when this happens. Promoting the same membership into the KNN as a true prefilter (the index traverses until `k` rows pass, removing the over-fetch) is the headline follow-up.
- **Within-generation top-k eviction** (a generation holds both a stale and a fresh copy of a PK, and the fresh one is evicted) is *not* pre-filtered — it shares a hash, so it can't be disambiguated by membership. It's the same bug class as the cross-source case, currently relying on the global dedup; closing it would need flush-time dedup (so flushed generations are internally deduped like the base table). The base table is assumed internally deduped.

## Tests

`test_vector_search_stale_read_when_fresh_falls_out_of_top_k` passes (with positive + `overfetch_factor=0.0` toggle-off assertions); plus over-fetch backfill, cross-flushed and composite-PK stale reads, same-L0 newest-wins, per-shard block-list isolation, and `PkHashFilterExec` / membership unit tests (incl. null and composite PKs). `cargo fmt`, `cargo clippy -p lance --tests -- -D warnings`, and the `lance` / `lance-jni` / `pylance` crate checks are clean.


jackye1995 (approved):
looks good to me!

PR #6901, opened 2026-05-21
fix(mem_wal): build secondary indexes when flushing active memtable

hamersaw:
## Summary

The background memtable flush handler (`MemTableFlushHandler::flush_memtable`) called `flush`, which persists the data file and bloom filter but **builds no secondary indexes**. The handler never even received the shard's index configs.

Two query-side consequences over flushed generations:

- **Point lookups** (`point_lookup.rs`) run a `filter_expr` scan. Lance can route that through a scalar index — but none existed, so lookups fell back to a full scan (perf regression).
- **Vector search** (`vector_search.rs`) uses index-only `fast_search()`. Its doc comment assumes "each flushed memtable has its own vector index built during flush", which was false — so flushed vector rows were invisible to KNN. This is a **correctness** bug, not just perf.

## Changes

- Thread the shard's `index_configs` into `MemTableFlushHandler`.
- Call `flush_with_indexes` when any indexes are configured so each flushed generation carries the same secondary indexes as the active memtable; keep plain `flush` when none are configured to avoid a needless dataset open.
- Box the `flush_with_indexes` future to keep the flush async block under the type-layout recursion limit.

## Testing

- New `test_flushed_generation_is_indexed`: writes through the real `ShardWriter` path, forces a flush, and asserts the flushed generation (a) carries the BTree index and (b) resolves `id = 5` via `ScalarIndexQuery` rather than a scan. HNSW/FTS flush and `fast_search` over an indexed flushed generation are already covered by existing tests.
- `cargo test -p lance --lib dataset::mem_wal::` — 316 passed, 0 failed.
- `cargo fmt --all` and `cargo clippy -p lance --tests -- -D warnings` — clean.

### DIFF
--- a/rust/lance/src/dataset/mem_wal/write.rs
+++ b/rust/lance/src/dataset/mem_wal/write.rs
@@ -1339,7 +1339,9 @@ impl ShardWriter {
         )?;
 
         // Background MemTable flush handler — frozen memtable to Lance file.
-        let memtable_handler = MemTableFlushHandler::new(state.clone(), flusher, epoch, stats);
+        // It rebuilds the same secondary indexes on each flushed generation.
+        let memtable_handler =
+            MemTableFlushHandler::new(state.clone(), flusher, epoch, index_configs.to_vec(), stats);
         task_executor.add_handler(
             "memtable_flusher".to_string(),
             Box::new(memtable_handler),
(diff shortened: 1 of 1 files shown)


jackye1995 (approved):

jackye1995:
cc @touch-of-grey you probably want to rerun the backpressure benchmark after this

PR #6903, opened 2026-05-22
fix(python)!: derive index type from details instead of opening the index

wjones127:
BREAKING CHANGE: `describe_indices()` now reports nested and special-character field names as full field paths (e.g. `meta.lang`, `` `user-id` ``) instead of just the leaf name.

`list_indices()` called the `load_indices()` binding, which opened each index to derive its type and reported `"Unknown"` on any failure.

`list_indices()` is now a thin wrapper over `describe_indices()`, which derives the type from index details without opening the index:

- `describe_indices()` no longer errors on indices without index details; it returns a best-effort degraded entry instead.
- When index details exist but no plugin is registered for the type URL, the type is derived from the type URL rather than `"Unknown"`.
- `field_names` now uses the full field path, so nested fields are reported as dotted paths instead of just the leaf name.
- `IndexSegmentDescription` gains a `base_id` field.
- The unused `load_indices()` Python binding is removed.

The `list_indices()` return type hint was incorrect (`List[Index]` — the method has always returned dicts). It now returns a typed `IndexInformation` `TypedDict`, so callers get key and value types instead of an opaque dict.

## Testing

- Rust: `cargo test -p lance --lib index::`, `lance-index` registry tests — new tests cover the degraded entry and the type-URL fallback.
- Python: `test_scalar_index.py`, `test_column_names.py`, `test_vector_index.py`, `test_optimize.py` — including new `list_indices()` characterization tests committed before the rework, plus index-without-details and legacy-vector cases.
- Lint: `cargo fmt`, `cargo clippy` (lance, lance-index, pylance), `ruff`, `pyright`.


python/python/lance/dataset.py
```diff
@@ -956,22 +956,34 @@ def checkout_latest(self):
...
+    def list_indices(self) -> List[IndexInformation]:
         """
         Returns index information for all indices in the dataset.
 
-        This method is deprecated as it requires loading the statistics for each index
-        which can be a very expensive operation.  Instead use describe_indices() to
-        list index information and index_statistics() to get the statistics for
-        individual indexes of interest.
+        This method is deprecated.  Use describe_indices() instead, which returns
+        richer per-index information.
```
westonpace, line 966: Should we actually mark it deprecated?

python/src/indices.rs
```diff
@@ -679,7 +684,8 @@ pub struct PyIndexDescription {
     pub index_type: String,
     /// The ids of the fields that the index is built on
     pub fields: Vec<u32>,
-    /// The names of the fields that the index is built on
+    /// The full paths of the fields that the index is built on
+    /// (dotted, with backtick-quoted segments for non-identifier names)
```
westonpace, line 687: The names don't seem to be backtick-quoted?

westonpace (approved):
It'll be great to finally put a lid on the old method 😆

python/python/lance/dataset.py
wjones127: Are you suggesting we should un-deprecate it? I think at this point it's just going to serve as an old alias. I think then end state we want it to remove it one day. But it's a pretty core API so we'll keep it around for a while.

python/src/indices.rs
wjones127: We call `Schema::field_path()` which in turn calls this:

https://github.com/lance-format/lance/blob/13854cc0fd9df9872effaad49078308e26e01588/rust/lance-core/src/datatypes/schema.rs#L1553-L1574

So it does get formatted with backticks.

PR #6904, opened 2026-05-22
fix(index): preserve range partition structure in BTree index update during optimize

Jay-ju:
## Problem

When a BTree index is built with range partitioning (distributed mode), the `update()` method previously called `combine_old_new()` which is unaware of the range partition structure. This caused the retrained index to lose its range partition layout during optimize, producing a single monolithic index instead of per-partition page data and lookup files.

## Root Cause

`BTreeIndex::update()` unconditionally calls `combine_old_new()` + `train_btree_index()` without checking whether the index was built with range partitioning. The `train_btree_index()` function with `range_id=None` generates a non-partitioned index, destroying the original range partition structure.

## Fix

Detect range-partitioned indices by checking `ranges_to_files` in `update()`. When present:

1. Combine old and new data as before via `combine_old_new()`
2. Collect all merged data into sorted batches
3. Re-train each partition independently with the correct `range_id`
4. Merge per-partition lookup files into a unified lookup with correct `page_idx` offsets via `merge_range_partition_lookups_in_place()`

## New Helper Functions

- `collect_sorted_batches()` — collects a `SendableRecordBatchStream` into `Vec<RecordBatch>`
- `slice_batches()` — slices a batch vector by row range `[start, end)`
- `batches_to_stream()` — converts `Vec<RecordBatch>` back to `SendableRecordBatchStream`
- `merge_range_partition_lookups_in_place()` — merges per-partition lookup files into a unified lookup, adjusting `page_idx` offsets and writing `pages_per_range_partition` metadata

## Tests

- `test_optimize_btree_after_append_preserves_data` — verifies regular BTree index data correctness after append + optimize
- `test_optimize_btree_index_update_preserves_range_partition_structure` — verifies range partition structure (`part_*` files) is preserved after optimize

### DIFF
--- a/rust/lance-index/src/scalar/btree.rs
+++ b/rust/lance-index/src/scalar/btree.rs
@@ -29,7 +29,7 @@ use crate::{metrics::NoOpMetricsCollector, scalar::registry::TrainingCriteria};
 use crate::{pbold, scalar::btree::flat::FlatIndex};
 use arrow_arith::numeric::add;
 use arrow_array::{Array, RecordBatch, UInt32Array, new_empty_array};
-use arrow_schema::{DataType, Field, Schema, SortOptions};
+use arrow_schema::{DataType, Field, Schema, SchemaRef, SortOptions};
 use async_trait::async_trait;
 use datafusion::physical_plan::{
     ExecutionPlan, SendableRecordBatchStream,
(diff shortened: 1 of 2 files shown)


Jay-ju:
@claude review

PR #6905, opened 2026-05-22
fix(index): skip scalar index optimize when no unindexed fragments

### LINKED ISSUE #6932: Useless Scalar Index Optimization When No New Fragment Is Present
## Phenomenon

Calling `Dataset.optimize_indices()` on a scalar index that is already in steady state (no unindexed fragments, `num_indices_to_merge <= 1`) is supposed to be a no-op, but instead it:

- rebuilds the existing index segment,
- assigns it a **new UUID**,
- writes a new manifest, advancing the **dataset version on every call**.

Effectively, every "idle" optimize call churns the dataset and the index, producing useless versions and invalidating any cache keyed by the index UUID.

## Reproduction

Repro script: `write/verify_optimize_noop.py`

```python
import lance, pyarrow as pa, tempfile, shutil

tmp = tempfile.mkdtemp()
uri = f"{tmp}/ds"

ds = lance.write_dataset(pa.table({"id": [f"song-{i}" for i in range(32)]}), uri)
ds.create_scalar_index("id", index_type="BTREE", name="id_idx")
ds = lance.dataset(uri)

def uuid_of(d):
    return next(i["uuid"] for i in d._ds.load_indices() if i["name"] == "id_idx")

print(ds.version, uuid_of(ds))

ds.optimize.optimize_indices(num_indices_to_merge=1)   # should be a no-op
ds = lance.dataset(uri)
print(ds.version, uuid_of(ds))

ds.optimize.optimize_indices()                          # should be a no-op
ds = lance.dataset(uri)
print(ds.version, uuid_of(ds))
```

### Observed output

| uuids | 
|---|
| `2 → 3 → 4` |

zhangyue19921010:
Closes: https://github.com/lance-format/lance/issues/6932

## Summary

When `optimize_indices` was called on a scalar index in steady state (no unindexed fragments, and num_indices_to_merge <=1 ), the scalar branch of `merge_indices_with_unindexed_frags` still rebuilt the segment and assigned it a new UUID, churning the dataset version on every call. **But in fact, the index content remains unchanged at all.**

This adds the same no-op short-circuit that already exists for the `Inverted` branch: when `unindexed` is empty and `num_indices_to_merge` resolves to `<= 1`, return `Ok(None)` and skip the rewrite.

## Behavior change

- `OptimizeOptions::default()` / `append()` / `merge(1)` with no pending fragments is now a true no-op
  (UUID and manifest version are preserved).

Same as the logic in inverted index
```
        match index_type {
            IndexType::Inverted => {
                let num_to_merge = options
                    .num_indices_to_merge
                    .unwrap_or(1)
                    .min(old_indices.len());
                if unindexed.is_empty() && num_to_merge <= 1 {
                    return Ok(None);
                }
```

### DIFF
--- a/rust/lance/src/index/append.rs
+++ b/rust/lance/src/index/append.rs
@@ -517,9 +517,6 @@ pub async fn merge_indices_with_unindexed_frags<'a>(
                 ))
             }
             it if it.is_scalar() => {
-                // Skip work when the caller has nothing to do: no unindexed
-                // fragments AND no explicit request to consolidate more than
-                // one old segment.
                 let num_to_merge = options
                     .num_indices_to_merge
                     .unwrap_or(1)
--- a/rust/lance/src/index/append.rs
+++ b/rust/lance/src/index/append.rs
@@ -517,6 +517,17 @@ pub async fn merge_indices_with_unindexed_frags<'a>(
                 ))
             }
             it if it.is_scalar() => {
+                // Skip work when the caller has nothing to do: no unindexed
+                // fragments AND no explicit request to consolidate more than
+                // one old segment.
+                let num_to_merge = options
+                    .num_indices_to_merge
+                    .unwrap_or(1)
+                    .min(old_indices.len());
+                if unindexed.is_empty() && num_to_merge <= 1 {
+                    return Ok(None);
+                }
+
                 // Use effective bitmap (intersected with existing dataset fragments)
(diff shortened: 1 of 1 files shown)


yanghua:
@claude review

wjones127 (approved):
Very straight forward. good work!

PR #6906, opened 2026-05-22
feat(java): speed up zonemap reads

summaryzb:
## Summary
The end-to-end time for reading zonemap for 1 billion rows (even including OSS network latency) is now 1/100 faster than before.
Replaces per-row JNI object construction with a single Arrow IPC byte stream returned across the JNI boundary, then decoded back to List<ZoneStats> in pure Java via the new ZonemapStatsCodec. The profiled hot path drops from O(N)  to O(1) — one byte[] crossing the boundary, regardless of zone count.

## Problem
The profiled hot path is caused by O(N) JNI calls (one new_object + 3 method invocations per zone, plus per-value scalar_value_to_java) 
image

## Tests:
- New ZonemapStatsCodecTest: 9 tests on synthetic IPC streams covering empty/null bytes, null-in-non-nullable across all id columns, date/ timestamp round-trip, and multi-batch ordering.
- ZonemapStatsTest extended with 6 integration tests covering VARCHAR, FLOAT8, BIT, DATEDAY, non-TZ TIMESTAMP, and all-null zones (was IntVector-only).


summaryzb:
@xuanwo @hamersaw PTAL

java/src/main/java/org/lance/index/scalar/ZonemapStatsCodec.java
```diff
@@ -0,0 +1,198 @@
...
+    if (ipcBytes == null || ipcBytes.length == 0) {
+      return Collections.emptyList();
+    }
+
+    ArrayList<ZoneStats> out = new ArrayList<>();
+    try (ByteArrayInputStream in = new ByteArrayInputStream(ipcBytes);
+        ArrowStreamReader reader = new ArrowStreamReader(in, allocator)) {
+      while (reader.loadNextBatch()) {
+        VectorSchemaRoot root = reader.getVectorSchemaRoot();
+        FieldVector fragmentIdVec = root.getVector("fragment_id");
```
Xuanwo: The decoder trusts the IPC schema and coerces structural fields from any numeric vector, so malformed or version-skewed zonemap data can produce unclear runtime failures or incorrect stats instead of a clear invalid-index error.

java/src/main/java/org/lance/index/scalar/ZonemapStatsCodec.java
```diff
@@ -0,0 +1,198 @@
...
+/**
+ * Decodes the Arrow IPC byte stream returned by {@code nativeGetZonemapStatsIpc} into a {@code
+ * List<ZoneStats>}.
+ *
+ * <p>This codec exists to keep {@code List<ZoneStats>} as the public surface of {@link
+ * org.lance.Dataset#getZonemapStats(String)} while letting the JNI hot path return a single Arrow
+ * IPC byte array instead of constructing N {@code ZoneStats} JNI objects per call. The type →
+ * {@code Comparable} mapping mirrors the Rust {@code scalar_value_to_java} helper.
+ */
+public final class ZonemapStatsCodec {
```
Xuanwo: This public codec exposes the internal zonemap IPC payload as a Java API surface, so downstream code can depend on an unversioned transport format and constrain future schema changes.

java/src/main/java/org/lance/index/scalar/ZonemapStatsCodec.java
```diff
@@ -0,0 +1,198 @@
...
+   *
+   * @param ipcBytes Arrow IPC stream, possibly empty (length 0) when the column has no zonemap
+   * @param allocator the {@link BufferAllocator} owning intermediate buffers; the caller retains
+   *     ownership and is responsible for closing it
+   * @return zone stats in the order the Rust side produced them
+   */
+  public static List<ZoneStats> decode(byte[] ipcBytes, BufferAllocator allocator)
+      throws IOException {
+    if (ipcBytes == null || ipcBytes.length == 0) {
+      return Collections.emptyList();
```
Xuanwo: The empty-payload path returns an immutable list while the previous JNI path returned a mutable list, so callers that append to empty results can now fail only in the no-index or empty-index case.

java/src/main/java/org/lance/index/scalar/ZonemapStatsCodec.java
summaryzb: Good point — addressed. The decoder now validates the IPC schema field-by-field before reading any row, against the Rust producer's contract

java/src/main/java/org/lance/index/scalar/ZonemapStatsCodec.java
summaryzb: The empty/null ipcBytes branch now returns new ArrayList<>(), matching the mutability of the previous JNI path so callers that append to an empty result keep working

java/src/main/java/org/lance/index/scalar/ZonemapStatsCodec.java
summaryzb: ZonemapStatsCodec is now package-private (default visibility) and was moved from org.lance.index.scalar to org.lance so it sits next to its only caller

java/src/main/java/org/lance/ZonemapStatsCodec.java
```diff
@@ -0,0 +1,261 @@
...
+    switch (minor) {
+      case TINYINT:
+      case SMALLINT:
+      case INT:
+      case BIGINT:
+      case UINT1:
+      case UINT2:
+      case UINT4:
+      case UINT8:
+        return ((Number) v.getObject(i)).longValue();
```
Xuanwo, line 203: Unsigned zonemap min/max values are decoded through `getObject(i)`, which crashes for UInt16 and returns signed values for UInt8/UInt32. Columns with unsigned zonemap stats can now fail or expose incorrect bounds through the unchanged public API.

PR #6907, opened 2026-05-22
refactor(io/exec): address #6799 review (drop helper, inline RAII timer)

brendanclement:
### Addresses Weston's review comments on #6799

- Revert: drop setup-time Instant::now() brackets that captured I/O time. I wrapped setup `.await`s in `MapIndexExec`, `FlatMatchFilterExec`, and `FlatMatchQueryExec` with `Instant::now() to add_duration(start.elapsed())` to attribute one-shot setup work to `elapsed_compute`. Weston pointed out those brackets capture I/O wait, not CPU.
- Replace InstrumentedChildInputStream with inline RAII timer in .then() bodies, Deletes the `InstrumentedChildInputStream` helper struct.
- Extend flat_bm25 scoring timer to pre- and post-await sync work In flat_bm25_search_stream_with_metrics the single scoring_start region only covered initialize_scorer + flat_bm25_score.

### Testing

Benchmarked (1M-row synthetic dataset with BTREE + INVERTED + IVF_PQ indexes, identical analyze plan queries on each side). Comparing against the commit prior to this work

| Scenario | Node | Before | After | Notes |
|---|---|---|---|---|
| KNN + BTREE scalar prefilter | KNNVectorDistance (520 rows) | 1.39 ms | 112 µs | ~12× smaller — the original double-count fix preserved |
| KNN + FTS post-filter | FlatMatchFilter (0 rows) | 10.95 ms | 42 µs | ~262× smaller |
| KNN + FTS prefilter | FlatMatchQuery (2.52 K rows) | 848 µs | 79.08 ms | ~93× larger; spawn_cpu tokenize CPU now correctly attributed |
| KNN + FTS prefilter (control) | MatchQuery (untouched code) | 3.80 ms | 4.63 ms | noise only |


wjones127 (approved):
Nice cleanup. I like the RAII timers.

PR #6882, opened 2026-05-21
refactor(bench): replace mem_wal_read/vector with focused LSM benchmarks

touch-of-grey:
Final piece of the #6856 split. Replaces the duplicated criterion-based `mem_wal_read.rs` and `mem_wal_vector.rs` benchmarks with two standalone CLI benchmarks that emit JSON output for panel-style trend analysis:

- `mem_wal_vector_bench`: KNN search across LSM levels with deterministic synthetic 384-dim embeddings, IVF-RQ base table index, and recall verification against brute-force ground truth.
- `mem_wal_point_lookup_bench`: PK-based point lookups across the base table, flushed generations, and active memtable.

Both accept `--flushed-generations` and `--max-memtable-rows` for sweeping the full matrix; results are written as individual JSON files.

This is the reusable bench template I want to extend for FTS benchmarking. Depends on the LSM vector-search API (`with_dataset` / `refine_factor`) landed in #6881, so it's the last of the three split PRs.

Part of splitting #6856 into focused PRs. Co-authored with @jackye1995.

PR #6910, opened 2026-05-22
feat(mem_wal): FTS search for LSM scanner with Local and global-rescore modes

touch-of-grey:
Adds full-text search to `LsmScanner`, spanning the base table, flushed
memtable generations, and the active/frozen in-memory memtables, with two
BM25 scoring modes from the multi-segment FTS discussion (#6789):

- **`Local`** — each source scores with its own corpus statistics; the
  coordinator unions per-source plans and merges by `_score` (per-partition
  top-k sort + sort-preserving merge). Single-pass, no cross-source coordination.
- **`LocalWithGlobalRescore`** — each source returns top-K' candidates with
  raw BM25 sufficient stats `(doc_len, term_freqs)`; the planner aggregates
  per-source `(N, sumdl, df_t)` into one global `MemBM25Scorer`, rescores every
  candidate with global stats, and selects the global top-k. `K' = max(rescore_factor·k, 100)`.

### Components
- `LsmFtsSearchPlanner` + `LsmScanner::full_text_search(column, query, k, mode)`.
- lance-index: `InvertedIndex::bm25_candidate_search` returns candidates with
  `(row_id, doc_len, term_freqs, local_score)` (parallel to `bm25_search`).
- mem_wal `FtsMemIndex`: `search_candidates` + `bm25_stats_for_terms` so the
  in-memory index can feed the global scorer.
- Rescore handles index-less flushed generations via a flat scan-and-tokenize
  fallback (mirrors the fallback Local mode already gets from `scanner.full_text_search`).

### Benchmark
`benches/mem_wal/fts/mem_wal_fts_read_bench.rs` (+ `run_fts_read_sweep.sh`),
following the #6882 CLI/JSON template: ShardWriter ingestion of flushed gens +
active memtable, FineWeb text, 200 single-term queries per config under both
modes. EC2 m7gd.4xlarge (ARM, 16 vCPU), local instance-store NVMe vs S3,
2 flushed generations + active, `rescore_factor=10`.

| storage | base | k   | Local p50 | Local p99 | Rescore p50 | Rescore p99 | Jaccard |
|---------|------|-----|-----------|-----------|-------------|-------------|---------|
| NVMe    | 100k | 10  | 16.9 ms   | 19.5 ms   | 31.7 ms     | 32.5 ms     | 0.859   |
| NVMe    | 100k | 100 | 19.3 ms   | 21.4 ms   | 36.2 ms     | 36.9 ms     | 0.927   |
| NVMe    | 1M   | 10  | 16.6 ms   | 18.6 ms   | 37.6 ms     | 40.8 ms     | 0.885   |
| NVMe    | 1M   | 100 | 19.3 ms   | 21.1 ms   | 44.4 ms     | 47.1 ms     | 0.939   |
| S3      | 100k | 10  | 905 ms    | 1475 ms   | 1556 ms     | 2391 ms     | 0.858   |
| S3      | 100k | 100 | 997 ms    | 1889 ms   | 1725 ms     | 2558 ms     | 0.927   |
| S3      | 1M   | 10  | 855 ms    | 1631 ms   | 1396 ms     | 2085 ms     | 0.885   |
| S3      | 1M   | 100 | 963 ms    | 1684 ms   | 1756 ms     | 2636 ms     | 0.939   |

Takeaways:
- **Storage dominates.** NVMe p50 is ~17–19 ms; the same query on S3 is
  ~0.9–1.0 s (Local) / ~1.4–1.8 s (Rescore). Per-query inverted-index reads
  from S3 are uncached across queries here, so each query pays object-store
  round-trips — the headline cost for a cold object-store FTS read tier.
- **Base size barely matters** (100k vs 1M within noise): BM25 over an
  inverted index is ~O(matches) with WAND top-k bounding the work, not corpus size.
- **Rescore costs ~1.7–2.3× over Local** (candidate search + global stat
  aggregation + a `take` to materialize winners). The multiplier shrinks on S3
  (~1.6–1.8×) because S3 read latency is the common denominator both modes pay.

**Jaccard** here = mean top-k overlap between the two modes,
`|A ∩ B| / |A ∪ B|` over the per-query result row-id sets (1.0 = identical
top-k, lower = global rescoring pulled different docs in). It rises with k
(0.86 at k=10 → 0.94 at k=100): rescoring swaps ~1–1.5 docs out of the top-10,
a smaller fraction of a top-100. This quantifies how much global-stats
rescoring actually changes results vs local scoring — the #6789 tradeoff.
It's a membership metric (ignores intra-top-k reorder); a rank/score metric
(NDCG / score correlation) is a useful follow-up.

Draft: opening for design feedback on the rescore planner shape and the
scoring-mode API before polishing.


touch-of-grey:
@jackye1995 @hamersaw — opening this as a draft for design feedback on the LSM FTS scoring-mode API and the rescore planner shape. PTAL when you have a chance.

jackye1995:
Thanks for sharing the draft! This is very helpful result, looks like the global bm25 scorer does have quite some penalty. Given that, I would propose let's get the local version implemented first, which should also introduce minimum change into the lance-index side of things, and then we can see if we can optimize the global option further.

PR #6915, opened 2026-05-22
docs: add new index page for integrations and fix broken build

prrao87:
The Lance-DuckDB documentation build is broken due to an incorrect URL. This PR fixes that to render correctly in the sidebar via the `.pages` card. The build scripts have been updated to only include the relevant docs files from the `lance-duckdb/docs` folder.

Also adds a new `index.md` page that users will now land on, rather than landing on the DataFusion page. The index page contains links to all the other relevant integrations.


jackye1995 (approved):
thanks for the fix!

PR #6920, opened 2026-05-23
feat: add free-form metadata map to DataFragment

jackye1995:
## Summary

- Add `map<string, string> metadata` field to `DataFragment` protobuf for lightweight per-fragment key-value metadata (e.g., creation time, process ID).
- Extend `UpdateConfig` transaction with `fragment_metadata_updates` to support updating fragment metadata, using existing `UpdateMap` semantics (merge or replace mode).
- Add conflict detection: two `UpdateConfig` transactions conflict if they modify metadata on the same fragment ID.
- Update transaction spec doc.

## Voting Discussion

This is a spec change. A corresponding [VOTE] discussion will be created separately.

### DIFF
--- a/protos/table.proto
+++ b/protos/table.proto
@@ -346,6 +346,10 @@ message DataFragment {
   // deletion tombstones. To compute the current number of rows, subtract
   // `deletion_file.num_deleted_rows` from this value.
   uint64 physical_rows = 4;
+
+  // Free-form key-value metadata for the fragment.
+  // Intended for lightweight metadata such as creation time, process ID, etc.
+  map<string, string> metadata = 11;
 }
 
 message DataFile {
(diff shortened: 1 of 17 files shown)


protos/table.proto
```diff
@@ -346,6 +346,10 @@ message DataFragment {
   // deletion tombstones. To compute the current number of rows, subtract
   // `deletion_file.num_deleted_rows` from this value.
   uint64 physical_rows = 4;
+
+  // Free-form key-value metadata for the fragment.
+  // Intended for lightweight metadata such as creation time, process ID, etc.
+  map<string, string> metadata = 11;
```
wjones127, line 352: Could you provide a concrete use case that isn't covered by metadata elsewhere?

* `creation time` -> that's just the timestamp in the manifest that created it
* `process id` -> similarly that just the provenance info from the manifest / transaction that created it.

If you want to make querying this stuff efficient, I favor putting this in some secondary structure rather than in the core manifest structure.

IMO one of the key features of Lance is that it's metadata layer is pretty compact and efficient. Many tables are represented by a manifest that's less than 5KB. The more fields we add, the more opportunity there is for bloat. This is especially true if we start writing fragment-level metadata. If we start injecting standard keys here, we now have `O(num_fragments)` copies of keys and their values.

PR #6925, opened 2026-05-23
feat(example): add generated dataset fallback for ivf_hnsw example

jiengup:
Summary
- Make the `ivf_hnsw` example dataset URI optional
- Generate a temporary local Lance dataset when no URI is provided
- Ensure generated vector column uses `FixedSizeList<Float32, 64>` so it can build a vector index
- Wire CLI HNSW params into `HnswBuildParams`

## Testing
- `cargo check -p lance-examples --example ivf_hnsw


rust/examples/src/ivf_hnsw.rs
```diff
@@ -68,20 +112,33 @@ fn ground_truth(mat: &MatrixView<Float32Type>, query: &[f32], k: usize) -> HashS
 async fn main() {
     env_logger::init();
     let args = Args::parse();
+    let tempdir;
+    let column = args.column.as_deref().unwrap_or("vector");
 
-    let mut dataset = Dataset::open(&args.uri)
-        .await
-        .expect("Failed to open dataset");
+    let uri = match args.uri.as_deref() {
```
Xuanwo, line 123: The no-argument fallback generates data but leaves index creation disabled, so the zero-argument ivf_hnsw example benchmarks flat KNN instead of IVF_HNSW.

rust/examples/src/ivf_hnsw.rs
```diff
@@ -68,20 +112,33 @@ fn ground_truth(mat: &MatrixView<Float32Type>, query: &[f32], k: usize) -> HashS
...
     let metric_type = MetricType::try_from(args.metric_type.as_str()).unwrap();
 
     let mut ivf_params = IvfBuildParams::new(128);
     ivf_params.sample_rate = 20480;
     let hnsw_params = HnswBuildParams::default()
-        .ef_construction(100)
-        .num_edges(15);
+        .ef_construction(args.ef)
+        .max_level(args.max_level)
+        .num_edges(args.max_edges);
```
Xuanwo, line 146: The `--max-edges` value is forwarded as HNSW `m`; the default `30` doubles the level-0 edge cap versus the previous example and makes benchmark results denser and more expensive.

rust/examples/src/ivf_hnsw.rs
```diff
@@ -68,20 +112,33 @@ fn ground_truth(mat: &MatrixView<Float32Type>, query: &[f32], k: usize) -> HashS
...
-    let column = args.column.as_deref().unwrap_or("vector");
     let metric_type = MetricType::try_from(args.metric_type.as_str()).unwrap();
 
     let mut ivf_params = IvfBuildParams::new(128);
     ivf_params.sample_rate = 20480;
     let hnsw_params = HnswBuildParams::default()
-        .ef_construction(100)
-        .num_edges(15);
+        .ef_construction(args.ef)
+        .max_level(args.max_level)
```
Xuanwo, line 145: The CLI accepts `--max-level 0` and forwards it to the builder, causing a panic during index construction instead of a controlled input error.

rust/examples/src/ivf_hnsw.rs
jiengup: Set the default value to `15` in the newset version.

rust/examples/src/ivf_hnsw.rs
jiengup: Reject `--max-level == 0` in the newest version.

rust/examples/src/ivf_hnsw.rs
jiengup: Set `--create_index` default to `true` in the newset version.

jiengup:
@Xuanwo 
Thanks for your advice! Please check it again.

PR #6929, opened 2026-05-24
fix(mem_wal): eliminate fresh-tier stale reads (dedup-on-scan + deletion-vector-on-flush)

hamersaw:
Unfortunately, we still have an issue where we can serve stale data from within a single source in the `mem_wal` implementation. The crux is that we dedup globally AFTER we run the filter at each source. So if the newer data does not satisfy the filter, then the algorithm maintains the stale data.

In this PR we updated the active memtable flush path the make a single pass over the in-memory data and compute a delete vector that gets written on the flushed memtable. Since we read using the native `LanceScanner` that respects the delete vector we essentially dedup the flushed memtable without having to invalidate and recompute the indicies that are built during inserts on the active memtable. As part of this we removed all of the now unnecessary `_generation` / `_freshness` field addition and datafusion dedup nodes.


jackye1995 (approved):
looks good to me! This is indeed a better approach compared to the reverse write

PR #6930, opened 2026-05-24
perf: fix very long row count times in ivf sampling

cmccabe:
Previously, when counting the number of rows in maybe_sample_training_data, if a list of fragment ids was supplied, we would read much more data than required.  This is because we simply used a scanner with a predicate. This change adds a count_rows method which simply calls count_rows on each fragment. Additionally, we parallelize this operation.


wjones127 (approved):
This seems fine as a stop-gap, but I'm going to file a follow up to make the original code efficient. Seems like something our query optimizer should be able to solve easily.

```rust
        scanner.with_fragments(resolve_scan_fragments(dataset, fragment_ids)?);
        scanner.count_rows().await? as usize
```

PR #6937, opened 2026-05-25
fix: apply filtered read limit after refine filter

### LINKED ISSUE #3436: scanner(filter=A AND B, limit=N)` returns 0 rows when A has a scalar index (BTree or Bitmap) and matches are not in the first fragment
# `scanner(filter=A AND B, limit=N)` returns 0 rows when A has a scalar index (BTree or Bitmap) and matches are not in the first fragment
## Environment
- `lance` 4.0.1
- `lancedb` 0.30.2
- `pyarrow` 24.0.0
- Python 3.11.14
- macOS 25 / Apple Silicon
## Summary
A scanner call like `ds.scanner(filter="topic = 'A' AND is_night = true", limit=10)` returns **0 rows** when `topic` has a scalar index (BTree or Bitmap, both reproduce) AND the matching rows are not in the first fragment scanned. The same predicate without `limit=` (or with a "large enough" limit only by accident) returns the correct count. Same surface bug via `tbl.search().where(pred, prefilter=True).limit(10)`.
The bug appears to be in the scalar-index-prefilter + limit interaction: Lance uses the scalar index to narrow to rows matching the indexed column, applies `limit` to that index-prefiltered row stream, then evaluates the second predicate on the truncated set. If the first `limit` index-matches happen to fail the second predicate, the result is empty even when there are matching rows further on.
## Minimal repro
Self-contained, no MCAP / external data:
```python
import tempfile, pyarrow as pa, lance, lancedb
    db = lancedb.connect(tmpdir)
    # Three fragments. The matching rows for `topic='A' AND is_night=true`
    # live in the LAST fragment.
    tbl = db.create_table(
        "repro",
        pa.table({"topic": ["A"]*100_000, "is_night": [False]*100_000}),
        mode="overwrite",
    )
    tbl.add(pa.table({"topic": ["B"]*100_000, "is_night": [False]*100_000}))
    tbl.add(pa.table({"topic": ["A"]*1_000,   "is_night": [True]*1_000}))
    # A scalar index on `topic` is required to trigger the bug —
    # both BTREE and BITMAP reproduce.
    tbl.create_scalar_index("topic", index_type="BTREE")
    ds = lance.dataset(f"{tmpdir}/repro.lance")
    pred = "topic = 'A' AND is_night = true"
    print("no limit:", ds.scanner(filter=pred).to_table().num_rows)        # 1000  ✓
    print("limit=10:", ds.scanner(filter=pred, limit=10).to_table().num_rows)   # 0    ✗  expected 10
    print("limit=100k:", ds.scanner(filter=pred, limit=100_000).to_table().num_rows)  # 0    ✗  expected 1000
```
Actual output:
```
no limit: 1000
limit=10: 0
limit=100k: 0
```
Same bug surfaces through `lancedb.Table.search()`:
```python
tbl.search().where(pred, prefilter=True).limit(10).to_arrow().num_rows
# -> 0  (expected 10)
```
## Behaviour matrix
Holding `pred = "topic = 'A' AND is_night = true"` and varying setup:
| Setup | full count | `limit=10` | OK? |
| --- | --- | --- | --- |
| 3 fragments (a-day, b-day, a-night), BTree on `topic` | 1000 | 0 | ❌ |
| 3 fragments (a-day, b-day, a-night), Bitmap on `topic` | 1000 | 0 | ❌ |
| 3 fragments (a-day, **a-night**, b-day), BTree on `topic` | 1000 | 0 | ❌ |
| 3 fragments (**a-night**, a-day, b-day), BTree on `topic` | 1000 | 10 | ✓ |
| 3 fragments (a-day, b-day, a-night), **no index** | 1000 | 10 | ✓ |
| 1 fragment with all rows interleaved (a-day before a-night), BTree on `topic` | 100 | 0 | ❌ |
| Single-column: `filter="topic = 'A'", limit=10` | 101 000 | 10 | ✓ |
| Single-column: `filter="is_night = true", limit=10` | 1000 | 10 | ✓ |
The bug fires whenever:
1. A scalar index (BTree or Bitmap) covers one of the AND'd columns, **and**
2. The first `limit` rows in the index-prefiltered stream all fail the secondary predicate.
It does **not** fire on single-column predicates, on tables without the index, or when the matching rows happen to come first in the index-prefilter scan order.
## Expected behaviour
`scanner(filter="A AND B", limit=N)` should return up to N rows where **both** `A` and `B` are true. The current behaviour drops post-filter matches that exist beyond the first N index-matches, which violates standard SQL `WHERE … LIMIT N` semantics.
## Suspected mechanism
Best guess from the outside (without reading the Lance source):
1. The query planner sees `A_indexed AND B`, picks the scalar-index-prefilter strategy for `A` (works for both BTree and Bitmap).
2. The scalar index returns a `Vec<RowAddress>` of all rows matching `A = X` (cheap, indexed).
3. **`limit` is applied to that row-address stream** before the secondary filter is evaluated.
4. Lance materialises only the first N row addresses, projects the requested columns, evaluates `B = Y` on those N rows, and returns whatever passes — which can be 0.
If that's the actual order, the fix is to apply `limit` *after* the secondary filter is evaluated: keep streaming index-matches and dropping non-`B = Y` rows until N have passed both filters (or the stream is exhausted).
A streaming early-stop on the user side proves the data is reachable:
```python
def stream_until(builder, limit):
    rows, n = [], 0
    for batch in builder.to_batches():
        if not batch.num_rows:
            continue
        if n + batch.num_rows >= limit:
            rows.append(batch.slice(0, limit - n))
            break
        rows.append(batch); n += batch.num_rows
    return pa.Table.from_batches(rows) if rows else pa.table({})
stream_until(tbl.search().where(pred, prefilter=True), 10).num_rows
# -> 10  ✓
```
So the rows are findable through the same builder; the issue is specifically how `limit` composes with the scalar-index-prefilter path (both BTree and Bitmap).
## Workaround
Drop `limit` from the scanner / query builder when the predicate touches an indexed scalar column, and stream batches with manual early-stop. The streaming form behaves correctly because each batch already has the secondary predicate applied before being returned.
## Why this matters
Filter-then-limit is the canonical UI pagination shape ("show me 50 night images") and is also the exact shape the LanceDB Streamlit demo uses on its Curation page. With even a single scalar index on `topic`, every compound predicate that touches it loses rows silently — the user sees an empty result and assumes their data isn't enriched, when in reality the rows exist but the limit truncated them out of the index-prefilter stream before the second predicate fired.

Xuanwo:
FilteredReadExec could spend a pushed-down LIMIT on scalar-index matches before evaluating a remaining refine filter, which allowed filtered scans to return fewer rows than requested even when enough rows matched the full predicate.

This keeps scalar-index range truncation behind the full-filter exactness boundary: indexed candidates are still narrowed by the scalar index, but scan_range_after_filter is only consumed early when no refine filter remains. Regression coverage exercises both the FilteredReadExec path and the scanner(filter, limit) end-to-end behavior.

Fixes lancedb/lancedb#3436.

### DIFF
--- a/rust/lance/src/dataset/scanner.rs
+++ b/rust/lance/src/dataset/scanner.rs
@@ -5530,6 +5530,56 @@ mod test {
         Ok(())
     }
 
+    #[tokio::test]
+    async fn test_limit_with_scalar_index_and_refine_filter() {
+        let schema = Arc::new(ArrowSchema::new(vec![
+            ArrowField::new("id", DataType::Int32, false),
+            ArrowField::new("topic", DataType::Int32, false),
+            ArrowField::new("is_night", DataType::Int32, false),
+        ]));
+        let batch = RecordBatch::try_new(
+            schema.clone(),
+            vec![
+                Arc::new(Int32Array::from_iter_values(0..20)),
+                Arc::new(Int32Array::from_iter_values(std::iter::repeat_n(1, 20))),
+                Arc::new(Int32Array::from_iter_values(
+                    (0..20).map(|i| if i < 10 { 0 } else { 1 }),
+                )),
+            ],
+        )
+        .unwrap();
+        let reader = RecordBatchIterator::new(vec![Ok(batch)], schema.clone());
+        let mut dataset = Dataset::write(reader, "memory://", None).await.unwrap();
+        dataset
+            .create_index(
+                &["topic"],
+                IndexType::BTree,
+                None,
+                &ScalarIndexParams::default(),
+                true,
+            )
+            .await
+            .unwrap();
+
+        let actual = dataset
+            .scan()
+            .filter("topic = 1 AND is_night = 1")
+            .unwrap()
+            .limit(Some(10), None)
+            .unwrap()
+            .try_into_batch()
+            .await
+            .unwrap();
+
+        assert_eq!(actual.num_rows(), 10);
+        let ids = actual
+            .column_by_name("id")
+            .unwrap()
+            .as_primitive::<Int32Type>()
+            .values();
+        assert_eq!(ids, &(10..20).collect::<Vec<_>>());
+    }
+
     #[test_log::test(tokio::test)]
     async fn test_limit_cancel() {
         // If there is a filter and a limit and we can't use the index to satisfy
--- a/rust/lance/src/io/exec/filtered_read.rs
+++ b/rust/lance/src/io/exec/filtered_read.rs
@@ -515,18 +515,31 @@ impl FilteredReadStream {
         evaluated_index: &Option<Arc<EvaluatedIndex>>,
         options: &FilteredReadOptions,
     ) -> FilteredReadInternalPlan {
-        // For pushing down scan_range_after_filter
+        // For pushing down scan_range_after_filter.
+        //
+        // This is only valid when there is no refine filter left to evaluate.  An exact scalar
+        // index result is exact for the indexed predicate, but not for the full predicate if a
+        // refine filter can still reject rows.
+        let can_push_down_scan_range_after_filter = options.refine_filter.is_none();
         let mut scan_planned_with_limit_pushed_down = false;
-        let mut to_skip = options
-            .scan_range_after_filter
-            .as_ref()
-            .map(|r| r.start)
-            .unwrap_or(0);
-        let mut to_take = options
-            .scan_range_after_filter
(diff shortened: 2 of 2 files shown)


Xuanwo:
@claude review once

wjones127 (approved):
Makes sense. Thanks for the fix!

Issue #6577, opened 2026-04-19
[Java/Spark] JNI dispatcher thread causes ClassNotFoundException and prevents JVM shutdown


summaryzb:
## Summary
  When using lance-java inside Apache Spark executors, two related JNI threading issues cause failures:

  1. **ClassNotFoundException on the dispatcher thread** — the lance JNI dispatcher thread cannot find `org.lance.ipc.AsyncScanner`
  2. **Spark process never exits** — the dispatcher thread keeps the JVM alive indefinitely after the Spark task completes

  Both stem from the same root cause: the `lance-jni-dispatcher` thread is a native thread attached to the JVM via `attach_current_thread_permanently()`, which gives it the wrong classloader **and** marks it as a non-daemon thread.

  ## Error log
```
  thread 'lance-jni-dispatcher' panicked at src/dispatcher.rs:44:22:
  AsyncScanner class not found: JavaException
  java.lang.NoClassDefFoundError: org/lance/ipc/AsyncScanner
  Caused by: java.lang.ClassNotFoundException: org.lance.ipc.AsyncScanner
```

summaryzb:
Introduced since https://github.com/lance-format/lance/pull/6102

Xuanwo:
Closing as completed. The JNI dispatcher uses a cached `GlobalRef` and attaches as a daemon, with custom-classloader coverage for both `ClassNotFoundException` and JVM shutdown behavior.

PR #6946, opened 2026-05-26
fix(java): resolve JNI classloader bug on dispatcher thread in Spark

sezruby:
Reopens the work from #6549, which was closed on 2026-05-20 with a "deduplicate with #6633" note. PR #6633 (merged) makes the dispatcher thread a daemon, but does **not** touch `find_class`. The classloader bug therefore remains unfixed on `main` — verified at `java/lance-jni/src/dispatcher.rs:45` on `lance-format/lance@main`:

```rust
.find_class("org/lance/ipc/AsyncScanner")
.expect("AsyncScanner class not found");
```

Tracking issue #6577 is still open. Detailed reproducer, full panic traces, and Databricks Runtime cluster setup at lance-format/lance-spark#538.

## Why this is broken on Databricks

On a Databricks cluster, the bundle JAR (`org.lance:lance-spark-bundle-*`) is installed via the cluster Libraries → Maven UI. Per Databricks' library-loading model, user-installed JARs end up under `/local_disk0/tmp/addedFile*` and are loaded into an **isolated** classloader chain (`MutableURLClassLoader` → `ChildFirstURLClassLoader`) installed per-task by `Executor.TaskRunner.run()`. Importantly, this classloader is installed as the **thread-context classloader on Java application threads only** — the JVM `AppClassLoader` is never updated.

When lance-jni does:

```rust
let dispatcher_thread = thread::Builder::new()
    .spawn(move || { /* ... */ env.find_class("org/lance/ipc/AsyncScanner") })
```

the spawned thread is a **native thread** that lance-jni then attaches to the JVM via `attach_current_thread_permanently()`. Per the JNI spec (https://docs.oracle.com/javase/8/docs/technotes/guides/jni/spec/invocation.html#wp1539), `FindClass` on an attached native thread uses the JVM's *system* classloader (`AppClassLoader`) — which on Databricks only sees `/databricks/jars/*` (Spark itself + DBR-managed JARs), **not** the user library directory. The lookup returns null, the Rust unwrap panics, the panic propagates to the JVM as a Java exception, `SparkUncaughtExceptionHandler` catches it and calls `System.exit(50)`, and the executor dies.

The same trap bites every JNI library that does `FindClass` from a Rust-spawned thread. The canonical fix (used by RocksDB JNI, netty-tcnative, Arrow's own JNI) is to resolve user classes once during `JNI_OnLoad` — which is invoked by `System.loadLibrary` on the *calling Java thread*, where the thread-context classloader is still the user's classloader chain — and pin the resulting `jclass` reference as a `GlobalRef`.

## What this PR does

This is a clean cherry-pick of @summaryzb's fix commit ([b9ab00c](https://github.com/lance-format/lance/commit/b9ab00c5)) onto current `main`, **without** the daemon-thread changes (already merged via #6633). Authorship preserved.

```
java/lance-jni/src/dispatcher.rs   |  +13/-2  use cached GlobalRef instead of FindClass
java/lance-jni/src/lib.rs          |  +20/-1  resolve AsyncScanner once in JNI_OnLoad
java/src/test/java/org/lance/AsyncScannerTest.java     | +232 (new)  classloader regression test
java/src/test/java/org/lance/ClassloaderBugBootstrap.java | +74 (new)
java/src/test/java/org/lance/ClassloaderBugHelper.java | +82 (new)
```

Changes in `lib.rs`:

```rust
pub extern "system" fn JNI_OnLoad(vm: JavaVM, _reserved: *mut c_void) -> jint {
    let env = vm.get_env().expect("Failed to get JNIEnv in JNI_OnLoad");
    let async_scanner_class = env
        .find_class("org/lance/ipc/AsyncScanner")     // runs on Java thread; user CL visible
        .expect("AsyncScanner class not found");
    let global_ref = env
        .new_global_ref(async_scanner_class)          // pin across thread boundary
        .expect("Failed to create GlobalRef for AsyncScanner class");
    DISPATCHER.set(Dispatcher::initialize(Arc::new(vm), global_ref)).unwrap();
    JNI_VERSION_1_8
}
```

Changes in `dispatcher.rs`: store the `GlobalRef` on the `Dispatcher` struct and use it in place of the `FindClass` call. The native thread no longer touches the JVM classloader.

The new tests under `java/src/test/java/org/lance/` (`AsyncScannerTest`, `ClassloaderBugBootstrap`, `ClassloaderBugHelper`) reproduce the classloader scenario in unit tests by loading lance classes through a custom `URLClassLoader` and exercising the dispatcher path — the same pattern that fails on Databricks.

## Validation

Cross-compiled this branch's `liblance_jni.so` to linux-x86-64, swapped it into `lance-spark-bundle-3.5_2.12:0.4.0-beta.4`, and re-ran the original repro on Databricks Runtime: stock bundle still hits the `AsyncScanner class not found` panic / executor exit 50; patched build resolves end-to-end. Reproducer details, panic traces, cluster setup, and the bundle-version matrix are in lance-format/lance-spark#538.

## Notes for reviewers

- The `FindClass` lookup in `JNI_OnLoad` itself is safe because `JNI_OnLoad` runs on a Java thread whose context classloader sees the user bundle JAR — same pattern as RocksDB / netty-tcnative / Arrow's JNI bindings.
- `attach_current_thread_as_daemon` (from #6633) is necessary for clean JVM shutdown but is **not sufficient** to fix the classloader bug. Both fixes are needed; the daemon fix alone leaves the JNI lookup broken.
- If lance-jni adds more `FindClass` lookups later (for other Java classes the dispatcher needs), they should follow the same `JNI_OnLoad` + `GlobalRef` pattern. Worth a code-review check on future JNI additions.

Addresses the classloader half of #6577 (the JVM-shutdown half is already fixed via #6633). Tracking from the lance-spark side at lance-format/lance-spark#538.

Co-Authored-By: zhoubin11 <zhoubin11@baidu.com>

### DIFF
--- a/java/lance-jni/src/dispatcher.rs
+++ b/java/lance-jni/src/dispatcher.rs
@@ -23,8 +23,12 @@ pub struct Dispatcher {
 }
 
 impl Dispatcher {
-    /// Initialize the dispatcher with a persistent JNI thread
-    pub fn initialize(jvm: Arc<JavaVM>) -> Arc<Self> {
+    /// Initialize the dispatcher with a persistent JNI thread.
+    ///
+    /// `async_scanner_class` must be a `GlobalRef` to the `AsyncScanner`
+    /// Java class, resolved on a thread that has the correct application
+    /// classloader (typically the `JNI_OnLoad` thread).
+    pub fn initialize(jvm: Arc<JavaVM>, async_scanner_class: GlobalRef) -> Arc<Self> {
         let (tx, mut rx) = mpsc::unbounded_channel::<DispatcherMessage>();
 
         // Spawn persistent dispatcher thread
@@ -40,10 +44,9 @@ impl Dispatcher {
 
                 log::info!("JNI dispatcher thread started");
 
-                // Cache method IDs for completeTask and failTask
-                let async_scanner_class = env
-                    .find_class("org/lance/ipc/AsyncScanner")
-                    .expect("AsyncScanner class not found");
+                // Cache method IDs for completeTask and failTask.
+                // Use the pre-resolved GlobalRef instead of find_class to
+                // avoid classloader issues on this native thread.
                 let complete_method = env
                     .get_method_id(&async_scanner_class, "completeTask", "(JJ)V")
                     .expect("completeTask method not found");
--- a/java/lance-jni/src/lib.rs
+++ b/java/lance-jni/src/lib.rs
@@ -162,10 +162,25 @@ pub extern "system" fn JNI_OnLoad(
     vm: jni::JavaVM,
     _reserved: *mut std::ffi::c_void,
 ) -> jni::sys::jint {
+    // Resolve AsyncScanner class on the current thread which has the correct
+    // application classloader. A newly spawned native thread only gets the
+    // system classloader after attach_current_thread_permanently(), which
+    // cannot find application classes in environments like Spark, web
+    // containers, or shaded JARs.
+    let mut env = vm.get_env().expect("Failed to get JNIEnv in JNI_OnLoad");
+    let async_scanner_local = env
+        .find_class("org/lance/ipc/AsyncScanner")
+        .expect("AsyncScanner class not found");
+    let async_scanner_class = env
+        .new_global_ref(async_scanner_local)
+        .expect("Failed to create GlobalRef for AsyncScanner class");
+
     let jvm_arc = Arc::new(vm);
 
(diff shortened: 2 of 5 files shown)


hamersaw (approved):
Looks great, thanks for the fix!

jackye1995 (approved):
thanks for the fix!

sezruby:
fixed cargo format style error from auto merging main

PR #6948, opened 2026-05-26
perf(knn): reduce memory for batch flat vector search

### LINKED ISSUE #6940: perf: High memory usage for batched vector query on flat KNN
https://github.com/lance-format/lance/pull/6828 introduced batched vector query, which supports passing multiple query vectors in single query request.

The implementation stores the record batches that referred by topk candidates, which could require `m * k * d * 4` bytes at worst case, where `m` is the number of query vectors, `k` is the number of results for each query vector, `d` is the dimension of vector, `4` is `sizeof(f32)`.

We don't need to store the raw vectors unless the query selects vector column, in most cases, vector column won't be selected, so this can significantly reduce the memory footprint

LeoReeYang:
## Summary
- Reduce memory usage of **batch flat KNN** when the final projection does **not** include the vector column (Closes #6940).
- Heap candidates no longer retain whole scan `RecordBatch`es (and the full vector column) in the common `columns=[...]` case.
- When the vector column *is* projected, retain only a per-row copy of the vector (not the entire scan batch).

## Implementation notes
- `Scanner::flat_knn` computes `retain_vector` from the final projection and plumbs it into `KnnBatchParams`.
- `KNNVectorDistanceExec` batch mode uses a two-tier candidate representation:
  - `RowIdOnly { query_index, distance, row_id }` when `retain_vector=false`
  - `WithVector { ..., slim_batch, row_index, vector_row }` when `retain_vector=true`

## Memory (unit test)
Synthetic workload matching `test_batch_knn_heap_memory`: `num_rows=4096`, `dim=512`, `m=10`, `k=10`.

**What is measured:** sum of Arrow `get_array_memory_size()` over candidate heap entries (not peak RSS or end-to-end query memory).

| Strategy | Candidate heap (array bytes) | vs baseline |
|----------|------------------------------|-------------|
| Baseline (full scan batch per candidate) | 842,167,200 (~803 MiB) | 1× |
| **RowIdOnly** (no vector in projection) | 0 | ≪ baseline (`< baseline/50`) |
| **WithVector** (per-row vector copy) | 257,664 (~252 KiB) | ~3,270× smaller (`< baseline/10`) |

Notes:
- Baseline simulates pre-fix behavior: each of `m×k` candidates holds a shallow clone of the full scan batch (vector + `_rowid` for all `num_rows`). In this synthetic setup the same batch is cloned repeatedly, so the counter sums shared buffers per candidate (worst-case retention model).
- `WithVector` is on the order of `m×k×d×4` (= 204,800 bytes for vector data) plus one deduplicated slim batch (`_rowid` only).
- Indexed batch KNN and other large columns are out of scope for this PR.

## Latency (local, secondary)
Release build (`uv run maturin develop --uv --release`), `use_index=false`, `with_row_id=true`.

Workload: `rows=200_000`, `dim=256`, `m=10`, `k=10` (medians over 8 rounds, 2 warmup).

| | separate | batch | batch speedup |
|---|----------|-------|---------------|
| Baseline (`upstream-main`) | 0.1262s | 0.09365s | 1.35× |
| This PR | 0.1272s | 0.05466s | 2.33× |

Batch latency improved ~1.71× vs baseline in this workload; separate-query path unchanged.

## Test plan
- [x] `cargo test -p lance --lib batch_knn`
- [x] `cd python && uv run pytest python/tests/test_vector_index.py -k batch`

### DIFF
--- a/rust/lance/src/dataset/scanner.rs
+++ b/rust/lance/src/dataset/scanner.rs
@@ -4424,6 +4424,14 @@ impl Scanner {
         } else {
             input
         };
+        let retain_vector = if self.is_batch_nearest {
+            let vector_field_id = self.dataset.schema().field_id(q.column.as_str())?;
+            self.projection_plan
+                .physical_projection
+                .contains_field_id(vector_field_id)
+        } else {
+            false
+        };
         let flat_dist = Arc::new(KNNVectorDistanceExec::try_new_batch(
             input,
             &q.column,
@@ -4435,6 +4443,7 @@ impl Scanner {
                 lower_bound: q.lower_bound,
                 upper_bound: q.upper_bound,
                 distance_type: metric_type,
+                retain_vector,
             },
         )?);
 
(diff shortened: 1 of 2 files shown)


LeoReeYang:
Superseded by #6950 after branch rename to `perf/batch-flat-knn-memory` and rebase onto latest main.

PR #6950, opened 2026-05-26
perf(knn): reduce memory for batch flat vector search

### LINKED ISSUE #6940: perf: High memory usage for batched vector query on flat KNN

LeoReeYang:
## Summary

Closes #6940.

Batch flat KNN previously kept a full scan `RecordBatch` (including the vector column) for each top-k candidate, which could use on the order of `m × k × d × 4` bytes when many query vectors and large `k`/`d` are used.

This PR changes batch flat KNN so that:

- If the **vector column is not in the final projection**, the top-k heap stores only row identifiers and distances (not whole batches or vector data).
- If the **vector column is projected**, each candidate keeps a single-row vector plus a shared scan batch **without** the vector column (for any other columns needed on output).
- If the flat scan already reads **non-vector columns** (for example columns referenced by a filter), only the vector column is dropped from what we retain; other columns can still be assembled in the KNN node without keeping vectors in the heap.

## Behavior

- [`Scanner::flat_knn`](rust/lance/src/dataset/scanner.rs) sets `retain_vector` from whether the user projection includes the vector field.
- [`KNNVectorDistanceExec`](rust/lance/src/io/exec/knn.rs) batch mode maintains a **per-query** top-k heap across all scan batches and emits one result batch with `query_index` and `_distance`.
- Indexed batch KNN is unchanged; this targets **flat** batch vector search only.

## Memory (qualitative)

| Projection / scan input | What top-k heap retains |
|-------------------------|-------------------------|
| No vector in projection; scan is `vec + _rowid` | Row id + distance only |
| Vector in projection | Shared batch without vector column + one row of vector per candidate |
| No vector in projection; scan includes other columns | Shared batch without vector column (no vector rows stored) |

## Test plan

- [x] `cargo test -p lance --lib test_batch_knn_flat`
- [x] `cargo test -p lance --lib test_batch_knn_flat_omits_vector_without_projection`
- [x] `cargo test -p lance --lib test_batch_knn_flat_filter_keeps_non_vector_columns`
- [x] `cargo test -p lance --lib test_batch_knn_flat_respects_distance_range`
- [x] `cargo fmt --all` and `cargo clippy -p lance --lib -- -D warnings`


LeoReeYang:
@BubbleCal Could you review this PR when you have time? It addresses #6940

rust/lance/src/dataset/scanner.rs
```diff
@@ -4424,6 +4424,14 @@ impl Scanner {
         } else {
             input
         };
+        let retain_vector = if self.is_batch_nearest {
+            let vector_field_id = self.dataset.schema().field_id(q.column.as_str())?;
```
BubbleCal, line 4485: does this work with nested vector field? can we add a test for this?

rust/lance/src/io/exec/knn.rs
```diff
@@ -286,13 +291,19 @@ impl KNNVectorDistanceExec {
                 "batch KNN cannot run when the input already contains reserved column '{QUERY_INDEX_COL}'"
             )));
         }
-        let input_schema = Arc::new(input_schema);
+
+        let stored_schema = if is_batch && !retain_vector {
+            Arc::new(Schema::new(vec![ROW_ID_FIELD.clone()]))
```
BubbleCal: something like `intput_schema.drop_column(vector_column)` would be better, we may have more columns other than row ID here

rust/lance/src/io/exec/knn.rs
```diff
@@ -329,19 +340,104 @@ impl KNNVectorDistanceExec {
...
             metrics: ExecutionPlanMetricsSet::new(),
         })
     }
 
+    fn take_vector_row(
+        batch: &RecordBatch,
+        column: &str,
+        row_index: u32,
+    ) -> DataFusionResult<ArrayRef> {
+        let vectors = batch.column_by_name(column).ok_or_else(|| {
```
BubbleCal: just do `batch[column]` is fine

rust/lance/src/io/exec/knn.rs
```diff
@@ -329,19 +340,104 @@ impl KNNVectorDistanceExec {
...
+            DataFusionError::Internal(format!(
+                "batch KNN expected vector column '{column}' in scan batch"
+            ))
+        })?;
+        let indices = UInt32Array::from(vec![row_index]);
+        arrow::compute::take(vectors, &indices, None)
```
BubbleCal: wonder whether there is an easier way to avoid constructing this `indices`, probably just directly copy the data out?

rust/lance/src/io/exec/knn.rs
```diff
@@ -373,6 +470,16 @@ impl KNNVectorDistanceExec {
                 .as_primitive::<UInt64Type>()
                 .clone();
 
+            let slim_batch = if retain_vector {
+                Some(Arc::new(
```
BubbleCal: `Arc` seems unnecessary

rust/lance/src/io/exec/knn.rs
```diff
@@ -673,11 +820,201 @@ impl PartialOrd for BatchKnnCandidate {
...
+                .columns()
+                .iter()
+                .map(|col| col.get_array_memory_size())
+                .sum::<usize>()
+        })
+        .sum()
+}
+
+#[cfg(test)]
+mod batch_knn_memory_tests {
```
BubbleCal: I guess we don't need these memory tests
probably only test that if the query doesn't select vector column, then the batch shouldn't contain it

BubbleCal (changes requested):

rust/lance/src/io/exec/knn.rs
```diff
@@ -397,13 +504,23 @@ impl KNNVectorDistanceExec {
...
-                    let row_index = row_index as u32;
-                    let candidate = BatchKnnCandidate {
-                        query_index,
-                        distance,
-                        row_id,
-                        batch: batch.clone(),
-                        row_index,
+                    let candidate = if retain_vector {
+                        let row_index = row_index as u32;
+                        let vector_row = Self::take_vector_row(&batch, &column, row_index)?;
```
BubbleCal: Don't copy the vector unless it will be pushed into the candidate heap

BubbleCal (changes requested):

rust/lance/src/io/exec/knn.rs
```diff
@@ -646,20 +764,49 @@ impl ExecutionPlan for KNNVectorDistanceExec {
 }
 
 #[derive(Clone)]
-struct BatchKnnCandidate {
-    query_index: i32,
-    distance: f32,
-    row_id: u64,
-    batch: RecordBatch,
-    row_index: u32,
+enum BatchKnnCandidate {
```
BubbleCal: keep it a struct and make only the diff fields into enum

BubbleCal (changes requested):

rust/lance/src/io/exec/knn.rs
LeoReeYang: minimize the diff and make a nested enum inside

rust/lance/src/io/exec/knn.rs
LeoReeYang: there will be a `would_enter_heap` check ahead and a slice for vector when necessary

rust/lance/src/io/exec/knn.rs
LeoReeYang: removed

rust/lance/src/io/exec/knn.rs
LeoReeYang: slim batch will be pinned when necessary

rust/lance/src/io/exec/knn.rs
LeoReeYang: would introducing `Array::slice` be a better choice？

rust/lance/src/io/exec/knn.rs
LeoReeYang: only vector column will be dropped now

rust/lance/src/io/exec/knn.rs
```diff
@@ -650,16 +796,44 @@ struct BatchKnnCandidate {
...
     distance: f32,
     row_id: u64,
-    batch: RecordBatch,
-    row_index: u32,
+    extra: BatchKnnExtra,
+}
+
+#[derive(Clone)]
+enum BatchKnnExtra {
+    RowIdOnly,
```
BubbleCal, line 1022: Can you add a test which selects row_id and row_addr but no vectors?
I think it can't work with the current implementation

LeoReeYang: test added

BubbleCal (approved):

PR #6952, opened 2026-05-27
feat(index): describe indices for fragment batches

hfutatzhanghb:
## Summary

Add `DatasetIndexExt::describe_indices_for_fragments` in Rust and expose it through `LanceDataset.describe_indices_for_fragments` in Python.

The new API filters optional index names and returns only index segments whose fragment bitmap is fully covered by the requested fragment ids. This lets callers describe disjoint fragment batches and concatenate the results without double-counting segment row counts or file sizes.

Rust tests cover subset segment selection, crossing-segment exclusion, full-fragment inclusion, and index-name filtering.

## Why

Distributed index workflows may need to inspect index metadata by fragment batch. Existing `describe_indices()` reports whole logical indices, which can double-count segment-level metadata when batches are processed independently.

## Validation

- `cargo fmt --all`


hfutatzhanghb:
Hi, @Xuanwo . Could you please help review this PR when have free time? Thanks very much.

PR #6954, opened 2026-05-27
fix(encoding): use block sampling in sample_is_near_unique to fix dict encoding on sorted data

### LINKED ISSUE #6941: String compression: Lance is 5-350x larger than Parquet on low-cardinality columns
**Problem**

When a string column has very low cardinality (e.g., 5,000 unique values across 
14 million rows, with data sorted by that column), Lance v2.0/2.1 produces files 
**350x larger** than Parquet with the same gzip compression. Lance v2.2 improves 
this to ~5x, but the gap remains

**Root Cause**

The v2 encoder (`lance-encoding/src/compression.rs`) handles string columns via 
`build_variable_width_compressor()` which uses **FSST + Binary encoding per row** — 
there is no automatic cardinality detection and no column-wide dictionary+RLE for 
string columns.

By contrast, `build_fixed_width_compressor()` for integer columns **does** have 
RLE via `try_rle_for_mini_block()`.

Parquet automatically detects low cardinality → builds a column-wide dictionary → 
encodes dictionary indices with RLE/Bit-Packing → applies gzip.  This is a 2-level 
optimization (dict dedup + RLE) that Lance v2 lacks for strings.

**Interestingly, Lance v1 (`previous/encoder.rs`) DID have this optimization:**
`check_dict_encoding()` used HyperLogLog to estimate cardinality and auto-switched 
to dictionary encoding for string columns with `< 100 unique values. This was 
removed in the v2 rewrite.

zhangyue19921010:
closes https://github.com/lance-format/lance/issues/6941

### DIFF
--- a/rust/lance-encoding/src/encodings/logical/primitive.rs
+++ b/rust/lance-encoding/src/encodings/logical/primitive.rs
@@ -5090,9 +5090,7 @@ impl PrimitiveStructuralEncoder {
     /// Probe whether a page looks near-unique before attempting dictionary encoding.
     ///
     /// Uses block sampling: takes consecutive samples from evenly-distributed blocks
-    /// across the page. This is robust against sorted data patterns (where stride
-    /// sampling fails because the step size exceeds run length) while maintaining
-    /// O(max_samples) cost.
+    /// across the page. This is robust against sorted data patterns
     fn sample_is_near_unique(
         data_block: &DataBlock,
         max_samples: usize,
@@ -5100,9 +5098,7 @@ impl PrimitiveStructuralEncoder {
     ) -> Option<bool> {
         use std::collections::HashSet;
 
-        // Split the budget into this many consecutive blocks. Block sampling keeps
-        // each sample run >= run_length, which avoids the stride-aliases-with-runs
-        // failure mode that uniform stride sampling has on sorted low-cardinality data.
+        // Split the budget into this many consecutive blocks.
         const NUM_SAMPLE_BLOCKS: usize = 32;
         // Below this many samples the ratio is too noisy to trust; treat as not near-unique.
         const MIN_RELIABLE_SAMPLES: usize = 1024;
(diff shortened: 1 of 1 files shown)


rust/lance-encoding/src/encodings/logical/primitive.rs
```diff
@@ -5109,71 +5113,80 @@ impl PrimitiveStructuralEncoder {
...
-                let ratio = unique.len() as f64 / sample_count as f64;
-                Some(sample_count >= 1024 && ratio >= unique_ratio_threshold)
+                unique.len() as f64 / actual_samples as f64
             }
-            _ => Some(false),
-        }
+            _ => return Some(false),
+        };
+
+        Some(actual_samples >= MIN_RELIABLE_SAMPLES && ratio >= unique_ratio_threshold)
```
Xuanwo, line 5189: The new block sampler only checks near-unique ratio, so sorted high-cardinality data with short duplicate runs is treated as dictionary-worthy and can trigger an expensive full-page dictionary probe before aborting.

rust/lance-encoding/src/encodings/logical/primitive.rs
```diff
@@ -7177,6 +7190,36 @@ mod tests {
...
+
+        let mut values = Vec::with_capacity(num_values as usize);
+        for i in 0..cardinality {
+            for _ in 0..run_length {
+                values.push(format!("value_{:016}", i));
+            }
+        }
+        let block = DataBlock::from_array(Arc::new(StringArray::from(values)) as ArrayRef);
+
+        let result = PrimitiveStructuralEncoder::sample_is_near_unique(&block, sample_count, 0.98);
```
Xuanwo, line 7215: The new test only asserts the private sampling helper, so it would not catch a regression where the public dictionary decision or final page layout still skips dictionary encoding for sorted low-cardinality strings.

zhangyue19921010:
A more elegant solution has been merged https://github.com/lance-format/lance/pull/7159 
Close this PR now.
Also Thanks for your time! @Xuanwo

PR #6956, opened 2026-05-27
feat: add ICU FTS tokenizer

Xuanwo:
Adds an optional ICU4X-based FTS tokenizer available as `base_tokenizer="icu"`, with bundled segmenter data so mixed-language text can be indexed without an external language model. The feature is wired through `lance-tokenizer`, `lance-index`, and the Python package, with docs, lockfiles, license metadata, and regression coverage updated accordingly.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3864,6 +3864,21 @@ dependencies = [
 ]
 
 [[package]]
+name = "icu_locale"
+version = "2.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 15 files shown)


BubbleCal (approved):
Nice to have this!
Maybe we can parameterize some existing tests to make them run with both `icu` and `simple` tokenizers, to see whether these tests still get passed.
If all pass, I think we can probably make `icu` the default

Xuanwo:
> Maybe we can parameterize some existing tests to make them run with both `icu` and `simple` tokenizers, to see whether these tests still get passed.
> If all pass, I think we can probably make `icu` the default

Seems to be a nice idea, will do as a follow-up PR

PR #6958, opened 2026-05-27
perf(index): reduce describe indices fragment scans

hfutatzhanghb:
## Summary

Optimize `describe_indices()` row-count aggregation by building a single `fragment_id -> logical_rows` map from manifest metadata and then iterating each segment's `RoaringBitmap` directly.

This avoids calling `Dataset::get_fragments()` inside the per-segment loop and adds trace-level instrumentation for the number of dataset fragments, segment fragment references, missing fragment references, and computed indexed rows.

## Results
We use the same test envirnment to compare the optimize result.  See below:

before optimization (425 seconds):
```
[2026-05-27T21:39:10] dataset_describe_indices describe_indices count=1 seconds=425.149
```

after optimization（0.701 seconds）
```
[2026-05-28T10:40:20] dataset_describe_indices describe_indices count=1 seconds=0.701
```

## Why

The previous implementation computed `rows_indexed` by looping over every dataset fragment for every index segment:

```text
for segment in index_segments:
    for fragment in all_dataset_fragments:
        if segment.fragment_bitmap.contains(fragment.id):
            rows_indexed += fragment.logical_rows
```

That makes the work proportional to `num_index_segments * total_dataset_fragments`. On a large table, for example 10,000 index segments and 1,000,000 fragments, this can turn a metadata-only description into billions of membership checks even when each segment covers only a small fragment bitmap.

There was also avoidable allocation in the inner loop. `Dataset::get_fragments()` clones the dataset into an `Arc` and clones each manifest fragment into a `FileFragment` wrapper. Because it was called once per segment, those clones and wrappers were rebuilt repeatedly while describing one logical index.

The new path builds fragment row metadata once with `Dataset::iter_fragments()` and then performs work proportional to:

```text
total_dataset_fragments + sum(segment.fragment_bitmap.cardinality)
```

For lance-ray style scheduling, this keeps the driver-side metadata planning step much lighter before slicing work across workers.

## Validation

- `cargo fmt --all`
- `cargo test -p lance test_describe_indices_total_size -- --nocapture`
- `git diff --check`

### DIFF
--- a/rust/lance/src/index.rs
+++ b/rust/lance/src/index.rs
@@ -664,7 +664,6 @@ impl IndexDescriptionImpl {
         }
 
         let details = IndexDetails(index_details.clone());
-        let mut rows_indexed = 0;
 
         let index_type = if details.is_vector() {
             derive_vector_index_type(index_details)
@@ -682,18 +681,42 @@ impl IndexDescriptionImpl {
                 .unwrap_or_else(|_| "Unknown".to_string())
         };
 
+        let mut fragment_rows = HashMap::with_capacity(dataset.manifest.fragments.len());
+        for fragment in dataset.iter_fragments() {
+            fragment_rows.insert(
+                fragment.id as u32,
+                fragment_logical_rows_from_metadata(fragment)?,
+            );
+        }
+        let mut rows_indexed = 0;
+        let mut indexed_fragment_refs = 0u64;
+        let mut missing_fragment_refs = 0u64;
+
         for shard in &segments {
             let fragment_bitmap = shard
             .fragment_bitmap
             .as_ref()
             .ok_or_else(|| Error::index("Fragment bitmap is required for index description.  This index must be retrained to support this method.".to_string()))?;
 
-            for fragment in dataset.get_fragments() {
-                if fragment_bitmap.contains(fragment.id() as u32) {
-                    rows_indexed += fragment.fast_logical_rows()? as u64;
+            indexed_fragment_refs += fragment_bitmap.len();
+            for fragment_id in fragment_bitmap.iter() {
+                if let Some(fragment_rows) = fragment_rows.get(&fragment_id) {
+                    rows_indexed += *fragment_rows;
+                } else {
+                    missing_fragment_refs += 1;
                 }
             }
         }
+        tracing::trace!(
+            index_name = name.as_str(),
+            index_type = index_type.as_str(),
+            segment_count = segments.len(),
+            dataset_fragment_count = fragment_rows.len(),
+            indexed_fragment_refs,
+            missing_fragment_refs,
+            rows_indexed,
+            "described index row coverage from fragment metadata"
+        );
 
         Ok(Self {
             name,
@@ -704,6 +727,15 @@ impl IndexDescriptionImpl {
             rows_indexed,
         })
     }
+}
+
+fn fragment_logical_rows_from_metadata(fragment: &Fragment) -> Result<u64> {
+    fragment.num_rows().map(|rows| rows as u64).ok_or_else(|| {
+        Error::internal(format!(
+            "Index description requires physical row count and deletion count in fragment metadata. Fragment id: {}",
+            fragment.id
+        ))
+    })
 }
 
 impl IndexDescription for IndexDescriptionImpl {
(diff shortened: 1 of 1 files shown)


hfutatzhanghb:
@Xuanwo @BubbleCal @jiaoew1991 Hi, could you please help review this PR when free? Thanks!

rust/lance/src/index.rs
```diff
@@ -682,18 +681,42 @@ impl IndexDescriptionImpl {
                 .unwrap_or_else(|_| "Unknown".to_string())
         };
 
+        let mut fragment_rows = HashMap::with_capacity(dataset.manifest.fragments.len());
+        for fragment in dataset.iter_fragments() {
+            fragment_rows.insert(
+                fragment.id as u32,
+                fragment_logical_rows_from_metadata(fragment)?,
```
LuciferYang, line 688: The `?` operator propagates `Error::internal` if any fragment in the manifest lacks complete row metadata (`physical_rows` or `num_deleted_rows`). This iterates *all* fragments upfront, including those never referenced by any index bitmap. The old code only called `fast_logical_rows()` for fragments that appeared in the bitmap, so fragments with incomplete metadata were tolerated as long as they were not indexed.

wjones127: Realistically, any modern Lance dataset will have `physical_rows` or `num_deleted_rows`. Generally, though, we try to use the async version to be backwards compatible. If it's not too much trouble, I'd prefer we kept with that.

> This iterates _all_ fragments upfront, including those never referenced by any index bitmap. The old code only called `fast_logical_rows()` for fragments that appeared in the bitmap, so fragments with incomplete metadata were tolerated as long as they were not indexed.

TBH I don't this this is a huge deal. Realistically, you will have indexes covering most if not all fragments, so you'll use all or most of the fragment row counts. I think the speed benefit of collecting them into a contiguous allocation (versus computing on demand) might outweigh the cost of slight over-computation.

rust/lance/src/index.rs
```diff
@@ -706,6 +729,15 @@ impl IndexDescriptionImpl {
     }
 }
 
+fn fragment_logical_rows_from_metadata(fragment: &Fragment) -> Result<u64> {
+    fragment.num_rows().map(|rows| rows as u64).ok_or_else(|| {
+        Error::internal(format!(
+            "Index description requires physical row count and deletion count in fragment metadata. Fragment id: {}",
```
wjones127: suggestion: if these statistics are missing, we can automatically run the migration on the write. Worth telling the users that.
```suggestion
            "Index description requires physical row count and deletion count in fragment metadata. To add this, make a write to this dataset using this library version. Fragment id: {}",
```

rust/lance/src/index.rs
wjones127: We could keep this is as-is if we do the error message change I suggest in: https://github.com/lance-format/lance/pull/6958/changes#r3314011721

hfutatzhanghb: @wjones127 Thanks for your valuable suggestion. Have applied.

hfutatzhanghb:
Hi, @wjones127 @LuciferYang  After apply this optimization, we reduce describe_indices cost time significantly from 400+ second to 0.7 seconds within 500B rows lance dataset .  I think this is a very useful PR.

rust/lance/src/index.rs
```diff
@@ -6884,6 +6916,7 @@ mod tests {
 
         let desc = &descriptions[0];
         assert_eq!(desc.name(), "test_idx");
+        assert_eq!(desc.rows_indexed(), 4);
```
LuciferYang, line 6919: The single `assert_eq!(desc.rows_indexed(), 4)` exercises one fragment, no deletions, no missing-bitmap entries. Per the project standard ("Include multi-fragment scenarios for dataset operations"), at least one of the following should be covered, ideally in a separate `#[tokio::test]` to keep the existing test focused:

- Multiple appended fragments — verifies the HashMap aggregation sums across fragments.
- A fragment with deletions — verifies `num_rows()` returns `physical_rows - num_deleted_rows`, not raw physical rows.
- A stale segment bitmap (e.g., after compaction) referencing a fragment that no longer exists in the manifest — exercises the new `missing_fragment_refs` branch.

hfutatzhanghb: @LuciferYang Thanks for suggestions. Will add unit tests laterly.

hfutatzhanghb:
@LuciferYang @wjones127 Have added unit tests. please check again. Thanks.

LuciferYang (approved):
+1, LGTM

jiaoew1991 (approved):

wjones127 (approved):

PR #6961, opened 2026-05-27
feat(namespace-dir): implement rename_table for DirectoryNamespace

### LINKED ISSUE #6922: feat(DirectoryNamespace): implement update_table and delete_from_table DML methods
## Summary

`DirectoryNamespace` (the filesystem-backed `LanceNamespace` implementation in
`rust/lance-namespace-impls/src/dir.rs`) currently falls back to the trait's
default `not_supported` implementation for two DML operations:

- `update_table(UpdateTableRequest) -> UpdateTableResponse`
- `delete_from_table(DeleteFromTableRequest) -> DeleteFromTableResponse`

Both methods are already wired through the Java side
(`DirectoryNamespace.java` `updateTable` / `deleteFromTable`) and exposed by the
`lance-namespace-reqwest-client` model, but calling them today returns
`OperationNotSupported`. This issue tracks adding the Rust-side implementation
so users of `dir://` namespaces can perform row-level updates and deletes
through the namespace API, on par with what `Dataset::update` /
`Dataset::delete` already provide for direct dataset access.

## Motivation

- DML is a baseline capability for any catalog; without it the directory
  namespace cannot back interactive workflows or downstream systems
  (e.g. GooseFS / LanceDB integrations) that drive Lance via the namespace
  trait.
- The underlying primitives (`UpdateBuilder`, `Dataset::delete`) are stable
  and well-tested in `lance` core — exposing them through the namespace layer
  is a small, isolated piece of work with no schema/metadata changes.
- Closes the first two unsupported operations listed in
  [`LANCE_UNSUPPORTED_OPS.md` §1.2](../docs/LANCE_UNSUPPORTED_OPS.md).

### `update_table`

| Field | Source / Validation |
|-------|---------------------|
| `id` | resolved via existing namespace path resolution; missing → `TableNotFound` |
| `updates: List<List<String>>` | each inner list must be `[column, expression]`; non-empty, no duplicate columns, no empty column names → `InvalidInput` |
| `predicate?` | optional; if present, applied through `UpdateBuilder::update_where` |
| Response.`updated_rows` | `metrics.rows_updated` from `UpdateBuilder::execute()` |
| Response.`version` | new dataset version after commit |

### `delete_from_table`

| Field | Source / Validation |
|-------|---------------------|
| `id` | resolved as above |
| `predicate` | **required**, non-empty / non-whitespace → otherwise `InvalidInput` |
| Response.`version` | new dataset version after `Dataset::delete` |

> Note: the namespace response model has no `deleted_rows` field, so the
> `DeleteResult.num_deleted_rows` returned by `lance` core is intentionally
> dropped at the namespace boundary.

### Error mapping

| Cause | Mapped error |
|-------|--------------|
| Predicate / expression parse failure (`DataFusionError`) | `InvalidInput` |
| `lance_core::Error::InvalidInput` from delete | `InvalidInput` |
| Missing table | `TableNotFound` (via existing `load_dataset`) |

## Test plan

Add unit tests in `rust/lance-namespace-impls/src/dir.rs#tests`:

- [x] `test_update_full_table` — full-table update bumps version by 1
- [x] `test_update_with_predicate` — only matching rows updated
- [x] `test_update_invalid_expression` — surfaces `InvalidInput`
- [x] `test_update_duplicate_columns` — `InvalidInput`
- [x] `test_delete_with_predicate` — row count decreases, version + 1
- [x] `test_delete_empty_predicate` — `InvalidInput`
- [x] `test_delete_table_not_found` — `TableNotFound`

CI gate: `cargo fmt --check`, `cargo clippy -D warnings`,
`cargo test -p lance-namespace-impls --lib dir::tests` all green.

XuQianJin-Stars:
Closes #6922

## Summary

`DirectoryNamespace` (the filesystem-backed `LanceNamespace` implementation in `rust/lance-namespace-impls/src/dir.rs`) currently falls back to the trait's default *"Not supported"* implementation for two DML operations. This PR wires them up to Lance core, on par with what `Dataset::update` / `Dataset::delete` already provide for direct dataset access.

| Method | Backed by |
|--------|-----------|
| `update_table(UpdateTableRequest) -> UpdateTableResponse` | `lance::dataset::write::UpdateBuilder` |
| `delete_from_table(DeleteFromTableRequest) -> DeleteFromTableResponse` | `Dataset::delete(predicate)` |

Both methods are already exposed on the Java side (`DirectoryNamespace.java` `updateTable` / `deleteFromTable`) and by `lance-namespace-request-client`, so this PR is **Rust-only** and unblocks the existing client surface.

### `update_table`

- Builds an `UpdateBuilder` by setting each `[column, expression]` pair from the request, applies the optional `predicate` via `update_where`, and returns the actual `rows_updated` and the new dataset `version`.
- Validates the `updates` list **up-front** (non-empty list, exactly two elements per pair, non-empty / duplicate-free column names) so callers get a clean `InvalidInput` instead of an opaque DataFusion parse error.

## Summary by CodeRabbit

* **New Features**
  * Added atomic, manifest-aware table rename support, including consistent updates to table locations across supported storage modes.
  * Added safeguards to validate rename requests (non-empty names/ids), reject cross-namespace renames, and prevent destination collisions.

* **Bug Fixes**
  * Improved correctness for concurrent rename scenarios by detecting stale source/destination state and failing with a clear conflict signal.
  * Ensures manifest metadata is preserved/updated appropriately during renames (including hybrid-mode location updates).

* **Tests**
  * Expanded rename test coverage for success, collisions, missing sources, invalid inputs, and cross-namespace rejection.

### DIFF
--- a/rust/lance-namespace-impls/src/dir.rs
+++ b/rust/lance-namespace-impls/src/dir.rs
@@ -71,10 +71,11 @@ use lance_namespace::models::{
     ListTableIndicesResponse, ListTableTagsRequest, ListTableTagsResponse,
     ListTableVersionsRequest, ListTableVersionsResponse, ListTablesRequest, ListTablesResponse,
     MergeInsertIntoTableRequest, MergeInsertIntoTableResponse, NamespaceExistsRequest,
-    QueryTableRequest, QueryTableRequestColumns, QueryTableRequestVector, RestoreTableRequest,
-    RestoreTableResponse, TableExistsRequest, TableVersion, TagContents as ModelTagContents,
-    UpdateTableRequest, UpdateTableResponse, UpdateTableSchemaMetadataRequest,
-    UpdateTableSchemaMetadataResponse, UpdateTableTagRequest, UpdateTableTagResponse,
+    QueryTableRequest, QueryTableRequestColumns, QueryTableRequestVector, RenameTableRequest,
+    RenameTableResponse, RestoreTableRequest, RestoreTableResponse, TableExistsRequest,
+    TableVersion, TagContents as ModelTagContents, UpdateTableRequest, UpdateTableResponse,
+    UpdateTableSchemaMetadataRequest, UpdateTableSchemaMetadataResponse, UpdateTableTagRequest,
+    UpdateTableTagResponse,
 };
 
 use lance_core::{Error, Result, box_error};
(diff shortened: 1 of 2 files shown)


rust/lance-namespace-impls/src/dir.rs
```diff
@@ -4285,6 +4286,146 @@ impl LanceNamespace for DirectoryNamespace {
...
+                message: format!("Destination table '{}' already exists", new_table_name),
+            }
+            .into());
+        }
+
+        // For manifest mode with cross-namespace rename, delegate to manifest
+        if let Some(ref manifest_ns) = self.manifest_ns
+            && request.new_namespace_id.is_some()
+        {
+            return manifest_ns.rename_table(request).await;
```
yanghua: This will return `Err(Error::not_supported("rename_table not implemented"))`?

rust/lance-namespace-impls/src/dir.rs
```diff
@@ -4285,6 +4286,146 @@ impl LanceNamespace for DirectoryNamespace {
...
+                    ),
+                })
+            })?;
+
+        // If manifest mode is enabled, update the manifest to reflect the rename
+        if let Some(ref manifest_ns) = self.manifest_ns {
+            // Drop the old table entry from manifest
+            let mut drop_req = DropTableRequest::new();
+            drop_req.id = request.id.clone();
+            let _ = manifest_ns.drop_table(drop_req).await;
```
yanghua: We need to check if it would fail.

rust/lance-namespace-impls/src/dir.rs
```diff
@@ -4285,6 +4286,146 @@ impl LanceNamespace for DirectoryNamespace {
...
+
+            // Register the new table in manifest
+            let relative_location = format!("{}.lance", new_table_name);
+            let _ = manifest_ns
+                .register_table(new_table_name, relative_location)
+                .await;
```
yanghua: ditto

rust/lance-namespace-impls/src/dir.rs
```diff
@@ -4285,6 +4286,146 @@ impl LanceNamespace for DirectoryNamespace {
...
+            let mut exists_req = TableExistsRequest::new();
+            exists_req.id = dest_id.clone();
+            self.table_exists(exists_req).await.is_ok()
+        };
+        if dest_exists {
+            return Err(NamespaceError::TableAlreadyExists {
+                message: format!("Destination table '{}' already exists", new_table_name),
+            }
+            .into());
+        }
```
yanghua: We should not trust all the exceptions that come from `table_exists` to mean that the table does not exist.

rust/lance-namespace-impls/src/dir.rs
```diff
@@ -4285,6 +4286,146 @@ impl LanceNamespace for DirectoryNamespace {
...
+            .remove_dir_all(source_path)
+            .await
+            .map_err(|e| {
+                lance_core::Error::from(NamespaceError::Internal {
+                    message: format!(
+                        "Failed to remove source table '{}' after rename: {:?}",
+                        source_name, e
+                    ),
+                })
+            })?;
```
yanghua: We may need to consider a `rollback` mechanism, here.

yanghua (commented):
Left some comments.

rust/lance-namespace-impls/src/dir.rs
```diff
@@ -4285,6 +4286,215 @@ impl LanceNamespace for DirectoryNamespace {
...
+        if request.new_namespace_id.is_some() {
+            return Err(NamespaceError::Unsupported {
+                message: "Cross-namespace rename is not supported by DirectoryNamespace"
+                    .to_string(),
+            }
+            .into());
+        }
+
+        // Verify source table exists (works with both manifest and dir-listing modes)
+        let _source_uri = self.resolve_table_location(&request.id).await?;
```
jackye1995: I’m still worried this path is not using the table’s actual physical location. We resolve the source URI here, but then discard it and rebuild the source path below from `request.id.last()` with `self.table_path(source_name)`. That only works for root tables stored as `<name>.lance`. Manifest-backed tables in child namespaces, manifest-only mode with `dir_listing_enabled(false)`, and custom registered locations can live under the manifest-recorded location instead. For those cases this can copy from the wrong prefix and then update the manifest to point at a different location. Could we derive the source path from the resolved/manifest location instead?

rust/lance-namespace-impls/src/dir.rs
```diff
@@ -4285,6 +4286,215 @@ impl LanceNamespace for DirectoryNamespace {
...
+            }
+        }
+
+        // If manifest mode is enabled, update the manifest to reflect the
+        // rename before we remove the source directory. Any failure here must
+        // roll back both the manifest changes we already made and the copied
+        // files, so the whole operation appears atomic to callers.
+        if let Some(ref manifest_ns) = self.manifest_ns {
+            // 1. Register the new table entry in the manifest.
+            let relative_location = format!("{}.lance", new_table_name);
```
jackye1995: This manifest registration still hard-codes the renamed table as a root table at `<new_table_name>.lance`. `ManifestNamespace::register_table(name, location)` builds the object id from an empty namespace and stores no metadata, so a rename of `['workspace', 'table']` would become a root-level `new_table_name`, and any existing table properties are lost. I think rename needs to preserve the original namespace and manifest metadata, or this logic should live in `ManifestNamespace` where it can update the manifest row consistently.

rust/lance-namespace-impls/src/dir.rs
```diff
@@ -4285,6 +4286,215 @@ impl LanceNamespace for DirectoryNamespace {
...
+                }));
+            }
+        }
+
+        // Remove the source directory. At this point the manifest (if any)
+        // already points at the new table, so if this cleanup fails we log a
+        // warning rather than failing the rename: the logical state is
+        // correct and only stale physical files remain, which can be cleaned
+        // up out-of-band.
+        if let Err(e) = self.object_store.remove_dir_all(source_path).await {
```
jackye1995: Returning `Ok` after source deletion fails is not safe in directory-only mode. When `manifest_ns` is `None`, the old table is still discoverable by directory listing, so a successful rename response can leave both the old and new table visible. I’d return the cleanup error or roll back the destination in the directory-only path; the best-effort cleanup behavior only makes sense if the manifest has already made the source table unreachable.

jackye1995 (changes requested):
Thanks for updating this. A few of the earlier issues are addressed now, but I think there are still some correctness problems around manifest-backed tables that should be fixed before merge. I only reviewed the source and did not build or run tests.

rust/lance-namespace-impls/src/dir.rs
```diff
@@ -4285,6 +4286,211 @@ impl LanceNamespace for DirectoryNamespace {
...
+        }
+
+        // Decide which mode this rename runs in, matching the routing used by
+        // `describe_table` / `create_table` / `table_exists` so that the row we
+        // read back after a rename is the one we actually mutated.
+        let is_root_level = source_id.len() == 1;
+        let skip_manifest_for_root = self.dir_listing_enabled
+            && is_root_level
+            && !self.dir_listing_to_manifest_migration_enabled;
+        let use_manifest = self.manifest_ns.is_some() && !skip_manifest_for_root;
```
jackye1995: Thanks for splitting the manifest-backed and directory-listing paths. I think there is still a hybrid-mode consistency issue here. In the default setup, `manifest_ns` exists and `create_table` still writes a manifest row for root tables, but this root/default rename path sets `use_manifest=false` and only moves the physical directory.

After renaming `a` to `b`, the manifest still has object id `a` pointing at `a.lance`, and there is no manifest row for `b`. That can show up later through manifest-delegated APIs or migration mode; for example, `drop_table(b)` delegates to the manifest and won’t find the renamed table, and migration/listing can expose both the stale old manifest name and the new directory name. Could we keep the manifest row in sync when `manifest_ns.is_some()` even if this branch is responsible for the physical move?

rust/lance-namespace-impls/src/dir/manifest.rs
```diff
@@ -2327,6 +2327,97 @@ impl ManifestNamespace {
...
+
+        let metadata_json = Self::serialize_metadata(
+            table_info.metadata.as_ref(),
+            ObjectType::Table.as_str(),
+            &dest_object_id,
+        )?;
+
+        // 1. Insert the new row (fail if it already exists as a defensive check
+        //    against races with a concurrent create).
+        self.insert_into_manifest_with_metadata(
```
jackye1995: I think this manifest rename should be a single manifest rewrite. Right now it commits the destination insert first and deletes the source row in a second commit. If the process crashes after the insert but before the delete, the manifest can permanently contain both object ids pointing at the same table location. Could we replace the source row with the destination row in one manifest mutation instead?

jackye1995 (changes requested):
Thanks for the updates. The previous review comments look addressed, but I found two remaining correctness issues around keeping the manifest state consistent during rename.

rust/lance-namespace-impls/src/dir/manifest.rs
```diff
@@ -2327,6 +2418,93 @@ impl ManifestNamespace {
...
+        // pointing at the same physical location, because either the whole
+        // rewrite commits or none of it does.
+        let dest_object_id_for_mutation = dest_object_id.clone();
+        let source_object_id_for_mutation = source_object_id.clone();
+        self.rewrite_manifest("Failed to rename table entry", move || {
+            RenameManifestMutation {
+                dest: ManifestEntry {
+                    object_id: dest_object_id_for_mutation.clone(),
+                    object_type: ObjectType::Table,
+                    location: Some(table_info.location.clone()),
```
jackye1995: Thanks for continuing to tighten this up. I think there’s one remaining detail in the hybrid root-table path: after the directory branch moves `old.lance` to `new.lance`, `rename_table_entry` renames the manifest object id but preserves `table_info.location`, so the new manifest row still points at `old.lance`, which was just removed. Root reads skip the manifest here, so the current tests can still pass, but manifest-routed operations like `drop_table(new)` or migration mode will use the stale/deleted location. Could this sync path update the manifest location to `new.lance` as part of the rename, or use a helper specifically for physical root-table renames?

rust/lance-namespace-impls/src/dir.rs
```diff
@@ -4285,6 +4286,259 @@ impl LanceNamespace for DirectoryNamespace {
...
+                message: "new_table_name cannot be empty".to_string(),
+            }
+            .into());
+        }
+
+        // Cross-namespace rename cannot be performed by the directory backend
+        // itself, and the manifest backend does not implement `rename_table`
+        // either, so bail out with a clear `Unsupported` error rather than
+        // returning the generic "not implemented" from the trait default.
+        if request.new_namespace_id.is_some() {
```
jackye1995: Small API compatibility issue: this rejects any request with `new_namespace_id`, but the field represents the target namespace, and omitting it just means “same namespace”. Some clients may send the current namespace explicitly, and the existing disabled REST tests do that. Could we only reject when `new_namespace_id` differs from the source namespace, and allow the explicit same-namespace case?

rust/lance-namespace-impls/src/dir/manifest.rs
```diff
@@ -583,6 +583,97 @@ pub struct NamespaceInfo {
...
+        index_data: &mut ManifestIndexAccumulator,
+    ) -> Result<()> {
+        if self.dest_collision {
+            // The source row was dropped (if present) above; we cannot append
+            // the destination row because that would overwrite the concurrent
+            // entry, and `process_existing_row` already returned the error
+            // that will surface to the caller.
+            return Ok(());
+        }
+        output.append(
```
jackye1995, line 674: Nice improvement making the manifest rename a single rewrite. One concurrency edge remains: `source_found` is set during the rewrite, but it isn’t enforced before appending the destination row. Since `rename_table_entry` reads `table_info` before entering `rewrite_manifest`, another writer can drop or deregister the source before this rewrite sees it; then this mutation can recreate the entry under the new name from stale metadata. Could `append_rows` or `finish` fail/no-op when `source_found` is false, so the destination is appended only when the source was actually removed in the same rewrite?

jackye1995 (changes requested):
Thanks for sticking with this through a few rounds. The latest changes are definitely closer; I found a few remaining correctness edges around the manifest sync and rename semantics.

XuQianJin-Stars:
hi @jackye1995 Thanks for the catch, I've updated the implementation to replace the source row with the destination row in a single manifest mutation, so there's no longer an intermediate state where both object ids point at the same table location. Could you take another look when you have a chance?

PR #6963, opened 2026-05-27
fix(java): rename DatasetDelta JNI bindings to match Java declarations

### LINKED ISSUE #6962: DatasetDelta JNI bindings throw UnsatisfiedLinkError
The Java side of `DatasetDelta` declares its native methods with the `nativeXxx` convention:

```java
private native List<Transaction> nativeListTransactions();
private native void nativeGetInsertedRows(long streamAddress);
private native void nativeGetUpdatedRows(long streamAddress);
```

But the Rust exports in `java/lance-jni/src/delta.rs` omit the `native` prefix:

```
Java_org_lance_delta_DatasetDelta_listTransactions
Java_org_lance_delta_DatasetDelta_getInsertedRows
Java_org_lance_delta_DatasetDelta_getUpdatedRows
```

The JVM looks up `Java_<package>_<class>_<javaMethodName>`, so the names never line up. Calling any of the three throws `UnsatisfiedLinkError` against the published jar (reproduced on 7.0.0-rc.1). `releaseNativeDelta` is the only one that happens to match.

`DeltaTest` was supposed to cover this surface, but every assertion is wrapped in `catch (UnsatisfiedLinkError) { Assumptions.assumeTrue(false, ...) }`, so CI silently skipped the broken tests rather than failing them.

LuciferYang:
# Summary

Fixes #6962. The Rust exports in `java/lance-jni/src/delta.rs` were named without the `native` prefix that the Java side declares, so every call into the three `DatasetDelta` native methods failed with `UnsatisfiedLinkError`. This renames the exports and fixes two unrelated DeltaTest fixture problems that the JNI bug had been silently masking.

# Changes

- Rename `Java_org_lance_delta_DatasetDelta_listTransactions` / `getInsertedRows` / `getUpdatedRows` in `java/lance-jni/src/delta.rs` to add the `native` prefix. `releaseNativeDelta` was already correct.
- Switch both DeltaTest cases from `memory://` URIs (which do not persist commits across separate `Dataset.write()` calls) to `@TempDir`-backed file URIs, matching every other test in the module.
- Add `.enableStableRowIds(true)` to the v1 write in `testInsertedRowsComparedAgainst` so the expected `_row_created_at_version` / `_row_last_updated_at_version` row-lineage columns appear in the delta output.
- Drop the now-dead `catch (UnsatisfiedLinkError) -> Assumptions.assumeTrue(false, ...)` skip pattern and the `Assumptions` import.
- Extract a `writeBatch()` helper to deduplicate the Arrow batch serialization between the two tests.

# Testing

- `./mvnw test -Dtest=DeltaTest -DskipNativeBuild=true` — both tests now pass instead of silently skipping.
- `nm -gU java/lance-jni/target/debug/liblance_jni.dylib | grep DatasetDelta` — confirms the four expected exports (`nativeListTransactions`, `nativeGetInsertedRows`, `nativeGetUpdatedRows`, `releaseNativeDelta`).

### DIFF
--- a/java/lance-jni/src/delta.rs
+++ b/java/lance-jni/src/delta.rs
@@ -109,7 +109,7 @@ fn inner_native_build<'local>(
 }
 
 #[unsafe(no_mangle)]
-pub extern "system" fn Java_org_lance_delta_DatasetDelta_listTransactions<'local>(
+pub extern "system" fn Java_org_lance_delta_DatasetDelta_nativeListTransactions<'local>(
     mut env: JNIEnv<'local>,
     j_delta: JObject<'local>,
 ) -> JObject<'local> {
@@ -140,7 +140,7 @@ fn inner_list_transactions<'local>(
 }
 
 #[unsafe(no_mangle)]
-pub extern "system" fn Java_org_lance_delta_DatasetDelta_getInsertedRows<'local>(
+pub extern "system" fn Java_org_lance_delta_DatasetDelta_nativeGetInsertedRows<'local>(
     mut env: JNIEnv<'local>,
     j_delta: JObject<'local>,
     stream_addr: jlong,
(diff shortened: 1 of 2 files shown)


LuciferYang:
cc @Xuanwo @hamersaw @jackye1995 FYI

yanghua (approved):
LGTM

PR #6563, opened 2026-04-17
fix: apply fragment bitmap allow-list to index search results

westonpace:
## Summary

- Restrict index search results to only fragments in the index's `fragment_bitmap`, filtering out stale entries from fragments whose data changed but whose index was not rewritten
- Apply the allow-list in `DatasetPreFilter` (vector/FTS queries) and `MapIndexExec` (scalar index queries used by merge_insert)
- Add regression tests for stale index entries after `DataReplacement` and partial-schema `merge_insert` (BTree, FTS, and vector indices)

## Test plan

- [x] `test_data_replacement_populates_invalidated_bitmap` — DataReplacement removes fragment from bitmap
- [x] `test_fts_stale_entries_after_data_replacement` — stale FTS entries blocked after DataReplacement
- [x] `test_vector_index_after_data_replacement` — stale vector index entries blocked after DataReplacement
- [x] `test_partial_merge_insert_stale_index_ambiguous` — repeated partial merge_insert doesn't cause ambiguous errors
- [x] `test_partial_merge_insert_stale_index_fragment_not_exist` — stale btree entries don't cause fragment-not-found errors
- [x] `test_partial_merge_insert_stale_index_batch_size_mismatch` — stale btree entries don't cause batch size mismatch
- [x] `test_partial_merge_insert_stale_vector_index_duplicates` — no duplicate rows from stale vector index
- [x] `test_partial_merge_insert_stale_fts_index_duplicates` — no duplicate rows from stale FTS index
- [x] `test_compaction_after_invalidated_fragment` — compaction succeeds after fragment invalidation
- [x] All 17 existing prefilter tests pass

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #6965, opened 2026-05-27
fix: apply fragment-bitmap allow-list to stable-row-id deletion mask

### LINKED ISSUE #6877: MergeInsertBuilder produces spurious "Ambiguous merge inserts" against rows previously rewritten by merge_insert
## Summary

Two sequential `merge_insert` operations against the same target row (or set of rows) fail on the second call with:

```
Invalid user input: Ambiguous merge inserts are prohibited:
multiple source rows match the same target row on (id = "<key>")
```

even when the source batch has exactly one row per `<key>`. Reproduces on Lance 4.0.0 and 4.0.1 (byte-identical in this code path).

## Minimal repro

```rust
use arrow_array::{RecordBatch, StringArray, Int32Array};
use lance::dataset::{
    Dataset, WriteParams, WriteMode,
    MergeInsertBuilder, WhenMatched, WhenNotMatched,
};
use lance_index::{DatasetIndexExt, IndexType};
use lance_index::scalar::ScalarIndexParams;
use std::sync::Arc;

#[tokio::main]
async fn main() -> Result<(), Box<dyn std::error::Error>> {
        Field::new("id", DataType::Utf8, false),
        Field::new("value", DataType::Int32, false),
    ]));

    let initial = RecordBatch::try_new(schema.clone(), vec![
        Arc::new(StringArray::from(vec!["A", "B", "C"])),
        Arc::new(Int32Array::from(vec![1, 2, 3])),
    ])?;

    let uri = "memory://test"; // or a tmpdir path
    let mut ds = Dataset::write(
        Box::new(arrow_array::RecordBatchIterator::new(
            vec![Ok(initial)], schema.clone()
        )),
        uri,
        Some(WriteParams {
            mode: WriteMode::Overwrite,
            enable_stable_row_ids: true,
            ..Default::default()
        }),
    ).await?;

    // Build BTREE on \`id\` (mirrors what downstreams do on every commit).
    ds.create_index(
        &["id"],
        IndexType::Scalar,
        None,
        &ScalarIndexParams::default(),
        false,
    ).await?;

    // First merge_insert: update A's value from 1 to 11.
    let update_a = RecordBatch::try_new(schema.clone(), vec![
        Arc::new(StringArray::from(vec!["A"])),
        Arc::new(Int32Array::from(vec![11])),
    ])?;

    let mut builder = MergeInsertBuilder::try_new(
        Arc::new(ds.clone()),
        vec!["id".to_string()],
    )?;
    builder.when_matched(WhenMatched::UpdateAll);
    builder.when_not_matched(WhenNotMatched::DoNothing);
    let job = builder.try_build()?;
    let reader = arrow_array::RecordBatchIterator::new(
        vec![Ok(update_a.clone())], schema.clone(),
    );
    let (ds, _) = job.execute(
        lance_datafusion::utils::reader_to_stream(Box::new(reader))
    ).await?;
    // OK so far.

    // Second merge_insert: update A's value from 11 to 22.
    let update_a_again = RecordBatch::try_new(schema.clone(), vec![
        Arc::new(StringArray::from(vec!["A"])),
        Arc::new(Int32Array::from(vec![22])),
    ])?;

    let mut builder = MergeInsertBuilder::try_new(
        Arc::new(ds.clone()),
        vec!["id".to_string()],
    )?;
    builder.when_matched(WhenMatched::UpdateAll);
    builder.when_not_matched(WhenNotMatched::DoNothing);
    let job = builder.try_build()?;
    let reader = arrow_array::RecordBatchIterator::new(
        vec![Ok(update_a_again)], schema,
    );
    let (_, _) = job.execute(
        lance_datafusion::utils::reader_to_stream(Box::new(reader))
    ).await?;
    // Errors here:
    // "Ambiguous merge inserts are prohibited: multiple source rows
    //  match the same target row on (id = \"A\")"

    Ok(())
}
```

## Workarounds that suppress the symptom

Either of these on the second `MergeInsertBuilder` is sufficient (cross-validated downstream — each alone suppresses the bug; we ship both as defense in depth):

```rust
builder.use_index(false);
// OR
builder.source_dedupe_behavior(SourceDedupeBehavior::FirstSeen);
```

## Internal analysis (best guess; please correct)

Two Lance internals appear to be involved; we have not fully disambiguated which is causal:

1. **Indexed-scan join path.** `create_indexed_scan_joined_stream` (`src/dataset/write/merge_insert.rs:616+`) appears to UNION a BTREE-on-`id` lookup with a full scan of unindexed fragments. After the first `merge_insert` rewrites the matched row, the BTREE's entry still points at the now-tombstoned slot in the original fragment, while the rewritten row lives in a new (unindexed) fragment. Both sides of the UNION surface the same stable `_rowid` (since `enable_stable_row_ids = true`).
2. **Source-dedup `HashSet`.** `processed_row_ids: Mutex<HashSet<u64>>` (`src/dataset/write/merge_insert.rs:2099`) reports the spurious duplicate. Setting `SourceDedupeBehavior::FirstSeen` silences it; setting `use_index(false)` (which presumably bypasses the UNION entirely) also silences it.

These could be one bug (UNION causes the dedup HashSet to see duplicates) or two interacting bugs. We have not traced the actual data flow.

## Possible fixes

- DV-filter the indexed-scan side of the UNION so tombstoned `_rowid`s don't reach the join.
- Default `SourceDedupeBehavior::FirstSeen` for datasets with stable row IDs enabled — the dedup is then a no-op invariant of the row-id machinery.
- Extend Fragment Reuse Index (FRI) coverage to `Operation::Update` transactions (it currently tracks compaction `Rewrite` only — https://lance.org/format/table/index/system/frag_reuse/).

ragnorc:
## Summary

Closes #6877.

`do_create_deletion_mask_row_id` previously iterated every fragment in the dataset, so the resulting allow-list included stable row ids whose **current** physical home was outside the index's `fragment_bitmap`. On stable-row-id datasets this bypassed the allow-list added in #6563: `MapIndexExec` requests `create_restricted_deletion_mask`, but the stable-row-id branch at `prefilter.rs:270` ignored the restriction.

### Effect on merge_insert
After a `merge_insert UpdateAll` rewrites a row into a new fragment, a second merge_insert against the same key saw that row twice — once via the BTREE (which holds the row's stable row id and resolves to its new fragment through `TakeExec`) and once via the unindexed-fragments scan that also covered the new fragment. Both branches emitted the same `_rowid`, tripping the source-dedup HashSet:

```
Invalid user input: Ambiguous merge inserts are prohibited:
multiple source rows match the same target row on (id = "A").
```

### Effect on FTS
A `full_text_search` after a `merge_insert` returned each moved row twice, because the inverted index's hits went through the same prefilter and the post-filter scan picked up the new fragment. The new regression test `test_issue_6877_fts_no_duplicates_stable_row_ids` empirically fails with "30 unique out of 40 total" before this fix.

## Fix

- `prefilter.rs`: thread `restrict_to: Option<RoaringBitmap>` through `do_create_deletion_mask_row_id`. When set, only iterate fragments in the bitmap. The resulting allow-list now means "stable row ids whose current home is inside the restriction" — semantically equivalent to the non-stable-row-id branch.
- `prefilter.rs:270`: pass `Some(fragments)` when `restrict_to_fragments=true`, `None` otherwise.
- `caches.rs`: add `restrict_hash: Option<u64>` to `RowAddrMaskKey` so two consumers asking for different fragment subsets don't poison each other's cached mask.

## Mechanism, validated

Instrumentation on the unfixed code shows the BTREE returning stable row id `4` for the matched row, the deletion-mask allow-list containing `4` (because the new unindexed fragment was iterated globally), `mask & 4 = {4}` passing the filter, both branches emitting `_rowid=4`, dedup HashSet tripping. After the fix, `mask & 4 = {}` correctly excludes the BTREE side and only the unindexed-fragment scan delivers the row.

## Test plan

- [x] **Reproduces pre-fix**: stashed the prefilter + cache changes, re-ran the two new regression tests — both FAIL with the documented errors.
- [x] **Fixed post-fix**: restored the stash, both tests pass.
- [x] `cargo test -p lance --lib prefilter` — 18/18 pass (17 existing + 1 new).
- [x] `cargo test -p lance --lib merge_insert` — 145/145 pass (143 existing + 2 new).
- [x] `cargo fmt --all`

New regression tests:
- `test_restricted_deletion_mask_stable_row_id_honors_bitmap` — pins the prefilter contract: full / restricted-subset / empty-bitmap cases.
- `test_issue_6877_repeated_merge_insert_stable_row_ids` — mirrors the issue's exact repro: two sequential merge_insert UpdateAll against the same key on a stable-row-id dataset with a BTREE scalar index.
- `test_issue_6877_fts_no_duplicates_stable_row_ids` — companion FTS regression: an inverted index + merge_insert sequence that previously returned duplicate hits.

## Notes

- The hash on `RowAddrMaskKey` uses `DefaultHasher` over the bitmap's sorted u32s. The cache is process-local in-memory, so cross-version stability isn't required.

### DIFF
--- a/rust/lance/src/dataset/write/merge_insert.rs
+++ b/rust/lance/src/dataset/write/merge_insert.rs
@@ -8589,6 +8589,131 @@ MergeInsert: on=[id], when_matched=DoNothing, when_not_matched=InsertAll, when_n
         }
     }
 
+    // Regression test for GitHub issue #6877.
+    //
+    // Two sequential full-schema merge_insert UpdateAll calls against the same
+    // target row, on a dataset with stable_row_ids enabled and a BTREE scalar
+    // index on the join column, used to fail on the second call with
+    // "Ambiguous merge inserts are prohibited" — even though each call's
+    // source had exactly one row per key.
+    //
+    // Mechanism: with stable row ids the BTREE stores stable_row_ids (not
+    // physical addresses). After the first merge_insert, A's stable_row_id is
+    // preserved but its physical home moves to an unindexed fragment. The
+    // BTREE-side TakeExec resolves the stable_row_id to A's new location and
+    // emits a row; the unindexed-fragments scan also covers the new fragment
+    // and emits the same logical row. Both surface the same `_rowid`, so the
+    // merge_insert source-dedup HashSet sees a duplicate and aborts.
+    //
+    // Fix: thread `restrict_to_fragments` into `do_create_deletion_mask_row_id`
+    // so the allow-list only contains stable_row_ids whose current physical
+    // home is inside the index's fragment_bitmap.
+    #[tokio::test]
+    async fn test_issue_6877_repeated_merge_insert_stable_row_ids() {
+        use arrow_array::Int32Array;
+
+        let schema = Arc::new(Schema::new(vec![
+            Field::new("id", DataType::Utf8, false),
+            Field::new("value", DataType::Int32, false),
+        ]));
+
+        let initial = RecordBatch::try_new(
+            schema.clone(),
+            vec![
+                Arc::new(StringArray::from(vec!["A", "B", "C"])),
+                Arc::new(Int32Array::from(vec![1, 2, 3])),
+            ],
+        )
+        .unwrap();
+
+        let mut ds = Dataset::write(
+            Box::new(RecordBatchIterator::new([Ok(initial)], schema.clone())),
+            "memory://test_6877",
+            Some(WriteParams {
+                mode: WriteMode::Overwrite,
+                enable_stable_row_ids: true,
+                ..Default::default()
+            }),
+        )
+        .await
+        .unwrap();
+
+        ds.create_index(
+            &["id"],
+            IndexType::Scalar,
+            None,
+            &ScalarIndexParams::default(),
+            false,
+        )
+        .await
+        .unwrap();
+
+        // First merge_insert: A 1 -> 11.
+        let update_a = RecordBatch::try_new(
+            schema.clone(),
+            vec![
+                Arc::new(StringArray::from(vec!["A"])),
+                Arc::new(Int32Array::from(vec![11])),
+            ],
+        )
+        .unwrap();
+        let (ds, _) = MergeInsertBuilder::try_new(Arc::new(ds), vec!["id".to_string()])
+            .unwrap()
+            .when_matched(WhenMatched::UpdateAll)
+            .when_not_matched(WhenNotMatched::DoNothing)
+            .try_build()
+            .unwrap()
+            .execute_reader(Box::new(RecordBatchIterator::new(
+                [Ok(update_a)],
+                schema.clone(),
+            )))
+            .await
+            .unwrap();
+
+        // Second merge_insert: A 11 -> 22. Used to fail before the fix.
+        let update_a_again = RecordBatch::try_new(
+            schema.clone(),
+            vec![
+                Arc::new(StringArray::from(vec!["A"])),
+                Arc::new(Int32Array::from(vec![22])),
+            ],
+        )
+        .unwrap();
+        let (ds, _) = MergeInsertBuilder::try_new(ds, vec!["id".to_string()])
+            .unwrap()
+            .when_matched(WhenMatched::UpdateAll)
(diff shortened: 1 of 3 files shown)


Xuanwo (approved):

PR #6968, opened 2026-05-27
feat: make ICU the default FTS tokenizer

Xuanwo:
This changes the default native FTS tokenizer from `simple` to `icu` so new inverted indexes handle mixed-language text without requiring users to opt into multilingual tokenization. Legacy missing tokenizer metadata continues to resolve to `simple`, while ICU is now always compiled into the native tokenizer stack.

Benchmark summary from the 100M-row runs:

| Dataset | Retrieval impact | Build cost | Index size | Query latency |
|---|---|---:|---:|---|
| English-only | Recall unchanged | +15.4% | +0.6% | Common terms flat; rare terms slightly slower but still small |
| Mixed-language | ZH / JP / TH rare recall improves from 0.0 to 1.0 | +20.4% | +25.7% | EN / FR flat; multilingual rare queries roughly flat |

The key tradeoff is that ICU has modest overhead for English-only data, but it fixes default recall for unspaced CJK, Japanese, and Thai text.

<details>
<summary>Detailed benchmark numbers</summary>

### English-only 100M rows

| Metric | simple | icu | Difference |
|---|---:|---:|---:|
| Build time | 15.31s | 17.66s | +15.4% |
| Index size delta | 964.0MB | 969.7MB | +0.6% |
| Common-term latency | 2370.1ms / 1179.4ms | 2372.5ms / 1188.8ms | Flat |
| Rare-term recall | 1.0 / 1.0 | 1.0 / 1.0 | No change |
| Rare-term latency | 11.7ms / 14.2ms | 23.3ms / 17.7ms | Slightly slower |

### Mixed-language 100M rows

| Metric | simple | icu | Difference |
|---|---:|---:|---:|
| Build time | 33.35s | 40.16s | +20.4% |
| Index size delta | 800.1MB | 1005.7MB | +25.7% |
| EN common latency | 1026.7ms | 953.8ms | Slightly faster |
| FR common latency | 945.9ms | 946.1ms | Flat |
| CJK common query | 0 rows / 1.1ms | 10 rows / 944.1ms | `simple` misses |
| ZH / JP / TH rare recall | 0.0 / 0.0 / 0.0 | 1.0 / 1.0 / 1.0 | ICU recovers recall |
| FR / EN rare recall | 1.0 / 1.0 | 1.0 / 1.0 | No change |
| Rare-query latency | 13.7-24.6ms | 11.1-23.1ms | Roughly flat |

</details>

### DIFF
--- a/docs/src/format/index/scalar/fts.md
+++ b/docs/src/format/index/scalar/fts.md
@@ -56,7 +56,7 @@ The metadata file contains JSON-serialized configuration and partition informati
 
 | Field               | Type    | Default   | Description                                                    |
 |---------------------|---------|-----------|----------------------------------------------------------------|
-| `base_tokenizer`    | String  | "simple"  | Base tokenizer type (see Tokenizers section)                   |
+| `base_tokenizer`    | String  | "icu"     | Base tokenizer type (see Tokenizers section)                   |
 | `language`          | String  | "English" | Language for stemming and stop words                           |
 | `with_position`     | Boolean | false     | Store term positions for phrase queries (increases index size) |
 | `max_token_length`  | UInt32? | None      | Maximum token length (tokens longer than this are removed)     |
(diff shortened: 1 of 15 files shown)


westonpace (approved):
Seems like good rationale.  The only thing that makes me slightly hesitant is having `default_base_tokenizer` depend on a feature flag since it might be confusing.

Still, it's in the default feature list, and it's always going to be on for wheels / pylance, so I guess the vast majority of users will never be clearing this flag.

Xuanwo:
> The only thing that makes me slightly hesitant is having `default_base_tokenizer` depend on a feature flag since it might be confusing.

I'm open to just remove this feature and make it always enabled.

PR #6974, opened 2026-05-28
feat(namespace-dir): implement alter_transaction for DirectoryNamespace

### LINKED ISSUE #6973: feat(namespace-dir): implement alter_transaction for DirectoryNamespace
## Summary

The `alter_transaction` method in the `LanceNamespace` trait is not yet implemented for `DirectoryNamespace`. It currently falls back to the default trait implementation which returns "not supported".

## Motivation

The `alter_transaction` API is needed to allow clients to:
- Update the status of an existing transaction (e.g., mark as `Canceled`, `Failed`, `Running`)
- Set or unset custom properties on a transaction for metadata tracking

This is required for full LanceNamespace trait coverage in the DirectoryNamespace implementation, which is used by COS/GooseFS connectors.

## Expected Behavior

`DirectoryNamespace::alter_transaction` should:

1. **Resolve the transaction** by parsing the request ID (table ID + transaction identifier)
2. **Process actions atomically**, supporting:
   - `SetStatusAction` — change transaction status to one of: `Queued`, `Running`, `Succeeded`, `Failed`, `Canceled` (case-insensitive)
   - `SetPropertyAction` — set a transaction property with modes: `Overwrite` (default), `Fail` (error if exists), `Skip` (no-op if exists)
   - `UnsetPropertyAction` — remove a transaction property with modes: `Skip` (default, no-op if missing), `Fail` (error if missing)
3. **Return** the final status and merged properties map

## Error Cases

- Missing or incomplete transaction ID → `InvalidInput`
- Transaction not found → `NotFound`
- Invalid status value → `InvalidInput`
- Property conflict in `Fail` mode → `ConcurrentModification`
- Missing property in `Fail` mode for unset → `InvalidInput`

## Related PRs

- PR: feat(namespace-dir): implement alter_transaction for DirectoryNamespace

XuQianJin-Stars:
Closes #6973

Implement the `alter_transaction` method for `DirectoryNamespace`, which
allows altering a transaction's status and properties. This completes one
of the remaining unimplemented methods in the `LanceNamespace` trait.

## Supported actions

- `SetStatusAction`: change transaction status (Queued / Running / Succeeded / Failed / Canceled)
- `SetPropertyAction`: set transaction properties with `Overwrite` / `Fail` / `Skip` modes
- `UnsetPropertyAction`: remove transaction properties with `Skip` / `Fail` modes

All actions in a request are applied atomically — either all succeed, or
the operation fails with an appropriate error.

## Persistence via sidecar

Since Lance embeds the `Transaction` into the manifest and treats it as
immutable, we cannot rewrite the transaction file in place. Previously,
`alter_transaction` only mutated properties in memory and returned them,
so subsequent `describe_transaction` / `alter_transaction` calls could
not observe the modifications.

To fix this, alterations are now persisted in a namespace-owned sidecar
file at:

```
{table}/_alter_transactions/{uuid}.json
```

This sidecar records the accumulated alterations (status, added /
overwritten properties, tombstoned properties). `describe_transaction`
and follow-up `alter_transaction` calls merge the sidecar on top of the
immutable transaction, so the altered state survives across calls.

## Reserved keys

`uuid`, `version`, `read_version`, `operation`, and `tag` are treated as
reserved keys that cannot be modified via `alter_transaction`, since they
are derived from the immutable `Transaction` metadata. They are still
surfaced in the response for the caller's convenience.

## Tests added

- `test_alter_transaction_set_status`
- `test_alter_transaction_set_property`
- `test_alter_transaction_set_property_fail_mode`
- `test_alter_transaction_unset_property`
- `test_alter_transaction_invalid_status`
- `test_alter_transaction_not_found`
- `test_alter_transaction_missing_id`

### DIFF
--- a/rust/lance-namespace-impls/Cargo.toml
+++ b/rust/lance-namespace-impls/Cargo.toml
@@ -13,8 +13,8 @@ rust-version.workspace = true
 
 [features]
 default = ["dir-aws", "dir-azure", "dir-gcp", "dir-oss", "dir-huggingface"]
-rest = ["dep:reqwest"]
-rest-adapter = ["dep:axum", "dep:tower", "dep:tower-http"]
+rest = ["dep:reqwest", "dep:serde"]
+rest-adapter = ["dep:axum", "dep:tower", "dep:tower-http", "dep:serde"]
 # Cloud storage features for directory implementation - align with lance-io
 dir-gcp = ["lance-io/gcp", "lance/gcp"]
 dir-aws = ["lance-io/aws", "lance/aws"]
@@ -24,8 +24,8 @@ dir-huggingface = ["lance-io/huggingface", "lance/huggingface"]
 dir-goosefs = ["lance-io/goosefs", "lance/goosefs"]
 # Credential vending features
 credential-vendor-aws = ["dep:aws-sdk-sts", "dep:aws-config", "dep:sha2", "dep:base64"]
-credential-vendor-gcp = ["dep:reqwest", "dep:sha2", "dep:base64", "dep:ring", "dep:rustls-pki-types"]
-credential-vendor-azure = ["dep:reqwest", "dep:sha2", "dep:base64", "dep:chrono", "dep:hmac", "dep:quick-xml"]
+credential-vendor-gcp = ["dep:reqwest", "dep:serde", "dep:sha2", "dep:base64", "dep:ring", "dep:rustls-pki-types"]
+credential-vendor-azure = ["dep:reqwest", "dep:serde", "dep:sha2", "dep:base64", "dep:chrono", "dep:hmac", "dep:quick-xml"]
 
 [dependencies]
 lance-namespace.workspace = true
@@ -58,12 +58,12 @@ datafusion-physical-plan = { workspace = true }
 axum = { workspace = true, optional = true }
 tower = { workspace = true, optional = true }
 tower-http = { workspace = true, optional = true, features = ["trace", "cors", "normalize-path"] }
+serde = { workspace = true, optional = true }
 
 # Common dependencies
 async-trait.workspace = true
 bytes.workspace = true
 tokio = { workspace = true, features = ["sync", "full"] }
-serde = { workspace = true, features = ["derive"] }
 serde_json = { workspace = true }
 futures.workspace = true
 log.workspace = true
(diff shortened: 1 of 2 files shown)


rust/lance-namespace-impls/src/dir.rs
```diff
@@ -3764,6 +3765,165 @@ impl LanceNamespace for DirectoryNamespace {
         Ok(Self::transaction_response(version, &transaction))
     }
 
+    async fn alter_transaction(
```
yanghua, line 3937: Did you persist or save the changes?

XuQianJin-Stars: > Did you persist or save the changes?

I updated it.

rust/lance-namespace-impls/Cargo.toml
```diff
@@ -13,8 +13,8 @@ rust-version.workspace = true
 
 [features]
 default = ["dir-aws", "dir-azure", "dir-gcp", "dir-oss", "dir-huggingface"]
-rest = ["dep:reqwest", "dep:serde"]
```
yanghua, line 16: Do we need to change this file?

rust/lance-namespace-impls/src/dir.rs
```diff
@@ -12968,4 +13293,322 @@ mod tests {
...
+        use lance_namespace::models::{
+            AlterTransactionAction, AlterTransactionRequest, AlterTransactionSetStatus,
+            DescribeTransactionRequest,
+        };
+
+        let (namespace, _temp_dir) = create_test_namespace().await;
+        create_scalar_table(&namespace, "users").await;
+        let transaction_id = create_scalar_index(&namespace, "users", "users_id_idx").await;
+
+        if let Some(txn_id) = transaction_id {
```
yanghua: This style may cause false positives if it is `None`. It would be better to use `expect(...)` or `assert!(transaction_id.is_some())`

rust/lance-namespace-impls/src/dir.rs
```diff
@@ -12968,4 +13293,322 @@ mod tests {
...
+                .await
+                .unwrap();
+            assert_eq!(response.status, "SUCCEEDED");
+            let props = response.properties.unwrap();
+            assert_eq!(props.get("custom_key"), Some(&"custom_value".to_string()));
+        }
+    }
+
+    #[tokio::test]
+    async fn test_alter_transaction_set_property_fail_mode() {
```
yanghua, line 13420: Using the reserved field `uuid` as the key results in zero coverage of the mode='Fail' branch.

yanghua (commented):
Left some comments.

yanghua (approved):
+1

PR #6837, opened 2026-05-19
feat(rust): support update by _row_id

yanghua:
In stable row ID mode, we can not update via `_row_id` currently. It throws:

```
Traceback (most recent call last):
  File "/data00/home/xxx/workspace/clientlance/read_lance.py", line 19, in <module>
    ds.update(dict(name_1 = '123'), where="_rowid=1")
  File "/data00/home/xxx/workspace/clientlance/.venv/lib/python3.11/site-packages/lance/dataset.py", line 2259, in update
    return self._ds.update(updates, where, conflict_retries, retry_timeout)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ValueError: Invalid user input: LanceError(Schema): Schema error: No field named _rowid. Valid fields are name_1, name_2, name_3., /home/runner/work/lance/lance/rust/lance-datafusion/src/planner.rs:929:20, /home/runner/work/lance/lance/rust/lance/src/dataset/write/update.rs:92:18
```

PR #6976, opened 2026-05-28
feat: support update by _row_id for python and java

yanghua:
The function of updating by `_row_id` has been supported via #6837 . This is a follow-up PR to expose the feature for Python and Java. The changes for the two upper modules are listed here:

* Python: added test cases;
* Java: supported the basic update API and also update by `_row_id` ;


Xuanwo (approved):
Thank you!

PR #6977, opened 2026-05-28
perf: parallel column decode for wide structs

WenDing-Y:
## Summary

When decoding a struct with 4+ child columns, spread the per-column decode
work across threads via `std::thread::scope`. Each column's page decoding is
independent CPU work and was previously done sequentially, leaving cores idle
on wide analytical tables.

## Motivation

`RepDefStructDecodeTask::decode()` iterates over child columns one at a time,
calling each child's `decode()` sequentially. For wide tables (common in
analytics workloads), this serializes CPU-bound decompression and
deserialization work that has no data dependency between columns.

## Approach

- Threshold: `PARALLEL_DECODE_MIN_CHILDREN = 4` — narrow tables are unaffected
- Uses `std::thread::scope` (zero dependencies, stable since Rust 1.63)
- Narrow tables (< 4 columns) take the existing sequential code path unchanged

## Benchmark (8 columns × 1M rows, M1 Pro)

|                | Sequential (before) | Parallel (after) | Improvement |
|----------------|---------------------|-------------------|-------------|
| **Latency**    | 7.32 ms             | 4.41 ms           | **-40%**    |
| **Throughput** | 5.34 GiB/s          | 8.86 GiB/s        | **+66%**    |

p = 0.00 (highly significant). Three independent runs confirmed stability.

## Changes

- `lance-encoding/src/encodings/logical/struct.rs` — parallelize
  `RepDefStructDecodeTask::decode()` when 4+ children
- `lance-encoding/benches/decoder.rs` — new `bench_decode_wide_struct` benchmark

## Testing

- ✅ 83/83 existing struct tests pass (no regressions)
- ✅ New `test_wide_struct_parallel_decode` added (6-column round-trip)
- ✅ `cargo bench -- decode_wide_struct` passes

### DIFF
--- a/rust/lance-encoding/Cargo.toml
+++ b/rust/lance-encoding/Cargo.toml
@@ -14,6 +14,7 @@ rust-version.workspace = true
 [dependencies]
 lance-arrow.workspace = true
 lance-core.workspace = true
+rayon.workspace = true
 arrow-arith.workspace = true
 arrow-array.workspace = true
 arrow-data.workspace = true
--- a/rust/lance-encoding/benches/decoder.rs
+++ b/rust/lance-encoding/benches/decoder.rs
@@ -3,7 +3,7 @@
 use std::{collections::HashMap, sync::Arc};
 
 use arrow_array::{RecordBatch, UInt32Array};
-use arrow_schema::{DataType, Field, Schema, TimeUnit};
+use arrow_schema::{DataType, Field, Fields, Schema, TimeUnit};
 use arrow_select::take::take;
 use criterion::{Criterion, criterion_group, criterion_main};
 use futures::StreamExt;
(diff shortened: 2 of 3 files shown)


rust/lance-encoding/src/encodings/logical/struct.rs
```diff
@@ -354,20 +354,45 @@ struct RepDefStructDecodeTask {
...
+        // When a struct has many child columns, each column's page decoding is
+        // independent CPU work.  Spreading it across threads improves scan
+        // throughput for wide analytical tables.
+        const PARALLEL_DECODE_MIN_CHILDREN: usize = 4;
+
+        let arrays = if children.len() >= PARALLEL_DECODE_MIN_CHILDREN {
+            std::thread::scope(|s| {
+                let handles: Vec<_> = children
+                    .into_iter()
+                    .map(|task| s.spawn(move || task.decode()))
```
Xuanwo: Maybe we can juse rayon for this?

Xuanwo (commented):

I have some concerns about this change. I understand that performing parallel decode can benefit a single end-to-end query, but what is the impact on overall performance? Most of our workload will have many queries running at the same time, so we need to consider not only latency but also the overall QPS.

Could you run some benchmarks to verify this? For example, measure the QPS on the same 100-million-row dataset.

Also, if we have some super-wide tables, such as those with 6,000 columns (which is a real case), what would happen?

rust/lance-encoding/src/encodings/logical/struct.rs
```diff
@@ -354,43 +355,50 @@ struct RepDefStructDecodeTask {
...
                 repdef: CompositeRepDefUnraveler::new(vec![]),
                 data_size: 0,
             });
         }
 
-        let arrays = self
-            .children
-            .into_iter()
+        let arrays: Vec<_> = children
+            .into_par_iter()
```
Xuanwo, line 374: This adds uncontrolled child-level Rayon fan-out on top of batch-level scan parallelism. `batch_readahead` and `LANCE_CPU_THREADS` no longer bound decode CPU usage, so concurrent scans or very wide structs can trade single-query latency for lower QPS and worse tail latency.

rust/lance-encoding/Cargo.toml
```diff
@@ -14,6 +14,7 @@ rust-version.workspace = true
 [dependencies]
 lance-arrow.workspace = true
 lance-core.workspace = true
+rayon.workspace = true
```
Xuanwo, line 17: The manifest adds a direct dependency without the corresponding lockfile update. Clean `--locked` CI or release builds can fail because the workspace package dependency list is stale.

rust/lance-encoding/src/encodings/logical/struct.rs
```diff
@@ -32,6 +32,7 @@ use itertools::Itertools;
 use lance_arrow::FieldExt;
 use lance_arrow::{deepcopy::deep_copy_nulls, r#struct::StructArrayExt};
 use lance_core::{Error, Result};
+use rayon::prelude::*;
```
Xuanwo, line 35: The PR is not rustfmt-clean; `cargo fmt --all --check` fails on the import order. This blocks the standard formatting check before tests can validate the change.

rust/lance-encoding/benches/decoder.rs
```diff
@@ -434,6 +434,67 @@ fn bench_decode_compressed(c: &mut Criterion) {
...
+///
+/// Encodes a struct with 8 columns (triggering the parallel decode path via
+/// `std::thread::scope`) and measures decode throughput in bytes/sec.
+fn bench_decode_wide_struct(c: &mut Criterion) {
+    let rt = tokio::runtime::Runtime::new().unwrap();
+    let mut group = c.benchmark_group("decode_wide_struct");
+
+    const NUM_ROWS: u64 = 1_000_000;
+
+    // 8 columns of diverse types to trigger parallel decode (threshold is 4)
```
Xuanwo, line 447: The benchmark still claims a 4-child threshold while the implementation parallelizes every non-empty struct. Narrow-table regressions are unmeasured, and the benchmark evidence no longer matches the behavior being merged.

PR #6978, opened 2026-05-28
perf(index): reduce HNSW build distance overhead

BubbleCal:
## Performance Improvement

HNSW index construction spends a large amount of time creating distance calculators from graph node ids and pruning graph neighbors. The hot path previously copied vectors or reconstructed PQ centroids when the query was already an indexed vector, and PQ `dist_between` rebuilt full centroid vectors for both endpoints.

This PR reduces that overhead by:

- Borrowing flat and SQ query vectors directly from storage when creating `dist_calculator_from_id`.
- Reusing owned candidate buffers in HNSW neighbor pruning and avoiding an unused backlink collection.
- Lazily building and sharing a PQ centroid pairwise distance table, then deriving per-query distance rows directly from PQ codes.
- Computing PQ code-to-code distance from the shared pairwise table instead of reconstructing centroid vectors.

### Benchmarks

GCP `c4-standard-16`, `RUSTFLAGS="-C target-cpu=native" cargo bench -p lance-index --bench hnsw -- create_hnsw --sample-size 10 --warm-up-time 1 --measurement-time 5`, 100k x 128 benchmark cases.

| Benchmark | main median | PR median | Change |
| --- | ---: | ---: | ---: |
| `create_hnsw(100000x128)` | 6.7919 s | 4.1944 s | 38.2% faster |
| `create_hnsw_sq(100000x128)` | 3.5711 s | 3.5269 s | 1.2% faster |
| `create_hnsw_pq(100000x128)` | 30.129 s | 13.457 s | 55.3% faster |

The PQ pairwise table is initialized lazily and shared across remapped storage instances. Criterion warmup pays the one-time initialization cost before measurement; production first use will pay that setup once per storage instance.

### Validation

- `cargo fmt --all`
- `cargo test -p lance-index vector::hnsw::builder::tests::test_builder_write_load --lib`
- `cargo test -p lance-index vector::sq::storage::tests::test_get_chunks --lib`
- `cargo test -p lance-index vector::pq::storage::tests::test_dist_calculator_from_id_matches_reconstructed_distance --lib`

### DIFF
--- a/rust/lance-index/src/vector/flat/storage.rs
+++ b/rust/lance-index/src/vector/flat/storage.rs
@@ -183,11 +183,7 @@ impl VectorStore for FlatFloatStorage {
     }
 
     fn dist_calculator_from_id(&self, id: u32) -> Self::DistanceCalculator<'_> {
-        Self::DistanceCalculator::new(
-            self.vectors.as_ref(),
-            self.vectors.value(id as usize),
-            self.distance_type,
-        )
+        Self::DistanceCalculator::new_from_id(self.vectors.as_ref(), id, self.distance_type)
     }
 }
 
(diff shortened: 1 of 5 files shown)


rust/lance-index/src/vector/flat/storage.rs
```diff
@@ -349,22 +345,33 @@ impl VectorStore for FlatBinStorage {
...
 pub struct FlatDistanceCal<'a, T: ArrowPrimitiveType> {
     vectors: &'a [T::Native],
-    query: Vec<T::Native>,
+    query: QueryVector<'a, T>,
     dimension: usize,
     #[allow(clippy::type_complexity)]
     distance_fn: fn(&[T::Native], &[T::Native]) -> f32,
 }
 
+enum QueryVector<'a, T: ArrowPrimitiveType> {
```
Xuanwo: What's difference with a `Cow`? I'm trying to understand why we need to write this enum by ourselves.

Xuanwo (approved):
Other changes LGTM

rust/lance-index/src/vector/flat/storage.rs
BubbleCal: changed to `Cow`

PR #6983, opened 2026-05-28
perf(fts): defer DocSet load until wand actually needs it

wkalt:
Make `InvertedPartition::load` defer the per-partition `DocSet` work
(row_id + num_tokens) until the wand walk actually needs it, instead
of materializing the entire DocSet up front.

### DIFF
--- a/rust/lance-index/src/scalar/inverted/index.rs
+++ b/rust/lance-index/src/scalar/inverted/index.rs
@@ -1341,6 +1341,10 @@ impl InvertedPartition {
     }
 
     #[instrument(level = "debug", skip_all)]
+    // Deferred-DocSet adds the `docs` param (caller materializes it) on top of
+    // the cross-partition `shared_threshold`, tipping this hot-path search fn
+    // one over the limit. Bundling args isn't worth the churn here.
+    #[allow(clippy::too_many_arguments)]
     pub fn bm25_search(
         &self,
         docs: &DocSet,
--- a/rust/lance-index/src/scalar/inverted/index.rs
+++ b/rust/lance-index/src/scalar/inverted/index.rs
@@ -4379,7 +4379,20 @@ impl DocSet {
         let batch = reader.read_range(0..reader.num_rows(), None).await?;
         let row_id_col = batch[ROW_ID].as_primitive::<datatypes::UInt64Type>();
         let num_tokens_col = batch[NUM_TOKEN_COL].as_primitive::<datatypes::UInt32Type>();
+        Self::from_columns(row_id_col, num_tokens_col, is_legacy, frag_reuse_index)
+    }
 
+    /// Build a `DocSet` from already-loaded `row_id` and `num_tokens`
+    /// arrow columns. Lets callers that have one column already in hand
+    /// (e.g. [`crate::scalar::inverted::lazy_docset::LazyDocSet`] after
+    /// `total_tokens_num` pre-fetched `num_tokens`) skip re-reading that
+    /// column.
+    pub fn from_columns(
+        row_id_col: &UInt64Array,
+        num_tokens_col: &arrow_array::UInt32Array,
+        is_legacy: bool,
+        frag_reuse_index: Option<Arc<FragReuseIndex>>,
+    ) -> Result<Self> {
         // for legacy format, the row id is doc id; sorting keeps binary search viable
         if is_legacy {
             let (row_ids, num_tokens): (Vec<_>, Vec<_>) = row_id_col
--- a/rust/lance-index/src/scalar/inverted/index.rs
+++ b/rust/lance-index/src/scalar/inverted/index.rs
@@ -1211,11 +1211,21 @@ impl InvertedPartition {
         let tokens = TokenSet::load(token_file, token_set_format).await?;
         let invert_list_file = store.open_index_file(&posting_file_path(id)).await?;
         let inverted_list = PostingListReader::try_new(invert_list_file, index_cache).await?;
-        // Defer the per-doc row_id/num_tokens read. Construction only
-        // calls reader.num_rows() (no IO); the bulk load happens on first
-        // scoring use, and partitions that never score skip it entirely.
-        let docs_file = store.open_index_file(&doc_file_path(id)).await?;
-        let docs = Arc::new(LazyDocSet::new(docs_file, false, frag_reuse_index));
+        // Defer the per-doc row_id/num_tokens read. Construction reads only
+        // the doc count (one footer read) and then drops the reader; the bulk
+        // load happens on first scoring use, re-opening the docs file on
+        // demand, and partitions that never score skip it entirely. Storing
+        // the store + path instead of an open reader keeps a cached partition
+        // from pinning a docs-file handle for its whole lifetime.
+        let docs_path = doc_file_path(id);
+        let num_docs = store.open_index_file(&docs_path).await?.num_rows();
+        let docs = Arc::new(LazyDocSet::new(
+            store.clone(),
+            docs_path,
+            num_docs,
+            false,
+            frag_reuse_index,
+        ));
 
         Ok(Self {
             id,
--- a/rust/lance-index/src/scalar/inverted/index.rs
+++ b/rust/lance-index/src/scalar/inverted/index.rs
@@ -1035,6 +1035,11 @@ impl InvertedIndex {
                 part.inverted_list
                     .prewarm_posting_lists(with_position)
                     .await?;
+                // Materialize the deferred DocSet too: prewarm's contract is
+                // that subsequent queries do no IO, so the per-doc row_ids /
+                // num_tokens must be resident, not lazily faulted in at query
+                // time. `ensure_loaded` opens, reads, and drops the reader.
+                part.docs.ensure_loaded().await?;
                 Result::Ok(())
             });
         stream::iter(prewarm_futures)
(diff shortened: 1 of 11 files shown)


wkalt:
Here are some results on different storage media for single partition, varying by dataset size (reddit comments):
```
   docs | DocSet   | shm main  | shm patched | shm x  | nvme main | nvme patched | nvme x | GCS main  | GCS patched | GCS x
  ------|----------|-----------|-------------|--------|-----------|--------------|--------|-----------|-------------|------
   1M   |  3.6 MB  |    571 ms |       66 ms |  8.6x  |    745 ms |        82 ms |  9.1x  |  3,859 ms |    3,520 ms |  1.10x
   2M   |  7.5 MB  |  1,177 ms |      118 ms | 10.0x  |  1,347 ms |       136 ms |  9.9x  |  4,396 ms |    3,901 ms |  1.13x
   5M   | 19.5 MB  |  3,100 ms |      252 ms | 12.3x  |  3,361 ms |       295 ms | 11.4x  |  7,314 ms |    4,678 ms |  1.56x
   10M  | 40   MB  |  6,301 ms |      477 ms | 13.2x  |  6,587 ms |       537 ms | 12.3x  | 11,433 ms |    6,375 ms |  1.79x
   25M  | 105  MB  | 16,377 ms |    1,093 ms | 15.0x  | 16,698 ms |     1,201 ms | 13.9x  | 25,170 ms |   10,132 ms |  2.48x
   50M  | 216  MB  | 33,771 ms |    2,092 ms | 16.2x  | 34,413 ms |     2,258 ms | 15.2x  | 50,107 ms |   19,845 ms |  2.52x
   100M | 560  MB  | 69,734 ms |    3,831 ms | 18.2x  | 69,556 ms |     4,140 ms | 16.8x  | 92,808 ms |   24,569 ms |  3.78x
```

Here is a fixed 100M docs, with LANCE_FTS_PARTITION_SIZE varying:
```
   LANCE_FTS_PARTITION_SIZE | partitions | docs/partition | nvme main  | nvme patched | nvme x  | GCS main   | GCS patched | GCS x
  --------------------------|------------|----------------|------------|--------------|---------|------------|-------------|------
   2048 (default)           |          1 |           100M |  72,160 ms |     3,903 ms | 18.49x  |  92,010 ms |   22,454 ms | 4.10x
   1024                     |          9 |            11M |  26,220 ms |     2,934 ms |  8.94x  |  53,006 ms |   28,601 ms | 1.85x
    256                     |         48 |             2M |  15,770 ms |     2,669 ms |  5.91x  |  43,926 ms |   29,684 ms | 1.48x
     64                     |        330 |           300k |   7,596 ms |     3,537 ms |  2.15x  |  40,107 ms |   43,061 ms | 0.93x
     16                     |      2,291 |            44k |   9,564 ms |    10,628 ms |  0.90x  |  93,016 ms |  142,290 ms | 0.65x
      4                     |     18,862 |           5.3k |  38,036 ms |    66,672 ms |  0.57x  | 441,792 ms |  477,156 ms | 0.93x
```

the second paste demonstrates that this causes regression for small partitions. This could hit us if the default were overridden, or if compaction were behind (lots of uncompacted small writes). I don't have a clear intuition yet for whether/where this would get bad.

wkalt:
partition-sweep
size-sweep
pictures from above tables

rust/lance-index/src/scalar/inverted/index.rs
```diff
@@ -516,18 +516,12 @@ impl InvertedIndex {
         })
     }
 
-    /// Build a single-segment [`MemBM25Scorer`] whose per-term IDF table
-    /// covers every token that the per-partition scoring loop will look
-    /// up. For fuzzy queries that means the union of Levenshtein
-    /// expansions, not just the raw query tokens — otherwise
-    /// `query_weight(expanded_token)` returns 0 and the BM25 contribution
-    /// of every expanded match is discarded.
```
westonpace, line 529: Why remove this comment?

rust/lance-index/src/scalar/inverted/index.rs
```diff
@@ -1085,7 +1147,10 @@ pub struct InvertedPartition {
     store: Arc<dyn IndexStore>,
     pub(crate) tokens: TokenSet,
     pub(crate) inverted_list: Arc<PostingListReader>,
-    pub(crate) docs: DocSet,
+    /// Per-doc row_id + num_tokens. Wrapped in `LazyDocSet` so partitions
+    /// that don't contribute hits to a query never pay the full-array
+    /// download. Scoring paths call `ensure_loaded` before walking wand.
+    pub(crate) docs: Arc<crate::scalar::inverted::lazy_docset::LazyDocSet>,
```
westonpace: Minor nit: let's not use full path for a struct member?  Looks kind of out-of-place

rust/lance-index/src/scalar/inverted/lazy_docset.rs
```diff
@@ -0,0 +1,320 @@
...
+/// Pre-materialized DocSet view -- no reader, no IO.
+pub struct LoadedDocSet {
+    docs: Arc<DocSet>,
+    num_rows: usize,
+    total_tokens: u64,
+}
+
+/// Reader-backed DocSet view that loads on demand and caches.
+pub struct DeferredDocSet {
+    reader: Arc<dyn IndexReader>,
```
westonpace: Is this something that gets cached?  Does it mean we will cache an open file handle?  I think we moved to try and avoid that at one point

rust/lance-index/src/scalar/inverted/lazy_docset.rs
```diff
@@ -0,0 +1,320 @@
...
+        }
+        if let Some(arr) = self.row_ids_col.get() {
+            return Ok(doc_ids.iter().map(|&d| arr.value(d as usize)).collect());
+        }
+        // Targeted per-doc reads -- the deferred path's whole point.
+        // Lance v2 coalesces nearby reads at the page level.
+        let mut row_ids = Vec::with_capacity(doc_ids.len());
+        for &d in doc_ids {
+            let d = d as usize;
+            let batch = self.reader.read_range(d..d + 1, Some(&[ROW_ID])).await?;
```
westonpace: Coalescing will only happen within a call.  We will not coalesce across multiple calls to `read_range`.  It should be easy enough to add a `read_ranges` which takes in a `Vec<Range>` though.

westonpace (changes requested):
CI is failing and I have a few suggestions.  I think adding a `read_ranges` is probably the most blocking.

rust/lance-index/src/scalar/inverted/index.rs
wkalt: thanks, reintroduced

rust/lance-index/src/scalar/inverted/lazy_docset.rs
wkalt: done

rust/lance-index/src/scalar/inverted/lazy_docset.rs
wkalt: updated

rust/lance-index/src/scalar/inverted/lazy_docset.rs
wkalt: Sorry I meant to put the comment above on your other comment.

This will cache an open file handle but this is an existing pattern in PostingListReader. Do you think there is a better way to go?

rust/lance-index/src/scalar/inverted/index.rs
wkalt: done

wkalt:
@westonpace thanks. Addressed/responded to your comments.

rust/lance-index/src/scalar/inverted/lazy_docset.rs
wkalt: maybe we should be opening/dropping at time of use. I can make that change.

rust/lance-index/src/scalar/lance_format.rs
```diff
@@ -223,6 +223,85 @@ impl IndexReader for current_reader::FileReader {
...
+            sorted_pos[oi] = sp;
+        }
+        let mut take_indices = Vec::with_capacity(total_rows as usize);
+        for &sp in &sorted_pos {
+            for k in 0..sorted_sizes[sp] {
+                take_indices.push(sorted_offsets[sp] + k);
+            }
+        }
+        let take_arr = arrow_array::UInt32Array::from(take_indices);
+        Ok(arrow_select::take::take_record_batch(&merged, &take_arr)?)
```
westonpace, line 318: Forcing a take here can be unfortunate.  Do you think the ranges are likely to come in out of order?  Or is this a prevenative measure?

If it isn't likely, or if there is a good chance they will be in-order, then you should add a fast-path for when the sort is not required.  It's a quick linear scan to see if the sort is required and if the values you are pulling back are large it can be worth it to avoid the take.

westonpace (commented):
One minor nit.  I'm not really aware of the code here in enough depth to verify the new algorithm would have better performance.  I'm assuming you've done some benchmarking though and observed improvements?  If so, then let's go for it.

westonpace (approved):
Sorry, meant that previous review to be an approving review

rust/lance-index/src/scalar/lance_format.rs
wkalt: it's handled above here: https://github.com/lance-format/lance/pull/6983/changes#diff-48e13331871cb7b43ee9af747224addfd4fe810d16656fa9ffc0f72b92ec6b5eR294-R296

PR #6984, opened 2026-05-28
fix(encoding): bound per-chunk rep/def overflow in too-sparse miniblock heuristic

Ali2Arslan:
## Summary
The previous `repdef_too_sparse_for_miniblock` heuristic compared the
page-average levels-per-value ratio against a max-sized chunk, so pages
whose empty/null lists cluster into a single mini-block chunk slipped
past the check and tripped the `u16` per-buffer size limit at encode
time (see issue #6681). It also used a 16KiB combined rep+def budget,
but the on-disk per-chunk size fields are independent `u16`s (~64KiB
each).
This change reworks the heuristic to:
- Walk the level stream to bound the worst-case level count for any
  chunk of `MAX_MINIBLOCK_VALUES` visible values (matching how
  `RepDefSlicer::slice_next` consumes specials, including leading and
  trailing ones).
- Check each rep/def buffer's estimated compressed size **independently**
  against the `u16` cap (with a 4KiB margin and a 2× ideal-bitpacking
  pessimism factor) instead of a combined budget.

### DIFF
--- a/rust/lance-encoding/src/encodings/logical/primitive.rs
+++ b/rust/lance-encoding/src/encodings/logical/primitive.rs
@@ -3835,16 +3835,17 @@ impl PrimitiveStructuralEncoder {
 
     /// Checks if the rep/def levels are too sparse for miniblock encoding.
     ///
-    /// Miniblock chunks are limited to ~32KiB total. Data can use up to ~16KiB,
-    /// leaving ~16KiB for both rep and def buffers combined. Each chunk has at most
-    /// MAX_MINIBLOCK_VALUES (4096) data values, but when data has many empty/null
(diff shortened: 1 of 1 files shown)


kushudai (approved):
To add a little color, we're using this patch to work around a crash in lance-spark that looked like
```
java.lang.RuntimeException: Encountered internal error. Repetition buffer size (66809 bytes) too large,
  .../lance-encoding/src/encodings/logical/primitive.rs:3912:21
```
because `u16::try_from(rep.data.len())` would overflow u16. `repdef_too_sparse_for_miniblock` was not consistently falling back to fullzip for some pathological cases

Ali2Arslan:
no longer necessary

PR #6985, opened 2026-05-28
feat: support Utf8View and BinaryView in encoding and filter coercion

xuanyu-z:
## Problem

Arrow's `Utf8View` and `BinaryView` are in-memory layout optimisations introduced in Arrow 2.0. They have no Lance on-disk representation — Lance stores everything as classic offset-based `Utf8`/`Binary`. DataFusion 43+ started producing these types from string functions such as `md5()`, which caused two distinct failures:

1. **Encoder panic**: `DataBlock::from_arrays` hit a `todo!()` for view types, crashing any write that involved a `Utf8View`/`BinaryView` column.
2. **Silent wrong filter results**: `safe_coerce_scalar` did not handle `Utf8View`/`BinaryView` literals, so filter expressions comparing a stored `Utf8` column against a `Utf8View` literal produced no matches.

Previous discussion and direction mentioned that we want to keep the lance physical type simple and do not want to introduce view type on disk: https://github.com/lance-format/lance/issues/5172#issuecomment-4119512129

### Type mapping (`datatypes.rs`, `field.rs`)
`LogicalType::try_from` and `Field::try_from` now map `Utf8View` → `Utf8` and `BinaryView` → `Binary` at the schema level, so view types are stored and tracked as their base equivalents.

### `is_binary_like` (`lance-arrow/lib.rs`)
`Utf8View` and `BinaryView` are included so any code that checks binary-like types (e.g. index building) handles them correctly.

### Write path — `SchemaAdapter` (`lance/src/dataset/utils.rs`)
`SchemaAdapter` now downcasts view-type columns before any encoder sees them. A `physical_field` helper is extracted as the single source of truth for the view→base mapping rule — both `downcast_view_columns` and `requires_physical_conversion` delegate to it, so adding a new view type in future requires changing exactly one place. `SchemaAdapter` is the correct interception point because it runs before both the legacy encoder (`LanceFileVersion::Legacy`) and the new structural encoder.

### Encoding fallback (`lance-encoding/src/data.rs`)
`DataBlock::from_arrays` handles `Utf8View`/`BinaryView` directly via `arrow_cast::cast` as a safety net for non-stream write paths. Performance note: string bytes are copied exactly once — `arrow_cast` gathers scattered view buffers into contiguous storage, and the subsequent `LanceBuffer::from(Buffer)` wrap is zero-copy.

### Filter coercion (`lance-datafusion/src/expr.rs`)
`safe_coerce_scalar` now coerces `Utf8View` ↔ `Utf8`/`LargeUtf8` and `BinaryView` ↔ `Binary`/`LargeBinary` in both directions, so filter literals produced by DataFusion are matched to the stored column type when resolving expressions.

## Testing

**Unit tests added:**
- `test_string_view_to_data_block` — multi-array `StringViewArray` with nulls and without, verifying correct offsets, data bytes, and null bitmap in the encoded block
- `test_binary_view_to_data_block` — smoke test confirming `BinaryView` dispatch reaches the correct encoder path
- `test_string_view_coerce` — all coercion directions for `Utf8View`↔`Utf8`/`LargeUtf8` and `BinaryView`↔`Binary`
- `test_resolve_utf8view_literal_against_utf8_column` — end-to-end filter resolution with a `Utf8View` literal against a `Utf8` column
- `test_view_types_stored_as_base_types` — schema-level mapping for both view types

**Manual end-to-end testing:**
- `SELECT md5(col)` queries (DataFusion produces `Utf8View` results) — no longer panic on write
- Filter expressions on string columns against `Utf8View` literals — now return correct results

## Future Work

- **Nested view types**: `SchemaAdapter` and `physical_field` only operate on top-level schema fields. Nested view types inside `Struct`, `List`, or `Map` are not currently handled. This is a known limitation.


rust/lance-arrow/src/lib.rs
```diff
@@ -66,6 +66,8 @@ pub trait DataTypeExt {
     /// assert!(DataType::Binary.is_binary_like());
     /// assert!(DataType::LargeUtf8.is_binary_like());
     /// assert!(DataType::LargeBinary.is_binary_like());
+    /// assert!(DataType::Utf8View.is_binary_like());
+    /// assert!(DataType::BinaryView.is_binary_like());
     /// assert!(!DataType::Int32.is_binary_like());
```
westonpace: I get this is pre-existing but this comment is just odd.  We should just remove it.

rust/lance-arrow/src/lib.rs
westonpace:
```suggestion
    /// Returns true if the data type is binary-like, such as Utf8, Binary, or the large and/or view variants.
```

westonpace (approved):
This is nice.  Casting on write is the correct approach.

One or two minor nits but good otherwise.

Potential follow-ups:
* Allow reading into the view type on read, this is more complex than it sounds because we'd really want the query engine to tell us which type it would prefer.  Still, most systems allow some kind of flag to toggle which type gets read into globally (e.g. https://docs.rs/datafusion/latest/datafusion/config/struct.ParquetOptions.html#structfield.schema_force_view_types)
* Support for list-view / large-list-view (if we don't support it already)

I'm marking approve but CI still needs to pass before we can merge.  Please ping me when that has been addressed.

rust/lance-arrow/src/lib.rs
xuanyu-z: Thx

rust/lance-arrow/src/lib.rs
xuanyu-z: Removed in the new revision

xuanyu-z:
> * Allow reading into the view type on read,

From me reading in the threads from Datafusion, the View types should be the default, at least for utf8? Are we considering even changing the disk format to be more aligned with View type and be the only on disk format, that may make encoding and decoding cheaper and may benefit the downstream engines?

xuanyu-z:
> * Support for list-view / large-list-view (if we don't support it already)

Will check that out

PR #6989, opened 2026-05-29
fix(encoding): handle u16 num_levels overflow in miniblock List encoding

touch-of-grey:
Surfaced while benchmarking mem_wal HNSW flush at scale (#6901 made flush
rebuild the persisted secondary indexes; the in-memory graph's
`__neighbors`/`__dists` columns are `List<UInt32>`/`List<Float32>`).
Flushing a generation larger than ~32k nodes at v2.1 panicked the
`lance-cpu` thread with `chunk_bytes <= max_chunk_size`; writing those
index files at v2.2 cleared the write panic but produced a corrupt read,
panicking the struct decoder with
`Mismatch in length (at offset=83300) expected 100 got 92` (counts vary
run-to-run with HNSW non-determinism). `HNSW::schema()` is shared with
every IVF_HNSW index, so this is a general codec correctness bug, not
specific to mem_wal.

Root cause: the miniblock structural codec stores each chunk's
`num_levels` as `u16` in the v2.2 on-disk header
(`encodings/logical/primitive.rs:467`). The HNSW persistence shape — a
dense level-0 prefix followed by ~6× as many empty higher-level rows —
has a healthy global levels/values ratio (~1.19), so
`repdef_too_sparse_for_miniblock` fell through to miniblock. The final
value chunk via `slice_rest` then absorbed all trailing empties (242 048
levels for a 40k-dense + 240k-empty case), `num_chunk_levels as u16`
silently truncated to 45 440, and the decoder short-read.

v2.2 is a stable on-disk format (`V2_2 < Next`), so the per-chunk header
cannot widen. Fix is code-only:

- **Heuristic.** Extend `repdef_too_sparse_for_miniblock` with
  `any_chunk_levels_overflow_u16`, a single linear pass over the def
  levels that mirrors the encoder's chunking (`MAX_MINIBLOCK_VALUES`
  visible values per non-last chunk; `slice_rest` for the final) and
  returns true if any chunk would carry more than `u16::MAX` levels.
  Routes affected shapes to fullzip through the existing dispatcher.
  Early-exits at `max_visible_level = None` and `def_levels.len() <=
  u16::MAX` keep dense / small-page paths zero-cost (encoder bench delta
  within criterion noise).
- **Codec safety net.** `as u16` becomes
  `u16::try_from(...).map_err(Error::invalid_input_source)?` so any
  future shape that slips past the heuristic surfaces a clear error
  rather than corrupting data.

A `log::warn!` records when the heuristic overrides an explicit
`STRUCTURAL_ENCODING_MINIBLOCK` request. New
`test_list_hnsw_shape_auto_routes_around_miniblock_overflow_v2_2`
reproduces the `Mismatch in length` panic on `HEAD~1` and round-trips
cleanly on this commit; existing `test_sparse_large_string_list`
parametrization still passes both encodings.

**Follow-ups (not in this PR):** (a) a real value chunk straddling the
simulation's 4096-value boundary can split an empty cluster across two
sim chunks; the codec safety net catches that as a clean error rather
than corruption — tightening to a sliding window is follow-up. (b)
`List<Dictionary<...>>` on the too-sparse path hits a pre-existing
`unreachable!` in `encode_full_zip`; not introduced here but the
heuristic broadens the surface.

### DIFF
--- a/rust/lance-encoding/src/encodings/logical/list.rs
+++ b/rust/lance-encoding/src/encodings/logical/list.rs
@@ -242,7 +242,9 @@ mod tests {
     use arrow_array::{
         Array, ArrayRef, BooleanArray, DictionaryArray, LargeStringArray, ListArray, StructArray,
         UInt8Array, UInt64Array,
-        builder::{Int32Builder, Int64Builder, LargeListBuilder, ListBuilder, StringBuilder},
+        builder::{
+            Int32Builder, Int64Builder, LargeListBuilder, ListBuilder, StringBuilder, UInt32Builder,
+        },
     };
 
     use arrow_buffer::{BooleanBuffer, NullBuffer, OffsetBuffer, ScalarBuffer};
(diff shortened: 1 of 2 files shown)


Xuanwo (approved):
Thank you!

PR #6992, opened 2026-05-29
fix: preserve zero-length buffers in binary copy compaction

### LINKED ISSUE #6991: Binary Copy Compaction Drops Zero-Length Buffers In V2.0
### Error
```
thread 'lance_background_thread' (2727090) panicked at xxxx/src/previous/encodings/physical.rs:39:41:
index out of bounds: the len is 1 but the index is 1

thread '<unnamed>' (2726968) panicked at xxxx/lance/rust/lance-encoding/src/decoder.rs:1998:30:
called Result::unwrap() on an Err value: JoinError::Panic(Id(1186576), "index out of bounds: the len is 1 but the index is 1", ...)
  File "/root/jinglun/compact/create_index.py", line 128, in <module>
    raise SystemExit(main())
                     ^^^^^^
  File "/root/jinglun/compact/create_index.py", line 122, in main
    create_indices(ds)
  File "/root/jinglun/compact/create_index.py", line 112, in create_indices
    ds.create_scalar_index(
  File "/opt/conda/lib/python3.11/site-packages/lance/dataset.py", line 3026, in create_scalar_index
    self._ds.create_index([column], index_type, name, replace, train, None, kwargs)
pyo3_runtime.PanicException: called Result::unwrap() on an Err value: JoinError::Panic(Id(1186576), "index out of bounds: the len is 1 but the index is 1", ...)
```
### Repro Shape
1. Write a v2.0 Lance dataset with a binary/string column that can produce zero-length data buffers, for example empty strings.
2. Run compaction with Binary Copy.
3. Create a scalar index on that column.
4. Decode panics while resolving a legacy page buffer index.

### Root Cause
Legacy v2.0 binary/string encoding can reference multiple page buffers by index. Empty strings may produce a zero-length bytes buffer, but the buffer slot is still valid and must be preserved in page metadata.

Binary Copy currently turns every source buffer into an I/O range:
```
offset..offset + size
```

For zero-length buffers this becomes an empty range. The I/O scheduler does not return a corresponding Bytes item for that empty range, so Binary Copy writes fewer buffer_offsets_and_sizes entries than the original page had.

After compaction:

encoding metadata still references buffer index 1
rewritten page metadata only contains buffer index 0

### Fix
Binary Copy should skip zero-length buffers for I/O, but still preserve their metadata slot by writing (current_pos, 0) into the rewritten page/column buffer metadata.

This keeps buffer indices stable after compaction.

zhangyue19921010:
closes: https://github.com/lance-format/lance/issues/6991

### DIFF
--- a/rust/lance/src/dataset/optimize/binary_copy.rs
+++ b/rust/lance/src/dataset/optimize/binary_copy.rs
@@ -362,10 +362,12 @@ pub async fn rewrite_files_binary_copy(
                         // Therefore `page_index - batch_pages + local_idx` yields the exact
                         // source page we are currently materializing, allowing us to access
                         // its metadata (encoding, row count, buffers) for the new page entry.
-                        let page =
-                            &src_column_info.page_infos[page_index - batch_pages + local_idx];
+                        let page_idx = page_index - batch_pages + local_idx;
+                        let page = &src_column_info.page_infos[page_idx];
                         let mut new_offsets = Vec::with_capacity(*buffer_count);
-                        for (_, size) in page.buffer_offsets_and_sizes.iter() {
+                        for (buffer_idx, (_, size)) in
+                            page.buffer_offsets_and_sizes.iter().enumerate()
+                        {
                             let writer = current_writer.as_mut().unwrap().as_mut();
                             current_pos =
                                 apply_alignment_padding(writer, current_pos, version).await?;
@@ -374,9 +376,10 @@ pub async fn rewrite_files_binary_copy(
                                 new_offsets.push((start, 0));
                             } else {
                                 let bytes = bytes_iter.next().ok_or_else(|| {
-                                    Error::execution(
-                                        "binary copy: missing page buffer bytes while rewriting data file",
-                                    )
+                                    Error::execution(format!(
+                                        "binary copy: missing page buffer bytes while rewriting data file \
+                                         (column {col_idx}, page {page_idx}, buffer {buffer_idx}, expected size {size})",
+                                    ))
                                 })?;
                                 writer.write_all(&bytes).await?;
                                 current_pos += bytes.len() as u64;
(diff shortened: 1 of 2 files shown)


yanghua:
@claude review

rust/lance/src/dataset/optimize/tests/binary_copy.rs
```diff
@@ -45,6 +45,57 @@ async fn do_test_binary_copy_merge_small_files(version: LanceFileVersion) {
...
 }
 
+#[tokio::test]
+async fn test_binary_copy_empty_string_scalar_index() {
+    for version in LanceFileVersion::iter_non_legacy() {
+        do_test_binary_copy_empty_string_scalar_index(version).await;
+    }
+}
+
+async fn do_test_binary_copy_empty_string_scalar_index(version: LanceFileVersion) {
```
yanghua, line 55: Can we also scan the data and assert the data correctness?

rust/lance/src/dataset/optimize/binary_copy.rs
```diff
@@ -363,12 +365,19 @@ pub async fn rewrite_files_binary_copy(
...
+                            let writer = current_writer.as_mut().unwrap().as_mut();
+                            current_pos =
+                                apply_alignment_padding(writer, current_pos, version).await?;
+                            let start = current_pos;
+                            if *size == 0 {
+                                new_offsets.push((start, 0));
+                            } else {
+                                let bytes = bytes_iter.next().ok_or_else(|| {
+                                    Error::execution(
+                                        "binary copy: missing page buffer bytes while rewriting data file",
```
yanghua: Can we add more context in the error msg? e.g. `col_idx `, `page index`, `buffer`...

yanghua (commented):
Left some comments—otherwise, LGTM.

zhangyue19921010:
Hi @yanghua Thanks a lot for your review. All comments are addressed. PTAL~

yanghua (approved):
+1

PR #6994, opened 2026-05-29
fix(storage): retry throttled fts metadata listing

BubbleCal:
### What is the bug?
Distributed FTS indexing can encounter transient object-store list failures during partition metadata discovery, especially Azure `ServerBusy` and account egress-limit responses. These failures were not retried consistently on the Lance side, and FTS metadata listing could swallow list stream errors and later report the misleading error `No partition metadata files found`.

Transient cloud throttling can fail FTS index builds instead of backing off and resuming. When the list failure happens while discovering partition metadata, the user can lose the original service error body, request ID, or throttle reason and see an empty-directory error instead. List streams also did not feed the AIMD limiter before the underlying object-store list request started.

- Add delayed retry with conservative jittered exponential backoff to object-store list retry streams, while preserving immediate return for non-retryable errors.
- Resume retried list streams from the last successfully yielded key.
- Acquire the AIMD list token before creating `list` and `list_with_offset` delegate streams, then continue observing yielded items and errors.
- Expand throttle classification for targeted cloud throttling messages, including Azure `ServerBusy`, account egress limits, HTTP 429, and known rate-limit phrases.
- Remove the provider-side `client_max_retries=0` AIMD-disable guard for AWS, Azure, and GCP. Users should set `lance_aimd_max_retries=0` to explicitly disable Lance AIMD throttling.
- Propagate real FTS metadata list stream errors immediately while preserving the existing no-files error for genuinely empty metadata directories.

No public Rust, Python, or Java APIs are added.

## Validation

- `cargo test -p lance-io list_retry`
- `CARGO_TARGET_DIR=/tmp/lance-fc2c-target cargo test -p lance-io throttle`
- `CARGO_TARGET_DIR=/tmp/lance-fc2c-target cargo test -p lance-index list_metadata_files`
- `CARGO_TARGET_DIR=/tmp/lance-fc2c-target cargo test -p lance merge_existing_index_segments_supports_fts_segments`
- `cargo fmt --all`
- `CARGO_TARGET_DIR=/tmp/lance-fc2c-target cargo clippy --all --tests --benches -- -D warnings`

### DIFF
--- a/rust/lance-io/src/object_store/throttle.rs
+++ b/rust/lance-io/src/object_store/throttle.rs
@@ -63,7 +63,21 @@ use tracing::{debug, warn};
 pub fn is_throttle_error(err: &object_store::Error) -> bool {
     // Only Generic errors can carry throttle responses
     if let object_store::Error::Generic { source, .. } = err {
-        source.to_string().contains("retries, max_retries")
+        let message = source.to_string();
+        let lowercase = message.to_ascii_lowercase();
+        lowercase.contains("retries, max_retries")
+            || lowercase.contains("serverbusy")
+            || lowercase.contains("server busy")
+            || lowercase.contains("egress is over the account limit")
+            || lowercase.contains("http 429")
+            || lowercase.contains("status code: 429")
+            || lowercase.contains("429 too many requests")
+            || lowercase.contains("too many requests")
+            || lowercase.contains("slowdown")
+            || lowercase.contains("please reduce your request rate")
+            || lowercase.contains("rate limit")
+            || lowercase.contains("throttling")
+            || lowercase.contains("throttled")
     } else {
         false
     }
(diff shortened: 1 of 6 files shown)


Xuanwo (approved):

Xuanwo:
It is possible for us to not rely listing?

BubbleCal:
let's remove listing in next PR!

rust/lance-io/src/object_store/throttle.rs
```diff
@@ -63,7 +63,21 @@ use tracing::{debug, warn};
...
+            || lowercase.contains("egress is over the account limit")
+            || lowercase.contains("http 429")
+            || lowercase.contains("status code: 429")
+            || lowercase.contains("429 too many requests")
+            || lowercase.contains("too many requests")
+            || lowercase.contains("slowdown")
+            || lowercase.contains("please reduce your request rate")
+            || lowercase.contains("rate limit")
+            || lowercase.contains("throttling")
+            || lowercase.contains("throttled")
```
westonpace, line 80: This could potentially lead to unnecessary retries if the bucket has `throttled` or `throttling` in the name 😦 

The old approach `retries, max_retries` was relying on one very particular string that would always be returned when `object_store` did a retry error.

I'm guessing the problem here is that object_store is not retrying on these and so we have to detect them ourselves?

It would be nice to upstream something like opendal's "is_temporary" into object_store so we don't have to do any of this.

PR #6995, opened 2026-05-29
fix: support multivector IVF centroids in segment builds

ddupg:
Distributed IVF index builds failed for multivector columns when using precomputed centroids:

```text
Failed to cast centroids to column type: Cast error: Casting from Float32 to FixedSizeList(4 x Float32) not supported
```

This fixes the centroid type handling for `list<fixed_size_list<...>>` columns and allows `IndicesBuilder` to infer the inner vector type for distributed builds.

### DIFF
--- a/python/python/lance/indices/builder.py
+++ b/python/python/lance/indices/builder.py
@@ -582,7 +582,8 @@ class IndicesBuilder:
             vector_type = self._describe_vector_type(field.type)
             if vector_type is None:
                 raise TypeError(
-                    f"Vector column {c} must be FixedSizeListArray "
+                    f"Vector column {c} must be FixedSizeListArray, "
+                    "list<FixedSizeList> (multivector), or "
                     f"1-dimensional FixedShapeTensorArray, got {field.type}"
                 )
             _, value_type = vector_type
@@ -602,7 +603,8 @@ class IndicesBuilder:
         if vector_type is not None:
             return vector_type[0]
         raise TypeError(
-            "Vector column must be FixedSizeListArray "
+            "Vector column must be FixedSizeListArray, "
+            "list<FixedSizeList> (multivector), or "
             f"1-dimensional FixedShapeTensorArray, got {data_type}"
         )
 
--- a/python/python/lance/indices/builder.py
+++ b/python/python/lance/indices/builder.py
@@ -56,7 +56,9 @@ class IndicesBuilder:
         """
         self.dataset = dataset
         self.column = self._normalize_column(column)
-        self.dimension = self.dataset.schema.field(self.column[0]).type.list_size
+        self.dimension = self._vector_dimension(
+            self.dataset.schema.field(self.column[0]).type
+        )
 
     def train_ivf(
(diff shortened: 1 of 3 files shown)


yanghua:
@claude review

python/python/lance/indices/builder.py
```diff
@@ -578,28 +580,48 @@ def _normalize_column(self, column):
...
+        ):
+            return data_type.value_type.list_size
+        if isinstance(data_type, pa.FixedShapeTensorType) and len(data_type.shape) == 1:
+            return data_type.shape[0]
+        raise TypeError(
+            "Vector column must be FixedSizeListArray "
+            f"1-dimensional FixedShapeTensorArray, got {data_type}"
+        )
+
+    def _vector_value_type(self, data_type):
```
yanghua: The two method bodies have the same judgment logic. Maybe we can extract them into a common place, e.g. `_describe_vector_type(data_type) -> Optional[(dim, scalar_type)]`.

python/python/lance/indices/builder.py
yanghua: Could there still be any problems here?

yanghua (commented):
Left two comments.

python/python/lance/indices/builder.py
```diff
@@ -578,28 +579,44 @@ def _normalize_column(self, column):
...
 
         return column
 
+    def _vector_dimension(self, data_type):
+        vector_type = self._describe_vector_type(data_type)
+        if vector_type is not None:
+            return vector_type[0]
+        raise TypeError(
+            "Vector column must be FixedSizeListArray "
+            f"1-dimensional FixedShapeTensorArray, got {data_type}"
```
yanghua: What about this? `Vector column must be FixedSizeListArray, list<FixedSizeList> (multivector), or 1-dimensional FixedShapeTensorArray, got {data_type}`?

python/src/dataset.rs
```diff
@@ -4337,17 +4337,27 @@ fn prepare_vector_index_params(
                 ));
             }
 
+            let centroid_type = match column_type {
+                DataType::List(field)
```
yanghua, line 4341: Maybe we can also support `LargeList<FixedSizeList>`. But not must have it in this PR.

yanghua (commented):
One comment left, otherwise LGTM.

python/src/dataset.rs
ddupg: Thanks for pointing this out. I checked the Rust and Python vector index paths, and LargeList<FixedSizeList> is not supported end-to-end today. Most multivector code only handles List<FixedSizeList>.

This PR keeps the scope to fixing the existing List<FixedSizeList> distributed/segment index build issue. LargeList support should be handled separately.

python/python/lance/indices/builder.py
ddupg: Thanks for reviewing, fixed this.

yanghua (approved):
LGTM

PR #6997, opened 2026-05-29
refactor!: remove index segment builder

Xuanwo:
Remove the deprecated `IndexSegmentBuilder` API and route staged index publishing through the existing `create_index_uncommitted`, `merge_existing_index_segments`, and `commit_existing_index_segments` workflow.

`IndexSegmentBuilder` was kept in the codebase after distributed vector indexing moved to the flatter merge-and-commit workflow. That leftover API created two competing ways to describe the same lifecycle, which made the codebase and bindings harder to reason about and caused confusion around whether callers should plan/build segments or just merge and commit existing segment metadata.

This PR removes the parallel plan/build surface from Rust, Python, and Java bindings while keeping direct uncommitted segment commit support. FTS staged segments now finalize during commit metadata construction, so single-source staged FTS outputs no longer need a separate builder step before publishing.


westonpace (approved):
These comments from Claude seem like they might be valid (I skipped point 1 which was not valid)...

---

**2. Validation regression: duplicate segment UUID check removed**

The builder previously rejected duplicate UUIDs with a clear error (`"duplicate segment uuid"`). That check is now gone and there's no equivalent in `commit_existing_index_segments`. The test `test_index_segment_builder_rejects_duplicate_segment_uuids` is deleted without replacement. If a caller accidentally passes the same segment twice, it will either silently produce a corrupt index or fail with a confusing downstream error. A duplicate-UUID guard should be added to `commit_existing_index_segments`.

**3. Missing `index_type` validation in `commit_existing_index_segments`**

The builder enforced an explicit index type, which meant callers had to consciously declare what they were building. That guardrail is now gone. If a caller passes segments with inconsistent or missing `index_details`, the error message they get (`"missing index details"`) is reasonable, but the check happens later and further down the stack. Consider whether the `commit_existing_index_segments` entry point should validate that all segments share the same index type.

**4. `target_segment_bytes` / merge-by-size capability silently dropped**

The builder had `with_target_segment_bytes`, which allowed callers to merge multiple small segments into larger physical segments up to a size target. `merge_existing_index_segments` handles explicit grouping, but the automatic size-based grouping is gone entirely. The PR description doesn't mention this capability being dropped. If any external caller relied on `build_all()` with `target_segment_bytes` to automatically pack segments, they'll silently get no merging after upgrading. This warrants at least a mention in the PR description and potentially a `CHANGELOG` or migration guide entry given the `breaking-change` label.

**5. Test `test_commit_existing_index_segments_rejects_vector_segments_without_index_details` behavior change**

The old test built segments first (which succeeded), then separately committed (also worked). The new test expects `commit_existing_index_segments` itself to return an error when `index_details` is missing. That's the correct behavior to test, and the test appears correct — but it's worth verifying that the error path actually propagates cleanly from `into_index_segment()` inside `build_index_metadata_from_segments` before this is merged.

**6. Codecov: 1 partial line in `rust/lance/src/index.rs`**

The codecov bot flags one partial line in the `build_index_metadata_from_segments` FTS-finalization block. Given that this is the most semantically significant new code in the PR (the moved FTS finalization), that partial coverage is worth closing before merge.

PR #6998, opened 2026-05-29
fix(python): validate BFloat16.from_bytes length

ddupg:
Small fix for a missing return when BFloat16.from_bytes rejects non-2-byte input.


yanghua (commented):
If we can add some tests, e.g.

```
import pytest
from lance.arrow import BFloat16


def test_bf16_from_bytes_roundtrip():
    val = BFloat16(1.5)
    assert BFloat16.from_bytes(b"\xc0\x3f") == BFloat16(1.5)


@pytest.mark.parametrize("bad", [b"", b"\x00", b"\x00\x00\x00", b"\x00" * 4])
def test_bf16_from_bytes_invalid_length_raises(bad):
    with pytest.raises(ValueError, match="expected 2 bytes"):
        BFloat16.from_bytes(bad)
```

That would be better.

ddupg:
> If we can add some tests, e.g.
> 
> ```
> import pytest
> from lance.arrow import BFloat16
> 
> 
> def test_bf16_from_bytes_roundtrip():
>     val = BFloat16(1.5)
>     assert BFloat16.from_bytes(b"\xc0\x3f") == BFloat16(1.5)
> 
> 
> @pytest.mark.parametrize("bad", [b"", b"\x00", b"\x00\x00\x00", b"\x00" * 4])
> def test_bf16_from_bytes_invalid_length_raises(bad):
>     with pytest.raises(ValueError, match="expected 2 bytes"):
>         BFloat16.from_bytes(bad)
> ```
> 
> That would be better.

Thanks for reviewing, added tests.

yanghua (approved):
LGTM

Issue #6986, opened 2026-05-28
Proposal: introduce A-* area labels with auto-labeler


wjones127:
## Goal

Make it easier to filter open PRs by code area (e.g., "show me all open encoding PRs"). Today the area-ish labels in this repo (`vector`, `indexes`, `file-storage`, `format`, `java`, `python`, etc.) are unprefixed, inconsistent, and partly overlap — there's no clean filter for "PRs in my area."

## Proposal

Adopt an `A-*` area-label convention (rust-lang style) plus an auto-labeler that applies labels based on changed paths.

### Labels

| Label | Description | Paths |
|---|---|---|
| `A-java` | Java bindings + JNI | `java/**`, `rust/lance-jni/**` |
| `A-namespace` | Namespace impls | `rust/lance-namespace*/**` |
| `A-index` | Vector index, linalg, tokenizer | `rust/lance-index/**`, `rust/lance-linalg/**`, `rust/lance-tokenizer/**` |
| `A-encoding` | Encoding, IO, file format | `rust/lance-encoding/**`, `rust/lance-io/**`, `rust/lance-file/**`, `protos/**` |
| `A-python` | Python bindings | `python/**` (excluding lockfiles) |
| `A-docs` | Documentation | `docs/**`, `**/*.md` |
| `A-ci` | CI / build workflows | `.github/**` |
| `A-deps` | Dependency updates | `**/Cargo.lock`, `**/uv.lock`, `deny.toml`, `**/pyproject.toml` |
| `A-core` | General Rust core (catch-all) | other `rust/**`, `Cargo.toml` |

A PR can carry multiple `A-*` labels (e.g. a Java + namespace PR gets both).

### Auto-labeler

Add `.github/labeler.yml` + a workflow using `actions/labeler@v5` that runs on PR open/sync and applies labels based on changed paths.

### Migration of existing labels

Rename in place via `gh label edit`. GitHub preserves the label on existing issues/PRs through rename, so nothing breaks:

- `vector`, `indexes` → `A-index` (merge)
- `file-storage`, `format` → `A-encoding`
- `java` → `A-java`
- `python` → `A-python`
- `ci` → `A-ci`
- `dependencies`, `python:uv` → `A-deps`
- `rust` → unclear; possibly drop in favor of `A-core`

Labels not on the map (`arrow`, `c++`, `wasm`, `PyTorch`/`Tensorflow`, `duckdb`, `ray`, OS labels, etc.) stay as cross-cutting tags.

[...]

PR #7000, opened 2026-05-29
ci: add A-* area auto-labeler and label migration script

wjones127:
Adds an `actions/labeler@v5` workflow that applies path-based `A-*` area labels (from #6986) so open PRs can be filtered by code area, e.g. `is:pr is:open label:A-encoding`.

Path-based labeling moves to the new labeler (`.github/labeler-area.yml`); the existing `srvaroa/labeler` keeps only the title/body conventional-commit labels that feed release notes. `.github/scripts/migrate-area-labels.sh` renames/merges the current labels onto the `A-*` names — it is **manually run by a maintainer when we merge**, not wired into CI (the labeler applies existing labels but doesn't create them).

Labels: `A-java`, `A-namespace`, `A-index`, `A-encoding`, `A-format`, `A-python`, `A-docs`, `A-ci`, `A-deps`. A PR can carry several. A few notable cases:

- `A-format` (`protos/**`, `docs/src/format/**`) tracks on-disk format changes, kept distinct from `A-encoding`; format docs are excluded from `A-docs`.
- `A-python` doesn't exclude lockfiles, so a dependency bump gets both `A-python` and `A-deps` (over-label rather than lose the area signal).
- `ci`/`documentation` are left as-is — they remain title-based release-notes labels, separate from `A-ci`/`A-docs`.

Refs #6986

### DIFF
--- a/.github/labeler-area.yml
+++ b/.github/labeler-area.yml
@@ -1,0 +1,62 @@
+# Area labels (A-*) applied by actions/labeler@v5 based on changed paths.
+# See .github/workflows/labeler.yml. Title/body conventional-commit labels are
+# handled separately by srvaroa/labeler via .github/labeler.yml.
+#
+# A PR may carry multiple A-* labels (e.g. a Java + namespace change gets both).
+
+A-java:
+  - changed-files:
+      - any-glob-to-any-file:
+          - "java/**"
+          - "rust/lance-jni/**"
+
+A-namespace:
+  - changed-files:
+      - any-glob-to-any-file:
+          - "rust/lance-namespace*/**"
+
+A-index:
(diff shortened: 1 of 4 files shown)


LuQQiu (approved):
A is confusing to me in the beginning, then realize it either be agent or auto ... kind of make sense

wjones127:
> A is confusing to me in the beginning, then realize it either be agent or auto ... kind of make sense

"A" is for "Area". It's just a prefix to say what the label is about.  It's inspired by the labelling system in rustlang repo: https://github.com/rust-lang/rust/labels?page=2 So they have `A` for area, and `E` for "effort" (`E-low`, `E-medium`).

PR #7004, opened 2026-05-29
fix: stream object copies larger than cloud's CopyObject limit

vivek-bharathan:
copy now HEADs the source, and when it exceeds 5 GiB, falls back to a streaming multipart copy

### DIFF
--- a/rust/lance-io/src/object_store.rs
+++ b/rust/lance-io/src/object_store.rs
@@ -47,7 +47,7 @@ pub mod throttle;
 mod tracing;
 use crate::object_reader::SmallReader;
 use crate::object_writer::{LocalWriter, WriteResult};
-use crate::traits::Writer;
+use crate::traits::{WriteExt, Writer};
 use crate::utils::tracking_store::{IOTracker, IoStats};
 use crate::{object_reader::CloudObjectReader, object_writer::ObjectWriter, traits::Reader};
 use lance_core::{Error, Result};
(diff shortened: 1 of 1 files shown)


rust/lance-io/src/object_store.rs
```diff
@@ -771,11 +771,32 @@ impl ObjectStore {
...
+    ///
+    /// A single-shot server-side copy (`CopyObject`) caps the source size at
+    /// 5 GiB. For larger objects we stream the source through a multipart write
+    /// instead.
+    async fn copy_impl(&self, from: &Path, to: &Path, max_single_copy: u64) -> Result<()> {
         if self.is_local() {
             // Use std::fs::copy for local filesystem to support cross-filesystem copies
             return super::local::copy_file(from, to);
         }
+        if self.inner.head(from).await?.size > max_single_copy {
```
cmccabe: we should only do this for AWS and GCP. Azure is fine (and probably other blob stores).

Also, please add a reference to the upstream bug that we're working around here: https://github.com/apache/arrow-rs-object-store/issues/563

rust/lance-io/src/object_store.rs
```diff
@@ -771,11 +771,32 @@ impl ObjectStore {
...
+            let reader = self.open(from).await?;
```
cmccabe, line 813: It would be more efficient to create the reader first, then call size() to get the size, rather than making a totally separate call to HEAD.

cmccabe (approved):

wjones127 (approved):
This looks reasonable to me.

cmccabe:
@vivek-bharathan : it looks like we need to remove this:
```
#[cfg_attr(coverage, coverage(off))]
```
because it's causing some Linux compiler errors:
```
    error[E0658]: the `#[coverage]` attribute is an experimental feature
    --> rust/lance-io/src/object_store.rs:1568:5
     |
1568 |     #[cfg_attr(coverage, coverage(off))]
     |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
     |
```

PR #7006, opened 2026-05-29
fix: restore simple FTS tokenizer default

Xuanwo:
This restores the native FTS default tokenizer to `simple` after ICU showed behavior differences that are too large for the default path. ICU remains available through explicit `base_tokenizer="icu"`, while docs and tests now describe the default as `simple` again.


Xuanwo:
> This PR touches the Lance format specification.

Not really

BubbleCal (approved):

wjones127 (commented):
+1

PR #7009, opened 2026-05-30
perf(mem_wal): match hnswlib throughput via runtime AVX-512 f32 distance

touch-of-grey:
## Summary

Makes the in-memory MemWAL HNSW (`rust/lance/src/dataset/mem_wal/hnsw/`) as fast as hnswlib on insert and search.

Decomposing the gap against hnswlib showed the index was never algorithmically slower — recall was equal-or-better and memory ~44% lower at every size. The only real gap was SIMD width: the shipped binary targets `target-cpu=haswell`, so `f32::l2`/`f32::dot` only autovectorize to AVX2, while a `-march=native` hnswlib uses AVX-512 on capable CPUs (a `target-cpu=native` control build confirmed parity).

This adds runtime-dispatched `l2_f32`/`dot_f32` in `lance-linalg` (`#[target_feature(avx512f)]` 16-wide kernels gated by `SIMD_SUPPORT`, AVX2 fallback via the existing autovectorized path — same pattern as `dot_u8`) and routes the MemWAL HNSW distance through them, so the shipped build uses AVX-512 at runtime.

## Results (c7i.12xlarge, 48 threads, dim=1024, m=12, ef=64, k=10)

Shipped build with this change vs hnswlib:

| rows | insert L/H | query L/H | recall@10 L/H | peak RSS L/H |
|------|-----------|-----------|---------------|--------------|
| 100k | 0.98×     | 1.11×     | 0.80 / 0.84   | 0.56×        |
| 500k | 0.96×     | 0.96×     | 0.51 / 0.52   | 0.56×        |
| 1M   | 0.98×     | 1.05×     | 0.49 / 0.43   | 0.56×        |

vs the pre-change shipped build: insert @1M +14%, query @1M +57%. `perf stat` @1M: cycles within 1% (Lance 0.99×), 41% fewer instructions (AVX-512 density + zero-copy Arrow vs hnswlib's per-vector memcpy), ~44% less RSS.

Net: matches or beats hnswlib on insert and query, with equivalent CPU cycles, lower memory, and comparable recall.

## Changes

- `lance-linalg`: new `l2_f32`/`dot_f32` runtime AVX-512 dispatch (+ unit tests asserting they match the scalar reference across 16-multiple and tail lengths).
- `mem_wal/hnsw/storage.rs`: route `compute_f32_distance` (L2 + Dot) through the dispatchers.
- `benches/mem_wal/vector/hnsw/`: parity-suite driver and `--query-repeats` for a stable query window.

Follow-ups: AVX-512 cosine for the memtable; adopt the same dispatch in the broader `f32::l2`/`f32::dot` for all vector search.

cc @jackye1995 — please review.

### DIFF
--- a/rust/lance-linalg/src/distance/dot.rs
+++ b/rust/lance-linalg/src/distance/dot.rs
@@ -63,6 +63,44 @@ pub fn dot<T: Dot>(from: &[T], to: &[T]) -> f32 {
     T::dot(from, to)
 }
 
+/// Dot product between two f32 slices, dispatched to the widest SIMD backend
+/// available at runtime. See [`crate::distance::l2::l2_f32`] for why this is
+/// needed on top of the generic [`dot`].
+#[inline]
+pub fn dot_f32(x: &[f32], y: &[f32]) -> f32 {
+    #[cfg(target_arch = "x86_64")]
+    {
+        use lance_core::utils::cpu::SimdSupport;
+        if matches!(*SIMD_SUPPORT, SimdSupport::Avx512 | SimdSupport::Avx512FP16) {
+            // SAFETY: guarded by the runtime AVX-512 detection above.
+            return unsafe { dot_f32_avx512(x, y) };
+        }
+    }
+    dot(x, y)
+}
+
+#[cfg(target_arch = "x86_64")]
+#[target_feature(enable = "avx512f")]
+unsafe fn dot_f32_avx512(x: &[f32], y: &[f32]) -> f32 {
+    use std::arch::x86_64::*;
+    debug_assert_eq!(x.len(), y.len());
+    let n = x.len();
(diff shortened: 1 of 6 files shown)


touch-of-grey:
@jackye1995 ping for review. Summary: the MemWAL HNSW was never algorithmically slower than hnswlib (equal-or-better recall, ~44% less memory); the only gap was that the shipped `target-cpu=haswell` build never emits AVX-512 for f32 L2/dot. Adding runtime AVX-512 dispatch brings insert/query to parity (sometimes faster) on AVX-512 hardware. perf @1M: cycles within 1%, 41% fewer instructions. Full numbers in the description.

jackye1995 (approved):
Looks good to me!

PR #7011, opened 2026-05-31
perf(mem_wal): faster MemTable point lookups

touch-of-grey:
## Summary

Speeds up MemWAL MemTable primary-key point lookups to match/beat RocksDB on single-thread latency and write throughput, and roughly doubles N-thread read scaling.

### Index data structure
- Replace the crossbeam-skiplist backing the in-memory BTree index with a purpose-built **single-writer, lock-free-read arena skiplist with no epoch reclamation**. The MemTable is append-only with one writer, so reads need no epoch pin — the dominant cost in profiling. Nodes are a **single allocation** with the key and forward-pointer tower laid out contiguously (RocksDB `InlineSkipList` style) for cache locality.
- **Compact small-node keys** (the value is recovered/decoded at flush, not stored in the node): `FixedKey` (order-preserving `u64`) for `Int8..Int64`/`UInt8..UInt64`/`Date32/64`; `BytesKey`/`InlineBytes` (bytes inline for small values such as a 16-byte UUID and short string PKs) for `Utf8`/`LargeUtf8`/`Binary`/`LargeBinary`/`FixedSizeBinary`. Other scalar types fall back to the original `OrderableScalarValue` key.

### Point-lookup read path
- Seek-and-stop probe (`get_newest_visible`) honoring the MVCC visibility watermark.
- `LsmPointLookupPlanner::lookup` / `lookup_many` (batch) plus scanner routing: `pk = lit` / `pk IN (..)` filters route to the fast path; everything else uses the existing union/dedup scan.
- N-thread fix: stop cloning shared schema `Arc`s per lookup (borrow via `schema_ref`, borrow `none_target`) — that refcount cache line was ping-ponging across cores under concurrency.

### Benchmark
- `mem_wal_kv_point_lookup`: head-to-head Lance MemTable vs RocksDB, identical key/value/query sets. RocksDB arm is behind the off-by-default `bench-rocksdb` feature; `--key-type int|uuid`.

### Results (1M rows, c7i.8xlarge)
- **Single-thread read**: int 1.45µs / uuid 1.61µs — beats RocksDB (1.53 / 1.61µs).
- **Write**: 757k–908k rows/s — beats RocksDB (~694k).
- **Memory**: ~1.4× RocksDB.
- **N-thread (32t)**: 0.50× RocksDB (up from 0.28× after the shared-`Arc` fix). The residual is structural (seek memory bandwidth + Lance's separate columnar value gather), not a contended lock.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1089,6 +1089,24 @@ dependencies = [
 ]
 
 [[package]]
+name = "bindgen"
+version = "0.72.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "993776b509cfb49c750f11b8f07a46fa23e0a1386ffc01fb1e7d343efc387895"
+dependencies = [
+ "bitflags 2.11.1",
+ "cexpr",
+ "clang-sys",
+ "itertools 0.13.0",
+ "proc-macro2",
+ "quote",
+ "regex",
+ "rustc-hash",
+ "shlex",
+ "syn 2.0.117",
+]
+
+[[package]]
 name = "bit-set"
 version = "0.8.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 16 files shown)


jackye1995 (approved):
thanks for the work!

PR #7013, opened 2026-05-31
refactor(index)!: move distributed BTree build to segmented index framework

zhangyue19921010:
## Overview

This is a **refactor** PR that wires the BTree scalar index's **distributed build flow** into
Lance's existing segmented-index framework, and retires the legacy BTree `range_id` sharding +
`merge_index_metadata` path.

The new distributed build flow: each worker builds **one canonical segment** for its assigned
`fragment_ids` **without committing it**, and the driver publishes them all at once via
`commit_existing_index_segments(...)`. Passing `fragment_ids` is what triggers the skip-commit
segment build; the per-language entry points are:

- Python: `create_index_uncommitted(column, "BTREE", fragment_ids=[...])` — returns the uncommitted
  segment `Index`. (`create_scalar_index(..., "BTREE",
  fragment_ids=...)` is no longer the distributed entry — see below.)
- Java: `createIndex(IndexOptions.builder(...).withFragmentIds([...]).build())` — returns the
  segment `Index`.
- Rust: `CreateIndexBuilder::new(...).fragments(vec![...]).execute_uncommitted()`.

The old "distributed write of `part_*` shards + a final `merge_index_metadata`" path is gone.

> This PR does **not** include the segmented planning / k-way merge / optimize-append
> implementations (`plan_segments` / `build_segment` / `merge_segments`, the BTree branches of
> `IndexSegmentBuilder` / `merge_existing_index_segments`, etc.); those land in a follow-up PR2. In
> PR1 the per-fragment segments are committed **directly** via `commit_existing_index_segments(...)`,
> with no `build_all` / segment-builder step (so `IndexSegmentBuilder.with_index_type("BTREE")` is
> intentionally not wired yet).

## Compatibility Impact (Review Focus)

- **API soft-break: `merge_index_metadata(BTree)`**
  `Dataset::merge_index_metadata(..., IndexType::BTree)` (and Java `Dataset.mergeIndexMetadata(BTREE)`,
  Python `merge_index_metadata(index_type="BTREE")`) now **returns an error directly**, with
  migration guidance. Callers that previously relied on "distributed write of `part_*` +
  `merge_index_metadata` to commit" are **broken explicitly** (rather than silently producing wrong
  results).
  Migration: build one segment per fragment group with `create_index_uncommitted(..., "BTREE",
  fragment_ids=...)` and publish them with `commit_existing_index_segments(...)`; to consolidate
  multiple segments into fewer/one, use `merge_existing_index_segments(...)` (BTree support comes in
  PR2).

- **Distributed BTree builds move to `create_index_uncommitted` (Python)**
  Distributed BTree segments are now built through `create_index_uncommitted(column, "BTREE",
  fragment_ids=[...])`.  `create_scalar_index(..., "BTREE",
  fragment_ids=...)` (and the `IndexConfig` equivalent) is **now rejected** with an error pointing to
  `create_index_uncommitted`. Non-distributed `create_scalar_index("BTREE")` (no `fragment_ids`) is
  **unchanged** and still commits a single canonical index. Java has no separate scalar entry, so its
  distributed entry stays `createIndex(...withFragmentIds...)`.

- **`range_id` deprecated**
  BTree `range_id` (Rust `BTreeParameters::range_id`, Java `BTreeIndexParams.rangeId`) is
  **deprecated**: if `Some(..)` is detected at build time, a `warn!` migration hint is logged and the
  value is then **ignored**, still producing a canonical segment (lossless, no error). The field is
  **retained** (not physically removed) so that stale inputs are flagged loudly rather than silently
  dropped by serde; it will be removed in a future release.
  A pre-sorted data stream is still supported via Rust `preprocessed_data` / Java
  `withPreprocessedData` (the Python API does not take a reader). Pass it together with `fragment_ids`
  so the resulting segment is uncommitted and scoped to those fragments.

- **Behavior change: distributed semantics change**
  In the segmented case, distributed builds move from "**a single index** written in a distributed
  fashion, then a final merge" to "**creating multiple independent segments** in a distributed
  fashion" — each worker produces one **canonical** segment with **exactly one `page_lookup.lance` +
  one `page_data.lance`** per segment, and no more internal `part_*` shards. At query time the
  segments of one logical index are **unioned automatically**.

- **Historical compatibility: old layouts remain readable and remappable**
  The **read** and **compaction remap** paths for already-persisted range / `part_*` legacy index
  layouts are **fully preserved**, to maintain backward compatibility.

## Example

```python
import lance

ds = lance.dataset(uri)

# 1. Build one uncommitted BTree segment per fragment (fragment_ids => skip commit;
#    exactly one page_lookup.lance + one page_data.lance per segment, no part_* shards).
segments = [
    ds.create_index_uncommitted(
        column="id",
        index_type="BTREE",
        name="id_btree",
        fragment_ids=[frag.fragment_id],
    )
    for frag in ds.get_fragments()
]

# 2. Commit all segments as one logical index.
ds = ds.commit_existing_index_segments("id_btree", "id", segments)

# 3. Query (segments are unioned automatically).
ds.scanner(filter="id = 100").to_table()

### CODE REVIEW

**Comment by yanghua** [comment]
@claude review

File: python/python/lance/dataset.py [hunk 1]
```diff
@@ -3072,11 +3065,11 @@ def create_scalar_index(
...
-            method returns metadata for one segment but does not commit
-            the index to the dataset. The segment can be planned, merged, and
-            committed later using the segment builder and commit APIs.
+            If provided, the index will be created only on the specified fragments
+            using the legacy distributed scalar-index path. BTREE indices do not
+            support this path; build distributed BTREE segments with
+            :meth:`create_index_uncommitted` (``index_type="BTREE"``) and publish
+            them with :meth:`commit_existing_index_segments` instead.
             This parameter is passed via kwargs internally.
         index_uuid : str, optional
```
Xuanwo, line 3185: Maybe we should prevent users from specifying the `index_uuid`.

python/python/lance/dataset.py
```diff
@@ -3951,6 +3957,32 @@ def create_index_uncommitted(
...
+                    "for distributed index build"
+                )
+            if not isinstance(column, str):
+                raise NotImplementedError(
+                    "Scalar indices currently only support a single column"
+                )
+
+            kwargs = dict(kwargs)
+            kwargs["fragment_ids"] = fragment_ids
+            if index_uuid is not None:
```
Xuanwo, line 4026: I used to think `index_uuid` should alwayds generated by lance instead of users. What do you think?

Xuanwo (commented):
It seems we don't support `merge_existing_index_segments` for now. What is the future plan?

Another thing I want to confirm is whether this change only affects the distributed path. I expect that the plain "create index" will continue to work.

zhangyue19921010:
Hi @Xuanwo Thanks a lot for your review
> It seems we don't support merge_existing_index_segments for now. What is the future plan?

The current PR focuses on underlying distributed creation and migrating commit btree indexes to the segmented index framework. As for the merging, optimize and planning of segmented Btree indexes, the relevant work is mainly handled in another PR at https://github.com/lance-format/lance/issues/6979. Is that make sense?

> Another thing I want to confirm is whether this change only affects the distributed path. I expect that the plain "create index" will continue to work.

This change only affects the old distributed BTREE path. The normal `create_scalar_index(..., index_type="BTREE")` / plain create-index path continues to work.

1. In Python, we only reject the request when `fragment_ids is not None` and the index type is BTREE. When `fragment_ids` is not provided, the code still calls the native `create_index` path normally.

2. In Rust, `CreateIndexBuilder` still treats `IndexType::BTree` as a normal scalar index type.

For example, in `test_distribute_btree_index_build`, the new distributed segmented BTREE path is compared against a normal full BTREE index. The reference full BTREE index is created with:

```
reference_ds.create_scalar_index(
    column="id",
    index_type="BTREE",
    name="reference_btree_idx",
)
```

The test then compares the result counts for both exact queries and range queries.

> Maybe we should prevent users from specifying the index_uuid.

I completely agree. Historically, the main reason for allowing users to specify `index_uuid` was the old distributed build flow: all workers shared one UUID, wrote `part_*` files into the same directory, and then `merge_index_metadata` merged those files. 

Under the segmented index framework, each worker should create its own segment with a unique UUID. So users no longer need to specify the UUID manually, and allowing them to do so is risky.

I did a quick investigation:

1. The segmented index framework already has a fallback guard: `commit_existing_index_segments` reports an error if the same commit contains duplicate segment UUIDs, for example `ValueError: ... duplicate segment uuid ...`.

2. Vector segmented indexes and Inverted segmented indexes currently have a similar issue: they also accept and use user-specified UUIDs. My personal preference is to open a separate issue to audit all scenarios where user-specified UUIDs are allowed today, and decide where the high-level APIs should reject them. https://github.com/lance-format/lance/issues/7032

python/python/lance/dataset.py
zhangyue19921010: This is indeed a very good point which i missed before. I completely agree and removed this `index_uuid ` for btree. Also create a issue to track it https://github.com/lance-format/lance/issues/7032

Xuanwo:
Thanks, the follow-up plan makes sense.

Before approval, I think two small fixes are still needed:

1. Please reject user-specified `index_uuid` for distributed/uncommitted BTree at the core/API boundary, not only in Python. Otherwise Rust/Java can still reuse one UUID and race/overwrite the canonical BTree files.

2. Please remove `merge_existing_index_segments(...)` from the BTree `merge_index_metadata` error message until BTree merge support lands. For now the migration hint should only point to `commit_existing_index_segments(...)`.

The Python `IndexConfig` / validation change looks good to me.

zhangyue19921010:
Hi @Xuanwo All comments are addressed and CI passed. PTAL~

Xuanwo (approved):
Great work! Please also update the migration notes for users that depending the current distributed index logic.

PR #7014, opened 2026-05-31
feat(python): add shared RaBitQ rotation for distributed IVF_RQ builds

### LINKED ISSUE #7012: Distributed IVF_RQ builds have no way of sharing the RaBitQ rotation across workers in Python
## Summary

The engine-level capability for distributed IVF_RQ segment builds can't be driven from Python today. A distributed build requires every per-fragment segment to share one RaBitQ rotation so the segments are mergeable, and there is currently no Python-facing way to pin that rotation. As a result, a distributed IVF_RQ build driven from pylance produces segments with mutually incompatible rotations, which merge into a silently corrupt index (degraded recall, no error).

## What already works
At the engine level the IVF family, including RQ, has distributed build, commit, merge, and segment-pruned query. On the Python side, create_index_uncommitted already lets callers pin the two other shared artifacts a distributed build needs:

1. The IVF model, via ivf_centroids / precomputed_partitions_file
2. The PQ codebook, via pq_codebook

Note: The one missing shared artifact for RQ is the rotation.

## The gap
The RaBitQ rotation is generated during quantizer construction time and is not exposed as a build input. Consequently every per-fragment create_index_uncommitted call for IVF_RQ constructs its own quantizer with a fresh random rotation. The codes across segments aren't in a comparable space, so merge_existing_index_segments (or cross-segment query) yields incorrect results. 

## Proposed fix (minimal, mirrors pq_codebook)
Thread one shared, serialized rotation into the build params, exactly the way pq_codebook is threaded today/

```python
rotation = build_rq_rotation(dimension=768, num_bits=1, dtype="float16")  # once, on the driver
seg_a = ds.create_index_uncommitted("vec", "IVF_RQ", name="idx",
                                    ivf_centroids=c, rq_rotation=rotation, fragment_ids=frags_a)
seg_b = ds.create_index_uncommitted("vec", "IVF_RQ", name="idx",
                                    ivf_centroids=c, rq_rotation=rotation, fragment_ids=frags_b)
merged = ds.merge_existing_index_segments([seg_a, seg_b])
ds = ds.commit_existing_index_segments("idx", "vec", [merged])
```

Note: Single-process create_index(index_type="IVF_RQ") is unaffected since the rotation never needs sharing there.

gstamatakis95:
Closes: #7012 

## What

Distributed `IVF_RQ` builds work in the Rust engine (#6359) but could not be driven from Python because the RaBitQ rotation could not be pinned across workers. Each per-fragment build generated its own random rotation, so segments rotated vectors differently, their binary codes were not comparable, and merging corrupted the index.

This adds a way to mint one rotation, broadcast it, and reuse it in every per-fragment build, mirroring how `pq_codebook` is injected.

  ## Changes

  - Add `build_rq_rotation(dimension, num_bits=1, rotation_type="fast", dtype="float32")` that returns one rotation as a JSON string.
  - Add an `rq_rotation` parameter to `create_index_uncommitted`, parsed into a new transient `RQBuildParams.rotation` field and consumed by `RabitQuantizer::build`.
  - `build()` reuses the supplied rotation instead of generating a random one, after validating `num_bits`, `code_dim`, and the signs length.
  
  ## Notes

  - Only the fast rotation is supported because its sign vector is JSON serializable. 
  - The matrix rotation keeps a dense matrix in a binary buffer that the JSON wire format drops, so it is rejected with a clear error.
  - The params proto, the segment builder, and the merge and commit paths are unchanged.

  ## Tests

  - Rust unit tests for shared-rotation reuse, identical codes across builds, mismatch and bad-input rejection, and the matrix-via-JSON rejection.
  - A Python integration test that builds two `IVF_RQ` segments on separate fragments with one shared rotation, merges, commits, and queries.


python/python/lance/lance/indices/__init__.pyi
```diff
@@ -67,6 +67,12 @@ def transform_vectors(
     pq_codebook: pa.Array,
     dst_uri: str,
 ): ...
+def build_rq_rotation(
+    dimension: int,
+    num_bits: int = 1,
+    rotation_type: str = "fast",
```
BubbleCal: let's remove this
`rotation_type` is an internal param for compatibilty

python/python/lance/dataset.py
```diff
@@ -3304,6 +3304,7 @@ def _create_index_impl(
         *,
         target_partition_size: Optional[int] = None,
         skip_transpose: bool = False,
+        rq_rotation: Optional[str] = None,
```
BubbleCal: rename to `rabitq_model`?

python/python/lance/lance/indices/__init__.pyi
```diff
@@ -67,6 +67,12 @@ def transform_vectors(
     pq_codebook: pa.Array,
     dst_uri: str,
 ): ...
+def build_rq_rotation(
```
BubbleCal: rename to `build_rq_model`

BubbleCal (approved):
Just some naming comments

This is nice and is what on our roadmap! Thanks for the contribution!

gstamatakis95:
@BubbleCal I believe we are ready.

gstamatakis95:
Fixed linting @BubbleCal. Sorry for the delay

gstamatakis95:
@BubbleCal Apologies for the spam, resolved conflicts, tests are passing and overall it looks OK. I don't know if I can trigger any of the workflows myself so please let me know.

BubbleCal:
> @BubbleCal Apologies for the spam, resolved conflicts, tests are passing and overall it looks OK. I don't know if I can trigger any of the workflows myself so please let me know.

No worries, it's a limitation that first-time contributor can't trigger the CI

Will keep eyes on it

gstamatakis95:
Perfect, unsure if I need to do something specific so I'm just waiting for now

BubbleCal:
@gstamatakis95 Thanks for the contribution!

PR #7017, opened 2026-05-31
feat(blobv2): support all BlobKind types in blob v2 compact_files

### LINKED ISSUE #6938: feat(blobv2):support for compaction
## Motivation

Datasets containing blob v2 columns (including external reference types) cannot be compacted. Calling `compact_files()` on such datasets raises a decoder error. Additionally, even if the decoder error were bypassed, external references would be incorrectly downgraded to inline/packed storage during re-encoding.

### test code

```python
import lance
import pyarrow as pa
from lance import Blob
import tempfile, os

with tempfile.TemporaryDirectory() as tmp_dir:
    local_file_path = os.path.join(tmp_dir, "external_blob.bin")
    with open(local_file_path, "wb") as f:
        f.write(b"hello world")

    file_uri = f"file://{local_file_path}"
    blob_ref = Blob.from_uri(file_uri, position=0, size=11)

    schema = pa.schema([
        pa.field("id", pa.int32()),
        lance.blob_field("blob")
    ])

    table = pa.table({
        "id": [1, 2, 3],
        "blob": lance.blob_array([blob_ref, blob_ref, blob_ref])
    }, schema=schema)

    ds = lance.write_dataset(
        table, os.path.join(tmp_dir, "test_dataset"),
        schema=schema,
        max_rows_per_file=1,
        data_storage_version="2.2",
        allow_external_blob_outside_bases=True
    )

    ds.optimize.compact_files(num_threads=1)  # 💥 Error
```

### error
```
Invalid user input: there were more fields in the schema than provided column indices / infos, /Users/bytedance/RustProject/community/lance/rust/lance-encoding/src/decoder.rs:454:13
```

yyzhao2025:
feat(blobv2): support all BlobKind types in blob v2 compact_files

`compact_files` did not support blob v2 columns. Running compaction on a dataset with blob v2 columns would corrupt blob data for all `BlobKind` types (Inline, Packed, Dedicated, and External). This PR adds full blob v2 support to the compaction path so that all 4 `BlobKind` types are correctly preserved after `compact_files`.

### Why compaction failed for blob v2

Blob v2 uses a multi-file storage layout where blob payloads live in sidecar `.blob` files and are referenced through a 5-field descriptor struct (kind, position, size, blob_id, blob_uri). The existing compaction path was designed for blob v1's single-file layout and did not account for this:

1. **Binary copy treated blob v2 as opaque bytes.** `can_use_binary_copy` copied raw page data without understanding the descriptor layout, out-of-line buffers, or sidecar file references. This lost the relationship between descriptors and their `.blob` files for Packed/Dedicated blobs, produced `size=0` for Inline blobs, and broke External blob URI resolution since `base_id` references to manifest `base_paths` were not preserved in the new fragment context.

2. **Re-encode path received only descriptors, no actual blob data.** `prepare_reader` used the default `BlobsDescriptions` mode, which unloads blob v2 columns into their descriptor view. Without actual blob bytes, `BlobPreprocessor` re-encoding produced `size=0` for all non-External kinds, and External blob URIs could not be correctly resolved without the original `base_id` → `base_paths` mapping.

3. **Descriptor-format pass-through branches were incorrect.** `BlobPreprocessor` and `BlobV2StructuralEncoder` had branches intended to handle descriptor-only data, but these produced wrong output without backing `.blob` files.

### Changes

- **Disable binary copy for all blob columns** in `can_use_binary_copy`, preventing raw page-level copying that ignores blob v2's multi-file layout.
- **Add `with_row_addr()` in `prepare_reader`** for blob v2 datasets, enabling row-address-based access to actual blob data via `take_blobs_by_addresses`.
- **Add `transform_blob_v2_batch`** to read real blob data per-row and convert descriptor columns to the `Struct<data, uri, position, size>` user-view format expected by `BlobPreprocessor` before re-encoding. This function:
  - Classifies each row as Null, External, or DataBlob (Inline/Packed/Dedicated) via `classify_rows`
  - Reads actual blob bytes for DataBlob rows using `take_blobs_by_addresses` (lazy, per-row loading to avoid OOM)
  - Reconstructs absolute URIs for External blobs from `base_id` and manifest `base_paths`
  - Assembles the user-view struct with correct null propagation
- **Remove incorrect descriptor-format handling** from `BlobPreprocessor`
- **Remove incorrect descriptor-format pass-through** from `BlobV2StructuralEncoder`
- **Resolve External blob URIs** by reconstructing absolute paths from `base_id` and manifest `base_paths` during compaction
- **Add shared `BLOB_V2_USER_FIELDS` / `BLOB_V2_USER_TYPE` constants** in `lance-core::datatypes` to eliminate schema hardcoding between the transform and write paths

### Testing

12 new blob v2 compact tests cover all `BlobKind` types and edge cases:

| Scenario | Test |
|---|---|
| External + Inline mixed | `test_compact_blob_v2_preserves_external_references` |
| Inline + Packed + Dedicated | `test_compact_blob_v2_packed_and_dedicated` |
| NULL rows | `test_compact_blob_v2_with_null_rows` |
| Deleted rows not resurrected | `test_compact_blob_v2_deleted_rows_not_resurrected` |
| External + DataBlob mixed | `test_compact_blob_v2_external_and_data_blob_mixed` |
| Multiple blob v2 columns | `test_compact_blob_v2_multiple_blob_columns` |
| External + NULL mixed | `test_compact_blob_v2_external_and_null_mixed` |
| All-NULL / All-External fragments | `test_compact_blob_v2_all_null_and_all_external_fragments` |
| Multiple base_ids | `test_compact_blob_v2_external_with_multiple_base_ids` |
| Large blob lazy loading | `test_compact_blob_v2_large_blobs` |
| BlobKind reclassification | `test_compact_blob_v2_blob_kind_reclassification` |
| Multi-batch processing | `test_compact_blob_v2_multi_batch` |

All tests verify actual blob bytes via `take_blobs`, not just descriptor metadata. 83 optimize tests and 50 blob tests pass with zero clippy warnings on changed files.

## Related Issue
Closes  #6938

### DIFF
--- a/rust/lance/src/dataset/optimize.rs
+++ b/rust/lance/src/dataset/optimize.rs
@@ -93,7 +93,7 @@ use super::transaction::{
     Operation, RewriteGroup, RewrittenIndex, Transaction, TransactionBuilder,
 };
 use super::utils::make_rowid_capture_stream;
-use super::{WriteMode, WriteParams, write_fragments_internal};
+use super::{WriteMode, WriteParams, cleanup_data_fragments, write_fragments_internal};
 use crate::Dataset;
 use crate::Result;
 use crate::dataset::utils::CapturedRowIds;
@@ -1654,26 +1654,39 @@ async fn rewrite_files(
 
     log::info!("Compaction task {}: file written", task_id);
 
-    let row_addrs = if let Some(row_ids_rx) = row_ids_rx {
-        let captured_ids = row_ids_rx
-            .try_recv()
-            .map_err(|err| Error::internal(format!("Failed to receive row ids: {}", err)))?;
-        let row_addrs = captured_ids.row_addrs(None).into_owned();
-        let mut serialized = Vec::with_capacity(row_addrs.serialized_size());
-        row_addrs.serialize_into(&mut serialized)?;
-        Some(serialized)
-    } else {
-        if dataset.manifest.uses_stable_row_ids() {
-            log::info!("Compaction task {}: rechunking stable row ids", task_id);
-            rechunk_stable_row_ids(dataset.as_ref(), &mut new_fragments, &fragments).await?;
-            recalc_versions_for_rewritten_fragments(
-                dataset.as_ref(),
-                &mut new_fragments,
-                &fragments,
-            )
-            .await?;
+    // Wrap in an async block so `?` returns into `row_addrs_result` and we can
+    // run cleanup before propagating the error.
+    let row_addrs_result: Result<Option<Vec<u8>>> = async {
+        if let Some(row_ids_rx) = row_ids_rx {
+            let captured_ids = row_ids_rx
+                .try_recv()
+                .map_err(|err| Error::internal(format!("Failed to receive row ids: {}", err)))?;
+            let row_addrs = captured_ids.row_addrs(None).into_owned();
+            let mut serialized = Vec::with_capacity(row_addrs.serialized_size());
+            row_addrs.serialize_into(&mut serialized)?;
+            Ok(Some(serialized))
+        } else {
+            if dataset.manifest.uses_stable_row_ids() {
+                log::info!("Compaction task {}: rechunking stable row ids", task_id);
+                rechunk_stable_row_ids(dataset.as_ref(), &mut new_fragments, &fragments).await?;
+                recalc_versions_for_rewritten_fragments(
+                    dataset.as_ref(),
+                    &mut new_fragments,
+                    &fragments,
+                )
+                .await?;
+            }
+            Ok(None)
         }
-        None
+    }
+    .await;
+
+    let row_addrs = match row_addrs_result {
+        Ok(v) => v,
+        Err(e) => {
+            cleanup_data_fragments(&dataset.object_store, &dataset.base, &new_fragments).await;
+            return Err(e);
+        }
     };
 
     metrics.files_removed = task
@@ -1995,6 +2008,13 @@ pub async fn commit_compaction(
         None
     };
 
+    // Collect new fragment paths before moving rewrite_groups into the transaction,
+    // so we can clean them up if the commit fails.
+    let all_new_fragments: Vec<Fragment> = rewrite_groups
+        .iter()
+        .flat_map(|g| g.new_fragments.iter().cloned())
+        .collect();
+
     let transaction = TransactionBuilder::new(
         // Use the version at which the compaction tasks were *planned*, not the
         // version of the dataset handle passed to this function.  In distributed
@@ -2013,9 +2033,13 @@ pub async fn commit_compaction(
     .transaction_properties(options.transaction_properties.clone())
     .build();
 
-    dataset
+    if let Err(e) = dataset
         .apply_commit(transaction, &Default::default(), &Default::default())
-        .await?;
+        .await
+    {
+        cleanup_data_fragments(&dataset.object_store, &dataset.base, &all_new_fragments).await;
+        return Err(e);
+    }
 
     Ok(metrics)
 }
@@ -5219,6 +5243,10 @@ mod tests {
             .count()
     }
 
+    /// Site 2 in PR #6320: when `commit_compaction` fails to apply the commit
+    /// after `rewrite_files` has already written new data files, those files
+    /// must be cleaned up. We force the commit failure by injecting an error on
+    /// writes to the `_transactions/` directory.
     #[tokio::test]
     async fn test_commit_compaction_cleans_up_data_on_commit_failure() {
         use crate::dataset::builder::DatasetBuilder;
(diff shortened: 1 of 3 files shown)


yanghua:
@claude review

rust/lance/src/dataset/optimize.rs
```diff
@@ -1643,13 +2013,9 @@ pub async fn commit_compaction(
     .transaction_properties(options.transaction_properties.clone())
     .build();
 
-    if let Err(e) = dataset
+    dataset
```
Xuanwo: This drops the cleanup path added for failed compaction commits. If `apply_commit` fails, the new fragments written by the rewrite are now left behind as orphan data files.

The existing test catches this:

`cargo test -p lance test_commit_compaction_cleans_up_data_on_commit_failure -- --nocapture`

It now fails with `left: 3, right: 2`.

Can we restore the `cleanup_data_fragments` wrapper around `apply_commit` failures, and keep the same cleanup for errors after `write_fragments_internal` has already written new fragments?

Xuanwo (commented):
Thanks for fixing blob v2 compaction. The main path looks good to me, but I found one failure-path regression that should be fixed before merge.

rust/lance/src/dataset/optimize.rs
yyzhao2025: > This drops the cleanup path added for failed compaction commits. If `apply_commit` fails, the new fragments written by the rewrite are now left behind as orphan data files.
> 
> The existing test catches this:
> 
> `cargo test -p lance test_commit_compaction_cleans_up_data_on_commit_failure -- --nocapture`
> 
> It now fails with `left: 3, right: 2`.
> 
> Can we restore the `cleanup_data_fragments` wrapper around `apply_commit` failures, and keep the same cleanup for errors after `write_fragments_internal` has already written new fragments?

@Xuanwo Thanks a lot for the reminder. I accidentally introduced a failure-path regression while resolving conflicts with PR #6320. I have now reverted back to the fixed state of PR #6320.

rust/lance/src/dataset/optimize.rs
```diff
@@ -1643,13 +2013,9 @@ pub async fn commit_compaction(
...
         .apply_commit(transaction, &Default::default(), &Default::default())
-        .await
-    {
-        cleanup_data_fragments(&dataset.object_store, &dataset.base, &all_new_fragments).await;
-        return Err(e);
-    }
+        .await?;
```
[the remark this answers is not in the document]
yyzhao2025: Thanks a lot for the reminder.PR #6320.Now I have now reverted back to the fixed state of PR #6320.

Xuanwo (commented):
Thanks, the main regression I pointed out is fixed now. I think the PR is close; I only have one remaining cleanup issue around blob v2 sidecars.

The original cleanup regression is fixed, but one blob-v2-specific cleanup gap still looks worth fixing before merge: the restored path cleans up the new `.lance` fragment files, while blob v2 rewrites can also create sidecars under `data/{data_file_key}/{blob_id}.blob`. `cleanup_data_fragments` only deletes the fragment `file.path`, so on an `apply_commit` failure with packed/dedicated blobs those sidecars appear to be left for later GC instead of being removed by the failure path.

Can we extend this cleanup to remove blob v2 sidecars for the new fragments too, and add a compact-failure test with packed/dedicated blob v2 data?

yyzhao2025:
> Thanks, the main regression I pointed out is fixed now. I think the PR is close; I only have one remaining cleanup issue around blob v2 sidecars.
> 
> The original cleanup regression is fixed, but one blob-v2-specific cleanup gap still looks worth fixing before merge: the restored path cleans up the new `.lance` fragment files, while blob v2 rewrites can also create sidecars under `data/{data_file_key}/{blob_id}.blob`. `cleanup_data_fragments` only deletes the fragment `file.path`, so on an `apply_commit` failure with packed/dedicated blobs those sidecars appear to be left for later GC instead of being removed by the failure path.
> 
> Can we extend this cleanup to remove blob v2 sidecars for the new fragments too, and add a compact-failure test with packed/dedicated blob v2 data?

@Xuanwo Thanks for the thorough review and for pointing out the blob v2 sidecar gap!
I've pushed an additional commit to address this cleanup issue:
1. Extended cleanup_data_fragments : It now extracts the data_file_key (via file stem) from the newly written .lance files and calls remove_dir_all to safely wipe out any associated .blob sidecar directories generated during the rewrite.
2. Added comprehensive test : Included a new test test_commit_compaction_cleans_up_blob_v2_sidecars_on_commit_failure . It specifically generates both a 100KB (Packed) and a 5MB (Dedicated) blob to ensure both sidecar layouts are fully covered and verified during an injected apply_commit failure.
All tests are passing locally.

Xuanwo (approved):

PR #7020, opened 2026-06-01
refactor!: separate memtable row position from row ids

yanghua:
## Summary

  Separates active MemTable internal row positions from public Lance `_rowid` / `_rowaddr` semantics.

  - Add an internal `__memwal_row_position` column for MemWAL dedup/order logic
  - Keep public active MemTable `_rowid` / `_rowaddr` outputs nullable instead of using row positions
  - Update active point lookup and vector search dedup to use the internal row position column
  - Generalize `WithinSourceDedupExec` from row-address ordering to ordering-column based dedup
  
  ## Breaking changes

  - `MemTableScanner::with_row_id()` and `MemTableScanner::with_row_address()` now return all-NULL system columns for active memtable rows instead of insertion-
  order surrogate values. Existing callers that used these columns as row positions will continue to compile but will observe NULL values instead.
  - `DedupDirection` variants were renamed. Downstream code matching on the old variant names must update to the new names.
  - `WithinSourceDedupExec::row_addr_column()` was renamed to `ordering_column()`. Callers using the old accessor must update their code.

### DIFF
--- a/rust/lance/src/dataset/mem_wal/memtable/scanner.rs
+++ b/rust/lance/src/dataset/mem_wal/memtable/scanner.rs
@@ -37,4 +37,5 @@ mod builder;
 mod exec;
 
 pub use builder::MemTableScanner;
+pub(crate) use exec::MEMWAL_ROW_POSITION_COLUMN;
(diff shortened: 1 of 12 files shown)


yanghua:
@claude review

PR #7026, opened 2026-06-01
feat(index): implement FM-Index scalar index for exact substring search

beinan:
## Motivation

Enable exact substring search at scale for AI pretraining data decontamination — detecting benchmark contamination in trillion-row text corpora, following the [Infini-gram Mini paper](https://arxiv.org/abs/2506.12229).

## Summary
- Implement FM-Index following the Infini-gram Mini paper architecture for exact substring search
- Huffman-shaped wavelet tree for entropy-compressed BWT rank queries (~0.26N bytes)
- Sampled suffix array (D=32) with LF-mapping locate for document resolution (~0.25N bytes)
- Partitioned index (10K docs/partition) with blocked storage (32KB blocks) and lazy loading
- Wire up `IndexType::FMIndex` in Lance's `create_index` and query paths (`contains()` filter)
- Index size ~0.95x of text (paper claims 0.44x; gap is Lance row overhead per block)

## Benchmark (100K gitlake source code files, 1.59 GB text)

| Metric | FM-Index | N-Gram |
|--------|----------|--------|
| Index size | 1,513 MB (0.95x) | 84 MB (0.05x) |
| Build time | 132s | 9s |
| Short queries (e.g. `fn `) | 9034ms/q | 448ms/q |
| Medium queries (e.g. `fn main()`) | **29ms/q** | 480ms/q |
| Long queries (~80 chars) | **34ms/q** | 206ms/q |

FM-Index is 17x faster than N-Gram on medium queries and returns exact results (N-Gram returns approximate candidates needing recheck). N-Gram cannot find queries shorter than 3 characters (e.g. `fn ` returns 0).

## Test plan
- [x] 9 unit tests covering search, locate, wavelet access, serialization, multi-document
- [x] End-to-end benchmark through Lance dataset API (`dataset.create_index`, `dataset.count_rows(filter)`)
- [x] Verified correct match counts against full-scan baseline on real source code


beinan:
Format vote: https://github.com/lance-format/lance/discussions/7027

westonpace:
Attached a review from Claude at the bottom (FWIW).  I agree with the first p0 point.  That call to ensure_block seems out of place, especially buried inside an inline method.  Can we ensure all the blocks we need are loaded before we start processing them?

Let's also avoid dead code.  Especially now with agents.  At read time dead code is just wasted tokens.

The remap limitation is fine to address in a follow-up.  Or will this index never be remappable?

Other follow-ups that we will need (to be fair, ngram index probably needs these same concerns addressed too):

* It appears the partitions are more-or-less independent.  Should we just make these segments?
* If the answer to the above is "yes" then we probably need a follow-up for merging segments.
* We need to document the performance characteristics.  Your comments say this performs poorly on short queries for example.  Also, how is the 10K partition size going to scale?  Would search still be feasible with 1M or 100M rows at that partition size?  Or would the search need to be distributed at that point?

### What is an FM-Index?

The FM-Index (Full-text Minute Index, named after Ferragina & Manzini) is a compressed full-text index that enables exact substring search in sub-linear space. Here's how it works conceptually:

**Burrows-Wheeler Transform (BWT):** The input text is sorted into all its rotations, and the last column of that sorted matrix is kept. The BWT has a property: characters preceding the same suffix cluster together, making it highly compressible and supporting a key operation called LF-mapping (Last-to-First).

**Backward search:** To find all occurrences of a pattern `P`, you maintain an interval `[lo, hi)` in the BWT. Starting from the last character of `P`, you narrow the interval one character at a time using LF-mapping:
```
for b in P.reversed():
    lo = C[b] + rank(b, lo)
    hi = C[b] + rank(b, hi)
```
where `C[b]` is a cumulative frequency table and `rank(b, pos)` counts occurrences of `b` in `BWT[0..pos]`. The interval size at the end is the match count.

**Wavelet tree for rank queries:** The BWT is stored in a wavelet tree — a binary tree that recursively partitions the alphabet. The PR uses a **Huffman-shaped** wavelet tree where frequent characters get shorter codes, giving entropy-compressed rank queries in ~0.26N bytes.

**Locate (BWT position → document):** Not every BWT position has an explicit suffix array (SA) entry stored — the SA is sampled every 32 positions. To resolve a BWT position to a document, you walk LF-mapping steps until landing on a sampled SA entry, then look up which document contains that text offset.

**Partitioning:** The index is split into 10K-document partitions. Each partition is an independent FM-Index with its own BWT, wavelet tree, and sampled SA. At query time all partitions are searched and results are unioned.

---

### Code Review

**P0 — Correctness / Panic risk**

- **`block_on` inside `block_in_place`** (`LazyRankBitVec::ensure_block`): The lazy block-load path bridges async I/O into the synchronous rank-query hot path using `block_in_place` + `block_on`. This will **panic on a `current_thread` Tokio runtime**, which Lance sometimes uses (e.g., in tests, Python bindings, and some integration paths). The tests explicitly use `#[tokio::test(flavor = "multi_thread")]` as a workaround, but nothing prevents this from being called from a single-threaded runtime in production. The lazy loading design needs rethinking — either make the query path fully async (propagate `async` through `rank()`), or pre-load blocks before entering the sync backward search loop.

- **Silent fallback in locate** (`steps >= n`): When the LF-mapping walk never hits a sampled SA entry (shouldn't happen in a correct index, but could due to corruption or bugs), the code silently returns position `0`, misattributing the match to document 0. Should return an error or at minimum emit a `tracing::warn!`.

**P1 — Design / Architecture**

- **Massive code duplication between `FMIndex`/`HuffmanWaveletTree` and `LazyFMIndex`/`LazyHuffmanWaveletTree`:** The `backward_search`, `locate`, `doc_for_position`, `search`, `access`, `rank`, and `rank_pair` methods are duplicated nearly verbatim between the in-memory (build-time) and lazy (query-time) structs. These should share a trait or be parameterized (e.g., `struct GenericFMIndex<W: WaveletTree>`). The duplication is flagged by the `#[allow(dead_code)]` suppressions.

- **`can_remap()` returns `false` / `remap()` returns `Err`:** This means the FM-Index cannot participate in Lance's compaction/fragment merging pipeline. As a result, deletes will never be reflected in the index. This is a significant limitation that should be prominently documented and ideally tracked as a follow-up issue.

- **`update_criteria` uses `only_new_data`** (no merge): Similarly, dataset updates/appends create a new partition but never merge. Over time with many appends, the partition count grows unbounded and query time degrades linearly with partitions. A merge/rebuild strategy is needed.

**P2 — Minor quality issues**

- `num_zeros` column is always written as `0` and never read. Dead weight — remove it.

- `test_sentinel_sanitization` doesn't actually test the sentinel sanitization path (it tests `build` directly, which skips sanitization). The test name is misleading.

- `protos/index.proto` has no trailing newline (`\\ No newline at end of file` in the patch).

- `walkdir` in the example binaries should be confirmed as `[dev-dependencies]`-only — not a runtime dependency.

**P3 — Observations**

- Index size of 0.95x input text vs the paper's claimed 0.44x is a large gap. The PR attributes this to "Lance row overhead per block." It would be worth understanding the breakdown — the 32KB block granularity for the wavelet tree seems coarse, and the per-partition `row_ids` and `doc_start_positions` stored in schema metadata as hex-encoded binary are unconventional storage choices.

- Short queries like `fn ` taking **9 seconds per query** is alarming. This is expected behavior for FM-Index when the match count is enormous (every result must be located), but it's a major usability footgun. Consider returning a result count cap or a `count_only` mode that skips the locate step.

**Summary:** The algorithm is well-chosen and the partitioned index architecture makes sense for large corpora. The main blockers before merging are the `block_on`/`current_thread` panic risk and the massive duplication between the in-memory and lazy structs. The inability to support `remap()` (and therefore compaction) should at minimum be documented clearly in the public API.

beinan:
Thanks for the thorough review! Addressed the key items:

**Fixed in this push:**
- **`block_on`/`block_in_place` panic risk**: Added `async fn prewarm()` on `LazyFMIndex` that pre-loads all wavelet tree blocks before sync search. Called from `ScalarIndex::search()` before entering the sync backward search + locate loop. The lazy `ensure_block()` fallback remains as a safety net but won't be hit in the normal path.
- **Dead `num_zeros` column**: Removed from schema — was always 0.
- **Silent `locate()` fallback**: Now emits `log::warn!` when SA locate exceeds N steps.

**Follow-ups (will track as issues):**
- **Remap/compaction**: FM-Index will need a full rebuild strategy. Not remappable by nature (SA/BWT are global structures).
- **Partitions → segments**: Agree this should align with Lance's segmented index model. Will investigate.
- **Code dedup**: `FMIndex`/`LazyFMIndex` share logic — will extract a trait in follow-up.
- **Short query perf**: Will add a `count_only` mode that skips locate for use cases that only need match counts.
- **Index size gap (0.95x vs 0.44x)**: The gap is from Lance row overhead per block. Will investigate larger blocks or alternative storage.

beinan:
@westonpace all the p0 issue addressed, could you help take a look again? thanks!

westonpace (approved):
Let's get some votes but this looks great to me for a first stab at a new index.  Thanks!

PR #7028, opened 2026-06-01
feat(python): expose external blob mode and outside-base option for fragments

plotor:
## Summary

Expose `external_blob_mode` and `allow_external_blob_outside_bases` on `lance.fragment.write_fragments`.

This brings the low-level fragment write API in line with `lance.write_dataset`, allowing distributed/manual fragment writers to control how BlobV2 external URIs are handled:

- `external_blob_mode="reference"` keeps external blob URI references.
- `external_blob_mode="ingest"` reads external blob bytes during fragment write and stores them in Lance-managed storage.
- `allow_external_blob_outside_bases=True` allows reference-mode writes to persist absolute external URIs outside registered base paths.

The native Python stub was also updated so the exposed kwargs are reflected in type declarations.

## Testing

Added BlobV2 fragment-write coverage for:

- Default reference-mode behavior rejecting external URIs outside registered base paths.
- Reference-mode writes allowing outside-base external URIs when `allow_external_blob_outside_bases=True`.
- Ingest-mode fragment writes copying external URI bytes into Lance-managed storage, verified by deleting the source file before reading the blob back.
- The `return_transaction=True` fragment-write path with `external_blob_mode="ingest"`.
- Invalid configuration rejection for `external_blob_mode="ingest"` combined with `allow_external_blob_outside_bases=True`.

### DIFF
--- a/python/python/lance/dataset.py
+++ b/python/python/lance/dataset.py
@@ -6938,9 +6938,10 @@ def write_dataset(
           Lance-managed storage using the normal inline / packed / dedicated
           thresholds.
     allow_external_blob_outside_bases: bool, default False
-        If False, external blob URIs must map to the dataset root or a registered
-        base path. If True, external blob URIs outside registered bases are allowed.
-        This option only applies when ``external_blob_mode="reference"``.
+        If False, external blob URIs must map to a registered non-dataset-root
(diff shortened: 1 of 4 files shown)


yanghua:
@claude review

yanghua (approved):
LGTM

PR #7029, opened 2026-06-01
perf(mem_wal): close the FTS MemTable performance gap with Lucene

touch-of-grey:
## Summary

This brings the in-memory full-text index (`FtsMemIndex`, `rust/lance/src/dataset/mem_wal/index/fts.rs`) to **parity-and-beyond with Apache Lucene** on read latency across term, multi-term OR, and phrase queries, while also improving write throughput and memory — and preserving the single-writer/multi-reader (SWMR) lock-free contract and **exact BM25 recall**.

Benchmarked on FineWeb (100k / 500k / 1M docs), `c7i.16xlarge`, Lucene 10.1.0 / JDK 21. Numbers below are **Run A** (pre-tokenized, whitespace) — the apples-to-apples index-vs-index comparison (Lucene also serves reads only from flushed segments). Harness: `rust/lance/benches/mem_wal/fts/` (`run_fts_compare.sh`).

## Latest performance — Lance vs Lucene (Run A, p50 µs, 100k / 500k / 1M)

| metric | Lance | Lucene | result |
| --- | --- | --- | --- |
| term | 22.6 / 58 / 81 | 216 / 235 / 364 | **4.0–9.6× faster** |
| OR (3-term) | 410 / 1496 / 2656 | 1076 / 1923 / 3442 | **1.3–2.6× faster** |
| phrase p50 | 120 / 325 / 244 | 501 / 760 / 799 | **2.3–4.2× faster** |
| phrase p95 | 1007 / 4915 / 7352 | 1548 / 5406 / 7645 | **faster at all sizes** |

- **Recall@10:** term 1.000, OR 1.000 / 1.000 / 0.998 — exact (Lucene ~0.95 due to quantized norms).
- **Write (build docs/s):** ~8.7–9.1k vs Lucene ~12.5–13.2k (~1.4× slower at the default freeze; the no-flush / large-tail mode ingests faster than Lucene builds segments after the O(1) append fix below).
- **Memory (1M):** 1079 MB vs 849 MB (~1.27×).

## Read optimizations

- **Per-block "impacts" (tight block-max bound).** Each block stores the Pareto `(freq, dl)` frontier instead of a decoupled `(max_freq, min_dl)` pair, giving an exact per-block score bound so block-skipping actually fires. Flips single-term latency from near-linear to sublinear (term 1.2× slower → 4–9× faster).
- **MaxScore** for multi-term OR (replaces WAND). Terms split into essential / non-essential by max contribution; candidates come only from the essential suffix, non-essential lists are probed per candidate with early exit. ~2× over WAND.
- **Tight per-term bound** for the MaxScore split (max over the term's block-impact frontiers, via a cheap metadata-only walk) — replaces the loose decoupled bound so the split fires earlier.
- **FST term dictionary** (compact, prefix-shared) and **single-term block-max skip**.
- **Lower per-iteration cost**: cache each WAND/MaxScore lane's current doc to avoid repeated `cursor.doc()` calls in the hot loop.
- **Phrase**: reuse position buffers across candidate docs (was a `Vec<Vec<u32>>` + `to_vec()` per candidate) — phrase p95 −39%.
- **Read model**: `include_tail` (read-your-writes vs immutable-only) + threshold-based tail-skip pruning.

## Write optimizations

- **O(1) persistent tail append**: the per-term tail postings are a structurally-shared cons-list instead of a copy-on-append `Vec` (was O(batches²) per generation). No-flush ingest 4138 → ~15.9k docs/s.
- **O(1) block-structured batch log** for the visible-batch metadata (removes the analogous per-publish quadratic copy; keeps O(1) index).
- **Parallel freeze collection**: the per-term posting collection during a freeze runs across cores (rayon), +24% build throughput with a warm pool, no single-thread regression.
- **Flat-CSR insert** (`BatchTermBuilder`), `FxHashMap`, and a writer-side term-slot cache to cut insert allocation churn.

## Memory

- Sorted-delta doc-id posting encoding (SIMD bit-packing), FST term dict, varint block-metadata layout, and the impacts frontier; removed now-dead `PostingRef` fields. Residual ~1.27× is posting data Lucene also stores.

## Correctness

- **Fixed a multi-term WAND recall bug**: the block-max prune bounded a pivot by `lanes[..=pivot]` while the scorer summed *every* lane at the pivot doc; ties at the pivot could under-bound and drop true top-k OR results (measured 0.806 → 1.000 recall). Bound by exactly the contributing lanes; regression test added.
- SWMR visibility contract preserved (lock-free `ArcSwap`); MaxScore exactness guarded by exhaustive-equality and multi-match regression tests. All `mem_wal::index::fts` unit tests pass; `clippy` + `fmt` clean.

## Notes

- Run B (each engine's native analyzer) is not directly comparable — Tantivy vs Lucene tokenization diverges (top-10 overlap ~0.24), and Lance's Run-B write is tokenizer-bound.
- Remaining gaps (scoped follow-ups): write ~1.4× at the default freeze, memory ~1.3×, and fully closing phrase p95 via a shingle index (already ≈ parity-to-faster today).


touch-of-grey:
@jackye1995 could you review when you get a chance? This closes the FTS MemTable read-performance gap with Lucene (term 4–9×, OR 1.3–2.6×, phrase faster including p95), exact recall, plus write + memory improvements. Summary and benchmarks in the description.

jackye1995 (approved):
looks good to me!

PR #7036, opened 2026-06-01
fix(python): clamp target partition sizing

ddupg:
Fix Python target partition size inference to clamp the derived IVF partition count to `1..=4096`, matching the Rust path.

The previous helper used `4096` as a lower bound, which produced oversized partition counts for small datasets and missed the upper bound for large datasets.


Xuanwo (approved):
Thanks for the fix1

PR #7039, opened 2026-06-01
feat: expose getters for ScalarIndexExec

LuQQiu:
Expose `dataset()` and `expr()` getters on `ScalarIndexExec` so that downstream physical optimizer rules can inspect the node and build optimized execution plans on top of it.


wjones127 (approved):
Seems reasonable!

PR #7042, opened 2026-06-01
perf(file): serve global-buffer reads from the captured tail bytes

### LINKED ISSUE #7041: perf(file): serve global-buffer reads from the tail bytes already captured at open
## Problem

When a v2 Lance file is opened, `FileReader::read_all_metadata` already reads from the start of the global-buffer region (`schema_start`, i.e. global buffer 0) through EOF in a single optimistic tail read:

```rust
// rust/lance-file/src/reader.rs
let all_metadata_bytes =
    Self::optimistic_tail_read(&tail_bytes, schema_start, scheduler, file_len).await?;   // ~:685
let schema_bytes = all_metadata_bytes.slice(0..schema_size as usize);                    // ~:688
```

Because all global buffers are laid out contiguously starting at `schema_start` (buffer 0 is the schema, followed by every user buffer, then column metadata, the CMO/GBO tables, and the footer), `all_metadata_bytes` **already contains the bytes of every global buffer** whenever that region fits within the captured window (and `optimistic_tail_read` extends it with one read otherwise).

But the reader slices out only the schema and column metadata and **drops the rest**. A later `read_global_buffer(index)` then issues a fresh, standalone IO:

```rust
// rust/lance-file/src/reader.rs ~:481
pub async fn read_global_buffer(&self, index: u32) -> Result<Bytes> {
    let buffer_desc = self.metadata.file_buffers.get(index as usize)...?;
    self.scheduler
        .submit_single(buffer_desc.position..buffer_desc.position + buffer_desc.size, 0)
        .await
}
```

So even a global buffer that physically lives inside the already-fetched tail region costs an extra round trip. This affects every index type that stores side-data in a global buffer (IVF model, PQ/RaBitQ codebooks, bitmap/label-list payloads, and an upcoming FTS posting-group-offsets buffer in lance-format/lance#7040), and those reads are currently issued sequentially with `await` between them.

## Proposed solution

Retain the captured tail region and serve global-buffer reads from it when possible:

1. In `CachedFileMetadata` (or `FileReader`), retain the `all_metadata_bytes` `Bytes` together with its starting file offset (`schema_start`). `Bytes` is cheap to hold and slices are zero-copy.
2. In `read_global_buffer`, if `[position, position + size) ⊆ [schema_start, file_len)` was captured, return a slice of the retained bytes — **zero IO**. Otherwise fall back to the existing `submit_single`.

This mirrors the trick `optimistic_tail_read` / `decode_gbo_table` already use for the GBO table and schema. Result:

* Small/medium files: global buffers (including the schema, which is buffer 0) are served from the read that open already performs — no extra round trips.
* Large files where the buffer sits outside the captured window: unchanged (one dedicated read).

### Considerations

* Memory: retaining `all_metadata_bytes` holds roughly the metadata+buffers tail region (often \~`block_size`, object-store dependent) per open reader for its lifetime. Scope it as needed (e.g. only retain when small, or always — it is modest), and document the trade-off.
* This is purely a read-path IO optimization; the on-disk format is unchanged.

### Tests

* A file with one or more global buffers within the tail window: assert `read_global_buffer` issues no additional IO (e.g. via an instrumented/counting scheduler or object store).
* A file with a global buffer outside the captured window: assert it falls back to a dedicated read and returns correct bytes.
* Round-trip correctness for buffers served from the retained region vs. a fresh read.

### Context

Split out from lance-format/lance#7040 (FTS small-posting-list grouping), which stores group boundaries as a global buffer in the posting-list file and benefits directly from this optimization. The two can land independently.

wjones127:
When a v2 Lance file is opened, `read_all_metadata` already reads the tail region of the file (bounded by the object store block size) to decode the footer, GBO table, schema, and column metadata. Because all global buffers are laid out contiguously starting at the schema, small/medium files keep every global buffer inside that window. Previously `read_global_buffer` still issued a fresh standalone IO for each buffer, costing an extra round trip even when the bytes were already in memory.

This retains the captured tail `Bytes` together with its starting file offset on `CachedFileMetadata`. In `read_global_buffer`, if the buffer lives entirely within the captured window, it returns a zero-copy slice with no additional IO; otherwise it falls back to the existing dedicated read.

The on-disk format is unchanged.

## Tests

Parametrized `test_read_global_buffer` asserts IO counts via the object store's IO tracker:
- buffer within the tail window → served from memory, 0 read IOPS
- buffer larger than `block_size` → falls back, 1 read IOP

Both assert byte-level round-trip correctness.

Fixes #7041

### DIFF
--- a/rust/lance-file/src/reader.rs
+++ b/rust/lance-file/src/reader.rs
@@ -105,6 +105,18 @@ pub struct CachedFileMetadata {
     pub minor_version: u16,
     /// The actual total file size in bytes, as reported by the object store.
     pub file_size_bytes: u64,
+    /// The tail region of the file captured during open, spanning
+    /// `[tail_offset, tail_offset + tail_bytes.len())` (which always ends at EOF).
+    ///
+    /// `read_all_metadata` already reads this window (bounded by the object
+    /// store block size) to decode the footer, GBO table, schema, and column
+    /// metadata.  Because all global buffers are laid out contiguously starting
+    /// at the schema, small/medium files keep every global buffer inside this
+    /// window.  Retaining it lets `read_global_buffer` serve those buffers with
+    /// zero additional I/O.  `Bytes` is cheap to hold and slices are zero-copy.
+    pub tail_bytes: Bytes,
+    /// The file offset at which `tail_bytes` begins.
+    pub tail_offset: u64,
 }
 
 impl CachedFileMetadata {
@@ -166,7 +178,12 @@ impl DeepSizeOf for CachedFileMetadata {
             })
             .sum();
 
-        schema_size + buffers_size + column_metadatas_size + column_infos_size
+        // tail_bytes retains the captured tail window for zero-IO global buffer reads.
+        schema_size
+            + buffers_size
+            + column_metadatas_size
+            + column_infos_size
+            + self.tail_bytes.len()
     }
 }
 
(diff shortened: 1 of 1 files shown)


westonpace (approved):
Sweet optimization

PR #7043, opened 2026-06-01
perf(fts): group small posting lists into shared cache entries

### LINKED ISSUE #7040: perf(fts): group small posting lists into a single cache entry to reduce per-entry overhead
## Problem

FTS posting lists for rare terms are very small — measured median \~20 B, mean \~150–550 B depending on corpus size. This is expected: natural language follows Zipf's law, so most tokens appear in only a handful of documents.

The in-memory cache stores one entry per term. Each entry carries a key (e.g. `"postings-42"`, \~15–20 bytes) plus the cache infrastructure overhead. For a 20-byte posting list the key alone is comparable to the value, and the per-entry bookkeeping overhead often exceeds the value entirely. Across millions of tokens this is a significant waste of memory capacity.

The root cause is the cache granularity: one cache entry per token. The fix is to group multiple consecutive token rows into a single cache entry, so small posting lists are amortized across their neighbors.

## Proposed solution

**Group consecutive posting lists into \~4 KiB cache entries. Grouping is decided at write time using actual serialized byte sizes, and the resulting group boundaries are stored as a global buffer in the posting-list file. No breaking on-disk format change.**

### Why not compute groups at read time (revised)

The original sketch assumed `PostingListReader` eagerly loads `lengths: Vec<u32>` at `try_new()`, so group boundaries could be computed with no extra IO. That is no longer true: v2 posting metadata (`max_scores`, `lengths`) is now loaded lazily via a `OnceCell`, and the scoring path deliberately avoids the bulk pull (`posting_metadata_for_token` / `posting_len_for_token` issue single-row reads for cold partitions). Computing byte-based groups at read time would either force the bulk metadata load (regressing the cold single-term path) or require an extra windowed metadata read per cold miss.

Instead, compute the grouping **once at write time**, where the exact serialized byte size of every token's posting blob is already known (each posting list is a fully-compressed `LargeBinary` blob before it is appended). Store the result — the group boundaries — and read it back at open.

### Write-time grouping

In the v2 posting-list builder, after the per-token posting blobs are serialized:

1. Walk tokens in row order, accumulating the actual blob byte size.
2. Cut a group when the accumulated size reaches a target (default 4 KiB) or a maximum token count (cap, e.g. 64). Both configurable via `LANCE_FTS_`-prefixed env vars.
3. Clamp: a single posting list that alone exceeds the target becomes its own group (`[t, t+1)`). This is the current behavior for common terms.
4. Record the resulting `group_starts: Vec<u32>` (the first `token_id`/row of each group; monotonically increasing).

### Storing the boundaries

Store `group_starts` as a **global buffer** in the posting-list file, recording the buffer index in that file's schema metadata under a new key (e.g. `posting_group_offsets_buf`). This matches how IVF/PQ/RaBitQ/bitmap/label_list store side-data (bytes in a global buffer, 1-indexed buffer id in schema metadata) — none of them inline-vs-spill by size, so we shouldn't either.

* Encode `group_starts` **varint-delta** (deltas are group sizes, 1..=cap, \~1 byte each), keeping the buffer ≤ \~1.4 MB even at 10 M unique terms, and KBs in the common case.
* Global buffers are written contiguously just before the schema/column-metadata region (global buffer 0 *is* the schema), so for small/medium indexes the offsets buffer is fetched as part of the same tail read that already loads the schema — effectively zero extra IO at open once the lance-file reader exposes that captured region (tracked separately).

### Read path

In `PostingListReader::try_new`, if the metadata key is present, `read_global_buffer` + varint-delta decode into an in-memory `group_starts: Vec<u32>`.

In `posting_list(token_id, ...)`:

1. `token_id → [start, end)`: `group_starts.partition_point(|&s| s <= token_id) - 1` gives the group; `start = group_starts[k]`, `end = group_starts[k+1]` (or `num_rows`). Note in v2 `token_id` *is* the row offset, so the group range is directly the physical row range.
2. Cache key encodes the range: `postings-{start}-{end}` (stable; a config change that re-shapes groups simply misses old entries rather than serving a differently-shaped group).
3. On miss, one `read_range(start..end, columns)` for the whole group; deserialize all `end-start` posting lists.
4. Cache the group value as a small struct wrapping `Vec<PostingList>` with its own `CacheCodecImpl`. On a hit, index into it by `token_id - start`.

### Backwards / forwards compatibility

* **Additive only**: the new metadata key + global buffer are ignored by older readers.
* **Old v2 indexes / legacy v1** (no boundaries present): fall back to the existing per-token path unchanged. (First pass: fallback only; no read-time grouping for old files.)
* No change to the posting-list column layout, so old and new readers both read the row data identically.

### IO comparison

| Scenario | Before | After |
| -- | -- | -- |
| N nearby rare terms all cold | N `read_range` calls | 1 `read_range` call |
| 1 common term (large posting list) | 1 `read_range` call | same — single-token group (clamp) |
| Warm cache hit | memory lookup | memory lookup + offset into group |
| Open | read schema | read schema (+ group offsets, riding the same tail read in the common case) |

### Key files

* `PostingListReader` (struct, `try_new`, `posting_list`, `posting_batch`, `posting_list_range`): `rust/lance-index/src/scalar/inverted/index.rs`
* v2 posting-list builder / schema metadata: `rust/lance-index/src/scalar/inverted/builder.rs`
* Cache codec for the new group value: `rust/lance-index/src/scalar/inverted/cache_codec.rs`

### Open questions / decisions

* Target group size: default 4 KiB, configurable via `LANCE_FTS_` env var. Max token cap also configurable.
* Group value deserialization: eager (`Vec<PostingList>` at cache-fill time) per the existing `PostingList` cache pattern — avoids holding raw Arrow memory alongside decoded structs.
* The lance-file reader change to slice global buffers from the already-captured tail region (so the offsets buffer costs zero extra IO at open) is tracked separately in lance-format/lance#7041.

wjones127:
## Problem

FTS posting lists for rare terms are tiny (median ~20 B, mean ~150–550 B). The in-memory cache stored one entry per term, so the key (~15–20 B) plus per-entry bookkeeping often exceeded the value itself. Across millions of tokens this wastes significant cache capacity.

## Solution

Group consecutive posting lists into ~4 KiB cache entries.

- **Write time** (`builder.rs`): walk tokens in row order accumulating actual serialized blob bytes; cut a group at the byte target or a max-token cap. A term that alone exceeds the target gets its own group. The resulting `group_starts` are varint-delta encoded and stored as a **global buffer**, with the buffer id recorded in schema metadata under `posting_group_offsets_buf` — matching the IVF/PQ/bitmap precedent.
- **Read time** (`index.rs`): `PostingListReader::try_new` decodes `group_starts`. `posting_list` resolves a token to its group's row range via `partition_point`, reads the whole group in one `read_range`, and caches it as a `PostingListGroup` (new `CacheCodecImpl`). Prewarm populates the same group keys.

Both group target size and max token count are configurable via `LANCE_FTS_POSTING_GROUP_TARGET_BYTES` (default 4096) and `LANCE_FTS_POSTING_GROUP_MAX_TOKENS` (default 256).

## Compatibility

Additive only. Indexes written before this change have no metadata key and fall back to the existing per-token cache path unchanged. No posting-column layout change.

## Tests

- Accumulator units (target/cap cut, clamp, empty/single), `group_starts` and `PostingListGroup` codec roundtrips.
- Grouped vs. legacy-fallback reader produce identical postings for every token across group boundaries.
- End-to-end BM25 over a multi-group index: correct results + warm-cache == cold-cache.
- Existing FTS suites pass, including multi-fragment, multi-segment optimize, empty/empty-string edges, phrase, fuzzy, and the serializing-cache-backend test (round-trips the new group codec through real eviction).

The lance-file tail-slice optimization that makes the offsets buffer cost zero extra IO at open is tracked separately in #7041; this works without it (one small extra read at open).

Closes #7040

### DIFF
--- a/rust/lance-index/src/scalar/inverted/index.rs
+++ b/rust/lance-index/src/scalar/inverted/index.rs
@@ -2101,13 +2101,12 @@ impl PostingListReader {
             return None;
         }
         let start = starts[k - 1];
-        // The last group runs to the final row. In v2 the token count equals
-        // the row count, so this matches the `end` prewarm derives from the
-        // posting count, keeping warm- and cold-cache group keys identical.
-        let end = starts
-            .get(k)
-            .copied()
-            .unwrap_or(self.reader.num_rows() as u32);
+        // The last group runs to the final posting list. `self.len()` is the
+        // authoritative posting-list count (offsets length for v1, row count for
+        // v2), and prewarm derives the same `end` from it — so warm- and
+        // cold-cache group keys are identical by construction, not by the
+        // incidental v2 `num_rows == token_count` equality.
+        let end = starts.get(k).copied().unwrap_or(self.len() as u32);
         Some((start, end))
     }
 
@@ -2275,11 +2274,11 @@ impl PostingListReader {
         // grouping is active (issue #7040), per-token entries otherwise.
         match self.group_starts.as_ref() {
             Some(starts) => {
+                // The read path derives the last group's `end` from `self.len()`;
+                // match it here so both produce identical `PostingListGroupKey`s.
+                debug_assert_eq!(postings_by_token.len(), self.len());
                 for (k, &start) in starts.iter().enumerate() {
-                    let end = starts
-                        .get(k + 1)
-                        .copied()
-                        .unwrap_or(postings_by_token.len() as u32);
+                    let end = starts.get(k + 1).copied().unwrap_or(self.len() as u32);
                     let group = PostingListGroup::new(
                         postings_by_token[start as usize..end as usize].to_vec(),
                     );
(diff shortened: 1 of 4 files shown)


jackye1995 (approved):
looks good to me

rust/lance-index/src/scalar/inverted/index.rs
```diff
@@ -2028,6 +2084,59 @@ impl PostingListReader {
...
+            return None;
+        }
+        let start = starts[k - 1];
+        // The last group runs to the final row. In v2 the token count equals
+        // the row count, so this matches the `end` prewarm derives from the
+        // posting count, keeping warm- and cold-cache group keys identical.
+        let end = starts
+            .get(k)
+            .copied()
+            .unwrap_or(self.reader.num_rows() as u32);
```
LuciferYang: Here it uses `self.reader.num_rows()` while line 2282 uses `postings_by_token.len()`. The two paths must produce the *same* `[start, end)` for the last group or their `PostingListGroupKey`s diverge and prewarm is silently wasted for that group. They are equal today only because `reader.num_rows() == token_count` in v2 — an invariant the code comments assert but nothing enforces. A future format tweak (padding rows, a layout change) would break it without any test failing (correctness is preserved — the read path just re-reads — so the regression would be invisible).

wjones127: This is is a good point. I suspect any change to this part of the format would break more than just this, but it's worth making it robust to it.

PR #7047, opened 2026-06-02
fix: support manifests >5 GB via size-aware copy

### LINKED ISSUE #7197: Bug: manifest commit fails with EntityTooLarge for manifests >5 GB (CopyObject hard cap)
## Summary

`ExternalManifestStore` calls `object_store.copy(staging, final)` unconditionally on the manifest commit path at [`external_manifest.rs:128`](https://github.com/lance-format/lance/blob/main/rust/lance-table/src/io/commit/external_manifest.rs#L128) and [`external_manifest.rs:250`](https://github.com/lance-format/lance/blob/main/rust/lance-table/src/io/commit/external_manifest.rs#L250). This routes to S3's `CopyObject` API, which has a **5 GB hard cap** on source object size. Any manifest above this fails with `EntityTooLarge` and the commit cannot complete.

## Reproduction

Production failure observed on a Pinterest CTAS workload:

- Manifest path: `_versions/<version>.manifest`
- Reported `ProposedSize`: `14961429442` bytes (~14 GB)
- Error: `EntityTooLarge` from S3 on the staging→final manifest copy step
- No workaround except shrinking the manifest (which isn't always feasible — manifests grow with table version count and fragment count)

## Why this matters

- Affects any user whose manifest grows past 5 GB. Manifest size scales with table version history and fragment count, so this is reachable on long-lived production tables, not a corner case.
- Crashes the commit; there is no graceful degradation. The CTAS or write workflow fails entirely.
- No `object_store`-layer fallback today — the upstream `object_store` crate doesn't expose `UploadPartCopy`, so a workaround inside Lance is needed.

## Proposed fix https://github.com/lance-format/lance/pull/7047  

A `copy_size_aware` helper that:
- Keeps the cheap server-side `store.copy()` for sources <5 GiB (the common case, no regression)
- Falls back to read+rewrite via multipart upload for sources ≥5 GiB
- Accepts a `size_hint` so callers that already know the source size can skip an extra `head()` round-trip on the small-file fast path

## Same bug class as #6750, different code path

[#6750](https://github.com/lance-format/lance/pull/6750) fixed the analogous bug for **transaction file writes** (`write_transaction_file` was using `inner.put()`, hitting S3's 5 GB single-PUT limit). That PR was scoped to txn files and did not touch the manifest commit path.

lixmgl:
Fixes #7197.

## Summary

- `object_store.copy()` is called unconditionally on the staging→final manifest copy path. This routes to S3's `CopyObject` API, which has a ~5 GB hard cap. Manifests above this fail with `EntityTooLarge` — production case was a ~14 GB manifest with `ProposedSize 14961429442` on `_versions/...manifest`.
- Add `copy_size_aware`: keeps the cheap server-side `store.copy()` for sources below the limit, falls back to read+rewrite via a multipart upload for larger sources. The required `size` argument lets the caller skip an extra `head()` round-trip.
- The 5 GiB threshold is backend-agnostic, not S3-specific: S3's `CopyObject` and GCS's single-shot `Objects.copy` both cap at ~5 GiB, so the constant is named `MAX_SERVER_SIDE_COPY_BYTES`. Stores without such a cap (e.g. local FS) take the read+rewrite fallback above 5 GiB too; correctness is preserved, only the rare large copy is slower.
- Also tighten `MAX_UPLOAD_PART_SIZE` from `5 GiB` to `5 GiB - 1` so `LANCE_INITIAL_UPLOAD_SIZE=5368709120` can't trigger a single PUT of exactly 5 GiB on shutdown — which S3 also rejects.

Same bug class as #6750 (multipart-aware put for txn file writes), different code path.

## Test plan

New tests in `rust/lance/src/io/commit/external_manifest.rs` covering both the >5 GB read+rewrite fallback and the small-file fast path.

Related downstream issue: https://github.com/lance-format/lance-spark/issues/529


rust/lance-table/src/io/commit/external_manifest.rs
```diff
@@ -213,6 +216,120 @@ pub(crate) fn detect_naming_scheme_from_path(path: &Path) -> Result<ManifestNami
...
+/// existing `Err(NotFound { .. })` arms.
+///
+/// This is a workaround for the missing `UploadPartCopy` primitive in the
+/// upstream `object_store` crate. Once that lands, this helper can be
+/// deleted and the call sites can go back to plain `store.copy()`.
+async fn copy_size_aware(
+    store: &dyn OSObjectStore,
+    from: &Path,
+    to: &Path,
+    size_hint: Option<u64>,
```
westonpace: This hint should not be optional.  We had some optional hints in other paths to ease backwards compatibility.  However, since this is a new path, and the only caller already knows the size, and the fallback is to issue an additional IOP anyways (which the caller can easily do) we should make it mandatory.

rust/lance-table/src/io/commit/external_manifest.rs
```diff
@@ -213,6 +216,120 @@ pub(crate) fn detect_naming_scheme_from_path(path: &Path) -> Result<ManifestNami
...
+    let mut stream = store.get(from).await?.into_stream();
+
+    // From here on, errors must `abort()` the upload to avoid leaving an
+    // orphan multipart in S3 (which would otherwise incur storage charges
+    // until the bucket's lifecycle policy cleans it up).
+    //
+    // Note: this does NOT cover task cancellation — `MultipartUpload`'s
+    // upstream Drop is documented as a no-op for S3/GCS. Callers that
+    // need cancellation cleanliness should run this with a guard or
+    // switch to `object_store::WriteMultipart` (planned follow-up).
```
westonpace, line 290: Is there actually a follow-up issue created?  Can we specify the issue directly?

rust/lance-table/src/io/commit/external_manifest.rs
```diff
@@ -213,6 +216,120 @@ pub(crate) fn detect_naming_scheme_from_path(path: &Path) -> Result<ManifestNami
...
+            part_buf.extend_from_slice(&chunk[offset..offset + take]);
+            offset += take;
+
+            if part_buf.len() >= COPY_REWRITE_PART_SIZE {
+                let payload =
+                    std::mem::replace(&mut part_buf, Vec::with_capacity(COPY_REWRITE_PART_SIZE));
+                if let Err(e) = upload.put_part(Bytes::from(payload).into()).await {
+                    let _ = upload.abort().await;
+                    return Err(e);
+                }
```
westonpace, line 320: This could maybe be optimized to allow for concurrent calls to `put_part` but maybe that will be addressed by the previously mentioned follow-up?

rust/lance-io/src/object_writer.rs
```diff
@@ -47,8 +47,12 @@ fn max_conn_reset_retries() -> u16 {
...
 }
 
-/// Maximum part size in GCS and S3: 5GB.
-const MAX_UPLOAD_PART_SIZE: usize = 1024 * 1024 * 1024 * 5;
+/// Maximum body size for a single S3 PUT: strictly less than 5 GiB.
+/// AWS rejects single-PUT bodies of exactly 5 GiB (= 5 * 1024^3) with
+/// `EntityTooLarge`, so we clamp `LANCE_INITIAL_UPLOAD_SIZE` one byte
+/// below that threshold to keep the buffer-fills-to-clamp single-PUT
+/// path safe. See lance#6750 for the related txn-file write fix.
+const MAX_UPLOAD_PART_SIZE: usize = 1024 * 1024 * 1024 * 5 - 1;
```
westonpace, line 55: Is this limit S3-specific or common across all object stores?  Should we limit the slow path to S3?

westonpace (commented):
A few questions but nothing too worrisome.  Go ahead and ping me when you're ready for another review.

rust/lance-table/src/io/commit/external_manifest.rs
lixmgl: Done.

rust/lance-table/src/io/commit/external_manifest.rs
lixmgl: I reworded this from a "deferred to a follow-up" TODO into a plain inline NOTE rather than referencing a tracked issue. 
It's a cold path (only triggers for >5 GiB manifests) and the whole read+rewrite helper is a stopgap until object_store exposes UploadPartCopy, so I didn't think a standalone tracking issue was warranted — a self-contained note explaining the tradeoff seemed more useful than an issue link that'd likely sit untouched. 
Happy to file one if you'd prefer it tracked.

rust/lance-table/src/io/commit/external_manifest.rs
lixmgl: Right — left sequential for now, covered by the reworded note just above. It can be parallelized with a bounded JoinSet (mirroring object_writer.rs's LANCE_UPLOAD_CONCURRENCY) or sidestepped by switching to object_store::WriteMultipart, which handles concurrency and abort-on-drop for free. Deferred since this only runs for >5 GiB manifests and the helper itself is interim until UploadPartCopy exists.

rust/lance-io/src/object_writer.rs
lixmgl: It's not actually S3-specific — GCS's single-shot Objects.copy has the same ~5 GiB limit. So rather than gate the slow path to S3, I reframed the threshold as the most-conservative server-side-copy limit across backends and renamed the constant to MAX_SERVER_SIDE_COPY_BYTES.

lixmgl:
> A few questions but nothing too worrisome. Go ahead and ping me when you're ready for another review.

Thanks @westonpace for reviewing it. Please take another look with the updates.

westonpace (approved):
Thanks for your help!  Will merge assuming CI passes.

PR #7048, opened 2026-06-02
feat: add index segment_seq metadata

jackye1995:
Adds optional, name-scoped `segment_seq` metadata to index segments and assigns it in the commit/rebase layer. The change stores `LogicalIndexMetadata.max_segment_seq` in the manifest index section so sequence numbers are not reused after the highest physical segment for an index name is removed.

This backfills legacy segments on the next segment commit, uses the single writer feature flag `FLAG_INDEX_SEGMENT_SEQ` for both physical `segment_seq` and logical high-water metadata, and preserves existing Rust/Python/Java describe/build call patterns by assigning sequence numbers at commit time.

Discussion: https://github.com/lance-format/lance/discussions/7044

### DIFF
--- a/protos/table.proto
+++ b/protos/table.proto
@@ -113,6 +113,8 @@ message Manifest {
   // * 2: row ids are stable and stored as part of the fragment metadata.
   // * 4: use v2 format (deprecated)
   // * 8: table config is present
+  // * 16: dataset uses multiple base paths
+  // * 32: transaction files are disabled
   uint64 reader_feature_flags = 9;
 
   // Feature flags for writers.
@@ -123,6 +125,9 @@ message Manifest {
   //
   // The flag identities are the same as for reader_feature_flags, but the values of
   // reader_feature_flags and writer_feature_flags are not required to be identical.
+  //
+  // Writer-only flags:
+  // * 64: index segments use segment_seq
   uint64 writer_feature_flags = 10;
 
   // The highest fragment ID that has been used so far.
@@ -286,6 +291,14 @@ message IndexMetadata {
   // of index sizes without extra IO.
   // If this is empty, the index files sizes are unknown.
   repeated IndexFile files = 10;
+
+  // Monotonically increasing sequence number for this index segment, scoped to
+  // the index name.
+  //
+  // Starts at 1 for each index name. Missing means the segment was written
+  // before this feature or is uncommitted metadata that has not reached the
+  // commit layer yet.
+  optional uint64 segment_seq = 11;
 }
 
 // Metadata about a single file within an index segment.
--- a/protos/table.proto
+++ b/protos/table.proto
@@ -309,9 +309,19 @@ message IndexFile {
   uint64 size_bytes = 2;
 }
 
+// Metadata about one logical index name.
+message LogicalIndexMetadata {
+  // Logical index name.
+  string index_name = 1;
+
+  // Highest segment_seq that has ever been assigned for this logical index.
+  optional uint64 max_segment_seq = 2;
+}
+
 // Index Section, containing a list of index metadata for one dataset version.
 message IndexSection {
   repeated IndexMetadata indices = 1;
+  repeated LogicalIndexMetadata logical_indexes = 2;
 }
 
 // A DataFragment is a set of files which represent the different columns of the same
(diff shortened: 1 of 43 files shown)


LuQQiu:
Several high level comments
1. this PR does not have `max_segment_seq`, is it expected?
2. backfill when new segment arrive. Do we want to support backfill when no new segment arrive? how about re-commit with exact old segments == no op commit but also trigger backfill?

rust/lance-table/src/feature_flags.rs
```diff
@@ -20,8 +20,12 @@ pub const FLAG_TABLE_CONFIG: u64 = 8;
 pub const FLAG_BASE_PATHS: u64 = 16;
 /// Disable writing transaction file under _transaction/, this flag is set when we only want to write inline transaction in manifest
 pub const FLAG_DISABLE_TRANSACTION_FILE: u64 = 32;
+/// Index metadata uses segment_seq.
+pub const FLAG_INDEX_SEGMENT_SEQ: u64 = 64;
+/// Index section stores logical high-water marks for segment_seq assignment.
+pub const FLAG_INDEX_SEGMENT_SEQ_HIGH_WATER: u64 = 128;
```
wjones127: issue(blocking): FLAG_INDEX_SEGMENT_SEQ_HIGH_WATER isn't documented in the protobuf.

Also, do we even need two separate flags? It seems like we could just have one flag for this and require both to be set. What do you think?

protos/table.proto
```diff
@@ -296,9 +309,19 @@ message IndexFile {
   uint64 size_bytes = 2;
 }
 
+// Metadata about one logical index name.
+message LogicalIndexMetadata {
```
wjones127, line 313: praise: I'm in favor of adding this. It's been a long-time coming.

Eventually, we should move index details over to this, so there's just one copy per index.

Also, people have requested other fields like a top-level UUID and created_at date. I think we can add those in a future follow up.

rust/lance/src/index.rs
```diff
@@ -6399,79 +6371,51 @@ mod tests {
...
+        segment_seqs.sort();
+        assert_eq!(segment_seqs, vec![Some(1), Some(2)]);
+        assert_ne!(
+            dataset.manifest.writer_feature_flags & FLAG_INDEX_SEGMENT_SEQ,
+            0
+        );
+        assert_ne!(
+            dataset.manifest.writer_feature_flags & FLAG_INDEX_SEGMENT_SEQ_HIGH_WATER,
+            0
+        );
```
wjones127: question(blocking): It's not clear to me from these test—how does a user opt-in to this segment sequence flag? We can't enable it by default, because that means that users can't downgrade their Lance version—any previous Lance library will (or should) error when trying to write to it, given it won't recognize the writer feature flags.

wjones127 (changes requested):
The overall idea seems fine, but I have a couple of blocking questions

rust/lance-table/src/feature_flags.rs
jackye1995: Done. I collapsed this back to a single writer flag: `FLAG_INDEX_SEGMENT_SEQ` now covers both physical `IndexMetadata.segment_seq` and logical `LogicalIndexMetadata.max_segment_seq`, and `FLAG_UNKNOWN` is back to 128.

The format docs and the vote/design discussion were updated with the single-flag compatibility model.

rust/lance/src/index.rs
jackye1995: The flag is not enabled by a plain dataset write. It is set only once the manifest contains committed index segment sequence metadata, either physical `segment_seq` or logical `max_segment_seq`.

So the behavior is:

- Existing datasets without this metadata keep the flag clear and remain writable by older writers.
- A new writer enables the flag when it commits index segment metadata and writes `segment_seq` / `logical_indexes`.
- After that point, older writers should reject writes because they do not understand the writer feature flag and could otherwise drop the metadata.

I also added a test assertion that a freshly written dataset has `FLAG_INDEX_SEGMENT_SEQ` clear before index segments are committed, and set after segment commit.

jackye1995:
Closing, @vivek-bharathan will carry on this work

PR #7052, opened 2026-06-02
fix(lance-index): fix some flaky tests

XuQianJin-Stars:
### 1. Fix flaky `test_simple_index_nearest_centroid::case_1_f16` (rust)

The `test_simple_index_nearest_centroid::case_1_f16` was flaky because `HNSW` approximate search with ef=15 could not reliably find the nearest centroid when querying with 42.1f32 against f16-precision centroids.

The `f16` cast to `f32` introduces subtle precision differences that alter the `HNSW` graph structure, causing the search to follow incorrect paths and return ID 45 instead of 42.

Fix by using an exact match query value (42.0f32) for the f16 case, ensuring zero distance to the target centroid so `HNSW` always finds it. The f32 case retains the original 42.1f32 query.

Fixes flaky test introduced in a57ec81.

### 2. Fix flaky `test_create_inverted_index_progress_callback_error_after_completion_is_ignored` (python)

The test was failing because the `complete:write_metadata` progress event was being dispatched (and its callback error propagated) during the pump loop **before** the future completed — the future still had commit work to do after the builder emitted `stage_complete("write_metadata")`.

The `block_on_pumping` function only ignores callback errors in the final pump **after** the future resolves. But since the event arrives in the channel before the commit step finishes, it gets drained in the loop where errors propagate.

Fix by making `IndexProgressDispatcher::drain()` tolerate callback errors on `Complete`-type events. Complete events are purely informational — the stage's actual work is already done, so a callback failure should never abort the operation. `Start` and `Progress` events still propagate errors normally, preserving the "error before completion propagates" semantics.

### 3. Fix flaky `test_list_acquires_token_before_starting_underlying_stream` (rust)

The test was flaky on Windows CI because it relies on real-time assertions with a 5ms timeout, but Windows system timer resolution (~15.6ms) makes such tight timing unreliable.

The root cause is that `TokenBucketState` used `std::time::Instant` which is not controllable in tests. When the token bucket has no available tokens, the test asserts that `stream.next()` should block (timeout after 5ms), but on Windows the elapsed time measurement is too coarse.

Fix by:
- Switching `TokenBucketState.last_refill` from `std::time::Instant` to `tokio::time::Instant`
- Adding `#[tokio::test(start_paused = true)]` to the two timing-sensitive list throttle tests
- Adding `tokio = { workspace = true, features = ["test-util"] }` to dev-dependencies

With `start_paused = true`, tokio fully controls time advancement, making the tests deterministic regardless of OS timer resolution.

### DIFF
--- a/python/src/dataset.rs
+++ b/python/src/dataset.rs
@@ -3686,7 +3686,7 @@ impl PyWriteDest {
     }
 }
 
-#[derive(Debug, Clone, Copy)]
+#[derive(Debug, Clone, Copy, PartialEq)]
 enum IndexProgressEventType {
     Start,
     Progress,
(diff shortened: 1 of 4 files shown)


Xuanwo (approved):
Thank you!

PR #7054, opened 2026-06-02
feat(mem_wal): configurable HNSW build params for MemTable writers

touch-of-grey:
## Summary

Makes the MemWAL MemTable HNSW index configurable and tunes its default, plus fixes that unblock large flushes. The HNSW build parameters are treated as **writer config** (they only affect the MemTable a writer builds; each flushed generation is independent), exposed across Rust, Python, and Java.

### Feature: configurable HNSW build params
- `ShardWriterConfig.hnsw_params` — a per-index map (`num_edges` / `ef_construction` / `max_level`). `num_edges` is the HNSW graph degree (level 0 keeps `2 * num_edges`), matching Lance's existing `HnswBuildParams::num_edges` naming.
- Persisted into the existing `writer_config_defaults` map under `hnsw.<index>.<field>` keys — no proto/format change.
- Python: `hnsw_params={"idx": {"num_edges": 32}}` on `initialize_mem_wal` and `mem_wal_writer`. Java: `ShardWriterConfig.withHnswParams([...])`.

### Tuned default: `num_edges=16, ef_construction=100`
MemTable graphs rebuild on every flush, so build speed matters. A dbpedia-1M sweep (500k flush) showed 16/100 **dominates** the prior generic default 20/150 — ~27% faster flush (26.6s vs 36.6s) with an equal-or-better recall ceiling (0.9525 vs 0.9510 @ ef=256, the SQ8 limit). `12/64` flushes faster but drops the ceiling below 0.95.

### Fixes
- **arc-swap reclamation** of HNSW read snapshots — the lock-free reader retired snapshots without reclaiming them (O(N²/batch)), OOM-killing the 1M-row flush. Now bounded.
- **Flushed generations inherit the base dataset's storage version** (data + index files stay on one format), falling back to the default when no base dataset exists.
- **Fullzip encoding** for the flushed HNSW graph `List` columns (avoids a v2.x miniblock mis-decode at scale).

### Benchmarks
- `mem_wal_recall_hnsw`: real ShardWriter-flush mode, direct-in-Rust read of the flushed generation, and a `RAW_READ` path that calls the index search directly to isolate DataFusion read overhead.
- `disk_ann_compare.py`: Lance vs FAISS (HNSW+SQ8) vs DiskANN harness.

@jackye1995 PTAL.

### DIFF
--- a/java/lance-jni/src/mem_wal.rs
+++ b/java/lance-jni/src/mem_wal.rs
@@ -1246,7 +1246,8 @@ fn build_writer_config(env: &mut JNIEnv, config: &JObject) -> Result<ShardWriter
                 .l()?;
             let index_name: String = JString::from(index_name).extract(env)?;
             let num_edges = env.call_method(&item, "numEdges", "()I", &[])?.i()? as usize;
-            let ef_construction = env.call_method(&item, "efConstruction", "()I", &[])?.i()? as usize;
+            let ef_construction =
+                env.call_method(&item, "efConstruction", "()I", &[])?.i()? as usize;
             let max_level = env.call_method(&item, "maxLevel", "()I", &[])?.i()? as u16;
             Ok((
                 index_name,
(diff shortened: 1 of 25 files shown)


jackye1995 (approved):
looks good to me!

PR #7056, opened 2026-06-02
fix: specify roaring's patch version

HuaHuaY:
Commit https://github.com/lance-format/lance/commit/43c37805866d3d7523c090c1a35528ea7c48bc8b adds these code which requires that `RoaringBitmap` should implement `Eq`. But `roaring` didn't implement it until version 0.11.4 https://github.com/RoaringBitmap/roaring-rs/pull/341.
```rust
#[derive(Debug, Clone, PartialEq, Eq, Default)]
pub struct UpdatedFragmentOffsets(pub HashMap<u64, RoaringBitmap>);
```


Xuanwo (approved):
Thank you!

PR #7059, opened 2026-06-02
refactor: remove scoring autoprojection deprecation and commit to new behavior

westonpace:
Scoring autoprojection (automatically appending `_distance`/`_score` to the output when an explicit projection is specified) has been removed. The new permanent behavior matches what `disable_scoring_autoprojection` previously opted into:

- No explicit projection (SELECT *): `_distance`/`_score` are still appended automatically.
- Explicit projection without `_distance`/`_score`: they are NOT added.
- Explicit projection that includes `_distance`/`_score`: they appear as requested.

The `disable_scoring_autoprojection` flag is kept to avoid breaking callers that had opted in.  However, the legacy behavior is being removed (we have warned about this for nearly a year):
- `True` / `None`: silently accepted (new default is already active).
- `False`: raises an error directing users to include the column explicitly.


python/python/tests/test_dataset.py
```diff
@@ -3843,31 +3798,33 @@ def test_scoring_columns(tmp_path: Path):
     assert tbl.schema.names == ["vec", "_distance", "_rowid", "_rowaddr"]
 
     # If _rowid or _rowaddr are specified via projection, they stay in the specified
-    # order
+    # order; _distance is not added automatically since autoprojection was removed
     tbl = dataset.scanner(
-        columns=["_rowid", "vec"], nearest=nearest, disable_scoring_autoprojection=False
+        columns=["_rowid", "vec", "_distance"], nearest=nearest
     ).to_table()
```
westonpace, line 3804: We can remove cases that explicitly set `disable_scoring_autoprojection=False`.  These were intended to test the flag.  We don't support the legacy behavior anymore so the test case isn't needed.

rust/lance/src/dataset/scanner.rs
```diff
@@ -1772,23 +1770,16 @@ impl Scanner {
...
-    /// is specified or if the user explicitly includes the _score or _distance column
-    /// in the projection.
+    /// Previously, this method opted into the behavior where _score/_distance columns
+    /// are only included when no projection is specified or explicitly requested.
+    /// This is now always the case.
+    #[deprecated(
+        since = "0.0.0",
+        note = "Scoring autoprojection has been removed. This method is a no-op."
+    )]
     pub fn disable_scoring_autoprojection(&mut self) -> &mut Self {
```
westonpace, line 1776: Go ahead and remove this method.  Rust has compile-time checks so we don't have to keep old flags / methods around.

PR #7061, opened 2026-06-02
fix: retry S3 multipart request timeouts

Xuanwo:
S3 can return HTTP 400 with a `RequestTimeout` XML body for active multipart PUT part requests.

Treat those responses like existing connection reset transient failures so `ObjectWriter` retries the saved part buffer instead of failing long-running writes. This addresses failures seen while writing large inverted index partition files.

### DIFF
--- a/rust/lance-io/src/object_writer.rs
+++ b/rust/lance-io/src/object_writer.rs
@@ -254,15 +254,7 @@ impl ObjectWriter {
                         match res {
                             Ok(Ok(())) => {}
                             Err(err) => return Err(std::io::Error::other(err)),
-                            Ok(Err(UploadPutError {
-                                source: OSError::Generic { source, .. },
-                                part_idx,
-                                buffer,
-                            })) if source
-                                .to_string()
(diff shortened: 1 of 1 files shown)


cmccabe (approved):

cmccabe:
Hmm, this should work. But given the way the error message looks, it still might be nice to key off off the string  'idle connection' just in case. I can't think of how it would hurt, and it might help?

LuQQiu (approved):

rust/lance-io/src/object_writer.rs
```diff
@@ -346,6 +338,15 @@ struct UploadPutError {
     source: OSError,
 }
 
+fn should_retry_upload_put(source: &OSError) -> bool {
```
xuanyu-z, line 341: new to rust sdk, isn't there supposed to be a retryable flag? and also a way to internally overload SDK's retry category, In your case what about throttling like 503 that is not classified as retryable but can be retried?

PR #7062, opened 2026-06-02
perf(fts): share a top-k threshold across partitions during WAND

wkalt:
Each partition ran WAND from a cold threshold and built its own local top-k, so a common term paid full block-max work in every partition. Share an Arc<AtomicU32> floor across a query's partitions: each publishes its local k-th (fetch_max) and reads it back as a pruning floor. The k-th of the union is >= any single partition's k-th, so the shared value is a lower bound on the global k-th and never drops a real top-k doc.

### DIFF
--- a/rust/lance-index/src/scalar/inverted/wand.rs
+++ b/rust/lance-index/src/scalar/inverted/wand.rs
@@ -578,14 +578,14 @@ pub struct Wand<'a, S: Scorer> {
     docs: &'a DocSet,
     scorer: S,
     // Shared cross-partition top-k floor. Each partition publishes its local
-    // k-th score (`atomic_max_f32`) and prunes against the running value -- a
-    // lower bound on the global k-th, so it never drops a real top-k document.
+    // k-th score (`atomic_store_max_f32`) and prunes against the running value
+    // -- a lower bound on the global k-th, so it never drops a real top-k doc.
     shared_threshold: Option<Arc<AtomicU32>>,
 }
 
 /// Monotonically raise an f32 stored in an `AtomicU32` to `val`. CAS loop (not a
 /// bit-max) so it stays correct for negative scores -- BM25 idf can go negative.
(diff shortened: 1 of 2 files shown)


wkalt:
I have seen success with this against a 1.5B row subset of the fineweb dataset, with a segmented V2 FTS index of 10 segments and the default partition size of 2GB. With this index I issued MatchQueries for "government", "data", and "united" and the patch dropped the reported index comparisons by 90% (79.1K -> 7.7K), 94% (84.3K -> 5.1K), and 88% (82.9K -> 9.8K).

wkalt:
offline suggestion from @BubbleCal - make the partition searches serial and drop the atomic. This may be a win a win for aggregate QPS but could hurt latencies. Need to do an analysis on a large dataset.

rust/lance-index/src/scalar/inverted/wand.rs
```diff
@@ -576,6 +577,22 @@ pub struct Wand<'a, S: Scorer> {
...
     scorer: S,
+    // Shared cross-partition top-k floor. Each partition publishes its local
+    // k-th score (`fetch_max`) and prunes against the running value -- a lower
+    // bound on the global k-th, so it never drops a real top-k document.
+    shared_threshold: Option<Arc<AtomicU32>>,
+}
+
+/// Monotonic atomic max for an f32 in an `AtomicU32`. CAS loop (not a bit-max)
+/// so it stays correct for negative scores -- BM25 idf can go negative.
+fn atomic_fetch_max_f32(slot: &AtomicU32, val: f32) {
```
wjones127: nitpick: I found this name to be confusing. It says "fetch", but the signature looks more like a put?

wjones127 (approved):
This looks like a nice optimization

rust/lance-index/src/scalar/inverted/wand.rs
wkalt: done, thanks

PR #7064, opened 2026-06-02
perf(index): avoid HEAD call when opening vector indexes

### LINKED ISSUE #6944: Can potentially avoid HEAD call when opening vector indexes
In the `open_vector_index_v2` method (lance/src/index/vector.rs) the index files are opened.  However, we are not using the known sizes that are present in the manifest.  This means we then do a HEAD call to get the size when we are reading those files.  This could be avoided if we used `open_with_size` instead of `open`:

For example:

```
let aux_path = index_dir.clone().join(uuid).join(INDEX_AUXILIARY_FILE_NAME);
// The below call should be open_with_size!
let aux_reader = object_store.open(&aux_path).await?;
```

gstamatakis95:
Closes #6944

## Problem

Opening a vector index made an extra network request to find out how big the index file was. We do not need that request. The file size is already saved in the manifest, in `IndexMetadata.files`.

`Dataset::open_vector_index` opens `index.idx` just to read the footer and work out the format version. The plain `open` call does not know the file size, so the reader sends a HEAD request to get it. This happens for every vector index, not only HNSW. On the modern reader path the file is then opened a second time with the size already known, so the first HEAD was wasted. The same wasted HEAD also happened on the HNSW auxiliary file open (the case named in the issue), on the legacy remap path, and on legacy detail inference.

##  Changes

Added one small helper, `open_index_file`. It reads the size from the manifest (`IndexMetadata::file_size_map()`) and opens the file with `open_with_size`. If the size is not recorded, which is the case for older indices, it falls back to the plain `open`.

The helper is now used everywhere a vector index file is opened:

- `index.rs`, the main open that detects the format version (IVF_PQ, IVF_RQ, FLAT). This also removes a duplicate `file_size_map()` call in the same path.
- `vector.rs`, the `IVF_HNSW_PQ` and `IVF_HNSW_SQ` auxiliary file opens.
- `ivf.rs`, the legacy v1 `remap_index_file`.
- `details.rs`, the legacy `infer_vector_index_details` fallback.

## Testing

- New test `test_open_index_file_skips_head_when_size_known`. It wraps the store in a proxy that counts metadata reads against the index file. The result is 0 HEAD requests when the size is known and 1 HEAD on the older fallback path.
- `cargo fmt --all`
- `cargo clippy -p lance --tests -- -D warnings`, clean.
- `cargo test -p lance --lib index::vector::`, 208 passed.

### DIFF
--- a/rust/lance/src/index/vector.rs
+++ b/rust/lance/src/index/vector.rs
@@ -1986,50 +1986,14 @@ mod tests {
     /// Regression test for <https://github.com/lance-format/lance/issues/6944>.
     #[tokio::test]
     async fn test_open_index_file_skips_head_when_size_known() {
-        use std::sync::Mutex;
-        use std::sync::atomic::{AtomicUsize, Ordering};
-
-        use lance_core::utils::testing::{ProxyObjectStore, ProxyObjectStorePolicy};
         use lance_index::INDEX_FILE_NAME;
-        use lance_io::object_store::{ObjectStoreParams, ObjectStoreRegistry, WrappingObjectStore};
+        use lance_io::assert_io_eq;
+        use lance_io::object_store::{ObjectStoreParams, ObjectStoreRegistry};
 
-        /// Counts metadata reads (`get_opts`) against the index file.
-        #[derive(Debug)]
-        struct HeadCounter(Arc<AtomicUsize>);
-
-        impl WrappingObjectStore for HeadCounter {
-            fn wrap(
-                &self,
-                _store_prefix: &str,
-                original: Arc<dyn object_store::ObjectStore>,
-            ) -> Arc<dyn object_store::ObjectStore> {
-                let counter = self.0.clone();
-                let mut policy = ProxyObjectStorePolicy::new();
-                policy.set_before_policy(
-                    "count_index_metadata_reads",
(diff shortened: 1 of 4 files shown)


rust/lance/src/index/vector.rs
```diff
@@ -1946,6 +1978,157 @@ mod tests {
...
+
+        use lance_core::utils::testing::{ProxyObjectStore, ProxyObjectStorePolicy};
+        use lance_index::INDEX_FILE_NAME;
+        use lance_io::object_store::{ObjectStoreParams, ObjectStoreRegistry, WrappingObjectStore};
+
+        /// Counts metadata reads (`get_opts`) against the index file.
+        #[derive(Debug)]
+        struct HeadCounter(Arc<AtomicUsize>);
+
+        impl WrappingObjectStore for HeadCounter {
```
wjones127: suggestion: Instead of creating a one-off IO tracker, can you reuse `ObjectStore::io_stats_incremental()` and `assert_io_eq!()`? Those are our standard utilities to assert how many IOs happen.

wjones127 (approved):
Thanks for working on this. I have one optional suggestion. Will merge on Monday.

PR #7065, opened 2026-06-02
feat(scalar-index): push scan limit into index search

### LINKED ISSUE #6949: Push limit into scalar index query
Sometimes when many rows can be returned a scalar index query might take some time.  For example, a large range query against a btree index that hits most or all of the data.  If there is a limit on the scan then we could potentially push this into the index itself so that we know we can short-circuit early.

gstamatakis95:
Closes #6949

A scalar index query that matches many rows does the full work even when the scan has a `LIMIT`, which is only applied later in the plan. This pushes the limit down into the index search so a B-tree can stop early.

## What changed

- `SearchOptions::with_limit(...)` carries a best-effort cap. An index may ignore it, so other index types need no changes.
- `BTreeIndex` searches matching pages in order and stops once it has at least `limit` matches, leaving later pages unread. A search that acted on the limit returns `SearchResult::AtLeast`, never `Exact`, so a short-circuited scan cannot be mistaken for the complete match set. The limit is declined when NULL tracking is on, because a partial scan cannot report a complete null set.
- `ScalarIndexExpr::evaluate_limited` passes the limit only to a lone top-level lookup. It drops it at `And`, `Or`, and `Not`, where a partial match set cannot be complemented, intersected, or unioned soundly.
- `ScalarIndexExec` (default path) and `MaterializeIndexExec` (legacy path) accept the limit through `with_limit(...)` and surface it in EXPLAIN. `MaterializeIndexExec` collapses an `AtLeast` result to its confirmed lower bound only when it received a pushed limit, so lower bounds from other sources still get a full recheck.
- `Scanner::index_search_limit()` computes the pushed value, `limit + offset`, and wires it into both paths. `LogicalScalarIndex` forwards it to every segment.

## Correctness

Pushing the limit is only an optimization. A `GlobalLimitExec` still applies the exact limit and offset at the top of the plan, so the index only needs to return at least `limit + offset` rows.

The pushdown is declined in every case where the first N matches are not as good as any N matches: ordered scans, `ORDER BY`, vector or full-text search, aggregates, a refine filter or an inexact (recheck) result, a compound index query, deletions or an overlay block on a covered fragment, and an index whose fragment coverage is unknown.

## Tests

Unit tests cover the B-tree early stop and its `AtLeast` result, the declined limit under NULL tracking, forwarding across index segments, and the `MaterializeIndexExec` mask choice with and without a pushed limit. Scanner tests assert the pushed value reaches the exec nodes on both paths, and that the limit is suppressed for fragment subsets, retired fragments, compound queries, unknown coverage, deletions, and overlay blocks.


rust/lance/src/dataset/scanner.rs
```diff
@@ -4068,6 +4077,55 @@ impl Scanner {
...
+    fn index_search_limit(
+        &self,
+        filter_plan: &ExprFilterPlan,
+        relevant_fragments: &[Fragment],
+    ) -> Option<usize> {
+        let limit = self.limit?;
+        if limit <= 0 {
+            return None;
+        }
+        if self.ordering.is_some()
```
Xuanwo: This pushes the scalar-index limit even when the scan is still in its default ordered mode. A B-tree stops after collecting matches in index-value page order, so LIMIT/OFFSET can return a different subset than the full filtered scan would return.

gstamatakis95: Thanks for the feedback, pushed changes.

gstamatakis95:
@Xuanwo I think this is ready.

rust/lance/src/dataset/scanner.rs
```diff
@@ -4080,6 +4089,69 @@ impl Scanner {
...
+            return None;
+        }
+        if filter_plan
+            .index_query
+            .as_ref()
+            .is_some_and(|query| query.needs_recheck())
+        {
+            return None;
+        }
+        if relevant_fragments
```
Xuanwo: The limit gate only checks live fragments for deletion files, but scalar index segments can still contain retired fragment entries that are filtered out later. Those stale hits can consume the early-stop budget and make an unordered filtered LIMIT return fewer rows than requested.

rust/lance-index/src/scalar/btree.rs
```diff
@@ -2200,19 +2210,56 @@ impl ScalarIndex for BTreeIndex {
...
+                }
+            } else {
+                results.push(page_result);
+            }
+        }
 
         // Merge matching row IDs
         let selection = NullableRowAddrSet::union_all(&results);
 
         Ok(SearchResult::Exact(selection))
```
Xuanwo: The limited B-tree path can stop before reading all matching pages but still returns the result as exact. That weakens the index result contract and makes future callers liable to treat a partial match set as complete.

rust/lance/src/dataset/scanner.rs
```diff
@@ -4111,6 +4128,80 @@ impl Scanner {
...
+            .filter(|fragment| fragment.deletion_file.is_none())
+            .map(|fragment| fragment.id as u32)
+            .collect();
+        let covered_frags = self.fragments_covered_by_index_query(index_query).await?;
+        if !covered_frags.is_subset(&live_undeleted) {
+            return Ok(None);
+        }
+        let offset = self.offset.unwrap_or(0).max(0) as usize;
+        Ok(Some((limit as usize).saturating_add(offset)))
+    }
```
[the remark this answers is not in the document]
gstamatakis95: Addressed in 2abbbcda4, but not via union coverage. The shared `fragments_covered_by_index_query()` is also used by `partition_frags_by_coverage` and the prefilter bitmap, which need intersection ("covered by ALL indices") — switching `Or` to union would make those skip the full-scan fallback and return incomplete results.

Instead the gate now only pushes a limit for a single top-level `Query`. `evaluate_nullable` already drops the limit for `And`/`Or`/`Not`, so a compound query never early-stops; restricting the pushdown to a lone `Query` closes the deleted/retired-fragment hole without touching the shared coverage semantics, and loses no optimization. Guarded by `test_limit_not_pushed_for_compound_index_query`.

rust/lance/src/dataset/scanner.rs
gstamatakis95: The gate no longer looks only at live fragments' deletion files. `index_search_limit` now builds `live_undeleted` and requires the index's covered fragments  to be a subset of it.

rust/lance-index/src/scalar/btree.rs
gstamatakis95: `BTreeIndex::do_search` now returns `SearchResult::AtLeast` (a lower bound) whenever a limit is set, never `Exact`, so a short-circuited page scan can no longer be mistaken for the complete match set. `MaterializeIndexExec` materializes the `AtLeast` `lower` mask (every row in it is a confirmed match) and a downstream `GlobalLimitExec` still enforces the exact limit. `test_search_limited_short_circuits` asserts the limited result is `AtLeast` and reads fewer pages than an unlimited search.

gstamatakis95:
@Xuanwo I think this is ready.

PR #7067, opened 2026-06-02
fix: dedup active-memtable predicate-crossing stale reads (vector + FTS)

hamersaw:
## What

Fixes a stale-read phantom shared by the **vector** and **FTS** index search arms over the MemWAL **active memtable**, and routes the in-memory newest-per-PK / membership decisions through a single maintained MVCC index.

## The bug

The active memtable is an append log; a PK update is a later append with the same key. The in-memory secondary indexes — HNSW (vector) and the inverted index (FTS) — are **append-only**, so an updated row's old entries stay live. Both arms deduped with `WithinSourceDedupExec`, which only suppresses a stale row when the fresh version is **also in the result set**. When an update moves a row out of the query's match set (vector: far from the query; FTS: new text no longer matches), the fresh version isn't returned, so the stale version leaks. (`point_lookup` was immune — it already did the MVCC recency seek.)

## The fix

Maintain a per-memtable **MVCC PK-position index**: a lock-free arena skiplist keyed on `(compute_pk_hash(pk_columns), row_position)`, enabled on the active memtable and carried through freeze. The row position *is* the version stamp, so this reuses the exact primitive point-lookup trusts (`get_newest_visible`).

- **`NewestPkFilterExec`** keeps an index hit iff `get_newest_visible(pk_hash, max_visible) == row_position` — predicate-independent, snapshot-exact (keys on the scanner's latched `max_visible`). Wired into the active vector arm (replacing `WithinSourceDedupExec`) and the FTS arm (adding `with_row_id`).
- **point_lookup** falls back to the index (hash + value-equality collision guard) when no scalar BTree exists; its plan-path active arm uses `SortExec(_rowid DESC).fetch(1)` instead of `WithinSourceDedupExec`.
- **Cross-source block-list** probes the index per candidate (`GenMembership::Index`, snapshot-bounded) with no per-query set; flushed/base keep cached sets. `contains_pks` probes too.
- **Cleanup:** `WithinSourceDedupExec` / `DedupDirection` and the per-query PK-hash set builders (`pk_hashes()`, `in_memory_pk_hashes`) are deleted. Net negative LOC.

Hash keying covers single **and composite** PKs uniformly. The snapshot-bounded probe also closes a latent over-block where a not-yet-visible newer write could shadow an older visible copy.

## Tests

Both `#[ignore]`d repros un-ignored and passing; new `PkPositionIndex` unit tests, point-lookup-without-btree, index-sourced block-list, and snapshot-bounded **vanished-row guard** tests (within- and cross-source). Full `mem_wal` suite green; `cargo fmt` + `clippy -D warnings` clean.

## Deferred follow-ups

- Migrate `MemTableDedupScanExec`'s reverse-walk `HashSet` (filtered-read scan path) onto the same probe — the last within-source mechanism off the index; benchmark-gated.
- In-graph HNSW within-gen eviction (perf end-game; correctness is now exact).


jackye1995:
I originally removed the skiplist on primary key columns because it impacts perf, trading it off with stale data. But now that we've made it even faster than RocksDB, I agree with the general direction of this PR to add it back for more accuracy.

However, if we're going to maintain a per-PK skiplist anyway, I think we should just always maintain a BTree index (which is a skiplist in the memtable) on the primary key columns rather than introducing a separate `PkPositionIndex`. `BTreeMemIndex` already provides `get_newest_visible` — the exact same MVCC seek-and-stop primitive this PR reimplements with a hash key. Since in most cases the PK already has a BTree definition anyway, adding a second skiplist doubles the memory overhead for PK tracking.

If we always maintain a BTree on the primary key columns, `NewestPkFilterExec` and the block-list probes can use the existing BTree directly — no duplication. And if we flush it as a BTree index, this also makes the whole structure a true LSM tree — we essentially always maintain the sort order of the memtable, and even when it is flushed we still know the sort order.

jackye1995 (commented):
This mostly looks good to me! A couple of design suggestions:

### 1. Composite-key BTree instead of per-column intersection

The PR creates one single-column BTree per PK column (`pk_btrees: Vec<Arc<BTreeMemIndex>>`) and intersects them by position in `pk_newest_visible()`. This works, but supporting multi-column BTree indexes has been a long-hanging task we need to complete anyway — it would be better to do that as part of this rather than building a per-column intersection layer on top of single-column BTrees.

With a composite-key `BTreeMemIndex` (lexicographically ordered tuple key), one seek answers the "newest visible" question for any PK arity, and we avoid the `pk_btrees` vector, `enable_pk_index` / `__pk__*` auto-creation, and the walk-leading-column algorithm.

That said, I think for the initial impl it's fine to limit to single-column PKs and add composite-key BTree support as a follow-up — just don't build the multi-column intersection machinery that will get thrown away.

### 2. Flush the PK BTree to disk and drop `GenMembership`

The `GenMembership` enum splits in-memory (BTree probe) vs on-disk (hash set) generations. But we already flush user-configured BTrees to disk via `flush_with_indexes()`. If we always flush the PK BTree too, then every non-base generation has a BTree index — in-memory for active/frozen, on-disk for flushed. The base table is always the shadowed side (never provides a membership), so we can uniformly probe the BTree for every block-list check and remove `GenMembership` entirely.

Benefits:
- **No upfront O(n) scan** — `scan_pk_hashes` currently reads every row of every flushed generation to build a `HashSet<u64>` at plan construction. With a flushed BTree, there's zero construction cost. Each probe is O(log n) instead of O(1), but given the size of individual memtables that's more than good enough.
- **Unified caching** — the flushed on-disk indexes go through the existing session cache / 2-tier cache mechanism instead of a separate `FlushedMemTableCache` PK-hash cache path.
- **Simpler code** — no `GenMembership` enum, no `needs_values` branching, no conditional `ScalarValue` extraction in `PkHashFilterExec`.

hamersaw:
I updated this to build a PK index if either (1) it doesn't exist on the base table or (2) it is multi-column, in which case we build a single concatenated PK value index. Then we maintain this index on flushed generations and use it for duplicate PK lookups on both cross source dedup operations. The main advantage is that these use the regular Lance `Session` for caching, so we can evict from memory under contention, etc. The main things that should be validated during review are:
(1) We still have the `GenMemebership` enum because the active memtable and flushed generations maintain a different struct (ie. active memtable is in-memory vs. flushed generations are the cacheable on-disk representation). I think this is reasonable, but interested in others thoughts.
(2) There are new function son the BTree index to allow singular value lookups. This doesn't change the Lance index interface in any way, but IIUC the existing index lookup uses a much heavier bitmap based solution. This allows us cheap and efficient PK existence checks.

jackye1995 (approved):
looks good to me!

PR #7068, opened 2026-06-03
fix(fts): size cached posting lists by referenced slice

vivek-bharathan:
Previously our accounting was off since we were counting the entire buffer for each reference.


Xuanwo (approved):
Thank you!

PR #7074, opened 2026-06-03
fix: advance kmeans redo random init rng

ddupg:
KMeans redos created one RNG before the redo loop, but each random initialization cloned that same initial RNG state. As a result, redos greater than one could repeatedly start from the same randomly selected centroids instead of exploring distinct initializations.

This reuses the RNG mutably across redo attempts so each random initialization consumes the advanced RNG state.


Xuanwo (approved):
thanks for the fix.

PR #7076, opened 2026-06-03
perf(scanner): answer unfiltered count_rows from fragment metadata

### LINKED ISSUE #6970: Optimize count query plans with just fragment filter to be metadata-only
This plan seems to actually scan a lot of data, even though it could be satisfied with fragment metadata:

```rust
scanner.with_fragments(resolve_scan_fragments(dataset, fragment_ids)?);
scanner.count_rows().await? as usize
```

https://github.com/lance-format/lance/pull/6930#pullrequestreview-4376136132

LuciferYang:
## Summary

`Scanner::count_rows` always built and executed a count plan, even when the count could be satisfied from fragment metadata alone. For a plain count with no row-level filter or search this scanned row data unnecessarily — especially wasteful when the scanner is restricted to a subset of fragments via `with_fragments` (#6970).

This adds a metadata-only fast path to `Scanner::count_rows`: when nothing in the scan needs to inspect row data, it sums each fragment's live row count (physical rows − deletions, both tracked in fragment metadata) instead of building and executing a plan. `Dataset::count_rows(None)` already had such a fast path via `count_all_rows`; this brings the same benefit to the `Scanner` path (and, crucially, to fragment-restricted counts).

The fast path falls back to the existing count plan — preserving its results **and** its errors — whenever any of these is set:

- a row-level filter, vector / full-text search, `index_segments`, `fast_search`, or `include_deleted_rows`;
- `order_by` or `limit` / `offset` (the plan rejects these when combined with the count aggregate);
- a dynamic-only projection such as `SELECT 1` (also rejected by the plan).

The shared per-fragment summing is factored into `Dataset::count_fragment_rows`, which `count_all_rows` now also uses. Its fan-out uses the module-standard `io_parallelism()` bound, matching the sibling `count_deleted_rows`. In the common (new-format) case the count is answered entirely from cached metadata with zero I/O; only legacy/uncached fragments fall back to per-fragment metadata reads.

Closes #6970.

## Test plan

- New `test_count_rows_metadata_only` covering: whole dataset, fragment subset, empty fragment list, deletions (whole-dataset and subset), `include_deleted_rows` fallback, `limit` / `offset` / `order_by` error preservation, filtered-count fallback, and all-rows-deleted. It asserts **zero read/write IOPS** on the metadata fast path and `> 0` reads on the plan fallback.
- `cargo test -p lance --lib` — count, scanner (142), and dataset_io (52) suites pass.
- `cargo fmt --all --check` clean.
- `cargo clippy -p lance --all-targets -- -D warnings` clean.

### DIFF
--- a/rust/lance/src/dataset.rs
+++ b/rust/lance/src/dataset.rs
@@ -1477,12 +1477,24 @@ impl Dataset {
     }
 
     pub(crate) async fn count_all_rows(&self) -> Result<usize> {
-        let cnts = stream::iter(self.get_fragments())
+        self.count_fragment_rows(self.get_fragments()).await
+    }
+
+    /// Sum the live row count (physical rows minus deletions) of the given fragments.
+    ///
+    /// Reads only fragment metadata where it is available, falling back to per-fragment
+    /// I/O only when the cached counts are absent (older fragments lacking `physical_rows`,
+    /// or deletion files lacking `num_deleted_rows`).  That fallback I/O is fanned out at
+    /// the object store's configured I/O parallelism, matching the rest of this module.
+    pub(crate) async fn count_fragment_rows(&self, fragments: Vec<FileFragment>) -> Result<usize> {
+        // `io_parallelism()` reflects `LANCE_IO_THREADS` and can be 0; `buffer_unordered(0)` never
+        // polls its input, so clamp to at least 1 to keep the metadata count from hanging.
+        let io_parallelism = self.object_store.io_parallelism().max(1);
+        stream::iter(fragments)
             .map(|f| async move { f.count_rows(None).await })
-            .buffer_unordered(16)
-            .try_collect::<Vec<_>>()
-            .await?;
-        Ok(cnts.iter().sum())
+            .buffer_unordered(io_parallelism)
+            .try_fold(0usize, |acc, count| futures::future::ready(Ok(acc + count)))
+            .await
     }
 
     /// Take rows by indices.
--- a/rust/lance/src/dataset/scanner.rs
+++ b/rust/lance/src/dataset/scanner.rs
@@ -83,6 +83,7 @@ use tracing::{Span, info_span, instrument};
 use uuid::Uuid;
 
 use super::Dataset;
+use super::fragment::FileFragment;
 use crate::dataset::row_offsets_to_row_addresses;
 use crate::dataset::utils::SchemaAdapter;
 use crate::index::DatasetIndexInternalExt;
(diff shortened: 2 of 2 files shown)


rust/lance/src/dataset.rs
```diff
@@ -1385,12 +1385,21 @@ impl Dataset {
...
+    /// or deletion files lacking `num_deleted_rows`).  That fallback I/O is fanned out at
+    /// the object store's configured I/O parallelism, matching the rest of this module.
+    pub(crate) async fn count_fragment_rows(&self, fragments: Vec<FileFragment>) -> Result<usize> {
+        stream::iter(fragments)
             .map(|f| async move { f.count_rows(None).await })
-            .buffer_unordered(16)
-            .try_collect::<Vec<_>>()
-            .await?;
-        Ok(cnts.iter().sum())
+            .buffer_unordered(self.object_store.io_parallelism())
```
Xuanwo: The per-fragment count now uses the object store I/O parallelism directly, and that value can be zero when `LANCE_IO_THREADS=0`. In that configuration the buffered stream never polls its input, so plain metadata counts can hang instead of returning.

LuciferYang: Good catch. Clamped it with `.max(1)` — `io_parallelism()` passes `LANCE_IO_THREADS` through directly so it can be 0, and `buffer_unordered(0)` just stalls. Added a regression test that sets `LANCE_IO_THREADS=0` and asserts `count_rows` returns the right count, wrapped in a timeout so a future regression fails fast instead of hanging.

The same unguarded `io_parallelism()` → `buffered`/`buffer_unordered` pattern shows up in several other spots in this module (the doc here even says "matching the rest of this module"). To keep this PR focused on the count path I've only clamped here — I'll open a separate PR to clamp at the source (`io_parallelism()` returning `.max(1)`) so the rest are covered in one place.

LuciferYang:
Upstream landed this as a plan rewrite rather than a pre-plan short-circuit. `CountPushdown` sits in `get_physical_optimizer`'s rule list (`rust/lance/src/io/exec/optimizer.rs:211`) and rewrites a `COUNT(*)` aggregate into `CountFromMaskExec`, whose `count_live_rows` (`rust/lance/src/io/exec/count_from_mask.rs:277`) sums `f.count_rows(None)` over the covered fragments under `io_parallelism()`. That is the same metadata sum this PR added, reached through the planner instead of ahead of it, so a plan is still built but it no longer scans column data.

The fragment-subset shape #6970 asked about is covered as well: `count_pushdown.rs:185` derives a `fragment_scope` from `options.fragments` and restricts the count to it. The `.max(1)` clamp this PR carried for `LANCE_IO_THREADS=0` is also already in place, at the end of `io_parallelism()` (`rust/lance-io/src/object_store.rs:802`).

Closing as superseded.

PR #7079, opened 2026-06-03
perf(index): fix O(N log N) warm-cache regression in BitmapIndex

westonpace:
## Problem

Commit 4de5ce67d ("feat(index): serializable cache for Bitmap and LabelList scalar indices #6874") introduced a performance regression in `BitmapIndexPlugin::get_from_cache`. Every warm-cache hit against a bitmap scalar index now pays O(N log N) cost where N is the number of unique values in the column, instead of O(1).

The regression: the new implementation stored only the serializable `BitmapIndexState` (an Arrow `RecordBatch`) in the cache and reconstructed the full `BTreeMap<OrderableScalarValue, usize>` on every cache hit by calling `parse_lookup_batch`. For a column with 10M unique values this rebuilds the map on every query — including `IS NULL`, whose actual bitmap lookup is `(*self.null_map).clone()` and is otherwise O(1).

`parse_lookup_batch` is expensive because:
1. It calls `ScalarValue::try_from_array` for every row — one heap allocation per unique value.
2. It inserts into a `BTreeMap` — O(log N) comparisons per insert, O(N log N) total.

## Fix

**`BitmapIndex.index_map`**: Changed from `BTreeMap<OrderableScalarValue, usize>` to `Arc<BTreeMap<OrderableScalarValue, usize>>`. The map is immutable after construction, so sharing it behind an `Arc` is safe, and cloning is O(1).

**`BitmapIndexState`**: Added an `index_map: Arc<BTreeMap<...>>` field that is **not serialized** — the wire format is unchanged. It is populated eagerly:
- `from_index` (called by `put_in_cache`): `Arc::clone`s the map from the live `BitmapIndex` — O(1).
- `deserialize` (disk-backed cache backends): calls `parse_lookup_batch` once at deserialization time, which is already paying disk I/O cost.

**`into_bitmap_index`**: Now takes `&self` and simply `Arc::clone`s `self.index_map` — always O(1), no reconstruction.

**`get_from_cache`**: The intermediate `(*state).clone()` is removed since `into_bitmap_index` no longer consumes `self`.

`LabelListIndex` had the same dual-entry patch applied in a prior iteration; that is also reverted to the original single-entry approach (its `BitmapIndexState` path is unchanged by this PR).

## Test

Added `test_bitmap_cache_fast_path` to `bitmap.rs`:
- Creates a high-cardinality bitmap index (1 000 unique integers + 5 null rows)
- Calls `put_in_cache`, then `get_from_cache`
- Asserts `get_from_cache` returns `Some`
- Runs `IS NULL` and asserts the correct 5 null rows are returned

To measure the end-to-end impact, run the `bitmap / is_null / warm` case in `python/python/ci_benchmarks/benchmarks/test_count_rows.py` — latency should be close to `btree / is_null / warm`.

### DIFF
--- a/rust/lance-index/src/scalar/bitmap.rs
+++ b/rust/lance-index/src/scalar/bitmap.rs
@@ -7,7 +7,7 @@ use std::{
     collections::{BTreeMap, BinaryHeap, HashMap},
     fmt::Debug,
     ops::Bound,
-    sync::{Arc, OnceLock},
+    sync::Arc,
 };
 
 use arrow::array::BinaryBuilder;
@@ -158,7 +158,7 @@ impl CacheKey for BitmapKey {
 /// cache handle, a lazy reader, a fragment-reuse index). `BitmapIndexState`
 /// captures just the data needed to rebuild it: the value→file-offset map,
 /// the null bitmap, and the value type.
-#[derive(Debug)]
+#[derive(Debug, Clone)]
 pub struct BitmapIndexState {
     /// Value-to-row-offset lookup, encoded as an Arrow `RecordBatch` so we can
     /// reuse the existing IPC utilities for zero-copy round trips.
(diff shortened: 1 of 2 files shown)


westonpace:
The benchmark that detected this regression went from ~3s -> 1.7ms.

Xuanwo (approved):

wjones127 (approved):
Looks like a clean fix. Thank you!

Issue #6967, opened 2026-05-27
Ensure Datafusion queries are constrained to Datafusion target partitions


wjones127:
We should make sure that the maximum number of CPUs used by a query is dictated by the target number of partitions in DataFusion. This lets a system that wants to use fewer than the number of available CPUs for a query. For example, in a multi-tenant system restricting each query to use 8 vCPUs even if 128 are available.

Some nodes don't use partitions yet; even if they use separate thread pools we should make sure they respect this setting in their max concurrent tasks. Moving those nodes to use partitioning is out-of-scope for this issue.

wjones127:
Migration plan broken into sub-issues:

**Option 1 (interim — cap internal concurrency):**
- #7082 — Cap `FilteredReadExec`, `LanceScanExec`, and ANN search concurrency to `target_partitions`

**Option 2 (long-term — native DataFusion partitions):**
- #7081 — Add `EnforceDistribution` to Lance's physical optimizer (prerequisite for the below)
- #7083 — Migrate `FilteredReadExec` to native DataFusion partitions
- #7084 — Migrate `LanceScanExec` to native DataFusion partitions
- #7085 — Migrate `ANNIvfPartitionExec` to native DataFusion partitions

PR #7086, opened 2026-06-03
feat: add EnforceDistribution to physical optimizer

### LINKED ISSUE #7081: Add EnforceDistribution to Lance's physical optimizer
Lance's physical optimizer (`get_physical_optimizer()`) only includes `CoalesceTake`, `SimplifyProjection`, and `LimitPushdown`. It omits DataFusion's `EnforceDistribution` rule, so partition-count mismatches between exec nodes are handled by manually inserting `RepartitionExec` and `CoalescePartitionsExec` nodes in `scanner.rs`.

Adding `EnforceDistribution` will let the optimizer automatically insert the correct exchange nodes as exec nodes are migrated to native multi-partition output (see linked issues), removing the need for manual exchange insertion in scanner.rs.

This also requires wiring the session config's `target_partitions` into the `ConfigOptions` passed to the optimizer — currently `Default::default()` is used, which means the optimizer doesn't see the user's concurrency setting.

Part of the migration tracked in #6967.

wjones127:
## Summary

- Adds `EnforceDistribution` to `get_physical_optimizer()` after `LimitPushdown`. This rule automatically inserts `CoalescePartitionsExec`/`RepartitionExec` exchange nodes where distribution requirements aren't satisfied, laying groundwork to remove manual exchange insertion in `scanner.rs` as exec nodes migrate to multi-partition output.
- Wires `target_partitions` into the `ConfigOptions` passed to the optimizer. Previously `Default::default()` was used, ignoring the user's concurrency setting. `Scanner` now carries an optional `target_partitions` field (defaulting to `get_num_compute_intensive_cpus()`).
- Fixes `KNNVectorDistanceExec` non-batch mode to correctly declare `required_input_distribution() = SinglePartition`. The execute path reads a single input partition, so all candidates must be coalesced before distance computation; the prior `UnspecifiedDistribution` declaration was incorrect.

All snapshot tests updated: manual `RepartitionExec(1)` nodes preceding `SinglePartition`-requiring execs are replaced by `CoalescePartitionsExec`; unnecessary repartitions under `UnspecifiedDistribution` parents are removed. `assert_plan_equals` sets `target_partitions=1` to make plan snapshots machine-independent.

Closes #7081. Part of #6967.


rust/lance/src/dataset/scanner.rs
```diff
@@ -1379,6 +1389,17 @@ impl Scanner {
...
+    /// Set the target number of partitions for the physical optimizer.
+    ///
+    /// Overrides the default (`get_num_compute_intensive_cpus()`). Used by
+    /// `EnforceDistribution` and similar rules to decide how many partitions to
+    /// target when inserting exchange nodes. Set to 1 in tests that assert
+    /// specific plan shapes.
+    pub fn target_partitions(&mut self, n: usize) -> &mut Self {
+        self.target_partitions = Some(n);
+        self
+    }
```
westonpace, line 1400: Minor nit as we don't worry about the rust API _too_ much but this field name and definition is going to be confusing to users that aren't Datafusion-native and don't know what partitions are.  Maybe a name like `target_parallelism` would be more informative?

westonpace (approved):
Looks like a good transitional plan

rust/lance/src/dataset/scanner.rs
wjones127: I think that's a good point. I'll change this.

PR #7087, opened 2026-06-03
fix: cap exec-node parallelism to DataFusion target_partitions

### LINKED ISSUE #7082: Cap FilteredReadExec, LanceScanExec, and ANN search concurrency to target_partitions
Several exec nodes manage their own parallelism using process-wide values (`get_num_compute_intensive_cpus()`, `object_store.io_parallelism()`), ignoring DataFusion's `target_partitions` setting. This makes it impossible to constrain a query's CPU usage in multi-tenant scenarios even if the caller sets `target_partitions` on the session.

Changes needed:

- **`FilteredReadExec`**: Read `target_partitions` from `TaskContext.session_config()` in `execute()` and use it as a cap on the `OnePartitionMultipleThreads` num_threads.
- **`LanceScanExec`**: Cap `frag_parallelism` and the CPU decode `try_buffered` size with `target_partitions` from `TaskContext`.
- **`ANNIvfPartitionExec`**: Thread `target_partitions` into `effective_query_parallelism()` as a ceiling; also cap the `tokio::spawn` fan-out over delta indices.

This is the interim fix described in #6967. The long-term migration to native DataFusion partitions is tracked separately (see linked issues).

wjones127:
## Summary

`FilteredReadExec`, `LanceScanExec`, and `ANNIvfPartitionExec` each managed their own concurrency using process-wide CPU counts (`get_num_compute_intensive_cpus()`, `io_parallelism`), ignoring DataFusion's `target_partitions` session config. This makes it impossible to constrain query CPU usage in multi-tenant scenarios even when the caller sets `target_partitions` on the session.

Changes:
- **`FilteredReadExec`**: cap `OnePartitionMultipleThreads` num_threads with `target_partitions` in `obtain_stream()`
- **`LanceScanExec`**: add `parallelism_cap: Option<usize>` to `LanceScanConfig`; set it from `target_partitions` in `execute()`; apply it to the CPU decode `try_buffered` (v2) and `batch_readahead` (v1) — IO-bound paths (`frag_parallelism`, `fragment_readahead`) are intentionally not capped
- **`ANNIvfPartitionExec`**: cap delta index fan-out `.buffered()` with `target_partitions`
- **`ANNIvfSubIndexExec`**: thread `target_partitions` through `initial_search`/`late_search` into `effective_query_parallelism()`, where it caps `get_num_compute_intensive_cpus()` before computing partition search parallelism

## Does this change default parallelism?

No. `get_num_compute_intensive_cpus()` returns `num_cpus::get() - IO_CORE_RESERVATION` (default reservation = 2), where `num_cpus::get()` on Linux already reads cgroup CPU limits. DataFusion's default `target_partitions` is `available_parallelism()`, which is also cgroup-aware and returns the same logical CPU count. Since `cpus - 2 ≤ cpus`, `min(get_num_compute_intensive_cpus(), target_partitions)` equals `get_num_compute_intensive_cpus()` — the existing value — in all configurations.

The cap only takes effect when a caller explicitly lowers `target_partitions` below the default, which is exactly the multi-tenant use case this change is intended to support.

Closes #7082

### DIFF
--- a/rust/lance/src/io/exec/knn.rs
+++ b/rust/lance/src/io/exec/knn.rs
@@ -875,10 +875,11 @@ impl ExecutionPlan for ANNIvfPartitionExec {
     fn execute(
         &self,
         partition: usize,
-        _context: Arc<datafusion::execution::TaskContext>,
+        context: Arc<datafusion::execution::TaskContext>,
     ) -> DataFusionResult<SendableRecordBatchStream> {
         let timer = Instant::now();
 
+        let target_partitions = context.session_config().target_partitions();
         let query = self.query.clone();
         let ds = self.dataset.clone();
         let metrics = Arc::new(AnnPartitionMetrics::new(&self.metrics, partition));
@@ -927,7 +928,7 @@ impl ExecutionPlan for ANNIvfPartitionExec {
                     Ok::<_, DataFusionError>(batch)
                 }
             })
-            .buffered(self.index_uuids.len())
+            .buffered(self.index_uuids.len().min(target_partitions).max(1))
             .finally(move || {
                 metrics_clone.baseline_metrics.done();
                 metrics_clone
@@ -1148,8 +1149,14 @@ impl PartitionSearchControl for LatePartitionSearchControl {
     }
 }
 
-fn effective_query_parallelism(query: &Query, index: &dyn VectorIndex) -> usize {
-    let cpu_pool_size = get_num_compute_intensive_cpus();
+fn effective_query_parallelism(
+    query: &Query,
+    index: &dyn VectorIndex,
+    target_partitions: usize,
+) -> usize {
+    let cpu_pool_size = get_num_compute_intensive_cpus()
+        .min(target_partitions)
+        .max(1);
     effective_query_parallelism_for(
         query,
         cpu_pool_size,
@@ -1211,6 +1218,7 @@ impl ANNIvfSubIndexExec {
             .boxed()
     }
 
+    #[allow(clippy::too_many_arguments)]
     fn late_search(
         index: Arc<dyn VectorIndex>,
         query: Query,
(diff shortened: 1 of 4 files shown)


rust/lance/src/io/exec/knn.rs
```diff
@@ -927,7 +928,7 @@ impl ExecutionPlan for ANNIvfPartitionExec {
                     Ok::<_, DataFusionError>(batch)
                 }
             })
-            .buffered(self.index_uuids.len())
+            .buffered(self.index_uuids.len().min(target_partitions).max(1))
```
westonpace, line 931: Does the stream here actually spawn anything?  I wonder if this `buffered` has any real meaning.

westonpace (approved):
This is definitely a compromise still.  I think truly respecting partitions would be migrating the filtered read mode away from `OnePartitionMultipleThreads`.

The KNN change seems harmless since I'm not sure we're actually doing any parallelism there.

The scan change is also harmless as that node is only used on extremely legacy paths.

rust/lance/src/io/exec/knn.rs
wjones127: Good question. The IO from `open_vector_index` just runs in the buffered concurrently, but the call to `find_partitions_on_cpu` does call `spawn_cpu`.

wjones127:
> This is definitely a compromise still. I think truly respecting partitions would be migrating the filtered read mode away from `OnePartitionMultipleThreads`.

Indeed. This is a partial solution. We'll deal with others soon with the following issues:

1. https://github.com/lance-format/lance/issues/7081 (ready for review!)
2. https://github.com/lance-format/lance/issues/7083
3. https://github.com/lance-format/lance/issues/7084
4. https://github.com/lance-format/lance/issues/7085

PR #7089, opened 2026-06-03
perf: block-max pruning on the disjunctive WAND path

wkalt:
A single-term match query runs ~3x slower as an OR than as the logically identical AND, returning byte-identical top-k. The conjunctive WAND path skips blocks whose block-max cannot reach the top-k threshold; the disjunctive path had the machinery but never used it, so it scored every posting entry one at a time.

This adds the same skip to the union path: when the block-max upper bound over every iterator overlapping the current window cannot beat the threshold, it advances to the next block boundary instead of scoring each document. The bound includes the `head` iterators (later documents still inside the window), so a skip only fires when no document in the window can qualify and results are unchanged.

Single-term OR drops to AND-level latency (20x+ faster on common terms in a Zipf corpus). Phrase queries run on the AND path and are unaffected.

Covered by `test_or_single_term_block_skip_matches_and`, which asserts OR and AND return the same top-k and that pruning skips a block.

### DIFF
--- a/rust/lance-index/src/scalar/inverted/wand.rs
+++ b/rust/lance-index/src/scalar/inverted/wand.rs
@@ -860,6 +860,19 @@ impl<'a, S: Scorer> Wand<'a, S> {
                 continue;
             }
 
+            // Block-Max WAND pruning: skip the whole window when its score upper
+            // bound cannot reach the top-k threshold.
+            if self.threshold > 0.0 && self.or_block_window_max() <= self.threshold {
+                // On the final block `up_to` is the `u64::MAX` sentinel; step once
+                // there to avoid seeking past the valid doc id range.
+                let skip_to = match self.up_to {
+                    Some(up_to) if up_to < u32::MAX as u64 => up_to + 1,
+                    _ => target + 1,
+                };
+                self.push_back_leads(skip_to);
+                continue;
+            }
+
             let Some(doc) = self.lead.first().and_then(|posting| posting.doc()) else {
                 self.push_back_leads(target + 1);
                 continue;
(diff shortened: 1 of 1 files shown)


wkalt:
Benchmark: 500k-doc Zipf corpus (vocab 50k, 1-100 words/doc), single-term `match` at limit 10, run as OR vs AND. Same machine, 30 iters/measurement. OR and AND returned byte-identical top-k row_ids and scores in every case.

| term (frequency)      | OR before | OR after | speedup | AND     |
|-----------------------|-----------|----------|---------|---------|
| commonest term        | 17.96 ms  | 0.79 ms  | 22.7x   | 0.55 ms |
| 2nd commonest         | 16.28 ms  | 0.74 ms  | 22x     | 0.48 ms |
| mid-frequency         | 7.12 ms   | 0.40 ms  | 18x     | 0.32 ms |
| rarer                 | 0.87 ms   | 0.21 ms  | 4x      | 0.26 ms |

After the change OR is at parity with AND. The gap scales with term frequency: the more common the term, the longer the posting list the old union path scanned in full.

Correctness is covered by `test_or_single_term_block_skip_matches_and`, which asserts OR and AND return the same top-k and that the counted scored-doc total stays under one block-window (it scores all 384 docs without the pruning, confirming the skip engages). Full `lance-index` inverted suite (89 tests) passes; fmt and clippy -D warnings clean.

rust/lance-index/src/scalar/inverted/wand.rs
```diff
@@ -1721,6 +1770,76 @@ mod tests {
...
+        let expected = (hot.start..hot.start + 10)
+            .map(u64::from)
+            .collect::<Vec<_>>();
+        assert_eq!(or_hits, expected, "OR must return the top-k");
+        assert_eq!(or_hits, and_hits, "OR and AND must agree for a single term");
+        // Without pruning OR scores all `total` docs; with it the cold blocks are skipped.
+        assert!(
+            or_scored <= 2 * BLOCK_SIZE,
+            "expected pruning to skip a block, but scored {or_scored} of {total}",
+        );
```
wjones127, line 1840: thought: this seems sufficient for this PR, but it makes me wonder if we should have a general query test where we assert brute force results always equal results with WAND enabled. 🤔

wjones127 (approved):

PR #7093, opened 2026-06-04
perf(pq): wrap l2_targets in Arc to eliminate per-partition clones

yuju-huang:
ProductQuantizer stores `l2_targets: Option<Vec<L2Prepared>>` — a pre-transposed SoA copy of the PQ codebook (~768 KB for a 768-dim, 48 sub-vector, 256-centroid index).  Every call to `PQIndex::load()` (one per cached IVF partition) clones the entire ProductQuantizer, deep-copying this Vec.  The data is logically identical across all partitions (derived from the same global codebook) but was physically duplicated N times.

With 8000 partitions fully warmed, this wastes ~6 GB (768 KB × 8000).  We measured this empirically: lance 4.0.1 used 2.9× more memory per cached partition than lance 0.39 (1894 KB vs 443 KB), with the excess tracing directly to l2_targets.

Fix: change the field type to `Option<Arc<Vec<L2Prepared>>>`. Cloning a ProductQuantizer now bumps a reference count instead of copying megabytes; all partitions share one allocation.

Changes (5 lines in one file):
- Field type: Option<Vec<L2Prepared>> → Option<Arc<Vec<L2Prepared>>>
- build_l2_targets return type updated to match
- Construction: wrap collected Vec with Arc::new
- DeepSizeOf: deref Arc before iterating (**v).iter()
- build_l2_distance_table: targets.as_slice() to pass &[L2Prepared]

### DIFF
--- a/rust/lance-index/src/vector/pq.rs
+++ b/rust/lance-index/src/vector/pq.rs
@@ -60,10 +60,9 @@ impl DeepSizeOf for ProductQuantizer {
             + self.num_bits.deep_size_of_children(_context)
             + self.dimension.deep_size_of_children(_context)
             + self.distance_type.deep_size_of_children(_context)
-            + self
-                .l2_targets
-                .as_ref()
-                .map_or(0, |v| (**v).iter().map(|t| t.size_bytes()).sum())
+            // deep_size_of_children on the Arc de-duplicates shared allocations
+            // via the context, so partitions sharing one l2_targets are counted once.
+            + self.l2_targets.deep_size_of_children(_context)
     }
 }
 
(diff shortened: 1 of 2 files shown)


rust/lance-index/src/vector/pq.rs
```diff
@@ -62,7 +63,7 @@ impl DeepSizeOf for ProductQuantizer {
             + self
                 .l2_targets
                 .as_ref()
-                .map_or(0, |v| v.iter().map(|t| t.size_bytes()).sum())
+                .map_or(0, |v| (**v).iter().map(|t| t.size_bytes()).sum())
```
westonpace: This will lead to double-counting the RAM used when `l2_targets` is shared.  Can we instead implement `DeepSizeOf` for `L2Prepared` (if it isn't already).  Then we can do `self.l2_targets.deep_size_of_children(_context)`.  The default impl for `Arc` should automatically de-duplicate.

westonpace (changes requested):
This is a good find!

It also seems that verifying the in-memory vector index size is missing from our regression suite.  I think that could be tackled in a follow-up.

One nit on `DeepSizeOf` which is important to get right since it controls when we evict from cache.

rust/lance-index/src/vector/pq.rs
yuju-huang: done

westonpace (approved):
Will merge once CI finishes

PR #7094, opened 2026-06-04
fix(index): invalidate stale index cache entries after index replacement

geserdugarov:
## What

`CreateIndex` commits can retire existing index metadata, for example when `optimize_indices(...retrain...)` replaces an IVF index with a new UUID. If the old index had been prewarmed, the session index cache could still hold root entries and IVF partition entries for the removed UUID.

A later read in the same session could then observe stale cached index state that no longer matches the manifest. This clears cache entries for removed index UUIDs after the commit succeeds.

## Changes

- Add exact-key invalidation to `LanceCache` via `invalidate_with_key` and `invalidate_unsized_with_key`.
- Extend `CacheBackend` with `invalidate_entry` and implement it for the Moka backend and existing serializing test backends.
- On successful `Operation::CreateIndex` commits, invalidate cache entries for each removed index:
  - legacy vector index root entry
  - IVF state root entry
  - IVF partition sub-cache under `uuid/`
  - corresponding `uuid-fri_uuid/` entries when a fragment-reuse index is active
- Use the rebased transaction’s `removed_indices`, so conflict resolution invalidates the indices actually retired by the committed manifest.

## Notes

- Exact-key invalidation is needed because root index cache entries are keyed by UUID, while partition entries live under a cache prefix. Prefix invalidation alone cannot remove one root entry without risking unrelated cache entries.
- Invalidation happens only after the manifest write succeeds.
- This does not change index metadata, params, protos, or Python / Java APIs.

## Tests

- Cache unit tests for exact-key invalidation, sibling key/type preservation, prefix scoping, nested sub-cache preservation, missing-key no-op behavior, and unsized cache keys.
- An index integration test that prewarms an IVF-PQ index, retrains it through `optimize_indices`, and verifies the removed index UUID no longer has cached root or partition entries.

### DIFF
--- a/rust/lance/src/index.rs
+++ b/rust/lance/src/index.rs
@@ -12,7 +12,7 @@ use async_trait::async_trait;
 use datafusion::execution::SendableRecordBatchStream;
 use futures::FutureExt;
 use itertools::Itertools;
-use lance_core::cache::CacheKey;
+use lance_core::cache::{CacheKey, LanceCache};
 use lance_core::datatypes::Field;
 use lance_core::datatypes::Schema as LanceSchema;
 use lance_core::utils::address::RowAddress;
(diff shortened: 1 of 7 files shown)


geserdugarov:
@Xuanwo, @wjones127, hi!
If you don't mind, could you please review this PR related to index cache entries?

rust/lance/src/index.rs
```diff
@@ -384,6 +384,41 @@ impl CacheKey for MemWalCacheKey<'_> {
...
+    cache: &LanceCache,
+    removed_indices: &[IndexMetadata],
+    current_fri_uuid: Option<&Uuid>,
+) {
+    for removed in removed_indices {
+        let uuid_str = removed.uuid.to_string();
+
+        // Bare-UUID entries from caches created without FRI.
+        cache
+            .invalidate_with_key(&LegacyVectorIndexCacheKey::new(&uuid_str, None))
```
vivek-bharathan, line 400: nit: Also invalidate ScalarIndexDetailsKey?

vivek-bharathan (approved):

wjones127 (changes requested):
> A later read in the same session could then observe stale cached index state that no longer matches the manifest.

I'm confused by this sentence. Does this suggest a later read could read the wrong index cache entry? Or just that those entries will take up memory when they are likely no longer needed?

This PR seems to presuppose that there are no readers intentionally reading an old version of the table. I think that's a bad assumption. This can happen in two ways:

1. Because of MVCC, a query that started earlier than the write, but is still ongoing, may still reference that index and want to use it.
2. Users can and do checkout older versions of tables. Sometimes they might pin their readers to a version while writers do multiple operations on their table.

I worry this PR is going to hurt performance in those two cases. I wonder if a better solution would instead be a TTL expiry on cache entries. What do you think of that?

geserdugarov:
> This PR seems to presuppose that there are no readers intentionally reading an old version of the table. I think that's a bad assumption. This can happen in two ways:
> 
>     1. Because of MVCC, a query that started earlier than the write, but is still ongoing, may still reference that index and want to use it.
> 
>     2. Users can and do checkout older versions of tables. Sometimes they might pin their readers to a version while writers do multiple operations on their table.
> 
> 
> I worry this PR is going to hurt performance in those two cases. I wonder if a better solution would instead be a TTL expiry on cache entries. What do you think of that?

Thanks for review! You're right, but I need some time to work on this properly.

wjones127:
Closing in favor of this issue: https://github.com/lance-format/lance/issues/7490

PR #7096, opened 2026-06-04
feat!: return write summaries from file writers

Xuanwo:
FileWriter and IndexWriter now return write summaries from finish so callers can access the final object size without issuing an extra size lookup. This lets dataset fragment metadata and index writer callers propagate file sizes directly from the completed write path while keeping Python's existing LanceFileWriter.finish row-count behavior.


westonpace (approved):
Definitely a breaking change but it should be a clean one (compiler failures if anyone was relying on this) and it seems valuable to me.

PR #7097, opened 2026-06-04
fix(lance-index): keep TokenSet::next_id consistent after remap

luizfonseca:
## Summary

`TokenSet::remap` shifts surviving token ids down to close the gaps left by removed ids, but it never decrements `next_id`. So after a remap (called from `InnerBuilder::remap` during compaction, when row deletes empty out a posting list) the invariant that `next_id == tokens.len()` is broken: `next_id` stays at its pre-remap high water mark.

The next time `InnerBuilder::merge_from` absorbs a partition that introduces a token unseen by the target, it does

```rust
self.posting_lists.resize_with(self.tokens.len(), ...);  // sized to N
// ...
let new_token_id = self.tokens.get_or_add(token);        // returns stale next_id, N+k
let merged_posting = &mut self.posting_lists[new_token_id as usize];  // panics
```

`get_or_add` returns `next_id` for new tokens, which is now greater than `tokens.len()`, so the index access panics:

```
panicked at lance-index/src/scalar/inverted/builder.rs:856:57:
index out of bounds: the len is 703894 but the index is 708550
```

We've been seeing this hit 9 tables during `optimize_indices` on the FTS `content` column.

## Fix

One line: decrement `next_id` by `removed_token_ids.len()` at the end of `TokenSet::remap` to restore the invariant. The values are guaranteed to be in range — \`next_id >= tokens.len() + removed_token_ids.len()\` going in (every surviving and removed token consumed an id), so the subtraction can't underflow.

## Test

Added `merge_from_tolerates_remapped_target_token_set` in `inverted/builder` that:
1. Builds a target with two tokens, each in its own document.
2. Runs `InnerBuilder::remap` to delete one row — this empties one posting list, drops it, and calls `TokenSet::remap`. Without the fix, \`next_id\` is now stale.
3. Calls `merge_from` with a delta partition containing a brand-new token.
4. Without the fix, step 3 panics with the OOB above. With the fix, the merge succeeds and the assertion confirms both surviving and newly-introduced tokens are present.

`cargo test -p lance-index --lib scalar::inverted` — 133 passed, 0 failed.

## Notes

- The bug predates v6.0.0 — same code in 6.x and 7.x. Suggesting `critical-fix` label since it's a crash during background optimize that requires a downgrade or manual intervention to recover from.
- I haven't dug into whether `TokenSet` is ever serialized with `next_id` and read back; if it is, existing on-disk indices may carry the stale value forward. Happy to extend the fix to clamp on load if maintainers think that's worth doing — let me know.

### DIFF
--- a/rust/lance-index/src/scalar/inverted/builder.rs
+++ b/rust/lance-index/src/scalar/inverted/builder.rs
@@ -3088,6 +3088,70 @@ mod tests {
         Ok(())
     }
 
+    // Regression test for an FTS optimize panic seen in production:
+    //   panicked at scalar/inverted/builder.rs ...
+    //   index out of bounds: the len is N but the index is N+k
+    //
+    // The setup is a target builder whose token set was previously shrunk by
+    // an `InnerBuilder::remap` (i.e. some posting lists became empty after
+    // row deletes during compaction). Before the accompanying fix,
+    // `TokenSet::remap` shifted the surviving ids down but left `next_id`
+    // unchanged, so a later `get_or_add` for an unseen token returned an id
+    // past `tokens.len()`. `merge_from` then sized `posting_lists` to
+    // `tokens.len()` and indexed it with that stale id.
+    #[tokio::test]
+    async fn merge_from_tolerates_remapped_target_token_set() -> Result<()> {
+        let mut target = InnerBuilder::new(0, false, TokenSetFormat::default());
+        // Seed with two tokens, each appearing in exactly one doc so that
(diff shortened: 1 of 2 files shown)


luizfonseca:
A note that this seems to affect `0.30.0` and `=7.0.0` most of the times; albeit the code path exists in earlier versions too.

luizfonseca:
Saw that it has been addressed on https://github.com/lance-format/lance/pull/7115 so we can close it.

PR #7099, opened 2026-06-04
feat(namespace): pluggable auth providers for REST Namespace with AWS SigV4

### LINKED ISSUE #6583: Support pluggable authentication providers for REST Namespace (e.g. AWS SigV4)
(Note: Migrated from https://github.com/lance-format/lance-namespace/issues/327 for better visibility.)

## Motivation

Cloud-managed catalog services are becoming the norm. AWS S3 Tables provides an Iceberg-compatible catalog directly from the storage layer. GCP BigLake offers similar capabilities. These services use their native IAM authentication (e.g. AWS SigV4) rather than OAuth2 or static Bearer tokens.

For the Lance REST Namespace to integrate with these services — or any catalog service fronted by cloud API gateways with IAM auth — it needs to support request-signing authentication at the protocol level. This is essentially an enrichment of the **Credential Vending** mechanism in Lance Namespace: beyond vending storage credentials for data access, the namespace itself needs flexible authentication when talking to catalog endpoints.

This was discussed in https://github.com/lance-format/lance-namespace/issues/10 where the conclusion was that REST only needs OAuth2 since native implementations handle other auth flows. I believe this deserves reconsideration given how the ecosystem has evolved — expecting every cloud catalog deployment to set up an OAuth2 token exchange layer creates unnecessary friction.

## Current State

The OpenAPI spec (`docs/src/rest.yaml`) defines three `securitySchemes`: OAuth2 (client credentials), Bearer Token, and API Key. No scheme exists for request-signing mechanisms. The existing client implementations (Rust, Java, Python) all mirror this limitation — none of them support request-level signing.

While native implementations (`lance-namespace-impls`) can work around this by talking to cloud services directly (e.g. the Glue implementation uses AWS SDK with IAM credentials), relying on native implementations as the sole solution for cloud auth has several drawbacks:

- **Misplaced responsibility**: Catalog service providers should focus on catalog semantics, not on reimplementing the full Lance Namespace interface. If a cloud catalog already speaks the REST protocol, requiring a dedicated native implementation just for auth is unnecessary overhead.
- **Community fragmentation**: As more cloud vendors offer catalog services, each requiring its own native implementation, the ecosystem risks becoming fragmented with many parallel implementations that are hard to maintain and keep consistent.
- **Undermines the value of REST as a standard**: The REST Namespace is designed to be a universal interop layer. If auth limitations push users toward vendor-specific native implementations even when the catalog already speaks REST, it effectively encourages the community to bypass the REST standard.
- **Duplicated effort**: Each native implementation must reimplement protocol handling, error mapping, pagination, and other concerns that the REST client already handles well. The only missing piece is authentication.

## Proposal

Introduce a **pluggable, registry-based authentication provider** mechanism for the REST Namespace. Rather than hardcoding specific auth schemes, allow users to register an auth provider that can intercept and mutate outgoing HTTP requests.

Key design points:

- **Pluggable architecture**: Define an auth provider interface (e.g. a `RequestSigner` or `AuthProvider` trait/interface) that has access to the full HTTP request and can sign or modify it before execution. Users and cloud vendors can implement this interface for their specific auth mechanism.
- **Registry-based**: Auth providers should be registerable by name (e.g. `sigv4`, `gcp-iam`), similar to how namespace implementations are registered via `ConnectBuilder`. Configuration would drive which provider is activated and with what parameters.
- **Spec-level extensibility**: The OpenAPI spec should acknowledge that the REST Namespace security model is extensible beyond the three built-in schemes, potentially via a convention for custom security schemes.
- **Built-in providers for major clouds**: Ship first-party providers for widely-used mechanisms (starting with AWS SigV4) behind feature flags, so common use cases work out of the box.

## Use Cases

- **AWS S3 Tables**: S3 Tables exposes an Iceberg REST catalog interface authenticated via SigV4. Although it does not yet support the Lance format, it represents the direction cloud-managed catalogs are heading. A Lance REST Namespace client with SigV4 support could connect to services like this directly without an intermediate auth proxy.
- **AWS API Gateway + IAM Auth**: Catalog services deployed behind API Gateway using IAM authorization require SigV4-signed requests from clients.
- **GCP BigLake / Managed Catalogs**: GCP-managed catalog services that rely on Google IAM and signed requests.
- **General cloud-native deployments**: Any environment where IAM-based auth is already the standard and adding OAuth2 infrastructure is unnecessary overhead.

## References

- Prior discussion: https://github.com/lance-format/lance-namespace/issues/10
- [AWS SigV4 documentation](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_sigv-authentication-methods.html)
- [Apache Iceberg REST Catalog AuthManager](https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/rest/auth/AuthManager.java) — prior art for pluggable auth in a REST catalog spec

---

I'm happy to discuss this further and contribute to the design and implementation.

shiwk:
## Summary

Implements pluggable authentication for REST Namespace, with AWS SigV4 as the first provider. Closes #6583.

- Introduces `RestAuthProvider` trait with per-request `authenticate()` and connect-time `initialize()` for fail-fast credential validation
- Adds `SigV4AuthProvider` using the `aws-sigv4` crate, with credentials resolved via explicit properties or the standard AWS chain (env vars, profile, IMDS)
- Supports explicit static credentials via `rest.auth.sigv4.access-key-id`, `rest.auth.sigv4.secret-access-key`, and `rest.auth.sigv4.session-token` properties, taking precedence over the AWS default chain
- Wires through Python (PyO3) and Java (JNI) bindings — properties like `rest.auth.type=sigv4` pass transparently to Rust
- Includes `x-amz-content-sha256` header for S3-compatible service support
- `rest.auth.*` and `header.Authorization` are mutually exclusive; misconfiguration is caught at build time

## Configuration

```properties
rest.auth.type = sigv4
rest.auth.sigv4.region = us-east-1
rest.auth.sigv4.service = execute-api          # default
rest.auth.sigv4.access-key-id = AKIA...        # optional, overrides env
rest.auth.sigv4.secret-access-key = wJal...    # optional, overrides env
rest.auth.sigv4.session-token = FwoGZX...      # optional
```

## Test plan

- [x] Rust unit tests (14): AWS official test vectors, session token signing, explicit credentials via properties, injected provider precedence over static credentials, partial credentials rejection, host:port handling, auth conflict detection
- [x] Signature correctness cross-verified against botocore (Python) for both GET and POST requests
- [x] Python e2e tests (7): connect + operate, missing region error, botocore signature cross-verification, explicit credentials, session token, precedence over env, partial credentials rejection
- [x] Java e2e tests (7): connect + operate, missing region error, header format verification, explicit credentials, session token, credentials regardless of env, partial credentials rejection
- [x] cargo clippy and cargo test across feature matrix including `--all-features`

## Summary by CodeRabbit

* **New Features**
  * Added AWS SigV4 REST authentication (`rest.auth.type=sigv4`) with `none` option, supporting region/service, access key/secret, and optional session token.
  * REST authentication is now warmed up during namespace/connection setup so misconfiguration fails fast.

* **Bug Fixes**
  * Enforced mutual exclusivity between `rest.auth.*` and `header.Authorization`, plus stricter validation for missing/partial SigV4 credentials.
  * Ensure credentials provided via config take precedence over environment credentials.
  * Improved request signing behavior for empty request bodies and header injection ordering.

* **Documentation**
  * Updated REST authentication Javadocs and examples for the `rest.auth.*` configuration.

* **Tests**
  * Added Rust/Java/Python coverage for signing correctness, header contents (including security token), precedence, and failure messages.


shiwk:
@jackye1995  Could you help review this when you have time? Thanks!

shiwk:
Friendly ping @jackye1995 — this PR is ready for review. Also, CI workflows need maintainer approval to run. Thanks!

rust/lance-namespace-impls/src/rest.rs
```diff
@@ -80,58 +86,127 @@ impl std::fmt::Debug for RestClient {
...
-                    )
-                {
-                    request_headers.insert(header_name, header_value);
-                }
-            }
+            let context_headers = provider
+                .provide_context(&info)
+                .into_iter()
+                .filter_map(|(k, v)| k.strip_prefix(HEADERS_PREFIX).map(|n| (n.to_string(), v)));
+            apply_string_headers(request.headers_mut(), context_headers);
```
Xuanwo: Dynamic context headers are applied after SigV4 signing, so any provider that returns headers.Authorization or x-amz-* can replace the signer’s output. The final request may bypass SigV4 or fail authentication even though SigV4 was configured.

rust/lance-namespace-impls/src/rest_auth/sigv4.rs
```diff
@@ -0,0 +1,613 @@
...
+use aws_sigv4::sign::v4;
+use lance_core::Result;
+use lance_namespace::error::NamespaceError;
+use tokio::sync::OnceCell;
+use url::Url;
+
+pub const REGION_KEY: &str = "rest.auth.sigv4.region";
+pub const SERVICE_KEY: &str = "rest.auth.sigv4.service";
+pub const ACCESS_KEY_ID_KEY: &str = "rest.auth.sigv4.access-key-id";
+pub const SECRET_ACCESS_KEY_KEY: &str = "rest.auth.sigv4.secret-access-key";
```
Xuanwo, line 29: This adds secret-bearing SigV4 properties that flow through the same properties map printed verbatim by ConnectBuilder’s Debug implementation. Debug logs or assertion failures can expose long-lived AWS secrets or STS tokens.

rust/lance-namespace-impls/src/rest.rs
```diff
@@ -434,12 +548,25 @@ impl RestNamespaceBuilder {
...
 
     /// Build the RestNamespace.
-    ///
-    /// # Returns
-    ///
-    /// Returns a `RestNamespace` instance.
-    pub fn build(self) -> RestNamespace {
-        RestNamespace::from_builder(self)
+    pub fn build(self) -> Result<RestNamespace> {
+        let has_auth = self.auth_provider.is_some()
```
Xuanwo: Provider-specific rest.auth.* settings are ignored unless rest.auth.type is present, and the documented mutual exclusion with Authorization does not catch those settings. A typo or omitted type can silently run without SigV4 while still accepting conflicting legacy auth.

shiwk:
@Xuanwo Thanks for the review! Addressed in the latest push:
 
1. Header ordering: Reordered to base → context → auth. Context headers now participate in signing; auth applies last and overrides conflicts.
2. Debug redaction: auth_properties is now shown as <N keys, redacted> in Debug output.
3. Orphaned properties: build() now rejects rest.auth.* properties when rest.auth.type is missing.

shiwk:
@Xuanwo The requested changes are done.  Please take a look whenever you're free.

shiwk:
@Xuanwo @jackye1995  I just brought this PR up to date, as the previous version was falling behind master. Could you please re-trigger the CI and take a look? Thanks!

rust/lance-namespace-impls/Cargo.toml
```diff
@@ -73,11 +74,14 @@ uuid.workspace = true
 
 # Shared credential vending dependencies
 sha2 = { version = "0.10", optional = true }
+hex = { version = "0.4", optional = true }
 base64 = { version = "0.22", optional = true }
 
 # AWS credential vending dependencies (optional, enabled by "credential-vendor-aws" feature)
 aws-sdk-sts = { version = "1.38.0", optional = true, default-features = false, features = ["default-https-client", "rt-tokio"] }
 aws-config = { workspace = true, optional = true }
+aws-sigv4 = { version = "1", optional = true }
```
Xuanwo, line 83: Can we avoid using aws-sigv4 and instead use `reqsign` instead? That will make it much esier to expand support to azure and gce.

shiwk: Makes sense! reqsign's already in the tree via object_store and keeps this pluggable instead of locking to aws-sigv4. I'll look into migrating.

shiwk: Thanks @Xuanwo. I prototyped the reqsign migration (reqsign-core + reqsign-aws-v4 3.0.1) — it compiles and matches aws-sigv4 byte-for-byte on simple paths.

   The blocker is canonical-path encoding. Per SigV4, S3 is the single-encode exception; every other service (e.g. execute-api, s3tables) double-encodes the canonical URI.

   `aws-sigv4` handles both: it exposes [PercentEncodingMode](https://github.com/awslabs/aws-sdk-rust/blob/c65da2275777d0a53a78e3dba2d06cab0c4157ed/sdk/aws-sigv4/src/http_request/settings.rs#L66-L75)::{Double, Single} and defaults to Double, and the SDK's SigningOptions [defaults double_uri_encode: true](https://github.com/awslabs/aws-sdk-rust/blob/c65da2275777d0a53a78e3dba2d06cab0c4157ed/sdk/aws-runtime/src/auth.rs#L46-L73), with only S3 opting into single-encoding — so non-S3 services like execute-api and s3tables are signed correctly.

   `reqsign-aws-v4`, as far as I can tell, only supports the S3 single-encode path: [canonical_request_string](https://github.com/apache/opendal-reqsign/blob/3a50c2c1f269109c5901b121791c4b4489cdafca/services/aws-v4/src/sign_request.rs#L163-L177) decodes then encodes the path exactly once, with no mode switch — so non-S3 services aren't covered.

   **Impact**: this PR targets execute-api and URL-encodes user-provided names into the path. A name containing a percent-encoded character (space, /, @, =, non-ASCII) makes reqsign single-encode where the service expects double-encoding, so the signature is rejected.

   If my understanding of `reqsign-aws-v4` is correct, I'd suggest keeping `aws-sigv4` in this PR for now and revisiting reqsign once it gains a double-URI-encode option (like `aws-sigv4` PercentEncodingMode).

shiwk: > Thanks @Xuanwo. I prototyped the reqsign migration (reqsign-core + reqsign-aws-v4 3.0.1) — it compiles and matches aws-sigv4 byte-for-byte on simple paths.
> 
> The blocker is canonical-path encoding. Per SigV4, S3 is the single-encode exception; every other service (e.g. execute-api, s3tables) double-encodes the canonical URI.
> 
> `aws-sigv4` handles both: it exposes [PercentEncodingMode](https://github.com/awslabs/aws-sdk-rust/blob/c65da2275777d0a53a78e3dba2d06cab0c4157ed/sdk/aws-sigv4/src/http_request/settings.rs#L66-L75)::{Double, Single} and defaults to Double, and the SDK's SigningOptions [defaults double_uri_encode: true](https://github.com/awslabs/aws-sdk-rust/blob/c65da2275777d0a53a78e3dba2d06cab0c4157ed/sdk/aws-runtime/src/auth.rs#L46-L73), with only S3 opting into single-encoding — so non-S3 services like execute-api and s3tables are signed correctly.
> 
> `reqsign-aws-v4`, as far as I can tell, only supports the S3 single-encode path: [canonical_request_string](https://github.com/apache/opendal-reqsign/blob/3a50c2c1f269109c5901b121791c4b4489cdafca/services/aws-v4/src/sign_request.rs#L163-L177) decodes then encodes the path exactly once, with no mode switch — so non-S3 services aren't covered.
> 
> **Impact**: this PR targets execute-api and URL-encodes user-provided names into the path. A name containing a percent-encoded character (space, /, @, =, non-ASCII) makes reqsign single-encode where the service expects double-encoding, so the signature is rejected.
> 
> If my understanding of `reqsign-aws-v4` is correct, I'd suggest keeping `aws-sigv4` in this PR for now and revisiting reqsign once it gains a double-URI-encode option (like `aws-sigv4` PercentEncodingMode).

Any thoughts on this? @Xuanwo

PR #7108, opened 2026-06-04
perf(core): track fragment ids without scanning

beinan:
## Summary

- Add a persisted `next_fragment_id` manifest/protobuf counter so writers can allocate fragment IDs in O(1).
- Use the counter in append/reserve/rewrite/merge/restore paths while preserving legacy fallback behavior for older manifests.
- Keep index fragment bitmap range calculations aligned with the exclusive next-fragment counter and add targeted regression coverage.

## Testing

- `PATH=/Users/beinan/.rustup/toolchains/1.94.0-aarch64-apple-darwin/bin:$PATH cargo fmt --all`
- `PATH=/Users/beinan/.rustup/toolchains/1.94.0-aarch64-apple-darwin/bin:$PATH cargo clippy --all --tests --benches -- -D warnings`
- `PATH=/Users/beinan/.rustup/toolchains/1.94.0-aarch64-apple-darwin/bin:$PATH cargo test -p lance-table fragment_id --lib`
- `PATH=/Users/beinan/.rustup/toolchains/1.94.0-aarch64-apple-darwin/bin:$PATH cargo test -p lance build_manifest_advances_counter_for_preassigned_fragment_ids --lib`
- `PATH=/Users/beinan/.rustup/toolchains/1.94.0-aarch64-apple-darwin/bin:$PATH cargo test -p lance reserve_fragments_backfills_legacy_fragment_counters --lib`

### DIFF
--- a/protos/table.proto
+++ b/protos/table.proto
@@ -133,6 +133,11 @@ message Manifest {
   // For a single fragment, will be zero. For no fragments, will be absent.
   optional uint32 max_fragment_id = 11;
 
+  // The next fragment ID to assign.
+  //
+  // Writers use this counter directly instead of scanning the fragment list.
+  optional uint64 next_fragment_id = 22;
+
   // Path to the transaction file, relative to `{root}/_transactions`. The file at that
   // location contains a wire-format serialized Transaction message representing the
   // transaction that created this version.
--- a/rust/lance-table/src/format/manifest.rs
+++ b/rust/lance-table/src/format/manifest.rs
@@ -73,6 +73,12 @@ pub struct Manifest {
     /// None means never set, Some(0) means max ID used so far is 0
     pub max_fragment_id: Option<u32>,
 
+    /// The next fragment id to assign.
+    ///
+    /// When present, this lets writers allocate fragment ids in O(1) time
+    /// without scanning the fragment list.
+    pub next_fragment_id: Option<u64>,
+
     /// The path to the transaction file, relative to the root of the dataset
     pub transaction_file: Option<String>,
 
@@ -188,6 +194,7 @@ impl Manifest {
             reader_feature_flags: 0,
             writer_feature_flags: 0,
             max_fragment_id: None,
+            next_fragment_id: None,
             transaction_file: None,
             transaction_section: None,
             fragment_offsets,
@@ -219,6 +226,7 @@ impl Manifest {
             reader_feature_flags: 0, // These will be set on commit
             writer_feature_flags: 0, // These will be set on commit
             max_fragment_id: previous.max_fragment_id,
+            next_fragment_id: previous.next_fragment_id,
             transaction_file: None,
             transaction_section: None,
             fragment_offsets,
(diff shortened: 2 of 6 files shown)


beinan (commented):
`rust/lance/src/dataset/transaction.rs:2383-2387` now derives the high-water mark from `fragment_id`, but `fragment_id` only advances when `fragments_with_ids()` assigns an id to a fragment whose `id == 0` (`rust/lance/src/dataset/transaction.rs:1706-1710`). That is not true for compaction: `rust/lance/src/dataset/optimize.rs:1926-1935` pre-reserves fragment ids before the transaction is built, so the rewrite path often arrives with already-assigned non-zero ids. In that case `fragment_id` never moves and this branch skips updating `max_fragment_id`/`next_fragment_id`, leaving the manifest with a stale counter even though new fragment ids were committed. The old `update_max_fragment_id()` scan covered this. Please either track the max over `final_fragments` regardless of whether ids were preassigned, or keep the scan on rewrite/compaction paths.

beinan (commented):
1. `rust/lance/src/dataset/transaction.rs:2383-2388` now computes the high-water mark only from the local `fragment_id` cursor, but `handle_rewrite_fragments()` preserves any fragment IDs already present in `group.new_fragments` (`fragment_id` only advances when `id == 0`). That means a rewrite/merge path that feeds pre-numbered fragments through this helper can commit a manifest whose `next_fragment_id` is stale even though `final_fragments` contains higher IDs. The old scan-based update caught this case. Please fold the actual `final_fragments` IDs into the max calculation, or otherwise account for preserved IDs here.

protos/table.proto
```diff
@@ -133,6 +133,11 @@ message Manifest {
   // For a single fragment, will be zero. For no fragments, will be absent.
   optional uint32 max_fragment_id = 11;
 
+  // The next fragment ID to assign.
+  //
+  // Writers use this counter directly instead of scanning the fragment list.
+  optional uint64 next_fragment_id = 22;
```
wjones127, line 139: question(blocking): What's the purpose of this? When would this be anything other than `max_fragment_id + 1`?

beinan: Good catch, `next_fragment_id` is now persisted so writers can allocate IDs in O(1), and it falls back to `max_fragment_id` or the fragment list only for older manifests.

beinan (commented):
## Security Review
No security issues identified in the added lines. The diff does not introduce authentication, authorization, secret handling, deserialization, network request, CSRF, or information-disclosure surfaces.

## Code Review
No blocking correctness issues found. I checked the counter propagation across manifest serialization, append/update/overwrite/rewrite/merge/reserve/restore paths, and the index fragment bitmap fallback behavior.

Verification run locally with the pinned Rust toolchain path:
- `PATH=/Users/beinan/.rustup/toolchains/1.94.0-aarch64-apple-darwin/bin:$PATH cargo fmt --all --check`
- `PATH=/Users/beinan/.rustup/toolchains/1.94.0-aarch64-apple-darwin/bin:$PATH cargo test -p lance-table fragment_id --lib`
- `PATH=/Users/beinan/.rustup/toolchains/1.94.0-aarch64-apple-darwin/bin:$PATH cargo test -p lance build_manifest_advances_counter_for_preassigned_fragment_ids --lib`
- `PATH=/Users/beinan/.rustup/toolchains/1.94.0-aarch64-apple-darwin/bin:$PATH cargo test -p lance reserve_fragments_backfills_legacy_fragment_counters --lib`

beinan:
I changed my mind, will make a new impl soon

PR #7110, opened 2026-06-04
fix(mem-wal): fence predecessor with a WAL sentinel on claim

hamersaw:
## Problem

A newly-claiming writer only bumps the manifest `writer_epoch` — it writes **nothing** into the WAL, and WAL slot keys are position-only (no epoch in the path). So until the successor actually writes a slot, a fenced predecessor's next `append` lands in the empty next slot, the `PUT-IF-NOT-EXISTS` succeeds, and `append` returns `Ok` **without** a fence check (the `check_fenced` call only fires on the `AlreadyExists` branch).

That window is a correctness hole: the predecessor false-acks a write that later dies at the seal-time manifest CAS — and if the successor already replayed past that position, the entry is orphaned (data loss for an acked write).

## Fix

On claim (`epoch >= 2`), drop a **data-less sentinel** WAL entry at the current tip, **before** replay:

- The predecessor's next `append` now collides at that slot and surfaces the fence via the existing `AlreadyExists -> check_fenced` path.
- Writing the sentinel *before* replay guarantees any predecessor entry that landed *below* the sentinel is recovered by replay rather than orphaned.
- A lost slot race (a predecessor/concurrent claimer wins the probed slot) re-probes one past the winner; that entry then sits below the sentinel and is still replayed.
- Sentinels carry zero batches (empty-schema Arrow IPC + `writer_epoch`/marker metadata) and are skipped by replay's existing empty-batch guard.

Epoch 1 (a fresh shard) has no predecessor, so the sentinel is skipped there.

This lets writers rely on collision for fencing instead of issuing a per-put fence-check GET on every append.

## Test

Adds `test_fence_sentinel_fences_predecessor_without_successor_write`, which exercises the exact race: a successor claims a higher epoch and drops a sentinel **without writing any data batch**, and the predecessor's next append is fenced. Also asserts the sentinel reads back as zero batches and the successor's own writes land after it.

### DIFF
--- a/rust/lance/src/dataset/mem_wal/wal.rs
+++ b/rust/lance/src/dataset/mem_wal/wal.rs
@@ -39,6 +39,10 @@ use super::memtable::batch_store::{BatchStore, StoredBatch};
 /// Key for storing writer epoch in Arrow IPC file schema metadata.
 pub const WRITER_EPOCH_KEY: &str = "writer_epoch";
 
+/// Marks a WAL entry as a data-less fence sentinel (observability only;
+/// replay skips sentinels via their empty batch list).
+pub const FENCE_SENTINEL_KEY: &str = "fence_sentinel";
+
 /// Watcher for batch durability using watermark-based tracking.
 ///
 /// Uses a shared watch channel that broadcasts the durable watermark.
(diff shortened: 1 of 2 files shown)


wjones127 (approved):
This makes sense to me. Thanks for the clear PR description 😄

PR #7115, opened 2026-06-04
fix(fts): reset TokenSet next_id and total_length after remap

vivek-bharathan:
During an FTS index build, merge_existing_segments() can panic with something like "index out of bounds: the len is 936 but the index is 1077". This happens when there has been an update/delete on the table that causes some token to be evicted from the posting lists. Fixed by updating the associated bookkeeping

### DIFF
--- a/rust/lance-index/src/scalar/inverted/index.rs
+++ b/rust/lance-index/src/scalar/inverted/index.rs
@@ -1638,17 +1638,26 @@ impl TokenSet {
             }
         };
 
+        let mut retained_length = 0;
         map.retain(
-            |_, token_id| match removed_token_ids.binary_search(token_id) {
+            |token, token_id| match removed_token_ids.binary_search(token_id) {
                 Ok(_) => false,
                 Err(index) => {
                     *token_id -= index as u32;
+                    retained_length += token.len();
                     true
                 }
             },
         );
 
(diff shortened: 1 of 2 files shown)


wjones127:
@vivek-bharathan In the future, if there is a bug, could you make a bug report issue first? If you make a report that has the original error message, this helps users find if the same issue has already been reported and/or fixed. Also helps us maintainers too when we get bug reports.

Also, could you write a PR description going forward? Give the reviewer a little context on why the change is being made. For context, in our release notes we link to all PRs that were merged and included in a release. So when a user reads and clicks through, they expect to see some description of what was changed and what the impact was.

rust/lance-index/src/scalar/inverted/index.rs
```diff
@@ -1638,17 +1638,26 @@ impl TokenSet {
             }
         };
 
+        let mut total_length = 0;
```
wjones127: nit: I wonder if this should be called `new_length` or `retained_length`. There's a bit of cognitive overhead of seeing this total variable but being in suspense about what it is for until we read the comment below.

wjones127 (approved):
This looks like a good fix. I have an optional comment you can address. Otherwise I can merge this soon.

PR #7123, opened 2026-06-05
feat(index): support configurable multi-segment FM-Index builds

beinan:
## Summary

- Add `num_segments` parameter to FM-Index creation that distributes dataset fragments across multiple independent index segments
- Each segment is a complete FM-Index covering a disjoint set of fragments, enabling incremental indexing and segment merge
- Wire FM-Index into Lance's segmented index infrastructure (`commit_existing_index_segments`, `merge_existing_index_segments`)
- Add `FMINDEX` type mapping and `num_segments` kwarg in Python bindings

### Usage
```python
# Single segment (default, backward compatible)
ds.create_index("text_col", index_type="FMIndex")

# Multi-segment: fragments distributed across 4 segments
ds.create_index("text_col", index_type="FMIndex", num_segments=4)
```

## Test plan
- [ ] Existing 20 FM-Index unit tests pass (verified locally)
- [ ] Build compiles clean (`cargo check -p lance -p lance-index`)
- [ ] `cargo fmt` passes
- [ ] CI passes
- [ ] Test multi-segment build with real dataset

### DIFF
--- a/rust/lance-index/src/scalar/fmindex.rs
+++ b/rust/lance-index/src/scalar/fmindex.rs
@@ -24,6 +24,8 @@ use std::cmp::Reverse;
 use std::collections::{BinaryHeap, HashMap};
 use std::sync::{Arc, OnceLock};
 
+use lance_select::RowSetOps;
+
 use arrow_array::RecordBatch;
 use arrow_schema::{DataType, Field};
 use async_trait::async_trait;
@@ -1366,9 +1368,17 @@ impl ScalarIndex for FMIndexScalarIndex {
         &self,
         new_data: SendableRecordBatchStream,
         dest: &dyn IndexStore,
-        _: Option<OldIndexDataFilter>,
+        old_data_filter: Option<OldIndexDataFilter>,
     ) -> Result<CreatedIndex> {
-        let texts = collect_texts(new_data).await?;
+        let mut texts = collect_texts(new_data).await?;
+        if let Some(filter) = old_data_filter {
+            texts.retain(|(row_addr, _)| match &filter {
+                OldIndexDataFilter::Fragments { to_keep, .. } => {
+                    to_keep.contains((*row_addr >> 32) as u32)
+                }
+                OldIndexDataFilter::RowIds(valid) => valid.contains(*row_addr),
+            });
+        }
         let files = write_partitioned_fmindex(&texts, dest).await?;
         Ok(CreatedIndex {
             index_details: prost_types::Any::from_msg(&pb::FmIndexIndexDetails {}).unwrap(),
@@ -1377,7 +1387,9 @@ impl ScalarIndex for FMIndexScalarIndex {
         })
     }
     fn update_criteria(&self) -> UpdateCriteria {
-        UpdateCriteria::only_new_data(TrainingCriteria::new(TrainingOrdering::None))
+        UpdateCriteria::requires_old_data(
+            TrainingCriteria::new(TrainingOrdering::None).with_row_addr(),
+        )
     }
     fn derive_index_params(&self) -> Result<ScalarIndexParams> {
         Ok(ScalarIndexParams::for_builtin(BuiltinIndexType::FMIndex))
--- a/rust/lance-index/src/scalar/fmindex.rs
+++ b/rust/lance-index/src/scalar/fmindex.rs
@@ -58,7 +58,7 @@ const SENTINEL_BYTE: u8 = 0xFF;
 const SA_SAMPLE_RATE: usize = 32;
 
 fn fmindex_partition_path(partition_id: u64) -> String {
-    format!("part_{partition_id}_fmindex.lance")
+    format!("part_{partition_id}_fm.lance")
 }
 
 // ── Bitvector with O(1) rank ─────────────────────────────────────────────────
@@ -1267,7 +1267,7 @@ impl FMIndexScalarIndex {
             if let Some(id) = f
                 .path
                 .strip_prefix("part_")
-                .and_then(|r| r.strip_suffix("_fmindex.lance"))
+                .and_then(|r| r.strip_suffix("_fm.lance"))
                 .and_then(|s| s.parse::<u64>().ok())
             {
                 pfiles.push((id, f.path.clone()));
@@ -1297,7 +1297,7 @@ impl Index for FMIndexScalarIndex {
     }
     fn as_vector_index(self: Arc<Self>) -> Result<Arc<dyn VectorIndex>> {
         Err(Error::invalid_input_source(
-            "FMIndex is not a vector index".into(),
+            "Fm is not a vector index".into(),
         ))
     }
     async fn prewarm(&self) -> Result<()> {
@@ -1305,14 +1305,14 @@ impl Index for FMIndexScalarIndex {
     }
     fn statistics(&self) -> Result<serde_json::Value> {
         Ok(serde_json::json!({
-            "type": "FMIndex",
+            "type": "Fm",
             "num_partitions": self.partitions.len(),
             "total_bwt_len": self.partitions.iter().map(|p| p.fm.wavelet.len).sum::<usize>(),
             "total_docs": self.partitions.iter().map(|p| p.fm.row_ids.len()).sum::<usize>(),
         }))
     }
     fn index_type(&self) -> IndexType {
-        IndexType::FMIndex
+        IndexType::Fm
     }
     async fn calculate_included_frags(&self) -> Result<RoaringBitmap> {
         let mut frags = RoaringBitmap::new();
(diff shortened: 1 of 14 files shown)


rust/lance/src/index/scalar_logical.rs
```diff
@@ -738,4 +738,213 @@ mod tests {
...
+        let result = logical.search(&query, &NoOpMetricsCollector).await.unwrap();
+        let row_addrs = match result {
+            SearchResult::Exact(row_addrs) => row_addrs,
+            other => panic!(
+                "expected exact result from segmented fmindex, got {:?}",
+                other
+            ),
+        };
+        let match_count = row_addrs.true_rows().row_addrs().unwrap().count();
+        assert!(
```
jackye1995: This assertion is too weak and appears to hide a correctness bug in segmented FMIndex search. FMIndex training currently does not request `_rowid` or `_rowaddr`, so `collect_texts` falls back to a local counter starting at 0 for each segment, and search returns those ids as row addresses. With this fixture, `"quick"` should match 5 rows, but segmented search can collapse local ids across segments and still satisfy `>= 3`.

Can we fix FMIndex to store global row addresses / stable row ids and make this test assert the exact expected rows or count?

rust/lance/src/index/create.rs
```diff
@@ -560,6 +568,155 @@ impl<'a> CreateIndexBuilder<'a> {
...
+                index_details: Some(Arc::new(created_index.index_details)),
+                index_version: created_index.index_version as i32,
+                created_at: Some(chrono::Utc::now()),
+                base_id: None,
+                files: created_index.files,
+            });
+        }
+
+        self.dataset
+            .commit_existing_index_segments(&index_name, &column, segment_metadatas)
```
jackye1995: This commit path does not preserve the normal `replace=true` behavior. The standard `execute` path removes all existing same-name indices when `replace` is set, but `commit_existing_index_segments` only removes overlapping same-type segments. Replacing an existing same-name BTree/bitmap/etc. index with segmented FMIndex can leave mixed metadata under one index name, and empty/untrained FMIndex coverage can leave old same-type segments too.

Can this path either build the transaction with the same `removed_indices` as the normal path, or extend the segment commit API with explicit replace semantics?

rust/lance/src/index/create.rs
```diff
@@ -560,6 +568,155 @@ impl<'a> CreateIndexBuilder<'a> {
...
+        }
+        let scalar_params = self.params.as_any().downcast_ref::<ScalarIndexParams>()?;
+        let params_json = scalar_params.params.as_deref()?;
+        let json: serde_json::Value = serde_json::from_str(params_json).ok()?;
+        json.get("num_segments")?.as_u64().map(|n| n as u32)
+    }
+
+    /// Build FM-Index with multiple segments, each covering a subset of fragments.
+    async fn execute_multi_segment_fmindex(&mut self, num_segments: u32) -> Result<IndexMetadata> {
+        let column_input = &self.columns[0];
```
jackye1995, line 570: This bypasses the standard `execute_uncommitted` validation that requires exactly one column. For `num_segments > 1`, an empty column list will panic here, and a multi-column request silently indexes only the first column instead of returning the existing error.

Can we run the same `self.columns.len() != 1` validation before entering the segmented FMIndex path?

rust/lance/src/index/scalar/fmindex.rs
```diff
@@ -0,0 +1,72 @@
...
+        Error::invalid_input(format!(
+            "CreateIndex: segment {} is missing field ids",
+            segments[0].uuid
+        ))
+    })?;
+    let column = dataset.schema().field_path(field_id)?;
+
+    let mut fragment_bitmap = RoaringBitmap::new();
+    for segment in &segments {
+        fragment_bitmap |= segment.fragment_bitmap.as_ref().cloned().ok_or_else(|| {
```
jackye1995, line 35: This merge path should mirror the btree merge behavior and intersect stored segment coverage with the dataset's current live fragments. As written, it unions historical fragment ids and passes them to `build_scalar_index`; after compaction retires one of those fragments, the rebuild can fail with `No fragment with id ...`.

Can we use effective live coverage here, and add a compaction/retired-fragment merge test similar to the btree coverage?

jackye1995 (changes requested):
Found a correctness issue in segmented FMIndex search and a few behavioral divergences from the existing segmented-index pattern. Details inline.

rust/lance/src/index/scalar_logical.rs
beinan: Good catch! I fixed this by:
1. Updating `FMIndexPlugin` to request global row addresses (`.with_row_addr()`) during training.
2. Updating `collect_texts` to parse the global `_rowaddr` (falling back to `_rowid` if needed), rather than falling back to local counter starting at 0.
3. Adding a new `search_row_addrs` method on `FMIndex` to return the full 64-bit row address during logical searches, avoiding any 32-bit offset truncation.
4. Tightening the test assertion to assert exactly `5` matching rows.

rust/lance/src/index/create.rs
beinan: Fixed. I replaced the `commit_existing_index_segments` calls with explicit `TransactionBuilder` + `apply_commit`, collecting all same-name indices into `removed_indices` when `replace=true`. This matches the standard `execute` path behavior and ensures replacing a BTree/bitmap index with a segmented FMIndex properly cleans up all old metadata.

rust/lance/src/index/create.rs
beinan: Added. `execute_multi_segment_fmindex` now runs the same `self.columns.len() != 1` check at entry, matching `execute_uncommitted` validation.

rust/lance/src/index/scalar/fmindex.rs
beinan: Fixed. `merge_segments` now intersects the unioned fragment bitmap with `dataset.fragment_bitmap` to drop retired/compacted fragments before rebuilding, mirroring the btree merge behavior. When the intersection is empty (all fragments retired), it produces an empty index instead of crashing. Added `test_fmindex_merge_after_compaction_drops_retired_fragments` to cover this scenario.

rust/lance-index/src/scalar/fmindex.rs
```diff
@@ -1349,20 +1388,25 @@ impl ScalarIndex for FMIndexScalarIndex {
...
     while let Some(batch) = stream.next().await {
         let batch = batch?;
-        let row_ids: Option<&arrow_array::UInt64Array> = batch
-            .column_by_name("_rowid")
-            .or_else(|| batch.column_by_name("_rowaddr"))
-            .and_then(|c| c.as_any().downcast_ref());
-        let value_col = batch.column(0);
+        // Prefer _rowaddr (global row address) over _rowid to ensure stable,
+        // globally unique identifiers across segments.
+        let row_addrs: &arrow_array::UInt64Array = batch
```
jackye1995, line 1399: [P1] This new hard requirement for `_rowaddr` also needs to be reflected in FMIndex update/optimize paths. `FMIndexScalarIndex::update_criteria()` still returns `TrainingCriteria::new(TrainingOrdering::None)`, so `optimize_indices()` / append maintenance can build a training stream with only the value column and fail here. Also, `FMIndexScalarIndex::update()` currently writes a fresh index from `new_data` only and ignores the existing index plus `old_data_filter`; if we only add `.with_row_addr()`, the generic single-segment scalar update path can replace the old FMIndex with one containing only appended rows. Can we either make FMIndex update merge existing rows correctly, or force FMIndex maintenance to rebuild from the full target fragment bitmap instead of taking the single-segment `update()` shortcut?

beinan: Good catch. Fixed both issues:
1. `update_criteria()` now returns `requires_old_data` with `.with_row_addr()`, so the training stream includes all existing + new rows with global row addresses.
2. `update()` now applies the `old_data_filter` to exclude deleted/compacted rows before rebuilding the BWT, so the single-segment update path produces a complete index covering both old and new data instead of silently dropping existing rows.

rust/lance-index/src/scalar/fmindex.rs
```diff
@@ -1327,9 +1368,17 @@ impl ScalarIndex for FMIndexScalarIndex {
         &self,
         new_data: SendableRecordBatchStream,
         dest: &dyn IndexStore,
-        _: Option<OldIndexDataFilter>,
+        old_data_filter: Option<OldIndexDataFilter>,
     ) -> Result<CreatedIndex> {
-        let texts = collect_texts(new_data).await?;
+        let mut texts = collect_texts(new_data).await?;
+        if let Some(filter) = old_data_filter {
+            texts.retain(|(row_addr, _)| match &filter {
```
jackye1995: [P1] This filter drops the appended rows in the optimize/append path. Because `update_criteria()` now uses `requires_old_data`, `load_unindexed_training_data()` passes `fragments=None` and scans all current rows, not just old rows. But the `old_data_filter` here only describes the selected old segment coverage, so retaining only `to_keep` / `valid` removes the unindexed appended fragments from `texts` before rebuilding. I think FMIndex either needs to rebuild from the intended union fragment set without this old-only filter, or the update path needs a filter that keeps both selected old rows and new/unindexed rows while excluding deleted/retired old rows.

beinan: Good catch. The filter now excludes `to_remove` fragments instead of keeping only `to_keep`, so both old valid rows and new unindexed/appended rows are retained in the rebuilt index.

jackye1995 (approved):

PR #7128, opened 2026-06-05
feat: support merging zonemap index segments

Xuanwo:
ZoneMap already supports fragment-scoped segment builds and multi-segment query, but its segments could not be consolidated through the canonical distributed indexing merge path. This adds a ZoneMap merge primitive that rewrites source segments into one self-contained physical segment while preserving fragment coverage and rows-per-zone metadata, and updates the distributed indexing guide to include zone map merge support.


Xuanwo:
@claude review once

rust/lance/src/index/scalar/zonemap.rs
```diff
@@ -0,0 +1,86 @@
...
+        &new_store,
+        &fragment_bitmap,
+    )
+    .await?;
+
+    Ok(IndexMetadata {
+        uuid: new_uuid,
+        fields: vec![field_id],
+        dataset_version: dataset.manifest.version,
+        fragment_bitmap: Some(fragment_bitmap),
```
BubbleCal, line 78: Could we add a merge-then-commit regression test here? Since the merged segment keeps only live fragment coverage, committing it under the same index name can compare against the old segment's original coverage and report retired fragments as orphaned.

BubbleCal (commented):
One concern about the merge-then-commit path.

rust/lance/src/index.rs
```diff
@@ -1214,7 +1224,8 @@ impl DatasetIndexExt for Dataset {
                     .is_none_or(|(details, expected)| details.type_url == expected)
             })
             .map(|idx| -> Result<Option<IndexMetadata>> {
-                let Some(existing_fragments) = idx.fragment_bitmap.as_ref() else {
+                let Some(existing_fragments) = idx.effective_fragment_bitmap(&dataset_fragments)
```
BubbleCal, line 1224: Potential issue: fully retired existing segments can reach this disjoint branch and be kept instead of removed. That leaves old zonemap segments referenced in the manifest after a merge, so the merge may not actually replace multiple physical segments with one.

BubbleCal (approved):

PR #7129, opened 2026-06-05
perf!: avoid listing index files after writes

Xuanwo:
Avoids listing index directories after new index writes by propagating `IndexFile` metadata from writer and builder APIs into manifest metadata. This covers scalar index writers, V3 vector builders, vector append/remap paths, and distributed vector merge while preserving legacy and extension fallbacks where the API cannot return files yet.

This reduces extra object-store list/stat IO on fresh index creation and maintenance paths, with IO-count regression coverage for scalar and V3 vector index creation.


Xuanwo:
@claude review once

BubbleCal (approved):
Nice work!

PR #7131, opened 2026-06-05
fix(python): avoid pandas requirement for add_columns UDFs

prrao87:
## Summary

This fixes `add_columns` UDF schema inference so Arrow-only callables do not require pandas to be installed.

`pd` is a lazy optional-dependency proxy. In a pandas-free environment, evaluating `pd.DataFrame` raises `ModuleNotFoundError` even when the user UDF returns a `pyarrow.RecordBatch`. The fix uses the existing pandas availability guard before the DataFrame `isinstance` checks, matching the conversion path used when executing batch UDFs.

The regression test simulates Lance's no-pandas lazy proxy state and verifies that an Arrow-only UDF still works with `batch_size`, covering both plain callables and `batch_udf()` schema inference.

## Validation

- `uv run pytest python/tests/test_schema_evolution.py::test_add_columns_arrow_udf_without_pandas_dependency`
- `uv run make lint` reached and passed Ruff; local Pyright then failed on existing optional `tensorflow` and `torch` imports in dependency typing on macOS.

🤖PR generated by Codex, but validated by @prrao87 and run with a local, fresh Python SDK build.


wjones127 (approved):
Thank you, this looks like a clean fix. 🙌

PR #7132, opened 2026-06-05
fix(mem-wal): surface fenced flush to durability waiters instead of hanging

hamersaw:
## Problem

A `durable_write` put waits on a `BatchDurableWatcher`, which only resolves when the durability watermark reaches its target batch position. The watermark is advanced **exclusively on a successful** WAL flush (`flush_from_batch_store`).

When a flush fails terminally — a successor has claimed a higher epoch and fenced this writer — the append returns `Err` before the watermark advances, the background dispatcher logs and swallows the error (by design, so a transient flush failure doesn't deadlock puts), and the watch channel is never closed. The watcher's `wait()` therefore **loops forever**: the watermark can never reach the target and the channel never closes. The put hangs until the client times out.

This became reachable with the fence-on-claim sentinel (#7110): a fenced predecessor's next flush now collides with the sentinel and fails, where it previously false-acked into an empty slot and advanced the watermark. Surfaced downstream as a WAL server's `merge_insert` HTTP handler hanging (client `operation timed out`) after a rolling-restart pod replacement, instead of returning a terminal verdict.

## Fix

Share a terminal-error slot between `WalFlusher` and every `BatchDurableWatcher` it hands out. On a fence error, `flush` records the error and wakes all waiters (via `send_modify`, without advancing the watermark); `wait()` re-checks the slot and returns the fence error. A fence is permanent, so failing fast is correct — and the error text is preserved so callers can map it to a terminal verdict (e.g. HTTP 410).

Only fences poison the slot; transient flush failures keep the existing log-and-retry-on-next-flush behavior.

## Test

`test_durable_watcher_aborts_on_fence_instead_of_hanging`: a predecessor's durable flush collides with a successor's fence sentinel; the watcher resolves with the fence error under a timeout rather than blocking forever.

### DIFF
--- a/rust/lance/src/dataset/mem_wal/wal.rs
+++ b/rust/lance/src/dataset/mem_wal/wal.rs
@@ -43,6 +43,13 @@ pub const WRITER_EPOCH_KEY: &str = "writer_epoch";
 /// replay skips sentinels via their empty batch list).
 pub const FENCE_SENTINEL_KEY: &str = "fence_sentinel";
 
+/// True if `error` is the terminal fence emitted by `ManifestStore::check_fenced`
+/// (a successor claimed a higher epoch). Matches the message it formats, since
+/// fences surface as a plain `Error::io` rather than a typed variant.
+fn is_fence_error(error: &Error) -> bool {
+    error.to_string().contains("Writer fenced")
+}
+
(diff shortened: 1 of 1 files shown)


jackye1995 (approved):
Looks good to me!

Xuanwo (approved):
Thank you!

PR #7138, opened 2026-06-06
feat(mem_wal): support prefiltered LSM vector and FTS search

touch-of-grey:
Adds true prefilter support for LSM vector and full-text search across base, flushed, and in-memory memtable sources.

This preserves newest-PK semantics before top-k for rewritten active rows, keeps HNSW/FTS limit pushdown for append-only PK memtables, and threads vector filters through Python and Java bindings with input validation.


jackye1995:
Thanks for the fix! Since we are already doing the refactoring here, I think let's also make sure the MemTable scanner is consistent with the other scanners.

hamersaw (approved):
This all looks great, my few observations:
- removing the `NewestPkFilterExec` is probably a good idea - this way we cannot underfill on the active memtable scan
- the sticky `pk_has_overrides` on the active memtable forces dedup logic. i'm interested to see the performance implications of this, but not concerned enough to block.

PR #7139, opened 2026-06-06
feat(index): accelerate regex and infix LIKE with the ngram index

### LINKED ISSUE #7130: Accelerate regular expressions with ngram index, if present
Some databases such as Postgres allow for some regular expressions to be accelerated with an ngram index.  We should be able to implement similar capability in lance.

wombatu-kun:
## What

Extends the ngram index (today only `contains(col, 'substr')`) to also accelerate `regexp_like(col, pat)` / `regexp_match(col, pat)` and infix `LIKE` (`col LIKE '%foo%bar%'`), which until now fell through to a full table scan. Closes #7130.

## How

Following Postgres `pg_trgm` and Russ Cox's [trigram-index approach](https://swtch.com/~rsc/regexp/regexp4.html), a pattern is compiled into a boolean trigram condition (`TrigramQuery`, an AND/OR tree) that is a *necessary* condition for any match. This maps onto the inverted index's set algebra: AND is posting-list intersection, OR is union. The index returns a candidate superset and the scan rechecks the true predicate, exactly as `contains` does. The derivation walks the `regex-syntax` HIR bottom-up, tracking per-node `(emptyable, exact, prefix, suffix)` sets and folding boundary trigrams across concatenation, with bounds that fold-then-discard so precision loss never drops a necessary trigram.

### Soundness

The derived condition never requires a trigram a matching string could lack, so no real match is dropped. Requirements come from the index's own tokenizer (so sub-trigram runs contribute nothing); character classes and case-insensitive folds are treated as a single unknown character (the index's normalization disagrees with Unicode case folding - `(?i)c` also matches U+2102); and patterns with no derivable trigram (`a.b`, `.*`) are left to a full scan.

### Why infix LIKE

A plain-literal `regexp_like(col, 'foo')` is rewritten to `col LIKE '%foo%'` by the optimizer before it reaches the index, so without infix-LIKE the most common "regex" query would not accelerate. The LIKE is translated to a loose regex for candidate generation only; the original LIKE stays as the recheck filter, so the candidate regex need only be a sound superset.

## Benchmark

`cargo bench -p lance --bench regex_ngram`, 200k rows, before (main) vs after:

| Query | Before | After | Change |
|---|---|---|---|
| `regexp_match(doc, 'zqxwvu.*needlexyz')` | 45.8 ms | 8.2 ms | -82% |
| `regexp_match(doc, '(zqxwvu\|qwerasdf\|needlexyz)')` | 51.7 ms | 11.6 ms | -77% |
| `regexp_match(doc, 'zqxwvu')` (rewritten to LIKE) | 36.6 ms | 8.0 ms | -78% |
| `regexp_match(doc, 'a.b')` (non-accelerable) | 77.8 ms | 76.2 ms | within noise |

## Testing

Unit tests for the regex-to-trigram derivation and the regex-flags folding; index-level search tests (AND across `.*`, alternation union, NULL exclusion, absent trigram); a multi-fragment end-to-end scan test asserting correct results and index engagement for `regexp_like`, `regexp_match`, and `LIKE`, plus a case-insensitive query that must fall back to a full recheck; and a regression test ensuring non-accelerable patterns still return all correct matches via full recheck. No binding changes are needed - Python/Java pass these filter strings through the existing scan API, so acceleration applies transparently.

### DIFF
--- a/rust/lance-index/src/scalar/expression.rs
+++ b/rust/lance-index/src/scalar/expression.rs
@@ -781,20 +781,28 @@ impl ScalarQueryParser for LabelListQueryParser {
     }
 }
 
-/// A parser for indices that handle string contains queries
+/// A parser for indices that handle string `contains` queries, and -- when
+/// `supports_regex` is set -- `regexp_like` / `regexp_match` queries.
 #[derive(Debug, Clone)]
 pub struct TextQueryParser {
     index_name: String,
     index_type: String,
     needs_recheck: bool,
+    supports_regex: bool,
 }
 
 impl TextQueryParser {
-    pub fn new(index_name: String, index_type: String, needs_recheck: bool) -> Self {
+    pub fn new(
+        index_name: String,
+        index_type: String,
+        needs_recheck: bool,
+        supports_regex: bool,
+    ) -> Self {
         Self {
             index_name,
             index_type,
             needs_recheck,
+            supports_regex,
         }
     }
 }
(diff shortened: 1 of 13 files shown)


rust/lance-index/src/scalar/expression.rs
```diff
@@ -837,31 +845,156 @@ impl ScalarQueryParser for TextQueryParser {
...
-                    Some(IndexedExpression::index_query_with_recheck(
-                        column.to_string(),
-                        self.index_name.clone(),
-                        self.index_type.clone(),
-                        Arc::new(query),
-                        self.needs_recheck,
-                    ))
-                } else {
+        // A non-string pattern cannot be handled.
+        let (ScalarValue::Utf8(Some(pattern)) | ScalarValue::LargeUtf8(Some(pattern))) =
```
wjones127, line 855: issue(non-blocking): this is pre-existing, but we should eventually also support `ScalarValue::Utf8View`.

rust/lance-index/src/scalar/ngram/ngram_regex.rs
```diff
@@ -0,0 +1,673 @@
...
+    }
+
+    #[test]
+    fn test_wildcard_splits_into_and() {
+        // `.*` breaks the literal run; both sides are required.
+        assert_eq!(
+            q("foo.*bar"),
+            TrigramQuery::and(vec![tri("foo"), tri("bar")])
+        );
+    }
```
wjones127, line 544: praise: these are some of the clearest test cases I've read in a while. Nice work!

wjones127 (approved):
This looks great! I'm excited to have this released :)

PR #7142, opened 2026-06-07
ci: pin license-header-checker to fix flaky install step

wombatu-kun:
The `license-header-check` job installs the checker via the upstream `install.sh` with no pinned tag, so it resolves "latest" by fetching the GitHub releases page and parsing `tag_name`. That fetch intermittently returns empty and fails the job with `unable to find ''`, unrelated to any PR's contents.

Previous failed run (the install step, not a header violation): https://github.com/lance-format/lance/actions/runs/27094024188/job/79962732768

Reading the upstream `install.sh` shows that passing a pinned tag would not help - `github_release()` still fetches the releases page to resolve the tag. So this pins to `v1.5.0` and downloads the release asset directly from the CDN, removing the dynamic tag/JSON lookup entirely. The job runs only on `ubuntu-latest`, so the `linux_amd64` asset is used.

### DIFF
--- a/.github/workflows/license-header-check.yml
+++ b/.github/workflows/license-header-check.yml
@@ -31,9 +31,16 @@ jobs:
         run: |
           set -euo pipefail
           mkdir -p ./bin
-          curl -sfSL "https://github.com/lluissm/license-header-checker/releases/download/v${LHC_VERSION}/license-header-checker_${LHC_VERSION}_linux_amd64.tar.gz" -o /tmp/lhc.tar.gz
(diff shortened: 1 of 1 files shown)


Xuanwo (commented):
The pinned install path now executes a downloaded third-party release binary without verifying its digest. CI loses the artifact-integrity check that protected the previous installer path.

wombatu-kun:
@Xuanwo Done `7df732c57` - the install step now downloads the release `checksums.txt` and runs `sha256sum -c` to verify the tarball before extracting, restoring the integrity check from the upstream `install.sh`.

Xuanwo (approved):

PR #7143, opened 2026-06-07
fix(index): cap IVF train prefetch memory for nullable columns

### LINKED ISSUE #7126: Memory consumption during IVF train prefetch
When a vector column is `nullable`, IVF training reads up to `2×` the requested sample size per prefetch round, while the non-nullable path reads exactly the requested size. The amplification is explicit in `rust/lance/src/index/vector/utils.rs::sample_training_data_scan_from_fragments`:
```rust
let target = sample_size_hint.saturating_mul(2)
```
and is the dominant peak-memory term for IVF training on nullable + fragment-limited inputs.

The over-fetch is also propagated into the consumer's output buffer.
`sample_nullable_fsl` only checks `num_non_null < sample_size_hint` before reading the next batch and then appends the whole filtered batch, so on low-null data the output `MutableBuffer` itself grows to roughly `2 × sample_size_hint × byte_width` before the post-loop truncate.

A second, smaller amplification lives in `sample_nullable_fallback` where every prefetched batch is retained in a `Vec<RecordBatch>` and then materialised into one combined batch via `concat_batches`, doubling peak memory at the moment of concat.

geserdugarov:
## What

Fixes IVF training sampling memory growth for nullable vector columns by bounding fragment-limited prefetch to the consumer's outstanding demand and clamping sampled output to the requested training sample size.

Closes https://github.com/lance-format/lance/issues/7126

## Changes

- Tracks the consumer's outstanding demand in a shared `still_needed` counter so each fragment-limited
prefetch round fetches at most `still_needed.min(remaining)` rows instead of `2 * sample_size_hint`;
consumers zero the counter when their sample is full so any further poll of the producer terminates instead
of reading another round.
- Replaces the `HashSet<u64>` of seen offsets with a `RoaringTreemap`, keeping the visited-offset state near
`num_rows / 8` bytes even when sparse or all-null columns force the stream to visit most selected rows.
- Samples unseen offsets via rejection sampling against the bitmap, switching to a dense shuffle of the
unseen set only when few offsets remain unseen (`remaining <= 4 * target`), so the dense allocation stays
proportional to the round size.
- Clamps `sample_nullable_fsl`, `sample_nullable_fallback`, and `accumulate_fsl_values` (new `max_rows`
parameter) so oversized batches cannot grow output past `sample_size_hint`.

## Notes

- No public API changes.
- The fallback path also slices accepted batches to the outstanding demand, keeping the retained batches and the final `concat_batches` bounded by the sample size.

## Tests

- `cargo test -p lance --lib index::vector::utils::tests`
- New: `test_maybe_sample_training_data_fsl_nullable_fragment_limited` (partial-null fills the sample, all-null terminates cleanly), `test_sample_fragment_scan_round_caps_at_still_needed` (producer round never over-reads the demand), `test_accumulate_fsl_values_respects_max_rows`.

### DIFF
--- a/rust/lance/src/index/vector/utils.rs
+++ b/rust/lance/src/index/vector/utils.rs
@@ -1,9 +1,9 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-FileCopyrightText: Copyright The Lance Authors
 
-use std::collections::HashSet;
 use std::pin::Pin;
 use std::sync::Arc;
+use std::sync::atomic::{AtomicUsize, Ordering};
 
 use arrow::array::ArrayData;
 use arrow::datatypes::DataType;
@@ -18,6 +18,7 @@ use log::{info, warn};
 use rand::rngs::SmallRng;
 use rand::seq::{IteratorRandom, SliceRandom};
 use rand::{Rng, SeedableRng};
+use roaring::RoaringTreemap;
 use tokio::sync::Mutex;
 
 use crate::dataset::{Dataset, ProjectionRequest, TakeBuilder, row_offsets_to_row_addresses};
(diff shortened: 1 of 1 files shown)


rust/lance/src/index/vector/utils.rs
```diff
@@ -543,6 +568,65 @@ fn sample_training_data_scan(
...
+    let remaining = num_rows.saturating_sub(seen_offsets.len());
+    let target = target.min(remaining);
+    if target == 0 {
+        return Vec::new();
+    }
+    if remaining > target.saturating_mul(4) {
+        let mut sampled = Vec::with_capacity(target);
+        while sampled.len() < target {
+            let index = rng.random_range(0..num_rows as u64);
+            if seen_offsets.insert(index) {
```
Xuanwo: The nullable fragment sampler still records every sampled offset in the persistent `seen_offsets` set until it has exhausted the selected rows. For all-null or extremely sparse nullable inputs, index training can still grow to O(selected rows) memory, CPU, and I/O before returning no training data.

geserdugarov: Addressed for the memory issue: `seen_offsets` is now a `RoaringTreemap` instead of `HashSet<u64>`, and each producer round is capped to the consumer's outstanding demand. The all-null path now terminates and is covered by the `all_nulls` case of `test_maybe_sample_training_data_fsl_nullable_fragment_limited`. The remaining O(selected rows) CPU/I/O on all-null input is inherent without fragment-level null statistics and could be a separate optimization.

Xuanwo (approved):
Thank you!

PR #7147, opened 2026-06-08
feat(rust): add cleanup explain API

yanghua:
## Summary

  - Add `Dataset::cleanup(policy)` with two terminal actions:
    - `explain()` returns a read-only `CleanupExplanation`
    - `execute()` re-evaluates current dataset/ref state before deleting files
  - Add explanation details including read version, aggregate removal stats, candidate files, truncation metadata,
  referenced branch details, and warnings.
  - Keep existing Rust execution APIs (`cleanup_old_versions`, `cleanup_with_policy`) compatible.

  ## Tests

  - `cargo fmt --all`
  - `cargo test -p lance dataset::cleanup::tests -- --nocapture`

  `cargo clippy -p lance --all-targets -- -D warnings` was attempted, but currently fails on unrelated pre-existing
  lint errors outside this change.

### DIFF
--- a/rust/lance/src/dataset.rs
+++ b/rust/lance/src/dataset.rs
@@ -24,8 +24,7 @@ use lance_core::datatypes::{OnMissing, OnTypeMismatch, Projectable, Projection};
 use lance_core::traits::DatasetTakeRows;
 use lance_core::utils::address::RowAddress;
 use lance_core::utils::tracing::{
-    DATASET_CLEANING_EVENT, DATASET_DELETING_EVENT, DATASET_DROPPING_COLUMN_EVENT,
-    TRACE_DATASET_EVENTS,
+    DATASET_DELETING_EVENT, DATASET_DROPPING_COLUMN_EVENT, TRACE_DATASET_EVENTS,
 };
 use lance_datafusion::projection::ProjectionPlan;
 use lance_file::datatypes::populate_schema_dictionary;
@@ -104,7 +103,7 @@ use self::scanner::{DatasetRecordBatchStream, Scanner};
 use self::transaction::{Operation, Transaction, TransactionBuilder, UpdateMapEntry};
 use self::write::{cleanup_data_fragments, write_fragments_internal};
 use crate::dataset::branch_location::BranchLocation;
-use crate::dataset::cleanup::{CleanupPolicy, CleanupPolicyBuilder};
+use crate::dataset::cleanup::{CleanupOperation, CleanupPolicy, CleanupPolicyBuilder};
 use crate::dataset::refs::{BranchContents, BranchIdentifier, Branches, Tags};
 use crate::dataset::sql::SqlQueryBuilder;
 use crate::datatypes::Schema;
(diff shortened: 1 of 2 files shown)


yanghua:
@claude review

Xuanwo (commented):
I think this should follow SQL `EXPLAIN` semantics: the cleanup plan should be a dry-run/audit report, not a materialized deletion plan.

Execution should re-evaluate cleanup from the current dataset/ref state instead of trusting an old file list. For example, a tag or branch can be added after planning without advancing the manifest version, so the current `read_version` check can still pass while the old plan deletes files that are now protected.

yanghua:
> I think this should follow SQL `EXPLAIN` semantics: the cleanup plan should be a dry-run/audit report, not a materialized deletion plan.
> 
> Execution should re-evaluate cleanup from the current dataset/ref state instead of trusting an old file list. For example, a tag or branch can be added after planning without advancing the manifest version, so the current `read_version` check can still pass while the old plan deletes files that are now protected.

Sounds reasonable. My original idea is also for `dry-run` purpose. I think I misunderstood your meaning of the `plan` when we talked offline before. Will refactor later.

Xuanwo (commented):
Thanks for refactoring this away from the previous executable `CleanupPlan` design. I think the current behavior is much closer to the semantics we want.

However, I’m still not sure `CleanupMode::{Execute, DryRun}` is the right public abstraction. This makes dry-run look like a mode of the destructive cleanup API, while what we want is closer to SQL `EXPLAIN`: a read-only explanation of what cleanup would do, not an alternate execution mode.

For Rust, I’d prefer a builder-style API that models cleanup as an operation with two terminal actions, for example:

```rust
let cleanup = dataset.cleanup(policy);

let explanation = cleanup.explain().await?;
let stats = cleanup.execute().await?;
```

`explain()` should return a `CleanupExplanation` with fields like `read_version`, `RemovalStats`, candidate files, referenced-branch details, warnings, and truncation info. It should not be accepted later as an executable deletion plan; `execute()` should still re-evaluate current dataset/ref state.

For Python, I think we should keep the existing Python style instead of exposing Rust builders/modes directly:

```python
explanation = ds.explain_cleanup_old_versions(
    older_than=timedelta(days=7),
    retain_versions=3,
    include_files=True,
    max_files=1000,
)

stats = ds.cleanup_old_versions(
    older_than=timedelta(days=7),
    retain_versions=3,
)
```

This also gives us a cleaner path to extend the same concept to other maintenance operations later, e.g. Rust builder-style `dataset.compaction(options).explain()` / `.execute()`, while Python can expose a Pythonic `ds.optimize.explain_compaction(...)` alongside `ds.optimize.compact_files(...)`. That feels more extensible than adding operation-specific mode enums such as `CleanupMode`.

yanghua:
> Thanks for refactoring this away from the previous executable `CleanupPlan` design. I think the current behavior is much closer to the semantics we want.
> 
> However, I’m still not sure `CleanupMode::{Execute, DryRun}` is the right public abstraction. This makes dry-run look like a mode of the destructive cleanup API, while what we want is closer to SQL `EXPLAIN`: a read-only explanation of what cleanup would do, not an alternate execution mode.
> 
> For Rust, I’d prefer a builder-style API that models cleanup as an operation with two terminal actions, for example:
> 
> ```rust
> let cleanup = dataset.cleanup(policy);
> 
> let explanation = cleanup.explain().await?;
> let stats = cleanup.execute().await?;
> ```
> 
> `explain()` should return a `CleanupExplanation` with fields like `read_version`, `RemovalStats`, candidate files, referenced-branch details, warnings, and truncation info. It should not be accepted later as an executable deletion plan; `execute()` should still re-evaluate current dataset/ref state.
> 
> For Python, I think we should keep the existing Python style instead of exposing Rust builders/modes directly:
> 
> ```python
> explanation = ds.explain_cleanup_old_versions(
>     older_than=timedelta(days=7),
>     retain_versions=3,
>     include_files=True,
>     max_files=1000,
> )
> 
> stats = ds.cleanup_old_versions(
>     older_than=timedelta(days=7),
>     retain_versions=3,
> )
> ```
> 
> This also gives us a cleaner path to extend the same concept to other maintenance operations later, e.g. Rust builder-style `dataset.compaction(options).explain()` / `.execute()`, while Python can expose a Pythonic `ds.optimize.explain_compaction(...)` alongside `ds.optimize.compact_files(...)`. That feels more extensible than adding operation-specific mode enums such as `CleanupMode`.

Sounds reasonable, good suggestion. Will refactor it.

Xuanwo (approved):
Nice work, thank you!

PR #7152, opened 2026-06-08
fix(dataset): resolve Blob v2 external URIs and robustly clean failed writes in add_columns

### LINKED ISSUE #7075: add_columns should support Blob v2 external URIs and clean up failed writes
## Problem

`add_columns` had two gaps compared with the main write path when working with Blob v2 datasets:

1. It could not resolve Blob v2 external URIs during the update/write path.
2. It could leave orphaned data files and Blob v2 sidecars behind when the operation failed after partial writes.

This made `add_columns` inconsistent with the expected Blob v2 behavior and could leave storage artifacts behind on failure.

### Blob v2 external URI resolution

`add_columns` now opens its update writer with the same Blob v2 external base resolution needed by the normal write path, so Blob v2 reference values can resolve dataset-registered external URIs correctly.

### Failed write cleanup

`add_columns` now cleans up files created by the current failed attempt, including:
- unfinished data files from the current writer
- completed but uncommitted fragment outputs
- Blob v2 sidecar directories created for those files

The cleanup logic also preserves the intended safety boundaries:
- do not delete external-base files
- do not delete fragments already recovered from / owned by checkpoint state

## Scope

This issue covers the `add_columns` path and its Blob v2 / cleanup behavior.

## Follow-up

`alter_columns` is **not** included in this change set.

`alter_columns` now shares some of the same lower-level machinery, but its commit-failure cleanup path should be handled in a **separate follow-up PR** to keep scope focused and reviewable.

## Notes

This work is intended to keep `add_columns` behavior aligned with Lance Blob v2 design expectations:
- consistent URI resolution behavior across write paths
- no orphaned internal files on failed operations
- no accidental deletion of external user-managed data

yyzhao2025:
## Description

This PR addresses two critical functional gaps in the `add_columns` pipeline when working with Blob v2 datasets. It ensures that the schema evolution path achieves feature parity with the main write path regarding external URI resolution, and it introduces a robust, leak-free cleanup mechanism for failed mutations.

### 1. Blob v2 External URI Resolution

**Context:** 
In Blob v2, external blobs (where data resides outside the dataset's native storage) rely on an `ExternalBaseResolver` to construct the correct absolute URIs from relative paths stored in the dataset manifest.
Previously, `add_columns` used a bare `open_writer` via the `Updater`, which lacked the context to initialize this resolver. As a result, appending or mutating Blob v2 columns with `external` kind would fail to resolve URIs.

**Fix:**
- Introduced `open_update_writer` in `write.rs`. When the `Updater` creates a new writer, it now checks if `storage_version >= V2_2` and if the schema contains `blob_v2` fields. 
- If so, it constructs the `ExternalBaseResolver` using the dataset's registered base paths and passes it to the writer alongside the `source_store_registry`. 
- This seamlessly enables operations like `BatchUDF` or `RecordBatchReader` to write `external` Blob v2 data during schema evolution, achieving strict parity with `write_fragments_internal`.

### 2. Comprehensive Cleanup for Failed Writes

**Context:**
The `add_columns` operation is complex and can fail mid-flight due to various reasons: UDF execution panics, stream ingestion errors, checkpoint lookup/insert failures, or schema merge conflicts. Previously, these failures would eagerly propagate via the `?` operator, leaving behind orphaned `.lance` data files and their corresponding Blob v2 sidecar directories.

**Fix & Architectural Guarantees:**
This PR overhauls the error handling and cleanup lifecycle in `add_columns_impl` and `add_columns_from_stream`.

- **Multi-stage Cleanup:** 
  We replaced concurrent `try_collect` streams with sequential processing augmented with scoped error handling. When a failure occurs, the pipeline now executes a two-stage cleanup:
  1. `updater.cleanup_unfinished_writer()`: Cleans up the currently active, unfinished data file that hasn't been finalized into a `Fragment` yet.
  2. `cleanup_new_column_data_files()`: Cleans up any fully written but uncommitted fragments generated in the current run.
  
- **Strict Safety Constraint 1: Preservation of External Data**
  The underlying `cleanup_data_fragments` logic now strictly checks `base_id.is_none()`. This guarantees that cleanup operations **never** attempt to delete files that belong to an external base, preventing catastrophic deletion of user-managed source data.
  
- **Strict Safety Constraint 2: Checkpoint Ownership**
  For long-running UDFs, fragments are incrementally saved to a `UDFCheckpointStore`. Once a fragment is successfully inserted into the checkpoint, it is explicitly `pop()`-ed from the local `fragments_to_cleanup` list. This ensures that if a subsequent step fails, we do not physically delete data files that the checkpoint relies on for resumption.

## Related Issues
Closes #7075

## Scope & Follow-up
This PR tightly scopes the fixes to the `add_columns` pipeline. 
**Note on `alter_columns`:** While `alter_columns` shares the newly improved `add_columns_impl` machinery (and thus benefits from the unfinished writer cleanup), its final `apply_commit` stage currently drops the `fragments_to_cleanup` vector. A commit failure there could still leave orphaned files. To keep this PR focused and reviewable, the commit-failure cleanup for `alter_columns` will be addressed in a separate, dedicated follow-up PR.

## Type of change
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Testing Performed
- **Rust Integration Tests:** Added parameterized tests covering `inline`, `packed`, `dedicated`, and `external` cleanup scenarios. Validated stream errors, UDF panics, and checkpoint lookup/insert anomalies to ensure zero orphan files and zero deleted external files.
- **Python Extension Tests:** Verified `add_columns` with `RecordBatchReader` and `BatchUDF` across all Blob v2 kinds. Asserted bit-for-bit data integrity upon reading back the mutated dataset.
- Passed all formatting (`cargo fmt`), linting (`cargo clippy`), and existing test suites.

### DIFF
--- a/python/python/tests/test_blob.py
+++ b/python/python/tests/test_blob.py
@@ -45,6 +45,34 @@ def _external_blob_table(blob_path, payload=b"hello"):
     return pa.table({"blob": lance.blob_array([blob_path.as_uri()])})
 
 
+def _add_columns_blob_v2_values(tmp_path):
+    external_base = tmp_path / "external_base"
+    external_blob = external_base / "external_blob.bin"
+    external_blob.parent.mkdir(parents=True, exist_ok=True)
+    external_blob.write_bytes(b"external")
+
+    payloads = [
+        b"inline",
+        b"p" * (64 * 1024 + 1024),
+        b"d" * (4 * 1024 * 1024 + 1024),
+        b"external",
+    ]
+    values = [payloads[0], payloads[1], payloads[2], external_blob.as_uri()]
+    initial_bases = [DatasetBasePath(external_base.as_uri(), name="external", id=1)]
+    return values, payloads, initial_bases
+
+
+def _assert_blob_v2_add_columns_result(dataset, column, payloads):
+    desc = dataset.to_table(columns=[column]).column(column).chunk(0)
+
+    assert desc.field("kind").to_pylist() == [0, 1, 2, 3]
+    assert desc.field("blob_id").to_pylist()[3] == 1
+    assert desc.field("blob_uri").to_pylist()[3] == "external_blob.bin"
+
+    blobs = dataset.take_blobs(column, indices=range(len(payloads)))
+    assert [blob.readall() for blob in blobs] == payloads
+
+
 def _out_of_order_blob_selection(dataset_with_blobs, selection_kind):
     addresses = _blob_row_addresses(dataset_with_blobs)
     expected = [(addresses[4], b"quux"), (addresses[0], b"foo")]
@@ -606,6 +634,54 @@ def test_blob_extension_write_external_ingest_rejects_reference_only_options(tmp
             external_blob_mode="ingest",
             allow_external_blob_outside_bases=True,
         )
+
+
+def test_blob_extension_add_columns_record_batch_reader_all_kinds(tmp_path):
+    values, payloads, initial_bases = _add_columns_blob_v2_values(tmp_path)
+    ds = lance.write_dataset(
+        pa.table({"id": range(4)}),
+        tmp_path / "test_add_columns_reader_blob_v2",
+        data_storage_version="2.2",
+        initial_bases=initial_bases,
+    )
+
+    ds.add_columns(pa.table({"blob": lance.blob_array(values)}).to_reader())
+
+    _assert_blob_v2_add_columns_result(ds, "blob", payloads)
+
+
+def test_blob_extension_add_columns_batch_udf_all_kinds(tmp_path):
+    values, payloads, initial_bases = _add_columns_blob_v2_values(tmp_path)
+    ds = lance.write_dataset(
+        pa.table({"id": range(4)}),
+        tmp_path / "test_add_columns_udf_blob_v2",
+        data_storage_version="2.2",
+        initial_bases=initial_bases,
+    )
+
+    @lance.batch_udf(output_schema=pa.schema([lance.blob_field("blob")]))
+    def make_blob_column(batch):
+        return pa.record_batch(
+            [lance.blob_array([values[row.as_py()] for row in batch["id"]])],
+            ["blob"],
+        )
+
+    ds.add_columns(make_blob_column, read_columns=["id"])
+
+    _assert_blob_v2_add_columns_result(ds, "blob", payloads)
+
+
+def test_blob_extension_add_columns_all_nulls_blob_v2(tmp_path):
+    ds = lance.write_dataset(
+        pa.table({"id": range(4)}),
+        tmp_path / "test_add_columns_all_nulls_blob_v2",
+        data_storage_version="2.2",
+    )
+
+    ds.add_columns(lance.blob_field("blob"))
+
+    assert ds.to_table(columns=["blob"]).column("blob").to_pylist() == [None] * 4
+    assert ds.take_blobs("blob", indices=range(4)) == []
 
 
 def test_blob_extension_write_fragments_external_denied_by_default(tmp_path):
--- a/python/python/tests/test_blob.py
+++ b/python/python/tests/test_blob.py
@@ -73,6 +73,28 @@ def _assert_blob_v2_add_columns_result(dataset, column, payloads):
     assert [blob.readall() for blob in blobs] == payloads
 
 
+def _dataset_file_set(dataset_path):
+    return {
+        path.relative_to(dataset_path)
+        for path in dataset_path.rglob("*")
+        if path.is_file()
+    }
+
+
+def _write_two_fragment_blob_v2_seed_dataset(tmp_path, name):
+    values, payloads, initial_bases = _add_columns_blob_v2_values(tmp_path)
+    dataset_path = tmp_path / name
+    ds = lance.write_dataset(
+        pa.table({"id": range(8)}),
+        dataset_path,
+        data_storage_version="2.2",
+        initial_bases=initial_bases,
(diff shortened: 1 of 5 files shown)


yanghua:
@claude review

rust/lance/src/dataset/schema_evolution.rs
```diff
@@ -404,13 +418,76 @@ pub(super) async fn add_columns(
     )
     .await?;
 
-    let operation = Operation::Merge { fragments, schema };
+    let operation = Operation::Merge {
+        fragments: fragments.clone(),
```
yanghua: Do we need to use it?

rust/lance/src/dataset/fragment.rs
```diff
@@ -1792,7 +1792,7 @@ impl FileFragment {
         read_columns: Option<Vec<String>>,
         batch_size: Option<u32>,
     ) -> Result<(Fragment, Schema)> {
-        let (fragments, schema) = schema_evolution::add_columns_to_fragments(
+        let (fragments, schema, _) = schema_evolution::add_columns_to_fragments(
```
yanghua, line 1795: Can we keep `fragments_to_cleanup`? So that the outer can do some cleanup work.

python/python/tests/test_blob.py
```diff
@@ -608,6 +636,54 @@ def test_blob_extension_write_external_ingest_rejects_reference_only_options(tmp
         )
 
 
+def test_blob_extension_add_columns_record_batch_reader_all_kinds(tmp_path):
```
yanghua, line 661: Can we at least add a test case about fail-then-cleanup?

yanghua (commented):
Left some comments.

rust/lance/src/dataset/schema_evolution.rs
yyzhao2025: Good catch, we don't need the clone here. `fragments` is not reused after constructing `Operation::Merge`, so I removed it and moved the value directly.

python/python/tests/test_blob.py
yyzhao2025: Makes sense. I'll add a fail-then-cleanup regression test so we cover the partial-write path, not just the success case.

rust/lance/src/dataset/fragment.rs
yyzhao2025: Good point. I think this is a broader follow-up than the current blob-v2 add_columns fix.

This PR is scoped to the dataset-level add_columns cleanup path. Keeping
`fragments_to_cleanup` in `Fragment::add_columns` would effectively extend the
fragment-level API / ownership boundary so the outer caller can decide how to
clean up on later failures.

I’d prefer to keep this PR focused and handle that as a separate follow-up PR,
unless you feel it is required to land this fix safely.

yanghua: Can you file a ticket to track this issue?

rust/lance/src/dataset/schema_evolution.rs
```diff
@@ -406,11 +420,71 @@ pub(super) async fn add_columns(
...
+        &first_fragment.dataset().base,
+        &fragments_to_cleanup,
+    )
+    .await;
+}
+
+struct AddColumnFragments {
+    fragments: Vec<Fragment>,
+    fragments_to_cleanup: Vec<Fragment>,
 }
```
yanghua, line 492: It would be better to add some doc for this two fields.

rust/lance/src/dataset/updater.rs
```diff
@@ -221,6 +215,37 @@ impl Updater {
         Ok(self.fragment.metadata().clone())
     }
 
+    /// Clean up any data file and blob sidecars created by the current unfinished writer.
+    pub(super) async fn cleanup_unfinished_writer(&mut self) {
```
yanghua, line 219: Shall we clear the `self.writer` after finishing the cleanup?

yanghua (commented):
Left some new comments.

yanghua (approved):
+1 cc @Xuanwo

rust/lance/src/dataset/fragment.rs
yyzhao2025: Good point. I filed a follow-up issue to track this: #7231 。
The current PR stays scoped to dataset-level `add_columns` cleanup, while the follow-up issue tracks the fragment-level cleanup ownership gap for `Fragment::add_columns` / `merge_columns` when a later outer merge commit fails.

rust/lance/src/dataset/schema_evolution.rs
yyzhao2025: Thanks for the suggestion. I added doc comments for both fields to clarify that `fragments` contains the fragments returned for the final merge commit, while `fragments_to_cleanup` tracks uncommitted fragments whose newly written data files should be removed if the operation fails before the merge commit completes.

rust/lance/src/dataset/updater.rs
yyzhao2025: The current callers return immediately after cleanup, so this does not affect the current error path. However, clearing the writer makes the Updater state match the filesystem state after cleanup and avoids leaving a stale writer that points to deleted files. I’ll update `cleanup_unfinished_writer` to take and drop `self.writer` after capturing the data file path.

yanghua:
Since there is no objection, I would merge it at this moment.

PR #7153, opened 2026-06-08
perf: reuse WAND lead buffer capacity in push_back_leads

wkalt:


wkalt:
wand_fix_curve

Test of concurrent OR MatchQuery on top 16 common terms from a generated dataset of 4M 80-word documents with 7950x processor.

BubbleCal (approved):

PR #7158, opened 2026-06-08
fix(dataset)!: fail-fast casting for columns with attached indices

WenDing-Y:
BREAKING CHANGE: Callers that previously depended on the silent index-drop behavior must now explicitly call `drop_index()` before `alter_columns` with a cast, then rebuild
the index afterward.

## Summary

Currently, calling `alter_columns` with a cast on a column that has an attached
index **silently drops the index** without any warning or error. This has caused
production incidents where vector search silently regressed to brute-force scan.

This PR adds a **fail-fast check** before any cast is performed: if a target
column has any index attached, the operation is rejected with a clear error
message that names the affected column(s), the index(es), and the required
remediation (`drop_index()` → cast → rebuild index).

## Changes

- Added index-awareness check in `alter_columns` that scans cast target columns
  for attached indices
- Returns a descriptive `Error::invalid_input` with column names, index names,
  and remediation steps when a conflict is detected
- Updated existing tests that previously relied on silent index-drop behavior
  to explicitly `drop_index()` before casting
- Added new test `test_alter_columns_cast_fails_with_attached_index` covering
  the IVF_PQ index path: verifies the error message, schema integrity, index
  preservation, and that the cast succeeds after a manual drop

### DIFF
--- a/python/python/tests/test_vector_index.py
+++ b/python/python/tests/test_vector_index.py
@@ -1747,6 +1747,8 @@ def test_index_cast_centroids(tmp_path):
     values = pa.array([x for arr in centroids for x in arr], pa.float32())
     centroids = pa.FixedSizeListArray.from_arrays(values, 128)
 
+    # Cast invalidates the attached index; drop it first per the new contract.
+    dataset.drop_index(index_name)
     dataset.alter_columns(dict(path="vector", data_type=pa.list_(pa.float16(), 128)))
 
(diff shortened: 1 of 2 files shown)


WenDing-Y:
@wjones127 please review

wjones127 (approved):
This makes sense. I agree with this change.

Thanks for writing a fix.

Issue #6802, opened 2026-05-15
Query the BTree lookup batch directly instead of building a separate BTreeLookup


wjones127:
## Background

`BTreeIndex` reads its `page_lookup.lance` file into a `RecordBatch` with columns `min | max | null_count | page_idx`, then walks the batch once in `try_from_serialized` to build a separate `BTreeLookup`:

- `BTreeMap<OrderableScalarValue, Vec<PageRecord{max, page_number}>>`
- `null_pages: Vec<u32>`, `all_null_pages: Vec<u32>`

After #6793 the lookup batch is also retained on `BTreeIndex` so the index can serialize itself into a cache (`BTreeIndexState`) without re-reading from disk. This means min/max values are now stored twice — once as Arrow buffers in the batch, once as owned `ScalarValue`s inside `BTreeLookup`. For wide value types or indices with many pages this is real memory.

## Proposal

Rewrite `BTreeLookup` to query the batch directly:

- Sorted iteration: the batch rows are already ordered by ascending `min` (training sorts the data and emits pages in order).
- `pages_between(lo, hi)` becomes a binary search on the `min` column to find the starting row, then a linear scan filtering by `max`. Same big-O as today, better constants (packed arrow buffers vs scattered `BTreeMap` nodes).
- Store a small precomputed `Vec<u32>` of all-null page indices (rows where `max IS NULL`) and a non-null row range for the binary-search bounds; the rest is derived from the batch.

### Type dispatch: `arrow_ord::ord::make_comparator`

```rust
pub fn make_comparator(left: &dyn Array, right: &dyn Array, opts: SortOptions)
    -> Result<DynComparator, ArrowError>;
pub type DynComparator = Box<dyn Fn(usize, usize) -> Ordering + Send + Sync>;
```

This is exactly the primitive we need. It handles every Arrow type out of the box: all primitives via `downcast_primitive!`, plus `Boolean`, `Utf8`/`LargeUtf8`/`Utf8View`, `Binary`/`LargeBinary`/`BinaryView`, `FixedSizeBinary`, and nested types (lists, structs, FSL) via a recursive fallback. Null ordering is configurable via `SortOptions`.

With it, the binary search is ~10 lines:

```rust
fn partition_point(mins: &dyn Array, query: &dyn Array /* 1-row */) -> Result<usize> {
    let cmp = make_comparator(mins, query, SortOptions::default())?;
    let mut lo = 0;
    let mut hi = mins.len();
    while lo < hi {
[...]

PR #7161, opened 2026-06-08
perf(index): cache parsed btree lookup state

LuQQiu:
## Summary
- cache BTree lookup state as parsed `Arc<BTreeLookup>` instead of reparsing `page_lookup.lance` on cache hits
- add a cache codec that regenerates the lookup RecordBatch only when serializing cache state
- keep cache size accounting tied to the resident parsed lookup tree and add coverage for codec round-trip, cache sizing, plugin cache reuse, range-partitioned cache, and frag-reuse reconstruction

## Testing
- `cargo test -p lance-index btree_index_state`
- `cargo test -p lance-index btree_plugin_cache -- btree_range_partitioned_plugin_cache_roundtrip`
- `cargo check -p lance-index`
- `cargo clippy -p lance-index --all-targets`


wjones127 (approved):
FWIW my plan was to implement https://github.com/lance-format/lance/issues/6802

This can work as an interim solution, though.

LuQQiu:
> FWIW my plan was to implement #6802
> 
> This can work as an interim solution, though.

That would be wonderful, if we can directly query from arrow lookup batch while preserving the perf

PR #7163, opened 2026-06-08
feat: stabilize cache codec with a versioned envelope

### LINKED ISSUE #7160: Make CacheCodec serialization format stable
Making it stable means you can put the entries somewhere that (a) isn't node specific and (b) survives node restarts.

wjones127:
Implements #7160. Cache entries (FTS posting lists, scalar/vector index state) were serialized with an ad-hoc, unversioned format only safe to read in the same process that wrote it. This stabilizes the format so entries can live in a **node-agnostic, restart-surviving** cache backend.

## Wire format

Each entry is an envelope followed by a body:

```text
[magic "LCE1"][envelope_version: u8][type_id][type_version: u32]   # envelope
<body: optional protobuf header, then sections in a fixed, version-keyed order>
```

Body sections, each self-delimiting:

```text
HEADER    : [len: u32][protobuf bytes]
ARROW_IPC : [pad to 64B][self-delimiting IPC stream]
RAW_BLOB  : [len: u64][bytes]
```

## Why this shape

- **The envelope is hand-framed, not protobuf.** It's the most stability-critical part: it must parse robustly against *any* bytes (including old, pre-stabilization blobs) and never change shape. The magic is chosen so no prior blob can collide with it.
- **Decode returns `Hit`/`Miss`, never a hard error.** Wrong/absent magic, an unknown envelope version, a `type_id` mismatch, a future `type_version`, or a body decode failure all become `Miss` → recompute. Old, foreign, or corrupt bytes self-heal with **zero migration code**.
- **Bodies use protobuf headers.** Field-number evolution lets us add fields without a format break; only changes protobuf can't express transparently (reordering sections, changing a raw-blob encoding) bump `type_version`, which the reader branches on.
- **Arrow IPC sections are 64-byte aligned** so concatenated sections decode zero-copy instead of a realigning `memcpy` on every read — this guards the FTS WAND hot path.
- **`RAW_BLOB` is reserved for payloads with their own portable, self-describing encoding** (roaring bitmaps, the shared position stream).

A codec with no scalar metadata (e.g. bitmap) simply omits the header — sections are positional, so nothing is written for an absent header.

## Scope

All cache codecs migrated: FTS posting lists (compressed/plain/positions + groups), scalar indices (BTree/Bitmap/Flat/LabelList/RowAddrTreeMap), and the five IVF quantizer partitions + IVF state. The cache protos live in `lance-index/protos-cache/cache.proto` (`package lance.index.cache`) — they describe *library serialization*, not the on-disk format spec.

## Tests

Envelope round-trip and every miss path; per-codec round-trip + through-envelope zero-copy alignment (incl. RabitQ Matrix rotation, multi-batch SQ, nested bitmap in a label-list entry); additive proto-field compat; existing IVF build+search suites pass through the migrated path.

Closes #7160.


protos/cache_fts.proto
wjones127, line 1: issue(blocking): Thinking about this, we don't want to make this a part of the format spec. All of the things in this folder are. Instead they should be attached to this library. Can we move them all into `lance-index`?

rust/lance-core/src/cache/codec.rs
```diff
@@ -5,44 +5,225 @@
...
+//!
+//! Every serialized entry begins with a small hand-framed **envelope** so the
+//! reader can validate it before trusting the body:
+//!
+//! ```text
+//! [magic: 4B = b"LCE1"]
+//! [envelope_version: u8]
+//! [type_id_len: u16 LE][type_id: utf8]   # stable, author-assigned
+//! [type_version: u32 LE]                 # per-type body schema version
+//! <body, written by the type's CacheCodecImpl::serialize>
```
westonpace, line 19: I suppose this could also just be a protobuf envelope?

```
{
  version: u32,
  value: Any,
}
```

Though this raw envelope is fine as well.

rust/lance-core/src/cache/codec.rs
```diff
@@ -5,44 +5,225 @@
...
+    /// Absent or wrong magic, unknown `envelope_version`, truncated framing, or
+    /// a non-utf8 `type_id`. Typically an entry written by a pre-stabilization
+    /// or otherwise foreign build.
+    InvalidEnvelope,
+    /// Well-formed envelope, but its `type_id` names a different entry type than
+    /// the codec reading it.
+    TypeMismatch,
+    /// Written by a newer build whose `type_version` this build does not
+    /// understand and must not attempt to interpret.
+    VersionTooNew,
```
westonpace, line 161: What happens in the version too new case?  I assume the older reader will replace the entry with an old version?  If we are in a mixed-version configuration will the older and newer reader just keep thrashing the entry?  Or do we expect this case to be quite rare as newer readers should generally read older versions?

rust/lance-core/src/cache/codec.rs
```diff
@@ -5,44 +5,225 @@
 //!
 //! Implement [`CacheCodecImpl`] on concrete types, then use
 //! [`CacheCodec::from_impl`] to produce a type-erased codec for the cache.
+//!
+//! # Wire format
+//!
```
westonpace, line 10: I realize we are well past this but it just recently struck me that we are basically heading towards a world where every index has _two_ different representations.  The cold representation and the cached representation.  Why do you think this is?  Why do we need two representations?

westonpace (approved):
A few thoughts.  Also Claude came up with this, not sure if relevant:

> Bug-ish gap: PostingListGroup serializes each nested PostingList into a fresh buffer (position resets to 0), then embeds those bytes at an arbitrary, non-64-aligned offset in the outer stream. So nested IPC sections for grouped postings won't actually land on real 64-byte boundaries and won't decode zero-copy — undermining the PR's own stated goal of protecting the FTS WAND hot path for this specific case. This pattern is pre-existing (not introduced by this PR), but it's exactly the path the PR's performance rationale calls out, and there's no test catching it. Worth a follow-up issue or at least a comment.

rust/lance-core/src/cache/codec.rs
wjones127: Yeah I suppose the one thing this allows is evolving into a non-protobuf envelope. I'm not sure I care much either way.

rust/lance-core/src/cache/codec.rs
wjones127: In the cases I'm working on, the cache is local to one particular process. So no thrashing in that case. So it's largely a one-time migration.

But you are right, that in cases like a cache backed by Redis, you might have a mix of library versions and they could thrash.

I suppose we could move the version of the envelope into the cache key itself, and that would address the thrashing.

rust/lance-core/src/cache/codec.rs
wjones127: That's true. Cold representation = "compressed format", cached representation = "uncompressed, ready to mmap format". The cached representation is optimized for minimal deserialization overhead, which is why it uses zero-copy Arrow IPC.

PR #7164, opened 2026-06-08
feat(lance-select): expose selected rows accessor on NullableRowAddrSet

LuQQiu:
`true_rows()` clones both `selected` and `nulls` to compute their difference. For wire serialization (e.g. distributed scalar index execution), we want to keep the original `selected` and `nulls` as intermediate result and sent them across nodes.

Expose `selected_rows() -> &RowAddrTreeMap` for a zero-copy view of the raw backing bitmap. Profiles on distributed scalar index queries show `RoaringBitmap::clone` is a measurable hot spot and this improves performance by reducing the need for cloning them.


westonpace (approved):
I'm fine merging this but the PR description and title are a little confusing.  Is the goal to speedup wire transfers?  Or is the goal to speedup the `union_all` method?  I think this does both?

LuQQiu:
@westonpace yep it does both
to remove as many clone as possible is the workload

PR #7166, opened 2026-06-08
feat: branch-aware table version ops in directory and rest namespaces

brendanclement:
### Description

Branch-aware table version operations. Table version operations in DirectoryNamespace and RestNamespace now operate on a branch's own version chain when a branch is supplied. Branches are already a first-class ref in the Lance table format; this brings the namespace operations in line. An absent branch (or "main") behaves exactly as before.

Branch lifecycle operations. Adds CreateTableBranch, ListTableBranches, and DeleteTableBranch to DirectoryNamespace and RestNamespace, exposed by the rest_adapter at POST /v1/table/{id}/branches/{create,list,delete}. These wrap Lance's first-class Dataset branch APIs and mirror the existing tag operations, so the three layers stay uniform.

Managed versioning now supports branches end to end. Previously the namespace-managed commit store (LanceNamespaceExternalManifestStore) was bound to one branch at construction and ignored the base path handed to it, and a branch-opened managed dataset stayed rooted at the table root.

Drive-by fix. batch_delete_table_versions constructed {version}.manifest to delete, which is V1-only naming and silently matched nothing under V2 (the default scheme). It now matches actual manifests via Lance's ManifestNamingScheme.

### Testing
- Branch round-trips for all version operations in DirectoryNamespace.
- Managed branch e2e (DirectoryNamespace and rest_adapter loopback): with_branch open and commit, data files under tree/<branch>/data/, unmanaged interop read, cross-branch checkout at overlapping version numbers, branch tags via with_tag and Ref::Tag, non-latest fork; asserts managed_versioning survives the REST round trip.

### DIFF
--- a/rust/lance-namespace-impls/src/dir.rs
+++ b/rust/lance-namespace-impls/src/dir.rs
@@ -2259,7 +2259,7 @@ impl DirectoryNamespace {
 
             for (&v, version_path) in &location_by_version {
                 let vi = v as i64;
-                if !te.ranges.iter().any(|&(s, e)| vi >= s && vi <= e) {
+                if !te.ranges.iter().any(|&(s, e)| vi >= s && (e < 0 || vi < e)) {
                     continue;
                 }
                 match self.object_store.inner.delete(version_path).await {
@@ -3203,23 +3203,22 @@ impl LanceNamespace for DirectoryNamespace {
         let ranges: Vec<(i64, i64)> = request
             .ranges
             .iter()
-            .map(|r| {
-                let start = r.start_version;
-                let end = if r.end_version > 0 {
-                    r.end_version
-                } else {
-                    start
-                };
-                (start, end)
-            })
+            .map(|r| (r.start_version, r.end_version))
             .collect();
 
-        // Reject pathological ranges up front: the manifest path below builds one
-        // id per version, so (0, i64::MAX) would exhaust memory.
+        // Reject pathological bounded ranges up front: the manifest path below
+        // builds one id per version, so (0, i64::MAX) would exhaust memory. A
+        // through-latest range (end < 0) is bounded by the manifests that exist.
         const MAX_VERSIONS_PER_REQUEST: i128 = 1_000_000;
         let requested: i128 = ranges
             .iter()
-            .map(|(s, e)| (*e as i128 - *s as i128 + 1).max(0))
+            .map(|(s, e)| {
+                if *e < 0 {
+                    0
+                } else {
+                    (*e as i128 - *s as i128).max(0)
+                }
+            })
             .sum();
         if requested > MAX_VERSIONS_PER_REQUEST {
             return Err(NamespaceError::InvalidInput {
@@ -3244,6 +3243,22 @@ impl LanceNamespace for DirectoryNamespace {
             && self.table_version_storage_enabled
             && let Some(ref manifest_ns) = self.manifest_ns
         {
+            // Through-latest ranges (end_version < 0) would require enumerating the
+            // __manifest chain up to the latest version, which is not wired up here.
+            // Reject rather than silently delete physical files while leaving the
+            // __manifest records in place.
+            if table_entries
+                .iter()
+                .any(|te| te.ranges.iter().any(|&(_, e)| e < 0))
+            {
+                return Err(NamespaceError::Unsupported {
+                    message: "through-latest delete (end_version < 0) is not supported \
+                              for managed-versioning tables"
+                        .to_string(),
+                }
+                .into());
+            }
+
             // Phase 1 (atomic commit point): Delete version records from __manifest
             // for ALL tables in a single atomic operation. This is the authoritative
             // source of truth — once __manifest entries are removed, the versions
@@ -3256,7 +3271,7 @@ impl LanceNamespace for DirectoryNamespace {
                     &te.table_id.clone().unwrap_or_default(),
                 );
                 for (start, end) in &te.ranges {
-                    for version in *start..=*end {
+                    for version in *start..*end {
                         let object_id = manifest::ManifestNamespace::build_version_object_id(
                             &table_id_str,
                             version,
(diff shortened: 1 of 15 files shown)


rust/lance-namespace-impls/src/dir.rs
```diff
@@ -1160,6 +1161,35 @@ impl DirectoryNamespace {
         ObjectStore::extract_path_from_uri(registry, uri)
     }
 
+    /// Normalize and validate a branch selector: `None`, empty, and `main` mean
+    /// the main branch; any other name is validated with lance's
+    /// `check_valid_branch` (lance skips this on the open path) so it cannot
+    /// escape the table root via `..`.
+    fn normalized_branch(branch: Option<&str>) -> Result<Option<&str>> {
+        match branch.filter(|b| !b.is_empty() && *b != "main") {
```
jackye1995, line 1209: no need to be in this PR, but we should probably just add it to the spec that "main" is reserved for the main branch so that we have consistent behavior.

jackye1995 (commented):
Could we avoid constructing branch storage paths in DirectoryNamespace and route this through Lance Dataset branch functionality instead?

Dataset already has first-class branch APIs for this: `DatasetBuilder::with_branch(branch, version)`, `Dataset::checkout_branch`, `Dataset::checkout_version((branch, version))`, and `dataset.branches().get/list` for branch metadata/source-of-truth checks. In particular, `Dataset::create_branch` documents that `BranchContents` is the source of truth for a branch.

This code currently adds `normalized_branch` / `branch_table_uri` and derives `<table>/tree/<branch>` directly. That makes the namespace implementation depend on Lance's internal branch layout and treats the physical directory as authoritative. For example, `create_table_version` can copy a manifest into `tree/ghost/_versions/...` without verifying that `_refs/branches/ghost.json` exists, after which namespace version APIs can report a branch that `Dataset::list_branches()` does not know about.

Can we instead open/resolve the requested branch through Dataset APIs, and only fall back to adding a small public Dataset helper if the namespace response needs manifest-location metadata that is not exposed today? At minimum, branch operations should validate branch existence via `dataset.branches().get(branch)` and derive branch location through Dataset/commit-handler semantics rather than string-appending `tree/<branch>`.

jackye1995:
Also I assume we will do the branch CRUD APIs implementation separately?

rust/lance-namespace-impls/src/dir.rs
```diff
@@ -2130,40 +2183,66 @@ impl DirectoryNamespace {
...
+                .into_iter()
+                .filter_map(|meta| {
+                    let version = Self::manifest_version_from_filename(meta.location.filename()?)?;
+                    Some((version, meta.location))
+                })
+                .collect();
+
+            for (&v, version_path) in &location_by_version {
+                let vi = v as i64;
+                if !te.ranges.iter().any(|&(s, e)| vi >= s && vi <= e) {
```
jackye1995: While this range handling appears to predate the branch support, the new branch delete path inherits it and the new tests now pin the wrong semantics. `VersionRange.end_version` is documented as exclusive, with `end_version = -1` meaning through latest, but this code treats the end as inclusive and normalizes `<= 0` to `start_version`. Can we fix the range normalization/matching here before adding branch coverage, so branch deletes follow the request contract?

jackye1995:
I think the branch-aware namespace version APIs also need to be wired through the normal managed-versioning Dataset path. `LanceNamespaceExternalManifestStore` still constructs describe/list/create table-version requests with `..Default::default()`, so `branch` is always omitted. That means a namespace-managed dataset checked out to a branch can still resolve or create namespace versions against the main chain instead of the active branch. Can we carry the dataset branch into `LanceNamespaceExternalManifestStore` and set it on the namespace requests, with a test for committing to a namespace-managed branch?

brendanclement:
> Also I assume we will do the branch CRUD APIs implementation separately?

was going to add to another PR but it's difficult to work with stacked PRs in a forked repo. I've added here as another commit instead.

brendanclement:
> I think the branch-aware namespace version APIs also need to be wired through the normal managed-versioning Dataset path. `LanceNamespaceExternalManifestStore` still constructs describe/list/create table-version requests with `..Default::default()`, so `branch` is always omitted. That means a namespace-managed dataset checked out to a branch can still resolve or create namespace versions against the main chain instead of the active branch. Can we carry the dataset branch into `LanceNamespaceExternalManifestStore` and set it on the namespace requests, with a test for committing to a namespace-managed branch?

i've been playing around with this on my workspace but wanted your thoughts on the approach first.

I can put a branch field on LanceNamespaceExternalManifestStore and it sets it on describe/list/create version requests

the part i'm less sure on is where the branch should come from. the way i'm thinking is that i read the branch off the builder's version ref (the with_branch value). from_namespace was already building the commit handler in the builder, the bug was just that it built it before with_branch ran so branch was always None. I'm thinking I can defer that construction to load where the branch is known and passes it through and no new param is needed (an alternative was adding a branch arg to from_namespace but that's redundant with with_branch, you'd be passing it in two places.)

one problem with this approach though is write_into_namespace is still main-only because WriteParams has no branch field so i can't thread it through. I'm thinking of creating another function write_into_namespace_on_branch() with a branch parameter that we can move callers to - what do you think?

rust/lance-namespace-impls/src/dir.rs
```diff
@@ -4273,6 +4455,156 @@ impl LanceNamespace for DirectoryNamespace {
...
+        // returns RefConflict up front.
+        if dataset.branches().get(&request.name).await.is_ok() {
+            return Err(NamespaceError::TableBranchAlreadyExists {
+                message: format!("branch '{}' for table at '{}'", request.name, table_uri),
+            }
+            .into());
+        }
+
+        dataset
+            .create_branch(
```
jackye1995, line 4500: I think this still clones from the table root even when `from_branch` is set. `Dataset::create_branch` resolves `(source_branch, version_number)` from this `(from_branch, from_version)` tuple, but the clone operation it creates uses `ref_path: self.uri()`. Since this namespace code opened `dataset` from the table root, `self.uri()` is main, so the commit resolves `from_version` under main's manifest path instead of under `from_branch`.

For example, if `dev` has diverged and we call `create_table_branch(name = "child", from_branch = "dev", from_version = 2)`, this will either clone main v2 or fail if main has no v2, rather than cloning dev v2. Can we make the clone source path branch-aware, e.g. have `Dataset::create_branch` set `ref_path` from `self.branch_location().find_branch(source_branch.as_deref())?` instead of always using `self.uri()`?

brendanclement: good catch, fixed exactly how you suggested and added tests

jackye1995 (approved):
looks good to me!

PR #7177, opened 2026-06-09
feat(python): expose zonemap segment builds

everySympathy:
## Summary

- allow Python `create_index_uncommitted(..., index_type="ZONEMAP", fragment_ids=...)` to use the scalar segment build path
- reject `create_scalar_index(..., index_type="ZONEMAP", fragment_ids=...)` with the same migration guidance used for BTREE/BITMAP segment-native scalar builds
- add Python regression coverage for ZoneMap fragment-id validation and for staging, merging, committing, and querying ZoneMap segments

## Context

ZoneMap segment merge support landed in #7128, but the Python public `create_index_uncommitted` helper still only routed BTREE/BITMAP/INVERTED scalar requests through the uncommitted scalar segment path. As a result, ZONEMAP fell through to vector validation and failed on scalar columns.

I searched for open duplicate PRs with `ZONEMAP create_index_uncommitted Python distributed` and `zonemap uncommitted python`; no open matches were found.

## Validation

- `UV_PYTHON=/usr/bin/python3.11 uv run pytest python/tests/test_scalar_index.py::test_fragment_ids_parameter_validation python/tests/test_scalar_index.py::test_segment_fts python/tests/test_scalar_index.py::test_zonemap_fragment_ids_parameter_validation python/tests/test_scalar_index.py::test_zonemap_segment_merge_and_commit_from_python python/tests/test_scalar_index.py::test_bitmap_uncommitted_segments_can_be_committed_from_python python/tests/test_scalar_index.py::test_btree_fragment_ids_parameter_validation` passed
- `UV_PYTHON=/usr/bin/python3.11 uv run ruff format --check --diff python/lance/dataset.py python/tests/test_scalar_index.py` passed
- `UV_PYTHON=/usr/bin/python3.11 uv run ruff check python/lance/dataset.py python/tests/test_scalar_index.py` passed
- `UV_PYTHON=/usr/bin/python3.11 uv run make lint` passed

### DIFF
--- a/python/python/lance/dataset.py
+++ b/python/python/lance/dataset.py
@@ -3074,6 +3074,38 @@ class LanceDataset(pa.dataset.Dataset):
         else:
             raise Exception("index_type must be str or IndexConfig")
 
+    @staticmethod
+    def _normalized_index_type(
+        index_type: Union[str, IndexConfig],
+    ) -> str:
+        if isinstance(index_type, IndexConfig):
+            index_type = index_type.index_type
+        return index_type.upper()
+
+    @classmethod
+    def _is_segment_native_scalar_index_type(
+        cls,
+        index_type: Union[str, IndexConfig],
+    ) -> bool:
+        return cls._normalized_index_type(index_type) in {
+            "BTREE",
+            "BITMAP",
+            "INVERTED",
+            "FTS",
+            "ZONEMAP",
+        }
+
+    @classmethod
+    def _requires_uncommitted_scalar_index(
+        cls,
+        index_type: Union[str, IndexConfig],
+    ) -> bool:
+        return cls._normalized_index_type(index_type) in {
+            "BTREE",
+            "BITMAP",
+            "ZONEMAP",
+        }
+
     def create_scalar_index(
         self,
         column: str,
@@ -3291,7 +3323,9 @@ class LanceDataset(pa.dataset.Dataset):
             column, index_type, kwargs
         )
 
-        if fragment_ids is not None and logical_index_type in {"BTREE", "BITMAP"}:
+        if fragment_ids is not None and self._requires_uncommitted_scalar_index(
+            logical_index_type
+        ):
             raise ValueError(
                 f"{logical_index_type} distributed indexing uses "
                 "create_index_uncommitted(..., "
@@ -4004,7 +4038,8 @@ class LanceDataset(pa.dataset.Dataset):
         Create one segment without publishing it and return its metadata.
 
         This is the public distributed-build API for vector, BTREE scalar,
-        canonical bitmap scalar, and INVERTED scalar index construction. Unlike
+        canonical bitmap scalar, INVERTED scalar, and ZONEMAP scalar index
+        construction. Unlike
         :meth:`create_index`, this method does not publish the index into the
         dataset manifest. Instead, it writes one segment under
         ``_indices/<segment_uuid>/`` and returns the resulting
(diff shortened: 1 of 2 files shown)


python/python/tests/test_scalar_index.py
```diff
@@ -4008,6 +4008,71 @@ def test_bitmap_uncommitted_segments_can_be_committed_from_python(tmp_path):
     )
 
 
+def test_zonemap_uncommitted_segments_can_be_merged_and_committed_from_python(tmp_path):
```
zhangyue19921010: Minor: Simplify the names of the test cases.

python/python/lance/dataset.py
```diff
@@ -3291,7 +3291,11 @@ def create_scalar_index(
             column, index_type, kwargs
         )
 
-        if fragment_ids is not None and logical_index_type in {"BTREE", "BITMAP"}:
+        if fragment_ids is not None and logical_index_type in {
```
zhangyue19921010: Minor: {"BTREE", "BITMAP", "ZONEMAP"} appears multiple times. Can we abstract it into a method?

python/python/tests/test_scalar_index.py
everySympathy: simplified to` test_zonemap_segment_merge_and_commit_from_python`

python/python/lance/dataset.py
everySympathy: extract these code into a new function: `_is_segment_native_scalar_index_type`

yanghua:
@claude review

python/python/tests/test_scalar_index.py
```diff
@@ -4008,6 +4008,71 @@ def test_bitmap_uncommitted_segments_can_be_committed_from_python(tmp_path):
     )
 
 
+def test_zonemap_segment_merge_and_commit_from_python(tmp_path):
+    ds = generate_multi_fragment_dataset(
+        tmp_path, num_fragments=4, rows_per_fragment=40
```
yanghua: Can we add the number about `rows_per_fragment` to be larger than 8192(e.g. > 2 * 8192), so that we can have two zones in one fragment? Because the default value of `rows_per_zone ` is 8192.

python/python/tests/test_scalar_index.py
```diff
@@ -4008,6 +4008,71 @@ def test_bitmap_uncommitted_segments_can_be_committed_from_python(tmp_path):
...
+    )
+
+    index_name = "id_zonemap_segments"
+    fragment_ids = [fragment.fragment_id for fragment in ds.get_fragments()]
+    with pytest.raises(ValueError, match="create_index_uncommitted"):
+        ds.create_scalar_index(
+            column="id",
+            index_type="ZONEMAP",
+            fragment_ids=[fragment_ids[0]],
+        )
```
yanghua, line 4068: IMO, it would be better to add a test case named .e.g `test_zonemap_fragment_ids_parameter_validation`?

yanghua (commented):
Left two comments.

python/python/tests/test_scalar_index.py
everySympathy: Adjusted the merge/commit test to use `rows_per_fragment = 20_000`, so each fragment contains multiple ZoneMap zones. The query now targets rows in the second zone.

python/python/tests/test_scalar_index.py
everySympathy: Split the `create_scalar_index(..., fragment_ids=...)` validation into `test_zonemap_fragment_ids_parameter_validation`, leaving the merge/commit test focused on the segment lifecycle.

yanghua (approved):
LGTM. Thanks for your contribution!

PR #7179, opened 2026-06-09
feat(vector)!: add approx mode for RaBitQ search

BubbleCal:
## Feature

This PR adds a public `approx_mode` setting for vector search with `fast`, `normal`, and `accurate` values. The public API avoids exposing RaBitQ/HACC terminology while still allowing callers to choose the speed/accuracy tradeoff when the backing index supports it.

## Implementation

- Adds `ApproxMode` to vector queries and threads it through Rust scanner, ANN proto serialization, Python query parsing, and FlatIndex distance calculator options.
- Implements RaBitQ behavior for each mode:
  - `fast`: force 1-bit query-time scoring for RaBitQ.
  - `normal`: preserve the existing search path.
  - `accurate`: use a 16-bit LUT accumulation path for the binary RaBitQ estimator.
- Extends query scratch with a wider accumulation buffer while keeping the existing `QueryScratchCapacity::new(...)` API compatible.
- Adds Python API support via `approx_mode="fast" | "normal" | "accurate"` and rejects invalid values.

## Benchmark

IVF_RQ on dbpedia, `num_bits=5`, no `refine_factor`. The plot shows `approx_mode=fast`, `normal`, and `accurate` as separate curves.

IVF_RQ dbpedia approx_mode benchmark, num_bits=5

## Breaking Change

BREAKING CHANGE: The ANN protobuf schema now includes `VectorApproxMode approx_mode` on vector query serialization. Consumers that regenerate or explicitly match Lance's serialized ANN query proto should update to the new schema.

## Validation

- `cargo fmt --all`
- `cargo test -p lance-linalg simd::dist_table`
- `cargo test -p lance-index vector::bq::storage`
- `cargo test -p lance-index vector::storage::tests`
- `cargo test -p lance --features substrait test_query_roundtrip`
- `cargo test -p lance test_knn_approx_mode_defaults_and_setter`
- `cargo check --workspace --tests --benches`
- `cd python && make install`
- `cd python && uv run pytest python/tests/test_vector_index.py::test_vector_index_with_approx_mode python/tests/test_vector_index.py::test_vector_index_invalid_approx_mode`
- `cd python && uv run pytest python/tests/test_vector_index.py::test_create_ivf_rq_multi_bit_searches_l2_and_cosine`
- `git diff --check`

### DIFF
--- a/protos/ann.proto
+++ b/protos/ann.proto
@@ -32,6 +32,8 @@ message VectorQueryProto {
   bool use_index = 11;
   optional float dist_q_c = 12;
   optional int32 query_parallelism = 13;
+  // Currently only affects RQ-quantized vector indexes, such as IVF_RQ.
+  // Other index types ignore this setting.
   VectorApproxMode approx_mode = 14;
 }
 
--- a/protos/ann.proto
+++ b/protos/ann.proto
@@ -14,11 +14,12 @@ import "index.proto";
 // This currently only affects RQ-quantized vector indexes, such as IVF_RQ.
 // Other index types ignore this setting.
 enum VectorApproxMode {
-  // Use the default Lance search path.
+  // Use all RQ bits for query-time scoring with u8-quantized lookup tables.
   Normal = 0;
-  // Prefer lower latency by forcing 1-bit RQ scoring at query time.
+  // Use only one RQ bit for query-time scoring, even for multi-bit indexes.
   Fast = 1;
-  // Prefer higher accuracy by using the high-accuracy RQ estimator.
+  // Use all RQ bits for query-time scoring with u16-quantized lookup tables
+  // to reduce estimator quantization error.
   Accurate = 2;
 }
 
--- a/protos/ann.proto
+++ b/protos/ann.proto
@@ -9,6 +9,12 @@ import "table_identifier.proto";
 import "table.proto";
 import "index.proto";
 
+enum VectorApproxMode {
+  Normal = 0;
+  Fast = 1;
+  Accurate = 2;
+}
+
 // Serialized vector query parameters.
 message VectorQueryProto {
   // Query vector as Arrow IPC bytes (supports Float16, Float32, Float64, UInt8, etc.)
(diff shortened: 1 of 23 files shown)


protos/ann.proto
```diff
@@ -9,6 +9,12 @@ import "table_identifier.proto";
 import "table.proto";
 import "index.proto";
 
+enum VectorApproxMode {
```
Xuanwo, line 16: What's the meaning of those three modes? Can you add comments for them?

protos/ann.proto
```diff
@@ -26,6 +32,9 @@ message VectorQueryProto {
   bool use_index = 11;
   optional float dist_q_c = 12;
   optional int32 query_parallelism = 13;
+  // Currently only affects RQ-quantized vector indexes, such as IVF_RQ.
+  // Other index types ignore this setting.
+  VectorApproxMode approx_mode = 14;
```
Xuanwo, line 45: I think we should mark this PR as a breaking change.

protos/ann.proto
Xuanwo, line 39: Do we need to define VectorApproxMode in `lance.index`?

protos/ann.proto
BubbleCal: Added comments for `VectorApproxMode` and each enum value. The comments also call out that this currently only affects RQ-quantized indexes such as IVF_RQ, and that other index types ignore it.

protos/ann.proto
BubbleCal: Done. I marked the PR title as `feat(vector)!: add approx mode for RaBitQ search` and added a `BREAKING CHANGE:` section to the PR body for the ANN protobuf schema update.

protos/ann.proto
BubbleCal: I kept this in `ann.proto` instead of `lance.index` because `approx_mode` is a query-time execution option serialized with `VectorQueryProto`, not persisted index metadata. It only changes RQ-quantized search behavior at execution time; non-RQ index types ignore it.

Xuanwo (approved):

PR #7180, opened 2026-06-09
fix: reserve main branch name

### LINKED ISSUE #7167: Mark `main` a reserved branch name for the main branch
I think it would be more clear at this point to just make it official, embed this logic to the dataset branch APIs, so that all the client behaviors are more consistent. What do we think? @majin1102 @brendanclement

majin1102:
Close #7167

### DIFF
--- a/docs/src/format/table/branch_tag.md
+++ b/docs/src/format/table/branch_tag.md
@@ -17,7 +17,9 @@ Branch names must follow these validation rules:
 4. Cannot contain `..` or `\`
 5. Segments must contain only alphanumeric characters, `.`, `-`, `_`
 6. Cannot end with `.lock`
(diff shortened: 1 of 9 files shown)


brendanclement (approved):

jackye1995:
I think this likely deserve a vote and also update the format spec accordingly?

majin1102:
> I think this likely deserve a vote and also update the format spec accordingly?

Yeah, sure.
I will update the format spec. This is actually an experience optimiziation. The real sematics has been established in earlier PRs. But I agree we coud have a vote to make this formal.

majin1102:
Related vote: #7229

majin1102:
Hey, @jackye1995 the voting has wrapped up. Could you please review this PR when you get a chance? 

Thanks

PR #7186, opened 2026-06-09
perf(index): query BTree lookup batch directly

### LINKED ISSUE #6802: Query the BTree lookup batch directly instead of building a separate BTreeLookup

wjones127:
Closes #6802.

`BTreeIndex` held both the `page_lookup.lance` batch and a parallel `BTreeLookup` (a `BTreeMap<OrderableScalarValue, Vec<PageRecord>>` plus null-page lists), duplicating every min/max value as owned `ScalarValue`s alongside the Arrow buffers.

This PR rewrites `BTreeLookup` to wrap the lookup batch as the single source of truth:

- Range searches binary-search the sorted `min` column via `arrow_ord::make_comparator`, then scan forward filtering by `max` and classify `Matches::Some`/`All` — same big-O as before, with packed Arrow buffers instead of scattered tree nodes. Only the small all-null / partial-null page index lists are precomputed.
- Equality / `IN` lookups go through a shared `candidate_pages_for_values` that compares query values against the page columns with a native, inlined comparator (no boxed `DynComparator` vtable call per comparison). Dispatch is on the **physical storage type**: logical types backed by the same native are reinterpreted to one path — zero-copy when already that type, otherwise an O(1) `ArrayData` relabel with no value copy — so every `Date32`/`Time32`/`Decimal32`/`IntervalYearMonth` reuses the `i32` path and every `Date64`/`Time64`/`Timestamp`/`Duration`/`Decimal64` reuses the `i64` path, rather than each generating its own. Byte-like columns (`Utf8`/`LargeUtf8`/`Binary`/`LargeBinary`/`FixedSizeBinary`) compare lexicographically via `ArrayAccessor`; intervals with struct natives and booleans fall back to `make_comparator`. Comparators are built once per query and reused across all values. This keeps the scan to ~19 monomorphizations instead of one per logical type.
- `BTreeIndex` no longer stores a separate `lookup_batch`; `statistics()` reads bounds from the batch and cache serialization clones the batch out of `page_lookup`.
- Float ordering uses `total_cmp` (`ArrowNativeTypeOp::compare` on the native path, `make_comparator` on the fallback), matching the previous `OrderableScalarValue` ordering — so the NaN caveat in the issue is a non-issue.

The first commit reverts #7161, which had taken the opposite approach (caching the parsed tree and regenerating the batch on serialize); that machinery is obsolete once the batch is the source of truth.

## Testing
- `cargo test -p lance-index --lib` (all pass, incl. NaN ordering, null handling, range/fragment consistency)
- `test_btree_lookup_pages_between` covers duplicate `min`s, a null-`min` straddling page, Some/All classification, and empty/inverted ranges.
- `test_btree_lookup_pages_eq_bytes` covers the native byte path for `Binary` and `FixedSizeBinary` (e.g. UUID columns).
- `test_btree_lookup_pages_eq_temporal` covers the physical-type reinterpret (`Date32`→`i32`, `Timestamp`→`i64`).

## Benchmarks

`cargo bench -p lance-index --bench btree` (the existing suite: numeric + string, high/low cardinality, equality / range / `IN`, cached + uncached). Compared against current `main` (which includes #7161) using criterion's baseline significance test — `--load-baseline <pr> --baseline main` — so the numbers below are criterion's own bootstrapped change estimate `[lower, upper]` around the median, with its p-value. Config: `sample_size = 10`, `measurement_time = 10s`, default 2% noise threshold. Run on a single dev machine (macOS arm64), so absolute timings are machine-specific; the verdicts are what matter. Equality/`IN` are from the current HEAD; `range_*` exercises `pages_between`, which the equality/IN dispatch work did not touch, so those numbers reflect the same code.

### Wins — low-cardinality and load/deserialize-bound cases

No longer rebuilding a `BTreeMap` on load, plus the native comparator inlining:

| case | change | p | verdict |
|---|---|---|---|
| equality/int_low_card/no_cache | −27.5% [−30.7, −24.3] | 0.00 | improved |
| equality/string_low_card/no_cache | −26.6% [−29.3, −24.3] | 0.00 | improved |
| range_few/int_low_card/no_cache | −23.7% [−26.5, −21.4] | 0.00 | improved |
| range_few/string_low_card/no_cache | −23.5% [−24.7, −21.9] | 0.00 | improved |
| equality/string_low_card/cached | −20.3% [−22.8, −18.5] | 0.00 | improved |
| range_few/string_low_card/cached | −18.0% [−19.4, −16.3] | 0.00 | improved |
| equality/int_low_card/cached | −6.64% [−7.90, −5.61] | 0.00 | improved |
| range_few/int_low_card/cached | −6.34% [−8.02, −4.49] | 0.00 | improved |

### Hot-path point / `IN` lookups (native physical-type dispatch)

An earlier iteration that queried the batch through the boxed `make_comparator` (one vtable call per comparison) regressed warm-cache high-cardinality lookups by up to +44% on a 30-value `IN`. The native comparators remove that. **Vs `main`, all 32 equality+IN benchmarks are improved or at parity — zero regressions.** High-cardinality warm-cache equality is back to parity; warm-cache `IN` is improved:

| case | change vs main | p | verdict |
|---|---|---|---|
| equality/int_unique/cached | parity | — | no change |
| equality/string_unique/cached | parity | — | no change |
| in_30/int_unique/cached | −9.67% [−10.7, −9.0] | 0.00 | improved |
| in_20/int_unique/cached | −3.81% [−4.11, −3.57] | 0.00 | improved |
| in_10/int_unique/cached | −2.80% [−4.84, −1.27] | 0.00 | improved |
| in_20/string_low_card/cached | −5.40% [−8.39, −2.96] | 0.00 | improved |
| in_10/string_low_card/cached | −3.05% [−4.36, −1.94] | 0.00 | improved |
| in_30/string_low_card/cached | −2.06% [−3.15, −1.23] | 0.00 | improved |

### Noise calibration

To measure the harness noise floor, comparing two runs of **identical code** (same binary behavior) still produces "significant" verdicts:

| case (identical code) | change | p | verdict |
|---|---|---|---|
| in_10/string_unique/cached | −4.29% [−7.41, −1.86] | 0.01 | "improved" |
| equality/int_unique/cached | −2.87% [−5.60, −0.68] | 0.02 | "improved" |
| equality/int_unique/no_cache | −3.14% [−4.42, −2.02] | 0.00 | "improved" |

The code did not change, so these p < 0.05 verdicts are run-to-run variance. Consistent with that, `in_10/int_unique/cached` has read anywhere from −4.2% to +9.3% across runs. At `sample_size = 10` on warm multi-µs benchmarks, swings of ±5% earn p < 0.05 from variance alone, so single-digit-percent deltas on the warm high-cardinality cases are not reliable signal.

### Residual cost

`range_*/unique/cached` shows small regressions — e.g. `range_few/string_unique/cached +2.7%`, `range_few/int_unique/cached +3.8%`. `pages_between` (the range path) still uses `make_comparator`; the native physical-type dispatch was added only to the equality/`IN` path. These sit at the edge of the calibrated noise band but are plausibly a small real cost — the deliberate tradeoff for not duplicating every min/max as an owned `ScalarValue` in memory, and a candidate for extending the physical-type dispatch to ranges in a follow-up. All other cases report "no change in performance detected."

### DIFF
--- a/rust/lance-index/src/scalar/btree.rs
+++ b/rust/lance-index/src/scalar/btree.rs
@@ -28,7 +28,7 @@ use crate::{
 use crate::{metrics::NoOpMetricsCollector, scalar::registry::TrainingCriteria};
 use crate::{pbold, scalar::btree::flat::FlatIndex};
 use arrow_arith::numeric::add;
-use arrow_array::{Array, RecordBatch, UInt32Array, new_empty_array};
+use arrow_array::{Array, ArrayAccessor, RecordBatch, UInt32Array, cast::AsArray, new_empty_array};
 use arrow_ord::ord::make_comparator;
 use arrow_schema::{DataType, Field, Schema, SortOptions};
 use async_trait::async_trait;
@@ -613,6 +613,26 @@ fn partition_point(lo: usize, hi: usize, mut pred: impl FnMut(usize) -> bool) ->
     lo
 }
 
+/// Builds a comparator over two array accessors of the same `Ord` item type,
+/// matching arrow's NULLs-first ascending order (`null < non-null`, `null == null`).
+///
+/// Unlike [`make_comparator`], the returned closure is generic (not boxed), so the
+/// element comparison inlines into the scan instead of dispatching through a vtable
+/// on every call.
+fn accessor_cmp<'a, T, L, R>(left: L, right: R) -> impl Fn(usize, usize) -> Ordering + 'a
+where
+    T: Ord,
+    L: ArrayAccessor<Item = T> + 'a,
+    R: ArrayAccessor<Item = T> + 'a,
+{
+    move |i, j| match (left.is_null(i), right.is_null(j)) {
+        (true, true) => Ordering::Equal,
+        (true, false) => Ordering::Less,
+        (false, true) => Ordering::Greater,
+        (false, false) => left.value(i).cmp(&right.value(j)),
+    }
+}
+
 /// Satisfies scalar queries by searching the `page_lookup.lance` batch directly.
 ///
 /// The batch holds one row per page with columns `min | max | null_count | page_idx`,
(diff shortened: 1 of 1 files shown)


rust/lance-index/src/scalar/btree.rs
```diff
@@ -594,46 +603,132 @@ impl Ord for OrderableScalarValue {
...
+            && self.null_pages == other.null_pages
+            && self.all_null_pages == other.all_null_pages
+            && self.search_start == other.search_start
     }
 }
 
-#[derive(Debug, Copy, Clone)]
+impl DeepSizeOf for BTreeLookup {
+    fn deep_size_of_children(&self, context: &mut lance_core::deepsize::Context) -> usize {
+        self.batch.get_array_memory_size()
```
wjones127: issue(blocking): this isn't how we handle record batches anymore. We should be able to do `deep_size_of_children` on it.

rust/lance-index/src/scalar/btree.rs
```diff
@@ -594,46 +603,132 @@ impl Ord for OrderableScalarValue {
...
 }
 
-/// An in-memory structure that can quickly satisfy scalar queries using a btree of ScalarValue
-#[derive(Debug, DeepSizeOf, PartialEq, Eq)]
+/// Satisfies scalar queries by searching the `page_lookup.lance` batch directly.
+///
+/// The batch holds one row per page with columns `min | max | null_count | page_idx`,
+/// sorted ascending by `min` with NULLs first (the order the index is trained in).
+/// The batch is the single source of truth, avoiding a parallel `BTreeMap` of owned
+/// `ScalarValue`s alongside the Arrow buffers. Both query paths binary-search the
```
wjones127: nitpick: We don't really have reason to reference the old implementation in the comments. That sort of thing belongs in the PR description.

LuQQiu (approved):
Great improvement, performance validated in benchmarking

PR #7200, opened 2026-06-10
fix(fts): recompute next_id when loading FST token set

wkalt:
Token ids in a TokenSet are dense [0, len), so next_id must equal the token count. load_fst restored next_id verbatim from the persisted _token_next_id column, so a stale value written by a binary built before #7115 -- whose remap dropped tokens without lowering next_id -- survived the reload. A later segment merge then minted an out-of-range token id and panicked in InnerBuilder::merge_from at self.posting_lists[new_token_id].

Recompute next_id from map.len() (O(1)) instead, mirroring load_arrow, which already does this. Heals existing indexes on read with no migration; writers and file schema are unchanged.

Complements 12f529f9 (#7115), which reset next_id after remap in memory but did not heal values already persisted by writers built before it.

### DIFF
--- a/rust/lance-index/src/scalar/inverted/index.rs
+++ b/rust/lance-index/src/scalar/inverted/index.rs
@@ -1620,15 +1620,13 @@ impl TokenSet {
         let map = fst::Map::new(bytes.to_vec())
             .map_err(|e| Error::index(format!("failed to load fst tokens: {}", e)))?;
 
-        let next_id_col = batch[TOKEN_NEXT_ID_COL].as_primitive::<datatypes::UInt32Type>();
         let total_length_col =
             batch[TOKEN_TOTAL_LENGTH_COL].as_primitive::<datatypes::UInt64Type>();
 
-        let next_id = next_id_col
-            .values()
-            .first()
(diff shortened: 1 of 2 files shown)


rust/lance-index/src/scalar/inverted/index.rs
```diff
@@ -1620,15 +1620,13 @@ impl TokenSet {
         let map = fst::Map::new(bytes.to_vec())
             .map_err(|e| Error::index(format!("failed to load fst tokens: {}", e)))?;
 
-        let next_id_col = batch[TOKEN_NEXT_ID_COL].as_primitive::<datatypes::UInt32Type>();
```
westonpace, line 1623: Do we even need to write this at all?

westonpace (approved):
Nice find.  Glad we were able to heal it without issue.

rust/lance-index/src/scalar/inverted/index.rs
wkalt: I checked, it's not read by the new code. But maybe we need it for old readers that don't have this fix?

PR #7203, opened 2026-06-10
feat(index): support xabi scalar index plugins

Xuanwo:
## Summary

This PR adds scalar index plugin support on top of `xabi`, allowing Lance to load session-scoped index implementations from dynamic libraries without making plugins depend on Lance internals.

`xabi` is used as the stable ABI layer between Lance and plugins. In this model, `#[xabi::data]` defines versioned wire data types, and `#[xabi::xabi]` defines trait contracts that are lowered into C-compatible vtables, handles, borrowed references, futures, and owned byte buffers. Lance and a plugin communicate through those generated ABI contracts instead of Rust trait objects from the same crate build. Larger payloads stay format-oriented: Arrow data crosses the boundary as Arrow IPC, and query expressions cross as Substrait bytes plus host-side JSON context.

The Lance session now owns plugin registrations and exposes scalar index integration through the existing index registry. A fixture mock btree plugin lives outside the workspace as a test-only xabi dynamic library, so the end-to-end dynamic loading path is covered without adding another published crate.

## Reviewing the ABI

The concrete ABI should be reviewed in these files:

- `rust/lance-index-plugin-abi/src/lib.rs`
  - Contract IDs and `ABI_VERSION`
  - `#[xabi::data]` request/response structs
  - `#[xabi::xabi]` traits for plugin, index, training data, index store, and build progress
  - Arrow IPC and little-endian encoding helpers used at the boundary

- `rust/lance-index-plugin-abi/xabi/snapshots/lance.*/*.txt`
  - Generated xabi layout snapshots for each public contract
  - Concrete field offsets, sizes, alignments, vtable sizes, and contract IDs for supported targets

The host-side adapter is in `rust/lance/src/session/xabi_index_plugin.rs`, and the test plugin implementation is in `rust/lance/tests/xabi_plugins/mock_btree/src/lib.rs`.

## ABI stability

The ABI stability story is intentionally centered on xabi rather than Lance internals:

- Public contracts use stable `lance.*` contract IDs and an explicit `ABI_VERSION`.
- xabi data layouts include size and ABI version fields, allowing prefix-compatible extension.
- xabi vtables track minimum and full sizes so older/newer sides can validate available methods.
- Boundary validation checks ABI version, vtable size, null pointers, and field availability before use.
- `xabi-assert` snapshots lock the generated ABI layout; intentional ABI changes require updating the snapshots.
- The ABI crate avoids depending on Lance internals, so plugins bind to stable wire contracts instead of private Rust types.

### DIFF
--- a/rust/lance-index-plugin-abi/src/lib.rs
+++ b/rust/lance-index-plugin-abi/src/lib.rs
@@ -1,0 +1,289 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright The Lance Authors
+
+//! Stable xabi contracts for Lance index plugins.
+//!
+//! This crate is the ABI boundary between Lance and dynamically loaded index
+//! plugins. It intentionally avoids depending on Lance internals. Data crosses
+//! the boundary as xabi data, Arrow IPC record batches, and Substrait bytes.
+
+use std::io::Cursor;
+
+use arrow_array::RecordBatch;
+use arrow_ipc::reader::StreamReader;
+use arrow_ipc::writer::StreamWriter;
+use arrow_schema::SchemaRef;
+
+pub const SCALAR_PLUGIN_TRAIT_ID: &str = "lance.scalar-index-plugin";
+pub const SCALAR_INDEX_TRAIT_ID: &str = "lance.scalar-index";
+pub const INDEX_STORE_TRAIT_ID: &str = "lance.index-store";
+pub const TRAINING_DATA_TRAIT_ID: &str = "lance.training-data";
+pub const BUILD_PROGRESS_TRAIT_ID: &str = "lance.index-build-progress";
+pub const ABI_VERSION: u32 = 1;
+
+pub const ORDERING_NONE: u32 = 0;
+pub const ORDERING_VALUES: u32 = 1;
+pub const ORDERING_ADDRESSES: u32 = 2;
+
+pub const EXACTNESS_EXACT: u32 = 0;
+pub const EXACTNESS_AT_MOST: u32 = 1;
+pub const EXACTNESS_AT_LEAST: u32 = 2;
+
+pub const UPDATE_REQUIRES_OLD_DATA: u32 = 0;
+pub const UPDATE_ONLY_NEW_DATA: u32 = 1;
+
+#[xabi::data]
+#[derive(Debug, Clone, Copy)]
+pub struct TrainingCriteriaAbi {
+    pub ordering: u32,
+    pub needs_row_ids: u8,
+    pub needs_row_addrs: u8,
+}
+
+#[xabi::data]
+#[derive(Debug, Clone, Copy)]
+pub struct UpdateCriteriaAbi {
+    pub mode: u32,
+    pub data_criteria: TrainingCriteriaAbi,
+}
+
+#[xabi::data]
+#[derive(Debug, Clone)]
+pub struct TrainingPlanAbi {
+    pub criteria: TrainingCriteriaAbi,
+}
+
+#[xabi::data]
+#[derive(Debug, Clone)]
+pub struct CreatedIndexAbi {
+    pub details_type_url: String,
+    pub details: Vec<u8>,
+    pub index_version: u32,
+}
+
+#[xabi::data]
+#[derive(Debug, Clone)]
+pub struct QueryPlanAbi {
+    pub accepted: u8,
+    pub query: Vec<u8>,
+    pub needs_recheck: u8,
+}
+
+#[xabi::data]
+#[derive(Debug, Clone)]
+pub struct SearchOutputAbi {
+    pub exactness: u32,
+    pub row_ids_le: Vec<u8>,
+    pub null_row_ids_le: Vec<u8>,
+}
+
+#[xabi::data]
+#[derive(Clone)]
+pub struct TrainInputAbi {
+    pub params_json: String,
+    pub fragment_ids_le: Vec<u8>,
+    pub data: BorrowedLanceTrainingDataAbi,
+    pub store: BorrowedLanceIndexStoreAbi,
+    pub progress: BorrowedLanceIndexBuildProgressAbi,
+}
+
+unsafe impl Send for TrainInputAbi {}
+unsafe impl Sync for TrainInputAbi {}
+
+#[xabi::data]
+#[derive(Debug, Clone)]
+pub struct SubstraitQueryAbi {
+    pub version: String,
+    pub schema_ipc: Vec<u8>,
+    pub expr_bytes: Vec<u8>,
+    pub host_query_json: String,
+}
+
+#[xabi::data]
+#[derive(Clone)]
+pub struct RemapInputAbi {
+    pub mappings_le: Vec<u8>,
+    pub mapped_row_ids_le: Vec<u8>,
+    pub has_mapped_row_id: Vec<u8>,
+    pub store: BorrowedLanceIndexStoreAbi,
+}
+
+unsafe impl Send for RemapInputAbi {}
+unsafe impl Sync for RemapInputAbi {}
+
+#[xabi::data]
+#[derive(Clone)]
+pub struct UpdateInputAbi {
+    pub new_data: BorrowedLanceTrainingDataAbi,
+    pub store: BorrowedLanceIndexStoreAbi,
+}
+
+unsafe impl Send for UpdateInputAbi {}
+unsafe impl Sync for UpdateInputAbi {}
+
+#[xabi::xabi(id = TRAINING_DATA_TRAIT_ID, version = ABI_VERSION)]
+pub trait LanceTrainingDataAbi {
+    async fn next_batch_ipc(&self) -> xabi::Result<Option<Vec<u8>>>;
+}
+
+#[xabi::xabi(id = INDEX_STORE_TRAIT_ID, version = ABI_VERSION)]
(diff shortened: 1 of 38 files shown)


rust/lance-index-plugin-abi/xabi/snapshots/lance.index-build-progress/aarch64-apple-darwin.txt
```diff
@@ -0,0 +1,136 @@
+format=xabi-contract-snapshot-v1
```
westonpace, line 1: What are these snapshots?  Why are they needed?

rust/lance-index-plugin-abi/src/lib.rs
```diff
@@ -0,0 +1,289 @@
...
+pub const UPDATE_REQUIRES_OLD_DATA: u32 = 0;
+pub const UPDATE_ONLY_NEW_DATA: u32 = 1;
+
+#[xabi::data]
+#[derive(Debug, Clone, Copy)]
+pub struct TrainingCriteriaAbi {
+    pub ordering: u32,
+    pub needs_row_ids: u8,
+    pub needs_row_addrs: u8,
+}
```
westonpace, line 41: I don't think training criteria should be part of the ABI.  I think there is little value in trying to centralize too much of the write path of indexing.  We should start by focusing entirely on the read path.  However, we could add some minimal write methods...

Index creation could receive the JSON parameters object from the user, the dataset URI, and a list of fragment IDs.  The return value would be the index metadata (if we want the main library to do the commit) or void (if we want the plugin to own the commit).

Index updates would just receive the dataset URI and a list of fragment IDs.

Stuff like training criteria, update criteria, etc. could go in a helper crate for encapsulating common index helper routines.  It would be something that indexes could import but wouldn't be part of the ABI itself.

rust/lance-index-plugin-abi/src/lib.rs
```diff
@@ -0,0 +1,289 @@
...
+    pub query: Vec<u8>,
+    pub needs_recheck: u8,
+}
+
+#[xabi::data]
+#[derive(Debug, Clone)]
+pub struct SearchOutputAbi {
+    pub exactness: u32,
+    pub row_ids_le: Vec<u8>,
+    pub null_row_ids_le: Vec<u8>,
```
westonpace, line 77: What is le here?

rust/lance-index-plugin-abi/src/lib.rs
```diff
@@ -0,0 +1,289 @@
...
+pub struct QueryPlanAbi {
+    pub accepted: u8,
+    pub query: Vec<u8>,
+    pub needs_recheck: u8,
+}
+
+#[xabi::data]
+#[derive(Debug, Clone)]
+pub struct SearchOutputAbi {
+    pub exactness: u32,
```
westonpace, line 75: What is exactness?

rust/lance-index-plugin-abi/src/lib.rs
```diff
@@ -0,0 +1,289 @@
...
+pub trait LanceIndexBuildProgressAbi {
+    async fn stage_start(&self, name: &str, total: u64, unit: &str) -> xabi::Result<()>;
+
+    async fn stage_progress(&self, name: &str, amount: u64) -> xabi::Result<()>;
+
+    async fn stage_complete(&self, name: &str) -> xabi::Result<()>;
+}
+
+#[xabi::xabi(id = SCALAR_INDEX_TRAIT_ID, version = ABI_VERSION)]
+pub trait LanceScalarIndexAbi {
```
westonpace, line 146: These two (LanceScalarIndexAbi and LanceScalarIndexPluginAbi) are the most important to start with.  We should fully document these methods and the contract that is expected.

rust/lance-index-plugin-abi/src/lib.rs
```diff
@@ -0,0 +1,289 @@
...
+    async fn prewarm(&self) -> xabi::Result<()>;
+
+    fn statistics(&self) -> xabi::Result<String>;
```
westonpace, line 149: Is this JSON?

rust/lance-index-plugin-abi/src/lib.rs
```diff
@@ -0,0 +1,289 @@
...
+
+    async fn calculate_included_frags(&self) -> xabi::Result<Vec<u8>>;
```
westonpace, line 151: In an ideal world we could leave this out?  I thought calculate_included_frags was only used on certain migration paths?

rust/lance-index-plugin-abi/src/lib.rs
```diff
@@ -0,0 +1,289 @@
...
+
+    async fn search(&self, query: &[u8]) -> xabi::Result<SearchOutputAbi>;
```
westonpace, line 153: How does xabi handle async?

westonpace, line 153: What is query here?

rust/lance-index-plugin-abi/src/lib.rs
```diff
@@ -0,0 +1,289 @@
...
+
+    fn can_remap(&self) -> bool;
+
+    async fn remap(&self, input: RemapInputAbi) -> xabi::Result<CreatedIndexAbi>;
+
+    async fn update(&self, input: UpdateInputAbi) -> xabi::Result<CreatedIndexAbi>;
+
+    fn update_criteria(&self) -> xabi::Result<UpdateCriteriaAbi>;
+
+    fn derive_index_params_json(&self) -> xabi::Result<String>;
```
westonpace, line 163: What is this used for today?

rust/lance-index-plugin-abi/src/lib.rs
```diff
@@ -0,0 +1,289 @@
...
+
+    fn provides_exact_answer(&self) -> bool;
+
+    fn new_training_plan(
+        &self,
+        params_json: &str,
+        field_json: &str,
+    ) -> xabi::Result<TrainingPlanAbi>;
+
+    async fn train_index(&self, input: TrainInputAbi) -> xabi::Result<CreatedIndexAbi>;
```
westonpace, line 182: As mentioned above, I think we can avoid this "training input" dance.  Instead of going lance->plugin->lance->plugin we can just go lance->plugin and let the creation of the training plan be something indexes can use a helper crate for if they want.

rust/lance-index-plugin-abi/src/lib.rs
```diff
@@ -0,0 +1,289 @@
...
+
+    async fn load_index(
+        &self,
+        details: &[u8],
+        store: BorrowedLanceIndexStoreAbi,
```
westonpace, line 187: Where is `BorrowedLanceIndexStoreAbi` defined?  I wonder if we can avoid this.  Perhaps not though 🤔

westonpace (commented):
I have a lot of questions, I think we might want some deliberation to get this right.  Thanks for starting this!

rust/lance-index-plugin-abi/src/lib.rs
wjones127: If it is JSON, I'd prefer it to be `fn statistics_json(&self) -> ...`. I partly say that because I hope we could eventually deprecate the JSON format and provide something better.

rust/lance-index-plugin-abi/src/lib.rs
wjones127, line 1: thought: I wonder if a better process might be to split into two stages:

1. Create a Rust trait that defines the interface for indexes
2. Create an ABI for implementing that trait

I think it makes sense to separate them, as they each are difficult problems that we need to sort through. But I think we'll be able to do a better job if we tackle them one at a time.

rust/lance-index-plugin-abi/src/lib.rs
wjones127, line 153: suggestion: should the index plugin itself handle search? I think that becomes hard. One thing I was thinking was that it would provide a DataFusion optimizer rule:

```rust
fn optimizer(&self) -> Arc<dyn PhysicalOptimizerRule>;
```

The downside is this might imply each plugin needs to statically link most of DataFusion, making them a bit big. But I think DuckDB has this same problem anyways.

The upside is that it means the plugin gets to define which queries it operates on, rather than the caller. I think this is important for many indexes.

rust/lance-index-plugin-abi/src/lib.rs
wjones127: Here's my first pass at a trait:

```rust
pub trait Index {
    fn as_any(&self) -> &dyn Any;
    /// A static list of supported data types for the index.
    fn supported_types() -> &'static [&'static DataType];
    /// Pass in the data to train the index. Must contain a column named "value" of the same type as the supported types.
    /// Must also contain a `_rowid` column of type `UInt64` that contains the row ids of the original data.
    async fn train(data: Box<dyn RecordBatchStream>) -> Arc<dyn IndexSegment>;
    /// Return the optimizer rule for this index. This will be used to rewrite queries to use the index.
    fn optimizer_rule(&self) -> Arc<dyn PhysicalOptimizerRule>;
    /// Return a table provider that can be used to scan the index.
    /// This might be used to train a new version of the index
    fn scan(&self) -> Arc<dyn TableProvider>;
}
```

There's some complexities I haven't handled that we want to handle:

1. Distributed training
2. A distinction between a segment of an index and a logical index.

wjones127 (commented):
Adding some thoughts. It might be good to have a sync meeting on this so we can discuss different approaches. Then out of that make a design doc recording the architecture decisions we make here.

PR #7204, opened 2026-06-10
feat(index): expose per-query I/O metrics on ANN operators

### LINKED ISSUE #7201: [Feature Request] Expose I/O metrics (bytes_read / iops) for vector-search index operators
## Summary

The vector-search execution operators `ANNSubIndex` (`ANNIvfSubIndexExec`) and `ANNIvfPartition` (`ANNIvfPartitionExec`) currently report only index metrics (`indices_loaded`, `parts_loaded`, `index_comparisons`) plus a few counters/timers. Unlike the data-path operators (`LanceScan`, `Take`, `FilteredRead`), they do not register `IoMetrics`, so the bytes and IOPS read from the vector index files are invisible in `EXPLAIN ANALYZE` / scan statistics. This issue proposes registering per-query I/O metrics (`bytes_read`, `iops`, `requests`) on these operators so index-file I/O becomes observable.

## Motivation

When a vector query is slow, the current plan metrics make it hard to tell whether the cost is I/O (reading IVF partitions and quantization storage from object storage) or compute (distance calculation, refine). The data-scan side of a query already exposes `bytes_read` / `iops`, but the index side reports nothing, so the index contributes 0 to the query's total I/O even in cases where it dominates.

Exposing these metrics would let users:

- Distinguish I/O-bound from compute-bound vector queries directly from `EXPLAIN ANALYZE`, instead of guessing.
- See cache effectiveness at a glance: a warm query that serves every partition from the index cache reads ~0 bytes, while a cold query shows the real volume pulled from storage. The gap is an immediate signal of whether the index cache is sized correctly for the workload.
- Reason about the storage cost of a query on object stores (S3/GCS/Azure), where byte and request counts map directly to latency and dollar cost.
- Complement the existing `parts_loaded` counter (how many partitions were loaded) with the actual byte and request volume behind those loads.

## Why this is not a trivial drop-in

The data-path operators own a fresh `ScanScheduler` per `execute()` and can simply read its cumulative `stats()`. The ANN operators cannot: they delegate reads to a cached, shared `IVFIndex` whose `ScanScheduler` is created once when the index is opened and then reused across queries. Reading that scheduler's cumulative stats would report lifetime totals rather than per-query I/O, and because the two ANN operators share one scheduler, naive attribution would be wrong.

The per-query carrier that already exists is the `MetricsCollector` threaded into the index search calls (a fresh one is created per `execute()` per operator). The natural design is to attribute the actual index reads to that per-query collector, so the numbers are exact, correctly scoped, and read ~0 on a warm cache.

## Proposed direction (sketch)

- Thread a lightweight I/O counter through the index read path so the bytes and IOPS read for a single query are attributed to that query's metrics collector. The counter is attached only while a partition is actually being loaded from storage (a cache miss); on a cache hit nothing is read, so nothing is counted. This is an internal implementation detail, not a user-facing toggle.
- Reuse the existing (cached) file reader without re-opening files, and do not disturb the scheduler's global I/O totals.
- Publish the per-query totals into the operators' `iops` / `requests` / `bytes_read` metrics.

## How users see it (no new option)

There is no new configuration flag, parameter, or API to enable this. The metrics are recorded automatically during execution, exactly like the existing `parts_loaded` and `index_comparisons` metrics, and they appear wherever plan metrics are already surfaced: `EXPLAIN ANALYZE` output, the scan-statistics callback, and the Python `analyze_plan` output. A query that never inspects plan metrics pays only a negligible, always-on recording cost, consistent with how the data-path operators already record `bytes_read` / `iops`.

## Scope

- In scope: IVF-based vector indexes (`IVF_PQ` / `IVF_SQ` / `IVF_FLAT` / `IVF_HNSW_*`) on the current (v3) reader path.

## Implementation

I plan to implement this myself and will open a PR for it. Filing this issue first to confirm the approach (and the "no new option / always-on" decision) with maintainers before sending the change.

wombatu-kun:
## Summary

Closes #7201. The vector-search operators `ANNSubIndex` and `ANNIvfPartition` only reported index metrics and registered no `IoMetrics`, so index-file I/O was invisible in `EXPLAIN ANALYZE`. This makes per-query `bytes_read` / `iops` / `requests` observable on both operators.

## Design

The ANN operators delegate reads to a cached, shared `IVFIndex` whose `ScanScheduler` is reused across queries, so the established "own a fresh scheduler and read its cumulative stats" pattern does not apply. A lightweight per-query sink (`IoStats`, backed by a new `IoStatsRecorder` trait in `lance-core`) rides on the already per-query `MetricsCollector` and is attached to the index / quantization-storage readers only on an actual cache-miss load, tapping the authoritative `FileScheduler` accounting point. Numbers are exact, correctly attributed, and ~0 on a warm cache. The one-time index-open I/O (file footers, IVF centroids, quantization metadata) is captured from the fresh `IVFIndex::try_new` scheduler and attributed to `ANNIvfPartition`.

Always on, consistent with the data-path operators; surfaces through the existing scan-statistics / EXPLAIN ANALYZE path with no DataFusion-layer change.

## Observability (EXPLAIN ANALYZE)

Cold query (empty index cache):
- `ANNSubIndex: ... parts_loaded=10, bytes_read=57.02 K, iops=40, requests=40`
- `ANNIvfPartition: ... bytes_read=8.11 K, iops=4, requests=4` (index-open I/O)

Warm query (partitions cached): both operators report `bytes_read=0, iops=0, requests=0`.

## Overhead

Adding metrics is observability, not a speedup. Measured via a same-binary A/B (feature on vs off, which isolates runtime cost from code-layout noise): `Ivf_PQ_NoCache` change `[-0.57%, -0.08%, +0.43%]`, p=0.76, "No change in performance detected".

## Tests

- `lance-io`: the per-query sink records identically to the scheduler and stays isolated from sibling file handles.
- `lance`: cold-vs-warm aggregate counts (`ExecutionSummaryCounts`); and an end-to-end `EXPLAIN ANALYZE` assertion that the metrics are non-zero (cold) and zero (warm) on both ANN operators.

### DIFF
--- a/rust/lance-core/src/utils/io_stats.rs
+++ b/rust/lance-core/src/utils/io_stats.rs
@@ -10,6 +10,18 @@ use std::ops::Range;
 /// another.  It lets a caller attach a lightweight counter to a file reader and
 /// measure the exact bytes/IOPS performed for a bounded scope (e.g. a single
 /// query); see `lance_io::scheduler::IoStats` for the concrete implementation.
+///
+/// # When to use this
+///
+/// Lance also exposes two *process-wide, cumulative* I/O accounting facilities:
+/// the global scheduler counters (`lance_io::scheduler::iops_counter` /
+/// `bytes_read_counter`) and the object-store `IOTracker` wrapper used in tests.
+/// Both aggregate every read in the process and cannot attribute I/O to a single
+/// bounded scope.  Prefer an `IoStatsRecorder` when you need the *exact* I/O of
+/// one operation (e.g. a single query): attach it to a reader with
+/// `with_io_stats`, then read the snapshot when the scope ends.  It re-uses the
+/// reader's cached metadata, so measuring costs no extra file opens and does not
+/// disturb the global counters.
 pub trait IoStatsRecorder: std::fmt::Debug + Send + Sync {
     /// Record one completed request, given the byte ranges as actually
     /// submitted to storage (i.e. after any coalescing/splitting), so the
--- a/rust/lance-core/src/utils/io_stats.rs
+++ b/rust/lance-core/src/utils/io_stats.rs
@@ -1,0 +1,18 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright The Lance Authors
+
+use std::ops::Range;
+
+/// A sink that records I/O requests as they are submitted to storage.
+///
+/// This lives in `lance-core` so that the encoding layer (`lance-encoding`) and
+/// the I/O layer (`lance-io`) can both refer to it without depending on one
+/// another.  It lets a caller attach a lightweight counter to a file reader and
+/// measure the exact bytes/IOPS performed for a bounded scope (e.g. a single
+/// query); see `lance_io::scheduler::IoStats` for the concrete implementation.
+pub trait IoStatsRecorder: std::fmt::Debug + Send + Sync {
+    /// Record one completed request, given the byte ranges as actually
+    /// submitted to storage (i.e. after any coalescing/splitting), so the
+    /// counts reflect physical I/O.
+    fn record_request(&self, ranges: &[Range<u64>]);
+}
--- a/rust/lance-index/src/vector/storage.rs
+++ b/rust/lance-index/src/vector/storage.rs
@@ -622,10 +622,6 @@ impl<Q: Quantization> IvfQuantizationStorage<Q> {
         self.ivf.num_partitions()
     }
 
-    pub async fn load_partition(&self, part_id: usize) -> Result<Q::Storage> {
-        self.load_partition_with_io_stats(part_id, None).await
-    }
-
     /// Load a partition's quantization storage, optionally measuring the exact
     /// I/O it performs into `io_stats`.
     ///
@@ -633,7 +629,7 @@ impl<Q: Quantization> IvfQuantizationStorage<Q> {
     /// scheduler also records into the sink (a cheap clone that shares all
     /// cached metadata, so no file is re-opened).  When `None`, the normal
     /// uninstrumented reader is used.
-    pub async fn load_partition_with_io_stats(
+    pub async fn load_partition(
         &self,
         part_id: usize,
         io_stats: Option<IoStats>,
--- a/rust/lance-index/src/vector/storage.rs
+++ b/rust/lance-index/src/vector/storage.rs
@@ -14,10 +14,12 @@ use lance_core::{Error, ROW_ID, Result};
 use lance_encoding::decoder::FilterExpression;
 use lance_file::reader::FileReader;
 use lance_io::ReadBatchParams;
+use lance_io::scheduler::IoStats;
 use lance_linalg::distance::DistanceType;
 use prost::Message;
 use std::{
     any::Any,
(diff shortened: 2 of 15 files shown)


rust/lance-core/src/utils/io_stats.rs
```diff
@@ -0,0 +1,18 @@
...
+use std::ops::Range;
+
+/// A sink that records I/O requests as they are submitted to storage.
+///
+/// This lives in `lance-core` so that the encoding layer (`lance-encoding`) and
+/// the I/O layer (`lance-io`) can both refer to it without depending on one
+/// another.  It lets a caller attach a lightweight counter to a file reader and
+/// measure the exact bytes/IOPS performed for a bounded scope (e.g. a single
+/// query); see `lance_io::scheduler::IoStats` for the concrete implementation.
+pub trait IoStatsRecorder: std::fmt::Debug + Send + Sync {
```
wjones127, line 25: question: why this any not any of the other IO recording functionality? Maybe we should at least mention in the docs when to use this over the IO stats on an object store.

rust/lance-index/src/vector/storage.rs
```diff
@@ -621,14 +623,32 @@ impl<Q: Quantization> IvfQuantizationStorage<Q> {
     }
 
     pub async fn load_partition(&self, part_id: usize) -> Result<Q::Storage> {
+        self.load_partition_with_io_stats(part_id, None).await
+    }
```
wjones127: issue(non-blocking): we expand the API a bit here. For internal packages (not lance) we don't care as much about API stability. So I think it would be fine if you just added `io_stats` as a parameter to this function directly.

wjones127 (approved):
I like what this PR is solving. The solution seems acceptable for now, but I'd like see a future refactor that moves this tracking to the object store level and re-uses IOTrackingStore.

https://github.com/lance-format/lance/blob/66ea533d4d87567afabde1dd37cab670e4e2ac3b/rust/lance-io/src/utils/tracking_store.rs#L242

Perhaps `IOTrackingStore` could hold a `Vec<Arc<IOTracker>>` and `ObjectStore` could have some method like: 

```rust
impl ObjectStore {
    fn with_io_tracker(tracker: Arc<IOTracker>) -> Self { ... }
}
```

So you can always wrap that.

I'll merge this in a few days to give you time to respond to the optional feedback.

rust/lance-core/src/utils/io_stats.rs
wombatu-kun: Done 6e2fa0c9a. The scheduler counters (`iops_counter`/`bytes_read_counter`) and the object-store `IOTracker` are process-wide cumulative totals and cannot attribute I/O to a single query, which is exactly what `IoStatsRecorder` provides (exact per-scope bytes/IOPS, no file re-open) - added a "when to use this" note to the trait doc.

rust/lance-index/src/vector/storage.rs
wombatu-kun: Done 6e2fa0c9a

wombatu-kun:
Tracked the object-store-level refactor (reuse `IOTrackingStore` / `IOTracker`) as a follow-up: #7224. This PR keeps the interim scheduler-level `IoStatsRecorder`.

PR #7205, opened 2026-06-10
perf(vector)!: add dedicated SIMD kernels for RaBitQ ex-code reranking

BubbleCal:
## Problem

Multi-bit RaBitQ (`num_bits >= 2`) reranking computes `sum_d query[d] * ex_code[d]` per candidate.
Before this change it had two slow paths:

- The top-k heap pruning path did scalar bit extraction per dimension plus a gather from a
  `dim * 2^ex_bits` f32 table (1.5 MiB at `ex_bits=8`, `dim=1536` — cache hostile, and the table
  was rebuilt per query).
- FastScan only covered `ex_bits` in {2, 4, 8} with extra u8 quantization error; `ex_bits` in
  {1, 3, 5, 6, 7} were fully scalar.

## Change

Two parts, following the kernel design of the RaBitQ reference library
([VectorDB-NTU/RaBitQ-Library](https://github.com/VectorDB-NTU/RaBitQ-Library), Apache-2.0):

**1. Dedicated SIMD inner-product kernels** (`vector/bq/ex_dot.rs`): unpack packed ex codes with
shifts/masks and FMA them against the f32 query directly — no LUT, no quantization error. All
`ex_bits` 1..=8, with scalar, AVX2, AVX-512, and NEON variants (runtime dispatch, resolved once
per dist calculator). Used by the heap-pruning rerank, `distance()`, and the bulk tail; the
quantized ex dist table is now only built for the FastScan widths.

**2. A new on-disk ex-code layout** (`__blocked_ex_codes` column): dims are grouped into 64-dim
blocks (last block zero-padded) and bit-interleaved within each block so the SIMD unpack emits
codes in natural dim order (3 bits = 2-bit plane + top-bit plane, 5 = 4+1, 6 = 6|2, 7 = 6|2+1).
Because unpack order is natural, the kernels read the rotated query **as stored** — there is no
per-query query permutation, no load-time repack, and no extra resident copy. An earlier
iteration of this PR kept the old format and paid for it on every query and partition load;
benchmarks below quantify the difference.

### Compatibility

- **Old indexes** (sequential `__ex_codes`) remain readable: rows are repacked once at load and
  the in-memory batch is normalized to the new column, so every rewrite path (remap, optimize
  merges, distributed shard merges — including mixed-version shards) emits the blocked format
  and legacy indexes upgrade on their next rewrite.
- **Old lance versions cannot read new multi-bit indexes**: they fail loudly with a
  missing-column error (`requires __ex_codes column`) instead of misinterpreting bytes.
- **`num_bits=1` indexes carry no ex codes and are unaffected in both directions.**
- Binary codes, factor columns, and index metadata are unchanged.

## Benchmarks

GCP `c3-standard-8` (Sapphire Rapids, AVX-512), 1024 rows per measurement.

Per-candidate rerank kernel vs the previous table-gather, production embedding dims:

| ex_bits | dim | kernel | table-gather | speedup |
|---|---|---|---|---|
| 3 | 1536 | 172.7 µs | 2.464 ms | 14.3x |
| 5 | 1536 | 169.9 µs | 2.534 ms | 14.9x |
| 3 | 2048 | 235.3 µs | 3.264 ms | 13.9x |
| 5 | 2048 | 232.7 µs | 3.377 ms | 14.5x |

All widths at dim=1536: 87.0–181.7 µs (ex_bits 8 fastest, 7 slowest); at dim=2048:
117.0–243.5 µs. Apple M-series NEON shows 9–14x over the gather at dim=1024.

Storage load path (`try_from_batch`, `num_bits=4`, dim=1536, 8192 rows):

| format | load time |
|---|---|
| blocked (new) | **572 ns** (width validation only, column aliased) |
| legacy sequential (repacked) | 21.1 ms |

This load-time repack and the per-query permutation were measured by an end-to-end workload as
+9.65% avg latency (normal mode, `num_bits=4`, where only 0.138% of candidates reach the exact
rerank); with the blocked format the query and load paths do strictly less work than before this
PR.

The FastScan ex artifacts are also built only when they can be consumed: the bulk LUT branch is
reachable only for legacy indexes without error factors (gated indexes rerank per candidate), so
fresh indexes skip the FastScan transpose (one resident copy of the ex codes plus a per-load
transpose) and no path materializes the `dim * 2^ex_bits` f32 table anymore — the u8 LUT is
computed directly from the rotated query, which also speeds the legacy bypass itself by 3-6%.
The LUT bulk path is kept for legacy indexes because it scores 2.5x (`ex_bits=4`) to 7x
(`ex_bits=2`) faster per row than the kernel loop at dim=1536 (16384-row partitions); gated
indexes calling `distance_all` fall through to the exact per-row kernels.

## End-to-end benchmark (search-benchmark suite)

Fresh GCP `c4-standard-16`, IVF_RQ, `num_bits=5`, partition size 4096, `nprobes=24`, `k=10`,
single-threaded, gist (960d) and dbpedia (1536d), `main` @ e256207 vs this branch @ da7f2dd:

| dataset | approx_mode | main avg | branch avg | latency | QPS | recall main / branch |
|---|---|---|---|---|---|---|
| dbpedia | fast | 3.321 ms | 3.175 ms | -4.4% | +4.6% | 0.8145 / 0.8142 |
| dbpedia | normal | 3.911 ms | 3.802 ms | -2.8% | +2.5% | 0.9620 / 0.9611 |
| dbpedia | accurate | 4.765 ms | 4.508 ms | -5.4% | +5.7% | 0.9661 / 0.9657 |
| gist | fast | 2.347 ms | 2.259 ms | -3.8% | +4.0% | 0.6539 / 0.6577 |
| gist | normal | 2.981 ms | 2.939 ms | -1.4% | +1.4% | 0.9345 / 0.9370 |
| gist | accurate | 3.439 ms | 3.434 ms | -0.1% | +0.1% | 0.9422 / 0.9428 |

p99 improves across the board (dbpedia accurate: -8.4%); recall and index build time are
unchanged. Note this benchmark rebuilds the index per run with randomized rotation/kmeans, so
single-run readings vary by up to ~10% — a pinned-core (taskset) steady-state query loop on the
same machine and index gives a cleaner paired measurement: **266.0/266.6 QPS (main) vs
295.3/298.9 QPS (this branch), +12%, run-to-run spread under 0.3%** (dbpedia, normal mode).

A cpu-clock profile of that loop attributes the gain directly to this PR: the exact ex rerank
drops from 8.07% of query time (table gather) to 1.15% (ex-dot kernel, ~8x in vivo at an exact
rerank rate of 0.14%), and the per-partition-load ex FastScan transpose (2.88%, background
threads) disappears entirely. An earlier iteration of this change (kernels on top of the
unchanged sequential format) measured **+9.65%** average latency on this kind of workload from
the per-load repack and per-query permutation; the blocked format turns that into the gains
above.

## Testing

All run on GCP: Sapphire Rapids (AVX2 + AVX-512 paths) **and** Ampere Altra (NEON), plus an
AddressSanitizer pass over the bq suite on x86_64 (no findings).

- Differential kernel tests vs an f64 reference: every variant x `ex_bits` 1..=8 x dims
  {7, 16, 60, 64, 100, 128, 1024, 1536, 2048}, dense dim sweep (1..=160) for the bit-plane
  widths, pack/unpack round-trips, legacy-repack == writer-pack equivalence.
- Storage-level reference test across all widths and **both formats** at dims 72 and 1536.
- FastScan ex LUT test extended to `num_bits` {3, 5, 9} x both formats with a per-dim varying
  query, an unaligned dim (padded-tail LUT entries), and a quantization-derived error bound.
- `remap` round-trip across widths {4, 6, 8, 9} with remap == fresh-reload equivalence;
  end-to-end IVF_RQ build+search+recall at `num_bits` {4, 6, 9}; distributed merger tests.
- `cargo fmt --all` and `cargo clippy --all --tests --benches -- -D warnings` are clean.

## Breaking changes

On-disk: multi-bit (`num_bits >= 2`) IVF_RQ indexes written by this version use the new
`__blocked_ex_codes` column and are rejected (loud error) by older lance versions. Existing
indexes keep working and upgrade on rewrite. `num_bits=1` is fully compatible both ways.
Rust API: `RabitDistCalculator::new` and `RabitRawQueryContext` gained parameters/fields (same
pattern as #7179). Python/Java bindings are unaffected.

### DIFF
--- a/rust/lance-index/src/vector/bq/ex_dot.rs
+++ b/rust/lance-index/src/vector/bq/ex_dot.rs
@@ -1,0 +1,1047 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright The Lance Authors
+
+//! Inner-product kernels between an `f32` query and bit-packed RaBitQ ex-codes.
+//!
+//! Multi-bit RaBitQ reranking reduces to `sum_d query[d] * ex_code[d]`, where
+//! `ex_code[d]` is an unsigned `ex_bits`-wide integer. Materializing a
+//! `dim * 2^ex_bits` lookup table and gathering one entry per dimension is
+//! cache-hostile (the table is 1MiB for `ex_bits=8`, `dim=1024`); these kernels
+//! instead unpack the codes with shifts and masks and FMA them against the
+//! query directly, following the kernel design of the RaBitQ reference library
+//! (<https://github.com/VectorDB-NTU/RaBitQ-Library>, Apache-2.0).
+//!
+//! Two code layouts are consumed:
+//!
+//! - `ex_bits` ∈ {1, 2, 4, 8}: the sequential LSB-first layout written by the
+//!   index builder is already byte-aligned, so rows are used as stored.
+//! - `ex_bits` ∈ {3, 5, 6, 7}: codes straddle byte boundaries in the sequential
+//!   layout, so rows are repacked once at load time ([`plane_pack_row`]) into
+//!   bit-planes that unpack with byte-wise shifts:
+//!
+//!   ```text
+//!   per 64-dim group (T = ex_bits - 1, the top bit):
+//!   3 bits: [16B 2-bit plane][8B top-bit plane]
+//!   5 bits: [32B 4-bit plane][8B top-bit plane]
+//!   6 bits: [48B: 3 blocks of "6 low bits | 2 stolen bits"]
+//!   7 bits: [48B: same as 6 bits][8B top-bit plane]
+//!   ```
+//!
+//! Kernels unpack each group into runs of 16 code bytes whose dimension order
+//! differs from the natural order, so the query is permuted once per search
+//! with [`build_ex_query_into`] and the kernels then read both sides
+//! sequentially. The permuted query is zero-padded to a multiple of 64 so that
+//! padded lanes contribute nothing.
+
+use std::sync::LazyLock;
+
+/// Dimensions are processed in groups; the permuted query is padded to this
+/// multiple so every kernel sees whole groups.
+pub const EX_DOT_GROUP_DIMS: usize = 64;
+
+/// `f32` length of the kernel-order query built by [`build_ex_query_into`].
+pub fn ex_query_len(dim: usize) -> usize {
+    dim.next_multiple_of(EX_DOT_GROUP_DIMS)
+}
+
+/// Whether the sequential ex-code layout must be repacked into bit-planes for
+/// the dot kernels. For the remaining widths codes are byte-aligned already.
+pub fn needs_plane_repack(ex_bits: u8) -> bool {
+    matches!(ex_bits, 3 | 5 | 6 | 7)
+}
+
+/// Bytes per row of the code layout consumed by the dot kernels.
+pub fn ex_dot_code_bytes(dim: usize, ex_bits: u8) -> usize {
+    debug_assert!((1..=8).contains(&ex_bits));
+    if needs_plane_repack(ex_bits) {
+        ex_query_len(dim) * ex_bits as usize / 8
+    } else {
+        (dim * ex_bits as usize).div_ceil(u8::BITS as usize)
+    }
+}
+
+/// Dimensions per unpacking group for the given code width.
+fn group_dims(ex_bits: u8) -> usize {
+    match ex_bits {
+        1 | 4 | 8 => 16,
+        _ => EX_DOT_GROUP_DIMS,
+    }
+}
+
+fn group_bytes(ex_bits: u8) -> usize {
+    group_dims(ex_bits) * ex_bits as usize / 8
+}
+
+/// Extract the `ex_bits`-wide code of `dim_idx` from a sequentially bit-packed
+/// row (LSB-first, codes may straddle byte boundaries).
+#[inline]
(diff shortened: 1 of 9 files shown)


rust/lance-index/src/vector/bq/ex_dot.rs
```diff
@@ -0,0 +1,1047 @@
...
+fn select_ex_dot_kernel(ex_bits: u8) -> ExDotFn {
+    #[cfg(target_arch = "x86_64")]
+    {
+        if std::arch::is_x86_feature_detected!("avx512f") {
+            return x86::avx512_kernel(ex_bits);
+        }
+        if std::arch::is_x86_feature_detected!("avx2") && std::arch::is_x86_feature_detected!("fma")
+        {
+            return x86::avx2_kernel(ex_bits);
+        }
```
wjones127, line 307: praise: I love the dynamic dispatch 😍

wjones127 (approved):
This is marked as a breaking change. I didn't see an API change in the Lance crate? Am I missing something?

BubbleCal:
> This is marked as a breaking change. I didn't see an API change in the Lance crate? Am I missing something?

Oops, I have one more commit to change the `exbits` layout, but it won't impact existing `num_bits=1` IVF_RQ indices

PR #7206, opened 2026-06-10
test(compat): search maintenance-op sequences between two arbitrary refs

wkalt:
Adds a cross-version index compat capability: take two arbitrary refs (version, sha, branch, or tag), age an index under the first, exercise it under the second, and search maintenance-op sequences for panics or correctness divergence -- without hand-coding the failing sequence. This is the test that would have caught ENT-1662 (FTS `load_fst` stale `next_id` panic, fixed in #7200).

## What's here

- `venv_manager.py`: a venv per ref -- published releases install the wheel, anything else builds from a git worktree via maturin. Cached and lock-guarded so parallel workers build each ref once. Sub-venv Rust panics are surfaced as the failure.
- `compat_sequence.py`: breadth-first search over op sequences (write/delete/compact/optimize) up to a max length, for INVERTED + BTREE, BITMAP, LABEL_LIST, NGRAM, ZONEMAP, BLOOMFILTER. Non-vacuous oracle per kind (FTS count for INVERTED; index-vs-`use_scalar_index=False` for scalar). Setups built once and snapshotted for speed.
- `test_fts_sequence.py`: env-driven entry, parallel via xdist sharding. Failures print the plain-English sequence and real error. Default refs are the two most recent stable releases.
- `.github/workflows/compat-pair.yml`: on a pull request, ages indexes under the latest release of each of the two previous majors and exercises them under the PR head; `workflow_dispatch` runs one explicit ref pair on demand.

## Test

`pytest python/tests/compat/test_fts_sequence.py --run-compat -n 128` reproduces ENT-1662 and runs all scalar kinds green. JSON is the one index kind not yet covered by the sequence search.

### DIFF
--- a/.github/workflows/compat-pair.yml
+++ b/.github/workflows/compat-pair.yml
@@ -1,0 +1,158 @@
+# Cross-version index compatibility suite between two refs.
+#
+# Each ref (version, commit sha, branch, or tag) is provisioned by the compat framework:
+# a published release installs a wheel, anything else is built from a worktree via maturin.
+# The suite ages an index under the writer ref and exercises it under the reader, searching
+# maintenance-op sequences and asserting the reader neither panics nor disagrees with a
+# full scan.
+#
+# Triggers:
+#   - workflow_dispatch: run one explicit writer -> reader pair on demand.
+#   - pull_request: age indexes under the latest release of each of the two previous
+#     majors and exercise them under the PR head, so a change that breaks reading older
+#     indexes fails the PR.
+name: Compat (index ref pair)
+
+on:
+  workflow_dispatch:
+    inputs:
+      from_ref:
+        description: "Writer ref (version / sha / branch). Blank = 2nd most recent release."
+        required: false
+        default: ""
+      to_ref:
+        description: "Reader ref (version / sha / branch). Blank = most recent release."
+        required: false
+        default: ""
+      kinds:
+        description: "Comma-separated index kinds (INVERTED,BTREE,...) or 'all'."
+        required: false
+        default: "all"
+      max_length:
+        description: "Max maintenance-op sequence length to search (deeper = slower)."
+        required: false
+        default: "5"
+  pull_request:
+    # Only changes that can alter index format or its readers need the cross-version check.
+    paths:
+      - "rust/lance-index/**"
+      - "rust/lance/**"
+      - "python/**"
+      - ".github/workflows/compat-pair.yml"
+
+# Each run builds refs from source, so supersede an in-flight run for the same PR/ref
+# rather than letting expensive builds pile up on rapid pushes.
+concurrency:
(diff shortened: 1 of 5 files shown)


.github/workflows/compat-pair.yml
```diff
@@ -0,0 +1,158 @@
...
+      - uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065 # v5
+        with:
+          python-version: "3.11"
+      # Toolchain for the build-from-source provisioning path (refs without a wheel).
+      - uses: actions-rust-lang/setup-rust-toolchain@a0b538fa0b742a6aa35d6e2c169b4bd06d225a98 # v1
+      - uses: Swatinem/rust-cache@779680da715d629ac1d338a641029a2f4372abb5 # v2
+      - name: Install build deps
+        run: |
+          sudo apt update
+          sudo apt install -y protobuf-compiler libssl-dev
```
wjones127, line 57: suggestion: A lot of time goes into building the package. Perhaps we can save time by having it re-use the wheel from the Python Linux build?

https://github.com/lance-format/lance/blob/66ea533d4d87567afabde1dd37cab670e4e2ac3b/.github/workflows/python.yml#L126-L131

That would likely mean placing this workflow within the python.yml workflow.

python/python/tests/compat/compat_sequence.py
```diff
@@ -0,0 +1,300 @@
...
+            try:
+                to_venv.execute_method(scenario, "exercise_and_check", env)
+            except Exception as e:
+                error = str(e).strip()
+                failures.append({"run": i, "sequence": label, "error": error})
+                print(f"FAIL: {label}\n    {error}")
+                if stop_on_first:
+                    break
+            else:
+                print(f"ok  : {label}")
```
wjones127: suggestion: pytest already does a good job of reporting results. For example, it will always put the failures at the end, so it's easier to see. Maybe we should see if we can just use the pytest reporting and not add our own stdout unless needed. The failure output looks good. I think you just need to make the pytest parameterized labels be your `label` and then it would be clear. There's an `ids` parameter you can pass in to set this up.

.github/workflows/compat-pair.yml
wjones127, line 1: suggestion: The log output is currently black and white. pytest supports color output and you can enable it like so:

https://github.com/lance-format/lance/blob/66ea533d4d87567afabde1dd37cab670e4e2ac3b/.github/workflows/python.yml#L35-L37

wjones127 (approved):
I wonder to what extent this makes the existing compat tests redundant: https://github.com/lance-format/lance/blob/main/python/python/tests/compat/test_scalar_indices.py

I think the way I'm thinking right now: The existing tests check for versions and fewer ops, this checks more ops on fewer versions. So they complement each other. Happy to hear your thoughts on this.

Overall this looks like a good addition. I just have optional suggestions.

PR #7208, opened 2026-06-10
perf(fts): stream prewarm posting lists in chunks

wkalt:
Stream each inverted-index partition's posting lists in bounded token-row chunks (read -> build -> insert -> drop) so prewarm's peak resident set is ~one chunk rather than the whole invert.lance, instead of loading whole positional posting segments at once. The chunk target is a fixed 32 MiB.

ENT-1663

### DIFF
--- a/rust/lance-index/src/scalar.rs
+++ b/rust/lance-index/src/scalar.rs
@@ -256,6 +256,11 @@ pub trait IndexReader: Send + Sync {
     fn num_rows(&self) -> usize;
     /// Return the metadata of the file
     fn schema(&self) -> &lance_core::datatypes::Schema;
+    /// Best-effort on-disk byte size of the file when the reader already knows it
+    /// without extra I/O, else `None`. Used to size prewarm chunks.
+    fn file_size_bytes(&self) -> Option<u64> {
+        None
+    }
 }
 
 /// Trait abstracting I/O away from index logic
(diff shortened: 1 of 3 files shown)


rust/lance-index/src/scalar.rs
```diff
@@ -256,6 +256,11 @@ pub trait IndexReader: Send + Sync {
     fn num_rows(&self) -> usize;
     /// Return the metadata of the file
     fn schema(&self) -> &lance_core::datatypes::Schema;
+    /// Best-effort on-disk byte size of the file if known cheaply (reported by the
+    /// object store on open, no extra I/O), else `None`. Used to size prewarm chunks.
```
wjones127: nit: This is a big misleading. All index files have a known size reported by the manifest now, so when we open we actually pass that size down into the file reader. So we don't get the on-disk byte size from the object store on open, we get it from the manifest.

rust/lance-index/src/scalar/inverted/index.rs
```diff
@@ -5841,6 +6072,253 @@ mod tests {
         );
     }
 
+    /// Prewarming a large partition in multiple chunks must end up holding exactly the
+    /// same per-token posting lists (doc ids and frequencies) as the whole-file path.
+    /// Parametrized over layout: the legacy-v1 chunk path rebases global offsets to
+    /// chunk-local rows, which the v2 one-row-per-token path never exercises.
```
wjones127, line 6078: praise: I appreciate this test!

wjones127 (approved):

wkalt:
@wjones127 thanks, addressed

PR #7210, opened 2026-06-10
fix(python): python binding of `Compaction.commit`

### LINKED ISSUE #7209: `Compaction.commit` Python binding silently discards compaction options, making `defer_index_remap` impossible for distributed compaction
The Rust `commit_compaction` accepts `&CompactionOptions` (`rust/lance/src/dataset/optimize.rs`), but the Python binding constructs `CompactionOptions::default()` and ignores whatever options the plan was built with (`python/src/dataset/optimize.rs`, see the `// TODO: pass compaction option from plan and execute time` comment).

Consequence: the documented distributed compaction workflow (`Compaction.plan` on a coordinator, `CompactionTask.execute` on workers, `Compaction.commit` on the coordinator) can never use `defer_index_remap`. The commit always remaps every covering index inline, which makes the commit critical section grow with index size and prevents the frag-reuse deferral path entirely for distributed callers. Single-process `Compaction.execute` parses the full options dict, so the two code paths behave differently for the same option.

Fix: add an optional `options` dict parameter to the `Compaction.commit` binding, parsed with the same `parse_compaction_options` helper that `plan` and `execute` use, defaulting to current behavior when absent. Includes a parametrized test asserting that `commit(..., options={"defer_index_remap": True})` records a `__lance_frag_reuse` system index while the default path does not.

gstamatakis95:
Closes  #7209 

- `python/src/dataset/optimize.rs`: `Compaction.commit` gains an optional `options` dict
  parameter (`#[pyo3(signature = (dataset, rewrites, options = None))]`), parsed through the same
  `parse_compaction_options` helper that `plan` and `execute` already use, seeded from the
  dataset's `manifest.config` exactly like `execute`. Absent or `None` preserves the old behavior.
  The TODO comment is removed.
- `python/python/lance/lance/optimize.pyi`: stub updated to
  `commit(dataset, rewrites, options: Optional[CompactionOptions] = None)`.
- `python/python/tests/test_optimize.py`: new parametrized test
  `test_defer_index_remap_via_commit_options`. With
  `Compaction.commit(dataset, rewrites, options={"defer_index_remap": True})` a
  `__lance_frag_reuse` system index appears in `describe_indices()`. Without options it does not
  (inline remap occurs). Mirrors the adjacent `test_defer_index_remap` setup.


wjones127 (approved):
Nice work. Thanks for implementing this.

PR #7211, opened 2026-06-10
feat: create vector indices without training

### LINKED ISSUE #4034: Top-level indices part 2: empty vector indices
Follow up to #3940

We are trying to make it possible to keep indices even though we don't have enough data to actually train them. In an earlier PR, we are adding a `train: bool` parameter to let the user choose whether they want to defer training an index. If `train=False`, then we just create an empty index and it will show `num_indexed_rows: 0` in stats. They can later call `optimize_indices()` to do the actual training.

Vector indices are more complex because of IVF partitions, since we need a minimum amount of data to create them.

Here is the behavior I think we want:

1. Can pass `train=False` to create an empty index. No IVF or PQ will be trained. Will save metadata as empty array.
2. If you call `create_index(..., train=True)` or `optimize_indices()`:
   * < 256 non-null vectors -> same as `train=False`
   * 256 <= number of non-null vectors < `num_partitions * 256` => train on smaller number of partitions
   * number of non-null vectors >= `num_partitions * 256` -> train full index

burlacio:
## Summary

  Enable `create_index(..., train=False)` for vector indices,
  completing the empty-index story that already worked for
  scalar/FTS indices. 

Previously this path hit `todo!("create empty
   vector index when train=false")` and panicked.

  The work touches three things:

  - **`rust/lance/src/index/create.rs`** — removes the `todo!()`
  and lets the existing `build_empty_vector_index` helper handle
  the train=false branch.
  - **`rust/lance/src/index.rs`** — empty (definition-only) indices
   have no on-disk files, so the stats and vector-detection paths
  needed adjustments:
    - `collect_regular_indices_statistics` now short-circuits when
  `fragment_bitmap` is empty and returns `{}` instead of trying to
  open a non-existent index file.
    - `is_vector_index` detection prefers `index_details` metadata
  over file-based probing, since empty indices have no
  `INDEX_FILE_NAME` to inspect. The file-based path is kept as a
  fallback for older indices.

Fixes: https://github.com/lance-format/lance/issues/4034

### DIFF
--- a/python/python/tests/test_create_empty_index.py
+++ b/python/python/tests/test_create_empty_index.py
@@ -3,25 +3,43 @@
 
 """Tests for creating empty indices with train=False."""
 
+import pytest
+
 import lance
 import pyarrow as pa
 import pyarrow.compute as pc
 
 
-def test_create_empty_scalar_index():
+@pytest.mark.parametrize("index_type", ["BTREE", "BITMAP"])
+def test_create_empty_scalar_index(index_type):
     data = pa.table({"id": range(100)})
     dataset = lance.write_dataset(data, "memory://")
 
-    # Passing train=False to create an empty index
-    dataset.create_scalar_index("id", "BTREE", train=False)
+    dataset.create_scalar_index("id", index_type, train=False)
 
-    # Verify index exists and has correct stats
     indices = dataset.describe_indices()
     assert len(indices) == 1
-    assert indices[0].index_type == "BTree"
     stats = dataset.stats.index_stats(indices[0].name)
     assert stats["num_indexed_rows"] == 0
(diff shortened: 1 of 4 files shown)


wjones127 (changes requested):
Apologies for taking a while to review this. This is the right direction, but I think we need some more tests before we can call this done.

Creating the initial index is good, but we also need to make sure the index is preserved. So testing these situations:

* Add data, create index, delete all data, make sure index is still there.
* Add 1024 rows, create IVF_PQ index, delete 800 rows (so under 256-row threshold described in https://github.com/lance-format/lance/issues/4034#issuecomment-5195384697), call optimize indices
* Add data, create ind3ex, update all rows, make sure index is still there, run optimize indices

PR #7215, opened 2026-06-10
feat(mem-wal): snapshot-consistent as-of cut for fresh-tier membership

hamersaw:
## What

Adds `FreshTierWatermark { active_generation, active_batch_count }` and `LsmScanner::contains_pks_at` (with `fresh_tier_block_list` threading the watermark) so a caller can evaluate fresh-tier PK membership against the **exact tier a prior scan observed**, instead of the live tier.

## Why

The WAL block-list runs as two independent RPCs (the read arm and a supersession check) that each snapshot the live fresh tier at their own call time. Under concurrent writes the two snapshots disagree, so a base row can be dropped as "superseded" by the check while the arm never delivered a replacement — a transient missing row. The fix pins both phases to the same watermark; this PR is the lance half that lets the check reconstruct the arm's snapshot.

## How the watermark works

The active memtable is the only fresh-tier source that grows between two reads; everything strictly below its generation (frozen memtables, flushed generations) is immutable. So the as-of filter:

- includes in-memory/flushed sources **below** `active_generation` whole (immutable, fully observed),
- bounds the **active** generation to its first `active_batch_count` batches (by append index),
- excludes in-memory sources **above** `active_generation` and flushed generations `>= active_generation` (produced after the snapshot).

It uses only `batch_store.len()` and the memtable generation — both always available on the read path — and only ever *excludes* rows the scan did not observe, so a stale watermark under-counts (a tolerable stale read) rather than over-counts (which would drop a row with no replacement).

> Note: an earlier approach keyed the watermark on per-batch WAL positions (`wal_batch_mapping`), but that map is only populated by `mark_wal_flushed`, which is test/bench-only — empty in production. The generation + batch-count watermark avoids any write-path dependency.

## Grace-period pinning

A flushed memtable could otherwise be evicted between the two reads, collapsing its per-batch boundaries and turning the active-generation bound into a wholesale `>=` exclusion (a stale read). Frozen memtables are now retained in memory for a configurable grace period (`frozen_memtable_grace`, default 3s) after flush and swept by an existing dispatcher ticker. The grace must be **strictly larger than the maximum query elapsed time** to guarantee snapshot isolation: while pinned, the generation is served batch-resolved from memory; once evicted, no in-flight read references it so the `>=` exclusion is safe.

## Tests

`fresh_tier_block_list` as-of unit tests: active-memtable batch-count bound, newer-gen-excluded / lower-gen-included-whole, flushed-gen at/above active excluded; collector suppression of a flushed generation pinned in memory; frozen retained during grace then swept.


hamersaw:
This PR adds a `FreshWatermarkTier` so that we can provide snapshot isolation for queries to occur within a brief time window. For example, if we provide the same watermark value the same query will produce identical results regardless of any modifying operations that could occur between them. This works by tracking the flushed generation, so what is the latest flushed generation we saw, and the active batch count, so what are the latest records in the active memtable that we saw. It's important to note that this is only valid for a brief bounded time-window where we keep the active memtable in memory even when we have already flushed it's contents to l0 cache. This is expected to be singular seconds.

westonpace (approved):
Marking approve without 100% understanding since you probably know WAL better than I do and it's not serving high use yet.  Here are Claude's suggestions:

---

## PR #7215 Review: `feat(mem-wal): snapshot-consistent as-of cut for fresh-tier membership`

### Overview

This PR fixes a two-snapshot consistency bug in the WAL block-list check: the read arm and the supersession check each snapshot the live fresh tier at their own call time, so concurrent writes could cause a base row to be considered "superseded" while the arm never delivered a replacement. The fix adds `FreshTierWatermark` to pin the supersession check to the same tier state the read arm observed.

---

### Code Quality

**Strengths:**
- The watermark design is conservative (under-counts on stale, never over-counts), which is the right bias for correctness.
- Converting `TriggerMemTableFlush` from struct to enum is clean; dispatcher-ticker design fits the existing pattern.
- `flushed_gen_pinned_in_memory` suppression is correct: a grace-pinned generation must be served from memory (batch boundaries preserved) and the on-disk copy skipped to avoid double-scanning.
- Tests are specific and cover the boundary cases well.

**Issues / Suggestions:**

1. **Grace-period contract is implicit and fragile** (`write.rs:177–199`). The doc comment says the grace "MUST be strictly greater than the maximum elapsed time of any query," but this is unenforced. A misconfigured or short grace silently degrades to a stale read. Consider adding a validator in `ShardWriterConfig::with_frozen_memtable_grace` that panics or returns an error if `grace == Duration::ZERO`, plus a prominent operator note that production values should be order-of-magnitude above P99 query latency.

2. **`now_millis()` used but not shown** (`write.rs:2289, 2379`). The diff references `now_millis()` in `sweep_expired_frozen` and `flush_memtable`, but the implementation isn't in the diff. If it wraps `SystemTime::now()` that's fine, but if it calls `Instant::now()` (non-wall-clock) the millisecond comparison with `flushed_at_ms` is still internally consistent — just confirm there's no mixing of clocks.

3. **`active_batch_count: u64::MAX` used as a sentinel in tests** (`block_list.rs:648`). The `FreshTierWatermark::active_batch_count` field is `u64`, and `u64::MAX` is used to mean "no batch-count limit." This is fine for tests but could be a footgun in production. Consider an `Option<u64>` or a named constant (`UNBOUNDED_BATCH_COUNT`) to make intent explicit, especially since the unbounded case is needed for the flushed-gen exclusion test.

4. **`pk_hashes_from_batch_store_bounded` silently skips missing indices** (`block_list.rs:220–226`):
   ```rust
   for i in 0..end {
       if let Some(stored) = store.get(i) {
           batches.push(stored.data.clone());
       }
   }
   ```
   Skipping a `None` silently under-counts. If `BatchStore::get` can return `None` for valid indices (e.g., concurrent eviction), a missing batch means a PK is incorrectly absent from the block list — a correctness bug. If it cannot return `None` for in-bounds indices, the `if let Some` should be replaced with an assertion or `expect`.

5. **`flushed_after_snapshot` condition** (`block_list.rs:162–165`):
   ```rust
   let flushed_after_snapshot = watermarks
       .and_then(|m| m.get(shard_id))
       .is_some_and(|watermark| generation.as_u64() >= watermark.active_generation);
   ```
   The `>=` means a flushed generation *equal* to `active_generation` is excluded. The PR description says "at or above," which matches. But the active memtable arm uses `>` for "newer than active gen" and `==` for "bounded by batch count." A flushed generation *equal* to `active_generation` means the active memtable was flushed *between* the two reads — excluding it is correct since its data is already captured by the in-memory arm (pinned via grace). This logic is subtle and worth a short inline comment here to help future readers.

6. **Sweep lag** (`write.rs:2258`). The tick is `grace / 3`, so a generation can live up to `~grace * 4/3` post-flush. The doc comment notes this but it is also a concern for memory pressure: if flush throughput is high, multiple grace-pinned generations stack up. Consider documenting the worst-case extra memory as a multiple of `max_memtable_size * (grace / flush_interval)`.

7. **Test timing sensitivity** (`write.rs:4322`). The test sleeps 1500ms for a 1s grace:
   ```rust
   tokio::time::sleep(Duration::from_millis(1_500)).await;
   ```
   The sweep tick is `grace / 3 = ~333ms`, so worst-case the generation lives `~1333ms`. 1500ms is tight on a loaded CI box. Using `tokio::time::pause()` + `advance()` would make this deterministic without the wall-clock dependency.

---

### Minor

- `FlushedMemTable` in `LsmDataSource` gains `shard_id` and `generation` fields that were apparently already there from the diff context (`block_list.rs:152–158`). If these are new fields, confirm all construction sites are updated.
- The `active_batch_count` field is `u64` but `pk_hashes_from_batch_store_bounded` takes `usize` for `limit`. The cast `watermark.active_batch_count as usize` could truncate on 32-bit targets (unlikely in practice but worth a `try_from` or a note).

---

### Test Coverage

The three new `block_list` unit tests cover the key cases (batch-count bound, generation ordering, flushed exclusion). The `test_frozen_retained_during_grace_then_swept` integration test covers the lifecycle. The collector suppression test is clear and well-structured.

Missing: a test that combines a flushed-gen-pinned-in-memory source with a watermark check — verifying that the in-memory copy is used (not the flushed one) for as-of evaluation. This is the critical path for correctness during the grace window.

---

### Summary

The design is sound and the core logic appears correct. The main risks are:
1. The silent correctness issue in `pk_hashes_from_batch_store_bounded` if `BatchStore::get` can return `None` for valid indices.
2. The grace-period contract being easy to misconfigure without runtime enforcement.
3. The timing-sensitive test in CI.

PR #7217, opened 2026-06-10
fix(rust): fixed IVF_RQ indexes ignoring the frag-reuse index

### LINKED ISSUE #7216: fix(rust): IVF_RQ indexes ignore the frag-reuse index at read time breaking queries after compaction with defer_index_remap
# Issue
When a dataset is compacted with defer_index_remap=true, Lance does not rewrite the covering vector indexes. Instead it records a __lance_frag_reuse system index that holds the mapping from old row addresses  to new ones, and every covering index is expected to apply that mapping when it is loaded. The IVF_PQ storage does this correctly. In rust/lance-index/src/vector/pq/storage.rs, try_from_batch walks the stored row ids through FragReuseIndex::remap_row_id and drops or rewrites entries as needed.

The IVF_RQ storage does not. RabitQuantizationStorage::try_from_batch in rust/lance-index/src/vector/bq/storage.rs receives the frag-reuse index as a parameter named _fri and discards it. The index loads with its original pre-compaction row addresses, which point at fragments that no longer exist. Any ANN query that needs to fetch row content from such a version fails at the scanner take stage with an error like thiss:

Execution error: The input to a take operation specified fragment id 0 but this fragment does not exist in the dataset

The error is raised from rust/lance/src/dataset/scanner.rs:4973.

Furthermore, the IVF state cache reconstruction path, reconstruct_typed in rust/lance/src/index/vector/ivf/v2.rs, passes a hard-coded None for the frag-reuse index when it rebuilds storage from a cached state. This means that even if the storage layer is fixed, the remap is silently skipped whenever the index is reconstructed from the cache. 
  
Execution error: The input to a take operation specified fragment id 0 but this fragment does not exist in the dataset

  The error is raised from rust/lance/src/dataset/scanner.rs:4973.

There is a second defect that hides the first one from a quick fix. The IVF state cache reconstruction path, reconstruct_typed in rust/lance/src/index/vector/ivf/v2.rs, passes a hard-coded None for the frag-reuse index when it rebuilds storage from a cached state. This means that even if the storage layer is fixed, the remap is silently skipped whenever the index is reconstructed from the cache. The cache key already includes the frag-reuse uuid, so threading the actual index through is safe and keeps cached states with and without remap debt distinct.

A note on why existing tests did not catch this. The existing test_read_ivf_pq_index_v3_with_defer_index_remap queries with an empty projection, so the take stage where this failure occurs is never exercised. A regression test for this bug needs a non-empty projection.

Steps to reproduce:

  1. Create a dataset with several fragments and build an IVF_RQ vector index on it.
  2. Run compact_files with defer_index_remap=true so the fragments are rewritten but the index is not.
  3. Run a nearest neighbor query that projects at least one column.
  4. The query fails with the take error above.

# Proposed fix 
I have a patch ready that fixes both halves. It builds the per-partition row id mapping from FragReuseIndex::remap_row_id and applies the storage's existing pack-aware remap method, which already handles the packed SIMD code layout and the multi-bit split code and factor columns. It also threads the frag-reuse index through the state cache reconstruction path and adds an IVF_RQ regression test that queries with a non-empty projection. When no frag-reuse index is present, or no row in a partition is affected, the new path is a no-op.

gstamatakis95:
Closes #7216

  After compaction with defer_index_remap=true, covering vector indexes are expected to remap their stored row addresses through the __lance_frag_reuse index at load time. The IVF_PQ storage does this, but
  RabitQuantizationStorage::try_from_batch received the frag-reuse index as a parameter named _fri and discarded it. As a result an IVF_RQ index loaded from a debt-carrying version kept its pre-compaction row  addresses, and any ANN query that fetched row content failed at the scanner take stage with take operation specified fragment id N but this fragment does not exist.

The IVF state cache reconstruction path in reconstruct_typed passed a hard-coded None for the frag-reuse index, so even with the storage fixed the remap was skipped whenever the index was rebuilt from a cached state. The cache key already includes the frag-reuse uuid, so cached states with and without remap debt stay distinct when the index is threaded through.

This PR makes the following changes:

In rust/lance-index/src/vector/bq/storage.rs, try_from_batch now builds a row id mapping for the partition from FragReuseIndex::remap_row_id, mirroring the PQ derivation, and applies the storage's existing
  remap method. Reusing remap instead of porting the PQ inline rewrite is deliberate, because RQ codes live in a packed and permuted SIMD layout and remap already performs the layout-correct filter and repack,
  including the multi-bit split ex code and factor columns. It is already pinned by the existing identity-remap regression tests. When no frag-reuse index is present, when its row id maps are empty, or when
  no row in the partition is affected, the path stays a no-op. The generic remap_row_ids_record_batch helper used by the flat and SQ storages was considered and does not fit here, both because it assumes a two
  column batch and because a plain row-wise take is not safe on the packed code layout.

  In rust/lance/src/index/vector/ivf/v2.rs and rust/lance/src/index.rs, IvfStateEntry::reconstruct gains a frag-reuse parameter and the caller passes open_frag_reuse_index, which is already cached in the index
  cache, closing the warm cache hole without adding load overhead.

  In rust/lance/src/dataset/optimize.rs, a regression test test_read_ivf_rq_index_v3_with_defer_index_remap is added next to the PQ analog. It queries with a non-empty projection because the take stage where
  this bug manifests is only exercised when row content is fetched. The existing PQ test uses an empty projection and never reaches that stage, which is why this went undetected.

  Verification: with the storage remap reverted, the new test fails with the exact defect signature from scanner.rs. With the fix it passes. cargo test -p lance defer_index_remap passes (the new RQ test plus
  the existing PQ tests), cargo test -p lance-index bq passes, and cargo fmt and cargo clippy -- -D warnings are clean on the touched crates.


Katomoto (commented):
Logic looks correct. One suggestion: the early return on line prevents the cleanup function from running — worth adding a finally block.

gstamatakis95:
@Katomoto Thanks, unfortunately I don't understand exactly which part of my code you are talking about so if possible make a comment on that section or refference it in some capacity.

xuanyu-z:
LGTM

xuanyu-z (approved):

xuanyu-z:
Follow-up: #7315 builds on this PR — it fixes a related IVF_PQ / IVF_HNSW_PQ stale-row-address bug under deferred compaction (PQ storage kept its pre-remap `row_ids` field) and adds deferred-compaction correctness coverage (fragment-reuse window, materialized deletions, and physical remap + trim) across the vector index types. It's stacked on this PR, so it'd merge after. 

@jackye1995 @wjones127 @Xuanwo

jackye1995 (approved):
looks good to me!

PR #7219, opened 2026-06-10
fix: branch support on object stores, python branch ops, s3/ddb integration tests

brendanclement:
Adds the first integration coverage for table branches (localstack S3 + DynamoDB) and fixes the four bugs it caught. All four were invisible to local-FS tests.

- Branch locations broke uris that carry query strings. BranchLocation appended tree/<branch> after the query string, so s3+ddb://...?ddbTableName=x produced a branch uri that re-parses to the main table path. Branches now work end to end over s3+ddb://.

- create_branch dropped the dataset's store params. It passed store_params.unwrap_or_default() to CommitBuilder, so the commit's existence probe built an object store with no endpoint or region. It now falls back to the dataset's own params, matching shallow_clone.

- The managed manifest store treated a missing chain as a fatal error. Probing a branch location before the branch exists surfaced the namespace's not-found instead of "no versions". get_latest_version now reports a missing table or branch chain as None, matching the ExternalManifestStore contract and the store's own behavior for empty main chains.

- The Python namespace adapter flattened error identity. Exceptions from custom Python namespaces all arrived in Rust as opaque IO strings. The adapter now reconstructs the NamespaceError from the lance_namespace exception's error code, gated on the exception actually being a LanceNamespaceError so a foreign exception that happens to carry an integer code is never reinterpreted (it propagates as itself). Custom namespaces now raise the same typed exceptions as the native ones.

### DIFF
--- a/python/pyproject.toml
+++ b/python/pyproject.toml
@@ -1,7 +1,7 @@
 [project]
 name = "pylance"
 dynamic = ["version"]
-dependencies = ["pyarrow>=14", "numpy>=1.22", "lance-namespace>=0.8.2,<0.9"]
+dependencies = ["pyarrow>=14", "numpy>=1.22", "lance-namespace>=0.8.5,<0.9"]
 description = "python wrapper for Lance columnar format"
 authors = [{ name = "Lance Devs", email = "dev@lance.org" }]
 license = { file = "LICENSE" }
--- a/python/pyproject.toml
+++ b/python/pyproject.toml
@@ -1,7 +1,7 @@
 [project]
 name = "pylance"
 dynamic = ["version"]
-dependencies = ["pyarrow>=14", "numpy>=1.22", "lance-namespace>=0.8.0,<0.9"]
+dependencies = ["pyarrow>=14", "numpy>=1.22", "lance-namespace>=0.8.2,<0.9"]
 description = "python wrapper for Lance columnar format"
 authors = [{ name = "Lance Devs", email = "dev@lance.org" }]
 license = { file = "LICENSE" }
--- a/python/python/tests/test_s3_ddb.py
+++ b/python/python/tests/test_s3_ddb.py
@@ -213,6 +213,58 @@ def test_s3_ddb_concurrent_commit_more_than_five(
 
 
 @pytest.mark.integration
+def test_s3_ddb_branches(s3_bucket: str, ddb_table: str):
+    """Branches on a table committed through the DynamoDB external manifest
+    store.
+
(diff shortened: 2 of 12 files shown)


python/python/tests/test_s3_ddb.py
```diff
@@ -212,6 +212,58 @@ def writh_dataset_with_start_barrier():
...
+    ds = lance.write_dataset(
+        pa.Table.from_pylist([{"a": 2}]),
+        table_dir,
+        mode="append",
+        storage_options=storage_options,
+    )
+
+    # Fork "dev" at v2 and commit on it, then diverge main to the same
+    # version number.
+    dev = ds.create_branch("dev", 2)
```
jackye1995, line 242: One issue this branch workflow exposes is that `LanceDataset.create_branch` returns a wrapper that is not initialized like the other `LanceDataset` construction paths: it does not set `_namespace_client_managed_versioning`. A caller that uses the returned branch dataset directly, for example `ds.create_branch("dev").checkout_version(("main", None))`, hits `AttributeError` in `__copy__` before Rust checkout runs. You can copy `self._namespace_client_managed_versioning` onto the returned wrapper in `LanceDataset.create_branch` so this public branch API works outside the write-then-reopen path.

python/pyproject.toml
```diff
@@ -1,7 +1,7 @@
 [project]
 name = "pylance"
 dynamic = ["version"]
-dependencies = ["pyarrow>=14", "numpy>=1.22", "lance-namespace>=0.8.0,<0.9"]
+dependencies = ["pyarrow>=14", "numpy>=1.22", "lance-namespace>=0.8.2,<0.9"]
```
jackye1995: The new branch APIs can surface Rust namespace errors with branch error codes 22/23, but `lance-namespace==0.8.2` only maps codes through 21. Because Python error conversion goes through `lance_namespace.errors.from_error_code`, branch failures are currently reclassified as `InternalError`; for example, deleting a missing branch returns `InternalError` instead of a branch-not-found error. You can either require a `lance-namespace` release that includes the branch error codes

brendanclement: https://github.com/lance-format/lance-namespace/pull/352

python/python/tests/test_s3_ddb.py
brendanclement: added self._namespace_client_managed_versioning in the create_branch. Thank you

jackye1995 (approved):
thanks for the fixes!

PR #7221, opened 2026-06-10
feat(python): add cleanup_frag_reuse_index

### LINKED ISSUE #7220: No Python API to prune frag-reuse generations, so deferred-remap users accumulate read-path debt forever
Deferred-remap compaction (`defer_index_remap=true`) appends a new generation to the `__lance_frag_reuse` system index on every commit. Rust already has a complete and race-safe pruner,  `cleanup_frag_reuse_index` in `rust/lance/src/dataset/index/frag_reuse.rs`, which drops every generation that all indexes have caught up to. However, this function is not exposed in Python  and is never invoked automatically, so a Python pipeline that defers remaps accumulates generations without bound. Because the frag-reuse index is loaded on every vector index open, the open-time cost grows  with compaction history even after every index has been rebuilt and no generation is needed anymore.

  I propose exposing the existing function as `LanceDataset.cleanup_frag_reuse_index()` so it can be scheduled as a periodic maintenance call alongside `cleanup_old_versions`. A patch is available that follows  the `migrate_manifest_paths_v2` wrapper pattern and includes a test asserting that `__lance_frag_reuse` reports zero versions once all indexes have caught up and the cleanup runs.

gstamatakis95:
Closes #7220

Exposes the existing Rust frag-reuse pruner to Python as `dataset.optimize.cleanup_frag_reuse_index()`.

Deferred-remap compaction (`compact_files(defer_index_remap=True)`) appends a generation to the `__lance_frag_reuse` index on every commit. Without pruning they accumulate forever and are all applied on every index open. The race-safe pruner from #3836 already existed in Rust but had no Python binding.

It lands on `DatasetOptimizer`, next to `compact_files` and `optimize_indices`, rather than on `LanceDataset` as the issue proposed.

Making that callable from Python turned every panic in the frag-reuse read path into a process abort instead of an exception, so this also closes those:

- Corrupt or unloadable index details, and undecodable `changed_row_addrs` bytes, now return errors instead of `unwrap()`ing.
- Fragment digests are validated at the open boundary before transposition: non-empty old fragments, fragment id within 32 bits, `physical_rows` below `FRAGMENT_SIZE` and non-zero, `num_deleted_rows` no greater than `physical_rows`. Each of these otherwise hit an assert, a `u64` overflow, a non-terminating loop, or a capacity-overflow panic.
- External-details `offset`/`size` are range-checked, and the concurrent-rebase paths in the conflict resolver propagate load errors rather than panicking.
- Retention now mirrors the load path: a rewrite group straddling indexed and non-indexed fragments is retained, not discarded as corrupt. Dropping it would leave the affected indices unable to remap.

Tests cover the Python round trip (a generation is retained while an index is behind, and pruned once it catches up) and each corrupt-metadata case in Rust.


Xuanwo (commented):
1. The new Python cleanup entry point can panic when fragment-reuse metadata cannot be loaded. Corrupt or stale metadata can surface as a panic instead of a normal Python exception.
2. The current test only exercises the already-caught-up index case. A cleanup implementation that drops all generations would still pass, so the regression can reappear.

gstamatakis95:
> 1. The new Python cleanup entry point can panic when fragment-reuse metadata cannot be loaded. Corrupt or stale metadata can surface as a panic instead of a normal Python exception.
> 2. The current test only exercises the already-caught-up index case. A cleanup implementation that drops all generations would still pass, so the regression can reappear.

Thanks for the feedback @Xuanwo, addressed both issues.

gstamatakis95:
Also moved to `dataset.optimize` which IMO is a better fit for this.

rust/lance/src/dataset/index/frag_reuse.rs
```diff
@@ -38,27 +38,25 @@ pub async fn cleanup_frag_reuse_index(dataset: &mut Dataset) -> lance_core::Resu
...
-            // If the check fails, the reuse version is likely corrupted, do not retain it.
-            continue;
+    'versions: for version in frag_reuse_details.versions.iter() {
+        let mut all_caught_up = true;
+        for idx in indices.iter() {
+            match is_index_remap_caught_up(version, idx) {
+                Ok(true) => {}
+                Ok(false) => all_caught_up = false,
+                // An InvalidInput error means the reuse version is likely corrupted; drop it.
+                Err(Error::InvalidInput { .. }) => continue 'versions,
```
Xuanwo: The cleanup path treats a partially indexed rewrite group as corrupt and drops the reuse generation, while the load path treats the same straddling state as recoverable. Calling the new Python cleanup API on an older dataset can remove the generation that keeps affected indexes safe.

rust/lance/src/dataset/index/frag_reuse.rs
```diff
@@ -38,27 +38,25 @@ pub async fn cleanup_frag_reuse_index(dataset: &mut Dataset) -> lance_core::Resu
         return Ok(());
     };
 
-    let frag_reuse_details = load_frag_reuse_index_details(dataset, frag_reuse_index_meta)
-        .await
-        .unwrap();
+    // Surface corrupt/stale metadata as a normal error instead of panicking on unwrap.
+    let frag_reuse_details = load_frag_reuse_index_details(dataset, frag_reuse_index_meta).await?;
```
Xuanwo, line 61: The cleanup path still reads external fragment-reuse details with unchecked offset/size arithmetic. Malformed metadata can still panic or produce an invalid object-store range instead of returning a Python exception.

gstamatakis95: Fixed in load_frag_reuse_index_details. The read now converts offset and size with usize::try_from so overflow errors instead of truncating, computes the end with checked_add so it errors instead   of wrapping, and stats the file via reader.size() to reject any range past EOF before calling get_range.

rust/lance/src/dataset/index/frag_reuse.rs
gstamatakis95: Good catch, fixed.

gstamatakis95:
@Xuanwo I think this is ready.

rust/lance/src/dataset/index/frag_reuse.rs
```diff
@@ -57,29 +57,15 @@ pub async fn cleanup_frag_reuse_index(dataset: &mut Dataset) -> lance_core::Resu
...
```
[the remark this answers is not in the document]
gstamatakis95: Skipping these three in this PR, with reasoning.

They are pre-existing issues in the frag-reuse read path rather than regressions from this change. `open_frag_reuse_index` is already reached by every scalar and vector index open, by `remap_column_index`, and by the existing Python `remap_row_ids` binding, so this PR adds one more caller to an already reachable path rather than widening the surface.

The earlier rounds were panics and hangs, where a corrupt payload aborts the process instead of raising. Those are fixed and covered by tests. The three remaining cases are silent wrong answers and a work bound. They are real, but they cannot abort a caller, so they do not block the contract this PR is claiming.

Validating version order, membership, cardinality, uniqueness, and aggregate work means defining what a coherent payload is for a stable on-disk format. Too strict and existing valid datasets stop opening, which is a worse outcome than the hand-crafted cases being defended against. That tradeoff should be a maintainer decision rather than something folded into a binding PR.

One point worth flagging on the ordering case. I am not convinced it is corruption only. `finish_create_index` extends `versions` while reconciling concurrent rewrites, and I have not proven that always yields ascending `dataset_version`. If it does not, out of order versions could arise from a legitimate race and would deserve separate investigation.

PR #7222, opened 2026-06-11
refactor!(dir-catalog): remove table_version_storage_enabled and __manifest version storage

jackye1995:
`__manifest` cannot sustain per-commit version-write TPS; it suits table
create/remove, not every commit. This removes the `table_version_storage_enabled`
option and the entire `__manifest`-backed table-version path. Version
list/create/describe/batch-delete now use the physical `_versions/` directory
exclusively, which was always the source of truth, so no data is lost.

Kept: `table_version_tracking_enabled` (managed versioning) and `manifest_enabled`
(table tracking) — both unaffected and served from physical storage.

The multi-table transaction work that motivated version storage will be
re-proposed on a separate mechanism.


LuQQiu (approved):
agree that multi_table_transaction is still helpful but may be not this approach

PR #7223, opened 2026-06-11
feat(index): support distributed LabelList scalar index builds

jackye1995:
## Summary

`LabelListIndexPlugin::train_index` unconditionally rejected fragment-scoped
training (`fragment_ids.is_some()`), so a LabelList index could not be built
through the distributed / segmented path (per-fragment `execute_uncommitted` +
`merge_existing_index_segments`) the way BTree, Inverted, Bitmap, and FM already
are. This makes LabelList a first-class distributed scalar index.

## Changes

- `lance-index/src/scalar/label_list.rs`: `train_index` ignores `fragment_ids` and
  builds over the already fragment-scoped stream (mirrors `FMIndexPlugin`); a partial
  index over a fragment subset is correct since it covers exactly those rows. Add
  `merge_label_list_indices`, which unions the per-segment bitmap states and the
  `list_nulls` row sets. LabelList wraps a `BitmapIndex` plus a null-row set and
  distributed segments cover disjoint rows, so this is a cheap union (the same
  operation `LabelListIndex::update` performs) — no source re-scan. Mirrors
  `merge_bitmap_indices` but also carries `list_nulls`.
- `lance/src/index/scalar/label_list.rs` (new): `merge_segments` opens each source
  segment as a `LabelListIndex` and calls `merge_label_list_indices` (mirrors
  `scalar/bitmap.rs`).
- `lance/src/index.rs`: `merge_existing_index_segments` routes `all_label_list`
  segments (details `LabelListIndexDetails`) to the new merge; add
  `segment_has_label_list_details`.
- `lance/src/index/scalar.rs`: declare the `label_list` module.

Covered by `test_label_list_merge_existing_index_segments`: one LabelList segment per
fragment, merged via `merge_existing_index_segments`, answers `array_has_any` across
both fragments with the same row count as a pre-index full scan.

### DIFF
--- a/rust/lance/src/index/scalar/label_list.rs
+++ b/rust/lance/src/index/scalar/label_list.rs
@@ -1,0 +1,81 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright The Lance Authors
+
+use lance_index::metrics::NoOpMetricsCollector;
+use lance_index::scalar::label_list::LabelListIndex;
+use lance_index::scalar::lance_format::LanceIndexStore;
+use lance_table::format::IndexMetadata;
+use roaring::RoaringBitmap;
+use std::sync::Arc;
+use uuid::Uuid;
+
+use crate::{Dataset, Error, Result, dataset::index::LanceIndexStoreExt};
+
+/// Merge one caller-defined group of source LabelList segments into a single segment.
+///
+/// A LabelList index is a bitmap over the unnested list values plus a `list_nulls`
+/// row set, so segments are recombined by unioning the underlying bitmap states and
+/// null sets (see [`lance_index::scalar::label_list::merge_label_list_indices`])
+/// rather than rebuilding from source text.
+pub(in crate::index) async fn merge_segments(
+    dataset: &Dataset,
+    segments: Vec<IndexMetadata>,
+) -> Result<IndexMetadata> {
+    if segments.is_empty() {
+        return Err(Error::index("No segment metadata was provided".to_string()));
+    }
+
+    let field_id = *segments[0].fields.first().ok_or_else(|| {
+        Error::invalid_input(format!(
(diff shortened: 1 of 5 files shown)


rust/lance/src/index/scalar/label_list.rs
```diff
@@ -0,0 +1,81 @@
...
+    let mut fragment_bitmap = RoaringBitmap::new();
+    for segment in &segments {
+        fragment_bitmap |= segment.fragment_bitmap.as_ref().cloned().ok_or_else(|| {
+            Error::invalid_input(format!(
+                "CreateIndex: segment {} is missing fragment coverage",
+                segment.uuid
+            ))
+        })?;
+        let scalar_index =
+            super::open_scalar_index(dataset, &field_path, segment, &NoOpMetricsCollector).await?;
```
Xuanwo, line 112: This can silently convert legacy nullable LabelList segments into a current-format index.

Please reject merging nullable LabelList segments that are missing the list-null metadata, or rebuild from source data instead of writing them as current-format output.

rust/lance/src/index/scalar/label_list.rs
```diff
@@ -0,0 +1,81 @@
...
+            "CreateIndex: segment {} is missing field ids",
+            segments[0].uuid
+        ))
+    })?;
+    let field_path = dataset.schema().field_path(field_id)?;
+
+    let mut source_indices = Vec::with_capacity(segments.len());
+    let mut fragment_bitmap = RoaringBitmap::new();
+    for segment in &segments {
+        fragment_bitmap |= segment.fragment_bitmap.as_ref().cloned().ok_or_else(|| {
```
Xuanwo: This can publish stale coverage and row addresses when any source segment covers retired fragments.

Please compute effective live coverage and filter the merged bitmap state plus `list_nulls`, or rebuild from live fragments when deleted coverage is present.

Xuanwo (approved):

PR #7228, opened 2026-06-11
fix: record TakeExec output and I/O metrics

### LINKED ISSUE #7227: Bug: TakeExec reports zero output and I/O metrics after successful take
## Summary
`TakeExec` can successfully read and emit rows while reporting zero output batch, output byte, and I/O metrics.

This shows up in `analyze_plan()` output as a `Take` node with rows and processed batches, but no output batch/byte or scheduler I/O metrics:

`Take: metrics=[output_rows=10, output_bytes=0.0 B, output_batches=0, batches_processed=1, bytes_read=0, iops=0, requests=0]`

## Observed behavior

A minimal TakeExec execution that reads 10 row addresses returns the expected 10 rows, but the metrics remain zero: `output_batches=0, output_bytes=0, bytes_read=0, iops=0, requests=0`

output_rows and batches_processed can be populated, which makes the plan misleading: the operator clearly emitted data, but the output and I/O metrics suggest it did no work.

  ## Expected behavior

  When TakeExec emits a batch, it should report:
  - positive output_batches
  - positive output_bytes
  - positive bytes_read
  - positive iops
  - positive requests
  - correct output_rows
  - correct batches_processed

  ## Cause

TakeExec records some metrics before the spawned take/read work has actually completed. The I/O metrics are sampled before try_buffered resolves the map_batch future, and output rows are recorded through a row-only path instead of DataFusion's final RecordBatch metrics path.

## Suggested fix
Record metrics after the take future resolves and the final RecordBatch is ready to emit:
  - use BaselineMetrics::record_poll(...) for the final batch so DataFusion records rows, batches, and bytes
  - record ScanScheduler I/O metrics after the read work completes
  - increment batches_processed only for successfully emitted take batches
  - record a final I/O snapshot when the stream finishes

I will create PR with fix myself. Creating issue to be aligned with contribution guide. 

 ## Regression test

Add a Rust test for TakeExec that builds a small dataset, feeds five _rowaddr values through a one-shot input stream, executes TakeExec, and asserts:

  output_rows == 5
  batches_processed == 1
  output_batches > 0
  output_bytes > 0
  bytes_read > 0
  iops > 0
  requests > 0

skyshineb:
# Problem
Closes #7227. The IoMetrics of the `take` operator are broken. The metrics such as `output_bytes=0.0 B`, `output_batches=0`, `bytes_read=0` are not visible. 

Example:
`... Take: elapsed=350.781983256s, columns="_distance, _rowid, (vec)", metrics=[output_rows=10.00 K, elapsed_compute=302.52s, output_bytes=0.0 B, output_batches=0, batches_processed=1, bytes_read=0, iops=0, requests=0] ...`

# Cause
`TakeExec` was recording output row metrics inside `map_batch`, before the spawned take work had completed and before the final `RecordBatch` was emitted through DataFusion's metrics path. As a result,
  `output_batches`, `output_bytes`, and `ScanScheduler` I/O metrics stayed at zero even when take work read data successfully.

# Solution
  This moves metric recording to the post-`try_buffered` result path, records the final `RecordBatch` with `BaselineMetrics::record_poll`, and records `ScanScheduler` I/O metrics after the actual take/read work
  completes. The stream also finalizes baseline metrics and records one final I/O snapshot when it finishes.

  Added a regression test covering output rows, output batches, output bytes, bytes read, IOPS, requests, and `batches_processed`.

### DIFF
--- a/rust/lance/src/io/exec/take.rs
+++ b/rust/lance/src/io/exec/take.rs
@@ -4,6 +4,7 @@
 use std::borrow::Cow;
 use std::collections::{HashMap, HashSet};
 use std::sync::{Arc, Mutex};
+use std::task::Poll;
 
 use arrow::array::AsArray;
 use arrow::compute::{TakeOptions, concat_batches};
@@ -27,6 +28,7 @@ use lance_arrow::RecordBatchExt;
 use lance_core::datatypes::{Field, OnMissing, Projection};
 use lance_core::error::{DataFusionResult, LanceOptionExt};
 use lance_core::utils::address::RowAddress;
+use lance_core::utils::futures::FinallyStreamExt;
 use lance_core::utils::tokio::get_num_compute_intensive_cpus;
 use lance_core::{ROW_ADDR, ROW_ID};
 use lance_io::scheduler::{ScanScheduler, SchedulerConfig};
@@ -353,10 +355,6 @@ impl TakeStream {
             (None, None) => {}
         }
 
-        self.metrics
-            .baseline_metrics
-            .record_output(new_data.num_rows());
-        self.metrics.batches_processed.add(1);
         Ok(batch.merge_with_schema(&new_data, self.output_schema.as_ref())?)
     }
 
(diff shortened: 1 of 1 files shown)


wombatu-kun (approved):

wjones127 (approved):
This looks good. Thanks for working on this.

PR #7233, opened 2026-06-11
fix: handle empty batch from deletion in add_columns_from_stream

### LINKED ISSUE #7232: panic in add_columns_from_stream: index out of bounds when entire read batch is deleted
## Bug

`add_columns` panics with `index out of bounds: the len is 0 but the index is 0` when a dataset fragment has a deletion file and the deleted rows happen to fill an entire read batch.

## Root cause

In `add_columns_from_stream` ([`schema_evolution.rs`](https://github.com/lance-format/lance/blob/main/rust/lance/src/dataset/schema_evolution.rs#L495-L528)), the updater reads the fragment in physical batches and applies the deletion vector via `filter_record_batch`. When **all rows in a physical batch are deleted**, the filter produces a **0-row batch**. The outer loop sets `rows_remaining = 0`, the inner `while rows_remaining > 0` loop never executes, `batches` stays empty, and the subsequent `concat_batches(&batches[0].schema(), ..)` panics:

```
thread 'lance_background_thread' panicked at rust/lance/src/dataset/schema_evolution.rs:526:62:
index out of bounds: the len is 0 but the index is 0
```

## Minimal reproduction

```rust
// 105 rows in one fragment; delete the trailing 5 so that with batch_size=50
// the last physical batch [100..105) is entirely deleted.
let mut dataset = Dataset::write(reader, uri, Some(WriteParams {
    max_rows_per_file: 200, ..Default::default()
})).await?;
dataset.delete("i >= 100").await?;

// panics on unpatched code:
dataset.add_columns(NewColumnTransform::Reader(Box::new(reader)), None, Some(50)).await?;
```

## Fix

When the updater yields a 0-row batch, feed an empty batch back to keep the updater in sync and continue:

```rust
if rows_remaining == 0 {
    updater.update(RecordBatch::new_empty(stream.schema())).await?;
    continue;
}
```

A PR with the fix and a regression test will follow.

zhangyang0418:
## What

Fixes a panic in `add_columns` when a fragment's deleted rows fill an entire read batch.

Closes #7232.

## Root cause

In `add_columns_from_stream`, the updater reads each fragment in physical batches and applies the deletion vector via `filter_record_batch`. When **every row in a physical batch is deleted**, the filter yields a **0-row batch**. The outer loop then sets `rows_remaining = 0`, the inner `while rows_remaining > 0` loop never runs, `batches` stays empty, and `concat_batches(&batches[0].schema(), ..)` panics with `index out of bounds: the len is 0 but the index is 0`.

This was hit in production using a merge-columns workflow over a dataset whose fragment had a deletion file, where the deleted rows happened to align with a read-batch boundary.

## Fix

When the updater yields a 0-row batch, feed an empty batch back to the updater (to keep it in sync) and continue:

```rust
if rows_remaining == 0 {
    updater.update(RecordBatch::new_empty(stream.schema())).await?;
    continue;
}
```

## Test

Adds `test_add_columns_with_fully_deleted_batch`: writes a single fragment with 105 rows, deletes the trailing 5 rows so that — read with `batch_size=50` — the last batch `[100..105)` is fully deleted, then verifies `add_columns` succeeds and the new column has the correct values. The test panics on the unpatched code and passes with the fix.


wjones127 (approved):
This looks like a clean fix, with a nice test. Thank you!

Good to merge once CI is passing.

PR #7235, opened 2026-06-11
fix: deduplicate BTree flat page row addresses

### LINKED ISSUE #7230: BTree scalar index can contain duplicate row addresses
## Summary

A BTree scalar index can contain duplicate row addresses after this public Python API flow:

1. Create a dataset.
2. Create a BTree scalar index.
3. Update the indexed row with `merge_insert`.
4. Create a delta BTree index segment with `optimize_indices(num_indices_to_merge=0)`.
5. Merge the old segment and delta segment with `optimize_indices(num_indices_to_merge=2)`.
6. Run an indexed filter.

The final indexed filter fails with:

```text
RowAddrTreeMap::from_sorted_iter called with non-sorted input
```

The error message is misleading. The row addresses are not necessarily out of order. In this repro, the BTree flat page can contain duplicate row addresses.

For example:

```text
[0, 0]
```

This is sorted, but not strictly increasing.

## Background

A BTree scalar index maps indexed values to row addresses.

A BTree flat page stores pairs like:

```text
(indexed value, row address)
```

For example, for an `id` column:

```text
values   row_ids
------   -------
1        0
```

In code, the flat page has two columns:

```text
BTREE_VALUES_COLUMN  -> indexed scalar values
BTREE_IDS_COLUMN     -> row addresses
```

When a flat page is loaded, `FlatIndex::try_new()` sorts by row id and then builds a row-address lookup:

```rust
let data = data.sort_by_column(IDS_COL_IDX, None)?;

let all_addrs_map = RowAddrTreeMap::from_sorted_iter(
    data.column(IDS_COL_IDX)
        .as_primitive::<UInt64Type>()
        .values()
        .iter()
        .copied(),
)?;
```

This requires the row ids to be strictly increasing after sorting.

So this is valid:

```text
[0, 1, 2]
```

But this is invalid:

```text
[0, 0]
```

even though it is sorted.

## Minimal repro

This uses only public Lance Python APIs.

```python
#!/usr/bin/env python3

import tempfile
from pathlib import Path

import pyarrow as pa

import lance


def print_rows(ds: lance.LanceDataset, label: str) -> None:
    rows = ds.to_table(with_row_id=True, with_row_address=True).to_pylist()
    print(f"{label}: {rows}")


with tempfile.TemporaryDirectory(prefix="lance-btree-repro-") as tmpdir:
    uri = str(Path(tmpdir) / "btree-duplicate-row-addr.lance")
    print(f"lance version: {lance.__version__}")
    print(f"dataset uri: {uri}")

    ds = lance.write_dataset(
        pa.table({"id": [1], "payload": [10]}),
        uri,
        mode="overwrite",
    )
    print_rows(ds, "initial rows")

    ds.create_scalar_index("id", index_type="BTREE")
    print("created BTree index on id")

    # The reordered source columns are important. With the same column order as
    # the dataset schema, this takes a different path and does not reproduce.
    source = pa.table({"payload": [100], "id": [1]})
    ds.merge_insert("id").when_matched_update_all().execute(source)
    print_rows(ds, "after merge_insert")

    print("creating a delta BTree segment for unindexed updated data")
    ds.optimize.optimize_indices(num_indices_to_merge=0)
    print_rows(ds, "after optimize_indices(num_indices_to_merge=0)")

    print("forcing old BTree segment + delta segment to merge")
    ds.optimize.optimize_indices(num_indices_to_merge=2)
    print_rows(ds, "after optimize_indices(num_indices_to_merge=2)")

    print("running indexed filter; affected builds fail here")
    print(ds.to_table(filter="id = 1").to_pylist())
```

## Actual result

Tested on latest `main`:

```text
109fa3cbc8badfa0239b544e64a36adba195c826
```

with:

```text
lance version: 8.0.0-beta.11
```

Output:

```text
lance version: 8.0.0-beta.11
dataset uri: /tmp/lance-btree-repro-442vi2_e/btree-duplicate-row-addr.lance
initial rows: [{'id': 1, 'payload': 10, '_rowid': 0, '_rowaddr': 0}]
created BTree index on id
after merge_insert: [{'id': 1, 'payload': 100, '_rowid': 0, '_rowaddr': 0}]
creating a delta BTree segment for unindexed updated data
after optimize_indices(num_indices_to_merge=0): [{'id': 1, 'payload': 100, '_rowid': 0, '_rowaddr': 0}]
forcing old BTree segment + delta segment to merge
after optimize_indices(num_indices_to_merge=2): [{'id': 1, 'payload': 100, '_rowid': 0, '_rowaddr': 0}]
running indexed filter; affected builds fail here
```

Then the final indexed filter fails:

```text
pyarrow.lib.ArrowInvalid: External error: Encountered internal error.
Please file a bug report at https://github.com/lance-format/lance/issues.
RowAddrTreeMap::from_sorted_iter called with non-sorted input,
rust/lance-index/src/scalar/btree/flat.rs:58:29
```

## Expected result

The final indexed filter should succeed and return:

```python
[{"id": 1, "payload": 100}]
```

## Why this happens

The initial BTree index contains one entry:

```text
(id=1, rowaddr=0)
```

Then `merge_insert` updates the row:

```text
before: id=1, payload=10,  rowaddr=0
after:  id=1, payload=100, rowaddr=0
```

The visible table state is correct. The row still has:

```text
_rowaddr=0
```

Next:

```python
ds.optimize.optimize_indices(num_indices_to_merge=0)
```

creates a delta BTree segment for the updated data.

That delta segment also contains:

```text
(id=1, rowaddr=0)
```

Now there are two BTree segments:

```text
old segment:
  (id=1, rowaddr=0)

delta segment:
  (id=1, rowaddr=0)
```

Then:

```python
ds.optimize.optimize_indices(num_indices_to_merge=2)
```

forces the old segment and delta segment to merge.

The merged BTree flat page can contain:

```text
values   row_ids
------   -------
1        0
1        0
```

So the row ids are:

```text
[0, 0]
```

This is sorted, but contains a duplicate.

When the final indexed filter loads the BTree flat page, `FlatIndex::try_new()` calls:

```rust
RowAddrTreeMap::from_sorted_iter(...)
```

which fails because the row ids are not strictly increasing.

## Why the flat page layer should handle this

`RowAddrTreeMap` only sees row ids:

```text
[0, 0]
```

It does not see the indexed values:

```text
values   row_ids
------   -------
1        0
1        0
```

So it cannot distinguish between:

```text
same-value duplicate:
(id=1, rowaddr=0)
(id=1, rowaddr=0)
```

and:

```text
conflicting duplicate:
(id=1, rowaddr=0)
(id=2, rowaddr=0)
```

The BTree flat page layer has both columns, so it can make the right decision.

## Suggested fix

Normalize or validate duplicate row addresses before constructing `RowAddrTreeMap`.

Suggested place:

```text
rust/lance-index/src/scalar/btree/flat.rs
```

Apply in:

```text
FlatIndex::try_new()
FlatIndex::remap_batch()
```

Suggested logic:

```text
sort by row id
for consecutive duplicate row ids:
  if indexed value is equal:
    keep one row
  else:
    return a clear internal error
```

Same-value duplicate entries can be safely deduplicated:

```text
(id=1, rowaddr=0)
(id=1, rowaddr=0)
```

Conflicting duplicate entries should be treated as index corruption:

```text
(id=1, rowaddr=0)
(id=2, rowaddr=0)
```

Suggested error message for the conflicting case:

```text
BTree flat index contains duplicate row id with conflicting values
```

## Why this fix is safe

Deduplicating identical `(value, rowaddr)` pairs does not change query results.

Before deduplication:

```text
id = 1 -> rowaddr 0
id = 1 -> rowaddr 0
```

After deduplication:

```text
id = 1 -> rowaddr 0
```

The query still returns the same row once.

But conflicting duplicates should not be silently deduplicated, because they mean the same row address appears under different indexed values. That would be a stale or corrupt index entry.

majin1102:
## Summary

Fixes #7230.

BTree segment merge preserves value ordering with `UnionExec -> SortPreservingMergeExec(value ASC) -> train_btree_index`. However, when an updated row contributes the same indexed value again, the old segment row and the new delta row can both survive as duplicate `(value, row address)` entries. Later, loading the flat page sorts row addresses and builds `RowAddrTreeMap`, which rejects duplicate row addresses with the misleading `from_sorted_iter called with non-sorted input` error.

This PR fixes the write path during BTree segment merge. After the existing value-ordered merge, it scans each equal-value group and drops duplicate row addresses for that value before training the new BTree pages. The read path stays unchanged, and the merge still avoids an extra global sort.

## Why this approach

- The bug is produced while writing merged BTree segments, so the fix belongs in `BTreeIndex::merge_segments()`.
- Deduplicating only exact `(value, row address)` duplicates is the minimal behavior needed for this repro.
- A simple previous-row check is not sufficient because the merge is ordered by value only; duplicate row addresses can be separated by other rows with the same value.
- Adding `row_id` as a secondary sort key would be heavier than a linear scan over the already value-ordered stream.

## Tests

- `cargo fmt --all`
- `cargo test -p lance-index scalar::btree -- --nocapture`
- `cargo test -p lance test_btree_merge_deduplicates_row_addrs -- --nocapture`
- `cargo clippy -p lance-index --tests -- -D warnings`
- `git diff --check`


wjones127 (commented):
This doesn't feel like the right solution. Two problems I see:

1. The real bug seems to me that training would have any duplicates in the input.
2. This only handles the case where the old and new scalar value for that row id are the same. But we could also have bugs where the value has been updated, and thus the row id is duplicated but across different unique values. Again, I think this should be properly fixed by (1). As you point out, deduplicating across the whole input would be expensive.

majin1102:
> This doesn't feel like the right solution. Two problems I see:
> 
> 1. The real bug seems to me that training would have any duplicates in the input.
> 2. This only handles the case where the old and new scalar value for that row id are the same. But we could also have bugs where the value has been updated, and thus the row id is duplicated but across different unique values. Again, I think this should be properly fixed by (1). As you point out, deduplicating across the whole input would be expensive.

Ye,  sorry I did't get this clearly.
This has been fixed by #7320

PR #7237, opened 2026-06-11
feat(compaction): introduce RowAddrRemap structure to avoid remap OOM caused by HashMap

### LINKED ISSUE #7150: Compaction remap OOM on Large Scale
## 1. The problem

In `commit_compaction`, the coordinator builds one giant `HashMap<u64, Option<u64>>` that maps **every affected row** (survivors *and* deletions) to its new address, then hands the whole map to each index for `remap`.

Relevant code: [`commit_compaction`](https://github.com/lancedb/lance/blob/main/rust/lance/src/dataset/optimize.rs) → [`transpose_row_ids_from_digest`](https://github.com/lancedb/lance/blob/main/rust/lance/src/dataset/optimize/remapping.rs).

```rust
let mut row_id_map: HashMap<u64, Option<u64>> = HashMap::default();
for task in completed_tasks {
    let row_addrs = RoaringTreemap::deserialize_from(&mut Cursor::new(&bytes))?;
    let transposed = remapping::transpose_row_addrs(row_addrs, &old_frags, &new_frags);
    row_id_map.extend(transposed); // ← inflates the compact RoaringTreemap into per-row entries
}
index_remapper.remap_indices(row_id_map, &affected_ids).await?;
```

### Why this hurts
- The map carries **one entry per affected row** (survivor *and* deleted), not per fragment. `hashbrown` costs ~24–40 B per entry.
- Workers already ship survivors as a compact `RoaringTreemap` in `RewriteResult.row_addrs`. The coordinator **inflates it back into a per-row hash table** — that's the regression.
- Affects datasets that do not enable `enable_move_stable_row_ids` (i.e., must use address-based remap).

### Measured impact (64 c / 247 GiB box, BTREE index, 10% deletes)

| Scale | `HashMap` size | RSS Δ during build | Process VmHWM | Build time |
|---|---:|---:|---:|---:|
| 500 M rows | **28.0 GB** | **+25.1 GB** | 64.7 GB | **159.9 s** |
| 1 B rows | **56.0 GB** | **+50.2 GB** | **122.1 GB** | **322.7 s** |

At 1 B rows the coordinator's working set already exceeds the RAM of common 64–96 GiB workers, so the compaction OOMs before it can commit.

## 2. Proposed fix

**Idea in one sentence:** keep the per-fragment survivor bitmaps the workers already ship and answer each lookup on the fly via per-fragment `RoaringBitmap::rank` plus a precomputed `rank_base` prefix sum, instead of inflating those bitmaps into an O(rows) `HashMap` on the coordinator.

Replace the materialized map with a **compact, queryable** structure that preserves the exact `HashMap::get(&addr).copied()` tri-state semantics (`None` = unaffected, `Some(None)` = deleted, `Some(Some(a))` = moved):

```rust
pub enum RowAddrRemap {
    Compact(CompactRowAddrRemap),         // computed on the fly; no per-row entries
    Explicit(HashMap<u64, Option<u64>>),  // adapter for callers that already hold a HashMap
}
```

**`get(addr)` is the single hot-path query.** The argument `addr: u64` is a row address (encoded as `(frag_id << 32) | offset`, produced by `RowAddress::new_from_parts(frag_id, offset)`); it returns the `Option<Option<u64>>` tri-state. The shorthand `get((11, 2))` used below is equivalent to `get(RowAddress::new_from_parts(11, 2).into())` — i.e. "where does the row at old `frag_id=11`, `offset=2` end up?". Each index `remap` calls it once per indexed row address.

`CompactRowAddrRemap` keeps, per rewrite group:

- one `RoaringBitmap` of survivor offsets per old fragment (already what workers send);
- a precomputed per-fragment `rank_base`: by ascending `frag_id`, the cumulative survivor count contributed by all earlier fragments in the group (a **prefix sum** of per-fragment survivor counts). Example: `frag=10` has 6 survivors, `frag=11` has 4 → `rank_base[10] = 0`, `rank_base[11] = 6`.

  **Why this matters.** `get((11, 2))` needs to answer "what is this address's index *i* in the group's full ascending list of survivor addresses?". The two implementations differ in **which `rank` they call**:

  | Naive | This PoC |
  |---|---|
  | Calls `RoaringTreemap::rank(addr)` on a cross-fragment treemap. The library implementation walks fragments from `frag_id=0` and **accumulates each earlier fragment's cardinality** before reaching the target — a group with 64 fragments scans 64 bitmaps, O(F), hundreds of ns up to a microsecond per `get`. | Calls `RoaringBitmap::rank(offset)` on the **target fragment's own bitmap only** (a few ns); the cross-fragment prefix is read straight from the precomputed `rank_base[frag]` (one O(F) precomputation, then O(1) lookup per `get`). |

  Combining the two — `rank_base[frag] + bitmap.rank(offset) - 1` — gives the global survivor index without **ever touching another fragment's bitmap**. Per-`get` is effectively O(1) (~15–25 ns, cache-friendly).

- a prefix-sum array over the new fragments, used to translate a global survivor index back into a new row address.

`get(addr)` in pseudocode:

```text
frag = addr >> 32; off = addr as u32
group = frag_to_group[frag]?                        // not in any group → unaffected
match group.frags[frag] {
    Some((bitmap, base)) if bitmap.contains(off) =>
        Some(Some(select_new(base + bitmap.rank(off) - 1))),
    _ => Some(None),                                // in the group but not a survivor → deleted
}
```

Two pieces in the snippet to clarify:

- **`base`** is `rank_base[frag]` — computed once at construction time and stored alongside the bitmap in `group.frags`.
- **`select_new(i)`** decodes a global survivor index back into a new row address. The earlier expression `base + bitmap.rank(off) - 1` yields "this address is the *i*-th survivor in ascending order across the group", but consumers need the *new* address. `select_new` binary-searches the `new_prefix: Vec<(frag_id, start, len)>` array, finds the new fragment with `i ∈ [start, start+len)`, and returns `RowAddress::new_from_parts(frag_id, i - start)`. Example: with `new_prefix = [(200, 0, 6), (201, 6, 4)]`, `select_new(8) → (frag=201, offset=2)`.

**Worked example.** Old `frag=10` (8 rows; offsets 2 and 5 deleted) and old `frag=11` (4 rows, all alive) are merged into new `frag=200`; old `frag=99` is not part of the compaction. The three states line up exactly:

| Query | Old HashMap entry | `Compact` on-the-fly | Result |
|---|---|---|---|
| `(99, 0)` | absent | `frag_to_group` miss | `None` (unaffected) |
| `(10, 5)` | `(10,5) -> None` | `frag_to_group[10]=0`; `bitmap{0,1,3,4,6,7}.contains(5)=false` | `Some(None)` (deleted) |
| `(11, 2)` | `(11,2) -> Some((200,8))` | `rank_base=6` + single-fragment `rank=3` − 1 = global index 8 → `select_new(8)=(200,8)` | `Some(Some((200,8)))` (moved) |

**Equivalence is provable.** "The *i*-th survivor in ascending order" ↔ "the *i*-th new offset in `new_fragments` order" = `rank_base + single-fragment rank − 1` ↔ `select_new(i)`, which is exactly the position-aligned `zip` that `transpose_row_ids_from_digest` performs today.

## 3. PoC results

Same hardware and dataset shape as above:

| Scale | Metric | Definition | `main` | PoC | Δ |
|---|---|---|---:|---:|---|
| 500 M | structure size | Self-reported byte estimate of the remap structure (HashMap: `capacity × 32`; PoC: sum of all `RoaringBitmap::serialized_size`). | 28.0 GB | **78 MB** | **−366×** |
| 500 M | RSS Δ during build | `VmRSS@remap_build_end − VmRSS@remap_build_begin`. Net physical memory the remap module adds, from the OS's viewpoint (the OOM-attribution metric). | +25.1 GB | **0** | OOM risk gone |
| 500 M | VmHWM | Process-wide RSS high-water mark from start to `remap_indices_end`. Drives instance sizing. | 64.7 GB | **14.2 GB** | −4.5× |
| 500 M | build time | Wall clock from receiving the task list to the remap structure being ready. | 159.9 s | **0.04 s** | −3845× |
| 500 M | `get` latency | Micro-bench: fixed-seed RNG, 80% hit / 20% miss, 2 M lookups, average per-`get` latency. | 154 ns | 590 ns | +3.8× |
| 500 M | end-to-end `remap_indices` | Wall clock of `index_remapper.remap_indices(...)` (includes IO / decode / encode). | 265.9 s | 338.2 s | +27% (IO-bound) |
| 500 M | **total compaction** | Wall clock of `compact_files` from start to commit. | 448.7 s | **363.5 s** | **−19%** |
| 1 B | structure size | Same | 56.0 GB | **147 MB** | **−391×** |
| 1 B | RSS Δ during build | Same | +50.2 GB | **0** | OOM risk gone |
| 1 B | VmHWM | Same | **122.1 GB** | **21.3 GB** | **−5.7×** |
| 1 B | build time | Same | 322.7 s | **0.08 s** | **−4015×** |
| 1 B | `get` latency | Same | 162 ns | 759 ns | +4.7× |
| 1 B | **total compaction** | Same | 1248.8 s | **1142.5 s** | **−8.5%** |

A single `get` is 3–5× slower (O(rank) vs amortized O(1)), but `remap_indices` is IO-bound (read / decode / encode / write of index pages), and the build-time savings dominate end-to-end. **Net: OOM is eliminated and total compaction is faster.**

### 1 B-row visualization

Memory (linear, GiB):

Image

Wall-clock time (log, seconds):

Image

zhangyue19921010:
Closes: https://github.com/lance-format/lance/issues/7150

# Compact Row-Address Remapping for Compaction

Compaction rewrites rows into new fragments, so indices that store physical row addresses need an old-address to new-address mapping without building an `O(total rows)` `HashMap<u64, Option<u64>>`.

### Old Rows

`old_fragment_id -> (old_offsets, old_rows_before)`

* `old_offsets`: rewritten old row offsets in this old fragment.
* `old_rows_before`: rewritten row count before this old fragment.

### New Rows

Ordered new-fragment ranges:

`(fragment_id, new_rows_before, physical_rows)`

* `new_rows_before`: rewritten row count before this new fragment.

## Lookup

* An address whose fragment was not rewritten returns `None`.

* For an address whose fragment was rewritten:

  1. Read `(old_offsets, old_rows_before)` from the old-row layout.

  2. If `offset` is not in `old_offsets`, return `Some(None)` because the row was deleted.

  3. Otherwise, `old_offsets.rank(offset) - 1` is this row's 0-based position among rewritten old rows in this old fragment.

  4. Add `old_rows_before` to get `k`, the row's 0-based position among all rewritten old rows.

  5. In the new-row layout, find the range:

     ```text
     (fragment_id, new_rows_before, physical_rows)
     ```

     where:

     ```text
     new_rows_before <= k < new_rows_before + physical_rows
     ```

  6. The new address is:

     ```text
     (fragment_id, k - new_rows_before)
     ```

## Ordering

Compact remap does not store each old-to-new row mapping. It computes `k` from the old-row layout, then maps it to the k-th row written to the new fragments.

This requires the reader-to-writer pipeline to preserve row order.

* `old_frag_ids` must match the order old fragments are read.

  * Within each old fragment, rewritten rows are interpreted by ascending old row offset.

* `new_frags` must match the order new rows are written.

* Current compaction satisfies this because it scans selected fragments in order and writes the resulting stream without reordering rows.

ig_03eb56d1be245827016a2a76f6c3588195a974dfbe6eea7146


zhangyue19921010:
Hi @wjones127 Would u mind to take a look? Thanks!

In the scenario of a billion-row large table, this remap OOM issue is one of the problems worth resolving.

rust/lance-core/src/utils/row_addr_remap.rs
```diff
@@ -0,0 +1,402 @@
...
+                frags.insert(frag_id, (bitmap, rewritten_rows_before));
+                rewritten_rows_before += num_rewritten_rows;
+            }
+        }
+        // Rewritten old row addresses must reference only fragments listed in `old_frag_ids`.
+        if !per_frag.is_empty() {
+            return Err(Error::invalid_input(format!(
+                "compaction rewritten old row addresses reference fragments {:?} not in the rewrite group's old fragments {:?}",
+                per_frag.keys().collect::<Vec<_>>(),
+                input.old_frag_ids,
```
yanghua, line 191: It would be better to check the row num between `new_frags ` and `rewritten_old_row_addrs `.

rust/lance/src/dataset/optimize.rs
```diff
@@ -2234,18 +2321,26 @@ mod tests {
...
+                    .map(|addr| (addr >> 32) as u32)
+                    .collect();
+                if index_map.affected_fragments() == expected_frags
+                    && expectation
+                        .expected
+                        .iter()
+                        .all(|(k, v)| index_map.get(*k) == Some(*v))
+                {
                     return Ok(expectation.answer.clone());
                 }
```
yanghua, line 2333: There is an assertion regression. It will miss the extra erroneous mappings within the same fragment, weakening the regression test protection for the newly introduced Compact remap semantics.

yanghua (commented):
Left two comments.

rust/lance/src/dataset/optimize.rs
zhangyue19921010: changed.

rust/lance-core/src/utils/row_addr_remap.rs
zhangyue19921010: Added.

yanghua (approved):
LGTM. Great optimization! cc @wjones127 @jackye1995

zhangyue19921010:
> LGTM. Great optimization! cc @wjones127 @jackye1995

Thanks! @yanghua further benchmark report can be found [[External] RowAddrRemap PR 1B Test Report.pdf](https://github.com/user-attachments/files/29549511/External.RowAddrRemap.PR.1B.Test.Report.pdf)

PR #7244, opened 2026-06-12
feat(index): support segmented ngram index

zhangyue19921010:


zhangyue19921010:
Hi @BubbleCal ~ This is a ngram segmented index adoption. Would u have time to take a look? Really appreciate it!

zhangyue19921010:
Hi @Xuanwo  ~ This is a ngram segmented index adoption. Would u mind to take look? Thanks in advance!

Xuanwo (commented):
`commit_existing_index_segments` can attach a staged segment to a different column snapshot and silently hide matching rows.

The segment preserves `dataset_version`, but `IntoIndexSegment for IndexMetadata` drops the source fields; commit resolves the current column and uses the coordinator's current manifest version as the transaction `read_version`. If the indexed column is dropped/recreated (the same issue applies to an indexed-field `Merge` or `DataReplacement`) after the segment was built and the coordinator has refreshed, those intervening transactions are never checked. The old NGram postings are then advertised as covering the current fragment, so the query has no flat fallback.

On `f2cfee815`, the public-API reproducer below builds the segment at V1, replaces `text` at V2/V3, and commits the V1 segment. `contains(text, 'dog')` returns 3 before the commit and 0 afterward.

<details>
<summary>Reproducer</summary>

```rust
use std::sync::Arc;

use arrow_array::{ArrayRef, Int32Array, RecordBatch, RecordBatchIterator, StringArray};
use lance::dataset::NewColumnTransform;
use lance::index::{CreateIndexBuilder, DatasetIndexExt};
use lance::Dataset;
use lance_index::scalar::{BuiltinIndexType, ScalarIndexParams};
use lance_index::IndexType;

#[tokio::main]
async fn main() -> lance::Result<()> {
        Field::new("text", DataType::Utf8, false),
    ]));
    let batch = RecordBatch::try_new(
        schema.clone(),
        vec![
            Arc::new(Int32Array::from(vec![0, 1, 2])) as ArrayRef,
            Arc::new(StringArray::from(vec!["cat-zero", "cat-one", "cat-two"])) as ArrayRef,
        ],
    )?;
    let reader = RecordBatchIterator::new(vec![Ok(batch)], schema);
    let tmp = tempfile::tempdir()?;
    let mut dataset = Dataset::write(reader, tmp.path().to_str().unwrap(), None).await?;

    let params = ScalarIndexParams::for_builtin(BuiltinIndexType::NGram);
    let segment = CreateIndexBuilder::new(
        &mut dataset,
        &["text"],
        IndexType::NGram,
        &params,
    )
    .name("text_ngram".to_string())
    .fragments(vec![0])
    .execute_uncommitted()
    .await?;
    let source_version = segment.dataset_version;
    let source_field_id = segment.fields[0];

    dataset.drop_columns(&["text"]).await?;
    dataset
        .add_columns(
            NewColumnTransform::SqlExpressions(vec![(
                "text".to_string(),
                "concat('dog-', cast(id as varchar))".to_string(),
            )]),
            Some(vec!["id".to_string()]),
            None,
        )
        .await?;
    let replacement_version = dataset.version().version;
    let replacement_field_id = dataset.schema().field("text").unwrap().id;
    let before_commit = dataset
        .count_rows(Some("contains(text, 'dog')".to_string()))
        .await?;

    dataset
        .commit_existing_index_segments("text_ngram", "text", vec![segment])
        .await?;
    let committed = dataset.load_indices_by_name("text_ngram").await?;
    let after_commit = dataset
        .count_rows(Some("contains(text, 'dog')".to_string()))
        .await?;

    assert_eq!(source_version, 1);
    assert_eq!(replacement_version, 3);
    assert_ne!(source_field_id, replacement_field_id);
    assert_eq!(committed[0].dataset_version, source_version);
    assert_eq!(before_commit, 3);
    assert_eq!(
        after_commit, before_commit,
        "committing an index built for the replaced column must not hide current matches"
    );
    Ok(())
}
```

The final assertion fails with `left: 0, right: 3`.

</details>

The delayed commit needs to reject or rebuild a segment when its source field/snapshot no longer matches the current indexed data, rather than publishing its old coverage.

Xuanwo (approved):
Thank you!

zhangyue19921010:
@Xuanwo and @jackye1995 Thanks a lot for all your help. Really appreciate it!

PR #7248, opened 2026-06-12
feat: support cleanup explain for python and java

yanghua:
## Summary

  Add Java and PyLance APIs for cleanup explain.

  - Add PyLance `Dataset.explain_cleanup_old_versions(...)` with cleanup explanation
  result types
  - Export PyLance cleanup explanation types from the top-level `lance` package
  - Add Java `dataset.cleanup(policy).explain()` / `.execute()` operation API
  - Keep Java `cleanupWithPolicy(policy)` as a compatibility wrapper for execute
  - Add JNI conversions and tests for cleanup explanations

  ## Testing

  - `cargo fmt --all`
  - `cargo fmt --manifest-path ./java/lance-jni/Cargo.toml --all`
  - `cargo check --manifest-path python/Cargo.toml`
  - `cargo check --manifest-path ./java/lance-jni/Cargo.toml`
  - `./mvnw compile`
  - `./mvnw -Dtest=CleanupTest test`
  - `uv run --frozen --python 3.12 pytest python/tests/
  test_dataset.py::test_explain_cleanup_old_versions`

### DIFF
--- a/java/lance-jni/src/blocking_dataset.rs
+++ b/java/lance-jni/src/blocking_dataset.rs
@@ -3133,34 +3133,33 @@ fn inner_explain_cleanup_with_policy<'local>(
 }
 
 fn extract_cleanup_policy(env: &mut JNIEnv<'_>, jpolicy: &JObject) -> Result<CleanupPolicy> {
-    let before_ts_millis =
-        env.get_optional_u64_from_method(&jpolicy, "getBeforeTimestampMillis")?;
+    let before_ts_millis = env.get_optional_u64_from_method(jpolicy, "getBeforeTimestampMillis")?;
     let before_timestamp = before_ts_millis.map(|millis| {
         let st = UNIX_EPOCH + Duration::from_millis(millis);
         DateTime::<Utc>::from(st)
     });
 
-    let before_version = env.get_optional_u64_from_method(&jpolicy, "getBeforeVersion")?;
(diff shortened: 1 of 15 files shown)


Xuanwo (commented):
I found two issues in the Java cleanup explain binding that should be fixed before this lands.

First, the JNI conversion should release local references while building cleanup explanation lists. `cleanup_candidate_files_to_java` can convert up to the Rust default of 1000 candidate files, and each entry creates local refs for the path string, kind string, and `CleanupCandidateFile`. Since the loop does not use a local frame or delete those temporary refs, `dataset.cleanup(policy).explain()` can overflow the JNI local reference table on larger explanations. The same pattern should be handled for referenced branches and warnings.

Second, Java should expose the candidate file limit that Rust core and Python already expose. Rust has `with_max_candidate_files`, and Python exposes this as `max_files`, but Java currently only has `cleanup(policy).explain()` and the JNI path calls `.explain()` directly. That fixes Java at the default 1000 candidate files, so users cannot request more details when `candidateFilesTruncated` is true, and cannot lower the amount of detail returned. A Java surface like `cleanup(policy).withMaxCandidateFiles(n).explain()` or `explain(n)` should pass the value through JNI to Rust `with_max_candidate_files(n)`.

yanghua:
@Xuanwo thanks, I have addressed your comments.

Xuanwo (approved):
Thank you!

PR #7250, opened 2026-06-12
fix: avoid versioned describe table for namespace opens

### LINKED ISSUE #609: Lance Spark writes fail with REST namespaces that don't support describing a specific table version
## Problem

When using Lance Spark with a REST namespace implementation and client-side storage options, writes can fail during driver commit if the namespace implementation does not support `describeTable` with a version parameter.

Example Spark catalog config:

```text
spark.sql.catalog.lance=org.lance.spark.LanceNamespaceSparkCatalog
spark.sql.catalog.lance.impl=rest
spark.sql.catalog.lance.uri=http://127.0.0.1:9101/lance
spark.sql.catalog.lance.parent=lance_client_storage

spark.sql.catalog.lance.storage.endpoint=http://127.0.0.1:9000
spark.sql.catalog.lance.storage.allow_http=true
spark.sql.catalog.lance.storage.access_key_id=...
spark.sql.catalog.lance.storage.secret_access_key=...
spark.sql.catalog.lance.storage.region=us-east-1
```

A simple insert can fail after task fragments are written:

```sql
CREATE TABLE lance.demo.orders (id INT, name STRING) USING lance;
INSERT INTO lance.demo.orders VALUES (1, 'alice'), (2, 'bob');
```

Error:

```text
Unsupported: Describing specific table version is not supported.
It should be null to indicate the latest version.

at org.lance.namespace.RestNamespace.describeTable
at org.lance.OpenDatasetBuilder.buildFromNamespaceClient
at org.lance.spark.utils.Utils$OpenDatasetBuilder.build
at org.lance.spark.write.LanceBatchWrite.commit
```

## Root Cause

`OpenDatasetBuilder.buildFromNamespaceClient()` currently copies `ReadOptions.version` into `DescribeTableRequest`:

```java
DescribeTableRequest request = new DescribeTableRequest();
request.setId(tableId);
options.getVersion().ifPresent(v -> request.setVersion(Long.valueOf(v)));

DescribeTableResponse response = namespaceClient.describeTable(request);
```

However, this version is the Lance dataset manifest version that should be used when opening the dataset. It does not need to be sent to `describeTable`, whose role here is to resolve table-level metadata such as location, storage options, and whether namespace-managed versioning is enabled.

The version is already passed later into the actual dataset open path:

```java
options.getVersion().ifPresent(optionsBuilder::setVersion);

return Dataset.open(
    allocator,
    selfManagedAllocator,
    location,
    optionsBuilder.build(),
    baseStoreParams,
    session,
    namespaceClient,
    tableId,
    Boolean.TRUE.equals(namespaceClientManagedVersioning));
```

On the Rust side, if `managed_versioning=false`, `DatasetBuilder::from_uri(location).with_version(version).load()` can resolve the requested version from the underlying Lance manifest files. If `managed_versioning=true`, the namespace-backed external manifest store can resolve versions via namespace table-version APIs.

## Suggested Fix

Do not set `DescribeTableRequest.version` in `OpenDatasetBuilder.buildFromNamespaceClient()`.

Suggested behavior:

```java
DescribeTableRequest request = new DescribeTableRequest();
request.setId(tableId);

DescribeTableResponse response = namespaceClient.describeTable(request);
```

Keep passing `ReadOptions.version` to the lower-level dataset open path:

```java
options.getVersion().ifPresent(optionsBuilder::setVersion);
```

This keeps responsibilities separate:

- `describeTable`: resolve table location, storage options, and managed versioning mode
- `ReadOptions.version`: select the Lance dataset manifest version
- `Transaction.readVersion`: provide OCC conflict detection during commit

## Validation

Observed behavior:

- Lance Spark 0.2.0 works with `impl=rest` and client-side storage options because its write path opens the dataset by URI and does not call `describeTable(version)`.
- Current main / 0.5.1 fails in `LanceBatchWrite.commit` because namespace-backed open passes the pinned version into `describeTable`.

FANNG1:
## What
Avoid passing the requested dataset version to namespace `describe_table` / `describeTable` when opening a dataset through a namespace client.

The namespace describe call is only used to resolve the table location and storage options. The requested dataset version is still applied by the lower-level dataset open path after namespace resolution.

## Why
Some namespace implementations do not support describing a table from a specific version. Passing `version` to `describe_table` can fail even though opening the dataset at that version would work once the location and storage options are resolved.

Fixes lance-format/lance-spark#609.

## Testing
- `JAVA_HOME=/opt/homebrew/Cellar/openjdk@17/17.0.18/libexec/openjdk.jdk/Contents/Home mvn test -Dtest=DirectoryNamespaceTest#testOpenSpecificVersionDoesNotPassVersionToDescribeTable` from `java/`
- `UV_HTTP_TIMEOUT=120 uv run pytest python/tests/test_namespace_dir.py::test_dataset_namespace_open_does_not_pass_version_to_describe_table` from `python/`
- `JAVA_HOME=/opt/homebrew/Cellar/openjdk@17/17.0.18/libexec/openjdk.jdk/Contents/Home ./mvnw spotless:check` from `java/`
- `UV_HTTP_TIMEOUT=120 uv run ruff format --check --diff python` from `python/`
- `UV_HTTP_TIMEOUT=120 uv run ruff check python` from `python/`
- `git diff --check`
- `UV_HTTP_TIMEOUT=120 uv run make lint` from `python/` (fails in pyright because local optional imports `tensorflow` and `torch` are not installed; `ruff format --check` and `ruff check` passed)

### DIFF
--- a/python/python/lance/__init__.py
+++ b/python/python/lance/__init__.py
@@ -230,7 +230,9 @@ def dataset(
                 "Both 'namespace_client' and 'table_id' must be provided together."
             )
 
-        request = DescribeTableRequest(id=table_id, version=version)
+        # Resolve the latest table metadata here. The requested dataset version is
+        # applied by the lower-level dataset open path after namespace resolution.
+        request = DescribeTableRequest(id=table_id, version=None)
         response = namespace_client.describe_table(request)
 
         uri = response.location
--- a/java/src/main/java/org/lance/OpenDatasetBuilder.java
+++ b/java/src/main/java/org/lance/OpenDatasetBuilder.java
@@ -216,8 +216,8 @@ public class OpenDatasetBuilder {
     // Call describe_table to get location and storage options
     DescribeTableRequest request = new DescribeTableRequest();
     request.setId(tableId);
-    // Only set version if present
-    options.getVersion().ifPresent(v -> request.setVersion(Long.valueOf(v)));
+    // Do not set the dataset version here. Some namespace implementations only support describing
+    // the latest table metadata; the requested version is applied when opening the dataset below.
 
     DescribeTableResponse response = namespaceClient.describeTable(request);
 
--- a/java/src/test/java/org/lance/namespace/DirectoryNamespaceTest.java
+++ b/java/src/test/java/org/lance/namespace/DirectoryNamespaceTest.java
@@ -190,6 +190,33 @@ public class DirectoryNamespaceTest {
   }
 
   @Test
+  void testOpenSpecificVersionDoesNotPassVersionToDescribeTable() throws Exception {
+    VersionRejectingNamespace versionRejectingNamespace =
+        new VersionRejectingNamespace(innerNamespaceClient);
+    namespaceClient = versionRejectingNamespace;
+    List<String> tableId = Arrays.asList("test_table");
+
+    namespaceClient.createTable(new CreateTableRequest().id(tableId), createTestTableData());
+    namespaceClient.insertIntoTable(
+        new InsertIntoTableRequest().id(tableId).mode("append"), createTestTableData());
+
+    try (Dataset versionOne =
+        Dataset.open()
+            .allocator(allocator)
+            .namespaceClient(namespaceClient)
+            .tableId(tableId)
+            .readOptions(new ReadOptions.Builder().setVersion(1L).build())
+            .build()) {
+      assertEquals(1, versionOne.version());
+      assertEquals(3, versionOne.countRows());
+    }
+
+    assertTrue(
+        versionRejectingNamespace.getDescribeTableCallCount() > 0,
+        "Expected describeTable to be called when opening through namespace");
+  }
+
+  @Test
   void testCreateAndListNamespaces() {
     // Create a namespace
     CreateNamespaceRequest createReq = new CreateNamespaceRequest().id(Arrays.asList("workspace"));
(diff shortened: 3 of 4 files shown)


python/python/lance/__init__.py
yanghua, line 139: Can we also update here?

yanghua (commented):
Can we also add a test for these change?

FANNG1:
Added Java and Python regression tests that verify namespace `describe_table` / `describeTable` does not receive the requested dataset version, while the actual dataset open still returns the requested version.

python/python/lance/__init__.py
FANNG1: The public behavior of `version` does not change here: it still opens the requested dataset version.

This change only adjusts the internal namespace resolution flow. We now use `describe_table` to resolve table metadata/location/storage options, and apply `version` later in the lower-level dataset open path. I'm not sure this internal detail should be exposed in the public parameter docs.

Do you think the current `version` doc is misleading for users, or were you suggesting documenting this namespace-specific implementation detail?

yanghua: > and apply version later in the lower-level dataset open path

Can you share more details about this?

FANNG1: Currently some namespace implementations can support describing a specific table version in `describe_table`, but not all implementations do. For example, REST-backed namespace implementations may only support resolving the current table metadata/location/storage options from `describe_table`. This is the background of lance-format/lance-spark#609.

This PR avoids requiring every namespace implementation to support versioned `describe_table` for the dataset-open path. We only use `describe_table` to resolve the table location and storage options. The requested dataset version is still passed to the actual dataset open path afterward.

So this keeps compatibility with namespace implementations that support only latest-table describe, while preserving the user-facing behavior of `version`: the opened dataset is still the requested version.

yanghua: Got it. Thanks for explaining.

yanghua (approved):
LGTM.

FANNG1:
> LGTM.

Thanks for the review!

PR #7256, opened 2026-06-12
fix(mem_wal): push LIMIT/OFFSET down to fresh-tier scan sources

hamersaw:
## Summary

The LSM scan planner (`LsmScanPlanner`) applied a query's `LIMIT`/`OFFSET` *above* the per-source union, so every flushed generation and the base table were fully scanned before the limit was imposed at the top. For a bounded read over a fresh tier with many flushed generations this materialized the entire tier on every query.

This threads an `overfetch_factor` through `LsmScanner` → `LsmScanPlanner` and pushes a per-source fetch of `(offset + limit) * overfetch` into each `Base`/`Flushed` source scan. The active memtable is exempt (its within-source dedup needs the full match set). Block-listed sources over-fetch by `overfetch_factor` so cross-generation PK dedup still leaves ~`limit` live rows; a `LocalLimit` re-imposes the exact `skip`/`fetch` cap above the merge. Unbounded reads (no limit) are unchanged.

## Changes
- `scanner/planner.rs`: `overfetch_factor` field + `with_overfetch_factor` on `LsmScanPlanner`; per-source limit pushdown in `plan_scan`; `fetch` parameter on `build_source_scan` (pushed into Base/Flushed scans, ActiveMemTable exempt).
- `scanner/builder.rs`: `overfetch_factor` on `LsmScanner` + `with_overfetch_factor`, threaded into both the scan and FTS planners.

## Validation
Validated end-to-end against a WAL benchmark on minikube with object storage behind a latency proxy: a bounded `offset+limit` read over an 18-generation fresh tier dropped from a full-tier scan (~360 generation scans) to a bounded per-source read.

### DIFF
--- a/rust/lance/src/dataset/mem_wal/scanner/builder.rs
+++ b/rust/lance/src/dataset/mem_wal/scanner/builder.rs
@@ -125,6 +125,9 @@ pub struct LsmScanner {
     /// Cache of opened flushed-generation datasets. When set, repeated
     /// queries against the same generation skip the manifest read entirely.
     flushed_cache: Option<Arc<FlushedMemTableCache>>,
+    /// Over-fetch multiple for block-listed sources in search plans
+    /// (see [`super::LsmFtsSearchPlanner::with_overfetch_factor`]).
+    overfetch_factor: Option<f64>,
 }
 
 impl LsmScanner {
@@ -160,6 +163,7 @@ impl LsmScanner {
             pk_columns,
             session,
             flushed_cache: None,
+            overfetch_factor: None,
         }
     }
 
@@ -198,6 +202,7 @@ impl LsmScanner {
             pk_columns,
             session: None,
             flushed_cache: None,
+            overfetch_factor: None,
         }
     }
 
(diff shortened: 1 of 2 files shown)


rust/lance/src/dataset/mem_wal/scanner/builder.rs
```diff
@@ -125,6 +125,9 @@ pub struct LsmScanner {
     /// Cache of opened flushed-generation datasets. When set, repeated
     /// queries against the same generation skip the manifest read entirely.
     flushed_cache: Option<Arc<FlushedMemTableCache>>,
+    /// Over-fetch multiple for block-listed sources in search plans
+    /// (see [`super::LsmFtsSearchPlanner::with_overfetch_factor`]).
+    overfetch_factor: Option<f64>,
```
LuQQiu, line 130: We call it refine factor?

LuQQiu: oh better to understand whether it's similar to refine factors in vector / fts
not saying we should use refine factor

LuQQiu (approved):

rust/lance/src/dataset/mem_wal/scanner/builder.rs
hamersaw: Great point. I'm trying to keep these separate because they are logically different things. IIUC the `refine_factor` is used because IVF distances may not be exact, so we over retrieve so that we can become more accurate. The proposed `overfetch_factor` is strictly for probabilistically ensuring we have K values after we remove duplicates. This is a simple solution for the problem, but ultimately we will probably want to fix it in another way. So by separating them, we can remove the `overfetch_factor` more easily.

PR #7257, opened 2026-06-12
perf(mem_wal): parallelize fresh-tier source planning and execution

hamersaw:
## Summary

The LSM FTS and vector search planners (`LsmFtsSearchPlanner`, `LsmVectorSearchPlanner`) built each source's plan in a sequential `for` loop and unioned the arms under a single `SortPreservingMergeExec`. The merge polls every union arm from one task, so per-arm CPU (posting/index decode, BM25 and distance scoring) serialized even though the underlying IO awaits interleave — wall time grew linearly with the flushed-generation count.

This:
- builds the per-source plans concurrently with `try_join_all` (FTS + vector),
- runs the cross-source block-list PK hashing concurrently (`block_list.rs`),
- wraps the union in a round-robin `RepartitionExec` via a new `spawn_union_arms` helper so each arm gets its own driver task.

Rows stay disjoint across partitions, so the per-partition TopK + sort-preserving merge semantics are unchanged.

## Changes
- `scanner/exec.rs`: `spawn_union_arms` helper (round-robin repartition over the union).
- `scanner/fts_search.rs`, `scanner/vector_search.rs`: concurrent per-source plan builds + `spawn_union_arms` over the union.
- `scanner/block_list.rs`: concurrent flushed-generation PK-hash loads.

## Validation
Validated end-to-end against a WAL FTS benchmark on minikube with object storage behind a 10ms/GET latency proxy. Read latency over a fresh tier as a function of flushed-generation count, p50:

| generations | before | after |
|---|---|---|
| 2  | 1,164ms | 565ms |
| 5  | 1,983ms | 610ms |
| 10 | 3,585ms | 660ms |
| 18 | 6,071ms | 759ms |

Per-generation slope dropped from ~290ms/gen to ~12ms/gen.

### DIFF
--- a/rust/lance/src/dataset/mem_wal/scanner/exec.rs
+++ b/rust/lance/src/dataset/mem_wal/scanner/exec.rs
@@ -28,3 +28,22 @@ pub use pk::{
 };
 pub use pk_hash_filter::PkHashFilterExec;
 pub use within_source_dedup::{DedupDirection, WithinSourceDedupExec};
+
+/// Give each union arm its own driver task. `UnionExec` exposes one
+/// partition per arm, but a downstream `SortPreservingMergeExec` polls them
+/// all from a single task, serializing per-arm CPU (index decode, scoring)
+/// even though the IO awaits interleave. Round-robin repartitioning spawns
+/// one pull task per input partition; rows stay disjoint across partitions,
+/// so per-partition TopK + merge semantics are unchanged.
+pub fn spawn_union_arms(
+    union: std::sync::Arc<dyn datafusion::physical_plan::ExecutionPlan>,
+) -> lance_core::Result<std::sync::Arc<dyn datafusion::physical_plan::ExecutionPlan>> {
+    use datafusion::physical_plan::repartition::RepartitionExec;
+    let n = union.properties().partitioning.partition_count();
+    let repart = RepartitionExec::try_new(
+        union,
+        datafusion::physical_plan::Partitioning::RoundRobinBatch(n),
+    )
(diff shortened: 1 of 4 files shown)


rust/lance/src/dataset/mem_wal/scanner/exec.rs
```diff
@@ -28,3 +28,22 @@ pub use pk::{
...
+/// all from a single task, serializing per-arm CPU (index decode, scoring)
+/// even though the IO awaits interleave. Round-robin repartitioning spawns
+/// one pull task per input partition; rows stay disjoint across partitions,
+/// so per-partition TopK + merge semantics are unchanged.
+pub fn spawn_union_arms(
+    union: std::sync::Arc<dyn datafusion::physical_plan::ExecutionPlan>,
+) -> lance_core::Result<std::sync::Arc<dyn datafusion::physical_plan::ExecutionPlan>> {
+    use datafusion::physical_plan::repartition::RepartitionExec;
+    let n = union.properties().partitioning.partition_count();
+    let repart = RepartitionExec::try_new(
```
LuQQiu: `RepartitionExec` may be a little heavy to solve concurrency issue? it's shuffle
does it improve the result a lot?
If we already do per source topK (seems like, could you confirm?), the final result candidates may not be that many? and can use `SortExec` directly?

hamersaw: Your right, the `RepartitionExec` is a bit heavy for what we actually want to achieve here. I removed this and am now relying on the underlying constructs to give us execution parallelism.

LuQQiu (approved):

PR #7259, opened 2026-06-12
feat(java): expose ArrowArrayStream export on LanceScanner

sezruby:
## Summary

Add `LanceScanner#exportArrowStream(ArrowArrayStream)` — a public wrapper around the existing private native `openStream(long)` JNI call. Lets callers populate a stream they allocated themselves instead of going through `scanBatches()`, which immediately imports the result into a Java `ArrowReader` backed by Lance's `BufferAllocator`.

## Why

Consumers loaded under a different classloader and/or pinned to a different Apache Arrow version cannot safely share `org.apache.arrow.vector.*` classes with Lance — the JVM treats them as distinct types even when the bytecode is identical. The C Data Interface struct is stable across Arrow versions, so handing the C struct's memory address across the boundary is the only correct integration shape.

A concrete consumer is the gluten-spark / Velox integration tracked at [apache/gluten#12263](https://github.com/apache/gluten/issues/12263). gluten-spark builds against Arrow 15 (matching what Spark 3.5 ships and Velox uses); Lance Java SDK is on Arrow 18. With this method, gluten can:

```java
try (ArrowArrayStream stream = ArrowArrayStream.allocateNew(glutenAllocator)) {
  scanner.exportArrowStream(stream);
  try (ArrowReader reader = Data.importArrayStream(glutenAllocator, stream)) {
    // import each batch into Velox via gluten's own Arrow 15 stack
  }
}
```

…where `glutenAllocator` is a Spark-task-managed `BufferAllocator` (`ArrowReservationListener` plumbing for memory accounting). Lance never sees Java Arrow on this side; ownership stays with the caller via the C Data Interface release callback.

## What changed

- `LanceScanner#exportArrowStream(ArrowArrayStream)` — new public method, ~7 lines + Javadoc with usage example. Mirrors the body of `scanBatches()` minus the local stream allocation and the `Data.importArrayStream` step.
- No native code touched. The underlying JNI hook already existed; it was just not reachable from outside the class.
- Test `testDatasetScannerExportArrowStream` exercises the full path: caller allocates the C stream from its own `RootAllocator`, scanner fills the C struct, caller imports into an `ArrowReader` and validates batch contents (40 rows over 2 batches of 20).

## Backwards compatibility

Pure addition. `scanBatches()`, `schema()`, `countRows()`, `getStats()`, `close()` all unchanged. No native ABI change.

## Test plan

- `./mvnw test -Dtest=ScannerTest#testDatasetScannerExportArrowStream` — passes locally (Java compile + spotless clean; full test run depends on a working `lance-jni` Rust build, which had an unrelated `aws-smithy-types` registry issue on my machine, so I'm relying on CI for the JNI-linked verification).
- Existing `testDatasetScannerColumns` covers the `scanBatches()` path so any regression in the shared `openStream` JNI call would surface there.


sezruby:
@hamersaw @jackye1995 Could you review the PR? The PR is to support Lance Reader in Gluten/Spark. I'll open a lance-spark PR after this PR is merged. Thanks!

sezruby:
Gentle ping — @wjones127 would you have a chance to take a look? Small Java-only change exposing the existing `openStream` JNI as a public `exportArrowStream(ArrowArrayStream)` on `LanceScanner`, same shape as the recent Java API exposures. (CI red is the unrelated `aws-smithy-types`/`time` E0119 issue from #7255, not this change.) Thanks!

java/src/main/java/org/lance/ipc/LanceScanner.java
```diff
@@ -146,6 +146,44 @@ public ArrowReader scanBatches() {
...
+   *       // ...
+   *     }
+   *   }
+   * }
+   * }</pre>
+   *
+   * @param stream the caller-allocated stream to populate
+   * @throws IOException if the native scan fails to start
+   */
+  public void exportArrowStream(ArrowArrayStream stream) throws IOException {
```
Xuanwo: The public API still requires Lance's `ArrowArrayStream` Java type, so callers using a different Arrow version or classloader cannot pass their own stream despite this being the integration boundary the method documents.

java/src/main/java/org/lance/ipc/LanceScanner.java
```diff
@@ -146,6 +146,44 @@ public ArrowReader scanBatches() {
...
+    Preconditions.checkNotNull(stream);
+    try (LockManager.ReadLock readLock = lockManager.acquireReadLock()) {
+      Preconditions.checkArgument(nativeScannerHandle != 0, "Scanner is closed");
+      openStream(stream.memoryAddress());
```
Xuanwo: This writes into the provided C stream without rejecting an already-populated stream, so a second export can overwrite the existing release callback and leak the first producer.

java/src/test/java/org/lance/ScannerTest.java
```diff
@@ -158,6 +160,51 @@ void testDatasetScannerSchema(@TempDir Path tempDir) throws Exception {
...
+            try (ArrowReader reader = Data.importArrayStream(allocator, stream)) {
+              VectorSchemaRoot root = reader.getVectorSchemaRoot();
+              int index = 0;
+              while (reader.loadNextBatch()) {
+                List<FieldVector> fieldVectors = root.getFieldVectors();
+                assertEquals(1, fieldVectors.size());
+                FieldVector fieldVector = fieldVectors.get(0);
+                assertEquals(
+                    ArrowType.ArrowTypeID.Int, fieldVector.getField().getType().getTypeID());
+                assertEquals(batchRows, fieldVector.getValueCount());
```
Xuanwo: This test assumes every non-strict scan batch has exactly the requested size, so a valid scanner change that returns smaller batches would fail this API test.

java/src/main/java/org/lance/ipc/LanceScanner.java
sezruby: Good catch — this defeated the method's whole purpose. Changed the signature to take a raw `long streamAddress` instead of a typed `ArrowArrayStream`. The caller now allocates the stream with its own Arrow runtime and passes `stream.memoryAddress()`, so only the address crosses into Lance and the two sides never share a Java Arrow type. Documented the reasoning in the Javadoc. Fixed in f725d96.

java/src/main/java/org/lance/ipc/LanceScanner.java
sezruby: Fixed. `inner_open_stream` now reads the existing struct's `release` field and rejects with `IllegalArgumentException` if it's already set (also rejects a null address). A freshly-allocated `ArrowArrayStream` has a null `release` per the C Data Interface, so that's the "empty" contract. The read is done unaligned via `addr_of!` without forming a reference, since the struct lives in a possibly-unaligned ArrowBuf (same reason the write below uses `write_unaligned`). Added a test that exports twice into the same stream and asserts the second call throws while the first producer stays intact and drainable. Fixed in f725d96.

java/src/test/java/org/lance/ScannerTest.java
sezruby: Agreed, that was brittle. Reworked the assertions into a shared `drainIdStream` helper that asserts only `batch size <= requested size` and that the full row set is correct, without assuming any batch count or that batches are full. Applied across all the new cases. Fixed in f725d96.

java/lance-jni/src/blocking_scanner.rs
```diff
@@ -589,6 +589,37 @@ pub extern "system" fn Java_org_lance_ipc_LanceScanner_openStream(
...
+    // earlier export and not yet released), overwriting it would drop that callback and leak the
+    // first producer's resources. A freshly-allocated `ArrowArrayStream` has a null `release`, per
+    // the Arrow C Data Interface, so requiring `release == None` is the contract for "empty".
+    //
+    // The struct is allocated by Arrow Java inside an ArrowBuf and is not guaranteed to be aligned
+    // (hence `write_unaligned` below), so we must not form a reference to it. We read only the
+    // `release` field through an unaligned read: `addr_of!` computes the field address without
+    // creating an intermediate (mis)aligned reference, and the field is an `Option<fn>` which is
+    // `Copy` with no destructor, so reading a copy of it leaves the caller's stream untouched.
+    let release_is_set = unsafe {
```
Xuanwo, line 611: The populated-stream guard checks the release callback before constructing and writing the new stream, so two concurrent exports to the same empty stream can both pass the check and then overwrite each other. That can leak the first producer and leave the caller draining whichever stream won the race.

java/src/test/java/org/lance/ScannerTest.java
```diff
@@ -158,6 +161,280 @@ void testDatasetScannerSchema(@TempDir Path tempDir) throws Exception {
...
+   * <p>Asserts the projected schema is exactly {@code id: int32} and that no batch exceeds {@code
+   * maxBatchRows}, but does not assume any particular batch count or that batches are full — batch
+   * size is a scanner hint, not a guarantee, so over-asserting on it makes the test brittle.
+   */
+  private static List<Integer> drainIdStream(
+      BufferAllocator allocator, ArrowArrayStream stream, int maxBatchRows) throws IOException {
+    List<Integer> ids = new ArrayList<>();
+    try (ArrowReader reader = Data.importArrayStream(allocator, stream)) {
+      VectorSchemaRoot root = reader.getVectorSchemaRoot();
+      while (reader.loadNextBatch()) {
```
Xuanwo, line 199: The schema assertions only run after a batch is loaded, so the empty-result case never validates the exported stream schema. A regression that returns the wrong schema for zero-row scans would still pass.

java/src/test/java/org/lance/ScannerTest.java
```diff
@@ -158,6 +161,280 @@ void testDatasetScannerSchema(@TempDir Path tempDir) throws Exception {
...
+            dataset.newScan(
+                new ScanOptions.Builder()
+                    .batchSize(batchRows)
+                    .columns(Arrays.asList("id"))
+                    .build())) {
+          try (ArrowArrayStream stream = ArrowArrayStream.allocateNew(allocator)) {
+            scanner.exportArrowStream(stream.memoryAddress());
+            List<Integer> ids = drainIdStream(allocator, stream, batchRows);
+            assertEquals(totalRows, ids.size());
+            Collections.sort(ids);
```
Xuanwo: The multi-fragment export test sorts the collected ids before asserting them, so it would pass even if the default ordered scan returned fragments out of order.

java/lance-jni/src/blocking_scanner.rs
sezruby: You're right that the check-then-write isn't atomic against a concurrent second export into the same stream. I don't think this one can be fixed on the Lance side, though: the `ArrowArrayStream` is a plain C struct in caller-owned memory with no synchronization, so two concurrent exports into the same struct are a data race regardless of what the guard does — the same constraint Arrow's C Data Interface itself imposes (a stream is single-producer, drained by one consumer). The guard's purpose is to catch the *sequential* "export twice into the same stream" mistake, which it does.

Rather than imply a safety we can't provide, I documented the contract on `exportArrowStream`: the stream must not be shared across concurrent exports; use a separate stream per export. (`scanBatches()` has the same shape — it allocates a fresh stream each call.) Let me know if you'd prefer a different framing. Updated in bc42f5c6.

java/src/test/java/org/lance/ScannerTest.java
sezruby: Good catch. Moved the schema assertions in the `drainIdStream` helper to run on the imported reader *before* the first `loadNextBatch()`, so they execute even when the scan returns zero batches — the empty-result test now validates the zero-row schema. (`ArrowReader#getVectorSchemaRoot()` exposes the schema as soon as the stream is imported, before any batch is loaded.) Fixed in bc42f5c6.

java/src/test/java/org/lance/ScannerTest.java
sezruby: Agreed — sorting defeated the point of a multi-fragment ordering test. Removed `Collections.sort` from both the multi-fragment and filter tests; they now assert the exact scan order (`0..39` / `0..19`), matching how the existing `validateScanResults` asserts order without sorting.

While here I also removed an unsound assertion in the shared helper that assumed every batch was <= the requested batch size — the export tests scan with the default (non-strict) batch size, which only guarantees the total row count, so that was the same brittleness in another spot. Added `testExportArrowStreamStrictBatchSize` to cover the per-batch-size contract properly via `strictBatchSize(true)`, plus `testExportArrowStreamPreservesNulls` for null round-tripping. Fixed in bc42f5c6.

sezruby:
@Xuanwo all review comments addressed across `f725d96` and `bc42f5c6`, CI green. Mind taking another look when you have a chance? (The concurrency one I handled via docs rather than code — flag me if you'd prefer otherwise.)

sezruby:
Friendly ping for review — this has been idle since late June and CI is green.

**What it does:** additive Java change (+540/−0, 3 files) exposing `LanceScanner#exportArrowStream(long streamAddress)`, so a caller can populate its own Arrow C Data Interface stream. This lets a consumer on a different Arrow version/classloader integrate without sharing `org.apache.arrow.vector.*` types. Downstream consumer: gluten-spark/Velox (apache/gluten#12263).

**State:** all of @Xuanwo's feedback is addressed — signature takes a raw `long` address; `inner_open_stream` rejects an already-populated (non-null `release`) or null stream; test assertions relaxed to `batch <= requested` with full-rowset checks; schema validated before the first `loadNextBatch()` so zero-row scans are covered; multi-fragment/filter tests assert exact scan order. One open thread remains on `blocking_scanner.rs`: the check-then-write on a caller-owned C struct isn't atomic against two concurrent exports into the same stream — since the struct is plain caller-owned memory, this is a caller contract rather than something synchronizable on the Lance side, and I'm happy to document it explicitly.

@wjones127 @jackye1995 @yanghua — would any of you have a few minutes to take a look? It's Arrow C Data Interface plus the Gluten integration, so your review would be a great fit. Thanks!

java/src/test/java/org/lance/ScannerTest.java
```diff
@@ -158,6 +162,450 @@ void testDatasetScannerSchema(@TempDir Path tempDir) throws Exception {
...
+   * number of batches and the rows per batch are not part of the contract being tested here; that
+   * dimension is covered separately by {@link #testExportArrowStreamStrictBatchSize}. Row ordering
+   * and exact values are asserted by the callers against the returned list.
+   */
+  private static List<Integer> drainIdStream(BufferAllocator allocator, ArrowArrayStream stream)
+      throws IOException {
+    List<Integer> ids = new ArrayList<>();
+    try (ArrowReader reader = Data.importArrayStream(allocator, stream)) {
+      VectorSchemaRoot root = reader.getVectorSchemaRoot();
+      List<Field> fields = root.getSchema().getFields();
```
yanghua, line 187: The comment for `drainIdStream` specifies that the exported stream schema should be `id: int32`, but currently only the field count, field names and `ArrowTypeID.Int` are checked. If an empty‑result stream is incorrectly exported as `id: int64` or with wrong nullability, `testExportArrowStreamEmptyResult` may still pass, leaving regressions for empty‑scan schemas uncovered. It is recommended to compare the full
Schema directly, or at least assert the bit width, signedness and nullability of `ArrowType.Int`.

sezruby: Good catch. You're right that `ArrowTypeID.Int` doesn't pin the bit width or signedness, and the `(IntVector)` cast that implicitly enforces int32 only runs for non-empty batches — so an empty scan exporting the wrong width or nullability would have slipped through `testExportArrowStreamEmptyResult`.

Fixed in f0c6a7c: `drainIdStream` now asserts the `id` field is nullable, bit width 32, and signed, so the schema is checked before the first `loadNextBatch()` for empty scans too. Thanks for the review!

yanghua (approved):
LGTM. @sezruby thanks for your contribution! cc @Xuanwo

sezruby:
Thanks for the review and approval, @yanghua. This is now approved, all 11 checks are green, and it's mergeable; the gatekeeper gate also recommended approve with no changes requested.

Could this be merged when you have a moment? It's the base Arrow C Data Interface export API that the downstream lance-spark and Gluten integration (apache/gluten#12263) build on, and there are follow-up PRs on that side that consume `exportArrowStream` directly — landing this lets them move forward.

cc @wjones127 @jackye1995 @Xuanwo in case anyone else has a moment to merge.

Issue #2606, opened 2024-07-17
Better determination of FSST output buffer size


westonpace:
Currently, the FSST output buffer is initialized to `8 * compressed_length`.  This should be safe (no FSST symbol is more than 8 bytes I think) but it is slow and excessive.  We can store the original length of the string as part of the encoded page and use that to size the output buffer exactly.

broccoliSpicy:
One idea is to store the original string size inside some header(maybe inside symbol table) and allocate the exact size before decompression

broccoliSpicy:
in our fsst implementation, symbol table and decompress output is decoupled, the user is responsible to associate which symbol table with which decompressed output, maybe we can have some code in both symbol table and decompressed output for matching,

broccoliSpicy:
> This should be safe (no FSST symbol is more than 8 bytes I think) but it is slow and excessive.

yes, the FSST symbol is of size [1, 8]

westonpace:
> One idea is to store the original string size inside some header(maybe inside symbol table) and allocate the exact size before decompression

That could work, I think that's fine for the FSST crate.  We always have encodings.proto too.  I don't see any big pros/cons to either approach so feel free to pick one.

broccoliSpicy:
ha @westonpace , I am starting to code `store the original size` and found a problem:
we are decoding it in decode batch size, not the original compression size, so the original size is a far overshot than `8 * compressed_length"

I checked the original implementation, they just allocate a big chunk of memory for decompression
```c++
#define FSST_MEMBUF (1ULL<<22)

---

   dstMem[0] = srcBuf[1] + (FSST_MEMBUF*(1ULL+decompress));
   dstMem[1] = dstMem[0] + (FSST_MEMBUF*(2ULL-decompress));

---

    dstLen[swap] = fsst_decompress(&decoder, srcLen[swap] - hdr, srcBuf[swap] + hdr, FSST_MEMBUF, dstBuf[swap] = dstMem[swap]);

```
[original implementation](https://github.com/cwida/fsst/blob/ef52cb324a661e4e76e5eff10ac6452232d852c0/fsst.cpp#L166-L167)

let me check out duckdb...

broccoliSpicy:
[...]

PR #7267, opened 2026-06-13
fix(fsst): grow decode buffer to prevent out-of-bounds write

### LINKED ISSUE #7266: FSST decompress writes out of bounds when data compresses better than 3:1
## Description

`fsst::fsst::decompress` documents and enforces a minimum output buffer of `3 * input`, but that bound is unsound. An FSST code byte can decode to a symbol of up to 8 bytes, and the decode loop in `decompress_bulk` writes 8 bytes at a time through raw pointers (`out.as_mut_ptr().add(out_curr)`). When the data compresses better than 3:1, the decoded output is larger than the buffer and the writes run past the end of the allocation.

A caller who allocates the documented minimum gets a heap out-of-bounds write. In-tree callers happen to allocate `8 * input`, so they avoid it, but the public API contract is wrong.

Related: #2606 (better determination of FSST output buffer size).

## Steps to Reproduce

Compress a highly repetitive input (so FSST builds 8-byte symbols), then decompress into the documented 3x buffer:

```rust
let row = "a".repeat(64);
let rows: Vec<&str> = (0..2000).map(|_| row.as_str()).collect();
let array = arrow_array::StringArray::from(rows);
let in_data = array.value_data();
let in_offsets = array.value_offsets();

let mut comp = vec![0u8; in_data.len()];
let mut comp_off = vec![0i32; in_offsets.len()];
let mut st = [0u8; FSST_SYMBOL_TABLE_SIZE];
compress(&mut st, in_data, in_offsets, &mut comp, &mut comp_off).unwrap();

// Documented minimum: 3x the compressed input.
let mut out = vec![0u8; 3 * comp.len()];
let mut out_off = vec![0i32; comp_off.len()];
decompress(&st, &comp, &comp_off, &mut out, &mut out_off).unwrap();
```

## Expected Behavior

Decompression does not write past the output buffer regardless of compression ratio.

## Actual Behavior

`STATUS_ACCESS_VIOLATION` (segfault) on Windows; an out-of-bounds write under Miri/ASan. With a generous output buffer the measured ratio is 8.0 (compressed 16000 bytes, decoded 128000 bytes), against the enforced minimum of 48000.

## Environment

- crate: `fsst` (`rust/compression/fsst`)
- branch: `main`

Noethix55555:
## Summary

`fsst::fsst::decompress` only required a `3 * input` output buffer, but an FSST code byte can decode to an 8-byte symbol and the decode loop writes 8 bytes at a time through raw pointers. On data that compresses better than 3:1, honoring the documented minimum overran the buffer (a heap out-of-bounds write, observed as a segfault).

This sizes the output buffer for the worst case before decoding, then shrinks it to the real length at the end. The change is internal to the decode path and does not change results for callers that already pass a large enough buffer (in-tree callers pass `8 * input`).

Closes #7266. Related: #2606.

## Test

Added `rust/compression/fsst/tests/buffer_bound.rs`: it compresses an ~8:1 input and decompresses into the documented 3x buffer. This segfaulted before the change and round-trips correctly after. `cargo test -p fsst` and `cargo clippy -p fsst --tests` pass.

### DIFF
--- a/rust/compression/fsst/src/fsst.rs
+++ b/rust/compression/fsst/src/fsst.rs
@@ -812,6 +812,19 @@ fn decompress_bulk<T: OffsetSizeTrait>(
 ) -> io::Result<()> {
     let symbols = decoder.symbols;
     let lens = decoder.lens;
+
+    // A code byte can decode to an 8-byte symbol and the loop below writes 8 bytes
+    // at a time via raw pointers, so size the buffer for the worst case (it is
+    // shrunk to the real size at the end).
+    let max_decoded = compressed_strs
+        .len()
+        .saturating_mul(MAX_SYMBOL_LENGTH)
+        .saturating_add(MAX_SYMBOL_LENGTH);
+    let needed = out_pos.saturating_add(max_decoded);
+    if out.len() < needed {
+        out.resize(needed, 0);
+    }
+
     let mut decompress = |mut in_curr: usize, in_end: usize, out_curr: &mut usize| {
         // Do SIMD operation here by 4 bytes
         while in_curr + 4 <= in_end {
--- a/rust/compression/fsst/tests/buffer_bound.rs
+++ b/rust/compression/fsst/tests/buffer_bound.rs
@@ -1,0 +1,54 @@
+// Regression test for the out-of-bounds write when data compresses better than
+// 3:1 (a code byte can expand to an 8-byte symbol, but decompress only required a
+// 3x output buffer).
+use arrow_array::StringArray;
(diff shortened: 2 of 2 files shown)


Xuanwo (commented):
The decompression path still trusts serialized symbol lengths and offsets before entering unsafe decode. Malformed input can write past the allocated buffer or read outside the compressed data.

Xuanwo:
Closing this because #7589 was merged with the complete 8x FSST output-buffer contract, overflow checks, and regression coverage for the same out-of-bounds root cause. Please reopen if a distinct FSST decoding case remains.

PR #7269, opened 2026-06-14
feat: configure blob inline threshold per column

### LINKED ISSUE #7268: Make blob v2 inline threshold configurable per column
## Problem

Blob v2 currently uses a fixed 64 KiB inline cutoff. Values larger than this cutoff are written to packed `.blob` sidecar files unless they exceed the dedicated blob threshold.

The dedicated blob threshold can already be overridden per field via metadata, but the inline cutoff is not configurable. This makes it difficult to tune blob storage layout for columns with different access patterns or object-size distributions.

## Proposed solution

Add a per-column configuration mechanism for the blob v2 inline threshold.

This should allow callers to tune when a blob value stays inline versus when it is moved to packed blob storage, without changing the threshold globally for every blob column in the dataset.

## Notes

Current behavior:

- inline cutoff: fixed at 64 KiB
- dedicated threshold: defaults to 4 MiB and can be overridden with field metadata
- pack file size threshold: controls `.blob` sidecar rollover, not inline-vs-packed placement

## Acceptance criteria

- Blob v2 inline threshold can be configured per blob column.
- The default behavior remains unchanged for existing callers.
- The API clearly distinguishes inline cutoff from dedicated cutoff and pack file size threshold.
- Rust and Python write paths expose or preserve the configuration consistently.
- Tests cover default behavior, per-column override, and interaction with the dedicated threshold.

Xuanwo:
This adds per-column blob v2 inline threshold metadata so callers can choose when a blob column moves from inline data-file storage to packed sidecar storage, without changing the existing packed sidecar rolling option.

The threshold is stored on the blob field metadata, matching the existing dedicated blob threshold model. Existing blob columns keep their policy in the dataset schema; appends that explicitly provide different threshold metadata are rejected instead of silently ignoring the input schema. The Python and Rust helpers validate threshold values at the API boundary so invalid values do not silently fall back to defaults.

Closes #7268.

### DIFF
--- a/rust/lance/src/dataset/blob.rs
+++ b/rust/lance/src/dataset/blob.rs
@@ -19,7 +19,9 @@ use arrow_schema::DataType as ArrowDataType;
 use bytes::Bytes;
 use futures::stream::BoxStream;
 use futures::{FutureExt, StreamExt, TryStreamExt, stream};
-use lance_arrow::{BLOB_DEDICATED_SIZE_THRESHOLD_META_KEY, FieldExt};
+use lance_arrow::{
+    BLOB_DEDICATED_SIZE_THRESHOLD_META_KEY, BLOB_INLINE_SIZE_THRESHOLD_META_KEY, FieldExt,
+};
 use lance_io::object_store::{ObjectStore, ObjectStoreParams, ObjectStoreRegistry};
 use lance_io::scheduler::{FileScheduler, ScanScheduler, SchedulerConfig};
 use object_store::path::Path;
@@ -206,6 +208,7 @@ pub struct BlobPreprocessor {
     local_counter: u32,
     pack_writer: PackWriter,
     blob_v2_cols: Vec<bool>,
+    inline_thresholds: Vec<usize>,
     dedicated_thresholds: Vec<usize>,
     writer_metadata: Vec<HashMap<String, String>>,
     external_base_resolver: Option<Arc<ExternalBaseResolver>>,
@@ -325,6 +328,10 @@ impl BlobPreprocessor {
         let arrow_schema = arrow_schema::Schema::from(schema);
         let fields = arrow_schema.fields();
         let blob_v2_cols = fields.iter().map(|field| field.is_blob_v2()).collect();
+        let inline_thresholds = fields
+            .iter()
+            .map(|field| inline_threshold_from_metadata(field.as_ref()))
+            .collect();
         let dedicated_thresholds = fields
             .iter()
             .map(|field| dedicated_threshold_from_metadata(field.as_ref()))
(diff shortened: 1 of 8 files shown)


rust/lance/src/dataset/write.rs
```diff
@@ -170,6 +172,42 @@ fn validate_external_blob_write_params(params: &WriteParams) -> Result<()> {
...
+
+        for key in [
+            BLOB_INLINE_SIZE_THRESHOLD_META_KEY,
+            BLOB_DEDICATED_SIZE_THRESHOLD_META_KEY,
+        ] {
+            let Some(input_value) = input_field.metadata.get(key) else {
+                continue;
+            };
+
+            if dataset_field.metadata.get(key) != Some(input_value) {
```
BubbleCal: This compares raw metadata, so appending a schema that explicitly sets the default threshold to an older/default dataset with no key will fail even though the effective threshold is unchanged. Please compare normalized threshold values.

rust/lance/src/dataset/blob.rs
```diff
@@ -700,13 +709,20 @@ impl BlobPreprocessor {
     }
 }
 
+fn inline_threshold_from_metadata(field: &arrow_schema::Field) -> usize {
+    field
+        .metadata()
+        .get(BLOB_INLINE_SIZE_THRESHOLD_META_KEY)
+        .and_then(|value| value.parse::<usize>().ok())
```
BubbleCal: Invalid threshold metadata is silently ignored here. Since callers can set these keys directly on the schema, a bad value can be persisted while writes fall back to the default layout. Please reject invalid values on the write path.

rust/lance/src/dataset/write.rs
```diff
@@ -170,6 +175,74 @@ fn validate_external_blob_write_params(params: &WriteParams) -> Result<()> {
...
+    dataset_schema: &Schema,
+) -> Result<()> {
+    for input_field in &input_schema.fields {
+        let Some(dataset_field) = dataset_schema.field(&input_field.name) else {
+            continue;
+        };
+        let is_blob_v2_field = input_field
+            .metadata
+            .get(ARROW_EXT_NAME_KEY)
+            .or_else(|| dataset_field.metadata.get(ARROW_EXT_NAME_KEY))
```
BubbleCal: This only falls back to the dataset extension when the input extension key is absent. If the input field has a non-blob extension plus blob threshold metadata, append validation skips the threshold checks and silently drops the input metadata. Please treat either side being blob v2 as requiring validation.

rust/lance/src/dataset/write.rs
```diff
@@ -1257,7 +1334,7 @@ async fn open_writer_with_options(
                 source_store_registry,
```
BubbleCal, line 1334: This still validates thresholds only when the blob preprocessor is created. A zero-batch create/overwrite never reaches this path, so invalid blob threshold metadata can be committed in an empty dataset. Please validate the final schema before writing fragments.

Xuanwo: I think this is overprotected.

BubbleCal (approved):

PR #7272, opened 2026-06-14
fix: make FTS finalization idempotent

jackye1995:
## Summary

Make distributed FTS finalization idempotent and object-store friendly. The old finalize flow was not retry-safe because it moved build-only `part_<old_id>_*` partition files before writing the final `metadata.lance`; if the process stopped in that window, a retry saw no final metadata but the original source files had already been deleted.

The new flow treats `metadata.lance` as the commit marker and keeps all pre-commit partition sources under stable `staging/` paths. Finalization copies staged files to dense final names without deleting the staged sources or canonical final paths first, writes final metadata, then removes staging best-effort. Any leftover `staging/` objects are filtered from committed FTS `IndexMetadata.files`, so cleanup failures do not make transient files part of the committed index.

## Implementation

- Write distributed build-only partition data and per-partition metadata under stable `staging/part_<old_id>_*` paths.
- Keep the non-distributed builder path writing final `part_<id>_*` files and final `metadata.lance` directly, with no dense rewrite.
- Add `IndexStore::copy_index_file_to(source, dest, dest_store)` for source-preserving copy-to-new-name; `LanceIndexStore` implements it with object-store copy and supports nested relative paths like `staging/...`.
- Copy existing root partition files into staging for the public distributed-from-existing path, `from_existing_index(..., fragment_mask=Some(...))`, before advertising those partitions in staged metadata.
- During `merge_index_files`, discover staged metadata, sort original partition IDs, map them densely to `0..N`, copy `staging/part_<old_id>_*` to final `part_<new_id>_*` without pre-deleting final paths, and then write final `metadata.lance`.
- Delete staged partition data and staged per-partition metadata only after final metadata is written, and keep that cleanup best-effort.
- Filter leftover `staging/` files from FTS segment `IndexMetadata.files` during segment commit.
- On retry, existing final `metadata.lance` makes finalize a no-op; without it, staged sources remain authoritative and any partial final files are replaceable from staging.

Added regressions for staged distributed writes, distributed-from-existing finalize, partial final-file retry without root `part_*` deletion, preserving staging when final metadata writing fails, and excluding stale staging files from committed metadata. Validated locally with Rust fmt/tests/clippy, segmented FTS integration tests, Python extension build, the targeted Python progress test, and `uv run make lint-rust`.

### DIFF
--- a/rust/lance-index/src/scalar/inverted/builder.rs
+++ b/rust/lance-index/src/scalar/inverted/builder.rs
@@ -407,9 +407,19 @@ impl InvertedIndexBuilder {
     ) -> Result<Vec<IndexFile>> {
         let partition_id = self.next_partition_id() | self.fragment_mask.unwrap_or(0);
         builder.set_id(partition_id);
-        let files = builder.write(dest_store).await?;
+        let files = builder
+            .write_to(dest_store, self.partition_write_target())
+            .await?;
         self.new_partitions.push(partition_id);
         Ok(files)
+    }
+
+    fn partition_write_target(&self) -> PartitionWriteTarget {
+        if self.fragment_mask.is_some() {
+            PartitionWriteTarget::Staged
+        } else {
+            PartitionWriteTarget::Final
+        }
     }
 
     fn next_partition_id(&self) -> u64 {
@@ -523,7 +533,11 @@ impl InvertedIndexBuilder {
             if let Some(builder) = merged_tail_partitions {
                 self.new_partitions.push(builder.id());
                 let mut builder = builder;
-                files.extend(builder.write(dest_store.as_ref()).await?);
+                files.extend(
+                    builder
+                        .write_to(dest_store.as_ref(), self.partition_write_target())
+                        .await?,
+                );
             }
             log::info!("wait workers indexing elapsed: {:?}", start.elapsed());
             Result::Ok(files)
@@ -550,12 +564,16 @@ impl InvertedIndexBuilder {
             .await?;
             let mut builder = part.into_builder().await?;
             builder.remap(mapping).await?;
-            files.extend(builder.write(dest_store).await?);
+            files.extend(
+                builder
+                    .write_to(dest_store, self.partition_write_target())
+                    .await?,
+            );
         }
         if self.fragment_mask.is_none() {
             files.push(self.write_metadata(dest_store, &self.partitions).await?);
         } else {
-            // in distributed mode, the part_temp_metadata is written by the worker
+            // in distributed mode, the staged partition metadata is written by the worker
             for &partition_id in &self.partitions {
                 files.push(self.write_part_metadata(dest_store, partition_id).await?);
             }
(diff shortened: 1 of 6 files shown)


Xuanwo (commented):
Two comments on FTS finalization.

rust/lance-index/src/scalar/inverted/builder.rs
```diff
@@ -1914,89 +2027,36 @@ async fn merge_metadata_files(
...
-            }
-            return Err(Error::index(format!(
-                "Failed to rename {} to {}: {}",
-                temp_path, final_path, e
-            )));
+    for &(old_id, new_id) in &id_mapping {
+        for suffix in PARTITION_FILE_SUFFIXES {
+            let staged_path = staged_partition_file_path(old_id, suffix);
+            let final_path = partition_file_path(new_id, suffix);
+            let _ = store.delete_index_file(&final_path).await;
```
Xuanwo: This is still unsafe with concurrent finalizers. `metadata.lance` is only checked at entry, so another finalizer may commit it before this one reaches `delete_index_file(final_path)`. If this task then crashes after the delete, retries will no-op on the committed metadata while a final `part_*` file is missing.

Can we avoid deleting canonical final files, or re-check the commit marker before any destructive final-path operation?

rust/lance-index/src/scalar/inverted/builder.rs
```diff
@@ -2023,10 +2083,15 @@ async fn merge_metadata_files(
     progress.stage_progress("write_merged_metadata", 1).await?;
     progress.stage_complete("write_merged_metadata").await?;
 
-    // Cleanup partition metadata files
+    // Cleanup staged partition metadata files
```
Xuanwo, line 2085: Ignoring staging cleanup errors can commit `staging/...` into `IndexMetadata.files`. The commit path finalizes, then lists the whole index dir, and that listing does not filter staging files.

Can we filter `staging/`, return the canonical final file list from finalize, or fail the commit when cleanup fails?

jackye1995: Addressed by filtering `staging/` from FTS segment `IndexMetadata.files` after finalization. Cleanup remains best-effort, but stale staged objects are treated as transient and are not recorded in committed metadata. I also extended `test_commit_existing_index_segments_finalizes_fts_segments` to create a leftover `staging/orphan.lance` after final metadata exists and assert the committed file list excludes it.

rust/lance-index/src/scalar/inverted/builder.rs
jackye1995: Fixed in `e9ccb116a` by removing the pre-copy delete of canonical final `part_<id>_*` files. I agree full atomic multi-object finalize is not possible with the current object-store abstraction, so the guarantee here is idempotence: concurrent/retry finalizers copy deterministic staged bytes to the same final names, `metadata.lance` remains the commit marker, and post-commit cleanup only touches `staging/`.

I also extended `test_merge_index_files_rewrites_partial_final_files_from_staging` so it still proves partial final files are repaired from staging, and now asserts merge does not delete root `part_*` files first.

@Xuanwo could you take another look?

Xuanwo (approved):
Thank you working for this!

PR #7273, opened 2026-06-15
perf(index): set ivf_rq target partition size to 4096

BubbleCal:

IVF_RQ currently falls through to the generic vector index target partition size of 8192. Recent IVF_RQ tuning indicates a smaller default partition target is preferred for this index type.

This PR gives `IndexType::IvfRq` an explicit `target_partition_size()` default of 4096. IVF_RQ index creation paths that do not set `target_partition_size` will now derive the number of IVF partitions from 4096 instead of the generic fallback.

It also adds a focused unit test so the IVF_RQ default cannot silently fall back to 8192 again.

No new benchmark run in this PR. This only changes the configured default and protects it with a unit test.

## Verification

- `cargo fmt --all`
- `cargo test -p lance-index test_ivf_rq_target_partition_size`

### DIFF
--- a/rust/lance-index/src/lib.rs
+++ b/rust/lance-index/src/lib.rs
@@ -312,6 +312,7 @@ impl IndexType {
             Self::IvfFlat => 4096,
             Self::IvfSq => 8192,
             Self::IvfPq => 8192,
+            Self::IvfRq => 4096,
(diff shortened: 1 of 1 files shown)


Xuanwo (approved):
Thank you!

PR #7275, opened 2026-06-15
fix(index): share FTS position stream offsets

BubbleCal:
### What is the bug?
FTS phrase-query position cache hits clone `SharedPositionStream`. The stream already used shared `Bytes` for the compressed position payload, but `block_offsets` was stored as `Vec<u32>`, so each clone copied one offset per compressed position block.

For large shared-position streams, repeated cache-hit clones can add avoidable O(N) CPU and memory traffic even when the position data itself is cached.

`SharedPositionStream` now stores `block_offsets` as `Arc<[u32]>`, keeping the existing `block_offsets() -> &[u32]` API while making clones share the offset allocation. A regression test verifies that cloned streams point at the same offsets allocation.

## Validation

- `cargo test -p lance-index scalar::inverted::index::tests::test_shared_position_stream_clone_shares_block_offsets`
- `cargo test -p lance-index shared_position`
- `cargo test -p lance-index shared_stream`
- `cargo test -p lance-index scalar::inverted::cache_codec::tests`
- `cargo test -p lance-index`
- `cargo fmt --all`
- `git diff --check`

### DIFF
--- a/rust/lance-index/src/scalar/inverted/index.rs
+++ b/rust/lance-index/src/scalar/inverted/index.rs
@@ -2947,10 +2947,9 @@ impl DeepSizeOf for CompressedPositionStorage {
 #[derive(Debug, Clone, PartialEq, Eq, Default)]
 pub struct SharedPositionStream {
     codec: PositionStreamCodec,
-    block_offsets: Vec<u32>,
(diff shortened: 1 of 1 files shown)


Xuanwo (approved):
Nice change

PR #7278, opened 2026-06-15
fix: reject nested legacy blobs in v2.2

Xuanwo:
This closes a v2.2 write-path hole where legacy blob metadata was rejected only on top-level fields. A nested field such as `summary_image_nested.image_bytes` could still carry `lance-encoding:blob` into a v2.2 dataset manifest.

The write guard now scans the full Lance schema preorder and reports the offending field path. This keeps v2.2 dataset creation from silently preserving BlobV1 columns while leaving valid `lance.blob.v2` inputs untouched.

### DIFF
--- a/rust/lance/src/dataset/write.rs
+++ b/rust/lance/src/dataset/write.rs
@@ -992,13 +992,10 @@ pub async fn write_fragments_internal(
     }
 
     if storage_version >= LanceFileVersion::V2_2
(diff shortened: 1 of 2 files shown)


geruh (approved):
Did a pass, LGTM!

BubbleCal (approved):

PR #7283, opened 2026-06-15
perf(fts): push prefilter through scalar index on flat FTS

LuQQiu:
## Summary

- Flat FTS (no inverted index on the text column) used `scan_fragments` + a manual `LanceFilterExec` to apply the prefilter, bypassing any scalar index on the filter column.
- Route the scan through `filtered_read` instead — same pattern as the brute-force KNN path at `scanner.rs:3839-3848` — so the pushable part of the prefilter is evaluated inside `FilteredReadExec` (using a scalar index when one exists) and only the unpushable `refine_expr` is reapplied on top.
- Requires `prefilter(true)` on the scanner. The postfilter branch passes an empty `ExprFilterPlan::default()` down the FTS path and is unaffected.

## Plan shape

Before (with `WHERE id = 1` and a BTree on `id`, no FTS index on `text`):

```
FlatMatchQueryExec
  └── LanceFilterExec(id = 1)            # post-scan filter, BTree unused
        └── LanceScan(columns=[text, id], with_row_id)
```

After:

```
FlatMatchQueryExec
  └── FilteredReadExec                   # full_filter=id = Int32(1), BTree used
        columns=[text], with_row_id
        # refine_expr (if any) reapplied as LanceFilterExec on top
```

## Test plan

- [x] New `test_fts_without_index_uses_scalar_index_for_prefilter` in `dataset_index.rs`: BTree on `id`, no FTS index, flat FTS + `id = 1` prefilter with `.prefilter(true)`. Asserts via `analyze_plan` that `LanceRead` shows `full_filter=id = Int32(1)`, no `LanceScan:` in the plan, and the result set is correct (2 rows).
- [x] Full FTS test suite passes (123 tests).
- [x] `cargo fmt --all`

### DIFF
--- a/python/python/tests/test_scalar_index.py
+++ b/python/python/tests/test_scalar_index.py
@@ -648,7 +648,10 @@ def test_partly_indexed_prefiltered_search(tmp_path):
     assert "ScalarIndexQuery" in plan
     assert "MaterializeIndex" not in plan
     assert "FlatMatchQuery" in plan
-    assert "LanceScan" in plan
+    # Flat FTS now reads via FilteredReadExec (prints as `LanceRead`) so the
+    # BTree on `id` pushes into the unindexed-fragment scan too.
+    assert "LanceRead" in plan
+    assert "LanceScan" not in plan
     assert make_fts_search(ds).to_table().num_rows == 12
 
     # Update vector index but NOT scalar index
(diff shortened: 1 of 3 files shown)


LuQQiu:
Duplicate; reopening against lancedb/lance via API.

BubbleCal (approved):

Issue #6852, opened 2026-05-19
Expose precomputed `RowAddrMask` as scanner prefilter input from Python API


JulianYG:
## Summary

Lance internally supports precomputed `RowAddrMask` as a first-class prefilter input via `DatasetPreFilter` and the existing `SelectionVectorToPrefilter` `FilterLoader`. The Arrow-binary serialization format is already defined for inter-process mask passing. However, there is no Python API to provide such a mask directly to a scanner — all prefilter inputs must go through expression evaluation (SQL string or pyarrow / Substrait `Expression`).

For workloads where row membership is precomputed externally — tag/labeling systems, business-logic-defined cohorts, materialized bitmaps maintained by separate services — this forces a round trip through expression construction and evaluation that becomes a bottleneck at scale.

This issue requests adding a parameter to `LanceDataset.scanner` (and `to_table`, `ScannerBuilder`) that accepts a serialized `RowAddrMask` and feeds it directly into the existing prefilter machinery, bypassing expression evaluation.

## Use case

We operate a Lance dataset at ~3B rows with an IVF vector index. A separate tag service is the source of truth for "datasets" — named subsets of rows identified by UUID — which mutate frequently and are managed independently of the main table. Encoding `dataset` membership as a column on the main table is not viable because:

1. Frequent membership changes would trigger fragment rewrites and invalidate scalar indexes.
2. The tag service is owned by a different system; coupling its updates to main-table writes is undesirable.
3. Membership is often computed dynamically (rule-based, user-curated) and not always representable as a stable scalar value.

A typical query is "vector search top-K, restricted to rows in the union of datasets {A, B, C}, optionally with additional scalar predicates like `city = 'NYC' AND ips_value < 100`." The dataset union ranges from a few hundred thousand rows to ~80M rows out of 3B.

The tag service can efficiently produce a serialized `RowAddrMask` (or its constituent `RowAddrTreeMap` bytes) describing allowed row IDs. But pylance has no way to consume one.

## Current workarounds and why they fall short

**1. SQL `_rowid IN (...)`** — Falls over at ~100k–1M values due to SQL parser cost. (Issue [#4115](https://github.com/lancedb/lance/issues/4115) proposes treating `_rowid` and `_rowaddr` as having an implicit scalar index, which would help but doesn't address the broader case of providing arbitrary bitmaps without going through expression evaluation.)

**2. Substrait `singular_or_list`** — Pushes the ceiling to a few million values, but at 50M+ values:
[...]

PR #7288, opened 2026-06-16
feat(scanner): add external row-address mask prefilter

JulianYG:
Addresses #6852.

## What
Adds `Scanner::with_row_addr_prefilter(RowAddrMask)`, letting callers pass a
precomputed row-address allow/block mask as a prefilter into vector and plain
scans, reusing the scanner's existing retrieval plan rather than re-deriving it.

## Motivation
Some pipelines precompute a set of eligible rows out-of-band (e.g. a stored
bitmap of rows belonging to a logical subset / dataset) and want to run KNN or a
plain scan restricted to that set -- without expressing it as a SQL filter. A
multi-hundred-thousand-element `IN (...)` is impractical to build and parse;
passing the row set directly is far cheaper.

## How
The mask threads into the existing prefilter machinery at three points:
- **ANN branch**: fed through `PreFilterSource` into `new_knn_exec`, ANDed with
  any deletion/SQL prefilter via a `MaskAndLoader`.
- **Flat / unindexed-fragment branch**: a new `RowAddrMaskFilterExec` filters
  scan output by `_rowid`, so rows appended after the index build are honored.
- **Plain (non-vector) scan**: the mask is supplied as the `FilteredReadExec`
  index input, so only masked rows are read; a SQL filter becomes a refine on top.

Deletions are still applied by `DatasetPreFilter`; illegal addresses are ignored.

## Status
Draft, pending API agreement on #6852. Behavior is exercised by an out-of-tree
PyO3 binding's test suite (based on v7.0.0, this PR is rebased); `cargo check -p lance` and `cargo fmt` are clean. I'd
appreciate guidance on the public API shape before finalizing in-tree tests.


JulianYG:
@coderabbitai review

JulianYG:
@coderabbitai review

JulianYG:
@coderabbitai review

JulianYG:
@coderabbitai review

JulianYG:
@coderabbitai review

JulianYG:
@coderabbitai review

JulianYG:
@coderabbitai review

JulianYG:
@coderabbitai review

JulianYG:
@coderabbitai Skipping this one — the premise doesn't hold against the current code.
FlatMatchQueryExec doesn't apply its own top-k. Its execute streams every scored unindexed row via flat_bm25_search_stream_* (supports_limit_pushdown() returns false), so there are no "limited results" to wrap. And RowAddrMaskFilterExec is the outermost node of the flat plan — the only top-k downstream is the SortExec…with_fetch(params.limit) in the mixed union case, or the plan's GlobalLimitExec, both applied strictly after the mask.
So masked rows are dropped before any k is applied and cannot consume a slot. The "excluded higher-ranked match consumes k, lower-ranked allowed match lost" scenario can't occur. Moving the mask below the scan would only avoid BM25-scoring masked rows (a minor cost saving), not change results, so I'm leaving the plan tree as-is.
Flat-branch masking is already tested: row_addr_mask_fts_search_only_allowed re-appends unindexed rows after indexing so the query matches on the flat branch, and asserts both the allowed-subset result and block-all-empty. A "consume k" variant isn't meaningful here because the single-match flat-only path returns before any score-sorted top-k — there's no k at that layer to exercise.

JulianYG:
@coderabbitai review

JulianYG:
@coderabbitai review

PR #7289, opened 2026-06-16
feat(schema_evolution): allow Dict <-> value-type casts in alter_columns

valkum:
Hey, got a small feature addition while messing around with dicts. I am afraid until BTree indexes support dict columns, we need to switch our column to `Utf8` and the cast was missing support for casting to or from `Dict<_, _>` columns.

## Summary

- `is_upcast_downcast` now treats `Dictionary` as a representation of its value type rather than a distinct family. This permits casts like `Dict(_, Utf8) <-> Utf8` (and `Dict(_, T1) <-> Dict(_, T2)` whenever `T1 <-> T2` was already allowed) via `alter_columns`.
- The function now takes a `LanceFileVersion`. Legacy storage is rejected upfront whenever the target requires materializing a fresh `Dictionary` column, because the legacy writer expects pre-populated `field.dictionary` metadata that the alter pipeline does not compute.
Without this, Legacy would fail mid-write with a cryptic `Lance field <name> misses dictionary info` error; now it fails fast with the same `Cannot cast column` message used for other rejected casts.
- The call site in `alter_columns` plumbs `dataset.manifest.data_storage_format.lance_file_version()?` through.

  ## Test plan

  - [x] `test_is_upcast_downcast_dictionary` — unit coverage for both directions, `Dict(_, Int)` peeling, cross-family rejection, and the Legacy rejection rule.
  - [x] `test_cast_dictionary_to_string` (Legacy + Stable) — Dict→Utf8 roundtrip, plus cross-family rejection through a dictionary.
  - [x] `test_cast_string_to_dictionary` (Legacy + Stable) — Utf8→Dict succeeds on Stable; on Legacy the alter returns `Cannot cast column` upfront instead of failing in the writer.
  - [x] `cargo fmt --all`
  - [x] Existing `test_cast_column` still passes on both file versions.

### DIFF
--- a/rust/lance/src/dataset/schema_evolution.rs
+++ b/rust/lance/src/dataset/schema_evolution.rs
@@ -150,24 +150,34 @@ impl ColumnAlteration {
 
 /// Limit casts to same type. This is mostly to filter out weird casts like
 /// casting a string to a boolean or float to string.
-fn is_upcast_downcast(from_type: &DataType, to_type: &DataType) -> bool {
+fn is_upcast_downcast(from_type: &DataType, to_type: &DataType, version: LanceFileVersion) -> bool {
     use DataType::*;
-    match from_type {
-        from_type if from_type.is_integer() => to_type.is_integer(),
-        from_type if from_type.is_floating() => to_type.is_floating(),
-        from_type if from_type.is_temporal() => to_type.is_temporal(),
-        Boolean => matches!(to_type, Boolean),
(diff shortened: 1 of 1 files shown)


wjones127 (approved):
This looks good. Once conflicts are resolved we can merge.

wjones127 (approved):
Looks good. Once conflicts are resolved we can merge.

valkum:
Ah sorry, had other work on my desk and forgot about this. Thanks for taking care of resolving the conflict.

PR #7295, opened 2026-06-16
perf(index): reduce TwoFileShuffler peak memory via interleave sort

### LINKED ISSUE #7299: Reduce TwoFileShuffler peak memory from 3-4x to 1x batch_size_bytes
The `TwoFileShuffler` (#6169) writes only two files (`shuffle_data.lance` and `shuffle_offsets.lance`) regardless of partition count: it accumulates `batch_size_bytes`, sorts that block by partition, appends it to `shuffle_data.lance`, and writes per-partition offsets to `shuffle_offsets.lance`.

Peak memory is currently **3-4x `batch_size_bytes`**. Reduce it to **~1x** by using zero-copy slicing of the input data and streaming the sorted data out, rather than materializing additional copies.

Draft PR: #7295

wjones127:
## Summary

- Replaces `rechunk_stream_by_size` + `concat_batches` + `take` (two full-data copies, peak ~3–4× `batch_size_bytes`) with a single-pass sort over the UInt32 part-id columns only, producing `(batch_idx, row_idx)` interleave indices.
- Sorted output is streamed to the data file via `interleave_batches` in 8 Ki-row chunks, so the interleave output adds only a small constant overhead above the accumulated source data.
- Peak memory drops to **~1× `batch_size_bytes`**, which enables setting `LANCE_SHUFFLE_BATCH_BYTES` much larger to reduce flush-group count and improve read-time I/O locality.

Closes #7299.

### DIFF
--- a/rust/lance-index/src/vector/v3/shuffler.rs
+++ b/rust/lance-index/src/vector/v3/shuffler.rs
@@ -410,6 +410,10 @@ impl TwoFileShuffler {
     }
 }
 
+/// `(batch_idx, row_idx)` pairs produced by [`sort_to_interleave_indices`], paired with
+/// per-partition row counts.
+type InterleaveResult = (Vec<(usize, usize)>, Vec<u64>);
+
 /// Sorts rows from multiple batches by partition ID and returns interleave indices.
 ///
 /// Builds a sort key of `(part_id, batch_idx, row_idx)` for every row across all
@@ -424,7 +428,7 @@ impl TwoFileShuffler {
 fn sort_to_interleave_indices(
     part_id_columns: &[&UInt32Array],
     num_partitions: usize,
-) -> Result<(Vec<(usize, usize)>, Vec<u64>)> {
+) -> Result<InterleaveResult> {
     let total_rows: usize = part_id_columns.iter().map(|a| a.len()).sum();
     let mut keys: Vec<(u32, u32, u32)> = Vec::with_capacity(total_rows);
     for (batch_idx, col) in part_id_columns.iter().enumerate() {
--- a/rust/lance-index/src/vector/v3/shuffler.rs
+++ b/rust/lance-index/src/vector/v3/shuffler.rs
@@ -419,10 +419,12 @@ impl TwoFileShuffler {
 ///
 /// Also returns per-partition row counts (derived from the same sorted keys at no
 /// extra cost).
+///
+/// Returns an error if any partition ID is out of range `[0, num_partitions)`.
 fn sort_to_interleave_indices(
     part_id_columns: &[&UInt32Array],
     num_partitions: usize,
-) -> (Vec<(usize, usize)>, Vec<u64>) {
+) -> Result<(Vec<(usize, usize)>, Vec<u64>)> {
     let total_rows: usize = part_id_columns.iter().map(|a| a.len()).sum();
     let mut keys: Vec<(u32, u32, u32)> = Vec::with_capacity(total_rows);
     for (batch_idx, col) in part_id_columns.iter().enumerate() {
@@ -437,18 +439,16 @@ fn sort_to_interleave_indices(
     let mut interleave_indices = Vec::with_capacity(total_rows);
     for (part_id, batch_idx, row_idx) in &keys {
         let pid = *part_id as usize;
-        if pid < num_partitions {
-            partition_counts[pid] += 1;
-        } else {
-            log::warn!(
-                "Partition ID {} is out of range [0, {})",
-                pid,
-                num_partitions
-            );
+        if pid >= num_partitions {
+            return Err(Error::invalid_input(format!(
+                "partition ID {} is out of range [0, {})",
+                pid, num_partitions
+            )));
         }
+        partition_counts[pid] += 1;
         interleave_indices.push((*batch_idx as usize, *row_idx as usize));
     }
-    (interleave_indices, partition_counts)
+    Ok((interleave_indices, partition_counts))
 }
 
 #[async_trait::async_trait]
@@ -591,13 +591,9 @@ async fn flush_shuffle_batch(
         .collect();
 
     let np = num_partitions;
-    let (interleave_indices, batch_partition_counts) = spawn_cpu(move || {
-        Ok::<_, Error>(sort_to_interleave_indices(
-            &part_id_cols.iter().collect::<Vec<_>>(),
-            np,
-        ))
-    })
-    .await?;
+    let (interleave_indices, batch_partition_counts) =
+        spawn_cpu(move || sort_to_interleave_indices(&part_id_cols.iter().collect::<Vec<_>>(), np))
+            .await?;
 
     // Drop part-id column from source batches before interleaving.
     let source_batches: Vec<RecordBatch> = accumulated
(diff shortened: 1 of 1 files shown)


wjones127:
## Benchmark: TwoFileShuffler memory & throughput vs batch size

Run against 500k rows × 128-dim float32 (244 MB vectors, 1024 partitions):

| batch_size | flush groups | elapsed | peak RSS |
|---|---|---|---|
| 16 MB | 16 | 227 ms | 323 MB |
| 32 MB | 8 | 188 ms | 348 MB |
| 64 MB | 4 | 358 ms | 350 MB |
| 128 MB | 2 | 200 ms | 346 MB |
| 256 MB | 1 | 178 ms | 352 MB |
| 512 MB | 1 | 157 ms | 364 MB |

### Flush groups (fewer = less random I/O at query time)

```
 16 MB │████████████████  16
 32 MB │████████           8
 64 MB │████               4
128 MB │██                 2
256 MB │█                  1
512 MB │█                  1
```

### Why peak RSS looks flat

The benchmark pre-generates all 244 MB of test data before any shuffle runs, so the OS RSS baseline already reflects that allocation. The _marginal_ memory attributable to the shuffler itself is small and gets lost in the noise. The real benefit shows up in the theoretical model:

| batch_size | new peak overhead | old peak overhead |
|---|---|---|
| 16 MB | ~16 MB | ~48–64 MB |
| 32 MB | ~32 MB | ~96–128 MB |
| 128 MB | ~128 MB | ~384–512 MB |
| 512 MB | ~512 MB | ~1.5–2 GB |

The old code did `concat_batches` (1× copy) + `take` (another 1× copy), giving a ~3–4× multiplier on `batch_size_bytes` at peak. The new code accumulates source batches (1×), sorts to index pairs, then streams `interleave_batches` in 8192-row chunks — the working set stays at ~1× batch_size throughout.

### Practical impact

With the old 3–4× overhead, setting `batch_size_bytes = 1 GB` on a 4 GB machine would require 3–4 GB just for the shuffle overhead, exhausting memory. With the new code, 1 GB batch size costs ~1 GB, meaning you can use the full available RAM. For this 244 MB dataset, that reduces flush groups from 16 → 1, cutting random-access I/O per query partition by 16×.

<details>
<summary>Reproduce</summary>

```bash
# (from lance-index/ directory, bench not checked in)
cargo bench --bench shuffler_mem 2>&1
```

</details>

wjones127:
## Memory benchmark: TwoFileShuffler interleave-sort

### Methodology

Two approaches:

1. **Rust micro-benchmark** (`benches/shuffler_mem.rs`, untracked): 500K rows × 128-dim float32 (244 MB), varying `LANCE_SHUFFLE_BATCH_BYTES`. Measures RSS overhead above baseline (input data already in memory). Compares the old `concat_batches + sort_to_indices + take` path (inlined) against the new `sort_to_interleave_indices + interleave_batches` path via `TwoFileShuffler`.

2. **Python end-to-end** (`test_indexing.py::test_io_mem_shuffler_batch_size`): IVF_PQ index build on ~1M rows × 3072-dim float32 (~204 MB PQ-encoded). Peak heap measured via `memtest` (LD_PRELOAD) over the full index build. Run on both `main` (8.0.0-beta.7) and this branch (8.0.0-beta.11).

---

### Rust micro-benchmark (shuffler phase in isolation)

Dataset: 500K rows × 128-dim float32, 244 MB, 1024 partitions.
Overhead = peak RSS above baseline (input already allocated before timing starts).

| Batch size | Flush groups | Old overhead | New overhead |
|---|---|---|---|
| 64 MB | 4 | 182 MB | ~0 MB |
| 128 MB | 2 | 382 MB | ~0 MB |
| 256 MB | 1 | 726 MB | ~0 MB |
| 512 MB | 1 | 728 MB | ~0 MB |

Old code scales at ~3× batch_size (accumulated input + concat copy + sorted-take copy). At 256 MB with 244 MB total data, old code peaks at ~3× 244 MB ≈ 732 MB extra above baseline.

New code overhead is essentially zero: the sort buffer (n × 12 bytes) and streaming 8192-row interleave chunks are negligible.

> 16 MB and 32 MB show noisy results for the new code (rapid flush cycles at 5 ms RSS sampling granularity). ≥64 MB cases are representative.

---

### Python end-to-end (main vs branch)

Dataset: ~1M rows × 3072-dim float32, ~204 MB PQ-encoded (index replaced between runs).

| Batch size | Main peak | Branch peak | Delta |
|---|---|---|---|
| 16 MB | 10.07 GB | 9.99 GB | −82 MB |
| 32 MB | 9.91 GB | 9.92 GB | +14 MB |
| 64 MB | 10.00 GB | 10.04 GB | +35 MB |
| 128 MB | 10.04 GB | 9.25 GB | **−790 MB** |

End-to-end peak is dominated by k-means training (~9–10 GB), which masks most of the shuffler improvement. The 128 MB case shows the clearest signal (−790 MB), consistent with the micro-benchmark prediction of ~382 MB shuffler overhead eliminated plus measurement variance. The smaller batch sizes show measurement noise in both directions.

The improvement is most impactful for production-scale workloads where users set `LANCE_SHUFFLE_BATCH_BYTES` to 2–8 GB to reduce flush groups and improve read locality.

rust/lance-index/src/vector/v3/shuffler.rs
```diff
@@ -468,72 +489,63 @@ impl Shuffler for TwoFileShuffler {
...
         let mut rows_processed: u64 = 0;
+        let mut total_loss = 0.0f64;
+        let mut accumulated: Vec<RecordBatch> = Vec::new();
+        let mut acc_bytes: usize = 0;
 
-        let mut rechunked = std::pin::pin!(rechunked);
-        while let Some(batch) = rechunked.next().await {
-            num_batches_ref.fetch_add(1, std::sync::atomic::Ordering::Relaxed);
+        let mut data = std::pin::pin!(data);
+        while let Some(batch) = data.next().await {
```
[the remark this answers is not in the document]
wjones127: If an incoming batch blows past memory limits, we can't do much about it. It's already in-memory. Slicing it and making a smaller copy just increases memory temporarily, so it's not helpful. If input chunks are too large, we need instead a way to tell the reader to use smaller chunks.

rust/lance-index/src/vector/v3/shuffler.rs
```diff
@@ -407,15 +410,59 @@ impl TwoFileShuffler {
...
+///
+/// Returns an error if any partition ID is out of range `[0, num_partitions)`.
+fn sort_to_interleave_indices(
+    part_id_columns: &[&UInt32Array],
+    num_partitions: usize,
+) -> Result<InterleaveResult> {
+    let total_rows: usize = part_id_columns.iter().map(|a| a.len()).sum();
+    let mut keys: Vec<(u32, u32, u32)> = Vec::with_capacity(total_rows);
+    for (batch_idx, col) in part_id_columns.iter().enumerate() {
+        let batch_idx = batch_idx as u32;
```
Xuanwo, line 435: Is it intentional that `batch_idx` and `row_idx` are narrowed to `u32` here instead of keeping them as `usize`?

rust/lance-index/src/vector/v3/shuffler.rs
```diff
@@ -468,72 +493,63 @@ impl Shuffler for TwoFileShuffler {
...
-                for count in &partition_counts {
-                    running += count;
-                    batch_offsets.push(running);
+            total_loss += batch
+                .metadata()
+                .get(LOSS_METADATA_KEY)
+                .and_then(|s| s.parse::<f64>().ok())
+                .unwrap_or(0.0);
+            acc_bytes += batch.get_array_memory_size();
+            accumulated.push(batch);
```
Xuanwo, line 513: Is it intentional that an oversized input batch is no longer split by `batch_size_bytes` before accumulation?

wjones127: Yes. I didn't seen the point in it. The point of `batch_size_bytes` it to keep memory constrained. But if you've already loaded an oversized batch, and you haven't run out of memory, then you should be fine to proceed as is. In fact, if you do a hard split, you are making additional copies and thus *increasing* the memory use. So I felt it best just to accept the batches as is, and accumulate.

Do you see an argument for splitting them?

rust/lance-index/src/vector/v3/shuffler.rs
wjones127: Yes. Just trying to keep the memory overhead low. We shouldn't have more than a billion batches or a billion rows per batch.

rust/lance-index/src/vector/v3/shuffler.rs
Xuanwo: Looks good to me. Just wanna keep the behavior compatible.

Xuanwo (approved):

PR #7296, opened 2026-06-16
feat(java): expose SourceDedupeBehavior in MergeInsertParams

sezruby:
## What

Expose the Rust core's `source_dedupe_behavior` (`Fail`/`FirstSeen`) on the Java `MergeInsertParams`. `MergeInsertBuilder` already supports it in lance-core, but the Java binding never wired it — Java callers were stuck on the default (`Fail`) with no way to opt into `FirstSeen`.

## Changes

- **`rust/lance/src/dataset.rs`** — re-export `SourceDedupeBehavior` from `lance::dataset` so the JNI crate can import it alongside the other merge types.
- **`MergeInsertParams.java`** — add `SourceDedupeBehavior { Fail, FirstSeen }` enum, `withSourceDedupeBehavior()` builder, getters, and `toString()` entry. Defaults to `Fail`, matching the Rust default.
- **`merge_insert.rs` (JNI)** — `extract_source_dedupe_behavior()` helper (mirrors `extract_when_matched`), passed through to the builder.
- **`MergeInsertTest.java`** — cover both enum values across the JNI boundary: `FirstSeen` keeps the first duplicate source row; `Fail` errors on duplicate source keys (asserting the "Ambiguous merge inserts are prohibited" cause) and leaves the dataset unchanged.

## Why

A merge source with duplicate join keys (e.g. a CDC stream with multiple updates per key) currently fails on the Java path with no recourse. `FirstSeen` is the documented escape hatch in lance-core; this makes it reachable from Java. Keeps the bindings consistent with the Python/Rust surface.

## Test

- `MergeInsertTest`: 11/11 pass (9 existing + 2 new)
- `cargo fmt` / `cargo clippy` (lance-jni) clean; `cargo check -p lance` clean
- `./mvnw spotless:check` clean

### DIFF
--- a/java/lance-jni/src/merge_insert.rs
+++ b/java/lance-jni/src/merge_insert.rs
@@ -12,7 +12,8 @@ use jni::objects::{JObject, JString, JValueGen};
 use jni::sys::jlong;
 use lance::dataset::scanner::ExprFilter;
 use lance::dataset::{
-    MergeInsertBuilder, MergeStats, WhenMatched, WhenNotMatched, WhenNotMatchedBySource,
+    MergeInsertBuilder, MergeStats, SourceDedupeBehavior, WhenMatched, WhenNotMatched,
+    WhenNotMatchedBySource,
 };
 use lance_core::datatypes::Schema;
 use lance_index::mem_wal::CompactedSsTable;
@@ -53,6 +54,7 @@ fn inner_merge_insert<'local>(
     let skip_auto_cleanup = extract_skip_auto_cleanup(env, &jparam)?;
     let use_index = extract_use_index(env, &jparam)?;
     let compacted_sstables = extract_compacted_sstables(env, &jparam)?;
+    let source_dedupe_behavior = extract_source_dedupe_behavior(env, &jparam)?;
 
     let (new_ds, merge_stats) = unsafe {
         let dataset = env.get_rust_field::<_, _, BlockingDataset>(jdataset, NATIVE_DATASET)?;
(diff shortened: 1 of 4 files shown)


java/lance-jni/src/merge_insert.rs
```diff
@@ -71,6 +73,7 @@ fn inner_merge_insert<'local>(
             .retry_timeout(Duration::from_millis(retry_timeout_ms as u64))
             .skip_auto_cleanup(skip_auto_cleanup)
             .use_index(use_index)
+            .source_dedupe_behavior(source_dedupe_behavior)
```
LuciferYang, line 77: `whenMatched=Delete` does not honor `sourceDedupeBehavior` yet. A Java caller can select `Fail` and still get a successful delete when the source contains duplicate keys, even though `Fail` should reject the ambiguous input. The Delete paths deduplicate by row address only, so the target row is deleted once instead of returning `Ambiguous merge inserts are prohibited`; `FirstSeen` also leaves `num_skipped_duplicates` unchanged.

Please apply the same source-key deduplication logic used by Update to both `FullSchema` and `DeleteOnly`, and add Java coverage for full-schema deletes and key-only delete-only plans.

sezruby: Thanks for the catch. This is resolved after rebasing onto `main`:

**Rust core:** the Delete path now honors `source_dedupe_behavior` for both the full-schema and delete-only plans, landed in #7484 — `Fail` returns `Ambiguous merge inserts are prohibited` (naming the duplicate key), and `FirstSeen` deletes the matched target row once while counting the duplicate in `num_skipped_duplicates`, matching Update. Covered there by the `full_schema_{fail,first_seen}` / `delete_only_{fail,first_seen}` rstest cases in `merge_insert.rs`. This PR rebases onto that and exposes the parameter to Java, so the binding inherits the behavior across all match modes.

**Java coverage:** added to `MergeInsertTest` — full-schema delete (`withMatchedDelete()` + `InsertAll`) and key-only delete-only (`withMatchedDelete()` + `DoNothing`, source carrying only the join key), each asserting `Fail` rejects with the ambiguous-merge error and leaves the target unchanged, and `FirstSeen` deletes once and skips the duplicate.

PR #7297, opened 2026-06-16
feat(java): add compaction io buffer size option

### LINKED ISSUE #7294: Expose `io_buffer_size` in Java in `CompactionOptions`
Follow up to #7226. Need to add that option for Java.

lalitium:
Added Java support for ioBufferSize in compaction options, then wired it through JNI to Rust, and added tests to verify it is preserved through planning and task serialization.

closes #7294


lalitium:
R: @wjones127 I've added ioBufferSize in compaction options. Please take a look. Thanks.

PR #7298, opened 2026-06-16
feat: expose session cache key inventory

jackye1995:
## Summary

- expose optional cache key inventory on cache backends and LanceCache
- implement Moka key snapshot support
- add Session helpers for index and metadata cache keys

### DIFF
--- a/rust/lance-core/src/cache/backend.rs
+++ b/rust/lance-core/src/cache/backend.rs
@@ -22,6 +22,9 @@ use super::CacheCodec;
 /// A type-erased cache entry.
 pub type CacheEntry = Arc<dyn Any + Send + Sync>;
 
+/// Iterator over cache keys currently known to a backend.
+pub type CacheKeyIterator<'a> = Box<dyn Iterator<Item = InternalCacheKey> + Send + 'a>;
+
 /// Structured cache key passed to [`CacheBackend`] methods.
 ///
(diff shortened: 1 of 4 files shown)


rust/lance-core/src/cache/backend.rs
```diff
@@ -116,6 +116,15 @@ pub trait CacheBackend: Send + Sync + std::fmt::Debug {
     /// Remove all entries.
     async fn clear(&self);
 
+    /// Return a snapshot of cache keys currently known to this backend.
+    ///
+    /// Backends that cannot enumerate keys cheaply or accurately should return
+    /// `None`. A returned empty vector means key inventory is supported and the
+    /// cache currently has no entries.
+    async fn keys(&self) -> Option<Vec<InternalCacheKey>> {
```
wjones127: issue(blocking): there could be hundreds of thousands of keys. Could we instead return an iterator? That would also compose better with the prefix filter.

wjones127 (changes requested):

rust/lance-core/src/cache/backend.rs
jackye1995: Done. Updated the cache inventory API to return a CacheKeyIterator instead of materializing Vec<InternalCacheKey>, and LanceCache::keys() now composes prefix filtering lazily over the backend iterator. Reran targeted tests/doctests and cargo clippy --all --tests --benches -- -D warnings.

wjones127 (approved):

PR #7303, opened 2026-06-16
perf: batch fragment-reuse index remap into a single rebuild + commit

xuanyu-z:
> 🟡 **DRAFT, stacked on #7286.** This branch is based on #7286, so the diff shows that fix's commit plus the one new `perf:` commit here — review the **`perf: batch fragment-reuse index remap…`** commit. I'll rebase to drop the #7286 commit once it merges.

## Summary

`remap_index` (the post-deferred-compaction catch-up that physically rewrites an index's stored row addresses) applied the fragment-reuse index **one version at a time** — rebuilding the index file and committing **once per reuse version**. An index touched by K deferred compactions paid **K full index rebuilds + K commits**, even though the final state is identical to applying all K at once. This is worst exactly when it matters most: when the reuse index has accumulated many versions before remap runs.

## Change

Compose the whole chain and rebuild once:

- **Row addresses:** `FragReuseIndex::remap_row_id` already chains every version and passes through addresses a version doesn't touch, so the union of all versions' map keys mapped through it yields a single **baseline → final** address map. One `index::remap_index` call applies it.
- **Coverage (fragment bitmap):** composed in a single pass over all versions, with the same all-or-nothing / straddle-error semantics as before (chaining is automatic — a version inserts its new fragments, which the next version sees as its old ones). The `data_predates_version` check is evaluated against the fixed baseline, since there are no longer any intermediate commits.
- One `CreateIndex` commit instead of one per version.

Semantically equivalent to the previous per-version loop (applying the composed map to a stored address gives the same result as applying each version's map in sequence), just without the intermediate rebuilds/commits.

## Why this is safe

- Fragment ids are never reused, so composing the union of version keys can't collide.
- The straddle invariant (#6610) is preserved — still errors on a partially-indexed rewrite group.
- All existing `remap` / `frag_reuse` / `compaction` tests pass unchanged, including `test_remap_index_after_compaction` (a 3-version chain) which now completes in one commit.

## Test

`test_remap_index_batches_multiple_reuse_versions`: accumulates ≥2 reuse versions (delete-prefix + deferred compaction in a loop), then asserts a single `remap_column_index` advances the dataset version by **exactly 1** (one commit, not one per version) and that the reuse index trims to zero afterward.

## Note

This optimizes the *cost per catch-up*; it's complementary to running remap regularly (which keeps the version count — and thus K — small in the first place).

### DIFF
--- a/rust/lance/src/dataset/index/frag_reuse.rs
+++ b/rust/lance/src/dataset/index/frag_reuse.rs
@@ -243,4 +243,91 @@ mod tests {
             Err(Error::RetryableCommitConflict { .. })
         ));
     }
+
+    /// With more than one index on the table, remapping every index must catch
+    /// all of them up so the reuse index can be trimmed.
+    ///
+    /// Regression: `remap_column_index` used to decide whether to remap an
+    /// index's data from the presence of the old fragments in its fragment
+    /// bitmap. But `load_indices` coverage-remaps the bitmap onto the new
+    /// fragments in memory, and remapping the *first* index commits a manifest
+    /// that persists that cleaned bitmap for the others — so remapping the
+    /// remaining indexes became a silent no-op (their data was never remapped
+    /// and their `dataset_version` never advanced), and the reuse index could
+    /// never be trimmed.
+    #[tokio::test]
+    async fn test_cleanup_frag_reuse_index_multiple_indices() {
+        let mut dataset = lance_datagen::gen_batch()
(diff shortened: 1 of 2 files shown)


xuanyu-z:
Superseded by a stacked PR based on the #7286 fix branch so the diff shows only the batching commit: https://github.com/xuanyu-z/lance/pull/1 . Will reopen against upstream main once #7286 merges.

PR #7310, opened 2026-06-17
feat(index): support Utf8View patterns in the ngram TextQueryParser

wombatu-kun:
Follow-up to #7139.

## Problem

The ngram `TextQueryParser` extracts string patterns and regex flags by matching only `ScalarValue::Utf8` and `ScalarValue::LargeUtf8`. When the indexed column is `Utf8View`, `safe_coerce_scalar` (`rust/lance-datafusion/src/expr.rs`) coerces the predicate literal to `ScalarValue::Utf8View(Some(..))`, which the parser then fails to match. As a result an ngram index on a `Utf8View` column silently does not accelerate `contains`, `regexp_like`, or infix `LIKE`; they fall back to a full scan.

## Change

Add the `Utf8View` arm to all three string-extraction sites in `TextQueryParser`: the `contains` / `regexp_like` pattern in `visit_scalar_function`, the infix-LIKE pattern in `visit_like`, and the regex-flags literal in `apply_regex_flags`. The bindings are unchanged because all three string `ScalarValue` variants carry an `Option<String>`.

## Tests

- New `test_text_query_parser_utf8view`: asserts `contains` and `regexp_like` over a `Utf8View`-typed ngram index route to the index (`StringContains` / `Regex` queries), and that infix `LIKE` with a `Utf8View` pattern is accelerated, with a `Utf8` parity control.
- Extended `test_apply_regex_flags` with a `Utf8View` flags literal.

Each test fails on the pre-change code and passes after.

## Out of scope

The identical pre-existing gap in `SargableQueryParser` (`starts_with` / LIKE-prefix) is left untouched - it is a separate feature that #7139 never modified, so folding it in would be unrelated scope. It can be a separate change if desired.

This addresses the non-blocking review comment https://github.com/lance-format/lance/pull/7139#discussion_r3422890853 from @wjones127.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved text search handling for `Utf8View` string values in `LIKE` predicates.
  * Enabled regex support for `Utf8View` inputs, including recognizing supported inline regex flags for `regexp_*` patterns.
  * Enhanced index-assisted matching and refinement for these string variants.
* **Tests**
  * Added and expanded automated coverage for `Utf8View` and related string cases, including `LIKE` pattern handling and regex flag parsing.

### DIFF
--- a/rust/lance-index/src/scalar/expression.rs
+++ b/rust/lance-index/src/scalar/expression.rs
@@ -897,8 +897,9 @@ impl ScalarQueryParser for TextQueryParser {
         }
         // A non-string pattern cannot be handled. `maybe_scalar` coerces the
         // literal to the indexed column's type, which for an ngram index is
-        // always `Utf8` (Lance normalizes `Utf8View` columns to `Utf8` at write
-        // time), so the coerced value here is never `Utf8View`.
+        // `Utf8` or `LargeUtf8` (the plugin rejects any other type at creation,
+        // and Lance normalizes a `Utf8View` column to `Utf8` at write time), so
+        // the coerced value here is never `Utf8View`.
         let (ScalarValue::Utf8(Some(pattern)) | ScalarValue::LargeUtf8(Some(pattern))) =
             maybe_scalar(&args[1], data_type)?
         else {
@@ -3126,34 +3127,41 @@ mod tests {
         // `Expr`), so a programmatically-built filter passed through
         // `scan.filter_expr` can carry a `Utf8View` literal that the parser must
         // still match. (A `contains` / `regexp_like` pattern is coerced to the
-        // indexed column's `Utf8` type first, so it never reaches the parser as
-        // `Utf8View`; an ngram-indexed column is never `Utf8View` either, because
-        // Lance normalizes it to `Utf8` at write time.)
+        // indexed column's type first, and an ngram index only ever covers a
+        // `Utf8` / `LargeUtf8` column, so that path never sees `Utf8View`.)
         //
         // `visit_like` is exercised directly so the test does not depend on
         // DataFusion's LIKE type coercion choosing `Utf8View` for the pattern. The
         // `Utf8` case is a parity control: the pre-existing path must keep behaving
-        // identically.
+        // identically. Each case builds its `Like` from the same literal it hands
+        // to the parser, mirroring `visit_like_expr`, which reads the pattern out
+        // of `like.pattern`.
         let parser = TextQueryParser::new("color_idx".to_string(), "NGram".to_string(), true, true);
-        let like = Like::new(
-            false,
-            Box::new(Expr::Column(Column::new_unqualified("color"))),
-            Box::new(Expr::Literal(
-                ScalarValue::Utf8View(Some("%foobar%".to_string())),
-                None,
-            )),
-            None,
-            false,
-        );
         for pattern in [
             ScalarValue::Utf8View(Some("%foobar%".to_string())),
             ScalarValue::Utf8(Some("%foobar%".to_string())),
         ] {
+            let like = Like::new(
+                false,
+                Box::new(Expr::Column(Column::new_unqualified("color"))),
+                Box::new(Expr::Literal(pattern.clone(), None)),
+                None,
+                false,
+            );
             let indexed = parser
                 .visit_like("color", &like, &pattern)
                 .expect("infix LIKE should use the ngram index");
-            assert!(indexed.scalar_query.is_some(), "index query missing");
-            assert!(indexed.refine_expr.is_some(), "LIKE recheck missing");
+
+            let Some(ScalarIndexExpr::Query(search)) = indexed.scalar_query else {
+                panic!("expected an index query for {pattern:?}");
+            };
+            assert_eq!(
+                search.query.as_any().downcast_ref::<TextQuery>(),
+                Some(&TextQuery::Regex(".*foobar.*".to_string())),
+            );
+            // The recheck is the original predicate verbatim, so its literal keeps
+            // the variant that was exercised.
+            assert_eq!(indexed.refine_expr, Some(Expr::Like(like)));
         }
     }
 
--- a/rust/lance-index/src/scalar/expression.rs
+++ b/rust/lance-index/src/scalar/expression.rs
@@ -2370,6 +2370,7 @@ mod tests {
     use lance_datafusion::exec::{LanceExecutionOptions, get_session_context};
     use lance_select::result::IndexExprResultWireFormat;
     use roaring::RoaringBitmap;
+    use rstest::rstest;
 
     use crate::scalar::json::{JsonQuery, JsonQueryParser};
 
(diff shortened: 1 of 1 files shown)


rust/lance-index/src/scalar/expression.rs
```diff
@@ -2887,6 +2900,84 @@ mod tests {
...
+        // Regression test for the follow-up to PR #7139: the ngram `TextQueryParser`
+        // must accept `Utf8View` patterns/flags, not only `Utf8` / `LargeUtf8`. When
+        // the indexed column is `Utf8View`, `safe_coerce_scalar` coerces the literal
+        // to a `ScalarValue::Utf8View`, so the parser has to match that variant or the
+        // index is silently skipped. The mock index declares `Utf8View` to drive that
+        // coercion regardless of the planner's schema type.
+        let index_info = MockIndexInfoProvider::new(vec![(
+            "color",
+            ColInfo::new(
+                DataType::Utf8View,
```
Xuanwo: The regression test models an ngram index on a Utf8View field, but the real NGram plugin still rejects Utf8View fields during index creation. This records an unreachable production contract and can hide regressions in the actual query path.

wombatu-kun: Done db595b8c2. Verified that no shipped path (native filter or DataFusion pushdown) delivers a Utf8View literal here - the column normalizes to Utf8 and the contains/regexp_like pattern is coerced to it - so that arm was unreachable; dropped it and the mock that modeled an impossible Utf8View-typed index. Kept Utf8View only on the uncoerced visit_like / regex-flags paths (still reachable via a filter built programmatically and pushed through scan.filter_expr), now covered by a direct visit_like test.

rust/lance-index/src/scalar/expression.rs
```diff
@@ -2887,6 +2906,44 @@ mod tests {
         );
     }
 
+    #[test]
+    fn test_text_query_parser_utf8view() {
+        // The infix-LIKE pattern reaches `visit_like` uncoerced (verbatim from the
+        // `Expr`), so a programmatically-built filter passed through
+        // `scan.filter_expr` can carry a `Utf8View` literal that the parser must
+        // still match - the same defensive contract #7139 / #7351 established for
```
Xuanwo: The test comment says the Utf8View LIKE behavior follows an already established BTree parser contract, but that contract is not present in this tree. This can mislead future changes by implying Utf8View LIKE-prefix support exists outside the ngram parser.

wombatu-kun: Done 25efdf7f1 - removed the BTree SargableQueryParser reference

rust/lance-index/src/scalar/expression.rs
```diff
@@ -3101,6 +3120,43 @@ mod tests {
...
+        // indexed column's `Utf8` type first, so it never reaches the parser as
+        // `Utf8View`; an ngram-indexed column is never `Utf8View` either, because
+        // Lance normalizes it to `Utf8` at write time.)
+        //
+        // `visit_like` is exercised directly so the test does not depend on
+        // DataFusion's LIKE type coercion choosing `Utf8View` for the pattern. The
+        // `Utf8` case is a parity control: the pre-existing path must keep behaving
+        // identically.
+        let parser = TextQueryParser::new("color_idx".to_string(), "NGram".to_string(), true, true);
+        let like = Like::new(
```
Xuanwo, line 3555: The parity case reuses a `LIKE` expression whose pattern is always `Utf8View` while passing a separate `Utf8` pattern into `visit_like`, so it can pass without proving the returned recheck expression matches the variant being exercised.

rust/lance-index/src/scalar/expression.rs
```diff
@@ -895,7 +895,10 @@ impl ScalarQueryParser for TextQueryParser {
         if args.len() < 2 {
             return None;
         }
-        // A non-string pattern cannot be handled.
+        // A non-string pattern cannot be handled. `maybe_scalar` coerces the
+        // literal to the indexed column's type, which for an ngram index is
+        // always `Utf8` (Lance normalizes `Utf8View` columns to `Utf8` at write
```
Xuanwo: The comment says the indexed column type is always `Utf8`, but ngram indexes also accept `LargeUtf8`, so future changes can incorrectly treat the `LargeUtf8` branch as unreachable.

wombatu-kun: Done b1d880f94 - the comment now says `Utf8` or `LargeUtf8`; the match arm already handled both.

rust/lance-index/src/scalar/expression.rs
wombatu-kun: Done b1d880f94 - each case now builds its own `Like` from the `ScalarValue` it passes in, and asserts the produced `TextQuery::Regex` plus the recheck literal, so the `Utf8` case can no longer pass on a `Utf8View` recheck.

rust/lance-index/src/scalar/expression.rs
```diff
@@ -3091,6 +3102,15 @@ mod tests {
...
         );
+        // A `Utf8View` flags literal is folded just like `Utf8` / `LargeUtf8`.
+        assert_eq!(
+            apply_regex_flags(
+                "foo",
+                &Expr::Literal(ScalarValue::Utf8View(Some("i".to_string())), None)
+            )
+            .as_deref(),
+            Some("(?i)foo")
+        );
```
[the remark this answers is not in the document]
wombatu-kun: Done ae830f128

wombatu-kun:
@Xuanwo all your comments are addressed. Could you take a look please?

rust/lance-index/src/scalar/expression.rs
```diff
@@ -981,8 +985,14 @@ impl ScalarQueryParser for TextQueryParser {
...
         }
+        // Unlike the `contains` / `regexp_like` pattern, the LIKE pattern reaches
+        // us uncoerced (verbatim from the `Expr`), so a programmatically-built
+        // filter passed through `scan.filter_expr` can carry any string variant
+        // here, including `Utf8View`.
         let pattern_str = match pattern {
-            ScalarValue::Utf8(Some(s)) | ScalarValue::LargeUtf8(Some(s)) => s.as_str(),
+            ScalarValue::Utf8(Some(s))
+            | ScalarValue::LargeUtf8(Some(s))
+            | ScalarValue::Utf8View(Some(s)) => s.as_str(),
```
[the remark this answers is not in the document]
wombatu-kun: Confirmed by a planner-level reproduction: the column arrives as `Cast(column, <wider>)` and index discovery drops it, so this arm cannot run. Fixed in #9054 rather than here, since the cast blocks every scalar index on a LIKE and not just ngram: discovery now looks through a cast that stays within the string types, and `visit_like_expr` normalizes the pattern to the indexed column's type so a wider literal cannot reach a narrower index. That normalization is what makes a `Utf8View` pattern work, so the arm in this PR is unnecessary and I am closing it.

wombatu-kun:
Superseded by #9054, which fixes the reachability problem in shared index discovery instead of adding an ngram-specific arm.

PR #7311, opened 2026-06-17
feat: add SpillStore trait with local-disk implementation

### LINKED ISSUE #7300: Add SpillStore trait with local-disk implementation
Index builds need a uniform place to put temporary, reclaimable intermediate state (posting-list segments, shuffle runs, btree pages) that can tier beyond memory when a machine is starved.

Add a **`SpillStore` trait**, available on `Session`, providing temporary scratch with an RAII lifecycle (handles delete on drop) and trafficking in opaque byte streams.

Provide a **local-disk implementation** that writes to temp dirs and returns a clean error when a configured size cap is exceeded (similar to DataFusion's `DiskManager`).

This is the foundation for memory-budgeted index builds: builders react to memory pressure (see the MemoryPool threading work) by spilling intermediate state through `SpillStore`. The trait should be a clean seam so alternative implementations can be injected via `Session`.

### Acceptance
- `SpillStore` trait + local-disk impl
- RAII cleanup of scratch handles
- Clean, typed error on disk-cap exhaustion (no panic)

wjones127:
## Summary

Adds a generic `SpillStore` — reclaimable RAII scratch storage for intermediate state that overflows memory (e.g. index-build posting lists, shuffle runs, BTree pages). Mechanism only; consumer migration (IVF shuffler) is a follow-up.

- `SpillStore` / `SpillFile` (lance-io): `create_spill_file()` vends a write-once RAII handle whose `writer()` / `reader()` hand back `Box<dyn Writer>` / `Box<dyn Reader>`, so callers feed spill files straight into `FileWriter::try_new` and a v2 `FileReader` without leaking an `ObjectStore` + path. Dropping the handle deletes the file and releases its bytes back to the store's budget.
- `LocalSpillStore`: writes to an OS temp directory; `with_cap` enforces an optional byte budget shared across all handles, returning a typed `Error::DiskCapExceeded` instead of silently filling the disk. Enforcement lives entirely in the spill store — the spill file decorates the writer with a `QuotaWriter` (reserve-on-write, release-on-drop-by-stat) rather than threading a field through `ObjectStore` and every provider, so it works for any backend the store opens.
- `From<io::Error>` recovers a wrapped lance `Error`, so typed errors such as `DiskCapExceeded` survive the `AsyncWrite` boundary.
- `ScanScheduler::open_reader` builds a `FileScheduler` over an already-open `Reader` (no path/size lookup), bridging a bare spill reader into the v2 reader path.
- `Session` gains a `spill_store` field (default: uncapped `LocalSpillStore`), a `with_spill_store()` builder for injection, and a `spill_store()` accessor.

Closes #7300

### DIFF
--- a/rust/lance-io/src/spill.rs
+++ b/rust/lance-io/src/spill.rs
@@ -1,0 +1,512 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright The Lance Authors
+
+//! Reclaimable scratch storage.
+//!
+//! A [`SpillStore`] hands out [`SpillFile`] handles for temporary state that is
+//! too large to keep in memory and is read back later in the same process (for
+//! example, posting lists or shuffle runs accumulated while building an index).
+//! The backing storage is reclaimed automatically when a handle is dropped.
+//!
+//! Key properties:
+//! - **`writer()`/`reader()` interface.** A [`SpillFile`] vends a [`Writer`]
+//!   and a [`Reader`] rather than exposing an [`ObjectStore`] and a path. The
+//!   writer feeds `FileWriter::try_new` directly; the reader feeds a v2
+//!   `FileReader` via [`crate::scheduler::ScanScheduler::open_reader`].
+//! - **Per-file RAII.** One [`SpillFile`] is one file; dropping it deletes the
+//!   file and releases its bytes back to the store's disk budget, so a caller
+//!   can hold N files and reclaim them individually.
+//! - **Disk cap enforcement.** [`LocalSpillStore::with_cap`] enforces a byte
+//!   budget shared across all handles, returning a typed
+//!   [`lance_core::Error::DiskCapExceeded`] rather than silently filling the
+//!   disk.
+//!
+//! # Usage contract
+//!
+//! Hold the [`SpillFile`] alive for the file's lifetime: the [`Writer`] and
+//! [`Reader`] reference the backing file, which the handle deletes on drop. The
+//! store's temp directory is the backstop for anything leaked when it drops.
+//!
+//! Accounting is reserve-on-write + release-on-drop (by stat), which is exact
+//! for the write-once contract this enforces: [`SpillFile::writer`] may be
+//! called only once, so the bytes reserved while writing match the file size
+//! released on drop. Two minor inexactnesses are not engineered around: a write
+//! aborted at the cap leaks its reservation until the store is dropped, and a
+//! file whose size cannot be stat-ed on drop is not released.
+
+use std::io;
+use std::path::PathBuf;
+use std::pin::Pin;
+use std::sync::Arc;
+use std::sync::atomic::{AtomicBool, AtomicU64, Ordering};
+use std::task::{Context, Poll};
+
+use async_trait::async_trait;
+use object_store::path::Path;
+use tokio::io::AsyncWrite;
+
+use lance_core::{Error, Result};
+
+use crate::object_store::ObjectStore;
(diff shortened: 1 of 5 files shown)


rust/lance-io/src/spill.rs
```diff
@@ -0,0 +1,512 @@
...
+//! Hold the [`SpillFile`] alive for the file's lifetime: the [`Writer`] and
+//! [`Reader`] reference the backing file, which the handle deletes on drop. The
+//! store's temp directory is the backstop for anything leaked when it drops.
+//!
+//! Accounting is reserve-on-write + release-on-drop (by stat), which is exact
+//! for the write-once contract this enforces: [`SpillFile::writer`] may be
+//! called only once, so the bytes reserved while writing match the file size
+//! released on drop. Two minor inexactnesses are not engineered around: a write
+//! aborted at the cap leaks its reservation until the store is dropped, and a
+//! file whose size cannot be stat-ed on drop is not released.
```
westonpace: What does "a file whose size cannot be stat-ed on drop" mean?  I guess it's not super important.

rust/lance-io/src/spill.rs
```diff
@@ -0,0 +1,512 @@
...
+    /// [`lance_core::Error::DiskCapExceeded`]. Spill files are write-once:
+    /// implementations may reject a second call (the [`LocalSpillStore`] impl
+    /// returns [`lance_core::Error::invalid_input`]).
+    async fn writer(&self) -> Result<Box<dyn Writer>>;
+
+    /// Open a reader over this spill file.
+    ///
+    /// The data must have been fully written (the writer shut down) first.
+    async fn reader(&self) -> Result<Box<dyn Reader>>;
+}
```
westonpace: It's not super important but it seems there is an implicit lifecycle here that could be captured at compile time.  The file must be fully written before it can be opened for reads and the writer can only be obtained once.

rust/lance-io/src/spill.rs
```diff
@@ -0,0 +1,512 @@
...
+///
+/// Data is written through [`SpillFile::writer`] and read back through
+/// [`SpillFile::reader`]. The backing storage is released when the handle is
+/// dropped.
+///
+/// The trait is object-safe so it can be returned as `Box<dyn SpillFile>` from
+/// [`SpillStore::create_spill_file`], allowing implementations not backed by a
+/// local file (e.g. in-memory buffers, remote object stores).
+#[async_trait]
+pub trait SpillFile: Send + Sync {
```
westonpace: Why a trait and not a struct?  Are you expecting multiple impls?

rust/lance-io/src/spill.rs
```diff
@@ -0,0 +1,512 @@
...
+            }
+            if self
+                .used
+                .compare_exchange(current, next, Ordering::Relaxed, Ordering::Relaxed)
+                .is_ok()
+            {
+                return Ok(());
+            }
+            // Another thread won the CAS — retry.
+        }
```
westonpace: Personal nit but anytime I start writing CAS loops I give up and just use a mutex 😆

westonpace (approved):
Just some nits

rust/lance-io/src/spill.rs
wjones127: That's a good point. I might refactor based on that.

rust/lance-io/src/spill.rs
wjones127: Yeah, potentially multiple custom impls. They might custom how and which disks are used. If we didn't do a trait, I would probably re-use DataFusions DiskManager rather than create out own.

PR #7317, opened 2026-06-17
perf: batch fragment-reuse index remap into a single rebuild + commit

xuanyu-z:
## Summary

`remap_index` — the catch-up that physically applies a fragment-reuse index after a deferred-remap compaction — applied the reuse index **one version at a time**, rebuilding the index file and committing **once per reuse version**. An index touched by K deferred compactions paid **K full index rebuilds + K commits** for a result identical to applying all K at once. This is worst exactly when the reuse index has accumulated many versions before a remap runs.

## Change

Compose the whole chain and rebuild once:

- **Row addresses:** `FragReuseIndex::remap_row_id` already chains every version (and passes through addresses a version does not touch), so mapping the union of all versions' keys yields a single **baseline → final** address map, applied in one rebuild.
- **Coverage bitmap:** composed in one pass with the same all-or-nothing / straddle-error semantics (chaining is automatic — a version's new fragments are the next version's old fragments). `data_predates_version` is evaluated against the fixed baseline since there are no intermediate commits.
- One `CreateIndex` commit instead of one per version.

## Why the composed map is not filtered by the fragment bitmap

A tempting way to keep the composed map small is to drop keys whose fragment isn't in the index's current `fragment_bitmap`. That optimization is **not** safe — the per-version loop never did it, and this PR keeps it that way.

In the sibling-coverage-remap case, remapping one index commits a manifest that coverage-remaps a *sibling* index's bitmap onto the new fragments and persists it *before* the sibling's own data is remapped. The sibling's on-disk bitmap then shows the new fragments while its data still holds old addresses. Filtering the map by that bitmap would drop exactly the keys the sibling needs, leaving an **empty** map — and `index::remap_index` treats an all-`None`/empty map as `RemapResult::Keep`, reusing the stale index files while the version is bumped and the reuse index trims, so the index would end up pointing at dead fragments.

So the composed map maps every old address the reuse index touched; addresses an index doesn't store are simply never looked up (the map stays bounded by the rows the reuse index touched).

## Tests

- `test_remap_index_batches_multiple_reuse_versions` — a multi-version reuse chain must rebuild + commit exactly once.
- `test_cleanup_frag_reuse_index_multiple_indices` — extended with a post-remap data-correctness scan so it asserts each remapped index resolves to **live rows**, not just that versions advance and the reuse index trims. (A bitmap-filtered map would make the sibling index return 0 of 1000 rows here.)

```
cargo test -p lance --lib frag_reuse::tests   # 3 passed
cargo test -p lance --lib remap               # 21 passed
```

### DIFF
--- a/rust/lance/src/dataset/index/frag_reuse.rs
+++ b/rust/lance/src/dataset/index/frag_reuse.rs
@@ -329,5 +329,112 @@ mod tests {
             .await
             .unwrap();
         assert_eq!(frag_reuse_details.versions.len(), 0);
+
+        // Data correctness, not just version bookkeeping: with the reuse index
+        // trimmed there is no auto-remap safety net, so each index must resolve
+        // to LIVE rows. An index whose data was not actually remapped (e.g. one
+        // whose bitmap was coverage-remapped by a sibling's commit before its
+        // own data remap) points at compacted-away fragments and errors on take.
+        use futures::TryStreamExt;
+        for col in ["i", "j"] {
+            let rows: usize = dataset
+                .scan()
+                .filter(&format!("{col} >= 2000 AND {col} < 3000"))
+                .unwrap()
+                .try_into_stream()
+                .await
+                .unwrap()
+                .try_collect::<Vec<_>>()
+                .await
+                .unwrap()
+                .iter()
+                .map(|b| b.num_rows())
+                .sum();
+            assert_eq!(
+                rows, 1000,
+                "index {col}_idx must resolve to live rows after remap+trim"
+            );
+        }
+    }
+
+    /// When the reuse index has accumulated several versions, a single remap
+    /// must compose them and rebuild + commit the index exactly ONCE, not once
+    /// per version.
+    #[tokio::test]
+    async fn test_remap_index_batches_multiple_reuse_versions() {
+        let mut dataset = lance_datagen::gen_batch()
+            .col("i", lance_datagen::array::step::<Int32Type>())
+            .into_ram_dataset(FragmentCount::from(8), FragmentRowCount::from(1000))
+            .await
+            .unwrap();
+        dataset
+            .create_index(
+                &["i"],
+                IndexType::Scalar,
+                Some("i_idx".into()),
+                &ScalarIndexParams::default(),
+                false,
+            )
+            .await
+            .unwrap();
+
+        // Accumulate multiple reuse versions: each round deletes a prefix, which
+        // shrinks fragments below target and forces another deferred compaction.
+        let options = CompactionOptions {
+            target_rows_per_fragment: 4_000,
+            defer_index_remap: true,
+            ..Default::default()
+        };
+        for round in 0..4 {
+            dataset
+                .delete(&format!("i < {}", 1_000 * (round + 1)))
+                .await
+                .unwrap();
+            compact_files(&mut dataset, options.clone(), None)
+                .await
+                .unwrap();
+        }
+
+        let frag_reuse_index_meta = dataset
+            .load_index_by_name(FRAG_REUSE_INDEX_NAME)
+            .await
+            .unwrap()
+            .expect("Fragment reuse index must be available");
+        let num_versions = load_frag_reuse_index_details(&dataset, &frag_reuse_index_meta)
+            .await
+            .unwrap()
+            .versions
+            .len();
+        assert!(
+            num_versions >= 2,
+            "test needs multiple reuse versions to exercise batching, got {num_versions}"
+        );
+
+        // A single remap must commit exactly once, regardless of version count.
+        let version_before = dataset.manifest.version;
+        remapping::remap_column_index(&mut dataset, &["i"], Some("i_idx".into()))
+            .await
+            .unwrap();
+        let commits = dataset.manifest.version - version_before;
+        assert_eq!(
+            commits, 1,
+            "batched remap must commit once, not once per reuse version ({num_versions})"
+        );
+
+        // ... and the reuse index then trims to zero.
+        cleanup_frag_reuse_index(&mut dataset).await.unwrap();
+        let frag_reuse_index_meta = dataset
+            .load_index_by_name(FRAG_REUSE_INDEX_NAME)
+            .await
+            .unwrap()
+            .expect("Fragment reuse index must be available");
+        assert_eq!(
+            load_frag_reuse_index_details(&dataset, &frag_reuse_index_meta)
+                .await
+                .unwrap()
+                .versions
+                .len(),
+            0
+        );
     }
 }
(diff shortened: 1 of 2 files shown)


westonpace (approved):
Sounds like a good optimization.  Attaching feedback from Claude

---

## PR #7317 Review: `perf: batch fragment-reuse index remap into a single rebuild + commit`

### Overview

This PR optimizes `remap_index` by collapsing K sequential index rebuilds + commits (one per deferred-remap version) into a single composed rebuild + commit. The approach is correct and the description is thorough.

---

### Code Quality

**Strengths:**
- The refactor is clean and well-scoped — no unrelated changes.
- The `remap_row_id` chain composition is leveraged rather than re-implemented.
- The small iterator cleanup (`group.new_frags.iter()` instead of `clone().into_iter()`) is a nice bonus.
- Comments explain the non-obvious "why" (the sibling-coverage-remap trap) clearly.

**Issues / Suggestions:**

1. **`unwrap()` on index metadata lookup** — The `.unwrap()` after `.find(|idx| idx.uuid == *index_id)` will panic if the index has been concurrently dropped. The old code had the same problem inside the loop, but this is a good opportunity to fix it since you're touching the code anyway. Consider returning a more informative error:
   ```rust
   .ok_or_else(|| Error::Internal { message: format!("index {} not found in manifest", index_id), .. })?;
   ```

2. **`composed_row_id_map` duplicates keys from multiple versions** — When multiple versions map the same `old_addr`, `flat_map` visits it multiple times and `.collect::<HashMap<_,_>>()` silently keeps the last one. Since `remap_row_id` already chains all versions via the full `frag_reuse_index`, the final value for each key is consistent regardless of order, so this is **not a bug** — but it wastes allocations. A more explicit approach using `keys().copied()` with deduplication via the `HashMap` insert semantics is fine in practice; just worth a brief comment that duplicates are intentional no-ops.

3. **Early return in `RemapResult::Drop` discards the bitmap update** — In the original loop, `continue` on `Drop` meant subsequent versions could still remap the index. The new code returns `Ok(())` on `Drop`, which is correct for the composed case (if the index is dropped, there's nothing more to do). This is a subtle behavioral change that's correct but could use a comment.

4. **`removed_indices: vec![curr_index_meta]` moves the struct** — This is fine, but note that in the `Keep` path `curr_index_meta` was cloned into `new_index_meta` fields already. A partial move is happening here; Rust allows it, but worth confirming the compiler is happy (it should be since the fields are cloned before the move).

---

### Test Coverage

- `test_remap_index_batches_multiple_reuse_versions`: Good — verifies the commit count is exactly 1 and that the reuse index trims correctly. This is a precise behavioral assertion.
- Extension of `test_cleanup_frag_reuse_index_multiple_indices`: The post-remap data scan catching the sibling-bitmap bug is exactly the right kind of regression guard. The filter `{col} >= 2000 AND {col} < 3000` assumes the dataset layout from the test setup — worth a brief comment tying it back to the compaction rounds.

---

### Performance

The improvement is real and proportional to the number of accumulated reuse versions. No new allocations of significance are introduced beyond the `composed_row_id_map` (which is bounded by the union of all version maps, the same data that was traversed in the old loop anyway).

---

### Summary

**Verdict: Approve with minor suggestions.** The logic is sound, the PR description documents the tricky sibling-coverage-remap invariant well, and the tests are meaningful. The main actionable suggestion is replacing the `unwrap()` on the index metadata lookup with a proper error, and optionally adding a comment about duplicate key handling in the composed map.

xuanyu-z:
> Sounds like a good optimization. Attaching feedback from Claude
> 
> ## PR #7317 Review: `perf: batch fragment-reuse index remap into a single rebuild + commit`
> ### Overview
> This PR optimizes `remap_index` by collapsing K sequential index rebuilds + commits (one per deferred-remap version) into a single composed rebuild + commit. The approach is correct and the description is thorough.
> 
> ### Code Quality
> **Strengths:**
> 
> * The refactor is clean and well-scoped — no unrelated changes.
> * The `remap_row_id` chain composition is leveraged rather than re-implemented.
> * The small iterator cleanup (`group.new_frags.iter()` instead of `clone().into_iter()`) is a nice bonus.
> * Comments explain the non-obvious "why" (the sibling-coverage-remap trap) clearly.
> 
> **Issues / Suggestions:**
> 
> 1. **`unwrap()` on index metadata lookup** — The `.unwrap()` after `.find(|idx| idx.uuid == *index_id)` will panic if the index has been concurrently dropped. The old code had the same problem inside the loop, but this is a good opportunity to fix it since you're touching the code anyway. Consider returning a more informative error:
>    ```rust
>    .ok_or_else(|| Error::Internal { message: format!("index {} not found in manifest", index_id), .. })?;
>    ```
> 2. **`composed_row_id_map` duplicates keys from multiple versions** — When multiple versions map the same `old_addr`, `flat_map` visits it multiple times and `.collect::<HashMap<_,_>>()` silently keeps the last one. Since `remap_row_id` already chains all versions via the full `frag_reuse_index`, the final value for each key is consistent regardless of order, so this is **not a bug** — but it wastes allocations. A more explicit approach using `keys().copied()` with deduplication via the `HashMap` insert semantics is fine in practice; just worth a brief comment that duplicates are intentional no-ops.
> 3. **Early return in `RemapResult::Drop` discards the bitmap update** — In the original loop, `continue` on `Drop` meant subsequent versions could still remap the index. The new code returns `Ok(())` on `Drop`, which is correct for the composed case (if the index is dropped, there's nothing more to do). This is a subtle behavioral change that's correct but could use a comment.
> 4. **`removed_indices: vec![curr_index_meta]` moves the struct** — This is fine, but note that in the `Keep` path `curr_index_meta` was cloned into `new_index_meta` fields already. A partial move is happening here; Rust allows it, but worth confirming the compiler is happy (it should be since the fields are cloned before the move).
> 
> ### Test Coverage
> * `test_remap_index_batches_multiple_reuse_versions`: Good — verifies the commit count is exactly 1 and that the reuse index trims correctly. This is a precise behavioral assertion.
> * Extension of `test_cleanup_frag_reuse_index_multiple_indices`: The post-remap data scan catching the sibling-bitmap bug is exactly the right kind of regression guard. The filter `{col} >= 2000 AND {col} < 3000` assumes the dataset layout from the test setup — worth a brief comment tying it back to the compaction rounds.
> 
> ### Performance
> The improvement is real and proportional to the number of accumulated reuse versions. No new allocations of significance are introduced beyond the `composed_row_id_map` (which is bounded by the union of all version maps, the same data that was traversed in the old loop anyway).
> 
> ### Summary
> **Verdict: Approve with minor suggestions.** The logic is sound, the PR description documents the tricky sibling-coverage-remap invariant well, and the tests are meaningful. The main actionable suggestion is replacing the `unwrap()` on the index metadata lookup with a proper error, and optionally adding a comment about duplicate key handling in the composed map.

Thanks, addressed a few of the comments

xuanyu-z:
~All checks passed (except a flaky test) after addressing comments, ready for merge

PR #7318, opened 2026-06-17
fix: restore blanks for empty batches and split oversized writes in updater

zhangyang0418:
## What

Follow-up to #7233. Fixes two more failures that occur when a fully deleted batch sits in the **middle** of a fragment and columns are added/merged (e.g. via `Fragment::merge_columns` + `LanceOperation::Merge`).

## Root cause

After #7233 handled the 0-row batch at a read boundary, two related issues remain in `updater.rs`:

1. **`add_blanks` rejects empty batches.** `DeletionRestorer` walks the deletion vector and may need to insert blanks for deleted rows that land on a batch whose live rows are all elsewhere, producing an empty input batch with non-empty offsets. `add_blanks` returned `NotSupported` ("Missing too many rows in merge, run compaction to materialize deletions first") in that case.
2. **Restored batches can exceed `batch_size`.** `DeletionRestorer` greedily appends trailing deleted rows to the preceding batch, so the restored batch can be larger than the requested `batch_size`, producing oversized output file batches.

## Fix

1. `add_blanks` now builds null rows via `new_null_array` for an empty input batch instead of erroring.
2. `Updater::update` slices the restored batch back down to the write batch size before writing.

## Tests

- `add_blanks` unit test on an empty batch.
- `add_columns` Rust test with a middle fully-deleted batch.
- Python regression test exercising `merge_columns` + commit over a dataset with a deletion file.

### DIFF
--- a/rust/lance/src/dataset/updater.rs
+++ b/rust/lance/src/dataset/updater.rs
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-FileCopyrightText: Copyright The Lance Authors
 
-use arrow_array::{RecordBatch, UInt32Array};
+use arrow_array::{new_null_array, RecordBatch, UInt32Array};
 use futures::StreamExt;
 use lance_core::datatypes::{OnMissing, OnTypeMismatch};
 use lance_core::utils::deletion::DeletionVector;
@@ -49,6 +49,8 @@ pub struct Updater {
     finished: bool,
 
     deletion_restorer: DeletionRestorer,
+
+    write_batch_size: Option<u32>,
 }
 
 impl Updater {
@@ -97,6 +99,7 @@ impl Updater {
             schema_adapter: None,
             finished: false,
             deletion_restorer: DeletionRestorer::new(deletion_vector, legacy_batch_size),
+            write_batch_size: Some(batch_size),
         })
     }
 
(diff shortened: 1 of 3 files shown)


rust/lance/src/dataset/updater.rs
```diff
@@ -390,11 +411,13 @@ pub(crate) fn add_blanks(batch: RecordBatch, batch_offsets: &[u32]) -> Result<Re
...
-        // TODO: implement adding blanks for an empty batch.
-        // This is difficult because we need to create a batch for arbitrary schemas.
-        return Err(Error::not_supported_source(
-            "Missing too many rows in merge, run compaction to materialize deletions first".into(),
-        ));
+        let arrays = batch
+            .schema()
+            .fields()
+            .iter()
+            .map(|field| new_null_array(field.data_type(), batch_offsets.len()))
```
Xuanwo, line 418: Restoring an empty batch with null placeholders still fails when the first read batch is fully deleted and the added column is non-nullable, so valid add/merge-column operations can error before returning any visible rows.

rust/lance/src/dataset/updater.rs
```diff
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-FileCopyrightText: Copyright The Lance Authors
 
-use arrow_array::{RecordBatch, UInt32Array};
+use arrow_array::{new_null_array, RecordBatch, UInt32Array};
```
Xuanwo, line 4: The new import order is not rustfmt-formatted, so the format and lint jobs fail before the PR reaches the rest of the Rust checks.

PR #7321, opened 2026-06-17
feat: distributed centroid training

summaryzb:
## Summary

This PR adds scheduler-neutral distributed IVF centroid training primitives so Spark, Ray, or custom RPC systems can train one global centroid set across many Lance fragments. It introduces a shared Arrow `PartialStats` wire format in Rust, dataset-aware IVF wrapp
ers, Python and Java bindings, and documentation for driving Round-0 sampling plus iterative Lloyd rounds outside Lance.

## Problem

Lance already supports distributed index segment construction, but callers did not have a reusable way to train shared IVF centroids across workers. That forced distributed builders either to centralize training data on one process or to train independent per-segm
ent centroid models, which prevents a single global model from being reused across worker-built vector index segments. The missing piece was a small, stable API surface for the k-means math itself: sample local data, compute per-worker E-step statistics, merge tho
se statistics, and finalize the next centroids.

## Approach

The implementation splits centroid training into two layers. `lance_index::vector::kmeans::distributed` provides pure Arrow-native primitives: `PartialStats`, `compute_partial_stats`, `merge_partial_stats`, `reduce_partial_stats`, `finalize_centroids`, reservoir s
ampling, and driver-side centroid initialization. `PartialStats` is represented as a fixed-schema `RecordBatch` with metadata for version, shape, distance type, and a centroid fingerprint so reducers reject stats from incompatible rounds.

`lance::index::vector::ivf::distributed` adds dataset-aware async wrappers that scan optional fragment slices, normalize cosine input when needed, filter non-finite training data, and delegate to the pure primitives. The existing streaming IVF refinement path is r
efactored to use these same partial-stats primitives, keeping the new distributed path aligned with the in-process training behavior. Python and Java expose the same operations through Arrow IPC-friendly APIs while leaving broadcast, tree reduction, convergence ch
ecks, and scheduling to the caller.

## Changes

- `rust/lance-index/src/vector/kmeans/distributed.rs`: adds the core distributed k-means primitives, `PartialStats` schema/metadata validation, centroid fingerprinting, E-step accumulation, merge/reduce, M-step finalization, reservoir sampling, and bootstrap helpe
rs.
- `rust/lance-index/src/vector/kmeans.rs`, `rust/lance-index/Cargo.toml`, `Cargo.toml`, `Cargo.lock`: wires the distributed k-means module into `lance-index` and adds the `sha2` dependency used for centroid fingerprints.
- `rust/lance/src/index/vector/ivf/distributed.rs`: adds dataset-aware distributed IVF training wrappers for worker-side sampling and partial-stat computation over optional fragment id slices.
- `rust/lance/src/index/vector/ivf.rs`: refactors streaming IVF refinement to accumulate and finalize centroids through the new `PartialStats` primitives and adds a regression test against the legacy sum/weight calculation.
- `python/src/indices.rs`, `python/python/lance/indices/distributed_kmeans.py`, `python/python/lance/indices/__init__.py`: registers PyO3 bindings and a Python facade for sampling, partial-stat computation, merge/reduce, centroid finalization, and centroid initial
ization.
- `java/lance-jni/src/distributed_kmeans.rs`, `java/lance-jni/src/lib.rs`, `java/src/main/java/org/lance/index/vector/DistributedKMeans.java`: adds JNI and Java public APIs using Arrow IPC for samples/stats and flat float arrays for centroid outputs.
- `docs/src/guide/distributed_indexing.md`: documents distributed centroid training roles, the Python workflow, and the Arrow `RecordBatch` interchange contract.

## Test Coverage

- Rust unit tests cover `PartialStats` schema/version validation, centroid fingerprint determinism, basic E-step counts/loss, Hamming rejection, all-NaN/empty handling, merge identity and fingerprint mismatch rejection, centroid finalization including empty-cluste
r fallback, deterministic reservoir sampling, initial centroid selection, and bootstrap k-means.
- Rust property tests cover merge commutativity and associativity across random splits.
- Rust IVF tests cover dataset-aware Round-0 sampling, layer-2 partial-stat computation, a simulated four-worker end-to-end training loop compared with single-machine IVF training, and streaming refinement parity with the legacy accumulation path.
- Python tests cover the public `lance.indices.distributed_kmeans` round trip, Arrow IPC serialization of partial stats, merge behavior after IPC restore, and initial centroid selection shape.
- Java JUnit tests cover sample/bootstrap/partial-stats/reduce/finalize flow through the public Java API and JNI Arrow IPC bridge.


rust/lance-index/src/vector/kmeans.rs
```diff
@@ -88,6 +90,12 @@ pub struct KMeansParams {
 
     /// Optional sync callback for iteration progress: (current_iteration, max_iterations).
     pub on_progress: Option<Arc<dyn Fn(u32, u32) + Send + Sync>>,
+
+    /// Optional RNG seed. When `Some`, `kmeans_random_init` uses
+    /// `SmallRng::seed_from_u64(seed)` instead of `SmallRng::from_os_rng()` so
+    /// random initialization is reproducible. Required by the distributed
+    /// kmeans `bootstrap_centroids` primitive (see `distributed.rs`).
+    pub seed: Option<u64>,
```
Xuanwo: Adding a public field to this public parameter struct changes the struct-literal contract, so downstream Rust code that constructs it by name will fail to compile until it is updated.

rust/lance-index/src/vector/kmeans/distributed.rs
```diff
@@ -0,0 +1,1259 @@
...
+pub fn bootstrap_centroids(
+    samples: Vec<RecordBatch>,
+    k: usize,
+    distance_type: DistanceType,
+    rng_seed: u64,
+) -> Result<FixedSizeListArray> {
+    let combined = concat_samples(samples)?;
+    let dim = combined.value_length() as usize;
+    let params = KMeansParams::default()
+        .with_distance_type(distance_type)
```
Xuanwo: Cosine bootstrap passes the cosine metric into the existing float kmeans path; ordinary-size runs can hit the unsupported-metric branch and panic instead of returning centroids.

rust/lance-index/src/vector/kmeans/distributed.rs
```diff
@@ -0,0 +1,1259 @@
...
+                Arc::new(loss),
+                Arc::new(radius),
+            ],
+        )
+        .expect("partial stats schema is internally consistent");
+        Self { batch }
+    }
+
+    /// Validate and adopt an externally-built RecordBatch.
+    pub fn from_record_batch(batch: RecordBatch) -> Result<Self> {
```
Xuanwo, line 166: Externally supplied stats are accepted after only version and column-name checks, while later merge/finalize paths unwrap concrete column types; malformed Arrow IPC can panic instead of returning a validation error.

rust/lance-index/src/vector/kmeans/distributed.rs
```diff
@@ -0,0 +1,1259 @@
...
+}
+
+/// Compute new centroids from accumulated stats.
+///
+/// `prev` provides the dtype and the fallback for empty clusters.
+pub fn finalize_centroids(
+    stats: &PartialStats,
+    prev: &FixedSizeListArray,
+) -> Result<FixedSizeListArray> {
+    if stats.k() != prev.len() {
```
Xuanwo, line 728: Finalization never checks that the stats fingerprint matches the previous centroids, so stale stats from another training round can be accepted and produce centroids for the wrong assignments.

rust/lance/src/index/vector/ivf/distributed.rs
```diff
@@ -0,0 +1,250 @@
...
+pub async fn compute_partial_stats(
+    dataset: &Dataset,
+    column: &str,
+    fragment_ids: Option<&[u32]>,
+    centroids: &FixedSizeListArray,
+    distance_type: DistanceType,
+) -> Result<PartialStats> {
+    // Pull all matching rows; the worker is expected to have a small enough fragment
+    // slice that scanning it whole is cheap.
+    let raw = maybe_sample_training_data(dataset, column, usize::MAX, fragment_ids).await?;
```
Xuanwo: The worker E-step materializes the entire fragment slice before computing stats, so large shards can create memory usage proportional to all rows times dimension instead of the bounded partial-stats state.

rust/lance/src/index/vector/ivf/distributed.rs
```diff
@@ -0,0 +1,250 @@
...
+    dataset: &Dataset,
+    column: &str,
+    fragment_ids: Option<&[u32]>,
+    target: usize,
+    distance_type: DistanceType,
+    rng_seed: u64,
+) -> Result<RecordBatch> {
+    // Round-0 oversamples to give the driver-side bootstrap enough material.
+    let raw =
+        maybe_sample_training_data(dataset, column, target.saturating_mul(2), fragment_ids).await?;
```
Xuanwo: The public seed only controls the second reservoir pass; the earlier dataset sampling remains nondeterministic, so retrying the same worker with the same seed can train from different samples.

java/lance-jni/src/distributed_kmeans.rs
```diff
@@ -0,0 +1,334 @@
...
+        .downcast_ref::<FixedSizeListArray>()
+        .ok_or_else(|| Error::input_error("centroids column must be FixedSizeList".to_string()))?
+        .clone())
+}
+
+fn fsl_to_jfloat_array<'a>(
+    env: &mut JNIEnv<'a>,
+    fsl: &FixedSizeListArray,
+) -> Result<jni::objects::JFloatArray<'a>> {
+    let values = fsl.values().as_primitive::<Float32Type>().values().to_vec();
```
Xuanwo: The Java return path unconditionally reads centroid values as Float32 even though the Rust primitive can return Float16 or Float64, so supported vector dtypes can fail across JNI.

java/lance-jni/src/distributed_kmeans.rs
```diff
@@ -0,0 +1,334 @@
...
+        let column: String = env.get_string(&column_jstr)?.into();
+        let dt = parse_distance_type(&mut env, &distance_type_jstr)?;
+        let frags = read_optional_fragment_ids(&mut env, &fragment_ids_arr)?;
+        let dataset_guard =
+            unsafe { env.get_rust_field::<_, _, BlockingDataset>(&dataset_obj, NATIVE_DATASET) }?;
+        let batch = RT.block_on(l2::sample_round_0(
+            &dataset_guard.inner,
+            &column,
+            frags.as_deref(),
+            target as usize,
```
Xuanwo, line 158: A negative Java target is cast to a huge unsigned size before sampling, so invalid input can turn Round-0 sampling into a full-slice scan and return.

rust/lance-index/src/vector/kmeans.rs
summaryzb: Removed the new pub seed

rust/lance-index/src/vector/kmeans/distributed.rs
summaryzb: fixed inside bootstrap_centroids.  When `distance_type == Cosine` we now L2-normalize the combined samples up front, dispatch the inner k-means with `DistanceType::L2` (mirroring the Layer-1 contract in compute_partial_stats), and then re-normalize the returned means so callers still see unit-norm centroids

rust/lance-index/src/vector/kmeans/distributed.rs
summaryzb: The fingerprint is the same 8-byte SHA-256 prefix already used to gate merge_partial_stats, so the trust boundary is consistent on both the merge and finalize paths

java/lance-jni/src/distributed_kmeans.rs
summaryzb: nativeSampleRound0 now validates `target < 0` up front

java/lance-jni/src/distributed_kmeans.rs
summaryzb: now return jbyteArray carrying an Arrow-IPC RecordBatch whose single column is   `FixedSizeList<original_dtype>`, so callers can dispatch on it instead of silently downcasting

rust/lance/src/index/vector/ivf/distributed.rs
summaryzb: The seed is now plumbed all the way through

rust/lance/src/index/vector/ivf/distributed.rs
summaryzb: Pick memory is O(k·d) instead of O(N·d). Empty input is handled by returning `PartialStats::empty(k, dim, distance_type, fp)` with the right fingerprint so downstream merge/finalize stays uniform

rust/lance-index/src/vector/kmeans/distributed.rs
```diff
@@ -0,0 +1,1808 @@
...
+                        "StreamingReservoir: schema mismatch across chunks (prev=FSL<{:?},{}> got=FSL<{:?},{}>)",
+                        prev_type, prev_dim, chunk_value_type, chunk_dim
+                    )));
+                }
+            }
+            _ => unreachable!(),
+        }
+
+        let batch_idx = self.held.len();
+        self.held.push(chunk.clone());
```
Xuanwo, line 936: The streaming reservoir still stores every input chunk even after selected rows from that chunk are replaced, so Round-0 sampling can retain the full streamed vector payload in memory.

rust/lance-index/src/vector/kmeans/distributed.rs
```diff
@@ -0,0 +1,1808 @@
...
+    // Cosine: normalize inputs once, then run inner kmeans as L2 (mirrors the
+    // assignment dispatch in `compute_partial_stats`).
+    let is_cosine = matches!(distance_type, DistanceType::Cosine);
+    let (combined, inner_dt) = if is_cosine {
+        (
+            normalize_fsl_owned(combined).map_err(arrow_error_to_lance)?,
+            DistanceType::L2,
+        )
+    } else {
+        (combined, distance_type)
```
Xuanwo, line 1081: Hamming requests still flow into the float k-means bootstrap path, so a public distributed bootstrap call can panic instead of returning a validation error.

rust/lance-index/src/vector/kmeans/distributed.rs
```diff
@@ -0,0 +1,1808 @@
...
+    }
+}
+
+fn concat_samples(samples: Vec<RecordBatch>) -> Result<FixedSizeListArray> {
+    if samples.is_empty() {
+        return Err(Error::index(
+            "select/bootstrap requires at least one sample batch",
+        ));
+    }
+    let arrays: Vec<&dyn Array> = samples.iter().map(|b| b.column(0).as_ref()).collect();
```
Xuanwo, line 1013: Externally supplied sample batches are indexed before validating that they contain a vector column, so malformed IPC can panic the driver process instead of producing an error.

rust/lance/src/index/vector/utils.rs
```diff
@@ -454,6 +479,118 @@ async fn scan_all_training_data(
...
+/// (Layer-2 partial stats) or (b) reservoir-sample (Layer-2 round 0), so the
+/// loss of the dense-path optimisation is irrelevant.
+pub(crate) async fn sample_training_data_stream(
+    dataset: &Dataset,
+    column: &str,
+    sample_size_hint: usize,
+    fragment_ids: Option<&[u32]>,
+    seed: Option<u64>,
+) -> Result<Pin<Box<dyn Stream<Item = Result<RecordBatch>> + Send>>> {
+    let num_rows = count_rows(dataset, fragment_ids).await?;
```
Xuanwo, line 527: Duplicate fragment filters are counted before the later deduplication path, so a repeated id can hit the ordered-fragment assertion and panic public distributed calls.

rust/lance-index/src/vector/kmeans/distributed.rs
```diff
@@ -0,0 +1,1808 @@
...
+    Ok(acc)
+}
+
+fn pairwise_sum_u64(a: &dyn Array, b: &dyn Array) -> Vec<u64> {
+    let a = a.as_any().downcast_ref::<UInt64Array>().unwrap();
+    let b = b.as_any().downcast_ref::<UInt64Array>().unwrap();
+    a.values()
+        .iter()
+        .zip(b.values().iter())
+        .map(|(x, y)| x + y)
```
Xuanwo, line 680: Merging externally supplied stats still adds counts with unchecked u64 arithmetic, so oversized counts can panic in debug builds or wrap in release and corrupt centroid averages.

python/src/indices.rs
```diff
@@ -748,6 +748,143 @@ impl PyIndexDescription {
...
+
+use lance::index::vector::ivf::distributed as l2;
+use lance_index::vector::kmeans::distributed as l1;
+
+fn parse_distance_type(s: &str) -> PyResult<DistanceType> {
+    DistanceType::try_from(s).map_err(|e| PyValueError::new_err(e.to_string()))
+}
+
+fn array_data_to_fsl(array: ArrayData) -> PyResult<FixedSizeListArray> {
+    Ok(FixedSizeListArray::from(array))
```
Xuanwo, line 763: Python centroid input is converted with a panic-on-wrong-type Arrow constructor, so malformed user input can abort the process instead of raising a Python error.

sezruby:
Thanks for this @summaryzb — the scheduler-neutral `PartialStats` design (Arrow wire format + centroid fingerprint so reducers reject mismatched stats) is exactly what distributed IVF training needs, and I'd like to build a Spark orchestration layer on top of it.

One suggestion that might help it land faster: would you be open to splitting it into stacked PRs? At +3959 / 20 files across four surfaces (Rust core, IVF integration, Java, Python) it's a lot to review in one pass, which may be part of why it's been quiet since early July. Splitting lets each surface get focused review and hit its own coverage bar (codecov currently flags 84.9%, 267 lines uncovered).

A natural split — core lands first, the rest fan out in parallel:

1. **Core primitive** — `lance-index/src/vector/kmeans/distributed.rs` + the `kmeans.rs` refactor: `PartialStats`, `compute`/`merge`/`reduce_partial_stats`, `finalize_centroids`, reservoir sampling, centroid init, Rust unit tests. Pure Arrow, no bindings — the foundation everything else depends on. Docs can ride along here.
2. **IVF / dataset integration** — `lance/src/index/vector/ivf/distributed.rs`, `ivf.rs`, `utils.rs`. Depends on #1.
3. **Java bindings** — JNI `distributed_kmeans.rs` + `DistributedKMeans.java` + tests. Depends on #1.
4. **Python bindings** — `distributed_kmeans.py` + `indices.rs` + tests. Depends on #1.

Once #1 is in, #2–#4 are independent and reviewable in parallel.

Happy to help however's useful — I can review, rebase against main, or take one of the binding/integration slices off your plate.

PR #7322, opened 2026-06-17
feat(blob-v2): configure packed blob file max size with field metadata

jo-migo:
# Fixes 
https://github.com/lance-format/lance/issues/7292

# Problem
There's currently no way to encode a threshold for maximum packed blob sidecar file size into a dataset in the way that you can encode inline blob and dedicated blob size thresholds via `lance-encoding:blob-dedicated-size-threshold` and `lance-encoding:blob-inline-size-threshold` respectively.

# Solution

Add a new `lance-encoding:blob-pack-file-size-threshold` field-level metadata key which informs the blob writer to only start new packed files when current packed file reaches that threshold.

It can be overridden by supplying another value via the existing `blob_pack_file_size_threshold` parameter to the `write_dataset` function.


rust/lance/src/dataset/blob.rs
```diff
@@ -225,7 +242,7 @@ impl PackWriter {
         let len = source.size();
         if self
             .current_blob_id
-            .map(|_| self.current_size + len > self.max_pack_size)
+            .map(|_| self.current_size + len > max_pack_size)
```
Xuanwo: This uses the incoming field's threshold to decide whether to append to the shared pack file. In datasets with multiple blob fields and different thresholds, a file that already contains data from a smaller-threshold field can be extended under a larger threshold, so the persisted field metadata no longer reliably bounds sidecar size.

rust/lance/src/dataset/write.rs
```diff
@@ -195,49 +196,32 @@ fn validate_blob_threshold_metadata_for_append(
...
-                return Err(Error::invalid_input(format!(
-                    "Cannot append data with blob threshold metadata {}={} for field '{}'; \
-                     the dataset schema has effective value {}. Blob thresholds for existing \
-                     columns are stored in the dataset schema.",
-                    BLOB_INLINE_SIZE_THRESHOLD_META_KEY,
-                    input_inline_threshold,
-                    input_field.name,
-                    dataset_inline_threshold,
-                )));
+        for (key, read_threshold) in [
```
Xuanwo, line 207: This append validation only checks top-level fields, even though blob threshold metadata is read recursively from nested blob fields. A nested blob append can silently ignore mismatched pack-threshold metadata and write using the persisted schema instead, so nested fields do not get the same explicit-mismatch protection as top-level blob columns.

rust/lance/src/dataset/blob.rs
jo-migo: Thanks for the review! Good point, I was not dealing with datasets containing different thresholds before. Now this case should be covered.

rust/lance/src/dataset/write.rs
jo-migo: I made this whole check recursive.

Xuanwo (approved):
Thank you!

PR #7323, opened 2026-06-17
fix: out-of-bounds write in fsst decompress when data compresses better than 3:1

### LINKED ISSUE #7266: FSST decompress writes out of bounds when data compresses better than 3:1

LuciferYang:
## Problem

`fsst::decompress` enforces a minimum output buffer of `3 * input`, but that bound is unsound. A code can decode to an 8-byte symbol, and `decompress_bulk` writes a full 8-byte word per code through raw pointers, so the decoded output can be as large as `8 * input`. When data compresses better than 3:1, decoding into a buffer sized at the documented minimum runs past the end of the allocation: a segfault under a strict allocator, an out-of-bounds write under Miri/ASan.

In-tree callers happen to allocate `8 * input`, so they were never affected. Only the public contract was wrong.

## Fix

`decompress` now grows its (owned) output buffer to `8 * in_buf.len() + 8` before decoding, instead of requiring 3x. The `+ 8` is one word of headroom for the final speculative write, which lands at `out_curr + 8` even when the last symbol is shorter than 8 bytes. The buffer is shrunk back to the exact decoded length afterwards, so callers see no change.

That bound only holds while every symbol is at most 8 bytes. Since the symbol table is read from untrusted on-disk data, `init` now rejects any table carrying a symbol length above 8 (`InvalidData`) rather than letting it break the bound.

## Tests

- `test_decompress_better_than_3x_does_not_overflow` and its `_64_bit_offsets` variant compress highly repetitive input (about 8:1) and decode it into a 3x buffer, checking a correct round trip for both 32- and 64-bit offsets. Both fail before the fix.
- `test_decompress_rejects_oversized_symbol_length` corrupts a symbol length byte and expects `InvalidData`.

`cargo test -p fsst`, `cargo clippy -p fsst --tests -- -D warnings`, and `cargo fmt --all` pass.

Closes #7266.

### DIFF
--- a/rust/compression/fsst/src/fsst.rs
+++ b/rust/compression/fsst/src/fsst.rs
@@ -1188,15 +1188,9 @@ impl<T: OffsetSizeTrait> FsstDecoder<T> {
         }
 
         self.decoder_switch_on = (st_info & (1 << 24)) != 0;
-        // when decoder_switch_on is true, we make sure the out_buf is at least 3 times the size of the in_buf,
-        if self.decoder_switch_on && in_buf.len() * 3 > out_buf.len() {
-            return Err(io::Error::new(
-                io::ErrorKind::InvalidInput,
-                "output buffer too small for FSST decoder",
-            ));
-        }
-
-        // when decoder_switch_on is false, we make sure the out_buf is at least the same size of the in_buf,
+        // The compressed path grows `out_buf` to the worst-case decoded size in `decompress`
+        // (see issue #7266), so it is sized there rather than validated here. The uncompressed
+        // path copies `in_buf` verbatim, so the output must be at least that large.
         if !self.decoder_switch_on && in_buf.len() > out_buf.len() {
             return Err(io::Error::new(
                 io::ErrorKind::InvalidInput,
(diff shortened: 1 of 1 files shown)


LuciferYang:
close this one since https://github.com/lance-format/lance/pull/7267  before this one.

PR #7325, opened 2026-06-17
fix(fts): keep inverted index correct under deletions in the fragment-reuse window

xuanyu-z:
## Summary

A deferred-remap compaction that materializes deletions writes a fragment-reuse index (FRI) that the inverted (full-text-search) index is read through at load time. The load-time path dropped the deleted rows and renumbered the surviving `doc_id`s — but the posting lists reference `doc_id`s **positionally** (a `doc_id` is an index into the `DocSet`'s `row_ids` / `num_tokens` arrays, fixed at build time) and are not regenerated at load. Dropping rows shifted every later `doc_id` out from under the posting lists, so a query would index `num_tokens` / `row_ids` out of bounds (panic) or score/return the wrong document.

This is deletion-specific: merge-only deferred compaction remaps every row to `Some(new_addr)`, so nothing is dropped and positions stay aligned. It only breaks when deletions are materialized (`remap_row_id` returns `None`).

## Fix: tombstone-preserve-positions

In `DocSet::from_columns` (the FRI load path), instead of dropping deleted rows:

- keep every doc slot so `doc_id`s stay aligned with the posting lists;
- put `RowAddress::TOMBSTONE_ROW` in the deleted slots, and leave them out of the `inv` reverse map so a `row_id` lookup never resolves to a deleted doc;
- keep `num_tokens` full-length, so `num_tokens(doc_id)` can't go out of bounds.

In `Wand::search`, skip docs whose resolved `row_id` is `TOMBSTONE_ROW` — placed right beside the existing prefilter-mask skip and using the same iterator-advance, so a tombstoned doc is stepped over exactly like a prefilter-rejected one and never surfaces in results.

The heavyweight physical remap (`DocSet::remap`) still does the real renumber + compact (and rebuilds the posting lists to match); this load-time path only needs to stay consistent until then.

### Note on stats

Tombstoned slots are still counted in `total_tokens` / `len()`, so BM25 `avgdl` in the FRI window is effectively the pre-deletion average. This only perturbs *scores* slightly, never the result set, and the physical remap restores exact stats. Excluding tombstones would require changing `len()` semantics (used by `idf`), which isn't worth it for a transient window.

## Test

`test_read_inverted_index_with_defer_index_remap_and_deletions`: delete a prefix, deferred-compact, then assert FTS returns exactly the surviving rows — both in the FRI window and after physical remap + trim. Without the fix it panics on the out-of-bounds `num_tokens` access.

## Scope

Independent change against `main` — touches only the inverted-index load and query paths (`scalar/inverted/{index,wand}.rs`) plus one new test. The analogous IVF_HNSW desync under deletions (#3993) is **not** addressed here: the HNSW graph traverses node ids positionally to compute distances *during* search (not just at result collection), so it needs a different approach across the SQ/PQ/flat/RQ storage load paths — a separate change.

### DIFF
--- a/rust/lance-index/src/scalar/inverted/index.rs
+++ b/rust/lance-index/src/scalar/inverted/index.rs
@@ -42,6 +42,7 @@ use lance_arrow::{RecordBatchExt, iter_str_array};
 use lance_core::cache::{CacheCodec, CacheKey, LanceCache, WeakLanceCache};
 use lance_core::deepsize::DeepSizeOf;
 use lance_core::error::{DataFusionResult, LanceOptionExt};
+use lance_core::utils::address::RowAddress;
 use lance_core::utils::tokio::{get_num_compute_intensive_cpus, spawn_cpu};
 use lance_core::utils::tracing::{IO_TYPE_LOAD_SCALAR_PART, TRACE_IO_EVENTS};
 use lance_core::{Error, ROW_ID, ROW_ID_FIELD, Result};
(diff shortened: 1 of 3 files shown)


westonpace (approved):
Claude had some minor suggestions but this seems like a clever way to fix the problem without slowing down queries.

---

> inv over-allocated. Vec::with_capacity(row_id_col.len()) but only non-tombstoned entries are pushed. For a high-deletion case this wastes memory until the Vec is dropped. A shrink_to_fit() after inv.sort_unstable_by_key(...) would recover it, or compute the capacity from a pre-pass. Low priority since it's transient.
> 
> Missing debug assert. After the loop, row_ids.len() must equal num_tokens.len(). Adding debug_assert_eq!(row_ids.len(), num_tokens.len()) before the return Ok(...) would cheaply enforce the invariant that the rest of the code relies on.
> 
> Comment verbosity. The 9-line block comment in from_columns explains the same reasoning already in the PR description. Per the project's standards ("only add a comment when the WHY is non-obvious"), the first 2–3 sentences earn their place; the last 3 restate the PR description and could be trimmed.

PR #7330, opened 2026-06-17
fix: grow posting_lists before indexed access in FTS with_position builder

### LINKED ISSUE #7313: bug: index out of bounds panic in inverted index builder when token_id > posting_lists.len()
## Description

`IndexWorker::process_batch()` in `rust/lance-index/src/scalar/inverted/builder.rs` panics with `index out of bounds` during `table.optimize()` when the FTS inverted index is rebuilt with position tracking enabled (`with_position: true`).

## Root Cause

In the position-tracking branch of `process_batch`, `tokens.add()` returns a token ID, then the code accesses `posting_lists[token_id]`. The posting_lists vector is resized only when `token_id == posting_lists.len()`, but `tokens.add()` can return an ID **greater than** the current length (e.g. when the token set was pre-populated from an existing index partition):

```
// Line 1325 in current main (line 856 in version at commit d88dde4):
let token_id = builder.tokens.add(token_text);
if token_id as usize == builder.posting_lists.len() {  // BUG: should be >=
    builder.posting_lists.push(
        PostingListBuilder::new_with_posting_tail_codec(true, posting_tail_codec),
    );
}
let posting_list = &mut builder.posting_lists[token_id as usize];  // PANICS here
```

The non-position branch (below) correctly handles this by calling `posting_lists.resize_with(tokens.len(), ...)` before accessing by index.

## Reproduction

Observed on three LanceDB tables during `table.optimize()`:

| Table | posting_lists.len() | token_id | Operation |
|---|---|---|---|
| conversations | 1731 | 4456 | optimize |
| cortex_engrams | 219 | 1663 | optimize |
| cortex_entities | 2398 | 2719 | optimize |

All three tables have FTS indexes with `with_position: true` (the default). The panic triggers when an existing token's ID exceeds the current posting_lists capacity because the token set was loaded from an old index partition during optimize.

## Fix

Change `==` to `>=` on the resize guard:

```
- if token_id as usize == builder.posting_lists.len() {
+ if token_id as usize >= builder.posting_lists.len() {
```

This matches the semantics used by the non-position branch which also handles the gap case.

## Environment

- lance commit: `d88dde4`
- Platform: Linux x86_64, Node.js 22

## Panic Output

```
thread 'tokio-rt-worker' panicked at .../lance-index/src/scalar/inverted/builder.rs:856:57:
index out of bounds: the len is 1731 but the index is 4456
```

Siriapps:
## What does this PR do?

Fixes an `index out of bounds` panic in `IndexWorker::process_batch()` when FTS indexing runs with `with_position: true` (the default). The `with_position` branch now grows `posting_lists` with `resize_with(token_idx + 1, ...)` before indexing by `token_id`, matching the pattern already used in the non-position branch and in `merge_from`.

## Why was this PR needed?

When `tokens.add()` returns a `token_id` greater than `posting_lists.len()` — e.g. after loading a legacy FTS partition with a stale `next_id` during `optimize_indices` — the old code only appended a posting list when `token_id == posting_lists.len()`. That skips growth for gaps and panics at `posting_lists[token_id]`.

Reported in production with `posting_lists.len()=1731` and `token_id=4456` (#7313). Changing `==` to `>=` with a single `push` is insufficient for that gap; `resize_with(token_idx + 1, ...)` is required.

## What are the relevant issue numbers?

Closes #7313

## Does this PR meet the acceptance criteria?

Per [CONTRIBUTING.md](https://github.com/lance-format/lance/blob/main/CONTRIBUTING.md):

- [x] Tests added for new/changed behavior (`test_process_batch_with_position_handles_token_id_gaps`)
- [x] All tests passing (`cargo test -p lance-index`)
- [x] Follows project style guide (`cargo fmt --all`)
- [x] Conventional Commits PR title (`fix:`)
- [x] No breaking changes introduced
- [x] Documentation updated — N/A (internal bug fix)

**Suggested label:** `critical-fix` (crash during `optimize_indices` on FTS indexes with `with_position: true`)

### DIFF
--- a/rust/lance-index/src/scalar/inverted/builder.rs
+++ b/rust/lance-index/src/scalar/inverted/builder.rs
@@ -3760,6 +3760,52 @@ mod tests {
         Ok(())
     }
 
+    /// Reproduces #7313: `with_position` indexing panics when `tokens.next_id` exceeds
+    /// `posting_lists.len()`, as observed during `optimize_indices` on legacy FTS partitions.
+    #[tokio::test]
+    #[should_panic(expected = "index out of bounds")]
+    async fn test_process_batch_with_position_panics_when_token_id_exceeds_posting_lists_len() {
+        const VOCAB_SIZE: usize = 1731;
+        const STALE_NEXT_ID: u32 = 4456;
+
+        let tokenizer = InvertedIndexParams::default().with_position(true).build().unwrap();
+        let store = Arc::new(CountingStore::new());
+        let id_alloc = Arc::new(AtomicU64::new(0));
+        let mut worker = IndexWorker::new(
+            tokenizer,
+            store,
+            id_alloc,
+            IndexWorkerConfig {
+                with_position: true,
+                format_version: InvertedListFormatVersion::V1,
+                fragment_mask: None,
+                token_set_format: TokenSetFormat::default(),
+                worker_memory_limit_bytes: u64::MAX,
+            },
+        )
+        .await
+        .unwrap();
+
+        let mut tokens = TokenSet::default();
+        for i in 0..VOCAB_SIZE {
+            tokens.add(format!("tok_{i}"));
+        }
+        tokens.next_id = STALE_NEXT_ID;
+        worker.builder.set_tokens(tokens);
+        let posting_tail_codec = worker.builder.posting_tail_codec;
+        worker
+            .builder
+            .posting_lists
+            .resize_with(VOCAB_SIZE, || {
+                PostingListBuilder::new_with_posting_tail_codec(true, posting_tail_codec)
+            });
+
+        worker
+            .process_batch(make_doc_batch("unseen_token_xyz", 0))
+            .await
+            .unwrap();
+    }
+
     #[test]
     fn test_resolve_worker_memory_limit_uses_default_when_unset() {
         let params = InvertedIndexParams::default();
--- a/rust/lance-index/src/scalar/inverted/builder.rs
+++ b/rust/lance-index/src/scalar/inverted/builder.rs
@@ -3761,8 +3761,12 @@ mod tests {
         Ok(())
     }
 
-    /// Regression test for #7313: `with_position` indexing must grow `posting_lists` when
-    /// `tokens.next_id` exceeds `posting_lists.len()`, as on legacy FTS partitions.
+    /// Regression test for #7313: `with_position` indexing must not panic when
+    /// `tokens.next_id` exceeds `posting_lists.len()`. The primary guard is in
+    /// `load_fst` (index.rs), which normalizes `next_id = max_id + 1` on load so
+    /// this gap never arises from a real partition. This test verifies `process_batch`
+    /// remains robust even when a stale `next_id` reaches it directly (e.g. in tests
+    /// or via direct builder construction).
     #[tokio::test]
     async fn test_process_batch_with_position_handles_token_id_gaps() {
         const VOCAB_SIZE: usize = 1731;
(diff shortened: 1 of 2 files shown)


Siriapps:
Hi @sinianluoye this addresses #7313. The fix grows posting_lists before indexed access when token_id exceeds the current vector length (stale next_id from legacy partitions). Would appreciate a review when you have time.

sinianluoye (approved):

rust/lance-index/src/scalar/inverted/builder.rs
```diff
@@ -1322,16 +1322,17 @@ impl IndexWorker {
...
                     let token_text = std::mem::take(&mut token.text);
                     let token_id = builder.tokens.add(token_text);
-                    if token_id as usize == builder.posting_lists.len() {
+                    let token_idx = token_id as usize;
+                    if token_idx >= builder.posting_lists.len() {
                         let old_posting_lists_overhead_size = (builder.posting_lists.capacity()
                             * std::mem::size_of::<PostingListBuilder>())
                             as u64;
-                        builder.posting_lists.push(
+                        builder.posting_lists.resize_with(token_idx + 1, || {
```
Xuanwo, line 1329: This turns a stale `next_id` into a persisted sparse token id while reload and tail-merge paths still assume dense token ids, so a later write/reload or multi-worker tail merge can reuse ids or hit another out-of-bounds access.

Siriapps: Thanks for the detailed feedback @Xuanwo. I traced through the failure: after `resize_with(token_idx + 1, ...)`, the FST ends up with 1732 tokens but max token ID 4456. On reload, `load_fst` already normalizes `next_id = map.len() = 1732`, which is correct for dense IDs but wrong here, the next free slot should be 4457. Subsequent writes then issue IDs 1732→4456, eventually colliding with the existing token at 4456 and silently corrupting the posting lists. The unit test missed this because it only covers the single process_batch call, not the serialize-reload-write cycle.

Should the fix prevent sparse IDs from being written at all: for example by normalizing `next_id` before `process_batch` runs so the gap is never created or would you prefer fixing the reload to use `max_token_id + 1` instead of `map.len()`? Happy to go either direction once I understand which invariant you'd like enforced.

sinianluoye (approved):

rust/lance-index/src/scalar/inverted/builder.rs
sinianluoye: @Siriapps Thank you for the PR  to help fix this issue.

@Xuanwo  Thanks for pointing out the issue - I think you're right that the original fix could create sparse token IDs, violating the dense invariant.

From my analysis, we shouldn't assume whether the current FST is dense or sparse - there could be other bugs that created sparse IDs beyond just the stale `next_id` case.

  The quick fix should be in `load_fst`, similar to how `load_arrow` already handles this correctly  . We should remap/detect sparse token IDs during load to ensure `next_id` is always correct.

  @Siriapps Could you help update the PR to fix `load_fst` instead? The approach would be:
  - In `load_fst`, compute the actual max token ID
  - Use `max_id + 1` as `next_id` instead of `map.len()`
  - This handles both dense and sparse FSTs correctly

  This is more robust than fixing it in `process_batch` and aligns with the existing `load_arrow` implementation.

Siriapps: Updated in 242cbc9c. Changed load_fst in index.rs to compute next_id as max_token_id + 1 by streaming the FST map, mirroring load_arrow. This corrects a stale next_id at load time so process_batch never sees the gap. Also updated the regression test comment to clarify that load_fst is now the primary guard. @Xuanwo @sinianluoye ready for another look when you have time!

sinianluoye (approved):

sinianluoye (approved):

sinianluoye:
hi @Xuanwo , @wjones127 Do we currently have any plan to fix this issue, which is preventing us from running the `OPTIMIZE` operation on tables in our production environment. If there is no plan in the short term, I may make some temporary modifications in our forked repository to resolve it. However, if the community has a fix planned, we would prefer to wait for the community's solution.

PR #7341, opened 2026-06-17
ci(java): parallelize macOS build, split publish; scope release lock updates

wjones127:
The Java publish workflow built the macOS native lib *and* published the multi-platform JAR in a single job, so the macOS build was gated behind the Linux builds (`needs: [linux-arm64, linux-x86]`) even though only the publish step depends on the Linux artifacts.

## Workflow restructure (`java-publish.yml`)

- **`build-linux`** — the two near-identical Debian 10 builds collapsed into a matrix (x86-64, arm64).
- **`build-macos`** — build-only, no `needs`, runs concurrently with the Linux builds.
- **`publish`** — a separate `ubuntu-latest` job that gathers all three native libs and packages/deploys with `-Dskip.build.jni=true`. No longer holds the expensive macOS runner during publish.

Critical path goes from `linux → (macos build + publish)` to `max(linux, macos) → publish`.

## Drop no-op `-P shade-jar`

That profile doesn't exist; shading is an always-on plugin bound to the `package` phase. The flag only produced a `could not be activated` warning.

## `cargo update --workspace` in release scripts

The release scripts ran bare `cargo update` after the version bump, which refreshed *all* transitive dependencies rather than just the local crates being re-versioned. This swept incidental dependency bumps into release commits — see [this run](https://github.com/lance-format/lance/actions/runs/26624032632/job/78456375994#step:5:2496), where the lock update pulled in `brotli`, `hyper`, `jiff`, `zerocopy`, and ~14 others on top of the intended `lance-*` version changes. `--workspace` re-pins only the workspace crates whose versions changed. Applied to all three release scripts (`create_release_branch.sh`, `release_common.sh`, `publish_beta.sh`) for consistency.

> [!NOTE]
> The PR's `pull_request` trigger exercises the full build + dry-run path. The `cargo update --workspace` change only takes effect on the next release-tooling run.

### DIFF
--- a/.github/workflows/java-publish.yml
+++ b/.github/workflows/java-publish.yml
@@ -28,10 +28,24 @@ permissions:
   contents: read
 
 jobs:
-  linux-arm64:
-    name: Build on Linux Arm64
-    runs-on: ubuntu-24.04-arm64-8x
+  build-linux:
+    name: Build on Linux ${{ matrix.arch }}
+    runs-on: ${{ matrix.runner }}
     timeout-minutes: 60
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - arch: x86-64
+            runner: ubuntu-24.04
+            docker_platform: linux/amd64
+            protoc_arch: x86_64
+            artifact: liblance_jni_linux_x86_64.zip
+          - arch: arm64
+            runner: ubuntu-24.04-arm64-8x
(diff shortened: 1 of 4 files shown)


xuanyu-z (approved):

jackye1995 (approved):
nice!!

PR #7350, opened 2026-06-18
fix(index): fail fast when FTS posting pipeline would deadlock

hfutatzhanghb:
## Summary

- Reject FTS inverted index builds early when fewer than 2 `lance-cpu` blocking threads are available, instead of hanging silently after logging `writing N posting lists`.
- Return a descriptive `invalid_input` error that explains the deadlock, mentions `LANCE_CPU_THREADS`, and suggests setting it to at least 2.
- Add a `tokio::sync::Semaphore` that limits concurrent `write_posting_lists_pipelined` calls to `available_cpu_threads - 1`, ensuring at least one `spawn_cpu` thread is always free for the consumer's nested page encoding inside `FileWriter::write_batch`. This prevents the concurrent-worker deadlock when `num_workers` equals all available CPU threads.
- Move the CPU-thread guard from `InvertedIndexBuilder::update_index` into `InnerBuilder::write_posting_lists` so that updates that never write posting lists (e.g., empty `new_data` with only deleted-fragment metadata) are not rejected.
- Add regression tests for the error message, for the concurrent-worker semaphore serialization, for semaphore permit scaling, and for the empty-update path.

## Root cause

`write_posting_lists` runs batch encoding on `spawn_cpu` while `FileWriter::write_batch` also submits column page encoding via `spawn_cpu`. With only one blocking thread, the producer blocks on the bounded channel while the consumer waits for nested encoding, causing a deadlock with no further log output.

With ≥2 threads but `num_workers` equal to all available threads, all producers occupy all `spawn_cpu` threads and consumers deadlock identically. The semaphore prevents this by serializing flush-phase concurrency.

## Test plan

- [x] `cargo test -p lance-index test_fts_posting_pipeline_cpu_threads_error_message`
- [x] `cargo test -p lance-index test_fts_posting_pipeline_write_posting_lists_deadlocks_with_one_cpu_thread`
- [x] `cargo test -p lance-index test_empty_update_with_one_cpu_thread_records_deleted_fragments`
- [x] `LANCE_CPU_THREADS=2 cargo test -p lance-index test_fts_posting_semaphore_serializes_with_two_cpu_threads`
- [x] `LANCE_CPU_THREADS=4 cargo test -p lance-index test_fts_posting_semaphore_permits_scale_with_threads`
- [x] `cargo fmt --all`
- [x] `cargo clippy -p lance-index --lib --tests -- -D warnings`


hfutatzhanghb:
@Xuanwo @BubbleCal @yanghua Hi, please help review this PR when have free time, Thanks very much!

hfutatzhanghb:
@claude review this pr.

rust/lance-index/src/scalar/inverted/builder.rs
```diff
@@ -181,6 +181,38 @@ fn resolve_worker_memory_limit_bytes(params: &InvertedIndexParams, num_workers:
         .unwrap_or(default_worker_memory_limit_bytes)
 }
 
+/// Minimum [`get_num_compute_intensive_cpus`] for the pipelined FTS posting write path.
+///
+/// `write_posting_lists` runs batch encoding on `spawn_cpu` while `FileWriter::write_batch`
+/// also submits column page encoding via `spawn_cpu`. With only one `lance-cpu` blocking
+/// thread the producer blocks on the bounded channel while the writer waits for encoding,
+/// which deadlocks with no further log output.
+const MIN_CPU_THREADS_FOR_FTS_POSTING_PIPELINE: usize = 2;
```
BubbleCal, line 190: A fixed threshold of 2 threads does not cover concurrent worker flushes. `resolve_num_workers` can still choose all available CPU threads, so with `LANCE_CPU_THREADS=2` two workers can each occupy a `spawn_cpu` producer and then both writers wait for nested page-encoding `spawn_cpu` work with no free thread. This should reserve capacity for nested encoding or limit the number of concurrent posting writers/workers.

rust/lance-index/src/scalar/inverted/builder.rs
```diff
@@ -437,6 +469,7 @@ impl InvertedIndexBuilder {
         stream: SendableRecordBatchStream,
         dest_store: &dyn IndexStore,
     ) -> Result<Vec<IndexFile>> {
+        check_fts_posting_pipeline_cpu_threads()?;
```
BubbleCal: This guard is broader than the deadlock path. The real risk is in `write_posting_lists`; checking here rejects updates that do not write posting lists, such as an empty `new_data` update that only records deleted-fragment metadata, even though that path cannot hit the pipelined writer deadlock.

BubbleCal (changes requested):
Requesting changes for two correctness issues in the CPU-thread deadlock guard.

rust/lance-index/src/scalar/inverted/builder.rs
hfutatzhanghb: Thanks for pointing that. Have fixed and add an unit test to cover this case.

rust/lance-index/src/scalar/inverted/builder.rs
hfutatzhanghb: @BubbleCal Thanks for the review. Addressed the threshold concern using a `tokio::sync::Semaphore`:

**Fix**: Added `fts_posting_write_semaphore()` with `available_cpu_threads - 1` permits (at least 1). `write_posting_lists` now acquires a permit before calling `write_posting_lists_pipelined`, so at most `available - 1` producers hold `spawn_cpu` threads concurrently, always leaving one free for the consumer's nested page encoding.

This handles the concurrent-worker case correctly:
- `LANCE_CPU_THREADS=2`, `num_workers=2`: semaphore has 1 permit → only 1 writer runs at a time → no deadlock.
- `LANCE_CPU_THREADS=4`, `num_workers=4`: semaphore has 3 permits → up to 3 writers concurrently, 1 thread remains free for consumers.

Two new tests verify the semaphore behavior in child processes with `LANCE_CPU_THREADS=2` and `LANCE_CPU_THREADS=4`.

rust/lance-index/src/scalar/inverted/builder.rs
hfutatzhanghb: Resolved in 0a3980d05 — the guard is now in `InnerBuilder::write_posting_lists` (the only entry point for the pipelined writer). Added `test_empty_update_with_one_cpu_thread_records_deleted_fragments` to verify the empty-update path succeeds with `LANCE_CPU_THREADS=1`.

Xuanwo (changes requested):
The new semaphore addresses the multi-worker flush case, but the one-CPU guard is still too broad. `write_posting_lists` calls `check_fts_posting_pipeline_cpu_threads()` before it knows whether the write can actually deadlock, so even a single-batch posting-list write is rejected under `LANCE_CPU_THREADS=1`.

A single-batch producer can send the only batch and release the sole `spawn_cpu` thread before the consumer encodes pages, so this rejects small valid FTS builds. In the review run, `LANCE_CPU_THREADS=1 cargo test -p lance-index scalar::inverted::builder::tests::test_write_posting_lists_batches_multiple_rows -- --exact --nocapture` failed from this guard.

Please either use a serial/non-pipelined fallback when only one CPU thread is available, or restructure the producer so it does not block inside the sole `spawn_cpu` thread before rejecting all small FTS builds in that configuration.

Xuanwo:
Closing this because #7423 was merged and removes the same FTS thread-pool deadlock root cause by using asynchronous producer sends. The fail-fast behavior here is a narrower mitigation. Please reopen if a distinct deadlock remains after #7423.

hfutatzhanghb:
Great Job!

PR #7351, opened 2026-06-18
feat(index): support Utf8View prefixes in SargableQueryParser

wombatu-kun:
Follow-up to #7310.

## Problem

#7310 added Utf8View support to the ngram TextQueryParser and noted the identical pre-existing gap in SargableQueryParser (starts_with / LIKE-prefix) as out of scope. The BTree/ZoneMap parser extracts string prefixes by matching only ScalarValue::Utf8 and ScalarValue::LargeUtf8. When the predicate literal is coerced to ScalarValue::Utf8View, the parser drops it, so starts_with(col, 'x') and col LIKE 'x%' do not use the scalar index and fall back to a full scan.

## Change

SargableQueryParser::visit_scalar_function (starts_with) and visit_like now accept a Utf8View literal/pattern and normalize the extracted prefix to ScalarValue::Utf8.

Normalize rather than preserve the variant: unlike the ngram path (where the pattern is just a regex string), the SargableQueryParser emits a SargableQuery::LikePrefix whose ScalarValue flows into the BTree. Page pruning (pages_between) compares the query bound against Utf8 page-statistics arrays with Arrow's type-dispatched make_comparator, which rejects a Utf8View bound vs Utf8 stats ("Can't compare arrays of different types"). Lance already normalizes Utf8View columns to Utf8 at write time, so the stored index data is always Utf8; normalizing the prefix to Utf8 matches that and needs no changes to the shared comparison code.

## Tests

New test_sargable_query_parser_utf8view exercises visit_scalar_function (starts_with) and visit_like directly with Utf8View literals/patterns, asserting the emitted LikePrefix(Utf8) query and recheck behavior, plus a Utf8 parity control. It fails on the pre-change parser and passes after.

### DIFF
--- a/rust/lance-index/src/scalar.rs
+++ b/rust/lance-index/src/scalar.rs
@@ -570,7 +570,9 @@ impl AnyQuery for SargableQuery {
             Self::IsNull() => col_expr.is_null(),
             Self::Equals(value) => col_expr.eq(Expr::Literal(value.clone(), None)),
             Self::LikePrefix(prefix) => match prefix {
-                ScalarValue::Utf8(Some(s)) | ScalarValue::LargeUtf8(Some(s)) => {
+                ScalarValue::Utf8(Some(s))
+                | ScalarValue::LargeUtf8(Some(s))
+                | ScalarValue::Utf8View(Some(s)) => {
                     // The prefix is a literal string. If it contains LIKE metacharacters
                     // (`_`, `%`, `\`) they must be escaped before appending the `%` wildcard;
                     // otherwise an inexact recheck (e.g. zone maps) would treat them as
@@ -578,6 +580,11 @@ impl AnyQuery for SargableQuery {
                     // When the prefix has no metacharacters we keep the plain
                     // `col LIKE 'prefix%'` form (no `ESCAPE`), identical to the prior behavior,
                     // so DataFusion's optimized prefix matcher still applies.
+                    // A `Utf8View` prefix is handled here too (rather than falling through to
+                    // the catch-all arm, which would rebuild the recheck without the trailing
+                    // `%` and silently turn prefix matching into equality matching) and is
+                    // normalized to a `Utf8` pattern below, mirroring how the parser normalizes
+                    // `Utf8View` to `Utf8` (see `expression.rs`).
                     let escaped = escape_like_pattern(s);
                     let needs_escape = escaped.as_str() != s.as_str();
                     let pattern = format!("{}%", escaped);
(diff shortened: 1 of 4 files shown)


rust/lance-index/src/scalar/expression.rs
```diff
@@ -386,6 +386,13 @@ impl ScalarQueryParser for SargableQueryParser {
                 Expr::Literal(ScalarValue::LargeUtf8(Some(s)), _) => {
                     ScalarValue::LargeUtf8(Some(s.clone()))
                 }
+                // Lance stores `Utf8View` columns as `Utf8` (normalized at write time), so a
+                // `Utf8View` literal is normalized to `Utf8` to match the indexed data: the
+                // BTree compares the query bound against `Utf8` page statistics at the Arrow
+                // level, which rejects a `Utf8View` bound.
+                Expr::Literal(ScalarValue::Utf8View(Some(s)), _) => {
```
Xuanwo, line 408: The new `Utf8View` prefix path can feed ZoneMap an inexact `LikePrefix` query, but the later recheck reconstructs the predicate as an unescaped LIKE pattern. Prefixes containing `_` or `%` can therefore match rows that do not satisfy the original `starts_with` predicate.

rust/lance-index/src/scalar/expression.rs
```diff
@@ -415,17 +422,21 @@ impl ScalarQueryParser for SargableQueryParser {
...
         let (prefix, needs_refine) = extract_like_leading_prefix(pattern_str, like.escape_char)?;
 
-        // Create the prefix ScalarValue with the same type as the pattern
+        // Create the prefix ScalarValue with the same type as the pattern. `Utf8View` is
+        // normalized to `Utf8` because Lance stores `Utf8View` columns as `Utf8`, and the
+        // downstream BTree compares the query bound against `Utf8` page statistics at the
+        // Arrow level (a `Utf8View` bound would fail that comparison).
         let prefix_value = match pattern {
-            ScalarValue::Utf8(_) => ScalarValue::Utf8(Some(prefix)),
+            ScalarValue::Utf8(_) | ScalarValue::Utf8View(_) => ScalarValue::Utf8(Some(prefix)),
```
Xuanwo, line 460: The shared sargable parser now emits `LikePrefix` for `Utf8View` prefixes even when the selected index is Bitmap, which rejects that query during search. A query that previously fell back to ordinary filtering can now fail once planned as an index scan.

rust/lance-index/src/scalar/expression.rs
wombatu-kun: Done 528b9f08a - `LikePrefix::to_expr` now escapes `_`/`%`/`\` and reconstructs the recheck as `LIKE 'prefix%' ESCAPE '\'` (only when the prefix actually contains metacharacters), so the inexact zone-map recheck no longer over-matches.

rust/lance-index/src/scalar/expression.rs
wombatu-kun: Done 528b9f08a - the bitmap index now configures the parser with `without_like_prefix`, so `LIKE`/`starts_with` fall back to ordinary filtering instead of emitting a `LikePrefix` that bitmap search would reject.

rust/lance-index/src/scalar.rs
```diff
@@ -553,16 +569,36 @@ impl AnyQuery for SargableQuery {
             )),
             Self::IsNull() => col_expr.is_null(),
             Self::Equals(value) => col_expr.eq(Expr::Literal(value.clone(), None)),
-            Self::LikePrefix(prefix) => {
-                let pattern = match prefix {
-                    ScalarValue::Utf8(Some(s)) => ScalarValue::Utf8(Some(format!("{}%", s))),
-                    ScalarValue::LargeUtf8(Some(s)) => {
-                        ScalarValue::LargeUtf8(Some(format!("{}%", s)))
+            Self::LikePrefix(prefix) => match prefix {
```
Xuanwo, line 572: The `LikePrefix` recheck still only preserves prefix semantics for `Utf8` and `LargeUtf8`; a public parser that constructs `LikePrefix(Utf8View(...))` would rebuild the recheck without the trailing wildcard, so prefix pruning can return different results.

wombatu-kun: Done a154dbdec - to_expr now matches Utf8View in the LikePrefix arm and normalizes it to a Utf8 pattern with the trailing %, so prefix pruning keeps starts-with semantics for a Utf8View prefix.

Xuanwo (approved):

Issue #7352, opened 2026-06-18
Low recall for IVF_SQ / IVF_HNSW_SQ with metric="dot" on MSMARCO


cwj0bzxg:
Observed unexpectedly low recall when using SQ-based indexes with dot‑product search on the MSMARCO 1M dataset.

**Environment**
- Lance tag: `v8.0.0-beta.18` (commit `909dea18b1`)
- Python package: `8.0.0-beta.18`
- Dataset: MSMARCO Web Search 1M – 1M base vectors, 9,376 queries, dim 768
- Metric: `dot`; ground‑truth: `msmarco-1M-gt100`; evaluated metric: `recall@10`

**Index Configurations**
- `IVF_HNSW_SQ`: `num_partitions=1`, `num_bits=8`, `m=16`, `ef_construction=200`, `nprobes=1`, `ef ∈ {20,40,80,160,320,640}`
- `IVF_SQ`: `num_partitions=1024`, `num_bits=8`, `nprobes ∈ {16,32,64,96,128}`
- `IVF_HNSW_FLAT` (baseline): `num_partitions=1`, `m=16`, `ef_construction=200`, `nprobes=1`, `ef ∈ {20,40,80,160,320,640}`

**Results**

| Index           | nprobes | ef   | recall@10 |
|-----------------|---------|------|-----------|
| `IVF_HNSW_SQ`   | 1       | 20   | 0.0250    |
| `IVF_HNSW_SQ`   | 1       | 40   | 0.0370    |
| `IVF_HNSW_SQ`   | 1       | 80   | 0.0490    |
| `IVF_HNSW_SQ`   | 1       | 160  | 0.0577    |
| `IVF_HNSW_SQ`   | 1       | 320  | 0.0641    |
| `IVF_HNSW_SQ`   | 1       | 640  | 0.0684    |
| `IVF_SQ`        | 16      | N/A  | 0.0918    |
| `IVF_SQ`        | 32      | N/A  | 0.0891    |
| `IVF_SQ`        | 64      | N/A  | 0.0862    |
| `IVF_SQ`        | 96      | N/A  | 0.0841    |
| `IVF_SQ`        | 128     | N/A  | 0.0827    |

**Baseline (`IVF_HNSW_FLAT`)**

| Index           | nprobes | ef   | recall@10 |
|-----------------|---------|------|-----------|
| `IVF_HNSW_FLAT` | 1       | 20   | 0.5179    |
[...]

PR #7355, opened 2026-06-18
fix: compute SQ dot distance from dequantized values

cwj0bzxg:
This PR fixes Dot distance computation for scalar-quantized vectors. #7352

Previously, the SQ Dot path computed the dot product directly from `u8` quantized codes and only applied a scale factor. This is incorrect when SQ uses a non-zero `lower_bound`, because each code represents an offset value:

```
value ≈ lower_bound + step * code
```

As a result, the old Dot path missed the offset-related terms and could produce a very different ranking from the actual vector values. This caused severe recall degradation for SQ indexes with `metric="dot"`.

The fix computes SQ Dot using the full expansion:

```
dot(x, q) ≈ sum_i((lower_bound + step * cx_i) * (lower_bound + step * cq_i))
```

Equivalently:

```
dot =
    step² · sum_i(cx_i · cq_i)
  + lower_bound·step · sum_i(cx_i)
  + lower_bound·step · sum_i(cq_i)
  + dim·lower_bound²
```

`distance = 1 - dot` is then used for Dot distance, matching the existing Dot distance convention.

## Changes

- Fix SQ Dot distance to include the `lower_bound` offset terms.
- Keep the existing SQ L2 and Cosine paths unchanged.
- Cache / compute the SQ query code sum needed by the Dot formula.
- Add Rust unit tests for:
  - Dot distance from a float query.
  - Dot distance from an indexed vector id.
  - Constant-bound SQ behavior.

## Validation

I rebuilt the Python extension with this patch and reran MSMARCO WebSearch 1M Dot benchmarks.

Dataset:
- 1M base vectors
- 9,376 queries
- dimension 768
- metric: Dot
- evaluated with recall@10

Before this fix, `IVF_HNSW_SQ` recall@10 was only around `0.0250` to `0.0684` across `ef=20..640`, while the `IVF_HNSW_FLAT` baseline reached `0.5179` to `0.9377`.

After this fix:

### IVF_HNSW_SQ

| ef | QPS | recall@10 |
|---:|---:|---:|
| 20 | 944.54 | 0.5176 |
| 40 | 880.87 | 0.6455 |
| 80 | 762.09 | 0.7448 |
| 160 | 607.29 | 0.8198 |
| 320 | 453.09 | 0.8716 |
| 640 | 305.57 | 0.9040 |

### IVF_SQ

| nprobes | QPS | recall@10 |
|---:|---:|---:|
| 16 | 578.43 | 0.6392 |
| 32 | 417.61 | 0.7352 |
| 64 | 272.83 | 0.8104 |
| 96 | 205.88 | 0.8456 |
| 128 | 165.96 | 0.8684 |

### IVF_HNSW_FLAT (baseline)

| ef | QPS | recall@10 |
|---:|---:|---:|
| 20 | 886.27 | 0.5195 |
| 40 | 773.42 | 0.6507 |
| 80 | 627.24 | 0.7563 |
| 160 | 470.82 | 0.8390 |
| 320 | 315.20 | 0.8983 |
| 640 | 193.39 | 0.9373 |

With the corrected distance formula, SQ Dot recall is restored to the same range as the Flat baseline. `IVF_HNSW_SQ` is close to `IVF_HNSW_FLAT` at the same `ef`, while generally providing higher QPS. The remaining recall gap at high `ef` is expected from quantization loss rather than an incorrect distance formula.

### DIFF
--- a/rust/lance-index/src/vector/sq/storage.rs
+++ b/rust/lance-index/src/vector/sq/storage.rs
@@ -16,7 +16,7 @@ use lance_core::deepsize::DeepSizeOf;
 use lance_core::{Error, ROW_ID, Result};
 use lance_file::previous::reader::FileReader as PreviousFileReader;
 use lance_io::object_store::ObjectStore;
-use lance_linalg::distance::{DistanceType, dot_distance, l2_u8::l2_u8};
+use lance_linalg::distance::{DistanceType, dot_u8::dot_u8, l2_u8::l2_u8};
 use lance_table::format::SelfDescribingFileReader;
 use object_store::path::Path;
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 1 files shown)


rust/lance-index/src/vector/sq/storage.rs
```diff
@@ -374,23 +374,56 @@ impl VectorStore for ScalarQuantizationStorage {
...
+fn sq_dequantized_dot_distance(
+    sq_code: &[u8],
+    sq_code_sum: u32,
+    query_sq_code: &[u8],
+    query_sq_code_sum: u32,
+    step: f32,
+    lower_bound: f32,
+) -> f32 {
+    let code_dot = dot_u8(sq_code, query_sq_code) as f32;
+    let dot = step * step * code_dot
```
Xuanwo, line 415: This expanded affine dot calculation is performed in `f32`, and the large offset terms can cancel in high-dimensional near-zero vectors enough to flip SQ Dot rankings.

rust/lance-index/src/vector/sq/storage.rs
```diff
@@ -592,4 +661,71 @@ mod tests {
...
+        let batch = create_record_batch_from_codes(
+            vec![10, 11],
+            2,
+            vec![
+                255, 255, // dequantized to [245, 245]
+                100, 0, // dequantized to [90, -10]
+            ],
+        );
+        let storage =
+            ScalarQuantizationStorage::try_new(8, DistanceType::Dot, -10.0..245.0, [batch], None)
```
Xuanwo, line 676: The new coverage only exercises unit-step and constant bounds, so regressions in the `step` and `step * step` terms can pass while arbitrary-range SQ Dot distances remain wrong.

Xuanwo:
Closing this because #7481 was merged and covers the same SQ dot-distance offset/lower-bound error in the same storage path, with IVF_SQ and HNSW_SQ regression coverage. Please reopen if this PR contains a distinct remaining case.

PR #7356, opened 2026-06-18
feat: allow tuning miniblock value chunks to 32k

### LINKED ISSUE #7326: Support 32K miniblock chunks
## Summary

Allow Lance miniblock chunks to contain up to 32K logical values when the existing metadata representation can safely express the chunk size.

This is a chunking capability, independent of RLE v2. It reduces chunk boundary fragmentation for encodings that benefit from longer logical runs.

Related: #6634

## Scope

- Allow `log_num_values <= 15` where the miniblock metadata format supports it.
- Raise the miniblock value cap from the current lower limit to 32768.
- Keep the compressed bytes-per-chunk limit intact.
- Add compatibility gating if older readers cannot safely reject or read 32K chunks.
- Add round-trip and metadata-boundary tests.

## Acceptance Criteria

- 32K-value miniblock chunks can be written and read by new Lance.
- Existing datasets remain readable.
- The implementation documents whether this requires a table-level feature flag.
- Tests cover max-size chunks, smaller chunks, and invalid metadata.

Xuanwo:
This allows miniblock writers to use up to 32K logical values per chunk when explicitly configured via `LANCE_MINIBLOCK_MAX_VALUES`, while keeping the default at 4096.

The file format already stores `log_num_values` in 4 bits, so the writer-side guard can allow values up to 15 without requiring the large-chunk metadata path. The compressed byte-size limits remain enforced.

Fixes #7326.

### DIFF
--- a/rust/lance-encoding/src/encodings/logical/primitive.rs
+++ b/rust/lance-encoding/src/encodings/logical/primitive.rs
@@ -3701,12 +3701,7 @@ struct SerializedFullZip {
 //
 // If we directly record the size in bytes with 12 bits we would be limited to
 // 4KiB which is too small.  Since we know each mini-block consists of 8 byte
-// words we can store the # of words instead which gives us 32KiB.  We want
-// at least 24KiB so we can handle even the worst case of
-// - values compressed into an 8186 byte buffer
-// - 4 bytes to describe rep & def lengths
-// - 16KiB of rep & def buffer (this will almost never happen but life is easier if we
-//   plan for it)
+// words we can store the # of words instead which gives us 32KiB.
 //
 // Second, each chunk in a mini-block is aligned to 8 bytes.  This allows multi-byte
 // values like offsets to be stored in a mini-block and safely read back out.  It also
--- a/rust/lance-encoding/src/encodings/logical/primitive.rs
+++ b/rust/lance-encoding/src/encodings/logical/primitive.rs
@@ -3703,7 +3703,7 @@ struct SerializedFullZip {
 // 4KiB which is too small.  Since we know each mini-block consists of 8 byte
 // words we can store the # of words instead which gives us 32KiB.  We want
 // at least 24KiB so we can handle even the worst case of
-// - 4Ki values compressed into an 8186 byte buffer
+// - values compressed into an 8186 byte buffer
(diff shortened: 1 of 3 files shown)


rust/lance-encoding/src/encodings/logical/primitive.rs
```diff
@@ -3703,7 +3703,7 @@ struct SerializedFullZip {
 // 4KiB which is too small.  Since we know each mini-block consists of 8 byte
 // words we can store the # of words instead which gives us 32KiB.  We want
 // at least 24KiB so we can handle even the worst case of
-// - 4Ki values compressed into an 8186 byte buffer
+// - values compressed into an 8186 byte buffer
 // - 4 bytes to describe rep & def lengths
 // - 16KiB of rep & def buffer (this will almost never happen but life is easier if we
 //   plan for it)
```
westonpace: This whole paragraph should probably be removed but it does explain my original reasoning for the 4Ki limit.

At the time we did not have a very good solution for the "there is too much rep/def information and it doesn't fit in 32KiB" problem.

I thought I could avoid this problem by limiting us to 4Ki values since that would limit how much rep/def we had (e.g. worst case 16KiB of rep/def).

However, as you know, I did not anticipate sparse lists.  With a sparse list we can have more than 4 bytes of rep/def per value!  So the limit of 4Ki values did not solve the problem anyways.

Since then, I think we have fixed the sparse list problem by providing escape valves if we end up with too much rep/def information.  Although maybe those escape valves are very narrowly scoped?

Anyways, this comment edit (removing 4Ki) doesn't work because then the last line (16KiB of rep & def buffer) doesn't make sense (that was 4 * 4Ki but if we allow up to 32Ki then it would be 128KiB).  We should just remove the entire paragraph.

westonpace (approved):
I think bumping the max up to 32Ki is fine.  I think there is some small chance we might end up seeing the "too much rep/def information and it doesn't all fit in the miniblock" error again but we can tackle that problem a different way when it surfaces.

rust/lance-encoding/src/encodings/logical/primitive.rs
Xuanwo: Got it, will remove.

PR #7360, opened 2026-06-18
feat: support COUNT(*) pushdown on stable row id datasets

wkalt:
`COUNT(*)` pushdown (`count_pushdown` -> `CountFromMaskExec`) was disabled on
datasets using stable row ids, so those counts fell back to a full scan.

The fast path intersects the scalar-index prefilter and the deletion mask (both
in stable-id space) with a fragments-allow universe -- but that universe was
built in row-address space. ANDing across the two id spaces dropped rows in
fragments > 0, so the rule was gated off entirely under stable row ids. The fix
builds the universe in stable-id space via the live-id deletion mask (restricted
to the covered fragments). An unfiltered count has no prefilter, so the universe
is never materialized -- the answer comes straight from fragment metadata.

Benchmark (synthetic: 5M rows, 50 fragments, ~1% scattered deletions, BTree on
the filter column), `cargo bench -p lance --bench count_pushdown`, this branch
vs `main`:

| benchmark | before | after | speedup |
|---|---|---|---|
| `count_unfiltered` | 44.5 ms | 69 us | ~640x |
| `count_filtered_1pct` | 1.53 ms | 258 us | ~5.9x |
| `count_filtered_50pct` | 28.3 ms | 5.5 ms | ~5.1x |

### DIFF
--- a/rust/lance/src/io/exec/count_from_mask.rs
+++ b/rust/lance/src/io/exec/count_from_mask.rs
@@ -31,7 +31,7 @@ use datafusion::physical_plan::{
 };
 use datafusion_physical_expr::EquivalenceProperties;
 use datafusion_physical_expr::aggregate::AggregateFunctionExpr;
-use futures::{StreamExt, TryStreamExt};
+use futures::{StreamExt, TryStreamExt, stream};
 use lance_core::{Error, Result};
 use lance_select::{RowAddrMask, RowAddrSelection, RowAddrTreeMap};
 use lance_table::format::Fragment;
@@ -40,6 +40,7 @@ use tracing::instrument;
 
 use super::utils::InstrumentedRecordBatchStreamAdapter;
 use crate::Dataset;
+use crate::dataset::rowids::load_row_id_sequences;
 use crate::index::prefilter::DatasetPreFilter;
 
 /// An execution node that computes a `COUNT(*)`-style aggregate from an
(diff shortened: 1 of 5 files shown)


rust/lance/src/io/exec/count_from_mask.rs
```diff
@@ -287,16 +268,118 @@ impl CountFromMaskExec {
...
+        let dataset_fragments: RoaringBitmap =
+            dataset.fragments().iter().map(|f| f.id as u32).collect();
+        let fragments_covered = match restrict_to_fragments {
+            Some(restrict) => dataset_fragments & restrict,
+            None => dataset_fragments,
+        };
 
-        let combined = Self::combine_masks(fragments_allow, prefilter, deletion_mask);
-        let count = Self::count_from_mask(&combined, dataset.as_ref())?;
+        // Under stable row ids the prefilter and deletion masks are in stable-id
```
westonpace, line 359: > Under stable row ids the prefilter and deletion masks are in stable-id space

Is this statement true?  Are scalar index search results in stable-id space when we are using stable row ids?  For some reason I thought we still used row addresses in the index even if we have stable row id.  However, the tests pass, so maybe I'm just wrong?

westonpace (approved):
I have one question I'd like you to follow-up on first but I could just be mistaken so I'll mark approve for now.

rust/lance/src/io/exec/count_from_mask.rs
wkalt: It uses row_id whether stable or not: https://github.com/lance-format/lance/blob/a13a760f8a901143a843c292474866cd55175cc7/rust/lance-index/src/scalar/btree.rs#L3180

PR #7362, opened 2026-06-18
feat(mem-wal): add ShardWriter::put_no_wait

hamersaw:
## What

Adds `ShardWriter::put_no_wait`, a variant of `put` that performs the visible in-memory insert and triggers the durable WAL flush, then returns the `BatchDurableWatcher` **without** awaiting it. A thin wrapper restores `put`'s behavior (await the watcher), so existing callers are unchanged.

`put_memtable` is split into:
- `put_memtable_no_wait` — the in-memory critical section (insert under `state_lock` + `track_batch_for_wal` + flush triggers) followed by `trigger_flush`, returning `(WriteResult, Option<BatchDurableWatcher>)`.
- `put_memtable` — calls the above, then `watcher.wait()`.

`BatchDurableWatcher` and `WriteResult` are re-exported from `mem_wal`.

## Why

Lets an external caller hold its own serialization lock across only the in-memory read-merge-insert critical section and await durability **after** releasing it, so concurrent durable flushes still coalesce. The in-memory insert stays guarded by the writer's `state_lock`, so `BatchStore`'s single-writer invariant holds regardless of the external lock — `state_lock` is intentionally **not** skipped.

This is the lance-side primitive for sophon's WAL partial-column-update path (read fresh tier → merge → insert under a per-bucket lock, durability awaited outside it).

## Tests

`test_put_no_wait_durable_visible_then_durable` (row visible before durability, watcher resolves) and `test_put_no_wait_non_durable_returns_no_watcher`.


jackye1995 (approved):
looks good to me

PR #7364, opened 2026-06-18
fix(python): handle empty filtered shards in ShardedBatchSampler

shiyan-xu-ai:
## Summary

- Fix `ShardedBatchSampler._shard_scan` so filtered scans with zero matches or empty per-rank shards yield an empty stream instead of raising.
- Preserve global round-robin sharding across filtered scanner batches by carrying `rows_to_skip` from the full batch length, not a sliced view.
- Skip `batch.take([])` when a rank owns no rows in a batch (PyArrow has no `array_take(int64, null)` kernel).
- Add regression tests for empty shard, zero-match filter, and cross-fragment carryover.

## Problem

When `ShardedBatchSampler` is called with a `filter`, the filtered path can crash in two cases:

1. **Global zero-match** — filter matches no rows -> `pa.Table.from_batches([])` raises `ValueError`.
2. **Empty per-rank shard** — sparse filter with `world_size > 1` leaves some ranks with zero rows -> `batch.take([])` raises `ArrowNotImplementedError`.

The unfiltered path already handles emptiness; this aligns the filtered path with that contract.

## Test plan

- [x] `test_sharded_batch_sampler_empty_filtered_shard` (randomize on/off)
- [x] `test_sharded_batch_sampler_filtered_carryover_across_fragments`
- [ ] `uv run pytest python/tests/test_sampler.py` (full sampler suite)

Found during distributed training work with sparse segment filters (AppliedIntuition core-stack).

### DIFF
--- a/python/python/lance/sampler.py
+++ b/python/python/lance/sampler.py
@@ -452,12 +452,23 @@ class ShardedBatchSampler(Sampler):
             filter=filter,
             scan_in_order=True,
         ).to_batches():
-            batch = batch.slice(rows_to_skip, batch.num_rows - rows_to_skip)
-            # Take every Nth row
-            indices = list(range(0, batch.num_rows, self._world_size))
-            rows_to_skip = (
-                self._world_size - (batch.num_rows % self._world_size)
-            ) % self._world_size
+            # Take this rank's rows by their global round-robin position. The
+            # scan is in order, so concatenating every batch reproduces the full
+            # filtered row sequence; this rank owns positions rank, rank + N,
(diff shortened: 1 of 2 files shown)


shiyan-xu-ai:
Superseded: reopening from personal fork xushiyan/lance now that SSH auth is configured for the xushiyan account.

shiyan-xu-ai:
Closing in favor of a new PR from personal fork `xushiyan/lance` (branch pushed; file via xushiyan GitHub account).

Issue #4583, opened 2025-08-27
Differentiate streaming and materialized inputs for add and merge insert


wjones127:
Most of our APIs assume inputs are a stream. This is nice in that it supports larger-than-memory writes. However, if data is fully materialized, we can often do things more optimally. To give a few examples:

1. If we have 2 million rows in memory we want to insert, we can write two data files in parallel. Currently we write them sequentially.
2. To support retries for write operations, we buffer data on disk. This could by bypassed if the data is in memory.
3. For merge_insert, we can compute basic statistics like `num_rows` and `num_bytes`, which can be used by DataFusion to optimize the join order. Currently we always use the table id column as the build side, but for large tables that is suboptimal.

Having an API would also support other downstream use cases: https://github.com/lancedb/lancedb/issues/2602

## API

In Rust, define an enum and conversion traits to take common input using generic APIs:

```rust
struct InputData {
    Stream(SendableRecordBatchStream)
    Materialized {
        batches: Vec<RecordBatch>,
        schema: SchemaRef,
    }
}

pub fn insert(data: impl Into<InputData>) -> { ... }

impl From<RecordBatch> for InputData { ... }
impl From<Vec<RecordBatch>> for InputData { ... }
impl From<Box<dyn RecordBatchReader>> for InputData { ... }
```

In Python, we want to make sure various inputs gets converted to the correct type.

Materialized:
* `pa.Table`
* `pd.DataFrame`
* `pa.RecordBatch`

Stream:
* `pa.RecordBatchReader`
* `pa.Dataset`
* `pa.Scanner`

## TODO

* [ ] Define `InputData` and conversion traits
* [ ] Change write APIs to take `impl Into<InputData>`
* [ ] Make sure `merge_insert` converts `InputData::Materialized` into [MemTable](https://docs.rs/datafusion/latest/datafusion/datasource/struct.MemTable.html) instead of `OneShotPartitionStream`.
     * This should solve use case 3
* [ ] Change `new_source_iter` to not spill when using `InputData::Materialized`
     * This should solve use case 2

(Note: we'll leave use case 1 for a follow up)

PR #7367, opened 2026-06-18
perf(merge_insert): defer reading non-source columns via a logical late-materialization rule

### LINKED ISSUE #7363: perf(merge_insert): defer reading non-source columns via late-materialization optimizer rule
## Background

Partial-schema (subset-of-columns) upserts now run through the v2 DataFusion plan path (`execute_uncommitted_v2` / `create_plan` in `rust/lance/src/dataset/write/merge_insert.rs`). Since Lance implements updates as delete + append, the plan must read the *missing* (non-source) columns for matched rows in order to rewrite full rows.

Today `create_plan` fills those columns via `with_column(field, col("target.<field>"))` above the join (around `merge_insert.rs:1521`). DataFusion projection pushdown faithfully keeps those referenced columns alive, so the **target scan reads the missing columns for every target row**, then the join discards all but the matched ones. For a selective backfill on a wide table this is the read-amplification problem: a ~1% match still scans 100% of the wide columns.

`explain_plan` test `test_merge_insert_subcols_v2_explain_plan` currently *asserts* this behavior — the target `LanceRead` projects the filled column and carries it through `HashJoinExec`.

This is the read-side half of the "efficient merge_insert" work. The write side still rewrites all columns of matched rows (delete + append); that is **out of scope here** (see #4193 / the column-replacement path).

## Approach: late materialization via a physical optimizer rule

Rather than hand-building a take node in `create_plan`, generalize Lance's existing late-materialization mechanism (today scoped to single-scan filters via `MaterializationStyle`, `scanner.rs:245`) into a plan-level **physical optimizer rule** that defers reads across row-reducing operators.

The rule, run after projection pushdown:

1. Find columns that originate from a `LanceScanExec` (fragment-addressable) and are first *used* above a row-reducing boundary (`HashJoinExec`, `FilterExec`).
2. Narrow the scan's projection to drop those columns, ensure it emits `_rowaddr`, and insert a `TakeExec` just above the boundary to fetch them by `_rowaddr`.
3. `CoalesceTake` (`io/exec/optimizer.rs`) already merges adjacent takes.

Consequence: **`create_plan` can stay naive** — its `col("target.<field>")` fill is exactly a "column read by the scan but only used above the join," so the rule defers it automatically. No merge-insert-specific logical node is needed.

### Scope: merge_insert only, for now

To keep blast radius small, apply the rule **only at the merge_insert call site** (in `create_plan`, after `create_physical_plan`), not in the session's global `PhysicalOptimizer` list. Write the rule generically (match any `LanceScanExec` feeding a row-reducing operator); generalizing later is just moving it into the global optimizer rules. A follow-up can enable it for generic queries once it's benchmarked across workloads.

## Components / implementation order

1. **Null-tolerant `TakeExec`** — prerequisite, independently testable. The merge_insert join is an *outer* join, so insert (source-only) rows have a null `target._rowaddr` and must yield NULL for the deferred columns. `TakeExec::map_batch` currently asserts `row_addrs.null_count() == 0` (`take.rs:222`); replace with real handling: take only non-null addresses, scatter back into full-length arrays with nulls at the insert positions.
2. **Late-materialization physical rule** — written generically, applied only at the merge_insert call site. Guards: only defer `LanceScanExec`-sourced columns (never source/one-shot-stream columns, which have no address); verify `_rowaddr` survives to the take point.
3. **Selectivity gate.** Even scoped to merge_insert, a ~100%-match update would regress (sequential scan → random take of every matched row's wide columns). Gate on column width / storage (reuse `MaterializationStyle::Heuristic`: cloud + wide → late) *and* join selectivity. The join-cardinality estimate from the materialized-input work (#4583) is the signal this gate should read; until that lands, fall back to the conservative width/storage heuristic (covers the inherently-selective backfill case).

## Tests

- Flip `test_merge_insert_subcols_v2_explain_plan`: the target `LanceRead` projection must *exclude* the deferred column, and a `Take` must appear above the join.
- Wide-table selective partial update: assert the wide deferred column is not read for all rows (read-bytes / column-not-loaded).
- Inserts (null `_rowaddr` take path), `UpdateIf` referencing a non-source target column (must stay alive via `__action`, not double-fetched), multi-fragment.

## Out of scope

- Write-side amplification (delete + append rewrites all columns of matched rows). The in-place column-replacement write path belongs to #4193.
- Enabling the rule for generic (non-merge_insert) queries.

Related: #4193 (parent), #4583 (materialized inputs / join cardinality), #3481 (non-indexed merge_insert loads too many columns).

wjones127:
Partial-schema upserts fill missing (non-source) columns from the target side of the join via `col("target.<field>")`, so projection pushdown keeps those columns in the target scan — reading wide columns for **every** target row even when only a few rows match.

This defers those reads to *after* the join shrinks the row count. It is done with a generic **logical** optimizer rule rather than a physical one, so the deferral is driven by which relation a column comes from (qualifier) and the scan-narrowing falls out of DataFusion's existing `OptimizeProjections` — no positional index remapping or hardcoded join side.

### How it works

A new logical extension node, `LateTakeNode`, is inserted above the row-reducing join:

```
LateTake(deferred=[wide_col])           # re-fetches wide_col by _rowaddr, appends it
  Projection(carried columns only)      # drops deferred + the redundant equi-key
    Join(target, source)
      LanceRead(target, projection=[key, _rowaddr])   # wide_col no longer scanned
      <source>
```

- `LateMaterializeJoin` (the rule) finds a join with a Lance scan side that emits `_rowaddr`, picks the wide data columns that are only carried through (not join keys, not used in a join filter), and wraps the join. A width/storage gate (`is_wide_column`) only defers columns wide enough that re-fetching by address beats scanning them for every row.
- The node reports via `necessary_children_exprs` that it does *not* need the deferred columns from its child, so `OptimizeProjections` prunes them from the scan while keeping `_rowaddr`. Downstream references resolve the appended columns by name — the plan's output schema is unchanged.
- `LateTakePlanner` lowers the node to the existing physical `TakeExec`.
- The rule is applied only at the merge_insert call site (between two `session_state.optimize` passes), not the session-wide optimizer, to bound its blast radius.

A **normalizing projection** is inserted between the join and the take. Without it the opaque extension node forces `HashJoinExec` to emit its full `left ++ right` output, which for an equi-join on a shared key name yields two arrow fields named `key` (qualifiers are erased at the physical level) — a duplicate `TakeExec` cannot merge by name. The projection keeps only the carried columns, so the physical join gets a tight output projection.

`TakeExec` is now null-tolerant: outer-join insert rows have a null `target._rowaddr`, so the take fetches only the non-null addresses and scatters results back with NULLs at the insert positions. The taken fields are marked nullable when the take sits above an outer join.

### Tests
- `late_take.rs` unit tests for the rule and node in isolation: wide-column deferred (scan narrowed, output schema unchanged), narrow/join-key/duplicate-name columns *not* deferred, qualifier-aware matching, outer-join nullable fields, `necessary_children_exprs`, the `is_wide_column` gate (local vs cloud thresholds, blob exclusion), and end-to-end execution parity vs the un-deferred plan (inner + outer with null `_rowaddr` scatter).
- `merge_insert` subcols tests: the target `LanceRead` excludes the deferred column and a `Take` appears above the join; on a selective update the wide column is materialized only for the matched rows (`Take` output_rows = matches, `LanceRead` still visits all rows); an `UpdateIf` condition on the deferred target column evaluates correctly; multi-fragment, multi-wide-column, and scalar-index cases pass.
- `TakeExec` null-address unit tests.

### Out of scope
Write-side amplification (delete + append rewrites all columns of matched rows) stays on #4193.

Closes #7363


rust/lance/src/io/exec/late_take.rs
```diff
@@ -0,0 +1,1019 @@
...
+/// worth deferring only if it is "wide" for the backing storage — a
+/// variable-width type (strings, lists, vectors) or a fixed-width type above the
+/// per-row byte threshold (1KB on cloud storage, 10 bytes on local). Narrow
+/// columns are cheaper to read in the sequential scan than to re-fetch by
+/// address.
+///
+/// Without a join-cardinality estimate (tracked in #4583) we cannot gate on
+/// match selectivity, so we fall back to width alone. This covers the
+/// inherently selective backfill case the feature targets; a follow-up can
+/// incorporate cardinality once it is available.
```
wjones127: issue(blocking): The comments in this PR carry a little too much use-case context. Please strip them down so the are generic across use cases and aren't specific to this issue.

rust/lance/src/io/exec/late_take.rs
```diff
@@ -0,0 +1,1019 @@
...
+        nullable_extra: bool,
+    ) -> DFResult<DFSchemaRef> {
+        let input_schema = input.schema();
+        let deferred_set: HashSet<&str> = deferred_columns.iter().map(|s| s.as_str()).collect();
+
+        let mut qualified_fields: Vec<(Option<TableReference>, Arc<ArrowField>)> = input_schema
+            .iter()
+            .filter(|(_, f)| !deferred_set.contains(f.name().as_str()))
+            .map(|(q, f)| (q.cloned(), f.clone()))
+            .collect();
```
wjones127, line 175: issue(blocking): I'm rather worried about the prospect of matching on string column names and ignoring the TableReference. I'm worried this could cause problems when are matching column names.

rust/lance/src/io/exec/late_take.rs
wjones127: Done in c2d76d0bc — stripped the merge_insert/issue-specific context (the `#4583` cardinality note, the backfill framing, the "non-source columns for every target row" wording) from the module doc and `is_wide_column` so the rule reads as generic.

rust/lance/src/io/exec/late_take.rs
wjones127: Good catch — fixed in c2d76d0bc. The node identified deferred columns by name alone, so a same-named column from another relation (e.g. `source.payload` when `target.payload` is deferred) would have been wrongly dropped from the passthrough. Both `build_output_schema` and `necessary_children_exprs` now match on `(qualifier, name)` against the deferred columns' relation (`self.qualifier`). Added `test_deferred_match_respects_qualifier` to lock it in.

wjones127:
## Summary

This PR added a logical optimizer rule (`late_take.rs`) that defers reading wide, non-source columns in merge_insert's join until after matching, using a width-only gate (assuming a fixed 0.1% match rate, borrowed from the scanner's filter heuristic). We looked at how to make that gate cardinality-aware instead.

Building a cost model (`Cost(take) = M·σ·L` vs `Cost(scan) = N·w/B`) surfaced that the hash-join shuffle (`σ`) — not just selectivity — is the dominant, previously-unaccounted-for cost, since `TakeExec` receives row addresses in scrambled order.

That led to a better path: DataFusion's dynamic filter pushdown for hash joins can eliminate the problem at the source by skipping wide-column decode during the scan itself, with no shuffle penalty. Lance's `FilteredReadExec` doesn't yet implement the pushdown side (filed #7562). Tracing DataFusion's `on_lr_is_preserved`/join-swap logic shows that once that lands, plus the source-stats PR (#7368) and a DataFusion bump to 54.x, dynamic filtering will cover merge_insert's two dominant shapes: pure backfill (`Inner`) and upsert (`Right`→`Left` after the build-side swap).

It fundamentally cannot cover delete-tracking joins (`Left`/`Full`, target preserved) — dropping unmatched probe rows there would incorrectly discard legitimate delete candidates. For that case, the fix isn't a smarter width gate but extending `LateTakeNode`'s existing `necessary_children_exprs` declaration to also require the opposite side's join key (a generic, sentinel-free match indicator), so `TakeExec` skips fetching unmatched rows there too.

Since the upstream work (#7562, #7368, the DataFusion bump) covers the common cases better than this PR's approach, closing this in favor of that, plus a smaller follow-up for the delete-tracking fix.

PR #7368, opened 2026-06-18
feat: accept TableProvider write inputs for merge_insert and insert

wjones127:
Implements #4583: make `Arc<dyn TableProvider>` the canonical internal write input behind ergonomic wrappers, so re-readable sources replay across retries without spilling and materialized sources expose statistics to the merge join.

## Changes

- **merge_insert**: `execute_provider` (canonical entry) and `execute_batches` (multi-partition `MemTable`); `execute(stream)` is now a thin wrapper. The retry loop re-scans the provider per attempt, replacing the `new_source_iter`/`SpillStreamIter` replay layer. A one-shot stream spills (memory→disk) only when `conflict_retries > 0`; `spill_for_retry(false)` fails fast on contention instead of buffering the stream.
- **Source statistics (use case 3)**: `execute_provider`/`execute_batches` plan against the provider via `read_table`, so a `MemTable`/file source's exact `num_rows`/`total_byte_size` reach the optimizer and drive join build-side selection. Stream sources keep the one-shot path — they carry no statistics anyway, and this preserves the source's original error type.
- **InsertBuilder**: adds `execute_provider`/`execute_uncommitted_provider` for input-shape parity (no retry/spill today).
- **Python**: materialized inputs (`pa.Table`, `pa.RecordBatch`, pandas/polars frames, dict/list-of-dict) route through the new in-memory `execute_batches` path; streams, readers, scanners, and datasets keep the spilling path.

## Tests

- Rust: new tests for `execute_provider`, `execute_batches`, source-stats-in-plan, `spill_for_retry(false)` fail-fast, and `InsertBuilder::execute_provider`; full `dataset::write::` suite + spill tests green; clippy/fmt clean.
- Python: new materialized-vs-streaming parity test; merge_insert suite green; ruff clean.

## Follow-ups (out of scope here)

- Re-scannable Python providers for `pa.dataset.Dataset`/`Scanner` (currently still spill — needs a Python-callback `TableProvider`).
- Fanning the fragment writer out over provider partitions so materialized inserts write data files in parallel (use case 1).

## Summary by CodeRabbit

* **New Features**
  * Merge-insert now supports materialized sources and batch-based execution, producing consistent results across fully materialized tables and streaming readers.
  * Added uncommitted batch execution paths for merge-insert workflows.
  * Added a control to enable/disable spilling input data for retry handling.

* **Bug Fixes**
  * Improved merge-insert retry behavior for one-shot streaming inputs to prevent incorrect retry outcomes.
  * Enhanced error handling for shared execution failures to preserve more specific error details.


rust/lance/src/dataset/write/merge_insert.rs
```diff
@@ -1455,14 +1615,22 @@ impl MergeInsertJob {
...
         //       indexed vs non-indexed cases. That should be handled by optimizer rules.
-        let session_config = SessionConfig::default();
+        let session_config = if scan_provider_directly {
+            // A provider may expose multiple partitions; keep the plan single-
+            // partition so it satisfies the merge write node's contract (the
+            // provider's statistics still drive join-side selection).
+            SessionConfig::default().with_target_partitions(1)
+        } else {
+            SessionConfig::default()
+        };
```
wjones127: issue(blocking): This seems unnecessary. If we label the `ExecutionPlan::required_input_distribution` to be `Distribution::SinglePartition`, then the optimizer should take care of this for us, right? Then we shouldn't need the `scan_provider_directly` parameter.

rust/lance/src/dataset/write/merge_insert.rs
```diff
@@ -1473,7 +1641,15 @@ impl MergeInsertJob {
...
-        let source_df = session_ctx.read_one_shot(source)?;
+        // Plan against the provider directly so its statistics reach the optimizer;
+        // otherwise adapt it to a one-shot stream (which carries no statistics but
+        // preserves the source's original error type).
+        let source_df = if scan_provider_directly {
+            session_ctx.read_table(provider)?
+        } else {
+            let source = provider_to_stream(provider).await?;
+            session_ctx.read_one_shot(source)?
+        };
```
wjones127: issue(blocking): this seems like unnecessary complexity again. Why isn't the error preserved? Can we adjust the error conversion traits so it is preserved?

rust/lance/src/dataset/write/merge_insert.rs
```diff
@@ -10058,4 +10274,306 @@ MergeInsert: on=[id], when_matched=DoNothing, when_not_matched=InsertAll, when_n
...
+    async fn test_merge_insert_execute_batches() {
+        let schema = id_value_schema();
+        let initial = RecordBatch::try_new(
+            schema.clone(),
+            vec![
+                Arc::new(UInt32Array::from(vec![0, 1, 2])),
+                Arc::new(UInt32Array::from(vec![0, 0, 0])),
+            ],
+        )
+        .unwrap();
```
wjones127: suggestion: when writing small test data like this, instead prefer the `record_batch!()` macro. It's both shorter and more explicit.


```suggestion
        let initial = record_batch!(
            ("id", UInt32, [0, 1, 2]),
            ("value", UInt32, [0, 0, 0])
        ).unwrap();
```

python/python/lance/types.py
```diff
@@ -52,6 +52,34 @@ def _casting_recordbatch_iter(
         yield batch
 
 
+def _is_materialized(data_obj: ReaderLike) -> bool:
```
hamersaw, line 55: Maintaining this list seems a little difficult. IIUC we support all `ArrowArrayStreamReader` parsable types (presumably huge?). And this function is attempting to decide which are oneshot and need to be spilled to disk. It may not be a huge concern because worst case scenario we just double store on ingest, but notably we are missing things like Scanner (Lances), `pa.dataset.Dataset`, or LanceDataset which can be reread without spilling.

rust/lance-datafusion/src/spill.rs
```diff
@@ -60,6 +64,101 @@ pub fn create_replay_spill(
...
+    mut source: SendableRecordBatchStream,
+    memory_limit: usize,
+) -> Result<Arc<dyn TableProvider>, DataFusionError> {
+    let schema = source.schema();
+    let tmp_dir = tokio::task::spawn_blocking(TempDir::try_new)
+        .await
+        .map_err(|e| DataFusionError::Execution(format!("Failed to spawn temp dir task: {e}")))?
+        .map_err(|e| DataFusionError::Execution(format!("Failed to create temp dir: {e}")))?;
+    let tmp_path = tmp_dir.std_path().join("spill.arrows");
+    let (mut sender, receiver) = create_replay_spill(tmp_path, schema.clone(), memory_limit);
```
hamersaw, line 127: So we create the persistent spill recovery (either complete in-memory or + spill file) before beginning any of the insert / merge_insert logic. I don't think it's useful to explore right now, but it feels like some level of parallelization could really help performance here.

rust/lance/src/dataset/write/merge_insert.rs
```diff
@@ -340,6 +343,12 @@ struct MergeInsertParams {
     // Controls whether data that is not matched by the source is deleted or not
     delete_not_matched_by_source: WhenNotMatchedBySource,
     conflict_retries: u32,
+    // When the source is a one-shot stream and `conflict_retries > 0`, the source
+    // is spilled (memory, then disk) so it can be replayed on each retry. Set to
+    // false to fail fast on contention instead of buffering the stream. Has no
+    // effect on re-scannable sources (materialized batches, files), which never
+    // spill.
+    spill_for_retry: bool,
```
hamersaw, line 351: It looks like the previous `new_source_iter` had an `enable_retries` function that was set if retries were configured. So users didn't have to manually set this, instead it was auto-inferred. I don't have great context, but would it be possible to use the same "enable on retries" and, similar to the python API, chose not to spill internally when we know the source is not oneshot? This could take the onus off users to define spill vs not.

hamersaw (approved):
Looks great! Just a few clarification questions more so for my understanding than anything.

python/python/lance/types.py
wjones127: Good point on the drift. The classifier answers "can I cheaply wrap this in an in-memory `MemTable`?" — a `Scanner`/`LanceDataset` is re-readable but *not* in-memory, so materializing it just to skip a spill could be worse than the spill. But you're right that re-scanning it (no spill, no double-store) is strictly better than what we do today. Filed #7895 to make source classification a `singledispatch` registry and add a re-scannable provider for exactly these types.

rust/lance/src/dataset/write/merge_insert.rs
wjones127: It's already automatic in that direction — spilling only kicks in when `conflict_retries > 0`, and re-scannable providers (materialized batches, files) never spill regardless. `spill_for_retry` is only an opt-*out* (default `true`) to fail-fast instead of buffering a genuine one-shot stream. The gap is that a plain `SendableRecordBatchStream` is opaque, so we can't tell re-scannable from one-shot at that boundary — which is what #7895 addresses on the Python side.

rust/lance-datafusion/src/spill.rs
wjones127: Worth noting the spill is already streaming: a reader can start pulling batches as soon as they land (it doesn't block on the full drain), and it stays in memory entirely unless it exceeds the 100 MB budget. So the "create spill before insert logic" step isn't a serial full-buffer barrier. Agreed there's more parallelism to chase later, though. (I've pulled these two properties up into the `spilling_table_provider` doc so they're visible at the call site.)

PR #7369, opened 2026-06-19
feat: add local SpillStore for index scratch

RitwijParmar:
## Summary

This extends the local spill store into the scalar index build path so temporary index files can use session-scoped, quota-aware scratch storage instead of ad hoc unbounded temp directories.

Changes include:
- expose the session spill store as a scratch-backed `LanceIndexStore`
- add a scratch-aware scalar plugin training hook with default fallback behavior
- wire ngram and rtree training scratch through the session spill store
- propagate scratch through JSON indexes when they delegate to another scalar plugin
- release quota on local delete and avoid copy-then-delete quota spikes during local rename
- clean up the final ngram spill file after copying it into the destination index
- add focused quota tests for local delete/rename and ngram scratch cleanup

This makes the spill primitive useful for real index builders while keeping the plugin API backward-compatible for builders that do not need scratch storage.

## Validation

- `rustup run nightly-2026-03-21 cargo fmt --all`
- `CARGO_PROFILE_DEV_DEBUG=0 CARGO_BUILD_JOBS=1 rustup run nightly-2026-03-21 cargo check -p lance-io --no-default-features`

I also tried `cargo check -p lance-index --no-default-features`; it reached `lance-encoding` and stopped because this machine does not have `protoc` installed.

Refs #7300

### DIFF
--- a/rust/lance-core/src/error.rs
+++ b/rust/lance-core/src/error.rs
@@ -169,6 +169,16 @@ pub enum Error {
         #[snafu(implicit)]
         location: Location,
     },
+    #[snafu(display(
+        "Disk spill quota exceeded: cap={cap_bytes} bytes, used={used_bytes} bytes, requested={requested_bytes} bytes, {location}"
+    ))]
+    DiskCapExceeded {
+        cap_bytes: u64,
+        used_bytes: u64,
+        requested_bytes: u64,
+        #[snafu(implicit)]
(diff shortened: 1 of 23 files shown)


RitwijParmar:
Closing this PR because it has been inactive for several months and is no longer merge-ready. Thanks for the consideration.

PR #7371, opened 2026-06-19
fix(index): drop stale per-segment rows in vector search after in-place column update

### LINKED ISSUE #7370: Vector search returns an updated row twice after update_columns + optimize_indices
## Summary

After overwriting an indexed vector column in place via the low-level `LanceFragment.update_columns` API and then delta-optimizing the index with `optimize_indices(num_indices_to_merge=0)`, a KNN vector query returns the updated row twice: once with the stale (pre-update) indexed vector and once with the new value.

## Reproduction

```python
import lance, numpy as np, pyarrow as pa, tempfile, os

ndim = 16
np.random.seed(42)
uri = os.path.join(tempfile.mkdtemp(), "ds")

def fsl(arr):
    return pa.FixedSizeListArray.from_arrays(pa.array(arr.reshape(-1), pa.float32()), list_size=ndim)

bulk = np.random.uniform(-1, 1, (1000, ndim)).astype(np.float32)
lance.write_dataset(pa.table({"id": pa.array(range(1000)), "vector": fsl(bulk)}), uri)
ds = lance.write_dataset(
    pa.table({"id": pa.array([10000]), "vector": fsl(np.full((1, ndim), 2.0, np.float32))}),
    uri, mode="append")

ds = ds.create_index("vector", index_type="IVF_PQ", num_partitions=1, num_sub_vectors=ndim)

frag = ds.get_fragment(1)
rowids = frag.to_table(columns=["id"], with_row_id=True)["_rowid"].to_pylist()
upd = pa.table({"_rowid": pa.array(rowids, pa.uint64()),
                "vector": pa.array([[10.8] * ndim] * len(rowids), pa.list_(pa.float32(), ndim))})
uf, fm = frag.update_columns(upd)
ds = lance.LanceDataset.commit(ds.uri, lance.LanceOperation.Update(updated_fragments=[uf], fields_modified=fm),
                               read_version=ds.version)

ds.optimize.optimize_indices(num_indices_to_merge=0)
ds = lance.dataset(ds.uri)

res = ds.to_table(columns=["id"], nearest={"column": "vector", "q": np.full(ndim, 10.8, np.float32), "k": 10},
                  with_row_id=True).to_pandas()
print(int((res["id"] == 10000).sum()))   # prints 2; expected 1
```

## Root cause

`update_columns` rewrites only the column data file, keeping the fragment id and row address. Committing the `Update` prunes the updated fragment from the existing index segment's `fragment_bitmap`, but that segment's index file still physically holds the row's old vector. `optimize_indices(num_indices_to_merge=0)` then appends a new delta segment for the updated fragment (with the new vector) without rewriting the old segment.

At query time `DatasetPreFilter` is built once from the union of all segment bitmaps, so it covers every fragment and applies no restriction. Both segments are searched and there is no cross-segment row-id dedup on the vector path, so the old segment returns the row with its stale vector and the new delta segment returns it with the new value: a duplicate. (The same stale segment also misranks the row for queries near its old vector.)

## Expected

The updated row appears exactly once, served by the segment that owns the fragment.

I'll open a PR with the fix.

wombatu-kun:
Closes #7370

## Problem
After an in-place column update (`LanceFragment.update_columns` + `LanceOperation.Update`) followed by `optimize_indices(num_indices_to_merge=0)`, a KNN query returns the updated row twice: once with the stale pre-update vector from the original index segment, once with the new value from the delta segment.

## Root cause
An in-place update keeps the fragment id and row address, and committing the `Update` prunes that fragment from the old segment's `fragment_bitmap`, but the old segment's index file still physically holds the row's old vector. The shared `DatasetPreFilter` is built from the union of all segment bitmaps, so it cannot express "fragment N is valid for the delta segment but stale for the old one", and there is no cross-segment row-id dedup on the vector path.

## Fix
Restrict each segment's search output to the fragments it still owns, using a mask from the existing `DatasetPreFilter::create_restricted_deletion_mask` so it is correct for both row-address and stable-row-id datasets.

The mask is applied wherever a partition's rows are produced or counted, before the shared early/late search coordinator sees them: `search_partition`, `instrument_sequential_partition_stream`, and `LatePartitionSearchControl::record_batch`, which the index calls on the CPU pool. Applying it later, as one post-filter on the combined stream, lets a stale row consume the `k` budget on its own, so the segment holding the fresh copy is never probed and the query returns fewer than `k` current rows.

The `late_search` "fewer than k results" shortcut previously emitted the whole prefilter set from the first delta. With a restriction active each delta emits only the addresses its own segment owns, so each is emitted exactly once; without one the original first-delta-only path is unchanged. Both are gated on every segment having a `fragment_bitmap` and are no-ops for normal append/merge indices.

## Tests
`test_no_stale_duplicate_after_partial_column_update` (Python) reproduces the duplicate end to end and asserts the row is returned once, the top-k is still full, and no id repeats. With a single IVF partition the late search returns before the budget is consulted, so it pins the filtering only.

`test_unowned_row_does_not_fill_the_shared_budget` (Rust) pins the accounting, parameterized over the sequential and parallel paths because they count rows in different places, with partitions that mix owned and unowned rows so the mask must keep part of a batch. Verified by mutation: it fails if any of the three application points is reverted, if the filter becomes an all-or-nothing drop, or if the row count becomes all-or-nothing. `test_segment_owning_nothing_skips_the_late_search` covers the skip below.

## Limits
- The mask is built per delta per query instead of once for the union, which is O(dataset fragments) and serialized ahead of that delta's search. Hoisting the shared `frag_map` out of `create_deletion_mask_impl` would remove it; left as a follow-up since it is not a correctness issue.
- A stale row no longer moves the shared counter, so a segment holding many stale rows probes more partitions than before. A segment that owns nothing is now skipped outright in `late_search`; the rest is bounded by sibling deltas filling the counter.
- The restriction is a post-filter on the index's already-truncated per-partition top-k, so a stale row can still displace an owned row inside a partition. Removing that means pushing the restriction into the sub-index search.

## Related
The merge path has an independent instance of the same invariant: `optimize_indices(num_indices_to_merge >= 1)` copies the old segment's rows into the merged segment unfiltered, so both copies end up in one segment that legitimately owns their fragment and no read-time mask can separate them. It reproduces on `main` without this PR and is fixed in #8342.


Xuanwo (commented):
1. Segment restrictions are applied only after early and late vector-search accounting has already counted stale rows. Stale candidates can consume the shared search budget and leave fewer than `k` valid results.
2. The commit metadata includes a disallowed co-author trailer. This violates the repository commit policy.

wombatu-kun:
1. This is the narrow limitation noted in the PR description, and even in this form the change is better merged than withheld: the amount of useful information returned does not change. The set of unique, current vectors is the same as before; what the per-segment restriction removes are the stale duplicate copies (the outdated rows the old segment should no longer serve). The single downside is a rare, bounded recall dip, limited to a selective prefilter coinciding with an in-place-update stale row whose fresh copy sits in an unsearched partition, and it never produces a duplicate or a wrong row. The restriction is a post-filter, so a stale candidate can occupy a k-slot in the early/late accounting and then be dropped; eliminating even that means moving the restriction into the shared cross-delta prefilter and its shortcut accounting, a substantially larger change. I can open a follow-up issue to track it if you would prefer.
2. Done ee0c7be4.

rust/lance/src/io/exec/knn.rs
```diff
@@ -1621,8 +1669,34 @@ impl ExecutionPlan for ANNIvfSubIndexExec {
...
                             state,
                             target_partitions,
+                            seg_mask.clone(),
                         );
-                        DataFusionResult::Ok(early_search.chain(late_search).boxed())
+                        let combined = early_search.chain(late_search);
+                        // Drop stale rows from this segment's search output (rows whose
+                        // fragment the segment no longer owns). The shortcut path in
+                        // late_search restricts its emitted rows with the same mask.
+                        let restricted = combined.map(move |batch_res| {
```
Xuanwo: This path applies the segment restriction only after the shared early/late search accounting has already counted the unmasked results. A stale hit can satisfy the coordinator and then be dropped, so the query can return fewer than k valid rows or miss the fresh row from the owning segment.

rust/lance/src/io/exec/knn.rs
```diff
@@ -2018,8 +2066,34 @@ impl ExecutionPlan for ANNIvfSubIndexExec {
...
```
Xuanwo: This mask cannot repair stale rows that have been merged into the same segment. With `num_indices_to_merge=1`, the rebuild path copies every physical batch from the old segment without applying that segment's fragment bitmap, concatenates the fresh rows, and unions the ownership bitmaps. A stale and fresh copy of row B can therefore land in one new segment that owns B; both copies pass this filter, so explicit index optimization can still preserve duplicate or stale data.

wombatu-kun: Confirmed. The two places are `take_partition_batches` (`rust/lance/src/index/vector/builder.rs:1024`) and the bitmap union in `merge_indices_with_unindexed_frags` (`rust/lance/src/index/append.rs:744-753`), and it reproduces on main independently of this PR: the vector build path has no equivalent of the scalar `select_segments_to_merge` filtering added in #7359.

The fix belongs on the build side - restrict each merged segment's rows to its effective fragment bitmap, plumbing per-segment masks from `LogicalIvfView::segments` through `optimize_vector_indices` into `take_partition_batches`, with a `num_indices_to_merge=1` regression test.

Do you want that in this PR, or as a follow-up so this one stays on the query path?

rust/lance/src/io/exec/knn.rs
wombatu-kun: Done bc8ab3dd8. The mask now runs where each partition's batch is produced or counted, so the shared counters and the emitted rows are the same set.

rust/lance/src/io/exec/knn.rs
```diff
@@ -1667,20 +1669,29 @@ impl ANNIvfSubIndexExec {
...
                         let initial_addrs = state.initial_ids.lock().unwrap();
                         let found_addrs = HashSet::<_>::from_iter(initial_addrs.iter().copied());
                         drop(initial_addrs);
-                        let mask_addrs = HashSet::from_iter(iter_addrs.map(u64::from));
-                        let not_found_addrs = mask_addrs.difference(&found_addrs);
-                        let not_found_addrs =
-                            UInt64Array::from_iter_values(not_found_addrs.copied());
+                        let not_found_addrs = UInt64Array::from_iter_values(
+                            iter_addrs.map(u64::from).filter(|addr| {
+                                !found_addrs.contains(addr)
```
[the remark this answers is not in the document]
wombatu-kun: Done bc8ab3dd8. The counters now see only rows that pass the mask, but the mask is applied where each partition's batch is produced rather than inside the segment prefilter: `create_restricted_deletion_mask` returns a non-empty mask for every segment of a multi-delta index, so folding it in would disable the no-prefilter fast path in the sub-index searches and turn an O(k) post-filter into an O(partition) masked search.

rust/lance/src/io/exec/knn.rs
wombatu-kun: Fixed in #8342, which filters each old segment's rows against its own effective coverage before the merge concatenates them. It reproduces on `main` without this PR, so it is a separate build-path fix.

rust/lance/src/io/exec/knn.rs
```diff
@@ -2027,8 +2074,34 @@ impl ExecutionPlan for ANNIvfSubIndexExec {
...
+                            };
+                            if batch.num_rows() == 0 {
+                                return Ok(batch);
+                            }
+                            let row_ids = batch[ROW_ID].as_primitive::<UInt64Type>();
+                            let keep = BooleanArray::from_iter(
+                                row_ids
+                                    .values()
+                                    .iter()
+                                    .map(|&id| Some(seg_mask.selected(id))),
```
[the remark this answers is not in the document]
wombatu-kun: Fixed in #8342.

rust/lance/src/io/exec/knn.rs
```diff
@@ -1591,11 +1642,13 @@ impl ANNIvfSubIndexExec {
         part_id: usize,
         pre_filter: Arc<DatasetPreFilter>,
         metrics: Arc<AnnIndexMetrics>,
+        seg_mask: Option<Arc<RowAddrMask>>,
     ) -> DataFusionResult<RecordBatch> {
         let batch = index
             .search_in_partition(part_id, &query, pre_filter, &metrics.index_metrics)
             .map_err(|e| DataFusionError::Execution(format!("Failed to calculate KNN: {}", e)))
             .await?;
+        let batch = restrict_to_segment(batch, seg_mask.as_deref())?;
```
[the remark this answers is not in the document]
wombatu-kun: Tracked in #8348. This reproduces identically on `main` at 6fe450bb7, so it is not introduced here:

```
main       OBSERVED_IDS=[100, 101, 102, 103, 104]  OBSERVED_DISTS=[0.0, 0.0, 0.0, 0.0, 0.0]
bc8ab3dd8  OBSERVED_IDS=[100, 101, 102, 103, 104]  OBSERVED_DISTS=[400.0, 400.0, 400.0, 400.0, 400.0]
```

Same ids either way. `main` ranks them by the stale pre-update vector and reports distance 0; this PR reports their current distance. The ids below 20 are missing in both, which is the limit already stated in the description: the old segment's top-k is consumed inside the sub-index by rows it no longer owns, before any post-filter can run.

Composing the mask into the sub-index prefilter does fix that, but `create_restricted_deletion_mask` returns `Some` for every segment of a multi-delta index, so it makes `DatasetPreFilter::is_empty()` false everywhere and disables the no-prefilter fast path in `pq.rs:282`, `flat/index.rs:151,272` and `hnsw/builder.rs:1356`, turning an O(k) post-filter into an O(partition) masked search on every delta query. That needs a benchmark behind it rather than a fold-in here.

Xuanwo (approved):

PR #7372, opened 2026-06-19
perf(index): add exact null-row bitmap to zone map and bloom filter

westonpace:
The queries `X IS NULL` and `X IS NOT NULL` are very common queries.  Because validity is scattered throughout a file it is not actually that easy to answer this query without scanning the column.  This means very wide columns can be very slow.  Very wide columns also tend to be columns that are not good candidates for btree (too much duplicated data) or bitmap (too much cardinality).

On the flip side, a validity bitmap is pretty small, and maintaining one for each column is probably affordable, even for the lightweight zone map and bloom filter indexes.  It would be nice to have consistent `IS NULL` speedup when a column is indexed, regardless of the index type.

---

Captures a `RowAddrTreeMap` of every null row address during index training.  IS NULL queries can now be answered in O(1) by returning `SearchResult::exact(null_rows)` instead of scanning all zones, which also eliminates the downstream recheck step.

Backward-compatible: older indexes without the `null_bitmap.lance` file load with `null_rows = None` and fall back to the previous zone-scan path.

### DIFF
--- a/rust/lance-index/src/scalar/bloomfilter.rs
+++ b/rust/lance-index/src/scalar/bloomfilter.rs
@@ -17,13 +17,14 @@ use crate::scalar::{
 use crate::{Any, pb};
 use arrow_array::{Array, UInt64Array};
 use arrow_schema::{DataType, Field};
-use lance_select::RowAddrTreeMap;
 use lance_arrow_stats::StatisticsAccumulator;
 use lance_core::utils::bloomfilter::as_bytes;
 use lance_core::utils::bloomfilter::sbbf::{Sbbf, SbbfBuilder};
 use lance_core::utils::row_addr_remap::RowAddrRemap;
+use lance_select::RowAddrTreeMap;
 use serde::{Deserialize, Serialize};
 
+use std::collections::HashMap;
 use std::sync::LazyLock;
 
 use datafusion::execution::SendableRecordBatchStream;
@@ -432,11 +433,12 @@ impl ScalarIndex for BloomFilterIndex {
         metrics: &dyn MetricsCollector,
     ) -> Result<SearchResult> {
         let query = query.as_any().downcast_ref::<BloomFilterQuery>().unwrap();
-        if let BloomFilterQuery::IsNull() = query {
-            if let Some(null_rows) = &self.null_rows {
-                return Ok(SearchResult::exact(null_rows.clone()));
-            }
+        if let BloomFilterQuery::IsNull() = query
+            && let Some(null_rows) = &self.null_rows
+        {
+            return Ok(SearchResult::exact(null_rows.clone()));
         }
+
         search_zones(&self.zones, metrics, |block| {
             self.evaluate_block_against_query(block, query)
         })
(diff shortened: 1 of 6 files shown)


rust/lance-index/src/scalar/zonemap.rs
```diff
@@ -617,19 +643,24 @@ impl ScalarIndex for ZoneMapIndex {
         let options = ZoneMapIndexBuilderParams::new(self.rows_per_zone);
         let processor = ZoneMapProcessor::new(value_type.clone())?;
         let trainer = ZoneTrainer::new(processor, self.rows_per_zone)?;
-        let updated_zones = rebuild_zones(&self.zones, trainer, new_data).await?;
+        let (updated_zones, new_null_rows) = rebuild_zones(&self.zones, trainer, new_data).await?;
+
+        // Merge existing and new null rows
+        let mut merged_null_rows = self.null_rows.clone().unwrap_or_default();
```
Xuanwo: `None` here means the loaded zonemap was created before exact null bitmaps existed, so the old null rows are unknown, not empty. Treating it as `RowAddrTreeMap::default()` and then writing `lance:null_bitmap` upgrades the updated index to an exact bitmap that only contains newly appended nulls, so `IS NULL` can silently miss nulls from the legacy part of the index.

rust/lance-index/src/scalar/bloomfilter.rs
```diff
@@ -446,18 +469,23 @@ impl ScalarIndex for BloomFilterIndex {
 
         let processor = BloomFilterProcessor::new(params.clone())?;
         let trainer = ZoneTrainer::new(processor, params.number_of_items)?;
-        let updated_blocks = rebuild_zones(&self.zones, trainer, new_data).await?;
+        let (updated_blocks, new_null_rows) = rebuild_zones(&self.zones, trainer, new_data).await?;
+
+        // Merge existing and new null rows
+        let mut merged_null_rows = self.null_rows.clone().unwrap_or_default();
```
Xuanwo: Same issue here: a legacy bloom filter with `null_rows = None` is being treated as if it had an exact empty null bitmap. After `write_index` persists the metadata, `IS NULL` will return `Exact(new_null_rows)` and can drop null rows that were already present in the old index.

rust/lance-index/src/scalar/zonemap.rs
```diff
@@ -681,18 +714,27 @@ pub async fn merge_zonemap_indices(
                 })
                 .cloned(),
         );
+        if let Some(null_rows) = &source.null_rows {
```
Xuanwo: This merge only preserves null rows from source segments that already have `null_rows`. If any merged segment is legacy, its null rows are unknown, but the merged output still gets written with exact null bitmap metadata. That can make `IS NULL` return an incomplete exact result after segment merge.

Xuanwo (commented):
The main concern is that `None` currently represents an unknown legacy null bitmap, but update and merge paths turn it into an exact empty or partial bitmap. That can create false negatives for `IS NULL`.

I think this needs a known/unknown state for null rows: only write `lance:null_bitmap` when the output bitmap covers all retained rows, or otherwise keep the legacy fallback / rebuild from data. It would also be good to add regression tests for legacy update, mixed/all-legacy zonemap merge, and non-empty exact `IS NULL` results.

One additional correctness issue is that `old_data_filter` is part of the `ScalarIndex::update` contract, but zonemap and bloomfilter ignore it while cloning the existing exact bitmap. Once null rows become exact, existing bitmap rows need to be filtered before unioning new null rows; otherwise deleted or rewritten old rows can remain in an `Exact` `IS NULL` result.

This also adds a new persistent format contract (`lance:null_bitmap` pointing at a global buffer), but the scalar index format docs still describe only the old metadata and still list `IS NULL` as `AtMost`. Please document that the bitmap is optional, absence means legacy/unknown fallback, and `IS NULL` is exact only when the bitmap is present and complete.

rust/lance-index/src/scalar/zonemap.rs
westonpace: Good catch.  I've changed the behavior.  Now, on update, if the old index was missing nulls then the updated index will also be missing nulls.

rust/lance-index/src/scalar/bloomfilter.rs
westonpace: Same fix

rust/lance-index/src/scalar/zonemap.rs
westonpace: I made a similar fix here.  If the any of the merged segments are missing nulls then the merged index will be missing nulls.

westonpace:
> I think this needs a known/unknown state for null rows: only write lance:null_bitmap when the output bitmap covers all retained rows, or otherwise keep the legacy fallback / rebuild from data. It would also be good to add regression tests for legacy update, mixed/all-legacy zonemap merge, and non-empty exact IS NULL results.

The `None` should have already meant "we don't know what the nulls are" and so I didn't introduce a new state.  I've added tests for the cases you described.

Xuanwo (approved):
Nice work, thank you!

PR #7373, opened 2026-06-19
fix: reject DataReplacement racing concurrent Update/Delete/Merge

wkalt:
A DataReplacement rewrites a column's data file positionally against the fragments it targets. The conflict resolver returned Ok unconditionally for a concurrent Update, Delete, or Merge, so a DataReplacement committed at a read version those operations had superseded was applied silently -- dropping or misaligning the rows the concurrent op moved or deleted, with no error raised. Merge was additionally asymmetric: check_merge_txn already treats a concurrent DataReplacement as a conflict, but not the reverse.

For Update/Delete, conflict when the other transaction's updated/removed fragment ids overlap our replacement fragment ids (mirrors the existing Rewrite handling). For Merge, which rewrites the entire fragment list, conflict unconditionally (mirrors check_merge_txn). All are retryable, so the committer rebuilds against the new layout.

Adds DataReplacement vs Update, Delete, and Merge cases (same and different fragment) to test_conflicts_data_replacement.

### DIFF
--- a/rust/lance/src/io/commit/conflict_resolver.rs
+++ b/rust/lance/src/io/commit/conflict_resolver.rs
@@ -904,13 +904,42 @@ impl<'a> TransactionRebase<'a> {
             match &other_transaction.operation {
                 Operation::Append { .. }
                 | Operation::Clone { .. }
-                | Operation::Delete { .. }
-                | Operation::Update { .. }
-                | Operation::Merge { .. }
                 | Operation::UpdateConfig { .. }
                 | Operation::ReserveFragments { .. }
                 | Operation::Project { .. }
                 | Operation::UpdateBases { .. } => Ok(()),
+                Operation::Merge { .. } => {
+                    // Merge rewrites the whole fragment list; always conflict
+                    // (symmetric with check_merge_txn).
+                    Err(self.retryable_conflict_err(other_transaction, other_version))
+                }
+                Operation::Update {
+                    updated_fragments,
+                    removed_fragment_ids,
+                    ..
+                }
+                | Operation::Delete {
+                    updated_fragments,
+                    deleted_fragment_ids: removed_fragment_ids,
+                    ..
+                } => {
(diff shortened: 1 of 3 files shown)


rust/lance/src/io/commit/conflict_resolver.rs
```diff
@@ -904,13 +904,42 @@ impl<'a> TransactionRebase<'a> {
...
+                            .iter()
+                            .map(|f| f.id)
+                            .chain(removed_fragment_ids.iter().copied())
+                            .any(|id| id == replacement.0);
+                        if touches_our_fragment {
+                            return Err(
+                                self.retryable_conflict_err(other_transaction, other_version)
+                            );
+                        }
+                    }
```
westonpace, line 940: Do we need to conflict if a delete removed a fragment that we updated?  Should this be asymmetric?

DataReplacement followed by Delete -> Disallowed?
Delete followed by DataReplacement -> Allowed (but changes discarded silently)

This is a genuine question, I don't really know the answer.

westonpace (commented):
Can you also update `/home/pace/dev/lance/main/docs/src/format/table/transaction.md`?

rust/lance/src/io/commit/conflict_resolver.rs
wkalt: for the specific case of Delete followed by DataReplacement, will throw an error here if you hit it today:
https://github.com/lance-format/lance/blob/main/rust/lance/src/dataset/transaction.rs#L2229,L2231

I assume we could probably work around this in the DataReplacement operation but handling it in the conflict mechanism seems like a more certain approach to me.

For the other direction we definitely need to conflict or the wrong rows may get deleted.

westonpace (approved):

westonpace:
> This PR touches the Lance format specification.

I think this gap in `docs/src/format` was a gap and does not represent a behavior change (as mentioned, we previously conflicted with delete in one direction anyways).

PR #7376, opened 2026-06-19
feat: add RLE v2 run length widths

### LINKED ISSUE #7327: Add opt-in RLE v2 run-length widths
## Summary

Add an opt-in RLE v2 codec path that supports `run_lengths = Flat(8|16|32)` without changing the Lance data file format or RLE protobuf shape.

## Scope

- Decode RLE run lengths from `Flat(8)`, `Flat(16)`, and `Flat(32)`.
- Keep `Flat(8)` as the default compatibility mode.
- Add strict validation for zero lengths, values/lengths entry mismatches, underflow, and overflow.
- Add a table-level reader/writer feature flag for datasets that may contain RLE v2 pages.
- Add an explicit writer opt-in path.

## Acceptance Criteria

- New Lance can write and read opt-in RLE v2 pages.
- Old `Flat(8)` RLE data remains readable.
- Old readers fail closed at the table level when RLE v2 pages may exist.
- Compatibility mode still writes `Flat(8)`.

Xuanwo:
## Summary

Adds RLE v2 run-length widths so newly created datasets can write RLE pages with `u16` or `u32` run lengths instead of splitting every run at 255 values. The capability is recorded as a reader feature flag and is only enabled when a new dataset is created with `WriteParams::enable_rle_v2`; existing unflagged datasets reject attempts to turn it on mid-stream.

Closes #7327.

## Benchmark

Ran on `xuanwo-lance-lazy-metadata-bench` with a #6941-style sorted low-cardinality `asset_id` workload.

| workload | Lance default | Lance RLE2 | reduction |
|---|---:|---:|---:|
| `150M rows / 5k assets / random5 value` | `167.36 MiB` | `164.57 MiB` | `1.67%` |
| `150M rows / 5k assets / by-asset5 value` | `7.62 MiB` | `2.03 MiB` | `73.34%` |

The first row keeps the random low-cardinality value column from the issue-like workload, which dominates total size. The second row isolates the long-run case RLE2 targets.

## Validation

Validated with focused RLE2 tests and full Rust clippy before publishing.

### DIFF
--- a/rust/lance-encoding/src/compression.rs
+++ b/rust/lance-encoding/src/compression.rs
@@ -143,8 +143,6 @@ pub struct DefaultCompressionStrategy {
     params: CompressionParams,
     /// The lance file version for compatibilities.
     version: LanceFileVersion,
-    /// Whether the writer may emit RLE v2 run length widths.
-    enable_rle_v2: bool,
 }
 
 fn try_bss_for_mini_block(
@@ -170,7 +168,7 @@ fn try_bss_for_mini_block(
 fn try_rle_for_mini_block(
     data: &FixedWidthDataBlock,
     params: &CompressionFieldParams,
-    enable_rle_v2: bool,
+    use_rle_v2: bool,
 ) -> Option<Box<dyn MiniBlockCompressor>> {
     let bits = data.bits_per_value;
     if !matches!(bits, 8 | 16 | 32 | 64) {
@@ -196,7 +194,7 @@ fn try_rle_for_mini_block(
 
     let num_values = data.num_values;
     let raw_bytes = (num_values as u128) * (type_size as u128);
-    let (run_length_width, rle_bytes) = if enable_rle_v2 {
+    let (run_length_width, rle_bytes) = if use_rle_v2 {
         select_run_length_width_and_size(&data.data, data.num_values, data.bits_per_value).ok()?
     } else {
         // Estimate the encoded size.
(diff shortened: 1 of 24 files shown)


westonpace (commented):
Why are we doing this as a feature flag and not part of the 2.3 file version?

Is the goal to somehow get something out faster than a 2.3 release?

Or maybe we want to move towards a more incremental release strategy?

I don't think it's wrong, but it is different, and I am curious why the change.

Xuanwo:
> Why are we doing this as a feature flag and not part of the 2.3 file version?

My first thought is that the RLE v2 change is not significant enough to be worth including in v2.3. However, I agree it is better to follow the same pattern. I have updated the code.

rust/lance-encoding/src/compression.rs
```diff
@@ -188,16 +192,22 @@ fn try_rle_for_mini_block(
...
+        //
+        // Compatibility RLE stores (value, u8 run_length) pairs. Long runs are split into
+        // multiple entries of up to 255 values. We don't know the run length distribution here,
+        // so we conservatively account for splitting with an upper bound.
+        let estimated_pairs = (run_count.saturating_add(num_values / 255)).min(num_values);
+        (
+            RunLengthWidth::U8,
+            (estimated_pairs as u128) * ((type_size + 1) as u128),
+        )
+    };
```
BubbleCal, line 210: can we move this block to be after line 220?
so that we don't need to pay for calling `select_run_length_width_and_size` if we don't use RLE

BubbleCal (approved):

rust/lance-encoding/src/compression.rs
Xuanwo: it's the cost we need to pay to get the estimate. To fully address this, we will need a better encoding planner.

PR #7377, opened 2026-06-19
feat(index): support flattened JSON sub-doc indexing

wirybeaver:
# Motivation
Lance JSON inverted indexes should preserve array/object structure so path queries over arrays of objects can distinguish exact positions from wildcards. Tantivy documents [a known JSON-index flaw](https://github.com/quickwit-oss/tantivy/blob/main/doc/src/json.md): because an array document is treated as a bag of terms, `cart.product_type:sneakers AND cart.attributes.color:red` can match a document where `sneakers` and `red` live in different array objects. Lance's current single-token-stream JSON index has the same class of problem.

This PR introduces `FlattenedSubDocs` mode for new JSON indexes, flattening each JSON row into one or more sub-documents so those terms must match inside the same flattened array element before the result is mapped back to the original Lance row id. 

It also keeps existing JSON indexes readable and provides two controls for documents that would produce many sub-docs: `disable_cross_array_unnest=true` trades cross-array query accuracy for lower growth, while `max_sub_docs_per_row` places an explicit per-row ingestion limit.

# Summary
- Add explicit JSON tokenizer modes: `SingleDocument` for existing JSON indexes and `FlattenedSubDocs` for new JSON indexes.
- Flatten JSON arrays into multiple internal inverted-index sub-docs, each mapped back to the original Lance row id through `DocSet`.
- Normalize bracketed JSON query paths into value tokens plus `$idx` constraint tokens, and deduplicate flattened JSON search results by row id using max score.
- Add persisted/index param `disable_cross_array_unnest`, default `false`. The default preserves exact Cartesian-product semantics. When set to `true`, sibling arrays are indexed independently to avoid sub-doc explosion, matching Pinot's `DisableCrossArrayUnnest` memory tradeoff.
- Add optional persisted/index param `max_sub_docs_per_row`. It is unlimited when unset. When set, Lance checks the flattened count before allocating the full Cartesian product.
- Add persisted enum param `max_sub_docs_per_row_exceed_action`, default `fail`. `fail` aborts ingestion and suggests increasing the cap or setting `disable_cross_array_unnest=true`; `skip_row` omits the offending source row and continues.
- Expose all JSON ingestion controls through Rust, Python, and Java inverted-index params.

# Error Handling Implementation
- The recursive flattener checks array accumulation and sibling-array expansion against `max_sub_docs_per_row` before allocating the complete set of flattened sub-docs. No limit check is performed when the option is unset.
- Limit failures are returned to the shared `JsonTokenizer` boundary, so immutable index builds and MemWAL indexing use the same behavior.
- With `max_sub_docs_per_row_exceed_action=fail`, tokenization returns an invalid-input error before any sub-doc from that source row is indexed. The error recommends increasing `max_sub_docs_per_row` or setting `disable_cross_array_unnest=true`.
- With `max_sub_docs_per_row_exceed_action=skip_row`, the tokenizer logs a warning and returns no token streams. Existing writer loops therefore omit the entire source row and continue indexing subsequent rows.

# Example 1 - Used to explain the idea

## Raw Documents
doc-0
```json
{"foo":[{"bar":["x","y"]}]}
```
doc-1
```json
{"foo":[{"bar":["y"]},{"bar":"z"}]}
```

## Flattened Documents
| Flatten ID | Flattened Path | Value|
|------------|----------------| ----------|
| 0 | `foo[0].bar[0]` | `"x"` |
| 1 | `foo[0].bar[1]` | `"y"` |
| 2 | `foo[0].bar[0]` | `"y"` |
| 3 | `foo[1].bar` | `"z"` |

## Token Dictionary
Each token is represented as: `(path, type, value)`

| Path | Type | Value | Flatten Posting |
|------|------|-------|-----------------|
| `foo..bar.` | str | `x` | `0` |
| `foo..bar.` | str | `y` | `1, 2` |
| `foo..bar` | str | `z` | `3` |
| `foo$idx` | num | `0` | `0, 1, 2` |
| `foo$idx` | num | `1` | `3` |
| `foo..bar$idx` | num | `0` | `0, 2` |
| `foo..bar$idx` | num | `1` | `1` |

> **Notes**
>
> - Prior array index `[<idx>]` is compressed to `.` when walking through the path from left to right.
> - `foo$idx` represents the index of the element in the `foo` array.
> - `foo..bar$idx` represents the index inside the `bar` array.

## Search Query A
Query:
```text
foo[0].bar[0] = "y"
```

Lookup tokens:
| Token | Flatten Posting |
|-------|---------|
| `foo..bar.`, `str`, `"y"` | `1, 2` |
| `foo..bar$idx`, `num`, `0` | `0, 2` |
| `foo$idx`, `num`, `0` | `0, 1, 2` |

Flatten Posting Intersection:
```text
(1, 2) intersection (0, 2) intersection (0, 1, 2) = {2}
```
Lookup original document:
```text
Flatten ID 2
    ->
Doc ID 1
```

## Search Query B
Query:
```text
foo[0].bar[*] = "y"
```

Lookup tokens:
| Token | Flatten Posting |
|-------|---------|
| `foo..bar.`, `str`, `"y"` | `1, 2` |
| `foo$idx`, `num`, `0` | `0, 1, 2` |

> **Notes**
> - When there's a `[*]` in the search query, we don't need to generate index constraint token `prefix$idx` to match the specific array index on that position.

Flatten Posting Intersection:
```text
(1, 2) intersection (0, 1, 2) = {1, 2}
```

Lookup original document:
```text
Flatten ID [1, 2]
    ->
Doc ID [0, 1]
```

# Example 2: sibling arrays

```json
{"foo":[{"bar":["x","y"]},{"bar":["a","b"]}],"foo2":["u"]}
{"foo":[{"bar":["y","z"]}],"foo2":["u"]}
```

Expected flattened sub-docs:
```text
row0/sub0: foo$idx=0, foo..bar$idx=0, foo..bar.=x, foo2$idx=0, foo2.=u
row0/sub1: foo$idx=0, foo..bar$idx=1, foo..bar.=y, foo2$idx=0, foo2.=u
row0/sub2: foo$idx=1, foo..bar$idx=0, foo..bar.=a, foo2$idx=0, foo2.=u
row0/sub3: foo$idx=1, foo..bar$idx=1, foo..bar.=b, foo2$idx=0, foo2.=u
row1/sub0: foo$idx=0, foo..bar$idx=0, foo..bar.=y, foo2$idx=0, foo2.=u
row1/sub1: foo$idx=0, foo..bar$idx=1, foo..bar.=z, foo2$idx=0, foo2.=u
```

Query behavior:
```text
foo[0].bar[0],str,y -> row1 only
foo[0].bar[*],str,y -> row0 and row1
```

# Example 3: `disable_cross_array_unnest`

Input:
```json
{"a":["x","y"],"b":["u","v"],"c":1}
```

Default `disable_cross_array_unnest=false` produces exact Cartesian-product sub-docs:
```text
sub0: a$idx=0, a.=x, b$idx=0, b.=u, c=1
sub1: a$idx=0, a.=x, b$idx=1, b.=v, c=1
sub2: a$idx=1, a.=y, b$idx=0, b.=u, c=1
sub3: a$idx=1, a.=y, b$idx=1, b.=v, c=1
```

With `disable_cross_array_unnest=true`, sibling arrays are indexed independently:
```text
sub0: a$idx=0, a.=x, c=1
sub1: a$idx=1, a.=y, c=1
sub2: b$idx=0, b.=u, c=1
sub3: b$idx=1, b.=v, c=1
```

This avoids combinatorial sub-doc growth. Queries that constrain values across multiple sibling arrays can sacrifice accuracy because no single flattened sub-doc contains terms from both sibling arrays.

# Example 4: Tantivy nested-object false positive

Tantivy documents this pitfall for JSON arrays: a document is a bag of terms, so `cart.product_type:sneakers AND cart.attributes.color:red` can match even when `sneakers` and `red` come from different objects in the same `cart` array. `FlattenedSubDocs` mode avoids that by preserving each array element as a separate internal inverted-index document.

## Raw Documents
row0 should **not** match because `sneakers` and `red` are in different `cart` objects:
```json
{"cart_id":3234234,"cart":[{"product_type":"sneakers","attributes":{"color":"white"}},{"product_type":"t-shirt","attributes":{"color":"red"}}]}
```

row1 should match because both terms are in the same `cart` object:
```json
{"cart_id":3234235,"cart":[{"product_type":"sneakers","attributes":{"color":"red"}}]}
```

## Flattened Sub-Docs
```text
row0/sub0: cart$idx=0, cart..product_type=sneakers, cart..attributes.color=white, cart_id=3234234
row0/sub1: cart$idx=1, cart..product_type=t-shirt, cart..attributes.color=red, cart_id=3234234
row1/sub0: cart$idx=0, cart..product_type=sneakers, cart..attributes.color=red, cart_id=3234235
```

## Correct Query Shape For Nested-Object Semantics
Use one JSON `MatchQuery` that contains all related JSON triplets separated by `;`, and set `Operator::And` on that `MatchQuery`:
```rust
FtsQuery::Match(
    MatchQuery::new(
        "cart[*].product_type,str,sneakers;cart[*].attributes.color,str,red".to_string(),
    )
    .with_column(Some("json_field".to_string()))
    .with_operator(Operator::And),
)
```

Expected result:
```text
row1 only
```

Here `;` only separates JSON triplets. `Operator::And` is what requires all generated tokens to match, and because they are inside one `MatchQuery`, the intersection happens on flattened sub-doc ids before Lance maps matches back to row ids.

## Anti-Example: Row-Level Boolean Composition
Do not express nested-object constraints as separate `BooleanQuery.must` children:
```rust
FtsQuery::Boolean(BooleanQuery {
    must: vec![
        MatchQuery::new("cart[*].product_type,str,sneakers".to_string())
            .with_column(Some("json_field".to_string()))
            .into(),
        MatchQuery::new("cart[*].attributes.color,str,red".to_string())
            .with_column(Some("json_field".to_string()))
            .into(),
    ],
    should: vec![],
    must_not: vec![],
})
```

Each child `MatchQuery` runs independently and returns Lance row ids. The outer boolean query then composes those row ids, so row0 can still match at row level even though no single `cart` object contains both terms.

## Apache Pinot JSON Index References

- JSON flattening behavior and array index keys: [JsonUtils.flatten](https://github.com/apache/pinot/blob/master/pinot-spi/src/main/java/org/apache/pinot/spi/utils/JsonUtils.java#L380-L586)
- Pinot `disableCrossArrayUnnest` config field: [JsonIndexConfig](https://github.com/apache/pinot/blob/master/pinot-spi/src/main/java/org/apache/pinot/spi/config/table/JsonIndexConfig.java#L34-L131)
- Mutable JSON index SPI, including write entrypoint and reader inheritance: [MutableJsonIndex](https://github.com/apache/pinot/blob/master/pinot-segment-spi/src/main/java/org/apache/pinot/segment/spi/index/mutable/MutableJsonIndex.java#L29-L55)
- Immutable JSON index writer SPI: [JsonIndexCreator](https://github.com/apache/pinot/blob/master/pinot-segment-spi/src/main/java/org/apache/pinot/segment/spi/index/creator/JsonIndexCreator.java#L30-L65)
- Immutable JSON index reader SPI: [JsonIndexReader](https://github.com/apache/pinot/blob/master/pinot-segment-spi/src/main/java/org/apache/pinot/segment/spi/index/reader/JsonIndexReader.java#L32-L110)
- Immutable writer implementation: [BaseJsonIndexCreator](https://github.com/apache/pinot/blob/master/pinot-segment-local/src/main/java/org/apache/pinot/segment/local/segment/creator/impl/inv/json/BaseJsonIndexCreator.java#L99-L175)
- Mutable implementation read/write behavior: [MutableJsonIndexImpl](https://github.com/apache/pinot/blob/master/pinot-segment-local/src/main/java/org/apache/pinot/segment/local/realtime/impl/json/MutableJsonIndexImpl.java#L547-L671)
- Immutable reader implementation: [ImmutableJsonIndexReader](https://github.com/apache/pinot/blob/master/pinot-segment-local/src/main/java/org/apache/pinot/segment/local/segment/index/readers/json/ImmutableJsonIndexReader.java#L846-L885)

## Test Plan

Rust:
- `cargo fmt --all --check`
- `cargo check --workspace --tests --benches` (passed)
- `cargo clippy --all --tests --benches -- -D warnings` (passed)
- `cargo test -p lance-index scalar::inverted::tokenizer:: --lib` (52 passed)

Python:
- rebuilt the extension
- `uv run pytest python/tests/test_scalar_index.py::test_json_inverted_match_query` from `python/` (1 passed)
- `uv run make lint` from `python/` (passed; existing missing-stub warnings only)

Java/JNI:
- `./mvnw spotless:check` from `java/` (passed)
- focused `InvertedIndexParamsTest` (14 passed)
- JNI Rust tests (22 passed)

### DIFF
--- a/protos/index_old.proto
+++ b/protos/index_old.proto
@@ -57,6 +57,11 @@ message InvertedIndexDetails {
     LIST_ELEMENT = 1;
   }
 
+  enum MaxSubDocsPerRowExceedAction {
+    FAIL = 0;
+    SKIP_ROW = 1;
+  }
+
   message CodeTokenizerConfig {
     /* Split one lexical identifier into subwords, e.g. getUserName ->
      * get/user/name.
@@ -121,4 +126,12 @@ message InvertedIndexDetails {
   // If true, avoid cross-array unnesting during flattened JSON tokenization.
   // The default false value preserves exact Cartesian-product semantics.
   bool disable_cross_array_unnest = 17;
+  /* Maximum flattened sub-documents emitted for one JSON row. Absence means
+   * unlimited.
+   */
+  optional uint64 max_sub_docs_per_row = 18;
+  /* Action taken when max_sub_docs_per_row is exceeded. FAIL aborts index
+   * ingestion; SKIP_ROW omits the source row from the index.
+   */
+  MaxSubDocsPerRowExceedAction max_sub_docs_per_row_exceed_action = 19;
 }
--- a/protos/index_old.proto
+++ b/protos/index_old.proto
@@ -115,4 +115,10 @@ message InvertedIndexDetails {
    * which identifies the overall inverted-index layout.
    */
   optional uint32 posting_format_version = 15;
+  // JSON document tokenization mode. Absent means SingleDocument JSON tokenization,
+  // which is how indexes written before flattened JSON sub-docs are interpreted.
+  optional string json_tokenizer_mode = 16;
+  // If true, avoid cross-array unnesting during flattened JSON tokenization.
+  // The default false value preserves exact Cartesian-product semantics.
+  bool disable_cross_array_unnest = 17;
 }
--- a/python/python/lance/dataset.py
+++ b/python/python/lance/dataset.py
@@ -3720,6 +3720,12 @@ class LanceDataset(pa.dataset.Dataset):
             ``[1, num_compute_cpus]``. If unset, Lance uses ``num_compute_cpus``
             workers unless ``LANCE_FTS_NUM_SHARDS`` is set. This parameter is
             only used for the current build and is not persisted with the index.
+        disable_cross_array_unnest: bool, default False
+            This is for the ``INVERTED`` index on JSON columns. If True, flattened
+            JSON tokenization indexes sibling arrays independently instead of
+            producing their Cartesian product. This reduces index build memory for
+            records with multiple arrays but can sacrifice result accuracy for
+            queries that constrain values across those arrays.
         base_tokenizer: str, default "simple"
             This is for the ``INVERTED`` index. The base tokenizer to use. The
             value can be:
--- a/python/python/lance/dataset.py
+++ b/python/python/lance/dataset.py
@@ -3726,6 +3726,14 @@ class LanceDataset(pa.dataset.Dataset):
             producing their Cartesian product. This reduces index build memory for
             records with multiple arrays but can sacrifice result accuracy for
             queries that constrain values across those arrays.
+        max_sub_docs_per_row: int, optional
+            This is for the ``INVERTED`` index on JSON columns. Maximum number of
+            flattened sub-documents one source row may produce. If unset, the
+            number is unlimited.
+        max_sub_docs_per_row_exceed_action: str, default "fail"
+            Action when ``max_sub_docs_per_row`` is exceeded. ``"fail"`` aborts
+            index ingestion with tuning guidance. ``"skip_row"`` omits the source
+            row from the index and continues.
         base_tokenizer: str, default "simple"
             This is for the ``INVERTED`` index. The base tokenizer to use. The
             value can be:
--- a/rust/lance-index/src/scalar/inverted/query.rs
+++ b/rust/lance-index/src/scalar/inverted/query.rs
@@ -2,7 +2,7 @@
 // SPDX-FileCopyrightText: Copyright The Lance Authors
 
 use crate::scalar::inverted::DocumentGranularity;
-use crate::scalar::inverted::document_tokenizer::DocType;
+use crate::scalar::inverted::document_tokenizer::{DocType, JsonTokenizerMode};
 use crate::scalar::inverted::tokenizer::document_tokenizer::LanceTokenizer;
 use lance_core::{Error, Result};
 use serde::ser::SerializeMap;
@@ -897,6 +897,25 @@ pub fn collect_query_tokens(text: &str, tokenizer: &mut Box<dyn LanceTokenizer>)
         }
     }
     Tokens::with_positions(tokens, positions, token_type)
+}
+
+#[doc(hidden)]
+pub fn effective_json_query_operator(
+    mode: Option<JsonTokenizerMode>,
+    query_tokens: &Tokens,
+    operator: Operator,
+) -> Operator {
+    if mode == Some(JsonTokenizerMode::FlattenedSubDocs)
+        && query_tokens.into_iter().any(|token| {
+            token
+                .split_once(',')
+                .is_some_and(|(path, value)| path.ends_with("$idx") && value.starts_with("number,"))
+        })
+    {
+        Operator::And
+    } else {
+        operator
+    }
 }
 
 pub fn has_query_token(
--- a/rust/lance-index/src/scalar/inverted/query.rs
+++ b/rust/lance-index/src/scalar/inverted/query.rs
@@ -904,13 +904,19 @@ pub fn has_query_token(
     tokenizer: &mut Box<dyn LanceTokenizer>,
     query_tokens: &Tokens,
 ) -> bool {
-    let mut stream = tokenizer.token_stream_for_doc(text);
-    while let Some(token) = stream.next() {
-        if query_tokens.contains(&token.text) {
-            return true;
+    match tokenizer.token_streams_for_doc(text) {
+        Ok(sub_docs) => {
+            for tokens in sub_docs {
+                for token in tokens {
+                    if query_tokens.contains(&token.text) {
+                        return true;
+                    }
+                }
+            }
+            false
         }
+        Err(_) => false,
     }
-    false
 }
 
 fn fill_match_query_columns(
--- a/rust/lance-index/src/scalar/inverted/tokenizer/document_tokenizer.rs
+++ b/rust/lance-index/src/scalar/inverted/tokenizer/document_tokenizer.rs
@@ -4,14 +4,50 @@
 use arrow_schema::{DataType, Field};
 use lance_arrow::ARROW_EXT_NAME_KEY;
 use lance_arrow::json::JSON_EXT_NAME;
+use lance_core::Error;
 use lance_tokenizer::{BoxTokenStream, TextAnalyzer, Token, TokenStream};
+use serde::{Deserialize, Serialize};
 use serde_json::Value;
+use std::str::FromStr;
 
 /// Document type for full text search.
 #[derive(Debug, Clone)]
 pub enum DocType {
     Text,
     Json,
+}
+
+/// Controls how JSON documents are represented inside the inverted index.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub enum JsonTokenizerMode {
+    /// Emit one token stream for each source JSON document.
+    SingleDocument,
+    /// Flatten arrays into multiple sub-doc token streams for each source JSON document.
+    FlattenedSubDocs,
+}
+
+impl AsRef<str> for JsonTokenizerMode {
+    fn as_ref(&self) -> &str {
+        match self {
+            Self::SingleDocument => "single_document",
+            Self::FlattenedSubDocs => "flattened_sub_docs",
+        }
+    }
+}
+
+impl FromStr for JsonTokenizerMode {
+    type Err = Error;
+
+    fn from_str(value: &str) -> std::result::Result<Self, Self::Err> {
+        match value {
+            "single_document" => Ok(Self::SingleDocument),
+            "flattened_sub_docs" => Ok(Self::FlattenedSubDocs),
+            _ => Err(Error::invalid_input(format!(
+                "unknown JSON tokenizer mode {value:?}; expected 'single_document' or 'flattened_sub_docs'"
+            ))),
+        }
+    }
 }
 
 impl AsRef<str> for DocType {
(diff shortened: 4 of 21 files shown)


wirybeaver:
@beinan @jja725 Would you mind taking a look at this Json Index improvement proposal shown in the PR summary? If no major concerns, I would go ahead to refine the PR.

protos/index_old.proto
```diff
@@ -39,4 +39,10 @@ message InvertedIndexDetails {
   uint32 min_ngram_length = 9;
   uint32 max_ngram_length = 10;
   bool prefix_only = 11;
+  // JSON document tokenization mode. Absent means SingleDocument JSON tokenization,
+  // which is how indexes written before flattened JSON sub-docs are interpreted.
+  optional string json_tokenizer_mode = 12;
```
westonpace: Should we make this an enum? There are only two choices at the moment and I wouldn't expect there to be many.

python/python/lance/dataset.py
```diff
@@ -3261,6 +3255,12 @@ def create_scalar_index(
             ``[1, num_compute_cpus]``. If unset, Lance uses ``num_compute_cpus``
             workers unless ``LANCE_FTS_NUM_SHARDS`` is set. This parameter is
             only used for the current build and is not persisted with the index.
+        disable_cross_array_unnest: bool, default False
+            This is for the ``INVERTED`` index on JSON columns. If True, flattened
+            JSON tokenization indexes sibling arrays independently instead of
+            producing their Cartesian product. This reduces index build memory for
+            records with multiple arrays but can sacrifice result accuracy for
+            queries that constrain values across those arrays.
         base_tokenizer: str, default "simple"
```
westonpace, line 3729: How does the user specify the json tokenizer mode?  Is it through the `base_tokenizer`?

westonpace (commented):
This seems like a useful addition to the JSON full text search.  I think `disable_cross_array_unnest` is going to be pretty technical for the user to judge.  It might be nice to have some guardrails.  For example, if a single document generates more than X subdocs (or maybe more than X subdocs per byte of the original doc) we could log a warning like "more than X subdocuments generated for a single document, if you have sibling arrays you may need to disable cross array unnesting to avoid a combinatorial explosion".  Or we could have a "max_subdocs_per_doc" variable that, if exceeded, would fail the index build.

In the other direction we could potentially add some kind of conservative one-time warning.  If the user disabled cross array unnesting and then queried the index using multiple constraints.  We wouldn't easily know they were multiple sibling array constraints but the warning could be something like "you are querying a JSON text index with multiple constraints and the index was built with cross array unnesting disabled.  If those constraints are on two different sibling arrays you may get false negatives.  To disable this warning set the environment variable ..."

Ideally we will want some examples added to the docs as well.  We could probably expand https://lance.org/guide/json/#full-text-search-on-json-documents

rust/lance-index/src/scalar/inverted/query.rs
```diff
@@ -811,13 +811,19 @@ pub fn has_query_token(
...
+            for tokens in sub_docs {
+                for token in tokens {
+                    if query_tokens.contains(&token.text) {
+                        return true;
+                    }
+                }
+            }
+            false
         }
+        Err(_) => false,
```
jja725, line 937: should we add some error trace here before return

rust/lance-index/src/scalar/inverted/tokenizer/document_tokenizer.rs
```diff
@@ -277,6 +433,168 @@ fn flatten_json(
...
+                }
+                sub_docs.extend(child_terms);
+            }
+            sub_docs
+        }
+        Value::String(text) => {
+            let mut token_texts = Vec::new();
+            let mut tokens = tokenizer.token_stream(text);
+            while let Some(token) = tokens.next() {
+                token_texts.push(format!("{prefix},str,{}", token.text));
```
jja725, line 557: maybe we can Add a max_sub_docs_per_row cap with a clear error or warning when exceeded

jja725 (commented):
Thanks for this great contribution! this would help a lot with FTS

wirybeaver:
@jja725 @westonpace Sorry for the late reply. Somehow I miss the email notification about your folks' feedback during vacation. I will work on the improvement this week. Appreciate you folks compliment on this idea.

In terms of adding the parameter "max_sub_docs_per_row", do you folks prefer failing the whole indexing process OR logging an error but keep the already flattened docs?

westonpace:
> In terms of adding the parameter "max_sub_docs_per_row", do you folks prefer failing the whole indexing process OR logging an error but keep the already flattened docs?

Either approach is fine.  Failing the entire operation is safest but could be quite annoying if there is only one bad row.  In vector indexing, if we encounter NaN / NULL embeddings, we just exclude that row from the index (so it will never appear in search) so there is some precedence for just skipping too.

wirybeaver:
Following up on `max_sub_docs_per_row`, I propose adding two JSON index params before implementation:

- `max_sub_docs_per_row`: optional cap on how many flattened sub-docs one source JSON row may produce. If unset, behavior remains unlimited, so Lance will keep flattening until normal memory/resource limits apply.
- `max_sub_docs_per_row_exceed_action`: enum scoped to this cap, with initial values `fail` and `skip_row`. `fail` aborts indexing with an error that suggests increasing `max_sub_docs_per_row` or setting `disable_cross_array_unnest=true`; `skip_row` omits the whole offending source row from the index and continues.

Default proposal: unset `max_sub_docs_per_row` and `max_sub_docs_per_row_exceed_action=fail`, so existing behavior is unchanged unless a user opts into the cap.

I'll wait for feedback from @westonpace and @jja725 before implementing this.

PR #7378, opened 2026-06-19
fix(compaction): prune fully-deleted fragments from the fragment-reuse index

### LINKED ISSUE #7374: Vector index can become corrupted when compaction is deferred
During remap we normally prune fragments that have been deleted.  When using the FRI (deferred compaction) we skip this remap phase and so the deleted fragment is not removed.  The index then returns references to the deleted fragment and we get "take received reference to fragment that does not exist" style errors.

Reproduction Script

```
#!/usr/bin/env python3
"""Reproduce index corruption via compaction with deferred remap."""

import numpy as np
import lance
import pyarrow as pa
import shutil
import os

DATASET_PATH = "/tmp/corruption_test"
ROWS_PER_FRAGMENT = 1000
NUM_FRAGMENTS = 10
DIM = 128
FRAGMENT_TO_DELETE = 4  # 0-indexed; delete all rows from fragment 4

def main():
    # Clean up any previous run
    if os.path.exists(DATASET_PATH):
        shutil.rmtree(DATASET_PATH)

    # Step 1: Create dataset with 10 fragments (1K rows each)
    print("=== Step 1: Creating dataset with 10 fragments ===")
    rng = np.random.default_rng(42)

    schema = pa.schema([
        pa.field("id", pa.int64()),
        pa.field("vec", pa.list_(pa.float32(), DIM)),
    ])

    ds = None
    for frag_idx in range(NUM_FRAGMENTS):
        start_id = frag_idx * ROWS_PER_FRAGMENT
        ids = np.arange(start_id, start_id + ROWS_PER_FRAGMENT, dtype=np.int64)
        vecs = rng.standard_normal((ROWS_PER_FRAGMENT, DIM)).astype(np.float32)

        table = pa.table({
            "id": ids,
            "vec": pa.array(vecs.tolist(), type=pa.list_(pa.float32(), DIM)),
        })

        if ds is None:
            ds = lance.write_dataset(table, DATASET_PATH, schema=schema)
        else:
            ds = lance.write_dataset(table, DATASET_PATH, mode="append")

    frags = ds.get_fragments()
    print(f"  Total rows: {ds.count_rows()}")
    print(f"  Fragments: {len(frags)}")
    for i, f in enumerate(frags):
        print(f"    Fragment {i}: id={f.fragment_id}, rows={f.count_rows()}")

    # Step 2: Create vector index
    print("\n=== Step 2: Creating vector index ===")
    ds.create_index(
        "vec",
        index_type="IVF_PQ",
        num_partitions=10,
        num_sub_vectors=8,
        replace=True,
    )
    print(f"  Indices: {ds.list_indices()}")

    # Step 3: Delete all rows from one fragment
    print(f"\n=== Step 3: Deleting all rows from fragment {FRAGMENT_TO_DELETE} ===")
    row_start = FRAGMENT_TO_DELETE * ROWS_PER_FRAGMENT
    row_end = row_start + ROWS_PER_FRAGMENT - 1
    ds.delete(f"id >= {row_start} AND id <= {row_end}")
    ds = lance.dataset(DATASET_PATH)
    print(f"  Rows after delete: {ds.count_rows()}")
    frags = ds.get_fragments()
    print(f"  Fragments after delete: {len(frags)}")
    for i, f in enumerate(frags):
        print(f"    Fragment {i}: id={f.fragment_id}, rows={f.count_rows()}")

    # Step 4: Compact with deferred index remap
    print("\n=== Step 4: Compacting with defer_index_remap=True ===")
    metrics = ds.optimize.compact_files(
        target_rows_per_fragment=ROWS_PER_FRAGMENT * NUM_FRAGMENTS,  # merge all into one
        materialize_deletions=True,
        defer_index_remap=True,
    )
    print(f"  Compaction metrics: {metrics}")
    ds = lance.dataset(DATASET_PATH)
    print(f"  Rows after compaction: {ds.count_rows()}")
    frags = ds.get_fragments()
    print(f"  Fragments after compaction: {len(frags)}")

    # Step 5: Search the dataset
    print("\n=== Step 5: Searching dataset (expect corruption / errors) ===")
    query = np.random.default_rng(99).standard_normal(DIM).astype(np.float32)
    try:
        results = ds.scanner(
            nearest={"column": "vec", "q": query, "k": 10},
        ).to_table()
        print(f"  Search returned {len(results)} rows")
        print(f"  Result ids: {results['id'].to_pylist()}")
        # Check for any ids that fall in the deleted fragment range
        deleted_ids = set(range(row_start, row_start + ROWS_PER_FRAGMENT))
        returned_ids = set(results['id'].to_pylist())
        ghost_ids = returned_ids & deleted_ids
        if ghost_ids:
            print(f"  CORRUPTION DETECTED: returned deleted ids: {ghost_ids}")
        else:
            print("  No ghost ids in results (may still be corrupt in other ways)")
    except Exception as e:
        print(f"  ERROR during search: {type(e).__name__}: {e}")

if __name__ == "__main__":
    main()
```

xuanyu-z:
Fixes #7374.

## Problem

When compaction is deferred (`defer_index_remap: true`), index entries are not physically remapped at compaction time; instead the fragment-reuse index (FRI) records old→new row-address maps and rewrites addresses on the fly via `remap_row_id`.

If a fragment is **fully deleted before** the compaction, `apply_deletions` removes it from the manifest entirely (no deletion file). It therefore never enters a compaction task and is absent from every rewrite group's `old_fragments`. An index that still covered that fragment keeps its stale entries, and because no group references the fragment, `remap_row_id` passes those addresses straight through — resurfacing as dangling references to a fragment that no longer exists.

The corruption is latent (`effective_fragment_bitmap` masks the dead fragment at query time) but surfaces once another rewrite touches the affected coverage.

## Fix

Track such orphaned fragments explicitly and prune their rows in the FRI:

- New additive proto field `FragmentReuseIndexDetails.Version.removed_fragments` (field 4) and `FragReuseVersion.removed_frags`.
- At `commit_compaction` (deferred branch), compute the orphaned set as `covered − live_fragments`, where `covered` is the union of every non-system index's coverage (read post-FRI-remap via `load_indices`, so fragments rewritten by earlier deferred compactions are already excluded) and `live_fragments` is the pre-commit fragment set. The difference is exactly the fragments an index still points at that have vanished from the manifest.
- `remap_row_id` maps any row whose **final** address (after chaining through all versions) lands in a `removed_fragments` fragment to `None`.

### Why track by fragment ID rather than per-row address

A fully-deleted fragment is gone from the manifest, so its `physical_rows`/address range can no longer be enumerated to build per-address `addr → None` entries. Since the entire fragment is gone, fragment-ID granularity is exact.

### Why the prune is applied after chaining, not before

A fragment can be removed at any generation. A row originally in fragment A may have been rewritten by an earlier deferred compaction into fragment B, which is then deleted. The orphaned fragment (B) shows up only at the *final* mapped address, not the input one. A removed fragment is always terminal (nothing compacts a manifest-removed fragment forward) and fragment IDs are never reused, so checking the final address never prunes a live row.

### Coverage is intentionally left intact

`remap_fragment_bitmap` does **not** strip orphaned fragments from index coverage. Their rows are already pruned from index data by `remap_row_id`, and a nonexistent fragment is masked at query time by `effective_fragment_bitmap`. Stripping coverage instead makes the planner treat the dead fragment's row range as an unindexed flat-scan fallback, which fails on `take`.

## Backward compatibility

`removed_fragments` is an additive proto field. The FRI is persisted as protobuf; an older index decodes the absent field as an empty list, so `remap_row_id` short-circuits and behaves exactly as before. No migration is required.

## Tests

The corruption is latent end-to-end (`create_restricted_deletion_mask` blocks missing fragments and `effective_fragment_bitmap` masks dead ones), so the core cases assert at the **FRI level** (`remap_row_id` must return `None`) via a shared `assert_fri_prunes` helper, and keep faithful end-to-end checks where a non-empty projection actually exercises the `take`. Shared helpers (`dataset_with_ivf_pq`, `dataset_with_btree`, `assert_fri_prunes`, `assert_knn_no_ghosts`) keep the cases small; FRI-level-only cases use a cheap scalar BTREE index (no vector training).

- `test_defer_index_remap_fully_deleted_fragment` — the #7374 repro: a fully-deleted fragment, deferred-compacted. Asserts the FRI prunes it and KNN returns no ghosts **during the FRI window**, then again **after the physical remap + FRI trim** that closes the window.
- `test_defer_index_remap_deleted_merged_fragment` — a fragment merged by one deferred compaction then fully deleted before a second; pruned at its post-merge ID while survivors are kept. (Fails if the prune is applied to the input address instead of the final one.)
- `test_defer_index_remap_multiple_fully_deleted_fragments` — several whole-fragment deletions before one compaction; verifies `removed_frags` is a union.
- `test_defer_index_remap_survivor_deleted_across_versions` — a fragment left untouched by an earlier deferred compaction and then deleted (pruned via pass-through, not chaining).
- `test_defer_index_remap_then_delete_during_window` — reverse ordering: compaction then delete during the FRI window; the new fragment's deletion vector is honored.

### DIFF
--- a/rust/lance-table/src/system_index/frag_reuse.rs
+++ b/rust/lance-table/src/system_index/frag_reuse.rs
@@ -7,6 +7,7 @@ use arrow_array::cast::AsArray;
 use arrow_array::types::UInt64Type;
 use arrow_array::{Array, ArrayRef, PrimitiveArray, RecordBatch, UInt64Array};
 use lance_core::deepsize::{Context, DeepSizeOf};
+use lance_core::utils::address::RowAddress;
 use lance_core::{Error, Result};
 use lance_select::RowAddrTreeMap;
 use roaring::{RoaringBitmap, RoaringTreemap};
@@ -105,6 +106,10 @@ impl TryFrom<pb::fragment_reuse_index_details::Group> for FragReuseGroup {
 pub struct FragReuseVersion {
     pub dataset_version: u64,
     pub groups: Vec<FragReuseGroup>,
+    /// Fragments an index covered that were gone from the manifest at commit
+    /// time and not part of any rewrite group (e.g. fully deleted before
+    /// compaction). Pruned by ID since their rows can't be enumerated by address.
+    pub removed_frags: Vec<u32>,
 }
 
 impl From<&FragReuseVersion> for pb::fragment_reuse_index_details::Version {
@@ -112,6 +117,7 @@ impl From<&FragReuseVersion> for pb::fragment_reuse_index_details::Version {
         Self {
             dataset_version: version.dataset_version,
             groups: version.groups.iter().map(|g| g.into()).collect(),
+            removed_fragments: version.removed_frags.clone(),
         }
     }
 }
@@ -127,6 +133,7 @@ impl TryFrom<pb::fragment_reuse_index_details::Version> for FragReuseVersion {
                 .into_iter()
                 .map(FragReuseGroup::try_from)
                 .collect::<Result<_>>()?,
+            removed_frags: version.removed_fragments,
         })
     }
 }
@@ -194,6 +201,15 @@ impl FragReuseIndexDetails {
                 .flat_map(|v| v.new_frag_ids().into_iter().map(|id| id as u32)),
         )
     }
+
+    /// Union of every version's `removed_frags`.
+    pub fn removed_frag_bitmap(&self) -> RoaringBitmap {
+        RoaringBitmap::from_iter(
+            self.versions
+                .iter()
+                .flat_map(|v| v.removed_frags.iter().copied()),
+        )
+    }
 }
 
 /// An index that stores row ID maps.
@@ -204,6 +220,9 @@ pub struct FragReuseIndex {
     pub uuid: Uuid,
     pub row_id_maps: Vec<HashMap<u64, Option<u64>>>,
     pub details: FragReuseIndexDetails,
+    /// Cache of `details.removed_frag_bitmap()`; rebuilt in `new`, so skipped.
+    #[serde(skip)]
+    removed_frags: RoaringBitmap,
 }
 
 impl DeepSizeOf for FragReuseIndex {
@@ -218,10 +237,12 @@ impl FragReuseIndex {
         row_id_maps: Vec<HashMap<u64, Option<u64>>>,
         details: FragReuseIndexDetails,
     ) -> Self {
+        let removed_frags = details.removed_frag_bitmap();
         Self {
             uuid,
             row_id_maps,
             details,
+            removed_frags,
         }
     }
 
@@ -234,6 +255,20 @@ impl FragReuseIndex {
                     .copied()
                     .unwrap_or(mapped_value);
             }
+        }
+
+        // Prune rows whose final address lands in a fully-deleted fragment
+        // (tracked by ID in `removed_frags`). Checked after chaining, since an
+        // earlier compaction may have rewritten the row into a fragment that was
+        // later deleted. Fragment IDs are never reused, so a live row is never
+        // pruned.
+        if let Some(addr) = mapped_value
+            && !self.removed_frags.is_empty()
+            && self
+                .removed_frags
+                .contains(RowAddress::new_from_u64(addr).fragment_id())
+        {
+            return None;
         }
 
         mapped_value
@@ -307,6 +342,13 @@ impl FragReuseIndex {
     }
 
     pub fn remap_fragment_bitmap(&self, fragment_bitmap: &mut RoaringBitmap) -> Result<()> {
+        // Orphaned fragments (fully deleted before compaction) are deliberately
+        // NOT removed from index coverage here. Their rows are already pruned
+        // from the index data by `remap_row_id`, and a fragment that no longer
+        // exists is masked out at query time by `effective_fragment_bitmap`'s
+        // intersection with the live fragments. Stripping a deleted fragment's
+        // coverage instead makes the planner treat its (now nonexistent) row
+        // range as an unindexed flat-scan fallback, which fails on `take`.
         for version in self.details.versions.iter() {
             for group in version.groups.iter() {
                 let mut removed = 0;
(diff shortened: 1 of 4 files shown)


xuanyu-z:
@claude review once

rust/lance-table/src/system_index/frag_reuse.rs
```diff
@@ -204,6 +220,9 @@ pub struct FragReuseIndex {
     pub uuid: Uuid,
     pub row_id_maps: Vec<HashMap<u64, Option<u64>>>,
     pub details: FragReuseIndexDetails,
+    /// Cache of `details.removed_frag_bitmap()`; rebuilt in `new`, so skipped.
+    #[serde(skip)]
+    removed_frags: RoaringBitmap,
```
westonpace, line 225: This is probably unrelated to this PR but why do we store the FRI as a collection of versions in the first place?  Why not merge them all together at write time?

rust/lance/src/dataset/optimize.rs
```diff
@@ -3279,6 +3309,357 @@ mod tests {
...
+            )
+            .await
+            .unwrap();
+        dataset
+    }
+
+    /// Top-200 KNN projecting `i` (the projection forces a `take`, so a stale
+    /// row address would error). Asserts no result id falls in `ghost_range`
+    /// and at least one live row is returned.
+    async fn assert_knn_no_ghosts(dataset: &Dataset, ghost_range: std::ops::Range<i32>) {
```
westonpace, line 3418: Ghosts is a good word for the phenomenon 👻

westonpace (approved):
Good catch for this error.  I have one minor question but this is good otherwise.

rust/lance-table/src/system_index/frag_reuse.rs
xuanyu-z: From the git history and code comments and my understandings, the main benefit is versioning allows the system to trim the FRI as different indices catch up so the size of the FRI is not unbounded before a full index job kicks in.

Then versioning also makes FRI ~append only which inherit all the consistency benefits

PR #7381, opened 2026-06-19
docs: specify data overlay files for the table format

wjones127:
Adds a specification for **data overlay files**: small files attached to a fragment that supply new values for a subset of `(row offset, field)` cells without rewriting the base data files. They make cell-level updates cheap when only a small fraction of rows and/or columns change.

This PR is **spec + proto only** — no read/write implementation yet. It is also explicitly *experimental*. The released libraries will not produce tables with this feature enabled. Once the implementation is done in the library, we will vote on the final design before releasing. This is similar to how we have done file format updates.

## Changes
- **`protos/table.proto`**
  - Rework `DataOverlayFile`: a `oneof coverage { bytes shared_offset_bitmap | FieldCoverage field_coverage }` to support both dense (rectangular) and sparse overlays; add the `FieldCoverage` message.
  - Rename `read_version` → `committed_version` (`uint64`), with effective/commit-stamped semantics so overlay-vs-index ordering is correct.
  - Drop the in-file offset key column in favor of rank-based addressing off the coverage bitmap.
  - Document reader feature flag `64` (and previously-undocumented `16`/`32`).
- **`docs/src/format/table/data_overlay_file.md`** (new): full specification — coverage/resolution, deletion precedence, NULL-override, layout + rank addressing, dense vs. sparse, versioning, field-aware index exclusion with flat re-evaluation, the correctness invariant, both compaction modes, row lineage, a worked example (write → read → index query → sparse write → read → compaction), and a guidance stub with open questions.
- **`docs/src/format/table/index.md`**: concise overview + link to the new spec (replacing the earlier inline sketch).

## Out of scope / follow-ups
- Write transaction shape (new `Operation` variant in `transaction.proto` + Rust).
- Writer support for unequal-length columns (needed for single-file sparse overlays).
- Coverage bitmap external spill for very large coverage.
- Per-fragment vs. per-table overlays / LSM analogy (open question in the doc).

## Summary by CodeRabbit

* **New Features**
  * Added documentation for experimental Data Overlay Files, including storage, versioning, querying, compaction, and transaction behavior.
  * Added format and transaction schema support for describing data overlays.
  * Added navigation links to the new documentation.

* **Bug Fixes**
  * Datasets using unsupported data overlays are now explicitly rejected instead of risking stale results.

* **Documentation**
  * Expanded guidance on invalidated index results and overlay-related filtering scenarios.


wjones127:
See discussion: https://github.com/lance-format/lance/discussions/7401

docs/src/format/table/data_overlay_file.md
```diff
@@ -0,0 +1,408 @@
...
+1. Run the index search as usual, producing candidate rows.
+2. Remove any candidate in the exclusion set. (Its indexed value may be stale.)
+3. **Re-evaluate** the excluded rows against their current values — the same flat
+   path already used for the unindexed tail of fragments. For a scalar predicate
+   this re-applies the filter; for a vector query it re-scores the row's current
+   vector. Rows that still match are added back to the result.
+
+Step 3 is what makes exclusion correct rather than merely safe: removing a row
+from index candidates without re-evaluating it would silently drop a row that
+should match under its new value.
```
hamersaw, line 161: IIUC this may not be enough to ensure accurate results in case of overlay updates -- if that is indeed the goal here. This covers the case where an query matching row was updated to no longer match, but what about one where the row previously didn't match and the updated value does? I think the only way to ensure 100% correctness is to brute force every update in the overlay.

wjones127: Yeah I see. We need to brute force a full exclusion set, not just the intersection between the candidate rows and the full exclusion set. Good point, I will update.

wjones127:
I've prototyped the basic write, scan, and take paths.

* Write: https://github.com/lance-format/lance/pull/7535#issuecomment-4847769769
  * Overlay files have a significant impact on manifest size when > 5% of rows have overlays and they are sparse rather than contiguous.
  * Overlays are significantly faster and more IO efficient for targeted updates than update or merge insert.
* Scan & Take: https://github.com/lance-format/lance/pull/7536#issuecomment-4848828467
  * Overlays have negligible impact on take performance
  * Overlays can have a *large impact on scan performance*. Having 16 overlays can make scanning 3.5x slower.

docs/src/format/table/data_overlay_file.md
```diff
@@ -0,0 +1,408 @@
...
+An overlay changes individual cells.
+
+## Concepts
+
+### Coverage and resolution
+
+Each overlay declares which cells it provides through a **coverage** bitmap (or,
+for sparse overlays, one bitmap per field). The bitmaps index **physical row
+offsets** — positions in the base data files, counting deleted rows — so they are
+stable across deletions, exactly like deletion vectors.
```
westonpace: Minor nit but maybe break this into two sentences.  Bitmaps refer to row offsets.  These offsets include deleted rows.

docs/src/format/table/data_overlay_file.md
```diff
@@ -0,0 +1,408 @@
...
+
+To resolve a cell `(offset, field)` on read, walk the fragment's overlays from
+**newest to oldest**. The first overlay that covers `(offset, field)` wins; its
+value is used. If no overlay covers the cell, the value falls through to the base
+data file (or is `NULL` if no base data file holds that field).
+
+Precedence among overlays is determined by:
+
+1. `committed_version` — higher wins (see [Versioning](#versioning-and-ordering)).
+2. Position in `DataFragment.overlays` as a tiebreaker — a later entry is newer.
```
westonpace, line 44: Does this mean a single commit can add multiple overlays that target the same field?  Is there a reason for this?  Or are we just being exhaustive?

docs/src/format/table/data_overlay_file.md
```diff
@@ -0,0 +1,408 @@
...
+- **Sparse.** A `FieldCoverage` carries one bitmap per field, used when different
+  fields cover different offset sets — for example a `MERGE` with multiple
+  `WHEN MATCHED` branches, where different rows update different columns. A dense
+  overlay would have to widen to the bounding rectangle and fill the untouched
+  cells with their current values (post-images), which for wide columns such as
+  embeddings means re-storing data that did not change. A sparse overlay stores
+  exactly the changed cells.
+
+A writer may always express a non-rectangular update as **multiple dense overlays
+in one transaction** (one per coverage group) instead of a single sparse overlay.
```
westonpace: Why have two ways to accomplish the same thing?  As a user, which should I prefer?  A sparse overlay or multiple dense overlays?

docs/src/format/table/data_overlay_file.md
```diff
@@ -0,0 +1,408 @@
...
+is the final word on a row, so a concurrent overlay against a row that was
+deleted needs no special conflict handling — its values are merely inert.
+
+### Physical layout
+
+An overlay's data file stores **one value column per field**, in the order of
+`data_file.fields`. It does **not** store a row-offset key column. The position of
+a covered offset's value within its column is the **rank** of that offset in the
+field's coverage bitmap — the number of set bits below it. For a Roaring bitmap
+this is an O(1) operation, so random access to any cell is a rank computation
```
westonpace: I'm not sure I believe that rank is `O(1)`.

docs/src/format/table/data_overlay_file.md
```diff
@@ -0,0 +1,408 @@
...
+<details>
+<summary>FieldCoverage protobuf message</summary>
+
+```protobuf
+%%% proto.message.FieldCoverage %%%
+```
+
+</details>
+
+## Versioning and ordering
```
westonpace, line 107: How do overlays participate in conflict resolution?  If I try and apply an overlay on row X and another transaction updates row X is this a conflict?  Wouldn't I need the read version to detect this kind of conflict?

docs/src/format/table/data_overlay_file.md
```diff
@@ -0,0 +1,408 @@
...
+only unrelated columns does not exclude anything from the index on `F`.
+
+The query then proceeds as:
+
+1. Run the index search as usual, producing candidate rows.
+2. Remove any candidate in the exclusion set. (Its indexed value may be stale.)
+3. **Re-evaluate** the excluded rows against their current values — the same flat
+   path already used for the unindexed tail of fragments. For a scalar predicate
+   this re-applies the filter; for a vector query it re-scores the row's current
+   vector. Rows that still match are added back to the result.
```
westonpace, line 157: How does this work?  Do we do a "take style" filtered read on these rows?  The current flat path is always a "scan style" filtered read.  Take style is probably fine, I'm just curious.

docs/src/format/table/data_overlay_file.md
```diff
@@ -0,0 +1,408 @@
...
+Step 3 is what makes exclusion correct rather than merely safe: removing a row
+from index candidates without re-evaluating it would silently drop a row that
+should match under its new value.
+
+### Correctness invariant
+
+> For every indexed field `F` and every row offset `o` in a fragment the index
+> covers, the index's entry for `(o, F)` is trusted unless `o` is excluded.
+> `o` is excluded iff some overlay with `committed_version > index.dataset_version`
+> covers `(o, F)`.
```
westonpace: This is rather formal and feels out of place and I think this invariant is stated fairly clearly above.  Maybe having it in formal logic makes it easier for LLM consumption though?  I'm not really opposed to it but if we're going to be formal 😛 ...

``o` is excluded iff some overlay with ...`

^-- this is not an `iff` (but an `if`) because there are other reasons `o` may be excluded.

I think a better paragraph might be one in the indexing section where we cover all of the various filters that we apply to index search results and why:

* Mask out deleted rows
* Apply the fragment white-list (to deal with invalidated fragments)
* Mask out overlay rows (new with this PR)

docs/src/format/table/data_overlay_file.md
```diff
@@ -0,0 +1,408 @@
...
+- **Writes** change a cell only by adding an overlay, and that overlay's
+  `committed_version` exceeds the version of any pre-existing index — so the
+  change is always covered by an exclusion.
+- **Compaction** may remove an overlay only if the index no longer relies on it
+  (see below).
+
+## Compaction
+
+Overlays accumulate read cost — every overlay is a bitmap to test and a possible
+file to open. Compaction bounds that cost in two modes:
```
westonpace: Not to mention we have to rewrite the Arrow arrays which is probably the most significant portion of the cost.

docs/src/format/table/data_overlay_file.md
```diff
@@ -0,0 +1,408 @@
...
+!!! warning "Folding an indexed field must update its index"
+
+    An overlay→base fold removes the overlay, which removes the exclusion signal
+    that kept an index correct. Folding an overlay that covers an indexed field
+    `F` is therefore equivalent to a column rewrite of `F` and must, in the same
+    commit, either rebuild the index to a `dataset_version` at least the folded
+    overlay's `committed_version`, or remove the fragment from the index's
+    coverage so the rows fall to the flat path. Otherwise the index would serve
+    stale values with no overlay to exclude them. This is the same rule that
+    already governs rewriting a column that an index is built on.
```
westonpace, line 199: Is "folding" overlays into the base a distinct operation from regular compaction (rewrite)?  If not, why call this out here?  This should already be covered in any section on compaction.

Follow-up question, folding overlays into a base fragment means we are making a new fragment (with a new fragment id) correct?

docs/src/format/table/data_overlay_file.md
```diff
@@ -0,0 +1,408 @@
...
+## Compaction
+
+Overlays accumulate read cost — every overlay is a bitmap to test and a possible
+file to open. Compaction bounds that cost in two modes:
+
+- **Overlay → overlay.** Merge several overlays into fewer, computing the
+  post-image per `(offset, field)` by walking the merged overlays newest-first.
+  The merged overlay takes the **maximum** `committed_version` of its inputs, so
+  the exclusion semantics are preserved. Indexes are unaffected. This is cheap and
+  does not touch the base.
```
westonpace: I think the result is still valid but I wouldn't say indexes are unaffected.  Imagine I have an index with read version 25.  If I have an old overlay (committed version is 10) and a new overlay (committed version is 50) and I fold them into one (commited version will be 50) I will now be excluding _more_ rows than before correct?

docs/src/format/table/data_overlay_file.md
```diff
@@ -0,0 +1,408 @@
...
+
+## Row lineage
+
+An overlay write updates the `last_updated_at_version` of every covered row, so
+change-data-feed and time-travel queries observe the update. Because overlays are
+addressed by physical offset, they do **not** require stable row IDs to be
+enabled; lineage updates apply only when those features are on.
+
+## Worked example
+
```
westonpace, line 215: Maybe give some introduction for this example.  At first I thought it was specific to `Row lineage` but now I think it is an example of the entire workflow?

docs/src/format/table/data_overlay_file.md
```diff
@@ -0,0 +1,408 @@
...
+`data/file1.lance` with those columns. In the old file, fields 2, 3, and 4 are
+tombstoned (`-2`); field 1 (`id`) remains. The fragment's `overlays` list is
+cleared. Row addresses are preserved (a column rewrite, not a row rewrite), so
+stable row IDs and the deletion vector are untouched.
+
+Because the fold removed the overlay that was excluding offset 1 from the `age`
+index, the same commit must reconcile that index: either rebuild it at
+`dataset_version >= 2`, or drop fragment `0` from its coverage so `age` queries
+fall to the flat path. After a rebuild at version 4, no overlay remains and the
+`age` index directly returns `26` for Bob with no exclusion needed.
```
westonpace: Is rebuilding really an option?  Wouldn't that mean that the index update and the rewrite would need to be part of the same transaction?  I don't think we can support that today.  I think you _must_ drop it from the fragment bitmap but you can then rebuild it to include it in the fragment bitmap going forwards.

docs/src/format/table/data_overlay_file.md
```diff
@@ -0,0 +1,408 @@
...
+different costs; a cost/benefit scheduler decides when each is worthwhile, using
+the version gap as a staleness signal.
+
+### Open questions
+
+*(To be resolved.)*
+
+- **Per-fragment vs. per-table overlays.** Overlays are attached per fragment.
+  Should there be a table-level overlay concept, and how would it interact with
+  fragment-level row addressing?
```
westonpace: Why would a table-level overlay be desired?  Are we just making up work here or is there a real need?

docs/src/format/table/data_overlay_file.md
```diff
@@ -0,0 +1,408 @@
...
+- **Relationship to LSM.** Overlays plus compaction resemble an LSM tree (newest
+  layer wins, periodic merge). How far should that analogy be taken, and what do
+  we deliberately do differently given Lance's random-access requirements?
```
westonpace: This seems like a study problem.  Does it need to be here?

docs/src/format/table/data_overlay_file.md
```diff
@@ -0,0 +1,408 @@
...
+- **Coverage bitmap spill.** Coverage bitmaps live inline in the manifest. Very
+  large coverage (an overlay touching many rows) may warrant external spill, as
+  the row-ID and last-updated-at sequences already do above a size threshold.
```
westonpace: Overlays are scoped to a fragment so I think the size is bounded in a way that it is not bounded for row-ID and last-updated-at.

Admittedly, the cumulative size of overlays across all fragments could get quite large but I think if you have that many overlays you have other problems beyond the size of the coverage bitmap.

docs/src/format/table/data_overlay_file.md
```diff
@@ -0,0 +1,408 @@
...
+### Scheduling compaction
+
+<!-- TODO: Fill in with a concrete cost/benefit policy once compaction is
+     implemented and benchmarked. -->
+
+*(To be expanded.)* The overlay→overlay and overlay→base modes have very
+different costs; a cost/benefit scheduler decides when each is worthwhile, using
+the version gap as a staleness signal.
+
+### Open questions
```
westonpace: Should we be persisting open questions in a spec?  Does this get removed when we stabilize the feature?

docs/src/format/table/index.md
```diff
@@ -168,6 +168,42 @@ However, this invalidates row addresses and requires rebuilding indices, which c
...
+!!! warning "Experimental"
+
+    This feature is currently experimental and not yet supported in any library.
+
+<!-- TODO: When overlay file support is implemented, update this note to state
+     the released version that first supports the feature. -->
+
+!!! note "Overlay files require feature flag 64 (data overlay files)"
+
+Overlay files supply new values for a subset of `(row offset, field)` cells within
```
westonpace:
```suggestion
Overlay files supply new values for a subset of cells within
```

Minor nit: this is an overview, the `(row offset, field)` verbiage is just noise at this resolution.

docs/src/format/table/index.md
```diff
@@ -168,6 +168,42 @@ However, this invalidates row addresses and requires rebuilding indices, which c
...
+a fragment without rewriting the base data files. They make updates cheap when only
+a small percentage of rows and/or columns change: a writer appends a small file
+carrying just the changed cells instead of rewriting whole columns or moving rows
+to a new fragment.
+
+On read, each cell is resolved by consulting the fragment's overlays from newest to
+oldest; the first overlay covering that `(offset, field)` wins, otherwise the value
+falls through to the base data file. Indices keep covering the fragment and reconcile
+overlays at query time through a field-aware exclusion set.
+
```
westonpace: We can probably drop this paragraph for an overview doc.

docs/src/format/table/index.md
```diff
@@ -168,6 +168,42 @@ However, this invalidates row addresses and requires rebuilding indices, which c
...
+[Data Overlay Files Specification](data_overlay_file.md).
+
+<details>
+<summary>DataOverlayFile protobuf message</summary>
+
+```protobuf
+%%% proto.message.DataOverlayFile %%%
+```
+
+</details>
```
westonpace: This is already in `data_overlay_file.md`.  Do we really want it here too?

protos/transaction.proto
```diff
@@ -315,6 +315,44 @@ message Transaction {
...
+    // to the new dataset version at commit time (re-stamped on retry), in the
+    // same way as the created-at / last-updated-at version sequences. The fields
+    // touched are read from each overlay's `data_file.fields`.
+    repeated DataOverlayFile overlays = 2;
+  }
+
+  // Attach overlay files to fragments, supplying new values for a subset of
+  // (row offset, field) cells without rewriting the fragments' base data files.
+  // See the DataOverlayFile message in table.proto and the Data Overlay Files
+  // specification for resolution, coverage, and versioning rules.
```
westonpace: What operation is used to merge overlays (overlay->overlay compaction)?  This operation or a different one?

protos/transaction.proto
```diff
@@ -315,6 +315,44 @@ message Transaction {
...
+  //  * Another DataOverlay (any fields): COMPATIBLE. Overlays stack; when two
+  //    overlays cover the same (offset, field) the one with the higher
+  //    `committed_version` wins, so independent backfills never conflict.
+  //  * Append / new fragments: COMPATIBLE.
+  //  * Delete: COMPATIBLE. A deletion takes precedence over an overlay, so an
+  //    overlay value for a deleted offset is inert (no special handling needed).
+  //  * DataReplacement or column-rewrite (Update with REWRITE_COLUMNS) of the
+  //    same field: COMPATIBLE. Both preserve physical row addresses, so overlay
+  //    offsets stay valid; the overlay is newer and wins its covered cells, and
+  //    the version gate excludes those cells from any rebuilt index.
```
westonpace: What if I have a column with an overlay file.  This has been the stable state for some time.  Later I want to replace the column with DataReplacement.  Doesn't this conflict rule mean that the overlay will still sit on top?  In other words, there is no way to overwrite an overlay with a `DataReplacment` until I've folded it in with a rewrite?

protos/transaction.proto
```diff
@@ -315,6 +315,44 @@ message Transaction {
...
+  //  * Row-rewrite, compaction, or an overlay->base fold of the fragment:
```
westonpace: Is this just three different ways to say compaction?

Is this fragment-level conflict detection?  In other words, can I apply an overlay to fragment 5 at the same time I compact fragments 6 & 7?

protos/transaction.proto
```diff
@@ -315,6 +315,44 @@ message Transaction {
...
+    // touched are read from each overlay's `data_file.fields`.
+    repeated DataOverlayFile overlays = 2;
+  }
+
+  // Attach overlay files to fragments, supplying new values for a subset of
+  // (row offset, field) cells without rewriting the fragments' base data files.
+  // See the DataOverlayFile message in table.proto and the Data Overlay Files
+  // specification for resolution, coverage, and versioning rules.
+  //
+  // Conflict semantics (intentionally permissive, like DataReplacement). Against
```
westonpace: Conflict semantics should probably be in a spec doc somewhere (transaction.md) and not here

westonpace (changes requested):
Looks like a cool feature.  I have some questions, some of which I think are probably blocking (but maybe I'm misunderstanding) but only because they need to be clarified.  I don't see anything insurmountable.

protos/table.proto
```diff
@@ -115,6 +115,9 @@ message Manifest {
   // * 1 << 3: table config is present
   // * 1 << 4: dataset uses multiple base paths
   // * 1 << 5: transaction file writes are disabled
+  // * 1 << 6: data overlay files are present (see DataOverlayFile). Readers that do
+  //   not understand overlays must refuse the dataset, since ignoring an overlay
+  //   would silently return stale base values.
```
westonpace, line 120: Should we also add something to `feature_flags.rs`?

Alternatively, do we need a feature flag?  Right now we are using the presence of overlays as the discriminator.  Doesn't that work?

docs/src/format/table/data_overlay_file.md
wjones127:
```suggestion
for sparse overlays, one bitmap per field). The bitmaps index **physical row
offsets**. They include deleted rows and are stable even as deletion vectors change.
```

docs/src/format/table/data_overlay_file.md
wjones127: It's possible, but not encouraged. This is mostly just trying to be exhaustive in the spec.

docs/src/format/table/data_overlay_file.md
wjones127: Honestly, I'm tempted to remove the multiple overlays suggestion. I think we generally prefer fewer files to keep metadata overhead low.

```suggestion
- **Sparse.** A `FieldCoverage` carries one bitmap per field, used when different
  fields cover different offset sets — for example a `MERGE` with multiple
  `WHEN MATCHED` branches, where different rows update different columns. A dense
  overlay would have to widen to the bounding rectangle and fill the untouched
  cells with their current values (post-images), which for wide columns such as
  embeddings means re-storing data that did not change. A sparse overlay stores
  exactly the changed cells.

```

docs/src/format/table/data_overlay_file.md
wjones127: > If I try and apply an overlay on row X and another transaction updates row X is this a conflict?

That will conflict.

> Wouldn't I need the read version to detect this kind of conflict?

You don't need the read version of committed transactions serialized in their protobuf messages. You just need the read version of the transaction you are trying to commit. For example, Update transactions don't serialize the read version they were going off of, but you can still resolve Update-Update conflicts. The Update-Overlay conflicts can be resolved as you might expect—checking whether they affect the same rows.

docs/src/format/table/data_overlay_file.md
wjones127: Yeah, should do a take-style filtered read.

docs/src/format/table/data_overlay_file.md
wjones127: I agree. I like the idea of having a consolidated list of filters in the indexing section. I will add that.

docs/src/format/table/data_overlay_file.md
wjones127: Fair point. My benchmarking confirms that is a significant cost.

```suggestion
Overlays accumulate read cost — every overlay is a bitmap to test, a possible
file to open, and additional work to interleave values. Compaction bounds that cost in two modes:
```

docs/src/format/table/data_overlay_file.md
wjones127: That's a good point. Another cost we need to consider.

wjones127:
```suggestion
- **Overlay → overlay.** Merge several overlays into fewer, computing the
  post-image per `(offset, field)` by walking the merged overlays newest-first.
  The merged overlay takes the **maximum** `committed_version` of its inputs, so
  the exclusion semantics are preserved. Indexes can still be re-used, but they
  may now need to exclude more rows. This is cheap to write and does not touch
  the base.
```

docs/src/format/table/data_overlay_file.md
wjones127: 🤔 I didn't think too hard about this. But I guess we have two possible routes:

1. Allow horizontal compaction "in-place". If there's no deletions, we really could just write one consolidated data file in place. It can make the indexing story more complicated. I'm not sure we should encourage that, but I don't see a strong case for making it *illegal*.
2. Always create a new fragment when compacting. This is much easier to implement, and is the right first pass.

So, to answer your questions, the library will probably make a new fragment always and be the same as normal compaction. But there's a future optimization we might consider of in-place horizontal compaction.

docs/src/format/table/data_overlay_file.md
wjones127: Yeah, this is meant to be a full example of how overlays work across their lifecycle.


```suggestion
## Worked example

The following example illustrates how overlays function across their lifecycle, to make the rules above concrete.

```

docs/src/format/table/data_overlay_file.md
wjones127: Yeah, true that's not possible today. I hope one day it will be possible. For now I can make those separate steps.


```suggestion
Because the fold removed the overlay that was excluding offset 1 from the `age`
index, the commit must drop fragment `0` from its coverage so `age` queries
fall to the flat path.
```

docs/src/format/table/data_overlay_file.md
wjones127: I'm happy to drop this section.

docs/src/format/table/index.md
wjones127:
```suggestion
```

docs/src/format/table/index.md
wjones127:
```suggestion
```

protos/table.proto
wjones127: > Right now we are using the presence of overlays as the discriminator. Doesn't that work?

Not really. An older reader with old protobuf definitions will not be able to detect the extra overlay fields, right?

protos/transaction.proto
wjones127: It can't be this one, as this one appends overlays and doesn't remove any. We'll create another operation for that.

protos/transaction.proto
wjones127: I think this is talking what if I try to add a data overlay but a concurrent operation already made a data replacement or update. In those cases it is compatible.

But I think you are right the other direction needs care: if I"m trying to do a data replacement but another operation just did an overlay, I either need to drop/rewrite that overlay or raise a conflict error.

protos/transaction.proto
wjones127: > Is this just three different ways to say compaction?

Row-rewrite meant an update affecting the same row. But the other two can be taken as compaction.

I think the row-rewrite case should be row-level.

> can I apply an overlay to fragment 5 at the same time I compact fragments 6 & 7?

Yeah, I see no reason not to.

protos/transaction.proto
westonpace: I'm not actually worried about "just did an overlay".  The overlay could be days or months old.

This isn't a "conflict error and retry the operation" failure but a more lasting "there is no way to paste over an overlay" problem.

If we had one overlay file per field (which I'm not recommending) we could say "as part of a DataReplacement commit you should remove any overlay files for that field" but I don't know how this works when there is an overlay file that covers multiple fields (how to say that overlay file is only valid for a subset of fields)

protos/table.proto
westonpace: Ah, good point.  In that case I think you do need the feature flag.  Looks like you added it here (https://github.com/lance-format/lance/commit/3b20e96b0fe5682d334324dde422fa61b55ba9e2) :+1:

docs/src/format/table/data_overlay_file.md
westonpace: Ok, I agree.  First we get new fragment path working.  Then we maybe get "same fragment but invalidate index".  I think "same fragment and rebuild index at the same time" is a very distant third in my mind.

protos/transaction.proto
wjones127: I see. This is actually something I addressed in https://github.com/lance-format/lance/pull/7535

> how to say that overlay file is only valid for a subset of fields

It's actually the same as when you need to invalidate a field in a datafile: you change the field id to -2 to mask it out. If all fields at -2 in the file, then you can drop the data file or overlay.

docs/src/format/table/data_overlay_file.md
wjones127: Yeah, that makes sense.

docs/src/format/table/data_overlay_file.md
wjones127: Yeah, I think I've left them to be inline for now and aimed for compaction to keep as few of them as possible. I'll try to do some more benchmarking to see if we can avoid any kind of spilling.

protos/transaction.proto
westonpace: Got it, and I see that in 7535.  I think that's probably worth mentioning in this specification doc.

westonpace (approved):
I don't have any further blocking questions

docs/src/format/table/data_overlay_file.md
```diff
@@ -0,0 +1,381 @@
...
+
+Overlays accumulate read cost — every overlay is a bitmap to test, a possible
+file to open, and additional work to interleave values. Compaction bounds that cost in two modes:
+
+- **Overlay → overlay.** Merge several overlays into fewer, computing the
+  post-image per `(offset, field)` by walking the merged overlays newest-first.
+  The merged overlay takes the **maximum** `committed_version` of its inputs, so
+  the exclusion semantics are preserved. Indexes can still be re-used, but they
+  may now need to exclude more rows. This is cheap to write and does not touch
+  the base.
```
wkalt: I think we can state that merged overlays must be contiguous in the version record?

docs/src/format/table/data_overlay_file.md
```diff
@@ -0,0 +1,381 @@
...
+(`dataset_version = 1`).
+
+### Step 1 — write an overlay
+
+```sql
+UPDATE users SET age = 26 WHERE id = 2;   -- Bob, offset 1
+```
+
+This touches one field (`age`) for one row, so the writer emits a dense overlay
+and commits it as version 2. Fragment `0` gains:
```
wkalt: nitpick: I think the single-cell case makes the choice of sparse vs dense a little arbitrary. We might be able to strengthen this example by updating more than one row/col.

wkalt (approved):
@wjones127 thanks for pushing this through! Very excited for this feature. Just two minor comments.

PR #7383, opened 2026-06-20
feat(python): add Pydantic model instance auto-conversion and from_pydantic_model() classmethod

### LINKED ISSUE #1106: working with pydantic model instances
1. Currently we still require an explicit call to pydantic_to_schema when creating a table. Instead, we can create a `LanceTable.from_pydantic_model()` method that infers both the name and the schema. For the name we'll convert to snake_case automatically.

2. To add pydantic model instances, currently we require the user to explicitly convert to a dictionary. Instead, we could just automatically do it under the hood.

vinaysurtani:
## Summary

Closes #1106

This PR adds first-class Pydantic model support to Lance so users no longer need to manually call `.model_dump()` before writing data.

**Two changes:**
- `write_dataset()` now auto-converts a list of Pydantic `BaseModel` instances by calling `.model_dump()` on each item and inferring the PyArrow schema from the resulting dicts
- New `LanceDataset.from_pydantic_model(model_class, data, uri=None, **kwargs)` classmethod that infers the dataset URI from the model class name (snake_case) and delegates to `write_dataset()`

**Before (required manual conversion):**
```python
from pydantic import BaseModel
import pyarrow as pa, lance

class MyModel(BaseModel):
    name: str
    score: float

data = [MyModel(name="alice", score=0.9)]
schema = pa.schema([pa.field("name", pa.string()), pa.field("score", pa.float64())])
lance.write_dataset([m.model_dump() for m in data], "/tmp/test.lance", schema=schema)
```

**After (just works):**
```python
lance.write_dataset(data, "/tmp/test.lance")
# or
ds = lance.LanceDataset.from_pydantic_model(MyModel, data, uri="/tmp/test.lance")
```

## Files Changed

- `python/python/lance/dependencies.py` — added `_PYDANTIC_AVAILABLE` flag and `_check_for_pydantic()` helper, following the same optional-dependency pattern used for pandas and HuggingFace
- `python/python/lance/types.py` — added Pydantic branch in `_coerce_reader()` before the generic `Iterable` branch; handles both Pydantic v1 (`.dict()`) and v2 (`.model_dump()`)
- `python/python/lance/dataset.py` — added `from_pydantic_model()` classmethod to `LanceDataset`
- `python/python/tests/test_dataset.py` — added Pydantic instances to the existing `test_input_data` parametrized suite and added `test_from_pydantic_model`

## Test Results

```
python/python/tests/test_dataset.py: 196 passed, 1 skipped, 4 failed
```

The 4 failures (`test_random_dataset_recall_accelerated`, `test_random_dataset_recall_accelerated_one_pass`, `test_count_index_rows_accelerated`, `test_count_index_rows_accelerated_one_pass`) all fail with `PermissionError: [Errno 13] Permission denied: 'nvcc'` — these require the CUDA compiler toolkit and are pre-existing failures unrelated to this change. They do not touch any code modified in this PR.

New tests added:
- `test_input_data[NoneType-pydantic]` — Pydantic instances auto-convert correctly in `write_dataset()`
- `test_from_pydantic_model` — classmethod writes data, infers schema, round-trips correctly

## Checklist

- [x] Pydantic instances can be passed directly to `write_dataset()` without manual `.model_dump()`
- [x] `LanceDataset.from_pydantic_model()` classmethod exists and works
- [x] Pydantic is treated as an optional dependency (no unconditional import)
- [x] Both Pydantic v1 and v2 are supported
- [x] All existing `test_input_data` parametrized cases still pass
- [x] New tests cover both new entry points
- [x] Rebased cleanly on upstream/main with no conflicts


prrao87:
Should we be supporting Pydantic v1 any more? It's deprecated (and even no longer supported) in many other well known frameworks like FastAPI.

EDIT: This question might be larger than the scope of this PR, but might be worth creating a new issue for it.

python/python/lance/types.py
```diff
@@ -116,6 +120,20 @@ def batch_iter():
...
         batch = pa.RecordBatch.from_pylist(data_obj, schema=schema)
         return pa.RecordBatchReader.from_batches(batch.schema, [batch])
+    elif (
+        isinstance(data_obj, list)
+        and len(data_obj) > 0
+        and _check_for_pydantic(data_obj[0])
+    ):
+        from pydantic import BaseModel
+
+        if isinstance(data_obj[0], BaseModel):
```
westonpace: We should probably have an else branch here.  The thing seems pydantic-ish (it passed `_check_for_pydantic` but it isn't a `BaseModel`.  Above, in the HF branch, we return a `Unknown HuggingFace dataset type` error in that case.  We should do something similar here.

However, I don't love this pattern for HF either 😛.

Maybe an even better approach would be to define an `is_pydantic_base_model` helper function which just does the import & isinstance check.  Then you can add that to the `elif` condition list.  That way, if it isn't a `BaseModel`, we just continue on the chain.  This is probably safest since `_check_for_pydantic` can return false positives.

python/python/lance/dataset.py
```diff
@@ -793,6 +793,45 @@ def __deserialize__(
...
+        model_class,
+        data,
+        uri: Optional[Union[str, Path]] = None,
+        mode: str = "create",
+        **kwargs,
+    ) -> "LanceDataset":
+        """Create a LanceDataset from a Pydantic model class and a list of instances.
+
+        The table name is inferred from the model class name converted to snake_case.
+        The schema is inferred from the data.
```
westonpace: Should we use `pydantic_to_schema` to infer the schema from the model instead of inferring it from the data?

For example, the model might have a field marked optional (nullable) but if the data for this particular batch doesn't have any nulls then `from_pylist` might give us a non-nullable field.

python/python/lance/dataset.py
```diff
@@ -793,6 +793,45 @@ def __deserialize__(
...
+
+        Parameters
+        ----------
+        model_class : type
+            A Pydantic BaseModel subclass.
+        data : list
+            A list of Pydantic model instances.
```
westonpace, line 874: We should return an error if `data` is an empty list.

python/python/lance/dataset.py
```diff
@@ -793,6 +793,45 @@ def __deserialize__(
...
+            The write mode. One of "create", "overwrite", or "append".
+        **kwargs
+            Additional arguments passed to write_dataset().
+        """
+        import re
+
+        if uri is None:
+            uri = re.sub(r"(?<!^)(?=[A-Z])", "_", model_class.__name__).lower()
+        dicts = [
+            item.model_dump() if hasattr(item, "model_dump") else item.dict()
```
westonpace: Why prefer `model_dump` over `dict`?  We have other places we convert pydantic to arrow (e.g. in the `_into_pyarrow_reader` function) and it would be good to be consistent.

python/python/tests/test_dataset.py
```diff
@@ -34,9 +34,17 @@
 from lance.file import stable_version
 from lance.schema import LanceSchema
 from lance.util import validate_vector_index
+from pydantic import BaseModel
```
westonpace: Should we use `pytest.importorskip("pydantic")`?  I don't know if we have other tests that assume `pydantic` or not.

python/python/lance/dataset.py
```diff
@@ -793,6 +793,45 @@ def __deserialize__(
...
+            A list of Pydantic model instances.
+        uri : str or Path, optional
+            The URI to write the dataset to. If not provided, the model class name
+            converted to snake_case is used as the path.
+        mode : str, optional
+            The write mode. One of "create", "overwrite", or "append".
+        **kwargs
+            Additional arguments passed to write_dataset().
+        """
+        import re
```
westonpace: I think we'd prefer this be added to top-level imports

.gitignore
```diff
@@ -125,3 +125,4 @@ docs/src/community/project-specific/namespace-impls.md
 docs/src/community/project-specific/ray.md
 docs/src/community/project-specific/spark.md
 docs/src/community/project-specific/trino.md
+reproduce.py
```
westonpace: Revert this change?

westonpace (changes requested):
The `_check_for_pydantic` and `_PYDANTIC_AVAILABLE` helpers are very welcome 😄.

I'm not entirely sold on the API however.  Your before code has:

```
data = [MyModel(name="alice", score=0.9)]
schema = pa.schema([pa.field("name", pa.string()), pa.field("score", pa.float64())])
lance.write_dataset([m.model_dump() for m in data], "/tmp/test.lance", schema=schema)
```

Wouldn't it work to do...

```
data = [MyModel(name="alice", score=0.9)]
ds = lance.write_dataset(data, "/tmp/test.lance")
```

In other words, don't we already automatically handle ingesting a list of pydantic model instances?

.gitignore
vinaysurtani: done.

python/python/lance/dataset.py
vinaysurtani: noted, moved the import to top-level

python/python/tests/test_dataset.py
vinaysurtani: Done, made the change as you suggested above. Also confirmed pydantic isn't actually declared anywhere in pyproject.toml's test deps

python/python/lance/dataset.py
vinaysurtani: I wanted to help support both Pydantic versions without creating any issues. I peeked at the function you mentioned (in lancedb repo) and matched its naming convention.
One difference from lancedb's version: I kept the per-call hasattr(obj, "model_dump") check rather than resolving v1-vs-v2 once via pydantic.__version__ at import time, since pydantic is an optional/lazy dependency in this repo (_PYDANTIC_AVAILABLE) rather than a hard one

python/python/lance/dataset.py
vinaysurtani: Added, it now raises a ValueError if data is empty.

python/python/lance/dataset.py
vinaysurtani: Got it, I fixed that as well. 
from_pydantic_model now derives the schema from model_class's own field annotations (via model_fields/__fields__) instead of guessing from the row data, so an Optional field that happens to be None across an entire batch is still typed correctly instead of collapsing to Arrow's null type.

However, I could not find pydantic_to_schema in this repo. I assume it could be in the lancedb repo I couldn't find an equivalent here, so I wrote a small local _pydantic_model_to_schema() helper function instead. Do let me know what would be the best action according to you

python/python/lance/types.py
vinaysurtani: Good catch, I went ahead with your second suggestion. 
Replaced _check_for_pydantic with a new _is_pydantic_base_model() helper in dependencies.py that does the lazy import and a real isinstance(obj, BaseModel) check in one step, then used that directly as the elif condition in _coerce_reader. 
So the nested if/local import is gone, and if the object isn't actually a BaseModel, the elif is just False and it falls through to the next branch in the chain - no dead branch, no false positives possible anymore.

westonpace (commented):
Ok, I had not realized in my original review that this was a PR for `lance-format/lance` and not for `lancedb/lancedb`.  That explains why the feature is needed (there is no existing pydantic handling in pylance).

However, it raises a new concern.  We have a fairly extensive pydantic-to-arrow conversion in lancedb (https://github.com/lancedb/lancedb/blob/f8dc2f78ee3219084d457a647e230b23e2f391b0/python/python/lancedb/pydantic.py#L415).  It does more than just check for optional types (I had mentioned that as an example), it also has custom pydantic types for Vector and MultiVector and has custom pydantic type to arrow type mappings.

This puts us in a bit of a quandry here:

* We could proceed as-is with this PR but would now have two different dialects of pydantic-to-arrow conversion (pylance and lancedb)
* We could move the pydantic-to-arrow code from lancedb into pylance (makes it a bit harder to change that code but it hasn't changed much in its lifetime)
  * We would still need the embedding function handling to stay in lancedb (lance has no concept of embedding functions) which means lancedb would still need its own pydantic_to_schema but it would proxy to this one.  It's a bit awkward but can work.
* We could just leave pydantic support as a lancedb-only feature (but that defeats the goal of this PR)

Can you take a look at the pydantic code in lancedb and let me know your opinion on migrating it into this repository?  Also, if we are going to do so, we might want to think about @prrao87 's question of whether we should deprecate pydantic v1 support at the same time.

vinaysurtani:
Went through lancedb/pydantic.py. The type-conversion logic - Vector, MultiVector, struct/nested-model, Enum, tz-aware datetime, pydantic_to_schema, model_to_dict - has no dependency on embedding functions at all, so it could move into pylance close to as-is.
The only piece that doesn't transplant is the embedding-function parsing in LanceModel.to_arrow_schema() (parse_embedding_functions, EmbeddingFunctionRegistry) - that's genuinely lancedb-only, since lance has no concept of embedding functions.

So I'd go with option 2: move the pure conversion code into pylance, lancedb proxies to it and keeps embedding-function handling on top of that.

That's a two-repo change though - landing it for real means adding the code here, then a separate PR in lancedb/lancedb to delete their copy and proxy to this one, plus sorting out the version coupling (lancedb pinning a minimum lance version) so it doesn't break existing users mid-transition. Bigger than this PR should carry. I'd suggest splitting it: I'll open a separate issue to track the migration, and keep this PR scoped to the initial conversion (scalars + Optional + List, which is what's already implemented here). Vector/MultiVector/struct/Enum support can follow once the migration lands.

On pydantic v1: I'd keep that separate too. lancedb still actively branches on PYDANTIC_VERSION.major < 2 throughout, so v1 looks like a live target there - deprecating it should get its own justification, not ride along with this.

vinaysurtani:
@westonpace I put together a rough branch showing what moving the conversion logic would look like, in case it's easier to react to than describing it in the abstract:
https://github.com/vinaysurtani/lance/compare/fix-issue-1106...pydantic-arrow-migration

It's the 3 commits on top of this PR: new lance/pydantic.py module, wiring it into from_pydantic_model(), and tests. Not proposing this as final/ready to merge, just want to know if this is the shape you had in mind before I take it further (and before opening the tracking issue). One thing worth flagging: it tightens nullability inference to match lancedb exactly, a field with a plain default but non-Optional type is no longer inferred as nullable, which is a small behavior change from what's already in this PR.

Do let me know what you think.

vinaysurtani:
hey @westonpace , please let me know once you have reviewed the suggested changes, and if this approach would be better compared to the previous approach. thank you for your time!

westonpace:
> One thing worth flagging: it tightens nullability inference to match lancedb exactly, a field with a plain default but non-Optional type is no longer inferred as nullable, which is a small behavior change from what's already in this PR.

This is fine (and for the best to match lancedb)

> hey @westonpace , please let me know once you have reviewed the suggested changes, and if this approach would be better compared to the previous approach. thank you for your time!

Yes, I think the approach you linked would be best.  Did you merge that approach into this PR?

westonpace:
By the way, if you hit re-request review when you are ready for more input then it will put this PR back into my queue:

image

vinaysurtani:
@westonpace oh i wasnt aware of that. I merged those changes here. Should I implement the suggested changes from CodeRabbit? Or is this a good point for it to go ahead?

vinaysurtani:
I made Coderabbit's suggested changes, only the dependencies file conflict might need a check, otherwise i think this is almost ready?

westonpace (approved):
Thanks.  I've rebased and resolved the dependencies conflict and will merge on green.

PR #7385, opened 2026-06-22
fix(fts): enforce required terms for and queries

BubbleCal:
### What is the bug?

FTS `AND` queries could return matches from a partition that only contained a subset of the required query terms. For fuzzy `AND`, expansions were also flattened without preserving the original query-position grouping, so missing required positions and same-position expansion scoring could produce incorrect results.

A query such as `alpha AND beta` could return rows from a partition that only had `alpha` because the missing term was skipped before WAND saw the query. Fuzzy `AND` could also treat expansions from one original position as separate required terms, or score grouped expansions using the wrong token IDF, which could affect top-k ordering.

This PR makes partition posting-list loading aware of the query operator. For `AND`, a partition now returns empty results when any required original position has no exact term or fuzzy expansion. For fuzzy `AND`, expansions are grouped by original query position, same-position expansions are unioned for candidate selection, and final scoring uses the actual matched expansion token frequencies.

## Validation

- `cargo fmt --all --check`
- `git diff --check`
- `CARGO_TARGET_DIR=... cargo test -p lance-index test_fuzzy_and_scores_grouped_expansions_by_matched_token -- --nocapture`
- `CARGO_TARGET_DIR=... cargo test -p lance-index test_and_query -- --nocapture`
- `CARGO_TARGET_DIR=... cargo test -p lance-index test_fuzzy_and_groups_expansions_by_original_position -- --nocapture`
- `CARGO_TARGET_DIR=... cargo test -p lance-index bm25_search -- --nocapture`
- `CARGO_TARGET_DIR=... cargo test -p lance-index scalar::inverted::wand::tests -- --nocapture`

### DIFF
--- a/rust/lance-index/src/scalar/inverted/index.rs
+++ b/rust/lance-index/src/scalar/inverted/index.rs
@@ -9,7 +9,7 @@ use std::{
     collections::BinaryHeap,
 };
 use std::{
-    collections::{HashMap, HashSet},
+    collections::{BTreeMap, HashMap, HashSet},
     ops::Range,
     time::Instant,
 };
@@ -207,6 +207,7 @@ impl FromStr for InvertedListFormatVersion {
 #[derive(Debug)]
 struct PartitionCandidates {
     tokens_by_position: Vec<String>,
+    grouped_expansions: Vec<GroupedExpansionTerms>,
     candidates: Vec<DocCandidate>,
 }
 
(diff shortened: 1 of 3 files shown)


rust/lance-index/src/scalar/inverted/index.rs
```diff
@@ -741,22 +800,37 @@ impl InvertedIndex {
                     let mask = mask.clone();
                     let metrics = metrics.clone();
                     let part_for_wand = part.clone();
-                    let mut partition_result = spawn_cpu(move || {
+                    let has_grouped_expansions = !grouped_expansions.is_empty();
+                    let wand_params = if has_grouped_expansions {
```
Xuanwo, line 819: With grouped fuzzy AND, this clears the requested limit before WAND runs. Since WAND treats `None` as `usize::MAX`, a `limit=1` query can still materialize every matching candidate in each partition and resolve all deferred row ids before the outer heap trims results, which removes the resource bound users expect from top-k search.

rust/lance-index/src/scalar/inverted/index.rs
```diff
@@ -1307,39 +1414,146 @@ impl InvertedPartition {
 
             let base_len = tokens.token_type().prefix_len(token) as u32;
             if let TokenMap::Fst(ref map) = self.tokens.tokens {
+                let mut expanded = Vec::new();
```
Xuanwo, line 1436: This makes `max_expansions` apply separately to each query token. The previous implementation accumulated expansions in one vector, so the same cap applied to the whole fuzzy query; multi-token fuzzy queries can now expand to `tokens.len() * max_expansions` terms, changing recall, scoring, and posting IO for existing queries.

Xuanwo (changes requested):
Two issues need changes before merging. Please add regression coverage for multi-token `max_expansions` and grouped fuzzy AND top-k/resource-bound behavior.

rust/lance-index/src/scalar/inverted/index.rs
BubbleCal: Addressed in afef60743: grouped fuzzy AND no longer clears the WAND limit to `None`; it now uses bounded oversampling based on grouped expansion terms while keeping final matched-expansion rescoring. Added regression coverage in `test_fuzzy_and_grouped_rescore_keeps_wand_limit_bounded`.

rust/lance-index/src/scalar/inverted/index.rs
BubbleCal: Addressed in afef60743: fuzzy expansion now applies `max_expansions` across the whole query again while preserving original token positions for grouped fuzzy AND. Added regression coverage in `test_fuzzy_expansion_cap_applies_to_whole_query`.

Xuanwo (approved):
LGTM, thank you!

PR #7390, opened 2026-06-22
perf(index): use NGram posting cardinalities for regex conjunction search

everySympathy:
## Summary

This PR improves regex conjunction search on the NGram index by making posting-list evaluation more cost-aware.

The existing regex NGram acceleration can derive required trigrams from regex patterns, but conjunction queries may still load and intersect posting lists without considering posting-list size. This PR adds posting-list cardinality metadata and uses rare-first ordering to reduce unnecessary work, especially when a required trigram is missing or when a regex contains many common required trigrams.

## Changes

| Area | Change |
| --- | --- |
| Index format | Adds a `cardinality` column to NGram postings, storing the row-count of each trigram posting list. |
| Query planning | Sorts required regex conjunction trigrams by cardinality before lookup. |
| Missing token handling | Short-circuits pure conjunction regex queries immediately when any required trigram is absent. |
| Bitmap intersection | Sorts loaded posting lists by actual bitmap cardinality before intersection, preserving rare-first CPU behavior after unordered async reads. |
| Compatibility | Keeps older two-column NGram postings readable when cardinality metadata is absent. |
| Remap/update | Recomputes cardinality when remapping posting lists and writes the destination in the current format. |
| Benchmarking | Extends `regex_ngram` with skewed conjunction cases for cardinality-aware posting-list planning and allows scaling row count with `LANCE_REGEX_NGRAM_TOTAL`. |

## Why

For a regex conjunction, every required trigram must be present. If one required trigram is absent, the candidate set is empty and there is no need to load/intersect other posting lists.

When all required trigrams are present, intersecting smaller posting lists first avoids cloning and intersecting large Roaring bitmaps early. This is the same selectivity principle used by inverted-index query planning: rare terms are more valuable filters than common terms.

## Benchmark Results

Measured with the final `rust/lance/benches/regex_ngram.rs` benchmark data: `HEAD^` plus the benchmark-only patch is compared against this PR.

Both runs use the same negative conjunction query:

```text
commonmarkerabcdefghijklmnopqrstuvwx.*missingmarker
```

Default 200k-row run:

```bash
cargo bench -p lance --bench regex_ngram -- many_common_missing_trigram
```

10M-row scale run:

```bash
LANCE_REGEX_NGRAM_TOTAL=10000000 cargo bench -p lance --bench regex_ngram -- many_common_missing_trigram
```

| Dataset | Baseline | This PR | Change |
| --- | ---: | ---: | ---: |
| 200k rows | `[783.43 us, 803.03 us, 829.14 us]` | `[417.02 us, 435.05 us, 454.90 us]` | ~46% faster |
| 10M rows | `[10.734 ms, 10.779 ms, 10.843 ms]` | `[465.45 us, 471.39 us, 482.43 us]` | ~22.9x faster |

The biggest improvement is in negative conjunction cases where a regex contains many common required trigrams plus a missing required trigram. The new path can return an empty candidate set before loading large common posting lists.

## Compatibility

| Index format | Behavior |
| --- | --- |
| New three-column postings: `tokens`, `cardinality`, `posting_list` | Uses cardinality metadata for regex conjunction planning. |
| Old two-column postings: `tokens`, `posting_list` | Still readable; `token_cardinalities` is `None`, and the index falls back safely. |

The old-format path checks the index file schema before choosing the projection. This avoids an intentional failed read on old-format postings and keeps compatibility covered in the NGram tests.

## Testing

- `cargo fmt --all`
- `cargo test -p lance-index ngram`
- `cargo test -p lance --bench regex_ngram`
- `cargo bench -p lance --bench regex_ngram -- many_common_missing_trigram`
- `LANCE_REGEX_NGRAM_TOTAL=10000000 cargo bench -p lance --bench regex_ngram -- many_common_missing_trigram`

### DIFF
--- a/rust/lance-index/src/scalar/ngram.rs
+++ b/rust/lance-index/src/scalar/ngram.rs
@@ -28,7 +28,7 @@ use crate::vector::VectorIndex;
 use crate::{Index, IndexType};
 use arrow::array::{AsArray, UInt32Builder};
 use arrow::datatypes::{UInt32Type, UInt64Type};
-use arrow_array::{BinaryArray, RecordBatch, UInt32Array};
+use arrow_array::{BinaryArray, RecordBatch, UInt32Array, UInt64Array};
 use arrow_schema::{DataType, Field, Schema, SchemaRef};
 use async_trait::async_trait;
 use datafusion::execution::SendableRecordBatchStream;
@@ -56,6 +56,7 @@ mod ngram_regex;
 pub(crate) use ngram_regex::regex_can_use_index;
 
 const TOKENS_COL: &str = "tokens";
+const CARDINALITY_COL: &str = "cardinality";
 const POSTING_LIST_COL: &str = "posting_list";
 const POSTINGS_FILENAME: &str = "ngram_postings.lance";
 const NGRAM_INDEX_VERSION: u32 = 0;
@@ -64,11 +65,14 @@ use std::sync::LazyLock;
 
 pub static TOKENS_FIELD: LazyLock<Field> =
     LazyLock::new(|| Field::new(TOKENS_COL, DataType::UInt32, true));
+pub static CARDINALITY_FIELD: LazyLock<Field> =
+    LazyLock::new(|| Field::new(CARDINALITY_COL, DataType::UInt64, false));
 pub static POSTINGS_FIELD: LazyLock<Field> =
     LazyLock::new(|| Field::new(POSTING_LIST_COL, DataType::Binary, false));
 pub static POSTINGS_SCHEMA: LazyLock<SchemaRef> = LazyLock::new(|| {
     Arc::new(Schema::new(vec![
         TOKENS_FIELD.clone(),
+        CARDINALITY_FIELD.clone(),
         POSTINGS_FIELD.clone(),
     ]))
 });
@@ -269,6 +273,9 @@ impl NGramPostingListReader {
 pub struct NGramIndex {
     /// The mapping from tokens to row offsets
     tokens: HashMap<u32, u32>,
+    /// The number of rows in each token's posting list. Older NGram indices do
+    /// not have this metadata and fall back to loading every required token.
+    token_cardinalities: Option<HashMap<u32, u64>>,
     /// The reader for the posting lists
     list_reader: Arc<NGramPostingListReader>,
     /// The tokenizer used to tokenize text.  Note: not all tokenizers can be used with this index.  For
@@ -293,6 +300,7 @@ impl std::fmt::Debug for NGramIndex {
 impl DeepSizeOf for NGramIndex {
     fn deep_size_of_children(&self, context: &mut lance_core::deepsize::Context) -> usize {
         self.tokens.deep_size_of_children(context)
+            + self.token_cardinalities.deep_size_of_children(context)
     }
 }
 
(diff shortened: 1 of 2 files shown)


rust/lance-index/src/scalar/ngram.rs
```diff
@@ -342,7 +383,10 @@ impl NGramIndex {
             .expect_ok()?
             .as_binary::<i32>();
 
-        let new_posting_lists = posting_lists_array
+        // The input batch may come from either the old two-column format or the
```
majin1102, line 386: I think we should expand what is actually `two-column format` and `current three-column ` here.  Or we could explain why we introduce `cardinalities` here.  People reading here might not know the two-column context.

What do you think?

rust/lance-index/src/scalar/ngram.rs
```diff
@@ -1680,6 +1830,119 @@ mod tests {
...
+            RecordBatch::try_new(schema.clone(), vec![Arc::new(data), Arc::new(row_ids)]).unwrap();
+        let data = Box::pin(RecordBatchStreamAdapter::new(
+            schema,
+            stream::once(std::future::ready(Ok(data))),
+        ));
+
+        let builder = NGramIndexBuilder::try_new(NGramIndexBuilderOptions::default()).unwrap();
+        let (index, _tmpdir) = do_train(builder, data).await;
+
+        assert!(index.token_cardinalities.is_some());
```
majin1102, line 1851: This verifies that the cardinality column is persisted correctly, but the query planner uses the in-memory `index.token_cardinalities` map rather than reading the column again. Could we also assert the loaded map contents here?

For example:

```rust
let cardinalities = index.token_cardinalities.as_ref().unwrap();
assert_eq!(
    cardinalities.get(&ngram_to_token("cat", NGRAM_N)),
    Some(&2)
);
assert_eq!(
    cardinalities.get(&ngram_to_token("dog", NGRAM_N)),
    Some(&2)
);

File: rust/lance-index/src/scalar/ngram.rs [hunk 3]
```diff
@@ -1680,6 +1830,119 @@ mod tests {
...
+        assert_eq!(read_cardinality_for_trigram(&index, "cat").await, 2);
+        assert_eq!(read_cardinality_for_trigram(&index, "dog").await, 2);
+    }
+
+    #[test_log::test(tokio::test)]
+    async fn test_ngram_regex_sorts_large_and_by_cardinality() {
```
majin1102, line 1857: This test does not really verify cardinality-based ordering. The result and `comparisons > 8` would still pass if `sorted_tokens_for_conjunction` ignored `token_cardinalities` and simply loaded all required posting lists.

Could we add a more direct assertion for the planner behavior, e.g. checking that a missing / rare trigram is ordered before a common one, or an end-to-end missing-token case with `metrics.parts_loaded == 0`?

majin1102 (commented):
Overall looks good to me. Left some nit comments.

This changes the on-disk NGram posting format by adding a `cardinality` column to `ngram_postings.lance`, but the format docs still describe the old two-column layout.

Could you update the NGram index format docs, especially `docs/src/format/index/scalar/ngram.md`, to document the new column and the compatibility behavior for older two-column indices?

Also, my understanding is that Lance format specification changes should be proposed in GitHub Discussions and go through the community vote process. Could you start / link the corresponding discussion vote from this PR?

PR #7391, opened 2026-06-22
ci: add Java 25 to the JNI build-and-test matrix

LuciferYang:
## Summary

Adds **Java 25** (the September 2025 LTS) to the `build-and-test-java` matrix in `.github/workflows/java.yml`, so the JNI bindings are built and tested on the current LTS alongside 11/17/21. The build targets `maven.compiler.release` 11, so the newer JDK only affects the build/test runtime, not the emitted bytecode.

With spotless excluded on Java 25 (see below), `mvn install` — the full build plus all JNI tests — passes cleanly on JDK 25, so the bindings themselves are JDK-25-ready.

## Why spotless is skipped on Java 25

The `spotless` / `google-java-format` format gate cannot run on JDK 25, and bumping the formatter is not a viable alternative while 11/17 stay in the matrix:

- `google-java-format` 1.22.0 (pinned via spotless) calls `com.sun.tools.javac.util.Log$DeferredDiagnosticHandler.getDiagnostics()`, whose signature changed in JDK 25 → `NoSuchMethodError`.
- The first gjf release with JDK 25 support is **1.34.0**, but gjf **≥ 1.29 requires JDK 21+ to run** — so it would fail the Java 11 and 17 jobs with `UnsupportedClassVersionError`. No gjf release runs on 11/17 **and** supports 25. (A trial bump 1.22 → 1.27 also reformatted 0 of 190 files, i.e. no formatting drift to gain.)

Since the format check is JDK-independent and stays enforced on 11/17/21, Java 25 is excluded from it. spotless is also bound into the Maven build lifecycle (`validate` phase), so it is skipped in **two** places on the 25 job:

- the standalone style step — `if: matrix.java-version != '25'`
- the build — `mvn install -Dspotless.skip=true` (via a matrix-conditioned arg)

## Test plan

- [x] `Build and Test with Java 11 / 17 / 21 / 25` all green in CI.


LuciferYang:
cc @Xuanwo

wjones127 (approved):

PR #7394, opened 2026-06-22
refactor(index): introduce RowIdRemapper trait to decouple ScalarIndexPlugin from FragReuseIndex

westonpace:
Replaces the concrete `FragReuseIndex` parameter in `ScalarIndexPlugin::load_index` and `get_from_cache` with an abstract `RowIdRemapper` trait, removing the direct dependency of scalar index infrastructure on the table-level fragment-reuse type. `FragReuseIndex` implements the new trait and the call site in lance casts before passing through.

### DIFF
--- a/rust/lance-index/src/frag_reuse.rs
+++ b/rust/lance-index/src/frag_reuse.rs
@@ -10,14 +10,39 @@
 use std::any::Any;
 use std::sync::Arc;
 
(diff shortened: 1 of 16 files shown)


Xuanwo (approved):
LGTM, need to address the conflicts.

PR #7395, opened 2026-06-22
refactor(index): extract BasicTrainer trait from ScalarIndexPlugin

westonpace:
Move the two-stage training API (new_training_request + train_index) out of ScalarIndexPlugin into a new BasicTrainer trait. ScalarIndexPlugin gains a basic_trainer() accessor (default: None) so callers can opt into training without coupling the core plugin interface to it.


.claude/scheduled_tasks.lock
Xuanwo, line 1: This file should be removed. I think we can add it as gitignore maybe.

Xuanwo (commented):
Mostly LGTM, thank you for working on this!

Xuanwo (approved):
Thank you for this work!

PR #7397, opened 2026-06-22
refactor!: rename FMIndexIndexDetails to FMIndexDetails

westonpace:
I'm not sure if the original naming was intentional or not.  However, it required a special case in `get_plugin_name_from_details_name` to rename `fmindex` to `fm` so I'm guessing it was accidental?

We are trying to convert indexes to be "generic plugins" and this means we cannot have special cases lying around.

An index's short-name is the name of the type URL minus the suffix `IndexDetails`.  So if we want `fm` then it should be `FmIndexDetails` (which this PR implements).  If we want `fmindex` then it should be `FmIndexIndexDetails` (which it had before).  If we want both then we should invent some kind of formal alias mechanism where an index plugin can register potential aliases.  However, I think it'd be simplest to avoid that.

This change would be a breaking change to any existing FM indexes!  That index type has not (AFAIK) been formally released yet so I think this is ok.  However, if this misses the 8.0.0 release then we will probably need to find a different way (and possibly forever be locked into carrying around this special case).

### DIFF
--- a/protos/index.proto
+++ b/protos/index.proto
@@ -248,4 +248,4 @@ message BloomFilterIndexDetails {}
 
 message RTreeIndexDetails {}
 
-message FMIndexIndexDetails {}
+message FMIndexDetails {}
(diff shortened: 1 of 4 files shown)


westonpace:
CC @beinan

hamersaw (approved):

beinan (approved):

PR #7399, opened 2026-06-22
ci: enforce PMC vote on format-spec changes via CI gate

wjones127:
Implements the accepted parts of #7112: the pull request becomes the canonical venue for a Lance format-specification proposal, and the required PMC vote is enforced by a CI gate instead of by convention.

## Governance changes

- **Format votes happen on the PR, not a GitHub Discussion.** `voting.md` and `contributing.md` are updated to say so. Cast +1 by approving, −1 by requesting changes. Discussions remain the venue for release votes and general design proposals; migrating the existing format discussions is follow-up work.
- **Format PRs should be strictly the format change.** Both docs now ask for the spec change (protos + `docs/src/format/`) plus only the library edits needed to keep the build green, with the implementation in follow-up PRs. The same rule is added to `protos/AGENTS.md` and `docs/src/format/AGENTS.md` so coding agents follow it. Rationale in the docs: voters need to read the contract, and an ordinary code review shouldn't be dragged through a 72-hour voting period.
- **Voting period: 1 week → 72 hours, excluding weekends.** A proposal opened Friday afternoon still gets three working days of attention. Weekend boundaries are fixed in UTC — no DST to reason about, and no member's local zone decides when everyone's clock pauses — and the gate's comment states the exact closing time in both UTC and Pacific.
- **The clock starts when the PR is ready for review**, not at labeling, so time spent drafting doesn't burn the period. The gate stays silent on drafts: there is no open vote to comment on.

## How the gate works

The path labeler (`.github/labeler-area.yml`) applies `format-change` to PRs touching `protos/**/*.proto` or `docs/src/format/**`. The gate (`ci/format_vote_gate.py`, run by `format-vote-gate.yml`) reads that label and publishes a `format-spec-vote` commit status that stays red until **all** hold:

- **3 binding +1 votes** — three PMC members have approved, excluding the author, counted only on the **latest commit** (new pushes invalidate stale approvals).
- **No veto** — no PMC member has an outstanding "Request changes" review (a −1 veto cannot be overruled).
- **72 weekday hours elapsed** — measured from the later of (`format-change` first applied, PR marked ready for review), with Saturday and Sunday UTC not counting.

It posts/updates a single live-tally comment carrying the exact deadline, and runs on PR events (including `ready_for_review` / `converted_to_draft`) and a 15-minute `cron`. The cron has no PR context, so it sweeps every open `format-change` PR — this is what observes approvals and re-checks the clock when no PR event fires.

There is deliberately **no `pull_request_review` trigger**: a run triggered by a fork PR's review event gets a read-only `GITHUB_TOKEN` regardless of the workflow's `permissions` block, so it cannot post the status (403 `Resource not accessible by integration`). Format proposals from non-committers are exactly the fork-PR case, so reviews are picked up by the sweep instead. The tally therefore trails a review by up to 15 minutes, which is immaterial against a 72-hour period; a voter who doesn't want to wait can trigger the gate from its **Run workflow** button (`workflow_dispatch`, optional PR number — blank re-checks every open format PR), which the gate's own comment links to. It uses `pull_request_target` so it can status/comment fork PRs, but never checks out or runs PR code: it reads the trusted base checkout and the API only.

PRs **without** the label are left alone: the gate posts only a passing `format-spec-vote` status (so the required check never blocks unrelated PRs) and does nothing else. A PMC member waives a trivial edit (typo, wording, formatting) with the `format-waived` label.

## PMC roster

`docs/src/community/pmc.yaml` is the source of truth. The roster table in `pmc.md` is rendered at docs build time by an MkDocs hook (`docs/hooks/pmc_roster.py`), and the gate reads the roster from the base checkout so a PR can't enlarge the electorate by editing it.

## Tests

`ci/test_format_vote_gate.py` (pytest, run by `ci-scripts.yml`) covers the pure logic: vote counting (latest-review-wins, stale-approval filtering, author/non-PMC/dismissed exclusion, verdict priority) and the deadline arithmetic (weekend exclusion from either side of a boundary, multi-weekend spans, non-UTC start zones, and when the clock does and doesn't open).

## Activation (after merge)

1. Create the `format-waived` label (`format-change` already exists) so PMC members can pick it in the UI.
2. Add `format-spec-vote` as a **required status check** on protected branches (`main` + release branches). It's posted on every PR — success immediately for non-format PRs — so it never leaves a required check pending.

> [!NOTE]
> The gate does not run on this PR itself: `pull_request_target` and `schedule` both resolve the workflow from the base branch, which doesn't yet contain it. So `format-spec-vote` is absent here rather than green — don't make the check required until this is merged.


ci/sync_pmc_docs.py
wjones127, line 1: question: does this script need to be run manually? Or can we integrate with the docs framework to create the table on-build?

ci/format_vote_gate.js
wjones127, line 1: issue(blocking): Could we rewrite this script in Python? More of the maintainers are familiar with Python so it would be easy to read. I believe GHA has the GH python package pre-installed.

.github/workflows/format-vote-gate.yml
wjones127, line 1: issue(blocking): do we not filter just for PRs that have no `format-change` label? I think we should leave those PRs alone.

.github/workflows/format-vote-gate.yml
```diff
@@ -0,0 +1,64 @@
...
+on:
+  pull_request_target:
+    types: [opened, reopened, synchronize, labeled, unlabeled]
+  pull_request_review:
+    types: [submitted, edited, dismissed]
+  schedule:
+    # Re-evaluate open format-change PRs so the voting-period clock is re-checked
+    # even when no PR event fires (e.g. the 1-week period elapses days after the
+    # third approval lands).
+    - cron: "0 */8 * * *"
```
wjones127: issue(blocking): how does this trigger find which PRs to look at? IIRC this trigger isn't PR specific, but runs on a schedule globally for the whole repo, am I wrong?

ci/sync_pmc_docs.py
wjones127: Good call — switched to generating the table at docs build time via an MkDocs hook (`docs/hooks/pmc_roster.py`), following the same build-time pattern as the existing `mkdocs_protobuf` plugin. `pmc.yaml` stays the source of truth, `pmc.md` just holds a `<!-- PMC_ROSTER_TABLE -->` placeholder, and `ci/sync_pmc_docs.py` (and its docs-check step) are gone. No manual run needed.

ci/format_vote_gate.js
wjones127: Done — rewritten in Python as `ci/format_vote_gate.py` using PyGithub (pip-installed in the workflow). Pure vote-counting logic (`tally_reviews`, `decide_verdict`) is unit-tested with pytest in `ci/test_format_vote_gate.py`.

.github/workflows/format-vote-gate.yml
wjones127: Agreed. The gate now keys off the `format-change` label only. Labeling moved to the existing path labeler (`.github/labeler-area.yml`), and the gate leaves unlabeled PRs alone — for non-format PRs it posts just a passing `format-spec-vote` status and nothing else (needed so the required check never blocks them). Trivial edits are now waived by a PMC applying a `format-waived` label, rather than removing `format-change` (the path labeler would otherwise re-add it on the next push).

.github/workflows/format-vote-gate.yml
wjones127: You're right — the `schedule` trigger runs repo-wide, with no PR in context. The script handles that by listing all open PRs carrying the `format-change` label and re-evaluating each (so the voting-period clock advances even if no PR event fires). The PR/review event paths use the PR from the event payload. See the `event_name == 'schedule'` branch in `main()`.

ci/format_vote_gate.py
```diff
@@ -0,0 +1,371 @@
...
+    def evaluate(self, number):
+        pr = self.repo.get_pull(number)
+        if pr.state != "open":
+            print(f"PR #{number} is {pr.state}; skipping.")
+            return
+        head_sha = pr.head.sha
+        labels = {label.name for label in pr.labels}
+
+        # Non-format PRs get a passing status and are otherwise left alone.
+        if FORMAT_LABEL not in labels:
```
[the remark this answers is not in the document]
wjones127: This is an acceptable limitation.

ci/format_vote_gate.py
```diff
@@ -0,0 +1,371 @@
...
+
+
+class Gate:
+    def __init__(self, repo, pmc, run_url):
+        self.repo = repo
+        self.pmc = pmc
+        self.run_url = run_url
+
+    def is_pmc(self, login):
+        return login is not None and login.lower() in self.pmc
```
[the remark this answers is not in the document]
wjones127: I think you are taking the security here much too seriously. There's still human review, and attention paid to changes. We can always revert changes that should not have gone through.

ci/format_vote_gate.py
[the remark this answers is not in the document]
wjones127: It's intentional that maintainers can edit those labels. We trust them to exercise judgement. If they show they can't be trusted, the PMC can always remove them.

But also keep in mind, the format-change label isn't not at all perfect. It misses some things that might be format changes, and it captures some things that just aren't meaningful changes. Given that, we don't need to take the safety of this label too seriously. It will always require judgement.

ci/format_vote_gate.py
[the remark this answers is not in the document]
westonpace: I believe it is ok to waive this finding.

ci/test_format_vote_gate.py
```diff
@@ -0,0 +1,153 @@
+"""Unit tests for the format-spec vote gate logic.
```
westonpace, line 1: Does any CI job actually run these unit tests?  Is this something that needs to be checked in?

docs/src/community/pmc.yaml
```diff
@@ -0,0 +1,90 @@
+# Source of truth for the Project Management Committee (PMC) roster.
+#
+# This file drives two things, so keep it accurate:
+#   1. The roster table in `pmc.md`, rendered at docs build time by the
+#      `docs/hooks/pmc_roster.py` MkDocs hook.
+#   2. The format-specification vote gate, which only counts PR approvals from
+#      the `handle`s listed here (see `.github/workflows/format-vote-gate.yml`).
+#
+# Adding or removing a member is itself a PMC vote (roster change). After the
+# vote passes, edit this file; the docs table updates automatically.
```
westonpace, line 10: Heh, if PMC votes weren't private we could just add pmc.yaml to the format vote gate and make the changes to this file be the vote 😛

docs/src/community/voting.md
```diff
@@ -47,7 +51,62 @@ A **-1** binding vote is considered a veto for all decision types. Vetoes:
...
+This is not just a tidiness preference. The vote is on the format — a durable
+compatibility contract that outlives any one implementation — and PMC members
+should be able to read the whole of what they are voting on. A PR that also
+carries the reader, writer, and test changes buries the contract in
+implementation detail, and it drags an ordinary code review through a 72-hour
+voting period it does not need.
+
+Discussion happens as review comments on the PR, so reviewers can respond to
+specific lines of the specification. Open the PR as a draft while it is still
+taking shape; the voting period starts when you mark it ready for review.
```
westonpace, line 82: Users can still open general discussion about a potential change if they want to correct?

westonpace (approved):

westonpace:
FYI, it looks like the job is currently failing due to 403: https://github.com/lance-format/lance/actions/runs/33172657935/job/98853353865?pr=7399

ci/test_format_vote_gate.py
wjones127: I don't think they are run. I'm fine removing them.

docs/src/community/voting.md
wjones127: They can. I don't like discussions personally, but they are always free to create a discussion as well. Or a google doc or similar.

ci/test_format_vote_gate.py
wjones127: Actually, they do run in CI. There is a `ci-scripts.yml` workflow that they are part of. So let's keep them.

PR #7400, opened 2026-06-22
docs: clarify local LTO guidance

Xuanwo:
This clarifies the repository agent instructions for Rust local workflows.

Local development, debugging, and performance testing should avoid LTO so those workflows do not accidentally pay release-artifact build costs or benchmark a build mode that was not explicitly requested. LTO remains available when release artifact validation explicitly asks for it.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -11,6 +11,7 @@ lto = "thin"
 
 [profile.release-no-lto]
 inherits = "release"
+debug = true
 lto = false
 # Prioritize compile time when LTO is not relevant to the measurement.
 codegen-units = 16
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -9,6 +9,12 @@ debug = true
 codegen-units = 16
 lto = "thin"
 
+[profile.release-no-lto]
+inherits = "release"
+lto = false
+# Prioritize compile time when LTO is not relevant to the measurement.
+codegen-units = 16
+
 [profile.bench]
 inherits = "release"
 lto = "thin"
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -9,10 +9,13 @@ debug = true
 codegen-units = 16
 lto = "thin"
 
+# Fast optimized profile for local investigations where LTO does not affect the
+# bottleneck. This gives agents a named profile instead of changing LTO flags ad hoc.
 [profile.release-no-lto]
 inherits = "release"
 lto = false
-# Prioritize compile time when LTO is not relevant to the measurement.
+# More codegen units split each crate into more parallel LLVM codegen jobs, which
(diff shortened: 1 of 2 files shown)


AGENTS.md
```diff
@@ -41,6 +41,7 @@ Key technical traits: async-first (tokio), Arrow-native, versioned writes with m
 * Coverage: `cargo +nightly llvm-cov -q -p <crate> --branch`
 * Coverage HTML: `cargo +nightly llvm-cov -q -p <crate> --branch --html`
 * Coverage for file: `python ci/coverage.py -p <crate> -f <file_path>`
+* Do not enable LTO for local development, debugging, or performance testing. Keep LTO reserved for explicitly requested release artifact validation.
```
wjones127: question: why provide this guidance, instead of baking in profiles that specify LTO, and telling the agent to use those profiles? As-is, the agent has to figure out what profiles there are and modify them in order to comply with this guidance.

wjones127 (commented):
I agree local development and testing don't need LTO. But I'm skeptical that turning it off for profiling in general is a good idea; it makes a meaningful difference in performance, which is the exact thing we are trying to measure in those builds. It would be like turning off SIMD during benchmarking—we'd be measuring the wrong thing.

Now, there are likely some benchmarks that aren't sensitive to LTO: if you are benchmarking IO-bound code, it's probably not necessary. But if you are doing CPU-bound benchmarks, such as distance function work, I would say keep the LTO on. Perhaps we can provide a guideline like that in the AGENTS.md?

AGENTS.md
Xuanwo: provide a dedicated profile seems to be a good idea. Will give a try.

.cargo/config.toml
```diff
@@ -9,6 +9,12 @@ debug = true
 codegen-units = 16
 lto = "thin"
 
+[profile.release-no-lto]
```
wjones127, line 12: Suggestion: if this is for debugging, why not enable debug symbols here too? I imagine we’ll want to profile here as well.

AGENTS.md
```diff
@@ -41,6 +41,10 @@ Key technical traits: async-first (tokio), Arrow-native, versioned writes with m
 * Coverage: `cargo +nightly llvm-cov -q -p <crate> --branch`
 * Coverage HTML: `cargo +nightly llvm-cov -q -p <crate> --branch --html`
 * Coverage for file: `python ci/coverage.py -p <crate> -f <file_path>`
+* Use repository-defined Cargo profiles instead of ad hoc LTO overrides.
+* Use `release` or `bench` for CPU-bound benchmarks and performance numbers intended for comparison.
```
wjones127: Suggestion: I would just say always use release with debug for benchmarks. It’s always nice to be able to profile without having to rebuild.

wjones127 (approved):

PR #7406, opened 2026-06-23
feat(file): v2 writer/reader support columns of unequal length

### LINKED ISSUE #7404: V2 File writers and readers can support columns of varying lengths
**Context**

Sparse data overlays (one file carrying a different set of rows per field) require the v2 file writer to emit columns whose item counts differ *within a single file*. The current v2 writer advances all columns from one global row counter and cannot do this. It is technically allowed by the file format but unused today. This is the foundation for [Data Overlay Files](<https://github.com/lance-format/lance/pull/7381>) (see "Physical layout" and "Writer support" in the spec).

**Acceptance criteria**

- [ ] The v2 file writer can write a single file in which columns have differing item counts (no shared global row counter).
- [ ] The v2 file reader can read each column back independently at its own length.
- [ ] Random access (by rank/position within a column) returns the correct value for a column shorter than the file's longest column.
- [ ] Per-column row counts are recorded in / derivable from file metadata.
- [ ] Round-trip tests: write columns of unequal length (including a zero-length column and a single-row column alongside longer ones); read each back fully and via random access; assert values and lengths.
- [ ] Existing equal-length files round-trip unchanged (backwards compatible).

**Out of scope**

* The `DataOverlayFile` concept, coverage bitmaps, and rank-vs-offset resolution semantics — this issue delivers only the file-level capability; overlay write/read is [OSS-1322](https://linear.app/lancedb/issue/OSS-1322/write-data-overlay-and-scan-data).
* Scanner / take / index / compaction integration.
* Dense (rectangular) overlays, which already write with the existing equal-length writer and do not depend on this work.

**Dependencies**

* Blocks [OSS-1322](https://linear.app/lancedb/issue/OSS-1322/write-data-overlay-and-scan-data). No blockers.

wjones127:
## What

The v2 file writer advanced every column from a single global row counter, so a single file could only hold columns of equal length. Sparse [data overlay files](https://github.com/lance-format/lance/pull/7381) need columns whose item counts differ *within one file* (each field covers a different set of rows). This is technically allowed by the file format but unused today.

This PR adds that capability:

- **`FileWriter::write_column(column_index, array)`** — writes one column, advancing only that field's row counter. Called across multiple calls a field appends; a field never written ends up as a zero-length column; an empty array is a no-op. `write_batch` is unchanged — it still advances all fields together, so ordinary rectangular files round-trip exactly as before.
- **`FileReader::column_num_rows(column_index)`** — exposes a single column's length (sum of its pages' row counts). Always derivable from page metadata; this just surfaces it.
- **Reader projection-length validation** — the reader combines a projection's columns into rectangular batches, so they must share a length. A projection whose columns differ in length is now rejected up front with a descriptive error (naming each column's length) instead of panicking inside the decoder. `RangeFull`/`RangeFrom` resolve to the projected common length rather than `self.num_rows` (the file's longest column), so a single short column reads back at its own length. Rectangular files are unaffected (the common length equals `num_rows`).

Per-field lengths are computed from each top-level field's root (first) physical column, which equals its top-level row count for primitive, struct, and list fields in both v2.0 and v2.1.

## Tests

`lance-file` lib suite (81 tests) + doctest pass; `cargo fmt` and `cargo clippy --tests -D warnings` clean. New/extended, all rstest over V2_0 + V2_1 where relevant:

- `test_write_columns_unequal_lengths`: writes a 5-row, a 1-row, and a 0-row column in one file; asserts per-column row counts; reads each back at its own length; random access (sorted indices) within the longer column; empty-array writes are a no-op on both written and unwritten columns.
- `test_read_unequal_length_projection`: equal-length projections full-scan; mismatched-length projections error before any batch (error names each column's length); single-column `RangeFull`/`RangeFrom`/`RangeTo` resolve to that column's own length; out-of-range bounds error.
- `test_read_nested_columns_under_validation`: struct and list columns (which map to multiple physical columns) still read under the new validation path.
- `test_write_column_validation_errors`: missing explicit schema, out-of-bounds field index, and a null written into a non-nullable field are each rejected.
- `test_blocking_read_unequal_length`: the blocking read path applies the same validation.
- `test_write_batch_keeps_equal_lengths`: rectangular files still produce equal-length columns (backwards compatible).

## Acceptance criteria (OSS-1323)

- [x] v2 writer can write a single file with columns of differing item counts (no shared global row counter)
- [x] v2 reader reads each column back independently at its own length
- [x] Random access by position within a shorter/longer column returns correct values
- [x] Per-column row counts derivable from file metadata (`column_num_rows`)
- [x] Round-trip tests incl. zero-length and single-row columns alongside longer ones
- [x] Existing equal-length files round-trip unchanged

Closes lance-format/lance#7404 (OSS-1323). Root of the Data Overlay Files stack; OSS-1322 builds on this.

### DIFF
--- a/rust/lance-file/src/reader.rs
+++ b/rust/lance-file/src/reader.rs
@@ -321,18 +321,15 @@ impl ReaderProjection {
     ) -> Result<()> {
         for field in fields {
             let is_structural = file_version >= LanceFileVersion::V2_1;
+            let (contributes, recurse) = field_column_shape(field, is_structural);
             // In the 2.0 system we needed ids for intermediate fields.  In 2.1+
             // we only need ids for leaf fields.
-            if (!is_structural
-                || field.children.is_empty()
-                || field.is_blob()
-                || field.is_packed_struct())
+            if contributes
                 && let Some(column_idx) = field_id_to_column_index.get(&(field.id as u32)).copied()
             {
                 column_indices.push(column_idx);
             }
-            // Don't recurse into children if the field is a blob or packed struct in 2.1
-            if !is_structural || (!field.is_blob() && !field.is_packed_struct()) {
+            if recurse {
                 Self::from_field_ids_helper(
                     file_version,
                     field.children.iter(),
@@ -520,6 +517,119 @@ struct Footer {
 
 const FOOTER_LEN: usize = 40;
 
+// How a field maps onto physical columns, shared by the projection-building and
+// projection-validation walks so they stay in lockstep. In the 2.0 layout every
+// field (including structs and lists) has its own column; in 2.1 only leaves do,
+// and blob/packed-struct fields are opaque (a single column with no descent).
+// Returns `(contributes, recurse)`: whether the field has its own column and
+// whether to walk into its children. The DFS order is the field's own column (if
+// any) followed by its children, so a field's root (first) column is always the
+// first entry of its sub-slice.
+fn field_column_shape(field: &Field, is_structural: bool) -> (bool, bool) {
+    let contributes =
+        !is_structural || field.children.is_empty() || field.is_blob() || field.is_packed_struct();
+    let recurse = !is_structural || (!field.is_blob() && !field.is_packed_struct());
+    (contributes, recurse)
+}
+
+// Whether a field's children each cover the same rows as the field itself. Struct
+// children do (one value per parent row), so they must share its length. List,
+// map, and fixed-size-list items have an independent cardinality (item count, not
+// row count) and are validated only against themselves.
+fn children_share_parent_length(field: &Field) -> bool {
+    field.logical_type.is_struct()
+}
+
+// Validate one field's slice of a projection's flat `column_indices`, returning
+// the field's top-level row count (the page-row sum of its root column). Walks the
+// same DFS order as `from_field_ids_helper`, advancing `cursor` past every column
+// the field contributes.
+//
+// `comparable` tracks whether the field's row count shares the read's top-level
+// cardinality. A struct's children must all match that count -- the decoders
+// combine them assuming equal lengths and would otherwise panic or read past a
+// shorter child -- so the equality check runs only while `comparable` holds. Once
+// the walk descends through a list/map/fixed-size-list its items have an
+// independent cardinality (item count, not row count), so `comparable` turns off
+// for that whole subtree and a nested struct's children are no longer compared.
+fn validate_field_length<F: Fn(usize) -> Result<u64>>(
+    field: &Field,
+    is_structural: bool,
+    comparable: bool,
+    column_indices: &[u32],
+    cursor: &mut usize,
+    column_len: &F,
+) -> Result<u64> {
+    let (contributes, recurse) = field_column_shape(field, is_structural);
+    let mut field_rows: Option<u64> = None;
+    if contributes {
+        let column = *column_indices.get(*cursor).ok_or_else(|| {
+            Error::invalid_input(format!(
+                "projection supplied fewer column indices than its fields require \
+                 (ran out at field '{}')",
+                field.name
+            ))
+        })?;
+        *cursor += 1;
+        field_rows = Some(column_len(column as usize)?);
+    }
+    if recurse {
+        // Only enforce equal-length children for a struct whose own count is still
+        // at the top-level cardinality; below a list/map/fixed-size-list the items
+        // have an independent cardinality, so neither this field nor its
+        // descendants are comparable.
+        let enforce_children = comparable && children_share_parent_length(field);
+        for child in &field.children {
+            let child_rows = validate_field_length(
+                child,
+                is_structural,
+                enforce_children,
+                column_indices,
+                cursor,
+                column_len,
+            )?;
+            // A struct that contributes no column of its own (the 2.1 layout)
+            // takes its row count from its first child.
+            let expected = *field_rows.get_or_insert(child_rows);
+            if enforce_children && child_rows != expected {
+                return Err(Error::invalid_input(format!(
+                    "cannot read field '{}': its children have differing lengths \
+                     (child '{}' has {} rows, but the field has {}); a struct's \
+                     children must all have the same length",
+                    field.name, child.name, child_rows, expected
+                )));
+            }
+        }
+    }
+    field_rows.ok_or_else(|| {
+        Error::invalid_input(format!(
+            "projected field '{}' maps to no columns",
+            field.name
+        ))
+    })
+}
+
+// The reader combines a projection's columns into rectangular batches, so they
+// must all have the same length.  Returns that common length, or a descriptive
+// error (naming each column's length) when they differ. Ordinary files always
+// pass; only files written with `FileWriter::write_column` whose columns ended up
+// unequal can fail, and those must be read separately.
+fn verify_uniform_lengths(field_lengths: &[(&str, u64)]) -> Result<u64> {
+    let first = field_lengths.first().map_or(0, |&(_, len)| len);
+    if field_lengths.iter().all(|&(_, len)| len == first) {
+        return Ok(first);
+    }
+    let columns = field_lengths
+        .iter()
+        .map(|(name, len)| format!("{name}={len}"))
+        .collect::<Vec<_>>()
+        .join(", ");
+    Err(Error::invalid_input(format!(
+        "cannot read columns of differing lengths together ({columns}); \
+         read each column (or equal-length group) separately"
+    )))
+}
+
 impl FileReader {
     pub fn with_scheduler(&self, scheduler: Arc<dyn EncodingsIo>) -> Self {
         Self {
@@ -547,6 +657,29 @@ impl FileReader {
 
     pub fn num_rows(&self) -> u64 {
         self.core.num_rows()
+    }
+
+    /// The number of rows stored in a single physical column.
+    ///
+    /// For ordinary (rectangular) files every column has the same length, equal
+    /// to [`num_rows`](Self::num_rows). Files written with
+    /// [`FileWriter::write_column`](crate::writer::FileWriter::write_column)
+    /// may have columns of differing lengths; this returns the length of one
+    /// such column, derived by summing its pages' row counts. Errors if
+    /// `column_index` is out of bounds.
+    pub fn column_num_rows(&self, column_index: usize) -> Result<u64> {
+        let column = self
+            .metadata
+            .column_metadatas
+            .get(column_index)
+            .ok_or_else(|| {
+                Error::invalid_input(format!(
+                    "column index {} is out of bounds (file has {} columns)",
+                    column_index,
+                    self.metadata.column_metadatas.len()
+                ))
+            })?;
+        Ok(column.pages.iter().map(|page| page.length).sum())
     }
 
     pub fn metadata(&self) -> &Arc<CachedFileMetadata> {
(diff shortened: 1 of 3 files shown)


rust/lance-file/src/writer.rs
```diff
@@ -507,19 +515,53 @@ impl FileWriter {
...
+                    row_number,
                     num_rows,
                 )
             })
             .collect::<Result<Vec<_>>>()
     }
 
+    /// Advance the per-field row counters after a set of columns has been
+    /// written, keeping `rows_written` (the file's logical length) in sync as the
+    /// maximum column length.
```
wjones127: nitpick: Don't write doc comments for internal-only functions. If their purpose is unclear from the signature, you can provide a standard comment.

rust/lance-file/src/writer.rs
```diff
@@ -557,18 +599,100 @@ impl FileWriter {
...
+    /// # use lance_file::writer::FileWriter;
+    /// # async fn example(writer: &mut FileWriter) -> lance_core::Result<()> {
+    /// // Field 0 gets three values, field 1 gets one — a non-rectangular file.
+    /// let a: ArrayRef = Arc::new(Int32Array::from(vec![1, 2, 3]));
+    /// let b: ArrayRef = Arc::new(Int32Array::from(vec![10]));
+    /// writer.write_columns(vec![(0, a), (1, b)]).await?;
+    /// # Ok(())
+    /// # }
+    /// ```
+    pub async fn write_columns(&mut self, columns: Vec<(usize, ArrayRef)>) -> Result<()> {
```
wjones127: question: what's the advantage of taking a Vec? Why not just `pub async fn write_column(&mut self, column_index: usize, array: ArrayRef) -> Result<()>`?

rust/lance-file/src/writer.rs
```diff
@@ -1040,6 +1165,196 @@ mod tests {
...
+        use lance_encoding::decoder::FilterExpression;
+
+        let projection = ReaderProjection::from_column_names(version, schema, &[name]).unwrap();
+        let batches: Vec<RecordBatch> = reader
+            .read_stream_projected(params, 1024, 16, projection, FilterExpression::no_filter())
+            .await
+            .unwrap()
+            .try_collect()
+            .await
+            .unwrap();
```
wjones127, line 1190: question: IIUC the API is: you can read varying length columns by just projecting each set of columns with matching length? And if you try to read a projection with mismatched lengths, does it error early?

rust/lance-file/src/reader.rs
```diff
@@ -491,6 +491,88 @@ impl FileReader {
...
 
+    /// The number of rows stored in a single physical column.
+    ///
+    /// For ordinary (rectangular) files every column has the same length, equal
+    /// to [`num_rows`](Self::num_rows). Files written with
+    /// [`FileWriter::write_column`](crate::writer::FileWriter::write_column)
+    /// may have columns of differing lengths; this returns the length of one
+    /// such column, derived by summing its pages' row counts. Returns `None` if
+    /// `column_index` is out of bounds.
+    pub fn column_num_rows(&self, column_index: usize) -> Option<u64> {
```
westonpace: I think elsewhere we'd just throw an error if the column index was out of bounds.  Any particular reason to make this `Option` instead of `Result`?

rust/lance-file/src/reader.rs
```diff
@@ -491,6 +491,88 @@ impl FileReader {
...
+            .get(column_index)
+            .map(|col| col.pages.iter().map(|page| page.length).sum())
+    }
+
+    // Number of physical columns a field (and its descendants) contributes to a
+    // projection. This applies the same shape rules as
+    // `ReaderProjection::from_field_ids_helper` and assumes every contributing
+    // field has a column-index entry (true for projections built by
+    // `ReaderProjection`'s constructors). Used to partition a projection's flat
+    // `column_indices` by top-level field.
```
westonpace: > Used to partition a projection's flat `column_indices` by top-level field.

Maybe I just haven't read far enough in but I don't understand this sentence.

rust/lance-file/src/reader.rs
```diff
@@ -491,6 +491,88 @@ impl FileReader {
...
+    fn count_projected_columns(field: &Field, is_structural: bool) -> usize {
+        let contributes = !is_structural
+            || field.children.is_empty()
+            || field.is_blob()
+            || field.is_packed_struct();
+        let recurse = !is_structural || (!field.is_blob() && !field.is_packed_struct());
```
westonpace: Minor nit: I wonder if we should put these rules in a helper function

rust/lance-file/src/reader.rs
```diff
@@ -491,6 +491,88 @@ impl FileReader {
...
+        }
+        count
+    }
+
+    // The top-level row count of each projected top-level field, paired with the
+    // field's name, in projection order. A field's length is the page-row sum of
+    // its root (first) physical column, which equals the field's top-level row
+    // count for primitive, struct, and list fields in both v2.0 and v2.1.
+    // Ordinary files have equal lengths across columns; files written with
+    // `write_column` may not.
```
westonpace: Do we have any guarantees for nested fields?  For example, if I have a struct with two children do both fields need to have the same length?

rust/lance-file/src/reader.rs
```diff
@@ -491,6 +491,88 @@ impl FileReader {
...
+    fn projected_field_lengths<'a>(&self, projection: &'a ReaderProjection) -> Vec<(&'a str, u64)> {
+        let is_structural = self.metadata.version() >= LanceFileVersion::V2_1;
+        let mut lengths = Vec::with_capacity(projection.schema.fields.len());
+        let mut cursor = 0usize;
+        for field in &projection.schema.fields {
+            let n = Self::count_projected_columns(field, is_structural);
+            if n == 0 {
+                continue;
+            }
```
westonpace: I don't even know if we support this in general but I wonder how this handles the empty struct case.  It looks like it would return an empty vec (which is probably correct?)

Not important at all, just idle musing

rust/lance-file/src/reader.rs
```diff
@@ -491,6 +491,88 @@ impl FileReader {
...
+            let Some(&root_column) = projection.column_indices.get(cursor) else {
+                break;
```
westonpace: I think it should be an error here and not a break?  This means the user supplied a projection with an unequal number of fields / indices which is invalid.

rust/lance-file/src/writer.rs
```diff
@@ -507,19 +515,53 @@ impl FileWriter {
...
     }
 
+    // Advance the per-field row counters after a set of columns has been
+    // written, keeping `rows_written` (the file's logical length) in sync as the
+    // maximum column length.
+    fn advance_columns(&mut self, items: &[(usize, ArrayRef)]) {
+        for (field_idx, array) in items {
+            self.field_rows_written[*field_idx] += array.len() as u64;
+        }
+        self.rows_written = self.field_rows_written.iter().copied().max().unwrap_or(0);
```
westonpace: Super minor nit but seems you could do something like...

```
for (field_idx, array) in items) {
  let new_total = self.field_rows_written[*field_idx] + array.len() as u64;
  self.field_rows_written[*field_idx] = new_total;
  self.rows_written = self.rows_written.max(new_total);
}
```

...and save yourself from looping through all the fields on each call.

rust/lance-file/src/writer.rs
```diff
@@ -557,18 +599,92 @@ impl FileWriter {
...
-        };
+        // `write_batch` advances every field by the same amount, keeping all
+        // columns equal length. Guard against overflowing the row counter.
+        if self.rows_written.checked_add(num_rows).is_none() {
+            return Err(Error::invalid_input_source(format!("cannot write batch with {} rows because {} rows have already been written and Lance files cannot contain more than 2^64 rows", num_rows, self.rows_written).into()));
+        }
+        for field_rows in self.field_rows_written.iter_mut() {
+            *field_rows += num_rows;
+        }
+        self.rows_written = self.field_rows_written.iter().copied().max().unwrap_or(0);
```
westonpace: Same comment as before.

rust/lance-file/src/writer.rs
```diff
@@ -507,19 +515,53 @@ impl FileWriter {
...
                     num_rows,
                 )
             })
             .collect::<Result<Vec<_>>>()
     }
 
+    // Advance the per-field row counters after a set of columns has been
+    // written, keeping `rows_written` (the file's logical length) in sync as the
+    // maximum column length.
+    fn advance_columns(&mut self, items: &[(usize, ArrayRef)]) {
```
westonpace: This method is called in the `write_column` case but not in the `write_batch` case (it has its own bookkeeping of `self.field_rows_written` and `self.rows_written`.  It might be nice to call that out.

Alternatively, does this really need to be a dedicated method?  It's only called once and similar logic is inlined in `write_batch` instead of a dedicated method.

rust/lance-file/src/writer.rs
```diff
@@ -507,19 +515,53 @@ impl FileWriter {
...
                         ))?;
+                Ok((field_idx, array.clone()))
+            })
+            .collect::<Result<Vec<_>>>()?;
+        self.encode_columns(&items, external_buffers)
+    }
+
+    // Encode a set of `(field index, array)` pairs, each advancing only its own
+    // column. The returned tasks must be written before the per-field row
+    // counters are advanced (see `advance_columns`).
```
westonpace: This says the returned tasks must be written before we advance the columns.  But in `write_columns` we have...

```
        let encoding_tasks = encoding_tasks
            .into_iter()
            .flatten()
            .collect::<FuturesOrdered<_>>();

        self.advance_columns(&columns);
        self.write_pages(encoding_tasks).await?;
```

So it looks like we aren't doing the writing until after `advance_columns`?

westonpace (approved):
Very cool to see this coming together.  Some minor nits but nothing blocking.  Biggest concern is probably whether we are calling `advance_columns` in the write spot.

rust/lance-file/src/reader.rs
wjones127: I'm open to it. I guess option is nice in cases where there's only one reason it wouldn't have a value, but it does mean you don't get an error message. I'll switch to an error with a clear message.

rust/lance-file/src/reader.rs
wjones127: That's a good question. I'll look into this. I think I'm inclined to say yes, otherwise it can get especially thorny with lists. But I guess in theory structs could be more flexible. Let me see what is possible and what we need to clear up.

rust/lance-file/src/reader.rs
wjones127: Maybe worth a simple unit test.

rust/lance-file/src/reader.rs
wjones127: Done — `column_num_rows` now returns a `Result` with an out-of-bounds message instead of `Option`.

rust/lance-file/src/reader.rs
wjones127: Removed `count_projected_columns` — the length walk is now `validate_field_length`, which should read more clearly, so the confusing sentence is gone.

rust/lance-file/src/reader.rs
wjones127: Done — extracted the shape rules into a shared `field_column_shape` helper, now used by both `from_field_ids_helper` and the validation walk.

rust/lance-file/src/reader.rs
wjones127: Yes, and it's now enforced on read. `write_column` can only target a top-level field, so a struct's children always advance together from a single array and stay equal-length — there's no way to write divergent children today. To guard against a future/foreign writer, I made the reader's length check structure-aware: a struct's children must share its row count, otherwise a clear error (the decoders would otherwise panic in v2.0 / underflow in v2.1). List/map/fixed-size-list items are exempt since their cardinality legitimately differs. Covered by `test_validate_struct_child_lengths`.

rust/lance-file/src/reader.rs
wjones127: Added `test_validate_length_list_and_empty_struct` — an empty struct validates to its single column's length.

rust/lance-file/src/reader.rs
wjones127: Done — a projection with the wrong number of column indices now errors instead of silently truncating.

rust/lance-file/src/writer.rs
wjones127: Done — `advance_columns` folds each written field's new total into `rows_written` directly instead of rescanning.

rust/lance-file/src/writer.rs
wjones127: Done — `write_batch` now grows `rows_written` by `num_rows` (every field advances uniformly).

rust/lance-file/src/reader.rs
westonpace: I think this is probably a reasonable restriction.  Things could get very complicated otherwise and the read/write amplification should still be pretty trivial with this requirement.  I'm sure I could dream up some counter-example but let's keep it simple until we hit something real.

PR #7407, opened 2026-06-23
feat: data overlay files — model, feature flag, and write/commit path

wjones127:
Data model, feature flag, and write/commit foundation for [Data Overlay Files](https://github.com/lance-format/lance/pull/7381) (OSS-1322).

**Stacked on #7381** (spec/proto) and **#7406** (OSS-1323 — its `FileWriter::write_columns` is needed for the sparse overlay sink). Both stacks' commits show in this diff until they merge.

## What's here

- **Data model** — `DataOverlayFile` + `OverlayCoverage` on `Fragment.overlays`, dense (`shared_offset_bitmap`) and sparse (per-field). Coverage bitmaps are parsed once on load into `Arc<RoaringBitmap>` (not re-deserialized per access), and a fragment's overlays are stable-sorted by `committed_version` (newest last) on load so resolution can rely on the ordering. Protobuf round-trip + `coverage_for_field` tested.
- **Feature flag 64** (`FLAG_DATA_OVERLAY_FILES`) — set when any fragment has overlays. Release-gated: treated as an unknown flag in release builds (so release readers/writers refuse overlay datasets) unless `LANCE_ENABLE_DATA_OVERLAY_FILES` is set; debug builds understand it. Tested.
- **`Operation::DataOverlay` transaction** — appends overlays to a fragment (preserving concurrently-written ones) and stamps `committed_version` at commit, re-stamped on retry. Protobuf round-trip tested.
- **Conflict resolution** (v2 `CommitConflictResolver`) — permissive like `DataReplacement`: compatible with append / column-rewrite / index-build / other overlays, and with deletes/updates that leave the overlaid fragment in place; conflicts when a concurrent op *removes* or row-rewrites an overlaid fragment (`Update`/`Delete` removal, `Rewrite`/`Merge`) and with whole-dataset `Overwrite`/`Restore`. Tested (matrix + remove-fragment conflict).

The fragment read path refuses overlays until the scan merge lands.

## Follow-ups on this stack

- [ ] `DataOverlayFileWriter` — streaming dense+sparse sink, used internally behind `update`/`merge_insert`.
- [ ] `validate()` overlay checks — value-column length vs. coverage cardinality, dtype vs. schema (I/O-bound; cheap structural invariants are enforced at parse/commit time).
- [ ] Scan/take merge (OSS-1324), resolved at **IO-scheduling time** (decide base-vs-overlay before reading); the post-read merge prototype was removed from this PR.

Blocks OSS-1324 (take/scan), OSS-1325 (index masking), OSS-1326 (compaction).


rust/lance-table/src/format/fragment.rs
```diff
@@ -233,6 +234,145 @@ impl TryFrom<pb::DataFile> for DataFile {
...
+    pub committed_version: u64,
+}
+
+impl DataOverlayFile {
+    /// The parsed coverage bitmap that applies to the field stored at
+    /// `field_pos` within `data_file.fields`.
+    ///
+    /// For a dense overlay the same shared bitmap is returned for every field;
+    /// for a sparse overlay the per-field bitmap at `field_pos` is returned.
+    pub fn coverage_for_field(&self, field_pos: usize) -> Result<RoaringBitmap> {
```
wjones127: issue(blocking): this deserialized the bitmap on each invocation. Can we deserialize the bitmaps on load instead?

rust/lance-table/src/format/fragment.rs
```diff
@@ -233,6 +234,145 @@ impl TryFrom<pb::DataFile> for DataFile {
...
+            Some(pb::data_overlay_file::Coverage::SharedOffsetBitmap(bytes)) => {
+                OverlayCoverage::Shared(bytes)
+            }
+            Some(pb::data_overlay_file::Coverage::FieldCoverage(fc)) => {
+                OverlayCoverage::PerField(fc.offset_bitmaps)
+            }
+            None => {
+                return Err(Error::invalid_input(
+                    "DataOverlayFile is missing its coverage",
+                ));
```
wjones127, line 430: suggestion: add a test for this branch.

rust/lance-table/src/feature_flags.rs
```diff
@@ -20,8 +20,13 @@ pub const FLAG_TABLE_CONFIG: u64 = 8;
...
 /// Disable writing transaction file under _transaction/, this flag is set when we only want to write inline transaction in manifest
 pub const FLAG_DISABLE_TRANSACTION_FILE: u64 = 32;
+/// Fragments contain data overlay files, which supply new values for a subset of
+/// cells without rewriting base data files. A reader that does not understand
+/// overlays must refuse the dataset, since ignoring an overlay would silently
+/// return stale base values.
+pub const FLAG_DATA_OVERLAY_FILES: u64 = 64;
 /// The first bit that is unknown as a feature flag
-pub const FLAG_UNKNOWN: u64 = 64;
+pub const FLAG_UNKNOWN: u64 = 128;
```
wjones127, line 34: suggestion: since we don't support all operations yet, I'm wondering if what we should do is make this default to 128 on debug builds, 64 on release builds, and maybe have an environment variable that let's you override in release builds for benchmarking. That way, we aren't releasing this into the wild just yet, but can test it.

rust/lance/src/dataset/overlay.rs
```diff
@@ -0,0 +1,242 @@
...
+use snafu::location;
+
+use lance_table::format::DataOverlayFile;
+
+/// One field's contribution from a single overlay: which physical offsets it
+/// covers, and the value column holding those offsets' values (indexed by rank).
+#[derive(Debug, Clone)]
+pub struct ResolvedFieldOverlay {
+    /// Physical offsets this overlay covers for the field.
+    pub coverage: RoaringBitmap,
```
wjones127: suggestion: this is immutable, right? Should we put this behind an Arc for sharing?

rust/lance/src/dataset/overlay.rs
```diff
@@ -0,0 +1,242 @@
...
+    /// `coverage.len()`; the value for a covered offset `o` is at `coverage`'s
+    /// rank of `o`.
+    pub values: ArrayRef,
+}
+
+/// Order a fragment's overlays from newest to oldest for read resolution.
+///
+/// Precedence is by `committed_version` (higher is newer); ties are broken by
+/// position in the fragment's `overlays` list, where a later entry is newer.
+/// Returns indices into `overlays`.
```
wjones127: suggestion: if position in the list is already load bearing, why don't we just assume it's already in version order. On load, we can validate this, and do a one-time sort if needed. Then we can assume it's sorted in all further operations.

wjones127:
Superseded by #7535 (same content, head + base in-repo) so the OSS-1324 PR can stack on it directly. Closing in favor of the in-repo PR.

PR #7409, opened 2026-06-23
feat: resolve data overlay files on the take (and scan) read path

wjones127:
Resolves the `take` random-access path against data overlay files, replacing the temporary "overlays not supported" error from #7407 (OSS-1322). Implements OSS-1324.

Because `take` and scan share `FragmentReader::new_read_impl`, the merge is wired there once: each row is addressed by its physical offset (from `ReadBatchParams::to_offsets_total`) and resolved against the overlays that cover its field. This necessarily also enables the scan-path merge that #7407 stubbed out.

### How it works
- `FileFragment::open` loads, for each projected field, the overlay value columns that cover it, ordered newest-first (by `committed_version`, list-position tiebreak).
- The merge runs on physical rows in read order, **before** deletion filtering, so:
  - deletions take precedence (an overlay value computed for a deleted row is dropped with the row),
  - NULL overrides apply (a covered offset with a NULL value resolves to NULL, distinct from fall-through),
  - fields resolve independently.
- The resolution core now takes explicit per-row physical offsets instead of a contiguous start, giving one code path for the contiguous scan range and arbitrary take indices.
- Sparse per-field overlays read each field's value column independently, so unequal-length value columns (#7406 / OSS-1323) need no rectangular batch. Rank-based addressing only (rank on the coverage bitmap + a value fetch; no offset key column, no binary search).

Overlays on nested (non-top-level) fields are not yet matched and are left for follow-up.

### Tests
take covered/uncovered offsets; multiple overlays (newest wins); per-field coverage with unequal-length columns; NULL override; overlay on a deleted row (inert); multi-fragment scan — each over v2.0 and v2.1. Plus unit tests for the offset-based core and the batch merge.

### Stacking
Stacked on #7406 (OSS-1323) and #7407 (OSS-1322). Until those merge, this PR's diff against `main` includes their commits; review only the final commit ("resolve data overlay files on the take and scan read paths").


wjones127:
Superseded by #7536, relocated into `lance-format/lance` and properly stacked on the OSS-1322 branch (#7535) so the diff is exactly this change. Closing in favor of the stacked PR.

PR #7412, opened 2026-06-23
fix(update): keep nested-field index correct when updating a struct column

jackye1995:
The update API rejects nested column references (`set` errors on a `.` in the column name), so a nested field can only be changed by setting its whole **top-level struct column** (e.g. `SET s = named_struct('x', …)`).

`commit_impl` built `fields_for_preserving_frag_bitmap` from the struct column's own (parent) field id via `field_id(column_name)`, so an index on a nested **child** field (a different leaf id) was absent from the set. `register_pure_rewrite_rows_update_frags_in_indices` then wrongly extended that child-field index over the rewritten fragment, which was therefore treated as indexed and never re-scanned — so the updated rows were silently dropped from queries on the index (false negatives).

The fix collects the full field subtree of each updated column, so a struct-column update marks all descendant field ids as modified. Flat columns are unaffected (no children).

This is the SQL `UPDATE` counterpart of the merge_insert fix in lance-format/lance#7410. Adds a regression test (`test_update_struct_column_keeps_nested_index`): a BTree index on `s.x`, an update of the struct column `s`, and an assertion that the index's effective fragment bitmap is not extended over the rewritten fragment so the updated value is still found.

### DIFF
--- a/rust/lance/src/dataset/write/update.rs
+++ b/rust/lance/src/dataset/write/update.rs
@@ -34,6 +34,16 @@ use lance_table::format::{Fragment, RowIdMeta};
 use roaring::RoaringTreemap;
 use snafu::ResultExt;
 
+/// Collect a field id and all of its descendant field ids (pre-order). A struct
+/// column update rewrites the whole subtree, so an index on any descendant must be
+/// treated as modified.
+fn collect_subtree_field_ids(field: &lance_core::datatypes::Field, out: &mut Vec<u32>) {
+    out.push(field.id as u32);
+    for child in &field.children {
+        collect_subtree_field_ids(child, out);
+    }
+}
+
 /// Build an update operation.
 ///
 /// This operation is similar to SQL's UPDATE statement. It allows you to change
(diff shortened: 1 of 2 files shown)


rust/lance/src/dataset/write/update.rs
```diff
@@ -386,10 +396,14 @@ impl UpdateJob {
         dataset: Arc<Dataset>,
         update_data: UpdateData,
     ) -> Result<UpdateResult> {
+        // Updated columns are top-level (nested references are rejected by `set`), but a
```
westonpace, line 399: I'm surprised that updated columns have to be top-level?  I would think we could do an update on a nested column.

westonpace (approved):

rust/lance/src/dataset/write/update.rs
jackye1995: yeah, this is still a TODO: https://github.com/lance-format/lance/blob/main/rust/lance/src/dataset/write/update.rs#L125-L135

jackye1995: tracked now in https://github.com/lance-format/lance/issues/7420

PR #7413, opened 2026-06-23
perf(index): add sparse ngram tokenization

everySympathy:
## What

Adds an optional sparse n-gram tokenization mode to the scalar NGram index.

The default index remains the existing fixed trigram mode. Users can opt into sparse tokenization with NGram index params:

```json
{"tokenization": "sparse"}
```

Sparse tokenization stores selected longer n-gram tokens instead of every fixed trigram. This is intended for long literal regex / contains predicates where each individual trigram is common, but the longer literal itself is rare. The implementation is inspired by the sparse n-gram idea described in Cursor's fast regex search write-up.

## How

- Adds `NGramTokenization::{Trigram, Sparse}` and persists the selected mode in postings metadata.
- Keeps existing trigram behavior as the default for compatibility.
- Builds sparse index tokens from normalized alphanumeric runs using a monotonic-stack sparse n-gram construction.
- Bounds sparse token length with `SPARSE_MAX_NGRAM_LEN` so build and query planning share the same token length contract.
- Uses stack-based query covering to emit a smaller necessary token subset instead of querying every valid sparse n-gram.
- Reuses the regex-to-ngram planner, but derives token requirements against the selected index tokenization mode.
- Extends `regex_ngram` benchmark coverage to compare fixed trigram and sparse n-gram indexes on the same dataset.

## Benchmark

Latest local `regex_ngram` Criterion run after the stack covering rewrite, 200k rows:

| Query | Trigram | Sparse | Result |
|---|---:|---:|---|
| `regexp_match(doc, 'zqxwvu.*needlexyz')` | 12.778 ms | 12.631 ms | roughly flat |
| `regexp_match(doc, '(zqxwvu\|qwerasdf\|needlexyz)')` | 16.033 ms | 15.986 ms | roughly flat |
| `regexp_match(doc, 'zqxwvu')` | 11.757 ms | 11.639 ms | roughly flat |
| `regexp_match(doc, 'a.b')` | 91.381 ms | 95.001 ms | within noise |
| `regexp_match(doc, 'sparsemarkerabcdefghijklmnopqrstuvwx')` | 50.832 ms | 3.666 ms | ~13.9x faster |

The last query is the target case: every row contains all fixed trigrams from the long literal, but only a small fraction contains the full literal. Fixed trigrams therefore produce a broad candidate set, while sparse longer n-grams stay selective.

The latest run also compared against Criterion history and did not show a regression; the target sparse long-literal case was reported as improved.

## Testing

- `cargo fmt --all`
- `cargo test --target-dir /tmp/lance-target-4ca9 -p lance-index ngram`
- `cargo clippy --target-dir /tmp/lance-target-4ca9 --all --tests --benches -- -D warnings`
- `regex_ngram` Criterion benchmark via the generated bench binary: `/tmp/lance-target-4ca9/release/deps/regex_ngram-d4cb3171b6e7c3d6 --bench --noplot`

### DIFF
--- a/rust/lance-index/src/scalar/ngram.rs
+++ b/rust/lance-index/src/scalar/ngram.rs
@@ -6,7 +6,7 @@ use std::collections::BTreeMap;
 use std::iter::once;
 use std::time::Instant;
 use std::{
-    collections::{HashMap, HashSet},
+    collections::{HashMap, HashSet, VecDeque},
     sync::Arc,
 };
 
@@ -20,8 +20,7 @@ use crate::metrics::NoOpMetricsCollector;
 use crate::pbold;
 use crate::scalar::expression::{ScalarQueryParser, TextQueryParser};
 use crate::scalar::registry::{
-    DefaultTrainingRequest, ScalarIndexPlugin, TrainingCriteria, TrainingOrdering, TrainingRequest,
-    VALUE_COLUMN_NAME,
+    ScalarIndexPlugin, TrainingCriteria, TrainingOrdering, TrainingRequest, VALUE_COLUMN_NAME,
 };
 use crate::scalar::{CreatedIndex, UpdateCriteria};
 use crate::vector::VectorIndex;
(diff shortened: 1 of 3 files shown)


westonpace:
Looks like a really cool idea!

everySympathy:
> Looks like a really cool idea!

 Thanks! The idea was inspired by Cursor’s fast regex search blog post. I’ll keep pushing this PR forward.

everySympathy:
@westonpace Hi, Could you take a review at this PR? Thanks!

PR #7414, opened 2026-06-23
fix(io): clamp ObjectStore::io_parallelism() to at least 1

LuciferYang:
## Summary

`ObjectStore::io_parallelism()` returns the `LANCE_IO_THREADS` override verbatim when that env var is set, so `LANCE_IO_THREADS=0` yields `0`. Callers feed this value straight into `buffered` / `buffer_unordered`, and a window of **0** makes those streams **never poll their input** — so a plain metadata-only `count_rows` (and ~8 other fan-out sites in `dataset.rs`) would hang instead of returning.

```diff
 pub fn io_parallelism(&self) -> usize {
     std::env::var("LANCE_IO_THREADS")
         .map(|val| val.parse::<usize>().unwrap())
         .unwrap_or(self.io_parallelism)
+        .max(1)
 }
```

Clamping at the source covers every caller in one place. The store's configured default is already `>= 1`, so this only changes the explicit `LANCE_IO_THREADS=0` case.

## Context

Follow-up to #7076, where the same hang was fixed locally in the count path (per review). This addresses the root cause so the other unguarded `io_parallelism()` → `buffered`/`buffer_unordered` sites are covered too.

## Test

`test_io_parallelism_clamped_to_nonzero` asserts `LANCE_IO_THREADS=0` clamps to `1`, a positive override (`8`) passes through unchanged, and the default is `>= 1`.

### DIFF
--- a/rust/lance-io/src/object_store.rs
+++ b/rust/lance-io/src/object_store.rs
@@ -578,10 +578,17 @@ impl ObjectStore {
         self.max_iop_size
     }
 
+    /// The amount of parallelism to use for I/O operations.
+    ///
+    /// Honors the `LANCE_IO_THREADS` override when set, otherwise the store's configured value.
+    /// Always at least 1: callers feed this straight into `buffered` / `buffer_unordered`, and a
+    /// window of 0 makes those streams never poll their input — e.g. a metadata-only `count_rows`
+    /// would hang rather than return.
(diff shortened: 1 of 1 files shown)


westonpace (approved):

LuciferYang:
Thanks @westonpace

PR #7415, opened 2026-06-23
fix(namespace): allow create_branch bootstrap on managed tables

geruh:
This PR picks up on the work in #7403 by @dantasse.

`create_branch` on managed tables calls `create_table_version` before `_refs/branches/<branch>.json` exists, and `resolve_branch_location` requires that ref. Bootstrap commit fails.

`create_table_version` now goes through `resolve_branch_for_commit` instead check the ref first, and if it's missing only allow the commit when the branch chain has no versions yet. Versions without a ref still get rejected (zombie).

`create_branch("exp")` used to fail on the first commit because `create_table_version` looked for `_refs/branches/exp.json` before phase 2 wrote it.

```
# branching isn't atomic so:
phase 1: manifest → create_table_version (ref doesn't exist yet ← was failing here)
phase 2: write _refs/branches/exp.json
```

### Testing
- Added `test_create_branch_on_managed_dataset_succeeds`. 
- Zombie case is already in `test_branch_ops_reject_zombie_branch`.

### DIFF
--- a/rust/lance-namespace-impls/src/dir.rs
+++ b/rust/lance-namespace-impls/src/dir.rs
@@ -1281,12 +1281,19 @@ impl DirectoryNamespace {
             .uri)
     }
 
-    /// Resolves the branch URI for a `create_table_version` commit.
+    /// Resolves a branch to its `(uri, object-store path)` for a
+    /// `create_table_version` commit, so the commit write and the zombie check
+    /// share one path.
+    ///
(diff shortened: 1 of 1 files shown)


dantasse (approved):
Cool, looks like a good solution to me!

jackye1995 (approved):
looks good to me!

PR #7416, opened 2026-06-23
perf(fts): remove OR hot path bookkeeping

BubbleCal:
Recent FTS changes added AND/fuzzy bookkeeping to shared OR query paths. Ordinary non-fuzzy OR searches were still allocating and checking grouped expansion state, maintaining matched-position tracking in posting-list loading, and recording AND-only WAND metrics even though those values are only meaningful for conjunction/grouped expansion paths.

This PR keeps ordinary OR on the lightweight path:

- only allocates and updates required/matched position tracking for AND or phrase queries
- skips grouped-position `HashSet` creation and per-frequency grouped checks unless grouped fuzzy expansions are present
- records AND-only WAND metrics only for `Operator::And`

Grouped fuzzy AND rescoring remains on its existing grouped path so matched-expansion scoring stays correct.

No benchmark was run in this PR. The change is intentionally scoped to hot-path bookkeeping removal; benchmark validation can run later on the existing benchmark branch/VM if needed.

### What is the bug?
The ordinary OR path was paying for bookkeeping intended for AND/fuzzy/grouped expansion correctness.

It caused avoidable overhead in OR query execution and showed up as a small regression in OR benchmark configurations after the recent FTS changes.

The OR path now bypasses the AND/fuzzy-only bookkeeping while preserving the grouped expansion branch used by fuzzy AND.

## Validation

- `cargo fmt --all`
- `cargo fmt --all --check`
- `git diff --check`
- `CARGO_TARGET_DIR=/tmp/lance-target-fts-or-hotpath cargo clippy -p lance-index --tests -- -D warnings`
- `CARGO_TARGET_DIR=/tmp/lance-target-fts-or-hotpath cargo test -p lance-index scalar::inverted::wand::tests -- --nocapture`
- `CARGO_TARGET_DIR=/tmp/lance-target-fts-or-hotpath cargo test -p lance-index scalar::inverted::index::tests::test_bm25_search_uses_global_idf -- --nocapture`
- `CARGO_TARGET_DIR=/tmp/lance-target-fts-or-hotpath cargo test -p lance-index scalar::inverted::index::tests::test_and_query_returns_empty_when_exact_term_missing -- --nocapture`
- `CARGO_TARGET_DIR=/tmp/lance-target-fts-or-hotpath cargo test -p lance-index scalar::inverted::index::tests::test_fuzzy_and -- --nocapture`

### DIFF
--- a/rust/lance-index/src/scalar/inverted/index.rs
+++ b/rust/lance-index/src/scalar/inverted/index.rs
@@ -769,6 +769,33 @@ impl InvertedIndex {
         if limit == 0 {
             return Ok((Vec::new(), Vec::new()));
         }
+
+        fn push_scored_candidate(
+            candidates: &mut BinaryHeap<Reverse<ScoredDoc>>,
+            limit: usize,
+            addr: CandidateAddr,
+            score: f32,
+        ) -> Result<()> {
+            // resolve_deferred_candidates ran upstream, so every candidate
+            // carries a real row_id at this point.
+            let row_id = match addr {
+                CandidateAddr::RowId(r) => r,
+                CandidateAddr::Pending(_) => {
+                    return Err(Error::internal(
+                        "bm25_search post-condition: deferred candidate left unresolved",
+                    ));
+                }
+            };
+
+            if candidates.len() < limit {
+                candidates.push(Reverse(ScoredDoc::new(row_id, score)));
+            } else if candidates.peek().unwrap().0.score.0 < score {
+                candidates.pop();
+                candidates.push(Reverse(ScoredDoc::new(row_id, score)));
+            }
+            Ok(())
+        }
+
         let mask = prefilter.mask();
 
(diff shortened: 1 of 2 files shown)


westonpace (approved):
One minor suggestion from Claude:

> [rust/lance-index/src/scalar/inverted/index.rs](https://github.com/lance-format/lance/blob/main/rust/lance-index/src/scalar/inverted/index.rs) ~line 884 — Scoring loop body is duplicated ~25 lines between the two branches. The fast path and full path share identical logic for addr resolution, score accumulation over freqs, and heap push/pop. Only the grouped_positions.contains guard and the grouped-expansion inner loop differ. Extracting the shared scaffolding (addr resolution + heap update) into a closure or small helper would halve the duplication and reduce maintenance risk when either scoring formula changes in the future. This is a cleanup concern, not a correctness bug.

PR #7417, opened 2026-06-23
feat(mem_wal): row-level delete via tombstone rows

hamersaw:
## Summary

Adds row-level **DELETE** to the in-memory WAL (`mem_wal`), expressed as **tombstone (sentinel) rows**: a delete appends the primary key plus `_tombstone = true` (null elsewhere). A tombstone is the newest value for its PK — it wins newest-per-PK resolution (suppressing the older real row) and is then silently dropped from query results.

This is the lance (OSS) half of the WAL delete feature. The sophon half (compaction hard-delete + serving) consumes this API separately.

## What's here

- **`_tombstone` column** — lance-owned, non-nullable `Boolean` appended to the memtable/generation schema only (never the base table). `ShardWriter::open` extends the caller's base schema internally; callers keep passing the base schema and never name the column.
- **Write path** — `ShardWriter::put` injects `_tombstone = false`; new **`ShardWriter::delete(keys)`** builds the tombstone batch from a key-only input. Legacy WAL entries (written before this change) are back-filled on replay.
- **HNSW null-vector fix** — mem_wal's HNSW rejected null vectors outright, so it couldn't handle a nullable vector column at all (a pre-existing bug, independent of deletes). `append_batch` now skips null rows like the base-table index, compacting survivors while preserving their original offsets in `row_ids` so search still resolves to the right row.
- **Read paths**
  - *Filtered read*: fold `NOT _tombstone` into the predicate per WAL arm (only when the source carries the column), so it runs before the per-source pushdown limit and post-dedup in the active arm.
  - *Point lookup*: carry `_tombstone` through and filter **after** `CoalesceFirstExec` (per-source filtering would let the coalesce fall through and resurrect the row); the plan-free BTree fast path short-circuits a tombstone hit to not-found.
  - *Vector / FTS*: no filter needed — a null-payload tombstone is never an index candidate; the deleted real row is suppressed by the existing block-list (cross-gen) / `NewestPkFilter` (within active).

## Notable design points

- Fold/carry `_tombstone` **only when the source schema has the column** — back-compat for legacy generations and existing tests for free.
- `_tombstone` is non-nullable so the point-lookup base arm can synthesize a matching `Literal(false)` for `CoalesceFirstExec`'s exact-schema check.
- Delete requires non-PK columns to be nullable (a tombstone nulls them); surfaced as a clear error otherwise.
- Caveat documented in code: a delete-heavy burst between compactions can under-deliver a `LIMIT`/top-k query (uncompensated block-list drops) — recall degradation, not wrong content; self-limiting as compaction drains tombstones.

## Testing

19 new tests; full `mem_wal` suite green (417 passed, 0 failed). `cargo fmt` + `cargo clippy -p lance --tests` clean. Coverage: filtered-read masking + filter + limit, point-lookup resurrection guards (fast + plan paths), delete-then-reinsert, delete-of-absent, HNSW null-vector handling (mid-batch + all-null), write-path injection/round-trip and validation.

Draft — opening for early review.

### DIFF
--- a/rust/lance/src/dataset/mem_wal.rs
+++ b/rust/lance/src/dataset/mem_wal.rs
@@ -43,6 +43,53 @@ pub mod util;
 mod wal;
 pub mod write;
 
+use std::sync::Arc;
+
+use arrow_schema::{DataType, Field as ArrowField, Schema as ArrowSchema};
+
+/// Column name for the mem_wal tombstone (delete sentinel) marker.
+///
+/// `_tombstone` is a *physical* column present only in mem_wal memtables and
+/// flushed generations — it is deliberately kept out of the base table (hard
+/// delete), so it is **not** a virtual [`is_system_column`](lance_core::is_system_column).
+/// A row with `_tombstone = true` is a delete sentinel: the newest value for
+/// its primary key, carrying null in every non-PK column, that wins
+/// newest-per-PK resolution and is then silently dropped from query results.
+///
+/// The column is owned end-to-end by lance: callers pass the base schema and
+/// lance injects the column on the write path ([`write::ShardWriter::put`] /
+/// [`write::ShardWriter::delete`]), so no caller ever constructs or names it.
+pub const TOMBSTONE: &str = "_tombstone";
+
+/// The mem_wal tombstone field appended to the base schema to form the
+/// memtable/generation schema.
+///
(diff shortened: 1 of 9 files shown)


jackye1995 (approved):
don't have much time to review for now, mostly looks good, will probably do some post-merge review to catch up

PR #7421, opened 2026-06-23
feat: overlay-aware compaction scheduler and rewrite

wjones127:
Implements OSS-1326. Overlay-aware compaction: a scheduler that reads per-fragment overlay state and decides when and in which mode to compact, plus the rewrite executor for both modes.

> **Stacked on #7409 (OSS-1324).** This PR branches off the 1324 branch (which carries 1322 + 1323 + 1324), because the compaction executor must *read* overlay cell values — a capability that lands in 1324, not 1322. Review the final commit (`feat: overlay-aware compaction scheduler and rewrite`); the rest is the stack below it.

## What it does

**Awareness & scheduling** (`OverlayFragmentState`, `plan_overlay_compaction`)
- Reads each overlaid fragment's state: overlay count, covered offsets/cells, and the version gap between an overlay's `committed_version` and the base (current dataset version) and each covering index's `dataset_version`.
- Picks a mode from the version-gap staleness signal (`choose_mode`): **fold** when an index on an overlaid field is stale, else **merge** once overlays accumulate, else **skip**. Field-aware — an overlay on an unrelated column never makes an index look stale.

**Execution** (`compact_overlays`)
- **Overlay → overlay merge**: walks the overlays newest-first and writes a single smaller overlay carrying the per-`(offset, field)` post-image, stamped with the **maximum** input `committed_version`. Base and indexes are untouched.
- **Overlay → base fold**: materializes the post-image of every covered cell into a fresh base data file, tombstones the folded fields in the old files, and clears the fragment's `overlays`. Row addresses are preserved (a column rewrite, not a row rewrite). The same commit drops the folded fragment from any index on a folded field, so an index is never left serving stale values.

**Read efficiency** (the core optimization)
- Each base/overlay cell is read at most once: overlay value columns are read only at the **ranks that win** (a cell a newer overlay supplies is never fetched from an older overlay), via `FileFragment::read_overlay_field_winners` using rank `take`s. Base columns are read whole — the cheaper coalesced read the spec permits.

The rewrite commits through the existing `Operation::Update` (`RewriteColumns`) path, so there's no new persisted operation or conflict-matrix change. Index reconciliation uses the **drop-from-coverage** option (the cheaper of the two the spec allows); rebuild-in-place is a possible follow-up.

## Tests
- Scheduler mode selection: version-gap computation, field-awareness, fold-on-stale-index, merge-on-count, skip.
- Fold: materialization + overlays cleared; index reconciliation (folded fragment removed from the stale index's coverage, query still correct); NULL override; multi-fragment / multi-overlay (newest wins).
- Merge: collapse to one overlay preserving the max `committed_version`; union coverage; read still correct.
- Read-each-cell-once: `assert_io` check that overlapping coverage reads strictly fewer value bytes than a disjoint negative control of the same per-overlay size.


wjones127:
Superseded by #7567, relocated into `lance-format/lance` and properly stacked on the OSS-1324 branch (#7536) so the diff is exactly this change. Closing in favor of the stacked PR.

PR #7423, opened 2026-06-23
fix(fts): use async send in FTS index builder to prevent thread-pool …

### LINKED ISSUE #3568: bug(python): create_fts_index deadlocks with the ngram tokenizer on CPU-limited hosts (1CPU K8S pod)
### LanceDB version

v0.33.0

### What happened?

create_fts_index(..., base_tokenizer="ngram", prefix_only=False) deadlocks permanently (the process sits at ~0% CPU with every thread asleep) once the input is large enough to produce more than one posting-list output batch, when the lance CPU runtime's blocking pool has only a single usable thread — i.e. on any host where num_cpus::get() <= 2 (small VMs, CPU-limited / pinned Kubernetes pods).

It is not a slow computation: during the hang, CPU is a sustained 0.0% and all threads are in futex_wait/epoll_wait. It reproduces with purely synthetic random text. Setting LANCE_CPU_THREADS=2 (or higher) fixes it.

Looks like source of the issue is in `rust/lance-index/src/scalar/inverted/builder.rs`
Deadlock cycle on a 1-thread blocking pool:

1. Producer sends batch B1 => cap-1 channel is full.
2. Producer's send_blocking(batch B2) parks the sole blocking thread.
3. Consumer receives B1, calls write_record_batch => do_flush => spawn_cpu => needs the same pool => none free.
Consumer can't complete the write, never drains the channel; producer never unparks. Permanent stall.

**Environment**

lancedb==0.33.0 (pins lance core =7.0.0), Python 3.12, pyarrow 24.0.0
Linux x86_64. Reproduces on a 1‑vCPU machine and on CPU‑limited pods. Local /tmp dataset (no object store involved).

**Suggested fixes**

Any of:

- Use an async tx.send(batch).await instead of send_blocking in write_posting_lists, so the producer yields instead of parking a pool thread.
- Ensure get_num_compute_intensive_cpus() returns at least 2 (the writer needs producer + consumer to make progress concurrently), and/or make CPU sizing cgroup-quota aware.
- Run the producer and the consumer-encode on distinct pools, or decouple the channel capacity from the pool size.

**Workaround**

Set LANCE_CPU_THREADS=2 (or higher) in the environment. This fully resolves the hang without changing index semantics (keeps prefix_only=False).

### Are there known steps to reproduce?

```python
import os, time, random, pyarrow as pa, lancedb
random.seed(1234)
alpha=[chr(c) for c in range(0x05D0,0x05EA)]+list("abcdefghijklmnopqrstuvwxyz0123456789")
rows,total=[],0
while total<800_000:
    L=random.randint(50,300); s="".join(random.choice(alpha) for _ in range(L)); rows.append(s); total+=len(s)
db=lancedb.connect("/tmp/lance_repro")
try: db.drop_table("t")
except: pass
tb=db.create_table("t",pa.table({"search_text":pa.array(rows,type=pa.string())}),mode="overwrite")
t0=time.time()
tb.create_fts_index("search_text",replace=True,use_tantivy=False,base_tokenizer="ngram",
  ngram_min_length=3,ngram_max_length=3,prefix_only=False)
print("done %.2fs"%(time.time()-t0))
```

On a host with ≤ 2 visible CPUs: hangs indefinitely (done never prints).
LANCE_CPU_THREADS=2 python repro.py: completes in ~1s.

a-agmon:
Fixes https://github.com/lancedb/lancedb/issues/3568 (the issue arises in lancedb indexing)

Building a full-text-search index hangs permanently at 0% CPU on hosts
whose Lance CPU pool has a single thread. 
The CPU compute pool is sized `max(1, num_cpus - LANCE_IO_CORE_RESERVATION)` (default reservation 2), so any machine with <= 3 visible CPUs (1-vCPU VMs, CI runners, CPU-limited Kubernetes pods) collapses to a 1-thread pool and deadlocks.

Root cause is in `write_posting_lists`. The posting-list producer runs on the CPU pool via `spawn_cpu` and pushes batches into a capacity-1 `async_channel` using the synchronous `tx.send_blocking()`. When the channel is full, `send_blocking` parks the OS thread it is running on. On a single-thread pool that is the only thread, and the async consumer's column encoder (`write_record_batch` -> `spawn_cpu`) needs that same pool to drain the channel. The parked producer and the starved consumer wait on each other forever: no timeout, no error, just a silent hang at 0% CPU. 
The hang only triggers once the posting lists span a second output batch (so the producer reaches a second, blocking send), which is why it appears as a data-size "cliff". 

The PR restructures the producer as an async task that builds each batch on the CPU pool via `spawn_cpu` and dispatches it with `tx.send(batch).await`. When the channel is full, `send().await` yields the task back to the runtime instead of parking a pool thread, so the consumer can always be scheduled to drain it. Between batches the producer holds no pool thread while waiting, making the pool size irrelevant. The builder and the remaining posting-list iterator are handed back out of each `spawn_cpu` call so the cross-batch cache-group accumulator is preserved.

In addition, it adds a regression test that writes a partition whose posting lists span many output batches (exercising channel back-pressure) under a timeout and verifies every batch is searchable.

(verbose comments added in the code intentionally for review purposes - can be removed if inappropriate. I just thought it might be helpful as the issue is somewhat confusing)

### DIFF
--- a/rust/lance-index/src/scalar/inverted/index.rs
+++ b/rust/lance-index/src/scalar/inverted/index.rs
@@ -7685,6 +7685,85 @@ mod tests {
         );
     }
 
+    // Enough distinct tokens that `write_posting_lists` emits several posting-list
+    // batches (the default batch size is 256 rows), exercising the restructured
+    // producer and async send path.
+    const MANY_BATCH_TOKENS: u64 = 1000;
+    const MANY_BATCH_ROW_ID_BASE: u64 = 1000;
+
+    // Writes a single partition whose posting lists span many output batches. Each
+    // token `tok{i:05}` maps to row id `MANY_BATCH_ROW_ID_BASE + i`.
+    async fn write_partition_spanning_many_batches(store: &dyn IndexStore) {
+        let mut builder = InnerBuilder::new(0, false, TokenSetFormat::default());
+        for i in 0..MANY_BATCH_TOKENS {
+            // Zero-padded so tokens are inserted in sorted order, as the set expects.
+            builder.tokens.add(format!("tok{i:05}"));
+            let doc_id = builder.docs.append(MANY_BATCH_ROW_ID_BASE + i, 1);
+            let mut posting_list = PostingListBuilder::new(false);
+            posting_list.add(doc_id, PositionRecorder::Count(1));
+            builder.posting_lists.push(posting_list);
+        }
+        builder
+            .write(store)
+            .await
+            .expect("writing posting lists should succeed");
+    }
+
+    // Correctness guard for the restructured posting-list writer. The producer now
+    // builds each batch in its own `spawn_cpu` call, handing the builder and the
+    // remaining posting lists back out so state (the cross-batch cache-group
+    // accumulator) is preserved, and dispatches with an async `send().await`. This
+    // verifies that path over many batches by checking representative tokens after
+    // they cross the producer/consumer boundary.
+    //
+    // Note: this does not reproduce the single-thread-pool deadlock the async send
+    // fixes -- that requires a 1-thread CPU pool (a process-global singleton) plus
+    // ~8MB of buffered posting data to trigger a consumer-side encoder flush, which
+    // is impractical as a lightweight unit test.
+    #[tokio::test]
+    async fn test_write_many_posting_list_batches_preserves_all_batches() {
+        let tmpdir = TempObjDir::default();
+        let store = Arc::new(LanceIndexStore::new(
+            ObjectStore::local().into(),
+            tmpdir.clone(),
+            Arc::new(LanceCache::no_cache()),
+        ));
+
+        write_partition_spanning_many_batches(store.as_ref()).await;
+
+        write_test_metadata(&store, vec![0], InvertedIndexParams::default()).await;
+        let cache = Arc::new(LanceCache::with_capacity(4096));
+        let index = InvertedIndex::load(store.clone(), None, cache.as_ref())
+            .await
+            .unwrap();
+
+        // Probe tokens from the first, a middle, and the last batch to confirm they
+        // remain queryable after crossing the producer/consumer boundary.
+        for token_idx in [0u64, MANY_BATCH_TOKENS / 2, MANY_BATCH_TOKENS - 1] {
+            let tokens = Arc::new(Tokens::new(
+                vec![format!("tok{token_idx:05}")],
+                DocType::Text,
+            ));
+            let params = Arc::new(FtsSearchParams::new().with_limit(Some(10)));
+            let (row_ids, _) = index
+                .bm25_search(
+                    tokens,
+                    params,
+                    Operator::Or,
+                    Arc::new(NoFilter),
+                    Arc::new(NoOpMetricsCollector),
+                    None,
+                )
+                .await
+                .unwrap();
+            assert_eq!(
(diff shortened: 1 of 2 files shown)


a-agmon:
Hi @westonpace  - would be happy for your review. 
This issue causes a nasty bug on K8S pods with one core, and it took my team quite some time to pin down. Especially as it occurs in native rust space. Submitting this PR to resolve this. 
Thanks!

rust/lance-index/src/scalar/inverted/index.rs
```diff
@@ -7685,6 +7685,79 @@ mod tests {
...
+    // Regression test for the FTS posting-list writer deadlock. A partition whose
+    // posting lists span many output batches forces the writer's producer to fill
+    // the bounded channel and wait for the consumer to drain it. The producer now
+    // dispatches batches with an async `send().await`, so a full channel yields the
+    // task instead of parking the CPU-pool thread the consumer's column encoder
+    // also needs. Before the fix this deadlocked at 0% CPU whenever the CPU pool
+    // had a single thread (hosts with <= 3 visible CPUs). The timeout guards
+    // against a regression, and the searches confirm every batch was written.
+    #[tokio::test]
+    async fn test_write_many_posting_list_batches_does_not_deadlock() {
```
wjones127: issue(blocking): one of the preconditions for triggering the deadlock is having a thread pool of a certain size. This doesn't configure the size of the thread pool. Does it actually fail in CI when run on a multi-cpu runner?

wjones127 (changes requested):
Thanks for doing this. It took me a bit to reason about this. I think we need to add better docs to the `spawn_cpu()` function and maybe check other usage. Basically it seems like we should never use `spawn_cpu()` for any work that will cause the thread to sleep. No channels, no IO, and no locks. It should just consume CPU and finish. If you need a blocking thread that does any of those things, we should instead just use a `spawn_blocking` against a unbounded threadpool. Does that sound right to you @westonpace ?

rust/lance-index/src/scalar/inverted/index.rs
a-agmon: thanks for the comment @wjones127 

The test doesn’t fail on a multi-core CI runner.

I tried to reproduce the deadlock in a unit test, but a faithful reproduction is fairly involved. It requires a fresh process with LANCE_CPU_THREADS=1, and the real writer only re-enters spawn_cpu() after its per-column encoding buffer crosses the current 8 MiB threshold. Until that flush occurs, the consumer does not schedule the spawn_cpu() work needed to complete the deadlock cycle. That makes an end-to-end regression substantially heavier.

 I have therefore removed the misleading timeout and renamed the test. It now verifies correctness of the restructured multi-batch producer path rather than claiming to reproduce the deadlock. In the changed path, the spawn_cpu() closure only builds a batch; channel backpressure is handled by send().await outside the CPU pool.

 I still think the test has value because it documents and exercises the multi-batch path affected by the fix, but I agree that presenting it as a deadlock regression was misleading.

I’d prefer to keep the broader spawn_cpu() contract and call-site audit separate from this focused fix.

wjones127: That sound good. We can create a follow up issue for that.

wjones127:
Created follow up: https://github.com/lance-format/lance/issues/7637

wjones127 (approved):

PR #7425, opened 2026-06-23
ci: detect Cargo.lock drift offline + harden java --locked

wjones127:
The repo keeps three lockfiles — root `Cargo.lock`, `python/Cargo.lock`, and `java/lance-jni/Cargo.lock` (the latter two are excluded from the workspace). A `workspace.dependencies` change can leave the excluded locks stale. python CI already runs `--locked`, but java CI did not, so a stale java lock went uncaught until a release build.

Changes:
- **`cargo-lock-sync` pre-commit hook** — verifies all three lockfiles with `cargo metadata --frozen` (`--locked` + `--offline`). No network, no compilation; fails fast with a message naming the stale manifest and the fix.
- **Empty `[workspace]` tables** in `python/` and `java/lance-jni/` manifests — declares each as its own workspace root so cargo resolves it standalone instead of climbing to an enclosing checkout. Without this, `--manifest-path` (and the hook) breaks when the checkout is nested inside another cargo project, e.g. a git worktree under the main checkout. The dependency graphs and lockfiles are unchanged.
- **`--locked`** added to the java clippy step (`java.yml`, runs on PRs) and the release build steps (`java-publish.yml`).
- **AGENTS.md** documents the three-lockfile rule.

Verified: offline check passes in-sync and detects an injected stale lock with exit 1 (no network); `[workspace]` additions leave all three lockfiles byte-identical; `pre-commit validate-config` passes and the hook ran green on this commit.

### DIFF
--- a/.github/workflows/java-publish.yml
+++ b/.github/workflows/java-publish.yml
@@ -112,7 +112,7 @@ jobs:
             export CXX=clang++
             ldd --version
 
-            cargo build --release
+            cargo build --release --locked
           "
       - uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4
         with:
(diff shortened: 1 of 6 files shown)


westonpace (approved):
+1 for the pre-commit hook alone 😄

PR #7426, opened 2026-06-23
feat(python): add list_with_delimiter, delete_file, and read_range to LanceFileSession

jmhsieh:
## Summary

Adds three primitives to `LanceFileSession` (`lance.file`), each a binding-exposure of an existing `object_store` capability (Rust → PyO3 → Python). They let callers route listing, deletion, and ranged reads through the dataset's `object_store` session instead of PyArrow's `AzureFileSystem`.

### 1. `list_with_delimiter(path=None) -> ListResult`

Non-recursive, path-segment-delimited list of a single directory level. Unlike the existing recursive `list()`, it returns only the immediate children of `path`: child "directories" as `common_prefixes` and direct child files as `objects` (a `ListResult` dataclass), all session-relative.

```python
result = session.list_with_delimiter("subdir")
result.common_prefixes  # immediate child "directories"
result.objects          # immediate child files
```

- `lance-io`: add `ObjectStore::list_with_delimiter(prefix) -> Result<ListResult>`.

### 2. `delete_file(path) -> None`

Delete an object by session-relative path. Deleting a missing path is a **no-op (idempotent)** rather than an error, so it is safe on best-effort cleanup paths.

```python
session.delete_file("subdir/checkpoint.lance")  # removes it
session.delete_file("already_gone.lance")        # no-op, no error
```

### 3. `read_range(path, offset, length) -> bytes`

Single ranged read of a byte slice from a file. Reading a missing object raises `OSError`, consistent with `download_file`.

```python
data = session.read_range("data/file-0.lance", offset=1024, length=256)  # -> bytes
```

## Why this matters

These remove the remaining PyArrow `AzureFileSystem` / `filesystem_from_uri` dependencies from geneva's checkpoint and range-blob paths:

- The recursive-only `list()` forced the flat-namespace bf-dir enumeration fallback to scan the whole checkpoint subtree; `list_with_delimiter` makes that fallback bounded again.
- `LanceFileSession` had no delete, so checkpoint purge fell back to PyArrow and silently became a no-op on flat / unreachable-DFS accounts; `delete_file` makes purge functional and blob-only.
- The range-blob reader's last PyArrow handle is replaced by `read_range`.

## Testing

- `test_session_list_with_delimiter` — non-recursive semantics (base level returns immediate files + child dirs but not their contents; subdir descends one level; trailing-slash equivalence; empty for a missing prefix).
- `test_session_delete_file` — top-level delete, nested delete, idempotent no-op on missing.
- `test_session_read_range` — mid-file range, start-of-file, up-to-end, zero-length read, and `OSError` on a missing object.
- All existing session tests pass.
- `cargo fmt`; `cargo clippy -- -D warnings` clean (default features).


python/python/lance/file.py
```diff
@@ -344,6 +365,70 @@ def list(self, path: Optional[str] = None) -> List[str]:
...
+        common_prefixes, objects = self._session.list_with_delimiter(path)
+        return ListResult(common_prefixes=common_prefixes, objects=objects)
+
+    def read_range(self, path: str, offset: int, length: int) -> bytes:
+        """
+        Read a byte range from a file (relative to this session's base path).
+
+        Issues a single ranged read. On Azure it talks to the blob endpoint only
+        and never probes the hierarchical-namespace (DFS) endpoint. Reading a
+        missing object raises ``OSError``, consistent with ``download_file``.
```
westonpace: No need to mention Azure specifically.  Can we just say "object storage" or just remove this sentence?

python/python/lance/file.py
```diff
@@ -344,6 +365,70 @@ def list(self, path: Optional[str] = None) -> List[str]:
...
+    def list_with_delimiter(self, path: Optional[str] = None) -> ListResult:
+        """
+        Non-recursively list a single directory level (relative to this
+        session's base path).
+
+        Unlike :meth:`list`, which recurses into the entire subtree, this
+        returns only the immediate children of ``path``: the child
+        "directories" as ``common_prefixes`` and the direct child files as
+        ``objects``. On Azure this is served entirely from the blob endpoint, so
+        it never probes the hierarchical-namespace (DFS) endpoint.
```
westonpace: Same comment, let's not mention Azure specifically.

python/python/lance/file.py
```diff
@@ -344,6 +365,70 @@ def list(self, path: Optional[str] = None) -> List[str]:
...
+        return self._session.read_range(path, offset, length)
+
+    def delete_file(self, path: str) -> None:
+        """
+        Delete a file (relative to this session's base path).
+
+        Deleting a path that does not exist is a no-op (idempotent) rather than
+        an error, so this is safe to call on best-effort cleanup paths. On Azure
+        it talks to the blob endpoint only and never probes the
+        hierarchical-namespace (DFS) endpoint.
```
westonpace: Once again, no need to mention Azure specifically.

python/src/file.rs
```diff
@@ -606,6 +659,68 @@ impl LanceFileSession {
         })?
     }
 
+    /// Delete a file from the object store.
+    ///
+    /// The path is interpreted relative to the session's `base_path`, matching
+    /// `contains`/`upload_file`/`download_file`. Deleting a missing object is a
+    /// no-op (idempotent) rather than an error, so this is safe to call on
+    /// best-effort cleanup paths. On Azure it talks to the blob endpoint only
+    /// and never probes the hierarchical-namespace (DFS) endpoint.
```
westonpace: Remove mention to Azure

python/src/file.rs
```diff
@@ -606,6 +659,68 @@ impl LanceFileSession {
...
+        })?
+    }
+
+    /// Read a byte range from a file in the object store.
+    ///
+    /// The path is interpreted relative to the session's `base_path`, matching
+    /// the other session methods. This issues a single ranged GET; on Azure it
+    /// talks to the blob endpoint only and never probes the
+    /// hierarchical-namespace (DFS) endpoint. Reading a missing object raises
+    /// `OSError`, consistent with `download_file`.
```
westonpace: Remove mention to Azure

python/src/file.rs
```diff
@@ -574,6 +574,59 @@ impl LanceFileSession {
...
     }
 
+    /// Non-recursive, delimited list of a single directory level.
+    ///
+    /// Returns a tuple `(common_prefixes, objects)` of paths relative to the
+    /// session's `base_path`, where `common_prefixes` are the immediate child
+    /// "directories" and `objects` are the immediate child files. Unlike
+    /// `list`, this does not recurse into the subtree. On Azure it is served
+    /// entirely from the blob endpoint, so it never probes the
+    /// hierarchical-namespace (DFS) endpoint.
```
westonpace: Remove mention to azure

python/src/file.rs
```diff
@@ -606,6 +659,68 @@ impl LanceFileSession {
...
+    /// ----------
+    /// path : str
+    ///     Path relative to `base_path` to delete.
+    pub fn delete_file(&self, path: String) -> PyResult<()> {
+        rt().block_on(None, async {
+            let full_path = self.base_path.child_path(&Path::from(path));
+            match self.object_store.inner.delete(&full_path).await {
+                Ok(()) => Ok(()),
+                // Deleting an object that does not exist is a no-op.
+                Err(object_store::Error::NotFound { .. }) => Ok(()),
```
westonpace: Do we want this to be a no-op?  Or would it be better to mirror object store's API directly?  Since we're just proxying calls it seems the less post-processing we do the better.

rust/lance-io/src/object_store.rs
```diff
@@ -849,6 +849,18 @@ impl ObjectStore {
...
 
+    /// Non-recursive, path-segment delimited list of a single directory level.
+    ///
+    /// Unlike [`Self::list`], which recurses into the entire subtree, this returns
+    /// only the immediate children of `prefix`: the child "directories" as
+    /// [`ListResult::common_prefixes`] and the direct child files as
+    /// [`ListResult::objects`]. On Azure this is served entirely from the blob
+    /// endpoint (`List Blobs` with `delimiter=/`), so it never probes the
+    /// hierarchical-namespace (DFS) endpoint.
+    pub async fn list_with_delimiter(&self, prefix: Option<&Path>) -> Result<ListResult> {
```
westonpace: Remove mention to Azure

westonpace (changes requested):
Looks great.  My only concern is all these comments specifically mention Azure but I think this feature is generic to any object storage.  Can we just remove these mentions to Azure entirely?  No need to specify exactly how the call performs on the backend.

Also, one small question on the delete.

python/python/lance/file.py
jmhsieh: done

python/python/lance/file.py
jmhsieh: done

python/python/lance/file.py
jmhsieh: done.

python/src/file.rs
jmhsieh: done.

python/src/file.rs
jmhsieh: done

python/src/file.rs
jmhsieh: done

rust/lance-io/src/object_store.rs
jmhsieh: done

python/src/file.rs
jmhsieh: I'll make it just mirror the object stores api.

westonpace (approved):
Thanks!

PR #7427, opened 2026-06-23
feat(index): write zone map seeds into data file footers during append

westonpace:
This PR introduces the concept of "index seeds".  Indexes can opt-in to planting seeds during ingestion.  The seeds are placed into the data files as global buffers.  Later, when updating the index to include these data files, we can harvest the seeds instead of scanning the data itself.

This is primarily intended to avoid a potentially expensive data scan to update the index.  As an example this PR adds index seeds for wide (binary, fixed-size-list, string) columns when creating a zone map index.  Now we calculate the min/max/nulls during ingestion, when the data is already present and flowing through the system.  Then, when we go to update the index, all we are doing is reading back those counts and adding them to the index (instead of scanning the large column all over again).

## Summary by CodeRabbit

* **New Features**
  * Zone-map indexes can now use stored seed statistics to accelerate incremental updates.
  * Zone-map configuration now supports optional `rows-per-zone` and seed usage (`use-seeds`) with backward-compatible behavior when omitted.
  * When appending to existing datasets, the system can persist zone-map seed metadata for eligible indexes.
  * Added a benchmark to compare incremental update performance with and without seeds.
* **Bug Fixes**
  * Incremental updates automatically fall back to full processing when seed data is unavailable or can’t be used.

### DIFF
--- a/protos/index_old.proto
+++ b/protos/index_old.proto
@@ -30,11 +30,14 @@ message ZoneMapIndexDetails {
   // is created for the index.
   optional uint64 rows_per_zone = 1;
   // Whether seed-based incremental updates are enabled for this index.
-  // Absent (old datasets) or false: seeds disabled.
-  // true: seeds enabled; the index will embed per-fragment seed buffers in
-  // data files and harvest them during incremental updates to skip full scans.
-  // Defaults to true for variable-length column types (strings, binary) and
-  // fixed-width types wider than 8 bytes when not explicitly set by the user.
+  // On-disk semantics: absent means seeds are disabled (old datasets written
+  // before this field was added). Present false means explicitly disabled.
+  // Present true means seeds are enabled: the index will embed per-fragment
+  // seed buffers in data files and harvest them during incremental updates
+  // to skip full column scans.
+  // Creation-time default: index creation code sets this to true for
+  // variable-length types (strings, binary) and fixed-width types wider than
+  // 8 bytes, and to false for narrow fixed-width types (e.g. Int64, Float64).
   optional bool use_seeds = 2;
 }
 message InvertedIndexDetails {
--- a/protos/index_old.proto
+++ b/protos/index_old.proto
@@ -24,7 +24,19 @@ message BTreeIndexDetails {}
 message BitmapIndexDetails {}
 message LabelListIndexDetails {}
 message NGramIndexDetails {}
-message ZoneMapIndexDetails {}
+message ZoneMapIndexDetails {
+  // Number of rows per zone. Optional for backwards compatibility: absent on
+  // datasets written before this field was added. When absent, no seed writer
+  // is created for the index.
+  optional uint64 rows_per_zone = 1;
+  // Whether seed-based incremental updates are enabled for this index.
+  // Absent (old datasets) or false: seeds disabled.
+  // true: seeds enabled; the index will embed per-fragment seed buffers in
+  // data files and harvest them during incremental updates to skip full scans.
+  // Defaults to true for variable-length column types (strings, binary) and
+  // fixed-width types wider than 8 bytes when not explicitly set by the user.
+  optional bool use_seeds = 2;
+}
 message InvertedIndexDetails {
   message CodeTokenizerConfig {
     // Split one lexical identifier into subwords, e.g. getUserName ->
--- a/rust/lance-index/src/scalar/seed.rs
+++ b/rust/lance-index/src/scalar/seed.rs
@@ -1,0 +1,59 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright The Lance Authors
+
+//! Index seed writers — compact per-fragment summaries embedded in data files.
+//!
+//! A seed writer observes column values as they are written to a data file,
+//! accumulates compact statistics in memory, and serializes them to a byte
+//! buffer that is embedded in the data file footer as a global buffer.
+//!
+//! The buffer can later be read back during index updates to reconstruct index
+//! statistics without re-scanning the column data.
+
+use arrow_array::ArrayRef;
+use bytes::Bytes;
+use lance_core::Result;
+
+/// Schema metadata key prefix for all seed buffers: `"lance.seed.<column_name>"`.
+pub const SEED_META_KEY_PREFIX: &str = "lance.seed.";
+
+/// A hook registered during data file writes that observes column values batch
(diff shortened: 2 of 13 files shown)


rust/lance-index/src/scalar/seed.rs
```diff
@@ -0,0 +1,49 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright The Lance Authors
+
+//! Index seed writers — compact per-fragment summaries embedded in data files.
+//!
+//! A seed writer observes column values as they are written to a data file,
+//! accumulates compact statistics in memory, and serializes them to a byte
```
westonpace, line 7: At the moment we assume the entire seed can comfortably fit in memory but in theory we could split it into multiple global buffers in the future if we needed to.

protos/index_old.proto
```diff
@@ -24,7 +24,19 @@ message BTreeIndexDetails {}
 message BitmapIndexDetails {}
 message LabelListIndexDetails {}
 message NGramIndexDetails {}
-message ZoneMapIndexDetails {}
+message ZoneMapIndexDetails {
+  // Number of rows per zone. Optional for backwards compatibility: absent on
+  // datasets written before this field was added. When absent, no seed writer
+  // is created for the index.
+  optional uint64 rows_per_zone = 1;
```
Xuanwo, line 31: We will need a vote for this change.

protos/index_old.proto
Xuanwo, line 1: Since we are going to touch this file. I'm thinking we need a better name for it.

protos/index_old.proto
westonpace: Hmm, I guess I'll start the discussion.

protos/index_old.proto
westonpace: 🤷 I'm open to suggestions.  It's basically the `.proto` for all index details that need to be in the `lance.table` package for backwards compatibility (newer indexes go in the `lance.index.pb` package and go in `index.proto`)

protos/index_old.proto
```diff
@@ -24,7 +24,19 @@ message BTreeIndexDetails {}
...
+  // Whether seed-based incremental updates are enabled for this index.
+  // Absent (old datasets) or false: seeds disabled.
+  // true: seeds enabled; the index will embed per-fragment seed buffers in
+  // data files and harvest them during incremental updates to skip full scans.
+  // Defaults to true for variable-length column types (strings, binary) and
+  // fixed-width types wider than 8 bytes when not explicitly set by the user.
+  optional bool use_seeds = 2;
+}
```
hamersaw, line 42: It looks like this data file seeds mechanism can be expanded to basically compute indexes at write time. Furthermore, it feels deliberately designed to support more than just ZoneMap in the future. Do we foresee every supported in type having a `use_seeds` config in the `*IndexDetails`?

westonpace: Possibly.  I think seeds make a lot of sense for small zonal indexes like zone map and bloom filter.  I think it's a bit more questionable for something like btree or bitmap as the "seed" would probably be similar in size to the data (so we aren't really saving I/O but maybe some compute).

It's also not strictly necessary to record it in the details.  I could just look at the data files and see if they have seeds and, if so, use them.  However, that felt like a bit of an expensive test.  I figured placing this flag in the details would make it easy to do a super cheap check based on the manifest before we start look in any file headers.

Or are you suggesting we promote this to a general index concept so we don't have to repeat ourselves in each index?  I think that could be fine too but I'd maybe want the idea to be a bit less experimental first.

rust/lance-index/src/scalar/seed.rs
```diff
@@ -0,0 +1,59 @@
...
+//!
+//! The buffer can later be read back during index updates to reconstruct index
+//! statistics without re-scanning the column data.
+
+use arrow_array::ArrayRef;
+use bytes::Bytes;
+use lance_core::Result;
+
+/// Schema metadata key prefix for all seed buffers: `"lance.seed.<column_name>"`.
+pub const SEED_META_KEY_PREFIX: &str = "lance.seed.";
```
wjones127, line 18: question: why column name? Isn't this supposed to be a column offset?

I remember elsewhere we were saying the field_ids nor field names should be hardcoded anywhere in the data files.

wjones127 (commented):
question: do you think index seeds will only be used for zone maps? Or could you see them be used for other things like FTS or BTREE?

westonpace:
> question: do you think index seeds will only be used for zone maps? Or could you see them be used for other things like FTS or BTREE?

@wjones127 

Maybe, but it will likely be different for each.  For example, for btree, we could store a sorted copy of the column, but this wouldn't actually reduce I/O at all at index build time.  It just moves the sort CPU cost from index time to write time and it isn't clear that's a win for anything.  Also, it would require caching the entire column in memory during the write.

Alternatively, we could capture some kind of n-tiles sketch, which we could use to inform a later indexer so that it knows the expected range of data and can setup a partitioned sort-merge (e.g. I think Spark does this when doing a sort and it requires reading the column twice so this could skip that first read).

I'm not sure about FTS, tokenizing the column is pretty expensive, and I don't think people will want to pay that cost at write time.

I do think it will be useful for something like a bloom filter index.

rust/lance-index/src/scalar/seed.rs
westonpace: Hmm, fair point, it could be an issue if the column was renamed after the write and before the index update for example.

rust/lance/src/dataset/write.rs
```diff
@@ -685,7 +697,9 @@ pub async fn do_write_fragments(
             if num_rows_in_current_file >= params.max_rows_per_file as u32
                 || writer.as_mut().unwrap().tell().await? >= params.max_bytes_per_file as u64
             {
-                let (num_rows, data_file) = writer.take().unwrap().finish().await?;
+                let mut w = writer.take().unwrap();
+                flush_seed_writers(w.as_mut(), &mut seed_writers).await?;
+                let (num_rows, data_file) = w.finish().await?;
```
wjones127, line 702: question: Below, if there's any error from `flush_seed_writers`, we drop the writer and call `cleanup_data_fragments`. But we don't here. Any particular reason?

wjones127 (approved):
I have a minor question but it otherwise looks good.

jja725:
Hi @westonpace , quick question regarding the seed. do we plan to support pruning using seed? Seeds could potentially provide temporary pruning for newly appended, not-yet-indexed fragments.

PR #528, opened 2023-02-06
[rust] make scanning in order configurable

changhiskhan:
Regular table scan go in order. KNN scan can be unordered

PR #7432, opened 2026-06-24
perf: serialize updated_fragment_offsets as RoaringBitmap bytes in proto

### LINKED ISSUE #7080: Operation::Update storing matched row offsets as flat repeated uint32 makes dense full-table column rewrites prohibitively large
## Context
PR #6650 added `Operation::Update.updated_fragment_offsets` so partial column-rewrite commits can refresh `_row_last_updated_at_version` only for matched rows. Followed by PR #6748 exposing this on the Java API.

In Rust the offsets are stored as `HashMap<u64, RoaringBitmap>,` which is space-efficient. The wire/proto encoding however is currently:

```
// transaction.proto
message Update {
  // ...
  map<uint64, UInt32List> updated_fragment_offsets = 9;
}

message UInt32List {
  // sorted distinct local physical row offsets within the fragment (0-based)
  repeated uint32 values = 1;
}
```
i.e. the in-memory bitmap is materialized as a flat `Vec<u32>` (and on the Java side a `long[]`) for transport.

## Problem
For dense full-fragment column rewrites — the common case for column-level migrations or any whole-table `update_columns_with_offsets` invocation — the wire-size is unbounded with respect to row count:

Concrete example from a dataset (1.6B rows, ~18,880 fragments, avg ~86k rows/fragment, all rows in every fragment match):

| encoding | bytes / fragment | bytes total |
| --- | --- | --- | 
| repeated uint32 packed varint (status quo) | ~242 KB | ~4.5 GB |
| serialized RoaringBitmap, portable format | ~30-300 B (single dense run container) | ~5 MB |

A ~4.5 GB transaction manifest is impractical in several places:
- Lance commit lock has a finite hold time; writing/reading multi-GB manifests pushes us against it.
- pylance / lance-jni readers materialize the full `Vec<u32>` / `long[]` in memory before constructing the `HashMap<u64, RoaringBitmap>` — peak memory is `O(total_matched_rows × 8 bytes)`, not `O(serialized_bitmap_bytes)`.
- Sparse cases (e.g. an operation that touched 10M rows out of 1.6B) work fine, so the bug is invisible until someone hits a dense-rewrite use case.

## Proposed change
Drop the flat `UInt32List` encoding and transmit only the serialized RoaringBitmap on the wire. The existing Rust in-memory representation (`HashMap<u64, RoaringBitmap>`) already maps 1:1 to this; current code paths spend the round-trip CPU/memory expanding it to/from `Vec<u32>` for no functional benefit.

```
message Update {
  // ...
  // Deprecated: writers stop populating; readers fall through to field 10.
  map<uint64, UInt32List> updated_fragment_offsets = 9 [deprecated = true];

  // Preferred. Value is the portable RoaringBitmap serialisation
  // (https://github.com/RoaringBitmap/CRoaring#format-specification),
  // i.e. `RoaringBitmap::serialize_into` / `deserialize_from`.
  map<uint64, bytes> updated_fragment_offsets_roaring = 10;
}
```

Rust write path:

```
let bytes: Vec<u8> = {
    let mut buf = Vec::with_capacity(bitmap.serialized_size_in_bytes());
    bitmap.serialize_into(&mut buf)?;
    buf
};
proto.updated_fragment_offsets_roaring.insert(frag_id, bytes);
```

Read path:
```
RoaringBitmap::deserialize_from(&bytes[..])?
```
`RoaringBitmap` is already a dependency of the crate after PR #6650 — no new third-party additions.

jerryjch:
Fixes: https://github.com/lance-format/lance/issues/7080

## Summary

Follow-up to #6650. The `updated_fragment_offsets` field (proto field 9) stores per-fragment
matched row offsets as `map<uint64, UInt32List>` -- one uint32 per matched row. For dense
rewrites this produces multi-GB manifests (e.g. 86k matched rows x 4 bytes x many fragments).

This PR adds proto field 10 (`map<uint64, bytes>`) using portable RoaringBitmap serialization,
which typically compresses the same data to tens of bytes per fragment. Writers emit field 10
only; readers prefer field 10, falling back to field 9 for manifests written before this change.

## Background

PR #6650 added `updated_fragment_offsets` to the `Update` transaction message so that
`build_manifest` can partially refresh `_row_last_updated_at_version` for matched rows only.
The encoding choice -- one uint32 per offset in a `UInt32List` -- was flagged post-merge as a
size regression for dense updates. The offsets are already stored internally as `RoaringBitmap`;
this PR aligns the proto encoding with that representation.

### protos/transaction.proto
- Deprecate field 9 (`map<uint64, UInt32List> updated_fragment_offsets`) with a comment
  pointing to field 10.
- Add field 10: `map<uint64, bytes> updated_fragment_offset_bitmaps` with documentation of
  the dual-read strategy.

### rust/lance/src/dataset/transaction.rs

Serialization (`From<&Transaction> for pb::Transaction`):
- Write field 10 only: `RoaringBitmap::serialize_into` produces portable bytes for each
  fragment's bitmap.
- Set field 9 to an empty `HashMap` (forward compat; old readers ignore unknown fields).

Deserialization (`TryFrom<pb::Transaction> for Transaction`):
- If field 10 is non-empty: deserialize each entry with `RoaringBitmap::deserialize_from`.
- Else if field 9 is non-empty: convert each `UInt32List` to `RoaringBitmap::from_iter`
  (legacy fallback).
- Same `if !new_field.is_empty() { ... } else { ... }` pattern used by the existing
  `Rewrite.groups` / `Rewrite.old_fragments` migration.

Invalid field 10 bytes fail deserialize with `Error::invalid_input`.

In-memory type unchanged: `UpdatedFragmentOffsets(HashMap<u64, RoaringBitmap>)`.

## Test plan

- `test_proto_round_trip_field_10` -- write a transaction with field 10, read back, verify
  offsets match for two fragments.
- `test_proto_legacy_field_9_read` -- construct a proto with only field 9 populated
  (simulating an old writer), deserialize, verify offsets are correctly recovered.
- `test_proto_field_10_takes_precedence_over_field_9` -- when both fields are present,
  field 10 values are used and field 9 is ignored.

Proto wire format change; team vote may be needed.

## Backward compatibility

- Proto field numbers: field 9 is kept (deprecated, not removed). Field 10 is new. No field
  number reuse.
- Old readers: ignore unknown field 10; they only read field 9, which is now empty on new
  commits. Old Lance versions deserializing commits written by this PR will not recover
  offsets from the txn; that only affects audit/`readTransaction()` on historical
  commits, not table data or OCC.
- New readers: prefer field 10; fall back to field 9 for manifests written by older Lance
  versions that predate this change.
- No JNI or Java changes. The in-memory type (`UpdatedFragmentOffsets`) is unchanged; only
  the proto wire encoding changes.

Independent of #6748 and lance-spark #528 (JNI wiring). No mutual merge dependencies.


jerryjch:
Hi, @wjones127 @pengw0048 I created a PR for https://github.com/lance-format/lance/issues/7080. Can you check if this is a good approach for the issue?

pengw0048:
Thanks, @jerryjch . I'm not very familiar with the Lance internals and it would be best if a maintainer can review this PR.

protos/transaction.proto
```diff
@@ -254,7 +254,12 @@ message Transaction {
     // Only tracks keys from INSERT operations during merge insert, not updates.
     optional KeyExistenceFilter inserted_rows = 8;
     // Per-fragment physical row offsets that matched an update_columns hash join (RewriteColumns).
+    // Deprecated: use updated_fragment_offset_bitmaps (field 10) instead.
     map<uint64, UInt32List> updated_fragment_offsets = 9;
+    // Per-fragment matched offsets as portable RoaringBitmap bytes (replaces field 9).
+    // Writers emit field 10 only. Readers prefer field 10; fall back to field 9 for
+    // manifests written before this change.
+    map<uint64, bytes> updated_fragment_offset_bitmaps = 10;
```
wjones127, line 262: question(blocking): if an older reader sees just field 10, it will ignore it. What are the consequences of that?

wjones127 (dismissed):
I have one question. Once that's addressed them I'm happy to open a PMC vote on this change.

protos/transaction.proto
jerryjch: @wjones127 Thanks for the review.
If an old reader deserializes a commit that only has field 10, it will not populate updated_fragment_offsets (field 9 is empty and field 10 is unknown to the old generated proto). That does not affect table data or OCC. Each writer—old or new—applies in-memory offsets in build_manifest at commit time and updates fragment.last_updated_at_version_meta in the manifest, which is what drives _row_last_updated_at_version at scan time. Both old and new readers read that from fragment metadata. The conflict resolver does not use updated_fragment_offsets. Field 9/10 is only the on-disk protobuf txn encoding.

The only consequence is on old Lance reading the stored txn (e.g. readTransaction()): offsets won’t round-trip for audit/history. Committed dataset versions remain correct. We chose single-write to field 10 to avoid persisting the large UInt32List on new commits.  New readers can fall back to field 9 for txns written by older Lance. Old binaries need an upgrade to decode field 10 from txn bytes. This is pretty normal for the kind of proto migration.

dshepelev15:
Hello @jerryjch, do you have any progress in the pr?

jerryjch:
@dshepelev15, The PR is pending review and approval.  cc @wjones127

wjones127:
Thanks, I've opened a PMC vote thread here: https://github.com/lance-format/lance/discussions/7705

That will close in one week, at which point we can merge this PR.

pengw0048:
Hi @wjones127 , are we good with the vote?

wjones127:
> Hi @wjones127 , are we good with the vote?

Hi @pengw0048 I'm still trying to get votes. We need 3 +1 votes. We currently have zero. I'm asking PMC members to weigh in.

jiaoew1991:
Hi @wjones127 , #7705 has enough votes 😄

Xuanwo (approved):
Let's move!

PR #7435, opened 2026-06-24
perf(fts): use block max for or tail bounds

BubbleCal:

FTS OR WAND kept lead postings that were moved back into `tail` using the posting list's full approximate upper bound. That made the tail bound too loose inside the active block-max window and reduced the pruning benefit available from existing per-block max-score metadata.

When OR WAND moves a lead posting back into `tail`, it now uses the current block max as the tail upper bound if the active block window still covers the next target. If the window is expired or not applicable, it falls back to the posting list's approximate upper bound so later high-score blocks remain reachable.

The PR also advances a headless OR tail to the next compressed block window when needed, while avoiding no-progress advancement for plain postings.

This intentionally does not implement impacts skip / ImpactsDISI, shared-floor import, MaxScoreCache, or score-only WAND fast paths.

### Benchmark context

Comparison between main and this OR tail block-max optimization:

| Case | Main QPS | This optimization QPS | Lift vs main |
|---|---:|---:|---:|
| match_len3_k10 | 352.1 | 439.5 | +24.8% |
| match_len3_k100 | 225.1 | 326.0 | +44.8% |
| phrase_len3_stop1_k10 | 639.1 | 682.2 | +6.7% |
| phrase_len3_stop1_k100 | 431.7 | 443.4 | +2.7% |

### Validation

All Cargo commands were run with an isolated target directory.

- `cargo fmt --all`
- `git diff --check`
- `cargo test -p lance-index scalar::inverted::wand::tests::test_or_ -- --nocapture` — 5 passed
- `cargo test -p lance-index scalar::inverted::wand::tests -- --nocapture` — 35 passed
- `cargo check -p lance-index --tests`
- `cargo check --workspace --tests --benches`

### DIFF
--- a/rust/lance-index/src/scalar/inverted/wand.rs
+++ b/rust/lance-index/src/scalar/inverted/wand.rs
@@ -1541,7 +1541,12 @@ impl<'a, S: Scorer> Wand<'a, S> {
         for mut tail_posting in tail.into_vec() {
             tail_posting.posting.shallow_next(target);
             let upper_bound = match tail_posting.posting.block_first_doc() {
-                Some(block_doc) if block_doc <= target => tail_posting.posting.block_max_score(),
+                Some(block_doc) if block_doc <= target => {
+                    tail_posting
+                        .posting
+                        .block_max_score_up_to_with_stats(up_to)
+                        .score
+                }
                 _ => 0.0,
             };
             if let Some(mut evicted) =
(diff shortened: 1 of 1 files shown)


rust/lance-index/src/scalar/inverted/wand.rs
```diff
@@ -1658,13 +1671,50 @@ impl<'a, S: Scorer> Wand<'a, S> {
...
+            .iter()
+            .any(|tail| tail.posting.has_next_compressed_block())
+        {
+            return false;
+        }
+
+        // A low-scoring tail can be the only iterator left in the current
+        // window. Move to the next window so a later high-scoring block is still
+        // reachable instead of ending the disjunction early.
+        self.update_max_scores(up_to + 1);
```
Xuanwo, line 1712: When this path advances a headless tail window, `update_max_scores()` derives the next `up_to` from `tail.peek()` in the no-head/no-lead case. The tail heap is ordered by upper bound, so if the top tail is already on its final block while a lower-bound tail still has later compressed blocks, `up_to` becomes `TERMINATED_DOC_ID` and the OR path stops before visiting those later blocks, dropping valid matches.

BubbleCal: fixed

Xuanwo (approved):

PR #7438, opened 2026-06-24
feat(python): expose fragment-reuse remap and delete-by-offset

wkalt:
Two thin bindings, no lance-core change, that let a caller follow and prune row addresses across a compaction without relying on stable row ids:

- Dataset.remap_row_addrs(addrs): translate row addresses to their current location across the compactions a fragment-reuse index still records (wraps open_frag_reuse_index + FragReuseIndex::remap_row_ids_array). Dropped rows become null; returns None when there is no fragment-reuse index.
- LanceFragment.delete_rows(offsets): delete rows by local (within-fragment) physical offset, exactly and without re-evaluating a predicate (wraps FileFragment::extend_deletions).

### DIFF
--- a/python/src/dataset.rs
+++ b/python/src/dataset.rs
@@ -60,7 +60,9 @@ use lance::dataset::{
     transaction::{Operation, Transaction},
 };
 use lance::index::vector::utils::get_vector_type;
-use lance::index::{DatasetIndexExt, IndexSegment, vector::VectorIndexParams};
+use lance::index::{
+    DatasetIndexExt, DatasetIndexInternalExt, IndexSegment, vector::VectorIndexParams,
+};
 use lance::{dataset::builder::DatasetBuilder, index::vector::IndexFileVersion};
 use lance_arrow::as_fixed_size_list_array;
 use lance_core::Error;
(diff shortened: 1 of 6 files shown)


python/src/dataset.rs
jackye1995, line 1: looks like these functions in dataset and fragment are missing corresponding `py.typed` stubs

python/src/fragment.rs
```diff
@@ -385,6 +385,26 @@ impl FileFragment {
...
 
+    /// Delete rows by their local (within-fragment) physical row offsets.
+    ///
+    /// Adds the given 0-based offsets to this fragment's deletion vector and
+    /// writes a new deletion file. Returns the updated fragment, or None if every
+    /// row is now deleted. Unlike `delete(predicate)`, this deletes exactly the
+    /// supplied rows without re-evaluating a filter -- useful when the caller
+    /// already knows which rows to delete (e.g. offsets collected from a prior
+    /// scan), avoiding a redundant predicate evaluation.
+    fn delete_rows(&self, offsets: Vec<u32>) -> PyResult<Option<Self>> {
```
jackye1995, line 396: I think we should add a check that the offsets are within the range of the fragment

jackye1995 (approved):
mostly looks good to me, some nits

wkalt:
@jackye1995 sorry I merged that without seeing your comments. I'll put up a patch to address those.

Issue #7274, opened 2026-06-15
Implement Blob v2 exact sharing for ingested blobs


Xuanwo:
## Problem

Blob v2 can already choose among inline, packed, dedicated, and external placement, but the write path still treats each ingested blob value as an independent physical object. This makes workloads with repeated source blobs store the same bytes multiple times even when the caller can identify that multiple rows refer to the exact same source value.

Discussion #7174 proposes an object-layer direction for Blob v2. This issue tracks the first narrow implementation step: exact sharing for ingested blobs.

## Scope

Add write-time exact sharing for blob values that are ingested into Lance-managed storage.

The sharing identity should be provided as a write-time source identity, scoped to a single write / data file. It must not become a dataset-global object id, and it must not create cross-table blob ownership or lifecycle dependencies.

Out of scope for this issue:

- delta blob encoding
- chunked blob encoding
- video or domain-specific blob packs
- cross-table blob sharing
- a public plugin system for blob object backends

## Expected behavior

When multiple rows in the same write provide the same source identity for an ingested blob, Lance should store the physical bytes once and write descriptors that resolve to the same logical bytes.

Existing callers that do not provide a source identity should keep the current behavior.

## Compatibility requirements

- Existing Blob v2 datasets remain readable.
- The default public read schema remains compatible with existing Blob v2 descriptor behavior.
- Any helper fields used to carry the write-time source identity must not leak into ordinary read schemas or persisted logical schemas.
- The implementation must fail clearly if exact sharing cannot be applied safely.

## Acceptance criteria

- Rust write path supports write-local exact sharing for ingested Blob v2 values.
- Python write path can pass or infer the source identity for ingested external blob values.
- Tests cover duplicate source identity, distinct source identity, missing source identity, and read-back correctness.
- Tests cover that ordinary read schemas do not expose the write-time helper identity.
- Documentation or examples explain the feature boundary and explicitly distinguish it from cross-table sharing.

## Related

- https://github.com/lance-format/lance/discussions/7174
- https://github.com/lance-format/lance/issues/4947
- https://github.com/lance-format/lance/issues/6372

PR #7442, opened 2026-06-24
feat: add blob v2 source id sharing

Xuanwo:
This PR implements Blob v2 object-layer plumbing and explicit write-side source sharing without changing the on-disk descriptor format.

Blob values can now carry an optional `source_id` as a write-only hint. The writer consumes this hint inside `BlobPreprocessor`, reuses the first materialized packed or dedicated descriptor within the same data file, and keeps `source_id` out of manifest and read schemas.

## API examples

Python:

```python
import lance
import pyarrow as pa
from lance import Blob

payload = b"x" * (1024 * 1024)

table = pa.table({
    "image": lance.blob_array([
        Blob.from_bytes(payload, source_id="image:001"),
        Blob.from_bytes(payload, source_id="image:001"),
    ])
})

lance.write_dataset(table, "dataset", data_storage_version="2.2")
```

External ingest can also use explicit source identity:

```python
table = pa.table({
    "image": lance.blob_array([
        Blob.from_uri("s3://bucket/image.jpg", source_id="image:001"),
        Blob.from_uri("s3://bucket/image.jpg", source_id="image:001"),
    ])
})

lance.write_dataset(
    table,
    "dataset",
    data_storage_version="2.2",
    external_blob_mode="ingest",
)
```

Rust:

```rust
use lance::blob::BlobArrayBuilder;

let payload = vec![0_u8; 1024 * 1024];

let mut builder = BlobArrayBuilder::new(2);
builder.push_bytes_with_source_id("image:001", &payload)?;
builder.push_bytes_with_source_id("image:001", &payload)?;

let field = builder.field("image", true);
let array = builder.finish()?;
```

## Benefits

This enables exact descriptor sharing for Lance-owned blob payloads. Within a single data file, repeated packed or dedicated blobs with the same `source_id` are written once, and later rows reuse the same descriptor.

This is useful when multiple rows refer to the same logical object, such as repeated images, documents, embeddings payloads, or externally ingested objects. It reduces duplicated blob bytes and avoids repeated external reads in ingest mode.

The sharing is explicit rather than content-based. Lance does not hash or compare payload bytes; `source_id` is the user's declaration that rows refer to the same logical source. The writer validates declared size when available, but same-size different-content inputs remain the user's responsibility.

The scope is intentionally data-file-local. Cross-fragment or cross-data-file physical sharing is not part of this contract. Inline blobs ignore `source_id`, and external reference mode does not materialize or deduplicate bytes.

## Compatibility

The on-disk Blob v2 descriptor and raw packed sidecar layout are unchanged. Shared rows are represented by multiple descriptors pointing at the same `(blob_id, position, size)` range.

`source_id` is a write-schema field only. It is consumed before encoding and is filtered from the manifest schema. Default reads still return the descriptor struct and never expose `source_id`.

Rust keeps the historical default `blob_field()` shape. Python upgrades its default write storage type to include `source_id`, while deserialization preserves older 4-field IPC storage types.

Validation covered Rust blob/object-layer tests, Lance schema conversion tests, Python blob tests, formatting, and diff whitespace checks.


Xuanwo:
Closing this in favor of building on the blob preparation APIs added in #7558. `PackedBlobWriter` and `BlobDescriptorArrayBuilder` already let advanced writers write a payload once and reuse its descriptor across rows, so we do not need a separate `source_id` mechanism to enable explicit sharing.

This does not provide automatic deduplication through `write_dataset`. We can revisit that convenience if a concrete workload needs it. The next step is to make explicit sharing work cleanly with independently addressed Managed blobs (#9293, still unmerged), and verify that sharing survives compaction and cleanup. #7274 remains open to track the broader work.

PR #7449, opened 2026-06-24
fix(index): give each FTS partition a distinct scheduler base priority

LuQQiu:
## Problem

FTS prewarm fans out over partitions through one shared `ScanScheduler`. Every partition opened its files at `base_priority = 0`, so all concurrent reads tied at the same priority.

The scheduler's deadlock-break admits the lowest-priority in-flight request without a byte-budget check (`task.priority <= min_in_flight`), which needs a **total order** to guarantee one request always advances. With all priorities tied at 0, freed IOP slots go to arbitrary tasks → no consumer drains → the I/O loop wedges (FTS index-cache prewarm hang on large multi-partition indexes).

## Fix

Stamp each partition's `IndexStore` with its index as `base_priority` at load time (mirrors how a scan prioritizes each fragment). Stores share the same scheduler, so backpressure is unchanged — only the priority differs, restoring a total order across partitions.

## Validation

Reproduced the hang on a 3-node cluster (600M-row multi-partition FTS index), verified the fix reaches `complete=3/3`. Adds a test for distinct, dense partition priorities.


westonpace (approved):
Looks good to me!

jackye1995 (approved):
looks good to me!

PR #7461, opened 2026-06-25
fix(index): record IVF_HNSW index file sizes after writing footer

Ali2Arslan:
## Summary

`optimize_ivf_hnsw_indices` captured the index/aux file sizes via `writer.tell()` **before** calling `finish()`. These are legacy `FileWriter` (`previous::writer::FileWriter`) instances, and `finish()` calls `write_footer()` which appends page tables, statistics, dictionaries, schema metadata, and magic bytes. As a result the captured `index_size` / `aux_size` were smaller than the actual on-disk files.

These undercounted values are persisted into `IndexMetadata.files` and surface through `describe_indices()` and `total_size_bytes()`. (Reads aren't affected today because legacy `(0,2)` vector indices are opened via `object_store.open()`, which ignores the cached sizes — but the recorded metadata is permanently wrong.)

This was introduced in #7129.

## Fix

Move both `tell()` calls to **after** `finish()`. The underlying object writer's cursor stays valid post-shutdown (`ObjectWriter::tell()` returns the running `cursor`; the local writer returns the finalized `result.size`), so it now reflects the full on-disk size including the footer.

```rust
writer.finish().await?;
let index_size = writer.tell().await? as u64;
```

## Test plan

- [x] Added `test_optimize_ivf_hnsw_records_actual_file_sizes`: builds a legacy IVF_HNSW_SQ index, appends unindexed data, runs `optimize_indices` (exercising `optimize_ivf_hnsw_indices`), then asserts each recorded `IndexFile.size_bytes` exactly equals the file size reported by the object store. This fails on the old code (recorded size < on-disk size) and passes with the fix.
- [x] `cargo fmt --all`
- [x] `cargo clippy -p lance --tests` clean
- [x] `cargo test -p lance --lib index::vector::ivf::tests::test_optimize_ivf_hnsw_records_actual_file_sizes` passes


wjones127 (approved):
Nice fix. Thank you!

PR #7462, opened 2026-06-25
fix(core): allow all-null Map columns in schema evolution

Ali2Arslan:
## Summary

`Schema::all_fields_nullable` walks every field in pre-order and requires each to be nullable. Arrow's `Map<K, V>` layout mandates a non-null `entries` struct and a non-null `key` (Lance enforces this when constructing a `Field` from an Arrow `Field`). As a result, **any** Map column fails the check, so `NewColumnTransform::AllNulls` rejects adding an all-null Map column with:

```
Invalid user input: All-null columns must be nullable.
```

…even though the inner `entries`/`key` fields are offset-addressed and completely inert when the whole Map value is NULL (the offsets pair collapses to a zero-length slice).

## Fix

Treat a **nullable** Map field as a leaf for this check — its mandatory non-null inner fields are never exercised for an all-NULL row. Behavior is otherwise unchanged:
- a non-nullable Map outer field is still rejected (the top-level `nullable` flag is still checked);
- the struct/list recursion for every other type is preserved.

## Test plan

- [x] Extended `test_all_fields_nullable` (lance-core) with nullable-Map (accept), non-nullable-Map (reject), and struct-containing-Map (accept) cases.
- [x] Added `test_add_column_all_nulls_map` (lance) — adds an all-null `Map<Utf8, Float64>` column end to end, asserts the schema/row count and that the column materializes fully null, and that a non-nullable Map is still rejected.
- [x] `cargo fmt --all`
- [x] `cargo clippy -p lance-core -p lance --tests` clean
- [x] Both tests pass.


rust/lance-core/src/datatypes/schema.rs
```diff
@@ -760,8 +760,23 @@ impl Schema {
         Ok(schema)
     }
 
+    /// Whether every field can hold a NULL — i.e. whether the schema is compatible
+    /// with columns added through `NewColumnTransform::AllNulls`.
```
westonpace: I don't love this comment because this method is in `lance-core` and we don't really have any concept of `NewColumnTransform::AllNulls` at that level (even "add columns" doesn't really make sense)

I wonder if we should move this method to `schema_evolution.rs` as its the only call site and its behavior seems tailored to schema evolution (e.g. the next caller might not expect that maps are leaves).

rust/lance-core/src/datatypes/schema.rs
```diff
@@ -760,8 +760,23 @@ impl Schema {
...
+        if !field.nullable {
+            return false;
+        }
+        // A nullable Map is a leaf here: Arrow mandates non-null `entries`/`key`, but
+        // those are offset-addressed and inert for an all-NULL row, so descending into
+        // them would wrongly reject an otherwise nullable Map column.
+        if field.logical_type.is_map() {
+            return true;
+        }
+        field.children.iter().all(Self::field_all_null_compatible)
```
westonpace: Why recurse into children at all?  I think "nullable list with non-nullable items" and "nullable struct with non-nullable child fields" would be compatible with schema evolution too.

I think this fix is perhaps too narrow.

westonpace (changes requested):
Let's make the fix a bit more of a general fix so we don't have to make another fix soon.  Or let me know if there is some case where we would want to recurse into children.

Ali2Arslan:
> have to make a

That's fair, I had AI start to port some of the standalone changes we made, which at the time were narrow on purpose (since we were only using the btree paths), I'll do a more thorough review and try to make them more general now (across the existing and new PR's).
Thank you!

rust/lance-core/src/datatypes/schema.rs
Ali2Arslan: I went a head and deleted the method all together from lance-core, since this method was public, technically it's a breaking change, not sure If I need to do anything to handle that.

rust/lance-core/src/datatypes/schema.rs
Ali2Arslan: yeah, we don't need to recurse at all, I in-lined the checked directly into `schema_evolution`, and had it check the new top-level field was nullable.

Ali2Arslan (commented):
@westonpace mind taking another look? I removed the `Schema::try_from` all together from `NewColumnTransform::AllNulls` branch since afaik it was only there to check the nullability of the new columns, but let me know if you guys would want to keep it.

westonpace (approved):
I like that approach even better, thanks!

PR #7463, opened 2026-06-25
feat(index): read a column's min/max from ZoneMap without a scan

Ali2Arslan:
## Summary

A ZoneMap scalar index already stores per-zone `min`/`max` summaries. This PR folds those summaries into a single global `(min, max)` for a column **without scanning any data**, and exposes it on the dataset:

- `ZoneMapIndex::value_range()` / `ZoneMapIndex::value_range_over(segments)` (lance-index) — fold one or more ZoneMap segments.
- `DatasetIndexExt::zonemap_value_range(column)` (lance) — the dataset-level accessor.

This is useful for cheap min/max stats and as a planning input (e.g. range pruning) without paying for a scan.

## Soundness

The folded range is designed to be **safe to prune with** — it never drops a matching row:

- **NaN bail.** `ScalarValue`'s total order ranks NaN above every finite value, so a NaN-bearing zone records `max = NaN`, hiding its true finite max. Folding only the finite maxes would yield a *subset* that could prune live rows, so any NaN zone makes the result `None`.
- **Joint fragment coverage.** The dataset accessor returns `None` unless the column's ZoneMap segments *jointly* cover every live fragment. Fragments appended after the index was built (or a segment set that doesn't span the table) leave a live fragment uncovered → `None`. Extra dead fragments in the union are fine.
- The disjoint segments of a multi-segment index are folded together.
- The result is a **superset** of live values, conservative under deletion vectors (a deleted extreme still bounds its zone): not guaranteed tight.

Reuses the existing `scalar_is_nan` helper rather than duplicating NaN logic.

## Test plan

lance-index (`value_range` / `value_range_over`):
- [x] spans multiple fragments/zones
- [x] all-null → `None`
- [x] NaN max → `None`
- [x] folds across segments; NaN in any segment → `None`; skips all-null segment

lance (`zonemap_value_range`):
- [x] basic range; column without a ZoneMap → `None`
- [x] `None` when an appended fragment isn't covered by the index
- [x] folds multiple committed segments of one logical index

- [x] `cargo fmt --all`
- [x] `cargo clippy -p lance-index -p lance --tests` clean

## Follow-up

Python bindings (`Dataset`-level accessor) can follow in a separate PR to keep this one focused on the Rust core + dataset API.

### DIFF
--- a/rust/lance/src/index.rs
+++ b/rust/lance/src/index.rs
@@ -1455,13 +1455,10 @@ impl DatasetIndexExt for Dataset {
             .await
     }
 
-    async fn zonemap_value_range(
-        &self,
-        column: &str,
-    ) -> Result<Option<(ScalarValue, ScalarValue)>> {
+    async fn column_value_range(&self, column: &str) -> Result<Option<(ScalarValue, ScalarValue)>> {
         let Some(field) = self.schema().field(column) else {
             return Err(Error::invalid_input(format!(
-                "zonemap_value_range: column '{column}' not found in dataset schema"
+                "column_value_range: column '{column}' not found in dataset schema"
             )));
         };
         let field_id = field.id;
--- a/rust/lance/src/index.rs
+++ b/rust/lance/src/index.rs
@@ -10,6 +10,7 @@ use std::sync::{Arc, OnceLock};
 use arrow_schema::DataType;
 use async_trait::async_trait;
 use datafusion::execution::SendableRecordBatchStream;
+use datafusion::scalar::ScalarValue;
 use futures::FutureExt;
 use itertools::Itertools;
 use lance_core::cache::CacheKey;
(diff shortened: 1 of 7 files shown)


rust/lance/src/index.rs
```diff
@@ -1453,6 +1455,65 @@ impl DatasetIndexExt for Dataset {
             .await
     }
 
+    async fn column_value_range(&self, column: &str) -> Result<Option<(ScalarValue, ScalarValue)>> {
```
wjones127: nitpick: I don't love adding more methods to Dataset. It would be nice if there was some sort of `Dataset::statistics()` handle this was applied to, rather than adding a method to a top-level object.

wjones127 (approved):
This seems good.

rust/lance/src/index.rs
Ali2Arslan: Thanks for taking a look Will! Are you saying you would prefer if there is some kind of extension trait for statics (smth like `DatasetStatisticsExt`) instead of having it in `DatasetIndexExt`?

Ali2Arslan: or something like a wrapper struct around Dataset and have that wrapper implement the functions logic?

Ali2Arslan: went with the latter impl, let me know if it addresses your comment

wjones127: I'm not sure honestly. I'll merge this for now, and we'll just keep that feedback in mind for a later refactor of the API.

westonpace:
Note that there is a larger picture for general "index-accelerated aggregates" here: https://github.com/lance-format/lance/issues/6765.  It would eventually expand to be able to handle things like "min/max with a filter".

I think this is fine for the short term but might become redundant if/when the more generic feature is implemented.

PR #7464, opened 2026-06-25
perf(index): coalesce concurrent scalar-index opens (single-flight)

Ali2Arslan:
## Summary

`open_scalar_index` checks the cache, and on a miss **every** caller independently re-reads `page_lookup.lance` and rebuilds the lookup. Under concurrent cold opens of the same index (common when many queries hit a freshly opened dataset) this is an N× I/O and transient-memory spike — each open loads the full index in parallel.

This PR routes the open through a new single-flight cache primitive so concurrent cold opens share one load:

- **`LanceCache::get_or_insert_unsized_with_key`** (lance-core) — the `UnsizedCacheKey` / `Arc<dyn Trait>` analogue of the existing `get_or_insert_with_key`. The cache backend dedups concurrent loads (moka `optionally_get_with`), so the loader runs at most once per key and all racing callers share the resulting `Arc`.
- **`open_scalar_index`** (lance) now wraps its L2 cache lookup → compatibility check → load → `put_in_cache` inside that loader. A warm open returns the live index with no I/O.

Behavior is otherwise unchanged; this is purely a coalescing/caching improvement on the open path.

## Test plan

- [x] `test_get_or_insert_unsized_dedup` (lance-core): 5 racing callers for one key → loader runs exactly once.
- [x] `test_open_scalar_index_coalesces_concurrent_cold_opens` (lance): 8 concurrent cold opens of a reopened dataset's BTree index → exactly one `record_index_load`, all opens return a usable index.
- [x] Full `index::scalar::tests` module passes (open-path regression).
- [x] `cargo fmt --all`
- [x] `cargo clippy -p lance-core -p lance --tests` clean


rust/lance/src/index/scalar.rs
```diff
@@ -448,29 +447,41 @@ pub async fn open_scalar_index(
         .index_cache
         .for_index(&index.uuid, frag_reuse_index.as_ref().map(|f| &f.uuid));
 
-    if let Some(index) = plugin
-        .get_from_cache(index_store.clone(), frag_reuse_index.clone(), &index_cache)
-        .await?
```
wjones127, line 467: issue(blocking): could we instead add `ScalarIndexPlugin::get_or_insert_in_cache()`? We are trying to move away from the `unsized` cache access variants, and the methods in the ScalarIndexPlugins give us a way to do so by letting the scalar index plugins provide caching storage implementations that downcast `Arc<dyn ScalarIndex>` before putting it into a cache. This is important for making sure `CacheCodec` works.

There's some background explained in: https://docs.rs/lance-index/7.0.0/lance_index/scalar/registry/trait.ScalarIndexPlugin.html#method.get_from_cache

wjones127 (changes requested):
This is something we should definitely fix, but I have a different direction I think we should go to preserve `CacheCodec` support.

rust/lance/src/index/scalar.rs
Ali2Arslan: ah, thanks Will, will take a look and update accordingly!

rust/lance-index/src/scalar/registry.rs
```diff
@@ -253,6 +280,48 @@ pub trait ScalarIndexPlugin: Send + Sync + std::fmt::Debug {
     }
 }
 
+/// A boxed, `Send` future performing the storage-level load of a scalar index
+/// (compat checks, `load_index`, metrics).
+///
+/// Passed to [`ScalarIndexPlugin::get_or_insert_in_cache`], which awaits it at
+/// most once on a cache miss and drops it un-awaited on a warm hit.
+pub type ScalarIndexLoad<'a> = BoxFuture<'a, Result<Arc<dyn ScalarIndex>>>;
```
Ali2Arslan, line 288: We can't really use a generic for this because of all the trait objects :(

rust/lance/src/index/scalar.rs
Ali2Arslan: updated Will, take a look once you get the chance, thanks!

wjones127 (approved):
Nicely done. Thanks for working on this!

PR #7465, opened 2026-06-25
perf(encoding)!: initialize only the page metadata a read will touch

Ali2Arslan:
## Summary

The structural decoder's `StructuralFieldScheduler::initialize` eagerly initializes **every** page in a column before scheduling, even when the read only touches one page. For a column with thousands of large, non-coalescing pages (e.g. a BTree leaf scan that reads one leaf), this means paying metadata IO proportional to the page count regardless of how few rows are actually requested.

This PR makes page-metadata initialization **range-scoped and per-page**:

- `DecodeBatchScheduler::try_new` and `StructuralFieldScheduler::initialize` now take an optional `requested_ranges: Option<Arc<[Range<u64>]>>` — the top-level row ranges that will later be scheduled. The structural path initializes only the pages those ranges overlap (`pages_overlapping_ranges`). `None` preserves the prior eager behavior.
- The page-scheduler `initialize` is split into `init_ranges()` (declare the byte ranges the page needs) and `init_from_buffers(buffers, io)` (finish initialization from those bytes). The field scheduler concatenates the misses' ranges into a **single** `submit_request` so adjacent ranges coalesce into shared GETs instead of one request per page.
- Page metadata is cached per page via `PageDataCacheKey { column_index, page_index, view_tag }` (replacing the per-column `FieldDataCacheKey`). `view_tag` is retained so a column decoded under two shapes (blob descriptor `Struct<pos,size>` vs raw `LargeBinary`) can't collide on cached state.
- Nested schedulers forward ranges consistently with their existing `schedule_ranges`: lists/maps/structs pass the top-level ranges through unchanged; fixed-size-list scales by `dimension`. The requested ranges are always a **superset** of what `schedule_ranges`/`schedule_take` later touch, so any page a read schedules was initialized here.

Net effect: a cold point/range read's metadata IO is invariant to the column's total page count.

## Test plan

- [x] `cargo test -p lance-encoding --lib` (393 passed) — the round-trip encoding harness now exercises **both** the eager path (full scan, `None`) and the lazy path (range reads and scattered takes pass their `requested_ranges`).
- [x] New unit test `test_initialize_coalesces_missed_page_metadata`: an N-page cache miss issues exactly **one** `submit_request` carrying every page's metadata range (not one request per page).
- [x] Updated FullZip page-scheduler tests to the `init_ranges` / `init_from_buffers` split.
- [x] `cargo test -p lance-file --lib` (76 passed).
- [x] `cargo test -p lance --lib dataset::blob::tests::test_blob_cache_key_distinguishes_views` — confirms the per-page key still distinguishes blob decoder views.
- [x] `cargo fmt --all` and `cargo clippy -p lance-encoding -p lance-file --tests --benches -- -D warnings` clean.


Ali2Arslan (commented):
Another option is to initialize the page's metadata on first touch (at schedule_ranges time - make it truley lazy) instead of this approach, but then we lose the possible coalescing benefits, so let me know if you guys have a preference on either. Thanks!

rust/lance-encoding/src/encodings/logical/primitive.rs
```diff
@@ -3477,75 +3502,173 @@ impl CachedPageData for NoCachedPageData {
...
-// that hit the same `column_index` with different shapes used to collide and
-// crash with a downcast failure when loading cached state.
+// `view_tag` distinguishes decode shapes of one physical column: a blob column
+// materializes as a `Struct<position, size>` descriptor in one variant and raw
+// `LargeBinary` in another, each caching a different `CachedPageData` type.
+// Without it, two readers at the same column/page but different shapes collide
+// and crash on a downcast.
 #[derive(Debug, Clone)]
-pub struct FieldDataCacheKey {
+pub struct PageDataCacheKey {
```
Ali2Arslan, line 4302: TODO: need to make the codec for this byte-compact now (and view_tag's encoding). Large datasets have a lot of these

rust/lance-encoding/src/encodings/logical/primitive.rs
```diff
@@ -3477,75 +3459,206 @@ impl CachedPageData for NoCachedPageData {
...
+impl CacheKey for PageDataCacheKey {
+    type ValueType = CachedPage;
 
     fn key(&self) -> std::borrow::Cow<'_, str> {
-        format!("{}:{}", self.column_index, self.view_tag).into()
+        format!(
+            "{}-{}:{}",
+            self.column_index, self.page_index, self.view_tag
+        )
+        .into()
```
wjones127, line 4318: issue(blocking): The actual content of this cache entry is likely to be smaller than the key. This is a pre-existing issue, but this PR makes it even worse. I worry this PR will bloat the metadata cache.

wjones127: I think it's worth exploring batching together a certain number of `CachedPageData` into a single MetadataCache entry. We did something similar with PostingList in https://github.com/lance-format/lance/pull/7043

Could you share more on what the typical sizes of these cache entries are? And whether it's viable to group these together into units based on byte size?

wjones127 (changes requested):

HolyMiss:
@Ali2Arslan Are there any plans to merge it? It would be very beneficial for our scenario.

Ali2Arslan:
@lance-gatekeeper reevaluate against new commit:

- #7878 replaced stored logical/string keys with allocation-free, fixed 16-byte `InternalCacheKey` digests.
- #7573 made both Moka and QuickCache eviction weights include that key footprint; capacity accounting is no longer value-only.
- `b41b972f5` skips `NoCachedPageData`, so stateless pages do not create cache entries.
- The session metadata cache now uses QuickCache by default.

Stateful pages still create individual backend records, so generic per-entry bookkeeping scales with the number of touched pages. However, that residual overhead is much smaller than the reviewed string-key/stateless-entry behavior.

Xuanwo:
Thank you @Ali2Arslan for working on this! I believe there is a tradeoff between scan and point read. I agree the review from lance-gatekeeper here. Can you try perform some bench to show the different compared to main branch?

Ali2Arslan:
@Xuanwo @wjones127 — benchmarks for the scan-vs-point-read tradeoff, plus the cache-entry/footprint evidence the gate asked for.

Headline: the selective-read win is large and unambiguous (cold point reads go from `P` metadata requests to 1, and even *warm* point reads get ~35% cheaper), the cache cost has an exact closed form of **+32 accounted bytes per stateful page**, and a full-scan wall-clock regression is **below measurement resolution** on my hardware — I would not claim one either way from this data.

## Method

`de2c80fe` (this PR's parent) vs `b41b972f`, with byte-identical harness source compiled separately against each revision. New `lance-file` bench:

* `memory://` object store, `ScanScheduler::default_for_testing()`, production `LanceEncodingsIo` wrapped in a counting `EncodingsIo` that records every `submit_request` (calls / ranges / bytes).
* Opened via `FileReader::try_open_with_file_metadata` so file metadata loads once outside the measured window; `inline_scheduling: Some(true)`; `LanceCache::with_capacity(512 MiB)` (Moka backend, so eviction weight includes the 16-byte `InternalCacheKey`).
* Fixture: one non-nullable `UInt64` column, `data_cache_bytes = 32 KiB`, 4096 rows per written batch. **Asserted** after opening: exactly one page per batch, exactly 4096 rows per page, uniform metadata-buffer size.
* Reads go through the public `FileReader::read_tasks` path with `ReadBatchParams::{Ranges, Indices, RangeFull}`; streams fully consumed, row counts asserted.
* 4 formats (2.0 / 2.1 / 2.2 / 2.3) × 8 page counts (1 → 4096) × 5 selections = 160 cases per revision, each measured in three cache states: disabled, cold-enabled, warm.
* Page-metadata I/O is attributed as **cold minus warm** for the same selection. Disabled-cache and cold-enabled I/O are asserted equal in every case, and a warm read is asserted never to miss.

Hardware: AMD Ryzen 9 3900X (12C/24T), WSL2, rustc 1.97.0, `bench` profile.

## Deterministic results

Let `P` = pages in the column, `T` = pages the requested rows touch, `b` = metadata bytes per page. Across all 120 structural cases these held with no exceptions:

| | parent `de2c80fe` | this PR `b41b972f` |
|---|---|---|
| metadata `submit_request` calls | `P` | 1 |
| metadata ranges | `P` | `T` |
| metadata bytes | `b·P` | `b·T` |
| cold cache misses | 1 | `T` |
| warm cache hits | 1 | `T` |
| retained cache entries | 1 | `T` |

v2.0 performs **zero** structural page-metadata I/O and retains **zero** cache entries on both revisions — it takes the `RootScheduler::Array` path, which never calls `initialize`. That makes it a clean control, which matters for the latency section below.

Pages touched on the 4096-page fixture: point 1, 1% range 42, sparse-64 64, 10% range 410, full scan 4096. Point-read metadata is invariant to `P` on this PR (16 B in 1 call at every page count from 1 to 4096) and linear in `P` on the parent (16 B / 1 call at `P=1`, rising to 65,536 B / 4,096 calls at `P=4096`).

## `b` depends on the data, not the format

Worth recording so these numbers aren't over-generalized: `b` = (miniblock chunks per page) × (2 bytes for v2.1, 4 for v2.2/v2.3). Chunks per page follows from `MAX_MINIBLOCK_BYTES` (8186), so it tracks compressibility:

| column | values/chunk | chunks/page | v2.1 `b` | v2.2 / v2.3 `b` |
|---|---|---|---|---|
| 64-bit incompressible | 512 | 8 | 16 B | 32 B |
| ≤63-bit, bit-packed | 1024 | 4 | 8 B | 16 B |

Both layouts are measured below, since the relative cache cost depends on which one you're in.

## Cache footprint: exactly +32 B per stateful page

Full scan, 4096 pages, accounted `LanceCache` bytes (Lance's own `DeepSizeOf` weight, not RSS):

| per-page state | parent | this PR | delta | per page: parent → PR |
|---|---|---|---|---|
| 172 B (4 chunks/page) | 770,104 | 901,120 | **+17.0%** | 188 → 220 |
| 188 B (8 chunks/page) | 835,640 | 966,656 | **+15.7%** | 204 → 236 |

The marginal cost is a constant, and it decomposes cleanly:

* parent, per page: `16` (the `Arc` slot in the whole-column `Vec`) `+ state`
* this PR, per page: `16` (`Arc` in `CachedPage`) `+ state + 16` (two `Arc` atomics) `+ 16` (`InternalCacheKey`)

Difference: **+32 accounted bytes per stateful page**, independent of format and data. The relative penalty is `32 / (state + 16)`, so it grows as pages get *smaller* — which is the honest statement of @wjones127's concern, and it also quantifies the bounded-grouping fix directly: batching `G` pages into one entry costs `32/G` bytes per page, so `G = 8` takes full-scan overhead from ~16% to ~2%. That is orthogonal to the range-scoping here and easy to add as a follow-up if we want it.

## Physical IOPs

Sampling `lance_io::iops_counter()` (post-coalescing storage requests) around each read, 4096-page fixture:

| selection | `T` | parent IOPs | PR IOPs |
|---|---|---|---|
| point | 1 | 4096 | **1** |
| 1% range | 42 | 4096 | 42 |
| sparse-64 | 64 | 4096 | 64 |
| 10% range | 410 | 4096 | 410 |
| full scan | 4096 | 4096 | 4096 |

One clarification on my own PR description, which claims the misses' ranges "coalesce into shared GETs": they do not, in the normal case. Page metadata buffers sit roughly a page apart, well outside the scheduler's 4 KiB coalescing window, so the batched `submit_request` issues exactly `T` physical requests. Full scan is 4096 → 4096 — unchanged, not batched. The benefit is entirely `T ≪ P`, not coalescing. That also means the win grows on real object storage, where per-request cost dominates, and that the single-request batching is a scheduling-overhead optimization rather than an I/O one.

## Latency

Two independent passes of both revisions, 20 samples, 2 s warmup, 5 s measurement, 4096 pages / 16,777,216 rows. Percentages are PR vs parent, shown per pass so the spread is visible:

| format | selection | empty cache (pass 1 / pass 2) | warm cache (pass 1 / pass 2) |
|---|---|---|---|
| **v2.0 — control** | point | **+43.4% / −31.1%** | **+25.9% / −44.2%** |
| **v2.0 — control** | 1% range | **+25.5% / −37.4%** | **+2.9% / −35.0%** |
| **v2.0 — control** | full scan | **+5.5% / −9.7%** | **−1.3% / −17.2%** |
| v2.1 | point | −92.0% / −94.6% | −30.9% / −44.0% |
| v2.1 | 1% range | −80.1% / −88.7% | −25.4% / −45.4% |
| v2.1 | sparse-64 | −88.7% / −93.5% | −35.8% / −73.4% |
| v2.1 | 10% range | −41.1% / −59.2% | −15.8% / −34.0% |
| v2.1 | full scan | +0.6% / −38.8% | +3.3% / −1.7% |
| v2.2 | point | −91.0% / −91.5% | −35.0% / −30.6% |
| v2.2 | 1% range | −80.5% / −86.1% | −21.2% / −31.9% |
| v2.2 | sparse-64 | −89.1% / −90.0% | −34.1% / −38.8% |
| v2.2 | 10% range | −31.9% / −45.9% | −4.0% / −15.2% |
| v2.2 | full scan | −10.0% / −5.8% | −1.5% / −5.6% |
| v2.3 | point | −90.9% / −91.0% | −41.9% / −45.6% |
| v2.3 | 1% range | −82.8% / −80.1% | −23.6% / −24.9% |
| v2.3 | sparse-64 | −89.8% / −88.4% | −37.0% / −33.1% |
| v2.3 | 10% range | −45.9% / −38.9% | −13.0% / −6.9% |
| v2.3 | full scan | +7.0% / −3.4% | +5.9% / −3.9% |

What this does and does not establish:

1. **Selective reads: solid.** −80% to −95% cold on every structural format, reproduced in both passes, while the control moves the *opposite* direction in pass 1. Far outside any plausible noise.
2. **Warm selective reads improve too — a second, separate win.** The parent's `initialize` calls `load()` on **all** `P` page schedulers even on a cache *hit*, so a warm point read on a 4096-page column still does 4096 `Arc` clones and downcasts before scheduling one page. This PR does `T`. That is the −25% to −55% warm column, and it is not an I/O effect at all.
3. **Full scan: no conclusion available.** Structural full-scan deltas span −38.8% to +7.0%, and the v2.0 control — provably executing identical code — spans +5.5% to −9.7% on the same benchmark. Same-binary drift between passes reached 50% on the full scan and >100% on the small cases. Any ±10% full-scan claim, in either direction, is below resolution here. The established parts of the full-scan cost are the deterministic ones: +32 B/page of accounted cache and 4096 → 4096 IOPs.
4. **The benefit does not vanish at 10% selectivity**, at least not where per-request cost is non-trivial: −32% to −59% cold. That is the expected shape, since the 10% read still reads 410 pages of metadata instead of 4096 — a 10× reduction in both bytes and storage requests.

## Where that leaves the PR

Range scoping is a large, reliable win for selective access on both cold and warm caches, and it is exactly neutral for v2.0. The cost is deterministic and small in absolute terms: one extra cache record and +32 accounted bytes per stateful page, with no full-scan latency penalty I can measure. If the ~16% full-scan footprint growth is judged too much, bounded grouping reduces it to ~2% at `G = 8` without touching the range-scoping logic — happy to fold that in here or do it as a follow-up, whichever you prefer.

I can push the harness as a bench target if it's worth keeping; the operation-count table above is exact, so it doubles as a cheap regression test for the initialization contract.

wjones127:
I admit I was a bit confused about this PR when I first read it, because I thought it was about Column Descriptors rather than Page-level metadata. Column descriptors are already at the end of the file and designed to be cheaply read, as described in https://lance.org/format/file/#column-descriptors.

And it's worth mentioning, there is actually a TODO comment to move the page-level metadata to the end of the file to make it quicker to read:

https://github.com/lance-format/lance/blob/31d78d1707c1d2cff01ed1ab111fe9cfc58d50d8/rust/lance-encoding/src/encodings/logical/primitive.rs#L4630-L4632

These fixes might be complementary, given that your ask isn't just for fewer IOPs but also less overall IO to read a small piece of a file with many pages.

FYI @westonpace who is the author of the above TODO comment.

rust/lance-encoding/src/decoder.rs
```diff
@@ -2234,6 +2274,45 @@ pub fn create_decode_iterator(
     }
 }
 
+fn requested_rows_to_ranges(requested_rows: &RequestedRows) -> Result<Arc<[Range<u64>]>> {
```
wjones127: suggestion: this seems to duplicate `indices_to_ranges` somewhat. Could we combine them? Also, could this be a method, `RequestedRows::to_ranges()`?

https://github.com/Ali2Arslan/lance/blob/b41b972f5d1061c7aa4acfb1b7ecca2d9938e655/rust/lance-encoding/src/decoder.rs#L1470-L1483

rust/lance-encoding/src/encodings/logical/primitive/constant.rs
```diff
@@ -101,16 +101,25 @@ enum ScalarSource {
     ValueBuffer(usize),
 }
 
+/// The (scalar, rep, def) file ranges a constant page reads, in the order
+/// `init_ranges` appends and `init_from_buffers` consumes them.  Any field is
+/// `None` when that buffer is absent. Computed once at construction so the two
+/// halves share one source of truth and their buffer order cannot drift.
```
wjones127, line 109: suggestion: this lack of drift is a nice property. Is there a way we could do something similar for other encodings? It looks like they have separate implementations that need to be kept in sync.

Ali2Arslan:
> I admit I was a bit confused about this PR when I first read it, because I thought it was about Column Descriptors rather than Page-level metadata. Column descriptors are already at the end of the file and designed to be cheaply read, as described in https://lance.org/format/file/#column-descriptors.
> 
> And it's worth mentioning, there is actually a TODO comment to move the page-level metadata to the end of the file to make it quicker to read:
> 
> https://github.com/lance-format/lance/blob/31d78d1707c1d2cff01ed1ab111fe9cfc58d50d8/rust/lance-encoding/src/encodings/logical/primitive.rs#L4630-L4632
> 
> These fixes might be complementary, given that your ask isn't just for fewer IOPs but also less overall IO to read a small piece of a file with many pages.
> 
> FYI @westonpace who is the author of the above TODO comment.

re: 
> there is actually a TODO comment to move the page-level metadata to the end of the file to make it quicker to read

I've been thinking about this recently, specifically having to pay 2 IOPS for reading the value, one for the page metadata, one for the actual value. I've not been able to come up with a solution I'm happy with, so very curious how you guys are thinking about it. Given that, for very large fragments with thousands of pages/fragment, this metadata can be quite large, but we need to know it ahead of time to index into the correct page... putting all of it into the end of the file would just move it, which I guess might get coalesced with the metadata read if it's small enough...

A bit of a tangent: Thoughts about putting the file metadata size (bytes) into the manifest to remove the (potentially) 2 IOPS for the (full) metadata of large fragment files? Similar to how the manifest already has the file size hint?

rust/lance-encoding/src/encodings/logical/map.rs
```diff
@@ -117,6 +117,18 @@ impl StructuralFieldScheduler for StructuralMapScheduler {
...
+        &'a mut self,
+        requested_ranges: &'a [Range<u64>],
+        filter: &'a FilterExpression,
+        context: &'a SchedulerContext,
+    ) -> BoxFuture<'a, Result<()>> {
+        // Like lists, the child (entries) column is scheduled with the same
+        // top-level row ranges, so forward them unchanged.
+        self.child
+            .initialize_ranges(requested_ranges, filter, context)
+    }
```
wjones127: suggestion: it would be a lot simpler if we just change the signature to take an optional range parameter. I think we'd be fine with the breaking change.

rust/lance-encoding/src/encodings/logical/primitive.rs
```diff
@@ -115,13 +115,28 @@ struct PageLoadTask {
 /// A trait for figuring out how to schedule the data within
 /// a single page.
 trait StructuralPageScheduler: std::fmt::Debug + Send {
-    /// Fetches any metadata required for the page
-    fn initialize<'a>(
+    /// Whether the requested page-local ranges require initialized metadata.
+    fn needs_initialization(&self, _ranges: &[Range<u64>]) -> Result<bool> {
```
wjones127: question(blocking): is there a case where ranges is used? As far as I can see it's either passed through to an inner impl but ignored at all the leaves.

rust/lance-encoding/src/encodings/logical/primitive.rs
```diff
@@ -115,13 +115,28 @@ struct PageLoadTask {
...
+    fn init_from_buffers<'a>(
         &'a mut self,
+        buffers: Vec<Bytes>,
         io: &Arc<dyn EncodingsIo>,
     ) -> BoxFuture<'a, Result<Arc<dyn CachedPageData>>>;
-    /// Loads metadata from a previous initialize call
-    fn load(&mut self, data: &Arc<dyn CachedPageData>);
+
+    /// Tries to attach metadata from a previous initialization.
+    fn try_load(&mut self, data: &Arc<dyn CachedPageData>) -> Result<()>;
```
wjones127: suggestion: it seems like `init_from_buffers` mutates this *and* returns some updated state, which can then be passed to `try_load`. What if we strictly made it return the state, and then had the caller use `try_load()` after? Would that be less confusing?

rust/lance-encoding/src/encodings/logical/primitive.rs
```diff
@@ -4048,10 +4043,25 @@ impl StructuralPrimitiveFieldScheduler {
...
+                PageCacheView::BlobPayload
+            }
+        } else {
+            PageCacheView::Native
+        };
         Ok(Self {
             page_schedulers,
             column_index: column_info.index,
-            view_tag: format!("{:?}", target_field.data_type()),
+            page_view,
```
wjones127: praise: I like the enum a lot better.

rust/lance-encoding/src/encodings/logical/primitive.rs
```diff
@@ -4048,10 +4043,25 @@ impl StructuralPrimitiveFieldScheduler {
...
+        });
+        let page_view = if contains_blob_pages {
+            if matches!(target_field.data_type(), DataType::Struct(_)) {
+                PageCacheView::BlobDescriptor
+            } else {
+                PageCacheView::BlobPayload
+            }
+        } else {
+            PageCacheView::Native
+        };
```
wjones127: suggestion: could we detect this by checking the type of the first scheduler in page_schedulers? IIRC we already detect what kind of blobs we are dealing with in `page_info_to_scheduler`, so might as well re-use that info.

rust/lance-encoding/src/encodings/logical/primitive.rs
```diff
@@ -4252,52 +4255,387 @@ impl CacheKey for FieldDataCacheKey {
...
+                ranges_in_page,
+            });
+        }
+        global_row_offset = page_end;
+    }
+    Ok(result)
+}
+
+impl StructuralPrimitiveFieldScheduler {
+    fn initialize_pages<'a>(
```
wjones127: nitpick: this method has gotten huge. Could you split it up? For example, the range concatenation, the sort/submit/unsort, cache insertion could be different private functions,

rust/lance-encoding/src/encodings/logical/primitive.rs
```diff
@@ -4252,52 +4255,387 @@ impl CacheKey for FieldDataCacheKey {
...
+                    Error::invalid_input_source(
+                        format!("Page index does not fit u32: {}", page.page_index).into(),
+                    )
+                })?;
+                let cache_key = PageDataCacheKey {
+                    column_index,
+                    page_index,
+                    view: page_view,
+                };
+                if let Some(cached_page) = cache.get_with_key(&cache_key).await {
```
wjones127: thought: I guess it didn't show up in your benchmarks, but does seem unfortunate to have `O(pages)` lookups here.

wjones127 (commented):
I'd like to see this move forward. I have several suggestions, but I think the most important thing is to rebase and make sure it's still passing tests.

rust/lance-encoding/src/encodings/logical/primitive.rs
Ali2Arslan: Done at the coordinator boundary: page initializers return cache state, then the field scheduler calls `try_load` and caches it uniformly. Constant, sparse, mini-block, and full-zip no longer install final state in `init_from_buffers`. Blob payload initialization necessarily loads its inner scheduler temporarily to decode descriptors, but its returned outer state is still installed by the coordinator and now caches only positions/sizes.

rust/lance-encoding/src/decoder.rs
Ali2Arslan: Implemented in 8f166d03d. `RequestedRows::to_ranges()` now returns a `Cow<[Range<u64>]>`, borrowing existing ranges and using the single checked `indices_to_ranges` conversion for indices. `schedule_take` uses that same conversion, so duplicate/descending indices and `u64::MAX` are handled consistently.

rust/lance-encoding/src/encodings/logical/primitive.rs
Ali2Arslan: Done. The coordinator is split into helpers for cache probing/miss collection, initialization-range collection, sorted read plus buffer-order restoration, concurrent page initialization/state attachment, and cache insertion.

rust/lance-encoding/src/encodings/logical/primitive.rs
Ali2Arslan: Addressed. Checked cumulative page bounds are built once; each requested range binary-searches its first overlapping page and then visits only touched pages. Cache probing iterates those `T` mappings, so sparse reads are `O(R log P + T)` rather than scanning or looking up all `P` pages (full scans naturally remain `O(P)`). The 64-page endpoint-range regression asserts exactly two cache lookups.

rust/lance-encoding/src/encodings/logical/primitive/constant.rs
Ali2Arslan: Implemented a shared `PageInitialization` / `PageInitializationBuffers` slot layout across constant, mini-block, full-zip, sparse, complex-all-null, and blob schedulers. Optional slots preserve semantic positions, and the same layout now drives both requested ranges and returned-buffer binding, with missing/extra buffers validated by tests.

rust/lance-encoding/src/encodings/logical/map.rs
Ali2Arslan: Done. `StructuralFieldScheduler` now has a single `initialize(Option<&[Range<u64>]>, ...)` method: `None` initializes the complete field and `Some` scopes initialization. Struct/list/map forward the option, while fixed-size lists scale present ranges with checked multiplication.

rust/lance-encoding/src/encodings/logical/primitive.rs
Ali2Arslan: Thanks!

rust/lance-encoding/src/encodings/logical/primitive.rs
Ali2Arslan: Confirmed: no leaf used the page-local ranges to decide what metadata to initialize. I removed ranges from the page-level initialization/needs APIs; requested ranges now only select touched pages at the field layer.

rust/lance-encoding/src/encodings/logical/primitive.rs
Ali2Arslan: Done. The constructed page scheduler now reports its cache view through an object-safe `cache_view()` method (default `Native`; blob wrappers override it), so there is no protobuf rescan. The key is per-page and also retains target-data-type isolation for native representation-dependent state.

rust/lance-encoding/src/encodings/logical/primitive.rs
Ali2Arslan: Addressed for the active per-page path. `PageDataCacheKey` now uses the versioned typed `KeyBuilder`, producing a fixed 128-bit `InternalCacheKey`; column/page IDs and the view enum are fixed-width fields, and the target type is streamed into the digest without enlarging the stored key. The former whole-field key remains only as an unused compatibility surface.

rust/lance-encoding/src/encodings/logical/map.rs
Ali2Arslan: This introduces the breaking change (as requested).

rust/lance-encoding/src/encodings/logical/primitive.rs
Ali2Arslan: @wjones127 an alternative follow-up, for your consideration here, is to add/implement `get_bulk` methods on the cache interface.

rust/lance-encoding/src/decoder.rs
Ali2Arslan: Follow-up in `c4efddae8`: this is now `RequestedRows::into_ranges()`. The coordinator consumes the request once, normalizes indices once, and reuses the same ranges for both page initialization and scheduling. Existing range requests move their `Vec` without cloning; index requests no longer run `indices_to_ranges` a second time through `schedule_take`.

rust/lance-encoding/src/encodings/logical/primitive.rs
wjones127: I agree it might be nice to provide some get_bulk. Something we should think about.

wjones127 (approved):
Thanks for addressing my feedback. This looks great!

PR #7468, opened 2026-06-25
fix(compaction): reject defer_index_remap with stable row IDs

zhangyue19921010:
Stable row IDs keep row IDs fixed across compaction, so indices never need remapping and there is nothing for defer_index_remap to defer.

**Previously this combination failed late in commit_compaction after data was already rewritten (wasting work and leaving orphaned files)**. 

Reject it at planning time instead.


zhangyue19921010:
Hi @yanghua Would u mind to take a look? Thanks. Minor fix for compaction conner case :)

yanghua (approved):
LGTM

PR #7470, opened 2026-06-25
fix: return Arrow JSON from FragmentSession.take

xloya:
## Problem
`FragmentSession.take` returns JSON columns as raw Lance JSONB (LargeBinary) instead of the logical Arrow JSON (`pa.json_()`) type.

## Root cause
The take path did not convert Lance JSON (LargeBinary/JSONB) back to Arrow JSON (Utf8) for user-facing output.

## Fix
Convert via `lance_arrow::json::convert_lance_json_to_arrow` when the batch contains JSON fields.

## Test
Rust `test_fragment_session_take_json`; Python `test_json_fragment_session_take`.

### DIFF
--- a/rust/lance/src/dataset/fragment/session.rs
+++ b/rust/lance/src/dataset/fragment/session.rs
@@ -54,7 +54,20 @@ impl FragmentSession {
         };
 
         // Then call take rows
-        self.take_rows(&row_ids).await
+        let batch = self.take_rows(&row_ids).await?;
+
+        // Convert Lance JSON columns (LargeBinary/JSONB) back to Arrow JSON (Utf8)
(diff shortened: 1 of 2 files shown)


rust/lance/src/dataset/fragment/session.rs
```diff
@@ -204,6 +217,78 @@ mod tests {
         );
     }
 
+    #[tokio::test]
+    async fn test_fragment_session_take_json() {
```
wjones127, line 221: praise: Good regression test — asserts both the extension metadata and the actual decoded values survive a partial take, not just the schema shape.

wjones127 (approved):

PR #7471, opened 2026-06-25
fix: return Arrow JSON from FileFragment.take

xloya:
## Problem
`FileFragment.take` (and dataset take) returns JSON columns as raw Lance JSONB instead of the logical Arrow JSON type.

## Root cause
The take path returned the stored Lance JSON (LargeBinary/JSONB) bytes without converting back to Arrow JSON (Utf8).

## Fix
After `take_rows`, convert Lance JSON columns back to Arrow JSON via `convert_lance_json_to_arrow` when JSON fields are present.

## Test
Python `test_take_with_json_column` (dataset) and `test_fragment_take_with_json_column`.

### DIFF
--- a/python/python/tests/test_dataset.py
+++ b/python/python/tests/test_dataset.py
@@ -964,6 +964,49 @@ def test_take_with_projection(tmp_path: Path):
     assert table3 == table2
 
 
+def test_take_with_json_column(tmp_path: Path):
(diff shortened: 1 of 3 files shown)


wjones127 (approved):
Thanks! This is a nice PR. I appreciate the regression tests.

PR #7472, opened 2026-06-25
fix: convert Arrow JSON when updating columns via merge/update path

xloya:
## Problem
Updating an Arrow JSON column through `fragment.update_columns` / the update merge path fails with a type-mismatch error.

## Root cause
The HashJoiner right-side stream did not convert Arrow JSON (Utf8) to Lance JSON (LargeBinary/JSONB) before joining.

## Fix
Wrap the right-side stream in a `JsonConvertingReader` that converts Arrow JSON fields to Lance JSON before the join.

## Test
Python `test_fragment_update_columns_with_json_column` plus dataset update tests.

### DIFF
--- a/rust/lance/src/dataset/fragment.rs
+++ b/rust/lance/src/dataset/fragment.rs
@@ -22,6 +22,7 @@ use datafusion::logical_expr::Expr;
 use datafusion::scalar::ScalarValue;
 use futures::future::{BoxFuture, try_join_all};
 use futures::{FutureExt, StreamExt, TryFutureExt, TryStreamExt, join, stream};
+use lance_arrow::json::{convert_json_columns, has_json_fields, is_arrow_json_field};
 use lance_arrow::{RecordBatchExt, SchemaExt};
 use lance_core::datatypes::{OnMissing, OnTypeMismatch, SchemaCompareOptions};
 use lance_core::utils::address::RowAddress;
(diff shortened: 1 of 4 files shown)


rust/lance/src/dataset/write/merge_insert.rs
```diff
@@ -1134,8 +1135,25 @@ impl MergeInsertJob {
                     // will be updates.
                     let mut source_batches = Vec::with_capacity(batches.len() + 1);
                     source_batches.push(batches[0].clone()); // placeholder for source data
```
wjones127: issue(blocking): You convert all the batches in this vector except the first one added here, which means the schema might be inconsistent between the batches. I think you need to convert this one as well.

rust/lance/src/dataset/write/merge_insert.rs
```diff
@@ -1134,8 +1135,25 @@ impl MergeInsertJob {
...
+                    // Check once whether JSON conversion is needed (all batches share the same schema)
+                    let needs_json_conversion = batches[0]
+                        .schema()
+                        .as_ref()
+                        .without_column(ROW_ADDR)
+                        .fields()
+                        .iter()
+                        .any(|f| is_arrow_json_field(f) || has_json_fields(f));
```
wjones127: suggestion: should we just make a helper `schema_has_json_fields(schema: &Schema) -> bool`?

wjones127 (changes requested):
Thanks for making a PR. I noticed one potential blocking issue when reviewing, but otherwise this looks like a nice fix.

rust/lance/src/dataset/write/merge_insert.rs
xloya: Good catch. I removed the `needs_json_conversion` flag entirely and now call `convert_json_columns` on every batch. Since `convert_json_columns` already no-ops (returns a clone) when there's nothing to convert, the separate schema scan was redundant — this fixes the schema-consistency bug and also handles nested Arrow JSON that the old top-level predicate missed. So the `schema_has_json_fields` helper is no longer needed.

rust/lance/src/dataset/fragment.rs
```diff
@@ -2946,6 +2958,59 @@ impl FragmentReader {
     }
 }
 
+/// A wrapper around a `RecordBatchReader` that converts Arrow JSON columns
+/// (Utf8/LargeUtf8 with `arrow.json` extension) to Lance JSON columns
+/// (LargeBinary with `lance.json` extension / JSONB format).
+///
+/// This is needed when user-provided data contains Arrow JSON fields but the
+/// dataset stores them in Lance's JSONB binary format.
+struct JsonConvertingReader {
```
wjones127, line 2967: suggestion: should we later consolidate this with `SchemaAdapter`?

https://github.com/lance-format/lance/blob/98efff9c77ff5d6eb752a1d6fa6ad722c327ab13/rust/lance/src/dataset/utils.rs#L194-L198

Can do that as a follow up.

wjones127 (approved):

PR #7473, opened 2026-06-25
fix: convert Arrow JSON in merge-insert full-fragment-rewrite fast path

xloya:
## Problem
A merge-insert touching a JSON column where ALL rows in a fragment are updated writes corrupt data and triggers a decoder panic on subsequent reads.

## Root cause
The "all rows updated" fast path in `handle_fragment` wrote Arrow JSON (Utf8, i32 offsets) into a column whose schema declares Lance JSON (LargeBinary, i64 offsets) without conversion.

## Fix
Convert Arrow JSON columns to Lance JSON before writing in that fast path.

## Test
Rust `test_merge_insert_full_fragment_rewrite_with_json_columns`; Python `test_merge_insert_full_fragment_rewrite_json_e2e`.

### DIFF
--- a/python/python/tests/test_dataset.py
+++ b/python/python/tests/test_dataset.py
@@ -2443,6 +2443,86 @@ def test_merge_insert_subcols(tmp_path: Path):
     assert dataset.to_table().sort_by("a") == expected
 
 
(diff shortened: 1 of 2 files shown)


wjones127 (approved):
This looks good. Thanks!

PR #7474, opened 2026-06-25
feat: add ICU split tokenizer variant

### LINKED ISSUE #7280: Add simple-style split control for the ICU FTS tokenizer
## Summary

The ICU tokenizer improves multilingual FTS, but it does not always split identifier-like or punctuation-delimited text the way Lance's `simple` tokenizer does.

For example, users may expect text such as `hello_world` to be searchable as `hello` and `world`, while ICU word segmentation can keep it as one token. We should add an explicit way to combine simple-style splitting with ICU segmentation for mixed multilingual text.

## Scope

* Add an opt-in tokenizer configuration that applies simple-style splitting before ICU segmentation, or otherwise gives ICU tokenizer users equivalent split behavior for punctuation-delimited tokens.
* Cover underscores, punctuation, ASCII words, CJK text, and mixed-script text in tokenizer-level tests.
* Preserve the current `base_tokenizer="icu"` behavior unless the new split control is enabled.
* Keep the scope limited to multilingual FTS tokenization. Do not fold in the broader code-aware tokenizer work from lance-format/lance#7195.

## Acceptance criteria

* With the new split control enabled, `hello_world` is indexed and queried as `hello` and `world` rather than only as one combined token.
* Mixed text such as `hello_world こんにちは世界` still benefits from ICU segmentation for non-whitespace-delimited languages.
* Existing ICU tokenizer tests continue to pass with the default behavior.
* Python-facing index creation can configure the behavior through `create_scalar_index` parameters or an equivalent public option.

Related: lance-format/lance#6661, lance-format/lance#7195

Xuanwo:
This adds an opt-in `icu/split` tokenizer variant for FTS so mixed-language text can keep ICU segmentation while also splitting identifier-like tokens on simple-style delimiters. The existing `icu` tokenizer remains unchanged, and the variant is configured through the existing `base_tokenizer` surface.

Closes #7280.

Validation note: `uv run make lint` currently fails at pyright in this local environment because optional `tensorflow` and `torch` imports cannot be resolved; ruff formatting and checks passed before that failure.

### DIFF
--- a/docs/src/format/index/scalar/fts.md
+++ b/docs/src/format/index/scalar/fts.md
@@ -81,6 +81,7 @@ The full text search index supports multiple tokenizer types for different text 
 | **raw**        | No tokenization, treats entire text as single token                       | Exact matching         |
 | **ngram**      | Breaks text into overlapping character sequences                          | Substring/fuzzy search |
 | **icu**        | ICU dictionary-based Unicode word segmentation                            | Mixed-language text    |
+| **icu/split**  | ICU segmentation with simple-style delimiter splitting                    | Mixed-language identifiers |
 | **jieba/***    | Chinese text tokenizer with word segmentation                             | Chinese text           |
 | **lindera/***  | Japanese text tokenizer with morphological analysis                       | Japanese text          |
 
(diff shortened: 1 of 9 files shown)


BubbleCal (approved):
Nice work!

PR #7475, opened 2026-06-25
docs: clarify Lance format stability

Xuanwo:
Lance releases frequently, which can make users worry that the Lance format itself is unstable. This PR updates the README to distinguish SDK/API release cadence from the dataset `data_storage_version` compatibility contract.

It explicitly states that stable storage versions remain readable by future Lance releases, that a dataset's storage version is fixed at creation, and that `next` is only for experimentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,15 +56,14 @@ For more details, see the full [Lance format specification](https://lance.org/fo
 > [!TIP]
 > Lance is in active development and we welcome contributions. Please see our [contributing guide](https://lance.org/community/contributing/) for more information.
 
-## Format stability
+## File format stability
 
-Lance releases frequently because the SDKs, integrations, and performance work are moving quickly. This does not mean the Lance format changes incompatibly in every release. The Lance format is identified by the `data_storage_version` stored in each dataset and stable storage versions are a long-term compatibility contract.
+Lance releases frequently because the SDKs, integrations, and performance work are moving quickly. This does not mean the Lance file format changes incompatibly in every release. The Lance file format is identified by the `data_storage_version` stored in each dataset, and stable storage versions are a long-term compatibility contract.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -56,6 +56,18 @@ For more details, see the full [Lance format specification](https://lance.org/fo
 > [!TIP]
 > Lance is in active development and we welcome contributions. Please see our [contributing guide](https://lance.org/community/contributing/) for more information.
 
+## Format stability
+
+Lance releases frequently because the SDKs, integrations, and performance work are moving quickly. This does not mean the Lance format changes incompatibly in every release. The Lance format is identified by the `data_storage_version` stored in each dataset and stable storage versions are a long-term compatibility contract.
```
westonpace: It's unclear here if you are speaking about the file format specifically or the entire scope of "the lance format" which includes things like table and namespace.  The table and namespace formats aren't related to `data_storage_version` and the namespace format is still changing quite rapidly.  Perhaps it would be best to mention that this is specific to the file format?

README.md
```diff
@@ -56,6 +56,18 @@ For more details, see the full [Lance format specification](https://lance.org/fo
...
+
+* Once a dataset is written with a stable `data_storage_version`, future Lance releases will continue to support reading that storage version.
+* A dataset's `data_storage_version` is fixed when the dataset is created. Later appends and updates keep using that dataset's storage version.
```
westonpace: This behavior may change in the future if we want to allow users to slowly migrate their data to a newer format.

westonpace (commented):
I think this is fine but we should probably be more specific that we are only talking about the file format here.

Xuanwo:
Updated, please review again

westonpace (approved):

PR #7476, opened 2026-06-25
fix(index): preserve schema metadata when re-serializing loaded HNSW

yanghua:
## Summary

`HNSW::to_batch()` on a disk-loaded graph dropped all schema metadata
except the HNSW key, breaking the IVF partition-cache round-trip with
`target schema is not superset of current schema`.

When an HNSW partition is loaded from disk, its `RecordBatch` schema
inherits metadata keys from the index file (`INDEX_METADATA_SCHEMA_KEY`,
`IVF_METADATA_KEY`, …). The `Loaded` branch of `to_batch()` replaced the
schema metadata with a single-key map (`HNSW_METADATA_KEY` only), so when
`RecordBatch::with_schema` ran its superset check, the new schema was not
a superset of the original and the call errored out. This affects any
`PartitionEntry<HNSW, *>` (all quantizers) that re-serializes a loaded
index through the partition cache.

The fix clones the existing schema metadata and inserts/updates the HNSW
key, instead of building a fresh single-key map — guaranteeing the new
metadata is always a superset.

## Changes

- `rust/lance-index/src/vector/hnsw/builder.rs`: in `to_batch()`'s
  `Loaded` branch, merge `HNSW_METADATA_KEY` into the retained batch's
  schema metadata rather than replacing it wholesale.
- Add `test_to_batch_loaded_preserves_extra_schema_metadata`: simulates
  the disk-load path by seeding extra schema metadata keys, then asserts
  `to_batch()` round-trips and preserves them.

### DIFF
--- a/rust/lance-index/src/vector/hnsw/builder.rs
+++ b/rust/lance-index/src/vector/hnsw/builder.rs
@@ -1239,17 +1239,23 @@ impl IvfSubIndex for HNSW {
                 // batch verbatim, re-stamped with up-to-date HNSW metadata.
                 // The IVF partition cache re-serializes loaded indices through
                 // here (`ivf/partition_serde.rs`), so this must round-trip.
+                //
+                // Merge into (not replace) the existing schema metadata: a
+                // disk-loaded batch inherits other keys from the index file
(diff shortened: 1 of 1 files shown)


jackye1995 (approved):
looks good to me!

PR #7477, opened 2026-06-25
perf: merge half-open range queries on the same BTree index

xloya:
## Problem
A filter like `fqdn = x AND log_time >= A AND log_time <= B AND channel = y` is compiled into two **half-open** range queries on the same BTree index (`log_time >= A` with an unbounded upper bound, and `log_time <= B` with an unbounded lower bound). Each half-open range matches nearly all BTree pages, so the entire index is loaded even though only the pages inside `[A, B]` are needed.

## Fix
Add a `ScalarIndexExpr::optimize()` pass that:
1. flattens the AND tree to collect leaf queries,
2. merges Range queries on the same index into a single closed range,
3. rebuilds the AND tree.

It is called from `ScalarIndexExec::new()` before execution.

## Benchmark

**Method:** build a real BTree index over 100,000 sorted unique `i32` values, split into ~100 pages (`batch_size = 1000`). Query a narrow 50-value window in the middle of the range. Measure both (a) the number of BTree pages loaded (via `LocalMetricsCollector.parts_loaded`) and (b) query wall-clock time, for:
- **Before** — the two half-open range searches the AND tree issues separately: `value >= a` (unbounded upper) and `value <= b` (unbounded lower);
- **After** — the single closed range `[a, b]` produced by `optimize()`.

**Pages loaded:**

| Execution | BTree pages loaded |
|---|---|
| Before (two half-open ranges) | 101 (≈ whole index) |
| After (merged closed range) | 1 |

≈ **101x** fewer index pages loaded for this query shape.

**Query latency:**

| Scenario | Before (two half-open) | After (merged) | Speedup |
|---|---|---|---|
| In-memory (CPU/decode bound) | 18.6 ms/q | 0.28 ms/q | ~67x |
| Per-page latency (2 ms/GET) | 76.7 ms/q | 10.8 ms/q | ~7x |

In-memory, the ~67x reflects the decode/scan cost of touching the whole index vs a single page. On remote storage the wall-clock gain depends on whether the workload is latency-bound (page fetches are parallelized, so the gain is smaller) or bandwidth-bound (scales with the ~100x reduction in bytes read).

## Test
`test_optimize_*` suite in `expression.rs` covers the merge logic (nested AND trees, exclusive bounds, no-merge cases for different indices / non-range queries, OR/NOT preservation, recheck propagation).

### DIFF
--- a/rust/lance-index/src/scalar/expression.rs
+++ b/rust/lance-index/src/scalar/expression.rs
@@ -1462,6 +1462,175 @@ impl PartialEq for ScalarIndexExpr {
     }
 }
 
+/// Returns the tighter (more restrictive) lower bound.
+/// Priority: Included/Excluded > Unbounded; Excluded > Included for same value.
+fn tighter_lower_bound(a: &Bound<ScalarValue>, b: &Bound<ScalarValue>) -> Bound<ScalarValue> {
+    match (a, b) {
+        (Bound::Unbounded, other) | (other, Bound::Unbounded) => other.clone(),
+        (Bound::Included(va), Bound::Included(vb)) => {
+            if va >= vb {
+                Bound::Included(va.clone())
+            } else {
+                Bound::Included(vb.clone())
+            }
+        }
+        (Bound::Excluded(va), Bound::Excluded(vb)) => {
+            if va >= vb {
+                Bound::Excluded(va.clone())
+            } else {
+                Bound::Excluded(vb.clone())
+            }
+        }
+        (Bound::Excluded(va), Bound::Included(vb)) => {
+            if va >= vb {
+                Bound::Excluded(va.clone())
+            } else {
+                Bound::Included(vb.clone())
+            }
+        }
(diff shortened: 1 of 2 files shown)


rust/lance-index/src/scalar/expression.rs
```diff
@@ -1462,6 +1462,175 @@ impl PartialEq for ScalarIndexExpr {
...
+        let mut result_leaves: Vec<Self> = Vec::new();
+
+        for i in 0..leaves.len() {
+            if merged_indices[i] {
+                continue;
+            }
+            let mut current = leaves[i].clone();
+
+            // Try to merge with subsequent leaves on the same index
+            for j in (i + 1)..leaves.len() {
```
wjones127, line 1569: nitpick: it would be nice to not have an `O(n^2)` algorithm here.

wjones127 (approved):
Looks good. This is a great optimization.

PR #7478, opened 2026-06-25
fix: avoid panic due to empty IO buffers on tokio runtime shutdown

ilya-zlobintsev:
Previously, shutting down an application with concurrent reads ongoing caused a panic that looks something like this:
```
 Panic!, thread_name: "tokio-rt-worker", msg: Some("range end out of bounds: 128 <= 0"), file_line: Some("/home/ilya/.local/share/cargo/registry/src/index.crates.io-1949cf8c6b5b557f/bytes-1.11.1/src/bytes.rs:392"), backtrace:   
    ...
   6: bytes::bytes::Bytes::slice
   7: lance_io::scheduler::FileScheduler::submit_request::{{closure}}
   8: <lance_file::io::LanceEncodingsIo as lance_encoding::EncodingsIo>::submit_request::{{closure}}
   9: <core::pin::Pin<P> as core::future::future::Future>::poll
  10: <lance_encoding::encodings::logical::primitive::MiniBlockScheduler as lance_encoding::encodings::logical::primitive::StructuralPageScheduler>::schedule_ranges::{{closure}}
  11: <core::pin::Pin<P> as core::future::future::Future>::poll
  12: <lance_encoding::encodings::logical::primitive::StructuralPrimitiveFieldSchedulingJob as lance_encoding::decoder::StructuralSchedulingJob>::schedule_next::{{closure}}::{{closure}}
  13: <core::pin::Pin<P> as core::future::future::Future>::poll
```

This was caused by `MutableBatch` returning an empty buffer on Drop, if it did not yet receive any data.

With this fix, it will be an I/O error instead.


rust/lance-io/src/scheduler.rs
```diff
@@ -266,6 +266,7 @@ struct MutableBatch<F: FnOnce(Response) + Send> {
     num_bytes: u64,
     priority: u128,
     num_reqs: usize,
```
wjones127, line 379: question: is `num_reqs` always equal to `data_buffers`?

wjones127: Looks like it is by construction. So this should be fine.

wjones127 (approved):
Nice work. Thanks for implementing this.

rust/lance-io/src/scheduler.rs
```diff
@@ -297,9 +299,16 @@ impl<F: FnOnce(Response) + Send> MutableBatch<F> {
 // data.
 impl<F: FnOnce(Response) + Send> Drop for MutableBatch<F> {
     fn drop(&mut self) {
-        // If we have an error, return that.  Otherwise return the data
-        let result = if self.err.is_some() {
-            Err(Error::wrapped(self.err.take().unwrap()))
+        // If we have an error, return that. Otherwise return the data, as long as the I/O requests have been processed.
+        let result = if let Some(err) = self.err.take() {
+            Err(Error::wrapped(err))
+        } else if self.num_delivered < self.data_buffers.len() {
```
Xuanwo, line 421: Counting delivered callbacks does not guarantee that each successful buffer satisfies its requested range. A `Reader::get_range` implementation can return a short `Ok` buffer; then `num_delivered == data_buffers.len()`, this branch returns `Ok`, and the scheduler later panics while slicing the requested range. Against this head, returning an empty buffer for `0..8` still produced `range end out of bounds: 8 <= 0`, so the public panic path remains.

<details>
<summary>Reproducer run against 67bb674fc1cdabadc29af230b5e834049c495ced</summary>

This was run as a small binary crate using the repository's `lance-core` and `lance-io`, plus `bytes`, `futures`, `object_store`, and `tokio`.

```rust
use std::sync::Arc;

use bytes::Bytes;
use futures::future::BoxFuture;
use lance_core::deepsize::{Context, DeepSizeOf};
use lance_io::{
    object_store::ObjectStore,
    scheduler::{ScanScheduler, SchedulerConfig},
    traits::Reader,
};
use object_store::path::Path;

#[derive(Debug)]
struct ShortReader {
    path: Path,
}

impl DeepSizeOf for ShortReader {
    fn deep_size_of_children(&self, _context: &mut Context) -> usize {
        0
    }
}

impl Reader for ShortReader {
    fn path(&self) -> &Path {
        &self.path
    }

    fn block_size(&self) -> usize {
        4096
    }

    fn io_parallelism(&self) -> usize {
        1
    }

    fn size(&self) -> BoxFuture<'_, object_store::Result<usize>> {
        Box::pin(async { Ok(1) })
    }

    fn get_range(
        &self,
        _range: std::ops::Range<usize>,
    ) -> BoxFuture<'static, object_store::Result<Bytes>> {
        Box::pin(async { Ok(Bytes::new()) })
    }

    fn get_all(&self) -> BoxFuture<'_, object_store::Result<Bytes>> {
        Box::pin(async { Ok(Bytes::new()) })
    }
}

#[tokio::main(flavor = "multi_thread")]
async fn main() {
    let scheduler = ScanScheduler::new(
        Arc::new(ObjectStore::memory()),
        SchedulerConfig {
            io_buffer_size_bytes: 1024,
            use_lite_scheduler: Some(false),
        },
    );
    let file = scheduler.open_reader(Arc::new(ShortReader {
        path: Path::parse("short").unwrap(),
    }));

    let task = tokio::spawn(async move { file.submit_request(vec![0..8], 0).await });
    match task.await {
        Ok(Err(error)) => println!("returned error: {error}"),
        Ok(Ok(bytes)) => panic!("expected an error, got {bytes:?}"),
        Err(join_error) if join_error.is_panic() => {
            eprintln!("scheduler panicked: {join_error}");
            std::process::exit(42);
        }
        Err(join_error) => panic!("scheduler task was cancelled: {join_error}"),
    }
}
```

```text
range end out of bounds: 8 <= 0
scheduler panicked: task ... panicked with message "range end out of bounds: 8 <= 0"
```

</details>

ilya-zlobintsev: Is there ever a situation where an underlying reader can return a range different from the requested one, where it is not a bug? I would expect that the answer is no, and it's safe to add a check for all reads that the returned range is of the expected size

wjones127: @Xuanwo how do you expect this PR should solve that? It's not obvious that this struct can do anything about it if the IO returns too little data, right?

Xuanwo: Ah, that makes sense. I used to perform a check at the IO layer to ensure the storage returned correct data. Now I realize there is no place for us to prevent it.

Xuanwo: Let's merge!

PR #7480, opened 2026-06-25
fix(rowids): tolerate sparse overlapping chunks in the stable row id index

wkalt:
An update-style merge_insert under stable row ids leaves the source fragment's id range overlapping the new fragment's, and a later delete punches holes in that range. RowIdIndex::new wrongly asserted that overlapping chunks densely tile their range, so a filtered with_row_id scan panicked under debug assertions. Overlap of ranges is not overlap of id sets: drop the assert and instead hard-error on the real invariant -- the same live id present in two fragments.


wjones127 (approved):
Thanks!

PR #7481, opened 2026-06-25
fix: account for SQ offset in dot distance

Xuanwo:
This fixes low recall for IVF_SQ and IVF_HNSW_SQ indexes using Dot distance when the scalar quantization bounds include negative values.

The previous SQ Dot path computed dot distance directly in quantized u8 code space. That drops the affine offset from scalar quantization, so zero-centered embeddings can be ranked incorrectly. The updated path computes Dot distance against the dequantized SQ value model while keeping the stored-code HNSW path allocation-free and able to use the u8 dot kernel.

The change also adds regression coverage for a crafted negative-bound SQ Dot case and end-to-end IVF_SQ / IVF_HNSW_SQ Dot recall on negative-valued vectors.


westonpace (commented):
I have no idea how to verify the math here.  What's your reference for this suggestion?  Do we have some kind of recall benchmark that this improves?

Xuanwo:
> I have no idea how to verify the math here. What's your reference for this suggestion? Do we have some kind of recall benchmark that this improves?

For the math part, I build a page: https://xuanwo.io/sketches/sq-affine-offset/

And I think it's a good idea to do a simple benchmark, let me check.

BubbleCal (approved):
Thanks for the fix!

PR #7483, opened 2026-06-25
feat(mem_wal): non-blocking ShardWriter::delete_no_wait

hamersaw:
## What

Adds `ShardWriter::delete_no_wait`, the delete analog of `put_no_wait`: it inserts the tombstone into the in-memory tier (visible to reads on this writer the instant it returns) and hands back the durability watcher *without* awaiting it. `delete` becomes a thin wrapper that awaits the watcher.

## Why

Lets a caller hold an external lock across only the in-memory tombstone insert and await durability after releasing it — matching how `put_no_wait` is already used so flushes still coalesce. A downstream WAL serializes deletes against partial-update merges under a per-bucket lock; with the blocking `delete` it would hold that lock across durability, and `delete_no_wait` lets it release first.

## Tests

- `test_shard_writer_delete_no_wait_visible_before_durability`

### DIFF
--- a/rust/lance/src/dataset/mem_wal/write.rs
+++ b/rust/lance/src/dataset/mem_wal/write.rs
@@ -1670,11 +1670,12 @@ impl ShardWriter {
     }
 
     /// Like [`Self::delete`], but returns the durability watcher *without*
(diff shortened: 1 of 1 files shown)


jackye1995 (approved):
Looks good to me

PR #7484, opened 2026-06-25
fix(merge-insert): apply Delete/Fail on the indexed-scan path

hamersaw:
## Problem

An indexed merge-insert delete by a composite primary key whose columns are **all** indexed silently removes nothing. `merge_insert(when_matched = Delete, use_index = true)` reports `deleted = 0`, the matched rows stay live in the table, and resurface in later reads.

## Root cause

`WhenMatched::Delete` is only implemented in the v2 plan (`DeleteOnlyMergeInsertExec`), which never uses a scalar index. When every join column is indexed, `can_use_create_plan` routes the merge to the legacy `Merger` instead (the only engine with the indexed-scan probe). That engine's matched-row handler only ever distinguished `DoNothing` from "update" — it folded **both** `Delete` and `Fail` into the update path:

- a fully-indexed `Delete` rewrote the matched rows in place → 0 deletes;
- a fully-indexed `Fail` silently updated instead of erroring.

A single indexed column hits the same path, so this was never composite-specific — composite keys just make it the common case (partial-column updates require an index on every PK column).

## Fix

Make the legacy `Merger` dispatch every `WhenMatched` variant explicitly, mirroring the v2 classifier (`merge_insert_action`):

- **`Delete`** collects matched row ids as deletions and emits no replacement batch.
- **`Fail`** aborts on any match, with the same message as the v2 path.
- A **delete-only commit branch** drains the merger and applies the deletions (resolving row ids → addresses via the row-id index for stable row ids) without writing fragments — this keeps the O(keys) indexed delete instead of falling back to a full table scan.
- The partial-schema (column-rewrite) commit branch physically cannot express row deletions, so combining `Delete` with inserts from a partial-schema source now returns a descriptive error rather than silently dropping the deletes.

## Tests

`cargo test -p lance --lib dataset::write::merge_insert` (152 passing), covering:

- composite-key indexed delete (index on first / second / both / neither — the both-indexed case is the original bug);
- single-column indexed delete;
- multi-fragment + stable-row-id variants, including an appended fragment neither index covers (unindexed-remainder union);
- `Delete` combined with `InsertAll` (not delete-only);
- `Fail` on an indexed key (match aborts, no-match inserts cleanly).

`cargo fmt` and `cargo clippy -p lance --lib --tests -- -D warnings` are clean.

### DIFF
--- a/rust/lance/src/dataset/write/merge_insert.rs
+++ b/rust/lance/src/dataset/write/merge_insert.rs
@@ -1785,7 +1785,66 @@ impl MergeInsertJob {
             .try_flatten();
         let stream = RecordBatchStreamAdapter::new(merger_schema, stream);
 
-        let (operation, affected_rows) = if !is_full_schema {
+        // A partial-schema source can patch columns or delete matched rows, but
+        // not both in one commit: the writer rejects subschema rows up front and
+        // the delete cannot be folded into the write. Reject that combination.
+        if !is_full_schema
+            && matches!(self.params.when_matched, WhenMatched::Delete)
+            && self.params.insert_not_matched
+        {
+            return Err(Error::not_supported_source("Combining when_matched(Delete) with inserts from a partial-schema source is not supported; provide the full target schema in the source".into()));
+        }
+
+        // The commit strategy follows what the merge does to matched rows. A
+        // pure delete (no inserts) writes nothing: the merger emits no batches
+        // and only records the matched row ids. This holds for any source schema
+        // width, so it is keyed on the operation rather than `is_full_schema`.
+        let is_delete_only = matches!(self.params.when_matched, WhenMatched::Delete)
+            && !self.params.insert_not_matched;
+
+        let (operation, affected_rows) = if is_delete_only {
+            // Consume the stream so the merger records the matched row ids in
+            // `deleted_rows`; it produces no batches.
+            let drained: Vec<RecordBatch> = Box::pin(stream).try_collect().await?;
+            debug_assert!(drained.is_empty(), "delete-only merge must not emit rows");
+
+            let removed_row_ids = Arc::into_inner(deleted_rows).unwrap().into_inner().unwrap();
+            let removed_row_addr_vec =
+                if let Some(row_id_index) = get_row_id_index(&self.dataset).await? {
+                    removed_row_ids
+                        .iter()
+                        .filter_map(|id| row_id_index.get(*id).map(|address| address.into()))
+                        .collect::<Vec<_>>()
+                } else {
+                    removed_row_ids
+                };
+            let removed_row_addrs = RoaringTreemap::from_iter(removed_row_addr_vec.into_iter());
+
+            let (updated_fragments, removed_fragment_ids) =
+                Self::apply_deletions(&self.dataset, &removed_row_addrs).await?;
+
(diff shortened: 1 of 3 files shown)


rust/lance/src/dataset/write/merge_insert.rs
```diff
@@ -1785,7 +1785,66 @@ impl MergeInsertJob {
             .try_flatten();
         let stream = RecordBatchStreamAdapter::new(merger_schema, stream);
 
-        let (operation, affected_rows) = if !is_full_schema {
+        // A partial-schema source can patch columns or delete matched rows, but
+        // not both in one commit: the writer rejects subschema rows up front and
+        // the delete cannot be folded into the write. Reject that combination.
+        if !is_full_schema
```
jackye1995, line 1801: This turns a previously valid partial-schema `Delete + InsertAll` merge into `NotSupported` whenever all join keys are indexed. The v2 path supports partial-schema inserts when omitted target columns are nullable, but the index routing sends this case into the legacy path and hits this guard. Can we fall back to the v2/full-scan path for this combination instead of rejecting it?

rust/lance/src/dataset/write/merge_insert.rs
```diff
@@ -2341,98 +2400,122 @@ impl Merger {
...
-                            matched = RecordBatch::new_empty(matched.schema());
+        // Each `WhenMatched` variant handles the matched rows (`in_both`)
+        // differently.
+        let match_filter_expr = self.match_filter_expr;
+        match &self.params.when_matched {
+            WhenMatched::DoNothing => {}
+            WhenMatched::Delete => {
+                // Matched rows are removed, not rewritten: record their row ids
+                // for the commit to delete and emit no replacement batch.
+                let matched_row_ids = arrow::compute::filter(batch.column(row_id_col), &in_both)?;
```
jackye1995: This branch bypasses the existing `source_dedupe_behavior` handling used for updates. If the source has duplicate keys matching the same target row, the default `Fail` behavior no longer errors, and `num_deleted_rows` is incremented for every joined duplicate even though the later `RoaringTreemap` deletes the row only once. Can we dedupe/apply the same duplicate policy before counting and recording deletes?

rust/lance/src/dataset/write/merge_insert.rs
hamersaw: I think that's the right call. I updated the earlier routing so that this case (ie. partial-schema delete + insert all) gets routed to v2 (which can support it). I left the check here to fail if this hits the v1 path as a safety net. Would be happy to remove it if it's concerning.

rust/lance/src/dataset/write/merge_insert.rs
hamersaw: Of course. I updated all arms so that the default `Fail` applies and fixed `num_deleted_rows` to correctly count the number of unique deletes. I think there is a potential follow up question here in that IIUC the `Fail` on duplicate input rows is necessary because we cannot determine which row would "win" in updates (application order). However, for the delete case we don't really care about duplicates because they're just deleting the row.

Like I said, I updated to respect the `Fail` default. But just commenting here for some historical context in case we decide to handle this differently in the future.

jackye1995 (approved):

Issue #7487, opened 2026-06-25
Default disable_transaction_file once v1.0.0+ readers are widespread


wjones127:
## Background

The transaction is already embedded inline in the manifest (`transaction_section`) on every commit, so the separate `_transactions/` file is redundant. `FLAG_DISABLE_TRANSACTION_FILE` (bit 32) lets a writer skip it and save one write per commit. Exposing a public toggle is tracked in #7486.

Inline-transaction writing and reader-side support for the flag shipped in **v1.0.0 (2025-12-12)**. A dataset written with the external file disabled cannot be read by releases before v1.0.0 (they don't understand bit 32 and will reject it).

## Ask

Once v1.0.0+ readers are sufficiently widespread (the usual ~3-6 month window after release, i.e. roughly mid-2026), flip the default so `disable_transaction_file` is on by default and we stop writing the redundant external file.

Before flipping:

- Confirm read support for inline `transaction_section` has been stable across Rust, Python, and Java since v1.0.0.
- Make sure the toggle from #7486 still allows opting back in to writing the external file for users who need older-reader compatibility.

See the [Feature Maturity](https://github.com/lance-format/lance/blob/main/docs/src/format/table/feature_maturity.md) doc for the rationale and the downgrade-floor framing.

Blocked by #7486.

Issue #7486, opened 2026-06-25
Expose a public toggle for disable_transaction_file (skip redundant external transaction file)


wjones127:
## Background

Every commit currently writes the transaction in **two** places:

- Inline in the manifest (`transaction_section`) — written unconditionally. `do_write_manifest` sets `manifest.transaction_section` whenever a transaction is passed (`rust/lance-table/src/io/manifest.rs:153-159`), and all commit paths pass `Some(transaction)` (`rust/lance/src/io/commit.rs:265,842,1047`).
- A separate `_transactions/` file — written whenever `!disable_transaction_file()` (`rust/lance/src/io/commit.rs:151,786,981`).

`FLAG_DISABLE_TRANSACTION_FILE` (bit 32, `rust/lance-table/src/feature_flags.rs:22`) and the commit-path support already exist to skip the now-redundant external file. But `disable_transaction_file` lives only on the internal `pub(crate) struct ManifestWriteConfig` (`rust/lance/src/dataset.rs:3438`) and is **never set to `true` anywhere** — no `WriteParams`/`CommitConfig` field, no Python/Java binding, no env var, no config key, not even in tests.

So the optimization is plumbed end-to-end except for the one switch that turns it on, and we pay one redundant write per commit.

## Ask

Wire up a public way to enable `disable_transaction_file`:

- Plumb it through a user-facing surface (e.g. `WriteParams` / `CommitConfig`, with Python and Java bindings kept as thin wrappers), keeping the parameter name consistent across languages.
- Add tests covering: a commit with the external file disabled produces a readable dataset from the inline `transaction_section`, and the writer feature flag (bit 32) is set.

Note: enabling it sets a writer feature flag, so older readers that don't understand bit 32 will reject the dataset. Defaulting it on is tracked separately (see follow-up issue).

PR #7488, opened 2026-06-25
docs: add feature maturity info to versioning & FTS docs

wjones127:
Adds version/maturity information to the existing per-layer versioning docs, rather than a standalone page, so each fact lives next to the spec it concerns and nothing is duplicated.

**`table/versioning.md`** — folds maturity data into the existing reader/writer feature-flags table:

- New **Introduced** column (release version + date) and **Enabled by** column for each flag.
- A *Feature Maturity* section explaining that the Introduced version is the downgrade floor (oldest release that can open a dataset using the flag), the ~3–6 month rule of thumb for defaulting, and the per-flag enable/opt-in/reserved behavior.

Only feature-flag-gated capabilities are tracked: feature flags are the format's downgrade-safety contract, so anything that would break older readers should *have* a flag. Non-flag capabilities (tags, branches, MemTable & WAL, change data feed) impose no downgrade floor and are intentionally omitted.

**`index/scalar/fts.md`** — adds a *Format Versions* table for the FTS index's legacy / v1 / v2 on-disk generations, mirroring the file-format versioning table's columns. Notes that v2 is selected only via `LANCE_FTS_FORMAT_VERSION=2` (no index-creation parameter yet) and that v2's downgrade floor is 6.0.0.

Version/date data was derived from git history (intro commit → first stable `vX.Y.Z` tag → release date).

Also corrects `table/versioning.md`: adds the missing `FLAG_DISABLE_TRANSACTION_FILE` (bit 32) row and fixes the unknown-flag threshold (64, not 32).

Related findings captured while writing this, tracked separately:
- Inline transactions: every commit writes the transaction both inline and to a separate `_transactions/` file; the flag to skip the redundant file is unwired — #7486 (expose a toggle), #7487 (default it once 1.0.0+ readers are widespread).

### DIFF
--- a/docs/src/format/table/versioning.md
+++ b/docs/src/format/table/versioning.md
@@ -15,20 +15,62 @@ they should return an "unsupported" error on any read or write operation.
 .feature-flags-table th:nth-child(2),
 .feature-flags-table td:nth-child(2) {
   white-space: nowrap;
-  min-width: 250px;
 }
 </style>
 
 <div class="feature-flags-table" markdown="1">
 
-| Flag Bit | Flag Name                       | Reader Required | Writer Required | Description                                                                                                 |
-|----------|---------------------------------|-----------------|-----------------|-------------------------------------------------------------------------------------------------------------|
-| 1        | `FLAG_DELETION_FILES`           | Yes             | Yes             | Fragments may contain deletion files, which record the tombstones of soft-deleted rows.                     |
-| 2        | `FLAG_STABLE_ROW_IDS`           | Yes             | Yes             | Row IDs are stable for both moves and updates. Fragments contain an index mapping row IDs to row addresses. |
-| 4        | `FLAG_USE_V2_FORMAT_DEPRECATED` | No              | No              | Files are written with the new v2 format. This flag is deprecated and no longer used.                       |
-| 8        | `FLAG_TABLE_CONFIG`             | No              | Yes             | Table config is present in the manifest.                                                                    |
-| 16       | `FLAG_BASE_PATHS`               | Yes             | Yes             | Dataset uses multiple base paths (for shallow clones or multi-base datasets).                               |
+| Flag Bit | Flag Name                       | Reader Required | Writer Required | Introduced          | Enabled by                                                      | Description                                                                                                 |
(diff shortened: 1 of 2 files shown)


docs/src/format/table/versioning.md
```diff
@@ -15,20 +15,62 @@ they should return an "unsupported" error on any read or write operation.
...
+  deletion file exists, the manifest config is non-empty, base paths are present). They are
+  in effect from their Introduced version onward and clear again if that data is later
+  removed (for example, after compaction materializes deletes).
+- **`FLAG_STABLE_ROW_IDS`** is opt-in and off by default. It must be set at dataset creation
+  and cannot be toggled on or off afterward. It landed in v0.12.0 but only became usable
+  end-to-end with secondary indices and compaction in v0.16.1 (2024-08-09); the write
+  parameter was named `enable_move_stable_row_ids` until v0.34.0.
+- **`FLAG_BASE_PATHS`** was introduced as `FLAG_SHALLOW_CLONE` in v0.34.0 and
+  renamed/generalized to multi-base datasets in v0.38.3 (2025-10-28); the bit value is
+  unchanged.
```
westonpace, line 69: Is this not an automatic flag?  I'm a little confused if this section is saying "automatic flags behave like X and non-automatic flags behave as described" or if it is saying "automatic flags have this behavior.  Also, there are a few other behaviors in addition to that behavior described below"

docs/src/format/table/versioning.md
```diff
@@ -15,20 +15,62 @@ they should return an "unsupported" error on any read or write operation.
...
+- **`FLAG_DISABLE_TRANSACTION_FILE`** is reserved: the flag and the commit-path support
+  exist, but no public API sets it yet. Every commit currently writes the transaction both
+  inline in the manifest (`transaction_section`) *and* to a separate `_transactions/` file;
+  wiring this flag on would let writers skip the now-redundant external file.
+
+For index (e.g. full-text search) format versions, see the
+[Full Text Search Index](../index/scalar/fts.md#format-versions) page.
```
westonpace, line 76: It seems a bit strange to call out FTS indexes here.  Every index is (in theory) independently versionable.  For example, I think we have multiple version of the IVF-PQ index at this point?

westonpace (approved):
Some thoughts but helpful information overall!

BubbleCal (approved):
Nice!

PR #7489, opened 2026-06-25
fix(mem-wal): apply cross-generation block-list to in-memory scan arms

hamersaw:
## What

The WAL fresh-tier vector and FTS search arms apply a per-generation recency filter (`NewestPkFilterExec`) to every in-memory memtable — but they only apply the cross-generation block-list (`PkBlockFilterExec`) to flushed/base sources, never to in-memory ones. This PR applies the block-list to in-memory arms too.

## Why

`LsmDataSourceCollector::in_memory_sources` emits **both** the live active memtable **and** every frozen-awaiting-flush memtable as `ActiveMemTable` sources, each with its own `index_store`. The recency filter queries only *that generation's* PK index, so it is purely within-generation.

When a PK is inserted in one generation and later deleted or updated in a **newer** in-memory generation, the older (frozen) generation's arm keeps the stale matching row: the newer generation's tombstone/update lives in a different `index_store` the frozen arm never consults. This happens whenever an insert and its later delete/update straddle a freeze boundary — common under a short flush interval, which freezes memtables continuously.

The cross-generation block-list (`NEWER(G)`) is the only mechanism that resolves this, and `compute_source_block_lists` already computes it for in-memory sources (snapshot-bounded by each generation's `max_visible_row`, so a not-yet-visible write can't over-block). The arms just discarded it.

## Change

- `vector_search.rs`: after `NewestPkFilterExec`, also wrap in `PkBlockFilterExec` when the source has a non-empty block set.
- `fts_search.rs`: unify the dedup loop so every source applies the block-list when blocked; the within-generation recency filter is already applied inside `build_source_plan` for in-memory sources.

The newest memtable per shard has no newer generation, so its block set is empty (`None`) and it is unaffected — only older frozen generations are now filtered. Each in-memory generation gets both within-generation recency and cross-generation supersession.

## Tests

- `cross_gen_stale_version_blocked_by_newer_memtable` (vector): a PK at a near vector in a frozen generation and at a far vector in the active generation. Before the fix the stale near copy leaks alongside the newest far version; after, only the newest survives.
- `cross_gen_stale_update_blocked_by_newer_memtable` (FTS): a PK matching "alpha" in a frozen generation, updated to "beta" in the active generation; an "alpha" search must not return the stale frozen row.

Both fail before this change and pass after. The existing single-generation recency tests (`keeps_only_the_newest_visible_position_per_pk`, `active_stale_update_predicate_crossing_leaks`) still pass.

### DIFF
--- a/rust/lance/src/dataset/mem_wal/scanner/fts_search.rs
+++ b/rust/lance/src/dataset/mem_wal/scanner/fts_search.rs
@@ -181,7 +181,6 @@ impl LsmFtsSearchPlanner {
         let arm_inputs: Vec<_> = sources
             .iter()
             .map(|source| {
-                let is_active = matches!(source, LsmDataSource::ActiveMemTable { .. });
                 let blocked = block_lists.get(&(source.shard_id(), source.generation()));
                 // Over-fetch a blocked source so the post-filter still yields k live
                 // rows. The active arm returns all matches (no builder limit), so its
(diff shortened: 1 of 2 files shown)


jackye1995 (approved):
looks good to me!

PR #7494, opened 2026-06-26
fix(arrow): avoid duplicate field when merging identical List<Struct> columns

LuciferYang:
## What

`RecordBatchExt::merge` produced a `StructArray` with a **duplicated field** when both batches had a `List<Struct>` column of identical type. The equal-types branch pushed the left field/column, then fell through (a missing `else`) and pushed the merged column again — so the merged batch gained a phantom extra column.

Guarding the merge path with `else` makes an identical `List<Struct>` column be taken from the left exactly once.

```rust
// Before: the if body ran, then execution fell through and pushed a second time.
if left_list.data_type() == right_list.data_type() {
    fields.push(left_field.as_ref().clone());
    columns.push(left_column.clone());
}
// This ran even when the types were identical, duplicating the field:
let merged_sub_array = merge_list_struct(&left_column, &right_column);
```

## Tests

- `test_merge_list_struct_identical_schema` — top-level identical `List<Struct>` merge. Left and right use distinct values so the equality assertion proves the *left* column is kept; asserts a single column equal to the input.
- `test_merge_nested_list_struct_identical_schema` — the same case reached through a recursive `merge` (a `List<Struct>` nested inside a struct), guarding the recursive path.

Both tests fail on `main` (the double-push yields two columns) and pass with the fix. `cargo test -p lance-arrow`, `cargo clippy -p lance-arrow --tests -- -D warnings`, and `cargo fmt -- --check` are green.

## Out of scope (separate follow-up)

While fixing this I noticed the `List<Struct>` arm of `merge` (both the identical-type and the structural-merge paths) does not call `adjust_child_validity`, unlike the sibling non-list arms — so a null parent-struct row is not propagated into a `List<Struct>` child during a recursive merge. This is a pre-existing gap that affects the whole arm, and doing it correctly on the merge path is non-trivial (which validity to propagate into a *merged* array). It is intentionally left out of this focused fix and tracked separately.

### DIFF
--- a/rust/lance-arrow/src/lib.rs
+++ b/rust/lance-arrow/src/lib.rs
@@ -1232,22 +1232,23 @@ fn merge(left_struct_array: &StructArray, right_struct_array: &StructArray) -> S
                         if left_list.data_type().is_struct()
                             && right_list.data_type().is_struct() =>
                     {
-                        // If there is nothing to merge just use the left field
+                        // Identical inner types: nothing to merge, keep the left column.
                         if left_list.data_type() == right_list.data_type() {
                             fields.push(left_field.as_ref().clone());
                             columns.push(left_column.clone());
-                        }
-                        // If we have two List<Struct> and they have different sets of fields then
-                        // we can merge them if the offsets arrays are the same.  Otherwise, we
-                        // have to consider it an error.
-                        let merged_sub_array = merge_list_struct(&left_column, &right_column);
+                        } else {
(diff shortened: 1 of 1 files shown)


LuciferYang:
cc @Xuanwo

wjones127 (approved):
Nice improvement and tests. Thank you!

LuciferYang:
Thank you @wjones127

PR #7495, opened 2026-06-26
fix(index): implement with_io_priority for FailNewFileStore test store

LuciferYang:
## What

The `lance-index` **test build** is currently broken on `main`: `error[E0046]: not all trait items implemented, missing: with_io_priority` at `rust/lance-index/src/scalar/fmindex.rs`.

`with_io_priority` was added to the `IndexStore` trait in #7449. The `FailNewFileStore` test fixture was added later in #7422 (which branched before #7449 landed), so it never implemented the new method. Each PR was green on its own; together on `main` the test build fails — a semantic merge conflict. It only surfaces in the `Build tests` step because the fixture is `#[cfg(test)]`.

## Fix

Implement `with_io_priority` on `FailNewFileStore` by delegating to the inner store and re-wrapping, mirroring how the fixture already delegates its other read methods while keeping the injected `new_index_file` failure.

## Verification

- `cargo check -p lance-index --tests` — compiles (was failing on `main`).
- `cargo test -p lance-index --lib fmindex` — 24 passed.
- `cargo fmt -p lance-index -- --check` and `cargo clippy -p lance-index --tests` — clean.

### DIFF
--- a/rust/lance-index/src/scalar/fmindex.rs
+++ b/rust/lance-index/src/scalar/fmindex.rs
@@ -2604,6 +2604,34 @@ mod tests {
         assert!(format!("{err}").contains("injected FMIndex write failure"));
     }
 
+    #[tokio::test]
+    async fn test_fail_new_file_store_with_io_priority_preserves_failure() {
+        // `with_io_priority` re-wraps in `FailNewFileStore`, so the reprioritized
+        // store must keep injecting the `new_index_file` failure.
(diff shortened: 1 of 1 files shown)


LuciferYang:
cc @hamersaw @Xuanwo main ci failed

LuciferYang:
also cc @jackye1995

Xuanwo (approved):

LuciferYang:
Thank you @Xuanwo

PR #7496, opened 2026-06-26
feat(bitpacking): add owned bitpacking codecs

BubbleCal:
## What changed

This PR narrows OSS-1346 to codec replacement work only. It adds Lance-owned `BitPacker4x` and internal `BitPacker8x` implementations in `lance-bitpacking`, and switches existing FTS posting-list users from the external `bitpacking` runtime dependency to `lance-bitpacking`.

The external `bitpacking` crate remains only as a dev-dependency for byte-for-byte compatibility tests against the legacy 4x/8x formats.

## Why

FTS currently depends on an external bitpacking crate for 128-value posting blocks. Keeping the implementation in `lance-bitpacking` lets Lance own the codec while preserving existing 128-block wire compatibility. The 8x implementation is included as an internal algorithm building block only; this PR does not expose a new block-size option.

## How it works

- Exposes `BitPacker` / `BitPacker4x` from `lance-bitpacking` with the same call shape used by existing FTS code.
- Adds owned backends for 128-value 4x blocks:
  - SSE3 on `x86_64`
  - NEON on `aarch64`
  - scalar fallback when SIMD backends are unavailable
- Adds an internal 256-value `BitPacker8x` algorithm:
  - AVX2 on `x86_64`
  - NEON on little-endian `aarch64`
  - scalar fallback when AVX2 or NEON is unavailable
  - not re-exported from `lance-bitpacking`
- Does not add or change FTS `block_size` parameters, metadata, Python, Java, or docs.

## Validation

- `cargo fmt --all`
- `cargo test -p lance-bitpacking -- --nocapture`
- `cargo check -p lance-index --tests`
- `cargo clippy -p lance-bitpacking --tests -- -D warnings`
- `cargo clippy -p lance-index --tests -- -D warnings`
- `cargo clippy -p lance --tests -- -D warnings`
- `cargo metadata --manifest-path python/Cargo.toml --locked --format-version=1 --no-deps`
- `CARGO_TARGET_DIR=/tmp/lance-target-oss1346-aarch64 cargo check -p lance-bitpacking --tests --target aarch64-apple-darwin`
- `CARGO_TARGET_DIR=/tmp/lance-target-oss1346-x86 cargo check -p lance-bitpacking --tests --target x86_64-apple-darwin`
- `git diff --check`

### DIFF
--- a/rust/compression/bitpacking/src/bitpacker_internal/bitpacker4x.rs
+++ b/rust/compression/bitpacking/src/bitpacker_internal/bitpacker4x.rs
@@ -1,0 +1,820 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright The Lance Authors
+
+use super::BitPacker;
+
+use crate::bitpacker_internal::{Available, UnsafeBitPacker};
+
+const BLOCK_LEN: usize = 32 * 4;
+
+#[cfg(target_arch = "x86_64")]
+mod sse3 {
+
+    use super::BLOCK_LEN;
+    use crate::bitpacker_internal::Available;
+
+    use std::arch::x86_64::__m128i as DataType;
+    use std::arch::x86_64::_mm_and_si128 as op_and;
+    use std::arch::x86_64::_mm_lddqu_si128 as load_unaligned;
+    use std::arch::x86_64::_mm_or_si128 as op_or;
+    use std::arch::x86_64::_mm_set1_epi32 as set1;
+    use std::arch::x86_64::_mm_slli_epi32 as left_shift_32;
+    use std::arch::x86_64::_mm_srli_epi32 as right_shift_32;
+    use std::arch::x86_64::_mm_storeu_si128 as store_unaligned;
+    use std::arch::x86_64::{
+        _mm_add_epi32, _mm_cvtsi128_si32, _mm_shuffle_epi32, _mm_slli_si128, _mm_srli_si128,
+        _mm_sub_epi32,
+    };
+
+    #[allow(non_snake_case)]
+    #[inline]
+    unsafe fn or_collapse_to_u32(accumulator: DataType) -> u32 {
+        let a__b__c__d_ = accumulator;
+        let ______a__b_ = _mm_srli_si128(a__b__c__d_, 8);
+        let a__b__ca_db = op_or(a__b__c__d_, ______a__b_);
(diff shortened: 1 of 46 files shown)


rust/compression/bitpacking/src/bitpacker_internal/mod.rs
```diff
@@ -0,0 +1,146 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright The Lance Authors
+
+// Lance-owned u32 SIMD bitpacking kernels.
+//
+// This is adapted from the Apache-2.0 `bitpacking` crate so Lance can keep the
```
Xuanwo: The copied/adapted code appears to come from the MIT-licensed `bitpacking` crate, but these new files only carry Apache/Lance notices and this text calls the upstream crate Apache-2.0. The upstream MIT notice needs to be preserved; otherwise the vendored codec may violate license notice requirements.

rust/compression/bitpacking/src/bitpacker_internal/bitpacker4x.rs
```diff
@@ -0,0 +1,820 @@
...
+            true
+        }
+    }
+}
+
+#[derive(Clone, Copy)]
+enum InstructionSet {
+    #[cfg(target_arch = "x86_64")]
+    SSE3,
+    #[cfg(target_arch = "aarch64")]
```
Xuanwo: `NEON` is compiled for every `target_arch = "aarch64"`, but the module and all match arms for that variant are only compiled on little-endian aarch64. On a big-endian aarch64 target this leaves an uncovered enum variant and prevents scalar fallback.

rust/compression/bitpacking/src/bitpacker_internal/mod.rs
BubbleCal: Fixed in af8763291: corrected the upstream license reference to MIT and added the quickwit-oss/bitpacking MIT notice to the root LICENSE.

rust/compression/bitpacking/src/bitpacker_internal/bitpacker4x.rs
BubbleCal: Fixed in af8763291: narrowed the BitPacker4x NEON enum cfg to little-endian aarch64 so it matches the NEON module and match arms; other aarch64 targets now fall back to scalar.

Xuanwo (approved):
Thank you!

PR #7497, opened 2026-06-26
refactor(arrow): simplify normalize_slicing to a no-op

LuciferYang:
## What

`StructArrayExt::normalize_slicing` folds a struct array's slice offset into its children. But in arrow-rs a `StructArray` cannot carry a slice offset in the first place — both `Array::slice` and `StructArray::from(ArrayData)` push any offset into the children (slicing them) at construction. So every `StructArray` that reaches `normalize_slicing` already satisfies the early-return condition (`offset() == 0 && all columns.len() == len()`), and the `normalized_struct_array_data` branch is dead code.

Verified by probe: `make_array(struct_data.into_builder().offset(1).len(2).build())` comes back as `offset 0, child len 2, validity preserved`. The crate's own `test_arrow_cpp_slicing` constructs exactly this shape and also hits the early-return.

## Change

- Remove the dead `normalized_struct_array_data` (~30 lines) and its `arrow_data` import.
- Collapse `normalize_slicing` to `Ok(self.clone())`, with a `debug_assert!` that trips in tests if a future arrow-rs change ever produces a non-normalized `StructArray`.
- Keep the public `normalize_slicing` method (no API break) and update its docs to describe the invariant honestly.

Behavior is unchanged: the existing `test_normalize_slicing_no_offset`, `test_arrow_rs_slicing`, and `test_arrow_cpp_slicing` pass without modification. `cargo test -p lance-arrow`, `cargo clippy -p lance-arrow --tests -- -D warnings`, and `cargo fmt -- --check` are green.

## Note for reviewers

The "dead code" claim rests on arrow-rs's `StructArray` representation (no struct-level offset). If you know of a path that produces a non-normalized `StructArray` (e.g. a specific FFI / C Data Interface import), the `debug_assert` will catch it in tests — and I'm happy to keep the original normalization logic instead if you'd prefer it as a defensive net.

### DIFF
--- a/rust/lance-arrow/src/struct.rs
+++ b/rust/lance-arrow/src/struct.rs
@@ -5,19 +5,20 @@
 
 use arrow_array::{Array, StructArray, cast::AsArray, make_array};
 use arrow_buffer::NullBuffer;
-use arrow_data::{ArrayData, ArrayDataBuilder};
 use arrow_schema::ArrowError;
 
 pub trait StructArrayExt {
-    /// Removes the offset / length of the struct array by pushing it into the children
+    /// Ensures the struct array's slicing is normalized: any offset is pushed into
+    /// the children rather than left on the struct itself.
     ///
-    /// In arrow-rs when slice is called it recursively slices the children.
-    /// In arrow-cpp when slice is called it just sets the offset/length of
-    ///   the struct array and leaves the children as-is
+    /// In arrow-rs a `StructArray` cannot carry a slice offset. Both
+    /// [`arrow_array::Array::slice`] and `StructArray::from(ArrayData)` push the
+    /// offset into the children (slicing them) at construction time, so any
+    /// `StructArray` reaching this method is already normalized and this is a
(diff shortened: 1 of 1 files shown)


LuciferYang:
cc @Xuanwo

wjones127 (changes requested):
I don't think this should be merged. We frequently take input Arrow data constructed by PyArrow, and thus can't just rely on implementation details of arrow-rs. It already says the justification for this code in the comment:

https://github.com/lance-format/lance/blob/621da36da951038ce2a6693de27faec8a78d5eeb/rust/lance-arrow/src/struct.rs#L14-L20

LuciferYang:
You're right, closing. My reachability check only covered arrow-rs's Rust construction path, where `StructArray::from(ArrayData)` folds parent offset into children at build. PyArrow structs arriving via FFI keep the arrow-cpp shape — offset on the parent, unsliced children — which is exactly what `normalize_slicing` is there to handle. The comment you cited already stated the reason; I should have taken it at face value. Thank‑you for pointing this out.@wjones127

PR #7498, opened 2026-06-26
fix(arrow): preserve inner nulls in convert_to_floating_point

LuciferYang:
## What

`FixedSizeListArrayExt::convert_to_floating_point` (used throughout the vector-indexing path — ivf, kmeans, residual, l2/dot/cosine distance) built each integer arm's float values with `from_iter_values(... .filter_map(|x| x.map(...)))`. `filter_map` **drops null inner elements**, so the values array gets shorter while the outer `FixedSizeList` null buffer is kept at full length (`self.nulls().cloned()`). For a nullable integer vector this either panics in `FixedSizeListArray::new` (when the surviving element count is not a multiple of the list size) or silently shifts every later element into the wrong list.

## Fix

Switch each arm from `from_iter_values(... filter_map(|x| x.map(...)))` to `from_iter(... map(|x| x.map(...)))`, which keeps nulls in place (same length, null positions preserved). For all-non-null data this is identical to the previous behavior, so existing callers are unaffected.

## Test

`test_convert_to_floating_point_preserves_inner_nulls` builds a `FixedSizeList<Int8>` with a null inner element and asserts the conversion keeps 2 lists / 4 values with the null still at position 1. It panics on `main` (3 surviving values, `3 % 2 != 0`) and passes with the fix.

`cargo test -p lance-arrow`, `cargo clippy -p lance-arrow --tests -- -D warnings`, and `cargo fmt -- --check` are green.

## Note

The per-type structure (Int8/16/32 → Float32, Int64/UInt8/UInt32 → Float64) is left as-is. A follow-up could collapse the whole match to `arrow_cast::cast`, but that would add an `arrow-cast` dependency to this crate, so it's intentionally kept out of this focused correctness fix.

### DIFF
--- a/rust/lance-arrow/src/lib.rs
+++ b/rust/lance-arrow/src/lib.rs
@@ -291,7 +291,7 @@ impl FixedSizeListArrayExt for FixedSizeListArray {
                         field.is_nullable(),
                     )),
                     *size,
-                    Arc::new(Float32Array::from_iter_values(
+                    Arc::new(Float32Array::from_iter(
                         self.values()
                             .as_any()
                             .downcast_ref::<Int8Array>()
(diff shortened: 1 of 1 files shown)


LuciferYang:
cc @Xuanwo

wjones127 (approved):
This is a nice, clean fix. Thank you!

LuciferYang:
Thank you @wjones127

PR #7503, opened 2026-06-26
fix: don't quote column names in list_indices

dantasse:
The bug: have a column with a hyphen, like "col-3". Build an index on it. Call list_indices, get a value that's double-quoted, like:
```
{'name': 'col-3_idx', 'type': 'BTree', 'uuid': '5db8a52c-1398-437a-9d1f-9b2570dd1b19', 'fields': ['`col-3`'], 'version': 3, 'fragment_ids': {0}, 'base_id': None}
```
The fix: use `field_path_minimal` to only quote the value if it has a dot or a backtick in it (and therefore needs it - otherwise it's impossible to tell if `my_loc.coords.lat` is `{my_loc: {coords: {lat: ...}}}` or `{my_loc: {"coords.lat": }}`

I think this change is purely cosmetic, because calls like `drop_column` and `update_field_metadata` seem to work even when passed
```
"`col-3`"
```
So I am fine with either merging this or not (and prettifying it on the UI side e.g. if I display these results)

### DIFF
--- a/python/python/tests/test_column_names.py
+++ b/python/python/tests/test_column_names.py
@@ -349,7 +349,7 @@ class TestSpecialCharacterColumnNames:
 
         indices = special_char_dataset.describe_indices()
         assert len(indices) == 1
-        assert indices[0].field_names == ["`user-id`"]
+        assert indices[0].field_names == ["user-id"]
         assert indices[0].name == "user-id_idx"
 
         # Query using the indexed column (requires backticks in filter)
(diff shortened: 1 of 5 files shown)


wjones127 (changes requested):
This looks thoughtfully done. If you get the remaining tests passing, I'd be open to merging this.

dantasse:
@wjones127 ok, thanks! yeah I go back and forth on if it's even worth it but I guess so. so I added some coverage and ran the tests locally, seemed to pass; I guess you have to approve the rest of the tests?

dantasse:
@wjones127 ok, I think I fixed the test that was breaking; ready for another test run (do you already get notified for this? I'll stop @-ing you if so)

wjones127 (approved):

PR #7505, opened 2026-06-27
docs: improve phrasing and update format spec image

prrao87:
Briefly improves the format spec intro wording and updates the lakehouse stack image.


docs/src/images/lakehouse_stack.png
wjones127, line 1: praise: wow, that is way better 😍

wjones127 (approved):
Thank you 🙌

PR #7510, opened 2026-06-29
fix: avoid stack overflow on deep logical filters

Xuanwo:
This fixes OSS-1232 by preventing deeply chained `AND` / `OR` SQL filters from overflowing the Rust stack during Lance's SQL filter planning.

The immediate root cause is in Lance's side of the SQL planning path. We first cloned deeply nested sqlparser AST expressions, and then recursively converted left-deep `AND` / `OR` trees into DataFusion expressions. A user-supplied flat logical filter could therefore abort the process before Lance returned an error.

This PR is intentionally a Lance-side workaround: it moves parsed SQL expressions out of the AST instead of recursively cloning them, then flattens same-operator logical chains and rebuilds a balanced DataFusion expression tree. That protects the known vulnerable path without waiting for upstream DataFusion/sqlparser behavior to change.

A more systematic follow-up should add shared expression complexity/depth limits across SQL, Substrait, DataFusion-expression inputs, merge-insert predicates, and any later optimization / planning passes, so Lance rejects pathological expressions consistently instead of fixing individual recursive call sites one by one.

### DIFF
--- a/rust/lance-datafusion/src/planner.rs
+++ b/rust/lance-datafusion/src/planner.rs
@@ -329,10 +329,82 @@ impl Planner {
         })
     }
 
+    fn is_logical_binary_op(op: &BinaryOperator) -> bool {
+        matches!(op, BinaryOperator::And | BinaryOperator::Or)
+    }
+
+    fn is_same_logical_binary_op(left: &BinaryOperator, right: &BinaryOperator) -> bool {
+        matches!(
+            (left, right),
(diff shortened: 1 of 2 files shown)


wjones127 (approved):
This looks like a good scoped fix. There are some things that it doesn't fix, like `a AND b OR c AND d OR ...`, but I think you are right now of this should be pushed into a follow up at the DataFusion level.

